PR #2312, opened 2026-04-03
Add SensorTiledCamera texture projection modes

nblauch:
## Description

Add projected base-color texture support to `SensorTiledCamera` for shapes without authored UVs.

This PR covers three related pieces:

- Preserve textures on UV-less USD full meshes and material subsets, with a white base color so textures render untinted.
- Add `SensorTiledCamera.TextureProjectionMode` with `CUBIC` and `TRIPLANAR` modes, configured through `SensorTiledCamera.RenderConfig.texture_projection_mode`.
- Default to cubic projection for UV-less shapes. Cubic samples the dominant local-axis projection once, matching the binary checkerboard behavior observed in OmniPBR `project_uvw`; triplanar blends samples across axes.

UV-authored meshes keep their existing interpolation path, and planes keep their existing plane coordinates. UV-less meshes and primitives share the ray-hit normal transformed into shape-local space. The texture sampler is specialized per projection mode when the render kernel is created.

The branch is rebased onto current `newton-physics/main`. It incorporates the newer USD material resolution from #3628 and extends it consistently to UV-less full meshes and material subsets.

## Checklist

## Test plan

```bash
uv run --extra dev python -m unittest newton.tests.test_sensor_tiled_camera
uv run --extra dev python -m unittest newton.tests.test_import_usd.TestImportSampleAssetsParsing
uv run docs/generate_api.py
uvx pre-commit run -a
```

Projection behavior is covered by four 64×64 `.npy` fixtures consumed by `unittest`: primitive and UV-less mesh renders in cubic and triplanar modes. CPU USD-import tests cover both full-mesh and material-subset paths.

The new tests were also applied to an isolated upstream-main checkout before the implementation: the two USD tests failed because the textures were dropped, and the projection test errored because the new API was absent.

One-off performance comparisons remain available in the [PR discussion](https://github.com/newton-physics/newton/pull/2312#issuecomment-4665255361); hardware-specific benchmark output and unconsumed PNG review artifacts are not committed.

## Bug fix

**Steps to reproduce:**

1. Import a USD full mesh or material subset with a base-color texture but no recoverable UV coordinates.
2. Render it with `SensorTiledCamera` textures enabled.
3. Without this PR, the importer drops the texture instead of making it available for projected rendering.

## New feature / API change

```python

render_config = SensorTiledCamera.RenderConfig(
    enable_textures=True,
    texture_projection_mode=SensorTiledCamera.TextureProjectionMode.TRIPLANAR,
)
sensor = SensorTiledCamera(
    model,
    default_render_config=render_config,
)
```

`TextureProjectionMode.CUBIC` is the default.


adenzler-nvidia:
Sorry for letting this go so stale - is this still needed/relevant?

adenzler-nvidia:
@nblauch are you ok with us postponing this to the 1.4 release given the lack of updates? code freeze for 1.3 is tomorrow.

adenzler-nvidia:
pushing to 1.4

nblauch:
It is still needed, yes. @daniela-hase , can you help me get this merged? Would be great to get your eyes on the fixes.

newton/_src/sensors/warp_raytrace/textures.py
```diff
@@ -47,6 +47,47 @@ def sample_texture_mesh(
...
+    the face normal, producing seamless results on arbitrarily oriented surfaces.
+    """
+    # Compute face normal from mesh triangle vertices
+    i0 = wp.mesh_get_index(mesh_id, face_id * 3 + 0)
+    i1 = wp.mesh_get_index(mesh_id, face_id * 3 + 1)
+    i2 = wp.mesh_get_index(mesh_id, face_id * 3 + 2)
+    p0 = wp.mesh_get_point(mesh_id, i0)
+    p1 = wp.mesh_get_point(mesh_id, i1)
+    p2 = wp.mesh_get_point(mesh_id, i2)
+    face_normal = wp.normalize(wp.cross(p1 - p0, p2 - p0))
```
daniela-hase: We could probably use the already computed normal from the ray-hit here, right? Just need to pass in `closest_hit.normal` to `sample_texture`.

Altough, we'd have to also transform it with the inverse transform below then .. so it is probably slower in comparison 🤔 On the other hand, this would then work for all shapes, not just meshes.

Can you do a performance comparison? And if it is similar, let's use the existing normal and make this available for all shapes, not just meshes :D

nblauch:
Added temporary PR-review artifacts under:

`newton/tests/golden_data/test_sensor_tiled_camera/projection_modes/`

The directory contains:

- `generate_projection_mode_artifacts.py`
- `projection_modes_primitives.png`
- `projection_modes_bunny.png`
- `projection_modes_contact_sheet.png`
- `projection_mode_benchmark.txt`

Regeneration command:

```bash
uv run --extra dev python newton/tests/golden_data/test_sensor_tiled_camera/projection_modes/generate_projection_mode_artifacts.py --benchmark
```

Benchmark setup:

- GPU: NVIDIA RTX PRO 6000 Blackwell Workstation Edition
- Warp: 1.14.0
- Output: albedo only, no readback inside timed loop
- Resolution: 1024x1024
- Warmup/timed updates: 20/500

Current implementation benchmark:

```text
scene                mode          mean_ms   relative
primitives           cubic           0.034      1.000x
primitives           triplanar       0.039      1.149x
uvless_bunny_mesh    cubic           0.483      1.000x
uvless_bunny_mesh    triplanar       0.529      1.094x
```

Follow-up A/B for Daniela's normal-reuse suggestion, using the same 1024x1024 bunny albedo benchmark with 20 warmup / 500 timed updates:

```text
implementation                 cubic_ms  triplanar_ms
previous split path              0.482         0.579
unified closest_hit.normal path  0.498         0.585
```

The unified path is effectively equivalent in this setup, so the branch now uses the simpler implementation: UV-less meshes and primitives both reuse the already-computed ray-hit normal, transformed into shape-local space. This removes the mesh-only per-hit face-normal reconstruction and keeps the render kernel on one texture-sampling call path.

The regenerated bunny images also look substantially better with this implementation. The previous split path forced UV-less mesh projection to use faceted face normals, while the unified path can use the smoother ray-hit normal already produced by the mesh intersection path. So this is a correctness/visual-quality improvement for smooth UV-less meshes, not just a simplification.

newton/_src/sensors/warp_raytrace/textures.py
nblauch: Updated this to use the simpler shared path you suggested.

UV-less meshes and primitives now both reuse `closest_hit.normal`, transform it into shape-local space, and route through the same cubic/triplanar projection helpers. UV-authored meshes still use the existing UV interpolation path, and planes keep their existing plane coordinate path. This also let me remove the mesh-only per-hit face-normal reconstruction and the projected-texture kernel state flag.

Fresh A/B on the same 1024x1024 bunny albedo benchmark, 20 warmup / 500 timed updates:

```text
implementation                 cubic_ms  triplanar_ms
previous split path              0.482         0.579
unified closest_hit.normal path  0.498         0.585
```

That is close enough that the simpler implementation is the better tradeoff here.

The regenerated bunny contact sheet also looks substantially better with this path, since UV-less mesh projection can now follow the smoother ray-hit normal instead of forcing a faceted face normal. So this improves visual correctness for smooth no-UV meshes too.

nblauch:
Pushed 5a51af10 to address the CodeRabbit feedback:

- UV-less textured meshes are now included when computing has_shape_projected_textures, including checkerboard-assigned textures.
- The Warp mesh sampler no longer uses an unsafe boolean or guard that could index mesh_data when mesh_data_index is negative.
- Added CUDA regression tests for cubic and triplanar checkerboard projection on a UV-less mesh-only scene.

Validation:

```bash
uv run --extra dev ruff check newton/_src/sensors/warp_raytrace/render_context.py newton/_src/sensors/warp_raytrace/textures.py newton/_src/sensors/warp_raytrace/utils.py newton/tests/test_sensor_tiled_camera.py
uv run --extra dev ruff format --check newton/_src/sensors/warp_raytrace/render_context.py newton/_src/sensors/warp_raytrace/textures.py newton/_src/sensors/warp_raytrace/utils.py newton/tests/test_sensor_tiled_camera.py
uvx pre-commit run -a
```

newton/_src/sensors/warp_raytrace/render.py
```diff
@@ -211,19 +211,53 @@ def render_megakernel(
...
-                        texture_index,
-                        shape_source_ptr[closest_hit.shape_index],
-                        mesh_data,
-                        shape_mesh_data_ids[closest_hit.shape_index],
-                        hit_point,
-                        closest_hit.bary_u,
-                        closest_hit.bary_v,
-                        closest_hit.face_idx,
-                    )
+                    if wp.static(state.render_shape_projected_textures):
```
daniela-hase: Can we move this whole block over to `textures.py`? So it doesn't blow up the code in the `render_megakernel` that much.

newton/_src/sensors/warp_raytrace/render_context.py
```diff
@@ -68,6 +69,11 @@ def __init__(self, world_count: int = 1, config: Config | None = None, device: s
         self.shape_source_ptr: wp.array[wp.uint64] | None = None
         self.shape_texture_ids: wp.array[wp.int32] | None = None
         self.shape_mesh_data_ids: wp.array[wp.int32] | None = None
+        self.shape_types_host: list[int] = []
+        self.shape_has_mesh_uvs: list[bool] = []
+        self.shape_uses_projected_texture: list[bool] = []
+        self.shape_texture_ids_host: list[int] = []
+        self.has_shape_projected_textures: bool = False
```
daniela-hase: Do we really need all those new variables on the render context? 
The places where they are used are usually only called once per run, so they can just be computed there on the fly, can't they?

nblauch: Agreed. After the unified `closest_hit.normal` path, these fields were only leftover bookkeeping from the earlier static projected-texture flag.

Removed them in `95d642d5`, along with the helper that computed `has_shape_projected_textures`. The UV-less mesh tests now assert rendered projection behavior directly instead of reading `RenderContext` internals.

newton/_src/sensors/warp_raytrace/render.py
nblauch: Handled in the latest branch state: `render_megakernel` now keeps a single `textures.sample_texture(...)` call for textured hits, and the shape/UV/projection routing lives in `textures.py`.

I also removed the now-dead projected-texture kernel state flag in follow-up commit `95d642d5`.

nblauch: hi @daniela-hase , are your issues resolved?

newton/_src/sensors/warp_raytrace/render.py
```diff
@@ -220,9 +220,11 @@ def render_megakernel(
                         mesh_data,
                         shape_mesh_data_ids[closest_hit.shape_index],
                         hit_point,
+                        closest_hit.normal,
                         closest_hit.bary_u,
                         closest_hit.bary_v,
                         closest_hit.face_idx,
+                        wp.static(config.texture_projection_mode),
```
daniela-hase: I think we could move this down into the function as well, then we can do a `if wp.static(config.texture_projection_mode == TextureProjectionMode.TRIPLANAR)` in there and staticly remove a bunch of code :)

You'll just have to wrap the `textures.sample_textures` function similarly like how for example the `lighting.compute_lighting` is wrapped in a `lighting.create_compute_lighting_function`.

So something like `sample_texture = textures.create_sample_textures_function(config, state)`

newton/_src/sensors/warp_raytrace/render.py
daniela-hase: Yep, I just added one more comment, sorry! 😅🫣

But otherwise this looks great now :D

newton/_src/sensors/warp_raytrace/render.py
nblauch: Implemented in `37ca8c68`.

`render_megakernel` now captures:

```python
sample_texture = textures.create_sample_texture_function(config)
```

and `textures.py` uses `wp.static(config.texture_projection_mode == TextureProjectionMode.TRIPLANAR)` inside that generated Warp function, so the unused cubic/triplanar branch can be compiled out. I left `state` out of the factory signature because texture sampling does not need it.

Validation:

```bash
uv run --extra dev ruff check newton/_src/sensors/warp_raytrace/render.py newton/_src/sensors/warp_raytrace/textures.py
uv run --extra dev ruff format --check newton/_src/sensors/warp_raytrace/render.py newton/_src/sensors/warp_raytrace/textures.py
uvx pre-commit run -a
```

newton/_src/sensors/warp_raytrace/render.py
nblauch: Handled that extra comment in `37ca8c68` by wrapping texture sampling in `textures.create_sample_texture_function(config)` and moving the `wp.static(config.texture_projection_mode == TextureProjectionMode.TRIPLANAR)` check into the generated Warp function. I also replied with validation details on the specific thread.

nblauch:
I reran the projection benchmark locally after specializing the texture sampler by `texture_projection_mode`.

Previous benchmark result:

| scene | mode | mean ms | relative vs cubic |
|---|---:|---:|---:|
| primitives | cubic | 0.034 | 1.000x |
| primitives | triplanar | 0.039 | 1.149x |
| uvless_bunny_mesh | cubic | 0.483 | 1.000x |
| uvless_bunny_mesh | triplanar | 0.529 | 1.094x |

Latest local result:

| scene | mode | mean ms | relative vs cubic |
|---|---:|---:|---:|
| primitives | cubic | 0.035 | 1.000x |
| primitives | triplanar | 0.037 | 1.078x |
| uvless_bunny_mesh | cubic | 0.420 | 1.000x |
| uvless_bunny_mesh | triplanar | 0.480 | 1.142x |

Direct pre/post timing delta:

| scene | mode | previous ms | latest ms | change |
|---|---:|---:|---:|---:|
| primitives | cubic | 0.034 | 0.035 | 0.97x speedup / +2.9% time |
| primitives | triplanar | 0.039 | 0.037 | 1.05x speedup / -5.1% time |
| uvless_bunny_mesh | cubic | 0.483 | 0.420 | 1.15x speedup / -13.0% time |
| uvless_bunny_mesh | triplanar | 0.529 | 0.480 | 1.10x speedup / -9.3% time |

The primitive cubic case is effectively noise-sized at this scale, but the mesh case shows the sampler specialization improving both cubic and triplanar paths. Triplanar still has a modest cost over cubic in the latest run, which is expected because it samples/blends across three projection axes.

daniela-hase (dismissed):

newton/_src/sensors/warp_raytrace/render.py
daniela-hase: Awesome, thank you! :)

newton/_src/utils/import_usd.py
```diff
@@ -540,9 +540,15 @@ def _get_mesh_with_visual_material(prim: Usd.Prim, *, path_name: str) -> Mesh:
...
-            mesh.texture = None
-        if material_props.get("color") is not None and mesh.texture is None:
+            warnings.warn(
+                f"Mesh {path_name} has a texture but no UVs; texture will use projected UVs.",
+                stacklevel=2,
+            )
+        if mesh.texture is not None:
+            # Texture provides albedo; use white so it renders at full brightness.
+            # (diffuse_color_constant is the untextured fallback, not a multiplier.)
+            mesh.color = (1.0, 1.0, 1.0)
```
adenzler-nvidia, line 964: This forces `mesh.color` to white for *all* textured meshes, not just the no-UV projected case this PR targets. For UV-authored textured meshes the previous behavior left `mesh.color = None`, which resolves to a palette color in `add_shape` and gets multiplied into the texture — so this is a (probably desirable, full-brightness) appearance change for existing assets too, not only for the new projection path.

Two questions:

1. Is applying this to UV-authored textured meshes intentional? If so, it's worth a `CHANGELOG.md` note since it changes how existing textured meshes render.
2. The GeomSubset path in `_make_visual_submesh` uses the opposite precedence (authored `color` wins, white only as a fallback). Should the two paths be reconciled so the same textured asset gets a consistent base color regardless of whether it uses material subsets?

newton/tests/golden_data/test_sensor_tiled_camera/projection_modes/generate_projection_mode_artifacts.py
```diff
@@ -0,0 +1,305 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
+
+"""Generate temporary PR-review artifacts for texture projection modes.
+
+This script renders cubic and triplanar checkerboard projection comparisons for
+``SensorTiledCamera``. The generated images are intended as PR review evidence
+and can be removed after the projection-mode change is accepted.
```
adenzler-nvidia: Since these are flagged as temporary, two questions on whether to make them durable instead of removing them:

**Golden images:** this test file already has pixel-level golden coverage — `test_golden_image` compares the render against `color.npy`/`depth.npy` via `__compare_images` with a tolerance. How important is it to protect the projection modes against regressions? If it's important enough, this should move there: save the cubic/triplanar renders as `.npy` goldens and compare them the same way, rather than committing PNGs that no test reads. That would also strengthen the new tests, which currently only assert color-set membership (a transposed UV or wrong projection axis would still pass). The 64×64 renders are ~16 KB each as `.npy` (~66 KB for all four) vs ~425 KB for the current PNGs.

**Benchmark:** the repo has an asv benchmark suite (`asv run`). Is the cubic-vs-triplanar timing meant to be tracked over time? If so, the generation/benchmark script would fit better there than under `golden_data/`. If it's purely a one-off A/B for this PR, the script and the committed `.txt` result (which is hardware-specific) can just be dropped after review.

Either way `golden_data/` is for fixtures consumed by tests, so a script + PNGs + a benchmark text file is a bit of a mismatch — mainly want to avoid leaving unreferenced artifacts in the tree. Happy to go whichever way you prefer.

newton/_src/utils/import_usd.py
```diff
@@ -540,9 +540,15 @@ def _get_mesh_with_visual_material(prim: Usd.Prim, *, path_name: str) -> Mesh:
         if texture is not None:
             mesh.texture = texture
         if mesh.texture is not None and mesh.uvs is None:
-            logger.info("Mesh %s: dropping texture because UVs could not be recovered.", path_name)
-            mesh.texture = None
-        if material_props.get("color") is not None and mesh.texture is None:
+            warnings.warn(
+                f"Mesh {path_name} has a texture but no UVs; texture will use projected UVs.",
```
adenzler-nvidia: This path now keeps the texture and projects it when UVs are absent. But the parallel material-subset path, `_make_visual_submesh` (around line 656), wasn't updated to match — it still does `logger.info("...dropping texture...")` + `submesh.texture = None`. So the same UV-less textured asset renders projected when loaded as a single material, but untextured when loaded via GeomSubset material subsets. The two paths also use opposite color precedence (the subset path keeps the authored `color`, this path forces white). Should the subset path route through projection too, so the result is consistent regardless of how the material happens to be authored?

newton/_src/utils/import_usd.py
```diff
@@ -540,9 +540,15 @@ def _get_mesh_with_visual_material(prim: Usd.Prim, *, path_name: str) -> Mesh:
...
+                stacklevel=2,
+            )
+        if mesh.texture is not None:
```
adenzler-nvidia, line 962: This behavior change (texture kept instead of dropped when UVs are absent, plus `mesh.color` forced to white) doesn't appear to have a regression test — `test_import_usd.py` is unchanged in this PR. Since this path is CPU-testable (USD parse, no GPU needed), a small test that imports a UV-less textured mesh and asserts `mesh.texture is not None` and `mesh.color == (1.0, 1.0, 1.0)` would lock in the new behavior. There are existing `assertWarnsRegex` examples in that file if you'd also like to assert the new warning.

jcarius-nv:

jcarius-nv:
@nblauch would you mind updating this PR again and bring it to completion?

nblauch:
@jcarius-nv I will try to rebase. Seems upstream has drifted a lot, but the changes here are mostly still necessary.

newton/_src/utils/import_usd.py
nblauch: Addressed in da15ab82 after rebasing. White is intentional for every base-color-textured mesh: upstream #3628 already established that policy for full meshes, and this update gives `_make_visual_submesh` the same texture-first precedence. The `[Unreleased]` changelog wording now covers full meshes and subsets, and the new fixtures include both an authored scalar color and a texture while asserting the imported base stays white.

newton/tests/golden_data/test_sensor_tiled_camera/projection_modes/generate_projection_mode_artifacts.py
nblauch: Addressed in 8fd39045. I removed the generator, unconsumed PNGs, and hardware-specific benchmark text. They are replaced by four 64×64 `.npy` fixtures consumed by `test_texture_projection_golden_images` through the existing pixel-comparison helper: primitive and UV-less mesh, each in cubic and triplanar mode. The timing comparison remains PR-only rather than becoming an ASV benchmark.

newton/_src/utils/import_usd.py
nblauch: Addressed in da15ab82. The GeomSubset path now retains UV-less textures for projection, logs the projected-UV fallback at info level, and uses the same white texture base as the full-mesh path. A CPU regression test imports an all-textured UV-less material subset and verifies retained texture, absent UVs, and white base color.

newton/_src/utils/import_usd.py
nblauch: Addressed in da15ab82. I added CPU tests for both the full-mesh and material-subset paths. They verify that the texture is retained, UVs remain absent, the base color is white, and the projected-UV diagnostic is logged. I kept the diagnostic at info level to match the current upstream importer and its strict-warnings coverage.

nblauch:
Rebased and force-pushed this PR onto current `newton-physics/main` as two clean commits. The update incorporates the newer USD material work from #3628, preserves UV-less textures for both full meshes and material subsets, replaces the temporary review artifacts with four test-consumed NPY goldens, and resolves the outstanding review threads.

Validation completed locally:

- 22 `SensorTiledCamera` tests
- 70 `TestImportSampleAssetsParsing` tests
- red/green verification against an isolated upstream-main checkout
- `uv run docs/generate_api.py`
- `uvx pre-commit run -a`

@adenzler-nvidia @daniela-hase, this is ready for another look when you have a chance.

nblauch:
@andrewkaufman it may be helpful to get your eyes on this too, given the rewrite based on your #3628 , cc @jcarius-nv

newton/tests/test_sensor_tiled_camera.py
```diff
@@ -631,6 +631,72 @@ def test_deprecated_checkerboard_material_to_all_shapes_warns(self):
...
+            ("primitive", SensorTiledCamera.TextureProjectionMode.CUBIC),
+            ("primitive", SensorTiledCamera.TextureProjectionMode.TRIPLANAR),
+            ("uvless_mesh", SensorTiledCamera.TextureProjectionMode.CUBIC),
+            ("uvless_mesh", SensorTiledCamera.TextureProjectionMode.TRIPLANAR),
+        )
+
+        for shape_kind, projection_mode in cases:
+            with self.subTest(shape_kind=shape_kind, projection_mode=projection_mode.name):
+                actual = self._render_checkerboard_projection(shape_kind, projection_mode)
+                golden = np.load(
```
jcarius-nv: These four `.npy` fixtures are appropriately test-consumed and follow the existing tiled-camera golden-data precedent, but I could not find a canonical regeneration path after the earlier artifact generator was removed. Could you document the exact command or workflow for regenerating all four files and inspecting an intentional visual change? An opt-in test flag, environment variable, or nearby comment would be enough; preview PNGs can remain local and uncommitted.

newton/tests/test_sensor_tiled_camera.py
```diff
@@ -631,6 +631,72 @@ def test_deprecated_checkerboard_material_to_all_shapes_warns(self):
...
+            width,
+            height,
+            camera_fovs=math.radians(45.0),
+        )
+        albedo_image = sensor.utils.create_albedo_image_output(width, height, camera_count=1)
+        sensor.update(model.state(), camera_transforms, camera_rays, albedo_image=albedo_image)
+        return albedo_image.numpy()
+
+    @unittest.skipUnless(wp.is_cuda_available(), "Requires CUDA")
+    def test_texture_projection_golden_images(self):
```
jcarius-nv, line 681: Newton's test guidelines require every test method to have a triple-double-quoted, imperative summary docstring. Please add one here, for example: `"""Verify cubic and triplanar projection against golden images."""`

newton/tests/test_import_usd.py
```diff
@@ -7042,6 +7043,78 @@ def test_textured_visual_mesh_uses_white_base_color(self):
...
+            shape_path = "/Body/VisualMesh/textured"
+        else:
+            UsdShade.MaterialBindingAPI.Apply(mesh.GetPrim()).Bind(material)
+            shape_path = "/Body/VisualMesh"
+
+        return stage, shape_path
+
+    @unittest.skipUnless(USD_AVAILABLE, "Requires usd-core")
+    def test_uvless_textured_visual_mesh_uses_projected_uvs(self):
+        """A full visual mesh retains its texture when UVs are unavailable."""
```
jcarius-nv: Newton's test guidelines require imperative summary docstrings. This test and `test_uvless_textured_visual_mesh_subset_uses_projected_uvs()` currently use declarative “A … retains” summaries, and `test_uv_length_mismatch_uses_info_logging()` says “diagnostics surface.” Please rewrite all three as concise `"""Verify ..."""` summaries.

newton/_src/sensors/warp_raytrace/types.py
```diff
@@ -40,6 +40,16 @@ class GaussianRenderMode(enum.IntEnum):
     """Quality Render Mode, collect hits until minimum transmittance is reached"""
 
 
+class TextureProjectionMode(enum.IntEnum):
```
jcarius-nv, line 43: Discussion prompt: do we need selectable triplanar projection as public API in this slice? The reported failure is addressed by preserving UV-less textures and providing a cubic fallback; exposing `TextureProjectionMode` and `RenderConfig.texture_projection_mode` makes both modes a lasting API commitment. If a concrete asset or robotics workload needs selectable triplanar projection, could we capture it here? Otherwise I wonder whether cubic should ship first while triplanar remains private until that demand is established. This is not a correctness objection—mainly a request to make the API trade-off explicit.

jcarius-nv (commented):
A few follow-ups from a current-head review. The public-API comment is intentionally a design discussion prompt rather than a correctness objection.

daniela-hase (approved):

daniela-hase (dismissed):
The changes to the renderer look good to me - but I'll leave the changes to the public API up to the Newton folks :)

newton/tests/test_sensor_tiled_camera.py
nblauch: Addressed in `413ccc69`.

The golden test now has an opt-in regeneration path:

```bash
NEWTON_UPDATE_TEXTURE_PROJECTION_GOLDENS=1 \
    uv run --extra dev -m newton.tests -k test_texture_projection_golden_images
```

That rewrites all four canonical `.npy` fixtures and writes four individual PNG previews plus a labeled contact sheet to the platform temporary directory (`/tmp/newton_texture_projection_previews` on Linux). Normal execution remains comparison-only.

I exercised both modes from the pushed head: regeneration left the tracked `.npy` fixtures unchanged, the normal golden comparison passes, and I inspected the generated contact sheet. The review copies are hosted separately from the PR diff and embedded in the public-API discussion below.

newton/tests/test_sensor_tiled_camera.py
nblauch: Addressed in `413ccc69`: `test_texture_projection_golden_images()` now has the required triple-double-quoted imperative summary, `"""Verify cubic and triplanar projection against golden images."""`

newton/tests/test_import_usd.py
nblauch: Addressed in `413ccc69`. All three summaries now use concise imperative `"""Verify ..."""` wording:

- full visual-mesh texture retention without UVs
- material-subset texture retention without UVs
- info-level UV-fallback diagnostics

The three focused USD tests pass.

newton/_src/sensors/warp_raytrace/types.py
nblauch: Thanks — this is a fair API trade-off to call out explicitly.

The concrete motivation is smooth UV-less assets. The bunny comparison makes the distinction much clearer than the small regression-test scenes: cubic projection switches abruptly between dominant axes, producing visible patch boundaries and broken checker continuity over the curved surface. Triplanar blends those transitions into a much more coherent mapping, at the expected cost of some softening where projections meet.

Close-up comparison of cubic and triplanar projection on the same smooth UV-less bunny

<details>
<summary>Full bunny comparison including shaded renders</summary>

Cubic and triplanar albedo and shaded renders of the UV-less bunny

</details>

<details>
<summary>64×64 regression-test goldens</summary>

These smaller images are the canonical test-consumed cases rather than the main visual-quality evidence.

| Cubic | Triplanar |
|---|---|
| Primitive with cubic projection | Primitive with triplanar projection |
| UV-less mesh with cubic projection | UV-less mesh with triplanar projection |

</details>

The sampler is compile-time-specialized by mode, and the PR benchmark measured triplanar at about 15% over cubic for primitives and 9% for the UV-less bunny. That makes selectable triplanar a useful visual-quality option without putting its unused path into cubic kernels.

That said, I do not have a named downstream robotics workload to cite, and I agree that the lasting public-API decision ultimately belongs to the Newton team. My main priority is getting UV-less texture support shipped. I have kept both modes public for now, but I am happy to collapse this PR to the cubic fallback if the team considers that the preferred and fastest route to merge.

These PNGs were regenerated from pushed head `413ccc69`; the canonical regression fixtures remain the `.npy` files consumed by the test. The PNG review copies live on a separate artifact branch and are not part of this PR's diff.

jcarius-nv: Thanks, this answers the API trade-off for me. Triplanar projection is a conventional renderer capability for UV-less geometry, and the bunny comparison demonstrates a meaningful improvement over cubic projection on smooth surfaces. For completeness, I think it makes sense to keep it public here.

The maintenance commitment also looks appropriately small: the implementation is a compact sampler plus a two-value enum and compile-time routing, reusing the existing texture sampler, hit normal, and render configuration. Cubic kernels do not retain the unused triplanar path. The algorithm and enum semantics are unlikely to need incompatible changes; likely future requests such as blend sharpness, mapping scale, or per-shape overrides would be additive. The four small test-consumed goldens and documented regeneration workflow provide proportionate regression coverage.

So I’m comfortable with this API and consider the discussion resolved.

jcarius-nv (approved):
Thank you!

PR #2327, opened 2026-04-06
Apply ALM hard constraints to VBD rigid bodies

### LINKED ISSUE #1062: Apply hard-constraint collision when consistent contact pairs are supported (AVBD)
Explore and implement a hard-constraint collision model to replace soft penalty forces when consistent contact pairs are available.

### LINKED ISSUE #1453: Use contact matcher for rigid body contact persistency.
Use contact matcher, and update warm start for contacts.

### LINKED ISSUE #2108: Overflow warning for VBD rigid-rigid contact.
When per-body contact buffer is overflowed, we should warn users.

### LINKED ISSUE #2326: Apply ALM hard constraints to VBD rigid bodies
Add option to use ALM hard constraints in addition to soft constraints for VBD rigid bodies.

### LINKED ISSUE #2450: [BUG] Collision pipeline unreasonable memory usage
### Bug Description

If i increase the number of elements in the `example_cable_pile.py` from 40 to 200, the example will not run on my 4080 Laptop GPU due to lack of memory. Hower, by simply calling `newton.CollisionPipeline(self.model)`  after `self.model = builder.finalize()`, but  before `newton.solvers.SolverVBD( self.model, iterations=self.sim_iterations, friction_epsilon=0.1,)`  the simulation runs without problems. The memory use is drastically reduced and the dedicated memory use is reported as less than 25%.

As the collision pipeline should not be changed by this modification, the behavior is unexpected.

The reproducible script is a modfied version of `example_cable_pile.py`. The memory problem arises if tline 167 is commented out.

### Reproduction Script

```python

###########################################################################
# Example Cable Pile
#
# Demonstrates complex cable-to-cable contact and settling behavior.
# Creates a pile of 40 cables (10 per layer x 4 layers) with alternating
# orientations (X/Y axis) and sinusoidal waviness. Tests multi-body contact
# resolution, stacking stability, and friction in dense cable assemblies.
#
###########################################################################

import math

import numpy as np
import warp as wp

import newton
import newton.examples


class Example:
    def __init__(
        self,
        viewer,
        args=None,
        slope_enabled: bool = False,
        slope_angle_deg: float = 20.0,
        slope_mu: float | None = None,
    ):
        # Store viewer and arguments
        self.viewer = viewer
        self.args = args

        # Simulation cadence
        self.fps = 60
        self.frame_dt = 1.0 / self.fps
        self.sim_time = 0.0
        self.sim_substeps = 10
        self.sim_iterations = 5

        # Cable pile parameters
        self.num_elements = 200 # CHANGED FROM 40 TO 200
        segment_length = 0.05  # 5cm segments
        self.cable_length = self.num_elements * segment_length
        cable_radius = 0.012

        # Layers and lanes
        layers = 4
        lanes_per_layer = 10
        # Increase spacing to accommodate lateral waviness without initial intersections
        lane_spacing = max(8.0 * cable_radius, 0.15)
        layer_gap = cable_radius * 6.0

        builder.rigid_gap = 0.05  # Default for all shapes

        rod_bodies_all: list[int] = []

        # Set default material properties before adding any shapes
        # Default config will be used by plane and any shape without explicit cfg
        builder.default_shape_cfg.ke = 1.0e6  # Contact stiffness (used by plane)
        builder.default_shape_cfg.kd = 1.0e-1  # Contact damping
        builder.default_shape_cfg.mu = 5.0  # Friction coefficient

        cable_shape_cfg = newton.ModelBuilder.ShapeConfig(
            density=builder.default_shape_cfg.density,
            ke=builder.default_shape_cfg.ke,
            kd=builder.default_shape_cfg.kd,
            kf=builder.default_shape_cfg.kf,
            ka=builder.default_shape_cfg.ka,
            mu=builder.default_shape_cfg.mu,
            restitution=builder.default_shape_cfg.restitution,
        )

        # Generate a ground plane (optionally sloped for friction tests)
        if slope_enabled:
            angle = math.radians(slope_angle_deg)
            rot = wp.quat_from_axis_angle(wp.vec3(1.0, 0.0, 0.0), angle)

            # Optionally override friction for the slope without changing defaults for other shapes
            slope_cfg = builder.default_shape_cfg
            if slope_mu is not None:
                slope_cfg = newton.ModelBuilder.ShapeConfig(
                    density=builder.default_shape_cfg.density,
                    ke=builder.default_shape_cfg.ke,
                    kd=builder.default_shape_cfg.kd,
                    kf=builder.default_shape_cfg.kf,
                    ka=builder.default_shape_cfg.ka,
                    mu=slope_mu,
                    restitution=builder.default_shape_cfg.restitution,
                )

            builder.add_shape_plane(
                width=10.0,
                length=10.0,
                xform=wp.transform(wp.vec3(0.0, 0.0, 0.0), rot),
                body=-1,
                cfg=slope_cfg,
            )
        else:
            builder.add_ground_plane()

        # Build layered lanes of cables with alternating orientations (no initial intersections)
        for layer in range(layers):
            orient = "x" if (layer % 2 == 0) else "y"
            z0 = 0.3 + layer * layer_gap
            # Generate parallel lanes along the orthogonal axis only
            for lane in range(lanes_per_layer):
                offset = (lane - (lanes_per_layer - 1) * 0.5) * lane_spacing
                if orient == "x":
                    start = wp.vec3(0.0, offset, z0)
                else:
                    start = wp.vec3(offset, 0.0, z0)

                # Regular waviness and no twist for repeatable layout across layers
                wav = 0.5
                twist = 0.0

                dir_vec = wp.vec3(1.0, 0.0, 0.0) if orient == "x" else wp.vec3(0.0, 1.0, 0.0)
                ortho_vec = wp.vec3(0.0, 1.0, 0.0) if orient == "x" else wp.vec3(1.0, 0.0, 0.0)

                # Center the cable around start to avoid initial overlaps with neighbors.
                cable_length = float(self.cable_length)
                start0 = start - 0.5 * cable_length * dir_vec
                pts = newton.utils.create_straight_cable_points(
                    start=start0,
                    direction=dir_vec,
                    length=cable_length,
                    num_segments=int(self.num_elements),
                )

                # Apply sinusoidal waviness along an orthogonal axis (optional).
                cycles = 2.0
                waviness_scale = 0.05  # amplitude fraction of length when wav=1.0
                if wav > 0.0:
                    for i in range(len(pts)):
                        t = i / self.num_elements
                        phase = 2.0 * math.pi * cycles * t
                        amp = wav * cable_length * waviness_scale
                        pts[i] = pts[i] + ortho_vec * (amp * math.sin(phase))

                edge_q = newton.utils.create_parallel_transport_cable_quaternions(pts, twist_total=float(twist))

                rod_bodies, _rod_joints = builder.add_rod(
                    positions=pts,
                    quaternions=edge_q,
                    radius=cable_radius,
                    cfg=cable_shape_cfg,
                    bend_stiffness=1.0e1,
                    bend_damping=5.0e-1,
                    stretch_stiffness=1.0e6,
                    stretch_damping=1.0e-4,
                    label=f"cable_l{layer}_{lane}",
                )
                rod_bodies_all.extend(rod_bodies)

        # Color bodies for VBD solver
        builder.color()

        # Finalize model

        # Create VBD solver for rigid body simulation
        newton.CollisionPipeline(self.model) # COMMENT OUT TO TRIGGER UNWANTED BEHAVIOR
        self.solver = newton.solvers.SolverVBD(
            self.model,
            iterations=self.sim_iterations,
            friction_epsilon=0.1,
        )

        # Initialize states and contacts
        self.state_0 = self.model.state()
        self.state_1 = self.model.state()

        self.contacts = self.model.contacts()

        # Optional capture for CUDA
        self.capture()

    def capture(self):
        """Capture simulation loop into a CUDA graph for optimal GPU performance."""
        if wp.get_device().is_cuda:
            with wp.ScopedCapture() as cap:
                self.simulate()
            self.graph = cap.graph
        else:
            self.graph = None

    def simulate(self):
        """Execute all simulation substeps for one frame."""
        for _substep in range(self.sim_substeps):
            self.state_0.clear_forces()

            # Apply forces to the model
            self.viewer.apply_forces(self.state_0)

            # Collide for contact detection
            self.model.collide(self.state_0, self.contacts)

            self.solver.step(
                self.state_0,
                self.state_1,
                self.control,
                self.contacts,
                self.sim_dt,
            )

            # Swap states

    def step(self):
        """Advance simulation by one frame (either via CUDA graph or direct execution)."""
        if self.graph:
            wp.capture_launch(self.graph)
        else:
            self.simulate()

    def render(self):
        """Render the current simulation state to the viewer."""
        self.viewer.log_state(self.state_0)
        self.viewer.log_contacts(self.contacts, self.state_0)
        self.viewer.end_frame()

    def test_final(self):
        """Test cable pile simulation for stability and correctness (called after simulation)."""
        cable_radius = 0.012
        cable_diameter = 2.0 * cable_radius  # 0.024m
        layers = 4

        # Use same tolerance for ground penetration and pile height offset
        tolerance = 0.1  # Soft contacts allow some penetration and gaps

        # Calculate maximum expected height for SETTLED pile
        # After gravity settles the pile, cables should be stacked:
        # 4 layers = 4 cable diameters high (approximately)
        # Plus compression tolerance and contact gaps
        max_z_settled = layers * cable_diameter + tolerance
        ground_tolerance = tolerance

        # Check final state after viewer has run 100 frames (no additional simulation needed)
        if self.state_0.body_q is not None and self.state_0.body_qd is not None:
            body_positions = self.state_0.body_q.numpy()
            body_velocities = self.state_0.body_qd.numpy()

            # Test 1: Check for numerical stability
            assert np.isfinite(body_positions).all(), "Non-finite positions"
            assert np.isfinite(body_velocities).all(), "Non-finite velocities"

            # Test 2: Check physical bounds - cables should stay within pile height
            z_positions = body_positions[:, 2]  # Z positions (Newton uses Z-up)
            min_z = np.min(z_positions)
            max_z_actual = np.max(z_positions)

            assert min_z > -ground_tolerance, (
                f"Cables penetrated ground too much: min_z={min_z:.3f} < {-ground_tolerance:.3f}"
            )
            assert max_z_actual < max_z_settled, (
                f"Pile too high: max_z={max_z_actual:.3f} > expected {max_z_settled:.3f} "
                f"(4 layers x {cable_diameter:.3f}m diameter + tolerance)"
            )

            # Test 3: Velocity should be reasonable (pile shouldn't explode)
            assert (np.abs(body_velocities) < 5e2).all(), "Velocities too large"


if __name__ == "__main__":
    # Parse arguments and initialize viewer
    viewer, args = newton.examples.init()

    # Create example and run
    # For sloped friction tests, use: Example(viewer, args, slope_enabled=True, slope_angle_deg=20.0, slope_mu=0.8)

    newton.examples.run(example, args)
```

### System Information

Newton: 1.0.0
GPU: NVIDIA RTX 4080 Laptop
OS: nvidia/cuda:12.6.2-devel-ubuntu22.04 Docker image
Python: 3.10.12
warp: 1.12.1

### LINKED ISSUE #2486: VBD cable: remove length normalization for better stiffness ratio.
With non-uniform segment lengths, add_rod's per-segment length normalization produces large per-joint stiffness ratios across the rod, which degrades VBD convergence. 
Propose: disable length normalization by default, and change the documented units of stretch_stiffness and bend_stiffness.

jumyungc:
## Description

Included:

- Body-body contact hard constraints.
- Joint hard constraints.
- Contact overflow warning.
- Fixed penalty by default, with optional AVBD penalty ramping.
- No implicit length normalization in cable stiffness, so add_rod() / add_rod_graph() use direct per-joint stiffness values consistent with other joint APIs.

Not included:

- Drive/limit hard constraints.
- Body-particle contact warm start / contact matching.
- Body-particle contact hard constraints.

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **New Features**
  * GPU graph-capturable kinematic kernels; contact warm-starting with improved contact-matching and interactive picking tweaks in examples.

* **Bug Fixes**
  * Removed a failing regression test for adaptive contact penalties.

* **Tests**
  * Added graph-capturable kinematic tests and migrated many tests to device-driven time.

* **Chores**
  * Retuned defaults: substantially lower contact/cable stiffness and damping, adjusted solver iterations and contact-refresh cadence across examples.

* **Documentation**
  * Changelog updated to reflect contact/joint runtime controls and tuning changes.


eric-heiden:
Great work bringing augmented-Lagrangian hard constraints to VBD rigid bodies. The main concern across this PR is that several public API symbols are renamed or removed without deprecation, and the CHANGELOG is missing entries for the new behavior.

🟡 **Missing CHANGELOG entries for API and behavior changes**

Per `AGENTS.md`, user-facing changes need entries in the `[Unreleased]` section of `CHANGELOG.md`. This PR currently has none. Suggested entries:

| Category | Entry |
|---|---|
| **Added** | `SolverVBD.set_joint_hard()`, `SolverVBD.JointSlot`, new constructor params (`rigid_contact_hard`, `rigid_contact_warmstart`, `rigid_contact_stabilization_alpha`, `rigid_joint_stabilization_alpha`) |
| **Changed** | Default `stretch_stiffness` reduced from `1e9` to `1e4` in `add_joint_cable()`, `add_rod()`, `add_rod_graph()` (ALM hard constraints now handle stretch enforcement). VBD solver uses augmented-Lagrangian hard constraints by default. Dahl friction is auto-detected instead of requiring an explicit flag. |
| **Deprecated** | Renamed constructor params (e.g. `rigid_body_contact_buffer_size` to `rigid_contact_buffer_size`). `set_rigid_history_update()` renamed to `set_rigid_contact_refresh()`. |

Please add these (with migration guidance for the deprecated items) so users upgrading can find them.

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -178,18 +197,20 @@ def __init__(
...
-        rigid_joint_linear_k_start: float = 1.0e4,  # AVBD: initial stiffness seed for linear joint constraints
-        rigid_joint_angular_k_start: float = 1.0e1,  # AVBD: initial stiffness seed for angular joint constraints
-        rigid_joint_linear_ke: float = 1.0e9,  # AVBD: stiffness cap for non-cable linear joint constraints (BALL/FIXED/REVOLUTE/PRISMATIC/D6)
-        rigid_joint_angular_ke: float = 1.0e9,  # AVBD: stiffness cap for non-cable angular joint constraints (FIXED/REVOLUTE/PRISMATIC/D6)
-        rigid_joint_linear_kd: float = 1.0e-2,  # AVBD: Rayleigh damping coefficient for non-cable linear joint constraints
-        rigid_joint_angular_kd: float = 0.0,  # AVBD: Rayleigh damping coefficient for non-cable angular joint constraints
-        rigid_body_contact_buffer_size: int = 64,
-        rigid_body_particle_contact_buffer_size: int = 256,
-        rigid_enable_dahl_friction: bool = False,  # Cable bending plasticity/hysteresis
+        rigid_penalty_decay: float = 0.9,  # AVBD gamma: per-step lambda decay factor
```
eric-heiden: 🔴 **Constructor parameters renamed/removed without deprecation**

Several `SolverVBD.__init__` keyword arguments were renamed or dropped here:

| Old name | New name |
|---|---|
| `rigid_avbd_beta` | *(removed, no replacement)* |
| `rigid_avbd_gamma` | `rigid_penalty_decay` |
| `rigid_contact_k_start` | *(removed)* |
| `rigid_joint_linear_k_start` | *(removed)* |
| `rigid_joint_angular_k_start` | *(removed)* |
| `rigid_body_contact_buffer_size` | `rigid_contact_buffer_size` |
| `rigid_body_particle_contact_buffer_size` | `rigid_particle_contact_buffer_size` |
| `rigid_enable_dahl_friction` | *(removed, auto-detected)* |

Any user code passing the old names will get a `TypeError` on upgrade. Per `AGENTS.md`: *"Breaking changes require a deprecation first. Do not remove or rename public API symbols without deprecating them in a prior release."*

Please add a `**kwargs` block that accepts the old names, emits `DeprecationWarning` with migration guidance, and forwards to the new parameters. For example:

<details>

<summary>Suggested deprecation shim</summary>

```python
_RENAMED_PARAMS = {
    "rigid_avbd_gamma": "rigid_penalty_decay",
    "rigid_body_contact_buffer_size": "rigid_contact_buffer_size",
    "rigid_body_particle_contact_buffer_size": "rigid_particle_contact_buffer_size",
}
_REMOVED_PARAMS = {
    "rigid_avbd_beta",
    "rigid_contact_k_start",
    "rigid_joint_linear_k_start",
    "rigid_joint_angular_k_start",
    "rigid_enable_dahl_friction",
}

for old, new in _RENAMED_PARAMS.items():
    if old in kwargs:
        warnings.warn(
            f"{old!r} is deprecated, use {new!r} instead.",
            DeprecationWarning, stacklevel=2,
        )
        locals()[new] = kwargs.pop(old)

for old in _REMOVED_PARAMS:
    if old in kwargs:
        warnings.warn(
            f"{old!r} is deprecated and no longer has any effect.",
            DeprecationWarning, stacklevel=2,
        )
        kwargs.pop(old)
```

</details>

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -1322,17 +1171,78 @@ def _compute_rigid_force_element_adjacency(self, model):
     # Main Solver Methods
     # =====================================================
 
-    def set_rigid_history_update(self, update: bool):
-        """Set whether the next step() should update rigid solver history (warmstarts).
+    def set_rigid_contact_refresh(self, refresh: bool):
```
eric-heiden: 🔴 **Public method renamed without deprecation shim**

`set_rigid_history_update()` is renamed to `set_rigid_contact_refresh()` here. The examples were updated, but any external user code calling the old name will get an `AttributeError`.

Please keep the old method as a thin deprecated wrapper:

```suggestion
    def set_rigid_contact_refresh(self, refresh: bool):
        """Control whether step() refreshes solver contact state.
```

And add a wrapper elsewhere in the class:

```python
def set_rigid_history_update(self, update: bool):
    """Deprecated: use :meth:`set_rigid_contact_refresh` instead."""
    warnings.warn(
        "set_rigid_history_update() is deprecated, use set_rigid_contact_refresh() instead.",
        DeprecationWarning,
        stacklevel=2,
    )
    self.set_rigid_contact_refresh(update)
```

newton/_src/sim/builder.py
```diff
@@ -4213,7 +4213,7 @@ def add_joint_cable(
 
         """
         # Linear DOF (stretch)
-        se_ke = 1.0e9 if stretch_stiffness is None else stretch_stiffness
+        se_ke = 1.0e4 if stretch_stiffness is None else stretch_stiffness
```
eric-heiden: 🔴 **Default `stretch_stiffness` silently reduced by 5 orders of magnitude**

This changes the default from `1.0e9` to `1.0e4`. The same change also appears in `add_rod()` (line 6424) and `add_rod_graph()` (line 6602).

Most users rely on the default since the examples omit `stretch_stiffness`. After this PR, every simulation using cables or rods will behave differently without any explicit opt-in. The rationale (ALM hard constraints now handle stretch enforcement) makes sense, but the silent nature of the change is risky.

At minimum this needs a `Changed` entry in `CHANGELOG.md` explaining the new default and why. Consider whether a one-release deprecation warning is warranted given the magnitude of the behavior shift, for example:

```python
if stretch_stiffness is None:
    warnings.warn(
        "Default stretch_stiffness changed from 1e9 to 1e4. "
        "Pass stretch_stiffness=1e4 explicitly to silence this warning.",
        FutureWarning, stacklevel=2,
    )
```

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -156,6 +161,20 @@ class SolverVBD(SolverBase):
             state_in, state_out = state_out, state_in
     """
 
+    class JointSlot:
```
eric-heiden, line 176: ⚪ **Minor documentation gaps in `JointSlot` and `set_joint_hard()`**

Two small things:

1. The `STRETCH = 0` / `BEND = 1` aliases are cable-specific, but that is easy to miss when reading the class. A short note like *"STRETCH and BEND are convenience aliases for cables only"* would help.
2. `set_joint_hard()` does not mention the defaults per joint type. Users need to know what they are changing *from*. Consider adding something like: *"Cable stretch defaults to hard; cable bend defaults to soft (for Dahl friction compatibility). All other structural slots default to hard."*

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -1322,17 +1171,78 @@ def _compute_rigid_force_element_adjacency(self, model):
...
+        differs from the simulation step frequency.
 
         Args:
-            update: If True, update rigid warmstart state. If False, reuse previous.
+            refresh: If True, refresh contacts from collision data.  If False, reuse previous contact state.
         """
-        self.update_rigid_history = update
+        self._rigid_contact_refresh = refresh
+
+    def set_joint_hard(self, joint_index: int, hard: bool, slot: int | None = None):
```
eric-heiden: Maybe something like `set_joint_constraint_mode` is more descriptive because "set_joint_hard" can also set a soft constraint mode?

eric-heiden: Also the type hint for `slot` should be `JointSlot | None` to make it easier to discover possible options.

jumyungc:
I’m working on reusing as many existing parameters as possible. I’ll share an update soon for your review. Thank you!

newton/_src/solvers/vbd/solver_vbd.py
gdaviet: Alternatively, could this be controlled by a custom attribute on the model? If I am reading correctly he current approach will to one device-host-device roundtrip of the whole array per joint, which might be inefficient for large scenes

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -1550,74 +1671,39 @@ def _initialize_rigid_bodies(
...
-                            self.joint_penalty_k_min,
-                            self.avbd_gamma,
-                            self.joint_penalty_k,  # input/output
-                        ],
-                        dim=self.joint_constraint_count,
-                        device=self.device,
-                    )
+            # Per-step joint lambda decay + C0 snapshot.
+            # Uses body_q_prev (previous-step pose for kinematic bodies) which
+            # provides implicit feedforward tracking force -- see avbd_rigid_review.md section 6.
```
gdaviet: Is this .md file commited somewhere? 
Would be great if the using `body_q_prev` was not required, this complicates coupling in general. Start of step boyd positions should always be `state_0.body_q`, even for kinematic bodies

newton/_src/solvers/vbd/rigid_vbd_kernels.py
```diff
@@ -527,51 +639,51 @@ def evaluate_rigid_contact_from_collision(
...
-            - h_al: Angular-linear block
-            - h_aa: Angular-angular block
+        10-tuple: (force_a, torque_a, H_ll_a, H_al_a, H_aa_a,
+                   force_b, torque_b, H_ll_b, H_al_b, H_aa_b)
     """
+    lam_n = wp.dot(contact_lam, contact_normal)
 
-    # Early exit: no penetration or zero stiffness
-    if penetration_depth <= 0.0 or contact_ke <= 0.0:
+    if penetration_depth <= _SMALL_LENGTH_EPS and lam_n <= 0.0:
```
gdaviet, line 606: Should _SMALL_LENGTH_EPS be configurable depending on scene scale?

newton/_src/solvers/vbd/rigid_vbd_kernels.py
```diff
@@ -527,51 +639,51 @@ def evaluate_rigid_contact_from_collision(
...
         zero_vec = wp.vec3(0.0)
         zero_mat = wp.mat33(0.0)
         return (zero_vec, zero_vec, zero_mat, zero_mat, zero_mat, zero_vec, zero_vec, zero_mat, zero_mat, zero_mat)
 
+    f_n = contact_ke * penetration_depth + lam_n
+    if f_n <= 0.0 or contact_ke <= 0.0:
+        if lam_n > 0.0 and hard_contact == 1:
+            f_n = 0.0
+            k_rescaled = lam_n / wp.max(wp.abs(penetration_depth), 1.0e-8)
```
gdaviet: The non-monotonicity around penetration-depth=0 is suprising

newton/_src/solvers/vbd/rigid_vbd_kernels.py
```diff
@@ -590,84 +702,76 @@ def evaluate_rigid_contact_from_collision(
...
+        if f_t_len > cone_limit and f_t_len > 0.0:
+            cone_ratio = cone_limit / f_t_len
+            f_t_vec = f_t_vec * cone_ratio
+            contact_ke_t = contact_ke_t * cone_ratio
+
+        f_total = contact_normal * f_n + f_t_vec
+        I3 = wp.identity(n=3, dtype=float)
+        K_total = contact_ke * n_outer + contact_ke_t * (I3 - n_outer)
+    else:
+        # Kinematic-hard or soft: normal penalty + IPC velocity-based friction
```
gdaviet: what's the reason for using a different model for kinematic vs dynamic for hard contacts?

newton/_src/solvers/vbd/rigid_vbd_kernels.py
```diff
@@ -1823,77 +1947,430 @@ def build_body_particle_contact_lists(
...
+    cell_size_inv: float,
+    match_tolerance_sq: float,
+    # Outputs
+    contact_penalty_k: wp.array[float],
+    contact_lambda: wp.array[wp.vec3],
+    contact_C0: wp.array[wp.vec3],
+    contact_material_kd: wp.array[float],
+    contact_material_mu: wp.array[float],
+):
+    """Warmstart body-body contacts using hash table-based contact history.
```
gdaviet: Isn't the collision pipeline supposed to provide contact matching for force warmstarting?  Or are there fundamental reasons for using a custom solution here?

newton/_src/solvers/vbd/solver_vbd.py
jumyungc: Resolved.

newton/_src/solvers/vbd/solver_vbd.py
jumyungc: No name change.

newton/_src/sim/builder.py
jumyungc: Mentioned in changelog.

newton/_src/solvers/vbd/solver_vbd.py
jumyungc: Applied

newton/_src/solvers/vbd/solver_vbd.py
jumyungc: Both the naming and custom attributes recommendation are applied.
JointSlot is a plain class with int class attributes. Would you guys still recommend it?

newton/_src/solvers/vbd/solver_vbd.py
jumyungc: I agree in principle that start-of-step body positions should be available from `state_0.body_q`, and I’m open to moving in that direction.
Currently, VBD (particle and rigid body) does not preserve `body_q` as a start-of-step snapshot; `state_in.body_q` is updated in place during the solve.

jumyungc: The comment mentioning md file is wrong.

newton/_src/solvers/vbd/rigid_vbd_kernels.py
jumyungc: I'd like to leave it fixed. This epsilon is only a numerical zero guard, not the user-facing contact tolerance. Making this configurable would add complexity, and I’d like to revisit it if we see real tiny-scale scene failures..

newton/_src/solvers/vbd/rigid_vbd_kernels.py
jumyungc: Outdated/test version. Now removed.

newton/_src/solvers/vbd/rigid_vbd_kernels.py
jumyungc: Outdated/test version. Now removed.

newton/_src/solvers/vbd/rigid_vbd_kernels.py
jumyungc: Previously, the direction was, each solver takes care of each warmstarting including contact matching.
It changed very recently, and I am also taking contact matching from collsiion pipeline now. (Not vbd internal)

jumyungc:
@eric-heiden @gdaviet  Thanks for the early and great reviews! I've moved this out of draft so it's ready for a full pass. Could you please take another look? :-) 

There are still a few items I want to verify and tighten up, but I'd rather get review feedback on the current base and merge it before it grows.

gdaviet (approved):

PR #2332, opened 2026-04-07
Fix multi-world joint indexing in warp-to-mj coord conversion

jsw7460:
## Description
convert_warp_coords_to_mj_kernel used per-world joint index for joint_type and joint_child instead of the global index. This caused wrong body CoM lookups for worldid > 0. The sibling kernel convert_mj_coords_to_warp_kernel already uses the global index.

## Checklist

## Test plan

     or manual steps. Example:
     ```
       uv run --extra dev -m newton.tests -k test_multiworld_body_properties
     ``` -->
  - Without fix: `test_perworld_com_produces_consistent_physics` FAILS
  - With fix: both tests PASS
 
## Bug fix

**Steps to reproduce:**

1. Create a multi-world model with an articulated body (free + revolute joint)
2. Set different `body_com` per world via `model.body_com` + `notify_model_changed(BODY_INERTIAL_PROPERTIES)`
3. Set non-zero initial angular velocity
4. Step physics and compare world 1 result against a single-world reference with the same parameters
5. World 1 diverges because `convert_warp_coords_to_mj_kernel` reads world 0's `joint_child`

**Minimal reproduction:**
See `newton/tests/test_multiworld_body_properties.py`


## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed joint indexing across multiple simulation worlds, resolving incorrect center-of-mass and joint handling when running concurrent worlds.

* **Tests**
  * Added regression tests that verify per-world body mass/center-of-mass produce simulation results consistent with equivalent single-world runs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -189,6 +189,10 @@
 - Fix `cloth_franka` example Jacobian broken by COM-referenced `body_qd` convention change; adjust robot base height, gripper orientations, and grasp targets for improved reachability (a follow-up PR will migrate the example to `newton.ik`)
 - Fix joint-synthesized CONNECT constraint anchors not updating when `dof_ref` or `joint_X_p` changes at runtime via `notify_model_changed()`
 - Fix WELD constraint data corruption when a model contains both FIXED and revolute/ball loop joints, caused by CONNECT anchor kernels overwriting WELD `eq_data`
+- Fix SDF hydroelastic broadphase scatter kernel using a grid-stride loop with binary search instead of per-pair thread launch
+- Fix `SensorRaycast` ignoring `PLANE` geometry
+- Fix box support-map sign flips from quaternion rotation noise (~1e-14) producing invalid GJK/MPR contacts for face-touching boxes with non-trivial base rotations
+- Fix multi-world coordinate conversion using the wrong body center of mass for replicated worlds
 
 ## [1.0.0] - 2026-03-10
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@
 - Fix MJCF importer creating finite planes from MuJoCo visual half-sizes instead of infinite planes
 - Fix ViewerViser mesh popping artifacts caused by viser's automatic LOD simplification creating holes in complex geometry
 - Fix degenerate zero-area triangles in SDF marching-cubes isosurface extraction by clamping edge interpolation away from cube corners and guarding against near-zero cross products
+- Fix multi-world coordinate conversion using the wrong body center of mass for replicated worlds
 
 ## [1.1.0] - 2026-04-13
 
@@ -189,10 +190,6 @@
 - Fix `cloth_franka` example Jacobian broken by COM-referenced `body_qd` convention change; adjust robot base height, gripper orientations, and grasp targets for improved reachability (a follow-up PR will migrate the example to `newton.ik`)
 - Fix joint-synthesized CONNECT constraint anchors not updating when `dof_ref` or `joint_X_p` changes at runtime via `notify_model_changed()`
 - Fix WELD constraint data corruption when a model contains both FIXED and revolute/ball loop joints, caused by CONNECT anchor kernels overwriting WELD `eq_data`
-- Fix SDF hydroelastic broadphase scatter kernel using a grid-stride loop with binary search instead of per-pair thread launch
-- Fix `SensorRaycast` ignoring `PLANE` geometry
-- Fix box support-map sign flips from quaternion rotation noise (~1e-14) producing invalid GJK/MPR contacts for face-touching boxes with non-trivial base rotations
-- Fix multi-world coordinate conversion using the wrong body center of mass for replicated worlds
 
 ## [1.0.0] - 2026-03-10
 
--- a/newton/_src/solvers/mujoco/kernels.py
+++ b/newton/_src/solvers/mujoco/kernels.py
@@ -644,17 +644,19 @@ def convert_warp_coords_to_mj_kernel(
 ):
     worldid, jntid = wp.tid()
 
+    joint_id = joints_per_world * worldid + jntid
+
     # Skip loop joints — they have no MuJoCo qpos/qvel entries
     q_i = mj_q_start[jntid]
     if q_i < 0:
         return
 
     qd_i = mj_qd_start[jntid]
-    type = joint_type[jntid]
-    wq_i = joint_q_start[joints_per_world * worldid + jntid]
-    wqd_i = joint_qd_start[joints_per_world * worldid + jntid]
+    jtype = joint_type[joint_id]
+    wq_i = joint_q_start[joint_id]
+    wqd_i = joint_qd_start[joint_id]
 
-    if type == JointType.FREE:
+    if jtype == JointType.FREE:
         # convert position components
         for i in range(3):
             qpos[worldid, q_i + i] = joint_q[wq_i + i]
@@ -683,7 +685,7 @@ def convert_warp_coords_to_mj_kernel(
         w_world = wp.vec3(joint_qd[wqd_i + 3], joint_qd[wqd_i + 4], joint_qd[wqd_i + 5])
 
         # Get CoM offset in world frame
-        child = joint_child[jntid]
+        child = joint_child[joint_id]
         com_local = body_com[child]
         com_world = wp.quat_rotate(rot, com_local)
 
@@ -702,7 +704,7 @@ def convert_warp_coords_to_mj_kernel(
         qvel[worldid, qd_i + 4] = w_body[1]
         qvel[worldid, qd_i + 5] = w_body[2]
 
-    elif type == JointType.BALL:
+    elif jtype == JointType.BALL:
         # change quaternion order from xyzw to wxyz
         ball_q = wp.quat(joint_q[wq_i], joint_q[wq_i + 1], joint_q[wq_i + 2], joint_q[wq_i + 3])
         ball_q_wxyz = quat_xyzw_to_wxyz(ball_q)
@@ -714,7 +716,7 @@ def convert_warp_coords_to_mj_kernel(
             # convert velocity components
             qvel[worldid, qd_i + i] = joint_qd[wqd_i + i]
     else:
-        axis_count = joint_dof_dim[jntid, 0] + joint_dof_dim[jntid, 1]
+        axis_count = joint_dof_dim[joint_id, 0] + joint_dof_dim[joint_id, 1]
         for i in range(axis_count):
             ref = float(0.0)
             if dof_ref:
--- a/newton/tests/test_multiworld_body_properties.py
+++ b/newton/tests/test_multiworld_body_properties.py
@@ -150,10 +150,10 @@ class TestMultiworldBodyProperties(unittest.TestCase):
     pass
 
 
-# The buggy code path is warp-to-mj coordinate conversion, which only
-# runs on CUDA (SolverMuJoCo with use_mujoco_cpu=False). Skip CPU devices
-# so CPU-only CI runners don't fail.
-devices = [d for d in get_test_devices() if d.is_cuda]
+# The buggy code path (convert_warp_coords_to_mj_kernel) runs on any
+# device when SolverMuJoCo is used with use_mujoco_cpu=False (default),
+# so this test must run on all available devices.
+devices = get_test_devices()
 for device in devices:
     add_function_test(
         TestMultiworldBodyProperties,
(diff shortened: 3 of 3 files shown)


eric-heiden:
🟡 Nice fix for the multi-world joint indexing in the coordinate conversion kernels.

One more instance of the same per-world indexing pattern was missed in the forward kernel, and the new mass regression test doesn't fully exercise the code path it targets.

---

🟡 **`joint_dof_dim` still uses bare `jntid` in the generic-joint branch**

In `convert_warp_coords_to_mj_kernel` (`newton/_src/solvers/mujoco/kernels.py`, line 641), the generic-joint `else` branch still indexes `joint_dof_dim` with the per-world `jntid`:

```python
axis_count = joint_dof_dim[jntid, 0] + joint_dof_dim[jntid, 1]
```

The reverse kernel (`convert_mj_coords_to_warp_kernel`, line 541) already uses the global `joint_id` for the same lookup:

```python
axis_count = joint_dof_dim[joint_id, 0] + joint_dof_dim[joint_id, 1]
```

Today this is functionally benign because `_validate_homogeneous_worlds()` ensures `joint_dof_dim` values are identical across worlds. But this is exactly the same indexing pattern this PR is correcting everywhere else, and leaving it creates an asymmetry between the forward and reverse kernels. Fixing it now costs nothing and prevents a latent bug if homogeneity validation is ever relaxed.

Suggested fix:

```python
axis_count = joint_dof_dim[joints_per_world * worldid + jntid, 0] + joint_dof_dim[joints_per_world * worldid + jntid, 1]
```

newton/tests/test_multiworld_body_properties.py
```diff
@@ -0,0 +1,208 @@
...
+
+    for _ in range(num_steps):
+        state_0.clear_forces()
+        solver.step(state_0, state_1, control, None, sim_dt)
+        state_0, state_1 = state_1, state_0
+
+    return state_0.joint_q.numpy()
+
+
+def test_perworld_mass_produces_consistent_physics(test, device):
```
eric-heiden: ⚪ Both links in `_build_articulated_model` use `com=(0, 0, 0)`, so the cross-product term in `v_origin = v_com - ω × com_world` vanishes no matter which `joint_child` index is read. This means the test would pass both before and after the indexing fix.

The companion `test_perworld_com_produces_consistent_physics` is the test that actually exercises the CoM-lookup regression. Additionally, `test_randomize_body_mass` in `test_mujoco_solver.py` already covers per-world mass propagation into MuJoCo.

Consider either:
- Giving the base link a non-zero CoM (e.g. `com=wp.vec3(0.02, 0.0, -0.01)`) so varying mass also stresses the CoM-lookup path, or
- Removing this test if the existing coverage is sufficient, saving ~300 simulation steps.

jsw7460:
Thanks for the review! Both comments addressed:

  1. Fixed `joint_dof_dim` indexing in the generic-joint branch (line 643). Extracted `joint_id = joints_per_world * worldid + jntid` once at the top of the kernel and reused it for all 5 lookups, matching the pattern in `convert_mj_coords_to_warp_kernel`.

  2. Removed `test_perworld_mass_produces_consistent_physics`. As you pointed out, the buggy kernel reads `body_com[wrong_child]`, so varying mass alone doesn't trigger the bug regardless of CoM offset (all worlds share the same CoM in the mass test, so the wrong index reads the same value). `test_randomize_body_mass` already covers per-world mass propagatioon. I kept only the CoM test that actually exercises the buggy lookup.

newton/tests/test_multiworld_body_properties.py
```diff
@@ -0,0 +1,149 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 The Newton Developers
```
adenzler-nvidia: SPDX year should be `2026` — this file is new in this PR.

```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
```

newton/tests/test_multiworld_body_properties.py
```diff
@@ -0,0 +1,149 @@
...
+
+import newton
+from newton.solvers import SolverMuJoCo, SolverNotifyFlags
+
+
+class TestMultiworldBodyProperties(unittest.TestCase):
+    """Verify that multi-world simulations with per-world body mass/CoM
+    produce physically consistent results across all worlds."""
+
+    def _build_articulated_model(self, num_worlds: int = 2, device: str = "cuda:0"):
```
adenzler-nvidia: Hardcoding `device="cuda:0"` breaks all CPU-only CI runners (`ValueError: Invalid device identifier: cuda:0`) — this is the sole test failure across ubuntu, ubuntu-arm, macos, and windows.

Please use the `get_test_devices()` / `add_function_test` pattern (like `test_mujoco_solver.py` and other tests in this repo) so the test auto-skips when CUDA is unavailable. An earlier commit (867c6a01) had this conversion but it was lost on the force-push.

newton/tests/test_multiworld_body_properties.py
```diff
@@ -0,0 +1,149 @@
...
+        solver.notify_model_changed(SolverNotifyFlags.BODY_INERTIAL_PROPERTIES)
+
+        state_0 = model.state()
+        state_1 = model.state()
+        control = model.control()
+
+        # Set initial height and angular velocity for all worlds.
+        # Non-zero angular velocity is critical: the bug in
+        # convert_warp_coords_to_mj_kernel affects the velocity
+        # conversion v_origin = v_com - ω × com. With ω=0 the cross
```
adenzler-nvidia: Unicode `×` (MULTIPLICATION SIGN) triggers Ruff RUF003, which is what fails `pre-commit.ci`. Use ASCII `x` instead.

```suggestion
        # convert_warp_coords_to_mj_kernel affects the velocity
        # conversion v_origin = v_com - w x com. With w=0 the cross
```

CHANGELOG.md
```diff
@@ -89,6 +89,7 @@
 ### Fixed
 
 - Fix GL viewer crash when enabling "Gap + Margin" for soft-body-only states with no rigid body transforms
+- Fix incorrect joint indexing in `convert_warp_coords_to_mj_kernel` that used per-world index instead of global index for `joint_type` and `joint_child` lookups, causing wrong CoM references for `worldid > 0`
```
adenzler-nvidia: Per repo guidelines, changelog entries should be user-facing and avoid internal implementation details. This references kernel names, parameter names, and indexing mechanics. Something like:

```suggestion
- Fix multi-world coordinate conversion using the wrong body center of mass for replicated worlds
```

newton/_src/solvers/mujoco/kernels.py
```diff
@@ -570,15 +570,17 @@ def convert_warp_coords_to_mj_kernel(
...
     # Skip loop joints — they have no MuJoCo qpos/qvel entries
     q_i = mj_q_start[jntid]
     if q_i < 0:
         return
 
     qd_i = mj_qd_start[jntid]
-    type = joint_type[jntid]
-    wq_i = joint_q_start[joints_per_world * worldid + jntid]
-    wqd_i = joint_qd_start[joints_per_world * worldid + jntid]
+    type = joint_type[joint_id]
```
adenzler-nvidia: Nit: `type` shadows the Python builtin (Ruff A001, enabled in this repo's `ruff.toml`). The rest of this file uses `jtype` for the same purpose — rename for consistency?

```suggestion
    jtype = joint_type[joint_id]
```

(Would need to update the `if type ==` checks below as well.)

adenzler-nvidia (changes requested):
Thanks for the fix — the core indexing change in `kernels.py` looks correct. A few things still need attention before this can land (the test device issue is the main CI blocker).

newton/tests/test_multiworld_body_properties.py
```diff
@@ -0,0 +1,167 @@
...
+    pass
+
+
+# The buggy code path is warp-to-mj coordinate conversion, which only
+# runs on CUDA (SolverMuJoCo with use_mujoco_cpu=False). Skip CPU devices
+# so CPU-only CI runners don't fail.
+devices = [d for d in get_test_devices() if d.is_cuda]
+for device in devices:
+    add_function_test(
+        TestMultiworldBodyProperties,
```
adenzler-nvidia: The CUDA-only filter here isn't necessary — `convert_warp_coords_to_mj_kernel` runs on any device when `use_mujoco_cpu=False` (the default). The bug manifests on Warp CPU as well. Could you change this to run on all test devices?

```suggestion
devices = get_test_devices()
```

adenzler-nvidia (commented):
Thanks for addressing all the feedback\! One more thing: the test should also run on CPU.

Sibling kernel bug filed as #2471.

newton/tests/test_multiworld_body_properties.py
```diff
@@ -0,0 +1,167 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
```
adenzler-nvidia, line 1: A few questions about the test scope — I wonder if it can be tightened.

The bug causes world-N to silently read world-0's per-world-replicated data. So with per-world-differing body properties and identical initial joint state, buggy code should produce **identical** trajectories across worlds while correct code produces **divergent** ones. Given that:

1. Do the single-world reference runs add anything beyond "world 0 and world 1 diverge"? It seems like a single multi-world run with a `not np.allclose(multi_q[0], multi_q[1])` assertion would catch the same regression at 1/3 the cost.

2. The comment "Using multiple joints/bodies is essential because the indexing bug only manifests when joint_child maps to different bodies across the flat array" — is this actually true? For any `worldid > 0`, buggy indexing reads slot `jntid` instead of `joints_per_world * worldid + jntid`, so the bug should fire with 1 joint per world as long as body properties differ across worlds. Am I missing something about why the child revolute joint is load-bearing here?

3. The `body_com.numpy().reshape(...).assign(...)` patching post-`finalize()` (with the `BODY_INERTIAL_PROPERTIES` notify) — is there a reason to do it this way rather than building each world with its target CoM directly via `add_builder`? The post-hoc mutation seems to be a consequence of using `replicate`, which forces identical worlds.

jsw7460: @adenzler-nvidia 
Thanks for the review. 

After implementing the `assertFalse(allclose(multi_q[0], multi_q[1]))` check, I found it doesn't actually detect the bug. Wanted to walk through why in case I'm missing something.

  **Root cause.** `notify_model_changed(BODY_INERTIAL_PROPERTIES)` syncs
  `model.body_com` into the per-world `mjw_model.body_ipos` array (see `_update_model_inertial_properties` launching `update_body_mass_ipos_kernel`). So MuJoCo's internal dynamics always use the correct per-world CoM, independent of the conversion kernel.

The kernel's role is the Newton <-> MuJoCo velocity convention transform `v_origin = v_com - ω × rotate(rot, body_com[child])`. With the bug, it makes `v_origin` identical across worlds (when initial `v_com` / `ω` / `rot` are set identically in the test), but MuJoCo then simulates each world with its own per-world `body_ipos`. Identical initial `v_origin` + different `body_ipos`  => different dynamics. Trajectories diverge in the buggy case too.

The options I'm considering is either
1. Keep the current test as-is.
2. Round-trip test at the kernel level: `joint_qd → warp_to_mj → mj_to_warp` should be identity; with the bug it drifts by `ω × rotate(delta_com)` per application. It has no stepping, and directly targets the kernel matth, but calls internal conversion methods rather than the public solver surface.

Any preference?

For your second and third points, I agree. Will modify it after we decide (1).

adenzler-nvidia: You're right — (1) is the right call, and my original (1) suggestion doesn't work. I missed that `notify_model_changed(BODY_INERTIAL_PROPERTIES)` syncs per-world `body_ipos` through a separately (and correctly) indexed kernel, so MuJoCo steps with the right per-world inertial state regardless of the conversion bug. Trajectories diverge either way, just to the wrong places in the buggy case. Apologies for the noise.

Between your two options, I'd go with (1). It tests the externally-observable behavior through the public solver surface, which is what we actually want to guard against regressing. (2) is tighter but tests an internal invariant between two kernels — a symmetric break in both would pass.

Happy to proceed with (1) as-is. Could you also resolve the merge conflicts on this branch when you get a chance?

jsw7460: Rebased onto latest main and simplified the test setup per points (2) and (3):
- Single free-floating body per world (removed the articulated revolute joint)
- Build each world directly via add_builder instead of replicate + post-hoc body_com mutation

Keeping the reference-based comparison as agreed in option (1).

adenzler-nvidia (dismissed):
All prior feedback addressed — multi-world indexing fix looks right, test covers the actual bug path (per-world CoM divergence). LGTM.

adenzler-nvidia:
@jsw7460 needs another conflict resolve in the changelog, but looks very good now! let's get this in!

adenzler-nvidia (approved):

PR #2341, opened 2026-04-07
Implement log_scalar() plotting in ViewerGL

### LINKED ISSUE #2097: Implement plotting functionality in ViewerGL
Our examples, like example_sensor_contact.py and [example_basic_plotting.py](https://github.com/newton-physics/newton/pull/2095) should use `Viewer.log_scalar()` for getting a time series plot.

vreutskyy:
## Description

Implement `ViewerGL.log_scalar()` so that calling it accumulates values in a rolling buffer and automatically renders live ImGui line plots in a dedicated "Plots" window. Each unique name gets its own collapsible plot with the latest value shown as overlay text.

Also simplifies two examples to use the new API:
- `example_basic_plotting` — replaces manual ImGui `plot_lines` calls with `log_scalar()`
- `example_sensor_contact` — removes the custom `ViewerPlot` helper class in favor of a single `log_scalar()` call

Closes #2097

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_basic.example_basic_plotting
uv run --extra dev -m newton.tests -k test_sensors.example_sensor_contact
```

Both pass on CPU and CUDA.

## New feature / API change

```python
import newton

viewer = newton.viewer.ViewerGL(plot_history_size=500)

# In your step() loop:
viewer.log_scalar("Kinetic Energy", kinetic_energy)
viewer.log_scalar("Contact Force", force_z)
# A "Plots" window appears automatically with live time-series plots.
```

## Summary by CodeRabbit

* **New Features**
  * Live scalar telemetry plots in the GL viewer with configurable rolling-history length and per-signal smoothing.

* **Refactor**
  * Example demos switched to the new scalar telemetry API; legacy per-demo plotting components removed.

* **API**
  * Viewer log_scalar signature extended to accept clear (discard previous samples) and smoothing parameters; viewers/backends updated to accept the new parameters.

### DIFF
--- a/newton/_src/viewer/viewer_gl.py
+++ b/newton/_src/viewer/viewer_gl.py
@@ -221,7 +221,6 @@ class ViewerGL(ViewerBase):
         self._scalar_accumulators: dict[str, list[float]] = {}
         self._scalar_smoothing: dict[str, int] = {}
         self._plot_history_size = plot_history_size
-        self._show_plots_window = True
 
         super().__init__()
 
@@ -1174,6 +1173,8 @@ class ViewerGL(ViewerBase):
             smoothing: Number of raw samples to average before committing
                 a point to the plot history.  Defaults to ``1`` (no smoothing).
         """
+        if smoothing < 1:
+            raise ValueError("smoothing must be >= 1")
         val = float(value.item() if hasattr(value, "item") else value)
         buf = self._scalar_buffers.get(name)
         if buf is None:
@@ -2240,7 +2241,7 @@ class ViewerGL(ViewerBase):
 
     def _render_scalar_plots(self):
         """Render an ImGui window with live line plots for all logged scalars."""
-        if not self._scalar_buffers or not self._show_plots_window:
+        if not self._scalar_buffers:
             return
 
         imgui = self.ui.imgui
@@ -2260,7 +2261,7 @@ class ViewerGL(ViewerBase):
             imgui.Cond_.appearing,
         )
 
-        expanded, self._show_plots_window = imgui.begin("Plots", self._show_plots_window)
+        expanded = imgui.begin("Plots")
         if expanded:
             graph_size = imgui.ImVec2(-1, 100)
             n = self._plot_history_size


newton/examples/sensors/example_sensor_contact.py
```diff
@@ -163,15 +158,14 @@ def step(self):
             self._set_shape_colors({plate_shape: self.shape_colors[counterpart_label]})
 
         self.flap_contact_sensor.update(self.state_0, self.contacts)
-        self.plot_window.add_point(np.abs(self.flap_contact_sensor.total_force.numpy()[0, 2]))
+        self.viewer.log_scalar("Flap Contact Force", np.abs(self.flap_contact_sensor.total_force.numpy()[0, 2]))
```
eric-heiden: 🟡 The old `ViewerPlot` that this replaces provided three behaviors that `log_scalar()` does not yet cover:

| Behavior | Old (`ViewerPlot`) | New (`log_scalar`) |
|---|---|---|
| **Smoothing** | `avg=10` averaged 10 raw samples into one plotted point, filtering noisy per-frame contact forces | No smoothing |
| **Episode reset** | `self.plot_window.reset()` cleared the buffer every 8 s reset | `_scalar_buffers` is never cleared, so stale data from the previous episode lingers |
| **History depth** | `n_points=100` x 10-sample avg = ~1 000 steps (~8 s episode) | Default `plot_history_size=250` at ~120 Hz covers only ~2 s |

This means the example's plot will look noticeably different (noisier, shorter window, and data bleeding across resets).

A couple of possible ways to close the gap:

1. Add a `clear: bool = False` argument to the log_scalar method that if set to true will clear the scalar history, and call it from `reset()` in this example.
2. Either bump `plot_history_size` when constructing the viewer here, or add an optional `smoothing` parameter to `log_scalar()`.

newton/_src/viewer/viewer_gl.py
```diff
@@ -2175,6 +2201,45 @@ def _edit_color3(
...
+        )
+
+        expanded, _opened = imgui.begin("Plots", True)
+        if expanded:
+            graph_size = imgui.ImVec2(-1, 100)
+            n = self._plot_history_size
+            for name, buf in self._scalar_buffers.items():
+                # Pad with NaN on the left so the x-axis scale is fixed
+                # but pre-history values are not drawn.
+                arr = np.full(n, np.nan, dtype=np.float32)
```
eric-heiden: ⚪ `np.full(n, np.nan)` and `np.array(buf)` are allocated fresh every frame for every logged scalar. With a handful of signals at 60 fps that is several hundred small allocations per second in the render hot path, adding GC pressure for no reason.

Consider pre-allocating one array per signal (stored alongside the deque in `_scalar_buffers`) and updating it in-place when new values arrive:

<details>

<summary>Sketch</summary>

```python
# In log_scalar(), after appending to the deque:
self._scalar_arrays[name] = None  # invalidate cache

# In _render_scalar_plots(), replace the two allocations:
arr = self._scalar_arrays.get(name)
if arr is None:
    arr = np.full(n, np.nan, dtype=np.float32)
    arr[n - len(buf):] = np.array(buf, dtype=np.float32)
    self._scalar_arrays[name] = arr
```

</details>

Not urgent, but easy to fix while the code is fresh.

PR #2343, opened 2026-04-07
Load per-shape visibility, color information from USD

eric-heiden:
* Consider effective visibility computed from `UsdGeom.Imageable(prim).ComputeVisibility()` for visual and collision shapes. This flag may turn off visibility explicitly for visual shapes which we did not parse previously.
* Load `diffuseColor` property from `UsdPreviewSurface` as shape color for all shape types (not just meshes)

## Summary by CodeRabbit

* **New Features**
  * Builder-level default shape color fallback; improved color resolution that imports authored per-shape colors and OmniPBR diffuse tint for both visual and collider shapes.

* **Bug Fixes**
  * USD import now respects authored visibility (including inherited invisibility) for visuals and colliders.
  * Fixed Gaussian splat transform so its world transform is computed correctly.

* **Tests**
  * Added regression tests for tint/color resolution and visibility/collider edge cases.

* **Documentation**
  * Changelog updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Add `ViewerBase.log_arrows()` for arrow rendering (wide line + arrowhead) in the GL viewer with a dedicated geometry shader
 - Add `enable_multiccd` parameter to `SolverMuJoCo` for multi-CCD contact generation (up to 4 contact points per geom pair)
 - Add `ViewerViser.log_scalar()` for live scalar time-series plots via uPlot
+- Honor `UsdGeomImageable` visibility (including inherited `invisible`) on USD prims imported via `ModelBuilder.add_usd()`; visual shapes, gaussian splats, and collider shapes are imported with `ShapeFlags.VISIBLE` cleared when the prim is effectively invisible, while collision behavior is preserved
 
 ### Changed
 


newton/_src/usd/utils.py
```diff
@@ -1502,6 +1502,41 @@ def _coerce_float(value: Any) -> float | None:
...
+            return value
+    return None
+
+
+def _resolve_diffuse_tint(
+    shader: UsdShade.Shader | None,
+    material: UsdShade.Material | None,
+) -> tuple[float, float, float] | None:
+    """Resolve an authored OmniPBR-style diffuse tint color."""
+    tint_value = _get_input_value(shader, ("diffuse_tint",)) if shader is not None else None
```
andrewkaufman: What shader is providing this diffuse_tint? I don't think that is a "normal" USD thing, its likely specific to some shading system... is it MDL & OmniPBR maybe based on the unittest?

imho we should not aim to support that, but instead support
- primvars:displayColor
  - this is the cannonical fallback color for low-fi visualization supported by every USD enabled app
- UsdPreviewSurface.inputs:diffuseColor
  - can be driven by texture as well. i think we already have support for it?
- The common MaterialX surface shaders (eg openpbr, standard_surface, etc)
  - Defer to the future when content starts showing up with these

newton/_src/utils/import_usd.py
```diff
@@ -447,6 +448,12 @@ def _has_visual_material_properties(material_props: dict[str, Any]) -> bool:
         # Require PBR-like material cues to avoid promoting generic displayColor-only colliders.
         return any(material_props.get(key) is not None for key in ("texture", "roughness", "metallic"))
 
+    def _is_effectively_visible(prim: Usd.Prim) -> bool:
+        imageable = UsdGeom.Imageable(prim)
+        if not imageable:
+            return True
```
andrewkaufman: shouldnt this be false? non-imagable prims are .... not renderable

newton/_src/usd/utils.py
```diff
@@ -1502,6 +1502,41 @@ def _coerce_float(value: Any) -> float | None:
...
+
+def _get_material_input_value(material: UsdShade.Material | None, names: Sequence[str]) -> Any | None:
+    """Return the first authored material input value matching any name in order."""
+    if material is None:
+        return None
+    for name in names:
+        inp = material.GetInput(name)
+        if not inp:
+            continue
+        value = inp.Get()
```
andrewkaufman: when resolving material & shader inputs you generally should use `GetValueProducingAttributes()` instead to follow connections in the material network & find the source input. Note you can also find an output this way, meaning there is no reliable value you could "get" here (as its a computed value by a shading system)

andrewkaufman: You have `_get_input_value` above for this

newton/_src/sim/builder.py
```diff
@@ -191,6 +191,12 @@ def _shape_palette_color(index: int) -> tuple[float, float, float]:
         color = ModelBuilder._SHAPE_COLOR_PALETTE[index % len(ModelBuilder._SHAPE_COLOR_PALETTE)]
         return (color[0] / 255.0, color[1] / 255.0, color[2] / 255.0)
 
+    @staticmethod
+    def _normalize_shape_color(color: Vec3 | None) -> tuple[float, float, float] | None:
```
huidongc: `normalize` is probably misused here. this function casts it to python tuple, which is not typical normlization.

newton/_src/usd/utils.py
```diff
@@ -1502,6 +1502,41 @@ def _coerce_float(value: Any) -> float | None:
         return None
 
 
+def _multiply_colors(
+    color: tuple[float, float, float] | None,
+    tint: tuple[float, float, float] | None,
+) -> tuple[float, float, float] | None:
+    """Multiply two RGB colors componentwise."""
+    if color is None or tint is None:
+        return color
```
huidongc: if color is None and tint is not None maybe we'd better return tint?

newton/_src/usd/utils.py
huidongc: >  is it MDL & OmniPBR maybe based on the unittest?

Yes, it is to support OmniPBR that has been used by the content of IsaacLab. 

Without this short-term approach, we would see notable pixel difference between RTX and newton.

CHANGELOG.md
```diff
@@ -99,6 +93,7 @@
 - Fix articulated bodies drifting laterally on the ground in XPBD solver by solving rigid contacts before joints
 - Fix viewer crash with `imgui_bundle>=1.92.6` when editing colors by normalizing `color_edit3` input/output in `_edit_color3`
 - Fix `hide_collision_shapes=True` not hiding collision meshes that have bound PBR materials
+- Fix USD import to honor authored `visibility = "invisible"` when deciding whether visual and collider shapes are rendered
```
huidongc: please also add doc for tint and default fallback color.

newton/_src/usd/utils.py
huidongc: Image

newton/_src/usd/utils.py
huidongc: I believe we should do this. Some shaders in the kuka.usd, e.g. the allegro_black, only have `diffuse_tint` but not `diffuse_color_constant`. In this case if we don't return the tint value, then the default fallback color or the color palette will be used, then the pixel would be notably different.

huidongc: Image

huidongc: In this case the object is expected to be black, but because we are turning None the default color of 18% gray will be used.

newton/_src/usd/utils.py
andrewkaufman: That just means the Lab content is authored incorreclty though, those assets should have displayColor approximating the diffuse_tint of the OmniPBR. And they shouldn't even use OmniPBR if they're meant to be usable anywhere other than OV apps, they should use UsdPreviewSurface or MaterialX.

huidongc: Fair enough. Let me bring this up to the IsaacLab rendering channel.

newton/_src/usd/utils.py
huidongc: Based on the latest discussion in slack, maybe we'd better limit this `diffuse_tint` workaround for OmniPBR only, something similar to below:

```py
ap = shader.GetPrim().GetAttribute("info:mdl:sourceAsset").Get()
is_omni_pbr = isinstance(ap, Sdf.AssetPath) and ap.authoredPath.endswith("OmniPBR.mdl")

if is_omni_pbr:
    if tint is None:
        return color

    if color is None:
        color = (0.2, 0.2, 0.2)

    return tuple(float(c * t) for c, t in zip(color, tint, strict=True))
```

newton/_src/usd/utils.py
huidongc: Based on the latest discussion in slack, maybe we'd better limit this diffuse_tint workaround for OmniPBR only, this can get us at least some level of render parity before the content moves away from MDL.

newton/_src/usd/utils.py
eric-heiden: Done

newton/_src/usd/utils.py
```diff
@@ -1643,6 +1702,15 @@ def _extract_shader_properties(shader: UsdShade.Shader | None, prim: Usd.Prim) -
...
                     break
 
+    if _is_omni_pbr_shader(shader):
+        tint = _resolve_diffuse_tint(shader, material)
+        color = properties["color"]
+        if tint is not None:
+            if color is None:
+                color = (0.2, 0.2, 0.2)
+            color = _multiply_colors(color, tint)
+        properties["color"] = color
```
huidongc: can move this overwrite into `if tint is not None`.

newton/_src/utils/import_usd.py
eric-heiden: `UsdGeom.Imageable(prim)` is a typed schema wrapper. Its truthiness means "this prim is compatible with the `Imageable` schema," not "this prim is visible." I verified that locally: an `Xform` gives `bool(UsdGeom.Imageable(...)) == True` while `ComputeVisibility()` returns `"inherited"` by default, and a `Material` gives `False`. So `if not imageable` is just a guard to avoid calling `ComputeVisibility()` on non-imageable prims.

newton/_src/utils/import_usd.py
```diff
@@ -635,7 +655,7 @@ def _load_visual_shapes_impl(
             if shape_id >= 0:
                 path_shape_map[path_name] = shape_id
                 path_shape_scale[path_name] = scale
-                if not is_site:
+                if not is_site and visual_shape_cfg_for_prim.is_visible:
                     bodies_with_visual_shapes.add(parent_body_id)
```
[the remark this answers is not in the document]
eric-heiden: Checked the current call sites before changing this. _load_visual_shapes_impl() is only entered from add_body() with a concrete body id, and its recursive calls preserve that same id, so bodies_with_visual_shapes is currently tracking owning bodies only; there is no reachable path here that adds the shared -1 sentinel. The -1 case mentioned in the helper docstring would be a latent hazard if we later reuse this helper for detached static visuals, but the suggested per-prim keying would also change the current body-level collider-visibility policy, so I am leaving this as-is for now.

eric-heiden:
Based on our discussions, we have decided to keep the Newton USD importer focused on only supporting displayColor and diffuseColor parsing, I've reverted this PR to this functionality.

huidongc:
> Based on our discussions, we have decided to keep the Newton USD importer focused on only supporting displayColor and diffuseColor parsing, I've reverted this PR to this functionality.

Sounds good. Thanks for patience!

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 - Add `SolverXPBD.update_contacts()` to populate `contacts.force` with per-contact spatial forces (linear force and torque) derived from XPBD constraint impulses
 - Raise process priority automatically in `--benchmark` mode for more stable measurements; add `--realtime` for maximum priority.
+- Add `ModelBuilder.default_shape_color` fallback so shapes without an explicit or imported color can use a user-configured default instead of the per-shape palette
```
huidongc: This change can be reverted too.

newton/_src/utils/import_usd.py
andrewkaufman: right but you then are returning true from the function, so the question `_is_effectively_visible(material)` is yielding `true` when it should yield `false` (non imagable prims are never visible)

CHANGELOG.md
eric-heiden: Done

newton/_src/utils/import_usd.py
eric-heiden: Makes sense, I've changed it now to return False, the exercised paths are not affected by this anyway.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Add `SolverXPBD.update_contacts()` to populate `contacts.force` with per-contact spatial forces (linear force and torque) derived from XPBD constraint impulses
 - Raise process priority automatically in `--benchmark` mode for more stable measurements; add `--realtime` for maximum priority.
+- Import per-shape authored color from USD stages into `ModelBuilder.shape_color`
```
huidongc, line 9: please also add doc that visibility on USD prims is honored.

huidongc:
I don't have approve permission to this repo. The change looks good to me.

CHANGELOG.md
eric-heiden: Done

PR #2353, opened 2026-04-08
Squash gh-pages to orphan commit on each deploy

### LINKED ISSUE #2352: gh-pages history bloats clone size (~418 MB)
## Problem

`git clone` of this repo transfers ~418 MB due to accumulated objects on the `gh-pages` branch. The branch has **433 commits**, each carrying a fresh Sphinx build (~11 MB `environment.pickle` alone). These objects are included in every full clone even though they're only on `gh-pages`.

### Top offenders by cumulative size across all revisions

| File path | Revisions | Cumulative size |
|---|---|---|
| `latest/.doctrees/environment.pickle` | 429 | ~4.1 GB |
| `latest/searchindex.js` | 429 | ~143 MB |
| `latest/.doctrees/…/newton.ModelBuilder.doctree` | 49 | ~86 MB |
| `latest/_static/recordings/00_introduction.viser` | 348 | ~73 MB |

## Root cause

Both `docs-dev.yml` and `docs-release.yml` append a normal commit to `gh-pages` on every deploy. Over 433 deploys, this accumulates gigabytes of near-duplicate binary/generated content that git can't delta-compress well.

shi-eric:
## Description

The `gh-pages` branch has accumulated 433 commits of Sphinx build artifacts (~4 GB cumulative), bloating clone size to ~418 MB. This changes both doc deploy workflows to force-push a single orphan commit instead of appending history.

Changes to `docs-dev.yml` and `docs-release.yml`:
- After staging, `git checkout --orphan` creates a parentless commit with the full deployed site
- `git push --force` replaces gh-pages with that single commit each time
- `git fetch --depth=1` for gh-pages (history no longer needed)
- `fetch-depth: 1` on initial checkout (main's full history not needed for doc builds)

After the first run, GitHub's server-side GC will prune the dereferenced objects and new clones will be much smaller.

Closes #2352

## Checklist

## Test plan

CI workflow change — verified by reviewing git behavior of the orphan approach:
- `git checkout --orphan` preserves the full index from the previous HEAD, so all existing versioned docs carry through
- First-ever deploy path (no gh-pages branch) still works via the existing `git checkout --orphan gh-pages` fallback
- `set -e` catches any git command failures
- Concurrency group `docs-deploy` prevents race conditions between dev and release deploys

## Summary by CodeRabbit

* **Chores**
  * Optimized documentation deployment workflow for faster, more reliable site updates.
  * Deployment now always publishes the latest build by using a fresh deployment commit and force-push, reducing stale-site issues and simplifying release behavior.
  * Improved error handling during deploy checks to avoid accidental data loss and fail safely on unexpected repository states.

### DIFF
--- a/.github/workflows/docs-dev.yml
+++ b/.github/workflows/docs-dev.yml
@@ -57,13 +57,22 @@ jobs:
           cp docs/_static/gh-pages-404.html /tmp/gh-pages-404.html
 
           # Switch to gh-pages branch (check existence first to avoid masking other fetch errors)
-          if git ls-remote --exit-code --heads origin gh-pages > /dev/null 2>&1; then
+          # Distinguish "branch not found" (exit 2) from fatal errors (exit 128)
+          # to prevent a transient failure from creating an orphan that overwrites
+          # all existing versioned docs on the force push below.
+          ls_remote_rc=0
+          git ls-remote --exit-code --heads origin gh-pages > /dev/null 2>&1 || ls_remote_rc=$?
+
+          if [ "$ls_remote_rc" -eq 0 ]; then
             git fetch --depth=1 origin gh-pages:gh-pages
             git checkout gh-pages
-          else
+          elif [ "$ls_remote_rc" -eq 2 ]; then
             echo "Creating new gh-pages branch"
             git checkout --orphan gh-pages
             git rm -rf . || true
+          else
+            echo "::error::git ls-remote failed with exit code $ls_remote_rc — aborting to prevent data loss"
+            exit 1
           fi
 
           # Remove old /latest/ and replace with new build
--- a/.github/workflows/docs-dev.yml
+++ b/.github/workflows/docs-dev.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # v6.0.1
         with:
-          fetch-depth: 0  # Need full history for gh-pages branch
+          fetch-depth: 1
 
       - name: Install uv
         uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78  # v7.6.0
@@ -58,7 +58,7 @@ jobs:
 
           # Switch to gh-pages branch (check existence first to avoid masking other fetch errors)
           if git ls-remote --exit-code --heads origin gh-pages > /dev/null 2>&1; then
-            git fetch origin gh-pages:gh-pages
+            git fetch --depth=1 origin gh-pages:gh-pages
             git checkout gh-pages
           else
             echo "Creating new gh-pages branch"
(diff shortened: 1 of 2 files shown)


.github/workflows/docs-release.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # v6.0.1
         with:
-          fetch-depth: 0  # Need full history for gh-pages branch
+          fetch-depth: 1
```
adenzler-nvidia, line 33: Might `fetch-depth: 1` on a tag-triggered run break line 115?

```bash
git show origin/main:scripts/ci/update_docs_switcher.py > /tmp/update_docs_switcher.py
```

With `fetch-depth: 0`, `actions/checkout` fetches all refs including `origin/main`. With `fetch-depth: 1` on a tag push, I think only the tag commit is fetched, so `origin/main` wouldn't exist locally and `git show` would fail. Not 100% sure though — could you double-check?

If it is an issue, adding `git fetch --depth=1 origin main` before that line should be enough to fix it without pulling full history.

adenzler-nvidia (commented):
One question about the `fetch-depth` change in the release workflow.

.github/workflows/docs-release.yml
shi-eric: Good catch! Verified in a /tmp experiment — shallow tag checkout indeed doesn't have `origin/main`, so `git show` fails. Added `git fetch --depth=1 origin main` before that line.

.github/workflows/docs-dev.yml
```diff
@@ -58,7 +58,7 @@ jobs:
 
           # Switch to gh-pages branch (check existence first to avoid masking other fetch errors)
           if git ls-remote --exit-code --heads origin gh-pages > /dev/null 2>&1; then
```
adenzler-nvidia: Nit / follow-up consideration: this `if` conflates "branch not found" (exit 2) with "fatal error" (exit 128 — network timeout, auth failure, etc.). With the new `--force` push, a transient `ls-remote` failure would create a fresh orphan and force-push it, silently dropping all versioned release docs. See review body for a suggested fix.

adenzler-nvidia (dismissed):
LGTM — the orphan approach is correct and the prior feedback about `origin/main` on tag-triggered runs is addressed.

One thing to consider as a follow-up: the `git ls-remote ... > /dev/null 2>&1` check treats "branch not found" (exit 2) and "fatal error" (exit 128, e.g. network failure) identically. With the old non-force push this was tolerable, but with `--force` a transient `ls-remote` failure would route to the `else` branch and force-push a fresh orphan containing only the newly built docs — silently dropping all versioned release content. Not a blocker for this PR, but worth hardening in a follow-up by capturing the exit code and failing on anything other than 0 or 2:

```bash
ls_remote_rc=0
git ls-remote --exit-code --heads origin gh-pages > /dev/null 2>&1 || ls_remote_rc=$?

if [ "$ls_remote_rc" -eq 0 ]; then
  git fetch --depth=1 origin gh-pages:gh-pages
  git checkout gh-pages
elif [ "$ls_remote_rc" -eq 2 ]; then
  echo "Creating new gh-pages branch"
  git checkout --orphan gh-pages
  git rm -rf . || true
else
  echo "::error::git ls-remote failed with exit code $ls_remote_rc — aborting to prevent data loss"
  exit 1
fi
```

.github/workflows/docs-release.yml
shi-eric: Applied this to both `docs-dev.yml` and `docs-release.yml` — the dev workflow had the same pattern. Now captures the exit code and aborts on anything other than 0 or 2.

.github/workflows/docs-dev.yml
shi-eric: Good call — applied the exit-code check to both workflows. Now aborts on anything other than 0 (branch exists) or 2 (not found).

adenzler-nvidia (approved):

PR #2405, opened 2026-04-10
Expose xpbd contact force

nvtw:
## Description

Connect XPBD to the contact force reporting API. Requested by Disney to simplify solver comparisons.

Implements `SolverXPBD.update_contacts()` so that `contacts.force` is populated after `step()`, matching the existing pattern used by `SolverMuJoCo`. Per-contact linear impulse is accumulated across XPBD iterations inside `solve_body_contact_positions` and converted to force (impulse / dt) in `update_contacts()`. Only the linear component is written; XPBD point contacts carry no intrinsic torque.

## Checklist

## Test plan

```
```

New tests (all run on CPU + CUDA):
- `test_xpbd_contact_force_sphere_on_plane` — sphere resting on ground, asserts Fz ≈ -mg (5%)
- `test_xpbd_contact_force_heavy_sphere_on_plane` — heavier sphere, same assertion
- `test_xpbd_contact_force_zero_when_no_contact` — free-fall sphere, no contacts, force is zero
- `test_xpbd_contact_force_zero_when_not_touching` — large gap creates contact pair but no penetration, force is zero
- `test_xpbd_update_contacts_requires_force_attribute` — ValueError when `contacts.force` not allocated

## New feature / API change

```python
import newton

model = newton.ModelBuilder()
model.add_ground_plane()
body = model.add_body(xform=wp.transform([0, 0, 0.25], wp.quat_identity()))
model.add_shape_sphere(body=body, radius=0.25)
model = model.finalize()
model.request_contact_attributes("force")

solver = newton.solvers.SolverXPBD(model)
state_in, state_out = model.state(), model.state()
contacts = model.contacts()

model.collide(state_in, contacts)
solver.step(state_in, state_out, model.control(), contacts, 1.0 / 60.0)
solver.update_contacts(contacts, state_out)

# contacts.force now contains per-contact spatial vectors (linear force + zero torque)
print(contacts.force.numpy()[:int(contacts.rigid_contact_count.numpy()[0])])
```

## Summary by CodeRabbit

* **New Features**
  * XPBD solver can now produce per-contact spatial forces (including torque) and populate contact force storage via a new update workflow; the update validates that force buffers are allocated.

* **Tests**
  * Added comprehensive tests covering contact-force computations and edge cases (various contact scenarios, no-contact/not-touching, and update validation).

* **Documentation**
  * Clarified Contacts.force behavior and noted XPBD per-contact forces are approximate when weighting is enabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Added
 
-- Add `SolverXPBD.update_contacts()` to populate `contacts.force` with per-contact linear forces derived from XPBD constraint impulses
+- Add `SolverXPBD.update_contacts()` to populate `contacts.force` with per-contact spatial forces (linear force and torque) derived from XPBD constraint impulses
 - Add repeatable `--warp-config KEY=VALUE` CLI option for overriding `warp.config` attributes when running examples
 - Add 3D texture-based SDF, replacing NanoVDB volumes in the mesh-mesh collision pipeline for improved performance and CPU compatibility.
 - Parse URDF joint `limit effort="..."` values and propagate them to imported revolute and prismatic joint `effort_limit` settings
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
+- Add `SolverXPBD.update_contacts()` to populate `contacts.force` with per-contact linear forces derived from XPBD constraint impulses
 - Add repeatable `--warp-config KEY=VALUE` CLI option for overriding `warp.config` attributes when running examples
 - Add 3D texture-based SDF, replacing NanoVDB volumes in the mesh-mesh collision pipeline for improved performance and CPU compatibility.
 - Parse URDF joint `limit effort="..."` values and propagate them to imported revolute and prismatic joint `effort_limit` settings
--- a/newton/_src/solvers/xpbd/kernels.py
+++ b/newton/_src/solvers/xpbd/kernels.py
@@ -2078,6 +2078,7 @@ def solve_body_contact_positions(
     # outputs
     deltas: wp.array[wp.spatial_vector],
     contact_inv_weight: wp.array[float],
+    contact_impulse: wp.array[wp.spatial_vector],
 ):
     tid = wp.tid()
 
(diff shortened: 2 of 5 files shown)


vastsoun (dismissed):
LGTM, thanks a lot @nvtw

newton/_src/solvers/xpbd/kernels.py
```diff
@@ -2285,6 +2286,38 @@ def solve_body_contact_positions(
     if body_b >= 0:
         wp.atomic_add(deltas, body_b, wp.spatial_vector(lin_delta_b, ang_delta_b))
 
+    if contact_impulse:
+        wp.atomic_add(contact_impulse, tid, wp.spatial_vector(lin_delta_a, ang_delta_a))
```
eric-heiden, line 2290: 🔴 **Contact force is inflated by N× when `rigid_contact_con_weighting` is enabled**

This line stores the raw `lin_delta_a` impulse, but `apply_body_deltas()` later divides the combined correction by `N` (the number of contacts on the body) when `rigid_contact_con_weighting` is active (the default). Because the per-contact impulse is captured *before* that `1/N` scaling, summing `contacts.force` over all contacts on a body overstates the true force by a factor of `N`.

For example, a box resting on a ground plane with 4 contact points will report a total contact force ~4× its weight instead of ~1×.

A couple of options to fix this:
- Divide each per-contact impulse by the body's contact count before storing it here.
- Defer the impulse accumulation into `apply_body_deltas()`, where the `1/N` weight is already known.

Either way the reported forces should match the actual correction applied to the body.

newton/_src/solvers/xpbd/solver_xpbd.py
```diff
@@ -675,3 +686,48 @@ def step(self, state_in: State, state_out: State, control: Control, contacts: Co
...
+            raise ValueError("No contact impulse data available. Call step() before update_contacts().")
+        if contacts.rigid_contact_max != self._contact_impulse_capacity:
+            raise ValueError(
+                f"Contacts capacity mismatch: update_contacts() received rigid_contact_max="
+                f"{contacts.rigid_contact_max}, but step() used {self._contact_impulse_capacity}. "
+                f"Pass the same Contacts instance to both step() and update_contacts()."
+            )
+
+        wp.launch(
+            kernel=convert_contact_impulse_to_force,
```
eric-heiden, line 770: 🟡 **Soft-contact portion of `contacts.force` is never cleared**

`contacts.force` is allocated for `rigid_contact_max + soft_contact_max`, but this kernel launch uses `dim=contacts.rigid_contact_max`, so only the rigid-contact slice is written. `Contacts.clear()` does not zero the `force` array either, so in scenes with soft contacts the tail of the buffer retains stale values from previous steps (or whatever happened to be in memory).

Consider zeroing the full `contacts.force` buffer at the top of `update_contacts()`, or at minimum the `[rigid_contact_max:]` tail. If soft-contact force reporting is planned for a later PR, it would still be good to zero defensively now so users don't accidentally read garbage values.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Added
 
+- Add `SolverXPBD.update_contacts()` to populate `contacts.force` with per-contact linear forces derived from XPBD constraint impulses
```
eric-heiden: ⚪ **CHANGELOG says "linear forces" but the API writes spatial forces (force + torque)**

The implementation writes 6-component spatial vectors containing both linear force and torque (including torsional and rolling friction). The method docstring describes this correctly, but this changelog entry will mislead users scanning release notes.

```suggestion
- Add `SolverXPBD.update_contacts()` to populate `contacts.force` with per-contact spatial forces (linear force and torque) derived from XPBD constraint impulses
```

newton/tests/test_solver_xpbd.py
```diff
@@ -466,6 +466,210 @@ def test_articulation_contact_drift(test, device):
     )
 
 
+def test_xpbd_contact_force_sphere_on_plane(test, device):
```
eric-heiden: ⚪ **Consider parametrizing the near-identical sphere-on-plane tests**

`test_xpbd_contact_force_sphere_on_plane` and `test_xpbd_contact_force_heavy_sphere_on_plane` share ~95% of their code and only differ in `radius` (0.25 vs 0.5) and `density` (1000 vs 2000). This makes it easy to update one and forget the other.

A `@pytest.mark.parametrize` over `(radius, density)` pairs (or a small helper) would remove the duplication and make it clearer what each configuration is actually testing.

Not a blocker, just a maintenance improvement.

eric-heiden (approved):
Thanks, LGTM!

PR #2409, opened 2026-04-10
Opt-in deterministic contacts

nvtw:
## Description

Adds an option to run contact detection deterministically to get bit-identical contacts when the contact detection is run multiple times on the same input.

Two mechanisms are introduced:
- **Fingerprint tiebreaking** — each contact carries a geometry-derived fingerprint (triangle/edge index) that is encoded into the `atomic_max` contact-reduction value, so the reduction winner is independent of GPU thread scheduling.
- **Radix sort** — after the narrow phase, all contact arrays are reordered by a 64-bit key encoding `(shape_a, shape_b, sub_key)` via a radix sort + gather pass, producing a canonical contact order.

Key changes:
- `ContactData` gains a `sort_sub_key` field propagated through all collision paths (convex, SDF, mesh-plane, mesh-triangle).
- `contact_reduction_global.py` introduces dual packing modes (fast / deterministic) for the hashtable `atomic_max` values, switchable via `enable_deterministic_contact_packing()`.
- New `ContactSorter` class (`contact_sort.py`) performs allocation-free, CUDA-graph-capturable radix sort + gather over the contact buffer.
- `CollisionPipeline` and `NarrowPhase` accept a `deterministic=True` flag that wires everything together.

Hydroelastic contacts are currently not yet deterministic. That will come in a future MR.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k TestFloatFlipPrecision
uv run --extra dev -m newton.tests -k test_contact_reduction_global
```

## New feature / API change

```python
import newton

# ... build scene ...
model = builder.finalize()

pipeline = newton.CollisionPipeline(model, deterministic=True)

state = model.state()
# contacts are now in a canonical, reproducible order
```

## Summary by CodeRabbit

* **New Features**
  * Added a global "deterministic" mode to produce reproducible contact ordering across runs (fingerprint tie‑breaking + post‑narrow‑phase radix sort and gather).
  * Exposed stable per-contact sort keys and a contact sorter to reorder contact arrays deterministically.

* **Tests**
  * Added comprehensive unit and end‑to‑end tests validating deterministic packing, sort key ordering, and repeatable reduction results.

* **Documentation**
  * Added “Deterministic Contact Ordering” guide describing usage, overhead, and current hydroelastic limitation.


newton/_src/geometry/contact_reduction_global.py
```diff
@@ -242,31 +242,188 @@ def make_contact_key(shape_a: int, shape_b: int, bin_id: int) -> wp.uint64:
...
+    """Switch to the deterministic contact-value packing.
+
+    Must be called **before** the first kernel launch.  Swaps the
+    module-level ``make_contact_value``, ``make_preprune_probe``, and
+    ``unpack_contact_id`` to the deterministic implementations.
+    """
+    global make_contact_value, make_preprune_probe, unpack_contact_id  # noqa: PLW0603
+    make_contact_value = _make_contact_value_det
+    make_preprune_probe = _make_preprune_probe_det
+    unpack_contact_id = _unpack_contact_id_det
```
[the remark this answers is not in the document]
nvtw: For now, this is by design. Pragmatic solution to keep the performance high when no determinism is needed.

newton/_src/geometry/contact_reduction_global.py
```diff
@@ -242,31 +242,191 @@ def make_contact_key(shape_a: int, shape_b: int, bin_id: int) -> wp.uint64:
...
+# These module-level names are what all callers import.  They start as the
+# fast variants and are swapped to the deterministic variants by
+# ``enable_deterministic_contact_packing()`` before the first kernel launch.
+
+make_contact_value = _make_contact_value_fast
+make_preprune_probe = _make_preprune_probe_fast
+unpack_contact_id = _unpack_contact_id_fast
+
+
+def enable_deterministic_contact_packing() -> None:
```
adenzler-nvidia: **[critical]** Process-global mutable state: `enable_deterministic_contact_packing()` mutates module-level globals shared by all `GlobalContactReducer` instances. Once called, *every* reducer — including non-deterministic ones created afterward — silently switches to 22-bit scores and 20-bit contact IDs.

Consequences:
1. A non-deterministic `GlobalContactReducer` created **after** a deterministic one silently loses score precision and is limited to ~1M contacts with no warning.
2. Python `from ... import` bindings (e.g. in `contact_reduction_hydroelastic.py`) are **not** updated by the module-level swap, creating a fragile divergence between hydroelastic and global packing.

Consider making the packing mode explicit per-reducer (via `GlobalContactReducerData` field or `wp.static()` flag) rather than relying on process-global state.

newton/_src/geometry/contact_data.py
```diff
@@ -52,6 +53,23 @@ class ContactData:
...
+
+    Layout (bit 63 kept zero so int64 order matches uint64 order):
+        [62:43] shape_a  (20 bits)
+        [42:23] shape_b  (20 bits)
+        [22:0]  sort_sub_key (23 bits — encodes edge/triangle index + mesh-side bit)
+    """
+    return (
+        ((wp.int64(shape_a) & wp.int64(0xFFFFF)) << wp.int64(43))
+        | ((wp.int64(shape_b) & wp.int64(0xFFFFF)) << wp.int64(23))
+        | (wp.int64(sort_sub_key) & wp.int64(0x7FFFFF))
```
adenzler-nvidia, line 82: **[critical]** Silent bit-overflow: `shape_a`/`shape_b` are masked to 20 bits and `sort_sub_key` to 23 bits, but there is no validation or warning when values exceed these widths. For `sort_sub_key` specifically, SDF contacts encode `(edge_idx << 2) | (mode << 1)` and mesh-triangle contacts encode `(tri_idx << 1) | 1`, so a mesh with >500K triangles will silently overflow the fingerprint field (22 bits in `FINGERPRINT_MASK`), causing distinct contacts to share fingerprints and making `atomic_max` tiebreaking non-deterministic — exactly what this feature is meant to prevent.

Suggestion: either document the mesh-size limits clearly or add a runtime `wp.expect_*` / warning when indices exceed their bit allocations.

adenzler-nvidia (commented):
## Review Summary

Well-designed feature — the dual-mechanism approach (fingerprint tiebreaking + radix sort) is solid and the ASCII art bit-layout diagrams are excellent. The `sort_sub_key` propagation covers all collision paths consistently, and the RST documentation is clear.

Main concerns are around the process-global mutable state in `enable_deterministic_contact_packing()`, silent bit-overflow for large meshes, and test coverage gaps. Details in inline comments.

newton/_src/geometry/contact_reduction_global.py
```diff
@@ -376,6 +541,13 @@ class GlobalContactReducerData:
     ht_capacity: int
     ht_values_per_key: int
 
+    # When non-zero, skip the speculative pre-prune in
```
adenzler-nvidia: **[critical]** This comment says "skip the speculative pre-prune" but the code does **not** skip the pre-prune. It *replaces* the non-deterministic floor probe with a deterministic ceiling probe (`make_preprune_probe`). The pre-prune logic still runs — the inline comment at the call site (~L1098) correctly describes this behavior, contradicting this field comment.

Suggested wording:
```python
# When non-zero, replace the speculative pre-prune probe with a
# deterministic variant (make_preprune_probe) so that the prune
# decision depends only on score and fingerprint, never on the
# non-deterministic contact_id.
```

newton/_src/geometry/contact_reduction_global.py
```diff
@@ -519,17 +692,31 @@ def __init__(
             store_hydroelastic_data: If True, allocate arrays for contact_area and entry_k_eff
             store_moment_data: If True, allocate moment accumulator arrays for friction
                 moment matching. Only needed when ``moment_matching=True``.
+            deterministic: If True, disable the speculative pre-prune optimization
```
adenzler-nvidia: **[critical]** Same inaccuracy as the struct field comment: this says "disable the speculative pre-prune optimization" but the pre-prune is not disabled — it uses a deterministic probe.

Suggested: `deterministic: If True, use deterministic fingerprint-based tiebreaking in contact reduction and replace the pre-prune probe with a deterministic variant.`

newton/_src/geometry/narrow_phase.py
```diff
@@ -1430,13 +1445,17 @@ def __init__(
...
+                independent of GPU thread scheduling.  Adds a radix sort + gather
+                pass.  Hydroelastic contacts are not yet covered.
         """
         self.max_candidate_pairs = max_candidate_pairs
         self.max_triangle_pairs = max_triangle_pairs
         self.device = device
         self.reduce_contacts = reduce_contacts
         self.has_meshes = has_meshes
         self.has_heightfields = has_heightfields
+        self.deterministic = deterministic
```
adenzler-nvidia, line 1471: **[important]** Inconsistent overflow handling: `NarrowPhase.launch()` silently reallocates the sort buffer and `ContactSorter` when capacity is exceeded (L2060), but `CollisionPipeline.collide()` raises `ValueError` in the same scenario. The silent reallocation here also breaks the "CUDA-graph-capturable" claim since GPU memory allocation during graph capture is not allowed.

Suggestion: unify to either both raise or both resize, and document graph-capture implications.

newton/_src/geometry/contact_reduction_global.py
```diff
@@ -242,31 +242,191 @@ def make_contact_key(shape_a: int, shape_b: int, bin_id: int) -> wp.uint64:
...
+# ---------------------------------------------------------------------------
+
+# 22-bit fingerprint is wide enough to distinguish any two contacts that share
+# the same truncated score within a single reduction slot.  The remaining 20
+# bits for contact_id support up to 1,048,575 buffered contacts.
+FINGERPRINT_BITS = wp.constant(wp.uint64(22))
+CONTACT_ID_BITS = wp.constant(wp.uint64(20))
+CONTACT_ID_MASK = wp.constant(wp.uint64((1 << 20) - 1))
+FINGERPRINT_MASK = wp.constant(wp.uint64((1 << 22) - 1))
+SCORE_SHIFT = 10
```
adenzler-nvidia, line 271: **[nit]** `SCORE_SHIFT` is a plain Python `int` while all sibling constants (`FINGERPRINT_BITS`, `CONTACT_ID_BITS`, etc.) use `wp.constant(wp.uint64(...))`. Consider using `wp.constant` for consistency or adding a comment explaining why it differs (e.g., because `wp.static()` needs a Python int).

newton/tests/test_contact_reduction_global.py
```diff
@@ -854,6 +869,165 @@ def store_one_dominated_contact_kernel(
...
+    tid = wp.tid()
+    f = values_in[tid]
+    flipped = float_flip(f)
+    truncated = (flipped >> wp.uint32(score_shift)) << wp.uint32(score_shift)
+    reconstructed = ifloat_flip(truncated)
+    values_out[tid] = reconstructed
+    truncated_out[tid] = truncated
+
+
+class TestFloatFlipPrecision(unittest.TestCase):
```
adenzler-nvidia, line 906: **[important]** The new `float_flip` roundtrip tests are well-written, but the overall test coverage has significant gaps:

1. **`ContactSorter`** (240 LOC, entirely new file) — zero test coverage for radix sort + gather, sentinel fill, or edge cases (0/1/max contacts).
2. **End-to-end deterministic test** — no test that same input → bit-identical contacts across multiple runs, which is the entire point of this feature.
3. **`make_contact_sort_key`** — no test for bit layout correctness, lexicographic ordering, or overflow masking.
4. **Deterministic packing functions** — `_make_contact_value_det`, `_unpack_contact_id_det`, `_make_preprune_probe_det` are untested.
5. **`sort_sub_key` propagation** — 5 collision paths assign sub_keys with different shift patterns, none tested.

At minimum, an end-to-end deterministic test (create pipeline with `deterministic=True`, collide the same scene N times, assert byte-identical contacts) would provide a meaningful safety net.

docs/concepts/collisions.rst
```diff
@@ -1629,6 +1629,37 @@ Custom collision properties can be authored in USD:
...
+
+1. **Fingerprint tiebreaking** — each contact carries a geometry-derived
+   fingerprint (triangle/edge index) that is used as a deterministic tiebreaker
+   in the ``atomic_max`` contact reduction, so the reduction winner is
+   independent of thread scheduling.
+2. **Radix sort** — after the narrow phase, all contact arrays are reordered by
+   a 64-bit key encoding ``(shape_a, shape_b, sub_key)`` via a radix sort +
+   gather pass.
+
+The overhead is small (one radix sort + gather over the contact buffer per
```
adenzler-nvidia: **[nit]** "The overhead is small (one radix sort + gather)" understates the full cost — the deterministic mode also adds fingerprint storage per contact and modified packing in the reduction. Consider: "The overhead is small: fingerprint storage per contact, modified packing in the reduction, and one radix sort + gather pass per frame."

eric-heiden:
🟡 **Prebuilt `NarrowPhase` bypasses deterministic reduction**

When a caller passes a prebuilt `narrow_phase` to `CollisionPipeline`, the constructor keeps it as-is. The `deterministic` flag only controls the final `ContactSorter` allocation later in the file; it is never propagated to, or validated against, the supplied `NarrowPhase`.

This means a non-deterministic narrow phase still uses schedule-dependent `GlobalContactReducer` behavior during mesh/heightfield contact reduction (`narrow_phase.py:1554-1561`). The post-sort cannot recover determinism once the wrong winner has already been chosen upstream.

Suggested fix: validate that the supplied `NarrowPhase.deterministic` matches `CollisionPipeline.deterministic`, and raise a `ValueError` when they conflict:

```python
if narrow_phase is not None and deterministic and not narrow_phase.deterministic:
    raise ValueError(
        "CollisionPipeline(deterministic=True) requires a deterministic "
        "NarrowPhase. Either omit narrow_phase or construct it with "
        "deterministic=True."
    )
```

newton/_src/geometry/contact_reduction.py
```diff
@@ -109,6 +109,15 @@
 def float_flip(f: float) -> wp.uint32: ...
 
 
+# Inverse of float_flip: convert an order-preserving uint32 back to float.
+@wp.func_native("""
+uint32_t mask = ((u >> 31) - 1) | 0x80000000;
+uint32_t i = u ^ mask;
+return reinterpret_cast<float&>(i);
+""")
+def ifloat_flip(u: wp.uint32) -> float: ...
```
eric-heiden: ⚪ `ifloat_flip` has no production callers. It is only imported in `test_contact_reduction_global.py` for roundtrip validation.

Consider moving it into the test file (or a shared test-utilities module) so this production module does not carry an extra symbol that only tests use.

docs/concepts/collisions.rst
```diff
@@ -1629,6 +1629,38 @@ Custom collision properties can be authored in USD:
...
 See :doc:`custom_attributes` and :doc:`usd_parsing` for details.
 
+.. _Deterministic Contacts:
+
+Deterministic Contact Ordering
+------------------------------
+
+GPU thread scheduling is non-deterministic, so the order in which contacts are
+written to the output buffer can vary between runs.  Pass ``deterministic=True``
+to :class:`~CollisionPipeline` (or :class:`~geometry.NarrowPhase`) to guarantee
```
eric-heiden: ⚪ These Sphinx cross-references will produce broken links. `` :class:`~CollisionPipeline` `` and `` :class:`~geometry.NarrowPhase` `` lack the `newton.` prefix, and without a `.. py:module::` directive in scope Sphinx resolves from the root namespace.

```suggestion
to :class:`~newton.CollisionPipeline` (or :class:`~newton.geometry.NarrowPhase`) to guarantee
```

newton/_src/geometry/contact_reduction_global.py
nvtw: That's on purpose because SCORE_SHIFT is used in unit tests where wp.constant causes trouble

newton/_src/geometry/narrow_phase.py
```diff
@@ -2170,7 +2199,15 @@ def launch(
         # Clear external contact count (internal counters are cleared in launch_custom_write)
         contact_count.zero_()
 
+        # Ensure sort-key buffer and sorter match the contact output capacity
+        if self.deterministic and self._sort_key_array.shape[0] < contact_max:
+            dev = self._sort_key_array.device
+            self._sort_key_array = wp.zeros(contact_max, dtype=wp.int64, device=dev)
+            self._contact_sorter = ContactSorter(contact_max, device=dev)
```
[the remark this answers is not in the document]
nvtw: Should be addressed with the latest commit

newton/_src/sim/collide.py
```diff
@@ -908,6 +931,13 @@ def collide(
         writer_data.out_stiffness = contacts.rigid_contact_stiffness
         writer_data.out_damping = contacts.rigid_contact_damping
         writer_data.out_friction = contacts.rigid_contact_friction
+        if self.deterministic and contacts.rigid_contact_max > self._sort_key_array.shape[0]:
+            raise ValueError(
+                f"Contacts buffer capacity ({contacts.rigid_contact_max}) exceeds the "
+                f"deterministic sort buffer size ({self._sort_key_array.shape[0]}). "
+                f"Create the CollisionPipeline with a larger rigid_contact_max."
+            )
+        writer_data.out_sort_key = self._sort_key_array
```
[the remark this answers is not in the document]
nvtw: Should be addressed with the latest commit

nvtw:
@adenzler-nvidia @eric-heiden your comments are addressed.

newton/_src/geometry/contact_reduction_global.py
```diff
@@ -242,31 +242,206 @@ def make_contact_key(shape_a: int, shape_b: int, bin_id: int) -> wp.uint64:
...
+# These module-level names are what all callers import.  They start as the
+# fast variants and are swapped to the deterministic variants by
+# ``enable_deterministic_contact_packing()`` before the first kernel launch.
+
+make_contact_value = _make_contact_value_fast
+make_preprune_probe = _make_preprune_probe_fast
+unpack_contact_id = _unpack_contact_id_fast
+
+
+def enable_deterministic_contact_packing() -> None:
```
eric-heiden: 🔴 `enable_deterministic_contact_packing()` mutates module-level globals (`make_contact_value`, `make_preprune_probe`, `unpack_contact_id`) for the entire process, and there is no `disable_deterministic_contact_packing()` to restore the originals.

This causes two concrete problems:

| Scenario | What happens |
|---|---|
| Kernel compiled **before** this toggle is called | Warp caches `@wp.func` references at compile time, so the kernel keeps the fast-packing variant. `deterministic=True` silently has no effect on those kernels. |
| Non-det reducer created **after** the toggle | It inherits `_unpack_contact_id_det`, which masks to 20 bits. Contact IDs above 1,048,575 are silently truncated. The capacity guard does not fire because `deterministic=False` on that reducer skips it. |

The root issue is that packing mode is process-global state rather than a per-reducer property.

Suggested fix: make the packing variant an explicit field on the reducer instance (or on `GlobalContactReducerData`) so each reducer carries its own function references. At minimum, add a restore function and a launch-time assertion that the active packing mode matches what the reducer expects.

newton/_src/geometry/contact_data.py
```diff
@@ -52,6 +53,29 @@ class ContactData:
     contact_stiffness: float
     contact_damping: float
     contact_friction_scale: float
+    sort_sub_key: int
+
+
+@wp.func
+def make_contact_sort_key(shape_a: int, shape_b: int, sort_sub_key: int) -> wp.int64:
```
eric-heiden, line 60: 🟡 The documented bit widths here are the raw field sizes, but they overstate the effective limits for mesh-triangle contacts.

The mesh-triangle contact path computes `sort_sub_key = ((tri_idx << 1) | 1) << 3`, which consumes 4 low bits before the value is stored. That leaves:

| Field | Raw width | Effective `tri_idx` bits | Effective limit |
|---|---|---|---|
| `FINGERPRINT_MASK` (22 bits) | 4,194,303 | 18 bits | **262,144** triangles |
| `sort_sub_key` (23 bits) | 8,388,607 | 19 bits | **524,288** triangles |

The docstring around line 70 claims overflow at "~4M triangles", which is only true for the raw 22-bit width and ignores the upstream shifts. Since `max_triangle_pairs` defaults to 1,000,000, meshes with more than 262k triangles (common in production) will alias in `atomic_max` tiebreaking and silently lose determinism.

Suggested fixes:
- Update the docstring to document the actual effective limits after upstream bit consumption.
- Consider adding a runtime check or warning when `max_triangle_pairs` exceeds the effective fingerprint range.
- Alternatively, widen the fingerprint field to accommodate the full intended range.

newton/_src/geometry/narrow_phase.py
```diff
@@ -1588,6 +1616,13 @@ def __init__(
             )
 
             self.empty_tangent = None
+            self._empty_sort_key = wp.zeros(0, dtype=wp.int64, device=device)
+            if deterministic:
+                self._sort_key_array = wp.zeros(max_candidate_pairs, dtype=wp.int64, device=device)
```
eric-heiden: 🟡 Both `_sort_key_array` and `ContactSorter` are sized to `max_candidate_pairs`, but a single candidate pair can emit multiple contacts (up to 4 for primitive paths, potentially more for convex manifolds). Later in `launch()`, the check at ~L2212 rejects any output buffer where `contact_max != _sort_key_array.shape[0]`.

This means callers who correctly provision a larger contact buffer (e.g. `contact_max = 4 * max_candidate_pairs`) to accommodate multi-contact emission will get a `ValueError` in deterministic mode, even though the same call succeeds in non-deterministic mode.

```suggestion
            if deterministic:
                det_capacity = max(max_candidate_pairs * 4, max_candidate_pairs)
                self._sort_key_array = wp.zeros(det_capacity, dtype=wp.int64, device=device)
                self._contact_sorter = ContactSorter(det_capacity, device=device)
```

Alternatively, accept a separate `contact_max` parameter in the constructor so callers can control the sort buffer size independently from the pair count.

newton/_src/geometry/contact_reduction_global.py
```diff
@@ -519,17 +706,32 @@ def __init__(
             store_hydroelastic_data: If True, allocate arrays for contact_area and entry_k_eff
             store_moment_data: If True, allocate moment accumulator arrays for friction
                 moment matching. Only needed when ``moment_matching=True``.
+            deterministic: If True, use deterministic fingerprint-based tiebreaking
+                in contact reduction and replace the pre-prune probe with a
+                deterministic variant.
         """
+        max_det_contacts = (1 << int(CONTACT_ID_BITS)) - 1
```
eric-heiden: ⚪ Minor off-by-one: `(1 << 20) - 1` gives 1,048,575, but contact IDs are 0-based (assigned via `atomic_add`), so a capacity of 1,048,576 produces IDs in `[0, 1,048,575]`, all of which fit in the 20-bit `CONTACT_ID_MASK`. The `- 1` rejects one valid capacity value.

```suggestion
        max_det_contacts = 1 << int(CONTACT_ID_BITS)
```

eric-heiden (approved):

PR #2414, opened 2026-04-11
Import TetMesh prims through add_usd

eric-heiden:
## Description

Add support for importing `UsdGeom.TetMesh` prims as soft meshes through `ModelBuilder.add_usd()`. When a USD stage contains TetMesh prims, they are now automatically converted to soft meshes via `add_soft_mesh()`, with proper handling of:

- World transforms (position, rotation, scale) including parent Xform chains
- Uniform and non-uniform scale decomposition
- Physics material binding (Young's modulus, Poisson's ratio, density)
- TetMesh data caching for instanced prims

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_add_usd_imports_tetmesh
uv run --extra dev -m newton.tests -k test_add_usd_imports_tetmesh_with_transforms
```

Two new tests verify:
1. A `.usda` file with a TetMesh prim and physics material is imported correctly — particle count, tet/tri indices, material parameters (µ, λ), and total mass are validated.
2. Prim-level and import-level transforms compose correctly onto particle positions.

## New feature / API change

```python
import newton
import warp as wp

# TetMesh prims in the USD are now automatically imported as soft meshes
builder.add_usd("scene_with_tetmesh.usda")

model = builder.finalize()
# model now contains soft-body particles and tet elements
```

## Summary by CodeRabbit

* **New Features**
  * USD TetMesh prims are now imported as soft meshes, honoring per-instance imports and composed transforms (translation/rotation/scale); source mesh custom attributes are preserved and filtered on export.

* **Documentation**
  * Changelog and import documentation updated to reflect TetMesh → soft-body import behavior.

* **Tests**
  * Added tests for TetMesh import, instancing, transform composition, and custom-attribute preservation/filtering.

### DIFF
--- a/newton/_src/utils/import_usd.py
+++ b/newton/_src/utils/import_usd.py
@@ -2425,7 +2425,7 @@ def parse_usd(
             if _is_uniform_scale(soft_mesh_scale):
                 add_soft_mesh_kwargs["scale"] = float(np.array(soft_mesh_scale, dtype=np.float32)[0])
             else:
-                add_soft_mesh_kwargs["vertices"] = tetmesh.vertices * np.array(soft_mesh_scale, dtype=np.float32)
+                add_soft_mesh_kwargs["vertices"] = tetmesh_for_builder.vertices * np.array(soft_mesh_scale, dtype=np.float32)
 
             builder.add_soft_mesh(**add_soft_mesh_kwargs)
 
--- a/newton/_src/utils/import_usd.py
+++ b/newton/_src/utils/import_usd.py
@@ -87,7 +87,7 @@ def parse_usd(
     force_position_velocity_actuation: bool = False,
     override_root_xform: bool = False,
 ) -> dict[str, Any]:
-    """Parses a Universal Scene Description (USD) stage containing UsdPhysics schema definitions for rigid-body articulations and adds the bodies, shapes and joints to the given ModelBuilder.
+    """Parses a Universal Scene Description (USD) stage and adds rigid bodies, soft bodies, shapes, and joints to the given ModelBuilder.
 
     The USD description has to be either a path (file name or URL), or an existing USD stage instance that implements the `Stage <https://openusd.org/dev/api/class_usd_stage.html>`_ interface.
 
@@ -353,6 +353,8 @@ def parse_usd(
     material_props_cache: dict[str, dict[str, Any]] = {}
     # cache for mesh data loaded from USD prims
     mesh_cache: dict[tuple[str, bool, bool], Mesh] = {}
+    # cache for TetMesh data loaded from USD prims
+    tetmesh_cache: dict[str, Any] = {}
 
     physics_scene_prim = None
     physics_dt = None
@@ -442,6 +444,18 @@ def parse_usd(
         if material_props.get("metallic") is not None:
             mesh.metallic = material_props["metallic"]
         return mesh
+
+    def _get_tetmesh_cached(prim: Usd.Prim):
+        """Load and cache TetMesh data to avoid repeated USD extraction."""
+        prim_path = str(prim.GetPath())
+        if prim_path not in tetmesh_cache:
+            tetmesh_cache[prim_path] = usd.get_tetmesh(prim)
+        return tetmesh_cache[prim_path]
+
+    def _is_uniform_scale(scale: wp.vec3) -> bool:
+        """Return whether a decomposed scale vector is effectively uniform."""
+        scale_np = np.array(scale, dtype=np.float32)
+        return bool(np.allclose(scale_np, scale_np[0], rtol=1e-6, atol=1e-6))
 
     def _has_visual_material_properties(material_props: dict[str, Any]) -> bool:
         # Require PBR-like material cues to avoid promoting generic displayColor-only colliders.
@@ -629,7 +643,7 @@ def parse_usd(
                     cfg=visual_shape_cfg,
                     label=path_name,
                 )
-            elif len(type_name) > 0 and type_name != "xform" and verbose:
+            elif len(type_name) > 0 and type_name not in {"xform", "tetmesh"} and verbose:
                 print(f"Warning: Unsupported geometry type {type_name} at {path_name} while loading visual shapes.")
 
             if shape_id >= 0:
@@ -2367,6 +2381,42 @@ def parse_usd(
                 for shape1 in builder.body_shapes[body1]:
                     for shape2 in builder.body_shapes[body2]:
                         builder.add_shape_collision_filter_pair(shape1, shape2)
+
+    root_prim = stage.GetPrimAtPath(root_path)
+    if root_prim and root_prim.IsValid():
+        for prim in Usd.PrimRange(root_prim, Usd.TraverseInstanceProxies()):
+            if not prim.IsA(UsdGeom.TetMesh):
+                continue
+
+            path = str(prim.GetPath())
+            if any(re.match(pattern, path) for pattern in ignore_paths):
+                continue
+
+            if collect_schema_attrs:
+                R.collect_prim_attrs(prim)
+
+            tetmesh = _get_tetmesh_cached(prim)
+            soft_mesh_mat = _get_prim_world_mat(prim, None, incoming_world_xform)
+            soft_mesh_pos, soft_mesh_rot, soft_mesh_scale = wp.transform_decompose(soft_mesh_mat)
+
+            add_soft_mesh_kwargs = {
+                "pos": soft_mesh_pos,
+                "rot": soft_mesh_rot,
+                "scale": 1.0,
+                "vel": wp.vec3(0.0, 0.0, 0.0),
+                "mesh": tetmesh,
+            }
+            if _is_uniform_scale(soft_mesh_scale):
+                add_soft_mesh_kwargs["scale"] = float(np.array(soft_mesh_scale, dtype=np.float32)[0])
+            else:
+                add_soft_mesh_kwargs["vertices"] = tetmesh.vertices * np.array(soft_mesh_scale, dtype=np.float32)
+
+            builder.add_soft_mesh(**add_soft_mesh_kwargs)
+
+            if verbose:
+                print(
+                    f"Added soft mesh {path} with {tetmesh.vertex_count} vertices and {tetmesh.tet_count} tetrahedra."
+                )
 
     # Load Gaussian splat prims that weren't already captured as children of rigid bodies.
     if load_visual_shapes:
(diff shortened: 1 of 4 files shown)


newton/_src/utils/import_usd.py
```diff
@@ -2368,6 +2382,51 @@ def _build_mass_info_from_shape_geometry(
...
+            }
+            if _is_uniform_scale(soft_mesh_scale):
+                add_soft_mesh_kwargs["scale"] = float(np.array(soft_mesh_scale, dtype=np.float32)[0])
+            else:
+                add_soft_mesh_kwargs["vertices"] = tetmesh.vertices * np.array(soft_mesh_scale, dtype=np.float32)
+
+            # Filter custom attributes to only those registered in the builder,
+            # matching the behavior of the rigid-body/joint import paths.
+            if tetmesh.custom_attributes:
+                tetmesh.custom_attributes = {
```
adenzler-nvidia: **Bug: cache mutation.** `_get_tetmesh_cached` returns a shared reference from `tetmesh_cache`. This assignment mutates `custom_attributes` on the cached object. If the same prim path is encountered again (e.g. instanced prims via `TraverseInstanceProxies`), the second import sees the already-filtered `custom_attributes`, silently losing attributes that were removed in the first pass.

Compare with the mesh cache pattern — `_get_mesh_cached` returns shared objects too, but no caller mutates them.

Additionally, this filtering is semantically redundant with `add_soft_mesh` (builder.py L8117-8124), which already validates `mesh.custom_attributes` against `builder.custom_attributes` and raises `ValueError` for unregistered ones. The pre-filtering silently swallows what the downstream API considers a programming error — the two layers disagree on the error contract.

Suggested fix — either remove the filtering entirely (let `add_soft_mesh` enforce the contract with a clear error), or filter into a local variable:
```python
if tetmesh.custom_attributes:
    filtered = {
        k: v for k, v in tetmesh.custom_attributes.items() if k in builder.custom_attributes
    }
    # pass filtered attrs without touching the cached object
```

newton/_src/utils/import_usd.py
```diff
@@ -2368,6 +2382,51 @@ def _build_mass_info_from_shape_geometry(
                     for shape2 in builder.body_shapes[body2]:
                         builder.add_shape_collision_filter_pair(shape1, shape2)
 
+    root_prim = stage.GetPrimAtPath(root_path)
+    if root_prim and root_prim.IsValid():
+        for prim in Usd.PrimRange(root_prim, Usd.TraverseInstanceProxies()):
+            if not prim.IsA(UsdGeom.TetMesh):
+                continue
+
+            path = str(prim.GetPath())
```
adenzler-nvidia, line 2394: **Missing deduplication guard.** The gaussian splat import path (L2386-2387 in the existing code) checks `if gaussian_path in path_shape_map: continue` before adding. This TetMesh path has no equivalent guard, so if a prim path were visited twice (e.g. direct child + instance proxy overlap), the same soft mesh would be added twice with doubled particles and tet elements.

Consider tracking imported paths:
```python
if path in tetmesh_imported:
    continue
tetmesh_imported.add(path)
```

newton/tests/test_import_usd.py
```diff
@@ -7733,6 +7733,63 @@ def test_get_tetmesh_no_material(self):
...
+
+        self.assertEqual(len(builder.particle_q), 4)
+        self.assertEqual(len(builder.tet_indices), 1)
+        self.assertEqual(len(builder.tri_indices), 4)
+        self.assertAlmostEqual(builder.tet_materials[0][0], 300000.0 / (2.0 * 1.3), places=0)
+        self.assertAlmostEqual(builder.tet_materials[0][1], 300000.0 * 0.3 / (1.3 * 0.4), places=0)
+        self.assertAlmostEqual(sum(builder.particle_mass), 40.0 / 6.0, places=5)
+
+    @unittest.skipUnless(USD_AVAILABLE, "Requires usd-core")
+    def test_add_usd_imports_tetmesh_with_transforms(self):
```
adenzler-nvidia, line 7783: **Test coverage gap for rotation and non-uniform scale.** This test only exercises a pure translation transform. The more complex code paths — rotation via `soft_mesh_rot` and especially the non-uniform scale branch (`_is_uniform_scale` returning `False` → pre-scaling vertices at L2414) — are not covered. Consider adding a case with a non-identity rotation and/or non-uniform scale on the Xform prim to lock in that behavior.

newton/_src/utils/import_usd.py
```diff
@@ -353,6 +353,8 @@ class PhysicsMaterial:
     material_props_cache: dict[str, dict[str, Any]] = {}
     # cache for mesh data loaded from USD prims
     mesh_cache: dict[tuple[str, bool, bool], Mesh] = {}
+    # cache for TetMesh data loaded from USD prims
+    tetmesh_cache: dict[str, Any] = {}
```
adenzler-nvidia: Nit: `dict[str, Any]` — the values are `TetMesh` objects (returned by `usd.get_tetmesh`). Other caches in this function use specific types (e.g. `mesh_cache: dict[..., Mesh]`). Consider `dict[str, TetMesh]` for consistency and type safety (with a `TYPE_CHECKING` import if needed).

adenzler-nvidia (commented):
## PR Review: Import TetMesh prims through `add_usd`

Nice addition — the feature follows existing patterns well (caching, transform decomposition, visual shape suppression, `/Prototypes/` filtering). The scope is tight and the happy-path test coverage is solid. A few issues worth addressing below.

### Strengths
- Smart uniform vs. non-uniform scale handling — pre-baking non-uniform scale into vertices since `add_soft_mesh` only supports uniform `scale` is the right approach.
- Good material propagation test — validates the full chain from USD physics material → Lamé parameters → total mass.
- Clean visual shape suppression — adding `"tetmesh"` to the warning exclusion set prevents noisy warnings without over-suppressing.

newton/_src/utils/import_usd.py
eric-heiden: Verified this with an instancing test: a USD stage with two `Usd.TraverseInstanceProxies()` instances of a prototype containing a 4-vertex `TetMesh`. The traversal yields each instance proxy exactly once, and the `/Prototypes/` skip at line 2392 correctly prevents the authored prototype from also being imported. Result is exactly 8 particles (2 × 4 vertices), no duplication.

The broader point about a defensive `path_shape_map`-style guard is still a reasonable hardening measure, but the current traversal + prototype skip does handle the instancing case correctly.

eric-heiden: I didn\u0027t add a separate dedup set here. The current `PrimRange(..., TraverseInstanceProxies())` walk plus the `/Prototypes/` skip already imports one TetMesh per instance proxy, and I added `test_add_usd_imports_instanced_tetmesh_once_per_instance` to keep that behavior covered.

newton/_src/utils/import_usd.py
eric-heiden: Kept `add_usd()`\u0027s current behavior of dropping unregistered TetMesh import attributes so it stays aligned with the rigid/joint import paths, but the filtering now happens on a shallow copy before `add_soft_mesh()`. The cached `TetMesh` is no longer mutated. Added `test_add_usd_does_not_mutate_loaded_tetmesh_custom_attributes` to lock that in.

newton/_src/utils/import_usd.py
eric-heiden: Updated the cache typing to `dict[str, TetMesh]` and annotated `_get_tetmesh_cached()` to return `TetMesh` as well.

newton/tests/test_import_usd.py
eric-heiden: Expanded `test_add_usd_imports_tetmesh_with_transforms` so it now exercises nested Xforms, rotation, non-uniform scale, and a non-identity import quaternion.

adenzler-nvidia (dismissed):
Looks good — prior review items are all properly addressed. The cache mutation fix via `copy.copy`, instancing test, expanded transform coverage, and type annotation update are solid.

One nit for a follow-up: in the non-uniform scale branch, `tetmesh.vertices` should be `tetmesh_for_builder.vertices`:

```python
add_soft_mesh_kwargs["vertices"] = tetmesh.vertices * np.array(soft_mesh_scale, dtype=np.float32)
```

This works today because `copy.copy` is shallow (both reference the same numpy array), but it's fragile — if the copy semantics ever change, this silently passes stale vertices. Using `tetmesh_for_builder.vertices` makes the intent clear and is more robust.

adenzler-nvidia (dismissed):

newton/_src/utils/import_usd.py
```diff
@@ -353,6 +356,8 @@ class PhysicsMaterial:
     material_props_cache: dict[str, dict[str, Any]] = {}
     # cache for mesh data loaded from USD prims
     mesh_cache: dict[tuple[str, bool, bool], Mesh] = {}
+    # cache for TetMesh data loaded from USD prims
+    tetmesh_cache: dict[str, TetMesh] = {}
```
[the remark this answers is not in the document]
eric-heiden: The observation about prototype-based cache keys is technically correct for shared-prototype instancing, but we intentionally kept `prim.GetPath()` here for two reasons:

1. **Consistency with existing caches.** `mesh_cache` (line 392) and `material_props_cache` (line 388) both key by `str(prim.GetPath())` using the same pattern. Switching only `tetmesh_cache` to prototype identity would create an inconsistency across the importer. If this optimization is worth pursuing, it should be done as a separate refactor across all caches.

2. **Correctness over performance for now.** `usd.get_tetmesh(prim)` is called with the instance proxy prim, which resolves attribute values in that proxy's context. Using a prototype-based key assumes all proxies of the same prototype will return identical geometry data, which is true for native instancing but could be surprising if someone later passes overridden attribute values through instance proxies. The current approach is safer.

3. **Negligible cost in practice.** `get_tetmesh` extracts a small vertex/tet-index buffer. For scenes with many instances of the same prototype this duplicates the extraction call, but the actual runtime cost is low, and the existing instancing test (`test_add_usd_imports_instanced_tetmesh_once_per_instance`) confirms correctness with the current key strategy.

Worth revisiting if profiling shows `get_tetmesh` is a bottleneck for heavily instanced scenes, but not needed for this PR.

eric-heiden:
> Looks good — prior review items are all properly addressed. The cache mutation fix via `copy.copy`, instancing test, expanded transform coverage, and type annotation update are solid.
> 
> One nit for a follow-up: in the non-uniform scale branch, `tetmesh.vertices` should be `tetmesh_for_builder.vertices`:
> 
> ```python
> add_soft_mesh_kwargs["vertices"] = tetmesh.vertices * np.array(soft_mesh_scale, dtype=np.float32)
> ```
> 
> This works today because `copy.copy` is shallow (both reference the same numpy array), but it's fragile — if the copy semantics ever change, this silently passes stale vertices. Using `tetmesh_for_builder.vertices` makes the intent clear and is more robust.

Thanks, addressed this now. Can you please reapprove @adenzler-nvidia?

adenzler-nvidia (approved):

PR #394, opened 2025-07-13
Add inverse kinematics module with interactive and benchmark examples

dylanturpin:
## Description
Adds a modular inverse kinematics (IK) system to Newton, inspired by [PyRoki](https://pyroki-toolkit.github.io/).

output

**Core features:**
- Levenberg-Marquardt solver leveraging Warp's tile API for cholesky linear solve
(with adaptive damping, step accept/reject logic)
- Analytic Jacobians for speed (option to use autodiff or mix instead)
- Modular objectives: position, rotation, and joint limit objectives included

**API:**
```
import newton.core.ik as ik

ik_solver = ik.IKSolver(
    model=model,
    joint_q=joint_q, # (n_problems, n_coords) 
    objectives=[
        ik.PositionObjective(link_index=16, target_positions=targets, ...),
        ik.RotationObjective(link_index=16, target_rotations=rotations, ...),
        ik.JointLimitObjective(weight=0.1, ...)
    ],
    jacobian_mode=JacobianMode.ANALYTIC # or MIXED/AUTODIFF
)
ik_solver.solve(iterations=10)

```

**Interactive example:**

- `newton/examples/example_ik_interactive.py` - drag gizmos to control H1 robot end-effectors
- Supports "tied controls" mode where dragging one robot controls all environments
- GUI utilities in newton/examples/gizmo_utils.py

**Benchmark example:**

- `newton/examples/example_ik_benchmark.py --robot [franka/h1]`
- Runs many seeds and selects best
- Generates benchmark tables like below

## Benchmarks
Accuracy stats averaged over 2M samples.

### Franka (Simple Arm): 3x Faster, 99.994% Success Rate
Results (98th percentile errors for successful solves)

<table>
  <thead>
    <tr>
      <th rowspan="2">Solver</th>
      <th rowspan="2">Pos&nbsp;Err&nbsp;(mm)</th>
      <th rowspan="2">Ori&nbsp;Err&nbsp;(rad)</th>
      <th rowspan="2">Success&nbsp;(%)</th>
      <th colspan="5">time&nbsp;(ms)&nbsp;for&nbsp;batch&nbsp;size</th>
    </tr>
    <tr>
      <th>b&nbsp;= 1</th>
      <th>b&nbsp;= 10</th>
      <th>b&nbsp;= 100</th>
      <th>b&nbsp;= 1 k</th>
      <th>b&nbsp;= 2 k</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>cuRobo</td>
[...]

PR #2415, opened 2026-04-11
Add Newton robotics tutorial notebook

### LINKED ISSUE #2173: Add more tutorials, update CI job
* Bring in more tutorial notebooks
* Create weekly CI job that runs on GPU just to evaluate tutorial notebooks and populate the docs
* Remove pandoc dependency and notebook evaluation from regular PR doc building job

eric-heiden:
This PR contributes the course notebook that has been adapted from the GTC hands-on lab, see recording at https://www.nvidia.com/en-us/on-demand/session/gtc26-dlit81700/.

This notebook and ViewerViser setup have been tested on Brev and confirmed to work with the Newton launchable:
https://brev.nvidia.com/launchable/deploy/now?launchableID=env-3CMMBtpCXzqJSEfSj6lZxNmCGaI


docs/tutorials/01_newton_course.ipynb
```diff
@@ -0,0 +1,1603 @@
...
+    "\n",
+    "\n",
+    "def simulate():\n",
+    "    \"\"\"Run one frame of simulation (multiple substeps).\"\"\"\n",
+    "    global state_0, state_1\n",
+    "\n",
+    "    for _ in range(sim_substeps):\n",
+    "        state_0.clear_forces()\n",
+    "        contacts = model.collide(state_0)\n",
+    "        solver.step(state_in=state_0, state_out=state_1, control=control, contacts=contacts, dt=sim_dt)\n",
```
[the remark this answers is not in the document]
eric-heiden: Fixed in ce48327a. The notebook now reuses the preallocated `contacts` buffer with `model.collide(state_0, contacts)`, matching the later simulation loops.

Verified with:
- `uv run python -m json.tool docs/tutorials/01_newton_course.ipynb > $null`
- `uv run --extra dev -m newton.tests -k test_viewer_viser_notebook`

docs/conf.py
```diff
@@ -395,21 +395,22 @@ def _copy_viser_client_into_output_static(*, outdir: Path) -> None:
...
+    # Installed viser package build. Prefer this so the copied docs assets stay in
+    # sync with the serializer version used for notebook playback.
     try:
-        import newton  # noqa: PLC0415
+        import inspect  # noqa: PLC0415
+
+        import viser  # noqa: PLC0415
 
-        src_candidates.append(Path(newton.__file__).resolve().parent / "_src" / "viewer" / "viser" / "static")
+        src_candidates.append(Path(inspect.getfile(viser)).resolve().parent / "client" / "build")
```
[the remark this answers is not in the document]
eric-heiden: Fixed in ce48327a. The docs copy path now checks both the installed `viser/client/build` directory and the `viser/static` fallback before warning.

Verified with:
- `uvx ruff check docs/conf.py newton/_src/viewer/viewer_viser.py newton/tests/test_viewer_viser_notebook.py`
- `uv run --extra dev -m newton.tests -k test_viewer_viser_notebook`

newton/_src/viewer/viewer_viser.py
```diff
@@ -184,13 +184,98 @@ def _call_scene_method(method, **kwargs):
...
-        """Get the URL of the viser server.
+        """Get the browser URL of the viser server.
 
         Returns:
-            str: Local HTTP URL for the running viser server.
+            str: Browser URL for the running viser server.
         """
+        if self.is_jupyter_notebook:
+            return self._build_browser_url(self._port)
         return f"http://localhost:{self._port}"
```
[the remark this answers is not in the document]
eric-heiden: Fixed in ce48327a. `ViewerViser` now initializes `_port` and notebook detection before logging, and the verbose startup message prints `self.url`, so notebook proxy URLs are shown.

Added `test_verbose_startup_log_uses_proxy_url_in_notebook` and verified with `uv run --extra dev -m newton.tests -k test_viewer_viser_notebook`.

newton/_src/viewer/viewer_viser.py
```diff
@@ -147,6 +147,8 @@ def __init__(
         self._meshes = {}
         self._instances = {}
         self._scene_handles = {}  # Track viser scene node handles
+        self._plane_meshes = {}
+        self._plane_handles = {}
```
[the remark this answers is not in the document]
eric-heiden: Fixed in ce48327a. `clear_model()` now removes tracked plane handles and clears the plane mesh metadata, with plane state initialized before the base viewer constructor calls `clear_model()`.

Added `test_clear_model_removes_plane_handles_and_meshes` and verified with `uv run --extra dev -m newton.tests -k test_viewer_viser_notebook`.

newton/_src/viewer/viewer_viser.py
```diff
@@ -494,6 +581,88 @@ def log_mesh(
...
+        if xforms_np is None or len(xforms_np) == 0:
+            return
+
+        xforms_np = np.asarray(xforms_np, dtype=np.float32)
+        positions = xforms_np[:, :3]
+        quats_wxyz = self._quats_xyzw_to_wxyz(xforms_np[:, 3:7])
+
+        width = float(plane_info["width"])
+        length = float(plane_info["length"])
+        grid_points = self._build_plane_grid_points(width, length)
```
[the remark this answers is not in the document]
eric-heiden: Fixed in ce48327a. The plane wireframe path now receives `scales` from `log_instances()` and applies each instance's X/Y scale to the local grid points before logging line segments.

Added `test_plane_instances_apply_per_instance_scales` and verified with `uv run --extra dev -m newton.tests -k test_viewer_viser_notebook`.

.github/workflows/docs-release.yml
```diff
@@ -1,5 +1,15 @@
...
 
+env:
+  AWS_REGION: us-east-2
+  AWS_INSTANCE_TYPE: g7e.2xlarge
+  AWS_VOLUME_SIZE: 92
+  AWS_VOLUME_TYPE: gp3
+  AWS_SECURITY_GROUP_IDS: sg-07807c44e7f2a368a
+  AWS_ROLE_ARN: arn:aws:iam::968945269301:role/newton-physics-newton-ec2-github-runner-role
+  AWS_ROLE_DURATION: 3600
+  HOME: /actions-runner
```
preist-nvidia: Setting `HOME=/actions-runner` at workflow scope also applies to the `deploy` job, which runs on `ubuntu-latest`. That job later calls `uv run --no-project` to update `switcher.json`; on GitHub-hosted runners `/actions-runner` is not the normal writable home, so uv will try to create `/actions-runner/.cache/uv` and fail. Please scope `HOME` to the EC2 runner jobs, override it in `deploy`, or call the script with `python` instead of `uv run`.

docs/tutorials/01_newton_course.ipynb
```diff
@@ -0,0 +1,1650 @@
...
+        "grip_closed = 0.0\n",
+        "\n",
+        "# Each phase: move to a target position with a gripper state, over N frames\n",
+        "sequence = [\n",
+        "    {\"name\": \"approach\", \"pos\": cube_pos + approach_offset, \"grip\": grip_open, \"frames\": 60},\n",
+        "    {\"name\": \"descend\", \"pos\": cube_pos, \"grip\": grip_open, \"frames\": 60},\n",
+        "    {\"name\": \"close\", \"pos\": cube_pos, \"grip\": (grip_open, grip_closed), \"frames\": 60},\n",
+        "    {\"name\": \"lift\", \"pos\": cube_pos + lift_offset, \"grip\": grip_closed, \"frames\": 60},\n",
+        "    {\"name\": \"move\", \"pos\": drop_pos + approach_offset, \"grip\": grip_closed, \"frames\": 60},\n",
+        "    {\"name\": \"drop\", \"pos\": drop_pos, \"grip\": (grip_closed, grip_open), \"frames\": 60},\n",
```
preist-nvidia: P2: Split the placement descent from the gripper opening here. In the generated replay the cube is released while the end-effector is still moving/settling down toward `drop_pos`, so it drops early instead of being deliberately placed. A cleaner sequence would keep `grip_closed` while moving to the drop pose, optionally hold for a few frames, then run a separate `open` phase at the same `drop_pos` before retracting. That would make the final pick-and-place result match the teaching intent of the section.

docs/tutorials/01_newton_course.ipynb
```diff
@@ -0,0 +1,1650 @@
...
+        "\n",
+        "def record_sequence():\n",
+        "    sim_time = 0.0\n",
+        "    phase_bar = tqdm(sequence, desc=\"phase: init\")\n",
+        "    for phase in phase_bar:\n",
+        "        phase_bar.set_description(f\"phase: {phase['name']}\")\n",
+        "        sim_time = run_phase(phase[\"pos\"], phase[\"grip\"], int(phase[\"frames\"]), sim_time)\n",
+        "\n",
+        "\n",
+        "grip_open = 0.06\n",
```
preist-nvidia: P2: Keep the gripper targets within the Franka finger joint limits. The URDF-imported model reports finger limits of `[0.0, 0.04]`, but this cell uses `grip_open = 0.06`, and `build_franka_scene()` earlier initializes the two finger targets to `1.0, 1.0`. The simulation may clamp or resist those values, but the tutorial is teaching target control, so the commanded targets should be valid, e.g. use `0.04` for the open target consistently.

docs/tutorials/01_newton_course.ipynb
```diff
@@ -0,0 +1,1650 @@
...
+      "source": [
+        "SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers\n",
+        "SPDX-License-Identifier: CC-BY-4.0"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "# Newton 101: From Concepts to Manipulation\n",
```
preist-nvidia: P3: Reframe this as the robotics tutorial rather than a second general intro. `00_introduction` should stay the minimal rigid-body quickstart, while this notebook can keep a short simulation-loop recap before moving into CUDA graphs, Franka control, IK, and manipulation. A title like `Robotics with Newton` or `Newton Robotics 101: From Concepts to Manipulation` would make that split clearer.

.github/workflows/docs-dev.yml
```diff
@@ -1,5 +1,15 @@
...
 
+env:
+  AWS_REGION: us-east-2
+  AWS_INSTANCE_TYPE: g7e.2xlarge
+  AWS_VOLUME_SIZE: 92
+  AWS_VOLUME_TYPE: gp3
+  AWS_SECURITY_GROUP_IDS: sg-07807c44e7f2a368a
+  AWS_ROLE_ARN: arn:aws:iam::968945269301:role/newton-physics-newton-ec2-github-runner-role
+  AWS_ROLE_DURATION: 3600
+  HOME: /actions-runner
```
preist-nvidia: P1: Do not set `HOME=/actions-runner` at workflow scope. This workflow also has `ubuntu-latest` jobs (`start-runner`, `stop-runner`, and especially `deploy`) where `/actions-runner` does not exist. On hosted runners, GitHub actions and git commonly write under `$HOME`; for example `HOME=/actions-runner git config --global --add safe.directory ...` fails with `could not lock config file /actions-runner/.gitconfig`. Scope this override only to the self-hosted `build-docs` job/step that needs the NVMe-backed cache, or create the directory in every hosted job before anything invokes git.

.github/workflows/pr.yml
```diff
@@ -66,12 +67,8 @@ jobs:
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6.2.0
         with:
           python-version-file: ".python-version"
-      - name: Install pandoc
-        uses: pandoc/actions/setup@86321b6dd4675f5014c611e05088e10d4939e09e  # v1.1.1
       - name: Build Sphinx documentation
         run: uv run --extra docs --extra sim sphinx-build -j auto -W -b html docs docs/_build/html
```
preist-nvidia, line 72: P2: Add a PR-time notebook smoke check now that notebook tutorials are excluded from the normal docs build. With the new `docs/conf.py` gating, this `pull-request-docs` command runs with `build_notebooks=False`, so it excludes `docs/tutorials/*.ipynb`; the notebook conversion, generated tutorial toctree, Viser recording paths, and copied Viser client are only exercised later by the main/tag deploy workflows. That means a PR can merge and then break the docs deployment. A lightweight check such as `NEWTON_BUILD_NOTEBOOKS=1 NEWTON_NBSPHINX_EXECUTE=never uv run --extra docs --extra notebook --extra sim sphinx-build -j auto -b html docs docs/_build/html-notebooks-smoke` would catch parse/conversion/static-asset regressions without executing the expensive GPU cells.

preist-nvidia:
@shi-eric if you could have a look at the automation changes that would be helpful.

I'll post what I found with Codex.

docs/tutorials/01_newton_course.ipynb
```diff
@@ -0,0 +1,1650 @@
...
+        "            wp.capture_launch(graph)"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "## 4) Load a Robot + Apply Joint Forces\n",
+        "\n",
+        "We will load a Franka arm from URDF and apply a sinusoidal torque using `Control.joint_f`. This is the lowest-level control interface, useful for direct torque control or learned policies."
```
preist-nvidia: P3: Please file a follow-up for adding a richer Franka source asset, ideally USD with the MuJoCo schema, or MJCF if that is the available intermediate. We do not currently have that in `newton-assets`, so URDF is fine for this PR.

docs/tutorials/01_newton_course.ipynb
```diff
@@ -0,0 +1,1650 @@
...
+        "FRANKA_HOME_Q = [\n",
+        "    -3.680e-03,\n",
+        "    2.390e-02,\n",
+        "    3.680e-03,\n",
+        "    -2.368e00,\n",
+        "    -1.292e-04,\n",
+        "    2.392e00,\n",
+        "    7.855e-01,\n",
+        "    0.05,\n",
+        "    0.05,\n",
```
preist-nvidia: P3: Please add provenance or a tuning note for these armature values. The effort limits are spec/URDF-derived, but URDF does not carry a MuJoCo-style joint armature field, so `[0.195] * 4 + [0.074] * 3 + [0.1] * 2` appears to be hand-authored simulation tuning. That is fine, but readers should know where it came from or why it is being added.

docs/tutorials/01_newton_course.ipynb
```diff
@@ -0,0 +1,1650 @@
...
+        "# Franka arm initial joint positions (7 arm joints + 2 gripper fingers)\n",
+        "FRANKA_HOME_Q = [\n",
+        "    -3.680e-03,\n",
+        "    2.390e-02,\n",
+        "    3.680e-03,\n",
+        "    -2.368e00,\n",
+        "    -1.292e-04,\n",
+        "    2.392e00,\n",
+        "    7.855e-01,\n",
+        "    0.05,\n",
```
preist-nvidia: P3: Please justify or soften these joint target gains. Values like `ke=3500-4500` and `kd=350-450` will saturate the realistic effort limits for very small tracking errors, so this behaves like an effort-limited stiff servo. If these are tuned for playback, say so; otherwise lower gains and/or gravity compensation may make the controller easier to reason about.

docs/tutorials/01_newton_course.ipynb
```diff
@@ -0,0 +1,1650 @@
...
+        "        include_table: Whether to add a table.\n",
+        "        include_cube: Whether to add a cube on the table.\n",
+        "        use_targets: Whether to enable joint target gains for PD control.\n",
+        "\n",
+        "    Returns:\n",
+        "        Tuple of (builder, cube_size, table_pos, cube_body).\n",
+        "    \"\"\"\n",
+        "    builder = newton.ModelBuilder()\n",
+        "    builder.default_shape_cfg.gap = 0.0\n",
+        "    newton.solvers.SolverMuJoCo.register_custom_attributes(builder)\n",
```
preist-nvidia: P3: Since this tutorial calls `register_custom_attributes()`, please author at least one MuJoCo-specific custom attribute as an example. `jnt_actgravcomp` would be a good fit here: it demonstrates how Newton carries MuJoCo-only joint data into `SolverMuJoCo`, and gravity compensation may also let the example use less aggressive PD gains. Right now the attributes are registered but not used.

docs/tutorials/01_newton_course.ipynb
```diff
@@ -0,0 +1,1650 @@
...
+        "    solver=\"newton\",  # Constraint solver (\"cg\" or \"newton\")\n",
+        "    integrator=\"implicitfast\",  # Integration method (\"euler\", \"rk4\", \"implicit\", \"implicitfast\")\n",
+        "    iterations=20,  # Number of solver iterations\n",
+        "    ls_parallel=True,  # Enable parallel line search\n",
+        "    ls_iterations=100,  # Number of line-search iterations\n",
+        "    nconmax=1000,  # Maximum number of contact points per world\n",
+        "    njmax=2000,  # Maximum number of constraints per world\n",
+        "    cone=\"elliptic\",  # Contact friction cone (\"pyramidal\" or \"elliptic\")\n",
+        "    impratio=1000.0,  # Frictional-to-normal constraint impedance ratio\n",
+        "    use_mujoco_contacts=False,  # Use Newton contacts instead of MuJoCo contacts\n",
```
preist-nvidia: P3: Please call out this choice in the surrounding text, not only as an inline code comment. `use_mujoco_contacts=False` is an important Newton/MuJoCo workflow: contacts are generated by Newton via `model.collide(...)` and then consumed by `SolverMuJoCo`, instead of using MuJoCo native contact generation. This tutorial is a good place to explain when/why users would choose the Newton contact pipeline.

docs/tutorials/01_newton_course.ipynb
```diff
@@ -0,0 +1,1650 @@
...
+        "viewer"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "## Inverse Kinematics (IK) in Newton\n",
+        "\n",
+        "Newton's IK module `newton.ik` is a GPU-first, batched inverse-kinematics system built on Warp that lets you compose multiple objectives into a single solve, then choose the optimization strategy (`LM` or `LBFGS`), Jacobian backend (`ANALYTIC`, `AUTODIFF`, or `MIXED`), and optional multi-seed sampling (`NONE`, `GAUSS`, `UNIFORM`, `ROBERTS`). Because it solves many IK problems in parallel and can leverage CUDA graph execution for low-overhead repeated solves, Newton scales especially well in large batched workloads, delivering throughput and solve quality that are competitive with modern GPU IK libraries such as PyRoKi and cuRobo."
```
preist-nvidia: The perf claim should ideally come with a link to a public benchmark.

docs/tutorials/01_newton_course.ipynb
preist-nvidia: P3: Could you pull a little more of the spoken context from the GTC transcript into the notebook prose so this reads better standalone?

In particular, move some of the later SolverMuJoCo motivation up to this first robot section: section 7 explains that MuJoCo is robust for rigid-body contact/joints and mentions the Newton contacts path, but by then the notebook has already used `SolverMuJoCo` several times.

preist-nvidia (commented):
The video in the intro is actually broken since 1.1, can you please have a look in the context of this PR? See https://newton-physics.github.io/newton/latest/tutorials/00_introduction.html#Running-the-Simulation

Or file a separate bug that's fine too.

shi-eric:
Is it really not possible to do the CI/CD changes in a follow-up merge request? I think we should narrow the blast radius.

eric-heiden:
> The video in the intro is actually broken since 1.1, can you please have a look in the context of this PR? See https://newton-physics.github.io/newton/latest/tutorials/00_introduction.html#Running-the-Simulation
> 
> Or file a separate bug that's fine too.

This should be fixed by this PR as well hopefully, this is the same ViewerViser output that is used in this new notebook.

eric-heiden:
> Is it really not possible to do the CI/CD changes in a follow-up merge request? I think we should narrow the blast radius.

I can move this to a follow-up PR. I can just include the notebook + ViewerViser changes here.

docs/tutorials/01_newton_course.ipynb
eric-heiden: Done: https://github.com/newton-physics/newton-assets/issues/40

.github/workflows/docs-release.yml
eric-heiden: Deferring the CI updates to https://github.com/newton-physics/newton/pull/2666. I've addressed this problem in that PR.

.github/workflows/pr.yml
eric-heiden: Deferring the CI updates to https://github.com/newton-physics/newton/pull/2666. But to be clear my intention is to evaluate all notebooks as part of every PR CI checks, so including this new notebook which should run on a GPU machine.

docs/tutorials/01_newton_course.ipynb
eric-heiden: Done

docs/tutorials/01_newton_course.ipynb
eric-heiden: Addressed in 77dc691b4. I expanded the notebook prose using the GTC lab transcript as context so it reads more like a standalone course, and moved the SolverMuJoCo/Newton contact-pipeline motivation up into the first robot section before the solver is used repeatedly.

Also tightened the writing throughout to explain the concrete data flow (`ModelBuilder` → `Model` → `State`/`Control`/`Contacts` → solver), contact reuse, FK, joint force/target semantics, IK objectives/Jacobians/sampling, and the open-loop nature of the pick-and-place example.

Verified with:
- `uvx pre-commit run -a`
- `nbformat` validation for `docs/tutorials/01_newton_course.ipynb`

docs/tutorials/01_newton_course.ipynb
eric-heiden: Addressed in 997e3a4b1. The final sequence now separates placement from release: `move` ? `place` ? `hold` keeps `grip_closed` at `drop_pos`, then a separate `open` phase interpolates only the gripper target at the same `drop_pos` before `retract`.

I also executed the notebook locally and checked the final manipulation result: the cube ended at `[0.015, -0.650, 0.126]` for a drop target of `[0.000, -0.650, 0.125]`, with the gripper target reopened to `[0.04, 0.04]`.

docs/tutorials/01_newton_course.ipynb
eric-heiden: Addressed in 997e3a4b1. I added `FRANKA_FINGER_OPEN = 0.04` and `FRANKA_FINGER_CLOSED = 0.0`, then used those constants for `FRANKA_HOME_Q`, the default target positions, the joint-target demo, and the final pick-and-place sequence. The old out-of-range `1.0, 1.0` and `0.06` targets are gone.

Local verification confirmed the imported finger limits are `[0.0, 0.04]`, the initial/default finger targets are `[0.04, 0.04]`, and all phase gripper commands stay inside those limits.

docs/tutorials/01_newton_course.ipynb
eric-heiden: Addressed in 997e3a4b1. I retitled the notebook to `Newton Robotics 101: From Concepts to Manipulation` and adjusted the intro so it frames the course as a robotics workflow rather than a second general Newton introduction.

docs/tutorials/01_newton_course.ipynb
eric-heiden: Addressed in 997e3a4b1. The notebook now calls this out explicitly: the effort limits match the robot data, but `FRANKA_MUJOCO_ARMATURE` is a MuJoCo-style simulation tuning value, not a URDF-provided robot specification. The prose explains that it adds a small diagonal inertia term to regularize the articulated solve.

docs/tutorials/01_newton_course.ipynb
eric-heiden: Addressed in 997e3a4b1. I lowered the target gains from the previous `3500?4500` / `350?450` range to `FRANKA_TARGET_KE = [900, 900, 700, 700, 400, 400, 400, 100, 100]` and `FRANKA_TARGET_KD = [90, 90, 70, 70, 40, 40, 40, 10, 10]`. The surrounding text now explains that gravity compensation handles the static load while the PD gains correct tracking errors under the effort limits.

The full notebook execution still passes with these lower gains, including the final pick-and-place check.

docs/tutorials/01_newton_course.ipynb
eric-heiden: Addressed in 997e3a4b1. `build_franka_scene(use_targets=True)` now authors MuJoCo custom attributes after registering them: `mujoco:gravcomp = 1.0` on the imported robot bodies and `mujoco:jnt_actgravcomp = True` on the controlled DOFs. The notebook prose introduces both attributes and explains why they are enabled only for the target-control scenes.

I verified the finalized model materializes these as `model.mujoco.gravcomp` and `model.mujoco.jnt_actgravcomp`, with the controlled DOFs enabled.

docs/tutorials/01_newton_course.ipynb
eric-heiden: Addressed in 77dc691b4 and still covered after the latest edits. The first robot-control section now explains that `use_mujoco_contacts=False` means contacts are generated by Newton through `model.collide(state_0, contacts)` and then passed into `SolverMuJoCo`, rather than using MuJoCo native contact generation. The final manipulation section repeats that choice before the pick-and-place setup.

docs/tutorials/01_newton_course.ipynb
eric-heiden: Addressed in 77dc691b4. The `newton.ik` introduction now links to the public benchmark tables in newton-physics/newton#394 and phrases the numbers as benchmark context rather than a blanket guarantee.

docs/tutorials/01_newton_course.ipynb
eric-heiden: Resolved via the follow-up filed at https://github.com/newton-physics/newton-assets/issues/40. The notebook keeps the URDF asset for this PR and now explicitly documents the MuJoCo-specific tuning/custom attributes we add around it.

docs/tutorials/01_newton_course.ipynb
eric-heiden: Done

docs/tutorials/01_newton_course.ipynb
eric-heiden: Done

docs/tutorials/01_newton_course.ipynb
eric-heiden: I cleaned up these magic numbers and also use gravity compensation now so the ke/kd values are lower now.

docs/tutorials/01_newton_course.ipynb
eric-heiden: Done, using gravity compensation now.

docs/tutorials/01_newton_course.ipynb
eric-heiden: The only benchmark results we have right now are shared in https://github.com/newton-physics/newton/pull/394 which I link to here now. But we should make an effort to publish them in the Newton docs.

eric-heiden:
@preist-nvidia I've renamed `docs/tutorials/01_newton_course.ipynb` to `docs/tutorials/01_robotics.ipynb` and improved the structure and content. It should be now much more self-contained with more explanations and cleaned up code.

camevor (approved):

PR #2419, opened 2026-04-13
Optimize hydroelastic contact performance

nvtw:
## Description

Optimize the hydroelastic broadphase scatter kernel (`broadphase_collision_pairs_scatter`) by replacing the per-pair thread launch with a grid-stride loop over the total block count. Each thread now uses binary search on the prefix-sum array to map its block index back to the owning shape pair, which eliminates out-of-bounds writes when block counts exceed pair counts and removes the need for the per-pair `num_blocks` input array.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_sdf_hydroelastic
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved performance and scalability of hydroelastic broadphase collision scattering, reducing runtimes for complex scenes and large shape counts.

* **Chores**
  * Updated CHANGELOG to document the broadphase scatter implementation change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -140,6 +140,7 @@
 - Fix O(W²·S²) memory explosion in `CollisionPipeline` shape-pair buffer allocation for NXN and SAP broad phase modes by computing per-world pair counts instead of a global N²
 - Clamp viewer picking force to prevent explosion when picking light objects near stiff contacts, configurable via `pick_max_acceleration` parameter on the `Picking` class (default 5g of effective articulation mass)
 - Fix `cloth_franka` example Jacobian broken by COM-referenced `body_qd` convention change; adjust robot base height, gripper orientations, and grasp targets for improved reachability (a follow-up PR will migrate the example to `newton.ik`)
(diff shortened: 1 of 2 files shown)


mzamoramora-nvidia:
Thanks for the PR! I think there might be one potential issue with the binary search bounds.

`shape_pairs_sdf_sdf_count[0]` can exceed `max_num_shape_pairs` (the counter is incremented unconditionally, only the write is guarded). The new kernel uses it unclamped as the binary search bound on `block_start_prefix`, which is sized `max_num_shape_pairs`:

```python
pair_count = shape_pairs_sdf_sdf_count[0]
hi = int(pair_count - 1)  # can exceed array bounds
```

The old kernel was immune — `dim=[self.max_num_shape_pairs]` structurally bounded all access.

Suggested fix:
```python
pair_count = wp.min(shape_pairs_sdf_sdf_count[0], max_num_shape_pairs)
```

https://github.com/newton-physics/newton/blob/f6018508faefc1694ee49722a2e251865ec134d0/newton/_src/geometry/sdf_hydroelastic.py#L946-L955

mzamoramora-nvidia (approved):

PR #2424, opened 2026-04-13
Fix degeneracies in hydroelastic contact surfaces

### LINKED ISSUE #2348: [BUG] Marching cubes produces degenerate (zero-area) triangles under deep interpenetration for hydroelastic contact
### Bug Description

The hydroelastic contact surface contains a significant fraction of degenerate triangles (zero or near-zero area) when two shapes are deeply interpenetrated. Under deep overlap, the stiffness-weighted SDF difference field `f(x) = k_a * sdf_a(x) - k_b * sdf_b(x)` has a vanishing gradient, causing marching cubes vertex interpolation to become numerically unstable. At extreme penetration, entire regions of the contact surface are missing (no triangles generated at all).

### What happened

When two hydroelastic shapes interpenetrate deeply, the contact surface extracted by marching cubes contains:
- **Degenerate triangles** (area < 1% of median) — up to 14% of all faces for equal-stiffness primitives, 1-3% for curved/mesh geometries.
- **Zero-area triangles** (exactly 0.0 m²) — 2-4% for primitive box-box contacts.
- **Missing surface regions** — at deep overlap, a gap appears in the center of the contact patch where no triangles are generated at all, because the SDF difference field is completely flat.

### What was expected

All isosurface triangles should have well-defined, non-zero area. The contact surface should be a clean triangulated mesh with smooth area variation and no gaps.

### Visualizations

**Isosurface degradation at increasing penetration (ellipsoid into box):**

An ellipsoid (40x40x60mm) is pushed into a flat box at 5mm, 15mm, 30mm, and 50mm overlap. Blue = well-formed triangles, red = degenerate, yellow = no triangles generated.

Image

**Dynamic collision (ellipsoid pushed into box through live pipeline):**

At **5mm** the contact patch is a clean filled circle. By **50mm** a large yellow gap appears at the center — the SDF difference field is completely flat there (both SDFs clamped to their narrow band limit), so marching cubes finds no zero-crossing. Degenerate faces (red) ring this dead zone.

A kinematic ellipsoid is driven at 60 mm/s into a stationary box, running the full broadphase → narrowphase → hydroelastic pipeline each frame. Contact surface snapshots at 5mm, 15mm, 30mm, 50mm, and 80mm overlap:

Image

The same center gap appears dynamically. At **5mm** the surface is a clean circle. By **50mm** a yellow gap has opened at the center, growing larger at **80mm**. This confirms the issue is not an artifact of pre-overlapped initial poses; it emerges through the live collision pipeline as well.

### Root cause

Marching cubes interpolates vertex positions along voxel edges:
```
t = (0 - val_0) / (val_1 - val_0)
vertex = p_0 + t * (p_1 - p_0)
```
When both shapes are deeply interpenetrated and `k_a * sdf_a ≈ k_b * sdf_b` across many voxels, the denominator `(val_1 - val_0)` approaches zero. The interpolation becomes unstable, producing collapsed vertices. At extreme depth, the field is flat over entire voxel regions, and no zero-crossing is found — resulting in gaps in the contact surface.

### Reproduction Script

```python
"""Minimal reproduction: marching cubes degeneracy under deep penetration.

Two hydroelastic boxes with controlled overlap. Run:
    python repro_mc_degeneracy.py
"""

import numpy as np
import warp as wp

import newton

wp.init()

DEVICE = "cuda:0"
BOX_HALF = 0.1  # 10cm half-extent → 20cm cube


def compute_triangle_areas(vertices: np.ndarray, num_faces: int) -> np.ndarray:
    v = vertices[: num_faces * 3].reshape(num_faces, 3, 3)
    e1 = v[:, 1] - v[:, 0]
    e2 = v[:, 2] - v[:, 0]
    return 0.5 * np.linalg.norm(np.cross(e1, e2), axis=1)


def run_overlap_test(overlap: float, kh_a: float = 1e10, kh_b: float = 1e10):
    # Scale proportionally to shape size (matches test_hydroelastic.py convention)
    narrow_band = BOX_HALF * 0.2
    contact_gap = BOX_HALF * 0.2

    def make_cfg(kh):
        return newton.ModelBuilder.ShapeConfig(
            mu=0.5, kh=kh, sdf_max_resolution=64, is_hydroelastic=True,
            sdf_narrow_band_range=(-narrow_band, narrow_band), gap=contact_gap,
        )

    body_a = builder.add_body(
        xform=wp.transform(wp.vec3(0.0, 0.0, BOX_HALF), wp.quat_identity()), label="box_a",
    )
    builder.add_shape_box(body=body_a, hx=BOX_HALF, hy=BOX_HALF, hz=BOX_HALF, cfg=make_cfg(kh_a))

    z_b = BOX_HALF + 2.0 * BOX_HALF - overlap
    body_b = builder.add_body(
        xform=wp.transform(wp.vec3(0.0, 0.0, z_b), wp.quat_identity()), label="box_b",
    )
    builder.add_shape_box(body=body_b, hx=BOX_HALF, hy=BOX_HALF, hz=BOX_HALF, cfg=make_cfg(kh_b))

    model = builder.finalize(device=DEVICE)
    state = model.state()

    hydro_config = HydroelasticSDF.Config(
        output_contact_surface=True, reduce_contacts=False,
        buffer_mult_iso=4, buffer_mult_contact=4,
    )
    collision_pipeline = newton.CollisionPipeline(
        model, rigid_contact_max=100000, broad_phase="explicit",
        sdf_hydroelastic_config=hydro_config,
    )

    contacts = collision_pipeline.contacts()
    collision_pipeline.collide(state, contacts)

    cs = collision_pipeline.hydroelastic_sdf.get_contact_surface()
    if cs is None:
        return None

    num_faces = int(cs.face_contact_count.numpy()[0])
    if num_faces == 0:
        return None

    vertices = cs.contact_surface_point.numpy()
    areas = compute_triangle_areas(vertices, num_faces)
    return num_faces, areas


def main():
    for label, kh_a, kh_b in [
        ("Equal stiffness (kh=1e10, kh=1e10)", 1e10, 1e10),
        ("Asymmetric stiffness (kh=1e10, kh=1e8)", 1e10, 1e8),
    ]:
        print(f"--- {label} ---")
        for overlap in [0.02, 0.05, 0.10, 0.15]:
            result = run_overlap_test(overlap, kh_a=kh_a, kh_b=kh_b)
            pct = overlap / (2 * BOX_HALF) * 100
            if result is None:
                print(f"  Overlap {overlap*100:.0f}cm ({pct:.0f}%): no contact surface")
                continue
            num_faces, areas = result
            median_area = np.median(areas)
            num_degen = int((areas < 0.01 * median_area).sum())
            num_zero = int((areas < 1e-20).sum())
            print(f"  Overlap {overlap*100:.0f}cm ({pct:.0f}%): {num_faces} faces, "
                  f"{num_zero} zero-area ({num_zero/num_faces*100:.1f}%), "
                  f"{num_degen} degenerate ({num_degen/num_faces*100:.1f}%)")
        print()


if __name__ == "__main__":
    main()
```

### System Information

- **Newton version:** 1.2.0.dev0 (commit `0910bd5c`, main branch as of 2026-04-07)
- **Warp version:** 1.12.0rc1
- **Python version:** 3.12
- **OS:** Ubuntu Linux 6.17.0
- **GPU:** NVIDIA RTX 6000 Ada Generation (2x, 49GB each)
- **CUDA Toolkit:** 12.9

nvtw:
## Description

Fixes https://github.com/newton-physics/newton/issues/2348. Adds small safeguards that improve the quality of triangles on the iso-pressure surface generated by the hydroelastic contact pipeline.

**Root cause:** When two SDF boxes overlap deeply, marching-cubes edge interpolation can place vertices exactly at cube corners (t ≈ 0 or t ≈ 1). Multiple vertices collapse to the same corner, producing degenerate zero-area triangles. The subsequent `wp.normalize()` on a zero-length cross product returns NaN normals.

**Changes (in both `sdf_mc.py` and `sdf_hydroelastic.py`):**
1. Clamp the edge interpolation parameter `t` to `[0.02, 0.98]` so vertices stay away from cube corners.
2. Tighten the near-zero `val_diff` threshold from `1e-8` to `1e-10` to avoid unnecessary midpoint snapping.
3. Guard the triangle normal computation: when the cross-product squared length is below `1e-30`, return zero area with a valid fallback normal instead of NaN.

## Checklist

## Test plan

     or manual steps. Example:
     ```
     ``` -->
```
uv run --extra dev -m newton.tests -k test_no_degenerate_triangles_deep_penetration
```
New test creates two hydroelastic boxes at multiple penetration depths and stiffness ratios, then verifies the contact surface contains no zero-area triangles and fewer than 2% degenerate triangles.

## Bug fix

<!-- DELETE this section if not a bug fix.
     Describe how to reproduce the issue WITHOUT this PR applied. -->

**Steps to reproduce:**

See https://github.com/newton-physics/newton/issues/2348



## Summary by CodeRabbit

* **Bug Fixes**
  * Improved marching‑cubes isosurface extraction to avoid degenerate/zero‑area triangles by tightening edge interpolation, clamping edge vertex positions away from cube corners, and skipping near‑degenerate faces during contact generation.

* **Tests**
  * Added a test validating stable, non‑degenerate contact surfaces under deep‑penetration scenarios.

* **Documentation**
  * Changelog updated noting the degenerate‑triangle fixes.


newton/_src/geometry/sdf_hydroelastic.py
```diff
@@ -132,10 +132,15 @@ def mc_calc_face_texture(
         p_0 = wp.vec3f(corner_offsets_table[v_idx_from])
         p_1 = wp.vec3f(corner_offsets_table[v_idx_to])
         val_diff = wp.float32(val_1 - val_0)
-        if wp.abs(val_diff) < 1e-8:
+        if wp.abs(val_diff) < 1.0e-10:
```
adenzler-nvidia: is the constant relevant here, or should this be minval/-maxval?

newton/_src/geometry/sdf_hydroelastic.py
```diff
@@ -132,10 +132,15 @@ def mc_calc_face_texture(
...
             t = float(0.5)
         else:
-            t = (0.0 - val_0) / val_diff
+            # Clamp t away from cube corners to prevent vertex collapse when
+            # corner values are near zero (e.g. at SDF ridge boundaries where
+            # both shapes share the same nearest face).  Without the clamp,
+            # t ≈ 0 or t ≈ 1 places multiple vertices at the same corner,
+            # producing degenerate (zero-area) triangles.
+            t = wp.clamp((0.0 - val_0) / val_diff, 0.02, 0.98)
```
adenzler-nvidia: The three new constants (`1.0e-10`, `0.02`/`0.98`, `1.0e-30`) are duplicated across `sdf_mc.py` and this file without derivation. Consider extracting them as named constants in `sdf_mc.py` (e.g., `MC_EDGE_CLAMP_MIN = 0.02`) and importing here — avoids silent drift and gives the rationale a single home.

Re the `val_diff` threshold on L135: `1.0e-10` is well below float32 precision (~1.2e-7) for typical SDF magnitudes, so this branch may rarely fire in practice. Should it relate to an existing constant like `MINVAL` (used as `1e-15` in `collision_primitive.py` / `raycast.py`), or is the specific value intentional?

newton/_src/geometry/sdf_hydroelastic.py
```diff
@@ -149,8 +154,15 @@ def mc_calc_face_texture(
             num_inside += 1
 
     n = wp.cross(face_verts[1] - face_verts[0], face_verts[2] - face_verts[0])
-    normal = wp.normalize(n)
-    area = wp.length(n) / 2.0
+    n_sq = wp.dot(n, n)
+    if n_sq < 1.0e-30:
```
adenzler-nvidia: `1.0e-30` only catches exact-zero cross products in float32. With t-clamping producing minimum edge lengths of ~0.02 voxel edges, a near-collinear triangle could have `n_sq` in the `1e-20`–`1e-14` range and still produce an unreliable normal after division. Raising to ~`1e-20` would be more conservative with negligible downside (false positives just zero out a tiny triangle’s force contribution via `area = 0`).

newton/_src/geometry/sdf_mc.py
```diff
@@ -169,10 +169,12 @@ def mc_calc_face(
         p_0 = wp.vec3f(corner_offsets_table[v_idx_from])
         p_1 = wp.vec3f(corner_offsets_table[v_idx_to])
         val_diff = wp.float32(val_1 - val_0)
-        if wp.abs(val_diff) < 1e-8:
+        if wp.abs(val_diff) < 1.0e-10:
             p = 0.5 * (p_0 + p_1)
         else:
-            t = (isovalue - val_0) / val_diff
+            # Clamp t away from cube corners to prevent vertex collapse when
+            # corner values are near zero (e.g. at SDF ridge boundaries).
```
adenzler-nvidia, line 202: nit: The `sdf_hydroelastic.py` version includes the consequence explanation (“Without the clamp, t ≈ 0 or t ≈ 1 places multiple vertices at the same corner, producing degenerate (zero-area) triangles.”) which applies equally here. The omission looks unintentional — suggest matching the longer version (minus the hydroelastic-specific “both shapes share the same nearest face” part).

adenzler-nvidia (commented):
Nice fix — the belt-and-suspenders approach (t-clamping + degenerate guard) is solid. A few observations:

### Degenerate triangles still flow into the contact buffer / hashtable

When `mc_calc_face_texture` returns `area = 0.0`, the caller in `generate_contacts_kernel` (~L1523) doesn’t skip the triangle. It still gets a hashtable slot via `hashtable_find_or_insert` and (when `pre_prune=False`) a contact buffer slot. The `area = 0.0` neutralizes the force, but wasted slots could contribute to “face contact overflow” warnings in the deep-penetration scenarios this PR targets. An early `if area <= 0.0: continue` in the face loop would be cleaner.

### Same interpolation pattern unfixed in `sdf_utils.py` and `sdf_texture.py`

There are four copies of the MC edge-interpolation + `wp.normalize(n)` pattern in the codebase. This PR fixes two (`sdf_mc.py` and `sdf_hydroelastic.py`), but `sdf_utils.py:1364` and `sdf_texture.py:1869` still use the old `1e-8` threshold with no t-clamping and no degenerate-triangle guard (`sdf_utils.py:1376` still calls `wp.normalize(n)` directly). Worth fixing here or in a follow-up.

newton/tests/test_hydroelastic.py
```diff
@@ -1237,5 +1237,109 @@ def _run_viewer_test(self, shape_type: ShapeType, solver_name: str = "xpbd", cub
 )
 
 
+def test_no_degenerate_triangles_deep_penetration(test, device):
+    """Verify marching cubes produces no zero-area triangles under deep interpenetration.
```
adenzler-nvidia: nit: The docstring says “no zero-area triangles” but the test also asserts `< 2%` degenerate triangles (area < 1% of median). The function name similarly implies zero tolerance. Suggest: “Verify marching cubes produces no zero-area triangles and fewer than 2% near-degenerate triangles under deep interpenetration.”

newton/tests/test_hydroelastic.py
```diff
@@ -1237,5 +1237,109 @@ def _run_viewer_test(self, shape_type: ShapeType, solver_name: str = "xpbd", cub
...
+            sdf_narrow_band_range=(-narrow_band, narrow_band),
+            gap=contact_gap,
+        )
+
+    configs = [
+        # (overlap, kh_a, kh_b, label)
+        (0.05, 1e10, 1e10, "equal stiffness 25% overlap"),
+        (0.10, 1e10, 1e10, "equal stiffness 50% overlap"),
+        (0.15, 1e10, 1e10, "equal stiffness 75% overlap"),
+        (0.10, 1e10, 1e8, "asymmetric stiffness 50% overlap"),
```
adenzler-nvidia, line 1272: suggestion: Consider adding a near-100% overlap case, e.g. `(0.19, 1e10, 1e10, "equal stiffness 95% overlap")`. Deep interpenetration right at SDF ridge boundaries is where vertex collapse is most likely — the current maximum of 75% may not stress the exact scenario from #2348.

newton/_src/geometry/sdf_hydroelastic.py
nvtw: Done

newton/_src/geometry/sdf_hydroelastic.py
nvtw: Done

newton/tests/test_hydroelastic.py
nvtw: Done

newton/_src/geometry/sdf_mc.py
nvtw: Should be improved now

newton/tests/test_hydroelastic.py
nvtw: Should be improved now

adenzler-nvidia (approved):
Thanks for the thorough follow-up -- all previous comments are addressed. The named constants with docstrings, the raised degenerate threshold, the `area <= 0.0` early-out in the contact kernel, and the 95% overlap test case all look good.

Two optional nits for a follow-up:
- Consider `MC_EDGE_CLAMP_MAX = 1.0 - MC_EDGE_CLAMP_MIN` to keep the coupling explicit.
- Adding `test.assertTrue(np.isfinite(v).all(), ...)` before the area checks would guard against NaN bypass.

Neither is blocking. LGTM.

Issue #2290, opened 2026-04-01
[BUG] SolverMuJoCo.notify_model_changed() does not update mj_model when SolverNotifyFlags.CONSTRAINT_PROPERTIES is raised.


gyeomannvidia:
### Bug Description

The core issue is that notify_model_changed directly updates mjw_model.  With MuJoCo CPU we first update mjw_model and then do some copying from mjw_model to mj_model.  This copy appears to be incomplete, meaning that mj_model does not reflect the updated properties.

We have this code for CONSTRAINT_PROPERTIES:

```
        if flags & SolverNotifyFlags.CONSTRAINT_PROPERTIES:
            self._update_eq_properties()
            self._update_mimic_eq_properties()
```

_update_eq_properties() updates the arrays stored in mjw_model.  We need to copy this back to mj_model but we do not appear to do this.  

We do copy for BODY_PROPERTIES and JOINT_DOF_PROPERTIES:

      if flags & (SolverNotifyFlags.BODY_PROPERTIES | SolverNotifyFlags.JOINT_DOF_PROPERTIES):
                self.mj_model.dof_armature[:] = self.mjw_model.dof_armature.numpy()[0]
                self.mj_model.dof_frictionloss[:] = self.mjw_model.dof_frictionloss.numpy()[0]
                self.mj_model.dof_damping[:] = self.mjw_model.dof_damping.numpy()[0]
                self.mj_model.dof_solimp[:] = self.mjw_model.dof_solimp.numpy()[0]
                self.mj_model.dof_solref[:] = self.mjw_model.dof_solref.numpy()[0]
                self.mj_model.qpos0[:] = self.mjw_model.qpos0.numpy()[0]
                self.mj_model.qpos_spring[:] = self.mjw_model.qpos_spring.numpy()[0]

We should do the same for CONSTRAINT_PROPERTIES eg

```
            if flags & SolverNotifyFlags.CONSTRAINT_PROPERTIES:
                self.mj_model.eq_data[:] = self.mjw_model.eq_data.numpy()[0]
                self.mj_model.eq_solref[:] = self.mjw_model.eq_solref.numpy()[0]
                self.mj_model.eq_solimp[:] = self.mjw_model.eq_solimp.numpy()[0]
                self.mj_data.eq_active[:] = self.mjw_data.eq_active.numpy()[0]
```

This issue may be bigger than described here so best to check if we need to copy more from mjw_model to mj_model.

### Reproduction Script

```python

```

[...]

PR #2430, opened 2026-04-13
Fix WELD eq_data corruption when CONNECT anchor kernels run on mixed …

gyeomannvidia:
…loop joint models

Joint-synthesized CONNECT constraints (from revolute/ball loop joints) now correctly update their anchors when dof_ref or joint_X_p changes at runtime. WELD entries (from FIXED loop joints) are excluded from mjc_eq_to_newton_jnt so the CONNECT kernels skip them, preventing overwrite of the WELD relpose data computed by spec.compile().

Add has_jnt_connect_constraints flag to avoid unnecessary kernel launches for models without joint-synthesized CONNECT constraints, and track actual eq_data writes with wrote_eq_data to fix a CPU sync asymmetry in _notify_connect_constraints_changed.

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **Bug Fixes**
  * CONNECT anchors now recompute correctly when joint parameters change at runtime.
  * CONNECT updates no longer overwrite or corrupt WELD constraint data in mixed FIXED/loop-joint models.

* **New Features**
  * Added support for CONNECT constraints synthesized from loop joints so anchors are maintained from joint frames.

* **Tests**
  * New tests verify runtime anchor recomputation across worlds and that WELD entries remain intact.

* **Documentation**
  * CHANGELOG updated with these fixes and additions.


newton/tests/test_equality_connect_constraint_with_sim_step.py
```diff
@@ -782,5 +782,690 @@ def _create_solver(self, model):
...
+                measured_a1_0 = wp.vec3(measured_eq_data[w][0][0], measured_eq_data[w][0][1], measured_eq_data[w][0][2])
+                measured_a2_0 = wp.vec3(measured_eq_data[w][0][3], measured_eq_data[w][0][4], measured_eq_data[w][0][5])
+                measured_a1_1 = wp.vec3(measured_eq_data[w][1][0], measured_eq_data[w][1][1], measured_eq_data[w][1][2])
+                measured_a2_1 = wp.vec3(measured_eq_data[w][1][3], measured_eq_data[w][1][4], measured_eq_data[w][1][5])
+
+                for k in range(3):
+                    self.assertAlmostEqual(float(anchor1_a[k]), float(measured_a1_0[k]), places=4)
+                    self.assertAlmostEqual(float(anchor2_a[k]), float(measured_a2_0[k]), places=4)
+                    self.assertAlmostEqual(float(anchor1_b[k]), float(measured_a1_1[k]), places=4)
+                    self.assertAlmostEqual(float(anchor2_b[k]), float(measured_a2_1[k]), places=4)
```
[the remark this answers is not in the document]
gyeomannvidia: Done

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -5979,8 +6044,120 @@ def _notify_connect_constraints_changed(
...
+                self.model,
+                self.mjw_model,
+                self.mjc_eq_to_newton_jnt,
+                self.jnt_eq_anchor1,
+                self.jnt_connect_constraint_q_rel,
+                self.jnt_connect_constraint_t_rel,
+            )
+            wrote_eq_data = True
+        if wrote_eq_data and self.use_mujoco_cpu:
+            self.mj_model.eq_data[:] = self.mjw_model.eq_data.numpy()[0]
```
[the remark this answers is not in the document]
gyeomannvidia: This is a pre-existing issue.  It is tracked in the following issue:

https://github.com/newton-physics/newton/issues/2290

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -2865,17 +2871,21 @@ def __init__(
         """Mapping from MuJoCo [world, eq] to Newton equality constraint index.
 
         Corresponds to the equality constraints that are created in MuJoCo from Newton's equality constraints.
-        A value of -1 indicates that the MuJoCo equality constraint has been created from a Newton joint, see :attr:`mjc_eq_to_newton_jnt`
-        for the corresponding joint index.
+        A value of -1 indicates the entry is unmapped -- the MuJoCo constraint was synthesized from a loop-closure
+        joint (CONNECT or WELD) rather than from an explicit Newton equality constraint. For CONNECT-only
+        entries see :attr:mjc_eq_to_newton_jnt.
```
eric-heiden: ⚪ Missing backticks in the Sphinx `:attr:` cross-reference. Without them this renders as plain text instead of a clickable link.

```suggestion
        entries see :attr:`mjc_eq_to_newton_jnt`.
```

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -2907,11 +2917,23 @@ def __init__(
...
         """Whether the model contains any CONNECT equality constraints."""
+        self.has_jnt_connect_constraints: bool = False
+        """Whether the model contains CONNECT constraints synthesized from loop-closure joints (revolute or ball)."""
         self.connect_constraint_q_rel: wp.array | None = None
         """Relative rotation ``inv(q2) * q1`` at the reference pose per equality constraint, ``wp.array[wp.quat]``, shape ``[equality_constraint_count]``."""
         self.connect_constraint_t_rel: wp.array | None = None
         """Relative translation [m] at the reference pose per equality constraint, ``wp.array[wp.vec3]``, shape ``[equality_constraint_count]``."""
 
+        # --- Internal state for anchor computation of joint-synthesized CONNECT constraints ---
+        self.jnt_eq_anchor1: wp.array | None = None
```
eric-heiden: ⚪ These four new fields (`jnt_eq_anchor1`, `jnt_eq_anchor1_has_axis_offset`, `jnt_connect_constraint_q_rel`, `jnt_connect_constraint_t_rel`) are allocated as 2D arrays with shape `[world_count, neq]` and their docstrings say `wp.array2d[...]`, but the type annotations use plain `wp.array | None`.

Nearby fields like `mjc_eq_to_newton_eq` already use `wp.array2d`, so updating these would keep things consistent:

```python
self.jnt_eq_anchor1: wp.array2d[wp.vec3] | None = None
self.jnt_eq_anchor1_has_axis_offset: wp.array2d[wp.int32] | None = None
self.jnt_connect_constraint_q_rel: wp.array2d[wp.quatf] | None = None
self.jnt_connect_constraint_t_rel: wp.array2d[wp.vec3] | None = None
```

newton/_src/solvers/mujoco/kernels.py
```diff
@@ -992,6 +992,184 @@ def update_connect_constraint_anchors_kernel(
...
+    data[1] = anchor[1]
+    data[2] = anchor[2]
+    data[3] = anchor2[0]
+    data[4] = anchor2[1]
+    data[5] = anchor2[2]
+    eq_data_out[world, mjc_eq] = data
+
+
 def create_convert_mjw_contacts_to_newton_kernel():
     """Create contact conversion kernel; deferred so ``wp.static`` doesn't import mujoco_warp at module load."""
```
adenzler-nvidia, line 1188: This kernel is structurally identical to `update_connect_constraint_anchors_kernel` above — both do:
```
anchor2 = quat_rotate(q_rel, anchor1) + t_rel
data[0:3] = anchor1; data[3:6] = anchor2
```

The only difference is where the anchor and rel-transform are looked up (per-constraint arrays vs per-`[world, eq]` arrays). Could these share a single kernel with a unified per-`[world, eq]` representation? That would eliminate one kernel, one static method, and simplify `_notify_connect_constraints_changed`. If that refactor is too large here, would you be open to tracking it as a follow-up?

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -4869,7 +4896,10 @@ def get_body_name(body_idx: int) -> str:
                 # Rotate axis into the parent body frame (anchor data[0:3] is
                 # in the parent body frame, so the offset must be too).
                 hinge_axis = wp.quat_rotate(parent_xform_tf.q, hinge_axis_local)
-                offset = 0.1  # offset along axis for second constraint point
+                offset = HINGE_CONNECT_AXIS_OFFSET
```
adenzler-nvidia, line 4908: Setting `has_connect_constraints = True` here (and in the ball branch below) changes the flag's meaning from "has explicit CONNECT constraints" to "has any CONNECT of either kind." This causes the explicit CONNECT kernel path in `_notify_connect_constraints_changed` to fire unnecessarily on models with only joint-synthesized CONNECT constraints (no-op but wasteful).

Consider keeping `has_connect_constraints` for explicit CONNECT only, and using `has_any_connect` (already computed in `notify_model_changed`) for the union.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -2865,17 +2871,21 @@ def __init__(
...
         Shape [nworld, neq], dtype int32."""
         self.mjc_eq_to_newton_jnt: wp.array2d[wp.int32] | None = None
-        """Mapping from MuJoCo [world, eq] to Newton joint index.
+        """Mapping from MuJoCo [world, eq] to Newton joint index for CONNECT constraints only.
 
-        Corresponds to the equality constraints that are created in MuJoCo from Newton joints that have no associated articulation,
-        i.e. where :attr:`newton.Model.joint_articulation` is -1 for the joint which results in 2 equality constraints being created in MuJoCo.
-        A value of -1 indicates that the MuJoCo equality constraint is not associated with a Newton joint but an explicitly created Newton equality constraint,
-        see :attr:`mjc_eq_to_newton_eq` for the corresponding equality constraint index.
+        Corresponds to the CONNECT equality constraints synthesized from
```
adenzler-nvidia, line 2884: Nit: Sphinx cross-ref should be `` :attr:`mjc_eq_to_newton_jnt` `` (backtick-wrapped target).

newton/tests/test_equality_connect_constraint_with_sim_step.py
```diff
@@ -782,5 +783,717 @@ def _create_solver(self, model):
...
+        else:
+            J0 = wp.transform(
+                wp.vec3(0.0, 0.0, 0.0),
+                wp.quat_from_axis_angle(axes_vec[joint0_axis], dof_ref0),
+            )
+
+        # Joint 1: T_body_b = T_body_a * X_p1 * J1(dof_ref1) * inv(X_c1)
+        X_p1 = wp.transform(
+            wp.vec3(
+                float(joint_X_p_np[joint1_idx][0]),
```
adenzler-nvidia, line 999: This ~25-line assertion block (read eq_data, construct 4 vec3s, assert 12 values, optional CPU check) appears identically three times in `_test_loop_joint_connect_constraint` — once for initial state, once after dof_ref change, once after joint_X_p change. Consider extracting a helper like:
```python
def _assert_connect_anchors(self, sim, world, expected_anchors, places=4):
    """Assert mjw_model.eq_data matches expected (anchor1_a, anchor2_a, anchor1_b, anchor2_b)."""
```
This would cut the method from ~200 lines to ~80, making the test structure (setup → assert → mutate → assert → mutate → assert) immediately visible.

newton/tests/test_equality_connect_constraint_with_sim_step.py
```diff
@@ -782,5 +783,717 @@ def _create_solver(self, model):
...
+
+        Bodies per world (3 total): root_body(0), body_a(1), body_b(2)
+        Joints per world (4 total): root_fixed(0), joint0(1), joint1(2), loop_joint(3)
+
+        Args:
+            loop_joint_axis: Axis for the loop revolute joint (0=X, 1=Y, 2=Z).
+            joint0_axis: Axis for joint0 (0=X, 1=Y, 2=Z).
+            joint1_axis: Axis for joint1 (0=X, 1=Y, 2=Z).
+            joint0_type: Type of joint0 (``"revolute"`` or ``"prismatic"``).
+            joint1_type: Type of joint1 (``"revolute"`` or ``"prismatic"``).
```
adenzler-nvidia, line 816: This method manually reimplements FK in ~120 lines, unpacking numpy transform rows element-by-element 6 times. The solver already has `_compute_body_poses_at_qref()` which is trusted, well-tested FK. Could that be used to get body poses, and then compute expected anchors in a short Python reference (~15 lines) rather than reimplementing the full FK chain by hand?

newton/tests/test_equality_connect_constraint_with_sim_step.py
```diff
@@ -782,5 +783,717 @@ def _create_solver(self, model):
...
+            wp.vec3(
+                float(joint_X_c_np[joint0_idx][0]),
+                float(joint_X_c_np[joint0_idx][1]),
+                float(joint_X_c_np[joint0_idx][2]),
+            ),
+            wp.quat(
+                float(joint_X_c_np[joint0_idx][3]),
+                float(joint_X_c_np[joint0_idx][4]),
+                float(joint_X_c_np[joint0_idx][5]),
+                float(joint_X_c_np[joint0_idx][6]),
```
adenzler-nvidia, line 983: The 4 joint-type combinations are tested inside a for-loop — if the first combo fails, the remaining three never run. Consider wrapping with `self.subTest(joint0=joint0_type, joint1=joint1_type)` so all combos are reported independently on failure.

adenzler-nvidia (commented):
Thanks for the fix — the WELD corruption bug fix (excluding WELD entries from `mjc_eq_to_newton_jnt`) is clean and correct. A few comments on the broader design and especially on simplifying the tests.

newton/_src/solvers/mujoco/solver_mujoco.py
gyeomannvidia: Done

newton/_src/solvers/mujoco/solver_mujoco.py
gyeomannvidia: Done

newton/_src/solvers/mujoco/solver_mujoco.py
gyeomannvidia: Done

newton/tests/test_equality_connect_constraint_with_sim_step.py
gyeomannvidia: Good idea.  Done.

newton/tests/test_equality_connect_constraint_with_sim_step.py
gyeomannvidia: Done

newton/_src/solvers/mujoco/kernels.py
gyeomannvidia: I'm not keen on reshaping the arrays in this PR. 

Tracked in 

https://github.com/newton-physics/newton/issues/2451

newton/_src/solvers/mujoco/solver_mujoco.py
gyeomannvidia: Wow, this was really ugly.  I separated has_connect_constraints from has_jnt_connect_constraints.

newton/tests/test_equality_connect_constraint_with_sim_step.py
gyeomannvidia: I deliberately avoided _compute_body_poses_at_qref() because it was introduced in this PR.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -2907,11 +2917,23 @@ def __init__(
...
         """Relative rotation ``inv(q2) * q1`` at the reference pose per equality constraint, ``wp.array[wp.quat]``, shape ``[equality_constraint_count]``."""
         self.connect_constraint_t_rel: wp.array | None = None
         """Relative translation [m] at the reference pose per equality constraint, ``wp.array[wp.vec3]``, shape ``[equality_constraint_count]``."""
 
+        # --- Internal state for anchor computation of joint-synthesized CONNECT constraints ---
+        self.jnt_eq_anchor1: wp.array2d[wp.vec3] | None = None
+        """Body1-local anchor [m] per ``[world, eq]`` for joint-synthesized CONNECT constraints, ``wp.array2d[wp.vec3]``, shape ``[world_count, neq]``."""
+        self.jnt_eq_anchor1_has_axis_offset: wp.array2d[wp.int32] | None = None
+        """Whether each ``[world, eq]`` entry is the second hinge CONNECT offset along the joint axis, ``wp.array2d[wp.int32]``, shape ``[world_count, neq]``."""
+        self.jnt_connect_constraint_q_rel: wp.array2d[wp.quatf] | None = None
```
adenzler-nvidia: nit: `wp.quatf` → `wp.quat` for consistency with the rest of the file and the docstring on this same line.

```suggestion
        self.jnt_connect_constraint_q_rel: wp.array2d[wp.quat] | None = None
```

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -5265,6 +5302,20 @@ def get_body_name(body_idx: int) -> str:
...
+            jnt_eq_anchor1_has_axis_offset_np = np.zeros((nworld, neq), dtype=np.int32)
+            for mjc_eq_id, anchor in jnt_eq_anchor1_dict.items():
+                has_offset = jnt_eq_anchor1_has_axis_offset.get(mjc_eq_id, False)
+                for w in range(nworld):
+                    jnt_eq_anchor1_np[w, mjc_eq_id, 0] = anchor[0]
+                    jnt_eq_anchor1_np[w, mjc_eq_id, 1] = anchor[1]
+                    jnt_eq_anchor1_np[w, mjc_eq_id, 2] = anchor[2]
+                    jnt_eq_anchor1_has_axis_offset_np[w, mjc_eq_id] = int(has_offset)
+            self.jnt_eq_anchor1 = wp.array(jnt_eq_anchor1_np, dtype=wp.vec3)
+            self.jnt_eq_anchor1_has_axis_offset = wp.array(jnt_eq_anchor1_has_axis_offset_np, dtype=wp.int32)
```
adenzler-nvidia, line 5326: suggestion: Since both `mjc_eq_to_newton_eq` and `mjc_eq_to_newton_jnt` write to the same `eq_data` array via separate kernels, it'd be worth asserting they never both map the same `[world, eq]` slot. If they overlap due to a construction bug, the two kernels would race on the same `eq_data` element on GPU — nondeterministic corruption that's very hard to diagnose.

```python
assert not np.any(
    (mjc_eq_to_newton_eq_np >= 0) & (mjc_eq_to_newton_jnt_np >= 0)
), "mjc_eq_to_newton_eq and mjc_eq_to_newton_jnt overlap — both kernels would write to the same eq_data slot"
```

adenzler-nvidia (dismissed):
Looks good — the WELD/CONNECT separation is clean and well-tested. Two minor items below.

newton/_src/solvers/mujoco/solver_mujoco.py
gyeomannvidia: Done

newton/_src/solvers/mujoco/solver_mujoco.py
gyeomannvidia: Done

newton/tests/test_equality_connect_constraint_with_sim_step.py
adenzler-nvidia: would it make sense to use mjwarp API then?

CHANGELOG.md
```diff
@@ -173,6 +173,11 @@
...
 - Clamp viewer picking force to prevent explosion when picking light objects near stiff contacts, configurable via `pick_max_acceleration` parameter on the `Picking` class (default 5g of effective articulation mass)
 - Fix `cloth_franka` example Jacobian broken by COM-referenced `body_qd` convention change; adjust robot base height, gripper orientations, and grasp targets for improved reachability (a follow-up PR will migrate the example to `newton.ik`)
+- Fix SDF hydroelastic broadphase scatter kernel using a grid-stride loop with binary search instead of per-pair thread launch
+- Fix `SensorRaycast` ignoring `PLANE` geometry
+- Fix joint-synthesized CONNECT constraint anchors not updating when `dof_ref` or `joint_X_p` changes at runtime via `notify_model_changed()`
+- Fix WELD constraint data corruption when a model contains both FIXED and revolute/ball loop joints, caused by CONNECT anchor kernels overwriting WELD `eq_data`
+- Fix box support-map sign flips from quaternion rotation noise (~1e-14) producing invalid GJK/MPR contacts for face-touching boxes with non-trivial base rotations
```
adenzler-nvidia: something seems off with the merge here.

gyeomannvidia: I did merge from main but there were no conflicts.  Very odd.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -4908,6 +4942,9 @@ def get_body_name(body_idx: int) -> str:
                 eq.name2 = child_name
                 eq.data[0:3] = joint_parent_xform[j][:3]
                 mjc_eq_to_newton_jnt[eq.id] = j
+                jnt_eq_anchor1_dict[eq.id] = list(joint_parent_xform[j][:3])
+                jnt_eq_anchor1_has_axis_offset[eq.id] = False
+                self.has_jnt_connect_constraints = True
```
adenzler-nvidia, line 4947: nit: `has_jnt_connect_constraints` is set in two separate branches (here and line 4910 for the hinge case). If a future loop joint type also produces CONNECT constraints but forgets to set this flag, the constraints would silently never update at runtime.

Consider deriving it from `jnt_eq_anchor1_dict` after the loop instead:
```python
self.has_jnt_connect_constraints = len(jnt_eq_anchor1_dict) > 0
```
More maintainable and automatically correct for new joint types.

newton/tests/test_equality_connect_constraint_with_sim_step.py
```diff
@@ -782,5 +783,682 @@ def _create_solver(self, model):
         )
 
 
+class TestLoopJointConnectConstraintBase(TestEqualityConstraintWithSimStepBase):
+    """Test that loop-joint-synthesized CONNECT constraints update when dof_ref changes.
+
+    Creates two articulations connected by a revolute loop joint. The loop
```
adenzler-nvidia: nit: "Creates two articulations connected by a revolute loop joint" — the test actually creates a **single** articulation with a loop joint closing back to its root body (as `_build_loop_joint_model` correctly documents). Suggest:

```
Creates a single articulation with a revolute loop joint closing back to
its root body. The loop joint generates 2 CONNECT constraints in MuJoCo.
```

newton/tests/test_equality_connect_constraint_with_sim_step.py
```diff
@@ -782,5 +783,682 @@ def _create_solver(self, model):
...
+            integrator="euler",
+        )
+
+
+class TestMixedWeldAndConnectLoopJointBase(TestEqualityConstraintWithSimStepBase):
+    """Test that WELD (FIXED) loop joint eq_data is not corrupted by CONNECT kernel updates.
+
+    Creates a model with both a revolute loop joint (2 CONNECT constraints) and
+    a FIXED loop joint (1 WELD constraint).  Verifies that after
+    ``notify_model_changed(JOINT_DOF_PROPERTIES)`` the WELD constraint's
```
adenzler-nvidia, line 1280: nit: This says the test verifies state "after `notify_model_changed(JOINT_DOF_PROPERTIES)`", but `test_weld_eq_data_not_corrupted_by_connect_update` actually validates state after solver construction (which calls `notify_model_changed(ALL)` internally). The test method's own docstring gets this right ("The corruption happens at init time"). Suggest aligning:

```
Verifies that after solver initialization (which calls
``notify_model_changed(ALL)`` internally) the WELD constraint's
``eq_data`` retains its anchor and relpose values.
```

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -6027,7 +6087,8 @@ def _notify_connect_constraints_changed(
 
         Optionally recomputes the relative body transforms at the reference
```
adenzler-nvidia, line 6088: nit: The `update_anchors` parameter (line 6099) only triggers recomputation for explicit Newton equality constraints, not joint-synthesized CONNECT constraints (those only respond to `update_anchor_rel_xform_at_ref_pose`). Worth noting in the docstring to prevent confusion:

```
update_anchors: Recompute anchors from
    ``equality_constraint_anchor``. Only affects explicit Newton
    equality constraints, not joint-synthesized CONNECT constraints.
```

adenzler-nvidia (commented):
Thanks for this fix — the approach of excluding WELD entries from `mjc_eq_to_newton_jnt` is clean and the new kernel/test structure is solid. A few minor suggestions below.

adenzler-nvidia (dismissed):
All feedback from previous reviews addressed or deferred with tracking issues (#2451).

The two new commits look good:
- Overlap assertion (`mjc_eq_to_newton_eq_np` vs `mjc_eq_to_newton_jnt_np`) is a solid safety net against future regressions
- `wp.quatf` → `wp.quat` nit fixed
- CHANGELOG merge artifacts cleaned up

Core fix is sound — excluding WELD entries from `mjc_eq_to_newton_jnt` prevents CONNECT anchor kernels from overwriting WELD `eq_data`, and the negative-index guard ensures WELD slots are never touched.

One open minor thread (test FK reimplementation vs mjwarp API) — not blocking.

LGTM, approving contingent on CI passing.

adenzler-nvidia (approved):
LGTM -- both nits addressed (wp.quat annotation, disjoint mapping assertion). The two new commits are clean: the ls_iterations cleanup removes unnecessary coupling in the test, and the merge from main has no conflicts. CI green.

PR #2433, opened 2026-04-14
Add array heatmap logging to ViewerGL

han-xudong:
## Description

Add support for visualizing logged numeric arrays in `ViewerGL`.

This implements `ViewerGL.log_array()` for scalar, 1-D, and 2-D numeric arrays.
2-D arrays are rendered as heatmaps in the existing ImGui diagnostics overlay,
alongside the current scalar plots.

This makes it easier to inspect grid-based or field-like debug data directly in
the GL viewer.

## Checklist

## Test plan

Manual verification in a local `ViewerGL` session.

Verified that:
- `ViewerGL.log_array()` accepts scalar inputs and displays them as a 1x1 array
- `ViewerGL.log_array()` accepts 1-D arrays and displays them as a single-row heatmap
- `ViewerGL.log_array()` accepts 2-D arrays and displays them as a heatmap
- array metadata (shape, min, max) is shown in the diagnostics overlay
- scalar plots continue to render correctly alongside logged arrays
- passing `None` removes a previously logged array

## New feature / API change

```python
import numpy as np

viewer.log_array("Scalar", np.array(1.0))
viewer.log_array("Series", np.array([1.0, 2.0, 3.0]))
viewer.log_array("Heatmap", np.array([[0.0, 1.0], [2.0, 3.0]]))
viewer.log_array("Heatmap", None)
```

## Summary by CodeRabbit

* **New Features**
  * Added heatmap visualization for logged 2D numeric arrays in the Plots window.
  * Supports adding, updating, and removing arrays; heatmaps refresh when data or display shape change.
  * Large arrays are downsampled for efficient display; NaN/non-finite values use a distinct color and an Inferno-like palette.
  * Shows shape/min/max overlays and provides collapsible per-array panels with dynamic sizing.

* **Bug Fixes**
  * Improved GPU/resource cleanup when arrays are removed or replaced.


eric-heiden:
Nice addition! Having inline heatmap visualization for logged arrays is a great debugging tool for ViewerGL.

The main concern is that the per-cell rendering loop runs entirely in Python every frame, which will bottleneck the viewer at moderate array sizes.

newton/_src/viewer/viewer_gl.py
```diff
@@ -2325,18 +2342,85 @@ def _edit_color3(
...
+            value_max = 0.0
+
+        denom = max(value_max - value_min, 1.0e-8)
+        draw_list = imgui.get_window_draw_list()
+        origin = imgui.get_cursor_screen_pos()
+
+        for row in range(rows):
+            for col in range(cols):
+                value = float(array[row, col])
+                if np.isfinite(value):
```
eric-heiden: 🟡 This nested loop runs `rows × cols` Python-level draw calls every frame. For a 64×64 array that is 4,096 iterations of `float()`, `np.isfinite()`, `_heatmap_color()`, two `ImVec2` constructions, and an `add_rect_filled()` call -- enough to drop below interactive frame rates.

`_heatmap_color` makes it worse by calling `np.clip` on a single scalar each iteration instead of plain `min`/`max`.

**Suggested fix (two levels):**

| Level | Change | Effort |
|---|---|---|
| Quick win | Replace `np.clip(value, ...)` in `_heatmap_color` with `max(0.0, min(1.0, value))`, and add a size cap that down-samples arrays larger than 128×128 before rendering | Small |
| Proper fix | Vectorize normalization and color mapping across the whole array in NumPy, rasterize into an RGBA texture, and draw it with a single `imgui.image()` call | Medium |

The proper fix moves all O(rows×cols) work out of Python and into one GPU texture upload per frame, which keeps the viewer responsive regardless of array size.

<details>

<summary>Quick-win example for <code>_heatmap_color</code></summary>

```python
# before
normalized = np.clip(value, 0.0, 1.0)

# after -- avoids numpy overhead on a single float
normalized = max(0.0, min(1.0, value))
```

</details>

han-xudong:
@eric-heiden Thanks for the thoughtful review and for calling out the Python-side rendering cost here. You were absolutely right that the per-cell draw loop would become a bottleneck at moderate array sizes.

I reworked the heatmap path toward the “proper fix” you suggested: the data is now normalized and color-mapped in NumPy, downsampled to the displayed resolution, uploaded as an RGBA texture, and rendered with a single `imgui.image()` call instead of one Python draw call per cell. That also let me add texture cleanup when arrays are replaced or removed.

I also folded in a few follow-up fixes that came up during review, including some CodeRabbit suggestions:
- preserving 2-D column vectors as column heatmaps instead of collapsing them through `squeeze()`
- reporting `min -- / max --` for all-non-finite inputs rather than showing misleading `0 / 0`
- fixing the OpenGL texture allocation call to use the correct pyglet ctypes pattern
- cleaning up a small redundant `round()` cast nit

Thanks again for the push here. The performance concern was the right one to fix properly.

newton/_src/viewer/viewer_gl.py
```diff
@@ -2325,18 +2378,192 @@ def _edit_color3(
...
+        downsampled = np.full((new_rows, new_cols), np.nan, dtype=np.float32)
+        np.divide(value_sum, value_count, out=downsampled, where=value_count > 0)
+        return downsampled
+
+    def _colorize_heatmap(self, array: np.ndarray) -> tuple[np.ndarray, float, float]:
+        finite_mask = np.isfinite(array)
+        if not np.any(finite_mask):
+            rgba = np.empty((*array.shape, 4), dtype=np.uint8)
+            rgba[...] = self._heatmap_nan_rgba
+            return np.ascontiguousarray(rgba[::-1]), float("nan"), float("nan")
```
eric-heiden: 🟡 The `rgba[::-1]` here flips the heatmap vertically before upload, so the top row of the source array is rendered at the bottom of the image. `imgui.image()` with the default UVs already maps texture row 0 to the top of the rect, so this inverts the array relative to the usual NumPy/Matplotlib orientation. The same reversal is present in the non-NaN return below. Unless `origin='lower'` is intentional, return `np.ascontiguousarray(rgba)` in both places, or do the flip through `uv0`/`uv1` and document why.

han-xudong: @eric-heiden Thanks, good catch. You were right that the extra `rgba[::-1]` was flipping the heatmap vertically. I removed the reversal in both return paths so the texture upload now preserves the usual array orientation.

eric-heiden (approved):
Thanks, LGTM!

PR #2435, opened 2026-04-14
Add enable_multiccd parameter to SolverMuJoCo

### LINKED ISSUE #2400: [REQ]  Expose `enableflags` in `SolverMuJoCo._convert_to_mjc` for `multiccd` support
### Current state

`SolverMuJoCo._convert_to_mjc` exposes `disableflags` (via the `opt_disableflags` parameter) but not `enableflags`. This prevents enabling `mjENBL_MULTICCD`, which generates up to 4 CCD contact points per geom pair instead of 1.

Without `multiccd`, a single MESH geom resting on a flat surface gets exactly 1 CCD contact point. Under any perturbation (e.g., constraint forces from a nearby articulated robot), the object pivots around that single contact and falls through the surface.

`SolverMuJoCo._convert_to_mjc` (line ~1800 in `solver_mujoco.py`) reads `opt_disableflags` from the solver config:

```python
# In _convert_to_mjc:
if self._disableflags is not None:
    self.mj_model.opt.disableflags = self._disableflags
```

There is no equivalent for `enableflags`. MuJoCo's `mjENBL_MULTICCD` flag (which generates up to 4 CCD contacts per MESH geom pair) can only be set if `enableflags` is exposed.

### Request

Users should be able to set MuJoCo `enableflags` (e.g., `mjENBL_MULTICCD`) through the solver configuration, the same way `disableflags` is already supported.

### Motivation / Use Case

Any manipulation task using MESH collision geoms with the MuJoCo contact pipeline (`use_mujoco_contacts=True`) is affected. This is a common configuration for:
- Dexterous manipulation with USD-loaded object meshes
- Scenes with multiple mesh-mesh contact pairs (e.g., peg insertion with a pegboard)

### Alternatives Considered

The workaround (switching to Newton's SDF collision pipeline) works but forfeits MuJoCo-native contact features.

vreutskyy:
## Description

Add `enable_multiccd` boolean parameter to `SolverMuJoCo` that enables MuJoCo's `mjENBL_MULTICCD` flag, generating up to 4 CCD contact points per mesh geom pair instead of 1. This prevents mesh objects from pivoting around a single contact and falling through surfaces under perturbation.

Follows the same pattern as the existing `disable_contacts` parameter — a Newton-level boolean that maps to a MuJoCo flag internally, keeping MuJoCo constants out of Newton's public API.

Closes #2400

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_enable_multiccd_passed_to_mujoco
```

## New feature / API change

```python
import newton

# ... build model ...
model = builder.finalize()

solver = newton.solvers.SolverMuJoCo(model, enable_multiccd=True)
```

## Summary by CodeRabbit

* **New Features**
  * Added an option to enable multi-CCD contact generation, allowing up to four contact points per mesh/geom pair for improved collision fidelity.

* **Tests**
  * Added tests ensuring the multi-CCD option is off by default and that enabling it activates the corresponding solver behavior.

* **Documentation**
  * Updated changelog to document the new multi-CCD option.


eric-heiden (dismissed):

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -2817,6 +2818,7 @@ def __init__(
             wind: Wind velocity vector (x, y, z) for lift and drag forces. If None, uses model custom attribute or MuJoCo's default (0, 0, 0).
             magnetic: Global magnetic flux vector (x, y, z). If None, uses model custom attribute or MuJoCo's default (0, -0.5, 0).
```
adenzler-nvidia, line 2819: "per mesh geom pair" is slightly too narrow — MuJoCo's `maxContacts()` in `engine_collision_convex.c` applies the 4-contact limit to any pair where both geoms are `mjGEOM_BOX` or `mjGEOM_MESH` (so box-mesh pairs also benefit, not just mesh-mesh). The MuJoCo XML reference says "geom pairs that use a general-purpose convex collider *e.g.,* mesh-mesh collisions" — mesh-mesh is given as an example, not the full scope.

Suggestion: replace "per mesh geom pair" with "per geom pair" here, in `_convert_to_mjc`, and in the CHANGELOG entry.

Also worth noting: `maxContacts()` returns 1 if either geom has `margin > 0`, silently disabling multi-contact regardless of the flag. Consider mentioning this non-obvious interaction in the docstring.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -3755,6 +3762,7 @@ def _convert_to_mjc(
             nconmax: Maximum number of contacts.
             solver: Constraint solver type ("cg" or "newton"). If None, uses model custom attribute or Newton's default ("newton").
```
adenzler-nvidia, line 3763: Nit: this line includes an `(e.g. ...)` example, but the adjacent `disableflags` docstring on the next line is just `"MuJoCo disable flags bitmask."` with no example. For consistency, consider either dropping the example here or adding one to `disableflags`.

newton/tests/test_mujoco_solver.py
```diff
@@ -6377,6 +6377,16 @@ def test_iterations_constructor_override(self):
         self.assertEqual(solver.mj_model.opt.iterations, 5, "Constructor value should override custom attribute")
         self.assertEqual(solver.mj_model.opt.ls_iterations, 3, "Constructor value should override custom attribute")
```
adenzler-nvidia, line 6466: The happy path is well-tested. Consider adding a small companion test verifying the flag is *not* set by default — it guards against accidental default flips in future refactors:

```python
def test_enable_multiccd_default_off(self):
    """Verify that mjENBL_MULTICCD is not set by default."""
    model = self._create_multiworld_model(world_count=1)
    solver = SolverMuJoCo(model)
    mujoco = SolverMuJoCo._mujoco
    self.assertFalse(
        solver.mj_model.opt.enableflags & mujoco.mjtEnableBit.mjENBL_MULTICCD,
        "mjENBL_MULTICCD should not be set when enable_multiccd is not specified",
    )
```

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@
 - Add `total_force_friction` and `force_matrix_friction` to `SensorContact` for tangential (friction) force decomposition
 - Add `compute_normals` and `compute_uvs` optional arguments to `Mesh.create_heightfield()` and `Mesh.create_terrain()`
 - Add RJ45 plug-socket insertion example with SDF contacts, latch joint, and interactive gizmo
+- Add `enable_multiccd` parameter to `SolverMuJoCo` for multi-CCD contact generation (up to 4 contact points per mesh geom pair)
```
adenzler-nvidia: Same note as on the docstring — "per mesh geom pair" should be "per geom pair" since the multi-contact behavior also applies to box-mesh pairs, not just mesh-mesh.

adenzler-nvidia (commented):
Clean, well-structured PR that follows the existing `disable_contacts` pattern nicely. A few comments below — one on technical accuracy, the rest are minor suggestions.

newton/_src/solvers/mujoco/solver_mujoco.py
vreutskyy: Good catch — fixed to "per geom pair" and added the margin > 0 caveat to the docstring.

newton/_src/solvers/mujoco/solver_mujoco.py
vreutskyy: Agreed — dropped the example to match `disableflags` style.

newton/tests/test_mujoco_solver.py
vreutskyy: Added `test_enable_multiccd_default_off` — thanks for the suggestion.

CHANGELOG.md
vreutskyy: Fixed — updated to "per geom pair" here as well.

adenzler-nvidia (approved):

Issue #2422, opened 2026-04-13
[BUG] Friction combination logic differs from `mujoco_warp`


nvtw:
### Bug Description

In `mujoco_warp`'s `contact_params` (`collision_core.py`), friction is resolved as follows:
- **Different priority**: use the higher-priority geom's friction directly
- **Same priority**: element-wise `max` of both geoms' friction

However, the Newton wrapper's `contact_params` unconditionally takes the element-wise `max` regardless of priority:

```python
# Newton wrapper — priority is ignored for friction
max_geom_friction = wp.max(geom_friction[worldid, g1], geom_friction[worldid, g2])
```

### Reproduction Script

See https://github.com/newton-physics/newton/discussions/2377

PR #2439, opened 2026-04-14
Fix material combination inconsistencies

nvtw:
## Description

MjWarp currently uses a different material combination logic. This MR applies the same logic into the Newton-to-MjWarp contact converter. See https://github.com/newton-physics/newton/issues/2422 and https://github.com/newton-physics/newton/discussions/2377

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Contact friction and parameter resolution in the MuJoCo solver now respect geometry priority ordering, correctly selecting higher-priority geometry values and applying component-wise clamping and stable mixing behavior for equal priorities.
* **Tests**
  * Added tests validating friction selection when priorities differ and element-wise mixing when priorities are equal.

### DIFF
--- a/newton/_src/solvers/mujoco/kernels.py
+++ b/newton/_src/solvers/mujoco/kernels.py
@@ -27,9 +27,15 @@ vec11 = wp.types.vector(length=11, dtype=wp.float32)
 
 # Constants
 MJ_MINVAL = 2.220446049250313e-16
+MJ_MINMU = 1e-5
 
 
 # Utility functions
+@wp.func
+def safe_div(x: float, y: float) -> float:
+    return x / wp.where(y != 0.0, y, MJ_MINVAL)
+
+
 @wp.func
 def orthogonals(a: wp.vec3):
     y = wp.vec3(0.0, 1.0, 0.0)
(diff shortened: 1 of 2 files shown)


adenzler-nvidia (approved):
Looks good — the priority/solmix/friction logic now matches MjWarp's `collision_core.contact_params` and the tests cover both the unequal-priority and equal-priority paths well.

Minor suggestions (non-blocking):
- Consider importing `MJ_MINMU` from `mujoco_warp._src.types` instead of hardcoding `1e-5`, so it stays in sync if MuJoCo ever changes the constant.
- Same for `safe_div` — could import from `mujoco_warp._src.math` to avoid the local copy, though it's trivial enough that either way is fine.

adenzler-nvidia (approved):
Looks good — the priority/solmix/friction logic now matches MjWarp's `collision_core.contact_params` and the tests cover both the unequal-priority and equal-priority paths well.

Longer-term, it might be worth factoring out the shared geom-level mixing logic into a common `@wp.func` upstream in MjWarp so both codebases stay in sync by construction, but that's outside the scope of this PR.

Issue #2248, opened 2026-03-28
[REQ] Stable rigid-contact manifold IDs for warmstarting and persistent friction


eric-heiden:
### Description

  The current rigid-contact output is frame-local rather than manifold-stable.

  `Contacts` allocates `rigid_contact_point_id`, but the rigid contact write path currently only stores shape IDs, body-local contact points, offsets, normals, and margins for the current frame. The same physical contact can move to a different slot or be replaced by a
  nearby point after manifold generation/reduction. As a result, solvers cannot reliably reuse previous-frame contact state.

  This shows up as:
  - poor settling / jitter in resting contacts
  - loss of static friction and post-impact sliding, especially visible for robotics use cases such as fallen G1 humanoids
  - extra solver iterations required because contact warm-starting is effectively impossible
  - inconsistent friction anchors when manifold points are re-ordered or replaced

  This affects any rigid-body solver that would benefit from persistent contact state, including `SolverSemiImplicit`, `SolverFeatherstone`, and `SolverMABD`.

  ### Current state in the pipeline

  - `newton/_src/sim/contacts.py` allocates `rigid_contact_point_id`, but it is currently unused.
  - `newton/_src/sim/collide.py` writes rigid contacts directly into the frame-local output buffers.
  - The collision output includes useful geometric data such as contact offsets and effective thickness/margins, but it does not currently expose a stable per-contact manifold key that survives reordering across frames.

  ### Proposed change

  1. Extend narrow phase / manifold generation to emit a stable per-contact key, not just the current output slot.
  2. Populate a rigid contact key/ID buffer in `Contacts`.
     - This could reuse `rigid_contact_point_id` or replace it with a wider key buffer if needed.
  3. Derive the key from shape pair + stable feature identifiers where available.
     - Examples: primitive face/edge/vertex ID, mesh triangle/edge ID, capsule segment endpoint/band, plane face ID.
  4. Add a fallback local-space patch key when exact feature IDs are unavailable.
     - For example, quantized local contact points on shape A/B hashed together with the shape pair.
  5. Preserve stable contact ordering per pair/manifold.
     - Keep a small per-pair manifold cache.
     - Match current contacts to previous contacts by key first, then nearest local-space point as a fallback.
  6. Expose the resulting key to solvers so they can warm-start:
     - accumulated normal impulses
     - accumulated tangential/static-friction impulses
     - torsional and rolling friction impulses
     - friction anchor state

### Motivation / Use Case

[...]

PR #2446, opened 2026-04-15
Opt-in support for contact matching - simplifies solver warm starting

nvtw:
## Description

Add frame-to-frame contact matching to `CollisionPipeline`. Each frame, contacts are matched against the previous frame's sorted contacts using binary search on deterministic sort keys, then verified against world-space position distance and surface normal dot-product thresholds. The result is a per-contact `match_index` on `Contacts` (`>= 0` matched, `-1` new, `-2` broken). Optional contact reports list newly formed and broken contact indices.

Enabled via `CollisionPipeline(contact_matching=True)`, which implies `deterministic=True` because they use the same underlying sort and the same contact fingerprints.

See https://github.com/newton-physics/newton/issues/2248

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_contact_matching
```

14 tests covering: first-frame sentinel values, stable-scene identity matching across multiple frames, new contact detection, position and normal threshold breaking, contact report index correctness, deterministic flag implication, and disabled-matching allocation.

## New feature / API change

```python
import newton

    model,
    contact_matching=True,
    contact_matching_pos_threshold=0.02,       # meters
    contact_matching_normal_dot_threshold=0.9,  # cos(~25°)
    contact_report=True,                        # optional new/broken lists
)

# Per-contact match index: >= 0 matched, -1 new, -2 broken
match_idx = contacts.rigid_contact_match_index.numpy()

# Optional: indices of new and broken contacts
matcher = pipeline.contact_matcher
new_indices = matcher.new_contact_indices.numpy()[:matcher.new_contact_count.numpy()[0]]
broken_indices = matcher.broken_contact_indices.numpy()[:matcher.broken_contact_count.numpy()[0]]
```

## Summary by CodeRabbit

* **New Features**
  * Frame-to-frame contact matching with configurable position and normal thresholds; per-contact match indices exposed (matched, new, broken).
  * Optional contact reporting that emits compact lists of newly formed and broken contacts.
  * Enabling contact matching forces deterministic contact ordering.

* **Documentation**
  * Added guidance and examples for enabling contact matching/reporting and interpreting match/report outputs.

* **Tests**
  * New test suite validating matching behavior, thresholds, reporting, and deterministic behavior.


newton/_src/geometry/contact_sort.py
```diff
@@ -205,6 +219,7 @@ def __init__(self, capacity: int, *, per_contact_shape_properties: bool = False,
             self._simple_normal_buf = wp.zeros(capacity, dtype=wp.vec3)
             self._simple_penetration_buf = wp.zeros(capacity, dtype=float)
             self._simple_tangent_buf = wp.zeros(capacity, dtype=wp.vec3)
+            self._simple_match_index_buf = wp.zeros(1, dtype=wp.int32)
```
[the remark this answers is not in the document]
nvtw: This should not be an issue

newton/_src/sim/collide.py
```diff
@@ -937,6 +976,20 @@ def collide(
...
+            self._contact_matcher.match(
+                sort_keys=self._sort_key_array,
+                contact_count=contacts.rigid_contact_count,
+                point0=contacts.rigid_contact_point0,
+                shape0=contacts.rigid_contact_shape0,
+                normal=contacts.rigid_contact_normal,
+                body_q=state.body_q,
+                shape_body=model.shape_body,
+                match_index_out=contacts.rigid_contact_match_index,
+                device=self.device,
```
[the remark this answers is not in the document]
nvtw: Should be addressed now

newton/_src/geometry/contact_match.py
```diff
@@ -0,0 +1,511 @@
...
+
+        old_pos = data.prev_pos_world[old_idx]
+        diff = new_pos_w - old_pos
+        dist_sq = wp.dot(diff, diff)
+        old_n = data.prev_normal[old_idx]
+        ndot = wp.dot(new_n, old_n)
+
+        if dist_sq < best_dist_sq and ndot > data.normal_dot_threshold:
+            best_dist_sq = dist_sq
+            best_idx = old_idx
```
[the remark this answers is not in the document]
nvtw: Should be addressed now

newton/_src/sim/contacts.py
```diff
@@ -188,6 +192,18 @@ def __init__(
...
+            self.contact_matching = contact_matching
+            if contact_matching:
+                self.rigid_contact_match_index = wp.full(rigid_contact_max, -1, dtype=wp.int32)
+                """Per-contact match index from :class:`ContactMatcher`.
+
+                Values: ``>= 0`` matched old contact index, ``-1`` new contact,
+                ``-2`` key matched but thresholds exceeded (broken).
+                Shape (rigid_contact_max,), dtype int32."""
+            else:
+                self.rigid_contact_match_index = None
```
camevor, line 218: Would it make sense to make this an extended contact attribute?

docs/concepts/collisions.rst
```diff
@@ -1669,6 +1673,91 @@ fully CUDA-graph-capturable.
...
+        model,
+        contact_matching=True,
+        contact_report=True,
+    )
+    contacts = pipeline.contacts()
+    pipeline.collide(state, contacts)
+
+    matcher = pipeline.contact_matcher
+
+    n_new = matcher.new_contact_count.numpy()[0]
```
camevor: Since `rigid_contact_match_index` is owned by the `Contacts` container, would it make sense to also allocate these on `Contacts`? This would remove the `contact_matcher` from the public API.

newton/_src/geometry/contact_match.py
```diff
@@ -0,0 +1,511 @@
...
+
+class ContactMatcher:
+    """Frame-to-frame contact matching using binary search on sorted keys.
+
+    Pre-allocates all buffers at construction time for CUDA graph capture
+    compatibility.  The typical per-frame call sequence is::
+
+        matcher.match(...)  # before ContactSorter.sort_full()
+        sorter.sort_full(...)  # match_index is permuted with contacts
+        matcher.save_sorted_state(...)  # after sorting
```
jumyungc: Should save_sorted_state go after builld_report?

jumyungc (commented):
Looks good to me, and looking forward to using it!

newton/_src/geometry/contact_match.py
```diff
@@ -0,0 +1,496 @@
...
+        self,
+        capacity: int,
+        *,
+        sorter: ContactSorter,
+        pos_threshold: float = 0.02,
+        normal_dot_threshold: float = 0.9,
+        contact_report: bool = False,
+        device: Devicelike = None,
+    ):
+        with wp.ScopedDevice(device):
```
adenzler-nvidia, line 451: Q on CPU support: all tests use `get_cuda_test_devices()`, and the sorter depends on `radix_sort_pairs` (CUDA-only in practice), but the matcher's own kernels don't depend on anything CUDA-specific. Is CPU intentionally unsupported, or just inherited from the sorter? If intentional, a runtime check here (when `device` is CPU) would be friendlier than a later kernel crash — something like:

```python
if not wp.get_device(device).is_cuda:
    raise RuntimeError("ContactMatcher requires a CUDA device (radix_sort_pairs is CUDA-only).")
```

(Or keep silent and document it as a known limitation — either works; the current state is the least kind.)

newton/_src/geometry/contact_match.py
```diff
@@ -0,0 +1,496 @@
...
+            self._capacity = capacity
+            self._pos_threshold_sq = pos_threshold * pos_threshold
+            self._normal_dot_threshold = normal_dot_threshold
+            self._sorter = sorter
+
+            # Only buffer we must own: sorted keys survive across frames
+            # (_sort_keys_copy is overwritten by _prepare_sort each frame).
+            self._prev_sorted_keys = wp.zeros(capacity, dtype=wp.int64)
+            self._prev_count = wp.zeros(1, dtype=wp.int32)
```
adenzler-nvidia, line 463: Cross-episode reset semantics question: `_prev_count` persists across `collide()` calls (the whole point). But in RL-style workflows where a user wants to "start fresh" after `state.reset()` or teleporting all bodies, there's no way to zero `_prev_count` without rebuilding the pipeline. `Contacts.clear()` resets `match_index=-1` but the matcher still has old previous-frame data, so the next frame produces spurious matches against bodies in their new poses.

Would you consider either (a) a `ContactMatcher.reset()` that zeros `_prev_count` (+ optionally `_prev_was_matched`), or (b) having `Contacts.clear()` coordinate with the matcher when `contact_matching=True`? If out of scope, a `.. note::` in the class docstring flagging the persistence behavior would help users notice.

newton/_src/geometry/contact_match.py
```diff
@@ -0,0 +1,496 @@
...
+    ):
+        with wp.ScopedDevice(device):
+            self._capacity = capacity
+            self._pos_threshold_sq = pos_threshold * pos_threshold
+            self._normal_dot_threshold = normal_dot_threshold
+            self._sorter = sorter
+
+            # Only buffer we must own: sorted keys survive across frames
+            # (_sort_keys_copy is overwritten by _prepare_sort each frame).
+            self._prev_sorted_keys = wp.zeros(capacity, dtype=wp.int64)
```
adenzler-nvidia: Nit: `wp.zeros(capacity, dtype=wp.int64)` — harmless but prev-keys are always overwritten by the first `save_sorted_state` before being read, so the initial zero values are never consumed. Initializing with `SORT_KEY_SENTINEL` from contact_sort.py would make a debug inspection of the buffer before the first `save_sorted_state` less confusing (zeros look like valid keys for shape_a=0, shape_b=0). Very minor.

newton/_src/geometry/contact_match.py
```diff
@@ -0,0 +1,496 @@
...
+                Written directly (no intermediate copy).
+            device: Device to launch on.
+        """
+        if self._has_report:
+            self._prev_was_matched.zero_()
+
+        data = _MatchData()
+        data.prev_keys = self._prev_sorted_keys
+        # Reuse sorter scratch buffers for prev-frame world-space data.
+        data.prev_pos_world = self._sorter._full_point0_buf
```
adenzler-nvidia: This (plus lines 382, 440, 441) reaches into `ContactSorter`'s private `_full_point0_buf` / `_full_normal_buf`. The scratch-reuse optimization itself is great and well-motivated in the module docstring — the leaky abstraction is the only concern. Would a narrow pair of `ContactSorter` properties (e.g. `scratch_pos_world`, `scratch_normal`) — public names, maybe with a short "for use by ContactMatcher; do not write outside the documented window" comment — make the coupling explicit and refactor-safe? Feel free to push back if the underscore convention already communicates enough.

newton/_src/geometry/contact_match.py
```diff
@@ -0,0 +1,496 @@
...
+
+from ..core.types import Devicelike
+
+if TYPE_CHECKING:
+    from .contact_sort import ContactSorter
+
+MATCH_NOT_FOUND = wp.constant(wp.int32(-1))
+"""Sentinel: no matching key found in last frame's contacts."""
+
+MATCH_BROKEN = wp.constant(wp.int32(-2))
```
adenzler-nvidia, line 94: Nit: `MATCH_NOT_FOUND` / `MATCH_BROKEN` are module-level `wp.constant`s but aren't re-exported through `newton.geometry` or the top-level. Users who'd rather write `match_idx == MATCH_NOT_FOUND` than `match_idx == -1` can't import them. Documenting the numeric values on the public `rigid_contact_match_index` docstring is already done — re-export is optional, but would make user code self-documenting.

newton/tests/test_contact_matching.py
```diff
@@ -0,0 +1,505 @@
...
+
+def test_broken_pos_threshold_all_contacts(test, device):
+    """Moving all spheres beyond pos_threshold must break ALL contacts (not just some)."""
+    with wp.ScopedDevice(device):
+        model, state = _build_simple_scene(device)
+        pipeline = newton.CollisionPipeline(
+            model,
+            broad_phase="nxn",
+            contact_matching=True,
+            contact_matching_pos_threshold=0.001,  # 1 mm — very tight
```
adenzler-nvidia: Using the tight `0.001` threshold here (vs the default `0.02`) means the test silently becomes a pass if the default ever gets re-tuned. Could this be rewritten to exercise the default? The scene uses an infinite ground plane, so shifting spheres by e.g. `0.2 m` (10× the default `0.02`) still keeps them in contact while exceeding the position threshold — same invariant, and if someone later changes the default the test follows along:

```python
    model,
    broad_phase="nxn",
    contact_matching=True,
    # rely on default contact_matching_pos_threshold (0.02 m)
)
...
# Shift all dynamic bodies by 0.2 m (10x default threshold).
for i in range(len(q)):
    q[i][0] += 0.2
```

Same story for the `# 10x the threshold` comment below at line 178.

adenzler-nvidia (commented):
Nice feature — well-scoped opt-in with strong invariant tests. The scratch-buffer reuse (point0/normal between sort_full and next-frame match) is genuinely clever and well-documented, and the deterministic binary-search-plus-verify shape is the right choice (O(n log n), graph-capture-safe). Graph-capture discipline throughout (preallocated buffers, `dim=self._capacity` launches, writing `MATCH_NOT_FOUND` to inactive slots so the gather-permuted result is well-defined on trailing capacity).

A few questions inline — mostly API-shape / doc-clarity, nothing blocking. One reminder: `mergeStateStatus: DIRTY` (40 ahead / 22 behind upstream/main), so you'll want to rebase before merge.

Independent of the inline threads, also worth engaging with:
- **camevor's API-shape question**: should `rigid_contact_match_index` be an extended contact attribute (`EXTENDED_ATTRIBUTES` + `requested_attributes={"match_index"}` like `"force"`), and should `new/broken_contact_indices` live on `Contacts` when `contact_report=True` so `contact_matcher` stays internal? The current split (match_index on Contacts, index lists on ContactMatcher) is asymmetric, and surfacing `pipeline.contact_matcher.new_contact_indices` leaks a lot of internal state (`_prev_sorted_keys`, `_prev_was_matched`, `_prev_count` are implementation details). Worth a discussion before locking the public shape.
- **jumyungc's `save_sorted_state` vs `build_report` ordering question** (line 266): the current order is correct — `save_sorted_state` overwrites `_prev_count` with the new count, and `_collect_broken_contacts_kernel` needs the *old* `_prev_count` to know how many `_prev_was_matched` slots to scan. There's a good comment at collide.py:1020-1021 explaining this. Could you reply with that rationale and lift a shorter form into the `ContactMatcher` class docstring (under "typical per-frame call sequence"), so the ordering constraint is visible from the class API?
- **Key-uniqueness invariant**: the matcher relies on `sort_key = (shape_a, shape_b, sub_key)` being unique per active contact. `make_contact_sort_key` silently masks overflow (mesh-triangle contacts drop to 19 effective bits → ~524K triangles after the `<<3` multi-contact expansion). In scenes exceeding those bit budgets two new contacts could map to the same key, the binary search lands on the same old range, both pick the same `best_idx` → duplicate `match_index` values plus an `atomic_max` on `prev_was_matched` conflating them. The invariant is inherited from the sorter, fine to treat as out of scope — but worth a one-liner in the module docstring so the next reader knows why `match_index` is assumed unique.
- **Broken-on-both-sides test missing**: `test_broken_pos_threshold_all_contacts` verifies the new side gets `-2` but not that the same old contact appears in `broken_contact_indices`. Combined assertion would close the most interesting semantic gap.
- `has_report: int` is the codebase's int-as-bool convention; no change needed — noted for future cleanup when Warp gains native struct-field bools.

newton/_src/sim/collide.py
```diff
@@ -477,6 +502,11 @@ def __init__(
         narrow_phase: NarrowPhase | None = None,
         sdf_hydroelastic_config: HydroelasticSDF.Config | None = None,
         deterministic: bool = False,
+        contact_matching: Literal["disabled", "latest", "sticky"]
+        | CollisionPipeline.ContactMatching = ContactMatching.DISABLED,
```
eric-heiden: Can't we just have `Literal["disabled", "latest", "sticky"]`? I don't see why we need an extra enum definition for this.

jumyungc (dismissed):

newton/_src/geometry/contact_match.py
```diff
@@ -0,0 +1,941 @@
...
+        """
+        self._prev_count.zero_()
+
+    # ------------------------------------------------------------------
+    # Public methods
+    # ------------------------------------------------------------------
+
+    def match(
+        self,
+        sort_keys: wp.array,
```
eric-heiden: ⚪ These public methods use bare `wp.array` annotations.
Please use bracketed Warp array annotations on `match`, `save_sorted_state`, `replay_matched`, and `build_report`, for example:

| parameter | type |
| --- | --- |
| `sort_keys` | `wp.array[wp.int64]` |
| `contact_count` | `wp.array[wp.int32]` |
| `point0`, `point1`, `normal` | `wp.array[wp.vec3]` |
| `shape0`, `shape1`, `shape_body`, `match_index_out` | `wp.array[wp.int32]` |
| `body_q` | `wp.array[wp.transform]` |

newton/_src/sim/collide.py
nvtw: Done. I think AI created it because we have something about enums in our Agents.md? At least the first time I asked to remove that enum, it was pushing pack.

newton/_src/sim/collide.py
```diff
@@ -512,6 +542,24 @@ def __init__(
             deterministic: Sort contacts after the narrow phase so that results
                 are independent of GPU thread scheduling.  Adds a radix sort +
                 gather pass.  Hydroelastic contacts are not yet covered.
+            contact_matching: Frame-to-frame contact matching mode.  Accepts
+                a string (``"disabled"`` / ``"latest"`` / ``"sticky"``) or the
+                equivalent :class:`CollisionPipeline.ContactMatching` enum
+                value.  Any non-disabled mode implies ``deterministic=True``
+                and populates :attr:`Contacts.rigid_contact_match_index`.
+                Defaults to ``"disabled"``.
+            contact_matching_pos_threshold: World-space distance threshold [m]
```
eric-heiden, line 525: 🔴 The documented default for `contact_matching_pos_threshold` does not match the constructor default.

`CollisionPipeline.__init__` defaults this value to `0.0005`, but the docstring says `0.005`. Since `ContactMatcher` squares this threshold before comparing midpoint distance, this is a 10x distance mismatch and a 100x squared-threshold mismatch. Users reading the docs will expect contacts to remain matched up to 5 mm, while the default actually treats moves over 0.5 mm as broken.

Please update the docstring to say `Defaults to ``0.0005``.` If `0.005` is intended instead, the constructor default and tests should be updated together.

nvtw: Done

newton/_src/geometry/contact_match.py
nvtw: Done

eric-heiden (approved):

PR #2449, opened 2026-04-15
Move modular design of newton-actuators to Newton repo

### LINKED ISSUE #2295: Move newton-actuators into newton
Move the https://github.com/newton-physics/newton-actuators repository into Newton and remove the separate repository.

jvonmuralt:
## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **New Features**
  * New integrated GPU actuator system: composable controllers (PD, PID, MLP, LSTM), multiple clamping modes (max-force, DC-motor, position-based), optional input delay, USD actuator parsing, and a consolidated public actuator API; docs and examples added.

* **Chores**
  * Removed external actuator dependency and migrated builder API, tests, and release checklist to the integrated controller-based workflow.


newton/_src/actuators/actuator.py
```diff
@@ -0,0 +1,294 @@
...
+        device = indices.device
+        self._sequential_indices = wp.array(np.arange(self.num_actuators, dtype=np.uint32), device=device)
+        self._computed_forces = wp.zeros(self.num_actuators, dtype=wp.float32, device=device)
+        self._applied_forces = (
+            wp.zeros(self.num_actuators, dtype=wp.float32, device=device) if self.clamping else None
+        )
+
+        controller.finalize(device, self.num_actuators)
+
+    def get_param(self, name: str) -> wp.array | None:
```
eric-heiden: Why do we need this method? It seems like a potential source of problems if the controller and clamping object share the same parameter name. The user should know from where to get the parameter, right?

newton/_src/actuators/delay.py
```diff
@@ -0,0 +1,158 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 The Newton Developers
```
eric-heiden: All these new files should have 2026 as copyright year.

newton/_src/actuators/controllers/controller_net_mlp.py
```diff
@@ -0,0 +1,169 @@
...
+    """MLP-based neural network controller.
+
+    Uses a pre-trained MLP to compute joint torques from position error
+    and velocity history.
+
+    The network receives concatenated position-error and velocity history
+    as input and is expected to return torques in physical units. Scaling is left to the user.
+    """
+
+    SHARED_PARAMS: ClassVar[set[str]] = {"network_path"}
```
eric-heiden, line 30: 🔴 `SHARED_PARAMS` is missing `input_order` and `input_idx`, which causes a `RuntimeError` during `finalize()`.

`resolve_arguments()` returns `input_order` (a string like `'pos_vel'`) and `input_idx` (`None`) as non-shared params. Because they are not in `SHARED_PARAMS`, `ModelBuilder` treats them as per-DOF numeric data and passes them through `_stack_args_to_arrays()` with `wp.float32`. That blows up with:

```
RuntimeError: could not convert string to float: 'pos_vel'
```

Both values are configuration shared across all DOFs, not per-DOF arrays.

```suggestion
    SHARED_PARAMS: ClassVar[set[str]] = {"network_path", "input_order", "input_idx"}
```

newton/_src/actuators/delay.py
```diff
@@ -0,0 +1,158 @@
...
+            self.write_idx = self.buffer_pos.shape[0] - 1
+            self.is_filled = False
+
+    @classmethod
+    def resolve_arguments(cls, args: dict[str, Any]) -> dict[str, Any]:
+        if "delay" not in args:
+            raise ValueError("Delay requires 'delay' argument")
+        return {"delay": args["delay"]}
+
+    def __init__(self, delay: int):
```
eric-heiden: 🟡 `delay=0` is accepted here but causes a `ZeroDivisionError` on the first actuator step.

`update_state()` computes `(current_state.write_idx + 1) % self.delay`, which divides by zero when `self.delay == 0`.

Either reject it early or treat zero as a pass-through:

```suggestion
    def __init__(self, delay: int):
        """Initialize delay.

        Args:
```

Option A -- fail fast:
```py
def __init__(self, delay: int):
    if delay < 1:
        raise ValueError(f"delay must be >= 1, got {delay}")
```

Option B -- treat `delay=0` as identity (skip the ring buffer entirely in `update_state`).

Either way, the current code will crash at runtime for any user who passes `delay=0`.

newton/_src/actuators/usd_parser.py
```diff
@@ -0,0 +1,181 @@
...
+@dataclass
+class ActuatorParsed:
+    """Result of parsing a USD actuator prim.
+
+    Each detected API schema produces a (class, kwargs) entry.
+    The controller is separated out; everything else goes into
+    component_specs (delay, clamping, etc.).
+    """
+
+    controller_class: type
```
eric-heiden: ⚪ Nit: bare `type` loses useful type information here and on line 84.

Since `controller_class` is always a `Controller` subclass and `component_specs` entries are always `Clamping` or `Delay` subclasses, tighter annotations would give better IDE autocompletion and catch misuse earlier:

```suggestion
    controller_class: type[Controller]
```

And on line 84:
```py
component_specs: list[tuple[type[Clamping] | type[Delay], dict[str, Any]]] = field(default_factory=list)
```

newton/_src/actuators/clamping/base.py
```diff
@@ -0,0 +1,61 @@
...
+
+        Args:
+            args: User-provided arguments.
+
+        Returns:
+            Complete arguments with defaults filled in.
+        """
+        raise NotImplementedError(f"{cls.__name__} must implement resolve_arguments")
+
+    def modify_forces(
```
eric-heiden, line 49: 🟡 `modify_forces()` silently does nothing when a subclass forgets to override it.

The `pass` body means `dst_forces` stays zeroed while `src_forces` is ignored, so a controller might compute the correct torque but the actuator applies zero force. This is inconsistent with `resolve_arguments()` on line 35, which already raises `NotImplementedError`.

Replace `pass` with `raise NotImplementedError` to match:

```py
def modify_forces(self, src_forces, dst_forces, positions, velocities):
    raise NotImplementedError(f"{type(self).__name__} must implement modify_forces")
```

eric-heiden:
🟡 **Bare `wp.array` annotations in public Python signatures**

AGENTS.md requires bracket-syntax Warp array annotations (`wp.array[float]`, `wp.array[wp.uint32]`, etc.), but roughly 15 Python-level signatures use bare `wp.array`. The Warp kernel signatures already use the typed form, so the correct dtypes are known.

Affected locations include:

| File | Method | Example fix |
|------|--------|-------------|
| `actuator.py` | `__init__` | `indices: wp.array` &rarr; `wp.array[wp.uint32]` |
| `controllers/base.py` | `compute` | `positions: wp.array` &rarr; `wp.array[float]` |
| `clamping/base.py` | `modify_forces` | `src_forces: wp.array` &rarr; `wp.array[float]` |
| `delay.py` | `update_state` | `target_pos: wp.array` &rarr; `wp.array[float]` |

Please update these to match the dtypes used in the corresponding kernel signatures.

newton/_src/actuators/delay.py
```diff
@@ -0,0 +1,158 @@
...
+    clamping list).
+    """
+
+    SHARED_PARAMS: ClassVar[set[str]] = {"delay"}
+
+    @dataclass
+    class State:
+        """Circular buffer state for delayed targets."""
+
+        buffer_pos: wp.array2d[float] = None
```
eric-heiden: ⚪ `buffer_pos`, `buffer_vel`, and `buffer_act` default to `None` but are annotated without `| None`.

Per PEP 604 and AGENTS.md, these should include the union type. This also matches the pattern already used in `ControllerPID.State.integral` and `ControllerNetLSTM.State`.

```suggestion
        buffer_pos: wp.array2d[float] | None = None
```

Same change needed for `buffer_vel` on line 62 and `buffer_act` on line 63.

newton/_src/actuators/actuator.py
```diff
@@ -0,0 +1,294 @@
...
+
+        def reset(self) -> None:
+            if self.controller_state is not None:
+                self.controller_state.reset()
+            if self.delay_state is not None:
+                self.delay_state.reset()
+
+    def __init__(
+        self,
+        indices: wp.array,
```
jeff-hough: In the original version of the library, we had both "input_indices" and "output_indices". I think we got rid of this because in the up-coming integration with Lab/Sim, we are:

- limiting ourselves to the case where we have one actuator per DOF of the robot, and 
- we expect that the input array (list of joint measurements) will have the same structure as the output array (list of joint force outputs)

However, isn't collapsing those two lists into just "indices" a bit limiting on the library beyond Sim/Lab? That requires that _every_ consumer of this library matches the structure of their input and output arrays, which may not always be the case.

For example, what if: 

- the input array contains passive and active joint measurements, and 
- the output array contains only actuated joint forces? 

That is a perfectly reasonable way for a user to organize their data, but if the user only has `indices` then they are being blocked from doing things that way.

Thoughts @jvonmuralt @eric-heiden @preist-nvidia ?

jvonmuralt: I was thinking to start it minimal for now and introduce output_indices as an optional parameter at a later iteration.

jvonmuralt: I also have a concern if we have some D6 joint in the robot, then joint_q and joint_qd indexing differs. should it be a users responsibility to construct every step joint_q_per_dof so that library reads correct values. Or should we then accept pos_indices, vel_indices, output_indices in actuator and then newton state and control can be passed as they are (it would also imply changing control.joint_target_pos to be consistent with state.joint_q)?

jeff-hough: I see. So I guess we could say we have three options:

1. Keep just `indices`: simplest (and what we already have) but most restrictive on the user data.
2. Put back `input_indices` and `output_indices`: I think this will work for all 1DOF acuators, including those with tranmission, but is slightly more complex.
3. Every array gets its own index array: most flexible and seems needed for D6 joints, adds the most indexing complexity.

newton/_src/actuators/actuator.py
```diff
@@ -0,0 +1,294 @@
...
+        delay: Optional Delay instance for input delay.
+        clamping: List of Clamping objects (post-controller force bounds).
+        state_pos_attr: Attribute on sim_state for positions.
+        state_vel_attr: Attribute on sim_state for velocities.
+        control_target_pos_attr: Attribute on sim_control for target positions.
+        control_target_vel_attr: Attribute on sim_control for target velocities.
+        control_feedforward_attr: Attribute on sim_control for control input. None to skip.
+        control_output_attr: Attribute on sim_control for clamped output forces.
+        control_computed_output_attr: Attribute on sim_control for raw (pre-clamp)
+            forces. None to skip writing computed forces.
```
eric-heiden: I would move these docstrings for constructor args to the constructor's docstring. That's at least what we have done in most other Newton classes.

newton/_src/actuators/actuator.py
```diff
@@ -0,0 +1,293 @@
...
+    indices: wp.array[wp.uint32],
+    output: wp.array[float],
+    computed_output: wp.array[float],
+):
+    """Scatter-add forces into output; optionally scatter computed forces too."""
+    i = wp.tid()
+    idx = indices[i]
+    output[idx] = output[idx] + forces[i]
+    if computed_output:
+        computed_output[idx] = computed_output[idx] + computed_forces[i]
```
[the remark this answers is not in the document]
jvonmuralt: actuator.step is launched in a loop. Current actuators are targeting 1 joint only, so we do not need atomic for now

newton/_src/actuators/actuator.py
eric-heiden: I'm confused - D6 joints have the same number of q and qd coordinates, maybe @jvonmuralt means free and ball joints which have differing coordinate counts?
I also think we need some separation of q and qd indices because for floating-base articulations (or articulations which have ball joints somewhere) the q indices will differ from qd/"output" indices.
This isn't really a question of input vs. output indices, right? It is more about `joint_q` and `joint_qd` indices because an actuator controller could read inputs from both `joint_q` and `joint_qd`. `state_pos_attr` and `state_vel_attr` are the raw `joint_q` and `joint_qd` arrays in Newton so there are going to be indexing problems unless we provide q/qd indexing information.

So maybe we actually need 3 index arrays?

| Axis | What it indexes | Example where collapsing fails |
|---|---|---|
| A. `joint_q` read index | positions for controller feedback | free/ball joint shifts coord vs. DOF numbering |
| B. `joint_qd` read index | velocities for controller feedback | same joints — q and qd diverge |
| C. `joint_f` write index | where clamped force gets scattered | passive vs. actuated DOFs, tendon/coupled transmissions |

If an actuator actuates exactly the joints that it reads from then joint_qd and joint_f indexing are the same.

jvonmuralt: yes, I meant free and ball joints:) and exactly the problem which you describe.
It looks like we would need 3 index arrays indeed. 
I also think we would need to change control.joint_target_pos to be the same shape as joint_q, so that index array A can be used for reading both position state and target.

jeff-hough: The analysis table is great! It's a very good summary of cases where having just one index array will fail. 

@jvonmuralt and @preist-nvidia , is the preference still to keep maximally simple on the first release and then add the other index arrays later?

Nitpick: I don't think we use the word "joint" except for on the output, where we are in fact always output joint forces.

On inputs, some actuators will depend on positions/velocities which are "actuator space" values, no always "joint-space" (i.e. length or speed of a retracting tendon).

andrewkaufman:
Schemas for this are ready for review [here](https://github.com/newton-physics/newton-usd-schemas/pull/45). Once that merges the parsing here should be updated to align. Or we can merge this first & align in a followup PR.

docs/concepts/actuators.rst
```diff
@@ -0,0 +1,237 @@
...
+
+**Delay**
+   Optionally delays the control targets (e.g. position or velocity) by *N*
+   timesteps before they reach the controller, allowing the actuator to model
+   communication or processing latency.  The delay always produces output;
+   when the buffer is still filling, the lag is clamped to the available
+   history so the most recent data is returned.
+
+**Controller**
+   Computes raw forces or torques from the current simulator state and control
```
preist-nvidia, line 47: Given my comments on the USD PR and having looked at PhysX perf envelope and lab API docs, should we standardize to effort to mean force/torque in this PR also?

Not going to comment further in this PR we can decide first then make a single refactor.

docs/concepts/actuators.rst
```diff
@@ -0,0 +1,237 @@
...
+   Actuator
+   ├── Delay       (optional: delays control targets by N timesteps)
+   ├── Controller  (control law that computes raw forces)
+   └── Clamping[]  (clamps raw forces based on motor-limit modeling)
+       ├── ClampingMaxForce        (±max_force box clamp)
+       ├── ClampingDCMotor         (velocity-dependent saturation)
+       └── ClampingPositionBased   (angle-dependent lookup table)
+
+**Delay**
+   Optionally delays the control targets (e.g. position or velocity) by *N*
```
preist-nvidia, line 40: The question I had in the USD PR applies here too, does delay apply also to FF terms and should we therefore prefer the control inputs instead of specific target language?

I'm aware I initially suggested the more narrow control targets, sorry.

newton/_src/actuators/delay.py
```diff
@@ -0,0 +1,325 @@
...
+
+    Delays targets using a circular buffer of depth
+    ``max_delay``.  Each DOF has its own lag stored in
+    :attr:`delays` (shape ``(N,)``).  The buffer is sized for the
+    maximum lag across all DOFs so that DOFs with different delays
+    can share the same actuator group.
+
+    The delay always produces output.  When the buffer is empty
+    (e.g. right after reset), the current targets are returned
+    directly.  When underfilled, the lag is clamped to the
```
preist-nvidia, line 115: Not sure the underfill behavior matches lab where the docstring says: https://github.com/isaac-sim/IsaacLab/blob/4df6560e187f2cc66685b41b21b259f4485d0c22/source/isaaclab/isaaclab/utils/buffers/delay_buffer.py#L164

newton/_src/actuators/clamping/clamping_position_based.py
```diff
@@ -0,0 +1,215 @@
...
+        for raw_line in table_path.read_text().splitlines():
+            stripped = raw_line.strip()
+            if not stripped or stripped.startswith("#"):
+                continue
+            parts = stripped.replace(",", " ").split()
+            angles.append(float(parts[0]))
+            torques.append(float(parts[1]))
+        if not angles:
+            raise ValueError(f"Lookup table file is empty: {path}")
+        if any(v < 0 for v in torques):
```
preist-nvidia, line 165: Should there be an additional check on monotonicity?

newton/_src/utils/import_usd.py
```diff
@@ -2752,35 +2752,45 @@ def _get_collision_mass_information(collider_prim: Usd.Prim):
...
+        path: builder.joint_qd_start[idx] for path, idx in path_joint_map.items() if idx < len(builder.joint_qd_start)
+    }
+    path_to_coord = {
+        path: builder.joint_q_start[idx] for path, idx in path_joint_map.items() if idx < len(builder.joint_q_start)
+    }
+    for prim in Usd.PrimRange(stage.GetPrimAtPath(root_path)):
+        parsed = parse_actuator_prim(prim)
+        if parsed is None:
+            continue
+        if len(parsed.target_paths) != 1 or parsed.target_paths[0] not in path_to_dof:
```
preist-nvidia, line 2769: In the schema we say: "  Currently only the first target is supported" so we don't need to fail for != 1, just ignore additional ones.

The dof detection failure should raise I think.

0 targets may be a valid disable mechanism and should just ignore, but depends on whether we add an enable: https://github.com/newton-physics/newton-usd-schemas/pull/45/changes#r3100480607

docs/concepts/actuators.rst
```diff
@@ -0,0 +1,237 @@
...
+The per-step pipeline is:
+
+.. code-block:: text
+
+   Delay read → Controller → Clamping → Scatter-add → State updates (controller + delay write)
+
+Controllers and clamping objects are pluggable: implement the
+:class:`Controller` or :class:`Clamping` base class to add new models.
+
+.. note::
```
preist-nvidia, line 68: I would make this more prominent in the intro.

Should this also have an experimental feature disclaimer as we discussed? We don't need it if you are fairly sure about the future-proofness of the API with upcoming changes like transmission.

newton/_src/actuators/actuator.py
```diff
@@ -0,0 +1,311 @@
...
+        controller: Controller,
+        delay: Delay | None = None,
+        clamping: list[Clamping] | None = None,
+        pos_indices: wp.array[wp.uint32] | None = None,
+        frc_indices: wp.array[wp.uint32] | None = None,
+        state_pos_attr: str = "joint_q",
+        state_vel_attr: str = "joint_qd",
+        control_target_pos_attr: str = "joint_target_pos",
+        control_target_vel_attr: str = "joint_target_vel",
+        control_feedforward_attr: str | None = "joint_act",
```
jeff-hough, line 92: @jvonmuralt question:
Is the feedforward intended to be "post-transmission" (once we have transmission)? I.e., forces forwarded directly to the joints?

In the future, should we also have a "pre-transmission" feed-forward? This would cover the case we discussed yesterday in slack:

1. A user has a single tendon which is routed such that it transmits forces 3 joints
2. They just want to apply a constant force through the tendon

I.e., more generally they could run a full control-law externally, and then just forward the actuator-space commands to their actuators.

newton/_src/actuators/clamping/clamping_dc_motor.py
```diff
@@ -0,0 +1,111 @@
...
+    def __init__(
+        self,
+        saturation_effort: wp.array[float],
+        velocity_limit: wp.array[float],
+        max_force: wp.array[float],
+    ):
+        """Initialize DC motor saturation.
+
+        Args:
+            saturation_effort: Peak motor torque at stall [N·m]. Shape (N,).
```
preist-nvidia, line 80: Should this define allowable ranges? And then input-check as well?

newton/_src/actuators/clamping/clamping_max_force.py
```diff
@@ -0,0 +1,59 @@
...
+            raise ValueError(f"max_force must be non-negative, got {max_force}")
+        return {"max_force": max_force}
+
+    def __init__(self, max_force: wp.array[float]):
+        """Initialize max-force clamp.
+
+        Args:
+            max_force: Per-actuator force limits [N or N·m]. Shape (N,).
+        """
+        self.max_force = max_force
```
preist-nvidia, line 41: A goal was for this library to be usable standalone, e.g. outside Newton with PhysX. In that usage, I would not expect the `builder -> resolve_arguments() -> constructor` path to be followed, so some of the current input validation/defaulting would be bypassed.

Should we make the standalone Python construction path validated as well, e.g. via a public factory that applies `resolve_arguments()`, or by moving the relevant checks into the component construction path?

newton/_src/utils/selection.py
```diff
@@ -1703,13 +1703,26 @@ def _get_actuator_dof_mapping(self, actuator: "Actuator") -> wp.array:
...
 
-    def _get_actuator_attribute_array(self, actuator: "Actuator", name: str) -> wp.array:
-        """Get actuator parameter array shaped (world_count, dofs_per_world), zeros for non-actuated DOFs."""
+    def get_actuator_parameter(self, actuator: "Actuator", component: Any, name: str) -> wp.array:
+        """Get actuator parameter values for actuators corresponding to this view's DOFs.
+
+        Args:
+            actuator: Actuator instance (used for DOF index mapping).
+            component: The component that owns the parameter — a
+                :class:`~newton._src.actuators.controllers.base.Controller`,
```
adenzler-nvidia, line 1712: `AGENTS.md`: *"Sphinx cross-refs… Prefer public API paths; never use `newton._src`."* These new docstrings reference `:class:`~newton._src.actuators.controllers.base.Controller`` etc. The public paths are `newton.actuators.Controller` / `newton.actuators.Clamping` / `newton.actuators.Delay`. Can we route through those instead? (Same check worth running over all the new files — I spot-checked the actuator module itself and it already uses `newton.actuators.*`, so it's just this pair of selection docstrings I noticed.)

newton/_src/actuators/controllers/controller_net_mlp.py
```diff
@@ -0,0 +1,186 @@
...
+        def reset(self, mask: wp.array[wp.bool] | None = None) -> None:
+            if mask is None:
+                self.pos_error_history.zero_()
+                self.vel_history.zero_()
+            else:
+                t = wp.to_torch(mask).bool()
+                self.pos_error_history[:, t] = 0.0
+                self.vel_history[:, t] = 0.0
+
+    @classmethod
```
adenzler-nvidia, line 50: `resolve_arguments()` returns only `network_path`, but the `__init__` signature has five more user-tunable params (`input_order`, `input_idx`, `pos_scale`, `vel_scale`, `torque_scale`). Because the builder does

```python
resolved_ctrl = controller_class.resolve_arguments(controller_kwargs)
...
controller = entry.controller_class(**ctrl_arrays, **entry.controller_shared_kwargs)
```

any of those kwargs passed through `ModelBuilder.add_actuator(ControllerNetMLP, ...)` (or via USD) will be silently dropped and the controller will use the defaults — including `pos_scale=vel_scale=torque_scale=1.0`, which is often not what a trained network needs.

Two options I can see:
1. Include them in `resolve_arguments()` and mark `input_idx`/`input_order` as `SHARED_PARAMS` (they don't vary per-DOF anyway).
2. Document explicitly that programmatic construction is required and raise if these kwargs are passed through `add_actuator`.

Either way, silent dropping of scale factors feels like a footgun. Same comment applies in spirit to `ControllerNetLSTM.resolve_arguments`, though that one has fewer tunables.

Related: the comment in `usd_parser.py:67-68` ("input_order / input_idx are intentionally left out of the schema; they should be set programmatically") acknowledges the issue for USD but the programmatic `add_actuator` path has the same hole.

newton/_src/actuators/actuator.py
```diff
@@ -0,0 +1,311 @@
...
+            else None
+        )
+
+        controller.finalize(self.device, self.num_actuators)
+        if delay is not None:
+            delay.finalize(self.device, self.num_actuators, requires_grad=requires_grad)
+        for clamp in self.clamping:
+            clamp.finalize(self.device, self.num_actuators)
+
+    @property
```
adenzler-nvidia, line 157: `Actuator.SHARED_PARAMS` is an instance property, but the component-level `Controller.SHARED_PARAMS` / `Clamping.SHARED_PARAMS` are `ClassVar[set[str]]`. The builder relies on reading it as a classvar (`getattr(controller_class, "SHARED_PARAMS", set())`). Reading `Actuator.SHARED_PARAMS` on the class returns the property object, not a set — so this method would silently break the same grouping logic if someone ever reached for it as a group key.

grep shows nothing else in the repo actually calls `Actuator.SHARED_PARAMS`. Is this still needed, or can it be dropped? (Noting @eric-heiden raised a related question in the existing thread — this is asking the same thing from the "consistency with the rest of the system" angle.)

newton/_src/actuators/delay.py
```diff
@@ -0,0 +1,325 @@
...
+                    their buffer columns zeroed and push count reset.
+                    ``None`` resets all.
+            """
+            if mask is None:
+                self.buffer_pos.zero_()
+                self.buffer_vel.zero_()
+                self.buffer_act.zero_()
+                self.num_pushes.zero_()
+                self.write_idx = self.buffer_pos.shape[0] - 1
+            else:
```
adenzler-nvidia, line 148: Two small things on masked reset:

a. The loop launches `_delay_masked_reset_kernel` three times (one per buffer), and each launch zeros `num_pushes[i]` where mask is true. That's redundant — the kernel takes `num_pushes` as output. Either pass `num_pushes` on only one launch, or have one kernel that zeros all three buffers and `num_pushes` once.

b. Related, `Delay.__init__` accepts `max_delay: int` with no validation. `ModelBuilder.add_actuator`'s `delay` path enforces `delay >= 1`, but `Delay(max_delay=0)` constructed directly would later divide by zero in `update_state`'s `(write_idx + 1) % self.buf_depth`. The builder-side validation in `resolve_arguments` should probably move (or be mirrored) into `Delay.__init__` so direct construction (the exact use case the concept doc advertises) doesn't have that trap.

docs/concepts/actuators.rst
```diff
@@ -0,0 +1,237 @@
...
+
+   class MyController(Controller):
+       @classmethod
+       def resolve_arguments(cls, args):
+           return {"gain": args.get("gain", 1.0)}
+
+       def __init__(self, gain: wp.array):
+           self.gain = gain
+
+       def compute(self, positions, velocities, target_pos, target_vel,
```
adenzler-nvidia, line 219: The example signature here is stale vs. `Controller.compute` in base.py:

```python
def compute(self, positions, velocities, target_pos, target_vel,
            feedforward, input_indices, target_indices, forces,
            state, dt, device=None):
```

The real signature uses `pos_indices`, `vel_indices`, `target_pos_indices`, `target_vel_indices` (four index arrays, not two). Anyone following this example to write a custom controller will get a `TypeError` on kernel launch. Could we update the example to match the 4-index layout?

newton/_src/sim/builder.py
```diff
@@ -1760,44 +1797,27 @@ def _stack_args_to_arrays(
         result = {}
```
adenzler-nvidia, line 1797:
```python
values = [args[key] for args in args_list]
if all(isinstance(v, int) for v in values):
    result[key] = wp.array(values, dtype=wp.int32, device=device)
else:
    result[key] = wp.array(values, dtype=wp.float32, device=device, requires_grad=requires_grad)
```

This is a nice simplification vs. the previous scalar-only stacking, but the `else` branch will do surprising things if `values` is a list of `wp.array` or numpy arrays: `wp.array` with a list of 0-d / ragged items tends to silently succeed with unintended shapes. Since `add_actuator`'s `**controller_kwargs` accepts anything, a user passing e.g. `kp=wp.array([1.0, 2.0])` (maybe thinking "multi-DOF at once") currently doesn't get a useful error.

Worth adding a `isinstance(v, (int, float))` check and raising a clear `TypeError("add_actuator expects scalar per-DOF params; got {type}")`?

newton/tests/test_actuators.py
```diff
@@ -22,252 +34,693 @@
...
+# ---------------------------------------------------------------------------
 
-        template.add_actuator(ActuatorPD, input_indices=[dof0], kp=100.0, kd=10.0)
-        template.add_actuator(ActuatorPID, input_indices=[dof1], kp=200.0, ki=5.0, kd=20.0)
-        template.add_actuator(ActuatorPD, input_indices=[dof1], output_indices=[dof2], kp=300.0)
 
-        num_worlds = 3
-        builder = newton.ModelBuilder()
-        builder.replicate(template, num_worlds)
+class TestDelay(unittest.TestCase):
```
adenzler-nvidia, line 295: The test file is thorough on the "happy path" (PD/PID/DC-motor/position-based/MLP/LSTM compute + full pipeline integration + selection API), which is great. Two gaps that match promises the code or docs make:

1. **`State.reset(mask=...)`** — `Delay.State.reset`, `ControllerPID.State.reset`, `ControllerNetMLP.State.reset`, `ControllerNetLSTM.State.reset`, and `Actuator.State.reset` are all part of the public API but not exercised by a single test. The masked path is the most interesting (e.g. resetting one env in a vectorized batch) and is the exact kind of logic that silently does the wrong thing.
2. **CUDA graph capture** — `Actuator.is_graphable()` is documented but there's no `wp.ScopedCapture`/`wp.capture_launch` test confirming PD+clamping actually graph-captures end-to-end. Given that this is explicitly called out as a design goal in the concept doc, a smoke test would be valuable.

At minimum, a `reset()` test for `Delay.State` + `ControllerPID.State` with a mask feels low-cost and catches the kind of regression that would otherwise silently corrupt a training run.

*Follow-up for a later PR (not blocking this one):* a `requires_grad=True` end-to-end test validating that a differentiable chain produces non-None grads after `backward()`. The gradient-enabled scratch-buffer path in `Actuator.__init__(requires_grad=...)` has no coverage today, but testing it well requires thinking through what a differentiable-actuator API contract should look like — bigger than this PR.

newton/_src/actuators/usd_parser.py
```diff
@@ -0,0 +1,195 @@
...
+    """Result of parsing a USD actuator prim.
+
+    Each detected API schema produces a (class, kwargs) entry.
+    The controller is separated out; everything else goes into
+    component_specs (delay, clamping, etc.).
+    """
+
+    controller_class: type[Controller]
+    controller_kwargs: dict[str, Any] = field(default_factory=dict)
+    component_specs: list[tuple[type[Clamping | Delay], dict[str, Any]]] = field(default_factory=list)
```
adenzler-nvidia, line 93: The dataclass keeps `target_paths: list[str]` "for future multi-target support" (comment in `# Currently only a single target is supported.`) and then `parse_actuator_prim` raises if `len != 1`. Meanwhile `import_usd.py:2769` also re-validates `len == 1`. Two questions:

- Should this just be `target_path: str` until multi-target is actually a thing? The list shape leaks into callers (see `import_usd.py`) and is easy to misuse.
- If we keep it as list for forward-compat, is it worth making `SCHEMA_REGISTRY` mutable/extensible via a public hook? Right now adding a custom controller schema requires monkey-patching a module-level dict; a `register_schema()` function would be a cleaner extension point.

docs/concepts/actuators.rst
```diff
@@ -0,0 +1,237 @@
+.. SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
```
adenzler-nvidia, line 1: Per `AGENTS.md`, user-facing changes (especially Changed/Removed) require a CHANGELOG entry with migration guidance. This PR (a) deletes the `newton-actuators` dependency, (b) replaces the old `ModelBuilder.add_actuator(actuator_class=..., input_indices=..., output_indices=...)` signature with a very different `add_actuator(controller_class, index, clamping=[...], delay=..., pos_index=..., **ctrl_kwargs)` API, and (c) changes `ArticulationView.get_actuator_parameter`/`set_actuator_parameter` to require a `component` argument.

All three are breaking for any existing user. Could we add entries under `Changed`/`Removed` with explicit before/after snippets? e.g.

```
- Remove `ActuatorPD`/`ActuatorPID` etc. in favor of composed
  `ControllerPD` + `ClampingMaxForce` etc. Migrate:
    builder.add_actuator(ActuatorPD, input_indices=[i], kp=..., kd=..., max_force=...)
  →
    builder.add_actuator(ControllerPD, index=i, kp=..., kd=...,
                         clamping=[(ClampingMaxForce, {"max_force": ...})])
```

docs/index.rst
```diff
@@ -35,6 +35,7 @@ Newton Physics
    Custom Attributes <concepts/custom_attributes>
    Extended Attributes <concepts/extended_attributes>
    Collisions and Contacts <concepts/collisions>
+   Actuators <concepts/actuators>
```
adenzler-nvidia, line 38: Don't forget to add `api/newton_actuators` to the API Reference toctree (further down in this file) and run `docs/generate_api.py` to generate the page. Per `AGENTS.md` — and without it, the `:mod:`newton.actuators`` / `:class:`ControllerPD`` / `:meth:`Actuator.state`` cross-refs in the new concept doc won't resolve.

docs/concepts/actuators.rst
```diff
@@ -0,0 +1,237 @@
...
+   Clamps raw forces based on motor-limit modeling.  This applies
+   post-controller output limits to the computed forces or torques to model
+   motor limits such as saturation, back-EMF losses, performance envelopes, or
+   angle-dependent torque limits.  Multiple clamping stages can be combined on
+   a single actuator.
+
+The per-step pipeline is:
+
+.. code-block:: text
+
```
adenzler-nvidia, line 62: Per Newton docs convention (`articulations.rst`, `collisions.rst`), runnable Python examples in concept pages use `.. testcode::` so they're exercised by the doctest builder and can't silently rot. This file uses `.. code-block:: python` for every Python example (lines 82, 104, 134, 146, 206) — none are executed.

Could we convert the first four (builder-level usage, manual construction, USD loading, selection API) to `.. testcode::` with appropriate named groups so setup state carries across? The custom-controller example at line 206 has a stub body (`...`) that won't execute cleanly — either give it a minimal concrete toy body so testcode works, or keep that one as `.. code-block::` with a note explaining why.

newton/_src/utils/selection.py
```diff
@@ -1652,15 +1643,15 @@ def _get_actuator_dof_mapping(self, actuator: "Actuator") -> wp.array:
         Build mapping from view DOF positions to actuator parameter indices.
 
         Note:
-            For selection we assume that input_indices is 1D (one input per actuator),
+            For selection we assume that indices is 1D (one input per actuator),
             not the general 2D case (multiple inputs per actuator) which is supported
```
preist-nvidia, line 1647: What is the general 2D case mentioned here? We are limiting to SISO first right?

newton/_src/utils/selection.py
preist-nvidia, line 282: Are these float arrays an issue for int delay gather/scatter?

newton/_src/utils/selection.py
```diff
@@ -1712,13 +1703,26 @@ def _get_actuator_dof_mapping(self, actuator: "Actuator") -> wp.array:
 
         return mapping
 
-    def _get_actuator_attribute_array(self, actuator: "Actuator", name: str) -> wp.array:
-        """Get actuator parameter array shaped (world_count, dofs_per_world), zeros for non-actuated DOFs."""
+    def get_actuator_parameter(self, actuator: "Actuator", component: Any, name: str) -> wp.array:
+        """Get actuator parameter values for actuators corresponding to this view's DOFs.
```
preist-nvidia, line 1707: Can you expand on this docstring a bit? I don’t understand what “this view’s DOFs” means here.

Is this returning parameter values for all DOFs in the view, with non-actuated ones ignored/defaulted, or only for the subset of DOFs that this actuator instance actually applies effort to? If the latter, it would help to say that explicitly.

newton/_src/actuators/actuator.py
jvonmuralt: We can discuss it further when we are working on adding transmission.
But I would say we should not need to have 2 feedforward terms and in the general case it is going to be pre-transmission. It is just joint actuators where we dont model transmission and assume everything to be in the joint space.

docs/concepts/actuators.rst
jvonmuralt: yes, delay is applied to all the inputs.

docs/concepts/actuators.rst
jvonmuralt: yes. I can use the word effort

newton/_src/actuators/delay.py
jvonmuralt: Behavior is consistent, it is implemented differently ( lab first writes then reads) and we do the other way around but the resulting targets used in controller are the same.

newton/_src/actuators/clamping/clamping_position_based.py
jvonmuralt: added, thanks!

newton/_src/utils/selection.py
jvonmuralt: cleaned this up, thanks!

newton/_src/utils/selection.py
jvonmuralt: fixed, thanks!

andrewkaufman:
@jvonmuralt update to `"newton-usd-schemas>=0.2.0rc1"` to pickup the actuator schemas, it is published now

newton/_src/actuators/clamping/clamping_max_force.py
jvonmuralt: I think the idea is that the client code should integrate construction into its parsing workflow, just how we do it with the builder with the user of helper functions parse_actuator_prim and resolve_arguments. If while we are integrating in Lab and Sim we notice the need of such a unified public factory, I will add it in another PR.

newton/_src/actuators/clamping/clamping_dc_motor.py
jvonmuralt: Check for allowed input values is done in resolve_arguments, in constructor the check for correct shapes is done.

newton/_src/sim/builder.py
```diff
@@ -1680,69 +1697,209 @@ def apply_indexed_values(
...
-        entry.output_indices.append(output_indices)
-        entry.args.append(array_params)
+        entry.indices.append(index)
+        entry.pos_indices.append(pos_index if pos_index is not None else index)
+        entry.controller_args.append(ctrl_array_params)
+        if delay_steps is not None:
+            entry.delay_args.append({"delay_steps": delay_steps})
+        entry.clamping_args.append(clamping_array_params_list)
 
     def _stack_args_to_arrays(
```
eric-heiden, line 1916: 🔴 `_stack_args_to_arrays()` currently picks `wp.int32` whenever a parameter column contains only Python ints. That means a natural call like `kp=100, kd=10, max_effort=50` builds integer arrays, but the controller and clamping kernels expect `wp.float32`, so the first `step()` crashes with a dtype mismatch. Maybe we can coerce numeric actuator parameters to float unless a parameter explicitly opts into an integer dtype? I'm not sure if this is straightforward, to check if a integer is wanted. Otherwise we could just always assume float32.

eric-heiden:
🔴 The USD actuator import path still feeds `add_actuator(..., pos_index=...)` values on the `joint_q` layout, but `Actuator.step()` later uses `pos_indices` to index `control.joint_target_pos`, which is `joint_qd`-sized. On floating-base articulations that pushes later joints out of bounds and the PD position term reads garbage. So here we actually need to fix it so that actuator outputs always follow dof indices (at least in Newton).

newton/_src/actuators/actuator.py
```diff
@@ -0,0 +1,303 @@
...
+                    self.pos_indices,
+                    self.indices,
+                    device=self.device,
+                )
+                src = self._applied_forces
+            output_forces = self._applied_forces
+        else:
+            output_forces = self._computed_forces
+
+        # --- 4. Scatter-add to output ---
```
eric-heiden, line 279: 🟡 `step()` scatter-adds into both `applied_output` and `computed_output` without checking whether those attributes alias the same buffer. If `control_computed_output_attr='joint_f'`, both writes hit the same array and the force is doubled. Please skip the second scatter when the two targets are the same object, or reject that configuration up front.

newton/_src/sim/builder.py
```diff
@@ -1680,69 +1697,209 @@ def apply_indexed_values(
...
+                post-controller. E.g. ``[(ClampingMaxEffort, {'max_effort': 50.0})]``.
+            delay_steps: Optional number of timesteps [timesteps] to delay inputs.
+            pos_index: DOF index into ``joint_q``-shaped arrays (positions,
+                position targets). Defaults to *index*. Differs from
+                *index* for floating-base or ball-joint articulations
+                where ``joint_q`` and ``joint_qd`` have different layouts.
+            **kwargs: Per-DOF controller parameters (e.g. ``kp``, ``kd``).
+        """
+        legacy_cls = kwargs.pop("actuator_class", None)
+        if legacy_cls is None and _LegacyActuator is not None and issubclass(controller_class, _LegacyActuator):
```
eric-heiden: 🟡 This `issubclass(controller_class, _LegacyActuator)` check runs before `controller_class` is validated. When a caller forgets `controller_class` and `newton_actuators` happens to be installed, they get `TypeError: issubclass() arg 1 must be a class` instead of the intended guidance that `add_actuator()` requires `controller_class`. Please guard this branch with a real class check first.

eric-heiden:
🟡 The public actuator docstrings still point Sphinx at private `_src` symbols such as `newton._src.actuators.clamping.base.Clamping` and `_src.actuators.utils.load_checkpoint`. That leaks internal module paths into the rendered docs, and the helper references are likely to warn as unresolved because those modules are not part of the public docs tree. Could we retarget these to public `newton.actuators.*` symbols where possible, and use plain text for private helpers that are not meant to be link targets?

eric-heiden:
🟡 The actuator changelog currently has some mistakes in the API that it notes:

| Changelog text | Actual API in this PR |
| --- | --- |
| `ClampingMaxForce` | `ClampingMaxEffort` |
| `delay=` | `delay_steps=` |
| `newton-actuators` removed | dependency and compatibility shim are still present |

Could we rewrite these entries with the real symbol names and the current compatibility story, ideally under Added/Changed/Deprecated headings with explicit migration guidance?

eric-heiden:
⚪ `ActuatorParsed` is now a public `newton.actuators` type, so its annotations should communicate what callers actually get back. `controller_class: type` and `component_specs: list[tuple[type, dict[str, Any]]]` erase the distinction between controllers and clamping or delay components, which means no autocomplete or static checking for the public API. Could we tighten these to `type[Controller]` and `type[Clamping | Delay]`?

eric-heiden:
⚪ `Delay.State` advertises `buffer_pos`, `buffer_vel`, and `buffer_act` as non-null `wp.array2d[float]`, but all three default to `None`. That makes the type hints inconsistent with the actual API and with the other state dataclasses in this PR. Could we update these to `wp.array2d[float] | None = None` so the annotations match the initialization state?

eric-heiden:
⚪ The parameter-name examples in `set_actuator_parameter()` and `get_actuator_parameter()` still list `gear`, but the new actuator API no longer exposes that parameter. Leaving it in the docstrings will send readers toward a knob that does not exist anymore. Could we drop `gear` from both examples, and maybe add `ki` so the PID case is represented?

newton/actuators.py
```diff
@@ -0,0 +1,43 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 The Newton Developers
```
eric-heiden: ⚪ `newton/actuators.py` is a new file in this 2026 PR, so the SPDX year should match the creation year. This header still says `2025`, which makes the copyright metadata incorrect for the file that is actually being introduced here.

eric-heiden:
Sorry some of these comments didn't end up as inline comments. I'm not sure if all of these findings are valid, better to double check. Otherwise it looks good to me!

newton/_src/actuators/usd_parser.py
jvonmuralt: 1. according to schema we have multiple targets
2. added, thanks!

jvonmuralt:
> Sorry some of these comments didn't end up as inline comments. I'm not sure if all of these findings are valid, better to double check. Otherwise it looks good to me!

Thanks @eric-heiden, I addressed your comments. For control.joint_target_pos layout I created an issue as it is a bigger change https://github.com/newton-physics/newton/issues/2556

newton/_src/actuators/controllers/controller_neural_mlp.py
```diff
@@ -0,0 +1,186 @@
...
+        def reset(self, mask: wp.array[wp.bool] | None = None) -> None:
+            if mask is None:
+                self.pos_error_history.zero_()
+                self.vel_history.zero_()
+            else:
+                t = wp.to_torch(mask).bool()
+                self.pos_error_history[:, t] = 0.0
+                self.vel_history[:, t] = 0.0
+
+    @classmethod
```
jvonmuralt: additional parameters are now read from the file metadata

newton/_src/actuators/delay.py
```diff
@@ -0,0 +1,338 @@
...
+    def __init__(self, delay_steps: wp.array[int], max_delay: int):
+        """Initialize delay.
+
+        Args:
+            delay_steps: Per-DOF delay values [timesteps], shape ``(N,)``.
+            max_delay: Maximum delay across all DOFs.  Determines the
+                circular-buffer depth.
+
+        Raises:
+            ValueError: If *max_delay* < 1.
```
preist-nvidia, line 188: This will not be compatible with the new 0 delay decision we made.

docs/concepts/actuators.rst
```diff
@@ -0,0 +1,259 @@
...
+Clamping
+^^^^^^^^
+
+* :class:`ClampingMaxEffort` — symmetric box clamp to ±max_effort per actuator.
+* :class:`ClampingDCMotor` — velocity-dependent effort saturation using the DC
+  motor effort-speed characteristic.
+* :class:`ClampingPositionBased` — position-dependent effort limits via
+  interpolated lookup table (e.g. for linkage-driven joints).
+
+Multiple clamping objects can be stacked on a single actuator; they are applied
```
preist-nvidia, line 212: So the sequence is strict here in impl, but the parser is not order-preserving, so we only need the order comment in the schema?

docs/api/newton_actuators.rst
```diff
@@ -0,0 +1,45 @@
+.. SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+.. SPDX-License-Identifier: CC-BY-4.0
+
+newton.actuators
+================
+
+GPU-accelerated actuator models for physics simulations.
+
+This module provides a modular library of actuator components — controllers,
+clamping, and delay — that compute joint forces from simulation state and
```
preist-nvidia: Still forces  - could state effort here and define as we did in USD.

docs/concepts/actuators.rst
```diff
@@ -0,0 +1,259 @@
...
+       ├── ClampingMaxEffort        (±max_effort box clamp)
+       ├── ClampingDCMotor         (velocity-dependent saturation)
+       └── ClampingPositionBased   (position-dependent lookup table)
+
+**Delay**
+   Optionally delays the control targets (e.g. position or velocity) by *N*
+   timesteps before they reach the controller, allowing the actuator to model
+   communication or processing latency.  The delay always produces output;
+   when the buffer is still filling, the lag is clamped to the available
+   history so the most recent data is returned.
```
preist-nvidia: THis is at odds with the delay and schema docstring.

newton/_src/utils/import_usd.py
```diff
@@ -3212,35 +3212,50 @@ def _get_collision_mass_information(collider_prim: Usd.Prim):
...
-    else:
-        # TODO: Replace this string-based type name check with a proper schema query
-        # once the Newton actuator USD schema is merged
-        for prim in Usd.PrimRange(stage.GetPrimAtPath(root_path)):
-            if prim.GetTypeName() == "Actuator":
-                raise ImportError(
-                    f"USD stage contains actuator prims (e.g. {prim.GetPath()}) but newton-actuators is not installed. "
-                    "Install with: pip install newton[sim]"
-                )
+    path_to_dof = {
```
preist-nvidia, line 3219: This mapping silently assumes every actuator target joint is 1-DOF — we take `joint_qd_start[idx]` and treat it as *the* actuator DOF without checking DOF count. Schema-compliant inputs (Revolute/Prismatic only) are fine, but a multi-DOF target like a D6 locked to one axis via LimitAPI — which the schema explicitly forbids — is accepted here and silently drives DOF 0 regardless of which axis the user intended. Can we gate on `joint_qd_start[idx+1] - joint_qd_start[idx] == 1` and raise otherwise? Would also let us drop the `coord_index != dof_index` branch on line 3240, which is only reachable for multi-DOF targets.

newton/_src/actuators/usd_parser.py
```diff
@@ -0,0 +1,277 @@
...
+        ValueError: If the prim is a ``NewtonActuator`` but:
+            - has no authored ``newton:targets`` relationship,
+            - has multiple controller schemas applied,
+            - has no controller schema, or
+            - has a ``NewtonNeuralControlAPI`` with an unsupported model.
+    """
+    if prim.GetTypeName() != "NewtonActuator":
+        return None
+
+    target_paths = _get_relationship_targets(prim, "newton:targets")
```
preist-nvidia, line 225: We only keep the target as a path string here — no check on the target prim's type or state. A few questions on the intended contract:

- Schema says the target must be a `PhysicsRevoluteJoint` or `PhysicsPrismaticJoint`. Should the parser (or importer) validate that, or do we rely on the caller? Today a D6/ball/free/fixed target passes straight through; fixed is particularly unfriendly since it has 0 DOFs and the importer's `joint_qd_start[idx]` lookup silently resolves to the next joint's first DOF.
- What's the intended behavior when the target joint is excluded from any articulation? Newton already supports orphan joints (with the `skip_validation_joints=True` opt-in and a warning), but an actuator on one will silently scatter effort into `joint_f` that most solvers will ignore. Reject, warn, or consider valid?

Happy to defer either check to the importer if parser-side feels too heavy — just want to pin down the contract.

newton/_src/actuators/delay.py
```diff
@@ -0,0 +1,338 @@
...
+            requires_grad: Allocate output arrays with gradient support.
+        """
+        self._device = device
+        self._num_actuators = num_actuators
+        self._requires_grad = requires_grad
+        self._out_pos = wp.zeros(num_actuators, dtype=wp.float32, device=self._device, requires_grad=requires_grad)
+        self._out_vel = wp.zeros(num_actuators, dtype=wp.float32, device=self._device, requires_grad=requires_grad)
+        self._out_act = wp.zeros(num_actuators, dtype=wp.float32, device=self._device, requires_grad=requires_grad)
+
+    def state(self, num_actuators: int, device: wp.Device) -> Delay.State:
```
preist-nvidia, line 218: Related to the standalone-usability point on `clamping_max_force.py`: this reads `self._requires_grad` which is only assigned in `finalize()`, so direct construction followed by `.state()` (without going through `Actuator`/`ModelBuilder`) will `AttributeError`. Worth either initializing `_requires_grad = False` in `__init__` or gating with `getattr(self, "_requires_grad", False)` so the standalone path works.

newton/_src/actuators/actuator.py
```diff
@@ -0,0 +1,304 @@
...
+            control_feedforward_attr: Attribute on sim_control for control input. None to skip.
+            control_output_attr: Attribute on sim_control for clamped output effort.
+            control_computed_output_attr: Attribute on sim_control for raw (pre-clamp)
+                effort. None to skip writing computed effort.
+            requires_grad: Allocate intermediate arrays with gradient support
+                for differentiable simulation.
+        """
+        self.indices = indices
+        self.pos_indices = pos_indices if pos_indices is not None else indices
+        self.frc_indices = frc_indices if frc_indices is not None else indices
```
preist-nvidia: We don't validate that `frc_indices.shape == indices.shape`. A mismatched output-index array silently corrupts the scatter (out-of-bounds writes or truncation depending on which is shorter) rather than raising. Cheap check to add alongside the other shape assertions.

preist-nvidia (commented):
Overall this is fine to merge today — the architecture is solid and the remaining items are edge cases / polish. Let's open follow-up issues for the unresolved threads so they don't get lost, especially the schema-contract items (`delay_steps=0`,  target-type validation, multi-DOF guard) and the standalone-usability gaps (`Delay.state()` pre-`finalize`, `frc_indices` shape check, clamping-order contract).                                                                                                
                                                                                                                                  
Test coverage also needs expansion. The existing suite is strong on the happy path but has zero `assertRaises` — none of the  `ValueError` paths in `resolve_arguments` / component `__init__` are exercised, and the standalone-construction path (building `Actuator` directly from Warp arrays without `ModelBuilder`) isn't covered end-to-end. Several of the bugs flagged here (`delay_steps=0`, pre-`finalize` `.state()`, shape mismatches) would have been caught by an error-path sweep and one standalone  integration test.

Maybe good to specifically prompt a delta to the new usd schema param ranges and design decisions.

newton/_src/actuators/delay.py
jvonmuralt: It is a max_delay which is a maximum delay in the whole simulation, this determined the size of the buffer. If all the dofs have 0 delay then there is no need to have delay.

preist-nvidia: Ah ok got it thanks!

newton/_src/actuators/actuator.py
```diff
@@ -0,0 +1,304 @@
...
+    output[idx] = output[idx] + forces[i]
+    if computed_output:
+        computed_output[idx] = computed_output[idx] + computed_forces[i]
+
+
+class Actuator:
+    """Composed actuator: delay → controller → clamping.
+
+    An actuator reads from simulation state/control arrays, optionally
+    delays the control targets, computes effort via a controller, applies
```
preist-nvidia: Per schema PR #45, the settled umbrella term for `target_pos`/`target_vel`/`feedforward` is **command inputs** — e.g. `NewtonActuatorDelayAPI` says *"Command inputs such as control targets and feedforward terms are delayed..."*. This docstring still says *"delays the control targets"*, which is narrower and leaves feedforward implicit even though delay already applies to it. Suggest reading *"delays the command inputs (control targets and feedforward)"*. Same swap applies in `delay.py` — class docstring at line 109, kernel doc at line 69, method docstrings at lines 245/296 all say "targets" where they mean all three command inputs. Also `actuator.py:116` has "Attribute on sim_control for control input" on the `control_feedforward_attr` param — worth saying "feedforward" explicitly there to avoid mixing umbrella and specific terms.

newton/_src/actuators/delay.py
```diff
@@ -0,0 +1,338 @@
...
+        delay_steps = args["delay_steps"]
+        if delay_steps < 0:
+            raise ValueError(f"delay_steps must be >= 0, got {delay_steps}")
+        return {"delay_steps": delay_steps}
+
+    def __init__(self, delay_steps: wp.array[int], max_delay: int):
+        """Initialize delay.
+
+        Args:
+            delay_steps: Per-DOF delay values [timesteps], shape ``(N,)``.
```
preist-nvidia: Schema consistently says *"actuator timesteps"* (e.g. line 545, 555) rather than bare *"timesteps"*. Once actuator dt may diverge from sim dt, this annotation becomes ambiguous. Suggest `[actuator timesteps]` here and on the matching class attribute docstring at line 195.

jvonmuralt: done

newton/_src/actuators/actuator.py
jvonmuralt: done

newton/_src/actuators/delay.py
jvonmuralt: done

newton/_src/actuators/actuator.py
jvonmuralt: done

newton/_src/actuators/usd_parser.py
jvonmuralt: 1. added a raise on unsupported joint types
2. I would suggest to defer this to the parser side

newton/_src/utils/import_usd.py
jvonmuralt: done

docs/concepts/actuators.rst
```diff
@@ -0,0 +1,260 @@
...
+
+An actuator is composed from three building blocks, applied in this order:
+
+.. code-block:: text
+
+   Actuator
+   ├── Delay       (optional: delays command inputs by N actuator timesteps)
+   ├── Controller  (control law that computes raw effort)
+   └── Clamping[]  (clamps raw effort based on motor-limit modeling)
+       ├── ClampingMaxEffort        (±max_effort box clamp)
```
preist-nvidia: symmetric clamp instead of box clamp

docs/concepts/actuators.rst
jvonmuralt: yes, in impl it is in the list order

docs/concepts/actuators.rst
```diff
@@ -0,0 +1,260 @@
...
+  PyTorch, stateful: position/velocity history buffers).
+* :class:`ControllerNeuralLSTM` — LSTM neural-network controller (requires
+  PyTorch, stateful: hidden/cell state).
+
+See the API documentation for each controller's control-law equations.
+
+Clamping
+^^^^^^^^
+
+* :class:`ClampingMaxEffort` — symmetric box clamp to ±max_effort per actuator.
```
preist-nvidia: Remove box here.

jvonmuralt:
> Overall this is fine to merge today — the architecture is solid and the remaining items are edge cases / polish. Let's open follow-up issues for the unresolved threads so they don't get lost, especially the schema-contract items (`delay_steps=0`, target-type validation, multi-DOF guard) and the standalone-usability gaps (`Delay.state()` pre-`finalize`, `frc_indices` shape check, clamping-order contract).
> 
> Test coverage also needs expansion. The existing suite is strong on the happy path but has zero `assertRaises` — none of the `ValueError` paths in `resolve_arguments` / component `__init__` are exercised, and the standalone-construction path (building `Actuator` directly from Warp arrays without `ModelBuilder`) isn't covered end-to-end. Several of the bugs flagged here (`delay_steps=0`, pre-`finalize` `.state()`, shape mismatches) would have been caught by an error-path sweep and one standalone integration test.
> 
> Maybe good to specifically prompt a delta to the new usd schema param ranges and design decisions.

Added a follow up task for testing https://github.com/newton-physics/newton/issues/2562

docs/concepts/actuators.rst
```diff
@@ -0,0 +1,260 @@
...
+.. testcode:: actuator-usage
+
+   builder.add_actuator(
+       ControllerPD,
+       index=dof_index,
+       kp=100.0,
+       kd=10.0,
+       delay_steps=5,
+       clamping=[(ClampingMaxEffort, {"max_effort": 50.0})],
+   )
```
jeff-hough, line 112: Question: for the clamping array, do we allow multiple clamps of the same kind? I.e.
```python
clamping=[
  (ClampingDCMotor, {params...}),
  (ClampingDCMotor, {other params...})
],
```
An actuator like the one above could be useful for somebody trying to model a more complicated torque-speed envelope, however we wouldn't be able to express it using our USD schema?

If this is a real mismatch, should we document it here (or somewhere else)? i.e., let the users know they will only be able to author their actuator to USD if they don't use multiple of the same clamping type?

@jvonmuralt @andrewkaufman

jvonmuralt: From the implementation side, it does not matter and it would allow multiple clamps of the same type. It is a limitation (or design choice) from the usd side

newton/_src/actuators/clamping/clamping_dc_motor.py
```diff
@@ -0,0 +1,124 @@
...
+    effort_min(vel) = max(saturation_effort * (-1 - vel / velocity_limit), -max_motor_effort)
+    """
+    i = wp.tid()
+    state_idx = state_indices[i]
+    vel = current_vel[state_idx]
+    sat = saturation_effort[i]
+    vel_lim = velocity_limit[i]
+    max_e = max_motor_effort[i]
+
+    effort_max = wp.min(sat * (1.0 - vel / vel_lim), max_e)
```
jeff-hough, line 56: Should this actually be:
```
effort_max = wp.min(wp.max(0, sat * (1.0 - vel / vel_lim)), max_e)
```
I.e., preventing the case where `vel` has become slightly larger than `vel_lim`, which would cause `effort_min > effort_max` in this line:

```
dst[i] = wp.clamp(src[i], effort_min, effort_max)
```

newton/_src/actuators/controllers/base.py
```diff
@@ -0,0 +1,136 @@
...
+            pos_indices: Indices into *positions* for each DOF.
+            vel_indices: Indices into *velocities* for each DOF.
+            target_pos_indices: Indices into *target_pos*.
+            target_vel_indices: Indices into *target_vel* and *feedforward*.
+            forces: Scratch buffer to write effort [N or N·m] to. Shape ``(N,)``.
+            state: Controller state (``None`` if stateless).
+            dt: Timestep [s].
+            device: Warp device for kernel launches.
+        """
+        raise NotImplementedError
```
jeff-hough: Missing error message?

newton/_src/actuators/clamping/clamping_dc_motor.py
jvonmuralt: Thanks, it is good point. Your max(0, ...) fix prevents the inversion but would also kill braking torque in the generator quadrant. I did with a corner velocity pre-clamp instead, clamps the velocity to where the torque-speed curve meets ±max_motor_effort before evaluating

newton/_src/actuators/controllers/base.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def is_stateful(self) -> bool:
+        """Return True if this controller maintains internal state."""
+        return False
+
+    def is_graphable(self) -> bool:
+        """Return True if compute() can be captured in a CUDA graph."""
+        return True
```
jeff-hough: I think there is an argument that `raise NotImplementedError(...)` is a better default than an explicit `return False` or `return True`.

For somebody implementing a new `Controller`, these defaults make it very easy to build a stateful controller which returns `False` on `is_stateful`, or a non-graphable controller which returns `True` on `is_graphable`. Further, reviewers of the new `Controller` might forget these functions exist (since they won't be in the diff).

If the default is `NotImplementedError`, then:

1. The person implementing the controller is forced to think about it.
2. The people reviewing the controller will see their choice, and will also have to think about it.

The actual implementation of these functions will always just be `return True` or `return False`. I don't feel forcing that implementation is a large burden.

Thoughts? @jvonmuralt @andrewkaufman @preist-nvidia

newton/_src/actuators/controllers/controller_pid.py
```diff
@@ -0,0 +1,197 @@
...
+            raise ValueError(f"kp must be non-negative, got {kp}")
+        ki = args.get("ki", 0.0)
+        if ki < 0:
+            raise ValueError(f"ki must be non-negative, got {ki}")
+        kd = args.get("kd", 0.0)
+        if kd < 0:
+            raise ValueError(f"kd must be non-negative, got {kd}")
+        integral_max = args.get("integral_max", math.inf)
+        if integral_max < 0:
+            raise ValueError(f"integral_max must be non-negative, got {integral_max}")
```
jeff-hough, line 99: Is it worth raising a warning if `kp==kd==ki==0`? That is probably not (though could be sometimes) the intention of the user?

preist-nvidia (dismissed):
Good from my side - can address more in follow-ups and when you are working on lab integration.

newton/_src/actuators/controllers/base.py
jvonmuralt: done

newton/_src/actuators/controllers/controller_pid.py
jvonmuralt: All-zero gains is valid,  it gives pure feedforward pass-through

jeff-hough (approved):
All of my comments are addressed, looks good to me!

PR #2427, opened 2026-04-13
Bump warp-lang to 1.13.0 dev nightly

adenzler-nvidia:
## Description

Bump warp-lang to 1.13.0.dev20260413 for 1.2 development cycle.

- Add `prerelease = "allow"` to `[tool.uv]` in `pyproject.toml` so `uv lock` (including the pre-commit hook) resolves the warp-lang nightly correctly
- Update `uv.lock` to warp-lang 1.13.0.dev20260413
- Update `asv.conf.json` benchmark pin to warp-lang 1.13.0.dev20260413

## Checklist

- [x] New or existing tests cover these changes
- [x] The documentation is up to date with these changes
- [ ] `CHANGELOG.md` has been updated (if user-facing change)

## Test plan

CI will validate that the nightly warp-lang wheel installs and all tests pass.

## Summary by CodeRabbit

* **Chores**
  * Updated configuration to enable pre-release package version support
  * Upgraded dependency to a more recent development version

PR #2452, opened 2026-04-15
Include CPU ISA hash in Warp kernel cache key

adenzler-nvidia:
## Summary

- Warp 1.13+ compiles CPU kernels with `-march=native`, which emits instructions
  specific to the compiling CPU. GitHub Actions runners vary in CPU model (Intel
  Ice Lake, AMD EPYC, etc.), so restoring a kernel cache built on one CPU onto a
  runner with a different ISA causes illegal-instruction crashes.
- Add `scripts/ci/cpu_isa_hash.py` that detects the host CPU's ISA feature set
  and prints a stable 16-char hex hash. Include this hash in the CI cache key so
  kernels are only reused on runners with matching instruction sets.
- Supports x86_64 (Linux/macOS/Windows via system C compiler), AArch64 Linux
  (via `/proc/cpuinfo`), and AArch64 macOS (via `sysctl`).

## Context

After the Warp 1.13 dev nightly bump (#2427), CI started hitting `Fatal Python
error: Illegal instruction` / `0xc000001d` on both Windows and Ubuntu runners.
The root cause: Warp 1.13 added a `cpu_compiler_flags` option that defaults to
`-march=native`, causing its bundled LLVM to emit CPU-specific instructions.
When the GH Actions cache restores kernel objects compiled on e.g. an Intel Ice
Lake runner (with AVX-512) onto an AMD EPYC runner (without AVX-512), the
process crashes on the first kernel invocation.

The previous cache key (`warp-kernels-OS-ARCH-<code-hash>`) did not account for
CPU differences, and the `restore-keys` prefix fallback made cross-CPU cache
reuse likely.

## Test plan

- [ ] Verify the `cpu-id` step runs and prints a hash on all four matrix runners
      (ubuntu-latest, ubuntu-24.04-arm, windows-latest, macos-latest)
- [ ] Verify the cache key in the logs includes the CPU hash
      (e.g. `warp-kernels-Linux-X64-0723c9b174ec6c08-...`)
- [ ] Verify no more illegal-instruction crashes on subsequent runs

### DIFF
--- a/scripts/ci/cpu_isa_hash.py
+++ b/scripts/ci/cpu_isa_hash.py
@@ -1,0 +1,114 @@
+#!/usr/bin/env python3
+# Copyright (c) 2025 NVIDIA CORPORATION. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Print a short hash of the host CPU's ISA feature set.
+
+Warp 1.13+ compiles CPU kernels with ``-march=native``, so cached object
+files are only safe to reuse on a CPU with the same instruction set.
+This script detects the ISA features and prints a 16-character hex hash
+suitable for use in a CI cache key.
+
+Supported platforms:
+  - x86_64 Linux / macOS / Windows (via system C compiler)
+  - AArch64 Linux (via /proc/cpuinfo)
+  - AArch64 macOS (via sysctl)
+"""
+
+import hashlib
+import platform
+import subprocess
+import sys
+
+# ISA-related macro keywords emitted by ``cc -march=native -dM -E``.
+# These correspond to the instruction set extensions that affect codegen.
+_X86_ISA_KEYWORDS = (
+    "ADX",
+    "AES",
+    "AVX",
+    "BMI",
+    "CLFLUSH",
+    "F16C",
+    "FMA",
+    "LZCNT",
+    "MMX",
+    "MOVBE",
+    "PCLMUL",
+    "POPCNT",
+    "RDRAND",
+    "RDSEED",
+    "SHA",
+    "SSE",
+    "VAES",
+    "VPCLMUL",
+)
+
+
+def _x86_features() -> str:
+    """Query ISA features by asking the system C compiler what -march=native enables."""
+    for cc in ("cc", "gcc", "clang"):
+        try:
+            out = subprocess.check_output(
+                [cc, "-march=native", "-dM", "-E", "-x", "c", "-"],
+                input="",
+                text=True,
+                stderr=subprocess.DEVNULL,
+            )
+        except (FileNotFoundError, subprocess.CalledProcessError):
+            continue
+
+        macros = sorted(
+            line.split()[1]
+            for line in out.splitlines()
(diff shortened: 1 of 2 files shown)


adenzler-nvidia:
Marking as draft — waiting on the upstream Warp fix that adds CPU ISA-aware module hashing and load-time feature validation directly in Warp. Once that ships, we should rework this PR to use Warp's own CPU feature detection for the cache key instead of rolling our own detection script.

adenzler-nvidia:
@shi-eric we won't need this anymore with your upstream changes, right?

shi-eric:
> @shi-eric we won't need this anymore with your upstream changes, right?

In principle we shouldn’t. Can you test a few times by repeatedly triggering a pr that updates Warp?

Note that you can’t update to the bleeding edge nightly due to the issue I mentioned on Slack. Have to choose a nightly from a few days ago.

scripts/ci/cpu_isa_hash.py
```diff
@@ -0,0 +1,114 @@
...
+    # Linux: kernel exposes HWCAP flags in /proc/cpuinfo.
+    try:
+        with open("/proc/cpuinfo") as f:
+            for line in f:
+                if line.startswith("Features"):
+                    return " ".join(sorted(line.split(":")[1].split()))
+    except FileNotFoundError:
+        pass
+
+    # macOS: sysctl exposes CPU features.
```
eric-heiden, line 82: 🟡 `machdep.cpu.features` only exists on Intel Macs, so on Apple Silicon runners the `sysctl -n machdep.cpu.features` call below raises `CalledProcessError`, `_aarch64_features()` returns `""`, and `main()` falls back to `platform.processor()` (which is `"arm"` on Apple Silicon). On `macos-latest` the cache key ends up not being derived from ISA features at all, so that runner effectively keeps the pre-change behavior and one of the four CI targets does not benefit from this change.

The fallback itself is safe (no crash, stable hash), but the module docstring on line 15 advertises "AArch64 macOS (via sysctl)" support that is not actually working on Apple Silicon.

On Apple Silicon, `sysctl hw.optional` enumerates per-feature flags such as `hw.optional.neon` and `hw.optional.armv8_2_sha3`. Parsing and sorting those keys gives a real ISA fingerprint.

<details>

<summary>Example macOS ARM branch</summary>

```python
# macOS Apple Silicon: hw.optional.* enumerates ISA features.
try:
    out = subprocess.check_output(
        ["sysctl", "-a"],
        text=True,
        stderr=subprocess.DEVNULL,
    )
    features = sorted(
        line.split(":", 1)[0].strip()
        for line in out.splitlines()
        if line.startswith("hw.optional.") and line.rstrip().endswith(": 1")
    )
    if features:
        return " ".join(features)
except (FileNotFoundError, subprocess.CalledProcessError):
    pass
```

</details>

Worth updating the docstring once the macOS ARM path actually contributes ISA features to the hash.

adenzler-nvidia:
closing as not needed with upstream changes.

PR #2260, opened 2026-03-30
Fix SolverMuJoCo NATIVECCD margin rejection

adenzler-nvidia:
## Description

Upstream `mujoco_warp` (post-3.6.0) rejects non-zero geom/pair margins at `put_model()` time when NATIVECCD is enabled, causing `NotImplementedError` in CI.

This PR zeroes margins in the MJCF spec so `put_model()` succeeds, and conditionally keeps them zero at runtime when MuJoCo handles collisions (`use_mujoco_contacts=True`). When Newton handles contacts (`use_mujoco_contacts=False`), margins are restored from `shape_margin` via the existing update kernel — the Newton collision pipeline is unchanged.

Closes #2106

## Checklist

- [x] New or existing tests cover these changes
- [x] The documentation is up to date with these changes
- [x] `CHANGELOG.md` has been updated (if user-facing change)

## Test plan

```bash
# New regression tests (5 tests)
uv run --extra dev -m newton.tests -k "test_mujoco_margin_zeroing.TestMuJoCoMarginZeroing"

# Updated existing tests
uv run --extra dev -m newton.tests -k "test_mujoco_solver.TestMuJoCoSolverGeomProperties.test_geom_margin_from_shape_margin"
uv run --extra dev -m newton.tests -k "test_import_mjcf.TestImportMjcfSolverParams.test_margin_gap_mujoco_solver"
uv run --extra dev -m newton.tests -k "test_mujoco_solver.TestMuJoCoSolverPairProperties.test_pair_properties_conversion_and_update"

# Full MuJoCo test suite (570 tests, 1 pre-existing hydroelastic failure)
uv run --extra dev -m newton.tests -k "mujoco"
```

## Bug fix

**Steps to reproduce:**

1. Create a model with shapes that have non-zero margin (e.g. via MJCF import)
2. Create `SolverMuJoCo(model)` with a `mujoco_warp` version that enforces NATIVECCD margin checks

**Minimal reproduction:**

```python
import newton

builder = newton.ModelBuilder()
builder.add_shape_box(
    body=-1, hx=1.0, hy=1.0, hz=0.01,
    cfg=newton.ModelBuilder.ShapeConfig(margin=1e-5),
)
b = builder.add_body(label="box")
builder.add_shape_box(
    body=b, hx=0.05, hy=0.05, hz=0.05,
    cfg=newton.ModelBuilder.ShapeConfig(margin=1e-5),
)
model = builder.finalize()

[...]

PR #2455, opened 2026-04-15
Bump mujoco/mujoco-warp to 3.7.0(.1), fix NATIVECCD margin rejection

### LINKED ISSUE #2106: SolverMuJoCo: NATIVECCD margin NotImplementedError from upstream mujoco_warp breaking change
## Description

Since 2026-03-14, nightly CI fails with a `NotImplementedError` from upstream `mujoco_warp` when NATIVECCD is enabled and geom pairs have non-zero margin.

## Error

```
NotImplementedError: geom pair (shape_1_1, shape_2_2) has non-zero margin
((9.999999747378752e-06, 9.999999747378752e-06)) with NATIVECCD enabled.
Set margin to 0 or disable NATIVECCD.
```

Raised in `mujoco_warp/_src/io.py`, line 389 — inside `mujoco_warp.put_model()`.

## Affected Tests

- `test_hydroelastic.TestHydroelastic.test_stacked_small_primitive_cubes_hydroelastic_mujoco_warp_cuda_0`
- `test_examples.TestContactsExamples.test_contacts.example_brick_stacking_cuda_0`

## Failing Nightly Runs

- [Run 23111038904](https://github.com/newton-physics/newton/actions/runs/23111038904) — 2026-03-15
- [Run 23088584287](https://github.com/newton-physics/newton/actions/runs/23088584287) — 2026-03-14

## Root Cause

Upstream `mujoco_warp` now enforces `margin=0` when NATIVECCD is enabled. A package bump between Mar 13–14 introduced this enforcement. Newton's `SolverMuJoCo` passes geom pairs with a small non-zero margin (~1e-5), which is no longer accepted.

## Proposed Fix

In Newton's MuJoCo binding (`SolverMuJoCo`), either:
1. Zero out margins for geom pairs before calling `mujoco_warp.put_model()` when NATIVECCD is enabled, or
2. Disable NATIVECCD for affected geom pairs that have non-zero margin.

A short-term workaround is to pin `mujoco_warp` to the previous version until a proper fix is in place.

adenzler-nvidia:
## Description

Update `mujoco` and `mujoco-warp` dependency pins from `~=3.6.0` to `~=3.7.0` (`mujoco-warp` requires `>=3.7.0.1` since 3.7.0 was broken).

Includes the NATIVECCD margin fix from #2260: upstream `mujoco_warp` rejects non-zero geom/pair margins at `put_model()` time when NATIVECCD is enabled, causing `NotImplementedError` in CI. Margins are now zeroed in the MJCF spec so `put_model()` succeeds, and conditionally kept zero at runtime when MuJoCo handles collisions (`use_mujoco_contacts=True`). When Newton handles contacts (`use_mujoco_contacts=False`), margins are restored from `shape_margin` via the existing update kernel — the Newton collision pipeline is unchanged.

Closes #2106
Supersedes #2260

## Checklist

## Test plan

```bash
# New regression tests (4 tests)
uv run --extra dev -m newton.tests -k "test_mujoco_margin_zeroing.TestMuJoCoMarginZeroing"

# Updated existing tests
uv run --extra dev -m newton.tests -k "test_mujoco_solver.TestMuJoCoSolverGeomProperties.test_geom_margin_from_shape_margin"
uv run --extra dev -m newton.tests -k "test_import_mjcf.TestImportMjcfSolverParams.test_margin_gap_mujoco_solver"
uv run --extra dev -m newton.tests -k "test_mujoco_solver.TestMuJoCoSolverPairProperties.test_pair_properties_conversion_and_update"

# Full MuJoCo test suite
uv run --extra dev -m newton.tests -k "mujoco"
```

## Bug fix

**Steps to reproduce:**

1. Create a model with shapes that have non-zero margin (e.g. via MJCF import)
2. Create `SolverMuJoCo(model)` with `mujoco-warp>=3.7.0`

**Minimal reproduction:**

```python
import newton

builder.add_shape_box(
    body=-1, hx=1.0, hy=1.0, hz=0.01,
    cfg=newton.ModelBuilder.ShapeConfig(margin=1e-5),
)
b = builder.add_body(label="box")
builder.add_shape_box(
    body=b, hx=0.05, hy=0.05, hz=0.05,
    cfg=newton.ModelBuilder.ShapeConfig(margin=1e-5),
)
model = builder.finalize()

# Raises NotImplementedError with mujoco_warp NATIVECCD enforcement:
# "geom pair has non-zero margin with NATIVECCD enabled"
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserve zeroed contact margins when MuJoCo contact handling is enabled to avoid native-CCD failures.

* **New Features**
  * Restore non-zero shape/pair margins when opting out of MuJoCo contact handling so previous margin behavior is retained.

* **Tests**
  * Added regression tests verifying geom/pair margin behavior across contact-backend configurations.

* **Chores**
  * Bumped MuJoCo and mujoco-warp requirement pins to the 3.7.x series (mujoco-warp requires >=3.7.0.1).

### DIFF
--- a/newton/_src/solvers/mujoco/solver_mujoco.py
+++ b/newton/_src/solvers/mujoco/solver_mujoco.py
@@ -1998,8 +1998,6 @@ class SolverMuJoCo(SolverBase):
         pair_solref = get_numpy("pair_solref")
         pair_solreffriction = get_numpy("pair_solreffriction")
         pair_solimp = get_numpy("pair_solimp")
-        pair_margin = get_numpy("pair_margin")
-        pair_gap = get_numpy("pair_gap")
         pair_friction = get_numpy("pair_friction")
 
         for i in range(pair_count):
@@ -2041,10 +2039,11 @@ class SolverMuJoCo(SolverBase):
                 pair_kwargs["solreffriction"] = pair_solreffriction[i].tolist()
             if pair_solimp is not None:
                 pair_kwargs["solimp"] = pair_solimp[i].tolist()
-            if pair_margin is not None:
-                pair_kwargs["margin"] = float(pair_margin[i])
-            if pair_gap is not None:
-                pair_kwargs["gap"] = float(pair_gap[i])
+            # Zeroed in the spec for NATIVECCD compatibility (#2106).
+            # When use_mujoco_contacts=False, real values are written back
+            # at runtime via notify_model_changed -> _update_pair_properties().
+            pair_kwargs["margin"] = 0.0
+            pair_kwargs["gap"] = 0.0
             if pair_friction is not None:
                 pair_kwargs["friction"] = pair_friction[i].tolist()
 
@@ -2916,6 +2915,12 @@ class SolverMuJoCo(SolverBase):
 
         self._viewer = None
         """Instance of the MuJoCo viewer for debugging."""
+
+        self._use_mujoco_contacts = use_mujoco_contacts
+        """Whether MuJoCo handles collision detection.
+
+        Controls margin zeroing: when True, geom/pair margins on the MuJoCo
+        model are kept at zero for NATIVECCD compatibility (#2106)."""
 
         enableflags = 0
         if enable_multiccd:
@@ -4014,7 +4019,6 @@ class SolverMuJoCo(SolverBase):
         shape_kd = model.shape_material_kd.numpy()
         shape_mu_torsional = model.shape_material_mu_torsional.numpy()
         shape_mu_rolling = model.shape_material_mu_rolling.numpy()
-        shape_margin = model.shape_margin.numpy()
 
         # retrieve MuJoCo-specific attributes
         mujoco_attrs = getattr(model, "mujoco", None)
@@ -4388,7 +4392,7 @@ class SolverMuJoCo(SolverBase):
                 if shape_geom_solmix is not None:
                     geom_params["solmix"] = shape_geom_solmix[shape]
                 geom_params["gap"] = 0.0
-                geom_params["margin"] = float(shape_margin[shape])
+                geom_params["margin"] = 0.0
 
                 body.add_geom(**geom_params)
                 # store the geom name instead of assuming index
@@ -6094,6 +6098,7 @@ class SolverMuJoCo(SolverBase):
                 shape_geom_solimp,
                 shape_geom_solmix,
                 self.model.shape_margin,
+                int(self._use_mujoco_contacts),
             ],
             outputs=[
                 self.mjw_model.geom_friction,
@@ -6130,8 +6135,9 @@ class SolverMuJoCo(SolverBase):
         pair_solref = getattr(mujoco_attrs, "pair_solref", None)
         pair_solreffriction = getattr(mujoco_attrs, "pair_solreffriction", None)
         pair_solimp = getattr(mujoco_attrs, "pair_solimp", None)
-        pair_margin = getattr(mujoco_attrs, "pair_margin", None)
-        pair_gap = getattr(mujoco_attrs, "pair_gap", None)
+        # Keep pair margin/gap at zero when MuJoCo handles collisions (#2106).
+        pair_margin = None if self._use_mujoco_contacts else getattr(mujoco_attrs, "pair_margin", None)
+        pair_gap = None if self._use_mujoco_contacts else getattr(mujoco_attrs, "pair_gap", None)
         pair_friction = getattr(mujoco_attrs, "pair_friction", None)
 
         # Only launch kernel if at least one attribute is defined
--- a/newton/_src/solvers/mujoco/solver_mujoco.py
+++ b/newton/_src/solvers/mujoco/solver_mujoco.py
@@ -2004,6 +2004,8 @@ class SolverMuJoCo(SolverBase):
         pair_solref = get_numpy("pair_solref")
         pair_solreffriction = get_numpy("pair_solreffriction")
         pair_solimp = get_numpy("pair_solimp")
+        pair_margin = get_numpy("pair_margin")
+        pair_gap = get_numpy("pair_gap")
         pair_friction = get_numpy("pair_friction")
 
         for i in range(pair_count):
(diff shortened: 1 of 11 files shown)


adenzler-nvidia:
@kenny-villela ready for review — merged upstream/main and cleaned up duplicate CHANGELOG entries.

newton/tests/test_equality_connect_constraint_with_sim_step.py
```diff
@@ -259,8 +259,8 @@ def _test_connect_constraint(self):
         offsets from the body origin.
         """
 
-        dt = 0.01
-        num_steps = 50
+        dt = 0.002
+        num_steps = 250
```
Kenny-Vilella, line 263: Why all of these changes?
Do we have less accuracy/stability after updating?

newton/tests/test_mujoco_margin_zeroing.py
```diff
@@ -0,0 +1,89 @@
...
+        builder.add_shape_box(
+            body=b,
+            hx=0.05,
+            hy=0.05,
+            hz=0.05,
+            cfg=newton.ModelBuilder.ShapeConfig(margin=margin),
+        )
+        return builder.finalize()
+
+    def test_solver_creation_with_nonzero_margin(self):
```
Kenny-Vilella: This test has very limited use no?

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -2939,6 +2938,12 @@ def __init__(
         self._viewer = None
         """Instance of the MuJoCo viewer for debugging."""
 
+        self._use_mujoco_contacts = use_mujoco_contacts
```
Kenny-Vilella, line 2960: It is a bit awkward that in some cases you will need to use: `disable_contacts=True, use_mujoco_contacts=False`.
I would expect that if contact is disabled, the value of `use_mujoco_contacts` does not matter.
Maybe I'm missing something?

Kenny-Vilella (commented):
Not sure to grasp the full implication of these changes.
First of all, do we agree that we are potentially ignoring user provided data by zeroing the margin?
If it is correct, then we should always be very careful with this, I would at least expect a warning.

Second, are we zeroing the margin in all cases? even in cases where it is not necessary because `NATIVECCD` is not used or support margin?

newton/tests/test_mujoco_margin_zeroing.py
adenzler-nvidia: yeah I agree it's just the opposite of the test below. I will remove.

adenzler-nvidia:
I do not see a solution where we would get away with not zeroing - we don't really have guaranteed explicit pairs in newton and anything with boxes/meshes could theoretically raise. The question is whether we should error out and tell the user, but then there are USD/MJCF models in the menagerie that would fail to compile like that.

We are a bit on the receiving end of a limitation here, not much we can do. Unless you have a better idea?

newton/tests/test_equality_connect_constraint_with_sim_step.py
adenzler-nvidia: The tests were failing after this update, so I had to reduce the timestep. 3.7.0 added new coriolis terms that somehow weren't liked by these tests.

newton/_src/solvers/mujoco/solver_mujoco.py
adenzler-nvidia: Good catch — I looked into this. The combination only shows up in margin-related tests (to trigger margin restoration so we can assert on `mjw_model.geom_margin`). No user-facing code combines them: `example_robot_ur10.py` and `example_selection_cartpole.py` use `disable_contacts=True` with the default `use_mujoco_contacts=True` and work fine. I think leaving this as-is is OK, but happy to consider alternatives if you feel strongly:

1. **Make `use_mujoco_contacts` a true no-op when `disable_contacts=True`** — gate the margin-zeroing on `use_mujoco_contacts and not disable_contacts`. Restores shape margins into `mjw_model` whenever contacts are disabled (harmless since nothing reads them). Tests would still work.
2. **Warn/raise when both flags are set together** — if `disable_contacts=True` and `use_mujoco_contacts=False`, emit a warning explaining that `use_mujoco_contacts` only controls margin bookkeeping in that case.
3. **Split the flag** — expose the margin-restoration side as its own parameter (`restore_shape_margins` or similar). Clean but adds API surface for something only tests care about.
4. **Leave as-is and document the dual role** of `use_mujoco_contacts` in the docstring.

My lean is (4) — the coupling is real (Newton's contact pipeline needs real margins, MuJoCo's NATIVECCD doesn't) and only tests hit the awkward combination. But let me know if (1) or something else appeals to you.

Kenny-Vilella:
I agree that we have to zero the margin, I do not question that.
My two points are:
- Except if it is against established convention, we should definitely have a warning when ignoring user provided value.
- Do we zero the margin only when necessary? For example if `NATIVECDD` is off, or if we have no box no mesh.

adenzler-nvidia:
Good points, both implemented in 3a9a0c5f6:

- **Conditional zeroing:** based on your observation, I checked `mujoco_warp/_src/io.py:386-394` — the check only fires for box-box pairs (NATIVECCD) or any box/mesh pair under MULTICCD. Newton now detects the shape types at solver construction and skips the entire workaround when no box shapes exist (and no mesh shapes, if `enable_multiccd=True`). Authored margins/gaps pass through to the spec unchanged for sphere/capsule/plane/etc. models.
- **Warning when overwriting user values:** when the workaround is active and `use_mujoco_contacts=True` (so margins are not restored at runtime), a per-geom and per-pair `UserWarning` is emitted identifying the geom/pair name, the authored value, and pointing users to Newton's collision pipeline (`use_mujoco_contacts=False` plus feeding Newton-generated contacts into `step()`) if they want the authored values honored.

The existing margin regression tests now also assert the warning fires via `assertWarnsRegex`.

newton/_src/solvers/mujoco/solver_mujoco.py
Kenny-Vilella: OK after taking a quick look, I feel that it is a `mujoco_warp` limitation.
`disable_contacts` is the equivalent in mjw world to `DisableBit.CONTACT`
`use_mujoco_contact` is the equivalent in mjw world to `opt.run_collision_detection`
The main issue is that from `mujoco_warp` side, the assert will be raised no matter the value of these flags.
We could fix that from newton side, but it is maybe better to keep the equivalence.

adenzler-nvidia: Agreed — keeping the `mujoco_warp` and classic `mujoco` paths in sync is the better option here. The classic-MuJoCo backend (`use_mujoco_cpu=True`) is primarily used for debugging, so diverging its spec behavior to preserve margins isn't worth the inconsistency. The current implementation already keys the zeroing off shape types and `MULTICCD` only, so both backends behave identically.

Kenny-Vilella (approved):

PR #2477, opened 2026-04-16
Collection of small collision pipeline performance improvements

nvtw:
## Description

Fuses a couple of clear operations into kernels, avoids some expensive operations that are not strictly needed etc.

## Checklist

## Test plan

     or manual steps. Example:
     ```
     ``` -->



## Summary by CodeRabbit

* **New Features**
  * Option to skip automatic zeroing of broad-phase pair counters.
  * Pipeline-wide toggle to enable/disable narrow-phase buffer verification.
  * Exposed control to optionally bump contact-generation when clearing contacts.
* **Performance**
  * Fused counter-clear and generation-bump into single GPU kernels to reduce launches.
  * Removed redundant counter zeroing during collision setup.
* **Documentation**
  * Changelog: convex-mesh AABB now uses precomputed local AABBs.


newton/_src/sim/collide.py
```diff
@@ -670,6 +681,7 @@ def __init__(
                 has_heightfields=has_heightfields,
                 use_lean_gjk_mpr=use_lean_gjk_mpr,
                 deterministic=deterministic,
+                verify_buffers=False,
```
adenzler-nvidia: What's the motivation for defaulting `verify_buffers=False` here? `verify_narrow_phase_buffers` is the only mechanism that warns users when any of the intermediate pair buffers or the final contact buffer overflow — silencing it for every `CollisionPipeline(model, ...)` user (which the class docstring calls "most users") means contacts and candidate pairs will get dropped silently with no hint that `rigid_contact_max` / `max_triangle_pairs` / `shape_pairs_max` need to be raised.

Was there a measured hotspot from the `dim=[1]` diagnostic kernel that justifies this, or is it mostly about launch overhead under graph capture? If the latter, would you be open to exposing `verify_buffers` on `CollisionPipeline.__init__` (defaulting to `True`) so power users who care about launch counts can opt out, rather than flipping the default for everyone?

newton/_src/sim/collide.py
```diff
@@ -159,24 +159,36 @@ def compute_shape_aabbs(
...
-    Uses support function for most shapes. Meshes and heightfields use the
-    pre-computed local AABB transformed to world frame. Infinite planes use
-    bounding sphere fallback.  AABBs are enlarged by per-shape effective gap
-    for contact detection.  Effective expansion is ``shape_margin + shape_gap``.
+    Fuses AABB computation, narrow-phase data preparation, contact counter
+    zeroing, and generation bumping into a single kernel launch.
     """
     shape_id = wp.tid()
 
+    # Thread 0: zero contact counters, bump generation, zero broad phase count.
```
adenzler-nvidia: Small invariant shift worth a sanity check: before this PR, `contacts.clear()` + `broad_phase_pair_count.zero_()` + the generation bump ran unconditionally at the top of `collide()`. They now live inside thread 0 of `compute_shape_aabbs`, whose launch is `dim=model.shape_count` — so when `shape_count == 0` the fused kernel doesn't run and those counters / generation don't get reset each frame.

`shape_count` is immutable after model-build, so in practice this only matters if someone calls `collide()` on a shape-less model (and even then the counters would already be zero from init). Is there any path that exercises this combination today, or is it degenerate-by-construction and fine to leave?

Also minor: the comment above says thread 0 "zeroes counter arrays", but it also bumps `contact_generation` — worth mentioning so the next reader isn't surprised.

newton/_src/sim/collide.py
```diff
@@ -159,24 +159,36 @@ def compute_shape_aabbs(
...
+    if shape_id == 0:
+        for c in range(num_contact_counters):
+            contact_counters[c] = 0
+        g = contact_generation[0]
+        if g >= 2147483647:
```
adenzler-nvidia: Nit: `>= 2147483647` here vs `== 2147483647` in `contacts.py:35` (`_clear_counters_and_bump_generation`). Both safe on int32 (you can't exceed INT32_MAX), but inconsistent — pre-PR `_increment_contact_generation` used `==`, so keeping `==` would match.

newton/_src/sim/collide.py
```diff
@@ -195,20 +207,19 @@ def compute_shape_aabbs(
     effective_gap = margin + shape_gap[shape_id]
     margin_vec = wp.vec3(effective_gap, effective_gap, effective_gap)
 
-    # Check if this is an infinite plane, mesh, or heightfield
+    # Check if this is an infinite plane or a shape with a pre-computed local AABB
     scale = shape_scale[shape_id]
     is_infinite_plane = (geo_type == GeoType.PLANE) and (scale[0] == 0.0 and scale[1] == 0.0)
-    is_mesh = geo_type == GeoType.MESH
-    is_hfield = geo_type == GeoType.HFIELD
+    has_local_aabb = geo_type == GeoType.MESH or geo_type == GeoType.HFIELD or geo_type == GeoType.CONVEX_MESH
```
adenzler-nvidia, line 214: This looks like a behavior change, not just a perf change — `CONVEX_MESH` previously fell through to the support-function AABB path and now takes the pre-computed-local-AABB branch. Is `shape_collision_aabb_lower/upper` guaranteed to be populated correctly for `CONVEX_MESH` shapes at model-build time? If yes, this is a nice speedup worth a CHANGELOG line and a mention in the PR description. If not, we have a silent correctness regression for convex-mesh scenes.

newton/_src/sim/collide.py
```diff
@@ -879,6 +895,7 @@ def collide(
                 self.broad_phase_shape_pairs,
                 self.broad_phase_pair_count,
                 device=self.device,
+                skip_count_zero=True,  # Already zeroed by compute_shape_aabbs
```
adenzler-nvidia, line 912: The fused `broad_phase_pair_count[0] = 0` in `compute_shape_aabbs` is only taken advantage of on the EXPLICIT path. `BroadPhaseSAP.launch()` and `BroadPhaseAllPairs.launch()` still unconditionally call `candidate_pair_count.zero_()` internally, so for those two backends we're now doing the zero twice per frame (once as a redundant thread-0 write, once in the broad-phase internal). Safe but wasteful.

Worth either plumbing `skip_count_zero=True` through to SAP and AllPairs, or dropping the zero from the fused kernel and keeping per-backend zeroing — otherwise the kernel-fusion story is incomplete for 2/3 broad-phase backends.

newton/_src/geometry/narrow_phase.py
```diff
@@ -1430,6 +1430,7 @@ def __init__(
         use_lean_gjk_mpr: bool = False,
         deterministic: bool = False,
         contact_max: int | None = None,
+        verify_buffers: bool = True,
```
adenzler-nvidia, line 1433: The new `verify_buffers` kwarg isn't in the `Args:` docstring block. Worth documenting it — what it gates (the `verify_narrow_phase_buffers` launch at the end of `launch()`) and the tradeoff (losing warnings on dropped candidate pairs / contacts).

newton/_src/geometry/broad_phase_nxn.py
```diff
@@ -410,6 +410,7 @@ def launch(
         candidate_pair: wp.array[wp.vec2i],  # Array to store overlapping shape pairs
         candidate_pair_count: wp.array[int],
         device: Devicelike | None = None,  # Device to launch on
+        skip_count_zero: bool = False,  # Skip candidate_pair_count.zero_() if already zeroed
```
adenzler-nvidia, line 419: The new `skip_count_zero` kwarg creates an invariant-by-naming: the caller must have pre-zeroed `candidate_pair_count`. Worth a one-liner in the docstring making the precondition explicit, so future callers don't set it without doing the zero themselves.

newton/_src/sim/contacts.py
```diff
@@ -229,14 +248,12 @@ def clear(self):
         If clear_buffers=True (conservative mode), performs full buffer clearing with sentinel
         values and zeros. This requires 7-10 kernel launches but may be useful for debugging.
         """
-        # Clear all counters with a single kernel launch (consolidated counter array)
-        self._counter_array.zero_()
-
-        # Bump generation so solvers know the contact set changed (graph-capture safe)
+        # Clear all counters and bump generation in a single kernel launch.
```
adenzler-nvidia: The `clear()` docstring above (line 245) still says the fast path "[requires] just 2 kernel launches" — after this PR it's 1 launch (the fused `_clear_counters_and_bump_generation`). Worth updating the number. Also check whether the "86-90% fewer kernel launches" claim in the class docstring is still calibrated.

adenzler-nvidia (commented):
Directionally good — kernel fusion for counter-zero / generation-bump is graph-capture friendly, and the `_clear_active_kernel` split in the second commit (with the docstring explaining why the intra-launch ordering needs the kernel boundary) is the right call and reads well. A few questions inline, mostly around `verify_buffers=False` becoming the new default for all `CollisionPipeline` users, plus some docstring / API-hygiene cleanups.

newton/_src/sim/collide.py
```diff
@@ -797,11 +819,10 @@ def collide(
...
 
-        contacts.clear()
-        # TODO: validate contacts dimensions & compatibility
-
-        # Clear counters
-        self.broad_phase_pair_count.zero_()
+        # Counter zeroing and generation bump are fused into compute_shape_aabbs.
+        # Only call contacts.clear() if clear_buffers mode is enabled (debug path).
+        if contacts.clear_buffers:
+            contacts.clear()
```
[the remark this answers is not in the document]

newton/_src/sim/contacts.py
```diff
@@ -223,27 +242,26 @@ def clear(self):
...
         wp.launch(
-            _increment_contact_generation,
-            dim=1,
-            inputs=[self.contact_generation],
+            _clear_counters_and_bump_generation,
+            dim=max(num_counters, 1),
+            inputs=[self._counter_array, self.contact_generation, num_counters],
             device=self.contact_generation.device,
             record_tape=False,
         )
```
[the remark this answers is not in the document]
nvtw: Done

nvtw:
@adenzler-nvidia your comments should be addressed

newton/_src/sim/collide.py
```diff
@@ -829,6 +853,10 @@ def collide(
                 model.shape_gap,
                 model.shape_collision_aabb_lower,
                 model.shape_collision_aabb_upper,
+                contacts._counter_array,
+                contacts.contact_generation,
+                self.broad_phase_pair_count,
+                contacts._counter_array.shape[0],
```
adenzler-nvidia: Reaching into `contacts._counter_array` (and its `.shape[0]`) from the pipeline crosses an underscore boundary — `Contacts` otherwise exposes counter state via sliced views (`rigid_contact_count`, `soft_contact_count`). Is that coupling intentional? If `Contacts` ever grows a third counter or changes layout, this call site would need updating in lockstep with nothing to remind us.

newton/_src/geometry/narrow_phase.py
```diff
@@ -2097,30 +2106,31 @@ def launch_custom_write(
...
+                    self.shape_pairs_mesh_plane.shape[0] if self.shape_pairs_mesh_plane is not None else 0,
+                    self.shape_pairs_mesh_mesh_count,
+                    self.shape_pairs_mesh_mesh.shape[0] if self.shape_pairs_mesh_mesh is not None else 0,
+                    self.shape_pairs_sdf_sdf_count,
+                    self.shape_pairs_sdf_sdf.shape[0] if self.shape_pairs_sdf_sdf is not None else 0,
+                    writer_data.contact_count,
+                    writer_data.contact_max,
+                ],
+                device=device,
+                record_tape=False,
```
adenzler-nvidia, line 2139: Two questions on this diagnostic kernel:

1. Would it be worth publicly documenting these buffer thresholds (and their meanings) so downstream users could implement their own overflow handlers? If newton gets integrated further downstream, a programmatic hook for buffer-overflow signals would likely be more useful than the `wp.printf` warnings.

2. Can you quantify the overhead of this check kernel? The docstring names the launch cost as the reason to disable in hot loops — a rough number would help users decide when to flip it off.

nvtw: The diagnostic kernel should now be on by default and the CI still shows good perf improvements. So I guess we can keep it on

newton/_src/sim/collide.py
nvtw: Should be improved now

adenzler-nvidia (approved):
Thanks for working through the two rounds of review — all 10 comments are addressed. Verified on 4945b7989:

- `verify_buffers=True` default flipped and propagates to `NarrowPhase`.
- `skip_count_zero` plumbed through SAP/AllPairs/Explicit with an explicit precondition docstring.
- `_counter_array` → `contact_counters` (public) with the "safe to zero every pass" invariant documented.
- `CONVEX_MESH` path has a CHANGELOG entry; builder populates `shape_collision_aabb_lower/upper` for it via the vertex-cloud AABB (matches the convex hull).
- INT32-wrap check is consistent (`==`) across both sites.
- `clear()` docstring reflects the single fused launch; stale "86–90%" claim removed.
- Diagnostic-kernel docstring lists counters and gives a rough overhead estimate.

A few non-blocking follow-ups worth noting for later (see my review thread): `shape_count==0` silently skips the fused reset; the CONVEX_MESH AABB correctness rests on an implicit "hull AABB == vertex-cloud AABB" invariant worth a one-line comment; the INT32-wrap block is now duplicated in two places (candidate for a `@wp.func`).

PR #2485, opened 2026-04-17
fix(security): 2 improvements across 1 files

tomaioo:
## Summary

fix(security): 2 improvements across 1 files

## Problem

**Severity**: `Medium` | **File**: `newton/_src/sensors/warp_raytrace/tiling.py:L6`

`tid_to_coord_tiled()` performs integer divisions by `tile_width` and `tile_height` without validating they are > 0. If configuration provides zero values, the kernel can fail at runtime (DoS/crash condition). Since render config values appear user-configurable, malformed input can reliably break rendering.

## Solution

Validate `tile_width > 0` and `tile_height > 0` before launching kernels (preferably when building/accepting `RenderConfig`), and fail fast with a clear error.

## Changes

- `newton/_src/sensors/warp_raytrace/tiling.py` (modified)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a robustness issue in ray tracing tiling where invalid tile dimensions could cause computation errors. Tile dimensions are now clamped to valid values to ensure correct behavior.

### DIFF
--- a/newton/_src/sensors/warp_raytrace/tiling.py
+++ b/newton/_src/sensors/warp_raytrace/tiling.py
@@ -13,6 +13,9 @@ def tid_to_coord_tiled(
     tile_width: wp.int32,
     tile_height: wp.int32,
 ):
+    tile_width = wp.max(tile_width, wp.int32(1))
+    tile_height = wp.max(tile_height, wp.int32(1))
+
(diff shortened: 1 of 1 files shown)


eric-heiden:
This is not a good solution, closing this.

PR #2487, opened 2026-04-18
Fix `State.assign` for namespaced extended and custom attributes

jsw7460:
## Description
`State.assign` skips `wp.array` attributes nested under attribute namespaces (e.g. `state.mujoco.qfrc_actuator`). The top-level loop only checks `isinstance(val, wp.array)` on `__dict__` entries, so namespace containers are skipped and never descended into. After `state_0.assign(state_1)`, `state_0.mujoco.qfrc_actuator` is still zero.

Fix: after the top-level loop, walk `EXTENDED_ATTRIBUTES` for `"namespace:attr"` entries and apply the same presence/mismatch checks.

## Checklist

## Test plan

uv run --extra dev -m newton.tests -k test_state_assign

## Bug fix

**Steps to reproduce:**

1. Request `mujoco:qfrc_actuator` on a model and step `SolverMuJoCo` so `state_1.mujoco.qfrc_actuator` is populated.
2. Call `state_0.assign(state_1)` (the pattern `State.assign`'s docstring recommends for odd-substep CUDA-graph loops).
3. Read `state_0.mujoco.qfrc_actuator`: values are still the zero-initialized array.

**Minimal reproduction:**

See test_state_assign.py

Note: I followed the current `"namespace:attr"` convention in `EXTENDED_ATTRIBUTES`.
If deeper nesting is needed in the future, let me know and I'll switch to a recursive approach.

## Summary by CodeRabbit

* **Bug Fixes**
  * State assignment now correctly copies namespaced extended and custom state attributes between states and raises on mismatches.

* **Tests**
  * Added tests that confirm namespaced attributes are copied correctly and that assignment fails when required namespace data is missing.

* **Refactor**
  * Internal state-copy logic reorganized to more reliably handle top-level and namespaced array transfers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@
 - Show prismatic joints in the GL viewer when "Show Joints" is enabled
 - Fix connect constraint anchor computation to account for joint reference positions when `SolverMuJoCo` is the chosen solver.
 - Fix `SolverMuJoCo` passing non-zero geom/pair margins to `mujoco_warp.put_model()`, which fails when NATIVECCD is enabled. Margins are forced to zero when MuJoCo handles collisions (`use_mujoco_contacts=True`); the Newton collision pipeline (`use_mujoco_contacts=False`) is unchanged
+- Fix `State.assign` not copying namespaced extended attributes such as `state.mujoco.qfrc_actuator`
 - Fix mesh-convex back-face contacts generating inverted normals that trap shapes inside meshes and cause solver divergence (NaN)
 - Fix finite plane geometry 2x too large in collision, bounding sphere, and raytrace sensor
 - Fix MPR convergence failure on large and extreme-aspect-ratio mesh triangles by projecting the starting point onto the triangle nearest the convex center
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - Fix viewer crash with `imgui_bundle>=1.92.6` when editing colors by normalizing `color_edit3` input/output in `_edit_color3`
 - Show prismatic joints in the GL viewer when "Show Joints" is enabled
 - Fix connect constraint anchor computation to account for joint reference positions when `SolverMuJoCo` is the chosen solver.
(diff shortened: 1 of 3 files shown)


newton/_src/sim/state.py
```diff
@@ -180,6 +180,34 @@ def assign(self, other: State) -> None:
 
             val_self.assign(val_other)
 
+        for ext in self.EXTENDED_ATTRIBUTES:
```
camevor: Ideally, we'd extend this to include custom attributes that are namespaced.

Maybe something like this:

```python
def copy_arrays(dst, src, prefix):  # prefix for error message
    # existing container assign, copies warp arrays

copy_arrays(self, other, "")
ns_names = { ... }  # Model.AttributeNamespace attributes
for ns in ns_names:
    # validate present on both
    copy_arrays(a, b, ns)
```

camevor (changes requested):
Thanks for spotting and fixing this, @jsw7460! Since the namespace issue also affects custom attributes (to an even larger extent), it would make sense to handle those in the same way. Discovering `AttributeNamespace` entries on the state object would cover both cases in one pass.

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@
 - Show prismatic joints in the GL viewer when "Show Joints" is enabled
 - Fix connect constraint anchor computation to account for joint reference positions when `SolverMuJoCo` is the chosen solver.
 - Fix `SolverMuJoCo` passing non-zero geom/pair margins to `mujoco_warp.put_model()`, which fails when NATIVECCD is enabled. Margins are forced to zero when MuJoCo handles collisions (`use_mujoco_contacts=True`); the Newton collision pipeline (`use_mujoco_contacts=False`) is unchanged
+- Fix `State.assign` not copying namespaced extended attributes such as `state.mujoco.qfrc_actuator`
```
camevor: This changelog entry should be updated to reflect the broadened scope
```suggestion
- Fix `State.assign` not copying namespaced extended and custom state attributes 
```

camevor (commented):
Thanks for updating, this looks good to me!

camevor (approved):

eric-heiden (approved):

PR #2488, opened 2026-04-19
Cleanup raycast tests

StafaH:
## Description

Cleanup of raycast tests. Functionally similar but substantially more concise, and less kernels (avoiding excessive kernel compilation). `ray_intersect_geom`is a dispatch function for the remaining ray functions and can be used across all tests.

## Checklist

## Test plan

 ```
 ``` 

## Summary by CodeRabbit

* **Tests**
  * Consolidated ray-intersection tests into a single, unified test harness for improved maintainability and consistency.
  * Replaced repetitive per-primitive test code with table-driven cases, including a new rotated-box scenario.
  * Simplified mesh validation by streamlining mesh-specific test paths.
  * Reduced redundant test imports and cleaned up duplicated coverage, lowering overall test complexity.

### DIFF
--- a/newton/tests/test_raycast.py
+++ b/newton/tests/test_raycast.py
@@ -1,7 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 The Newton Developers
 # SPDX-License-Identifier: Apache-2.0
 
-import math
 import unittest
 
 import numpy as np
@@ -61,14 +60,14 @@ def test_ray_intersect_box(test: TestRaycast, device: str):
     direction = wp.vec3(1.0, 0.0, 0.0)
 
     identity = wp.transform_identity()
-    rot_45_z = wp.quat_from_axis_angle(wp.vec3(0.0, 0.0, 1.0), math.pi / 4.0)
+    rot_45_z = wp.quat_from_axis_angle(wp.vec3(0.0, 0.0, 1.0), wp.pi / 4.0)
 
     # (name, xform, origin, expected)
     cases = [
         ("hit", identity, wp.vec3(-2.0, 0.0, 0.0), 1.0),
         ("miss", identity, wp.vec3(-2.0, 2.0, 0.0), -1.0),
         ("inside", identity, wp.vec3(0.0, 0.0, 0.0), 1.0),
-        ("rotated", wp.transform(wp.vec3(0.0, 0.0, 0.0), rot_45_z), wp.vec3(-2.0, 0.0, 0.0), 2.0 - math.sqrt(2.0)),
+        ("rotated", wp.transform(wp.vec3(0.0, 0.0, 0.0), rot_45_z), wp.vec3(-2.0, 0.0, 0.0), 2.0 - wp.sqrt(2.0)),
     ]
 
     for name, xform, origin, expected in cases:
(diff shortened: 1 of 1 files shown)


newton/tests/test_raycast.py
```diff
@@ -136,543 +31,334 @@ def kernel_test_geom(
...
+            wp.launch(
+                kernel_test_geom,
+                dim=1,
+                inputs=[out_t, geom_to_world, size, geomtype, origin, direction, 0],
+                device=device,
+            )
+            test.assertAlmostEqual(out_t.numpy()[0], expected, delta=delta)
 
 
 def test_ray_intersect_mesh(test: TestRaycast, device: str):
```
eric-heiden, line 229: 🟡 After this refactor, `test_ray_intersect_mesh` launches `kernel_test_geom` with `GeoType.MESH`, which is the same dispatcher path already covered by `test_mesh_ray_intersect_via_geom` at lines 308-333. That leaves no direct coverage of the `ray_intersect_mesh()` entry point in this file, even though renderer code calls it directly at `newton/_src/sensors/warp_raytrace/raytrace.py:439-446` and `680-687`.

This distinction matters because `ray_intersect_geom()` hardcodes `enable_backface_culling=False` and `_DEFAULT_MESH_MAX_T` when it forwards to the helper (`newton/_src/geometry/raycast.py:739-742`), so a regression in the direct helper plumbing would slip past both tests here. The current test name and docstring also still read as if this test targets `ray_intersect_mesh()` itself.

Would you consider either restoring `kernel_test_mesh` so this test calls the helper directly, or renaming it to reflect that it is another `ray_intersect_geom` case and adding a separate direct `ray_intersect_mesh()` smoke test?

newton/tests/test_raycast.py
```diff
@@ -136,543 +31,334 @@ def kernel_test_geom(
...
+
+    for name, xform, size, origin, direction, expected in cases:
+        with test.subTest(name):
+            wp.launch(
+                kernel_test_geom, dim=1, inputs=[out_t, xform, size, GeoType.PLANE, origin, direction, 0], device=device
+            )
+            test.assertAlmostEqual(out_t.numpy()[0], expected, delta=1e-5)
 
 
 def test_geom_ray_intersect(test: TestRaycast, device: str):
```
eric-heiden: ⚪ After the refactor, every per-type test (`test_ray_intersect_sphere`, `test_ray_intersect_box`, and so on) also launches `kernel_test_geom`, so this function overlaps with them substantially. The `sphere`, `box`, `capsule`, `cylinder`, `cone`, and `ellipsoid` cases here use origin `(-2, 0, 0)` and direction `(1, 0, 0)`, which match the `hit` case in each per-type test modulo delta. The only genuinely unique case is `plane` with origin `(0, 0, 5)` and direction `(0, 0, -1)` expecting `5.0`, since `test_ray_intersect_plane` only uses 3-4-5 triples and never a unit-length axial ray.

Since this PR is a cleanup, consolidation would match that intent. A few options:

- remove `test_geom_ray_intersect` entirely,
- trim it to just the plane case,
- or fold the plane case into `test_ray_intersect_plane` and delete this function.

If you keep it as-is, a short comment noting which coverage is unique would help future readers.

eric-heiden (approved):
Thanks!

PR #2527, opened 2026-04-22
Avoid gpu only features in the collision pipeline

nvtw:
## Description

Replaces explicit shared memory usage with new warp tile operations to make sdf collisions work on the CPU.

## Checklist

## Test plan

     or manual steps. Example:
     ```
     ``` -->




## Summary by CodeRabbit

* **New Features**
  * Mesh-mesh contact detection now available on CPU.
  * Mesh-heightfield SDF contact generation supported on CPU.
  * Contact reduction optimization runs on all platforms.

* **Improvements**
  * Simplified collision query pipeline for improved efficiency.
  * Unified contact kernel implementations across CPU and CUDA platforms.


newton/_src/geometry/multicontact.py
```diff
@@ -843,7 +837,7 @@ def build_manifold(
 
         # Allocate buffers: 5 for A, up to 10 for B (5 + clipping headroom)
         b_buffer = wp.zeros(shape=(10,), dtype=wp.vec2f)
-        a_buffer = wp.array(ptr=get_ptr(b_buffer) + wp.uint64(5 * 8), shape=(5,), dtype=wp.vec2f)
+        a_buffer = wp.array(ptr=b_buffer.ptr + wp.uint64(5 * 8), shape=(5,), dtype=wp.vec2f)
```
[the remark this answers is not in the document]
nvtw: No, this code is perfectly right, highly optimized and well tested. it works.

newton/_src/geometry/sdf_contact.py
```diff
@@ -1268,14 +1174,11 @@ def mesh_sdf_collision_kernel(
...
-                            contact_data.sort_sub_key = (edge_idx << 2) | (mode << 1)
+                            contact_data.sort_sub_key = (my_edge_idx << 2) | (mode << 1)
 
                             writer_func(contact_data, writer_data, -1)
 
-                    synchronize()
-                    if t == 0:
-                        selected_edges[edge_capacity] = 0
-                    synchronize()
+                    wp.tile_stack_clear(edge_stack)
```
adenzler-nvidia, line 1225: **[Critical — potential data loss on GPU]** I believe this loop can silently drop accepted edges on GPU. Trace:

- Inner loop exits once `wp.tile_stack_count(edge_stack) >= capacity` (= `MESH_SDF_BLOCK_DIM` = 256). Pre-push count ≤ 255, cooperative push adds up to 256 items, so post-push count ∈ [256, 511] — fits in `STACK_CAPACITY = 512`, good.
- `wp.tile_stack_pop` drains `min(count, block_dim)` items (one per thread; see `warp/_src/builtins.py:7146-7186`).
- Post-pop count ∈ [0, 255].
- This `wp.tile_stack_clear` then resets count to 0 (`warp/_src/builtins.py:7215`, confirmed by `warp/tests/tile/test_tile_stack.py:147-155`), discarding the residual.

Meanwhile `progress` has already advanced past those edges (`old_progress + capacity` each inner iter), so they're never re-evaluated. Net effect: up to `block_dim - 1` accepted edges lost per outer iteration on GPU.

The previous `add_to_shared_buffer_atomic` handled this explicitly via `wp.atomic_min(buffer, capacity + 1, edge_idx)`, which rolled the progress counter back for any overflowed push so it got retried on the next pass. The new code has no equivalent.

On CPU (`block_dim = 1`), at most 1 item is ever pushed between pop+clear, so the bug is latent there. All current tests use `Mesh.create_box` (12 edges), which stays well below capacity in a single cooperative pass, so they don't exercise the overflow path.

Fix options:
1. Drain the stack before clearing: `while wp.tile_stack_count(edge_stack) > 0: popped, slot = wp.tile_stack_pop(edge_stack); if slot >= 0: ...process...` — then the `clear` is a no-op.
2. Cap push acceptance so count never exceeds `block_dim` (e.g., a scan-based selection), eliminating both the overflow margin and the clear.
3. Re-introduce a progress-rollback equivalent to the old `atomic_min`.

newton/_src/geometry/sdf_contact.py
```diff
@@ -1573,9 +1512,6 @@ def mesh_sdf_collision_global_reduce_kernel(
                                 reducer_data,
                             )
 
-                    synchronize()
-                    if t == 0:
-                        selected_edges[edge_capacity] = 0
-                    synchronize()
+                    wp.tile_stack_clear(edge_stack)
```
adenzler-nvidia, line 1576: **[Critical — same issue as in `mesh_sdf_collision_kernel`]** Same `push → pop → clear` pattern in the global-reduce kernel; same concern about silently discarding up to `block_dim - 1` accepted edges per outer iteration on GPU. Fix here will presumably mirror whatever you choose for the first kernel — another reason to extract a shared helper (see comment on the duplicated block above).

adenzler-nvidia (commented):
Thanks for the rewrite — dropping the four `func_native` / `__shared__` helpers is a welcome simplification, and `MESH_SDF_BLOCK_DIM` + `EdgeCullResult` are both well-motivated.

One concern I'd like you to look at before merge is correctness of the new `push → pop → clear` loop on GPU (inline below at both `wp.tile_stack_clear` sites). I believe it can silently drop up to `block_dim - 1` accepted edges per outer iteration, because `tile_stack_pop` only drains `block_dim` items per call and `tile_stack_clear` unconditionally resets count to zero (verified against `warp/_src/builtins.py:7146-7186, 7215`). This wouldn't show on CPU (`block_dim=1`) or in the current tests (all meshes have <<256 edges), but it would manifest on dense meshes on GPU. Happy to be wrong — if I am, a short comment documenting the invariant would help the next reader.

The other items are smaller: a couple of stale comments, a hardcoded `target_blocks=64` on CPU, and test coverage gaps around `reduce_contacts` on CPU and large-mesh overflow. All inline below — no blocking opinions from me, just notes.

newton/_src/geometry/sdf_contact.py
```diff
@@ -1125,58 +998,91 @@ def mesh_sdf_collision_kernel(
...
+
+                if wp.static(enable_heightfields):
+                    sdf_is_heightfield = sdf_type == GeoType.HFIELD
+                else:
+                    sdf_is_heightfield = False
+                sdf_aabb_lower = texture_sdf.sdf_box_lower
+                sdf_aabb_upper = texture_sdf.sdf_box_upper
+
+                while wp.tile_extract(progress, 0) < num_edges:
+                    # -- begin: find interesting edges (inlined) --
```
adenzler-nvidia: **[Important — code duplication; also applies at line 1348]** This ~60-line inlined edge-culling block is byte-for-byte identical to the version at line 1348 except for the loop bound (`num_edges` vs `edge_end`). The PR explicitly dropped `find_interesting_edges_func`, which suggests the inlining was intentional — but if tile ops can live inside a `@wp.func`, extracting a shared helper would remove the duplication and give you one place to fix the C1 issue instead of two. If inlining is a hard constraint of Warp kernel codegen, at minimum please add a cross-reference comment at each copy so they don't drift.

**[Suggestion]** The `# -- begin: find interesting edges (inlined) --` / `# -- end: ... --` sentinels are refactoring artifacts — they describe what the code *was* rather than what it *is* or what invariant it maintains. Consider replacing with a single leading comment describing the loop's contract, e.g. *"Cooperative edge-culling pass: each thread evaluates one candidate edge and pushes it onto `edge_stack` if it may produce contact; loop exits once the stack has at least `block_dim` items or all edges scanned."*

newton/_src/geometry/narrow_phase.py
```diff
@@ -1521,7 +1511,9 @@ class (``gjk_candidate_pairs_count``, ``shape_pairs_mesh_count``,
             writer_func = contact_writer_warp_func
 
         self.tile_size_mesh_convex = 128
-        self.tile_size_mesh_mesh = 256
+        # Must match the tile stack capacity in sdf_contact.py, re-use
+        # the single source of truth so the two can't drift.
+        self.tile_size_mesh_mesh = MESH_SDF_BLOCK_DIM
         self.tile_size_mesh_plane = 512
         self.block_dim = 128
```
adenzler-nvidia, line 1527: **[Important]** `self.block_dim = 128` here coexists with `self.tile_size_mesh_mesh = MESH_SDF_BLOCK_DIM` (= 256) on the line above. The mesh-SDF kernels' tile-stack invariant assumes `wp.block_dim() == MESH_SDF_BLOCK_DIM` (the push overflow margin is sized for it). If anything ever feeds `self.block_dim` to a mesh-SDF launch by accident, the invariant breaks silently.

Suggest either:
- Make the coupling explicit via `assert self.tile_size_mesh_mesh == MESH_SDF_BLOCK_DIM`, or
- If Warp supports it, a `wp.static_assert(wp.block_dim() == MESH_SDF_BLOCK_DIM)` at the top of each kernel.

Also — `self.block_dim = 128` doesn't seem to be used for the mesh-SDF launches (the launches use `self.tile_size_mesh_mesh`), but a quick grep would confirm.

newton/_src/geometry/narrow_phase.py
```diff
@@ -1521,7 +1511,9 @@ class (``gjk_candidate_pairs_count``, ``shape_pairs_mesh_count``,
             writer_func = contact_writer_warp_func
 
         self.tile_size_mesh_convex = 128
-        self.tile_size_mesh_mesh = 256
+        # Must match the tile stack capacity in sdf_contact.py, re-use
```
adenzler-nvidia: **[Important]** The comment says *"Must match the tile stack capacity in sdf_contact.py"*, but you're assigning the **block dim** (`MESH_SDF_BLOCK_DIM` = 256), not the stack capacity (`STACK_CAPACITY` = 512). They happen to differ by 2×. Suggest: *"Must match `MESH_SDF_BLOCK_DIM` in sdf_contact.py — the mesh-SDF kernels assume `wp.block_dim()` equals that constant so the tile-stack overflow margin (`STACK_CAPACITY = 2 * MESH_SDF_BLOCK_DIM`) is correctly sized."*

newton/_src/geometry/narrow_phase.py
```diff
@@ -1996,7 +1984,7 @@ def launch_custom_write(
             if shape_edge_range is None:
                 shape_edge_range = self._empty_edge_range
 
-            if wp.get_device(device).is_cuda and self.mesh_mesh_contacts_kernel is not None:
+            if self.mesh_mesh_contacts_kernel is not None:
```
adenzler-nvidia, line 2000: **[Important / Suggestion]** Two small items about this block:

1. The comment ten lines above at line 1977 *"Launch mesh-mesh contact processing kernel on CUDA."* is now stale — the whole point of this PR is that this kernel runs on CPU too. Drop the "on CUDA" qualifier.
2. With the CUDA gating removed, `self.mesh_mesh_contacts_kernel is not None` is effectively a tautology (it's non-None whenever `self.has_meshes` is true). Consider tightening to `if self.has_meshes:` for clarity, or dropping the guard entirely if guaranteed by construction.

newton/_src/geometry/sdf_contact.py
```diff
@@ -1125,58 +998,91 @@ def mesh_sdf_collision_kernel(
...
-                    )
-
-                    has_edge = t < selected_edges[edge_capacity]
-                    synchronize()
+                wp.tile_scatter_masked(progress, 0, 0, t == 0)
+
+                if wp.static(enable_heightfields):
+                    sdf_is_heightfield = sdf_type == GeoType.HFIELD
+                else:
+                    sdf_is_heightfield = False
```
adenzler-nvidia: **[Suggestion — same at line 1343]** `sdf_is_heightfield` here duplicates `sdf_is_hfield` computed earlier in the same kernel (set at the outer scope and passed down to `do_edge_sdf_collision`). You're executing the same `wp.static` condition twice. Re-using the existing variable would remove one duplicate computation and one potential source of drift.

newton/_src/geometry/sdf_contact.py
```diff
@@ -9,17 +9,35 @@
...
+# ``mesh_sdf_collision_kernel`` and ``mesh_sdf_collision_global_reduce_kernel``.
+MESH_SDF_BLOCK_DIM = 256
+
+# Capacity of the cooperative edge-selection tile stack. Sized to
+# ``2 * MESH_SDF_BLOCK_DIM`` so that after an iteration leaves the stack
+# with up to ``MESH_SDF_BLOCK_DIM`` accepted edges, the next cooperative
+# push from ``MESH_SDF_BLOCK_DIM`` threads can always fit without overflow
+# — the stack can never overrun regardless of how many edges pass the
+# culling test.
+STACK_CAPACITY = 2 * MESH_SDF_BLOCK_DIM
```
adenzler-nvidia, line 32: **[Suggestion]** Small off-by-one in the prose: the comment says *"after an iteration leaves the stack with up to `MESH_SDF_BLOCK_DIM` accepted edges"* — but the inner loop gate `count < capacity` caps pre-push count at `capacity - 1`, and pop drains `block_dim`, so post-pop residual is strictly *less than* `MESH_SDF_BLOCK_DIM`. A tighter phrasing: *"the next cooperative push from `MESH_SDF_BLOCK_DIM` threads fits without overflow because pre-push count is at most `MESH_SDF_BLOCK_DIM - 1` and push adds at most `MESH_SDF_BLOCK_DIM` more."*

(Separately — see the C1 comment: the *no-overflow* invariant this comment describes is correct. But the consumer-side invariant — "every pushed edge is eventually processed" — does not currently hold; worth calling out explicitly once addressed.)

newton/_src/geometry/sdf_contact.py
```diff
@@ -1037,6 +907,9 @@ def mesh_sdf_collision_kernel(
 
         pair_count = wp.min(shape_pairs_mesh_mesh_count[0], shape_pairs_mesh_mesh.shape[0])
 
+        edge_stack = wp.tile_stack(capacity=STACK_CAPACITY, dtype=EdgeCullResult)
+        progress = wp.tile_zeros(shape=1, dtype=int, storage="shared")
```
adenzler-nvidia, line 922: **[Suggestion]** The old `find_interesting_edges_func` docstring spelled out the progress-counter semantics (*"`[block_dim+1]`: Progress counter (edges processed so far)"*). With the migration to `progress = wp.tile_zeros(shape=1, ...)`, a reader has to reverse-engineer from `base_edge_idx + t` / `old_progress + capacity` that `progress[0]` is the next edge index the next cooperative pass should start from (a high-water mark, not a count). One short line would help, e.g. *"`progress[0]` = next edge index to evaluate; each thread `t` processes edge `progress[0] + t` in a cooperative pass."*

Applies equally at line 1230.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 - Add `Mesh.is_watertight` property (cached) that reports whether every geometric edge is shared by exactly two triangles
 - Add `deterministic` flag to `CollisionPipeline` and `NarrowPhase` for GPU-thread-scheduling-independent contact ordering via radix sort and deterministic fingerprint tiebreaking in contact reduction
 - Add fast parity-based SDF construction path for watertight meshes in `SDF.create_from_mesh`, using `wp.mesh_query_point_sign_parity` instead of winding numbers; selected via the new `sign_method` argument (`"auto"` — the default — picks parity when `Mesh.is_watertight` is true, or `"parity"` / `"winding"` to force either strategy)
+- Enable full CPU execution of the collision pipeline, including mesh–mesh and mesh–heightfield SDF contacts that were previously CUDA-only; contact reduction (`reduce_contacts`) also runs on CPU
```
adenzler-nvidia: **[Suggestion]** The `warnings.warn("NarrowPhase running on CPU: mesh-mesh contacts will be skipped ...")` call in `NarrowPhase.__init__` is gone. That's the correct outcome (the warning is no longer accurate), but it's a user-observable change for anyone who used it as a CI signal to catch accidental CPU fallback. Consider either:
- Folding it into this entry: *"...; the previous CPU-mesh-skipped warning is no longer emitted."*, or
- Adding a one-liner in the `Changed` section.

newton/tests/test_heightfield.py
```diff
@@ -408,7 +408,6 @@ def _build_mesh_vs_heightfield(self, mesh: newton.Mesh, mesh_z: float = 0.15):
         builder.add_shape_mesh(body=mesh_body, mesh=mesh)
         return builder.finalize(), mesh_body
 
-    @unittest.skipUnless(_cuda_available, "mesh-heightfield collision requires CUDA")
     def test_non_convex_mesh_vs_heightfield(self):
```
adenzler-nvidia: **[Suggestion]** These tests now correctly run without the `skipUnless(_cuda_available, ...)` decorator, but they still execute on the default device only (not parametrized via `add_function_test(devices=get_test_devices(...))`). On CUDA-enabled machines they'll still only exercise the CUDA path, so the CPU coverage is only picked up on CPU-only CI runs. Parametrizing would give you both devices explicitly, same approach as `test_collision_pipeline.py`.

newton/tests/test_collision_pipeline.py
```diff
@@ -424,15 +424,35 @@ def test_mesh_mesh_bvh_vs_bvh(_test, device, broad_phase: str):
...
+]
+mesh_mesh_sdf_tests_all = [
     ("bvh_vs_bvh", test_mesh_mesh_bvh_vs_bvh),
 ]
 
-for mode_name, test_func in mesh_mesh_sdf_tests:
+for mode_name, test_func in mesh_mesh_sdf_tests_cuda:
+    for broad_phase_name, broad_phase in [
+        ("explicit", "explicit"),
+        ("nxn", "nxn"),
```
adenzler-nvidia, line 443: **[Important / Suggestion — test coverage gaps]** A few gaps worth closing:

1. **No CPU coverage for `reduce_contacts=True`.** The only caller is `test_deterministic_pipeline_500_steps` and it's still gated to `get_cuda_test_devices()`. The PR enables `reduce_contacts` on CPU in `narrow_phase.py:1490-1494` (and adds the hardcoded `target_blocks = 64` CPU branch at `narrow_phase.py:1673`), but nothing actually exercises the CPU path of `mesh_sdf_collision_global_reduce_kernel`. That's the same kernel where C1 lives, so any CPU-only regression in the reduce path would slip through. Suggest parametrizing that test or adding a lightweight CPU smoke test that asserts non-zero contacts with `reduce_contacts=True`.

2. **No CPU↔CUDA parity test.** The PR's central claim is "same code path, different device." A small test that builds one scene on both, copies contacts to host, sorts, and asserts contact counts / normals match within tolerance would guard against both CPU-only and GPU-only regressions (including C1). Current coverage relies on loose XPBD velocity-drift tolerances in `collision_pipeline_contact_tests` which would likely hide partial contact loss.

3. **`STACK_CAPACITY = 2 * MESH_SDF_BLOCK_DIM` invariant untested.** All test meshes are `Mesh.create_box` (12 edges). A scene with an icosphere or a non-trivial OBJ with hundreds of edges intersecting a heightfield would exercise the stack-overflow margin and — once C1 is fixed — verify the residual-drain path.

nvtw:
@adenzler-nvidia your comments should be addressed

adenzler-nvidia:
Thanks for the fixes — the drain-until-empty pattern on both kernels, the new overflow-margin and CPU↔CUDA parity tests, and the `narrow_phase` assertion + comment rewording all look good to me.

Two small nits still open that would be nice to clean up before merge:

1. **`narrow_phase.py:1990` — stale "on CUDA" comment.** Still reads `# Launch mesh-mesh contact processing kernel on CUDA.`; since the whole point of this PR is that the kernel runs on CPU too, the "on CUDA" qualifier is now misleading. (Separately — I was wrong about `self.mesh_mesh_contacts_kernel is not None` being a tautology: the enclosing `if self.has_meshes or self.has_heightfields:` means `is not None` actually is the correct guard. Ignore that half of my original comment.)

2. **`sdf_contact.py:1018` and `:1393` — `sdf_is_heightfield` duplicates `sdf_is_hfield`.** Both kernels still compute `sdf_is_heightfield = sdf_type == GeoType.HFIELD` when `sdf_is_hfield` is already in scope (and `sdf_type` is derived from it). One-liner each: `sdf_is_heightfield = sdf_is_hfield`.

Happy to approve once those are in.

nvtw:
> Thanks for the fixes — the drain-until-empty pattern on both kernels, the new overflow-margin and CPU↔CUDA parity tests, and the `narrow_phase` assertion + comment rewording all look good to me.
> 
> Two small nits still open that would be nice to clean up before merge:
> 
> 1. **`narrow_phase.py:1990` — stale "on CUDA" comment.** Still reads `# Launch mesh-mesh contact processing kernel on CUDA.`; since the whole point of this PR is that the kernel runs on CPU too, the "on CUDA" qualifier is now misleading. (Separately — I was wrong about `self.mesh_mesh_contacts_kernel is not None` being a tautology: the enclosing `if self.has_meshes or self.has_heightfields:` means `is not None` actually is the correct guard. Ignore that half of my original comment.)
> 2. **`sdf_contact.py:1018` and `:1393` — `sdf_is_heightfield` duplicates `sdf_is_hfield`.** Both kernels still compute `sdf_is_heightfield = sdf_type == GeoType.HFIELD` when `sdf_is_hfield` is already in scope (and `sdf_type` is derived from it). One-liner each: `sdf_is_heightfield = sdf_is_hfield`.
> 
> Happy to approve once those are in.

Done

adenzler-nvidia (approved):
Both nits from the previous round are addressed — the stale "on CUDA" comment in `narrow_phase.py` is updated and `sdf_is_heightfield` now reuses `sdf_is_hfield` in both kernels. Thanks!

PR #2533, opened 2026-04-22
Parse NewtonSDFCollisionAPI for SDF and hydroelastic collision configuration

### LINKED ISSUE #1343: Add USD Parsing Support for Hydro/SDF Contact Models
When https://github.com/newton-physics/newton-usd-schemas/issues/24 is done, add parsing support in the model builder.

mzamoramora-nvidia:
## Description

Adds USD parsing for `NewtonSDFCollisionAPI` (companion schemas PR https://github.com/newton-physics/newton-usd-schemas/pull/46) — the first USD authoring path for SDF-based and hydroelastic contact in Newton. Applying `NewtonSDFCollisionAPI` to a `Gprim` opts the shape in to SDF generation; hydroelastic contact is opt-in on top via `newton:hydroelasticEnabled`.

## What's added

**USD attributes parsed on `NewtonSDFCollisionAPI`** (via `ModelBuilder.add_usd()`):

| Attribute | Default | Effect |
|---|---|---|
| `newton:sdfMaxResolution` | `64` | Max grid dimension (multiple of 8) |
| `newton:sdfTargetVoxelSize` | `-inf` (sentinel) | When set, overrides `sdfMaxResolution`; otherwise falls back to it |
| `newton:sdfNarrowBandInner` | `-0.1` m | Inner narrow-band depth |
| `newton:sdfNarrowBandOuter` | `0.1` m | Outer narrow-band depth |
| `newton:sdfPadding` | `-inf` (sentinel) | SDF AABB padding; falls back to `newton:contactGap` |
| `newton:sdfTextureFormat` | `"uint16"` | Subgrid storage format (`uint8`/`uint16`/`float32`) |
| `newton:hydroelasticEnabled` | `false` | Hydroelastic opt-in (both contact partners must set true) |
| `newton:hydroelasticStiffness` | `1e10` N/m³ | Contact stiffness coefficient (force = k_eff · area · depth) |

`newton:hydroelasticStiffness` authored alone is a material parameter — it does **not** flip hydroelastic on without `newton:hydroelasticEnabled = true`.

**Python API additions**

- `ModelBuilder.ShapeConfig.sdf_padding: float | None` — per-shape SDF AABB padding in meters. When `None`, finalize falls back to `shape_gap`. Rejected on mesh shapes (where `mesh.sdf` is authoritative).
- `ModelBuilder.shape_sdf_padding` — accumulator list propagated to `Model.shape_sdf_padding`.

**Validation**

- Early `ValueError` at parse time when a mesh is `hydroelasticEnabled = true` but has no `sdfMaxResolution` / `sdfTargetVoxelSize` authored — hydroelastic meshes require an SDF source.

**Docs**

- New *Porting PhysX SDF Assets to Newton* section in `docs/concepts/usd_parsing.rst`, including a runnable `port_physx_sdf_to_newton()` helper exercised as a Sphinx doctest.

Depends on https://github.com/newton-physics/newton-usd-schemas/pull/46.

Closes #1343.

## Open questions

- `metersPerUnit` handling: distance attributes are read raw; confirm whether importer should scale.
- Default `newton:hydroelasticStiffness = 1e10` mirrors Newton's `default_shape_cfg.kh`. Worth documenting the choice in user-facing docs.

## Checklist

- [x] Tests cover these changes
- [x] Docs are up to date
- [x] `CHANGELOG.md` updated

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_sdf_usd
uv run --extra dev -m newton.tests -k test_hydroelastic
uv run --extra dev -m newton.tests -k test_menagerie
cd docs && uv run --extra docs make doctest
cd docs && uv run --extra docs make html
```


newton/_src/utils/import_usd.py
```diff
@@ -2637,11 +2812,35 @@ def _build_mass_info_from_shape_geometry(
...
                     )
+                    # Store SDF intent on the builder (deferred to finalize).
+                    builder.shape_sdf_max_resolution[shape_id] = sdf_max_resolution
+                    builder.shape_sdf_target_voxel_size[shape_id] = sdf_target_voxel_size
+                    builder.shape_sdf_narrow_band_range[shape_id] = sdf_narrow_band_range
+                    builder.shape_sdf_texture_format[shape_id] = sdf_texture_format
+                    builder.shape_sdf_margin[shape_id] = sdf_margin
+                    if is_hydroelastic:
+                        builder.shape_flags[shape_id] |= ShapeFlags.HYDROELASTIC
+                        builder.shape_material_kh[shape_id] = kh
```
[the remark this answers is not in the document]
mzamoramora-nvidia: Thanks for flagging — applied the change in commit ecf9e33f. The `kh` assignment now runs for every mesh, not just hydroelastic ones.

Small clarification on the underlying claim: `newton:kh` wasn't actually being dropped for meshes. The `replace()` call above strips only the SDF-specific cfg fields and leaves `kh` intact, so `add_shape_mesh()` → `add_shape()` still appends `cfg.kh` to `shape_material_kh` (`builder.py:5322`). The explicit write inside `if is_hydroelastic:` was just overwriting with the same value.

Moving the line out of the conditional is still the right call: it matches how the other SDF fields are written back after the cfg is stripped, and it keeps working if someone later changes the `replace()` call to also strip `kh`. The inline comment in the commit captures that intent.

mzamoramora-nvidia:
@coderabbitai review

nvtw:
@mzamoramora-nvidia @andrewkaufman let's say we merge this MR. Does this mean the SDF schema is then already set in stone? I would like to have the new SDF schema code on Newton main but would prefer to not advertise it yet (we can call it experimental). Having the code on main would make it easy for many people to experiment with it while we do not yet guarantee schema stability and make it clear that assets could break (we should probably clearly mark that in all relevant places). I would prefer to have a larger scale testing phase before the SDF schema is set into stone.

andrewkaufman:
> I would prefer to have a larger scale testing phase before the SDF schema is set into stone.

Can you elaborate? FWIW the only schemas currently explicitly marked experimental is the Actuator stuff. The rest is just released as is, but the entire schema module is v0.x still. So its all _implicitly_ experimental to those paying attention to semver.

I don't mind marking SDF as experimental as well, if its really unstable / shape-shifting on the runtime side too. But is that the case? I was under the impression that ShapeConfig is considered 1.x stable API, so the SDF part of it would be as stable.

nvtw:
> > I would prefer to have a larger scale testing phase before the SDF schema is set into stone.
> 
> Can you elaborate? FWIW the only schemas currently explicitly marked experimental is the Actuator stuff. The rest is just released as is, but the entire schema module is v0.x still. So its all _implicitly_ experimental to those paying attention to semver.
> 
> I don't mind marking SDF as experimental as well, if its really unstable / shape-shifting on the runtime side too. But is that the case? I was under the impression that ShapeConfig is considered 1.x stable API, so the SDF part of it would be as stable.

My only concern is that once something is in the schema, it's there forever (that's what happened in PhysX and it was quite a mess). Therefore I thought merging now and then having another MR before Newton 1.3 is released, in case testing shows something is missing or not handy in the schema, would still be possible.

It's only about having the SDF schema marked as experimental for now to not end up in this trap where you need to offer backward compatibility for old usd files that were created with your first schema draft that might turn out to have flaws - often these flaws are only found by using the schema in every day tasks.

andrewkaufman:
@nvtw 

>  I thought merging now and then having another MR before Newton 1.3 is released, in case testing shows something is missing or not handy in the schema, would still be possible.

This is definitely possible, any change between now & when 1.3 is released shouldn't be considered breaking (since it would never have been released). I think thats irrespective of whether its experimental or not.

> It's only about having the SDF schema marked as experimental for now to not end up in this trap where you need to offer backward compatibility for old usd files that were created with your first schema draft that might turn out to have flaws

We discussed this in the schema WG & concluded that we need to provide auto-upgrade fixes for older assets regardless (either scripted or agent assisted), & we need to allow even the more stable APIs to fluctuate. The entire schema package is still v0.x so technically all "experimental"

I've also added a notice about this to the schema README.md in https://github.com/newton-physics/newton-usd-schemas/pull/55

newton/_src/utils/import_usd.py
```diff
@@ -261,7 +283,9 @@ def parse_usd(
     if mesh_maxhullvert is None:
         mesh_maxhullvert = Mesh.MAX_HULL_VERTICES
 
-    if schema_resolvers is None:
+    if not schema_resolvers:
```
andrewkaufman: Switching from `is None` to `not schema_resolvers` silently changes the semantics for callers who pass `schema_resolvers=[]` to mean "skip all resolver-based attribute collection." After this, an empty list is indistinguishable from "use defaults."

If a caller intentionally passes `[]` (e.g. a pipeline that only wants physics-core attributes, no Newton extensions), they'd now get `SchemaResolverNewton()` injected without warning.

If the goal is to guard against accidental empty lists, we could consider logging a warning when `[]` is used.

newton/_src/sim/builder.py
```diff
@@ -10147,6 +10209,36 @@ def _create_primitive_mesh(stype: int, scale: Sequence[float] | None) -> NewtonM
...
+                    if mesh_sdf is None and (sdf_max_resolution is not None or sdf_target_voxel_size is not None):
+                        sdf_kwargs = {"narrow_band_range": tuple(sdf_narrow_band_range)}
+                        if sdf_max_resolution is not None:
+                            sdf_kwargs["max_resolution"] = sdf_max_resolution
+                        if sdf_target_voxel_size is not None:
+                            sdf_kwargs["target_voxel_size"] = sdf_target_voxel_size
+                        sdf_kwargs["margin"] = sdf_gen_margin
+                        sdf_kwargs["scale"] = tuple(shape_scale)
+                        sdf_kwargs["texture_format"] = sdf_tex_fmt
+                        deferred_key = (
```
andrewkaufman, line 10397: The cache key correctly includes `id(shape_src)` and `tuple(shape_scale)` to deduplicate identical (Mesh, params) combinations while distinguishing different scales. This is the right design for the shared-Mesh use case described in the comment.

However, I don't see a test that verifies two shapes sharing the same `Mesh` but with different scales produce *different* SDF entries in the finalized model. This is the motivating scenario for the clone cache — worth adding a regression test.

newton/_src/utils/import_usd.py
```diff
@@ -44,6 +44,28 @@
 _NEWTON_SRC_DIR = os.path.normpath(os.path.join(os.path.dirname(__file__), os.pardir)) + os.sep
 
 
+def _prim_has_applied_schema(prim, name: str) -> bool:
+    """Check whether an API schema is applied on a prim.
+
+    Reads the ``apiSchemas`` metadata directly so detection works even when
+    the schema is not registered with the USD runtime (codeless schemas
+    loaded only by plugInfo may still be unregistered in test environments).
```
andrewkaufman: Is this strictly necessary? Once released, we can assume the `NewtonSDFCollisionAPI` (and the other `Newton*` schemas) will always be available in the runtime. If it was already released, would you still have need of this `_prim_has_applied_schema` function?

newton/_src/utils/import_usd.py
mzamoramora-nvidia: Restore `is None` in 140939c

newton/_src/sim/builder.py
mzamoramora-nvidia: Added test in 72c912d

newton/_src/utils/import_usd.py
mzamoramora-nvidia: This is only necessary to make CI tests happy for now. We should be able to remove it once https://github.com/newton-physics/newton-usd-schemas/pull/46 has landed.

andrewkaufman: I just released 0.3.0b1 with the SDF schema & I've updated to that dep in #2951, so if we can land that then you can drop any WIP code

andrewkaufman: Or alternatively you can update to 0.3.0b1 in this PR & we merge this first. Go for it

mzamoramora-nvidia: Done in 3572ecd

docs/concepts/usd_parsing.rst
```diff
@@ -844,15 +844,25 @@ doubles as an end-to-end regression against Newton's importer.
                                               op.orderedItems) if items)
 
    def _bbox_diag(prim) -> float | None:
-       """Compute the local-space AABB diagonal [m] of a USD prim, or
-       ``None`` if the bbox is empty."""
+       """Compute the local-space AABB diagonal [m] of a USD prim, applying
+       ``physics:meshScale`` (and ``physxSDFMeshCollision:meshScale`` as a
```
andrewkaufman: Where does ` physics:meshScale` come from? I'm not familiar with any schema that provides this. Is it a physx one?

andrewkaufman (dismissed):

nvtw (dismissed):

docs/concepts/usd_parsing.rst
mzamoramora-nvidia: Removed in ee62fad. Using `BBoxCache.ComputeWolrdBound` to estimate diagonal.

nvtw (dismissed):

adenzler-nvidia:
**Mesh approximation bypasses the new hydroelastic/SDF guard**

The new guard in `add_shape` (`newton/_src/sim/builder.py:5519-5539`) only covers the direct `add_shape_convex_hull` path. The USD importer reaches the same end state by a different route: a `MeshShape` with `NewtonSDFCollisionAPI` + `hydroelasticEnabled=true` is added via `add_shape_mesh` as `GeoType.MESH`, with the `HYDROELASTIC` flag and deferred SDF params written straight onto the builder lists (`import_usd.py:2880-2890`). If that prim also authors `physics:approximation`, `approximate_meshes()` then rewrites `shape_type` in place (`builder.py:6589`, `6661`, `6672`, `6688`) — and for box/sphere also sets `shape_source = None` — without clearing the SDF params or the `HYDROELASTIC` flag.

In `finalize()` this yields a silently wrong result:

- convexHull/coacd/vhacd → `CONVEX_MESH`: skips the `GeoType.MESH` deferred-SDF branch, falls into `elif is_hydroelastic`, `_create_primitive_mesh` returns `None`, and the `else` appends an empty texture SDF — the collision geometry is silently lost.
- boundingBox/boundingSphere → `BOX`/`SPHERE`: builds a texture SDF of the wrong primitive from the fitted bound, silently substituting box/sphere geometry for the mesh's intended SDF.

Repro: author a `Mesh` with `NewtonSDFCollisionAPI`, `newton:hydroelasticEnabled=true`, `newton:sdfMaxResolution=64`, and `physics:approximation=convexHull`, then `parse_usd(...)` + `finalize()`.

Suggested fix: reject (or warn) at parse time when `is_hydroelastic`/SDF params are authored alongside a mesh approximation, or have `approximate_meshes()` clear `shape_sdf_*` and the `HYDROELASTIC` flag whenever it converts a shape away from `GeoType.MESH`. Worth a regression test for the approximation + hydroelastic combination.

mzamoramora-nvidia:
@adenzler-nvidia great catch! I opted for raising an error when approximation methods that rewrite `shape_type` are used ( `coacd`, `vhacd`, `convex_hull`, `bounding_box`, `bounding_sphere`). Users are now warned that the sdf will be built on the approximated mesh when using mesh approximation methods like `quadratic`, `ftetwild`, etc.

andrewkaufman:
> If that prim also authors physics:approximation

In the design meeting we'd said this can't happen, that the "sibling" `*CollisionAPI` would be mutually exclusive (in docs & validation, no way to enforce in schema directly). So you definitely still need this guard in the runtime.

But I am wondering about this case where you want both simplification & SDF to occur. Does it imply we need to revisit the mutually exclusive part of the schema design?

newton/_src/sim/builder.py
```diff
@@ -10191,6 +10291,36 @@ def _create_primitive_mesh(stype: int, scale: Sequence[float] | None) -> NewtonM
...
+                            tuple(sdf_narrow_band_range),
+                            sdf_target_voxel_size,
+                            sdf_max_resolution,
+                            sdf_tex_fmt,
+                            sdf_gen_margin,
+                        )
+                        mesh_sdf = deferred_mesh_sdf_cache.get(deferred_key)
+                        if mesh_sdf is None:
+                            mesh_copy = shape_src.copy()
+                            mesh_copy.build_sdf(**sdf_kwargs)
```
adenzler-nvidia, line 10409: The deferred build populates the simplified collision-edge set on `mesh_copy` (`build_sdf()` → `_build_collision_edges`, on by default at the 0.1° threshold), but only `mesh_copy.sdf` is retained — the clone is then discarded. The edge-consumption loop later in `finalize()` reads `mesh = self.shape_source[i]` (the original `shape_src`), whose `_collision_edges` is still `None`, so it falls back to the full `mesh.edges`.

Net effect: USD-imported SDF meshes silently use the unsimplified edge set, diverging from the directly-attached `mesh.sdf` path that gets the simplified edges. Not a crash (full edges are correct, just unsimplified), but the edge-simplification optimization never reaches meshes imported through this path.

Consider copying `mesh_copy._collision_edges` back onto `shape_src` after the deferred build, or keying the edge-consumption loop off the deferred result rather than `shape_src._collision_edges`.

newton/_src/utils/import_usd.py
```diff
@@ -2651,6 +2651,101 @@ def _build_mass_info_from_shape_geometry(
...
+                )
+
+                sdf_texture_format = R.get_value(
+                    prim, prim_type=PrimType.SHAPE, key="sdf_texture_format", verbose=verbose
+                )
+                if sdf_texture_format is None:
+                    sdf_texture_format = builder.default_shape_cfg.sdf_texture_format
+
+                sdf_padding = R.get_value(prim, prim_type=PrimType.SHAPE, key="sdf_padding", verbose=verbose)
+                if sdf_padding == float("-inf"):
```
adenzler-nvidia, line 2791: Only the exact `-inf` sentinel is normalized here. A negative-finite (e.g. `-0.05`) or `NaN` `newton:sdfPadding` passes straight through: `ShapeConfig.validate()` never inspects `sdf_padding`, and in `finalize()` the value becomes `sdf_gen_margin` and is passed as `margin=` to `create_texture_sdf_from_mesh`.

A `NaN` margin produces `NaN` AABB extents that slip past the `< 1e-10` degeneracy check, yielding a silently-garbage SDF; a negative-finite value shrinks the AABB inside the mesh. Both go undiagnosed.

Consider validating `sdf_padding` (finite and `>= 0`, sentinel excepted) at parse time and/or in `ShapeConfig.validate()`, emitting a warning that includes the prim path on an invalid authored value.

adenzler-nvidia (commented):
A couple of correctness notes on the deferred-SDF path and padding validation (details inline).

mzamoramora-nvidia:
@andrewkaufman Sometimes there are assets that have very complex meshes for which it could make sense to run mesh simplification and then build the sdf on top of that.

newton/_src/sim/builder.py
mzamoramora-nvidia: Done in 42c824b

newton/tests/test_sdf_usd.py
```diff
@@ -0,0 +1,824 @@
...
+            builder = newton.ModelBuilder()
+            result = parse_usd(builder, str(usd_path))
+            s1 = result["path_shape_map"]["/World/Body1/CollisionMesh"]
+            # shape_source still has no collision edges before finalize (the
+            # deferred build runs inside finalize on a Mesh clone).
+            self.assertIsNone(builder.shape_source[s1]._collision_edges)
+            full_edge_count = len(builder.shape_source[s1].edges)
+
+            model = builder.finalize(device=device)
+            simplified_edge_count = int(model.edge_count)
```
adenzler-nvidia: This asserts on the wrong attribute, so the test doesn't actually exercise the fix.

`Model.edge_count` is the count of **cloth bending edges** (`edge_count = len(edge_rest_angle)` in `builder.py`, and `edge_rest_angle` is only populated by the `add_edge`/`add_edges` cloth paths). A rigid `UsdGeom.Mesh` + `CollisionAPI` never touches `edge_rest_angle`, so `model.edge_count == 0` here. The assertion `0 < full_edge_count` passes trivially — and passes identically if this PR's fix is reverted, since the revert only affects `mesh_edge_indices`/`shape_edge_range`, never `edge_count`.

The simplified collision edges this change forwards land in `model.shape_edge_range` / `model.mesh_edge_indices`. Assert on those instead, e.g.:

```python
model = builder.finalize(device=device)
start, count = model.shape_edge_range.numpy()[s1]
simplified_edge_count = int(count)
self.assertLess(simplified_edge_count, full_edge_count, ...)
```

As written the test won't fail against an unfixed tree.

newton/_src/utils/import_usd.py
mzamoramora-nvidia: Added warnings to `sdfPadding` and to other attributes in 692aed6 . In general current parsing usd attributes does not guard against `NaN`. Not sure if that needs wider audit in the codebase  in a follow up PR or if should start implement `NaN` guards here.

newton/tests/test_sdf_usd.py
mzamoramora-nvidia: Done in 7ca6021

newton/_src/utils/import_usd.py
andrewkaufman: I think the limits warnings you added here are a good improvement.

Re nan, I didn't even realize it was possible to serialize a nan to USD, but I just checked & it is :scream: `double radius = nan`. That said, I would question whether anyone has ever done so... I generally don't think we need to guard against unlikely events that have never occured in practice. Certainly shoud be out of scope for this PR to start doing so.

andrewkaufman:
> Sometimes there are assets that have very complex meshes for which it could make sense to run mesh simplification and then build the sdf on top of that.

Could it make sense to expose an `approximation` attribute directly on SDFCollisionAPI with the allowed tokens for the different simplification methods? (we can consider this next release rather than try to cram it in for 1.3)

andrewkaufman:
The other PR has merged & 0.3.0 is released now, so you can drop the deps update from this branch. Unfortunately that also requires a conflict resolution in schemas.py & import_usd.py...

mzamoramora-nvidia:
@andrewkaufman Dep changes have been dropped and conflicts have been solved.  I agree that an `approximation` token could be a good path forward to enable building an sdf after remeshing in a follow up PR. 

For now, this PR enforces mutual exclusivity at parsing time and runtime.  As I see it, introducing the `approximation` token in the future will be a change of behavior and not a breaking change, but maybe @adenzler-nvidia can comment on that.

adenzler-nvidia (commented):
A few follow-ups, mostly around the raise-vs-warn policy (raise only when no workable model remains; otherwise warn and leave the prim importable) plus a couple of small correctness/doc nits. Details inline.

newton/_src/utils/import_usd.py
```diff
@@ -2689,6 +2689,137 @@ def _build_mass_info_from_shape_geometry(
...
+                # attached mesh.sdf or a resolution/voxel_size so one can be
+                # built deferred.
+                if (
+                    is_hydroelastic
+                    and key == UsdPhysics.ObjectType.MeshShape
+                    and sdf_max_resolution is None
+                    and sdf_target_voxel_size is None
+                ):
+                    raise ValueError(
+                        f"{prim.GetPath()}: hydroelastic mesh requires newton:sdfMaxResolution "
```
adenzler-nvidia, line 2840: Per the raise-vs-warn convention (raise only when no workable model remains), this should warn-and-degrade instead of aborting the whole import. A hydroelastic mesh with no SDF source is recoverable: warn naming the prim, disable hydroelastic on this shape, and import it as a normal mesh collider. As written, one mis-authored prim aborts `parse_usd()` for the entire stage with no partial model returned — and it's inconsistent with the `<= 0` checks just above (voxel/res/padding/kh), which already warn-and-default.

Reachability is narrow: this only triggers when `newton:hydroelasticEnabled=true` is authored as a bare attribute *without* applying `NewtonSDFCollisionAPI` (a properly applied schema auto-fills `sdfMaxResolution=64`), so it's an unusual hand-authoring case — all the more reason not to abort the whole import for it.

newton/_src/utils/import_usd.py
```diff
@@ -2805,24 +2943,66 @@ def _build_mass_info_from_shape_geometry(
...
+                    # so we strip SDF fields and pass a clean cfg. SDF building is
+                    # deferred to finalize() where instances can be deduplicated.
+                    # We write SDF params directly to the builder's per-shape lists
+                    # after the shape is added. Validate the SDF fields here so
+                    # invalid values (e.g. sdf_max_resolution not divisible by 8,
+                    # unknown sdf_texture_format) are caught before stripping
+                    # bypasses ShapeConfig.validate. shape_type=None skips the
+                    # hydroelastic-requires-SDF check (which only fires for
+                    # primitives) while still running the SDF-format checks.
+                    shape_params["cfg"].validate(shape_type=None)
```
adenzler-nvidia: Same raise-vs-warn point as the hydroelastic-mesh check below: `cfg.validate(shape_type=None)` raises on an unknown `sdfTextureFormat` or a `sdfMaxResolution` not divisible by 8, aborting the entire import for a single shape. A well-formed, schema-applied prim with one typo (`newton:sdfMaxResolution=100`) dies here. This is recoverable — warn (naming the prim + the offending attribute) and drop the SDF config for this shape, importing it as a plain collider — which would also match the `<= 0` warn-and-default checks a few lines above. Suggest validating the SDF fields inline with warn-and-degrade rather than calling the raising `validate()` in the import loop.

newton/_src/utils/import_usd.py
```diff
@@ -2689,6 +2689,137 @@ def _build_mass_info_from_shape_geometry(
...
+                )
+                if sdf_texture_format is None:
+                    sdf_texture_format = builder.default_shape_cfg.sdf_texture_format
+
+                sdf_padding = R.get_value(prim, prim_type=PrimType.SHAPE, key="sdf_padding", verbose=verbose)
+                if sdf_padding == float("-inf"):
+                    # Schema sentinel: builder.finalize falls back to shape_gap
+                    # when ShapeConfig.sdf_padding is None.
+                    sdf_padding = None
+                elif sdf_padding is not None and sdf_padding < 0:
```
adenzler-nvidia, line 2793: `sdf_padding < 0` is caught, but a NaN slips through: `NaN < 0` is `False`, so a NaN `newton:sdfPadding` passes to `create_texture_sdf_from_mesh(margin=NaN)`, yielding NaN AABB extents that clear the `< 1e-10` degeneracy guard and produce a silently-garbage SDF. Add a `math.isfinite()` check alongside the `< 0` test (e.g. warn + fall back to default when not finite).

newton/_src/utils/import_usd.py
```diff
@@ -2689,6 +2689,137 @@ def _build_mass_info_from_shape_geometry(
...
+                # Resolve target_voxel_size first because it overrides
+                # sdf_max_resolution and the two are mutually exclusive in
+                # ShapeConfig.validate().
+                sdf_target_voxel_size = R.get_value(
+                    prim, prim_type=PrimType.SHAPE, key="sdf_target_voxel_size", verbose=verbose
+                )
+                # Schema sentinel is -inf meaning "use sdfMaxResolution instead";
+                # the <= 0 check also rejects any authored non-positive value as
+                # out of the documented (0, inf) range.
+                if sdf_target_voxel_size == float("-inf"):
```
adenzler-nvidia, line 2712: The `== float("-inf")` sentinel branch here (and the matching one for `sdf_padding` at line 2766) is effectively dead. `R.get_value` reads through `usd.get_attribute`, which returns the authored value or `None` (it gates on `HasAuthoredValue()`), and the resolver mapping default for these keys is `None`, not `-inf`. So this branch only fires if a user literally authors `-inf`; the normal unauthored case already falls through to the `None` handling below.

Harmless (both paths converge to `None`), but the comment implies a schema-fallback mechanism that doesn't actually flow through `get_attribute`. Either drop the `-inf` branches and treat `None` as the sole "unset" signal, or correct the comment. (Contrast `gap` / `shell_thickness`, whose resolver mapping defaults genuinely *are* `-inf`.)

newton/_src/sim/builder.py
```diff
@@ -302,6 +302,19 @@ class ShapeConfig:
             For MuJoCo, stiffness values will internally be scaled by masses.
             Users should choose kh to match their desired force-to-penetration ratio.
         """
+        sdf_padding: float | None = None
+        """SDF AABB padding [m] for primitive shapes. When a texture SDF is
+        generated from a primitive (box, sphere, capsule, cylinder, cone) during
```
adenzler-nvidia: This list omits `ellipsoid` ("box, sphere, capsule, cylinder, cone"), but `_create_primitive_mesh` does build a texture SDF for `GeoType.ELLIPSOID`, and the doc tip in `usd_parsing.rst` includes it. Worth matching here so the docstring and docs agree.

newton/tests/test_sdf_usd.py
```diff
@@ -0,0 +1,817 @@
...
+"""Tests for SDF USD attribute parsing."""
+
+import tempfile
+import unittest
+from pathlib import Path
+
+import warp as wp
+
+import newton
+from newton._src.utils.import_usd import parse_usd
```
adenzler-nvidia: Minor: this imports the internal `parse_usd` directly, whereas every sibling USD test uses the public `builder.add_usd(stage)` (same return dict, so `result['path_shape_map']` works identically). Prefer the public entry point to avoid coupling the test to `newton._src`.

mzamoramora-nvidia: Done in  4e95cd6

newton/_src/sim/builder.py
mzamoramora-nvidia: Done in 4e95cd6

newton/_src/utils/import_usd.py
mzamoramora-nvidia: Attributes were switched to have the `-inf` sentinel by default in 2a84cf7 to respect the current schema contract.

newton/_src/utils/import_usd.py
mzamoramora-nvidia: Warned and degrade was implemented in 5fc7e8a

newton/_src/utils/import_usd.py
mzamoramora-nvidia: Warned and degrade was implemented here 5fc7e8a

newton/_src/utils/import_usd.py
mzamoramora-nvidia: This is similar as the previous discussion on nans (https://github.com/newton-physics/newton/pull/2533#discussion_r3325504373), I would leave the nan guarding to a general audit of the schema parsing dealing with nans, if that's ok.

newton/tests/test_sdf_usd.py
```diff
@@ -0,0 +1,868 @@
...
+)
+add_function_test(
+    TestSDFUSDParsing,
+    "test_deferred_sdf_distinguishes_shape_scales",
+    TestSDFUSDParsing.test_deferred_sdf_distinguishes_shape_scales,
+    devices=devices,
+)
+# The two test_usd_hydroelastic_mesh_*_raises methods validate parse-time
+# ValueErrors on the host; they do not call wp.launch and need no CUDA
+# device, so they are registered as plain unittest methods.
```
adenzler-nvidia: This comment is stale. There are no `*_raises` methods anymore — they're `test_usd_hydroelastic_mesh_without_sdf_config_warns_and_disables` and `..._with_kh_without_sdf_config_warns_and_disables` — and they assert a `UserWarning` via `assertWarnsRegex`, not a parse-time `ValueError`. The warn-and-degrade behavior is exactly what they test, so the comment now contradicts the code.

Suggest something like:
```python
# The hydroelastic-mesh validation tests run host-side (no wp.launch / CUDA),
# so they're registered as plain unittest methods.
```
or just drop it.

newton/_src/sim/builder.py
```diff
@@ -302,6 +302,19 @@ class ShapeConfig:
             For MuJoCo, stiffness values will internally be scaled by masses.
             Users should choose kh to match their desired force-to-penetration ratio.
         """
+        sdf_padding: float | None = None
+        """SDF AABB padding [m] for primitive shapes. When a texture SDF is
```
adenzler-nvidia: This docstring runs ~12 lines, and the `a tight gap therefore implies a tight SDF ... accurate near-surface contact resolution` part reads as design-note prose rather than reference docs. Consider trimming to the essentials: padding [m] for primitive texture SDFs, falls back to `gap` when `None`, distinct from `gap`/`margin`, and rejected on `MESH`/`CONVEX_MESH` (use `Mesh.build_sdf(margin=...)` there).

newton/_src/sim/builder.py
```diff
@@ -10302,6 +10378,13 @@ def _create_primitive_mesh(stype: int, scale: Sequence[float] | None) -> NewtonM
                 sdf_target_voxel_size = self.shape_sdf_target_voxel_size[i]
                 sdf_max_resolution = self.shape_sdf_max_resolution[i]
                 sdf_tex_fmt = self.shape_sdf_texture_format[i]
+                sdf_padding = self.shape_sdf_padding[i]
+                # SDF AABB padding: use the per-shape sdf_padding when the
```
adenzler-nvidia: This fallback rationale duplicates the `ShapeConfig.sdf_padding` docstring almost verbatim — a one-liner suffices, e.g. `# Fall back to shape_gap when sdf_padding is unset (see ShapeConfig.sdf_padding).`

Same story for the deferred-build comment just below (~L10397): the `if SDF params were stored but mesh.sdf was not built yet` sentence restates the `if` condition. Keep the load-bearing *why* (build on a clone so shapes sharing one Mesh at different scale/params get distinct SDFs) and drop the restatement.

newton/_src/utils/import_usd.py
```diff
@@ -2805,24 +2973,62 @@ def _build_mass_info_from_shape_geometry(
                         default=mesh_maxhullvert,
                         verbose=verbose,
                     )
+                    # add_shape_mesh() rejects SDF params in ShapeConfig for meshes,
```
adenzler-nvidia: This 6-line comment and the `# Store SDF intent on the builder` comment at L2998 say the same thing twice, and most of it restates the visible `replace(...)` / assignment lines. Suggest condensing to a single line and dropping the L2998 one, e.g.:
```python
# add_shape_mesh() rejects SDF cfg fields on meshes; strip them and write the
# SDF intent to the builder lists, deferring the build to finalize().
```

adenzler-nvidia (commented):
Re-review focused on comment quality. One stale comment that needs fixing, plus several spots where comments run heavier than Newton's usual terse style — restating the code or repeating the same rationale across a file. Details inline.

(Breaking changes all check out and are documented in the CHANGELOG. The NaN-guard point is still on its separate thread.)

newton/_src/utils/import_usd.py
```diff
@@ -2689,6 +2689,167 @@ def _build_mass_info_from_shape_geometry(
...
+                        f"(must be > 0); falling back to default.",
+                        stacklevel=2,
+                    )
+                    kh = None
+                if hydroelastic_enabled is True:
+                    is_hydroelastic = True
+                elif hydroelastic_enabled is False:
+                    is_hydroelastic = False
+                elif has_sdf_api:
+                    # The API is applied but newton:hydroelasticEnabled is
```
adenzler-nvidia: The `hydroelasticEnabled` schema-default rationale is spelled out at length here, and the `kh authored alone is a material parameter and does NOT flip the feature on` rule is stated twice in this function (the block near L2803 and again at L3004). The tri-state `is True / is False / has_sdf_api / else` is already explicit. Suggest one short comment at the `elif has_sdf_api:` branch (`# API applied but hydroelasticEnabled unauthored -> schema default False, not builder default`) and a single `# kh is a material param; persist regardless of hydro state.`

docs/concepts/usd_parsing.rst
```diff
@@ -568,6 +568,516 @@ After importing the USD file with the custom attributes shown above, they become
...
+   _PHYSX_DEFAULTS = {
+       "physxSDFMeshCollision:sdfResolution": 256,
+       "physxSDFMeshCollision:sdfNarrowBandThickness": 0.01,
+       "physxSDFMeshCollision:sdfMargin": 0.01,
+       "physxSDFMeshCollision:sdfBitsPerSubgridPixel": "BitsPerPixel16",
+   }
+
+
+   class PortDefaults(enum.Enum):
+       """Controls how unauthored PhysX attributes are handled during porting.
```
adenzler-nvidia: In the doctest helper, the `PortDefaults` docstring (here) and the `_bbox_diag` docstring (~L866) largely restate the surrounding .rst prose, and the World-vs-Local-bound rationale is repeated a third time in the Xform-scale demo comment (~L1029). The hardcoded default values baked into these docstrings (`sdfMaxResolution=64`, etc.) will also drift from the real defaults over time. Suggest trimming each docstring to a single line, keeping the rationale in one place, and reducing the per-demo narration (~L946/990/1011) to one-line labels.

mzamoramora-nvidia: Done in [f81bef0](https://github.com/newton-physics/newton/pull/2533/commits/f81bef0c3fa998809678c496431352bd81587b7a)

newton/tests/test_sdf_usd.py
mzamoramora-nvidia: Done in [62a4a76](https://github.com/newton-physics/newton/pull/2533/commits/62a4a76664ff0de4aa67fbd156c1fbd28af5a864)

newton/_src/sim/builder.py
mzamoramora-nvidia: Done in [893a7d9](https://github.com/newton-physics/newton/pull/2533/commits/893a7d9aedb4e7e0c98210597d0f0ebe185eb1c5)

newton/_src/utils/import_usd.py
mzamoramora-nvidia: Done in [893a7d9](https://github.com/newton-physics/newton/pull/2533/commits/893a7d9aedb4e7e0c98210597d0f0ebe185eb1c5)

newton/_src/utils/import_usd.py
mzamoramora-nvidia: Done in [893a7d9](https://github.com/newton-physics/newton/pull/2533/commits/893a7d9aedb4e7e0c98210597d0f0ebe185eb1c5)

newton/_src/sim/builder.py
mzamoramora-nvidia: Done in [893a7d9](https://github.com/newton-physics/newton/pull/2533/commits/893a7d9aedb4e7e0c98210597d0f0ebe185eb1c5)

docs/concepts/usd_parsing.rst
mzamoramora-nvidia: [f81bef0](https://github.com/newton-physics/newton/pull/2533/commits/f81bef0c3fa998809678c496431352bd81587b7a)

adenzler-nvidia (approved):
LGTM. All prior feedback is addressed, including the deferred-mesh edge-set correctness bug (simplified collision edges now forwarded from the SDF clone in `finalize()`, with a genuine regression test) and the raise-vs-warn cleanup across the SDF/hydroelastic config path.

One non-blocking nit if you ever revisit: a NaN-authored `newton:sdfPadding` / `newton:hydroelasticStiffness` / `newton:sdfTargetVoxelSize` still slips past the `< 0` / `<= 0` guards (NaN compares false) into the SDF build. Narrow reachability; a `math.isfinite()` check folded into each guard would close it. Fine to defer.

mzamoramora-nvidia:
@adenzler-nvidia An issue was filed to consider an importer-wide audit for `NaN`guards in https://github.com/newton-physics/newton/issues/3024

PR #2539, opened 2026-04-22
Unify SolverFeatherstone on the public Newton twist convention

eric-heiden:
## Description

`SolverFeatherstone` previously maintained its own parent-origin twist basis for FREE/DISTANCE joints and converted into/out of it around every step via dedicated velocity-conversion kernels. That internal convention diverged from the public Newton `joint_qd` layout (child-COM linear velocity + angular velocity in the joint parent frame), which made the solver hard to compose with the rest of the stack (IK, semi-implicit contact, kinematics tests) and required duplicated motion-subspace helpers.

This PR makes `SolverFeatherstone` consume the public twist convention directly and keeps its internal floating-base solve numerically better centered:

- Replace the parent-origin FREE/DISTANCE twist basis with the public child-COM / parent-frame `joint_qd` convention; drop the solver-level velocity translation stages.
- Share a single FREE/DISTANCE motion-subspace helper between Featherstone and IK (LBFGS, LM); remove the now-dead Featherstone-only variants.
- Center Featherstone's internal solve frame at the root body COM for floating-root FREE/DISTANCE articulations, reducing large moment arms and improving stability when those articulations are far from the world origin.
- Fuse Featherstone FK into the `eval_rigid_id` pass, cache `body_q_com` for `eval_rigid_tau`, and keep public body twists/wrenches in the COM/world-frame contract.
- Fix the FREE/DISTANCE `jcalc_integrate` child-anchor COM conversion and descendant FREE/DISTANCE transport update.
- Align `articulation.py`, semi-implicit contact kernels, and kinematics/body-velocity tests with the public convention.

Net effect: ~300 fewer lines in the solver, one shared motion-subspace helper, no public/internal FREE/DISTANCE velocity basis translation, and better floating-root stability for Featherstone articulations far from the world origin.

## Checklist

- [x] New or existing tests cover these changes (`test_kinematics.py`, `test_body_velocity.py` updated to the public convention; FREE/DISTANCE FK FD regression reinstated)

## Test plan

```
uv run --extra dev -m newton.tests -k test_kinematics
uv run --extra dev -m newton.tests -k test_featherstone_
uv run --extra dev -m newton.tests -k test_control_force
uvx pre-commit run -a
```

## Summary by CodeRabbit

* **Changed**
  * FREE/DISTANCE joint velocities now use child-COM linear velocity + angular velocity in the parent frame (six values per joint); internal velocity-conversion stages removed.
  * FREE/DISTANCE `joint_f` remains the public COM-referenced wrench in world frame, matching the existing control-force tests.
  * Dynamics, integration and FK are fused into consolidated passes for more consistent COM-referenced twist/wrench handling and corrected FREE/DISTANCE behavior.
  * Floating-root Featherstone articulations use a root-COM-centered internal solve frame to improve stability far from the world origin while preserving public COM/world state conventions.

* **Tests**
  * Added/updated regression tests covering FREE/DISTANCE integration, rooted FK finite-difference velocity checks, and root-body stability.

* **Changelog**
  * Documented the behavioral/contract change for the solver in CHANGELOG.


jcarius-nv:

newton/_src/solvers/featherstone/solver_featherstone.py
```diff
@@ -502,12 +518,13 @@ def step(
...
 
-            # particle shape contact
-            eval_particle_body_contact_forces(model, state_in, contacts, particle_f, body_f, body_f_in_world_frame=True)
+            # particle shape contact for non-articulated models; articulated models run this after ID
+            # so contacts see the freshly reconstructed public COM twist.
+            if not model.joint_count:
+                eval_particle_body_contact_forces(model, state_in, contacts, particle_f, body_f)
 
             # muscles
-            if False:
```
gyeomannvidia, line 509: Is this a deliberate change?  I don't believe that we have a test that runs the Featherstone solver with a non-zero muscle count. Would it make more sense to disable this to match main?

newton/_src/solvers/featherstone/solver_featherstone.py
```diff
@@ -376,23 +387,37 @@ def _allocate_state_aux_vars(self, model, target, requires_grad):
...
+            # Body spatial acceleration/bias recurrence value expressed in the internal solve frame.
             target.body_a_s = wp.empty(
                 (model.body_count,), dtype=wp.spatial_vector, device=model.device, requires_grad=requires_grad
             )
+            # Per-body inertial bias minus gravity wrench in the internal solve frame.
             target.body_f_s = wp.zeros(
                 (model.body_count,), dtype=wp.spatial_vector, device=model.device, requires_grad=requires_grad
             )
-            target.body_f_ext = wp.zeros(
+            # External/contact body-force buffer kept public until eval_rigid_tau shifts it to the solve frame.
```
gyeomannvidia: I don't quite follow the buffer being "public". I think the point being made is that body_f_ext adopts the public wrench convention until eval_rigid_tau changes it to the solve frame. Can we make this clearer?

newton/_src/solvers/featherstone/kernels.py
```diff
@@ -818,14 +824,17 @@ def compute_link_velocity(
...
-    f_g_s = wp.spatial_vector(f_g, wp.cross(r_com, f_g))
+    f_g_s = wp.spatial_vector(f_g, wp.cross(x_com_s, f_g))
 
     # body forces
-    I_s = transform_spatial_inertia(X_sm, I_m)
+    X_sm_s = wp.transform(x_com_s, wp.transform_get_rotation(X_sm))
+    I_s = transform_spatial_inertia(X_sm_s, I_m)
 
     f_b_s = I_s * a_s + spatial_cross_dual(v_s, I_s * v_s)
+    omega_world = wp.spatial_bottom(v_s)
```
gyeomannvidia, line 834: Could we instead use origin_twist_to_com_twist here?

gyeomannvidia (dismissed):

newton/_src/solvers/featherstone/solver_featherstone.py
eric-heiden: Indeed, this change was unintentional. I will revert it.

newton/_src/solvers/featherstone/solver_featherstone.py
eric-heiden: Done in `bdf6ec937`. I clarified the comment to say that `body_f_ext` stores public COM/world wrenches before `eval_rigid_tau`, and that `eval_rigid_tau` shifts articulated entries into the solve frame.

newton/_src/solvers/featherstone/kernels.py
eric-heiden: I looked into this. `origin_twist_to_com_twist()` assumes the incoming twist is referenced at the body origin, using `X_wb` plus the body-local COM offset to find the COM shift.

Here `v_s` is referenced at Featherstone's solve origin, which can be translated to the floating root COM rather than the child body origin. The current expression is the solve-frame equivalent of shifting the twist to the child COM: `v_com = v_s.linear + omega x (x_com - solve_origin)`. I left this explicit so the solve-origin offset stays visible and we do not accidentally shift from the body origin instead.

gyeomannvidia (approved):

newton/_src/solvers/featherstone/kernels.py
gyeomannvidia: Thanks for checking. Definitely makes sense to leave as is.

newton/_src/solvers/featherstone/solver_featherstone.py
gyeomannvidia: Thanks!

Issue #2531, opened 2026-04-22
[BUG] [isaac-sim6.0][newton][rc18] Menu assets: Simple Room Qhull error


preist-nvidia:
### Bug Description

**Summary**
Synopsis: [isaac-sim6.0][newton][rc18] while trying to load the asset given at menu options with isaac-sim.newton app observing multiple type of error logs in isaac sim

Summary: When loading assets from menu options with ./isaac-sim.newton, multiple errors appear; some flows lead to a consistent application crash (see Issue 2).

**Build / environment**
Sim build: Isaac Sim 6.0 rc18
GPU: NVIDIA GeForce RTX 5090
Driver: 590.95.05
OS: Linux
NVBugs OperatingSystems: Linux - Ubuntu 22.04
Version (NVBugs): Kit 109.0.1
Prev passing version: Kit 107.3.3
Regression (NVBugs field): Not a regression
Found In: isaacsim6.0-rc2
Fixed In: TBD

### Reproduction Script

```python
Issue 1 — Simple Room (Qhull precision error)
Steps

Launch Isaac with Newton: ./isaac-sim.newton.sh
Create → Environments → Simple Room
Press Play
Observed

2026-03-10T09:24:06Z [102,375ms] [Error] [isaacsim.physics.newton.impl.simulation_functions] [Newton] Simulation step error: QH6154 Qhull precision error: Initial simplex is flat (facet 1 is coplanar with the interior point)
```

preist-nvidia:
@rthaker01 - can we change the ground floor to use a triangle-mesh collider? And can you kindly provide the repro asset for this, ideally flattened or isolated to the component/mesh that triggers above?

@nvtw - can you comment on whether we should robustify the convex hull gen / reject flat meshes?

nvtw:
I researched a bit and the only solution I found is to analyze the input before passing it into QHull. We would need a helper function that does this analysis and gives the points to a 2d convex hull algorithm in case they all lie in the same plane and then triangulate the 2d convex hull. Non-degenerate input could be forwarded to QHull.

preist-nvidia:
Thanks @nvtw - if you think it's useful for us to be able to create convex hulls from 2d meshes, we should add.

Otherwise we could also detect and raise letting the user know that they should use the tri-mesh directly as collider or provide a suitable mesh.

@eric-heiden for comment as well.

nvtw:
[...]

PR #2540, opened 2026-04-22
Allow zero volume convex hulls

nvtw:
## Description

See https://github.com/newton-physics/newton/issues/2531

Zero volume convex hulls work for collision detection. The question is what inertia and mass they should get in case they are used on dynamic bodies. Maybe the convex hull inertia calculation code should throw an exception for zero volume convexes?

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **Enhancements**
  * Added an eps parameter for stability, deterministic fallback meshes for degenerate point clouds (coincident, collinear, coplanar), preserved outward-facing winding, and a retry path for robust 3D hull generation.
* **Bug Fixes**
  * Empty input now raises an error; degenerate cases emit warnings and return consistent zero-volume outputs.
* **Tests**
  * Comprehensive suite covering full-rank, degenerate, near-flat, vertex-limited, randomized "never-raises", and warning behavior.
* **Changelog**
  * Entry describing degenerate handling, warning behavior, empty-input error, and retry behavior.

### DIFF
--- a/newton/_src/geometry/utils.py
+++ b/newton/_src/geometry/utils.py
@@ -702,9 +702,15 @@ def remesh_convex_hull(vertices: np.ndarray, maxhullvert: int = 0, eps: float = 
         - verts: A numpy array of shape (M, 3) containing the vertex positions of the convex hull.
         - faces: A numpy array of shape (K, 3) containing the vertex indices of the triangular faces of the convex hull.
 
+    Raises:
+        ValueError: If ``vertices`` is empty. Empty input has no geometric
+            interpretation; the caller must decide whether to skip the hull
+            computation or supply a fallback rather than having this function
+            fabricate a point at the origin.
+
     Guarantees:
-        - Never raises on degenerate input; always returns a well-formed
-          ``(verts, faces)`` pair with ``M >= 3`` and ``K >= 2``.
+        - Never raises on non-empty degenerate input; always returns a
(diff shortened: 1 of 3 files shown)


eric-heiden:
⚪ This change makes degenerate convex hull handling more complete. The main remaining risks are empty inputs turning into real geometry and flat mesh remeshing silently changing shape semantics.

newton/_src/geometry/utils.py
```diff
@@ -611,38 +611,222 @@ def remesh_quadratic(
...
+          by ~1e-11 and returns the convex hull of the *perturbed* input.
     """
 
-    from scipy.spatial import ConvexHull
+    from scipy.spatial import ConvexHull, QhullError
+
+    vertices = np.asarray(vertices, dtype=np.float64)
 
+    # Degenerate: empty or single point.
+    if vertices.shape[0] == 0:
```
eric-heiden, line 766: 🔴 This empty-input fallback fabricates a real collider at the origin.

For callers like `Mesh.compute_convex_hull()` and `ModelBuilder.approximate_meshes(..., "convex_hull")`, an empty mesh now becomes three zero vertices plus two triangles instead of staying empty or failing. That can create phantom geometry in a simulation when an imported mesh is accidentally empty.

Please keep the empty case empty, or raise and let the higher-level caller decide whether to skip or provide a fallback.

eric-heiden (approved):

PR #2544, opened 2026-04-22
Disentangle BVH from Warp Raytracer move into model with public BVH API

StafaH:
## Description

This PR moves the BVH fields from the Tiled Camera Sensor to the Model, in preparation for sharing this with the Raycast Sensors.

A couple important things are happening:
- BVH fields have been moved to Model
- `bvh.py` is migrated to geometry, and a new public API is available for buildings and refitting the BVH structures. These new functions take Model and State as arguments
- Tiled Camera Sensor now does not handle BVH logic and the user must now call these operations to initialize and update the BVH every step

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Public BVH build/refit helpers added for shape and particle acceleration structures.
* **Documentation**
  * API docs updated to include the new BVH helpers.
* **Breaking Changes**
  * Camera/render APIs now require callers to provide an explicit simulation state and prepare BVHs externally; automatic in-call BVH refit behavior removed.
* **Examples & Tests**
  * Examples and tests updated to build/refit BVHs before sensor/render calls.


newton/geometry.py
```diff
@@ -17,6 +19,8 @@
     collide_sphere_capsule,
     collide_sphere_cylinder,
     collide_sphere_sphere,
+    refit_particle_bvh,
+    refit_shape_bvh,
```
eric-heiden: I wonder if we could make these `build_*` and `refit_*` functions simply methods on Model? Ideally the prefixes should be shared so consider naming them `build_bvh_shape()`, `refit_bvh_particle()`, etc.

StafaH: I think it will be better to avoid mixing bvh api and logic with the model classes. The existing functions in the model class focus on getting and setting the fields of the class, better to keep it that way and keep bvh logic seperate.

newton/_src/sensors/sensor_tiled_camera.py
```diff
@@ -191,8 +195,12 @@ def update(
         ``[world_id, camera_id, y, x]`` corresponds to the ray in ``camera_rays[camera_id, y, x]``. Each output
         channel is optional -- pass None to skip that channel's rendering entirely.
 
+        Shape and particle BVHs on :attr:`model` must be refit for *state* via
+        :func:`~newton.geometry.refit_bvh_shape` and
+        :func:`~newton.geometry.refit_bvh_particle` before calling this method.
+
         Args:
-            state: Simulation state with body transforms. If not None, calls :meth:`sync_transforms` first.
```
eric-heiden, line 195: 🔴 This change breaks the public `SensorTiledCamera.update()` call pattern without a compatibility window. Callers that currently do `update(None, ..., refit_bvh=False)` will now fail with a signature error, and there is no equivalent fallback because `sync_transforms()` no longer maintains the rigid-shape BVH path by itself.

Can we keep the old arguments for one release, emit a `DeprecationWarning`, and translate them to the new explicit BVH lifecycle internally? For example, `state: State | None = None` and `refit_bvh: bool = True` would let existing code keep working while pointing users at the new `build_bvh_*()` / `refit_bvh_*()` calls.

newton/_src/sensors/warp_raytrace/render_context.py
```diff
@@ -412,6 +245,8 @@ def render(
                 render_kernel = create_kernel(self.config, self.state, clear_data)
                 self.kernel_cache[kernel_cache_key] = render_kernel
 
+            particle_count = state.particle_q.shape[0] if has_particles else 0
```
eric-heiden, line 270: 🔴 `particle_count` now comes from `state.particle_q`, but this path does not verify that `model.build_bvh_particle()` has been called. A particle scene can enter the kernel with a nonzero particle count and unbuilt `model.bvh_particles_group_roots`, which means `closest_hit_particles()` may index invalid BVH data on first render.

The same lifecycle gap exists for shapes: `bvh_shape_count_enabled == 0` is treated like an empty scene, so rigid geometry can be silently skipped when the model BVH has not been built yet. Please either auto-build on first render to preserve the old behavior, or fail fast when geometry exists but the matching model BVH is still unbuilt.

newton/_src/sensors/sensor_tiled_camera.py
```diff
@@ -162,18 +162,23 @@ def __init__(self, model: Model, *, config: Config | RenderConfig | None = None,
                 self.utils.assign_random_colors_per_shape()
 
     def sync_transforms(self, state: State):
-        """Synchronize shape transforms from the simulation state.
+        """Synchronize triangle-mesh points from the simulation state.
 
         :meth:`update` calls this automatically when *state* is not None.
 
+        Shape and particle BVHs on :attr:`model` must be refit separately
+        via :func:`~newton.geometry.refit_bvh_shape` and
```
eric-heiden: 🟡 This documentation tells users to call `refit_bvh_shape()` / `refit_bvh_particle()` before `update()`, but that sequence fails on a fresh model because the refit helpers require an existing BVH. First use needs `build_bvh_shape()` / `build_bvh_particle()`, then later frames can call the refit helpers.

Please update the class example and these method docs to show the full lifecycle: build once after setup, then refit before each frame that changes geometry.

eric-heiden:
🟡 This PR changes the public `SensorTiledCamera` BVH lifecycle and adds new public `newton.geometry` BVH helpers, but the `[Unreleased]` changelog does not mention either change.

Please add a `Changed` entry with migration guidance for users moving from implicit camera BVH maintenance to explicit `build_bvh_*()` / `refit_bvh_*()` calls. It would also be useful to add an `Added` entry for the new public BVH helper functions.

eric-heiden (approved):
Thanks!

PR #2547, opened 2026-04-22
Add Mesh Quality Validation Feature

### LINKED ISSUE #2222: Implement option to validate meshes passed to `add_cloth_mesh()`, `add_soft_mesh()`
It often happens that assets are parsed that have poor quality, such as small areas in certain elements, or sliver elements, etc.

We should add an option, such as `validate_mesh: bool = False` to these ModelBuilder methods that will perform some checks on the input triangle / tet mesh to check for these deficiencies and raise an error to point the user to use some one of the remeshing methods in `newton.utils.remesh_mesh()`.

AnkaChan:
## Description

Add opt-in mesh quality validation for `add_cloth_mesh()` and `add_soft_mesh()`. When `validate_mesh=True` is passed, the builder inspects input geometry for common defects (degenerate elements, slivers, non-manifold topology) and emits `warnings.warn()` with actionable diagnostics. No exceptions are raised — the mesh is always added regardless.

Closes #2222

## Checklist

## Test plan

```
uv run --extra dev -m unittest newton.tests.test_mesh_validation -v
```

18 tests covering all quality checks for both triangle and tet meshes: clean mesh (no warning), empty mesh, each threshold individually, borderline pass, mixed issues, and public API export.

## New feature / API change

```python
import newton

# Opt-in validation on cloth mesh — warns if quality issues detected
builder.add_cloth_mesh(
    pos=(0, 0, 0),
    rot=(0, 0, 0, 1),
    scale=1.0,
    vertices=verts,
    indices=tris,
    validate_mesh=True,  # NEW
)

# Same for soft bodies
builder.add_soft_mesh(
    pos=(0, 0, 0),
    rot=(0, 0, 0, 1),
    scale=1.0,
    vertices=verts,
    indices=tets,
    validate_mesh=True,  # NEW
)

# Or call validators directly
from newton.utils import validate_triangle_mesh, validate_tet_mesh

validate_triangle_mesh(verts_np, tris_np)
validate_tet_mesh(verts_np, tets_np)
```

### Triangle mesh checks (all assume SI meters)

| Check | Threshold |
|---|---|
| Small area | area < 1 mm² (1e-6 m²) |
| Sliver (aspect ratio) | longest_edge / shortest_altitude > 10 |
| Extreme angle | min interior angle < 5° |
| Non-manifold edge | edge shared by > 2 faces |

### Tet mesh checks

| Check | Threshold |
|---|---|
| Inverted | signed volume < 0 |
| Small volume | \|volume\| < 1 mm³ (1e-9 m³) |
| Sliver (eta quality) | η < 0.01 (regular tet = 1.0) |
| Non-manifold face | face shared by > 2 tets |

### Example warning output

```
my_sim.py:42: UserWarning: Mesh quality warning (873 vertices, 1544 triangles):
  - 4 sliver triangle(s) with aspect ratio > 10 (worst: 18.9)
  - 1 triangle(s) with minimum angle < 5° (smallest: 4.0°)
Consider remeshing the input geometry.
  builder.add_cloth_mesh(pos=pos, rot=rot, ..., validate_mesh=True)
```

### Files changed

| File | Change |
|---|---|
| `newton/_src/utils/mesh.py` | `validate_triangle_mesh()`, `validate_tet_mesh()` |
| `newton/utils.py` | Export both functions |
| `newton/_src/sim/builder.py` | `validate_mesh` param on `add_cloth_mesh`, `add_soft_mesh` |
| `newton/_src/solvers/style3d/cloth.py` | `validate_mesh` param on Style3D `add_cloth_mesh` |
| `newton/tests/test_mesh_validation.py` | 18 unit tests |

## Summary by CodeRabbit

* **New Features**
  * Optional mesh validation when adding cloth and soft meshes (opt-in flag) to detect degenerate geometry/topology before use
  * New public mesh diagnostics for triangle and tetrahedral quality that emit consolidated warnings and guidance

* **Tests**
  * Comprehensive tests covering mesh-quality/topology warnings, edge cases, threshold behavior, and builder integration

* **Documentation**
  * Changelog and API docs updated to include the new validation utilities and opt-in behavior


gdaviet (commented):
Looks good to me. 
For usability I wonder it it could make sense to add an optional `label: str` parameter to the `Builder.add_xxx_mesh` functions, like it already exist for the `add_shape_xxx`. This would allow to map the warning to a well identified mesh, for instance by usd path

adenzler-nvidia:
Thanks for putting this together — the sliver / min-angle / tiny-area thresholds for triangles, and the η-quality / inversion / tiny-volume checks for tets, are genuinely new and useful. Before merging though, I'd like us to reconcile this with existing mesh-validation pieces that already live in the tree, so we don't end up with parallel implementations.

**Already present:**

- `newton.utils.MeshAdjacency` (`newton/_src/utils/mesh.py`) already builds triangle edge adjacency and emits `warnings.warn("Detected non-manifold edge", ...)` when an edge would be shared by more than two faces. The triangle non-manifold check in `validate_triangle_mesh` reimplements this from scratch.
- `Mesh.is_watertight` (`newton/_src/geometry/types.py`) already does the "every edge shared by exactly two triangles" check, with vertex welding at 100 nm. The SDF construction path in `sdf_utils.py` already branches on it (fast parity vs. winding-number fallback), so the notion of "non-manifold / non-watertight" is established.
- `newton.utils.remesh_mesh` is the public dispatcher for the existing repair tools (`remesh_ftetwild`, `remesh_alphashape`, `remesh_quadratic`, `remesh_convex_hull`). There's also an SDF-based `remesh_poisson` pipeline in `newton/_src/geometry/remesh.py` (multi-view raycast → Poisson surface reconstruction → `extract_largest_island`) whose docstring literally advertises "remesh a problematic mesh to get a clean, watertight version" — but it's not exposed publicly today.

**What I'd like to see before merging:**

1. **Reuse the existing topology check.** The non-manifold edge detection in `validate_triangle_mesh` should delegate to `MeshAdjacency` (or a small shared helper factored out of it) rather than reimplement edge-pair counting. If we want a watertight check, `Mesh.is_watertight` is the precedent.
2. **Make the warning message actionable.** Instead of "Consider remeshing the input geometry", point at a concrete entry point — e.g. `newton.utils.remesh_mesh(...)`. If we'd like to recommend the SDF-based pipeline, let's promote `remesh_poisson` (and `extract_largest_island`) to public as part of this PR; it's what the advice really means.
3. **Keep the genuinely new checks.** The sliver / angle / tet-quality thresholds don't duplicate anything and are the core value add — those stay as-is.

Optional follow-up (doesn't have to block this PR): consider consolidating these pieces behind a single `validate_mesh`-style entry point or a `MeshQualityReport` that composes topology checks (from `MeshAdjacency` / `is_watertight`) with the new geometric-quality checks, so future users have one obvious place to look.

eric-heiden:
> 2. **Make the warning message actionable.** Instead of "Consider remeshing the input geometry", point at a concrete entry point — e.g. `newton.utils.remesh_mesh(...)`. If we'd like to recommend the SDF-based pipeline, let's promote `remesh_poisson` (and `extract_largest_island`) to public as part of this PR; it's what the advice really means.

Note that `poisson` is already an available method that can be selected from `remesh_mesh` - no need to expose `remesh_poisson` as a separate entry point and shouldn't be done as part of this PR. So just a message pointing the user to `newton.utils.remesh_mesh(...)` is sufficient.

AnkaChan:
@adenzler-nvidia @eric-heiden The `remesh_mesh` cannot solve the quality issues here. 

First, it does not work on tet meshes. Second, it cannot improve the quality of a cloth mesh.

Methods "ftetwild", "alphashape" and "convex_hull" are mesh converters rather than quality improvers. "quadratic" actually usually makes quality worse rather than better. "Poisson" may improve mesh quality, but it only works for **water tight** triangular meshes, which is almost always not the case for cloth meshes. 

This PR aims for improving the quality of simulation meshes for soft bodies & cloth. It only checks the input to `add_cloth_grid/mesh` and `add_soft_grid/mesh`, those are not the main application of the SDF-based pipeline. To really improve the quality of deformable meshes we need something like "[Isotropic Explicit Remeshing](https://graphics.stanford.edu/courses/cs164-10-spring/Handouts/isotropic.pdf)". But that will introduce new dependencies, do we want that?

That is why I intentionally avoided pointing the user to `newton.utils.remesh_mesh(...)`. Because I don't think that will help.

AnkaChan:
Hi @gdaviet,

I went a bit further than warning-text-only. `ModelBuilder` now keeps a tiny deformable registry — two parallel arrays, one entry per cloth/soft body, with offsets in CSR form so deformable `i` covers particles `[deformable_offset[i], deformable_offset[i+1])`:

```python
# newton/_src/sim/builder.py (ModelBuilder.__init__)
self.deformable_label:  list[str] = []        # length = deformable_count
self.deformable_offset: list[int] = [0]       # length = deformable_count + 1

# newton/_src/sim/model.py (Model.__init__) — finalized form
self.deformable_label:  list[str] = []                       # plain list
self.deformable_offset: wp.array[wp.int32] | None = None     # kernel-readable

# newton/_src/sim/builder.py (finalize)
m.deformable_label  = list(self.deformable_label)
m.deformable_offset = wp.array(self.deformable_offset, dtype=wp.int32)
```

Each `add_cloth_mesh` / `add_cloth_grid` / `add_soft_mesh` / `add_soft_grid` (plus the Style3D variants) calls a single helper after its particle-add work is done. The helper is the only place that mutates the registry, which keeps the N+1 invariant trivially correct and the default-label format in one spot:

```python
def _register_deformable(self, kind: str, particle_start: int, label: str | None) -> None:
    if not self.deformable_label:
        # First registration: anchor offset[0] to where this deformable actually starts,
        # so prior free particles don't shift the range.
        self.deformable_offset[0] = particle_start
    name = label if label is not None else f"{kind}_{len(self.deformable_label)}"
    self.deformable_label.append(name)
    self.deformable_offset.append(self.particle_count)
```

`kind` is the source-function tag (`"cloth_mesh"`, `"cloth_grid"`, `"soft_mesh"`, `"soft_grid"`) so default labels read like `cloth_mesh_0`, `soft_grid_3`. Single global counter — `len(self.deformable_label)` — so IDs are unique within a builder regardless of kind. Grids that internally delegate to the mesh path (`add_cloth_grid` → `add_cloth_mesh`) compute their own `cloth_grid_<id>` default first and pass it through as `label=`, so the mesh path uses the explicit label rather than its own default.

`add_builder` merges with a particle-index shift; the only non-obvious bit is resetting `offset[0]` when `self` had no prior deformables, so the merged offsets still anchor to wherever the sub-builder's first deformable actually lands:

```python
if builder.deformable_count:
    if not self.deformable_label:
        self.deformable_offset = [start_particle_idx + builder.deformable_offset[0]]
    self.deformable_label.extend(builder.deformable_label)
    for off in builder.deformable_offset[1:]:
        self.deformable_offset.append(start_particle_idx + off)
```

Validator side, `validate_triangle_mesh` / `validate_tet_mesh` gained `label: str | None = None` and a tiny prefix-or-don't block — `None` produces no `[None]` artefact for direct callers:

```python
prefix = "Mesh quality warning"
if label is not None:
    prefix += f" [{label}]"
msg = f"{prefix} ({n_verts} vertices, {n_faces} triangles):\n" + ...
```

USD soft bodies pick up labels automatically. `add_usd` was already routing `UsdGeom.TetMesh` prims through `add_soft_mesh`, so they auto-register; I just wired the prim path through as `label=`:

```python
# newton/_src/utils/import_usd.py (single-key addition to the kwargs dict)
add_soft_mesh_kwargs = {
    "pos": soft_mesh_pos, "rot": soft_mesh_rot, "scale": 1.0,
    "vel": wp.vec3(0.0, 0.0, 0.0),
    "mesh": tetmesh_for_builder,
    "label": path,        # prim.GetPath(), e.g. "/World/Characters/Hero/Body"
}
```

USD cloth isn't currently imported by `add_usd` — if/when that lands, the same `label=prim.GetPath()` pattern applies. Commits: `64ccdca8` (registry + plumbing), `85f71b97` (Model propagation), `903d1bde` (`test_model.py::TestModelDeformables`), `b2734dcf` (USD prim-path forwarding); 53 tests across `test_mesh_validation`, `test_model`, and `test_import_usd` pass.

AnkaChan:
@

> Thanks for putting this together — the sliver / min-angle / tiny-area thresholds for triangles, and the η-quality / inversion / tiny-volume checks for tets, are genuinely new and useful. Before merging though, I'd like us to reconcile this with existing mesh-validation pieces that already live in the tree, so we don't end up with parallel implementations.
> 
> **Already present:**
> 
> * `newton.utils.MeshAdjacency` (`newton/_src/utils/mesh.py`) already builds triangle edge adjacency and emits `warnings.warn("Detected non-manifold edge", ...)` when an edge would be shared by more than two faces. The triangle non-manifold check in `validate_triangle_mesh` reimplements this from scratch.
> * `Mesh.is_watertight` (`newton/_src/geometry/types.py`) already does the "every edge shared by exactly two triangles" check, with vertex welding at 100 nm. The SDF construction path in `sdf_utils.py` already branches on it (fast parity vs. winding-number fallback), so the notion of "non-manifold / non-watertight" is established.
> * `newton.utils.remesh_mesh` is the public dispatcher for the existing repair tools (`remesh_ftetwild`, `remesh_alphashape`, `remesh_quadratic`, `remesh_convex_hull`). There's also an SDF-based `remesh_poisson` pipeline in `newton/_src/geometry/remesh.py` (multi-view raycast → Poisson surface reconstruction → `extract_largest_island`) whose docstring literally advertises "remesh a problematic mesh to get a clean, watertight version" — but it's not exposed publicly today.
> 
> **What I'd like to see before merging:**
> 
> 1. **Reuse the existing topology check.** The non-manifold edge detection in `validate_triangle_mesh` should delegate to `MeshAdjacency` (or a small shared helper factored out of it) rather than reimplement edge-pair counting. If we want a watertight check, `Mesh.is_watertight` is the precedent.
> 2. **Make the warning message actionable.** Instead of "Consider remeshing the input geometry", point at a concrete entry point — e.g. `newton.utils.remesh_mesh(...)`. If we'd like to recommend the SDF-based pipeline, let's promote `remesh_poisson` (and `extract_largest_island`) to public as part of this PR; it's what the advice really means.
> 3. **Keep the genuinely new checks.** The sliver / angle / tet-quality thresholds don't duplicate anything and are the core value add — those stay as-is.
> 
> Optional follow-up (doesn't have to block this PR): consider consolidating these pieces behind a single `validate_mesh`-style entry point or a `MeshQualityReport` that composes topology checks (from `MeshAdjacency` / `is_watertight`) with the new geometric-quality checks, so future users have one obvious place to look.

@adenzler-nvidia  As to other points you mentioned:

- **on-manifold edge **: the `MeshAdjacency` does the check in a pythonic way when adding edges. It's very hard to have a shared manifold-checking method between  `MeshAdjacency`'s python set/list based implementation  and the mesh validator's `np.ndarray` based implementation. So I dropped the validator's redundant non-manifold block for the triangle path, and defer to `MeshAdjacency`'s existing per-edge warning so users still see *something* but we stop running the same scan twice on the builder path.
- **`Mesh.is_watertight`** does not apply here. It checks `counts == 2`, i.e. *no boundary edges and no non-manifold edges*. Cloth meshes legitimately have boundary edges (`counts == 1`); the validator must not flag those or we'd warn on every open garment.

eric-heiden (commented):
Let's wait with merging, I need to confirm some API changes here first.

eric-heiden (dismissed):
Thanks, LGTM! We will defer the introduction of labels for deformables to a follow-up.

eric-heiden (approved):

eric-heiden:
@AnkaChan can you merge the main branch into this PR? Then the docs build should work and we can merge it.

PR #2549, opened 2026-04-23
Add rigid body, static friction ramp test

Alex-Omar-Nvidia:
## Description

This MR adds static friction tests based on rigid bodies (cubes) falling down ramps of varying `mu` and `theta`.

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **Tests**
  * Added a comprehensive rigid-body friction ramp benchmark covering static vs sliding classification, kinetic stopping-distance checks, parameter sweeps, tolerance and NaN/sanity checks, and optional visual/debug runs.
  * Consolidated prior Coulomb-friction checks into this focused benchmark and removed them from the older validation suite.
* **Documentation**
  * Added a guideline recommending one-test-per-phenomenon benchmarks using parameterized analytic predicates applied across solvers.


adenzler-nvidia:
Before we go further, can you help me understand what gap this test fills that isn't already covered?

`newton/tests/test_physics_validation.py:505` (`test_static_friction`) already verifies Coulomb stick/slip: two boxes on a ground plane, one pushed with `F < µmg` (must not slide), one with `F > µmg` (must slide). It runs across `xpbd` / `mujoco_cpu` / `mujoco_warp`.

`newton/tests/test_physics_validation.py:571` (`test_dynamic_friction`) verifies kinetic friction analytically: a box with initial `v₀` must stop at `d = v₀²/(2µg)` within 1%. Same solver matrix.

The physics invariant your test asserts — `θ > atan(µ) ⇒ slides`, `θ < atan(µ) ⇒ stays` — is mathematically equivalent to the force-balance invariant those tests already cover. Projecting gravity onto a tilted surface doesn't exercise a different code path in the solvers; the contact normal and friction cone math is the same.

Concretely, could you answer:

1. **What does this test catch that the existing ones don't?** If a solver regressed on Coulomb friction, would it fail `test_static_friction` before failing this one? If not, what's the failure mode you're protecting against?

2. **Is this really about adding VBD to the friction matrix?** That's genuinely missing coverage — but the cleaner fix is to register VBD in `test_physics_validation.py`'s existing `solvers` dict at `:1619`, not a new 243-line file.

adenzler-nvidia:
Thanks for the feedback — let's align on a path forward.

I agree #2581 isn't the right shape: the VBD overrides on `test_static_friction` / `test_dynamic_friction` (`stop_distance_rel=3.0`, `final_vel_max=2.5 m/s`) are wide enough that the assertions stop testing anything meaningful — a box still moving at 2.5 m/s would pass as "stopped". That's because `μ=0.4` / `μ=0.5` puts AVBD's penalty friction in its saturating regime, which is what your code comment in this PR sidesteps by capping `μ ≤ 0.30`. Let's drop that direction.

Instead, what do you think about positioning **this** test as the canonical friction benchmark for Newton — covering all our solvers, and serving as the standard correctness check anyone (us or external/research solvers) would run when validating friction? The grid sweep over `(μ, θ)` is a much better fit for that role than the single-point tests we have today. Two questions:

1. Does this same framing work for the rest of the VBD tests you're planning from Miles' doc? If we can land each of those as a per-phenomenon canonical benchmark (one for friction, one for restitution, etc.) covering all applicable solvers, that's a much stronger end state than VBD-specific tests in parallel with the existing ones.
2. If the answer is yes, can we document this somewhere discoverable so new devs land here first and any new validation work plugs into the same infrastructure rather than spawning sibling test files?

If the direction makes sense, I'd want to extend this PR a bit before it can replace `test_static_friction` / `test_dynamic_friction`:

1. **Tighten `_DEFAULT_THRESHOLDS` for non-VBD solvers** to roughly `(eps_pos=0.01 m, v_rest=0.01 m/s)`, matching the strictness of the current static-friction check. VBD keeps its widened thresholds and capped `μ ≤ 0.30` with a comment explaining the saturation regime.
2. **Add a quantitative assertion on cells expected to slide**: terminal acceleration `a ≈ g(sinθ − μcosθ)` measured over a late simulation window. This is the same kinetic-friction invariant `test_dynamic_friction` checks (`d_stop = v₀² / (2μg)`), just with a different boundary condition — and it folds into the existing grid for free.

With (1) + (2) in place, the ramp grid covers everything `test_static_friction` and `test_dynamic_friction` cover for XPBD / MuJoCo, plus a much wider μ sweep, plus VBD — so I'd fold the deletion of those two tests into this PR rather than a follow-up. Featherstone / SemiImplicit currently have no friction coverage at all (they're not in `test_physics_validation.py:1619`); a viscous-friction sibling test could fix that in another follow-up.

Looping in @Kenny-Vilella as the author of the tests we'd be retiring, and @mmacklin / @eric-heiden for thoughts on the canonical-benchmark framing.

Alex-Omar-Nvidia:
Thanks — agree on the framing, and yes to both questions. Pushed an update that implements (1) + (2) and folds in the deletion of `test_static_friction` / `test_dynamic_friction`.

There are a few more tests my team was hoping to implement from Miles' doc
* Friction Area Independence
* Capsule Stack and Pyramid Stack (probably in the same test/scene)
* Friction Coefficient Sliding
* Friction Area Independence

For those last 2, I'd appreciate your feedback on if you think those should be unit tests, as I don't want to cause an unnecessary increase in test times.

I set the default thresholds as tightly as I could:

1. **Static thresholds landed at `eps_pos=0.02 m, v_rest=0.10 m/s`** rather than 0.01 / 0.01. The 0.01 v_rest target wasn't reachable on `mujoco_warp` — even with `cone="elliptic"`, `impratio=10`, `iterations=200, ls_iterations=100`, near-critical cells (e.g. μ=0.7 / θ=30°, crit=35°) intermittently show |v| ≈ 0.05–0.08 m/s across runs. This is the same constraint-solver creep `test_static_friction` sidestepped by checking only position. `eps_pos=0.02` keeps a small budget for initial pop-out; `v_rest=0.10` sits above the observed creep band.

2. **Kinetic-friction tolerance is `0.50·|a_expected| + 0.50 m/s²`,** not 1–10%. Starting from rest, MuJoCo's stick-to-slip transition is the dominant error term — at μ=0.7 / θ=50° on `mujoco_warp` I saw `a_measured` swing between ~0.85 and ~4.3 m/s² across runs vs an expected 3.10. `test_dynamic_friction` hits 1% because v₀=2 m/s puts the box well into the kinetic regime; from rest, a constraint solver needs much longer evolution to escape the regularized zone, which the grid can't easily afford (fast cells slide off the ramp). Worth tightening if we ever want a stricter kinetic check — probably needs a per-cell adaptive ramp length / measurement window.

3. **Angle sweep capped at 40°.** θ=50° was flaky on `mujoco_warp` no matter what solver settings I tried — looks like a stick-slip / contact-loss interaction at steep slopes. Practical effect: μ=1.00 only exercises the static side (no above-crit cell), which felt like an acceptable tradeoff vs nondeterminism in CI.

4. **VBD uses a softer `min_slide` displacement check** instead of the kinetic check, even within the μ ≤ 0.30 cap. AVBD's penalty friction still saturates on the borderline cells we'd most want to check (e.g. μ=0.20 / θ=17.5°, ~6° above crit), locking the box near zero velocity. Cap keeps the static cells meaningful; the Coulomb-acceleration invariant just isn't a fair bar for AVBD. The kinetic check stays in place for XPBD / MuJoCo across the full default sweep.

Default sweep is now μ ∈ {0.10, 0.30, 0.50, 0.70, 1.00} × θ ∈ {3°, 10°, 20°, 30°, 40°}. Static side covers the full grid for all four solvers; kinetic covers everything except the VBD rows and the μ=1.00 row.

adenzler-nvidia:
Thanks - this all sounds reasonable, but it looks like the update is not pushed?

Also, I think the google doc links probably aren't meant for public sharing - in the current state they won't be accessible by outsiders, but let's remove the links anyway.

Alex-Omar-Nvidia:
Oops, I pushed the new code and removed the gdocs links. Thanks

eric-heiden:
Recommendation: keep the ramp-grid direction, but make it the canonical friction coverage before merging. The new `test_rigid_friction_ramp.py` covers the same Coulomb threshold behavior that `test_static_friction` exercised and broadens it across `(mu, theta)` values and solver variants, so it is a reasonable replacement for the flat static-friction case once the active assertions are tightened.

I would not treat it as a complete replacement for the removed `test_dynamic_friction` yet, because that test asserted the kinetic stopping-distance oracle `v0^2 / (2 * mu * g)`, which the ramp grid does not currently exercise. I recommend keeping the ramp grid, closing the skipped or weak edge-cell assertions, and either adding a small stopping-distance case to the same canonical friction file or explicitly scoping this PR to static friction only.

Alex-Omar-Nvidia:
Thanks, your concern makes sense. The from-rest acceleration check was visibly weaker than `test_dynamic_friction`'s `v₀² / (2μg)` oracle (±50% relative tolerance, no kinetic cells at μ=1.00, VBD routed around it entirely).

I pushed an update to add `test_friction_stopping_distance` to the canonical friction file and dropped the weak kinetic check from the ramp grid. Roles are now cleanly split:

- **`test_friction_ramp`** — (μ, θ) static-threshold map. Below-crit cells assert at-rest; above-crit cells assert `disp ≥ min_slide` (a sanity bound, not a precision check).
- **`test_friction_stopping_distance`** — kinetic-friction oracle. Three boxes at μ ∈ {0.20, 0.40, 0.70} with `v₀ = 2 m/s` on per-box ground patches with matching μ. Asserts `d_measured ≈ v₀² / (2μg)` per box, plus a `|v_final| < 0.05 m/s` sanity bound.

Same boundary condition as the deleted `test_dynamic_friction` (sliding from `v₀`, not from rest), so it sidesteps the from-rest stick-slip transient that forced the loose tolerance on the acceleration check.

## Measured per-solver tolerances

Ran the scene before locking in the assertion to set tolerances against actual solver output:

| Solver | μ=0.20 | μ=0.40 | μ=0.70 | rel_tol | Headroom |
|---|---|---|---|---|---|
| XPBD | −0.026% | −0.043% | +0.048% | 1% | ~20× |
| MuJoCo Warp | +0.183% | +0.137% | +0.076% | 1% | ~5.5× |
| MuJoCo CPU | −0.034% | −0.031% | +0.028% | 1% | ~33× |
| VBD | +12.99% | +13.66% | +17.69% | 25% | ~1.4× |

XPBD / MuJoCo (CPU + Warp) all hit `test_dynamic_friction`'s 1% bar with substantial headroom — across three μ values rather than the deleted test's single point.

VBD systematically overshoots by 13–18% with positive bias growing in μ. Reproducible across runs, not noise — AVBD's low-velocity friction regularization lets the box coast past the Coulomb stopping point. 25% leaves ~7% margin over the worst observed; still meaningful (catches sign errors, off-by-2× bugs, NaN, lock-up), just acknowledges AVBD isn't a pure Coulomb-cone solver. If you'd rather I drop VBD from this test entirely and leave the ramp grid as VBD's only kinetic coverage, that's a one-line change. Alternatively, I can drop mu=0.7 from the VBD solver test.

## Ramp-grid cleanup

With the kinetic oracle now in its own test, the ramp's old `a_measured` branch was redundant *and* weaker than what it covered. Dropped it: `_Thresholds` collapses from six fields to four (`margin_deg`, `v_rest`, `eps_pos`, `min_slide`), and both `_DEFAULT_THRESHOLDS` and `_VBD_THRESHOLDS` now use the same shape — they differ only in numerical strictness.

## One unrelated fix folded in

The `main` merge picked up #2327 (ALM hard constraints on VBD rigid bodies). Two borderline cells in the VBD ramp sweep (μ=0.20 / θ=17.5°, μ=0.25 / θ=20.0°) now creep ~0.011–0.014 m in 0.25 s instead of sliding past 0.02 m. Lowered VBD's `min_slide` from 0.02 m → 0.005 m so they pass; still asserts "the box moved measurably". Happy to split this into a separate PR if you'd prefer it not ride along.

Kenny-Vilella:
No problem for me to remove the tests.

Alex-Omar-Nvidia:
I wanted to follow up on this PR and see if there is anything else I can update!

adenzler-nvidia:
Sorry, this was just lost in the noise of the release, I'll make another pass today.

adenzler-nvidia (dismissed):
LGTM.

Alex-Omar-Nvidia:
Updates:

There was a regression in the VBD solver's rigid friction.

Changes
* Kept XPBD and MuJoCo on the strict analytic Coulomb stopping-distance check with 1% relative tolerance.
* Added a configurable final-speed threshold for the stopping-distance test instead of using one global STOPPING_V_FINAL_MAX.

* Gave VBD its own looser regression bounds:
  * stopping_distance_rel_tol = 3.0
  * VBD_STOPPING_V_FINAL_MAX = 1.75

* Updated comments/docstrings to say that VBD is using an empirical regression envelope, not the precise Coulomb stopping-distance oracle.

Why:

* The failures were deterministic and isolated to VBD stopping-distance cases on both CPU and CUDA.
* VBD’s penalty-friction model was overshooting the analytic Coulomb distance substantially, and the low-mu case was still decelerating at the measurement point.
* XPBD and MuJoCo still match the analytic expectation closely, so their stricter checks should stay intact.

Alex-Omar-Nvidia:
It looks like there has been a regression the VBD solver. I created https://github.com/newton-physics/newton/issues/2881 to capture this.  I think this warrants a deeper investigation and then we should come back and tighten these bounds in these VBD tests.

AGENTS.md
```diff
@@ -39,6 +39,7 @@ uv run --extra dev --extra torch-cu12 -m newton.tests                  # with Py
 ### Testing guidelines
 
 - Never call `wp.synchronize()` or `wp.synchronize_device()` right before `.numpy()` on a Warp array. This is redundant as `.numpy()` performs a synchronous device-to-host copy that completes all outstanding work.
+- Per-phenomenon canonical benchmarks: one test file per physical phenomenon (friction, restitution, …), parameter-sweep over an analytic predicate, applied to all applicable solvers. See `newton/tests/test_rigid_friction_ramp.py`. Extend an existing one before adding a sibling test.
```
adenzler-nvidia, line 42: Did you verify that this helps? We can leave it in for now, but it seems oddly-specific for an agents.md, and I wonder whether it's not more confusing than it helps. Especially since per-phenomenon benchmarks seems not very well defined.

adenzler-nvidia (dismissed):

AGENTS.md
Alex-Omar-Nvidia: I haven't A/B tested this. I wanted to try to provide some structure to the agent and codify some of the decisions we made about structure, but maybe this isnt the right place to make this call. I can remove this if you want

Alex-Omar-Nvidia: I ended up removing this bit from agents.md

Alex-Omar-Nvidia:
In `newton/tests/test_rigid_friction_ramp.py,` I applied JC's `kd = 0` suggestion specifically to the stopping-distance scene. That made VBD much closer to the analytic Coulomb result: it only missed the strict 1% bound by about 1.06-1.07% on the low-mu case, and no longer needed the loose final-speed threshold.

Based on that, I tightened VBD’s stopping-distance regression bounds from 3.0 relative tolerance and 1.75 m/s final speed down to 0.02 relative tolerance and the shared 0.05 m/s final-speed bound. XPBD and MuJoCo remain on the strict 0.01 analytic tolerance.

I also tried kd = 0 on the ramp-grid scene, but VBD failed one sliding cell by a tiny margin, so I kept kd = 1.0e3 there and scoped kd = 0 to the stopping-distance test only.

Verified:
`uv run --extra dev -m newton.tests -k test_friction_stopping_distance`
`uv run --extra dev -m newton.tests -k test_friction_ramp_vbd`

adenzler-nvidia (approved):

PR #1732, opened 2026-02-23
Margin and Gap rename

nvtw:
## Description

Margin and Gap rename - so far only for rigid bodies (soft bodies and particles keep to use the old terms)

The MR should close https://github.com/newton-physics/newton/issues/1655

## Newton Migration Guide

Please ensure the migration guide for **warp.sim** users is up-to-date with the changes made in this PR.

- [x] The migration guide in ``docs/migration.rst`` is up-to date

## Before your PR is "Ready for review"

- [x] Necessary tests have been added and new examples are tested (see `newton/tests/test_examples.py`)
- [x] Documentation is up-to-date
- [x] Code passes formatting and linting checks with `pre-commit run -a`

## Summary by CodeRabbit

* **Breaking Changes**
  * Public collision config renamed: contact_margin → gap, thickness → margin, rigid_contact_margin → rigid_gap; contact data fields renamed to margin_* and gap-based semantics used for contact detection and resting separation.

* **Documentation**
  * Docs, examples and attribute mappings updated to reflect margin/gap terminology and new defaults.

* **Tests & Examples**
  * Tests and sample code updated to the new public names and semantics.

PR #2553, opened 2026-04-23
Make ShapeConfig strict with slots=True

hujc7:
## Description

`ModelBuilder.ShapeConfig` is a plain `@dataclass`. Python silently accepts writes to undeclared attribute names on such classes, so a downstream `cfg.old_name = value` becomes a dead no-op when `old_name` is renamed upstream.

Seen in practice in [isaac-sim/IsaacLab#5289](https://github.com/isaac-sim/IsaacLab/pull/5289): PR #1732 renamed `ShapeConfig.contact_margin` → `gap`, and Isaac Lab's stale write `builder.default_shape_cfg.contact_margin = 0.01` continued to run for weeks after the pin bump, created a dead attribute, and the intended 1 cm shape gap was never applied.

## Change

Soft-deprecate undeclared writes on `ShapeConfig` (replacing the original `slots=True` hard fail per maintainer feedback).

`ShapeConfig` keeps its `__dict__` storage, but overrides `__setattr__` to emit a `DeprecationWarning` when a write targets a name that is not:

- a declared dataclass field, or
- a property (e.g. the `flags` setter), or
- a private / dunder attribute.

The write still goes through, so any out-of-tree caller that legitimately stashes extra metadata on `ShapeConfig` keeps working unchanged — they just get a loud warning at the write site. Stale writes after a rename (e.g. the IsaacLab `cfg.contact_margin = 0.01` case) become attributable instead of silent.

Scheduled to flip to a hard `AttributeError` in a future release; the CHANGELOG entry under `Deprecated` tracks the migration window.

Scope is intentionally narrow to `ShapeConfig`. Happy to apply the same to `ActuatorEntry`, `CustomAttribute`, `CustomFrequency` in a follow-up.

## Compatibility

Sanity-checked locally:

- Valid writes (`cfg.gap`, `cfg.margin`, `cfg.density`, …) are silent — no warning.
- `cfg.flags = ...` (property setter) is silent — no warning.
- `cfg.contact_margin = 0.01` (the IsaacLab#5289 case) raises a `DeprecationWarning` naming the bad attr and listing the declared fields; the write still lands so today's behavior is byte-identical (just noisy).
- `dataclasses.replace`, `copy.copy`, the existing `ShapeConfig.copy()` method, and `pickle` round-trips all continue to work.

No behavior change for existing valid callers; no API surface added or removed.

## Newton Migration Guide

- [ ] The migration guide in `docs/migration.rst` is up-to-date

(CHANGELOG entry added under `[Unreleased]` → `Deprecated`. Happy to mirror the note into `docs/migration.rst` if maintainers prefer.)

## Before your PR is "Ready for review"

- [x] Necessary tests have been added and new examples are tested (no new tests required; existing `ShapeConfig`-touching tests in `newton/tests/test_mesh_backface.py` continue to pass and the deprecation path is exercised by the in-PR sanity checks)
- [x] Documentation is up-to-date (CHANGELOG `Deprecated` entry added)
- [x] Code passes formatting and linting checks with `pre-commit run -a`

## Summary by CodeRabbit

* **Chores**
  * Writing undeclared attributes on shape/config objects now emits a DeprecationWarning; such writes will become disallowed in a future release.

* **Documentation**
  * CHANGELOG updated under "Deprecated" with a deprecation notice and guidance to migrate stale attribute writes to the canonical declared field names.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2553?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@
 - Increase conveyor rail roughness in `example_basic_conveyor` to reduce mirror-like reflections
 - Migrate all raycast logic to `geometry.raycast`, all raycast functions now return distance and normal information
 
+### Deprecated
+
+- Writing undeclared attributes on `ModelBuilder.ShapeConfig` now emits a `DeprecationWarning` and is scheduled to raise `AttributeError` in a future release. Migrate stale writes (e.g. `cfg.contact_margin = ...` left over from the rename to `gap`) to current field names; see the `ShapeConfig` declared fields for the canonical list.
+
 ### Fixed
 
 - Fix Sphinx docs builds to auto-discover bundled ``pypandoc_binary`` pandoc so notebook tutorials build without manual PATH configuration
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@
 
 ### Deprecated
 
-- Writing undeclared attributes on `ModelBuilder.ShapeConfig` now emits a `DeprecationWarning` and is scheduled to raise `AttributeError` in a future release. Migrate stale writes (e.g. `cfg.contact_margin = ...` left over from the rename to `gap`) to current field names; see the `ShapeConfig` declared fields for the canonical list.
+- Deprecate writing undeclared attributes on `ModelBuilder.ShapeConfig`; attempting to set attributes not in the declared fields emits a `DeprecationWarning` and will raise `AttributeError` in a future release. Migrate stale writes (e.g. `cfg.contact_margin = ...` left over from the rename to `gap`) to current field names; see the `ShapeConfig` declared fields for the canonical list.
 
 ### Fixed
 
--- a/newton/_src/sim/builder.py
+++ b/newton/_src/sim/builder.py
@@ -210,7 +210,7 @@ class ModelBuilder:
         output_indices: list[list[int]]  # Per-actuator output indices
         args: list[dict[str, Any]]  # Per-actuator array params (scalar params in dict key)
 
-    @dataclass
+    @dataclass(slots=True)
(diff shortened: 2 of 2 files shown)


eric-heiden (commented):
Thanks, @hujc7, this is a good contribution! Can you please sign the CLA?

eric-heiden:
Thanks @hujc7. We had some discussion and also think the behavior first requires a deprecation mechanism before we throw an error unconditionally. So better raise a deprecation warning when a user is trying to access an undefined dataclass attribute on `ShapeConfig` (what CodeRabbit also suggests). Do you think you can get this implemented here?

adenzler-nvidia:
@hujc7 are still planning to update this? otherwise we will close it soon.

hujc7:
Sorry I lost track on this earlier. Will make an update.

preist-nvidia:
@adenzler-nvidia can you please help shepherd this into 1.3, or push to 1.4 if too tight?

adenzler-nvidia:
Thanks for following up on this. After looking at it again, I think the original issue was that `contact_margin -> gap` shipped without a targeted deprecation alias.

Since then, we have committed to handling public API changes with proper deprecations and clear migration documentation, so this kind of silent rename should not happen again. Because of that, I do not think we need a generic undeclared-attribute warning on `ShapeConfig`; it feels broader than the issue we are trying to address and changes the behavior policy for this dataclass.

I would prefer to close this PR on that basis. What do you think?

hujc7:
Sounds good. Thanks all for look into this!

PR #2559, opened 2026-04-23
Fix free joint anchor transform handling

### LINKED ISSUE #2558: Unify treatment of free joint with other joints
In Newton currently, the free joint (in particular in `ModelBuilder.add_joint_free()` is receiving special treatment, where `joint_q[0:7]` is initialized directly to the child body's world pose (`body_q[child]`). Every other joint type leaves joint_q at its neutral (identity-like) value and encodes the parent-to-child geometry through the joint's anchor transforms `joint_X_p` and `joint_X_c`, with the body world pose reconstructed by forward kinematics:
```body_q[child] = body_q[parent] * joint_X_p * transform(joint_q) * inv(joint_X_c)```

This special treatment must then be propagated to importers and downstream code. In the common case of a free joint attached to the root body with identity transform, this shortcut is correct, but things start to break otherwise, for instance when creating a free joint programmatically (in particular with non-trivial transforms).

As another example, together with issue #2556 this creates the following inconsistency:
- Model.control() will clone the joint target positions (currently all-zero and in dof space)
- free joints are added with a non-zero `joint_q` to account for the child body's pose
so the default control in neutral pose will not have the joint coords equal to their target (even converting between dofs and coords)

Ideally, we'd be able to remove this silent assumption about the free joint.

I've created a PR with Claude (#2559) to help judge the scope of changes this would imply since I'm not very familiar with main Newton code (to take with a big grain of salt).

### PR DESCRIPTION
**Pull request by Guirec-Maloisel**
## Description

Preserve the existing free-joint coordinate behavior while fixing the anchor-transform cases uncovered by the original PR:

- Initialize `ModelBuilder.add_joint_free()` coordinates from the relative parent/child body poses and `joint_X_p` / `joint_X_c`, so forward kinematics reproduces the authored child pose.
- Compose and decompose free-joint anchor transforms at the `SolverMuJoCo` coordinate boundary so non-identity anchors round-trip correctly.
- Keep the existing user-facing convention that floating-root poses remain in `joint_q` when the parent and joint anchors are identity.

The selection API issue remains tracked separately in #3226. Changing the free-joint force/wrench convention is also out of scope for this coordinate-conversion fix and should be handled as a follow-up.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k TestModelJoints
uv run --extra dev -m newton.tests -k TestMuJoCoSolverQpos0
uv run --extra dev -m newton.tests -k test_floating_base_
uvx pre-commit run -a
```

## Bug fix

**Steps to reproduce:**

1. Create a free joint with a parent body or non-identity parent/child anchor transforms.
2. Initialize it from an authored child body pose or round-trip it through `SolverMuJoCo`.
3. Observe that the old code ignores the transforms when initializing or converting `joint_q`, so forward kinematics or MuJoCo qpos does not reproduce the body pose.

**Minimal reproduction:**

```python
import newton
import warp as wp

parent = builder.add_link(xform=wp.transform(wp.vec3(1.0, 0.0, 0.0)))
child = builder.add_link(xform=wp.transform(wp.vec3(2.0, 0.0, 0.0)))
builder.add_joint_free(
    parent=parent,
    child=child,
    parent_xform=wp.transform(wp.vec3(0.5, 0.0, 0.0)),
)
```


eric-heiden:
Thanks for looking into this @Guirec-Maloisel!
I think generally this approach looks correct - we should indeed make sure the kinematics
```
body_q[child] = body_q[parent] * joint_X_p * transform(joint_q) * inv(joint_X_c)
```
are maintained also for free joints. We had the assumption of free joints being mostly root joints where the case `joint_q == body_q` was correct but yes this assumption breaks if the parent != -1. It is also good to fix the handling of `joint_X_p` and `joint_X_c` for these root joints in selection.py.

I think the problem here in this PR is that we shouldn't move away from assigning `joint_q` for these free joints. Particularly the examples this PR modifies worked fine before because the free joints were root joints, so why complicate things with the assignment to `joint_X_p` and then also assigning `body_q` (which is not needed because `eval_fk` takes care of that)? We should keep using ` joint_q` as the true configuration for free joints for reduced-coordinate solvers at least. I would argue it is more confusing to assign to `joint_X_p` because now `body_q` is updated but users of reduced-coordinate solvers will not see any change in `joint_q`.

Guirec-Maloisel:
@eric-heiden yes your point makes sense to me, I have only skimmed through the AI-proposed changes and missed that part, but I fully agree that resetting the root pose locally should be done by editing `joint_q`, rather than `joint_X_p` (which amounts to changing the model itself rather than its state).

I can try to ask the AI to make edits, but maybe it's better if someone more familiar with the affected code takes charge?

preist-nvidia:
Closes #2558

Guirec-Maloisel:
I was able to pick this up again and update the branch w.r.t. main. I also reverted most of the changes to the examples so they keep assigning to joint_q for resetting the root transformation. So I think this is now ready to get reviewed in details.

jcarius-nv:

Guirec-Maloisel:
Hey, just a reminder that this PR is awaiting a review.

jcarius-nv:
I'll notify the team again, thanks @Guirec-Maloisel

eric-heiden (changes requested):
Hi @Guirec-Maloisel, sorry for the delay! We just had a discussion with @nvlukasz and @mmacklin about this PR and we think it is better to keep free joint coordinates assigned to `joint_q` rather than routing them to `joint_X_p` (which itself would also not be consistent with how we handle other joint types). That means the downstream changes in this PR should also be reverted.

What we should keep from this PR is the conversion in SolverMuJoCo to account for the free joint anchor transform so that we respect the Model's `joint_X_p` and `joint_X_c`.

While reviewing this PR we noticed a bug in how we compute the `joint_q` from the child body's `body_q` transform in `ModelBuilder.add_joint_free()` which currently always assumes the free joint has no parent body when in fact we should be computing the `joint_q` as the relative transform between the parent (if it exists) and child body while taking into account parent + child joint transforms. We also found some issue thanks to your PR in the selection API that should be fixed in a follow-up: https://github.com/newton-physics/newton/issues/3226.

I'm also happy to take over this PR if you like to address these few things?

Guirec-Maloisel:
@eric-heiden thanks for having a look at this, I'm happy to hand this PR over to you, as you seem to have a clearer plan for this.

One additional related thing, that we noticed independently since first looking at this, is that the free joint gets a separate treatment also for joint forces (https://newton-physics.github.io/newton/stable/concepts/conventions.html#spatial-wrench-conventions), where the force is considered to be applied to the child body rather than to the joint's location (would differ for a non-trivial joint frame). Do you think that we could also get rid of that special case with this PR, or do you consider this a separate issue?

eric-heiden:
> One additional related thing, that we noticed independently since first looking at this, is that the free joint gets a separate treatment also for joint forces (https://newton-physics.github.io/newton/stable/concepts/conventions.html#spatial-wrench-conventions), where the force is considered to be applied to the child body rather than to the joint's location (would differ for a non-trivial joint frame). Do you think that we could also get rid of that special case with this PR, or do you consider this a separate issue?

Thanks, @Guirec-Maloisel. I've created a follow-up issue for it, I think it's better to address it separately: https://github.com/newton-physics/newton/issues/3308.

eric-heiden (dismissed):

newton/_src/sim/builder.py
```diff
@@ -4646,8 +4646,11 @@ def add_joint_free(
             custom_attributes=custom_attributes,
         )
         q_start = self.joint_q_start[joint_id]
-        # set the positional dofs to the child body's transform
-        self.joint_q[q_start : q_start + 7] = list(self.body_q[child])
+        # Initialize the coordinates so FK preserves the authored child pose.
+        parent_body_xform = wp.transform_identity() if parent == -1 else self.body_q[parent]
+        parent_anchor_world = parent_body_xform * self.joint_X_p[joint_id]
+        joint_q = wp.transform_inverse(parent_anchor_world) * self.body_q[child] * self.joint_X_c[joint_id]
```
vreutskyy, line 4499: [P1] This relative coordinate is correct, but `ModelBuilder.add_builder(xform=...)` still left-multiplies every FREE `joint_q` by the world offset. That is only valid for a root free joint with identity `joint_X_p`; with a rotated root anchor it composes the offset in the wrong frame, and for a parented free joint it applies the offset twice because the parent body is transformed too. A targeted test on this head makes `eval_fk()` disagree with the copied `body_q`. Please make the copy/replication handling frame-aware and add regressions for a rotated root anchor and a parented FREE joint.

newton/_src/solvers/mujoco/kernels.py
```diff
@@ -623,39 +623,42 @@ def convert_mj_coords_to_warp_kernel(
...
                 qpos[worldid, q_i + 5],
                 qpos[worldid, q_i + 6],
             )
         )
-        joint_q[wq_i + 3] = rot[0]
-        joint_q[wq_i + 4] = rot[1]
-        joint_q[wq_i + 5] = rot[2]
-        joint_q[wq_i + 6] = rot[3]
+        world_xform = wp.transform(world_pos, world_rot)
+        joint_xform = wp.transform_inverse(joint_X_p[joint_id]) * world_xform * joint_X_c[joint_id]
```
vreutskyy, line 800: [P2] This establishes anchor-relative FREE coordinates at the MuJoCo boundary, but `build_ref_q_kernel` still fills FREE `ref_q` directly from the child `body_q`. `_compute_body_poses_at_qref()` then applies `joint_X_p` / `joint_X_c` again, producing the wrong reference pose and therefore incorrect CONNECT-constraint reference anchors across the world or separate articulations. I reproduced the pose mismatch directly on this head. Please make the FREE reference-coordinate construction anchor-aware and add a CONNECT-constraint regression test.

vreutskyy (changes requested):
The new transform algebra is correct, but two existing integration paths still interpret FREE `joint_q` as a world body pose. I verified both failures with targeted `unittest` reproductions run directly against this head. Please update both paths before merge.

newton/_src/sim/builder.py
eric-heiden: Addressed in 05e5ad11. `add_builder(xform=...)` now changes only root FREE coordinates and expresses the world offset in the unchanged root-anchor frame: `joint_X_p^-1 * xform * joint_X_p * joint_q`. Parented FREE `joint_q` values remain unchanged, so their transformed parent and child poses stay consistent. I added regressions for both a rotated root anchor and a parented FREE joint; each verifies the copied `joint_q` and the resulting FK body pose.

newton/_src/solvers/mujoco/kernels.py
eric-heiden: Addressed in 05e5ad11. `build_ref_q_kernel` now copies FREE/DISTANCE reference coordinates directly from `model.joint_q` instead of using the child `body_q`, so `_compute_body_poses_at_qref()` applies `joint_X_p` / `joint_X_c` exactly once. The new CONNECT regression uses two separate FREE articulations with non-identity parent and child frames and verifies `ref_q`, the reconstructed reference body poses, and both constraint anchors.

vreutskyy (approved):
Re-reviewed at `29375d7`. Both requested fixes are correct, their targeted regressions pass locally, and the full CI suite is green.

jcarius-nv:
@Guirec-Maloisel if you're good with the changes and the state of the PR, then please go ahead and merge. From teh comments I take it as Viktor and Eric approving as well.

PR #2560, opened 2026-04-23
Fix SensorRaycast/picking ignoring HFIELD geometry

### LINKED ISSUE #2412: [BUG] `SensorRaycast` and `ray_intersect_geom` ignore HFIELD geometry
### Bug Description

`ray_intersect_geom` has no `GeoType.HFIELD` branch, so `SensorRaycast` silently returns `-1` (no hit) for heightfield shapes.

### Reproduction Script

```python

```

### System Information

_No response_

vreutskyy:
## Description

`ray_intersect_geom` had no `GeoType.HFIELD` branch, so `SensorRaycast` and viewer picking silently reported no hit for heightfield shapes. This PR adds heightfield support to the ray dispatch.

- `utils/heightfield.py` gains `ray_intersect_heightfield_local(hfd, elevations, ro, rd)` — local-space intersection that slab-clips the ray against the heightfield AABB and walks the XY cells with 2D DDA, testing the two triangles per cell with Möller–Trumbore. Co-located with the other local-space heightfield queries (`sample_sdf_heightfield`, `get_triangle_shape_from_heightfield`, ...).
- `geometry/raycast.py` gains a thin `ray_intersect_heightfield` wrapper that maps the ray to local, delegates, and rotates the normal back to world. `ray_intersect_geom` now dispatches `HFIELD` to it alongside the other primitives.
- `raycast_kernel` and `sensor_raycast_kernel` forward the model's heightfield metadata (`shape_heightfield_index`, `heightfield_data`, `heightfield_elevations`) into `ray_intersect_geom`. The launch sites in `picking.py` and `sensor_raycast.py` pass these through — `None` is fine when the model has no heightfields (warp's ternary short-circuits).

Closes #2412

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_sensor_raycast
uv run --extra dev -m newton.tests -k test_heightfield
```

New `test_ray_intersect_heightfield_via_geom` (5 subcases: flat hit, sloped-face hit, outside-extent miss, parallel-above miss, translated hit) runs on CPU and CUDA. All 22 raycast / 22 sensor_raycast / 35 heightfield / 69 viewer tests pass locally.

## Bug fix

**Steps to reproduce:**

1. Build a Model containing a `Heightfield` shape via `builder.add_shape_heightfield(...)`.
2. Point a `SensorRaycast` at it (or attempt viewer picking).
3. Observe the returned depth image is all `-1.0` (no hits) even though rays clearly cross the heightfield surface.

**Minimal reproduction:**

```python
import numpy as np
import newton
from newton.sensors import SensorRaycast

hfield = newton.Heightfield(
    data=np.zeros((4, 4), dtype=np.float32),
    nrow=4, ncol=4, hx=2.0, hy=2.0, min_z=0.0, max_z=1.0,
)
builder.add_shape_heightfield(heightfield=hfield)
model = builder.finalize()
state = model.state()

sensor = SensorRaycast(
    model,
    camera_position=(0.0, 0.0, 5.0),
    camera_direction=(0.0, 0.0, -1.0),
    camera_up=(0.0, 1.0, 0.0),
    fov_radians=1.0,
    width=8,
    height=8,
)
sensor.update(state)
print(sensor.get_depth_image_numpy())  # previously all -1; now reports real distances
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Raycasting and viewer picking behavior corrected for heightfields; heightfield geometry is properly excluded where appropriate and per-instance scale is now applied during collision and raycasts.

* **New Features**
  * World- and local-space heightfield ray intersection added, producing accurate hit distances and normals.

* **Tests**
  * New tests for heightfield intersections, normals, scaling, and traversal behavior.

* **Documentation**
  * Changelog updated with the fixes.


camevor (commented):
Thanks, this looks good to me! One thing worth flagging is that `ModelBuilder.add_height_field` accepts a scale, but raycast does not apply it. Since collision kernels also don't apply it, I'd recommend either:
* implementing scaling for the raycast, but warn when scale is not 1, or
* documenting the lack of scaling support in height fields.

vreutskyy:
Thanks for the review, @camevor\!

I went with your second suggestion (document the lack of scaling) but formalized it as a deprecation since the current behavior of `scale` on `add_shape_heightfield` is internally inconsistent:

- broad-phase collision AABB ([`builder.py:9736-9743`](https://github.com/newton-physics/newton/blob/6245747e/newton/_src/sim/builder.py#L9736-L9743)) **does** apply `scale` (`shape_src.hx * shape_scale[0]`, etc.)
- narrow-phase collision and the new raycast path read extents from `HeightfieldData` directly and ignore `shape_scale`

Fixing only the raycast would leave it inconsistent with the narrow-phase, so in commit 6245747e I:

1. Kept `scale` being forwarded to `add_shape()` so the AABB path is unchanged for this release (no silent breakage for anyone relying on it).
2. Emit a `DeprecationWarning` when a caller passes `scale`, pointing them at `Heightfield.hx` / `hy` / `min_z` / `max_z`.
3. Added a `Deprecated` entry in `CHANGELOG.md`.
4. Docstring now spells out the inconsistency and the planned removal.

The parameter (and the residual AABB scaling) can be removed in a follow-up release once the warning has been live.

vreutskyy:
Update, @camevor: after your Slack nudge about getting wider buy-in before deprecating, I took another pass at the code and found the existing state was more mixed than "collision kernels don't apply it" -- see the audit below -- so I pivoted to implementing ``scale`` properly instead (pushed in 9c20dd5b + 0bab8c07).

Audit of pre-existing ``scale`` handling for HFIELD:

| Path | Reads from | Applies ``shape_scale``? |
|---|---|---|
| Broad-phase AABB (``builder.py`` local AABB precompute) | ``shape_src`` | ✓ |
| Narrow-phase triangle path (``get_triangle_shape_from_heightfield``) | ``HeightfieldData`` only | ✗ |
| Narrow-phase edge path (``get_edge_from_heightfield``) | ``HeightfieldData`` + ``wp.cw_mul(p, mesh_scale)`` | ✓ |
| Narrow-phase SDF-side for HFIELD | ``HeightfieldData``, ``sdf_scale = (1,1,1)`` forced | ✗ |

The new approach:

1. **Bake per-instance ``scale`` into the ``HeightfieldData`` extents** (``hx``, ``hy``, ``min_z``, ``max_z``) at ``ModelBuilder.finalize()``. Every kernel that reads ``HeightfieldData`` now sees scaled extents consistently. Pure multiplication, no ``abs`` / ``min-max``: a ``Heightfield`` with ``min_z > max_z`` already represents an inverted surface and that's exactly what a negative z scale should produce.
2. **Drop the ``wp.cw_mul(p, mesh_scale)`` in ``get_edge_from_heightfield``** since ``HeightfieldData`` now already carries the scaled vertices -- the multiplication would be a double-scale.
3. **Reverted the deprecation** (no ``DeprecationWarning``, no CHANGELOG ``Deprecated`` entry); replaced with a ``Fixed`` entry.

New regression coverage:

- ``test_ray_intersect_heightfield_scaled`` in ``test_raycast.py`` -- drives ``ModelBuilder -> raycast_kernel`` and asserts a hit at XY=(1.5, 0) that would be outside the unscaled extent.
- ``test_heightfield_native_collision_scaled`` in ``test_heightfield.py`` -- same XY=(1.5, 0) scenario through ``CollisionPipeline`` to cover the narrow-phase side.

Full ``test_raycast`` / ``test_heightfield`` / ``test_mesh_backface`` / ``test_mesh_aabb`` suites pass locally.

newton/_src/geometry/raycast.py
```diff
@@ -822,7 +870,24 @@ def raycast_kernel(
     else:
         mesh_id = wp.uint64(0)
 
-    t, _normal = ray_intersect_geom(geom_to_world, geom_size[shape_idx], geomtype, ray_origin, ray_direction, mesh_id)
+    # Resolve per-shape heightfield metadata (zero-initialised dummy for non-HFIELD shapes).
```
eric-heiden: 🟡 This adds measurable overhead to every `(pixel, shape)` raycast thread even when the scene has no HFIELD shapes. On non-HFIELD scenes, each thread now zero-initializes `HeightfieldData`, checks `shape_heightfield_index`, and passes the extra heightfield state through `ray_intersect_geom`, but the HFIELD branch can never be used.

Re-measured at PR HEAD `d42a597e` against `origin/main` `bc5df8c1` on CUDA L40 with 512x512 rays and identical hit counts:

| Scene without HFIELD | Main mean | PR mean | Delta |
|---|---:|---:|---:|
| primitives | 0.130 ms | 0.173 ms | +33% |
| spheres_4 | 0.085 ms | 0.095 ms | +12% |
| spheres_16 | 0.133 ms | 0.188 ms | +41% |
| spheres_64 | 0.355 ms | 0.578 ms | +63% |

The slowdown scales with shape count, so this looks like per-shape thread overhead rather than a fixed launch cost. Please gate the heightfield path for scenes with no heightfields, for example by compiling separate kernel variants with `wp.static` or otherwise avoiding the dummy `HeightfieldData` setup and extra parameters when `shape_heightfield_index` is empty.

newton/_src/geometry/sdf_contact.py
```diff
@@ -450,7 +450,6 @@ def get_edge_from_mesh(
 def get_edge_from_heightfield(
```
eric-heiden, line 389: 🔴 `HeightfieldData.hx`, `hy`, `min_z`, and `max_z` are now scale-baked before they reach the heightfield helpers, and this change updates `get_edge_from_heightfield` to match that contract. The sibling helper `get_triangle_from_heightfield` still takes `mesh_scale` and applies `wp.cw_mul` to vertices that are already based on the baked extents, so any future caller following the old pattern will silently apply shape scale twice.

Please either remove `get_triangle_from_heightfield` if it is unused, or update it to the same scale-baked contract by dropping `mesh_scale` and the final `wp.cw_mul` block. The docstring should also be updated so it no longer points readers at the mesh scaling model.

vreutskyy:
Thanks for the careful review and the perf measurements, @eric-heiden — pushed e665b483 to address both points:

**Triangle-helper inconsistency:** ``get_triangle_from_heightfield`` is now removed (it had no callers in the repo). ``get_edge_from_heightfield``'s contract (vertices come out of the scale-baked ``HeightfieldData`` already) is now the only one, so future readers can't follow a stale model.

**Per-thread perf overhead:** mirrored the ``_create_sdf_contact_funcs(enable_heightfields)`` pattern — ``_make_raycast_funcs(enable_heightfields)`` emits ``ray_intersect_geom`` + the two raycast kernels with the HFIELD branch, ``ray_intersect_heightfield`` callee, and per-shape lookup all gated by ``wp.static``. Two pre-compiled variants (with/without HFIELD) live at module scope. ``Model.has_heightfields`` (set by the builder) tells the launch sites in ``picking.py`` and ``sensor_raycast.py`` which pair to launch.

I also restructured the HFIELD lookup itself: ``ray_intersect_geom`` now does the ``shape_heightfield_index[shape_idx]`` + ``heightfield_data[hf_idx]`` reads internally (gated by ``wp.static``), so the kernel body has zero per-thread HFIELD setup -- no dummy ``HeightfieldData`` zero-init, no extra branches in non-HFIELD threads. The arrays are still kernel parameters (parallel to ``shape_source_ptr`` for meshes), but with ``wp.static(False)`` every read of them is stripped from the no-HFIELD variant's compile unit.

Would you mind re-running your benchmarks against ``e665b483``? I'd like to confirm the regression's gone end-to-end on your hardware before considering this resolved.

---

A larger architectural point your review surfaced: the only reason this perf gating exists is that HFIELD storage doesn't yet match the mesh model -- meshes hand a single ``wp.Mesh`` handle through ``shape_source_ptr`` and dispatch through ``wp.mesh_query_ray``, while HFIELD currently needs three side tables (``shape_heightfield_index`` + ``heightfield_data`` + ``heightfield_elevations``) because warp doesn't have a first-class ``wp.Heightfield`` type. We sketched a follow-up that would back ``Heightfield`` with ``wp.Texture2D`` + ``shape_source_ptr`` + ``shape_size`` and eliminate the special casing entirely (and the perf gating with it):

#2624

It's a substantial refactor (touches every HFIELD consumer across raycast, narrow-phase SDF contact, AABB build, midphase) so it's well outside the scope of this PR. But filed so we don't lose the design and so this conversation has a tracked next step.

eric-heiden (approved):
Thanks @vreutskyy, reran the benchmark and can confirm the performance regression is resolved now!

PR #2561, opened 2026-04-23
Backfill actuator_biasprm to eliminate G1 precision diff

### LINKED ISSUE #2495: Small actuator_biasprm precision diff on Unitree G1 MJCF
Newton and native MuJoCo produce slightly different `actuator_biasprm` values for the Unitree G1 MJCF, with a max diff of ~1.7e-5. Likely a float precision issue in the import path (MJCF parses as float64, Newton may cast to float32 at different points).

## Impact

- Model comparison requires skipping `actuator_biasprm`
- Dynamics test works but needs slightly looser tolerance (1e-4 instead of 1e-6)

## Context

Discovered during Epic #1401 menagerie testing. Test class uses `model_skip_fields = DEFAULT_MODEL_SKIP_FIELDS | {"actuator_biasprm"}` as a workaround.

vreutskyy:
## Summary

`TestMenagerie_UnitreeG1.test_model_comparison` previously skipped `actuator_biasprm` because Newton and native differ by `~2e-5` on G1's `<position kp="500" dampratio="1"/>` actuators.

Root cause is the same class as other inertia-derived backfills: MuJoCo's `mj_setConst` encodes `-kd = -2·dampratio·sqrt(kp·M_eff)` in `biasprm[2]`, where `M_eff` is a joint-space inertia diagonal entry. Newton and native compute slightly different `M_eff` because of inertia re-diagonalization — the same reason `body_invweight0`, `dof_invweight0`, and `actuator_acc0` are already backfilled. The `~1e-6` `M_eff` diff propagates through the square root into a `~2e-5` `actuator_biasprm` diff. MuJoCo itself does the math correctly on both sides; only the inertia input differs.

Fix: add `actuator_biasprm` to both `DEFAULT_MODEL_SKIP_FIELDS` (strict compare skips it) and `MODEL_BACKFILL_FIELDS` (copied from native before dynamics stepping). Remove G1's `model_skip_fields` override.

## Test plan

- [x] G1 `actuator_biasprm` diff drops from `2.29e-5` → exactly `0.0` after backfill
- [x] `uv run python -m unittest discover -s newton/tests -p "test_menagerie_mujoco.py"` — 48/48 pass, 50 skipped

Fixes #2495

## Summary by CodeRabbit

* **Bug Fixes**
  * UnitreeG1 now imports actuator bias parameters from MuJoCo so actuator settings are fully preserved during model initialization; validation now allows backfilled actuator values.

* **Documentation**
  * Clarified actuator parameter behavior for position actuators, noting expected differences driven by joint-space inertia representation and damping ratio.

### DIFF
--- a/newton/tests/test_menagerie_mujoco.py
+++ b/newton/tests/test_menagerie_mujoco.py
@@ -349,6 +349,10 @@ DEFAULT_MODEL_SKIP_FIELDS: set[str] = {
     # Computed from mass matrix and actuator moment at qpos0; differs due to inertia
     # re-diagonalization. Backfilled instead.
     "actuator_acc0",
+    # Position actuators with `dampratio` encode -kd = -2*dampratio*sqrt(kp*M_eff) in
+    # biasprm[2]; M_eff is joint-space inertia which differs when inertia representation
+    # differs. Backfilled instead.
+    "actuator_biasprm",
     "actuator_lengthrange",  # Derived from joint ranges, computed by set_length_range
     "stat",  # meaninertia derived from invweight0
     # Meshes: Newton / trimesh may create a different number of meshes (nmesh differs),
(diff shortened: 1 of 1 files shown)


vreutskyy:
Good catch — fixed in bf69eb27. G1 was defaulting to `backfill_model = False`, which short-circuited `_backfill_and_recompute()` and made the `MODEL_BACKFILL_FIELDS` addition a no-op for G1's `test_dynamics`. Flipping it to `True` (matching H1) now copies `actuator_biasprm` from native as intended. G1 tests still pass locally.

@coderabbitai resolve

adenzler-nvidia (approved):

PR #2566, opened 2026-04-23
#2431 Adding single frame forward stepping when paused bounded to '.' key

### LINKED ISSUE #2431: [REQ] Add single step option to viewers
### Description

When pausing the simulation it would be good for debugging to be able to step the simulation forward a single frame. In ViewerGL this could be a keypress.

### Motivation / Use Case

Useful for debugging, this is available in almost all existing simulators.

### Alternatives Considered

_No response_

aneangel:
## Description

Closes #2431, when simulation is paused using the '.' key it now allows forward stepping one frame. One note is that to custom simulations loops 

## Checklist

- [ x] New or existing tests cover these changes
- [x ] The documentation is up to date with these changes
- [x ] `CHANGELOG.md` has been updated (if user-facing change)

## Test plan

uv run --extra dev -m newton.tests -k test_basic as well as going into various newton examples and stepping through single simulation frames while paused

we also created test_viewer_controls.py where we intentionally the potential bug described in the feedback to ensure the behavior is as intended

## New feature / API change

```python
import newton
import warp as wp

# Works automatically in the example runner.
# For a custom simulation loop, consume the step request yourself:

viewer = newton.viewer.ViewerGL(model)

while viewer.is_running():
    # Advance when running normally, or one frame at a time when paused
    # (press '.')
    if not viewer.is_paused() or viewer.pop_step_request():
        state_0.clear_forces()
        solver.step(state_0, state_1, control, contacts, dt)

    viewer.render(model, state_0)

```

## Summary by CodeRabbit

* **New Features**
  * Frame-by-frame stepping: press `.` while paused or use the new "Step" button to advance one simulation frame.
  * Run controls grouped together; Reset and controls moved into the Run controls area (Reset shown when available).
* **Bug Fixes**
  * Prevent stale single-step from firing when switching from running to paused.
  * Disable the Step button while the simulation is running.
* **Documentation**
  * Viewer docs and keyboard-shortcut reference updated for per-frame stepping.
* **Tests**
  * Added unit tests covering stepping behavior and edge cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,12 +30,11 @@
 - Honor `UsdGeomImageable` visibility (including inherited `invisible`) on USD prims imported via `ModelBuilder.add_usd()`; visual shapes, gaussian splats, and collider shapes are imported with `ShapeFlags.VISIBLE` cleared when the prim is effectively invisible, while collision behavior is preserved
 - Add more solver options to implicit MPM: `gs-soa` (or `gauss-seidel-soa`) for improved memory coalescing, `gs-batched` (or `gauss-seidel-batched`) merging GS colors with Jacobi-style mass-split parallelism, plus `cr` (Conjugate Residual) and `gmres` linear solver options.
 - Add frame-by-frame step support to `ViewerGL`: press `.` while paused to advance one simulation frame
-- Add `ViewerBase.should_step()` — call once per frame to determine whether the simulation loop should advance; returns `True` when running and consumes a pending single-step request when paused, replacing the `is_paused()` / `pop_step_request()` composition
+- Add `ViewerBase.should_step()` — call once per frame to determine whether the simulation loop should advance; returns `True` when running and consumes a pending single-step request when paused.
 
 
 ### Changed
 
-- Replace `ViewerBase.pop_step_request()` with `ViewerBase.should_step()`; simulation loops should change from `if not viewer.is_paused() or viewer.pop_step_request():` to `if viewer.should_step():`
 - Use pre-computed local AABB for `CONVEX_MESH` shapes in `compute_shape_aabbs`, avoiding a per-frame support-function AABB computation
 - Build mesh SDFs via the texture-based sparse path only; sample via `SDF.texture_data` instead of `SDF.sparse_volume` / `SDF.coarse_volume`.
 - Change implicit MPM default `solver` from `"gs"` to `"auto"`, which selects `"gs"` for trilinear bases and `"gs-batched"` for higher-order ones. Set `solver="gs"` explicitly to restore the previous behavior.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,10 +30,12 @@
 - Honor `UsdGeomImageable` visibility (including inherited `invisible`) on USD prims imported via `ModelBuilder.add_usd()`; visual shapes, gaussian splats, and collider shapes are imported with `ShapeFlags.VISIBLE` cleared when the prim is effectively invisible, while collision behavior is preserved
 - Add more solver options to implicit MPM: `gs-soa` (or `gauss-seidel-soa`) for improved memory coalescing, `gs-batched` (or `gauss-seidel-batched`) merging GS colors with Jacobi-style mass-split parallelism, plus `cr` (Conjugate Residual) and `gmres` linear solver options.
 - Add frame-by-frame step support to `ViewerGL`: press `.` while paused to advance one simulation frame
+- Add `ViewerBase.should_step()` — call once per frame to determine whether the simulation loop should advance; returns `True` when running and consumes a pending single-step request when paused, replacing the `is_paused()` / `pop_step_request()` composition
 
 
 ### Changed
 
+- Replace `ViewerBase.pop_step_request()` with `ViewerBase.should_step()`; simulation loops should change from `if not viewer.is_paused() or viewer.pop_step_request():` to `if viewer.should_step():`
 - Use pre-computed local AABB for `CONVEX_MESH` shapes in `compute_shape_aabbs`, avoiding a per-frame support-function AABB computation
 - Build mesh SDFs via the texture-based sparse path only; sample via `SDF.texture_data` instead of `SDF.sparse_volume` / `SDF.coarse_volume`.
 - Change implicit MPM default `solver` from `"gs"` to `"auto"`, which selects `"gs"` for trilinear bases and `"gs-batched"` for higher-order ones. Set `solver="gs"` explicitly to restore the previous behavior.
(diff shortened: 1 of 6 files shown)


adenzler-nvidia:
Thanks for picking this up — nice debugging win. One API-shape suggestion before it locks in as public surface. @eric-heiden would appreciate your take on the API question too.

`_step_requested` is only latched while paused, only meaningful while paused, and every caller writes the same composition: `if not viewer.is_paused() or viewer.pop_step_request(): step()`. That line appears verbatim in the example runner and the custom-loop snippet, and will be copy-pasted downstream. I'd fold it into the API:

```python
# ViewerBase
def should_step(self) -> bool:
    """Return True if the loop should advance one step this frame.

    Consumes a pending single-step request, so call exactly once per frame.
    """
    return not self.is_paused()

# ViewerGL
@override
def should_step(self) -> bool:
    if not self._paused:
        self._step_requested = False  # clear stale request
        return True
    if self._step_requested:
        self._step_requested = False
        return True
    return False
```

Call site collapses to `if viewer.should_step(): example.step()`.

Why I think it's worth the churn:

1. **Removes a real bug.** If a user presses `.` then `SPACE` before the loop consumes the flag, `_step_requested` persists and fires a spurious step next time they pause (CodeRabbit flagged this). `should_step()` clears the flag on any non-paused frame, so the bug is structurally impossible — no manual cleanup in pause-toggle sites.
2. **Smaller surface.** One method, no `pop_step_request()` override in `ViewerNull`/`ViewerViser` needed — the base default falls out of `is_paused()`.
3. **Foot-guns go away.** `pop_step_request()` breaks silently if you forget `or is_paused()` or call it twice. `should_step()` has neither.

If someone needs "is a step pending?" later for a HUD indicator, add a non-destructive `is_step_pending` property then — YAGNI for now.

Small drive-bys:

- A unit test would be good (new public API, PR test checkbox is unchecked). `ViewerBase.should_step()` defaults to `True`; `ViewerGL` returns `True` once after latching, `False` after.
- The imgui "Step" button is fully clickable when running, but the handler's `and self._paused` silently drops the click. `imgui.begin_disabled(not self._paused)` / `end_disabled()` greys it out. Orthogonal to the API change.
- `.` is fine — matches mpv/VLC/YouTube. Optional: also bind `Right Arrow` for parity with MuJoCo `simulate`, dm_control, Blender.

Would love your thoughts — happy to iterate if any of this doesn't hold up.

preist-nvidia:
Thank you @aneangel ! 

As a follow-up to this, we should move the reset next to the step button, so all run controls are grouped together: https://github.com/newton-physics/newton/issues/2571

adenzler-nvidia:
wouldn't it make sense to address the grouping in this PR as well, as this PR introduces the step button?

preist-nvidia:
Sure, but it's up to @aneangel of course.

aneangel:
I would not mind addressing the grouping in this PR as well if needed, is there already an open issue on github for that as well or was that in reference to #2571?

@adenzler-nvidia thank you for the feedback on the implementation, I will address it accordingly.

preist-nvidia:
Thanks @aneangel - yeah #2571 is the follow up with the grouping that you could address here as well.

aneangel:
@adenzler-nvidia I see that you mentioned potentially adding the binding to the right arrow key, I noticed that the arrow keys are also bounded to the camera movement through alias checking for e.g W or UP, 

I am assuming that this behavior was intentional so to allow users to use the right arrow key to step forward one frame would we want to have them use key combo i.e "Ctrl + Right Arrow Key" or is there a better suggestion on how we should enable familiarity with MuJoCo and Blender that allow stepping with the right arrow key

aneangel:
I added addressing the grouping issue in this PR as well, I didn't create a unit test because I didn't know if those are needed for UI components.

I left out binding to the RIGHT key until further specification on the direction the implementation should take.

adenzler-nvidia:
Thanks, I forgot about the existing mapping for the arrow key, makes sense to drop it then.

adenzler-nvidia (commented):
Thanks for the update — overall this looks good to me. Two small items before merge:

**1. CHANGELOG references a `pop_step_request()` API that never landed on `main`.**

The "Changed" entry implies users have a migration to do, but `pop_step_request()` only existed in an intermediate iteration of this PR (it was renamed to `should_step()` in `fdd92d48a` before any release). I'd suggest:

- Drop the "Changed" entry entirely.
- Reword the "Added" entry to remove the trailing "replacing the `is_paused()` / `pop_step_request()` composition" clause. Something like:
  > Add `ViewerBase.should_step()` — call once per frame to determine whether the simulation loop should advance; returns `True` when running and consumes a pending single-step request when paused.

**2. Base-class `should_step()` docstring overpromises.**

In `newton/_src/viewer/viewer.py`, the docstring says "Consumes a pending single-step request, so call exactly once per frame." — but the base implementation has nothing to consume, and `ViewerNull` / `ViewerRerun` / etc. inherit the no-op default. Could you tighten the base-class docstring to describe the contract (subclasses *may* consume a pending request) and keep the stronger wording on the `ViewerGL` override?

Deferring to @eric-heiden for the API sign-off.

aneangel:
@adenzler-nvidia thanks for the feedback, the necessary changes have been made. 

@eric-heiden please let me know if you see any other issues to address on this PR or if I am missing something needed to be completed from the PR standpoint so that it can be signed-off.

devshahofficial:
@aneangel I think this is close. I’d handle @adenzler-nvidia’s two comments by removing the `pop_step_request` migration wording from the changelog and making the base `should_step()` docstring describe the generic contract rather than promise request consumption. The GL subclass can keep the stronger single-step behavior. I can help with that cleanup if useful.

aneangel:
Hi @devshahofficial I appreciate the feedback, but I am not sure which commit you are viewing as there have been multiple. On my end the changelog has been addressed and the docstring no longer promises request consumption; however after looking at it again I have reworded it to make it clearer to avoid confusion, thanks for pointing that out.

docs/guide/visualization.rst
```diff
@@ -21,6 +21,7 @@ All viewer backends inherit from :class:`~newton.viewer.ViewerBase` and share a
 - :meth:`~newton.viewer.ViewerBase.end_frame` — finish the frame and present it
 - :meth:`~newton.viewer.ViewerBase.is_running` — check whether the viewer is still open (useful as a loop condition)
 - :meth:`~newton.viewer.ViewerBase.is_paused` — check whether the simulation is paused (toggled with ``SPACE`` in :class:`~newton.viewer.ViewerGL`)
+- :meth:`~newton.viewer.ViewerBase.should_step` — call exactly once per frame; returns ``True`` when running, or ``True`` once after a single-step request (triggered with ``.`` or the "Step" button in :class:`~newton.viewer.ViewerGL`) and ``False`` otherwise; prefer this over composing ``is_paused()`` manually
```
adenzler-nvidia, line 24: Since this documents the new `.` shortcut, could you also add it to the ViewerGL controls table below? That table currently lists `SPACE` but not `.`.

CHANGELOG.md
```diff
@@ -36,10 +39,13 @@
 - Migrate all raycast logic to `geometry.raycast`, all raycast functions now return distance and normal information
 - Disable process reuse in the test runner on multi-GPU systems to prevent CUDA errors from cascading across test suites, keeping process reuse enabled on single-GPU systems for faster throughput
 - Default `python -m newton.examples` with no argument to launch `basic_pendulum`; use `--list` to print available examples
+- Adjusted the grouping of `reset`, `step`, and `pause` so that they are all grouped together
 
 ### Fixed
 
 - Fix `remesh_convex_hull` raising `QhullError` on degenerate (coincident, collinear, or coplanar) point clouds; it now returns a zero-volume fallback mesh with a `UserWarning`, raises `ValueError` on empty input, and retries Qhull with `QJ` joggle as a last resort on the 3D path
+- Fix `ViewerGL` spurious single-step firing when `.` is pressed while the simulation is running and is subsequently paused; `should_step()` clears the stale request on any non-paused frame
```
adenzler-nvidia: I think this entry can be dropped. This stale-request bug only existed during this PR iteration; `.` stepping has not shipped on `main`, so there is no user-visible regression to call out in the changelog. The `Added` entry for frame-by-frame stepping covers the release note.

aneangel:
@adenzler-nvidia I dropped the entry, and also re-included the control in the table, I must have mistakenly removed it resolving a merge conflict I had earlier on this branch.

adenzler-nvidia (approved):

PR #2227, opened 2026-03-25
Document registered custom attributes and Mjc USD schemas for SolverMuJoCo

adenzler-nvidia:
## Description

Document which custom attributes, custom frequencies, and `mjc:` USD attributes are supported by `SolverMuJoCo`.

Closes #2080

**Docstring** (`solver_mujoco.py`): Expand the `register_custom_attributes` docstring to catalog all 5 custom frequencies and ~60 custom attributes grouped by category (geom, joint/DOF, body, equality constraints, solver options, pairs, actuators, tendons).

**Integration docs** (`mujoco.rst`): Add a "Supported `mjc:` USD Attributes" section covering both code paths that consume `mjc:` attributes:
- Attributes resolved through `SchemaResolverMjc` (bindings defined in `newton/_src/usd/schemas.py`).
- Custom attributes registered via `register_custom_attributes` (bindings defined on the solver).

Tables are grouped by prim type (Scene, Joint, Shape, Material, Body, Actuator, Equality, Tendon). Newton-specific `mjc:` extensions (14 attributes not present in the mjcPhysics schema) are marked `:sup:`ext``; cross-prim-type reuse is marked `:sup:`reuse``.

**Coverage tests** (`test_mujoco_solver.py`):
- `TestRegisterCustomAttributesDocstringCoverage` — every registered custom attribute and frequency name must appear in the docstring.
- `TestMujocoRstCoverage` — every registered custom attribute and every `mjc:` `SchemaResolverMjc` attribute must appear in `mujoco.rst`, so the documentation cannot drift out of sync.

## Checklist

- [x] New or existing tests cover these changes
- [x] The documentation is up to date with these changes
- [x] `CHANGELOG.md` has been updated (if user-facing change)

## Test plan

```
uv run --extra dev -m newton.tests -k test_register_custom_attributes_docstring_coverage
uv run --extra dev -m newton.tests -k test_custom_attributes_in_rst
uv run --extra dev -m newton.tests -k test_schema_resolver_in_rst
cd docs && make html
```

PR #2574, opened 2026-04-24
Rewrite MuJoCo integration docs

### LINKED ISSUE #2080: [DOCS] Document which schemas, attributes are supported in SolverMuJoCo
### Category

Missing documentation

### Description

We should document which custom attributes and frequencies are registered.
This should happen in the docstring for `SolverMuJoCo.register_custom_attributes` and ideally we refer to this function once more in https://newton-physics.github.io/newton/latest/integrations/mujoco.html.
Then in this document we also define which Mjc USD schemas we support.

adenzler-nvidia:
## Description

Rewrite `docs/integrations/mujoco.rst` around Newton ↔ SolverMuJoCo API compatibility and mapping. The previous draft in #2227 had grown to ~1,300 lines, largely duplicating the `SolverMuJoCo.register_custom_attributes()` docstring and tying the file to strict RST-vs-code coverage tests.

The new layout covers:

- **Concept-mapping tables per family** (joints, geoms, actuators, equality, tendons/pairs) with unsupported rows folded into the tables themselves (no separate "what's supported / what's not" list).
- **Runtime state synchronisation** — a new section narrating the push / integrate / pull cycle inside `step()`, with explicit Newton-field ↔ `mjData`-field references (`joint_q ↔ qpos`, `joint_target_pos/vel → ctrl`, `joint_f → qfrc_applied`, `body_f → xfrc_applied`, …).
- **Custom attributes** — a compact section that explains the `mujoco` namespace, the `mjc:*` USD schema, and `SchemaResolverMjc`, and defers to `SolverMuJoCo.register_custom_attributes` as the single source of truth for the attribute list.
- **Unsupported MuJoCo features** — a section listing what does not round-trip from MuJoCo into Newton (sensors, keyframes, composite/flex, plugins, user data, …).
- **Where the conversion lives** — a final code-anchors section pointing at `_convert_to_mjc`, `_init_pairs`/`_init_actuators`/`_init_tendons`, `_apply_mjc_control` / `_update_newton_state`, `kernels.py`, and `SchemaResolverMjc`, with a note that underscore-prefixed symbols are internal.

Closes the conversation around #2227; that PR can be closed in favour of this one.

## Checklist

- [x] New or existing tests cover these changes — docs-only change; no test changes.
- [x] The documentation is up to date with these changes.
- [ ] `CHANGELOG.md` has been updated (if user-facing change) — docs-only, no CHANGELOG entry.

## Test plan

- `uvx pre-commit run --files docs/integrations/mujoco.rst`
- Sphinx build: `cd docs && uv run --extra docs sphinx-build -a -b html . _build/html` — clean, no warnings pointing at `mujoco.rst`.
- Manually reviewed the rendered HTML at `docs/_build/html/integrations/mujoco.html` for layout, cross-refs, and code-block formatting.

## Summary by CodeRabbit

* **Documentation**
  * Reworked MuJoCo integration into a structured solver overview with version pinning, field-name conventions, and expanded unsupported-feature list (e.g., DISTANCE/CABLE, CONE/GAUSSIAN).
  * Added collision pipeline, tendons, contact-pair overrides, custom-attribute/USD mappings, multi-world validations, and a three-phase state sync (contacts require explicit update).
  * Replaced solver-options table with precedence rules and clarified integrator default.
  * Clarified heightfields are always static (world-body sentinel).


docs/integrations/mujoco.rst
```diff
@@ -6,21 +6,44 @@
...
+    import newton
+    from newton.solvers import SolverMuJoCo
+
+    builder = newton.ModelBuilder()
+    SolverMuJoCo.register_custom_attributes(builder)
+    builder.add_mjcf("robot.xml")        # or build the model directly
+    model = builder.finalize()
+    solver = SolverMuJoCo(model)
+    # ...then call solver.step(state_in, state_out, control, contacts, dt)
+
```
preist-nvidia: Suggest dropping this in favor of the Custom attributes section, where the same snippet sits next to the explanation of why `register_custom_attributes` must come before asset loading.

```suggestion
:class:`~newton.solvers.SolverMuJoCo` wraps `mujoco_warp
<https://github.com/google-deepmind/mujoco_warp>`_ behind Newton's standard
solver interface.
```

docs/integrations/mujoco.rst
```diff
@@ -6,21 +6,44 @@
...
+Newton concepts the solver supports, how each is mapped to MuJoCo, how
+state is exchanged at every step, and where each piece of the conversion
+lives in the source. MuJoCo-specific behaviour that has no Newton-core
+equivalent is exposed through the :ref:`custom-attribute namespace
+<mujoco-custom-attributes>`. A :ref:`code pointers <mujoco-code-pointers>`
+section at the bottom collects the most useful anchor points.
+
 
 Coordinate conventions
 ----------------------
```
preist-nvidia: Rename to `Conventions` (or `Convention differences`) — the section already covers body ordering in addition to quaternion order, and is likely to grow.

```suggestion
Conventions
-----------
```

docs/integrations/mujoco.rst
```diff
@@ -6,21 +6,44 @@
...
-  the ``mj_model``/``mj_data`` attributes on the solver).
+  The solver converts between the two automatically (via
+  ``quat_xyzw_to_wxyz`` / ``quat_wxyz_to_xyzw`` in ``kernels.py``). Be aware
+  of this when inspecting raw MuJoCo objects — e.g. through ``save_to_mjcf``
+  or the solver's ``mj_model``/``mj_data`` attributes.
+
+**Body ordering.**
+  The solver sorts joints in depth-first topological order when building
+  MuJoCo's kinematic tree. If the Newton model's joint order differs, a
+  warning is emitted because kinematics may otherwise diverge.
```
preist-nvidia: Does this need to be called out in the docs? Push/pull (`convert_mj_coords_to_warp_kernel`, `kernels.py:505`) does indexed gather/scatter either way, so no correctness or perf impact on `step()`. Only user-visible in `save_to_mjcf` / direct `mj_data` introspection — where the runtime warning already explains itself when it fires. Am I missing a case?

docs/integrations/mujoco.rst
```diff
@@ -220,114 +206,233 @@ Equality constraints
...
+  MPM, and FEM; they are not part of the MuJoCo integration.
+- **Skinned meshes** (``<skin>``) — visualisation-only, not imported.
+- **User plugins** (``<plugin>``) — MuJoCo's plugin mechanism for custom
+  passive forces or dynamics is not supported.
+- **User data and arbitrary custom elements** (``<custom>``, ``<numeric>``,
+  ``<text>``) — not imported. Newton-specific user data should use the
+  Newton custom-attribute system instead.
+- **Actuator transmission types beyond joint / tendon / site / body /
+  slidercrank** — notably, MuJoCo's ``adhesion`` transmission is not
+  supported.
```
preist-nvidia: This bullet enumerates the supported transmissions in prose, which drifts on both sides: `jointinparent` is already missing from the list (it's in `SolverMuJoCo.TrnType` and handled by `parse_trntype` in `solver_mujoco.py:968`), and any future change — upstream adding a new transmission, or Newton gaining `adhesion` support — requires a doc edit to stay accurate.

Suggest dropping the enumeration entirely and pointing to the enum as the single source of truth:

```suggestion
- **Actuator transmissions** — not all MuJoCo transmission types are
  supported. See :class:`~newton.solvers.SolverMuJoCo.TrnType` for the
  full list.
```

`SolverMuJoCo.TrnType` is a nested public `IntEnum` at `newton/_src/solvers/mujoco/solver_mujoco.py:193`. This mirrors how the custom-attributes section defers to `register_custom_attributes` rather than listing attributes in prose.

docs/integrations/mujoco.rst
```diff
@@ -52,8 +75,15 @@ Joint types
...
-       connects to the world, the body is created as a **mocap** body.
+     - The child body is nested directly under its parent. A fixed joint
+       connecting to the world produces a **mocap** body, driven via
+       ``mjData.mocap_pos`` / ``mjData.mocap_quat``.
+   * - ``DISTANCE``, ``CABLE``
+     - *unsupported*
+     - Not forwarded to MuJoCo.
+
+Per-DOF ``joint_velocity_limit`` values have no MuJoCo equivalent and are
+ignored.
```
preist-nvidia: Duplicates the `Velocity limits are not forwarded.` caveat at rst:428. Drop.

```suggestion
```

docs/integrations/mujoco.rst
adenzler-nvidia: Done — dropped from the intro. Snippet stays in the *Custom attributes* section.

docs/integrations/mujoco.rst
adenzler-nvidia: Ended up dropping the section entirely rather than renaming. The body-ordering bullet went per your other comment, and on reflection the quaternion-order bullet is in the same boat: both are transparent during `step()` (indexed gather/scatter, joints re-sorted internally) and only surface when you're already inspecting raw `mj_model` / `mj_data` or a `save_to_mjcf` file. Once both bullets were gone, a heading with no content didn't earn its keep.

docs/integrations/mujoco.rst
adenzler-nvidia: You weren't missing a case. `SolverMuJoCo._convert_to_mjc` topologically sorts joints itself before emitting them to MuJoCo (`solver_mujoco.py:4232-4240`), and `ModelBuilder.validate_joint_ordering` is opt-in (not run from `finalize`). Combined with the indexed gather/scatter in push/pull, user-side joint order doesn't reach MuJoCo's kinematic tree at all. Dropped the section.

Side note: given the above, the runtime warning at `solver_mujoco.py:4238` ("may lead to diverging kinematics between MuJoCo and Newton") is overstated — happy to send a follow-up softening or removing it if you agree.

docs/integrations/mujoco.rst
adenzler-nvidia: Done — replaced with the pointer to `SolverMuJoCo.TrnType`, verbatim from your suggestion.

docs/integrations/mujoco.rst
adenzler-nvidia: Dropped, thanks for catching it.

docs/integrations/mujoco.rst
```diff
@@ -339,29 +415,60 @@ user-prescribed rather than dynamically integrated:
...
 for a fixed-root articulation after constructing the solver, call
 ``solver.notify_model_changed(newton.solvers.SolverNotifyFlags.JOINT_PROPERTIES)``
-to synchronize the updated fixed-root poses into MuJoCo.
+to synchronise the updated fixed-root poses into MuJoCo.
+
+
+.. _mujoco-code-pointers:
+
+Where the conversion lives
+--------------------------
```
preist-nvidia: `Code pointers` reads better in retrieval / TOC than `Where the conversion lives` — terser, matches the `.. _mujoco-code-pointers:` anchor, and the section also lists `step` / `register_custom_attributes` (not pure conversion).

```suggestion
Code pointers
-------------
```

docs/integrations/mujoco.rst
```diff
@@ -220,114 +177,233 @@ Equality constraints
...
+
+1. **Constructor argument** passed to :class:`~newton.solvers.SolverMuJoCo`
+   — one value, applied to all worlds. The full list of kwargs, their
+   types, and their defaults is documented on the class itself.
+2. **Custom attribute** (``model.mujoco.<option>``) — supports per-world
+   values. Typically populated automatically by USD or MJCF import.
+3. **Default** — if neither of the above is set, the MuJoCo default is
+   used, with one Newton-opinionated exception: ``integrator`` defaults
+   to ``implicitfast`` (MuJoCo's default is ``euler``) for better
+   stability on stiff systems.
```
preist-nvidia, line 332: The section names every parameter but doesn't link to MuJoCo's docs explaining what they do. An agent told to tune solver parameters has the names but no semantics. Add at the end of the section:

    See MuJoCo's `solver documentation
    <https://mujoco.readthedocs.io/en/stable/computation/index.html>`_ and
    `<option> XML reference
    <https://mujoco.readthedocs.io/en/stable/XMLreference.html#option>`_ for
    what each parameter does and when to tune it.

docs/integrations/mujoco.rst
```diff
@@ -31,7 +28,7 @@ Joint types
    :widths: 25 35 40
 
    * - Newton type
-     - MuJoCo type(s)
+     - MuJoCo equivalent
      - Notes
    * - ``FREE``
```
preist-nvidia: Plain literals here violate AGENTS.md: *"Sphinx cross-refs (`:class:`, `:meth:`) with shortest possible targets. Prefer public API paths."* `newton.JointType` and `newton.GeoType` are both publicly exported, so agents writing Newton code currently have to guess the import path. Use `:attr:` refs:

```suggestion
   * - :attr:`~newton.JointType.FREE`
```

Same pattern for every row in this table and in `Geometry types` (`:attr:`~newton.GeoType.SPHERE``, etc.).

docs/integrations/mujoco.rst
preist-nvidia: Yeah makes sense to fix the warning.

docs/integrations/mujoco.rst
```diff
@@ -83,96 +84,50 @@ Geometry types
...
 
-Non-colliding shapes (no ``COLLIDE_SHAPES`` flag, or ``collision_group == 0``)
-get ``contype = conaffinity = 0``.
+**Sites** (shapes with the ``SITE`` flag) are converted to MuJoCo sites —
+non-colliding reference frames used for sensor attachment and spatial
+tendon wrap anchors.
 
-Additionally, body pairs for which all shape-shape combinations are filtered
-are registered as ``<exclude>`` elements.
+SDF-based and hydroelastic collisions are not part of the MuJoCo geometry
```
preist-nvidia: Newton's pipeline supports more than just SDF / hydroelastic — non-convex trimesh too. Phrasing the list more openly so it doesn't claim completeness:

```suggestion
Several Newton collision features — for example non-convex trimesh,
SDF-based contacts, and hydroelastic contacts — are not part of the
MuJoCo geometry model. They are only available through Newton's
collision pipeline (see `Collision pipeline`_ below).
```

docs/integrations/mujoco.rst
```diff
@@ -83,96 +84,50 @@ Geometry types
...
 
-Bodies with positive mass have their mass, center-of-mass offset (``ipos``),
-and inertia tensor set explicitly (``explicitinertial="true"``).  When the
-inertia tensor is diagonal the solver uses ``diaginertia``; otherwise it uses
-``fullinertia``.
+Bodies with positive mass are exported with ``explicitinertial=True``,
+forwarding mass, centre-of-mass offset (``ipos``), and inertia tensor.
+When the inertia tensor is diagonal it is passed as a 3-element vector
+(preserving MuJoCo's sameframe optimisation, ``body_simple=1``); otherwise
+it is passed as the full 6-element symmetric form.
```
preist-nvidia: I don't follow the interplay between zero-mass bodies and the `inertiafromgeom = auto` flag here. Can you clarify what happens for (a) a body with no Newton mass and no child geoms (e.g. a pure sensor frame) vs. (b) a body with no Newton mass but child geoms?

preist-nvidia (changes requested):

adenzler-nvidia:
Pushed two new commits.

`1ec25cd3` (already up) folds in the round-2 review feedback:
- `:attr:` cross-refs in Joint types / Geometry types / Equality constraints (also the existing inline `:class:` ref to `JointType.FIXED`)
- Geometry note broadened to cover non-convex trimesh
- Mass and inertia section dropped — `inertiafromgeom = auto` is MuJoCo's MjSpec default and only diverged from Newton's intent in a pathological corner; documenting it overclaimed
- MuJoCo solver-doc + `<option>` XML reference links added at the end of Solver options
- "Where the conversion lives" → "Code pointers"
- New: loop-closure synthesis (`joint_articulation == -1` → `mjEQ_*`, with the 0.1 m hinge-axis offset for revolute) and multi-CCD / margin-zeroing notes under Collision pipeline

`2127c667` is a section-by-section walkthrough where I cross-checked every numerical and behavioural claim in the page against `solver_mujoco.py` / `kernels.py` / `schemas.py` / `builder.py`. The material changes:

- Intro: note the `~=` pin on `mujoco` and `mujoco-warp`; switched repo-relative paths (`kernels.py`, `schemas.py`, `pyproject.toml`) to the `:github:` extlink so they track the release ref
- Geometry types: PLANE row uses Newton-side terminology (`body=-1`) and mentions the conversion-time `ValueError`; drive-by fix to the heightfield docstring in `builder.py` to drop "world body"
- Equality constraints: WELD row now lists anchor / relpose / torque-scale slots, matching the detail in CONNECT and JOINT
- Collision pipeline: margin-zeroing warning now correctly gated on `use_mujoco_contacts=True` (with `use_mujoco_contacts=False` the runtime kernel restores margins and no warning fires)
- Multi-world support: global-world restriction is enforced as a hard `ValueError` for bodies/joints/eq/mimic, not optional — reworded
- Runtime state synchronisation: explicit note that the joint-state push is gated by `update_data_interval`
- Solver options: dropped the inline kwargs enumeration; the autoclass docstring is the source of truth and the inline list drifts

Open follow-up I noted in an earlier reply: the "kinematics may otherwise diverge" wording in `solver_mujoco.py:4238` is overstated given that joint order is internally re-sorted and push/pull is indexed; happy to send a small follow-up softening it. No code changes in this PR.

docs/integrations/mujoco.rst
```diff
@@ -64,115 +67,56 @@ Geometry types
...
+MuJoCo geometry model. They are only available through Newton's
+collision pipeline (see `Collision pipeline`_ below).
 
 
 Actuators
 ---------
 
-Newton's per-DOF ``joint_target_mode`` determines which MuJoCo general
-actuators are created:
+Newton's per-DOF ``joint_target_mode`` creates MuJoCo general actuators:
```
preist-nvidia:
```suggestion
Newton's per-DOF :attr:`~newton.Model.joint_target_mode` creates MuJoCo general actuators:
```

docs/integrations/mujoco.rst
```diff
@@ -191,143 +135,308 @@ actuators are created:
    * - ``NONE``
      - No actuator created for this DOF.
```
preist-nvidia: Fold `EFFORT` into the no-actuator row — it falls through `solver_mujoco.py:4679-4716` to the same outcome as `NONE`.

```suggestion
   * - :attr:`~newton.JointTargetMode.NONE`,
       :attr:`~newton.JointTargetMode.EFFORT`
     - No MuJoCo actuator created.
```

The POSITION / VELOCITY / POSITION_VELOCITY rows above (lines 127, 129, 131) are unchanged context so I can't suggestion them — same `:attr:` substitution applies:
- ``POSITION`` → `:attr:`~newton.JointTargetMode.POSITION``
- ``VELOCITY`` → `:attr:`~newton.JointTargetMode.VELOCITY``
- ``POSITION_VELOCITY`` → `:attr:`~newton.JointTargetMode.POSITION_VELOCITY``

docs/integrations/mujoco.rst
```diff
@@ -191,143 +135,308 @@ actuators are created:
...
 
-``joint_effort_limit`` is forwarded as ``actfrcrange`` on the joint (for
-prismatic, revolute, and D6 joints) or as ``forcerange`` on the actuator (for
-ball joints).
+``joint_effort_limit`` is forwarded as ``actfrcrange`` on the joint
```
preist-nvidia:
```suggestion
:attr:`~newton.Model.joint_effort_limit` is forwarded as ``actfrcrange`` on the joint
```

docs/integrations/mujoco.rst
```diff
@@ -191,143 +135,308 @@ actuators are created:
...
+   * - :attr:`~newton.EqType.JOINT`
      - ``mjEQ_JOINT``
      - Polynomial coefficients forwarded in ``data[0:5]``.
    * - Mimic
      - ``mjEQ_JOINT``
-     - ``coef0`` / ``coef1`` mapped to polynomial coefficients.
-       Only ``REVOLUTE`` and ``PRISMATIC`` joints are supported.
+     - ``coef0`` / ``coef1`` mapped to polynomial coefficients. Only
+       :attr:`~newton.JointType.REVOLUTE` and
+       :attr:`~newton.JointType.PRISMATIC` joints are supported.
```
preist-nvidia, line 200: Link to the public builder for mimic constraints.

```suggestion
   * - Mimic
     - ``mjEQ_JOINT``
     - Added via :meth:`~newton.ModelBuilder.add_constraint_mimic`. Maps
       ``coef0`` / ``coef1`` to polynomial coefficients. Only
       :attr:`~newton.JointType.REVOLUTE` and
       :attr:`~newton.JointType.PRISMATIC` joints are supported.
```

docs/integrations/mujoco.rst
```diff
@@ -191,143 +135,308 @@ actuators are created:
...
 
 By default :class:`~newton.solvers.SolverMuJoCo` uses MuJoCo's built-in
-collision detection (``use_mujoco_contacts=True``).  Alternatively, you can set
-``use_mujoco_contacts=False`` and pass contacts computed by Newton's own
-collision pipeline into :meth:`~newton.solvers.SolverMuJoCo.step`.  Newton's
-pipeline supports non-convex meshes, SDF-based contacts, and hydroelastic
-contacts, which are not available through MuJoCo's collision detection.
+collision detection (``use_mujoco_contacts=True``). Alternatively, you can
+set ``use_mujoco_contacts=False`` and pass contacts computed by Newton's
+own collision pipeline into :meth:`~newton.solvers.SolverMuJoCo.step`.
```
preist-nvidia: Make it explicit that this is a constructor kwarg.

```suggestion
:class:`~newton.solvers.SolverMuJoCo` uses MuJoCo's built-in collision
detection by default. Construct it with ``use_mujoco_contacts=False``
to feed contacts computed by Newton's own collision pipeline into
:meth:`~newton.solvers.SolverMuJoCo.step` instead.
```

docs/integrations/mujoco.rst
```diff
@@ -191,143 +135,308 @@ actuators are created:
...
+Newton's pipeline supports non-convex meshes, SDF-based contacts, and
+hydroelastic contacts, which are not available through MuJoCo's collision
+detection.
+
+**Multi-contact CCD.** With ``enable_multiccd=True`` the solver allows
+up to four contact points per geom pair instead of one. Pairs where
+either geom has ``margin > 0`` still fall back to a single contact
+regardless of the flag.
```
preist-nvidia: Same constructor-kwarg phrasing, and clarify that this is MuJoCo's `geom_margin` (not the authored `shape_margin`) and cross-link to the next paragraph that covers the Newton↔MuJoCo margin mapping.

```suggestion
**Multi-contact CCD.** Constructing
:class:`~newton.solvers.SolverMuJoCo` with ``enable_multiccd=True``
allows up to four contact points per geom pair instead of one. Pairs
where either geom has non-zero MuJoCo ``geom_margin`` still fall back
to a single contact regardless of the flag (see `Margin zeroing`_ for
how Newton's :attr:`~newton.Model.shape_margin` is forwarded to it).
```

docs/integrations/mujoco.rst
```diff
@@ -191,143 +135,308 @@ actuators are created:
...
+
+**Margin zeroing.** ``mujoco_warp`` rejects non-zero geom margins on
+box-box pairs (its default NATIVECCD path) and on any box/mesh pair
+when ``enable_multiccd=True``. To stay compatible the solver zeroes
+``geom_margin`` model-wide at compile time whenever a box geom exists,
+or whenever ``enable_multiccd=True`` is combined with mesh geoms; geoms
+with non-zero authored margins emit a warning when
+``use_mujoco_contacts=True``. The Newton model's ``shape_margin`` array
```
preist-nvidia:
```suggestion
``use_mujoco_contacts=True``. The Newton model's :attr:`~newton.Model.shape_margin` array
```

docs/integrations/mujoco.rst
```diff
@@ -191,143 +135,308 @@ actuators are created:
...
+
+Each call to :meth:`~newton.solvers.SolverMuJoCo.step` goes through the
+same three-phase cycle:
+
+1. **Push Newton → MuJoCo.** Joint positions and velocities
+   (``state_in.joint_q`` / ``joint_qd``) are copied into MuJoCo's
+   ``qpos`` / ``qvel`` with quaternion reordering. Control is applied via
+   ``control.joint_target_pos`` / ``joint_target_vel`` (→ MuJoCo
+   ``ctrl``), ``control.joint_f`` (→ ``qfrc_applied``), and
+   ``state_in.body_f`` (→ ``xfrc_applied``). Direct MuJoCo actuator
```
preist-nvidia: Make it strictly a generalization — drop the field-level routing detail (lives in `solver_mujoco.py:3420`) and trim the gating sentence to a substepping pointer.

```suggestion
1. **Push Newton → MuJoCo.** ``SolverMuJoCo._apply_mjc_control`` and
   ``SolverMuJoCo._update_mjc_data`` transfer the Newton ``State``
   and ``Control`` inputs to MuJoCo's working data. When
   ``use_mujoco_contacts=False``, Newton-side contacts are also
   converted before the integrator runs. The joint-state re-sync
   frequency can be controlled via the ``update_data_interval``
   kwarg for substepping schemes.
```

docs/integrations/mujoco.rst
```diff
@@ -191,143 +135,308 @@ actuators are created:
...
+   ``mjData.contact`` before the integrator runs. The state sync
+   (``joint_q`` / ``joint_qd`` → ``qpos`` / ``qvel``) is gated by the
+   ``update_data_interval`` kwarg, which defaults to ``1`` (sync every
+   step); larger values amortise the sync, and ``0`` disables it after
+   construction so MuJoCo simulates from its own internal state.
+2. **Integrate.** ``mujoco_warp`` (or the CPU MuJoCo backend when
+   ``use_mujoco_cpu=True``) steps the MuJoCo model forward by ``dt``.
+3. **Pull MuJoCo → Newton.** Integrated ``qpos`` / ``qvel`` are copied
+   back into ``state_out.joint_q`` / ``joint_qd``, with quaternion
+   reordering; ``body_q`` / ``body_qd`` are recomputed from the new joint
```
preist-nvidia: Generalize the field-level enumeration — auxiliary outputs (body_qdd, body_parent_f, qfrc_actuator) are documented in the code.

```suggestion
3. **Pull MuJoCo → Newton.** ``SolverMuJoCo._update_newton_state``
   populates the output ``State`` from the integrated MuJoCo data.
   Kinematic roots pass through unchanged from ``state_in`` (see
   `Kinematic links and fixed roots`_).
```

docs/integrations/mujoco.rst
```diff
@@ -191,143 +135,308 @@ actuators are created:
...
+  :attr:`~newton.ModelBuilder.ShapeConfig.gap` value. MJCF/USD ``gap``
+  values are still imported into
+  :attr:`~newton.ModelBuilder.ShapeConfig.gap` on the Newton model, but
+  they are not forwarded to the MuJoCo solver.
 
 **shape_collision_radius is ignored.**
-  MuJoCo computes bounding-sphere radii (``geom_rbound``) internally from the
-  geometry definition.  Newton's ``shape_collision_radius`` is not forwarded.
+  MuJoCo computes bounding-sphere radii (``geom_rbound``) internally from
+  the geometry definition. Newton's ``shape_collision_radius`` is not
```
preist-nvidia:
```suggestion
  the geometry definition. Newton's :attr:`~newton.Model.shape_collision_radius` is not
```

docs/integrations/mujoco.rst
```diff
@@ -191,143 +135,308 @@ actuators are created:
...
+  effective collision boundary.
 
 **Velocity limits are not forwarded.**
-  Newton's ``joint_velocity_limit`` has no MuJoCo equivalent and is ignored.
-
-**Body ordering must be depth-first.**
-  The solver sorts joints in depth-first topological order for MuJoCo's
-  kinematic tree.  If the Newton model's joint order differs, a warning is
-  emitted because kinematics may diverge.
+  Newton's ``joint_velocity_limit`` has no MuJoCo equivalent and is
```
preist-nvidia:
```suggestion
  Newton's :attr:`~newton.Model.joint_velocity_limit` has no MuJoCo equivalent and is
```

docs/integrations/mujoco.rst
```diff
@@ -339,29 +448,60 @@ user-prescribed rather than dynamically integrated:
...
+  no joint coordinates for a fixed root, so Newton drives the pose through
   ``mjData.mocap_pos`` and ``mjData.mocap_quat`` instead.
 - **World-attached shapes that are not part of an articulation** remain
   ordinary static MuJoCo geometry rather than mocap bodies.
 
 If you edit :attr:`newton.Model.joint_X_p` or :attr:`newton.Model.joint_X_c`
 for a fixed-root articulation after constructing the solver, call
 ``solver.notify_model_changed(newton.solvers.SolverNotifyFlags.JOINT_PROPERTIES)``
-to synchronize the updated fixed-root poses into MuJoCo.
+to synchronise the updated fixed-root poses into MuJoCo.
```
preist-nvidia:
```suggestion
for a fixed-root articulation after constructing the solver, call
:meth:`~newton.solvers.SolverBase.notify_model_changed` with the
:attr:`~newton.solvers.SolverNotifyFlags.JOINT_PROPERTIES` flag to
synchronise the updated fixed-root poses into MuJoCo.
```

docs/integrations/mujoco.rst
```diff
@@ -191,143 +135,308 @@ actuators are created:
...
-custom attributes and are appended after the joint-target actuators.
+The full MuJoCo general-actuator model (arbitrary gain/bias/dynamics types
+and parameters, explicit transmission targets, ctrl/force/act ranges) is
+only reachable through the :ref:`mujoco-custom-attributes` namespace.
+Additional actuators declared this way are appended after the joint-target
+actuators — see ``SolverMuJoCo._init_actuators``.
 
 
 Equality constraints
 --------------------
```
preist-nvidia, line 173: Anchor before the table so the `data[...]` references make sense to readers without MuJoCo internals knowledge.

```suggestion

Each row's ``data[...]`` reference below points into MuJoCo's
`equality.data <https://mujoco.readthedocs.io/en/stable/XMLreference.html#equality>`_
array; slot layout depends on the constraint type.

```

docs/integrations/mujoco.rst
```diff
@@ -191,143 +135,308 @@ actuators are created:
...
+``contype``. ``conaffinity`` is set to the complement so same-color geoms
+never match. Up to 32 colors are supported (one per ``contype`` bit). If the
+graph requires more than 32 colors, shapes with color index ≥ 32 fall back
+to MuJoCo defaults (``contype=1``, ``conaffinity=1``) and will collide with
+all other shapes, silently bypassing the intended filtering.
+
+Non-colliding shapes (no ``COLLIDE_SHAPES`` flag, or
+``collision_group == 0``) get ``contype = conaffinity = 0``. Body pairs for
+which all shape-shape combinations are filtered are emitted as
+``<exclude>`` elements.
```
preist-nvidia: Most of this section is impl detail (graph-coloring algorithm, bitmask formula, `contype/conaffinity` zeroing for non-colliders, `<exclude>` emission). The user-facing piece — the silent >32-color fallback — fits better as a `Caveats` bullet (separate suggestion below). Drop the whole section.

@adenzler-nvidia — does the >32-color fallback also have measurable perf impact (excess shapes joining the default-collision bucket)? If yes, worth surfacing in the new caveat.

```suggestion
```

docs/integrations/mujoco.rst
```diff
@@ -191,143 +135,308 @@ actuators are created:
...
 
 **Velocity limits are not forwarded.**
-  Newton's ``joint_velocity_limit`` has no MuJoCo equivalent and is ignored.
-
-**Body ordering must be depth-first.**
-  The solver sorts joints in depth-first topological order for MuJoCo's
-  kinematic tree.  If the Newton model's joint order differs, a warning is
-  emitted because kinematics may diverge.
+  Newton's ``joint_velocity_limit`` has no MuJoCo equivalent and is
+  ignored.
```
preist-nvidia, line 454: Two new caveats: collision-filtering bitmask fallback (folded from the dropped Collision filtering section) + a CPU+Newton-contacts gotcha I noticed in the runtime-sync review.

```suggestion
  ignored.

**Collision filtering bitmask fallback.**
  Newton's :attr:`~newton.Model.shape_collision_group` (see
  :ref:`Collision Groups`) is translated to MuJoCo's ``contype`` /
  ``conaffinity`` via graph coloring
  (:github:`newton/_src/sim/graph_coloring.py`). Up to 32 colors are
  supported. If the filtering graph requires more, shapes with color
  index ≥ 32 fall back to ``contype=1`` / ``conaffinity=1`` and
  silently collide with every other default-bucket shape, bypassing
  the intended filtering.

**MuJoCo CPU + Newton contacts.**
  ``use_mujoco_contacts=False`` requires the ``mujoco_warp`` backend
  (the default). Combining ``use_mujoco_cpu=True`` with
  ``use_mujoco_contacts=False`` silently drops Newton-side contacts —
  only MuJoCo's own CPU collision detection runs.
```

docs/integrations/mujoco.rst
```diff
@@ -191,143 +135,308 @@ actuators are created:
...
+detection.
+
+**Multi-contact CCD.** With ``enable_multiccd=True`` the solver allows
+up to four contact points per geom pair instead of one. Pairs where
+either geom has ``margin > 0`` still fall back to a single contact
+regardless of the flag.
+
+**Margin zeroing.** ``mujoco_warp`` rejects non-zero geom margins on
+box-box pairs (its default NATIVECCD path) and on any box/mesh pair
+when ``enable_multiccd=True``. To stay compatible the solver zeroes
```
preist-nvidia: Disambiguate "the solver" from `mujoco_warp` since both are referred to as solvers.

```suggestion
when ``enable_multiccd=True``. To stay compatible :class:`~newton.solvers.SolverMuJoCo` zeroes
```

docs/integrations/mujoco.rst
```diff
@@ -6,21 +6,20 @@
...
-**Quaternion order.**
-  Newton stores quaternions as ``(x, y, z, w)``; MuJoCo uses ``(w, x, y, z)``.
-  The solver converts between the two automatically.  Be aware of this when
-  inspecting raw MuJoCo model or data objects (e.g. via ``save_to_mjcf`` or
-  the ``mj_model``/``mj_data`` attributes on the solver).
+:class:`~newton.solvers.SolverMuJoCo` wraps `mujoco_warp
+<https://github.com/google-deepmind/mujoco_warp>`_ behind Newton's standard
+solver interface. Newton compatible-release-pins (``~=``) both ``mujoco``
+and ``mujoco-warp`` to keep the two version-aligned; see
+:github:`pyproject.toml` for the current pins.
```
preist-nvidia, line 13: The doc is currently mujoco_warp-centric; the CPU backend (`use_mujoco_cpu=True`) only surfaces in passing in the runtime-sync section. A skimming reader wouldn't know there's a CPU path.

That said — not sure if we want to keep supporting it long-term, so this is more a question than a recommendation. If we do plan to keep CPU support, the suggestion below adds a brief intro mention; if not, leave the doc as-is and let the kwarg quietly fade.

```suggestion
:class:`~newton.solvers.SolverMuJoCo` wraps `mujoco_warp
<https://github.com/google-deepmind/mujoco_warp>`_ behind Newton's standard
solver interface; constructing it with ``use_mujoco_cpu=True`` swaps the
backend to the MuJoCo C reference implementation, useful for CPU-only
environments and debugging numerical mismatches. Newton
compatible-release-pins (``~=``) both ``mujoco`` and ``mujoco-warp`` to
keep the two version-aligned; see :github:`pyproject.toml` for the
current pins.
```

docs/integrations/mujoco.rst
```diff
@@ -191,143 +135,308 @@ actuators are created:
...
+
+Multi-world support
+-------------------
+
+When ``separate_worlds=True`` (the default for GPU mode with multiple
+worlds), the solver builds a MuJoCo model from the **first world** only and
+replicates it across all worlds via ``mujoco_warp``. This requires all
+Newton worlds to be structurally identical (same bodies, joints, and
+shapes); the solver validates this at construction and raises
+``ValueError`` on a mismatch.
```
preist-nvidia: Same constructor-kwarg phrasing as the use_mujoco_contacts / enable_multiccd paragraphs, plus disambiguate "the solver" → SolverMuJoCo (mujoco_warp does the replication, the wrapper does the structural-identity check).

```suggestion
Constructing :class:`~newton.solvers.SolverMuJoCo` with
``separate_worlds=True`` (the default for GPU mode with multiple
worlds) builds a MuJoCo model from the **first world** only and
replicates it across all worlds via ``mujoco_warp``. This requires
all Newton worlds to be structurally identical (same bodies, joints,
and shapes); :class:`~newton.solvers.SolverMuJoCo` validates this at
construction and raises ``ValueError`` on a mismatch.
```

docs/integrations/mujoco.rst
```diff
@@ -6,21 +6,20 @@
...
+:github:`pyproject.toml` for the current pins.
+
+Because MuJoCo has its own modelling conventions, many Newton properties
+are mapped differently or not at all. The sections below describe which
+Newton concepts the solver supports, how each is mapped to MuJoCo, how
+state is exchanged at every step, and where each piece of the conversion
+lives in the source. MuJoCo-specific behaviour that has no Newton-core
+equivalent is exposed through the :ref:`custom-attribute namespace
+<mujoco-custom-attributes>`. A :ref:`code pointers <mujoco-code-pointers>`
+section at the bottom collects the most useful anchor points.
```
preist-nvidia, line 22: Add a note clarifying the `mjModel` / `mjData` naming throughout the page so the doc isn't pinned to one backend's struct names.

```suggestion
section at the bottom collects the most useful anchor points.

.. note::

   References to ``mjModel`` / ``mjData`` throughout this page apply
   to both backends — the CPU MuJoCo types and ``mujoco_warp``'s
   ``mjwModel`` / ``mjwData`` analogues. The active structure depends
   on ``use_mujoco_cpu``.
```

docs/integrations/mujoco.rst
```diff
@@ -191,143 +135,308 @@ actuators are created:
...
+See :doc:`/concepts/custom_attributes` for how Newton's custom-attribute
+system works in general.
+
+A subset of ``mjc:*`` USD attributes is additionally mapped onto Newton's
+built-in properties during USD import by
+:class:`~newton.usd.SchemaResolverMjc` (:github:`newton/_src/usd/schemas.py`) —
+for example, joint-limit stiffness/damping from ``mjc:solreflimit``, and
+torsional/rolling friction from ``mjc:torsionalfriction`` /
+``mjc:rollingfriction``. Attributes not handled by the schema resolver
+land in the ``mujoco`` namespace unchanged.
```
preist-nvidia: Combine the existing two paragraphs under one section that's honest about covering both surfaces.

```suggestion
.. _mujoco-custom-attributes:

MuJoCo-specific parameters in USD and MJCF
------------------------------------------

MuJoCo has parameters with no counterpart in Newton's core API.
:class:`~newton.ModelBuilder` handles them during MJCF / USD import
via two mechanisms.

**Custom-attribute namespace.** A dedicated ``mujoco`` custom-attribute
namespace (``model.mujoco.<name>``) is populated from MJCF elements
and from attributes in the OpenUSD MuJoCo schema (``mjc:*``). To
enable the namespace, call
:meth:`~newton.solvers.SolverMuJoCo.register_custom_attributes` on the
:class:`~newton.ModelBuilder` **before** adding anything to it::

    import newton
    from newton.solvers import SolverMuJoCo

    builder = newton.ModelBuilder()
    SolverMuJoCo.register_custom_attributes(builder)
    # ...then add anything (e.g. import MJCF / USD, add joints, ...)
    model = builder.finalize()

The authoritative list of registered attributes — names, defaults,
dtypes, MJCF / USD source names, and the category each belongs to —
is the body of
:meth:`~newton.solvers.SolverMuJoCo.register_custom_attributes`
itself. See :doc:`/concepts/custom_attributes` for how Newton's
custom-attribute system works in general.

**Direct mapping to Newton built-ins.** Some MuJoCo-specific
attributes are mapped onto Newton's built-in properties during import
(rather than the ``mujoco`` namespace) — for example, joint-limit
stiffness and damping derived from ``solreflimit``. The MJCF parser
handles this inline (:github:`newton/_src/utils/import_mjcf.py`); USD
goes through :class:`~newton.usd.SchemaResolverMjc`
(:github:`newton/_src/usd/schemas.py`).
```

docs/integrations/mujoco.rst
```diff
@@ -191,143 +135,308 @@ actuators are created:
...
-prismatic, revolute, and D6 joints) or as ``forcerange`` on the actuator (for
-ball joints).
+``joint_effort_limit`` is forwarded as ``actfrcrange`` on the joint
+(prismatic, revolute, and D6) or as ``forcerange`` on the actuator (ball).
 
-Additional MuJoCo general actuators (motors, etc.) can be attached through
-custom attributes and are appended after the joint-target actuators.
+The full MuJoCo general-actuator model (arbitrary gain/bias/dynamics types
+and parameters, explicit transmission targets, ctrl/force/act ranges) is
+only reachable through the :ref:`mujoco-custom-attributes` namespace.
```
preist-nvidia: The bare `:ref:` would render with the renamed section title; use explicit display text.

```suggestion
only reachable through the ``mujoco`` :ref:`custom-attribute namespace <mujoco-custom-attributes>`.
```

docs/integrations/mujoco.rst
```diff
@@ -191,143 +135,308 @@ actuators are created:
...
-  emitted because kinematics may diverge.
+  Newton's ``joint_velocity_limit`` has no MuJoCo equivalent and is
+  ignored.
 
 
 .. _mujoco-kinematic-links-and-fixed-roots:
 
-Kinematic Links and Fixed Roots
+Kinematic links and fixed roots
 -------------------------------
```
preist-nvidia, line 478: "link" is unusual in Newton (entity is "body"). Anchor `.. _mujoco-kinematic-links-and-fixed-roots:` left as-is to preserve the cross-ref from `concepts/articulations.rst:263`.

```suggestion
Kinematic and fixed roots
-------------------------
```

docs/integrations/mujoco.rst
```diff
@@ -339,29 +448,60 @@ user-prescribed rather than dynamically integrated:
...
+  joints with the same Newton joint type and DOFs. The solver assigns a
+  very large internal armature to those DOFs so MuJoCo treats them like
+  prescribed, effectively infinite-mass coordinates.
+- **Roots attached to world with a fixed joint** are exported as MuJoCo
+  mocap bodies. This applies to both kinematic and non-kinematic Newton
+  roots attached to world by :attr:`~newton.JointType.FIXED`. MuJoCo has
+  no joint coordinates for a fixed root, so Newton drives the pose through
   ``mjData.mocap_pos`` and ``mjData.mocap_quat`` instead.
 - **World-attached shapes that are not part of an articulation** remain
   ordinary static MuJoCo geometry rather than mocap bodies.
```
preist-nvidia: Drop "exporter"/"export" terminology — it's not defined elsewhere and conflates SolverMuJoCo's at-construction conversion with the optional `save_to_mjcf` file export. Reword "are exported as" to "become" throughout the export bullets and tighten the kinematic/non-kinematic clause on the mocap row.

Note: the intro paragraph (lines 442-444) needs the same treatment ("MuJoCo exporter" → drop the term, "kinematic link" → "kinematic body", "exported normally" → "converted normally"), but those lines are unchanged context outside this suggestion's anchor — they have to be edited manually.

```suggestion

- When converting MuJoCo state back to Newton, the previous Newton
  :attr:`newton.State.joint_q` and :attr:`newton.State.joint_qd` values
  are passed through for kinematic roots instead of being overwritten
  from MuJoCo's integrated ``qpos`` and ``qvel``.
- Applied body wrenches and joint forces targeting kinematic bodies
  are ignored on the MuJoCo side.
- Kinematic bodies still participate in contacts, so they can act as
  moving or fixed obstacles for dynamic bodies.

During Newton-to-MuJoCo conversion (at
:class:`~newton.solvers.SolverMuJoCo` construction), roots are mapped
by joint type:

- **Kinematic roots with non-fixed joints** become ordinary MuJoCo
  joints with the same Newton joint type and DOFs. A very large
  internal armature is assigned to those DOFs so MuJoCo treats them
  as prescribed, effectively infinite-mass coordinates.
- **Roots attached to world with a fixed joint** become MuJoCo mocap
  bodies (whether kinematic or not). MuJoCo has no joint coordinates
  for a fixed root, so Newton drives the pose through
  ``mjData.mocap_pos`` and ``mjData.mocap_quat`` instead.
- **World-attached shapes that are not part of an articulation**
  remain ordinary static MuJoCo geometry rather than mocap bodies.
```

preist-nvidia (changes requested):
A round of :attr: / :meth: cross-ref nits in the same vein as the earlier table refs, plus the missing EFFORT case folded into the actuator table.                                       

Minor higher-level note: Tendons and contact pairs bundles two unrelated concepts. Worth splitting: tendons belong near Actuators / Equality constraints (transmission/coupling), contact pairs near Collision filtering / Collision pipeline (collision overrides).

And makes sense to double-check again for british spelling.

docs/integrations/mujoco.rst
adenzler-nvidia: Applied.

docs/integrations/mujoco.rst
adenzler-nvidia: Applied.

docs/integrations/mujoco.rst
adenzler-nvidia: Applied.

docs/integrations/mujoco.rst
adenzler-nvidia: Applied.

docs/integrations/mujoco.rst
adenzler-nvidia: Applied.

docs/integrations/mujoco.rst
adenzler-nvidia: Applied.

docs/integrations/mujoco.rst
adenzler-nvidia: Applied.

docs/integrations/mujoco.rst
adenzler-nvidia: Applied.

docs/integrations/mujoco.rst
adenzler-nvidia: Applied.

docs/integrations/mujoco.rst
adenzler-nvidia: Applied.

docs/integrations/mujoco.rst
adenzler-nvidia: Applied.

docs/integrations/mujoco.rst
adenzler-nvidia: Applied.

docs/integrations/mujoco.rst
adenzler-nvidia: Applied.

docs/integrations/mujoco.rst
adenzler-nvidia: Applied.

docs/integrations/mujoco.rst
adenzler-nvidia: Applied.

docs/integrations/mujoco.rst
adenzler-nvidia: Applied.

docs/integrations/mujoco.rst
adenzler-nvidia: Applied. Switched `` `Margin zeroing`_ `` to "see *Margin zeroing* below" — the bold-prefix paragraph isn't a real section anchor, so `:ref:` resolves to nothing and Sphinx warns. Italics + "below" gets the same intent across without an explicit target.

docs/integrations/mujoco.rst
adenzler-nvidia: Section dropped. Yes, the >32-color fallback does have measurable perf impact: overflow shapes get MuJoCo defaults (`contype=1`, `conaffinity=1`), which means `(1 << N) & 1 ≠ 0` for any color N>0 and `1 & ~(1<<N) ≠ 0`, so an overflowed shape collides with every other shape — including other overflows of the same nominal color. Surfaced this in the new Caveats bullet alongside the silent-filtering point.

docs/integrations/mujoco.rst
adenzler-nvidia: First caveat (bitmask fallback) applied with an added sentence on the perf impact (extra broadphase pairs from the everyone-collides-with-everyone fallback). Second caveat (CPU + Newton contacts) skipped — decided to drop all CPU-backend references from the page instead, so this gotcha no longer has a home here.

docs/integrations/mujoco.rst
adenzler-nvidia: Going the other direction: dropped the one remaining CPU mention (in the Runtime sync section) so the page consistently describes the `mujoco_warp` path. The `use_mujoco_cpu` kwarg stays available but is no longer surfaced in the docs.

docs/integrations/mujoco.rst
adenzler-nvidia: Applied as a `.. note::` but rewrote the body. The original text referenced the CPU-vs-warp split and `mjwModel` / `mjwData`, which doesn't fit now that we've dropped CPU-backend mentions. Replaced with a pointer to MuJoCo's `mjModel` / `mjData` reference pages so readers know what `mjData.contact` etc. mean.

docs/integrations/mujoco.rst
adenzler-nvidia: Pushing back on this one. `ModelBuilder.add_link()` (`newton/_src/sim/builder.py:3533`) is real public API — "Adds a link (rigid body) to the model within an articulation." `articulations.rst` itself uses "base link", "body link positions", and `add_link()` examples. "Link" is the natural term for a body inside an articulation, which is exactly what this section is about (kinematic-root behavior only applies to articulation roots). Keeping the title as "Kinematic links and fixed roots".

docs/integrations/mujoco.rst
adenzler-nvidia: Bullet rewrite applied verbatim. Reworded the intro paragraph to drop "MuJoCo exporter" while keeping "kinematic link" (see thread on the section title — `add_link()` exists, "link" is the articulation-context term). Also caught one stray "at export time" in the Unsupported MuJoCo features intro and reworded it to "during conversion".

adenzler-nvidia:
Pushed `42bca7eb5`. All `:attr:` / `:meth:` nits applied, EFFORT folded into the actuator no-actuator row, Tendons and Contact pairs split into separate sections (Tendons after Equality constraints, Contact pairs after Collision pipeline). On spelling: the rest of `docs/` is American (`behavior` 17 vs `behaviour` 3, `color` 12 vs `colour` 0, `modeling` 3 vs `modelling` 2), so flipped this page to American to match — `modelling` → `modeling`, `behaviour` → `behavior`, `synchronisation` / `synchronise` → `synchronization` / `synchronize`. See per-thread replies for items where I deviated or pushed back (collision-filtering perf note, CPU backend dropped, kinematic-links rename declined).

docs/integrations/mujoco.rst
preist-nvidia: That's a good call, thanks.

docs/integrations/mujoco.rst
preist-nvidia: Great.

docs/integrations/mujoco.rst
preist-nvidia: Ok - in theory this section is only about roots, but makes sense to keep links for agents/users wondering about kinematic links.

docs/integrations/mujoco.rst
```diff
@@ -64,304 +78,443 @@ Geometry types
...
+  solver.
+- **Cameras and lights** declared in MJCF/USD. Newton uses its own viewer
+  and lighting pipeline; camera/light primitives in the source asset are
+  ignored.
+- **Keyframes** (``<keyframe>``) — MuJoCo's saved-state / reset mechanism
+  is not imported.
+- **Composite and flex** (``<composite>``, ``<flex>``) — MuJoCo's built-in
+  deformables and soft bodies. Newton has dedicated solvers for cloth,
+  MPM, and FEM; they are not part of the MuJoCo integration.
+- **Skinned meshes** (``<skin>``) — visualisation-only, not imported.
```
preist-nvidia: nit: missed in the British→American spelling flip — only "visualisation" remains.

```suggestion
- **Skinned meshes** (``<skin>``) — visualization-only, not imported.
```

docs/integrations/mujoco.rst
```diff
@@ -6,21 +6,31 @@
...
-----------------------
-
-**Quaternion order.**
-  Newton stores quaternions as ``(x, y, z, w)``; MuJoCo uses ``(w, x, y, z)``.
-  The solver converts between the two automatically.  Be aware of this when
-  inspecting raw MuJoCo model or data objects (e.g. via ``save_to_mjcf`` or
-  the ``mj_model``/``mj_data`` attributes on the solver).
+:class:`~newton.solvers.SolverMuJoCo` wraps `mujoco_warp
+<https://github.com/google-deepmind/mujoco_warp>`_ behind Newton's standard
+solver interface. Newton compatible-release-pins (``~=``) both ``mujoco``
```
preist-nvidia: nit: "compatible-release-pins" reads as a compound noun used as a verb.

```suggestion
solver interface. Newton uses compatible-release pins (``~=``) on both ``mujoco``
```

preist-nvidia (dismissed):
Two nits left and then let's get this in, thank you for the massive improvement.

docs/integrations/mujoco.rst
adenzler-nvidia: Done.

docs/integrations/mujoco.rst
adenzler-nvidia: Done.

docs/integrations/mujoco.rst
```diff
@@ -64,304 +83,472 @@ Geometry types
...
-   * - ``PLANE``
+   * - :attr:`~newton.GeoType.PLANE`
      - ``mjGEOM_PLANE``
-     - Must be attached to the world body.  Rendered size defaults to
-       ``5 × 5 × 5``.
-   * - ``HFIELD``
+     - Must be attached to a static body (``body=-1``); attaching to a
+       non-static body raises ``ValueError`` at conversion time. Planes
+       are infinite for collision in MuJoCo regardless of size; the
+       configured ``shape_size`` only affects rendering, defaulting to
```
preist-nvidia: `shape_size` isn't a public attribute. `solver_mujoco.py:4152` aliases `model.shape_scale.numpy()` as a local `shape_size`; the public field is `Model.shape_scale`.

```suggestion
       configured :attr:`~newton.Model.shape_scale` only affects rendering, defaulting to
```

preist-nvidia (changes requested):

docs/integrations/mujoco.rst
```diff
@@ -64,304 +83,472 @@ Geometry types
...
-
-- **Kinematic roots with non-fixed joints** are exported as ordinary MuJoCo
-  joints with the same Newton joint type and DOFs. The solver assigns a very
-  large internal armature to those DOFs so MuJoCo treats them like prescribed,
-  effectively infinite-mass coordinates.
-- **Roots attached to world with a fixed joint** are exported as MuJoCo mocap
-  bodies. This applies to both kinematic and non-kinematic Newton roots
-  attached to world by :class:`~newton.JointType.FIXED`. MuJoCo has no joint
-  coordinates for a fixed root, so Newton drives the pose through
+  :attr:`newton.State.joint_q` and :attr:`newton.State.joint_qd` values
```
preist-nvidia: nit: missing `~` shortener; rest of the file uses `:attr:` `~newton...` form.

```suggestion
  :attr:`~newton.State.joint_q` and :attr:`~newton.State.joint_qd` values
```

docs/integrations/mujoco.rst
```diff
@@ -64,304 +83,472 @@ Geometry types
...
+- **Roots attached to world with a fixed joint** become MuJoCo mocap
+  bodies (whether kinematic or not). MuJoCo has no joint coordinates
+  for a fixed root, so Newton drives the pose through
   ``mjData.mocap_pos`` and ``mjData.mocap_quat`` instead.
-- **World-attached shapes that are not part of an articulation** remain
-  ordinary static MuJoCo geometry rather than mocap bodies.
+- **World-attached shapes that are not part of an articulation**
+  remain ordinary static MuJoCo geometry rather than mocap bodies.
 
 If you edit :attr:`newton.Model.joint_X_p` or :attr:`newton.Model.joint_X_c`
```
preist-nvidia: nit: missing `~` shortener.

```suggestion
If you edit :attr:`~newton.Model.joint_X_p` or :attr:`~newton.Model.joint_X_c`
```

docs/integrations/mujoco.rst
```diff
@@ -64,304 +83,472 @@ Geometry types
...
+- ``SolverMuJoCo._apply_mjc_control``,
+  ``SolverMuJoCo._update_mjc_data``, and
+  ``SolverMuJoCo._update_newton_state`` — per-step control, data, and
+  state sync between Newton and MuJoCo.
+- :meth:`~newton.solvers.SolverMuJoCo.update_contacts` — explicit pull
+  of MuJoCo's resolved contacts into a Newton ``Contacts`` object
+  (default per-step path does not pull contacts back).
+- :meth:`~newton.solvers.SolverBase.notify_model_changed` —
+  re-synchronize MuJoCo state after editing the Newton ``Model`` (e.g.
+  fixed-root pose changes via ``joint_X_p`` / ``joint_X_c``).
```
preist-nvidia: nit: same symbols are `:attr:` refs four lines above (L512); use refs here too for consistency.

```suggestion
  fixed-root pose changes via :attr:`~newton.Model.joint_X_p` / :attr:`~newton.Model.joint_X_c`).
```

preist-nvidia (approved):

Issue #13, opened 2025-04-29
Implement environment vectorization


preist-nvidia:

eric-heiden:
This will be handled by https://github.com/newton-physics/newton/pull/119

PR #2578, opened 2026-04-24
Add Viewer.log_image for displaying batched sensor output in ViewerGL

### LINKED ISSUE #2232: [REQ] Tiled Camera view in NewtonViewer
### Description

Tiled camera view in NewtonViewer, similar to TiledCamera in IsaacSim, to help monitoring and recording. 

Below is a screenshot of an example of this feature that @daniela-hase shared in Slack:
Image

### Motivation / Use Case

To help improve monitoring environments in Isaac Lab in the Newton Visualizer, particularly when robots are sparsely placed, it'd be great to have a tiled camera view in the Newton Viewer, similar to the TiledCamera in Isaac Sim.

### Alternatives Considered

_No response_

adenzler-nvidia:
## Description

Implements a tiled-camera view for `ViewerGL` matching the Isaac Sim `TiledCamera` UX referenced in the feature request. Adds `Viewer.log_image()` for displaying images — single or batched — as dockable windows in `ViewerGL`, plus `SensorTiledCamera.utils.to_batched_rgba_from_*` adapters that convert raw camera output into the format `log_image` accepts.

Closes #2232.

Sensor Tiled Camera with log_image

## New feature / API change

Users push any image tensor into the viewer under a stable name; each name becomes its own dockable window with a tile grid for batched inputs.

```python
import newton

# ... build scene ...
model = builder.finalize()
viewer.set_model(model)

sensor = SensorTiledCamera(model=model)
W, H, camera_count = 256, 256, 1
color_image = sensor.utils.create_color_image_output(W, H, camera_count)
rgba = wp.empty((model.world_count * camera_count, H, W, 4), dtype=wp.uint8)

# Each frame:
sensor.update(state, camera_transforms, camera_rays, color_image=color_image)
sensor.utils.to_batched_rgba_from_color(color_image, out_buffer=rgba)
viewer.log_image("color", rgba)  # dockable window, one tile per world
```

Four adapters convert sensor output to `(N, H, W, 4)` uint8: `to_batched_rgba_from_color`, `_from_normal`, `_from_depth` (with `near/far` normalization), and `_from_shape_index` (hash palette by default, or a caller-provided RGB lookup for semantic coloring).

The existing `flatten_*_image_to_rgba` methods still work (pre-tiled 2D output for disk writes) but now emit `DeprecationWarning`; viewer use cases should migrate to the batched adapters.

## Accepted formats and why

`log_image` accepts `(H, W)`, `(H, W, C)`, `(N, H, W)`, `(N, H, W, C)` with `C ∈ {1, 3, 4}` and dtype `uint8` (0–255) or `float32` (0–1, clipped). These are the conventions shared across image-logging APIs in the ecosystem, so a user coming from any of them finds the same rules:

- **Channels-last with `C ∈ {1, 3, 4}`**: matches [TensorBoard `tf.summary.image`](https://www.tensorflow.org/api_docs/python/tf/summary/image) (`[batch, H, W, channels]`, channels in `{1, 3, 4}`), [rerun `Image`](https://rerun.io/docs/reference/types/archetypes/image), [matplotlib `imshow`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.imshow.html), and W&B `wandb.Image`. The tensor layout also matches [Isaac Lab `TiledCamera`](https://isaac-sim.github.io/IsaacLab/main/source/api/lab/isaaclab.sensors.html#isaaclab.sensors.TiledCamera) (`(num_envs, H, W, num_channels)`), the direct use case from #2232.
- **Batch dim `(N, H, W, C)` required for multi-image**: mirrors TensorBoard and rerun, where single images are 3D and batches are 4D. For Newton, `N = world_count * camera_count` with world as the slower-changing axis.
- **3D disambiguation via last-axis channel count**: matches matplotlib's `imshow`: a 3D array with last-dim in `{1, 3, 4}` is a single color image `(H, W, C)`; otherwise it's a batch of grayscale frames `(N, H, W)`. Pass a 4D array when the distinction matters.
- **Dtype `uint8`/`float32` with those ranges**: the intersection of conventions across matplotlib (`uint8 ∈ [0, 255]`, `float ∈ [0, 1]`), TensorBoard, PIL, and rerun. No other dtypes are accepted — callers should normalize upstream so the display result is unambiguous.

## Architecture

- **`Viewer.log_image` on `ViewerBase`** is a no-op, matching `log_gaussian` / `log_wireframe_shape`. Concrete backends override.
- **`ViewerGL._image_logger: ImageLogger | None`** is constructed after device + renderer are ready. Initialized to `None` first so `close()` can safely run on a half-constructed viewer.
- **`ImageLogger`** (in `newton/_src/viewer/gl/image_logger.py`) owns per-name GL state (`LoggedImage`: texture, PBO, Warp GL-interop registration). It holds only the viewer's CUDA device, no viewer reference. Two upload paths:
  - **GPU** — same-device CUDA `wp.array` → map PBO via Warp GL-interop → pack kernel writes RGBA bytes → `glTexSubImage2D` from the PBO. Zero host copies.
  - **CPU** — numpy or cross-device/CPU `wp.array` → `_convert_to_packed_rgba_numpy` → `glTexSubImage2D` from host memory.
- **Display**: selected name is drawn as one ImGui window; `compute_grid_layout` picks tile rows/cols that maximize rendered area for the current window aspect. Selection is driven by a sidebar dropdown.
- **Input validation**: `_validate` normalizes any accepted shape to a canonical `(N, H, W, C)` tuple, with error messages that name the problem (dtype, dims, channel count). Called at the top of `ImageLogger.log`.
- **Error handling**: upload failures leave `entry.n/h/w/c` unchanged so the next call starts from a consistent state; `_ensure_pbo` rolls back a half-created GL buffer if `wp.RegisteredGLBuffer` registration fails; `_free_entry` warns rather than silently swallowing (except during interpreter shutdown); cross-device CUDA uploads emit a one-shot warning per `(name, device)` and re-warn when the source device changes.

## Checklist

## Test plan

```
# ImageLogger orchestration, rollback, cleanup, cross-device warnings (stubbed pyglet.gl)
uv run --extra dev -m newton.tests -k test_viewer_image_logger_class

# Pure helpers: compute_grid_layout, _validate, NumPy-vs-Warp kernel parity
uv run --extra dev -m newton.tests -k test_viewer_image_logger

# Sensor adapter kernels + public Utils wrappers
uv run --extra dev -m newton.tests -k test_sensor_to_batched_rgba

# End-to-end integration via the migrated example
uv run --extra dev -m newton.tests -k test_basic.example_sensor_tiled_camera

# Documentation snippet (newly doctest-runnable)
uv run --extra docs --extra sim sphinx-build -j auto -W -b doctest docs docs/_build/doctest
```

Manual:

- [ ] `python -m newton.examples sensor_tiled_camera` — cycle each image stream (color / depth / normal / shape_index / semantic / albedo) from the sidebar dropdown; verify tiles re-layout when resizing the image window.

## Summary by CodeRabbit

* **New Features**
  * Viewer supports image logging for single/batched images with a dockable image panel and selection controls; images accept uint8/float32 and common shapes.
  * New utilities convert sensor outputs (color, normal, depth, shape-index) into batched RGBA for logging.

* **Deprecation**
  * Previous flatten-to-RGBA helpers deprecated; migrate to new batched conversion APIs.

* **Documentation**
  * Visualization guide expanded with logging rules, accepted formats, and usage examples.

* **Examples & Tests**
  * Sensor example logs multiple streams; extensive tests added for conversions and the image logger.


newton/_src/viewer/viewer.py
```diff
@@ -1202,6 +1202,30 @@ def log_gaussian(
...
+                * ``(N, H, W, C)`` -- batch of N color images, ``C in (1, 3, 4)``
+
+                Accepted dtypes: ``uint8`` (values in ``[0, 255]``) or
+                ``float32`` (values in ``[0, 1]``). Values outside the range
+                are clipped.
+
+        The base implementation is a no-op. Backends that render images
+        (currently only :class:`ViewerGL`) override this method.
+        """
+        return
```
adenzler-nvidia, line 1227: as far as the input data goes, I looked up existing viewers (rerun, matplotlib, tensorboard) on what formats they support and it came down to that list. Happy to discuss this - we can add more overloads or tighten the accepted data as needed.

newton/_src/sensors/warp_raytrace/utils.py
```diff
@@ -370,6 +520,196 @@ def flatten_color_image_to_rgba(
         )
         return out_buffer
 
+    def to_batched_rgba_from_color(
```
mmacklin: Hmm, I think this can be implemented as a reshape + view operation rather than copy can't it?

Something like:

`return image.view(wp.vec4ub).reshape((n, h, w)).view(wp.uint8)`

As long as the input is contiguous.

newton/examples/sensors/example_sensor_tiled_camera.py
```diff
@@ -271,13 +233,30 @@ def render_sensors(self):
...
             color_image=self.tiled_camera_sensor_color_image,
+            albedo_image=self.tiled_camera_sensor_albedo_image,
             depth_image=self.tiled_camera_sensor_depth_image,
             normal_image=self.tiled_camera_sensor_normal_image,
             shape_index_image=self.tiled_camera_sensor_shape_index_image,
-            albedo_image=self.tiled_camera_sensor_albedo_image,
             clear_data=SensorTiledCamera.GRAY_CLEAR_DATA,
         )
-        self.update_texture()
+        utils = self.tiled_camera_sensor.utils
```
mmacklin, line 248: Not directly related to this PR but I kinda feel like the methods on this `utils` class should just be on the `SensorTiledCamera` class itself as static methods.

newton/examples/sensors/example_sensor_tiled_camera.py
```diff
@@ -298,161 +277,36 @@ def get_camera_transforms(self) -> wp.array[wp.transformf]:
             dtype=wp.transformf,
         )
 
-    def create_texture(self):
```
mmacklin, line 306: Great to see all this stuff coming out of the example code.

newton/_src/sensors/warp_raytrace/utils.py
adenzler-nvidia: Good catch — done in 32ebaa4. `to_batched_rgba_from_color` is now exactly what you suggested:

```python
return image.view(wp.vec4ub).reshape((n, h, w)).view(wp.uint8)
```

Zero-copy view of the input. Eliminates a redundant full-image GPU pass per frame. Dropped the `out_buffer` parameter since it had no allocation to avoid anymore. The other `to_batched_rgba_from_*` methods keep `out_buffer` — they still launch real kernels.

newton/examples/sensors/example_sensor_tiled_camera.py
daniela-hase: I personally would prefer to keep them in the utils class, keeps them cleanly separated from the rest of the sensor since they are really just utility functions. Makes auto-complete in IDEs more structured I think.

newton/_src/sensors/warp_raytrace/utils.py
```diff
@@ -337,11 +456,23 @@ def flatten_color_image_to_rgba(
...
+            pre-tiled 2D image, reshape the batched output yourself
+            (``batched.reshape(rows, cols, H, W, 4).transpose(0, 2, 1, 3, 4).reshape(rows*H, cols*W, 4)``).
+
         Args:
-            image: Color output from :meth:`~SensorTiledCamera.update`, shape ``(world_count, camera_count, height, width)``.
+            image: Color output from :meth:`~newton.sensors.SensorTiledCamera.update`, shape ``(world_count, camera_count, height, width)``.
             out_buffer: Pre-allocated RGBA buffer. If None, allocates a new one.
             worlds_per_row: Tiles per row in the grid. If None, picks a square-ish layout.
         """
+        warnings.warn(
```
daniela-hase: Why are we deprecating the flatten-functions?
What if someone wants to flatten the images without using Newton's viewer?

newton/_src/sensors/warp_raytrace/utils.py
```diff
@@ -370,6 +501,184 @@ def flatten_color_image_to_rgba(
         )
         return out_buffer
 
+    def to_batched_rgba_from_color(
```
daniela-hase: I am not a fan of the `to_batched_*` naming. Why "batched"?

I'd prefer to name them `convert_rgba_to_uint8` or similar

newton/_src/sensors/warp_raytrace/utils.py
adenzler-nvidia: Good point — those still serve a different use case (writing a single pre-tiled image to disk, no viewer needed). Dropped the deprecation in 50d6caf5; both `flatten_*_to_rgba` and the new viewer-oriented adapters are now first-class. Updated the docstrings to point at the right tool for each use case.

newton/_src/sensors/warp_raytrace/utils.py
adenzler-nvidia: Renamed to `to_rgba_from_color` / `_normal` / `_depth` / `_shape_index` in 50d6caf5. You're right that "batched" was redundant — the input is already batched along `(world_count, camera_count)`, the function just collapses those into a single `N` axis on the output. The new name describes what it does (sensor format → RGBA) without the noise.

newton/_src/sensors/warp_raytrace/utils.py
```diff
@@ -370,6 +493,180 @@ def flatten_color_image_to_rgba(
...
+                contiguous; arrays returned by
+                :meth:`~newton.sensors.SensorTiledCamera.update` always satisfy this.
+
+        Returns:
+            Array of shape ``(world_count * camera_count, H, W, 4)``,
+            dtype ``uint8``, aliasing *image*.
+        """
+        world_count, camera_count, h, w = image.shape
+        n = world_count * camera_count
+        return image.view(wp.vec4ub).reshape((n, h, w)).view(wp.uint8)
```
daniela-hase, line 530: I wonder if it would be better to return the arrays as 5d, instead of 4d - so they keep the original shape and just split the packed value into 8bit rgba.

As it is now, these functions also change the shape from `world, camera, height, width` to `world * camera, height, width, 4`.

If the viewer downstream needs it in that other shape, it could just reshape it itself.

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@
...
 - Add fast parity-based SDF construction path for watertight meshes in `SDF.create_from_mesh`, using `wp.mesh_query_point_sign_parity` instead of winding numbers; selected via the new `sign_method` argument (`"auto"` — the default — picks parity when `Mesh.is_watertight` is true, or `"parity"` / `"winding"` to force either strategy)
+- Add `Viewer.log_image()` for displaying single or batched images in `ViewerGL`; other backends inherit a no-op. Also add `SensorTiledCamera.utils.to_rgba_from_color()`, `to_rgba_from_normal()`, `to_rgba_from_depth()`, and `to_rgba_from_shape_index()` (hash palette or caller-provided RGB lookup) adapters producing output consumable by `log_image()`.
+- Enable full CPU execution of the collision pipeline, including mesh–mesh and mesh–heightfield SDF contacts that were previously CUDA-only; contact reduction (`reduce_contacts`) also runs on CPU. The previous `"NarrowPhase running on CPU: mesh-mesh contacts will be skipped"` warning is no longer emitted — CPU runs now execute the same kernels as CUDA.
```
eric-heiden: This isn't part of this PR

newton/_src/sensors/warp_raytrace/utils.py
```diff
@@ -145,14 +145,134 @@ def convert_ray_depth_to_forward_depth_kernel(
...
+        out[n, y, x, 0] = wp.uint8(0)
+        out[n, y, x, 1] = wp.uint8(0)
+        out[n, y, x, 2] = wp.uint8(0)
+        out[n, y, x, 3] = wp.uint8(255)
+        return
+    denom = far - near
+    if denom <= 0.0:
+        t = wp.float32(0.0)
+    else:
+        t = wp.clamp((d - near) / denom, 0.0, 1.0)
```
eric-heiden: 🔴 `clear_depth` still defaults to `0.0`, so untouched depth pixels reach this normalization path and get clamped as the nearest possible hit. That makes no-hit pixels render near-white instead of black, which is a behavioral regression from the old flatten path.

Please treat the sensor clear-depth sentinel as a miss here too, not only negative depths. For example, handle `d <= 0.0` as black before computing `t`, or compare against the configured clear-depth value if that is available in this adapter.

newton/_src/viewer/gl/image_logger.py
```diff
@@ -0,0 +1,658 @@
...
+
+        tex_w = w
+        tex_h = n * h
+
+        if self._can_use_gpu_path(image):
+            src_4d = _to_canonical_4d_wp(image, n, h, w, c)
+            self._upload_gpu(entry, src_4d, n, h, w, c, tex_w, tex_h, needs_realloc)
+        else:
+            self._maybe_warn_cross_device(name, image)
+            self._upload_cpu(entry, image, n, h, w, c, tex_w, tex_h, needs_realloc)
```
eric-heiden: 🔴 The current packing always uploads the batch as one vertical strip with `tex_h = n * h`. That can exceed `GL_MAX_TEXTURE_SIZE` for moderate batches even when the same images would fit in a 2D atlas, for example `65 x 256 x 256` becomes a `256 x 16640` texture.

Please pack batched images into a grid or atlas before allocating the GL texture, and size the texture from that atlas layout instead of a single strip.

newton/_src/sensors/warp_raytrace/utils.py
```diff
@@ -370,6 +493,180 @@ def flatten_color_image_to_rgba(
         )
         return out_buffer
 
+    def to_rgba_from_color(
+        self,
+        image: wp.array4d[wp.uint32],
+    ) -> wp.array4d[wp.uint8]:
+        """Reinterpret packed ``uint32`` RGBA color sensor output as ``uint8`` RGBA.
```
eric-heiden, line 506: 🔴 `worlds_per_row=0` now raises `ValueError` in the shared flatten helper, but this helper backs the public `SensorTiledCamera.Utils.flatten_*_image_to_rgba` API. Before this change, `0` took the same auto-layout path as `None` because the code used a falsy check.

Please preserve the existing `0` behavior for now, preferably with a `DeprecationWarning` that asks callers to pass `None` for auto layout, then reject `0` in a later release if needed.

newton/_src/sensors/warp_raytrace/utils.py
```diff
@@ -370,6 +493,180 @@ def flatten_color_image_to_rgba(
...
+            Array of shape ``(world_count * camera_count, H, W, 4)``, dtype
+            ``uint8``. Suitable for :meth:`~newton.viewer.ViewerBase.log_image`.
+        """
+        world_count = image.shape[0]
+        camera_count = image.shape[1]
+        h = image.shape[2]
+        w = image.shape[3]
+        n = world_count * camera_count
+
+        if depth_range is None:
```
eric-heiden, line 603: 🟡 Defaulting `depth_range=None` to `(0.0, 10.0)` loses contrast for scenes beyond 10m. Depths like `20..50m` all clamp to the same luma value, while the previous auto-range behavior preserved visible differences.

Please keep the old auto-range behavior when `depth_range is None`, or require callers that migrate from the flatten helper to pass an explicit range and document that requirement at the call sites.

newton/_src/sensors/warp_raytrace/utils.py
adenzler-nvidia: Good catch — fixed in e159c7ed8. `unpack_depth_to_rgba_kernel` now treats `d <= 0.0` as miss (matches the `ClearData.clear_depth = 0.0` sentinel and the legacy `flatten_depth_image` `if depth > 0` gate). Added a regression test in `test_depth_normalizes_to_grayscale` that pins a `d=0.0` input to black.

newton/_src/sensors/warp_raytrace/utils.py
adenzler-nvidia: Fixed in e159c7ed8. `__reshape_buffer_for_flatten` now keeps `worlds_per_row=0` on the auto-layout path behind a `DeprecationWarning` asking callers to pass `None`. `< 0` still raises. Added `test_worlds_per_row_zero_warns_and_auto_layouts` to verify the warning fires and the output shape matches `worlds_per_row=None`.

newton/_src/sensors/warp_raytrace/utils.py
adenzler-nvidia: Fixed in e159c7ed8. `to_rgba_from_depth(depth_range=None)` now launches `find_depth_range` on device to compute a tight per-frame `[min(d>0), max(d>0)]` range, matching `flatten_depth_image_to_rgba`. To make the auto-range path work without a host sync, I also changed `unpack_depth_to_rgba_kernel` to take a 2-element `wp.array[wp.float32]` instead of scalar `near, far` parameters, so the reduction result feeds straight into the unpack kernel.

newton/_src/viewer/gl/image_logger.py
adenzler-nvidia: Fixed in 6b045a552. ``log_image`` now packs the batch as a square-ish atlas (`cols = ceil(sqrt(N))`, `rows = ceil(N/cols)`, `tex_w = cols*W`, `tex_h = rows*H`) instead of a 1xN vertical strip. The pack kernels and the NumPy reference both take `atlas_cols` so the GPU and CPU paths stay byte-identical, `draw()` samples each tile via 2D UVs derived from its slot, and trailing slots in the last row are zero-filled (GPU path zeroes the mapped PBO on tile-shape change). Your 65 x 256 x 256 example now sizes to a 9x8 atlas of 2304 x 2048 instead of 256 x 16640. Added `TestAtlasLayout` with a regression assertion for that case.

CHANGELOG.md
adenzler-nvidia: Fixed in 75a132118. The previous upstream merge had reworded the `Enable CPU execution of the collision pipeline` entry on the PR side, leaving two near-identical bullets in `[Unreleased]/Added`. Re-merged upstream/main and dropped the local rewrite so this PR's only delta against `upstream/main` is now the `Viewer.log_image()` line.

newton/_src/sensors/warp_raytrace/utils.py
adenzler-nvidia: 5D isn't reachable with Warp arrays — `wp.array5d` doesn't exist (only `array2d`/`3d`/`4d`) and `wp.zeros((2,2,2,2,2), dtype=wp.uint8)` raises `RuntimeError: Failed to create array with shape (2, 2, 2, 2, 2), the maximum number of dimensions is 4`. So preserving the original `(world, camera, H, W)` axes plus an RGBA channel axis would have to fall back to NumPy or a flat byte buffer with manual indexing, neither of which composes with Warp kernels.

Given that hard 4D cap, `(N, H, W, 4)` (with `N = world * camera`) is the cheapest option that stays a Warp array end-to-end and matches what `Viewer.log_image()` and every comparable image-logging API in the ecosystem (TensorBoard `tf.summary.image`, rerun `Image`, Isaac Lab `TiledCamera`, matplotlib `imshow`, W&B `wandb.Image`) consume. Callers who need the 4D `(world, camera, H, W)` view of the *source* images can keep using the sensor's existing per-stream outputs; this adapter is specifically for the viewer's input shape.

newton/_src/sensors/warp_raytrace/utils.py
```diff
@@ -145,14 +145,151 @@ def convert_ray_depth_to_forward_depth_kernel(
...
+    world, camera, y, x = wp.tid()
+    camera_count = image.shape[1]
+    n = world * camera_count + camera
+    idx = image[world, camera, y, x]
+    # Knuth multiplicative hash, masked to 24 bits.
+    h = (idx * wp.uint32(2654435761)) & wp.uint32(0xFFFFFF)
+    out[n, y, x, 0] = wp.uint8((h >> wp.uint32(16)) & wp.uint32(0xFF))
+    out[n, y, x, 1] = wp.uint8((h >> wp.uint32(8)) & wp.uint32(0xFF))
+    out[n, y, x, 2] = wp.uint8(h & wp.uint32(0xFF))
+    out[n, y, x, 3] = wp.uint8(255)
```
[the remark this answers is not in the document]
adenzler-nvidia: Already addressed in an earlier revision: `unpack_shape_index_hash_to_rgba_kernel` now hashes `(idx + 1) * 2654435761` so shape index 0 gets a non-zero color, and the miss sentinel `0xFFFFFFFF` wraps back to 0 (which intentionally renders black). See `newton/_src/sensors/warp_raytrace/utils.py:215-222` and the regression test `test_shape_index_hash_zero_is_not_black` in `newton/tests/test_sensor_to_rgba.py`.

newton/_src/sensors/warp_raytrace/utils.py
```diff
@@ -145,14 +145,151 @@ def convert_ray_depth_to_forward_depth_kernel(
...
+    expected_device: wp.Device,
+) -> None:
+    """Raise ``ValueError`` if *out_buffer* is not a canonical RGBA sink."""
+    if tuple(out_buffer.shape) != expected_shape:
+        raise ValueError(f"{name}: out_buffer shape {tuple(out_buffer.shape)} does not match expected {expected_shape}")
+    if out_buffer.dtype != wp.uint8:
+        raise ValueError(f"{name}: out_buffer dtype must be wp.uint8, got {out_buffer.dtype}")
+    if out_buffer.device != expected_device:
+        raise ValueError(f"{name}: out_buffer is on {out_buffer.device} but input is on {expected_device}")
+
```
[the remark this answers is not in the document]
adenzler-nvidia: Already addressed: `Any` is imported at `newton/_src/sensors/warp_raytrace/utils.py:8` (`from typing import TYPE_CHECKING, Any`). `typing.get_type_hints()` and Sphinx autodoc resolve cleanly. Ruff (`uvx pre-commit run -a`) passes.

newton/_src/viewer/viewer_gl.py
```diff
@@ -228,10 +231,15 @@ def __init__(
...
 
+        # Initialized below once self.device is available; declared here so
+        # close() can safely run if __init__ raises before that point.
+        self._image_logger: ImageLogger | None = None
+
         super().__init__()
 
         self.renderer = RendererGL(vsync=vsync, screen_width=width, screen_height=height, headless=headless)
         self.renderer.set_title("Newton Viewer")
+        self._image_logger = ImageLogger(device=self.device, sidebar_width_px=_SIDEBAR_WIDTH_PX)
```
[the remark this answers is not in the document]
adenzler-nvidia: Already addressed: `ViewerGL.set_model()` rebinds the image logger if `self.device` changed during the `super().set_model()` call. See `newton/_src/viewer/viewer_gl.py:506-511`:

```python
# ``ViewerBase.set_model`` may have switched ``self.device`` to the
# model's device. Rebind the image logger so its GPU path tests against
# — and registers PBO interop with — the correct CUDA context.
if self._image_logger is not None and self._image_logger.device != self.device:
    self._image_logger.clear()
    self._image_logger = ImageLogger(device=self.device, sidebar_width_px=_SIDEBAR_WIDTH_PX)
```

newton/examples/sensors/example_sensor_tiled_camera.py
```diff
@@ -230,13 +181,26 @@ def __init__(self, viewer: ViewerGL, args):
...
         )
-        self.tiled_camera_sensor_albedo_image = self.tiled_camera_sensor.utils.create_albedo_image_output(
-            self.sensor_render_width, self.sensor_render_height, self.camera_count
+
+        # Palette for the "semantic" debug view: looked up by shape index.
+        self.semantic_palette = wp.array(
+            np.asarray(semantic_colors, dtype=np.uint8),
+            dtype=wp.uint8,
+            device=self.tiled_camera_sensor_color_image.device,
         )
```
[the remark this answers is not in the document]
adenzler-nvidia: Already addressed in the example: there's an explicit length check before constructing the palette to catch the misalignment you describe. See `newton/examples/sensors/example_sensor_tiled_camera.py:185-190`:

```python
# Palette for the "semantic" debug view: looked up by shape index.
# Indices written into shape_index_image come from builder shape order,
# so the palette must be one entry per shape in that same order.
assert len(semantic_colors) == self.model.shape_count, (
    f"semantic_colors out of sync with model: {len(semantic_colors)} vs {self.model.shape_count}"
)
```

If builder shape ordering ever changes silently this assertion still wouldn't catch a *reordering* (only a length mismatch), but the only way to fix that fully is the `colors` palette plumbed through `to_rgba_from_shape_index` — which is what the example demonstrates.

newton/_src/sensors/warp_raytrace/utils.py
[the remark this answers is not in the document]
adenzler-nvidia: There's only one hash kernel in the current branch:

```
$ grep -rn 'unpack_shape_index_hash' newton/
newton/_src/sensors/warp_raytrace/utils.py:212:def unpack_shape_index_hash_to_rgba_kernel(
newton/_src/sensors/warp_raytrace/utils.py:665:                unpack_shape_index_hash_to_rgba_kernel,
newton/tests/test_sensor_to_rgba.py:189: ...
```

`unpack_shape_index_hash_to_batched_rgba_kernel` was the original name in commit ac26e4f60; it was renamed (no semantic change) to `unpack_shape_index_hash_to_rgba_kernel` in commit 50d6caf5 ("Rename to_rgba_from_*; keep flatten_*_to_rgba public") in response to daniela-hase's [comment #13](https://github.com/newton-physics/newton/pull/2578#discussion_r3156007514). I think your diff base predates that rename, so the tool is treating the renamed kernel as a new addition without the +1 bias and the original (which always had it) as separate.

Current kernel body, with the +1 bias intact:

```python
# newton/_src/sensors/warp_raytrace/utils.py:221-224
# Knuth multiplicative hash, masked to 24 bits. ``idx + 1`` keeps shape 0
# away from the all-zero hash that collides with the miss color; the
# miss sentinel ``0xFFFFFFFF`` wraps back to 0 and intentionally renders black.
h = ((idx + wp.uint32(1)) * wp.uint32(2654435761)) & wp.uint32(0xFFFFFF)
```

No action needed.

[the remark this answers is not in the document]
adenzler-nvidia: Thanks for the quick turnaround on dismissing — appreciate it.

eric-heiden (approved):
Thanks!

PR #2582, opened 2026-04-25
Fix MuJoCo geom_dataid indexing for mesh geoms

han-xudong:
## Description

Fix MuJoCo mesh geom property updates to use the correct `geom_dataid`
layout from `mujoco_warp`.

`mujoco_warp` stores `geom_dataid` as a model-level 1-D array, but
Newton's bridge kernel treated it as a per-world 2-D array. For mesh
geoms, that can read the wrong mesh id during property updates.

This PR aligns the kernel signature and indexing with the actual
`mujoco_warp` model layout and adds a changelog entry for the fix.

## Checklist

## Test plan

```bash
uvx pre-commit run -a
uv run python -m unittest newton/tests/test_body_velocity.py
uv run python -m unittest newton/tests/test_examples.py
```

## Bug fix

**Steps to reproduce:**

1. Run a MuJoCo-backed scene that includes mesh geoms.
2. Trigger the bridge path that updates MuJoCo geom properties.
3. Observe that `geom_dataid` is indexed as `[world, geom]` even though
   `mujoco_warp` stores it as a model-level 1-D array.

**Minimal reproduction:**

```python
import newton

model = builder.finalize()
solver = SolverMuJoCo(model)
```

## New feature / API change

Not applicable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed mesh geometry property updates in MuJoCo simulations where incorrect mesh IDs could be selected due to improper indexing. Mesh lookup now follows model-level semantics, ensuring geometries receive the correct mesh assignment. This improves visual/physical consistency of mesh geoms, prevents mis-assigned meshes during simulation updates, and reduces related selection errors and instability.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
 
 - Fix Sphinx docs builds to auto-discover bundled ``pypandoc_binary`` pandoc so notebook tutorials build without manual PATH configuration
 - Fix viewer crash with `imgui_bundle>=1.92.6` when editing colors by normalizing `color_edit3` input/output in `_edit_color3`
+- Fix MuJoCo mesh geom property updates indexing `geom_dataid` as a per-world array instead of the model-level array used by `mujoco_warp`, which could select the wrong mesh id for mesh geoms
 - Show prismatic joints in the GL viewer when "Show Joints" is enabled
 - Fix connect constraint anchor computation to account for joint reference positions when `SolverMuJoCo` is the chosen solver.
 - Fix `SolverMuJoCo` passing non-zero geom/pair margins to `mujoco_warp.put_model()`, which fails when NATIVECCD is enabled. Margins are forced to zero when MuJoCo handles collisions (`use_mujoco_contacts=True`); the Newton collision pipeline (`use_mujoco_contacts=False`) is unchanged
--- a/newton/_src/solvers/mujoco/kernels.py
+++ b/newton/_src/solvers/mujoco/kernels.py
@@ -2244,7 +2244,7 @@ def update_geom_properties_kernel(
     mjc_geom_to_newton_shape: wp.array2d[wp.int32],
     geom_type: wp.array[int],
     GEOM_TYPE_MESH: int,
-    geom_dataid: wp.array2d[int],
+    geom_dataid: wp.array[int],
     mesh_pos: wp.array[wp.vec3],
     mesh_quat: wp.array[wp.quat],
     shape_mu_torsional: wp.array[float],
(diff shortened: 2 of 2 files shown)


adenzler-nvidia:
Thanks for the PR!

This change came with the 3.7.0 `mujoco_warp` upgrade — see commit bffb57951 ("Update geom_dataid to 2D for mujoco_warp 3.7.0 compatibility"), which landed alongside the 3.7.0 bump in #2455. In 3.7.0, `mujoco_warp` made meshes randomizable per world and changed `geom_dataid` from `(ngeom,)` to `(nworld, ngeom)`. The current 2-D indexing in the kernel matches that layout; this PR would re-break it.

j3soon:
Encountered the same issue, it's due to forgetting to re-install and update dependencies after pulling from main:

```
$ git pull
$ uv pip install -e ".[examples]"
...
 - mujoco==3.6.0
 + mujoco==3.7.0
 - mujoco-warp==3.6.0
 + mujoco-warp==3.7.0.1
 ~ newton==1.2.0.dev0 (from file:///home/johnson/projects/newton)
 - newton-usd-schemas==0.1.0
 + newton-usd-schemas==0.2.0rc1
```

The issue disappears after updating the dependencies.

adenzler-nvidia:
how hard did this bite you - should we warn when initializing SolverMuJoCo if there is a version mismatch?

han-xudong:
> how hard did this bite you - should we warn when initializing SolverMuJoCo if there is a version mismatch?

@adenzler-nvidia It was a local version mismatch on my side after pulling `main` without reinstalling dependencies. Once I updated `mujoco` / `mujoco-warp`, the issue disappeared.

So it was confusing, but only because my environment was stale rather than because of a bug in the current code. A warning for clearly incompatible versions during `SolverMuJoCo` init could still be helpful for this exact failure mode, especially in editable installs.

j3soon:
> how hard did this bite you - should we warn when initializing SolverMuJoCo if there is a version mismatch?

@adenzler-nvidia I encountered the exact same issue with identical steps leading to the issue. I initially `pip install -e ".[examples]"` an earlier version, and then some time after I pulled latest main, but didn't re-run `pip install` or update dependencies.

When observing the error in the `mpm_twoway_coupling` example (I recall that `mpm_viscous` seems to work fine), I rolled back to v1.1.0 and confirmed the example works in that version. I've bisected the issue with the help of Claude Code and found the commit a0fadced introduced the change, and then Claude Code proposed a similar fix in this PR, making the example working again. While I'm about to submit an issue report, I looked up existing issues/PRs and stumble across this PR. (thanks @han-xudong!)

I should've think of even with `pip install -e .`, I should still re-install dependencies after updating code, but this is easily overlooked. So yes, adding a warning would definitely help in this case.

PR #2583, opened 2026-04-25
Adding test viewer mixin class and refactor hydroelastic tests

Alex-Omar-Nvidia:
## Description

This MR adds a mixin class that provides a simple viewer framework for unit testing. This makes it simpler for new tests to implement viewers.

This also refactors `test_hydroelastic.py` to use this new framework as a demonstration.

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Consolidated the viewer-based hydroelastic test harness into a reusable mixin-driven flow and a scene abstraction, reducing duplication and exposing sensible default viewer parameters.

* **Tests**
  * Added utilities to package simulation scenes and run frame-by-frame viewer tests with stable state swapping, per-frame hooks, optional extra logging, graceful run/skip behavior, and interrupt handling.

### DIFF
--- a/newton/tests/test_hydroelastic.py
+++ b/newton/tests/test_hydroelastic.py
@@ -1,7 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 The Newton Developers
 # SPDX-License-Identifier: Apache-2.0
 
-import time
 import unittest
 from enum import Enum
 
@@ -14,6 +13,7 @@ from newton.tests.unittest_utils import (
     add_function_test,
     get_selected_cuda_test_devices,
 )
+from newton.tests.viewer_test_utils import ViewerScene, ViewerTestMixin
 
 # --- Configuration ---
 
@@ -1063,7 +1063,10 @@ def test_mujoco_hydroelastic_penetration_depth(test, device):
 # --- Test class ---
 
 
-class TestHydroelastic(unittest.TestCase):
+class TestHydroelastic(ViewerTestMixin, unittest.TestCase):
+    DEFAULT_NUM_FRAMES = VIEWER_NUM_FRAMES
+    DEFAULT_SIM_DT = SIM_DT
+
     @unittest.skip("Visual debugging - run manually to view simulation")
     def test_view_stacked_primitive_cubes(self):
         """View stacked primitive cubes simulation with hydroelastic contacts."""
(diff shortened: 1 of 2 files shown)


Alex-Omar-Nvidia:
Please let me know if I should update the changelog. I wasn't sure because this is just a testing feature, not really a newton feature.

adenzler-nvidia:
Throwing an alternative shape into the conversation in case it's useful — not a request to change this PR, just to compare designs.

**Sketch: a `NewtonTestCase` base class.** The test keeps its own loop and data flow; the base contributes one indirection and a small convention:

- `for _ in self.frames(N):` instead of `range(N)`. In headless mode it's `range(N)`; in viewer mode it drives a render loop.
- Convention that handles live on `self` (`self.model`, `self.state_0`, `self.contacts`, …) so the base can find them without per-test viewer calls.
- `_on_frame(viewer)` override for scene-specific overlays.
- Layered enable hooks: `NEWTON_TEST_VIEWER=1` env var, `ENABLE_VIEWER` class attr, `@view` decorator. Env var wins so CI can force off.
- `ViewerNull` is the headless backend, so the same code path runs in both modes.

Tests in this shape read as ordinary tests — no viewer imports, no callbacks, no skipped-by-default methods. Visualization is a flag flip, not a code change.

**The reason a base class is interesting beyond the viewer.** Once it exists, it's the natural place to enforce things uniformly across the suite:

- Wrap each test in `CheckOutput` so stray `wp.printf` / unexpected stdout fails the test.
- Device-aware skip helpers (`self.skipIfNoCuda()`, …).
- `tearDown` sanity (no NaN states, no leaked GPU allocations).

The viewer is one tenant of that base; the others are independent and can land incrementally.

Curious what folks think of the broader direction — fully aware this would be a much larger change than what's in flight here.

Alex-Omar-Nvidia:
I see this just got pulled into the 1.4 release. I currently have my hands full with some pose estimation work. Can I hand this off to another engineer?

This PR was a good path-finding example but I think a large rewrite would be best in order to incorporate @adenzler-nvidia's recommendations.

+cc @preist-nvidia for viz

adenzler-nvidia:
yes we can take this over - as this was mostly explorative work anyway. I'll leave this one open until I decide what direction to go, and the main test base class PR is merged.

Issue #1373, opened 2026-01-15
Supporting different per-world time-step and gravity for heterogeneous models.


vastsoun:
Currently gravity is represented in `newton.Model` as a single-element array so that it is uniformly applied to all worlds, and the `SolverBase.step(...)` function accepts a single scalar value for the `dt` argument.

To work towards supporting heterogeneous models, which the current implementation of Kamino already supports, we should consider mechanism to facilitate this.

## Time-step
Two reasonable options for heterogeneous time-steps would be: 
- change the function signature of `SolverBase.step(...)` to:
```python
def step(
        self, 
        state_in: State, 
        state_out: State, 
        control: Control, 
        contacts: Contacts, 
        dt: float | wp.array
    ):
        """
        Simulate the model for a given time step using the given control input.

        Args:
            state_in (State): The input state.
            state_out (State): The output state.
            control (Control): The control input.
                Defaults to `None` which means the control values from the
                :class:`Model` are used.
            contacts (Contacts): The contact information.
            dt (float | wp.array): The time step (typically in seconds).
        """
        raise NotImplementedError()
```

- or introduce an array of per-world  time-steps to the `newton.Model`.

## Gravity
I think this one makes sense to handle directly in the `newton.ModelBuilder` by introducing an appropriate set of functions, that result in making the `newton.ModelBuilder.gravity` attribute array a per-world array instead of a single element.

Moreover, currently in Kamino, per-world gravity is represented by the `kamino.Model.gravity.vector` member attribute that is an array of `shape=(num_worlds,), dtype=vec4f`. The first three elements are the 3D `(g_x, g_y, g_z)` vector of gravity direction x gravity acceleration, and the fourth element is a `0.0/1.0` value indicating if gravity is enabled, so all computations effectively use gv.w * vec3f(gv.x, gv.y, gv.z) for the final value. We did this because we wanted to support an easy per-world gravity toggle without losing the current gravity settings.

eric-heiden:
Thanks @vastsoun! The gravity part I think is addressed now by https://github.com/newton-physics/newton/pull/1381, can you confirm?
Having a per-world timestep is a good feature. We will discuss whether it makes sense to add this the Newton Model rather than passing it in to the step() function of the solver...

vastsoun:
[...]

PR #2584, opened 2026-04-25
Fix MuJoCo timestep option import

han-xudong:
## Description

Fix MuJoCo MJCF `<option timestep="...">` import so per-world timesteps
are preserved when constructing `SolverMuJoCo`.

This change:

- registers `mujoco:option:timestep` as a WORLD-frequency custom
  attribute
- resolves that attribute during `SolverMuJoCo` conversion
- propagates the value to both `mj_model.opt.timestep` and
  `mujoco_warp`'s per-world `opt.timestep`
- adds focused importer and solver regressions for timestep parsing and
  propagation

## Checklist

## Test plan

```bash
uvx pre-commit run -a
uv run python -m unittest newton.tests.test_import_mjcf.TestImportMjcfSolverParams.test_option_scalar_world_parsing newton.tests.test_import_mjcf.TestImportMjcfSolverParams.test_option_scalar_per_world newton.tests.test_mujoco_solver.TestMuJoCoSolver.test_timestep_option_from_mjcf
```

## Bug fix

**Steps to reproduce:**

1. Import an MJCF model that sets `<option timestep="...">`.
2. Finalize the model and inspect `model.mujoco`.
3. Construct `SolverMuJoCo` from that model.
4. Observe that the imported timestep is not preserved, so solver setup
   falls back to the default timestep instead of the MJCF-authored
   value.

**Minimal reproduction:**

```python
import newton

robot_a = newton.ModelBuilder()
robot_a.add_mjcf("""
<mujoco>
    <option timestep="0.002"/>
    <worldbody>
        <body name="a" pos="0 0 1">
            <joint type="hinge" axis="0 0 1"/>
        </body>
    </worldbody>
</mujoco>
""")

robot_b = newton.ModelBuilder()
robot_b.add_mjcf("""
<mujoco>
    <option timestep="0.005"/>
    <worldbody>
        <body name="b" pos="0 0 1">
            <joint type="hinge" axis="0 0 1"/>
        </body>
    </worldbody>
</mujoco>
""")

builder.add_world(robot_a)
builder.add_world(robot_b)
model = builder.finalize()
solver = SolverMuJoCo(model, disable_contacts=True)

print(model.mujoco.timestep.numpy())
print(solver.mjw_model.opt.timestep.numpy())
```

## New feature / API change

Not applicable.

## Summary by CodeRabbit

* **Documentation**
  * Updated the changelog to clarify that MJCF `<option timestep="...">` preserves per-world MuJoCo timesteps, and `step(..., dt=None)` uses them while an explicit `dt` overrides.
* **New Features**
  * Added MuJoCo WORLD-frequency `timestep` support with proper per-world resolution and multi-world expansion.
* **Bug Fixes**
  * Ensured solver-option synchronization propagates per-world MJCF timesteps into solver reporting and runtime stepping behavior.
* **Tests**
  * Extended MJCF import and MuJoCo solver tests to verify per-world timestep values, ordering, and `dt` override semantics.


adenzler-nvidia:
Thanks for the contribution - the implementation looks correct on a first glance. However, we intentionally did not implement this yet because it's unclear how it will interact with the API of calling step() with a dt, and how it would work for per-world dts. It becomes a bit weird with core newton in the sense that worlds don't really exist, it's just collision filtering and constraint partitioning, but mujoco then has the explicit concept.

I will put this discussion back on the map because you are not the first one needing this - we should have that API discussion.

han-xudong:
@adenzler-nvidia Thanks for the clarification. That makes sense to me.

I agree this touches a broader API design question around how a MuJoCo-side timestep override should interact with `step(dt)` and what the expected semantics should be for per-world timesteps. Given that this is still unresolved at the API level, I’ll stop here and not push this further as an implementation-only change.

Thanks again for taking a look and for putting the discussion back on the map.

jcarius-nv:
@adenzler-nvidia how do you suggest we proceed here to come to a conclusion? I'll backlog it for now.

adenzler-nvidia:
This needs an architecture discussion with all parties involved - I know Disney also had requests for this.

jcarius-nv:

han-xudong:
Thanks for the context here. I took another look at the current `SolverMuJoCo.step()` behavior and the older discussion in #1373.

@adenzler-nvidia @dsmither-lgtm @jcarius-nv, it looks like the remaining blocker is not the MJCF parsing itself, but the ownership/precedence of timestep values: this PR can import per-world `<option timestep="...">` values into the model/MuJoCo option state, but the current `step(..., dt)` contract still treats the scalar `dt` as an explicit runtime value and overwrites `mj_model.opt.timestep` / `mjw_model.opt.timestep` on every step.

Before I update this PR further, could you confirm which direction you would prefer?

1. Keep `step(dt)` as the explicit override, and narrow this PR to importing/storing the MJCF timestep while documenting/testing that normal `step(dt)` calls override it.
2. Move toward model-owned/per-world timestep semantics, e.g. allow MuJoCo to step from stored per-world timestep values when no explicit scalar override is requested. This would likely be a broader API decision related to #1373.

Once the intended precedence is clear, I can update the implementation accordingly. In the meantime, I will bring the branch up to date with latest `upstream/main`, move the changelog entry to the current `[Unreleased]` section if needed, and handle the float32 timestep assertion nit.

han-xudong:
@adenzler-nvidia @jcarius-nv @camevor I pushed an update for the timestep precedence issue raised by CodeRabbit and merged the latest upstream/main.

The branch now preserves the model/per-world timestep when `SolverMuJoCo.step(..., dt=None)` is used, while an explicit `dt` still overrides both CPU and MJWarp timestep state. I also added regression coverage for the preserve-vs-override behavior and clarified the changelog entry.

Verified locally with:
- `uv run --extra dev -m newton.tests -k timestep`
- `uvx pre-commit run -a`

Could you please take another look when you have a chance?

jcarius-nv:
Since this needs more architectural discussions that we don't have the bandwidth for right now, closing this PR for now to defer this change for the future.

PR #2591, opened 2026-04-26
canonicalize collision pairs by GeoType instead of index

### LINKED ISSUE #1281: suggestion: organize shape_contact_pairs according to shape type ordering in the first place
In `sim/builder.py`: `find_shape_contact_pairs`, the a vs. b elements are sorted by shape index:

```Python
 # Ensure canonical order (smaller_element, larger_element)
                if s1 > s2:
                    shape_a, shape_b = s2, s1
                else:
                    shape_a, shape_b = s1, s2
```

But when these are actually consumed in various places, e.g., `geometry/kernels.py`: `broadphase_collision_pairs`, there is redundant code that sets the a vs. b ordering based on the `GeoType` ordering of shape types. Why not just do that once at the start?

4-legends:
closes #1281

## Description

This PR refactors the canonicalization logic for collision shape pairs throughout the ecosystem. Previously, pairs were ordered based solely on their creation index: (min(id1, id2), max(id1, id2)). This created a logic mismatch with solver kernels and contact capacity logic that require GeoType sorting for efficient contact math, leading to redundant allocations and potential filtering failures.

Key Changes:

Unified Canonicalization: All shape pairs are now ordered by GeoType first, with the Shape Index acting as a tie-breaker.

Builder Robustness: Fixed an IndexError in add_shape() where collision filters were being generated before shape metadata (like type) was appended to the builder's internal lists.

Solver Synchronicity: Updated Kamino conversions and MuJoCo solver logic to ensure that geom-pair exclusions and contact manifolds match the new engine-wide ordering.

Logic Simplification: Removed redundant type-swapping logic in model.py and collide.py, as the builder now guarantees the incoming order.

## Checklist

## Test plan

uv run --extra dev -m newton.tests.test_collision_pipeline
uv run --extra dev -m newton.tests.test_model
uv run --extra dev -m newton.tests.cable
uv run --extra dev -m newton.tests.test_enviornment_group_collision

## New feature / API change

New Feature / API Change

Internal API change to shape pair ordering. Downstream code relying on model.shape_contact_pairs or model.shape_collision_filter_pairs should expect the following sort priority:

1. GeoType (Ascending)
2. Shape Index (Ascending)

## Summary by CodeRabbit

* **Refactor**
  * Collision-filter pair canonicalization now orders by shape type first, then index, yielding consistent behavior for mixed-shape interactions.
  * Solver contact-capacity logic aligned with the new type-aware pair ordering to ensure consistent contact handling.

* **Tests**
  * Collision-filtering and contact-pair tests updated and expanded to validate the type-first canonical ordering and consistent exclusion across pipelines.

### DIFF
--- a/newton/_src/sim/model.py
+++ b/newton/_src/sim/model.py
@@ -244,7 +244,7 @@ class Model:
         self.shape_collision_group: wp.array[wp.int32] | None = None
         """Collision group of each shape, shape [shape_count], int. Array populated during finalization."""
         self.shape_collision_filter_pairs: set[tuple[int, int]] = set()
-        """Pairs of shape indices (s1, s2) that should not collide. Pairs are in canonical order: s1 < s2."""
+        """Pairs of shape indices (s1, s2) that should not collide. Pairs are in canonical order (First GeoType then Index): t2 < t1, s1 < s2."""
         self.shape_collision_radius: wp.array[wp.float32] | None = None
         """Collision radius [m] for bounding sphere broadphase, shape [shape_count], float. Not supported by :class:`~newton.solvers.SolverMuJoCo`."""
         self.shape_contact_pairs: wp.array[wp.vec2i] | None = None
--- a/newton/_src/sim/model.py
+++ b/newton/_src/sim/model.py
@@ -244,7 +244,7 @@ class Model:
         self.shape_collision_group: wp.array[wp.int32] | None = None
         """Collision group of each shape, shape [shape_count], int. Array populated during finalization."""
         self.shape_collision_filter_pairs: set[tuple[int, int]] = set()
-        """Pairs of shape indices (s1, s2) that should not collide. Pairs are in canonical order (First GeoType then Index): t2 < t1, s1 < s2."""
+        """Pairs of shape indices (s1, s2) that should not collide. Pairs are in canonical order (First GeoType then Index): t1 < t2, t1 == t2 and s1 < s2, s1 < s2."""
         self.shape_collision_radius: wp.array[wp.float32] | None = None
         """Collision radius [m] for bounding sphere broadphase, shape [shape_count], float. Not supported by :class:`~newton.solvers.SolverMuJoCo`."""
         self.shape_contact_pairs: wp.array[wp.vec2i] | None = None
--- a/newton/_src/solvers/kamino/_src/core/conversions.py
+++ b/newton/_src/solvers/kamino/_src/core/conversions.py
@@ -237,9 +237,7 @@ def compute_required_contact_capacity(
         s2 = int(shape_pair[1])
         type1 = int(shape_type_np[s1])
         type2 = int(shape_type_np[s2])
-        if type1 > type2:
-            s1, s2 = s2, s1
-            type1, type2 = type2, type1
+
         num_contacts_a, num_contacts_b = max_contacts_for_shape_pair(
             type_a=type1,
             type_b=type2,
--- a/newton/_src/solvers/mujoco/solver_mujoco.py
+++ b/newton/_src/solvers/mujoco/solver_mujoco.py
@@ -3721,7 +3721,11 @@ class SolverMuJoCo(SolverBase):
             excluded = True
             for shape_1 in shapes_1:
                 for shape_2 in shapes_2:
-                    if shape_1 > shape_2:
+                    type_1 = model.shape_type[shape_1]
+                    type_2 = model.shape_type[shape_2]
+                    if type_1 > type_2:
+                        s1, s2 = shape_2, shape_1
+                    elif type_1 == type_2 and shape_1 > shape_2:
                         s1, s2 = shape_2, shape_1
                     else:
                         s1, s2 = shape_1, shape_2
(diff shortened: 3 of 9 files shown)


newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -3721,7 +3721,11 @@ def _find_body_collision_filter_pairs(
             excluded = True
             for shape_1 in shapes_1:
                 for shape_2 in shapes_2:
-                    if shape_1 > shape_2:
+                    type_1 = model.shape_type.numpy()[shape_1]
```
vastsoun, line 3724: This call to `model.shape_type.numpy()` is invoking a a CPU-to-GPU copy multiple times in the for-loop. Consider creating a `shape_type_np = model.shape_type.numpy()` outside the loop to first cache the data to CPU to avoid this.

newton/_src/solvers/kamino/_src/core/conversions.py
```diff
@@ -237,9 +237,7 @@ def compute_required_contact_capacity(
         s2 = int(shape_pair[1])
         type1 = int(shape_type_np[s1])
         type2 = int(shape_type_np[s2])
-        if type1 > type2:
-            s1, s2 = s2, s1
-            type1, type2 = type2, type1
+
```
vastsoun, line 240: This is likely to break the CD logic in `kamino/_src/geometry/*`, please run Kamino unit tests in `kamino/tests/` to ensure that everything works. Unfortunately these are not part of the main UT suite, as we're still in the process of migrating them.

newton/_src/sim/model.py
```diff
@@ -244,7 +244,7 @@ def __init__(self, device: Devicelike | None = None):
         self.shape_collision_group: wp.array[wp.int32] | None = None
         """Collision group of each shape, shape [shape_count], int. Array populated during finalization."""
         self.shape_collision_filter_pairs: set[tuple[int, int]] = set()
-        """Pairs of shape indices (s1, s2) that should not collide. Pairs are in canonical order: s1 < s2."""
+        """Pairs of shape indices (s1, s2) that should not collide. Pairs are in canonical order (First GeoType then Index): t1 < t2, t1 == t2 and s1 < s2, s1 < s2."""
```
vastsoun, line 247: Please explain the logic in the docstring, it is not obvious to users, nor to us devs, why this change is necessary. What exactly is wrong with the current shape pair data format?

vastsoun (changes requested):
Thanks @4-legends for the contribution. However, as you are touching very sensitive back-end code, these changes can be quite dangerous, so a lot of care must be taken with this PR. Please consider my PR comments.

newton/_src/solvers/mujoco/solver_mujoco.py
4-legends: Thank you updated
My apologies I missed this.

newton/_src/sim/model.py
4-legends: I have updated the comment to this
Pairs of shape indices (s1, s2) that should not collide. 

Pairs are stored in a unified canonical order based on Shape Type priority:
1. Primary Sort: Lowest GeoType index first (t1 < t2).
2. Secondary Sort: Lowest Shape Index first if types are identical (t1 == t2 and s1 < s2).

This ordering synchronizes the Model's filter list with the Solver's internal contact
manifolds, ensuring that collision exclusion checks and contact capacity estimation 
use a single consistent representation for any shape pair.


More in depth explanation:
1. The "Dual Canonical" Problem

Previously, we had two competing definitions of a "canonical pair":

1. The Builder/Model: Used min(ID1, ID2) to store filters.
2. The Solver/Kernels: Used type-based sorting because collision math is asymmetric (e.g., in a Sphere-Box collision, the math is entirely different than Box-Sphere). To avoid writing two kernels for every pair, the solver forces a specific type order.

2. Redundancy & Memory Waste

1. Because the Model and the Solver were "out of sync," the engine often couldn't tell that Filter (Sphere_ID, Box_ID) and Contact (Box_ID, Sphere_ID) were the same thing without performing an extra swap and check. This led to:
2. Double-counting capacity: compute_required_contact_capacity would sometimes overestimate needed memory because it couldn't efficiently match sorted filters against unsorted potential contacts.

- Kernel Complexity: GPU kernels had to perform "double-swapping" logic to check if a contact should be excluded, wasting registers and cycles.

3. Solving Broad-phase Inconsistency

- Broad-phase algorithms (SAP, BVH, NxN) generate pairs in different orders based on their traversal logic. By enforcing a Type-First canonical order at the root (the ModelBuilder), we ensure that no matter how a pair is discovered, it is instantly converted into a format that matches the pre-sorted exclusion list.

Summary of the new logic:
By moving the "canonicalization" burden to the Builder, we ensure that the shape_collision_filter_pairs array is a perfect mirror of how the GPU will actually "see" those shapes during a collision event. This makes binary searches for excluded pairs faster and 100% reliable across different broad-phase implementations.

nvtw: I do see the logic behind this but I would prefer to keep the filter key solely based on sorted shape ids and rather fix any bugs in case there are some. The swapping based on types when needed in collision kernels should not have a huge impact on performance but feel free to prove me wrong with a benchmark.

nvtw:
@4-legends can we close this issue? I would prefer not to take this change unless you can provide examples where the change is absolutely needed.

PR #2593, opened 2026-04-27
Pin floors for real in minimum-deps nightly tests

adenzler-nvidia:
## Description

The minimum-deps nightly was effectively testing against the **latest** versions of our dependencies, not the floors declared in `pyproject.toml`. The workflow ran:

```yaml
- run: uv lock --resolution lowest-direct
- run: uv run --extra dev -m newton.tests ...
```

`uv lock --resolution lowest-direct` writes the lockfile with floor versions, but `uv run --extra dev` then re-resolves with the default `highest` mode (uv even logs `Ignoring existing lockfile due to change in resolution mode: 'lowest-direct' vs. 'highest'`), installing whatever's currently latest on PyPI.

Surfaced when a newly-released `usd-core` started raising a `DeprecationWarning` from `LoadUsdPhysicsFromRange`. The nightly went red on a SHA whose `pyproject.toml` floor is `usd-core>=25.5` — a version that does not have the deprecation. The floor was never the problem; the workflow just wasn't using it.

This PR replaces the split `uv lock` + `uv run --extra dev` with a single `uv sync --extra dev --resolution lowest-direct` that writes the lockfile and provisions the venv together, then runs the tests via `uv run --no-sync` so they reuse that venv as-is.

`uv tree --depth 1 --frozen` is preserved (now reading the just-written lock without re-resolving) so the resolved versions remain visible in the job log.

## Checklist

- [x] `CHANGELOG.md` has been updated (if user-facing change) — N/A, CI-only

## Test plan

- [ ] Watch the next scheduled minimum-deps nightly (or trigger via `workflow_dispatch`) and confirm `uv tree --depth 1 --frozen` reports `usd-core` at its 25.5 floor and the previously-failing example-import tests pass.

## Bug fix

1. Wait for any direct dependency to publish a new version after the previous nightly run.
2. Trigger the `Minimum Dependency Version Tests on AWS EC2 (Reusable)` workflow.
3. Observe that `uv tree --depth 1` (and the actually-installed packages) reflect the **newest** compatible versions, not the floors declared in `pyproject.toml`.

Concretely, the nightly runs that motivated this fix (`24928773418` on 2026-04-25 and `24954413671` on 2026-04-26, both on SHA `c7b4ddee`) installed a newer `usd-core` than the `>=25.5` floor and failed on `LoadUsdPhysicsFromRange` deprecation.

## Summary by CodeRabbit

* **Chores**
  * CI updated to sync/install a minimal test environment, make dependency inspection deterministic, and reuse the prepared environment for test runs.

* **Changed**
  * Raised optional minimums: meshio → >=5.3.5; Open3D → >=0.19.0 for affected Python/platform combinations; Pillow (examples) → >=11.3.0.

* **Documentation**
  * Changelog updated with new minimum-version notes and a meshio compatibility note regarding NumPy 2.0.


shi-eric:
Minor nit: the `PYTHONFAULTHANDLER` env var was removed from the "Run Tests" step. Other CI workflows (`ci.yml`, `aws_gpu_tests.yml`) consistently set this for traceback dumps on fatal signals. Was this intentional, or just collateral from the reformatting?

shi-eric:
Re: the `open3d` and `Pillow` floor bumps — these force Python 3.10/3.11 users to upgrade dependencies they didn't need to upgrade. From their perspective, `open3d>=0.18.0` and `Pillow>=9.0.0` worked fine; we're bumping the floor purely because our CI runs on Python 3.12.

Worth considering conditional floors instead?

```toml
"open3d>=0.18.0; python_version < '3.12'",
"open3d>=0.19.0; python_version >= '3.12'",
```

Same pattern for Pillow. The `meshio` bump is different — that's about NumPy 2.0 compatibility, not Python version, so a single floor makes sense there.

(Not blocking, just flagging the tradeoff.)

adenzler-nvidia:
Thanks Eric, both fair points.

**`PYTHONFAULTHANDLER`**: re-added at the step level for explicitness. The job-level declaration kept it set in practice (every step in `minimum-deps-tests` inherits `env`), but you're right that anyone running this step in isolation later would benefit from the explicit copy. Now matches the pattern in `ci.yml` / `aws_gpu_tests.yml`.

**Conditional floors**: applied. Now:

```toml
"open3d>=0.18.0; python_version < '3.12' and (sys_platform != 'linux' or platform_machine != 'aarch64')",
"open3d>=0.19.0; python_version >= '3.12' and python_version < '3.13' and (sys_platform != 'linux' or platform_machine != 'aarch64')",

"Pillow>=9.0.0;  python_version < '3.12'",
"Pillow>=10.1.0; python_version >= '3.12'",
```

Both original failures were pure wheel-availability — `open3d 0.18.0` only ships `cp310`/`cp311` wheels and `Pillow 9.0.0` only `cp310`, so on Python 3.12 uv either refuses or falls back to an sdist build that fails for missing system deps. No API break in either: Newton's usage (`PIL.Image` open/save, `o3d.geometry.PointCloud`, `Vector3dVector`, `TriangleMesh.create_from_point_cloud_poisson`) has been stable across these versions, so leaving the original floors in place on Python 3.10/3.11 is safe.

`meshio` stayed unconditional at `>=5.3.5` since the `np.string_` removal is a NumPy-2 API break, not Python-version-conditional.

shi-eric:
LGTM — thanks for the quick turnaround on both points. The conditional floors are clean.

pyproject.toml
```diff
@@ -80,7 +82,8 @@ examples = [
     "imgui_bundle>=1.92.0", # for viewer GUI
     "pyyaml>=6.0.2",
     "cbor2>=5.7.0, <6",     # for binary recording format (.bin files) - more efficient than JSON
-    "Pillow>=9.0.0",        # for image processing (viewer icons, heightfield loading, textures)
+    "Pillow>=9.0.0; python_version < '3.12'",
+    "Pillow>=10.1.0; python_version >= '3.12'",  # for image processing (viewer icons, heightfield loading, textures); 10.1.0 is the first release with cp312 wheels
```
eric-heiden: 🟡 This marker still lets Python 3.13 and 3.14 pick `Pillow==10.1.0` under `--resolution lowest-direct`. That release has cp312 wheels, but not cp313 or cp314 wheels, so the minimum-dependency job can still force a source build or fail on supported Python versions.

Please split the floor by Python version and regenerate `uv.lock`, for example:

```suggestion
    "Pillow>=10.1.0; python_version >= '3.12' and python_version < '3.13'",  # for image processing (viewer icons, heightfield loading, textures); 10.1.0 is the first release with cp312 wheels
    "Pillow>=10.4.0; python_version >= '3.13' and python_version < '3.14'",  # 10.4.0 is the first release with cp313 wheels
    "Pillow>=11.3.0; python_version >= '3.14'",  # 11.3.0 is the first release with cp314 wheels
```

adenzler-nvidia:
Stepped back on the conditional split. Newton's PIL surface is just `Image.open/new/frombytes` + `ImageDraw`/`ImageFont` for viewer icons, and the open3d surface is `PointCloud`, `Vector3dVector`, `TriangleMesh.create_from_point_cloud_poisson`, and `t.io.read_point_cloud` — all rock-stable across these versions. The "don't force 3.10/3.11 users to upgrade" argument doesn't really earn its keep against the maintenance cost of a three-way conditional pin.

Collapsed to single floors:

- `Pillow>=11.3.0` (released 2025-07-01) — first release with cp314 wheels, also has cp310–cp313 wheels, so one line covers all supported Pythons.
- `open3d>=0.19.0; python_version < '3.13' and (sys_platform != 'linux' or platform_machine != 'aarch64')` (released 2025-01-08) — latest open3d still ships no cp313/cp314 wheels, so the `< '3.13'` marker stays load-bearing.

`uv.lock` regenerated; CHANGELOG entries shortened to match.

shi-eric (dismissed):

eric-heiden (dismissed):

nvtw (approved):

PR #2594, opened 2026-04-27
Bump usd-core to >=26.5; drop physics thread-safety workarounds

adenzler-nvidia:
## Description

OpenUSD 26.05 (usd-core 26.5, released 2026-04-24) shipped two related changes that let us delete two long-standing workarounds:

1. Renamed `UsdPhysics.LoadUsdPhysicsFromRange` → `UsdPhysicsLoadStageFromPrimRange` (the old name still exists as deprecated; it raises a `DeprecationWarning` on call).
2. **Fixed a race condition in `UsdPhysicsParsingUtility` that crashed when parsing multiple colliders under the same rigid body in a multi-threaded context** (OpenUSD [PR #4002](https://github.com/PixarAnimationStudios/OpenUSD/pull/4002)).

Newton was carrying matching workarounds for both:

- `PXR_WORK_THREAD_LIMIT=1` set in `newton/tests/thirdparty/unittest_parallel.py` before any `pxr` import, to dodge the race.
- Continuing to call the now-deprecated `LoadUsdPhysicsFromRange` at three sites (`import_usd.py`, the kamino USD reader, and the USD parser test).

This PR bumps `usd-core` to `>=26.5`, migrates the three call sites to `UsdPhysicsLoadStageFromPrimRange`, and drops the `PXR_WORK_THREAD_LIMIT=1` shim — restoring multi-threaded USD work for the parallel test runner.

## Checklist

## Test plan

Verified locally on a Blackwell GPU with `uv sync --extra dev` (default highest resolution, picks `usd-core 26.5`) and the full suite via `python -m newton.tests`:

- All 14 USD-import example tests that fail on the current cloud nightly (`example_robot_*`, `example_sensor_contact_*`, `example_selection_cartpole_*`) pass.
- `test_import_body_newton_armature_warns_deprecated` passes (the extra `LoadUsdPhysicsFromRange` deprecation no longer leaks into the warning count assertion).
- The parallel runner (8 workers) completes without thread-safety crashes — the path the `PXR_WORK_THREAD_LIMIT=1` shim was guarding now relies on the upstream fix.
- Two pre-existing flakes (`test_deterministic_pipeline_500_steps_cuda_0`, `test_robot.example_robot_panda_hydro_cuda_0`) reproduce with and without this change on the same machine and are unrelated; the determinism one looks driver-version-sensitive (cloud `Driver 13.0`, local `12.9`).

## Bug fix

1. Sync the project with `usd-core>=26.5` available on PyPI (e.g. via `uv sync --extra dev` on a fresh checkout today).
2. Run `python -m newton.tests`.
3. Observe 14 example tests crash on `LoadUsdPhysicsFromRange(...)` raising `DeprecationWarning` and `test_import_body_newton_armature_warns_deprecated` failing with a `2 != 1` warning-count mismatch.

The deprecation-side symptom is also addressed by [#2593](https://github.com/newton-physics/newton/pull/2593) (which keeps the floor at `25.5` so the deprecation isn't tripped); this PR is the upstream fix-and-cleanup once we accept the floor bump.

## Summary by CodeRabbit

* **Dependencies**
  * Require USD core >= 26.5 for improved thread-safety and physics utilities.

* **Bug Fixes**
  * Removed startup thread-limit workaround from the parallel test runner.

* **Other Updates**
  * Switched physics-stage loading to the newer OpenUSD API for more robust parsing.

* **Documentation**
  * Removed legacy workaround guidance related to the prior USD physics loading limitation.


adenzler-nvidia:
Ok this one looks blocked on the usd-exchange update, not sure if I should guard for ARM or just wait with this.

There is a deprecation warning now for any user that resolves to latest so that's not great.

andrewkaufman:
@adenzler-nvidia usd-exchange aside, this is a pretty risky change to make, Kit 110 is locked to USD 25.11 & that is what Isaac Sim 6.0 will ship with. If newton becomes dependant on API surface on a newer USD (or even on the threading fix), then Isaac Sim will not be able to adopt newton without also matching that USD minspec.

andrewkaufman:
For the usd-exchange part of it, our 3.x series was planned to bump to USD 26.5, but that release is still a monther or two out...

adenzler-nvidia:
Thanks for the heads-up - Let's postpone this change. We should still address the deprecation warning when resolving to latest, but otherwise I agree no point in upgrading the min version floor.

Issue #1289, opened 2025-12-23
xpbd restitution likely issues


rcoreilly:
in `solvers/xpbd/kernels.py, apply_rigid_restitution` there are a couple of likely issues. However, even fixing these still leaves issues with this code -- seems like a longer time integral is needed to properly manage this logic.

In general, _a and _b are using the same normal n, but given their relationship on opposite sides of the contact, the _a side should use the negated norm (as is done in other such code).

e.g.:

```Python
        rxn_a = wp.quat_rotate_inv(q_a, wp.cross(r_a, n))
...
        wp.atomic_add(deltas, body_a, wp.spatial_vector(n * m_inv_a * dv_a, dq))
```

tueboesen:
I have been testing out xpbd restitution and there are definitely some issues with it.
If I set restitution to 1.0 and drop a mesh on top of a box from a few centimeters height, then it bounces against the box and flies more than a meter up in the air.

If I set the restitution very low instead 0.01, then the mesh still flies significantly higher than the initial drop height, upon the second rebounce the mesh simply disappears (looks like it went to 1e29 in height from one frame to the next.

(I attached a usd file for the second case with .txt extension since uploading usd files is not allowed.)
[physics_model_00.usd.txt](https://github.com/user-attachments/files/24637357/physics_model_00.usd.txt)

aneangel:
@adenzler-nvidia can I also be put as the assignee on this as well as this is touched on by a PR that i have open for review.

PR #2602, opened 2026-04-27
Deprecate SolverXPBD enable_restitution=False and fix energy explosion when enabled

### LINKED ISSUE #1137: Rigid bodies bouncing even with zero restitution
I am currently in the process of testing out Newton, and noticed this rather unphysical behaviour when dropping one shape on top of another.
Both objects involved in this bounce are shape meshes with a friction coefficient of 0.5 and a restitution of 0.0, which should resolve in a completely inelastic collision as far as I am aware and should result in no bounce at all.

However as you can see in the attached usd file this is not what actually happens. (I had to rename the usd file to .txt since .usd is not allowed)

The usd file shows a box being droppped on another box from a height of roughly 9 cm, and it bounces around 9 cm up in the air again.

[physics_model_07_usd.txt](https://github.com/user-attachments/files/23771119/physics_model_07_usd.txt)

Am I doing something wrong? 
Is this a known bug?

The code that produces the above is rather big, but I have made the following small code that reproduces the problem
```

import newton
import numpy as np
import warp as wp
from plyfile import PlyData, PlyElement

class PhysicsSimulator:
    def __init__(self, mesh0,mesh1, fps=50, steps=50,substeps=32, restitution=0.0, output_name='test'):

        self.up_axis = "Z"
        self.fps = fps
        self.steps = steps
        self.frame_dt = 1.0 / fps
        self.sim_substeps = substeps
        self.gravity = -9.8156
        self.use_cuda_graph = False
        self.sim_time = 0.0
        self.contact_margin = 0.01
        self.output_file = f"{output_name}_{restitution:2.2f}.usd"


        self.viewer = newton.viewer.ViewerUSD(output_path=self.output_file, fps=self.fps, scaling=1.0, up_axis=self.up_axis)

        builder = newton.ModelBuilder(gravity=self.gravity, up_axis=self.up_axis)
        builder.add_ground_plane()


        origin = wp.transform(p=[0.0,0.0,0.1], q=[0.0,0.0,0.0,1.0])
        b = builder.add_body(xform=origin)
        cfg = newton.ModelBuilder.ShapeConfig(density=100, restitution=restitution)
        builder.add_shape_mesh(body=b, mesh=mesh0, cfg=cfg)

        origin = wp.transform(p=[0.0,0.0,0.3], q=[0.0,0.0,0.0,1.0])
        b = builder.add_body(xform=origin)
        cfg = newton.ModelBuilder.ShapeConfig(density=1000, restitution=restitution)
        builder.add_shape_mesh(body=b, mesh=mesh1, cfg=cfg)

        self.solver = newton.solvers.SolverXPBD(self.model)

        self.state_0 = self.model.state()
        self.state_1 = self.model.state()
        self.contacts = self.model.collide(state=self.state_0, rigid_contact_margin=self.contact_margin)

        newton.eval_fk(self.model, self.model.joint_q, self.model.joint_qd, self.state_0)

        for i in range(self.steps):
            self.step()
            self.render()
        self.viewer.close()

    def capture_graph(self):
        if self.use_cuda_graph:
                self.simulate()
            self.graph = capture.graph
        else:
            self.graph = None

    def simulate(self):
        for _ in range(self.sim_substeps):
            self.state_0.clear_forces()

            # apply forces to the model
            self.viewer.apply_forces(self.state_0)

            self.contacts = self.model.collide(state=self.state_0,
                                               rigid_contact_margin=self.contact_margin)
            self.solver.step(self.state_0, self.state_1, self.control, self.contacts, self.sim_dt)

            # swap states

    def step(self):
        with wp.ScopedStream() if self.use_cuda_graph else wp.ScopedTimer("step"):
            if self.use_cuda_graph:
                wp.capture_launch(self.graph)
            else:
                self.simulate()

    def render(self):
        with wp.ScopedStream(self.stream_render) if self.use_cuda_graph else wp.ScopedTimer("render"):
            self.viewer.log_state(self.state_0)
            self.viewer.log_contacts(self.contacts, self.state_0)
            self.viewer.end_frame()


def get_mesh_from_ply_data(ply_data: PlyData, box_sizes: tuple[float, float, float], center_box: bool) -> newton.Mesh:
    """
    Create a Warp mesh from PLY data.

    Args:
        ply_data (PlyData): The PLY data containing the mesh information
        box_sizes (tuple): The sizes of the bounding box
        center_box (bool): Whether to center the box at the origin
    Returns:
        newton.Mesh: The generated Warp mesh
    """

    # Extract vertices from PlyData
    vertex_data = ply_data['vertex'].data
    verts = np.column_stack([
        vertex_data['x'].astype(np.float32),
        vertex_data['y'].astype(np.float32),
        vertex_data['z'].astype(np.float32)
    ])

    if center_box:
        verts -= np.array(box_sizes, dtype=np.float32) * 0.5

    # Extract faces
    faces_raw = ply_data['face'].data['vertex_indices']
    tri_list = []
    for face in faces_raw:
        face_idx = list(face)
        if len(face_idx) == 3:
            tri_list.append(face_idx)

    if not tri_list:
        raise ValueError("No triangle faces could be extracted from PLY data")

    # Convert to Warp arrays
    wp_points = np.array(verts, dtype=np.float32)

    # Warp expects a flat int array of indices (multiple of 3)
    wp_indices = np.array(tri_list, dtype=np.int32)

    return newton.Mesh(vertices=wp_points, indices=wp_indices)

def get_ideal_box_mesh(box_size: tuple[float, float, float], edge_length: float,
                       center_box: bool = False) -> newton.Mesh:
    """
    Create an Nvidia Warp 3D box mesh with uniform vertex spacing.

    Args:
        box_sizes (tuple): (dx, dy, dz) defining the box dimensions
        center_box (bool): If True, center the box at the origin
        mean_edge_length (float): Approximate desired distance between neighboring vertices

    Returns:
        newton.Mesh: The generated box mesh
    """
    sx, sy, sz = box_size

    # Grid counts (at least 2 per axis (corners))
    nx = max(2, int(np.ceil(sx / edge_length)) + 1)
    ny = max(2, int(np.ceil(sy / edge_length)) + 1)
    nz = max(2, int(np.ceil(sz / edge_length)) + 1)

    xs = np.linspace(-sx / 2, sx / 2, nx)
    ys = np.linspace(-sy / 2, sy / 2, ny)
    zs = np.linspace(-sz / 2, sz / 2, nz)

    vertices = []  # will contain floats (may have duplicates before dedupe)
    faces = []  # list of (a,b,c) indices into vertices list

    def add_face(coord1: np.ndarray, coord2: np.ndarray, const_val: float, axis: str):
        """
        Add a grid face.
        Triangles are added as (v0,v1,v3) and (v0,v3,v2) per quad.

        Args:
            coord1 (np.ndarray): 1D array for the first varying coordinate in a mesh grid
            coord2 (np.ndarray): 1D array for the second varying coordinate in a mesh grid
            const_val (float): Fixed coordinate value for the face
            axis (str): 'x', 'y' or 'z' indicating which axis is fixed
        """
        nonlocal vertices, faces

        g1, g2 = np.meshgrid(coord1, coord2, indexing='ij')
        n1, n2 = g1.shape
        start_idx = len(vertices)

        if axis == 'x':
            pts = np.stack([np.full_like(g1, const_val), g1, g2], axis=-1)
        elif axis == 'y':
            pts = np.stack([g1, np.full_like(g1, const_val), g2], axis=-1)
        else:  # 'z'
            pts = np.stack([g1, g2, np.full_like(g1, const_val)], axis=-1)

        # Append vertices for this face (will deduplicate later)
        vertices.extend(pts.reshape(-1, 3))

        # Triangulate grid
        for i in range(n1 - 1):
            for j in range(n2 - 1):
                v0 = start_idx + i * n2 + j
                v1 = v0 + 1
                v2 = v0 + n2
                v3 = v2 + 1
                faces.append((v0, v1, v3))
                faces.append((v0, v3, v2))

    # Add the six faces (each at +/- half-extent)
    add_face(ys, zs, -sx / 2, 'x')  # -X
    add_face(ys, zs, sx / 2, 'x')  # +X
    add_face(xs, zs, -sy / 2, 'y')  # -Y
    add_face(xs, zs, sy / 2, 'y')  # +Y
    add_face(xs, ys, -sz / 2, 'z')  # -Z
    add_face(xs, ys, sz / 2, 'z')  # +Z

    # Convert to numpy and deduplicate
    verts_arr = np.asarray(vertices, dtype=np.float64)
    verts_rounded = np.round(verts_arr, decimals=8)
    unique_verts, inv_idx = np.unique(verts_rounded, axis=0, return_inverse=True)

    # Remap faces to unique vertex indices
    faces_mapped = np.array([(inv_idx[a], inv_idx[b], inv_idx[c]) for (a, b, c) in faces], dtype=np.int32)

    # Ensure outward normals: For each triangle, compute normal and flip if it points inward
    for i in range(len(faces_mapped)):
        a, b, c = faces_mapped[i]
        v0 = unique_verts[a]
        v1 = unique_verts[b]
        v2 = unique_verts[c]

        normal = np.cross(v1 - v0, v2 - v0)
        centroid = (v0 + v1 + v2) / 3.0

        # If normal dot centroid < 0 then normal points toward center and we flip
        if np.dot(normal, centroid) < 0:
            faces_mapped[i] = (a, c, b)

    # Build PLY elements
    vertex_el = np.array([tuple(v.astype(np.float32)) for v in unique_verts],
                         dtype=[('x', 'f4'), ('y', 'f4'), ('z', 'f4')])
    face_el = np.array([(list(face),) for face in faces_mapped], dtype=[('vertex_indices', 'i4', (3,))])

    ply = PlyData([PlyElement.describe(vertex_el, 'vertex'), PlyElement.describe(face_el, 'face')])

    box_mesh = get_mesh_from_ply_data(ply, box_size, center_box)

    return box_mesh


def generate_box_mesh(box_sizes, center_box=False):
    """
    Creates a box mesh.
    """
    point_cloud = np.zeros((8, 3), dtype=np.float32)
    point_cloud[1] = [box_sizes[0], 0, 0]
    point_cloud[2] = [box_sizes[0], box_sizes[1], 0]
    point_cloud[3] = [0, box_sizes[1], 0]
    point_cloud[4] = [0, 0, box_sizes[2]]
    point_cloud[5] = [box_sizes[0], 0, box_sizes[2]]
    point_cloud[6] = [box_sizes[0], box_sizes[1], box_sizes[2]]
    point_cloud[7] = [0, box_sizes[1], box_sizes[2]]
    if center_box:
        point_cloud -= box_sizes / 2

    indices = np.zeros((12, 3), dtype=int)
    indices[0] = [0, 2, 1]
    # indices[0] = [0, 1, 2]
    indices[1] = [0, 3, 2]
    indices[2] = [4, 5, 6]
    indices[3] = [4, 6, 7]
    indices[4] = [0, 1, 5]
    indices[5] = [0, 5, 4]
    indices[6] = [1, 2, 6]
    indices[7] = [1, 6, 5]
    indices[8] = [2, 3, 7]
    indices[9] = [2, 7, 6]
    indices[10] = [3, 0, 4]
    indices[11] = [3, 4, 7]
    return point_cloud, indices




if __name__ == "__main__":
    use_simple = False
    restitution = 0.0
    resolution = 0.0375  # TODO Is this a good value? Arrived by smallest box (0.15 m) / 4
    output_name = 'test'

    box_sizes = np.array([0.385, 0.385, 0.09], dtype=np.float32)
    if use_simple:
        box_mesh_parameters = generate_box_mesh(box_sizes, center_box=True)
        mesh0 = newton.Mesh(*box_mesh_parameters)
    else:
        mesh0 = get_ideal_box_mesh(box_sizes,edge_length=resolution,center_box=True)

    box_sizes = np.array([0.385, 0.385, 0.09], dtype=np.float32)
    if use_simple:
        box_mesh_parameters = generate_box_mesh(box_sizes, center_box=True)
        mesh1 = newton.Mesh(*box_mesh_parameters)
    else:
        mesh1 = get_ideal_box_mesh(box_sizes,edge_length=resolution,center_box=True)

    sim = PhysicsSimulator(mesh0,mesh1, restitution=restitution,output_name=output_name)

```
Changing the value of the restitution from 0.0 to 1.0 seems to have no effect on the actual collisions.

aneangel:
## Description
Fixes two bugs in `SolverXPBD` that caused per-shape restitution values to be silently ignored and, when restitution was manually enabled, caused energy explosion.

**Bug 1 (issue #1137):** `enable_restitution` defaulted to `False`, so the restitution pass never ran regardless of per-shape restitution settings. Rather than flip the default immediately, this PR deprecates `enable_restitution=False` — constructing `SolverXPBD` without passing `enable_restitution=True` now emits a `DeprecationWarning`. The default will flip to `True` in a future release; after that, the flag will be removed so behavior is fully material-driven. Pass `enable_restitution=True` to opt in now, and set material restitution coefficients to `0.0` for no-bounce contacts.

**Bug 2 (issue #1289):** When `enable_restitution=True`, the restitution kernel read stale pre-correction velocities, producing positions on the order of 1e29 m after the first impact. Fixed by computing body velocities from position deltas before the restitution pass, so the kernel always sees physically bounded, post-correction velocities.

Closes #1137 and #1289 

## Checklist

## Test plan

Created two unit tests `test_rigid_restitution_zero_settles and test_rigid_restitution_elastic_no_explosion` checking specifically that we have address the behavior noted in the two open issues

## Bug fix

Steps to reproduce (Bug 1 — restitution silently ignored):

Create a SolverXPBD with default arguments and assign restitution=0.0 to one shape and restitution=1.0 to another.
Drop both shapes from the same height.
Observe both shapes behave identically — neither bounces — because the restitution pass never ran.
Steps to reproduce (Bug 2 — energy explosion):

Create a SolverXPBD(model, enable_restitution=True).
Drop a body with restitution=1.0 onto the ground.
Observe the body position reaching ~1e29 m after the first impact.

Minimal reproduction (Bug 2):

```
import warp as wp
import newton

builder.add_ground_plane()
body = builder.add_body(xform=wp.transform(wp.vec3(0, 0, 0.3), wp.quat_identity()))
builder.add_shape_box(body, hx=0.1, hy=0.1, hz=0.05,
                      cfg=newton.ModelBuilder.ShapeConfig(density=500.0, restitution=1.0))
model = builder.finalize()

solver = newton.solvers.SolverXPBD(model, enable_restitution=True)  # explodes before this PR
control, contacts = model.control(), model.contacts()
newton.eval_fk(model, model.joint_q, model.joint_qd, state_0)

dt = 1.0 / 60 / 16
for _ in range(60 * 16):
    state_0.clear_forces()
    solver.step(state_0, state_1, control, contacts, dt)

print(state_0.body_q.numpy()[0][2])  # was ~1e29, now ~0.2–0.3 m
```

## Summary by CodeRabbit

* **New Features**
  * Rigid-body restitution is now enabled by default in the XPBD physics solver for more realistic bounces.

* **Bug Fixes**
  * Prevents unbounded/energy-explosive rebounds when restitution is active and ensures per-shape restitution is respected.

* **Tests**
  * Added regression tests for inelastic (no bounce) and fully elastic scenarios and stability bounds for elastic simulations.

* **Documentation**
  * Changelog updated with default change and opt-out guidance.


devshahofficial:
@aneangel I think the remaining items split cleanly: the duplicate changelog heading is mechanical, and the restitution path needs one focused grad-mode test. For restitution, I’d keep the position-delta velocity calculation available when `enable_restitution` is true, but route it through temporary non-grad state so the rebound logic still sees the impact velocity. Happy to help with that patch if useful.

aneangel:
@devshahofficial, I want to make sure that I understand correctly, are you suggesting to run update_body_velocities into a detached temporary array (not on the tape), then feed that temp velocity into apply_rigid_restitution. The restitution impulse would then be written to a new clone of state_out.body_qd (required anyway since you can't overwrite a recorded array under Warp's autodiff rules)?

would that not potentially have a gradient problem where it ends up with this dep chain: state_out.body_q (detached) ->  tmp_body_qd -> restitution_impulse  -> new_body_qd

Because tmp_body_qd is detached, the gradient of any downstream loss through the bounce is silently zero with respect to positions. A user doing gradient-based trajectory optimization would get no gradient signal from contact restitution (and no error), just wrong gradients.

For XPBD + requires_grad (in cases like gradient-based robot learning, trajectory optimization), we get physically correct forward bouncing at the price of silently incorrect gradients. 

That trade off is probably not worth it (unless there is reason for the former). The current approach which is skip restitution in grad mode, keep the gradient path clean is a conservative choice. 

If restitution in grad mode becomes a real need, would it make more sense to make update_body_velocities itself differentiable on the tape, however that is a larger and more principled change than routing through detached temps.

Would love to discuss this more I may very well be wrong/misunderstanding something so would love your take on this, @adenzler-nvidia / @preist-nvidia would also love your input. Want to make sure the implementation is moving forward in the correct direction.

devshahofficial:
I prepared a focused follow-up for the remaining grad-enabled restitution thread here:

https://github.com/devshahofficial/newton/commit/5dadb2c4

Branch: https://github.com/devshahofficial/newton/tree/devshahofficial/local-pr2602-followups

It keeps the position-delta velocity update available on grad-enabled rigid steps by writing into cloned buffers, then feeds those corrected velocities into `apply_rigid_restitution` without mutating recorded arrays in place. The regression now requires bounded rebound for both grad and non-grad paths.

Validation I ran locally:
```bash
uv run --extra dev -m newton.tests -k rigid_restitution
uvx pre-commit run -a
```

I tried pushing directly to the PR head branch because maintainer edits are enabled, but GitHub rejected that with a 403 for my account, so I left this as a branch/commit you can pull or cherry-pick.

aneangel:
@devshahofficial That makes sense. I will take a look at integrating your inputs, the code you provided cleared up my understanding.

aneangel:
@coderabbitai review

devshahofficial:
@jcarius-nv @aneangel let me know if I can help!

aneangel:
@dylanturpin when you get a chance this PR is also ready for review to resolve and push if no additional feedback, thank you!

dylanturpin:
Gotcha, will do. Thank you!

dylanturpin (dismissed):
LGTM.

Validation:
`uv run --extra dev -m newton.tests -k restitution` passed locally.

Small nit:
the title/description still mention defaulting enable_restitution to True , but I believe this currently keeps default as False. I'm not sure if the code should be changed or the PR text, but those should align.

Default line: `enable_restitution: bool = False`
https://github.com/newton-physics/newton/blob/b5c01da89d48b73639f8a61a955bb39b64bf7f43/newton/_src/solvers/xpbd/solver_xpbd.py#L108

adenzler-nvidia:
**Stage XPBD restitution toward material-driven behavior**

The current code keeping `enable_restitution: bool = False` is the right compatibility choice for this PR. It avoids flipping `SolverXPBD(model)` behavior immediately.

The end goal should be that XPBD honors material restitution by default and the solver-level `enable_restitution` flag eventually goes away. To get there without a breaking change, we should deprecate the current disabled-by-default behavior now and give users a clear opt-in path.

Suggested staging:

- keep `enable_restitution=False` in this PR;
- emit a `DeprecationWarning` when `enable_restitution` is `False`;
- tell users to pass `enable_restitution=True` to opt into the future behavior;
- tell users who want no-bounce contacts to set material restitution coefficients to `0.0`;
- later flip the default to `True`;
- after that, deprecate/remove the flag so behavior is fully material-driven.

Suggested warning text:

```python
warnings.warn(
    "SolverXPBD(enable_restitution=False) is deprecated. XPBD will enable "
    "restitution by default in a future release. Pass enable_restitution=True "
    "to opt into the future behavior, and set material restitution coefficients "
    "to 0.0 for no-bounce contacts.",
    DeprecationWarning,
    stacklevel=2,
)
```

Please also add a `Deprecated` changelog entry with that migration guidance, and update the PR title/body so they describe the staged deprecation rather than saying the default has already flipped.

**Trim regression-test comments to match Newton guidance**

Newton guidance says code comments should be brief, only for non-obvious code, and explain why rather than restating what the code does. The new regression tests add long before/after narratives, and some of that prose is already drifting from the implementation.

Please trim these to the regression intent and keep detailed setup mostly in failure messages where it helps debug failures.

Specific spots:

- `newton/tests/test_solver_xpbd.py:372`: the docstring is much longer than needed and mixes the default-disabled issue with the explicit `enable_restitution=True` path.
- `newton/tests/test_solver_xpbd.py:634`: “gated behind not requires_grad” is misleading; the old code still launched restitution. The issue was stale/corrected velocity handling and grad-safe output buffers.
- `CHANGELOG.md:87`: same “silently skipped when requires_grad=True” wording should be tightened.

jcarius-nv:

adenzler-nvidia:
I pushed an update to this branch to get it over the line:

- Merged latest `main`.
- Implemented the staged deprecation from my earlier comment: `enable_restitution` keeps its `False` default, but constructing `SolverXPBD` with restitution disabled now emits a `DeprecationWarning` pointing users at `enable_restitution=True` and material restitution coefficients. The default flips in a future release, after which the flag can go away so behavior is fully material-driven.
- Added a `Deprecated` changelog entry with migration guidance and moved the two `Fixed` entries into the current top-level `[Unreleased]` section (the merge had left them under the released 1.2.0 heading).
- Trimmed the regression-test docstrings and corrected wording that misstated the old behavior: the restitution kernels always launched under `requires_grad`; the actual bugs were stale pre-correction velocities and in-place writes to tape-recorded arrays.

All `TestSolverXPBD` tests pass locally on CPU and GPU.

@aneangel could you update the PR title/description to describe the staged deprecation instead of a default flip? Everything else should be ready for review.

aneangel:
@adenzler-nvidia went ahead and changed what you requested, thank you for helping get this over the line have been backed up with a few other things, thanks!

devshahofficial:
@aneangel @dylanturpin I tracked the new CI failures to the deprecation warning firing for default `SolverXPBD()` construction. The hosted and GPU unit jobs run examples with `-W error::DeprecationWarning`, so the warning from omitted `enable_restitution` was enough to fail broad example coverage.

I pushed a focused helper commit here since I still cannot push to the PR head branch directly:

https://github.com/devshahofficial/newton/commit/e009cf2a

Branch: https://github.com/devshahofficial/newton/tree/devshahofficial/pr2602-ci-warning-fix

What it does:
- keeps omitted `enable_restitution` as the silent legacy default during the deprecation window
- still warns for explicit `enable_restitution=False`
- adds tests for default/no-warning and explicit-false/warning behavior
- tightens the changelog wording to match that behavior

Validated locally on macOS CPU:
- `uv run --extra dev -m newton.tests -k test_enable_restitution`
- `uv run --extra dev python -W error::DeprecationWarning -m newton.examples.basic.example_basic_pendulum --device cpu --test --quiet --viewer null`
- `uv run --extra dev python -W error::DeprecationWarning -m newton.examples.basic.example_basic_conveyor --device cpu --test --quiet --viewer null --num-frames 100`
- `uv run --extra dev -m newton.tests -k rigid_restitution`
- `uv run --extra dev -m newton.tests -k particle_shape_restitution`
- `uvx pre-commit run -a`

Small note for the AWS workflow: use “Re-run all jobs” after applying this, not “Re-run failed jobs”, because the failed-job path queues against the terminated EC2 runner.

devshahofficial:
@aneangel @dylanturpin quick nudge: CI is still failing on the default-construction deprecation warning. The focused fix remains ready at devshahofficial/newton@e009cf2a; happy to refresh it onto the latest PR head if that would make it easier to pick up.

dylanturpin:
@devshahofficial thanks — your diagnosis is right: the warning fires on default construction and the example/test jobs run with `-W error::DeprecationWarning`, hence the ~152 failures. Rather than picking up the fork commit directly, the team should call the semantics first: (a) sentinel-`None` default that stays silent when omitted (your approach — CI-green, but users relying on the implicit default never see a warning before it flips to `True`), or (b) keep warning on omission and update our own call sites/examples to pass `enable_restitution` explicitly, like the `body_frame_origin` deprecation did. I'll route that decision and follow up here. Either way the branch needs a merge from main first (`solver_xpbd.py` conflicts with #2993/#3300).

devshahofficial:
Thanks, that tradeoff is clear. I’d choose (b): warning on omission reaches the users whose behavior will change, while Newton’s own examples and tests can pass the value explicitly during the transition. I’ll hold off on refreshing the helper branch until the semantics call lands; once it does, I’m happy to rebase and help with the call-site updates.

dylanturpin:
@devshahofficial quick update to save you effort: (b) is confirmed, and the call-site work is already done on a local branch — all ~100 internal constructions flipped, plus the merge from main. One heads-up for whenever you rebase: git auto-merges `apply_rigid_restitution` incorrectly (it drops the margin-aware separation and silently kills rigid restitution — `test_rigid_restitution_zero_settles` catches it; we have a fix). The flip also surfaced some solver-behavior questions we're reviewing internally before pushing — enabling restitution changes more than the restitution pass. So: no need to refresh your branch or start on call sites; I'll post the branch and details here shortly.

dylanturpin:
As promised, the branch: https://github.com/dylanturpin/newton-collab/tree/dturpin/pr2602-callsite-updates (this PR's head + merge of current main + the call-site work). Three things in it:

- **All ~100 internal `SolverXPBD` constructions now pass `enable_restitution=True` explicitly** (tests, examples, benchmarks, docs). The `--strict-warnings` deprecation failures are gone.
- **Merge trap + fix** (`cea33ca5`): merging main auto-resolves `apply_rigid_restitution` to #3287's raw separation while keeping this branch's restitution flow, which silently kills all rigid restitution — contacts are generated while surfaces are still within the collision margin, so the `d >= 0` gate rejects exactly the impact contacts restitution must handle. `test_rigid_restitution_zero_settles` catches it (passes on this PR's head, fails on a naive merge). The fix restores the margin-aware `contact_surface_separation` gate and extends the #3287 regression test with explicit zero margins; all 8 rigid-restitution tests (this branch's and main's) pass together.
- **Open design issue — why the branch is not green yet**: with every call site flipped, 16 physics tests fail, and most are *not about restitution*. `enable_restitution=True` also switches body velocity integration to position-delta-derived values for all bodies (`update_body_velocities`, OR-coupled at `solver_xpbd.py:782`), contacts or not: a contact-free pendulum trajectory goes 86× out of tolerance, linear momentum drift is 6× over bound, hydroelastic cube stacks explode, and a friction ramp stops sliding. An `e == 0 → skip` early-out in `apply_rigid_restitution` fixes the resting-contact instabilities but not the integration-mode failures, so the two effects are separable. Net: the future `True` default currently changes core rigid dynamics for every user, and the team needs to decide (being routed internally) whether to decouple the velocity mode from restitution before call sites — or user code — can flip.

Please don't spend time on call sites or the main merge; happy to hand any of this over once the design call lands.

jcarius-nv:
@dylanturpin can you update this PR now that Eric's one is merged? Or are you planning to open a new one based on the changes in your fork mentioned above?

aneangel:
Hi @jcarius-nv I noticed there has been quite a lot of work surrounding this PR and the original issue, based off your last comment and Dylan's work is there anything needed from me on this PR? I just want to make sure that any work that is left gets done from my end. I see that Dylan already has done some work which is why I ask.

jcarius-nv:
@aneangel, I've asked @dylanturpin to provide an update here, I'm also a bit confused because the linked PRs are just on Dylan's and Eric's forks. Let's wait for this before proceeding.

dylanturpin:
Current story, since the linked fork PRs make this look more mysterious than it is:

- #2602 found two separable problems: the API migration (`enable_restitution` default/flag) and the restitution implementation. Flipping Newton's internal call sites to `True` exposed that the flag also switched the solver's body-velocity integration path even without active restitution, causing broad core-dynamics regressions; this head is not mergeable as-is and is now conflict-stale.
- Eric's dylanturpin/newton-collab#14 is merged and decouples restitution from body integration, keeping contact-free trajectories bit-identical and making the all-zero-`e` path cheap.
- dylanturpin/newton-collab#15 is the per-manifold GS follow-up for the low-iteration flat-manifold problem. Eric's original coefficient/timestep review points are addressed at `5a7500b5`, but his Aug 12 `e=0` already-separating regression still needs one fix and test. After that I'll rebase the complete chain onto current `main` and open a clean upstream replacement; #2602 should be superseded rather than revived.

@aneangel nothing is left on your side right now. I own the remaining `e=0` fix plus the upstream rebase/replacement, and I'll close the loop here when that is ready.

dylanturpin:
Superseded by #4103, which carries this PR plus the decoupling and per-manifold GS follow-up onto current `main` with the original authorship preserved.

#4103 now includes the remaining `e=0` already-separating fix, grad-buffer lifetime correction, runtime-toggle consistency, particle resting dead band, compatibility deprecation, and focused CPU/CUDA regressions. I reviewed and approved the current replacement head.

Nothing remains on Anthony's branch, so I'm closing this stale/conflicting PR in favor of #4103. #1137 remains open separately because its exact dynamic mesh/body-body reproducer is not fully resolved.

aneangel:
Thanks for clarifying @dylanturpin , just wanted to double check :)

jcarius-nv:
>https://github.com/newton-physics/newton/issues/1137 remains open separately because its exact dynamic mesh/body-body reproducer is not fully resolved.

what is the remaining issue here to get that also closed?

PR #2606, opened 2026-04-27
Add example render FPS cap

### LINKED ISSUE #2605: [REQ] Add `--render-fps` flag to cap example rendering rate
### Description

Add `--render-fps` flag for all examples.

Sample implementation: https://github.com/newton-physics/newton/compare/main...j3soon:newton:feat/render-fps

I can open a pull request if suitable.

### Motivation / Use Case

Capping the render FPS can be useful on legacy systems. For example, on a system using x11vnc, running applications with an OpenGL viewer without an FPS cap can make the VNC session extremely laggy / frozen. There are also other scenarios where limiting the rendering FPS may improve overall usability and performance.

Note that this is different from the `--fps` flag, which will affect simulation `frame_dt`.

### Alternatives Considered

_No response_

devshahofficial:
## Description

Add a shared `--render-fps` option for Newton examples so users can cap rendering throughput without changing simulation `frame_dt` or example `--fps` behavior.

The option is parsed by the common example CLI and applied in the shared run loop after each render call. Invalid values such as `0`, negative numbers, and `nan` are rejected at parse time.

Fixes #2605

## Checklist

## Test plan

```bash
python -m newton.tests -k test_render_fps_cli
python -m newton.tests -k test_warp_config_cli
SKIP=uv-lock uvx pre-commit run -a
```

## New feature / API change

```bash
python -m newton.examples basic_pendulum --render-fps 30
```

## Summary by CodeRabbit

* **New Features**
  * Added `--render-fps` CLI option to cap example rendering frame rate independently of simulation timing. Accepts only positive finite values.

* **Tests**
  * Added comprehensive test suite for the new CLI feature, validating argument parsing, input validation, and rendering throttle behavior.

### DIFF
--- a/newton/examples/__init__.py
+++ b/newton/examples/__init__.py
@@ -397,6 +397,7 @@ def run(example, args):
         if example is None:
             viewer.begin_frame(0.0)
             viewer.end_frame()
+            _throttle_render_fps(frame_start_time, render_fps)
             continue
 
         if viewer.should_step():
--- a/newton/examples/__init__.py
+++ b/newton/examples/__init__.py
@@ -5,7 +5,9 @@ import ast
 import copy
 import gc
 import importlib
+import math
 import os
+import time
 import warnings
 from collections import defaultdict
 from collections.abc import Callable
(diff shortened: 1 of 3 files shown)


devshahofficial:
@eric-heiden could you review this when you have a chance?

I re-read the thread and the prior CodeRabbit feedback. The idle-path throttling issue was already addressed in the follow-up commit, and the latest CodeRabbit pass reported no actionable comments. I rebased the branch onto current `upstream/main` to clear the previous dirty merge state.

Validation:
```bash
uv run --extra dev -m newton.tests -k test_render_fps_cli
uvx pre-commit run -a
```

devshahofficial:
@eric-heiden WDYT ?

eric-heiden:
@devshahofficial Apologies for the delay! I only have one last comment, then this should be good to merge.

newton/examples/__init__.py
```diff
@@ -283,9 +285,62 @@ def _format_fps(fps: float) -> str:
...
+    render_fps: float | None,
+    *,
+    time_fn: Callable[[], float] = time.perf_counter,
+    sleep_fn: Callable[[float], None] = time.sleep,
+) -> float:
+    """Sleep to cap a render loop to ``render_fps``.
+
+    Args:
+        frame_start_time: Wall-clock time at the start of the current frame.
+        render_fps: Maximum render rate in frames per second, or ``None`` for
```
eric-heiden, line 337: 🟡 The new shared `--render-fps` option should also be listed in the README's `Example Options` table. That section is where users see the common flags accepted by examples, so leaving this out makes the new render cap discoverable only through `--help`.

Please add a row for `--render-fps` in `README.md` alongside `--viewer`, `--device`, `--num-frames`, and `--output-path`, for example:

```markdown
| `--render-fps` | Maximum render rate in frames per second (loop throughput cap). | `None` (uncapped) |
```

jcarius-nv:

newton/examples/__init__.py
adenzler-nvidia: Sorry for the churn here — this one's on me. @eric-heiden's request was correct at the time, and @devshahofficial did act on it in the "Document render FPS example option" commit (the README row is there). The issue is timing: a couple of days later I merged #3070, which intentionally slimmed this table down to just the most commonly used options (`--viewer`, `--device`, `--num-frames`, `--output-path`) and pointed users at `python -m newton.examples <example> --help` as the always-current, complete list.

`--render-fps` is a niche throughput cap, so under that policy it shouldn't go in the table — and that added row is now what's causing the merge conflict against `main`. Could you drop the README change in the rebase? The argparse `help=` string already documents the flag, which is exactly what `--help` surfaces. Apologies again for sending you back and forth on it.

adenzler-nvidia:
Re-reviewed — the code and tests look good, and the idle-path throttle concern is handled. Two mechanical things before this can merge:

1. **Rebase onto current `main`** — it's currently conflicting. The conflict is the README row; see my reply on that thread, it should be dropped rather than re-applied.
2. **Changelog placement** — following up on @jcarius-nv's note: now that #3145 has merged, the entries your `--render-fps` line currently sits beside have been moved down into a released section. After rebasing, please move the entry into the top-level `[Unreleased] → Added` (it currently holds just the `cloth_stiff_material_*` entry) so it doesn't land under `1.3.0`.

Thanks for your patience here!

devshahofficial:
@adenzler-nvidia I pushed the requested mechanical follow-up.

Changes:
- Rebased onto current `main`.
- Dropped the README `--render-fps` row per the updated example-options table policy.
- Moved the `--render-fps` changelog entry into the top-level `[Unreleased] -> Added` section.

Validation:
```bash
uv run --extra dev -m newton.tests -k test_render_fps_cli
uvx pre-commit run -a
```

adenzler-nvidia (approved):
Re-reviewed the latest revision. All earlier asks are addressed: rebased cleanly onto current main, the README options-table row dropped per the slimmed-down table policy (#3070), and the changelog entry moved into the top-level `[Unreleased] → Added`.

The throttle logic is correct — both the idle and render paths are capped, and frames that already exceed the target period correctly skip the sleep. The injected clock/sleep functions make the behavior deterministically testable, and the test suite passes locally.

LGTM.

j3soon:
I can confirm this works by pulling the latest main, thanks for the PR!

PR #2432, opened 2026-04-13
Simplify private FeatherPGS API to matrix-free-only path

dylanturpin:
## Description

This PR narrows the private FeatherPGS API line to the current matrix-free path instead of keeping the older dense and split ablation surface alive.

It removes the private solver constructor knobs for `pgs_mode` and per-stage kernel selection, renames the remaining articulated-row tuning surface from `dense_*` to mode-neutral `contact_compliance` and `max_constraints`, and runs the private FeatherPGS step path as matrix-free only.

The rationale matches the published nightly ablations from `2026-04-01T20-49-30Z` on `gh-pages`: the matrix-free path materially outperforms split on the tabletop ablation, and the tiled `hinv_jt` / tiled PGS / parallel-stream path already represents the current winner direction for the private line.

## Checklist

- [x] New or existing tests cover these changes
- [ ] The documentation is up to date with these changes
- [x] `CHANGELOG.md` has been updated (if user-facing change)

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_feather_pgs
uvx pre-commit run -a
```

## New feature / API change

```python
from newton._src.solvers.feather_pgs import SolverFeatherPGS

solver = SolverFeatherPGS(
    model,
    contact_compliance=0.0,
    max_constraints=32,
    mf_max_constraints=512,
)
```

## Summary by CodeRabbit

* **New Features**
  * FeatherPGS solver now enforces matrix-free-only computation, removing previously available matrix-based configuration options.

* **Refactor**
  * Updated velocity handling in kernel computation to align with matrix-free implementation requirements.

* **Tests**
  * Added test coverage validating FeatherPGS operates exclusively in matrix-free mode and verifies solver constraint enforcement.

PR #2608, opened 2026-04-27
Add private pure Warp FeatherPGS

dylanturpin:
## Description

This draft PR adds a private FeatherPGS solver package under `newton._src.solvers.feather_pgs` and keeps it narrowed to the current pure-Warp matrix-free path.

It exposes only the private import:

```python
from newton._src.solvers.feather_pgs import SolverFeatherPGS
```

The solver uses the fused Warp tile backend contributed by @zcorse. It packs each contact triplet into the matrix-free path, computes articulated/contact response without materializing the full dense Delassus solve path, and runs the PGS contact update through Warp tile operations such as `wp.tile_load`, `wp.tile_dot`, `wp.tile_axpy`, scatter helpers, and tiled `J H^-1 J^T` work. The branch keeps the actuator-only effort-limit semantics and joint velocity-limit row support, while removing public exports and path-selection/ablation controls from the private API.

This supersedes draft PR #2432 (`dturpin/fpgs-private-api-matrix-free`).

## Checklist

## Test plan

Targeted checks run during branch prep:

```bash
uv run ruff check newton/_src/solvers/feather_pgs newton/_src/solvers/__init__.py newton/tests/test_feather_pgs_effort_limit.py newton/tests/test_feather_pgs_velocity_limit.py newton/tests/test_feather_pgs_fused_warp.py newton/tests/test_feather_pgs_private_api.py newton/tests/test_feather_pgs_free_root_velocity.py
uv run --extra dev -m newton.tests -k test_feather_pgs_effort_limit
uv run --extra dev -m newton.tests -k test_feather_pgs_velocity_limit
uv run --extra dev -m newton.tests -k test_feather_pgs_fused_warp
uv run --extra dev -m newton.tests -k test_feather_pgs_private_api
uv run --extra dev -m newton.tests -k test_feather_pgs_free_root_velocity
```

Additional manual benchmark on an RTX PRO 6000 Blackwell Server Edition, using the integration branch before the private-API trim:

| Scenario | Worlds | Existing matrix-free | Pure-Warp fused tile |
|---|---:|---:|---:|
| `g1_flat` | 16,384 | 969,889 env-FPS | 945,066 env-FPS |
| `h1_tabletop` | 8,192 | 121,493 env-FPS | 108,015 env-FPS |

## New feature / API change

Private import only:

```python
from newton._src.solvers.feather_pgs import SolverFeatherPGS
```

## Summary by CodeRabbit

* **New Features**
  * Added a GPU-accelerated FeatherPGS solver with fused/matrix-free contact & joint solving, multi-articulation/world batching, improved free-root integration, and explicit handling of joint effort/velocity limits and contact-force updates.

* **Tests**
  * Added extensive regression and CUDA smoke tests covering effort/velocity limits, free-root velocity semantics, fused execution, body-parent force outputs, invalid-model validation, and private-API checks.

* **Chores**
  * Made the FeatherPGS solver type available via the package namespace.


newton/_src/solvers/feather_pgs/__init__.py
```diff
@@ -0,0 +1,20 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 The Newton Developers
```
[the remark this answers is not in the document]
dylanturpin: Fixed in 875c7037: updated the new package initializer SPDX year to 2026.

newton/tests/test_feather_pgs_velocity_limit.py
```diff
@@ -0,0 +1,623 @@
...
+    # Solver B: constructed with enable_joint_velocity_limits=False (explicit).
+    model_b, s0_b, s1_b, ctl_b = _build_driven_pendulum(
+        device, velocity_limit=velocity_limit, target_ke=2000.0, target_kd=40.0
+    )
+    solver_b = SolverFeatherPGS(model_b, enable_joint_velocity_limits=False)
+    peak_b = _step_and_read_qdot(model_b, solver_b, s0_b, s1_b, ctl_b, dt=1.0 / 120.0, n_steps=20)
+
+    # Bit-exact match: flag-off is a strict no-op.
+    test.assertAlmostEqual(peak_a, peak_b, delta=1e-6)
+
```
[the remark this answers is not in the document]
dylanturpin: Fixed in 875c7037: added a per-step joint_qd history helper and changed the flag-off no-op test to compare the full velocity trace rather than only the peak.

newton/tests/test_feather_pgs_velocity_limit.py
```diff
@@ -0,0 +1,623 @@
...
+   into the grouped Jacobian and sets the correct constraint metadata
+   (row type, zero Baumgarte bias, ``target_vel = -qdot_max``).
+4. **Flag-off is a strict no-op** — constructing the solver without the flag
+   matches the baseline behaviour bit-for-bit.
+
+The end-to-end Isaac-Lab-level smoke comparison (``fpgs`` vs ``fpgs_vlim`` vs
+``physx`` vs ``mjwarp``) lives in
+``skild-IL-solver/scripts/benchmarks/velocity_spike_smoke.py`` and is reported
+in ``notes/investigations/velocity-spike/smoke-scenario.md``.
+"""
```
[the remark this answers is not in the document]
dylanturpin: Fixed in 875c7037: removed the monorepo/private-path references from the velocity-limit test module docstring.

dylanturpin:
CodeRabbit follow-up in 875c7037:

- Fixed SPDX year in the new FeatherPGS package initializer.
- Tightened the effort-limit test wording so it does not overclaim about a buffer the kernel does not receive.
- Removed stale monorepo/private-path references from the velocity-limit test docstring.
- Replaced the flag-off peak-only comparison with a full per-step joint_qd trace comparison.
- Replaced the hard-coded fused-Warp contact-triplet count with max_constraints // 3.

I intentionally left the lower-bound suggestion on the flag-on peak test unchanged: the companion flag-on-vs-flag-off test already proves the scenario is an active witness by requiring the baseline to overshoot and the constrained solver to be tighter. I also left raw device suffixes in generated test names because this file follows the existing Newton unittest pattern.

dylanturpin:
Follow-up pushed:
- Narrowed FeatherPGS to the single fused Warp matrix-free path.
- Removed dead modes/kernels and stale knobs; velocity limits are always on for finite limits.
- Added body_parent_f support, kinematic guard coverage, and tighter tests.

dylanturpin:
Addressed CodeRabbit follow-ups: validate FeatherPGS constructor capacities, reject invalid step dt, and clamp dense contact row counts through the finalize path. Local checks: private API tests, fused Warp tests, pre-commit.

daedalus5:
The constructor docstring for `SolverFeatherPGS` duplicates types in its `Args:` entries:

```python
angular_damping (float, optional): Angular damping factor. Defaults to 0.05.
```

`AGENTS.md` states: *"Google-style docstrings. Types in annotations, not docstrings. `Args:` use `name: description`."* This matches what `SolverFeatherstone.__init__` already does:

```python
angular_damping: Angular damping factor. Defaults to 0.05.
```

Affects ~15 parameter entries in `solver_feather_pgs.py`. Suggest dropping the `(type, optional)` prefixes throughout the constructor docstring.

daedalus5:
The class-level docstring on `SolverFeatherPGS` is much lighter than other Newton solvers. For comparison, `SolverFeatherstone`'s class docstring covers:

- Algorithm description + paper reference
- Sphinx cross-refs to sibling solvers (e.g. `` :class:`~newton.solvers.SolverSemiImplicit` ``)
- Supported / unsupported joint type matrix
- Documentation of extended state attributes (`body_parent_f`) with units
- A `.. code-block:: python` usage example
- A `` :ref:`Joint feature support` `` link

Suggest expanding to include at minimum: a usage example, supported joint types, and Sphinx cross-refs to `` :class:`~newton.solvers.SolverFeatherstone` `` since FeatherPGS is structurally a sibling. Even if the solver stays a "private prototype," keeping the docstring shape consistent with the rest of the solver library helps future readers.

daedalus5:
`TiledKernelFactory` introduces a class-based, `ClassVar`-cached factory pattern for size-specialized kernels (key: `(dimensions, device.arch)`). No other Newton solver currently uses this pattern — the closest existing precedent is `featherstone/kernels.py`'s `create_inertia_matrix_kernel(num_joints, num_dofs)`, a plain function called once at init.

The class-based caching is justified by heterogeneous multi-articulation needs (per-DOF specialization across articulations in the same model), but it's establishing a new convention in `newton/_src/solvers/`. We should get approval from other Newton devs before this lands as a precedent — I'll follow up and ping specific people.

daedalus5:
The PR adds 5 feature-split test files (`test_feather_pgs_effort_limit.py`, `test_feather_pgs_velocity_limit.py`, `test_feather_pgs_fused_warp.py`, `test_feather_pgs_private_api.py`, `test_feather_pgs_free_root_velocity.py`) rather than a single primary test file. There's no `test_solver_feather_pgs.py` / `test_feather_pgs.py` covering general smoke or integration cases.

Existing test naming across Newton is already inconsistent (`test_solver_xpbd.py`, `test_mujoco_solver.py`, `test_implicit_mpm.py`, `test_menagerie_mujoco.py`), so there's no single "right" pattern to match. But splitting one solver's coverage across 5 narrow files is novel for the project. We should get approval from other Newton devs on the test layout convention — I'll follow up and ping specific people.

dylanturpin:
Converted to draft while I add fixes from [here](https://notes-dturpin.dylanturpin.workers.dev/fpgs-change-inventory/#newton-changes). With all fixes FPGS training on IsaacLab franka cube lift and g1 locoomtion tasks matches physX. Need to write tests for the bug fix changes (and clean up implementations) + ablate which feature additions are actually required for successful training (or otherwise useful).

preist-nvidia:
@dylanturpin - setting to 1.4 for tracking for now, but of course you can merge earlier and make 1.3 too.

dylanturpin:
Refreshed the PR branch to `ba87fb6d` on top of current `upstream/main`.

Summary:
- Ported Zach/Daedalus heterogeneous-world fixes from `dylanturpin/newton-collab#1`: per-world DOF counts, padded `world_velocity`, contact triplet padding, unsupported free-joint / zero-DOF guards, contact metadata capacity checks, same-`Contacts` update guard, and heterogeneous velocity-limit coverage.
- Replaced `TiledKernelFactory` with private module-level cached kernel factory functions, then pre-resolved the size-specialized kernels in `SolverFeatherPGS.__init__` so per-step launches use stored `wp.Kernel` objects.
- Consolidated the five split FeatherPGS test files into `newton/tests/test_solver_feather_pgs.py`; the PR file surface is back to the private solver package plus one test module.
- Expanded the class docstring with algorithm shape, joint support limits, `body_parent_f` semantics/units, usage, and sibling solver cross-refs. Cleaned constructor Args to Google style without duplicated types.
- Preserved the previous CodeRabbit fixes: SPDX year, effort-limit wording, stale external path removal, per-step no-op velocity trace comparison, constructor/dt validation, and dense contact row count clamping.
- Kept finite joint-limit activation-gap work out of this PR, and did not add public exports, IsaacLab integration, broad PGS modes, or Buildkite scaffolding.

Validation:
- `uvx pre-commit run -a` passed.
- `uv run --extra dev -m newton.tests -k solver_feather_pgs` passed: 35 tests in 79.871s on CPU plus `cuda:0` (RTX 3090).

newton/_src/solvers/feather_pgs/kernels.py
```diff
@@ -0,0 +1,4167 @@
...
+    """
+    world = wp.tid()
+    m = world_constraint_count[world]
+
+    for i in range(m):
+        if world_row_type[world, i] == PGS_CONSTRAINT_TYPE_CONTACT:
+            world_diag[world, i] += contact_alpha
+
+
+# =============================================================================
```
daedalus5, line 3451: Old comments?

newton/_src/solvers/feather_pgs/kernels.py
```diff
@@ -0,0 +1,4167 @@
...
+    Thread dimension: n_arts_of_size (one thread per articulation in this size group)
+    """
+    idx = wp.tid()
+    art = group_to_art[idx]
+    dof_start = articulation_dof_start[art]
+    for i in range(n_dofs):
+        joint_qdd[dof_start + i] = qdd_group[idx, i, 0]
+
+
+# =============================================================================
```
daedalus5, line 3589: Also nothing following this note.

daedalus5 (commented):
Review notes from a solver-path/performance pass.

The live step path looks like the intended fused Warp tiled velocity-space Jacobi implementation. I did not find any `func_native` use in the new Feather PGS files, and the heterogeneous-world regression from dylanturpin/newton-collab#1 appears to be included and passing. I would not block this on numerical behavior from this review; the items below are about keeping the implementation single-path and tightening avoidable per-step work.

Findings:

1. Medium: unused leftover kernels remain in the new package.

   PR #2608 appears to launch only the fused velocity-space Jacobi path, but `newton/_src/solvers/feather_pgs/kernels.py` still carries kernels/helpers that are not reachable from `SolverFeatherPGS`:

   - `compute_velocity_predictor`
   - `build_contact_rows_normal`
   - `accumulate_contact_jacobian_matrix_free`
   - `clamp_contact_counts`
   - `apply_augmented_mass_diagonal`

   The solver imports and launches the replacement/current kernels instead, for example `compute_velocity_predictor_and_seed_world_velocity`, `allocate_world_contact_slots`, `populate_world_J_for_size`, `build_mf_contact_rows`, `finalize_world_constraint_counts`, `pack_contact_triplets_vec3`, and `apply_augmented_mass_diagonal_grouped`.

   This conflicts with the goal of maintaining a single solver path. Even if these symbols are private, they make future solver maintenance and performance review harder because stale contact-build/mass-update paths can drift or be accidentally revived. I recommend deleting these unused kernels/helpers and any associated stale fallback comments unless there is a concrete launch path and test coverage for them.

2. Medium: matrix-free setup work runs too broadly.

   In `SolverFeatherPGS.step()`, the MF setup block is gated by `self._has_free_rigid_bodies`. That means any model with free rigid bodies launches `_mf_pgs_setup`, `compute_mf_world_dof_offsets`, and `pack_mf_meta` even when there are no active MF contact rows. `pack_mf_meta` also writes the full `(world_count, mf_max_constraints)` range with no active-row guard.

   With the default `mf_max_constraints=512`, free-body scenes without free/free or free/ground MF rows still pay extra kernels and full metadata traffic each step. At minimum, this should skip when `contacts is None` or `rigid_contact_max == 0`, and `pack_mf_meta` should guard on `mf_constraint_count`. Ideally, offset/meta packing would be fused into the active-row effective-mass setup so inactive capacity is not touched.

3. Low: comments reference an internal note that is not present in the tree.

   Several comments cite `notes/investigations/velocity-spike/physx-deep-dive.md`, but that path is not present. Because these comments justify solver ordering/math, future maintainers need either the note checked in, a public/repo doc link, or enough inline rationale to preserve the reasoning.

Verification run locally on the PR branch:

```bash
uv run --extra dev -m newton.tests -k test_solver_feather_pgs
uvx ruff check newton/_src/solvers/feather_pgs newton/tests/test_solver_feather_pgs.py
git diff --check main...review-pr-2608
git grep -n "func_native" review-pr-2608 -- newton/_src/solvers/feather_pgs newton/tests/test_solver_feather_pgs.py
```

Results: tests passed, `uvx ruff` passed, whitespace check passed, and no `func_native` hits were found in the PR files.

newton/_src/solvers/feather_pgs/solver_feather_pgs.py
```diff
@@ -0,0 +1,2540 @@
...
+            import newton
+            from newton._src.solvers.feather_pgs import SolverFeatherPGS
+
+            builder = newton.ModelBuilder()
+            model = builder.finalize(device="cuda")
+            solver = SolverFeatherPGS(model, pgs_iterations=12)
+            state_0 = model.state()
+            state_1 = model.state()
+            solver.step(state_0, state_1, model.control(), None, 1.0 / 240.0)
+    """
```
adenzler-nvidia, line 139: could you label this as experimental, similar to the PRs in-flight for Kamino/VBD? using the sphinx experimental tag.

newton/_src/solvers/feather_pgs/solver_feather_pgs.py
```diff
@@ -0,0 +1,2540 @@
...
+    update_articulation_origins,
+    update_articulation_root_com_offsets,
+    update_body_qd_from_featherstone,
+    update_qdd_from_velocity,
+    update_qdd_from_world_velocity,
+)
+
+
+class SolverFeatherPGS(SolverBase):
+    """Private CUDA-only FeatherPGS prototype using one fused Warp path.
```
adenzler-nvidia, line 96: should probably be GPU-only. What prevents this from running on CPU?

newton/_src/solvers/feather_pgs/solver_feather_pgs.py
```diff
@@ -0,0 +1,2540 @@
...
+            solver = SolverFeatherPGS(model, pgs_iterations=12)
+            state_0 = model.state()
+            state_1 = model.state()
+            solver.step(state_0, state_1, model.control(), None, 1.0 / 240.0)
+    """
+
+    def __init__(
+        self,
+        model: Model,
+        angular_damping: float = 0.05,
```
adenzler-nvidia, line 144: angular_damping looks like something that should be per-body. Should we promote this to a custom attribute?

newton/_src/solvers/feather_pgs/solver_feather_pgs.py
```diff
@@ -0,0 +1,2540 @@
...
+        update_mass_matrix_interval: int = 1,
```
adenzler-nvidia, line 145: are you actively exercising non-1 paths for this? does it help, what are the tradeoffs?

newton/_src/solvers/feather_pgs/solver_feather_pgs.py
```diff
@@ -0,0 +1,2540 @@
...
+        friction_smoothing: float = 1.0,
+        enable_contact_friction: bool = True,
```
adenzler-nvidia, line 147: not sure if long-term it's a good idea to have a global switch for friction, couldn't the same be achieved more locally? is it really desirable to have simulations without friction?

newton/_src/solvers/feather_pgs/solver_feather_pgs.py
```diff
@@ -0,0 +1,2540 @@
...
+        enable_joint_limits: bool = False,
```
adenzler-nvidia, line 148: some point for joint limits, I think long-term it should just be a limited config on each joint? and for debugging, it could be a more local flag?

newton/_src/solvers/feather_pgs/solver_feather_pgs.py
```diff
@@ -0,0 +1,2540 @@
...
+        pgs_iterations: int = 12,
+        pgs_beta: float = 0.2,
+        pgs_cfm: float = 1.0e-6,
+        contact_compliance: float = 0.0,
```
adenzler-nvidia, line 152: is contact compliance something global or should it be per shape (pair?)

newton/_src/solvers/feather_pgs/solver_feather_pgs.py
```diff
@@ -0,0 +1,2540 @@
...
+        self.pgs_omega = pgs_omega
+        self.max_constraints = max_constraints
+        self._max_constraints_padded = ((max_constraints + 2) // 3) * 3
+        self.mf_max_constraints = mf_max_constraints
+        self._double_buffer = True
+        self._nvtx = False
+        if not model.device.is_cuda:
+            raise ValueError("SolverFeatherPGS requires a CUDA device.")
+        if not enable_contact_friction:
+            raise ValueError("SolverFeatherPGS currently requires enable_contact_friction=True.")
```
adenzler-nvidia, line 218: shouldn't we just remove the option from the constructor based on this condition?

newton/_src/solvers/feather_pgs/kernels.py
```diff
@@ -0,0 +1,4167 @@
...
+    # compute body forces
+    X_sm = body_q_com[child]
+    X_sm_local = wp.transform(
+        wp.transform_get_translation(X_sm) - origin,
+        wp.transform_get_rotation(X_sm),
+    )
+    I_m = body_I_m[child]
+
+    # gravity and external forces (expressed in frame aligned with s but centered at body mass)
+    m = I_m[0, 0]
```
adenzler-nvidia, line 854: looks like you're computing the spatial matrix just to fetch component [0, 0] ? can we remove compute_spatial_inertia?

newton/_src/solvers/feather_pgs/kernels.py
```diff
@@ -0,0 +1,4167 @@
...
+
+@wp.kernel
+def compute_com_transforms(
+    body_com: wp.array[wp.vec3],
+    # outputs
+    body_X_com: wp.array[wp.transform],
+):
+    tid = wp.tid()
+    com = body_com[tid]
+    body_X_com[tid] = wp.transform(com, wp.quat_identity())
```
adenzler-nvidia, line 82: Is realize this kernel just transforms to a wp.transform, but wouldn't it make more sense to construct that in the downstream user kernels? I don't see a place where you ever write to body_X_com, so the transform part will always be identity?

newton/_src/solvers/feather_pgs/kernels.py
```diff
@@ -0,0 +1,4167 @@
...
+    body_v_s: wp.array[wp.spatial_vector],
+    body_f_s: wp.array[wp.spatial_vector],
+    body_a_s: wp.array[wp.spatial_vector],
+):
+    type = joint_type[i]
+    child = joint_child[i]
+    parent = joint_parent[i]
+    articulation = joint_articulation[i]
+    qd_start = joint_qd_start[i]
+    origin = wp.vec3()
```
adenzler-nvidia, line 802: an explicit zero-initialization would signal the fallback a bit better here, I think.

newton/_src/solvers/feather_pgs/kernels.py
```diff
@@ -0,0 +1,4167 @@
...
+
+
+@wp.kernel
+def convert_root_free_qd_local_to_world(
+    articulation_root_is_free: wp.array[int],
+    articulation_root_dof_start: wp.array[int],
+    articulation_root_com_offset: wp.array[wp.vec3],
+    # in/out
+    qd: wp.array[float],
+):
```
adenzler-nvidia, line 173: would it make more sense to precompute a list of free joints, and then only launch as many threads as free joints here?

newton/_src/solvers/feather_pgs/kernels.py
adenzler-nvidia: edit, forget about this my search didn't hit the case below.

newton/_src/solvers/feather_pgs/kernels.py
```diff
@@ -0,0 +1,4167 @@
...
+            body_q_com,
+            joint_X_p,
+            articulation_origin,
+            gravity,
+            joint_S_s,
+            body_I_s,
+            body_v_s,
+            body_f_s,
+            body_a_s,
+        )
```
adenzler-nvidia, line 925: could we exploit topology here, try not reloading values if we move from parent->child? I'm not sure if actually guarantee an ordering, but mujoco for sure expects one.

newton/_src/solvers/feather_pgs/kernels.py
```diff
@@ -0,0 +1,4167 @@
...
+    joint_count = joint_end - joint_start
+
+    M_offset = articulation_M_start[index]
+
+    for l in range(joint_count):
+        I = body_I_s[joint_start + l]
+        block = M_offset + l * 36
+        for row in range(6):
+            for col in range(6):
+                M_blocks[block + row * 6 + col] = I[row, col]
```
adenzler-nvidia, line 1068: would it make sense to build a one-time remapping from flat indices of M_blocks to body_I_s indices and launch a flat kernel to run more or less at memcopy speed?

newton/_src/solvers/feather_pgs/kernels.py
```diff
@@ -0,0 +1,4167 @@
...
+    joint_q: wp.array[float],
+    joint_qd: wp.array[float],
+    joint_qdd: wp.array[float],
+    dt: float,
+    # outputs
+    joint_q_new: wp.array[float],
+    joint_qd_new: wp.array[float],
+):
+    # one thread per-articulation
+    index = wp.tid()
```
adenzler-nvidia, line 1118: this comment confused me, as it's parallelized over joints

newton/_src/solvers/feather_pgs/kernels.py
```diff
@@ -0,0 +1,4167 @@
...
+@wp.kernel
+def compute_velocity_predictor(
+    joint_qd: wp.array[float],
+    joint_qdd: wp.array[float],
+    dt: float,
+    # outputs
+    v_hat: wp.array[float],
+):
+    tid = wp.tid()
+    v_hat[tid] = joint_qd[tid] + joint_qdd[tid] * dt
```
adenzler-nvidia, line 1153: looks like this kernel is unused

newton/_src/solvers/feather_pgs/kernels.py
```diff
@@ -0,0 +1,4167 @@
...
+    if local_d >= world_dof_count[world]:
+        return
+    v_new[world_dof_start[world] + local_d] = world_velocity[world, local_d]
+
+
+@wp.func
+def contact_tangent_basis(n: wp.vec3):
+    # pick an arbitrary perpendicular vector and orthonormalize
+    tangent0 = wp.cross(n, wp.vec3(1.0, 0.0, 0.0))
+    if wp.length_sq(tangent0) < 1.0e-12:
```
adenzler-nvidia, line 1229: I think we have a minval constant for things like these.

newton/_src/solvers/feather_pgs/kernels.py
```diff
@@ -0,0 +1,4167 @@
...
+    mu = 0.0
+    mat_count = 0
+    if shape_a >= 0:
+        mu += shape_material_mu[shape_a]
+        mat_count += 1
+    if shape_b >= 0:
+        mu += shape_material_mu[shape_b]
+        mat_count += 1
+    if mat_count > 0:
+        mu /= float(mat_count)
```
adenzler-nvidia, line 1439: you can assume that there is no contact with a negative shape here, so the math becomes a bit simpler.

newton/_src/solvers/feather_pgs/kernels.py
```diff
@@ -0,0 +1,4167 @@
...
+    n = -contact_normal[tid]
+    shape_a = contact_shape0[tid]
+    shape_b = contact_shape1[tid]
+
+    body_a = -1
+    body_b = -1
+    if shape_a >= 0:
+        body_a = shape_body[shape_a]
+    if shape_b >= 0:
+        body_b = shape_body[shape_b]
```
adenzler-nvidia, line 1411: same thing here, the shape cannot be negative for a valid contact

newton/_src/solvers/feather_pgs/kernels.py
```diff
@@ -0,0 +1,4167 @@
...
+    point_b_local = contact_point1[tid]
+    point_a_world = wp.vec3(0.0)
+    point_b_world = wp.vec3(0.0)
+
+    if body_a >= 0:
+        X_wb_a = body_q[body_a]  # World-from-Body transform
+        # Contact points are stored in body frame by collision detection
+        point_a_world = wp.transform_point(X_wb_a, point_a_local) - thickness_a * n
+    else:
+        point_a_world = point_a_local - thickness_a * n
```
adenzler-nvidia, line 1451: maybe makes sense to just branch for the transform init here and keep the actual math a one-liner?

newton/_src/solvers/feather_pgs/kernels.py
```diff
@@ -0,0 +1,4167 @@
...
+            tangent_base_1,
+            Jc_out,
+        )
+
+        # Friction row 2 at base_slot + 2
+        row_index_2 = articulation * max_constraints + base_slot + 2
+        tangent_base_2 = row_index_2 * max_dofs
+
+        for col in range(max_dofs):
+            Jc_out[tangent_base_2 + col] = 0.0
```
adenzler-nvidia, line 1586: did you try making some of these max_* a wp.static?

newton/_src/solvers/feather_pgs/kernels.py
```diff
@@ -0,0 +1,4167 @@
...
+            coord_index = coord_start + axis
+
+            ke = joint_target_ke[dof_index]
+            kd = joint_target_kd[dof_index]
+            if ke <= 0.0 and kd <= 0.0:
+                continue
+
+            K = ke * dt * dt + kd * dt
+            if K <= 0.0:
+                continue
```
adenzler-nvidia, line 1698: should we really check for negative stiffness/damping here? I don't think that's defined

Issue #2009, opened 2026-03-09
shape_material_ke/kd units mismatch for MuJoCo solver: mass-normalized vs force-based


preist-nvidia:
## Summary

When importing USD assets with `SchemaResolverMjc`, the contact stiffness and damping stored in `Model.shape_material_ke` / `shape_material_kd` are **mass-normalized** (`[1/s²]` and `[1/s]`), derived from MuJoCo's `solref = [timeconst, dampratio]` via:

```python
ke = 1 / (timeconst² × dampratio²)   # [1/s²]
kd = 2 / timeconst                    # [1/s]
```

However, `Model.shape_material_ke` is documented as `[N/m]` and `shape_material_kd` as `[N·s/m]` — force-based, not mass-normalized. This discrepancy also affects `joint_limit_ke`/`kd` (documented as `[N/m or N·m/rad]` / `[N·s/m or N·m·s/rad]`).

## Why the simulation is consistent in the default case

`SolverMuJoCo` converts `ke`/`kd` back to `solref` via `convert_solref(ke, kd, d_width=1.0, d_r=1.0)` — the inverse of the import formula. This makes the roundtrip exact **under the approximation `d_width = d_r = 1.0`** (i.e., flat solimp).

The roundtrip breaks for any user that:
1. **Domain-randomizes `shape_material_ke`/`kd`** — perturbations applied in `[N/m]` semantics land at wrong `solref` values because the stored baseline is in `[1/s²]`.
2. **Uses non-default `solimp`** — the `d_width` / `d_r` factors are approximated as 1.0 at import time but used with their actual values at solve time, breaking the inverse.
3. **Uses `ke`/`kd` with any other solver** (SemiImplicit, Featherstone) after a MuJoCo import — those solvers treat the values as force-based and get incorrect contact response.

## Proposed fixes

Two complementary directions:

1. **Introduce mass-normalized (acceleration spring) variants** of `shape_material_ke`/`kd` — e.g., `shape_material_ke_normalized [1/s²]` / `shape_material_kd_normalized [1/s]` — with clear unit documentation. MuJoCo import writes here; `SolverMuJoCo` reads here.

2. **Expose `mjc:solref` directly as a custom solver attribute** so users domain-randomize in the native `[timeconst, dampratio]` space rather than going through the ke/kd conversion at all. This is the most faithful representation and avoids any solimp approximation.

## Related

[...]

PR #2610, opened 2026-04-27
Fix MuJoCo joint-limit ke/kd scaling

eric-heiden:
## Summary
Fix the MuJoCo joint-limit path so Newton force-space `joint_limit_ke` / `joint_limit_kd` produce the expected `jnt_solref`, including runtime updates after inertia changes and `set_const`.

## Scope
Partially addresses #2009 for joint limits only.
`shape_material_ke` / `shape_material_kd` are intentionally out of scope and will follow in a separate PR.

## What changed
- Scale `jnt_solref` by `dof_invweight0 * (1 - dmax)` when converting Newton joint-limit stiffness and damping to MuJoCo `solref`.
- Recompute the scaling on runtime model updates so the Warp and CPU MuJoCo paths stay in sync after inertia changes or `set_const`.
- Add regression coverage for initial export, runtime mass changes, disabling limit stiffness, CPU parity, and torque response.
- Keep the shape-material work out of this PR and update the changelog entry to match the reduced scope.

## Touched areas
- `CHANGELOG.md`
- `newton/_src/solvers/mujoco/kernels.py`
- `newton/_src/solvers/mujoco/solver_mujoco.py`
- `newton/tests/test_mujoco_solver.py`

## Validation
- `uv run --extra dev -m newton.tests -k joint_limit`

## Summary by CodeRabbit

* **Bug Fixes**
  * Joint-limit stiffness/damping are now scaled by per-DOF inverse-weight and margin so MuJoCo’s effective response matches configured force-space gains; falls back to MuJoCo defaults when limits are disabled.

* **New Features**
  * Preserves authored per-joint MuJoCo solref values on import/export and uses them as overrides; runtime recomputation ensures exported models reflect runtime solref.

* **Documentation**
  * Changelog and MuJoCo integration docs updated to explain scaling and import/export semantics.

* **Tests**
  * Added regression tests covering initialization, runtime recompute, export, backend recompute, disabled limits, and edge cases.


devshahofficial:
@eric-heiden I’d probably avoid forcing a shared helper between Python and the Warp kernel here. A small parity table test may be a cleaner contract: cover `ke <= 0`, `invw == 0`, `dmax >= 1`, and normal scaling, then assert CPU/kernel behavior stays aligned. Happy to sketch that if useful.

eric-heiden:
@devshahofficial Addressed in the latest commit. I added parity coverage for the CPU and Warp scaling cases you called out, including `ke <= 0`, `invw == 0`, `dmax >= 1`, and the normal scaling path.

newton/_src/solvers/mujoco/kernels.py
```diff
@@ -1745,8 +1745,17 @@ def update_body_inertia_kernel(
     if newton_body < 0:
         return
 
+    inertia = body_inertia[newton_body]
+    off_diag = wp.abs(inertia[0, 1]) + wp.abs(inertia[0, 2]) + wp.abs(inertia[1, 2])
+    if off_diag <= 1.0e-12:
```
adenzler-nvidia: can this be maxval?

adenzler-nvidia (commented):
One follow-up question on the MJCF import path — can wait for the prior thread to settle if you'd rather batch.

newton/_src/utils/import_mjcf.py
```diff
@@ -1591,8 +1592,11 @@ def parse_body(
                     parsing_mode="mjcf",
                     context={"use_degrees": use_degrees, "joint_type": joint_type_str},
                 )
-                # assemble custom attributes for each DOF (dict mapping DOF index to value)
-                # Only store values that were explicitly specified in the source
+                if builder.has_custom_attribute("mujoco:solreflimit"):
+                    dof_attr.setdefault("mujoco:solreflimit", raw_solreflimit)
```
adenzler-nvidia: `solreflimit` is registered as a `mujoco:solreflimit` custom attribute with `mjcf_attribute_name="solreflimit"`, which already gives `parse_custom_attributes(parsing_mode="mjcf")` automatic parsing — and the auto-parser intentionally only fills in attributes that were explicitly authored, leaving unauthored joints at the registered default `(0.0, 0.0)` sentinel.

The hand-rolled `parse_vec("solreflimit", (0.02, 1.0))` a few lines up plus this `dof_attr.setdefault("mujoco:solreflimit", raw_solreflimit)` duplicate that parsing and replace the registered sentinel with `(0.02, 1.0)` for *every* MJCF joint, even when the XML didn't author `solreflimit`. Net effect:

- MJCF-imported joint, no `solreflimit` in XML → stored as `(0.02, 1.0)`, kernel takes the `raw_solref != 0` branch, `joint_limit_ke`/`joint_limit_kd` mutations are silently no-ops.
- USD-imported / direct-builder-API joint, no `mjc:solreflimit` authored → stored as `(0.0, 0.0)` sentinel, kernel takes the `ke`/`kd` scaling branch, mutations propagate.

If preserving MuJoCo's native `(0.02, 1.0)` default for unauthored joints is the goal, would it be cleaner to do that on the kernel side (e.g. fall through to `wp.vec2(0.02, 1.0)` when both `raw_solref` is sentinel and `ke <= 0`), or by changing the registered `default=`, so MJCF/USD/builder all behave the same? Otherwise this asymmetry will surprise users who randomize `joint_limit_ke` at runtime.

devshahofficial:
I updated the follow-up branch for the remaining review items:

Branch: https://github.com/devshahofficial/newton/tree/devshahofficial/local-pr2610-followups
Latest commit: https://github.com/devshahofficial/newton/commit/52bdf70d

Changes:
- `kernels.py`: use the max off-diagonal magnitude for the axis-aligned inertia fast path.
- `import_mjcf.py`: preserve authored `solreflimit`, while leaving unauthored joints at the registered `(0, 0)` sentinel.
- `test_import_mjcf.py`: add a `SolverMuJoCo` runtime regression for the unauthored MJCF case. The test mutates `joint_limit_ke/kd`, calls `notify_model_changed(SolverNotifyFlags.JOINT_DOF_PROPERTIES)`, and checks that runtime `jnt_solref` follows the scaled Newton gains while authored solref joints remain unchanged.

Validation:
```bash
uv run --extra dev -m newton.tests -k test_solreflimit_parsing
uv run --extra dev -m newton.tests -k solreflimit
uv run --extra dev -m newton.tests -k body_inertia_eigendecomposition_determinant
uvx pre-commit run -a
```

newton/_src/solvers/mujoco/kernels.py
eric-heiden: Done

newton/_src/utils/import_mjcf.py
eric-heiden: Addressed in 97ed825. The latest commit removed the manual `dof_attr.setdefault("mujoco:solreflimit", raw_solreflimit)` path and leaves `parse_custom_attributes()` as the only source for `mujoco:solreflimit`. So unauthored MJCF joints keep the `(0.0, 0.0)` sentinel and still take the scaled `joint_limit_ke` / `joint_limit_kd` update path, while explicitly authored `solreflimit` values are preserved as native MuJoCo parameters.

I also added regression coverage for both branches: `test_mjcf_unauthored_solreflimit_ke_updates_are_not_shadowed` and `test_mjcf_authored_solreflimit_is_preserved_as_native_parameter`.

eric-heiden:
@devshahofficial Thanks, I've already made the changes, just forgot to push them earlier. See my replies in the earlier comments.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -5806,6 +5917,14 @@ def get_body_name(body_idx: int) -> str:
             # now complete the data from the Newton model
             self.notify_model_changed(SolverNotifyFlags.ALL)
 
+            if target_filename:
+                for mjc_jnt, solref in enumerate(self.mj_model.jnt_solref):
```
adenzler-nvidia, line 5953: This export path seems to make saved MJCF non-reimportable for default/implicit solref limits. When a limited joint keeps MuJoCo's default `(0.02, 1.0)`, this loop writes `spec.joints[mjc_jnt].solref_limit = solref`. In my repro (`add_mjcf` on a hinge that omits `solreflimit`, then `SolverMuJoCo(..., save_to_mjcf=path)`, then `ModelBuilder().add_mjcf(path)`), `spec.to_xml()` serialized the default as `solreflimit="0.02"`. Re-import then fails in `parse_vec(..., DEFAULT_LIMIT_SOLREF)` with `ValueError: Invalid number of arguments in vector constructor, expected 2 elements, got 3`.

Could we either avoid authoring default solref values here, or teach the MJCF parser to accept MuJoCo's one-value `solref/solreflimit` shorthand before `save_to_mjcf` starts producing it?

newton/_src/solvers/mujoco/kernels.py
```diff
@@ -2351,6 +2365,88 @@ def update_geom_properties_kernel(
...
+        return
+
+    solref_mode = SOLREF_MODE_FORCE_SPACE
+    if joint_limit_solref_mode:
+        solref_mode = joint_limit_solref_mode[newton_dof]
+
+    if joint_limit_solref:
+        raw_solref = joint_limit_solref[newton_dof]
+        raw_solref_is_set = raw_solref[0] != 0.0 or raw_solref[1] != 0.0
+        if solref_mode == SOLREF_MODE_RAW or raw_solref_is_set:
```
adenzler-nvidia: `SOLREF_MODE_FORCE_SPACE` does not actually override a nonzero `mujoco.solreflimit` here because `raw_solref_is_set` still forces the raw branch. I reproduced this by setting `model.mujoco.solreflimit = [[0.03, 0.7]]` and `model.mujoco.solreflimit_mode = SOLREF_MODE_FORCE_SPACE`; `SolverMuJoCo` still produced `jnt_solref [0.03, 0.7]` instead of the scaled `joint_limit_ke/kd` value. The CPU path has the same fallback at `solver_mujoco.py` line 6897.

If the mode attribute is present, I think it should be authoritative: take the raw branch only for `SOLREF_MODE_RAW`, and keep the `raw_solref_is_set` compatibility fallback only when `solreflimit_mode` is absent.

newton/_src/solvers/mujoco/kernels.py
```diff
@@ -2347,6 +2352,99 @@ def update_geom_properties_kernel(
...
+        # user edits those Newton gains, then fall through to force-space
+        # scaling below.
+        jnt_solref[world, mjc_jnt] = wp.vec2(DEFAULT_LIMIT_SOLREF_TIMECONST, DEFAULT_LIMIT_SOLREF_DAMPRATIO)
+        return
+
+    if ke <= 0.0:
+        # Restore MuJoCo's compiled default so runtime ``ke -> 0`` updates
+        # behave the same as a fresh model built without a custom limit solref.
+        jnt_solref[world, mjc_jnt] = wp.vec2(DEFAULT_LIMIT_SOLREF_TIMECONST, DEFAULT_LIMIT_SOLREF_DAMPRATIO)
+        return
```
adenzler-nvidia: **I1 — `kd <= 0` is asymmetric with `ke <= 0`.** The reset to MuJoCo's default solref only fires on `ke <= 0`. A user with `ke=2000, kd=0` falls through to the force-space branch and writes `solref = (-ke·factor, 0)`. Mixed signs break MuJoCo's negative-convention assumption: MuJoCo interprets the result as `(timeconst, dampratio)` with `dampratio=0`, which triggers a divide-by-zero in `k_eff = 1/(τ²·ζ²)` and can NaN out the constraint. Suggest extending this guard to `(ke <= 0.0) or (kd <= 0.0)` and emitting a one-shot `warnings.warn` from `_update_solref_from_invweight0` when the asymmetric case is detected (kernel can't warn from inside Warp).

newton/_src/solvers/mujoco/kernels.py
```diff
@@ -2347,6 +2352,99 @@ def update_geom_properties_kernel(
...
+    values are forwarded unchanged when present; MJCF joints that rely on the
+    implicit default keep MuJoCo's native ``(0.02, 1.0)`` until
+    ``joint_limit_ke``/``joint_limit_kd`` are changed by the user.
+    ``dof_invweight0`` is only valid after MuJoCo's ``set_const_0`` /
+    ``mj_setConst`` has run, so this kernel must be launched from
+    ``notify_model_changed`` after those calls (and once at initialisation right
+    after ``put_model``).
+    """
+    world, mjc_jnt = wp.tid()
+    newton_dof = mjc_jnt_to_newton_dof[world, mjc_jnt]
```
adenzler-nvidia, line 2415: **I2 — Multi-DOF joint coverage is missing.** The kernel relies on `jnt_dofadr[mjc_jnt]` per-DOF lookup, and the MJCF importer at `import_mjcf.py:1556-1559` writes `solreflimit_mode` across three DOF offsets for ball joints. None of the 16 new tests build a ball / free / universal / D6 joint, so the per-`mjc_jnt`-to-`newton_dof` mapping and the multi-DOF replication path are untested in this PR. Suggest adding a ball-joint pendulum (or free + revolute combo) that asserts the per-DOF `mjc_jnt` solref entries are independently scaled with the correct `jnt_dofadr` lookup. Also worth a case where `mjc_jnt_to_newton_dof` returns `-1` so the early-return branch is exercised.

newton/_src/solvers/mujoco/kernels.py
```diff
@@ -2347,6 +2352,99 @@ def update_geom_properties_kernel(
...
+    if joint_limit_solref_mode:
+        mode_present = True
+        solref_mode = joint_limit_solref_mode[newton_dof]
+
+    if joint_limit_solref:
+        raw_solref = joint_limit_solref[newton_dof]
+        if mode_present:
+            if solref_mode == SOLREF_MODE_RAW:
+                jnt_solref[world, mjc_jnt] = raw_solref
+                return
```
adenzler-nvidia, line 2434: **I3 — Authored `solreflimit` is forwarded with no domain validation.** When `solref_mode == SOLREF_MODE_RAW`, the raw vec2 is written verbatim. MuJoCo's solref domain is `(timeconst, dampratio)` with both > 0 (standard mode) OR both < 0 (negative stiffness/damping mode). Combined with the `parse_vec` shorthand change, `solreflimit="-2500"` becomes `(-2500, 1.0)` — mixed signs, which MuJoCo silently treats as `timeconst=2500s` and effectively disables the limit. Same risk for `timeconst==0` (div-by-zero in MuJoCo) and `dampratio==0`. Suggest validating same-sign + non-zero components in `_update_solref_from_invweight0` and warning on invalid entries; the kernel can't validate from inside Warp.

newton/_src/utils/import_mjcf.py
```diff
@@ -516,8 +525,18 @@ def parse_vec(attrib, key, default):
...
+            # ``save_to_mjcf`` round-trip relies on this for ``solreflimit``:
+            # ``spec.to_xml()`` serialises ``(0.02, 1.0)`` as ``solreflimit="0.02"``
+            # and the importer must reconstruct the full vector. Historically
+            # vec3 callers used the single value as a uniform replicate; vec2
+            # callers (solref) need the trailing component to come from the
+            # MuJoCo default, not from ``out[0]``.
+            if len(default) == 3:
+                return wp.types.vector(3, wp.float32)(out[0], out[0], out[0])
+            padded = (out[0], *(float(default[i]) for i in range(1, len(default))))
+            return wp.types.vector(len(default), wp.float32)(*padded)
```
adenzler-nvidia: **I4 — `parse_vec` shorthand silently changes semantics for ALL vec2+ MJCF attributes.** The new padding branch generalises to any `len(default) != 1`, but the justification (round-tripping `solreflimit="0.02"`) only holds when `default` is a MuJoCo solref pair. `parse_vec` is shared by every vec2/vec4/vec5 MJCF attribute; any caller relying on the historic "replicate to fill" behaviour now silently gets `(out[0], default[1], default[2], ...)` instead. No log, no whitelist, no warning. Suggest restricting the shorthand path to a whitelist of solref-shaped keys, or at minimum gating a `warnings.warn(..., stacklevel=2)` on `verbose` so unexpected callers surface in CI.

newton/_src/utils/import_mjcf.py
```diff
@@ -1538,13 +1570,13 @@ def parse_body(
...
-                solreflimit = parse_vec(joint_attrib, "solreflimit", (0.02, 1.0))
+                solreflimit = parse_vec(joint_attrib, "solreflimit", DEFAULT_LIMIT_SOLREF)
                 limit_ke, limit_kd = solref_to_stiffness_damping(solreflimit)
                 # Handle None return values (invalid solref)
                 if limit_ke is None:
-                    limit_ke = 2500.0  # From MuJoCo's default solref (0.02, 1.0)
+                    limit_ke = DEFAULT_LIMIT_KE  # From MuJoCo's default solref.
                 if limit_kd is None:
-                    limit_kd = 100.0  # From MuJoCo's default solref (0.02, 1.0)
+                    limit_kd = DEFAULT_LIMIT_KD  # From MuJoCo's default solref.
```
adenzler-nvidia, line 1631: **I5 — `solref_to_stiffness_damping` failure silently coerces to MuJoCo defaults.** When the helper returns `None` for a malformed `solreflimit`, the code substitutes `DEFAULT_LIMIT_KE/KD` and continues. Because the new branch below sets `solreflimit_mode = SOLREF_MODE_RAW` whenever `"solreflimit" in joint_attrib`, the same bad MJCF will subsequently forward the original (still invalid) `joint_limit_solref` to `jnt_solref` via the RAW path. So `ModelBuilder.joint_limit_ke` ends up at the safe default while `jnt_solref` ends up at the invalid raw value — two different physics outcomes from the same input, no diagnostic either way. Suggest `warnings.warn(f"Joint {joint_name}: invalid solreflimit {solreflimit_raw!r}; falling back to MuJoCo default")` and either skip storing the mode flag (so MJCF_DEFAULT is used) or hard-fail.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -5809,6 +5926,14 @@ def get_body_name(body_idx: int) -> str:
...
+                    if self.mj_model.jnt_limited[mjc_jnt]:
+                        spec.joints[mjc_jnt].solref_limit = solref
+                with open(target_filename, "w") as f:
+                    f.write(spec.to_xml())
+                    print(f"Saved mujoco model to {os.path.abspath(target_filename)}")
```
adenzler-nvidia, line 5968: **I6 — `save_to_mjcf` round-trip permanently changes runtime semantics.** This loop writes the post-scaled `(-ke·factor, -kd·factor)` into the spec for every limited joint. On re-import, the MJCF parser sees an explicit `solreflimit`, the new code path at `import_mjcf.py:1556` sets `solreflimit_mode = SOLREF_MODE_RAW`, and `_update_solref_from_invweight0` will then forward the raw value unchanged — so subsequent edits to `joint_limit_ke`/`joint_limit_kd` on the reloaded model are silently ignored. Save→load is not semantically idempotent: pre-save the gains flow through, post-load they don't. Suggest documenting this trap in `docs/integrations/mujoco.rst` (in the new "Joint-limit stiffness and damping" section), and considering whether to emit unscaled `(-ke, -kd)` so round-tripping keeps Newton as the source of truth.

newton/_src/solvers/mujoco/constants.py
```diff
@@ -0,0 +1,45 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
+
+"""Shared constants for Newton's MuJoCo solver integration."""
+
+from __future__ import annotations
+
+DEFAULT_LIMIT_GAIN_ATOL = 1.0e-4
```
adenzler-nvidia: **I7 — `DEFAULT_LIMIT_GAIN_ATOL = 1e-4` is an absolute tolerance.** On `DEFAULT_LIMIT_KE ≈ 2500` that's 4e-8 relative — the MJCF_DEFAULT → FORCE_SPACE promotion threshold is invisible to users. A user typing `ke=2499.9999` keeps the MJCF default; `ke=2499.999` silently switches to force-space scaling. Suggest a relative tolerance (or `np.isclose` defaults) and at minimum a regression test demonstrating the threshold. Same constant is reused in the Warp kernel guard at `kernels.py:2421-2422`.

adenzler-nvidia (commented):
Review of the joint-limit `ke`/`kd` scaling fix. The mode-field design and the kernel/helper docstrings are excellent. Below are the items I'd like to see addressed before merge: a few silent fallbacks, round-trip semantics around `save_to_mjcf`, and a doc naming consistency fix. Findings flagged as Important — none are blocking on their own, but they all hide misconfiguration or surprise users.

newton/_src/solvers/mujoco/kernels.py
```diff
@@ -2347,6 +2352,99 @@ def update_geom_properties_kernel(
...
+    MuJoCo's default ``(0.02, 1.0)`` pair so runtime disablement matches a fresh
+    model compiled without ``solref_limit``. MJCF-imported raw ``solreflimit``
+    values are forwarded unchanged when present; MJCF joints that rely on the
+    implicit default keep MuJoCo's native ``(0.02, 1.0)`` until
+    ``joint_limit_ke``/``joint_limit_kd`` are changed by the user.
+    ``dof_invweight0`` is only valid after MuJoCo's ``set_const_0`` /
+    ``mj_setConst`` has run, so this kernel must be launched from
+    ``notify_model_changed`` after those calls (and once at initialisation right
+    after ``put_model``).
+    """
```
adenzler-nvidia, line 2413: **I8 — Kernel docstring lacks an `Args:` block.** AGENTS.md mandates Google-style docstrings with `Args:` sections (`name: description`) and SI units for physical quantities in public API. This kernel has 9 array parameters and the prose explains the math but never says what each parameter is. At minimum the load-bearing inputs (`joint_limit_solref_mode`, `jnt_dofadr`, `dof_invweight0`, `jnt_solimp`) and the output `jnt_solref` should be documented so reviewers can verify semantics without reading the call site. The wrapper at `solver_mujoco.py:6831-6862` has the same gap and paraphrases units differently — consider a single `Args:` block on the kernel and a `:meth:` cross-ref from the wrapper.

CHANGELOG.md
```diff
@@ -174,10 +174,14 @@
...
+- Fix `SolverMuJoCo` to preserve authored zero-valued USD `mjc:solreflimit` values as raw MuJoCo data and avoid writing limit `solref` values to unlimited joints in saved MJCF
 - Fix box support-map sign flips from quaternion rotation noise (~1e-14) producing invalid GJK/MPR contacts for face-touching boxes with non-trivial base rotations
 - Fix USD import of multi-DOF joints from MuJoCo-converted assets where multiple revolute joints between the same two bodies caused false cycle detection; merge them into D6 joints with correct DOF label mapping for MjcActuator target resolution
 - Fix USD `MjcActuator` import so position and velocity actuators populate Newton's joint target arrays and can be driven via `Control.joint_target_pos` / `Control.joint_target_vel`
 - Fix MJCF importer creating finite planes from MuJoCo visual half-sizes instead of infinite planes
+- Fix `SolverMuJoCo` ignoring `mujoco.solreflimit_mode = SOLREF_MODE_FORCE_SPACE` when a non-zero `mujoco.solreflimit` was also authored; the explicit mode is now authoritative so force-space scaling applies as requested
```
adenzler-nvidia: **I9 — CHANGELOG entry references private constant `SOLREF_MODE_FORCE_SPACE`.** That constant lives in `newton/_src/solvers/mujoco/constants.py` and is not part of the public API. CHANGELOG entries are user-facing and should describe behaviour, not internals. Suggested rewording: "Honor authored `mujoco.solreflimit_mode` even when a non-zero `mujoco.solreflimit` is also authored, so the explicit mode is authoritative."

docs/integrations/mujoco.rst
```diff
@@ -135,6 +135,45 @@ MuJoCo geometry model. They are only available through Newton's
 collision pipeline (see `Collision pipeline`_ below).
 
 
+Joint-limit stiffness and damping
+---------------------------------
+
+:attr:`~newton.Model.joint_limit_ke` and
+:attr:`~newton.Model.joint_limit_kd` are force-space gains (for example,
+``N·m/rad`` and ``N·m·s/rad`` for revolute joints). MuJoCo's negative
+``solref_limit`` convention stores stiffness and damping values, but the
```
adenzler-nvidia: **I10 — Inconsistent naming: `solref_limit` vs `solreflimit`.** This section mixes `solref_limit` (the C struct field name used in `mj_model.jnt_solref`) and `solreflimit` (the XML attribute / `model.mujoco.solreflimit` custom-attribute name) within the same paragraph. Users searching the MJCF for `solref_limit` won't find anything — the XML attribute is `solreflimit` (no underscore). Suggest settling on `solreflimit` everywhere the user-facing storage is meant, reserving `solref_limit` for the C field if it's referenced at all.

newton/_src/solvers/mujoco/solver_mujoco.py
vreutskyy: The MJCF parser side of this is now handled — `parse_vec` accepts MuJoCo's one-value `solreflimit="0.02"` shorthand (commit `b7b0d9a7`, restricted to a `solref`-shaped whitelist in this batch per I4). The export side has been narrowed further in this batch (I6): `spec.joints[mjc_jnt].solref_limit` is now only written for `SOLREF_MODE_RAW` joints — `SOLREF_MODE_FORCE_SPACE` and `SOLREF_MODE_MJCF_DEFAULT` joints intentionally don't author `solreflimit` in the saved MJCF, so the round-trip is idempotent and a re-imported model recovers the same MJCF_DEFAULT state instead of being frozen into RAW by a scaled serialisation. Behavior trade-off documented in `docs/integrations/mujoco.rst`.

newton/_src/solvers/mujoco/kernels.py
vreutskyy: Addressed in `b3025ac5`. Added `test_joint_limit_solref_scales_per_dof_on_multi_dof_joint`: a D6 joint with three angular DOFs, each given distinct `(limit_ke, limit_kd)` pairs. The test verifies that each MuJoCo joint row in `jnt_solref` is independently scaled by its DOF's `dof_invweight0` via `jnt_dofadr[mjc_jnt]`, and that the three rows are not aliased to a single DOF's gains.

The early-return branch for `mjc_jnt_to_newton_dof[..] < 0` is not separately exercised — it's a defensive guard for MuJoCo-injected internal constraints that don't have a Newton DOF mapping. Happy to add a targeted fixture if you have a reproducible way to inject one without re-implementing the mapping setup.

newton/_src/solvers/mujoco/kernels.py
vreutskyy: Addressed in `b3025ac5`. Added a one-shot validation pass in `_update_solref_from_invweight0` that scans `mujoco.solreflimit_mode == SOLREF_MODE_RAW` entries and warns on mixed signs or zero components (excluding the documented `(0, 0)` sentinel, which is preserved verbatim per `test_mjcf_authored_zero_solreflimit_is_preserved_as_native_parameter`). The validation flag is reset on every `notify_model_changed(JOINT_DOF_PROPERTIES)` so re-assigning `mujoco.solreflimit` re-validates.

newton/_src/utils/import_mjcf.py
vreutskyy: Addressed in `b3025ac5`. The MJCF importer now warns when `solref_to_stiffness_damping` rejects an authored `solreflimit` (mixed signs, etc.), making the divergence between the default-coerced `joint_limit_ke`/`kd` and the forwarded RAW solref visible to the user. The warning excludes the `solreflimit="0 0"` sentinel — that pair is intentionally preserved as a RAW-mode marker (regression-covered by `test_mjcf_authored_zero_solreflimit_is_preserved_as_native_parameter`).

I3's runtime-side validator (in `_update_solref_from_invweight0`) catches the same conditions for non-MJCF sources (USD, builder-API).

newton/_src/solvers/mujoco/solver_mujoco.py
vreutskyy: Addressed in `b3025ac5`. Two changes:

1. The `_convert_to_mjc` seeding of `spec.joints[..].solref_limit = (-joint_limit_ke, -joint_limit_kd)` for FORCE_SPACE joints has been removed. Only `SOLREF_MODE_RAW` joints seed the spec now; `SOLREF_MODE_FORCE_SPACE` / `SOLREF_MODE_MJCF_DEFAULT` joints leave the spec at MuJoCo's compile-time default, and `_update_solref_from_invweight0` rescales `jnt_solref` post-compilation as before.

2. The `save_to_mjcf` export loop only writes `spec.joints[mjc_jnt].solref_limit` for `SOLREF_MODE_RAW` joints. So the saved MJCF no longer carries scaled FORCE_SPACE solref values that would freeze a re-imported model into RAW mode.

Trade-off documented in the new note under "Joint-limit stiffness and damping" in `docs/integrations/mujoco.rst`: MJCF format cannot express "use Newton force-space scaling with these gains", only the resulting solref. Users re-loading a saved FORCE_SPACE model must reapply `joint_limit_ke`/`kd` (and the FORCE_SPACE mode) on the rebuilt model.

newton/_src/solvers/mujoco/kernels.py
vreutskyy: Addressed in `b3025ac5`. Full Google-style `Args:` block added to `update_jnt_solref_from_invweight0_kernel` covering all nine array parameters (`mjc_jnt_to_newton_dof`, `joint_limit_ke`/`kd`, `joint_limit_solref`/`_mode`, `jnt_dofadr`, `dof_invweight0`, `jnt_solimp`) plus the `jnt_solref` output, with SI units (`[N/m or N·m/rad]`, `[1/kg]`) per AGENTS.md.

The wrapper at `solver_mujoco.py:_update_solref_from_invweight0` retains its narrative docstring; cross-ref to `:meth:` added.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -4357,6 +4449,25 @@ def get_custom_attribute(name: str) -> np.ndarray | None:
...
         joint_ref = get_custom_attribute("dof_ref")
 
+        def joint_has_raw_limit_solref(dof_idx: int) -> bool:
+            if joint_solref_limit is None:
+                return False
+            if joint_solref_limit_mode is not None:
+                return int(joint_solref_limit_mode[dof_idx]) == SOLREF_MODE_RAW
+            return bool(np.any(joint_solref_limit[dof_idx] != 0.0))
+
+        def joint_uses_mjcf_default_limit_solref(dof_idx: int) -> bool:
```
adenzler-nvidia: `joint_uses_mjcf_default_limit_solref` is defined here but never called — only `joint_has_raw_limit_solref` is used (the two `solref_limit` seed sites). The 3-line comment documents logic that never runs, so a future maintainer will read it as live behavior. Either wire it in or drop the function along with its comment.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -6703,6 +6871,169 @@ def _update_geom_properties(self):
...
+                if np.any(invalid):
+                    bad = np.flatnonzero(raw_mask)[invalid]
+                    warnings.warn(
+                        f"Authored mujoco.solreflimit has invalid components at DOF indices "
+                        f"{bad.tolist()}: expected two same-sign non-zero values; MuJoCo will "
+                        "silently misbehave (divide-by-zero or disabled limit) until corrected.",
+                        stacklevel=2,
+                    )
+            # One-shot guard: model edits trigger a fresh notify which will
+            # reset this flag via ``_invalidate_contact_fast_path`` callers,
```
adenzler-nvidia: This comment is inaccurate: `_raw_solreflimit_validated` is reset only in `notify_model_changed` under `JOINT_DOF_PROPERTIES` (the `self._raw_solreflimit_validated = False` line), not "via `_invalidate_contact_fast_path` callers" — that method never touches this flag. Since `_update_solref_from_invweight0` also runs on `BODY_INERTIAL_PROPERTIES` and the other `need_const_0` triggers (where the flag stays `True`), please correct the comment to state the actual reset condition. Minor: consider initializing `_raw_solreflimit_validated` in `__init__` rather than relying on `getattr(..., False)`, so the attribute's lifecycle is explicit.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -6703,6 +6871,169 @@ def _update_geom_properties(self):
...
+            mode_np = joint_limit_solref_mode.numpy()
+            raw_np = joint_limit_solref.numpy()
+            raw_mask = mode_np == SOLREF_MODE_RAW
+            if np.any(raw_mask):
+                tc = raw_np[raw_mask, 0]
+                dr = raw_np[raw_mask, 1]
+                invalid = (tc == 0.0) | (dr == 0.0) | (np.sign(tc) != np.sign(dr))
+                if np.any(invalid):
+                    bad = np.flatnonzero(raw_mask)[invalid]
+                    warnings.warn(
```
adenzler-nvidia, line 6941: This invalid-RAW-solref warning has no test coverage, and neither do the two warnings added in `import_mjcf.py` (invalid authored `solreflimit` on import, and the `parse_vec` single-value shorthand fallback). The mixed-sign / zero-component detection here is non-trivial, and the one-shot `_raw_solreflimit_validated` dedup (warns once, re-arms after a `JOINT_DOF_PROPERTIES` notify) is exactly the kind of stateful behavior that regresses silently. Could you add assertWarns coverage for these three paths, plus a check that the one-shot guard re-arms after a JOINT_DOF_PROPERTIES notify?

adenzler-nvidia (commented):
A few maintainability notes on the joint-limit solref work — all minor, none blocking.

newton/_src/solvers/mujoco/solver_mujoco.py
vreutskyy: Dropped in `d9d0f38e`. After I6's switch to only-RAW spec seeding, `joint_uses_mjcf_default_limit_solref` had no remaining call sites. The function and its comment are gone.

newton/_src/solvers/mujoco/solver_mujoco.py
vreutskyy: Fixed in `d9d0f38e`:

- Moved `self._raw_solreflimit_validated: bool = False` to `__init__` so the attribute's lifecycle is explicit; the `getattr(self, "_raw_solreflimit_validated", False)` indirection is gone.
- Corrected the inline comment to state the actual reset condition: only `notify_model_changed(SolverNotifyFlags.JOINT_DOF_PROPERTIES)` re-arms the flag; the other `need_const_0` triggers (e.g. `BODY_INERTIAL_PROPERTIES`) leave it `True` because they cannot change the authored solreflimit values themselves.

newton/_src/solvers/mujoco/solver_mujoco.py
vreutskyy: Added in `d9d0f38e`:

- `test_invalid_raw_solreflimit_warns_in_update_solref` — builds a revolute joint, promotes it to `SOLREF_MODE_RAW` with a mixed-sign `(-2500, 1)` solref, and asserts (a) the validator warns on init, (b) a subsequent `notify_model_changed(BODY_INERTIAL_PROPERTIES)` does NOT re-warn (one-shot dedup), and (c) `notify_model_changed(JOINT_DOF_PROPERTIES)` re-arms and re-warns.

- `test_mjcf_invalid_solreflimit_emits_import_warning` — covers the MJCF importer side: `solreflimit="-1 1"` triggers the importer-time warning.

- `test_parse_vec_unknown_shorthand_warns` — two sub-cases:
  * `solreflimit="0.04"` (whitelisted) must NOT warn — silent shorthand padding.
  * `range="0.5"` (non-whitelisted multi-component) MUST warn and fall back to replicate.

Also fixed `parse_vec` to handle `default=None` (used by `actuatorfrcrange`), which surfaced during the warn-coverage work.

newton/_src/solvers/mujoco/constants.py
```diff
@@ -0,0 +1,54 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""Shared constants for Newton's MuJoCo solver integration."""
+
+from __future__ import annotations
+
+DEFAULT_LIMIT_GAIN_RTOL = 1.0e-5
+"""Relative tolerance for detecting imported MuJoCo default joint-limit gains.
+
+Used to recognise the ``shape_material_ke`` / ``shape_material_kd`` values
```
adenzler-nvidia: This docstring describes the tolerance as recognizing `shape_material_ke` / `shape_material_kd`, but `DEFAULT_LIMIT_GAIN_RTOL` is used only for joint-limit gains — every call site compares `joint_limit_ke` / `joint_limit_kd` against `DEFAULT_LIMIT_KE` / `DEFAULT_LIMIT_KD` (`kernels.py:2445-2446`, `solver_mujoco.py:6914-6915`, `:6990-6991`). Shape materials are explicitly out of scope for this PR, so this reads like a copy-paste from the broader #2009 context. Suggest swapping to `joint_limit_ke` / `joint_limit_kd` to avoid misleading a future maintainer.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -5809,6 +5924,52 @@ def get_body_name(body_idx: int) -> str:
...
+                #
+                # Note: ``save_to_mjcf`` is *not* a fully semantic
+                # round-trip for ``FORCE_SPACE`` joints — MJCF cannot
+                # express "use Newton force-space scaling with these
+                # gains", only the resulting solref. Users who need to
+                # reload force-space joint dynamics must reapply
+                # ``joint_limit_ke``/``kd`` (and the FORCE_SPACE mode) on
+                # the rebuilt model.
+                jnt_to_newton_dof = self.mjc_jnt_to_newton_dof.numpy()[0]
+                solreflimit_mode_np = joint_solref_limit_mode if "joint_solref_limit_mode" in locals() else None
```
adenzler-nvidia: `joint_solref_limit_mode` is unconditionally bound at line 4447 in this same `_convert_to_mjc` scope, so `"joint_solref_limit_mode" in locals()` is always true and the `else None` re-fetch from `self.model.mujoco` just below is unreachable. Suggest dropping either the `locals()` guard or the dead fallback so the code doesn't document a path that never runs.

adenzler-nvidia (dismissed):
Re-reviewed the current state end-to-end: the `jnt_solref` scaling kernel, its CPU mirror, the MJCF import/export paths, and the `FORCE_SPACE` / `RAW` / `MJCF_DEFAULT` mode machine. Everything I raised earlier (I1–I10 plus the three follow-ups) is addressed, the divide-by-zero guard is in place, `save_to_mjcf` only seeds `solref_limit` for `RAW` joints, and the GPU and CPU scaling paths agree. Approving.

Two optional doc/cleanup nits below — neither blocking.

newton/_src/solvers/mujoco/constants.py
vreutskyy: Fixed in `1dc03ca1`. Docstring now references `joint_limit_ke` / `joint_limit_kd` (the actual call sites in `kernels.py` and `solver_mujoco.py`). The `shape_material_ke` / `kd` wording was a leftover from #2009's broader scope and didn't belong here.

newton/_src/solvers/mujoco/solver_mujoco.py
vreutskyy: Fixed in `1dc03ca1`. The dead `else None` branch and the `"joint_solref_limit_mode" in locals()` guard are gone — the export loop now references `joint_solref_limit_mode` directly. Replaced the previous comment with a short note pointing back at the unconditional `get_custom_attribute` binding earlier in `_convert_to_mjc`.

adenzler-nvidia (approved):

PR #2614, opened 2026-04-28
Fix URDF package URI fallback resolution

han-xudong:
## Description

Fix URDF `package://` mesh fallback resolution when `resolve-robotics-uri-py` is not installed.

Previously, the fallback path logic matched package names by substring inside the URDF folder path. That could incorrectly resolve assets from unrelated directories whose names merely contained the package name. This change restricts the fallback to full path-component matches only.

## Checklist

## Test plan

```bash
uvx pre-commit run -a
uv run --extra dev -m unittest newton.tests.test_import_urdf.TestUrdfUriResolution
```

## Bug fix

**Steps to reproduce:**

1. Load a URDF from a file path in a directory whose name merely contains the target package name as a substring, for example `.../notpkg/urdf/robot.urdf`.
2. Reference a mesh using `package://pkg/meshes/link.obj`.
3. Do not install `resolve-robotics-uri-py`.
4. Place an unrelated mesh at a path that matches the substring-derived fallback, such as `.../not/pkg/meshes/link.obj`.
5. Observe that Newton incorrectly loads the unrelated mesh instead of warning that the package URI could not be resolved.

**Minimal reproduction:**

```python
import tempfile
from pathlib import Path
from unittest.mock import patch

import newton

MESH_OBJ = """v 0 0 0
v 1 0 0
v 0 1 0
f 1 2 3
"""

with tempfile.TemporaryDirectory() as td:
    base = Path(td)
    accidental = base / "not" / "pkg" / "meshes"
    accidental.mkdir(parents=True)
    (accidental / "link.obj").write_text(MESH_OBJ)

    misleading = base / "notpkg"
    (misleading / "urdf").mkdir(parents=True)
    urdf_path = misleading / "urdf" / "robot.urdf"
    urdf_path.write_text(
        '<robot name="r"><link name="base"><visual><geometry>'
        '<mesh filename="package://pkg/meshes/link.obj"/>'
        '</geometry></visual></link></robot>'
    )

    with patch("newton._src.utils.import_urdf.resolve_robotics_uri", None):
        builder.add_urdf(str(urdf_path), up_axis="Z")
        print(builder.shape_count)  # incorrectly becomes 1 without this PR
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed URDF `package://` mesh resolution fallback when the robotics URI resolver is unavailable, ensuring package matching uses full path components to avoid incorrect substring-based matches.

* **Tests**
  * Added coverage to verify `package://` fallback does not resolve using misleading path fragments and that unresolved meshes are handled as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -270,6 +270,7 @@
 - Fix `State.assign` not copying namespaced extended and custom state attributes 
 - Fix mesh-convex back-face contacts generating inverted normals that trap shapes inside meshes and cause solver divergence (NaN)
 - Fix triangle-mesh-vs-convex collisions silently dropping all contacts under non-uniform (and even large uniform) mesh scale: the BVH AABB query in `mesh_vs_convex_midphase` is now performed in unscaled mesh-local space (matching the BVH built over `mesh.points`), with the per-axis contact gap converted accordingly. Previously the query was performed in scaled mesh-local space, so any convex shape whose unscaled-space AABB lay outside the BVH bounds would receive 0 triangles and 0 contacts.
+- Fix URDF `package://` mesh fallback resolution without `resolve-robotics-uri-py` so package names only match full path components instead of unrelated directory-name substrings
 - Fix finite plane geometry 2x too large in collision, bounding sphere, and raytrace sensor
 - Fix MPR convergence failure on large and extreme-aspect-ratio mesh triangles by projecting the starting point onto the triangle nearest the convex center
 - Fix MPR/GJK reporting wrong contacts for `CONVEX_MESH` shapes whose authoring origin lies outside the hull, and tighten heightfield-vs-convex midphase to use the convex's local AABB instead of an origin-centered bounding sphere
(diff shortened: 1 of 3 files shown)


jcarius-nv:

adenzler-nvidia:
Thanks for the fix — switching to full path-component matching is the right call. I checked that it stays equivalent to the old logic on valid ROS layouts (including when the package name appears as a nested or repeated path component) while correctly rejecting the substring false-positive, and the added test covers the regression well.

Two things before this can merge:

1. **Merge the latest `main`** — the branch is currently in conflict (`CHANGELOG.md` is the culprit; the source change still applies cleanly).

2. **Re-place the CHANGELOG entry** — it currently sits next to the "Fix mesh-convex back-face contacts…" and "Fix finite plane geometry 2x…" lines, but those were released in `1.2.0`. After merging `main`, please move your entry into the current `[Unreleased]` → `Fixed` section (any position within it is fine).

han-xudong:
Rebased onto the latest `main` and moved the changelog entry into the current top-level `[Unreleased] -> Fixed` section.

I also reran:
- `uv run --extra dev -m unittest newton.tests.test_import_urdf.TestUrdfUriResolution`

Thanks for the review.

adenzler-nvidia (approved):
Thank you!

PR #2615, opened 2026-04-28
Fix MJCF visual geom explicit mass

han-xudong:
## Description

Honor explicit `mass=` on visual geoms loaded via `parse_visuals=True`.

This is a follow-up to the earlier explicit-geom-mass support: visual geoms
already contribute default-density mass when visuals are parsed, but authored
explicit mass was still skipped by the `not just_visual` guard. That left
visual-only mass and inertia at zero even though the geom itself was loaded.

## Checklist

## Test plan

```bash
uv run --extra dev -m unittest \
  newton.tests.test_import_mjcf.TestImportMjcfGeometry.test_visual_geom_density_with_parse_visuals \
  newton.tests.test_import_mjcf.TestImportMjcfGeometry.test_visual_geom_explicit_mass_with_parse_visuals

uv run --extra dev -m unittest \
  newton.tests.test_import_mjcf.TestImportMjcfGeometry.test_explicit_geom_mass \
  newton.tests.test_import_mjcf.TestImportMjcfGeometry.test_visual_geom_density_with_parse_visuals \
  newton.tests.test_import_mjcf.TestImportMjcfGeometry.test_visual_geom_explicit_mass_with_parse_visuals

uvx pre-commit run -a
```

## Bug fix

**Steps to reproduce:**

1. Build a body with one visual geom carrying `mass="5"` and one collider carrying `mass="0"`.
2. Import the MJCF with `parse_visuals=True`.
3. Finalize the model and inspect `body_mass` and `body_inertia`.
4. Observe that both remain zero without this fix.

**Minimal reproduction:**

```python
import newton
import numpy as np

mjcf = """<?xml version=\"1.0\" ?>
<mujoco>
  <worldbody>
    <body name=\"test\" pos=\"0 0 0.5\">
      <joint type=\"hinge\" axis=\"0 0 1\"/>
      <geom name=\"vis\" type=\"box\" size=\"0.1 0.1 0.1\"
            contype=\"0\" conaffinity=\"0\" group=\"2\" mass=\"5\"/>
      <geom name=\"col\" type=\"box\" size=\"0.1 0.1 0.1\"
            mass=\"0\" group=\"3\"/>
    </body>
  </worldbody>
</mujoco>"""

builder.add_mjcf(mjcf, parse_visuals=True)
model = builder.finalize()

print(model.body_mass.numpy()[0])
print(np.trace(model.body_inertia.numpy()[0]))
```

## Summary by CodeRabbit

* **Bug Fixes**
  * MJCF importer now correctly applies explicit mass values specified on visual geometry elements when visual parsing is enabled. Visual-only geoms with authored mass now contribute to body mass and inertia calculations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@
 - Fix MJCF importer creating finite planes from MuJoCo visual half-sizes instead of infinite planes
 - Fix USD import of joint limit stiffness/damping from `MjcJointAPI`: `SchemaResolverMjc` now reads the schema-correct `mjc:solreflimit` attribute instead of the generic `mjc:solref`, which was never authored on joints
 - Fix MJCF importer in `compiler.angle="degree"` mode: (1) stop multiplying joint `damping`/`stiffness` by `180/π` (MuJoCo stores these in `N·m·s/rad` and `N·m/rad` regardless of `angle`); (2) stop `deg2rad`-scaling the default `±MAXVAL` sentinel for joints without an explicit `range=`, which was turning unlimited hinges into bounded joints with `~1.75e8 rad` range
+- Fix MJCF importer ignoring explicit `mass=` on visual geoms loaded via `parse_visuals=True`; authored visual-only mass now contributes to body mass and inertia like visual-only density already does
 - Fix ViewerViser mesh popping artifacts caused by viser's automatic LOD simplification creating holes in complex geometry
 - Fix degenerate zero-area triangles in SDF marching-cubes isosurface extraction by clamping edge interpolation away from cube corners and guarding against near-zero cross products
 - Fix multi-world coordinate conversion using the wrong body center of mass for replicated worlds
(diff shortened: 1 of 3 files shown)


adenzler-nvidia:
Thanks. I remember being confused about this before. Just to clarify - what is your expectation here: if a body has both collision and visual geoms with explicit mass, are they added? Or which one is preferred?

vreutskyy (approved):
LGTM. Small, well-motivated fix that closes an inconsistency between the density and explicit-mass paths for visual geoms when `parse_visuals=True`. The `not just_visual` guard was the only thing keeping authored visual mass from contributing, and the `body_lock_inertia` check still suppresses the contribution when an explicit `<inertial>` is present, so the existing inertial-lock semantics are preserved.

Optional nit: the new test only checks `trace(I) > 0`; asserting the closed-form box inertia (`I_xx = m·(hy² + hz²)/3`) would catch silent regressions in inertia scaling, but that's a follow-up at most.

PR #2616, opened 2026-04-28
Fix collapsed URDF articulation starts

### LINKED ISSUE #2589: ModelBuilder.finalize() crashes with 3+ articulations
# G16 — `ModelBuilder.finalize()` crashes with 3+ articulations

**Severity:** blocker for multi-env simulation
**Affects:** Newton 1.1.0
**Filed from:** NVIDIA kitless-isaacsim project (ANYmal C multi-env walking)
**Related:** G15 (`add_urdf` no `env_count`), G17 (`SolverMuJoCo` hang)

## Summary

`ModelBuilder.finalize()` crashes when the builder contains 3 or more
articulations, due to an unsorted / non-monotonic `articulation_start`
array.  2 articulations work; 3+ fail.

## Expected behavior

```python
for i in range(4):
    builder.add_urdf(urdf, xform=wp.transform((i * 2.0, 0, 0.62),
                                              wp.quat_identity()),
                     floating=True)
model = builder.finalize(device="cuda:0")   # should succeed
```

## Actual behavior

`finalize()` raises inside the articulation-range validation (or produces
a corrupt `Model` whose solver invocation later segfaults / raises on
array-bounds checks), depending on exact input order.

## Reproduction (minimal, no URDF)

```python
import newton
import warp as wp

wp.init()

# Create 4 tiny 2-body articulations
for i in range(4):
    pos = (i * 2.0, 0.0, 0.5)
    b0 = builder.add_body(xform=wp.transform(pos, wp.quat_identity()))
    builder.add_shape_box(body=b0, hx=0.2, hy=0.2, hz=0.2)
    b1 = builder.add_body(
        xform=wp.transform((pos[0], pos[1], pos[2] + 0.5), wp.quat_identity())
    )
    builder.add_shape_box(body=b1, hx=0.1, hy=0.1, hz=0.3)
    builder.add_joint_revolute(
        parent=b0, child=b1,
        parent_xform=wp.transform((0, 0, 0.2), wp.quat_identity()),
        child_xform=wp.transform((0, 0, -0.3), wp.quat_identity()),
        axis=(1, 0, 0),
    )

builder.add_ground_plane()

# Fails with 3+ articulations
```

## Reproduction (ANYmal URDF)

```python
import newton, warp as wp, math
wp.init()

urdf = str(newton.utils.download_asset("anybotics_anymal_c") / "urdf" / "anymal.urdf")

for i in range(3):                     # 3 is enough to trigger
    xform = wp.transform(
        wp.vec3(i * 2.0, 0.0, 0.62),
        wp.quat_from_axis_angle(wp.vec3(0, 0, 1), math.pi * 0.5),
    )
    builder.add_urdf(urdf, xform=xform, floating=True,
                     collapse_fixed_joints=True)

builder.add_ground_plane()
model = builder.finalize(device="cuda:0")   # crashes
```

2 envs: finalize succeeds.  3 envs: finalize crashes.

## Root cause (best guess, please confirm)

When multiple articulations are added interleaved with the ground plane /
free-body shapes, `ModelBuilder.articulation_start` ends up with indices
that are not strictly increasing, or that don't match the actual joint /
body insertion order.  The finalize pass that walks articulations for
kinematic-tree construction reads past the end of one articulation into
the next, which blows the body-range invariant downstream.

Sorting `articulation_start` (and reindexing `articulation_end` /
per-articulation metadata) in the finalize pass should resolve the crash
— but we haven't traced this in Newton sources directly; would appreciate
confirmation from Newton maintainers.

## Workaround

Downstream (`kitless.loader.cloner.KitlessCloner`) sidesteps the crash by
calling `add_urdf()` per environment *and* reordering so the ground plane
is added last.  This works in practice but is brittle — it relies on the
exact insertion order being what finalize expects.

## Environment

- Newton: 1.1.0
- Warp: (as shipped with Newton 1.1.0)
- GPU: NVIDIA L40
- OS: Ubuntu 22.04
- Python: 3.10

## Notes

Blocks the idiomatic multi-env workflow
(`builder.replicate()`) for any articulated robot, which in turn blocks
the RL training path in kitless-isaacsim.

devshahofficial:
## Description

Closes #2589.

This fixes `ModelBuilder.collapse_fixed_joints()` so articulation starts keep their original order after retained joints are remapped. It also drops fully-collapsed empty articulations consistently across start, label, world, custom articulation values, and per-joint articulation ids.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_import_urdf.TestImportUrdfComposition.test_repeated_collapse_fixed_joints_preserves_articulation_order
uvx pre-commit run -a
```

Manual smoke check: ANYmal C URDF imported three times with `collapse_fixed_joints=True` now finalizes with `model.articulation_start == [0, 13, 26, 39]`.

## Bug fix

**Steps to reproduce:**

1. Import the same multi-joint URDF three times into one `ModelBuilder` with `collapse_fixed_joints=True`.
2. Inspect `builder.articulation_start` before `finalize()`.
3. Without this PR, the starts can be reordered, e.g. `[0, 26, 13]`, and `finalize()` raises `ValueError` because the starts are not monotonically increasing.

**Minimal reproduction:**

See `test_repeated_collapse_fixed_joints_preserves_articulation_order` for the minimal inline URDF regression.

## Summary by CodeRabbit

* **Bug Fixes**
  * Repeated URDF imports now preserve articulation order and stable indexing when fixed-joint collapse is enabled; articulation identifiers and related attributes are consistently remapped.
* **Tests**
  * Added regression tests covering repeated URDF imports, articulation ordering, and remapping of articulation-scoped attributes.
* **Documentation**
  * Changelog updated to document the fix.


newton/_src/sim/builder.py
```diff
@@ -5131,23 +5131,57 @@ def dfs(parent_body: int, child_body: int, incoming_xform: wp.transform, last_dy
...
-        # update articulation_start
-        for i, old_i in enumerate(self.articulation_start):
-            start_i = old_i
-            while start_i not in joint_remap:
-                start_i += 1
-                if start_i >= self.joint_count:
-                    break
-            self.articulation_start[i] = joint_remap.get(start_i, start_i)
-        # remove empty articulation starts, i.e. where the start and end are the same
-        self.articulation_start = list(set(self.articulation_start))
```
preist-nvidia, line 5146: Why not apply a minimal fix and just make sure that this list(set( is fixed to be sorted by doing sorted(set(

preist-nvidia: Ok got it - we need the rebuild to catch the all-fixed degenerate case.

eric-heiden:
Thanks @devshahofficial, your changes look good to me! Can you please address the comments from CodeRabbit? Those are actionable.

newton/tests/test_import_urdf.py
```diff
@@ -1140,6 +1140,50 @@ def test_floating_false_with_parent_body_succeeds(self):
...
+            parse_urdf(
+                robot_urdf,
+                builder,
+                floating=True,
+                collapse_fixed_joints=True,
+                up_axis="Z",
+                xform=wp.transform(wp.vec3(float(i), 0.0, 0.0), wp.quat_identity()),
+            )
+
+        self.assertEqual(builder.articulation_start, [0, 13, 26])
```
eric-heiden, line 1181: ⚪ This test checks the numeric articulation starts, but it does not verify the remapped per-articulation metadata that this change now updates.

A regression could still corrupt `builder.articulation_label`, `builder.articulation_world`, or `builder.joint_articulation` while preserving `articulation_start`, so the repeated-import case would pass even though articulation ownership is wrong. Please add assertions for those fields here. For three imported robots, `builder.joint_articulation` should be `[0] * 13 + [1] * 13 + [2] * 13`.

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@
 
 ### Fixed
 
+- Fix repeated URDF imports with fixed-joint collapse preserving articulation order.
```
eric-heiden: ⚪ The new changelog entry ends with a period, while the surrounding entries in this `### Fixed` block do not.

Please remove the trailing period so this matches the existing changelog style.

```suggestion
- Fix repeated URDF imports with fixed-joint collapse preserving articulation order
```

devshahofficial:
yep @eric-heiden on it!

newton/tests/test_import_urdf.py
devshahofficial: Thanks @eric-heiden, added those assertions here for `builder.articulation_label`, `builder.articulation_world`, and `builder.joint_articulation` with the expected `[0] * 13 + [1] * 13 + [2] * 13` ownership. I also added the same checks after `builder.finalize()` so the test covers both the builder state and the finalized model state.

newton/_src/sim/builder.py
```diff
@@ -5131,23 +5131,78 @@ def dfs(parent_body: int, child_body: int, incoming_xform: wp.transform, last_dy
...
+                return value
+            if isinstance(value, int):
+                return articulation_remap.get(value, -1) if value >= 0 else value
+            if isinstance(value, list):
+                return [remap_articulation_reference(v) for v in value]
+            if isinstance(value, tuple):
+                return tuple(remap_articulation_reference(v) for v in value)
+            return value
+
+        for custom_attr in self.get_custom_attributes_by_frequency([Model.AttributeFrequency.ARTICULATION]):
```
eric-heiden, line 5186: 🟡 This remaps `ARTICULATION`-frequency custom attributes only when `custom_attr.values` is a `dict`.

If a custom attribute ever reaches this path with list-backed values, from manual assignment or a future API change, those values keep the old articulation indices while the rest of the builder state is remapped. Please either handle the non-dict case here, or add an assertion/comment that explains why enum-frequency attributes are guaranteed to use dict storage.

newton/_src/utils/import_usd.py
```diff
@@ -3065,9 +3065,18 @@ def _get_collision_mass_information(collider_prim: Usd.Prim):
         body_merged_parent = collapse_results["body_merged_parent"]
         body_merged_transform = collapse_results["body_merged_transform"]
         body_remap = collapse_results["body_remap"]
+        articulation_remap = collapse_results["articulation_remap"]
         # remap body ids in articulation bodies
-        for art_id, bodies in articulation_bodies.items():
-            articulation_bodies[art_id] = [body_remap[b] for b in bodies if b in body_remap]
+        articulation_bodies = {
```
eric-heiden, line 3070: ⚪ These remapped `articulation_bodies` and `articulation_has_self_collision` maps are not read after this block.

That means this remap has no user-visible effect: it is too late to affect the earlier self-collision filtering, and the remapped maps are not returned as metadata. Please either remove this dead remap, or move/return the remapped data where it is actually consumed.

eric-heiden:
Closing since https://github.com/newton-physics/newton/pull/2688 includes these changes.

PR #2620, opened 2026-04-28
Sdf file caching for fast startup times

nvtw:
## Description

Adds the possibility to specify a cache directory for SDFs. The sdf format is not to be supposed to be stable across Newton versions, it's only a cache.

## Summary by CodeRabbit

* **New Features**
  * On-disk caching for cooked texture-based SDFs to reuse work across runs
  * Optional cache_dir parameter on SDF.create_from_mesh and Mesh.build_sdf
  * Optional ability to return cooked sparse payloads when constructing texture SDFs

* **Documentation**
  * Updated docs and changelog describing cache behavior, cache keys, and automatic invalidation

* **Tests**
  * Added unit, concurrency, and end-to-end tests validating cache correctness, provenance, and invalidation

* **Examples**
  * Example updated to demonstrate persistent SDF caching across executions


mzamoramora-nvidia:
Nice PR. The cache path is well scoped, the tests cover the important hit/miss and invalidation behavior, and I like that the feature stays opt-in through `cache_dir` without changing the default SDF build path.

I agree with CodeRabbit's comment about using a unique temp file before `os.replace()` for concurrent cache writers. This is probably not the common single-script path, but it seems worth fixing for robustness: CI, parallel test runners, or multi-process simulation/training jobs can have multiple workers cook the same SDF into the same shared cache directory. A unique temp file keeps those workflows from racing on `<hash>.sdf.npz.tmp.npz`.

One small issue I found locally: `_resolve_newton_version()` currently returns `"unknown"` because `from .. import __version__` resolves to  `newton._src`, not package root `newton`. This does not affect cache validity, since validity is handled by the content hash and `CACHE_FORMAT_VERSION`, but it makes  the diagnostic `__newton_version__` metadata less useful when debugging cache files. Locally I see `newton.__version__ == "1.2.0.dev0"`, while  `_resolve_newton_version()` returns `"unknown"`.

  Validated locally:

  ```bash
  uv run --extra dev -m newton.tests -k test_sdf_disk_cache
  uv run --extra dev -m newton.tests -k test_sdf_texture

mzamoramora-nvidia (approved):
Thanks for the updates. I re-tested the latest PR head (`72c525f0`) locally, and the two issues from my previous review look addressed:

- Concurrent cache writes now use per-writer temp files.
- `_resolve_newton_version()` now resolves the actual Newton version.

Validated locally:

```bash
uv run --extra dev -m newton.tests -k test_sdf_disk_cache
uv run --extra dev -m newton.tests -k test_sdf_texture
```

Both passed.

Non-blocking follow-ups to consider:

1. Document the USD/imported-asset workflow. `add_usd()` creates mesh objects, but it does not take a `cache_dir` or build cached mesh SDFs automatically. Users who want cached SDFs for imported mesh collision assets need to build them after import and before `finalize()`:

```python
result = builder.add_usd("asset.usda")
sdf_cache_dir = "./sdf_cache"

for shape_idx in result["path_shape_map"].values():
    if builder.shape_type[shape_idx] == newton.GeoType.MESH:
        mesh = builder.shape_geo[shape_idx]
        if mesh.sdf is None:
            mesh.build_sdf(max_resolution=64, cache_dir=sdf_cache_dir)

model = builder.finalize()
```

On later runs, `build_sdf()` should load the cooked SDF data from disk when the mesh and build parameters match, while still recreating the runtime GPU textures for the current process.

2. Consider a future disk-cache path for builder-generated primitive hydroelastic SDFs. Those appear to use only the builder's in-memory SDF cache during one `finalize()` call today.

PR #2623, opened 2026-04-28
Attempt to improve mpr stability with large size differences

nvtw:
## Description

Draft MR. Mainly here to check for performance regressions through the CI

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved false-penetration cases in triangle vs. convex-hull collision initialization, improving contact detection reliability.

* **New Features**
  * Added precise closest-point surface projection for convex primitives and meshes, improving surface queries and collision refinement for varied shapes.


devshahofficial:
@nvtw I think the CodeRabbit point is real: the convex-mesh face projection looks like it is mixing reference points. Querying B from the contact center, e.g. using `center_a - position_b` in B's local frame, should keep `candidate_btoa` and `current_btoa` comparable. If useful, I can help with a small patch plus a regression for the separated triangle-vs-convex case.

Issue #1260, opened 2025-12-17
Investigate differences in properties of the ground plane


Kenny-Vilella:
It looks to me that we do not import the ground plane from mjcf.
This could potentially lead to different contact behavior due to diverging properties such as geom_friction, geom_condim, etc.

Moreover, I did not find an easy way to modify properties of the ground plane within Newton, which could be a limiting factor on some advanced cases.

adenzler-nvidia:
Planes should be imported from MJCF according to the code, can you clarify what the issue is here?

Kenny-Vilella:
This is an old issue that seems to have been solved.
Closing the ticket.

PR #2631, opened 2026-04-29
Honor WARP_CACHE_ROOT in unittest runner main-process clear

adenzler-nvidia:
## Description

The parallel test runner (`newton/tests/thirdparty/unittest_parallel.py`)
clears the Warp kernel cache on every invocation unless `--no-cache-clear`
is passed.

`WARP_CACHE_ROOT` was already honored inside the per-worker setup
(`initialize_test_process` at line 583-598), but the main-process clear at
line 182 ran *before* the env var was applied — so the clear always hit
Warp's default cache directory (`~/.cache/warp/<version>/` on Linux).

Because that default cache is keyed only on the Warp version, two
worktrees pinned to the same Warp version share one cache, and running
tests in either worktree silently wipes the other's cache. The next
example run elsewhere then has to recompile every kernel from scratch.

This PR pulls the `WARP_CACHE_ROOT` resolution up to the main process so
the clear is scoped to the same per-checkout root that workers already
use. Default behavior (no env var set) is unchanged.

The log message also now prints the resolved cache path so operators can
see what got cleared.

## Checklist

- [x] `CHANGELOG.md` has been updated (if user-facing change) — internal
      test-runner fix, not user-facing.

## Test plan

```bash
# WARP_CACHE_ROOT now scopes the clear instead of nuking the default cache
WARP_CACHE_ROOT=/tmp/wcr-test uv run --extra dev -m newton.tests \
    -k test_viewer_world_offsets
# → "Cleared Warp kernel cache: /tmp/wcr-test/<version>"

# Default path unchanged when no env var is set
uv run --extra dev -m newton.tests -k test_viewer_world_offsets
# → "Cleared Warp kernel cache: /home/<user>/.cache/warp/<version>"
```

Both paths verified locally; 12 tests in `test_viewer_world_offsets` pass
in each configuration.

## New feature / API change

No API change. This is a developer-facing fix to the test-runner cache
scoping; no exported symbol or simulation behavior is affected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Cache handling now respects WARP_CACHE_ROOT, avoids cross-worker cache conflicts, and logs the resolved cache directory.
  * Per-worker and shared-cache directory selection behavior made consistent when WARP_CACHE_ROOT is set.

* **Behavior**
  * Importing the package no longer eagerly initializes the Warp runtime.

* **Tests**
  * Added a unit test confirming lazy initialization and adjusted cache-related test setup.

* **Refactor**
  * Standardized module-level mathematical constants and trig computations for consistency.


newton/tests/thirdparty/unittest_parallel.py
```diff
@@ -176,11 +176,17 @@ def main(argv=None):
 
     import warp as wp  # noqa: PLC0415 NVIDIA Modification
 
-    # Clear the Warp cache (NVIDIA Modification)
+    # Clear the Warp cache (NVIDIA Modification).  Honor WARP_CACHE_ROOT
```
shi-eric: Nitpick: The "NVIDIA modification" thing can probably be dropped here as Newton is not an NVIDIA product, the file was a carry-over from Warp

newton/tests/thirdparty/unittest_parallel.py
```diff
@@ -176,11 +176,17 @@ def main(argv=None):
...
+    # before the clear so concurrent worktrees do not nuke each other's
```
shi-eric: Nitpick: "nuke" is too colloquial for me for code comments

shi-eric:
Thanks for tracking down the worktree-clobbering. I think this MR replicates a pattern that broke in Warp 1.13 ([GH-1260](https://github.com/NVIDIA/warp/issues/1260), commit `328a48cc8`), and the right fix is broader than the main-process clear.

**What changed in Warp 1.13.** `init_kernel_cache(path)` now appends `warp.config.version` to the path you give it, including paths set via `wp.config.kernel_cache_dir`, `WARP_CACHE_PATH`, or an explicit `path=` argument. Previously the path was used as-is. The goal was symmetry with the default cache, so `WARP_CACHE_PATH` users get per-version isolation automatically and don't need to bump the path on upgrade.

**Consequence in Newton.** Anywhere we pre-join `wp.config.version` ourselves before handing the path to Warp, the version now gets doubled. Three sites:

1. This MR at \`unittest_parallel.py:182-184\`. We set \`kernel_cache_dir = <ROOT>/<VERSION>\`, then \`wp.clear_kernel_cache()\` triggers init which resolves to \`<ROOT>/<VERSION>/<VERSION>\` and clears that. The \`os.makedirs(<ROOT>/<VERSION>)\` creates a directory we never use.
2. Worker setup (pre-existing) at \`unittest_parallel.py:590, 604\`. Same pre-join, same double on init. Both sides happen to agree, which is why the MR's clear targets the right path.
3. \`test_examples.py:114-118\`. We pass the worker's resolved \`wp.config.kernel_cache_dir\` as \`WARP_CACHE_PATH\` to the example subprocess. Post-1.13 the subprocess appends another version, lands one level below the worker, and recompiles every kernel from scratch. The intended cache-sharing is silently broken.

**Suggested fix.** Drop the manual version join and let \`init_kernel_cache\` add it once.

\`\`\`python
# unittest_parallel.py main process
if "WARP_CACHE_ROOT" in os.environ:
    wp.config.kernel_cache_dir = os.environ["WARP_CACHE_ROOT"]

# unittest_parallel.py shared-cache worker (line 604)
wp.config.kernel_cache_dir = os.getenv("WARP_CACHE_ROOT")

# test_examples.py
if warp_cache_path is not None:
    env_vars["WARP_CACHE_PATH"] = os.path.dirname(warp_cache_path)
\`\`\`

Per-worker branch at line 590 needs more thought since Warp's append doesn't naturally produce \`<ROOT>/<VERSION>-<worker>\`. Easiest is to leave the worker suffix logic alone and stop pre-joining \`<VERSION>\`.

Warp upstream \`unittest_parallel.py:688, 700\` has the same bug. I can file that separately.

Happy either way on bundling vs splitting, but landing this MR alone risks cementing the doubled path as load-bearing across files.

shi-eric:
FYI, a related fix landed upstream in Warp for its own parallel test runner: https://github.com/NVIDIA/warp/commit/694608a059e1f87e351ed7c5d54a89e54bc22b3c. Sharing for context; doesn't affect this PR since Newton uses its own unittest runner.

adenzler-nvidia:
Right call on test_examples.py — that's where the doubling actually mattered. The subprocess starts fresh, so init does run there and the worker's already-versioned `WARP_CACHE_PATH` got the version appended again. Fixed via `os.path.dirname` so the subprocess's append lands on the worker's directory.

For the runner main/worker, the doubling didn't fire on this branch because newton's imports were forcing `wp.init()` before `main()` ran: `multicontact.py:31` had `wp.static(wp.sin(...))`, which goes through Warp's builtin resolver and unconditionally calls `init()` to make sure the runtime is alive. Once `Runtime` existed, later `wp.config.kernel_cache_dir = ...` was a direct override and `clear_kernel_cache` never re-resolved through `init_kernel_cache`.

Brittle guarantee, so I split it:

- 31597dc4 makes `import newton` lazy — `math.sin`/`math.cos` for the `wp.static`-folded constants in `multicontact.py` and `collision_core.py`, plus a regression test that asserts `warp._src.context.runtime is None` after `import newton`.
- e0206607 mirrors your suggested runner pattern: hand `WARP_CACHE_ROOT` to Warp as a base path, rename the per-worker leaf to `worker-NNN`, pass `os.path.dirname(kernel_cache_dir)` as `WARP_CACHE_PATH` from `test_examples`. Nitpicks addressed in the same commit.

Bundled rather than split because the runner change isn't safe without the lazy-import change first.

newton/tests/thirdparty/unittest_parallel.py
adenzler-nvidia: Dropped in e0206607.

newton/tests/thirdparty/unittest_parallel.py
adenzler-nvidia: Changed to "wipe" in e0206607.

shi-eric (approved):

PR #2632, opened 2026-04-29
Small fixes for the CollisionPipeline

nvtw:
## Description

Fix non-determinism when sticky mode is used in contact matching and allow to user control more collision buffer sizes (relevant for very big scenes)

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added shape_pairs_max to CollisionPipeline to cap broad‑phase candidate sizing, with validation for invalid values and runtime overflow warnings.

* **Bug Fixes**
  * Deterministic sticky contact matching: tie‑break behavior and narrow‑phase buffer sizing corrected to ensure stable, memory‑bounded collision results.

* **Tests**
  * Added a long deterministic regression test verifying identical sticky contact matching across parallel pipelines.


newton/_src/geometry/contact_match.py
```diff
@@ -157,14 +157,26 @@ def _float_flip(f: float) -> wp.uint32:
...
+        tie-break (smallest wins).  Using the sort key (rather than the
+        unsorted thread id) keeps the resolution invariant under
+        non-deterministic narrow-phase slot assignment: two new contacts
+        racing for the same prev contact get the same packed claim
+        regardless of which unsorted slot the narrow phase happened to
+        hand them this run.
+
+    Within a single shape pair the upper 40 bits of every contact's sort
+    key are identical, so the low 32 bits hold the (shape_b LSBs +
+    sort_sub_key) which uniquely identifies each contact in the pair as
```
adenzler-nvidia, line 184: The fix's correctness relies on `sort_sub_key` being unique within a shape pair, which is true for the multi-contact and mesh/SDF paths I checked. But what about the reduced-contact path in `contact_reduction_global.py`, where `sort_sub_key = contact_fingerprints[contact_id]`? Two reduced contacts in the same shape pair can in principle land in different reduction slots and still get the same fingerprint (the dedup inside `export_reduced_contacts_kernel` is on `contact_id`, not fingerprint). Have you ruled that out, or is it just considered geometrically improbable enough that the previous `tid` scheme was overkill?

Related: when `sort_sub_key` does collide, both the deterministic radix sort and this new tiebreak degrade together (the sort can't disambiguate either), so the new scheme isn't *worse* than what already exists upstream — but the docstring "the same invariant the deterministic sort relies on" reads like a guarantee, when really it's "if sort breaks, this also breaks." Worth softening?

newton/_src/sim/collide.py
```diff
@@ -417,6 +417,29 @@ def _compute_per_world_shape_pairs_max(model: Model) -> int:
...
+    use ``None`` instead -- and values larger than the natural bound
+    are silently clamped down (allocating beyond the bound is just
+    burning memory, never required).
+    """
+    natural = _compute_per_world_shape_pairs_max(model)
+    if override is None:
+        return natural
+    if override <= 0:
+        raise ValueError(f"shape_pairs_max must be a positive integer or None, got {override}")
+    if override > natural:
```
adenzler-nvidia: The silent clamp here surprised me. If a user passes `shape_pairs_max=10_000_000` thinking they're getting headroom and the natural bound is 1M, they silently get 1M with no signal. The docstring justifies it with "allocating beyond the bound is just burning memory, never required" — agreed in absolute terms, but that's also true of any oversized capacity hint, and we don't normally clamp those silently.

Two alternatives that preserve the user's stated intent:
- Clamp + log/`warnings.warn` once so the user knows their value was ignored.
- Just accept it — they explicitly asked for that capacity.

What's the case for clamping silently rather than one of these?

newton/_src/sim/collide.py
```diff
@@ -514,6 +538,18 @@ def __init__(
                 "nxn"/"sap" modes, ignored.
             sdf_hydroelastic_config: Configuration for
                 hydroelastic collision handling. Defaults to None.
+            shape_pairs_max: Override for the broad-phase candidate-pair
+                buffer capacity used by the ``"nxn"`` and ``"sap"`` modes.
+                Defaults to the worst-case ``N*(N-1)/2`` per-world bound,
+                which is correct for ``"nxn"`` but typically 10-100x
```
adenzler-nvidia: Minor wording question: NXN also goes through AABB overlap, group filtering, and the excluded-pairs filter inside `BroadPhaseAllPairs` before writing, so it typically emits a tiny fraction of `N*(N-1)/2` too — same as SAP, just for slightly different reasons. The phrasing here implies the per-world bound is tight for NXN, which I don't think is generally the case (only for "every shape can collide with every other" scenarios). Would something like "an upper bound that is rarely hit by either NXN or SAP in practice" be more accurate?

adenzler-nvidia (commented):
Thanks for picking this up. The sticky-determinism fix is well-targeted and the regression test (two parallel pipelines, sticky on, identical state) is the right shape. The `shape_pairs_max` knob addresses a real multi-GB allocation issue on large SAP scenes.

A few questions inline — none blocking.

newton/_src/geometry/contact_match.py
nvtw: Done

newton/_src/sim/collide.py

newton/_src/sim/collide.py

adenzler-nvidia (approved):

PR #2633, opened 2026-04-29
Remove CONNECT smoke simulations

adenzler-nvidia:
## Description

Remove the explicit simulation loops from the CONNECT equality-constraint matrix test. The 30-case matrix still checks the generated MuJoCo inputs and runtime update paths directly; the removed work was mostly re-testing MuJoCo CONNECT solver convergence.

Current main CI timing for `test_equality_connect_constraint_with_sim_step.py` on 2026-04-29:

| Job | Current file span | Dominant CONNECT test |
| --- | ---: | ---: |
| `ubuntu-latest` | 287.9s | MuJoCo-Warp: 286.0s |
| `macos-latest` | 104.7s | MuJoCo-Warp: 104.7s |
| AWS GPU | 398.8s | MuJoCo-Warp: 398.6s |

Local measurements on Threadripper PRO 5975WX + RTX PRO 6000 Blackwell:

| Variant | CPU CONNECT | GPU CONNECT | Full file |
| --- | ---: | ---: | ---: |
| Original | 17.972s | 58.350s | - |
| Final | 1.158s | 1.581s | 4.814s |
| Speedup | 15.5x | 36.9x | - |

Projected CI impact, applying the local GPU CONNECT speedup to the current dominant CI test:

| Job | Projected CONNECT time | Projected file span | Estimated savings |
| --- | ---: | ---: | ---: |
| `ubuntu-latest` | ~7.8s | ~9.7s | ~278s |
| `macos-latest` | ~2.8s | ~2.8s | ~102s |
| AWS GPU | ~10.8s | ~11.0s | ~388s |

Exact CI savings will depend on runner speed and cache state, but the current 100-400s tail should disappear.

## Checklist

- [x] `CHANGELOG.md` has been updated (if user-facing change) - N/A, test-only change

## Test plan

```bash
uv run ruff format --check newton/tests/test_equality_connect_constraint_with_sim_step.py
uv run ruff check newton/tests/test_equality_connect_constraint_with_sim_step.py
git diff --check
uv run --extra dev -m newton.tests -p test_equality_connect_constraint_with_sim_step.py --no-cache-clear --serial-fallback -q
```

Fresh isolated-worktree result:

```text
Ran 6 tests in 42.614s
OK
elapsed 48.13
```

## Summary by CodeRabbit

* **Tests**
  * Simplified constraint handling tests to focus on core data-path behavior.
  * Updated test infrastructure for anchor assignment validation.
  * Removed simulation-driven residual convergence checks from the test suite.

### DIFF
--- a/newton/tests/test_equality_connect_constraint_with_sim_step.py
+++ b/newton/tests/test_equality_connect_constraint_with_sim_step.py
@@ -24,36 +24,6 @@ class Sim:
         self.control = control
 
 
-def connect_residual(body_poses, connect_body_indices, leafbody1_anchor, leafbody2_anchor):
-    """Compute the world-space residual of a CONNECT constraint.
-
-    Transforms anchor points on leafbody1 and leafbody2 (in their respective
-    body frames) to world space using the body poses, then returns the
-    distance between them.
-
-    Args:
-        body_poses: Array of body transforms (from ``state.body_q.numpy()``).
-        connect_body_indices: ``[leafbody1_index, leafbody2_index]`` into
-            ``body_poses``.
-        leafbody1_anchor: Anchor on leafbody1 in leafbody1's local frame
-            (``wp.vec3``).
-        leafbody2_anchor: Anchor on leafbody2 in leafbody2's local frame
-            (``wp.vec3``).
-
-    Returns:
-        Euclidean distance between the two world-space anchor points.
-    """
-    leafbody1 = connect_body_indices[0]
-    leafbody2 = connect_body_indices[1]
-    bq1 = body_poses[leafbody1]
-    bq2 = body_poses[leafbody2]
-    T1 = wp.transform(wp.vec3(bq1[0], bq1[1], bq1[2]), wp.quat(bq1[3], bq1[4], bq1[5], bq1[6]))
-    T2 = wp.transform(wp.vec3(bq2[0], bq2[1], bq2[2]), wp.quat(bq2[3], bq2[4], bq2[5], bq2[6]))
-    P1 = wp.transform_get_translation(T1) + wp.quat_rotate(wp.transform_get_rotation(T1), leafbody1_anchor)
-    P2 = wp.transform_get_translation(T2) + wp.quat_rotate(wp.transform_get_rotation(T2), leafbody2_anchor)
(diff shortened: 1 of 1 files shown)


gyeomannvidia:
Can we not keep the test and introduce the system we had on physx to distinguish between nightly testing and pre-merge testing? PhysX marks some tests as slow and these are only run by the nightly test. 

Running the full sim found quite a few bugs in Newton.

adenzler-nvidia:
no, we can't. I'll be the one fixing everyones tests then. We should avoid moving things to nightly as long as possible.

adenzler-nvidia:
I realize you found a lot of bugs running actual mujoco sims. But then you added regression tests, which cover the actual fixes, so running the sim was purely a debugging aid in that sense.

gyeomannvidia:
I don't follow how you would end up fixing everyone's test. We know who wrote the test so we can just assign it to them to investigate.

The goal of a test is to stop breakages in the future. We can never test enough.

I'm happy to investigate if we can make these tests run faster.

adenzler-nvidia:
I'm speaking from the experience of the last few months. We should test as much as we can, but there's also no need to double-cover everything and run long simulations "just to be safe".

feel free to propose an alternate solution.

gyeomannvidia:
I would argue this is not double cover. We typically test that a value ends up in mjw_model but how do we know that is everything that we need to do?  How do we know we are not overwriting something else that should not be overwritten? Ideally, the test would be extended to cover other solvers, which will obviously make the tests even longer to run.

I think we should first investigate if we can make the tests faster.

If we can't do that then I would recommend having a nightly build that includes tests marked as slow.

If we can't do that then we should look at removing the test.  I would recommend another reviewer at that point.

adenzler-nvidia:
Because mjw_model is the interface to mujoco, and that's where newton ends? Yes we don't know whether someone else overwrites it, but that's not the point of a unittest. The point here is to test that your function is writing the correct values into the model. If someone else would overwrite it, that part should have their own unittest checking things.

gyeomannvidia:
I believe that we have options that we can explore before removing these tests.

Having said that, I don't want to stand in anyone's way so I recommend finding another reviewer.

PR #2634, opened 2026-04-29
Add Newton's-cradle loading splash to ViewerGL

adenzler-nvidia:
## Description

Warp kernel compilation can block the main thread for several seconds when an example starts, leaving the viewer window unresponsive with no visible feedback. This PR shows a stylized Newton's-cradle splash with a "Loading..." sub-label during that window so the user has something to look at instead of a frozen black frame.

The splash is implemented on `ViewerGL` only; `newton.examples.init` raises it for visible GL viewers (skipping `--headless` and non-GL backends), and `newton.examples.run` lowers it before the main loop via a `hasattr` guard that matches the existing `register_ui_callback` pattern at `_ExampleBrowser.__init__:195`.

The splash is a static one-frame snapshot, not an animation: a true animated splash would need example construction on a worker thread, which is left for a follow-up. Sizes scale with the current ImGui font size so the splash stays legible across DPI settings.

## Checklist

- [x] New or existing tests cover these changes (20 cases in `newton/tests/test_viewer_loading_splash.py`)

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_viewer_loading_splash
# Ran 20 tests in 5.637s -- OK
```

Test surface:
- ViewerGL state machine (show/hide/replace/idempotent).
- API-shape pin: `ViewerGL.show_loading_splash` / `hide_loading_splash` exist by name — guards against silent rename-blindness in the `hasattr` call site.
- Direct draw-routine coverage with a recording fake ImGui: short-circuit on inactive/no-UI, draw counts (1 dim rect + 1 bar + 5 strings + 5 balls + optional text), required window flags, full-viewport dim rect including non-zero `viewport.pos` (multi-monitor), leftmost-ball-lifted geometry, font-size scaling, collapsed-window `begin()=(False, …)` path.
- Integration: `init` shows splash before the three frame pumps, `--headless` skips both, USD/Rerun/Null/Viser skip both.
- `run` ordering: hide precedes `is_running` (no-loop case) and precedes `step`/`render` (with-loop case via a one-shot stub).

Manual smoke-test: ran `python -m newton.examples basic_pendulum`; cradle visible during kernel compilation, replaced by the simulation once `Example.__init__` returns.

## New feature / API change

Adds two public methods on `ViewerGL`:

- `ViewerGL.show_loading_splash(text: str | None = None) -> None`
- `ViewerGL.hide_loading_splash() -> None`

Not added to `ViewerBase`: only `ViewerGL` has a UI surface where a splash makes sense. The other viewers (`ViewerNull`, `ViewerUSD`, `ViewerRerun`, `ViewerViser`) don't need the API and a no-op base method would have been dead surface. The `hasattr` guard at the one cross-viewer call site in `newton.examples.run` matches the existing project convention for capability-probing on heterogeneous viewer types.

## Summary by CodeRabbit

* **New Features**
  * GL viewer now has a loading splash overlay with a stylized Newton’s-cradle animation; can be shown/hidden and display optional text. It appears automatically during example initialization and is used around example switching and reset flows.

* **Documentation**
  * CHANGELOG updated to document the loading splash feature.

* **Tests**
  * New tests cover splash show/hide behavior and example init/run lifecycle interactions.

### DIFF
--- a/newton/tests/test_viewer_loading_splash.py
+++ b/newton/tests/test_viewer_loading_splash.py
@@ -1,0 +1,479 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
+
+import unittest
+from types import SimpleNamespace
+from unittest import mock
+
+import newton.examples
+from newton._src.viewer.viewer_gl import ViewerGL
+
+
+class TestViewerGLLoadingSplashAPIShape(unittest.TestCase):
+    """Pin ``ViewerGL``'s splash methods by name.
+
+    ``newton.examples.run`` probes for the method via
+    ``hasattr(viewer, "hide_loading_splash")``.  If the method is ever
+    renamed on ``ViewerGL`` and the call site is not updated together,
+    ``hasattr`` would silently flip to ``False`` and the splash would
+    linger across the entire example run with no diagnostic — none of
+    the integration tests would notice because they use
+    ``_RecordingViewer`` (which always defines the method by name).
+    """
+
+    def test_show_loading_splash_exists(self):
+        self.assertTrue(callable(getattr(ViewerGL, "show_loading_splash", None)))
+
+    def test_hide_loading_splash_exists(self):
+        self.assertTrue(callable(getattr(ViewerGL, "hide_loading_splash", None)))
+
+
+class TestViewerGLLoadingSplashState(unittest.TestCase):
+    """``show_loading_splash`` / ``hide_loading_splash`` mutate ViewerGL state correctly.
+
+    Constructed via ``__new__`` so the test does not require a GL context;
+    only the state-machine behavior is exercised.
+    """
+
+    def _make_viewer(self):
+        # Deliberately bypass ``ViewerGL.__init__`` (which would open a
+        # GL window) and hand-initialize only the state the splash API
+        # touches.  Do not call ``__init__`` here.
+        viewer = ViewerGL.__new__(ViewerGL)
+        viewer._loading_splash_active = False
+        viewer._loading_splash_text = None
+        return viewer
+
+    def test_show_sets_active_and_text(self):
+        viewer = self._make_viewer()
+        viewer.show_loading_splash("Loading basic_pendulum...")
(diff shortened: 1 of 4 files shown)


newton/tests/test_viewer_loading_splash.py
```diff
@@ -0,0 +1,479 @@
...
+
+import unittest
+from types import SimpleNamespace
+from unittest import mock
+
+import newton.examples
+from newton._src.viewer.viewer_gl import ViewerGL
+
+
+class TestViewerGLLoadingSplashAPIShape(unittest.TestCase):
```
eric-heiden: ⚪ Thanks for adding coverage for the splash lifecycle. This file feels heavier than the feature needs, though: most of the 479 lines validate fake ImGui draw calls and mocked geometry rather than behavior that is likely to regress in practice. Consider trimming this to the lifecycle checks around `newton.examples.init()` showing/pumping the splash and `newton.examples.run()` hiding it, plus a small direct state test for `show_loading_splash()` / `hide_loading_splash()`.

eric-heiden:
Thanks, this is great! Would it be possible to show this splash screen also while switching examples in the GUI from the example selector?

newton/tests/test_viewer_loading_splash.py
adenzler-nvidia: Trimmed in 06ec7e17: dropped the FakeImGui mock + draw-routine assertions and the ordering minutiae. Kept the direct `show_loading_splash` / `hide_loading_splash` state tests and a minimal `init()` / `run()` lifecycle check (3 cases). Visual behavior is now covered by manual smoke-testing on real GL rather than by mocked ImGui counts. 479 -> 117 lines, 20 -> 5 tests.

adenzler-nvidia:
Done in 7055ea78 — `_ExampleBrowser.switch()` and `.reset()` now raise the splash before the blocking `importlib` + `Example` construction (which can trigger several seconds of Warp kernel recompilation) and lower it afterwards. Switch uses `Loading <example_name>...`, reset uses `Resetting...`. Both call sites are `hasattr`-guarded so non-GL viewers keep working.

eric-heiden (approved):
Thanks, looks good!

PR #2646, opened 2026-04-29
Add changed test duration CI check

adenzler-nvidia:
## Description

Add a PR CI guard that reports slow newly changed tests without running an additional test suite.

The existing PR unit-test job now uploads its generated `rspec.xml` files as per-OS artifacts. A dependent `Check changed test durations` job downloads those artifacts, filters JUnit timing records to tests affected by the PR diff, reports tests at or above 10 seconds, and fails the check at or above 30 seconds.

Smoke validation was done on this branch with a temporary commit, `Add slow test duration smoke case`, that added an intentionally slow passing test. The new duration job failed as expected and reported `TestChangedTestDurationCheck::test_ci_duration_smoke_slow_test` at roughly 31 seconds on all four OS artifacts. That temporary commit has been reverted.

## Checklist

## Test plan

```bash
uv run ruff check scripts/ci/check_test_durations.py newton/tests/test_check_test_durations.py
uv run ruff format --check scripts/ci/check_test_durations.py newton/tests/test_check_test_durations.py
uv run --extra dev -m newton.tests -k test_check_test_durations
git diff --check -- .github/workflows/ci.yml .github/workflows/pr.yml scripts/ci/check_test_durations.py newton/tests/test_check_test_durations.py
```

Smoke validation while the temporary slow-test commit was present:

```bash
uv run --extra dev -m newton.tests -k test_check_test_durations --junit-report-xml rspec.xml
uv run python scripts/ci/check_test_durations.py --junit-root . --base-ref upstream/main --head-ref HEAD --warn-seconds 10 --fail-seconds 30
```

The second smoke command failed as expected, matching the GitHub Actions smoke result.

## New feature / API change

Not a public API change.

## Summary by CodeRabbit

* **New Features**
  * PR-only check that compares changed tests against recent test run results and flags slow or warning-level tests.

* **Chores**
  * Uploads per-run test result artifacts to CI for cross-job analysis.
  * Writes a summarized report into the PR workflow summary and fails the job for blocking slow tests.

* **Tests**
  * Added unit coverage for the changed-test duration check and parsing behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2646?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


.github/workflows/pr.yml
```diff
@@ -48,6 +48,42 @@ jobs:
...
     secrets: inherit
 
+  check-changed-test-durations:
+    name: Check changed test durations
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    needs: run-newton-tests
+    permissions:
+      contents: read
+      pull-requests: read
```
Kenny-Vilella: Why do you need this permission?

.github/workflows/pr.yml
```diff
@@ -48,6 +48,42 @@ jobs:
...
+        uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131  # v7.0.0
+        with:
+          pattern: test-results-*
+          path: test-results
+      - name: Check changed test durations
+        env:
+          BASE_REF: ${{ github.event.pull_request.base.sha }}
+          HEAD_REF: ${{ github.sha }}
+        run: |
+          python scripts/ci/check_test_durations.py \
```
Kenny-Vilella: You do not set the python version so you get whatever the runner gives you.
Will it possibly cause issue?

Kenny-Vilella (commented):
I just review the CI change.
Before I review the script itself, could you please make a second round and try to simplify the script?
This seems to be over-engineered, and why does the main function is in the middle?

Also does your script handles renaming? Of files, class or test. Should it handle it?

adenzler-nvidia:
Updated.

I simplified the duration check substantially:
- removed line-level diff parsing and dynamic-registration parsing
- now checks whole changed test classes using Newton's `Test*` class convention
- removed the custom CLI; the CI job now runs `python scripts/ci/check_test_durations.py`
- uses fixed CI-local thresholds and `BASE_REF` from the workflow environment
- removed the extra `pull-requests: read` permission
- added pinned `actions/setup-python` using `.python-version`

For renames, the job uses `git diff --name-only --diff-filter=AMR` and parses the current changed test file, so renamed test files/classes are handled by checking the tests declared in the resulting file/class.

I also validated the duration gate with a temporary slow passing test: normal unit tests passed, and the changed-test duration job failed on that slow changed test. The probe has now been removed from the branch.

shi-eric:
The duration check appears to key changed tests by bare class name, but class names are not unique in the test suite.

For example, `TestControlForce` exists in both `test_control_force.py` and `test_up_axis.py`, and `TestOverrideRootXform` exists in both `test_import_usd.py` and `test_import_mjcf.py`. Since the JUnit XML also records only the bare class name, a PR changing one file can cause timings from the unrelated duplicate class to be evaluated and potentially fail the gate.

Could this use a module/class or path/class key instead?

shi-eric:
A viable compromise might be to make this advisory and attach it only to the AWS GPU PR workflow, `.github/workflows/pr_target_aws_gpu_tests.yml` (`Pull Request - AWS GPU Tests`).

I’m concerned the standard PR workflow may give noisy timing data from the parallel GitHub-hosted matrix. If the goal is to flag slow changed tests without blocking PRs, it may be better to run the duration evaluation from the AWS GPU test result only, then post/update a PR comment from the existing `pull_request_target` workflow.

One possible shape:
- Remove the hard-failing duration gate from `.github/workflows/pr.yml`.
- Run the duration evaluation in `.github/workflows/aws_gpu_tests.yml` after it has checked out the PR head.
- Have the duration script run in advisory mode and emit a rendered Markdown report artifact.
- Upload the AWS GPU timing report artifact from `.github/workflows/aws_gpu_tests.yml`.
- Have `.github/workflows/pr_target_aws_gpu_tests.yml` post or update a PR comment from that artifact.
- Keep the comment-writing step on trusted workflow code; treat the report artifact as data, not executable code.

That keeps the signal visible without turning parallel timing variability into a required gate.

adenzler-nvidia:
My goal was to keep CPU time in check as well as this is what's currently the bottleneck. But I'm not sure if I should even go ahead with this and if we care enough.

PR #2220, opened 2026-03-25
Add site-targeted actuator support to MuJoCo solver

oursland:
`resolve_actuator_target` now checks `builder.shape_label` for shapes with `ShapeFlags.SITE` so that `MjcActuator` prims targeting sites are correctly resolved as `TrnType.SITE` during USD parsing.

Pass `site_mapping` into `_init_actuators`, build a `site_label_to_name` reverse lookup, and handle `TrnType.SITE` actuators by matching the actuator target label to the exported MuJoCo site name.

## Description

Add actuators that can be attached to `SITE`s.

## Checklist

- [x] New or existing tests cover these changes
- [x] The documentation is up to date with these changes
- [x] `CHANGELOG.md` has been updated (if user-facing change)

## Test plan

```sh
# All site actuator tests (new)
uv run --extra dev -m newton.tests -k TestMuJoCoSiteActuators
```

## New feature / API change

```python
import warp as wp
from newton import ModelBuilder
from newton.solvers import SolverMuJoCo

builder = ModelBuilder()
builder.add_mjcf("""
<mujoco>
  <worldbody>
    <body name="arm" pos="0 0 1">
      <joint name="shoulder" type="hinge" axis="0 1 0"/>
      <geom type="capsule" size="0.05" fromto="0 0 0 0.3 0 0" mass="1"/>
      <site name="end_effector" pos="0.3 0 0"/>
    </body>
  </worldbody>
  <actuator>
    <general name="ee_force" site="end_effector" gainprm="50 0 0" biasprm="0 -50 -5"/>
  </actuator>
</mujoco>
""", ctrl_direct=True)

model = builder.finalize()
solver = SolverMuJoCo(model, iterations=10)
```

## Summary by CodeRabbit

* **New Features**
  * MuJoCo solver now supports actuators that target sites and ensures site-referenced shapes are preserved during export (even when sites are disabled).

* **Tests**
  * Added unit tests validating site-targeted actuator parsing, export, parameter parity with native MuJoCo, and behavior when sites are excluded.

* **Documentation**
  * CHANGELOG updated to document site-targeted actuator support.

PR #2649, opened 2026-04-29
Vectorize SolverMuJoCo init site-actuator scan

adenzler-nvidia:
## Description

Fix a solver-init regression introduced in #2220 (site-targeted actuator support) where `SolverMuJoCo.__init__` time scales with `world_count * per-world size`. Visible across all dashboard runners on `setup.bench_model.KpiInitializeSolver` ([asv regressions](https://newton-physics.github.io/newton-asv/#regressions)).

The new code did two Python-level passes during init:

- A dict comprehension over the entire replicated `model.shape_label` (~400k entries for `ant` x 8192) to build `site_shape_by_label`.
- A loop over every entry of `actuator_trntype.numpy()` (~170k for `humanoid` x 8192) just to filter to the template world before resolving site targets.

Both are restricted to the template world here: `site_shape_by_label` is built from `selected_shapes`, the per-actuator loop is replaced by a vectorized `np.flatnonzero` on `actuator_world` plus a numpy mask for trnid targets, and the USD-deferred label path becomes a label-set intersection. Cost is back to O(per-world size).

### Timings (RTX PRO 6000, `KpiInitializeSolver`, world_count=8192, median of 3 measured runs)

| variant | `ant` | `humanoid` |
|---|---|---|
| `488e5e002` (last good, pre-#2220) | 0.264s | 0.661s |
| `main` (regressed) | 0.542s | 0.826s |
| this PR | **0.260s** | **0.601s** |

### Behavior note

The original code preferred label resolution and skipped trnid for the same actuator if the label matched. The vectorized version unions both paths. For well-formed actuators label and trnid agree, so the result is identical. In the pathological case of disagreement, the vectorized version retains both shapes; this is harmless because `actuator_required_shapes` is purely a "do not drop these shapes" set, never used to pick which shape an actuator targets.

## Checklist

## Test plan

\`\`\`
uv run --extra dev -m newton.tests -k test_mujoco_general_actuators
\`\`\`

All 22 tests pass on the patched code. Solver-init perf was measured with a script that mirrors `KpiInitializeSolver` on `ant` and `humanoid` at 8192 worlds against three commits (last-good, current main, and this PR) on the same machine.

## Bug fix

**Steps to reproduce:**

1. Run \`KpiInitializeSolver\` on any commit at or after #2220 (e.g. via the asv dashboard) with \`world_count=8192\`.
2. Compare against any commit before #2220.
3. Observe ~50-120% slower solver init across `humanoid`, `g1`, `cartpole`, and `ant`.

## Release Notes

* **Bug Fixes**
  * Fixed time scaling calculation in MuJoCo solver initialization where actuator counts were incorrectly computed

* **Performance Improvements**
  * Optimized actuator-target detection during MuJoCo solver initialization to reduce computational overhead and improve startup performance

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@
 - Fix `SensorRaycast` and viewer picking ignoring `HFIELD` (heightfield) geometry
 - Fix `ModelBuilder.add_shape_heightfield` `scale` being ignored by narrow-phase collision and raycast
 - Fix multi-world `qfrc_actuator` conversion using the wrong body center of mass for worlds with `worldid > 0`
+- Fix `SolverMuJoCo.__init__` time scaling with `world_count × actuators_per_world` instead of `actuators_per_world` by vectorizing the template-world filter for site-targeted actuators
 - Fix compressed tets in `evaluate_volumetric_neo_hookean_force_and_hessian` producing an indefinite Hessian by clamping the cofactor-derivative coefficient to `max(0, s)`, removing a contribution that could corrupt the VBD inner solve
 - Fix SDF hydroelastic broadphase scatter kernel using a grid-stride loop with binary search instead of per-pair thread launch
 - Fix box support-map sign flips from quaternion rotation noise (~1e-14) producing invalid GJK/MPR contacts for face-touching boxes with non-trivial base rotations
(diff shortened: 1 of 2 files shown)


vreutskyy (approved):
Looks good

PR #2651, opened 2026-04-29
Fix MPR/GJK bug that assumed convex hulls are centered around the origin

nvtw:
## Description
Fix incorrect MPR/GJK contact reporting for convex hulls whose authoring origin lies outside the hull's AABB. 
- `geometric_center` (MPR/GJK initial interior point `v0`) now uses the on-the-fly AABB center for `CONVEX_MESH` shapes instead of the local origin, avoiding degenerate portals where supports collapse onto a single partner vertex.

## Checklist

## Test plan
```
uv run --extra dev -m newton.tests -k test_off_origin_convex_hull_contacts
```

## Bug fix
For a `CONVEX_MESH` whose local origin sits outside its AABB (common in imported collision assets), MPR picks `v0` at the local origin. The resulting initial portal is degenerate against thin partners (e.g. a single mesh triangle) and MPR converges on a wrong normal/penetration even when the shapes are clearly separated. Heightfield mid-phase additionally misses triangles because the origin-centered bounding sphere does not enclose the hull.

**Steps to reproduce:**
1. Import a convex hull whose vertices are offset ~20+ cm from the shape's local origin.
2. Place a small mesh triangle ~1.4 mm away from the hull along world Z.
3. Run the collision pipeline and inspect the reported contact.
4. Without the fix: ~7.1 mm penetration along +Y. With the fix: +1.42 mm separation along -Z.

**Minimal reproduction:**
See `newton/tests/test_off_origin_convex_hull_contacts.py`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected contact generation for off-origin convex meshes and improved heightfield-vs-convex accuracy by using per-shape local AABBs.
  * Improved bounding-sphere computation for mesh geometries using local AABBs.

* **API Changes**
  * Narrow-phase and heightfield midphase interfaces now accept per-shape local AABB inputs to drive triangle overlap and broadphase.

* **Tests**
  * Added a regression test verifying correct separation for off-origin convex-hull scenarios.

* **Docs**
  * Updated docstrings to clarify local-frame conventions and AABB-based logic.


adenzler-nvidia (commented):
Two small follow-ups.

Per `AGENTS.md`, user-facing behavior changes need a `CHANGELOG.md` entry under `[Unreleased]`. Off-origin convex hulls reporting wrong contacts is squarely user-facing — would you mind adding a `Fixed:` entry, e.g. "Fix MPR/GJK reporting wrong contacts for off-origin `CONVEX_MESH` shapes by initialising the v0 portal to the local AABB centre instead of the local origin"?

newton/tests/test_off_origin_convex_hull_contacts.py
```diff
@@ -0,0 +1,266 @@
...
+    def _to_world(shape_idx: int, local: np.ndarray) -> np.ndarray:
+        body_idx = int(shape_body[shape_idx])
+        if body_idx < 0:
+            return local
+        bx = body_q[body_idx]
+        pos = bx[:3]
+        q = bx[3:]
+        # Rotate v by quaternion (xyzw): v + 2*q_xyz x (q_xyz x v + q_w v)
+        qv = q[:3]
+        t = 2.0 * np.cross(qv, local)
```
adenzler-nvidia, line 192: Tiny: there's already a `wp.quat_rotate` you could call from numpy via `wp.quat_rotate(...)` / `wp.transform_point(...)` (or via numpy quaternion code in `unittest_utils` if there's a helper). Re-deriving the rotation here means future bugs in `v + 2*q_xyz × (q_xyz × v + q_w v)` hide in test code rather than production.

nvtw: Done

newton/_src/geometry/mpr.py
```diff
@@ -174,11 +191,45 @@ def geometric_center(
...
+            in world space; ``BtoA = center_a - center.B`` mixes A-local
+            with world space, which is the convention used by the
+            ``solve_mpr_core`` / ``solve_gjk_core`` callers.
         """
         center = Vert()
 
         center_a = wp.vec3(0.0, 0.0, 0.0)
+        center_b_local = wp.vec3(0.0, 0.0, 0.0)
+
+        if geom_a.shape_type == int(GeoType.CONVEX_MESH):
```
camevor, line 204: Question: since this is being called for each collision detection step, could we precompute and cache the center instead?

nvtw: I thought about that but I'm not sure if it is worth it: This initial direction guess is called once per MPR/GJK iteration. The support function is called up to (with default values) 30 times and the support function for convexes iterates over all convex points as well. So I think precomputing it would not help much and lead to increased memory consumption. I tried to pass in the local aabb bounds and compute it from them but it just makes the code bloated without helping much.

camevor (approved):
Approving as discussed, thanks!

PR #2654, opened 2026-04-29
Vbd Neohookean Material Tests

AnkaChan:
## Description

Two new examples cover the regression scenarios the previous StVK form could not handle:

- **`VBD/Cloth/example_cloth_stiff_material_hanging.py`** — square cloth at `tri_ke = tri_ka = 1e8` pinned along one edge under gravity. Under StVK these parameters drove the membrane Hessian non-PSD under compression and VBD diverged; under StableNH the cloth swings to a stable hanging curtain. `test_final` adds a per-particle velocity bound on top of the existing position/volume checks.
- **`VBD/Cloth/example_cloth_stiff_material_stretch.py`** — five sheets pinned at +/-X edges and stretched by λ = 2, each with a different effective Poisson ratio set via `tri_ka = 2 · tri_ke · ν / (1 − ν)`. `test_final` checks the bulk area ratio (centerline-based to avoid corner-pinning artifacts) against the closed-form StableNH 2D-membrane prediction
  `A(λ) / A_0 = 2λ² / [(1 − ν) + (1 + ν) λ²]`
  within 10% per sheet. NH lands within ~5% of theory across `ν ∈ {0.10, 0.20, 0.30, 0.40, 0.49}`; the same example on StVK deviates up to +22% at low ν.

https://github.com/user-attachments/assets/e93c3068-9986-43b7-85b2-7cd2bb988720

https://github.com/user-attachments/assets/41ef9bc8-9c9b-4b9f-8de7-a6e0b0917720

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_VBD.Cloth.example_cloth_stiff_material_hanging
uv run --extra dev -m newton.tests -k test_VBD.Cloth.example_cloth_stiff_material_stretch
uv run --extra dev -m newton.tests -k test_softbody          # NH tet autodiff coverage
```

Both new example tests pass on CUDA (`Ran 1 test in 11.951s — OK` for stretch, ~59 s for hanging).

The stretch example was also run against `origin/main` (StVK) to confirm the area-deviation gap that motivates the new material:

| ν | NH measured | NH theory | StVK measured |
|---:|---:|---:|---:|
| 0.10 | 1.505 | 1.509 | 1.837 |
| 0.20 | 1.403 | 1.429 | 1.669 |
| 0.30 | 1.327 | 1.356 | 1.491 |
| 0.40 | 1.243 | 1.290 | 1.316 |
| 0.49 | 1.176 | 1.236 | 1.184 |

## New feature / API change

The user-facing API does not change shape; `tri_ke` / `tri_ka` continue to set the membrane material per triangle. The interpretation of those numbers does change — they're now the Lamé parameters of the StableNH energy, so the same numerical values describe a different (but more robust and predictable) constitutive response.

```python
import warp as wp
import newton

# Effective Poisson ratio on the new material:
#   nu = tri_ka / (tri_ka + 2 * tri_ke)
# So to target nu, choose tri_ka = 2 * tri_ke * nu / (1 - nu).

builder.add_cloth_grid(
    pos=wp.vec3(0.0, 0.0, 1.0),
    rot=wp.quat_identity(),
    vel=wp.vec3(0.0, 0.0, 0.0),


    dim_x=20, dim_y=20, cell_x=0.05, cell_y=0.05,
    mass=1.0,
    fix_left=True, fix_right=True,
    tri_ke=1.0e3,                # mu (Lamé)
    tri_ka=2.0 * 1.0e3 * 0.30 / (1.0 - 0.30),  # ~ nu = 0.30
    tri_kd=1e-4,
)
```

For an end-to-end demonstration, see `newton/examples/VBD/Cloth/example_cloth_stiff_material_stretch.py`, which sweeps five Poisson ratios in a single scene and the `test_final` check verifies the resulting area ratios against the closed-form prediction.

## Release Notes

* **New Features**
  * Added cloth hanging simulation example to test stiff material stability under gravity
  * Added cloth stretch simulation example to validate bulk area preservation

* **Tests**
  * Added tests for the new cloth simulation examples

* **Documentation**
  * Updated changelog documenting new cloth example regressions


newton/examples/vbd/example_cloth_stiff_material_hanging.py
eric-heiden, line 1: I think these examples should go under `newton/examples/cloth/`. Can we also integrate this example into `example_cloth_hanging.py`? Just wondering if we really need that many cloth examples. The stretching example could maybe also be combined with the twist example somehow, or converted into a unit test?

newton/examples/VBD/Cloth/example_cloth_stiff_material_stretch.py
```diff
@@ -0,0 +1,244 @@
...
+        self.control = self.model.control()
+        self.contacts = self.model.contacts()
+
+        # Right edge starts unstretched and is ramped to STRETCH * rest_side
+        # over RAMP_FRAMES — easing avoids the instant force spike that breaks
+        # VBD at low Poisson ratios.
+        self._frame_index = 0
+
+        # Triangle index ranges per sheet, for area measurement.
+        tri_ind = self.model.tri_indices.numpy()
```
eric-heiden: ⚪ This host copy and `self.sheet_tri_ranges` bookkeeping look stale now that `measure_area_ratios()` measures the centerline column directly. Since nothing reads `self.sheet_tri_ranges`, this adds setup work without affecting the example.

Please remove the `tri_ind` copy and range construction unless a later assertion is going to use them.

newton/examples/VBD/Cloth/example_cloth_stiff_material_hanging.py
```diff
@@ -0,0 +1,164 @@
...
+        mesh_indices = cloth_mesh.indices
+
+        vertices = [wp.vec3(v) for v in mesh_points]
+
+        builder = newton.ModelBuilder()
+        builder.add_ground_plane()
+        builder.add_cloth_mesh(
+            pos=wp.vec3(0.0, 0.0, 2.0),
+            rot=wp.quat_from_axis_angle(wp.vec3(1, 0, 0), np.pi / 2),
+            # rot=wp.quat_identity(),
```
eric-heiden: ⚪ This commented-out alternative looks like local tuning scaffolding in a public example. Please drop it, or turn it into an actual documented option if users are expected to switch orientations.

newton/examples/VBD/Cloth/example_cloth_stiff_material_stretch.py
```diff
@@ -0,0 +1,244 @@
...
+#
+#     A(lambda) / A_0 = (lmbd_nh + mu) * lambda^2
+#                       ---------------------------------
+#                          mu + lmbd_nh * lambda^2
+#
+# where mu = tri_ke, lmbd_nh = tri_ka + tri_ke, and Poisson ratio
+# nu = tri_ka / (tri_ka + 2 * tri_ke).
+#
+# Run on `main` (StVK) and on `vbd_stiff_material_stretch` (Neo-Hookean)
+# to compare against the same theoretical curve.
```
eric-heiden: ⚪ This header references `vbd_stiff_material_stretch`, which appears to be a PR-local branch name. After merge, users reading the shipped example will not be able to follow this comparison, and `main` will already contain the Neo-Hookean code.

Please rephrase this as self-contained documentation, for example by saying the example validates the Neo-Hookean membrane against the StableNH closed-form prediction.

newton/examples/vbd/example_cloth_stiff_material_hanging.py
mmacklin: @eric-heiden and I discussed and will build `examples/vbd` as the set of robustness and correctness tests

(All lowercase, no subdirectories)

AnkaChan: Done — moved both examples to `examples/vbd/` (flat, all lowercase) and updated test registrations. Removed the `VBD/Cloth/` path.

newton/examples/VBD/Cloth/example_cloth_stiff_material_stretch.py
AnkaChan: Removed — the `tri_ind` copy and `sheet_tri_ranges` construction were leftover from before `measure_area_ratios()` switched to centerline measurement.

newton/examples/VBD/Cloth/example_cloth_stiff_material_hanging.py
AnkaChan: Dropped the commented-out line.

newton/examples/VBD/Cloth/example_cloth_stiff_material_stretch.py
AnkaChan: Rephrased — the header now says the example validates the Neo-Hookean membrane against the StableNH closed-form prediction, no branch reference.

AnkaChan:
@eric-heiden Can we merge this?

eric-heiden:
> @eric-heiden Can we merge this?

@AnkaChan it looks good, we just need these new examples with screenshots in the example gallery of README.md.

AnkaChan:
> > @eric-heiden Can we merge this?
> 
> @AnkaChan it looks good, we just need these new examples with screenshots in the example gallery of README.md.

Can we merge this after our discussion on slack?

CHANGELOG.md
```diff
@@ -19,6 +19,9 @@
...
+- Add `cloth_stiff_material_hanging` example: stiff square cloth (`tri_ke = tri_ka = 1e8`) pinned along one edge under gravity, exercising the new Neo-Hookean triangle material at parameters where the previous StVK form would explode. `test_final` checks for non-explosion via a per-particle velocity bound in addition to the existing position-volume checks
+- Add `cloth_stiff_material_stretch` example: five cloth sheets pinned at +/-X edges and stretched, each with a different effective Poisson ratio set via `tri_ka / tri_ke`. `test_final` checks bulk area against the closed-form StableNH 2D-membrane prediction within 10%, regression-guarding the new triangle material model
+- Add `cloth_stiff_material_hanging` and `cloth_stiff_material_stretch` examples regression-guarding the new Neo-Hookean triangle material (stability under gravity at extreme stiffness, and bulk area-preservation across a Poisson-ratio sweep)
```
eric-heiden: I think the other 2 entries already capture the new examples, so we don't need to repeat this bullet point?

newton/tests/test_examples.py
```diff
@@ -355,6 +355,20 @@ class TestClothExamples(unittest.TestCase):
...
+    name="vbd.example_cloth_stiff_material_hanging",
+    devices=cuda_test_devices,
+    test_options={"usd_required": True, "num-frames": 360},
+    use_viewer=True,
+)
+add_example_test(
+    TestClothExamples,
+    name="vbd.example_cloth_stiff_material_stretch",
+    devices=cuda_test_devices,
+    test_options={"num-frames": 600},
```
eric-heiden: The runtime is not bad (about 14s on CI) compared to other examples but just double check if we need to roll out that many frames.

newton/examples/vbd/example_cloth_stiff_material_hanging.py
AnkaChan: Confirming the final layout honors this: both examples live directly under `newton/examples/vbd/` (flat, all lowercase) — `example_cloth_stiff_material_hanging.py` and `example_cloth_stiff_material_stretch.py`, with no subdirectory. The test registrations use the matching `vbd.example_cloth_stiff_material_*` names.

newton/examples/vbd/example_cloth_stiff_material_hanging.py
```diff
@@ -0,0 +1,163 @@
...
+        self.model.soft_contact_kd = 1.0e0
+        self.model.soft_contact_mu = 1.0
+
+        cloth_size = 50
+        fixed_side = [cloth_size - 1 + i * cloth_size for i in range(cloth_size)]
+
+        flags = self.model.particle_flags.numpy()
+        for fixed_vertex_id in fixed_side:
+            flags[fixed_vertex_id] = flags[fixed_vertex_id] & ~ParticleFlags.ACTIVE
+        self.model.particle_flags = wp.array(flags)
```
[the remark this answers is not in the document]
AnkaChan: `cloth_size` is now derived from the loaded mesh instead of hardcoded: `cloth_size = int(round(len(vertices) ** 0.5))`, with an assert that the vertex count is a perfect square. The pinned column now tracks the asset resolution instead of assuming 50×50.

CHANGELOG.md
AnkaChan: Consolidated to a single `[Unreleased]` entry covering both examples and removed the redundant per-example bullets. (The merge with main — which had cut the 1.2.0 release — also duplicated these into the released [1.2.0] section; that's cleaned up in the same commit.)

newton/tests/test_examples.py
AnkaChan: Reduced the stretch example test to 360 frames (matching the hanging example). `test_final` still passes — the area-ratio-within-10% check holds for all five Poisson-ratio sheets at 360 frames (200-frame ramp + 160 settle).

AnkaChan:
Pushed an update resolving the conflict with `main` and addressing the open review feedback:

- **Merge conflict:** merged current `main` (which had cut the 1.2.0 release). The merge had pulled the example bullets and a duplicated `validate_mesh` entry into the released `[1.2.0]` CHANGELOG section — moved to a single `[Unreleased]` entry instead.
- **`cloth_size`** (hanging): now derived from the loaded vertex count with a square-mesh assert, instead of hardcoding `50`, so the pinned column tracks the asset resolution.
- **Docstrings** (stretch helpers): added Google-style `Args:`/`Returns:` sections.
- **Stretch ramp:** skip the per-frame `particle_q` round-trip once the ramp completes (the right edge is then held by `fix_right`).
- **Stretch test:** `600 → 360` frames; `test_final` still passes for all five Poisson-ratio sheets.

The earlier CodeRabbit CHANGELOG notes on the StVK → Neo-Hookean *Changed* entry and the tet `max(0, s)` *Fix* entry are now on `main` (that material change shipped separately in 1.2.0), so they're outside this PR's diff.

Both `vbd.example_cloth_stiff_material_*` example tests pass on CUDA, and pre-commit is clean.

jcarius-nv:

AnkaChan:
@jcarius-nv done — merged current `main` (post-#3145), and the entry now sits under the top-level `[Unreleased]` → `Added`, not under the released `[1.3.0]`. The auto-merge had initially pulled it (plus some already-released bullets) into `[1.3.0]`; I reset `CHANGELOG.md` to `main` and re-added just the single example entry to `[Unreleased]`.

newton/examples/vbd/example_cloth_stiff_material_stretch.py
```diff
@@ -0,0 +1,252 @@
...
+    def render(self):
+        self.viewer.begin_frame(self.sim_time)
+        self.viewer.log_state(self.state_0)
+        self.viewer.log_contacts(self.contacts, self.state_0)
+        self.viewer.end_frame()
+
+
+if __name__ == "__main__":
+    parser = newton.examples.create_parser()
+    viewer, args = newton.examples.init(parser)
```
eric-heiden, line 250: 🟡 `create_parser()` keeps the shared `--num-frames` default of 100, but this example ramps the stretch over `RAMP_FRAMES = 200`. That means the default command stops halfway through the ramp, and `--test` then runs `test_final()` against the full `STRETCH = 2.0` theory before the setup reaches that state.

Please set this example's frame count before `newton.examples.init(parser)`, like the other long-running examples do. Matching the test registration would be enough:

```suggestion
    parser.set_defaults(num_frames=360)
    viewer, args = newton.examples.init(parser)
```

eric-heiden (approved):

PR #2657, opened 2026-04-30
Unified Solver Reset and Model Flags

### LINKED ISSUE #1266: Extending the solver API to handle reset operations
@mmacklin @eric-heiden 

As we're preparing the integration of Kamino into Newton, the topic of how sim resets happen and how to best provide a front-end has come up once more. The question is particularly important, and gets especially challenging, for the case of models with closed kinematic loops. 

One simple solution could be to extend the API of `SolverBase` to define a `reset()` method that each solver can override.
This could look something like:
```python
class SolverBase:
    """..."""

    def __init__(self, model: Model):
        self.model = model

    @property
    def device(self) -> wp.context.Device:
        """..."""
        return self.model.device

    def reset(
        self,
        state_out: State,
        joint_q: wp.array | None = None,
        joint_qd: wp.array | None = None,
        base_q: wp.array | None = None,
        base_qd: wp.array | None = None,
        world_mask: wp.array | None = None,
    ):
        """
        Resets the simulation state given a combination of desired base body
        and joint states, as well as an optional per-world mask array indicating
        which worlds should be reset. The reset state is written to `state_out`.

        For resets given absolute quantities like base body poses, the 
        `state_out` must initially contain the current state of the simulation.

        Args:
            state_out (State):
                The output state container to which the reset state data is written.
            joint_q (wp.array, optional):
                Optional array of target joint coordinates.
                Shape of `(num_joint_coords,)` and type :class:`wp.float32`
            joint_qd (wp.array, optional):
                Optional array of target joint DoF velocities.
                Shape of `(num_joint_dofs,)` and type :class:`wp.float32`
            base_q (wp.array, optional):
                Optional array of target base body poses.
                Shape of `(num_worlds,)` and type :class:`wp.transformf`
            base_qd (wp.array, optional):
                Optional array of target base body twists.
                Shape of `(num_worlds,)` and type :class:`wp.spatial_vectorf`
            world_mask (wp.array, optional):
                Optional array of per-world masks indicating which worlds should be reset.
                Shape of `(num_worlds,)` and type :class:`wp.int8 | wp.bool`
        """
        raise NotImplementedError()

    def step(self, state_in: State, state_out: State, control: Control, contacts: Contacts, dt: float):
        """..."""
        raise NotImplementedError()
    ...
```

This type of reset API could support all types of reset operations AFAIK. Specifically, it can account for the following cases:
1. The per-world mask array `world_mask`, if specified, only resets the target worlds. I not specified, it resets all worlds.
2. If none of the joint/base-body arrays are provided, then it resets to the default/initial state defined in the `Model`.
3. If only `joint_q` is specified, it resets the the joint configuration and leaves everything else as-is.
4. If `joint_q` and `joint_qd` are specified, it resets the full reduced-coordinate state of the joints.
5. If `base_q` is specified, then it leaves the joints state as-is and uniformly transforms all bodies and/or `FREE` joints to achieve the target pose.
6. If `base_q` and `base_qd` are specified, then it leaves the joints state as-is and uniformly transforms all bodies and/or `FREE` joints to achieve the target base pose and twist.
7. Combines joint-only and base-only behaviors given that the combination is valid.
8. Raises appropriate errors if combinations are not supported by the given solver.

Additional notes:
* The joint arrays can be over all joints, both passive and actuated, and we can leave it up to the users to know how to fill them in.
* The `base` body is identical to the `root` body of an articulation, but we avoid using the latter term to generalize to systems w/o an explicit articulation as well as those containing kinematic loops. 
* `base` body pose/twist arrays are provided explicit arguments to also consider the case of models which may still be "floating-base", but may not define `FREE` joints.

eric-heiden:
## Summary
Introduces a shared in-place solver reset API with state-reset flags, and standardizes model-change flags so solver integrations stay consistent and backward compatible.

## What changed
- Add SolverBase.reset() and SolverStateFlags for solver state resets

## Touched areas
- `CHANGELOG.md`
- `newton/_src/solvers/__init__.py`
- `newton/_src/solvers/flags.py`
- `newton/_src/solvers/kamino/solver_kamino.py`
- `newton/_src/solvers/solver.py`
- `newton/solvers.py`

## Validation
No tests were run while creating this PR.

Closes #1266

## Summary by CodeRabbit

* **New Features**
  * Added SolverBase.reset() for in-place solver state resets with optional world-mask and SolverStateFlags.
  * Added SolverStateFlags and SolverModelFlags enums.

* **Changed**
  * SolverKamino.reset() signature standardized to accept state (positional compatibility retained).
  * Solvers now interpret model-change bitmasks using SolverModelFlags.

* **Deprecated**
  * SolverNotifyFlags deprecated in favor of SolverModelFlags.
  * SolverVBD.rigid_enable_dahl_friction deprecated/ignored; Dahl friction now auto-detected.

* **Documentation**
  * API docs and examples updated to reference SolverModelFlags/SolverStateFlags.

* **Tests**
  * New test ensuring no remaining references to SolverNotifyFlags.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2657?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


newton/_src/solvers/solver.py
```diff
@@ -329,17 +356,17 @@ def notify_model_changed(self, flags: int) -> None:
...
+        ``SolverModelFlags.BODY_INERTIAL_PROPERTIES``    Rigid-body mass or inertia tensors have changed.
+        ``SolverModelFlags.SHAPE_PROPERTIES``            Shape transforms or geometry have changed.
+        ``SolverModelFlags.MODEL_PROPERTIES``            Model global properties (e.g., gravity) have changed.
         ==============================================  =============================================================
 
         Args:
-            flags (int): Bit-mask of model-update flags indicating which model
-                properties changed.
+            flags (int): Bit-mask of :class:`SolverModelFlags` indicating which
+                model properties changed.
```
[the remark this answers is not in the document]
eric-heiden: Fixed in 2ca6ba4ac by adding the missing SolverModelFlags.CONSTRAINT_PROPERTIES, SolverModelFlags.TENDON_PROPERTIES, and SolverModelFlags.ACTUATOR_PROPERTIES rows to the notify_model_changed() valid-flags table.

eric-heiden:
Addressed the remaining CodeRabbit review items on the current branch:

- notify_model_changed() valid-flags table: fixed in 2ca6ba4ac by documenting SolverModelFlags.CONSTRAINT_PROPERTIES, SolverModelFlags.TENDON_PROPERTIES, and SolverModelFlags.ACTUATOR_PROPERTIES.
- Outside-diff SolverKamino.reset(flags=...) item: no additional code change needed on the current branch. SolverKamino.reset() now applies the mask via state_flags and _preserve_if_unset(...) before/after the underlying Kamino reset, preserving state arrays outside the requested SolverStateFlags mask.

docs/conf.py
```diff
@@ -171,7 +171,7 @@ def _ensure_pandoc_on_path() -> str | None:
     "python": ("https://docs.python.org/3", None),
     "numpy": ("https://numpy.org/doc/stable", None),
     "jax": ("https://docs.jax.dev/en/latest", None),
-    "pytorch": ("https://pytorch.org/docs/stable", None),
+    "pytorch": ("https://docs.pytorch.org/docs/2.10", None),
```
adenzler-nvidia: This looks like a leftover from a bad merge.

CHANGELOG.md
```diff
@@ -101,6 +101,9 @@
 - Add site-targeted actuator support to MuJoCo solver
 - Add USD parsing support for equality constraints based on the `MjcEquality` schema
 - Add more solver options to implicit MPM: `gs-soa` (or `gauss-seidel-soa`) for improved memory coalescing, `gs-batched` (or `gauss-seidel-batched`) merging GS colors with Jacobi-style mass-split parallelism, plus `cr` (Conjugate Residual) and `gmres` linear solver options.
+- Add `SolverBase.reset()` method for in-place solver state resets with optional `world_mask` and `SolverStateFlags`; default implementation is a no-op
+- Add `SolverStateFlags` enum to control which state attributes are reset
+- Add `SolverModelFlags` as the canonical name for model-change notification flags
```
adenzler-nvidia: I'm assuming the API has been discussed and approved, can you confirm?

adenzler-nvidia:
**CHANGELOG.md (~line 142):** the new `### Deprecated` block is inserted mid-`### Changed`, leaving a duplicate `### Deprecated` heading and pushing the existing `- Reduce collision-pipeline overhead …` entry under the wrong section — please fold the new deprecation lines into the existing `### Deprecated` section. Also, the `SolverKamino.reset()` entry should call out that the reset is now **in place on `state`** (not written to a separate output container), since that's the behavioral change for existing callers.

newton/_src/solvers/kamino/solver_kamino.py
```diff
@@ -437,43 +437,101 @@ def __init__(
...
         self._control_kamino.finalize(self._model_kamino)
 
+    @override
     def reset(
         self,
-        state_out: State,
+        state: State | None = None,
         world_mask: wp.array | None = None,
+        flags: SolverStateFlags | None = None,
+        *legacy_targets: wp.array | None,
```
Guirec-Maloisel: Sorry for the late review, last week was taken by a deadline on the Kamino side.

After a quick discussion in the Kamino side, we seemed to agree that since Kamino is still in beta, we'd prefer keeping the interface readable and avoiding the added machinery for backward compatibility, if you're comfortable with this (so taking out `legacy_targets` and `state_out` from the signature). The only caveat is that integration code into Isaac Lab would need to update as well, but the changes should be minimal.

newton/_src/solvers/kamino/solver_kamino.py
```diff
@@ -485,12 +543,19 @@ def reset(
...
         # TODO: fix brittle in-place update of arrays after conversion
-        # Create a zer-copy view of the input state_out as a StateKamino
-        # to interface with the Kamino solver's reset operation
-        state_out_kamino = self._kamino.StateKamino.from_newton(self._model_kamino.size, self.model, state_out)
+        state_out_kamino = self._kamino.StateKamino.from_newton(self._model_kamino.size, self.model, state)
+        restore_after_reset: list[tuple[wp.array, wp.array]] = []
+
+        def _preserve_if_unset(array: wp.array | None, flag: SolverStateFlags) -> None:
+            if array is not None and not (state_flags & flag):
+                restore_after_reset.append((array, wp.clone(array)))
```
Guirec-Maloisel: This code will break graph capture, which is undesirable as reset operations are part of the main graph that runs in a loop within RL training. I don't see an immediate fix though, and this would only affect partial resets which I don't think are in use currently, so we could potentially add a comment for now, and fix this internally in Kamino in a later PR.

Regardless of this, the call to `wp.clone()` should specify an explicit device (e.g. the state's or model's device)

newton/_src/solvers/kamino/solver_kamino.py
```diff
@@ -485,12 +543,19 @@ def reset(
             base_q = base_q_com
 
         # TODO: fix brittle in-place update of arrays after conversion
-        # Create a zer-copy view of the input state_out as a StateKamino
```
Guirec-Maloisel, line 488: Could you restore this comment and the two other removed comments below? and rather add a new comment above the added lines

newton/_src/solvers/kamino/solver_kamino.py
eric-heiden: The code itself doesn't break graph capture - `wp.clone` is allowed even if we are creating a local temporary array here that goes out of scope before graph replay. I've also tested this locally.

This explicit device thing should not be needed because `wp.clone()` will use the device of the source array which already is on the device where it should be.

CHANGELOG.md
eric-heiden: Yes, this is what we agreed on in the architecture meeting.

newton/_src/solvers/kamino/solver_kamino.py
Guirec-Maloisel: My bad, that's good to know!!

Guirec-Maloisel (dismissed):
This is good to go now from the Kamino side!

newton/_src/solvers/solver.py
```diff
@@ -298,6 +299,32 @@ def integrate_particles(
                 device=model.device,
             )
 
+    def reset(
+        self,
+        state: State,
+        world_mask: wp.array | None = None,
+        flags: SolverStateFlags | None = None,
```
eric-heiden: Rename to `StateFlags`. Investigate how custom solvers could accept their own custom attributes here, bitmask is fine.

eric-heiden: Other solvers can add their own integers here.

CHANGELOG.md
```diff
@@ -44,21 +47,27 @@
...
+- Deprecate `SolverNotifyFlags` in favor of `ModelFlags`; migrate calls such as `SolverNotifyFlags.MODEL_PROPERTIES` to `ModelFlags.MODEL_PROPERTIES`
 - Deprecate implicit positive Dahl defaults in `SolverVBD.register_custom_attributes()`. Pass `dahl_defaults_enabled=False` and explicitly author positive `model.vbd.dahl_eps_max` and `model.vbd.dahl_tau` values when Dahl cable friction is desired, instead of relying on registered default values.
 - Deprecate `newton:contact_ke`/`newton:contact_kd`/`newton:contact_kf`/`newton:contact_ka` custom attributes on shape prims; author `newton:contactStiffness`/`newton:contactDamping`/`newton:contactFrictionGain`/`newton:contactAdhesion` on the bound `NewtonMaterialAPI` material instead
 - Deprecate `mjc:solref` on material prims for contact stiffness/damping; author `newton:contactStiffness`/`newton:contactDamping` on the bound `NewtonMaterialAPI` material, or use per-shape `mjc:solref` (`MjcGeomAPI`) which is unchanged
 
 ### Removed
 
+- Remove `SolverModelFlags` and `SolverStateFlags`; use `ModelFlags` and `StateFlags` instead.
```
mmacklin: Is this out of date? I don't think the `SolverModelFlags` got merged did it?

adenzler-nvidia: yes looks like an accidental incremental changelog entry while this branch evolved.

adenzler-nvidia (dismissed):
Approving. The flag consolidation reads cleanly: `ModelFlags`/`StateFlags` are promoted to public top-level API and exported, internal solvers are all migrated, and `SolverNotifyFlags` is kept as a `DeprecationWarning`-emitting alias with a 3.10-safe `EnumMeta` shim. Deprecation/removal entries carry migration guidance, and the CHANGELOG churn is cleaned up.

Two optional, non-blocking follow-ups:

- The two `SolverKamino.reset()` CHANGELOG entries (one under Changed, one under Removed) describe the same edit and overlap; consider collapsing into a single entry since the solver is beta.
- The existing test suite still references the deprecated `SolverNotifyFlags` (test_mujoco_solver.py and others); migrating those call sites to `ModelFlags` would avoid emitting deprecation warnings from our own tests.

adenzler-nvidia (dismissed):
Re-reviewed the full diff against `main` at `8784ccb`. The flag consolidation is clean and complete: `ModelFlags`/`StateFlags` are promoted to public top-level API, all six internal solvers and every `SolverKamino.reset` call site are migrated, and `SolverNotifyFlags` is retained as a deprecation-emitting alias rather than removed. The `IntEnum` (not `IntFlag`) choice correctly lets custom solver bits survive `|`, and `test_custom_solver.py` guards that contract. No correctness issues found.

Approving. A few optional, non-blocking nits for a follow-up if convenient:

- The deprecation message says "use ModelFlags instead" but the symbol moved namespaces (`newton.solvers.SolverNotifyFlags` -> top-level `newton.ModelFlags`, which isn't re-exported from `newton.solvers`). Spelling it `newton.ModelFlags` in the warning would smooth migration.
- The seven `testing/joints/*.usda` fixtures each only drop a trailing blank line, unrelated to the refactor — could be reverted to keep the diff focused.

adenzler-nvidia (approved):

chschuma-disney (approved):

PR #2658, opened 2026-04-30
Preserve ViewerGL state across compatible model switches

han-xudong:
## Description

Preserve `ViewerGL` camera and wind state when switching between models that use the same `up_axis`.

`ViewerGL.set_model()` needs to rebuild model-dependent viewer state, but it was also recreating `Camera` and `Wind` unconditionally. That discarded interactive/headless viewer state such as camera position, yaw, pitch, pivot, and configured wind parameters during otherwise compatible model switches.

This change restores prior camera state when the old and new models share the same `up_axis`, while still rebuilding the default camera when the axis changes. It also preserves wind parameters across model switches.

## Checklist

## Test plan

```bash
uv run --extra dev -m unittest newton.tests.test_viewer_camera.TestViewerGLSetModelState.test_set_model_preserves_camera_and_wind_for_same_up_axis
uv run --extra dev -m unittest newton.tests.test_viewer_camera
uv run --extra dev pre-commit run -a
```

## Bug fix

**Steps to reproduce:**

1. Create a `ViewerGL(headless=True)` instance.
2. Load one model, then set custom camera position/yaw/pitch/pivot and wind parameters.
3. Call `viewer.set_model()` with another model that uses the same `up_axis`.
4. Observe that the old behavior resets the camera and wind state to defaults.

**Minimal reproduction:**

```python
import warp as wp
import newton


def build_model(z: float) -> newton.Model:
    body = builder.add_body(
        xform=wp.transform(wp.vec3(0.0, 0.0, z), wp.quat_identity()),
        mass=1.0,
        inertia=wp.mat33(1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0),
    )
    builder.add_shape_box(body=body, hx=0.5, hy=0.5, hz=0.5)
    return builder.finalize()


viewer = ViewerGL(headless=True)
try:
    viewer.set_model(build_model(1.0))
    viewer.camera.pos = viewer.camera._as_vec3((3.0, 4.0, 5.0))
    viewer.camera.yaw = 37.0
    viewer.camera.pitch = -12.0
    viewer.camera.set_pivot((1.0, 2.0, 3.0))
    viewer.wind.amplitude = 5.0
    viewer.wind.frequency = 7.0

    viewer.set_model(build_model(2.0))

    print(tuple(float(x) for x in viewer.camera.pos))
    print(viewer.camera.yaw, viewer.camera.pitch)
    print(tuple(float(x) for x in viewer.camera.pivot))
    print(viewer.wind.amplitude, viewer.wind.frequency)
finally:
    viewer.close()
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Camera state (position/orientation/pivot and view settings) and wind simulation parameters are now preserved when switching between models that share the same up-axis.
  * When switching to a model with a different up-axis, the camera is reset to defaults as expected.
* **Tests**
  * Added unit coverage to verify state preservation and correct reset behavior across model changes.
* **Documentation**
  * Updated the changelog with an “Unreleased → Fixed” entry for this issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 - Fix MJCF worldbody static geoms bypassing the visual/collider class filter, so `parse_visuals=False` drops visual-class geoms attached directly to `<worldbody>` too. (#3030)
 - Fix `cable_cross_slide_table` example stability so the cable-driven table reliably tracks its rectangular path and catches drift during regression runs.
 - Fix URDF `package://` mesh fallback resolution without `resolve-robotics-uri-py` so package names only match full path components instead of unrelated directory-name substrings
+- Fix `ViewerGL.set_model()` resetting headless/interactive camera and wind state when switching between models that use the same up-axis. (#2658)
 
 ### Removed
 
--- a/newton/_src/viewer/viewer_gl.py
+++ b/newton/_src/viewer/viewer_gl.py
@@ -529,7 +529,16 @@ class ViewerGL(ViewerBase):
 
         Args:
             model: The Newton model instance.
+
+        Note:
+            Switching between models with the same up-axis preserves the
+            existing camera state. Wind settings are preserved across
+            non-``None`` model switches because they are independent of the
+            model up-axis.
         """
+        prev_camera = self.camera
+        prev_wind = self.wind
+
         super().set_model(model)
 
         # ``ViewerBase.set_model`` may have switched ``self.device`` to the
@@ -598,6 +607,19 @@ class ViewerGL(ViewerBase):
 
         fb_w, fb_h = self.renderer.window.get_framebuffer_size()
         self.camera = Camera(width=fb_w, height=fb_h, up_axis=model.up_axis if model else "Z")
+
+        if prev_camera is not None and model is not None and prev_camera.up_axis == self.camera.up_axis:
+            # Reuse the compatible camera so future Camera fields survive model switches too.
+            prev_camera.update_screen_size(fb_w, fb_h)
+            self.camera = prev_camera
+
+        if prev_wind is not None and model is not None:
+            # Wind parameters are model-agnostic, so keep them across model swaps.
+            self.wind.time = prev_wind.time
+            self.wind.period = prev_wind.period
+            self.wind.amplitude = prev_wind.amplitude
+            self.wind.frequency = prev_wind.frequency
+            self.wind.direction = prev_wind.direction
 
     def _build_packed_vbo_arrays(self):
         """Build write-index + output arrays for batched shape transform computation.
--- a/newton/tests/test_viewer_camera.py
+++ b/newton/tests/test_viewer_camera.py
@@ -5,8 +5,11 @@ import math
 import unittest
 
 import numpy as np
+import warp as wp
 
+import newton
 from newton._src.viewer.camera import Camera
+from newton.viewer import ViewerGL
 
 
 def _as_np(value):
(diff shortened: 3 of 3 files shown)


jcarius-nv:

han-xudong:
Updated this branch on top of the latest `main`, moved the changelog entry into the current top-level `[Unreleased] -> Fixed` section, and addressed the camera zoom/FOV preservation follow-up.

I reran:
- `uv run --extra dev -m unittest newton.tests.test_viewer_camera.TestViewerGLSetModelState.test_set_model_preserves_camera_and_wind_for_same_up_axis`
- `uv run --extra dev -m unittest newton.tests.test_viewer_camera`

newton/_src/viewer/viewer_gl.py
```diff
@@ -599,6 +602,20 @@ def _ensure_indices_wp(model_shapes) -> wp.array:
         fb_w, fb_h = self.renderer.window.get_framebuffer_size()
         self.camera = Camera(width=fb_w, height=fb_h, up_axis=model.up_axis if model else "Z")
 
+        if prev_camera is not None and model is not None and prev_camera.up_axis == self.camera.up_axis:
+            self.camera.fov = prev_camera.fov
+            self.camera.pos = self.camera._as_vec3(prev_camera.pos)
+            self.camera.yaw = prev_camera.yaw
+            self.camera.pitch = prev_camera.pitch
+            self.camera.set_pivot(prev_camera.pivot)
```
adenzler-nvidia: This restores only 5 of `Camera`'s 10 state fields — `near`/`far` are silently reset to defaults, and any field added to `Camera` later will be silently dropped on a model switch (no test would catch it). Since the only things that legitimately change across a compatible same-up-axis switch are the framebuffer size and the model, consider keeping the previous camera and just refreshing its size via the existing `Camera.update_screen_size()`:

```python
self.camera = Camera(width=fb_w, height=fb_h, up_axis=model.up_axis if model else "Z")
if prev_camera is not None and model is not None and prev_camera.up_axis == self.camera.up_axis:
    prev_camera.update_screen_size(fb_w, fb_h)
    self.camera = prev_camera
```

This preserves all view state and is robust to future fields. A one-line comment on the rationale would also match the commenting density of the rest of `set_model()`.

newton/_src/viewer/viewer_gl.py
```diff
@@ -599,6 +602,20 @@ def _ensure_indices_wp(model_shapes) -> wp.array:
...
+
+        if prev_wind is not None:
+            self.wind.time = prev_wind.time
+            self.wind.period = prev_wind.period
+            self.wind.amplitude = prev_wind.amplitude
+            self.wind.frequency = prev_wind.frequency
+            self.wind.direction = prev_wind.direction
```
adenzler-nvidia: Wind is preserved unconditionally, while the camera above is gated on `model is not None` and matching up-axis. So on an up-axis change the camera resets but wind carries over, and on `set_model(None)` wind state is copied onto a model-less `Wind`. If intentional, a short comment would help; otherwise gate wind on the same compatibility condition as the camera.

newton/_src/viewer/viewer_gl.py
```diff
@@ -530,6 +530,9 @@ def set_model(self, model: nt.Model | None):
         Args:
             model: The Newton model instance.
```
adenzler-nvidia, line 531: `set_model()` now has a meaningful side effect — preserving camera/wind across compatible (same up-axis) switches. Worth a one-line `Note:` in the docstring documenting the contract this PR's test asserts.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 - Fix MJCF `euler` producing wrong orientations for multi-component angles by treating angles as intrinsic rotations. (#3030)
 - Fix MJCF parsing so attributes from multiple `<compiler>` elements, including `<include>`-expanded children, are merged in document order. (#3030)
 - Fix MJCF worldbody static geoms bypassing the visual/collider class filter, so `parse_visuals=False` drops visual-class geoms attached directly to `<worldbody>` too. (#3030)
+- Fix `ViewerGL.set_model()` resetting headless/interactive camera and wind state when switching between models that use the same up-axis
```
adenzler-nvidia: Nit: the sibling `### Fixed` entries end with a period and a PR reference (e.g. `… document order. (#3030)`). Add a trailing period and `(#2658)` for consistency.

adenzler-nvidia (commented):
Functional behavior is correct — preserving interactive camera/wind across compatible model switches is the right fix, and the regression test fails without it for the right reason. The comments below are about robustness, convention adherence, and test coverage, not the core logic.

One item that can't be left inline: the commit message body uses literal `\n` escape sequences instead of real newlines, so the whole body renders as one unwrapped line. Please rewrap it at 72 columns with real line breaks.

newton/tests/test_viewer_camera.py
```diff
@@ -118,5 +121,51 @@ def test_dolly_moves_camera_toward_pivot_without_moving_pivot(self):
...
+    builder.add_shape_box(body=body, hx=0.5, hy=0.5, hz=0.5)
+    return builder.finalize()
+
+
+class TestViewerGLSetModelState(unittest.TestCase):
+    def test_set_model_preserves_camera_and_wind_for_same_up_axis(self):
+        model_a = _build_viewer_model(1.0)
+        model_b = _build_viewer_model(2.0)
+
+        viewer = ViewerGL(headless=True)
```
adenzler-nvidia: Two gaps here:

1. **No CUDA guard.** `ViewerGL(headless=True)` constructs on CPU, but `set_model()` then raises `RuntimeError: Failed to allocate … on device 'cpu'` from `_build_packed_vbo_arrays`, so this test hard-errors (rather than skips) on CPU-only/displayless lanes. The sibling viewer test guards with both `@unittest.skipUnless(wp.is_cuda_available(), …)` and a `try/except → self.skipTest(...)` (`test_viewer_image_logger_class.py`). Please add the same.
2. **Only the preserve branch is covered.** Both models are Z-up, so the `prev_camera.up_axis == self.camera.up_axis` guard is never exercised on its false path — deleting the guard would still pass. Add a second case with a Y-up model asserting the camera resets to its default (e.g. `fov == 45`).

han-xudong:
@adenzler-nvidia Thanks for review! Updated the branch for the latest review.

I switched the camera restoration to reuse the previous compatible `Camera` and refresh only its screen size, documented the camera/wind preservation contract, tightened wind preservation to non-`None` model switches, added CUDA/ViewerGL skips, added different-up-axis reset coverage, fixed the changelog entry, and rewrote the commit message body with real line breaks.

Reran:
- `uv run --extra dev -m unittest newton.tests.test_viewer_camera`

adenzler-nvidia (approved):
LGTM

PR #2662, opened 2026-04-30
Reduce the usage of pytorch

nvtw:
## Description

Replaces PyTorch with a minimal Warp-backed ONNX runtime where possible. Kamino is excluded and keeps using PyTorch for now. Avoids repeated 2.5GB+ PyTorch downloads when running examples that need neural network inference.

Adds `newton.utils.OnnxRuntime`, a graph-capturable ONNX inference engine implemented entirely with Warp kernels (no `onnxruntime` or `torch` runtime dependency). Migrates `ControllerNeuralMLP`, `ControllerNeuralLSTM`, and the robot policy examples (`example_robot_anymal_c_walk`, `example_robot_policy`, `example_mpm_anymal`) to load `.onnx` policies via the new runtime. Selection examples drop their unused torch code paths. A one-shot `scripts/convert_policies_to_onnx.py` is included to convert existing `.pt` policies in newton-assets to `.onnx`.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_examples.TestRobotExamples
uv run --extra dev -m newton.tests -k test_examples.TestRobotPolicyExamples
uv run --extra dev -m newton.tests -k test_examples.TestAdvancedRobotExamples
```

## Summary by CodeRabbit

* **New Features**
  * OnnxRuntime: run ONNX policies on-device; neural controllers and examples now support ONNX models.

* **Improvements**
  * Replaced PyTorch inference with on-device ONNX+Warp execution across controllers, examples, and tests for lower host/device sync and improved CUDA-graph capture.
  * Examples now use persistent device buffers for observations/actions and on-device kernels for preprocessing and action assembly.

* **Tests**
  * Tests updated to use ONNX fixtures and gate execution on ONNX availability.

* **Chores**
  * Added optional onnx extra, docs/changelog updates, and deprecated legacy TorchScript checkpoint loading.


newton/examples/assets/rl_policies/anymal_walking_policy_physx.onnx
eric-heiden, line 1: Can we move all these policy files to newton-assets, including the torch checkpoints? This can be done in a follow-up, but I don't see a reason why we have the USD assets in the assets repo but the policy files here in the main repo.

scripts/convert_policies_to_onnx.py
eric-heiden, line 1: I don't see why we need this file, it should be removed?

newton/_src/utils/onnx_runtime.py
```diff
@@ -0,0 +1,602 @@
...
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Any
+
+import numpy as np
+import onnx
+import warp as wp
+from onnx import numpy_helper
```
eric-heiden: This makes `onnx` now a required dependency for all of Newton which is not what we want. Please limit the import to only where `onnx` is needed. See also comment below on pyproject.toml dependency.

newton/_src/actuators/utils.py
```diff
@@ -6,69 +6,59 @@
...
-        pass
-
-    checkpoint = torch.load(path, map_location="cpu", weights_only=False)
-    if isinstance(checkpoint, dict):
-        meta = checkpoint.get("metadata", {})
-        if "model" not in checkpoint:
-            raise ValueError(f"Cannot load checkpoint at '{path}'; dict checkpoint has no 'model' key")
-        return checkpoint["model"], meta
-
-    raise ValueError(f"Cannot load checkpoint at '{path}'; expected a TorchScript archive or a dict with a 'model' key")
```
eric-heiden, line 36: 🔴 **Neural controllers drop .pt checkpoint compatibility**

The shared neural-controller loader now only calls `onnx.load()` / `OnnxRuntime`, while the previous implementation accepted TorchScript `.pt` archives and dict checkpoints. `ControllerNeuralMLP` and `ControllerNeuralLSTM` are public via `newton.actuators`, so existing user code that passes a `.pt` `model_path` will fail immediately without a prior deprecation release. The changelog provides migration guidance, but the project API convention says breaking changes require deprecation first.

**Suggested change:**
Keep the old TorchScript/dict loading path behind an optional `torch` import and emit a deprecation warning for `.pt` checkpoints, then remove it in a later release; or introduce a new ONNX-only controller/API while preserving the existing behavior.

pyproject.toml
```diff
@@ -83,14 +84,16 @@ examples = [
     "Pillow>=11.3.0",
 ]
 
-# Extra dependency needed to run examples that inference an RL policy
+# PyTorch is no longer required to run RL policies (examples and actuators
+# load ONNX files via Newton's Warp-backed runtime).  These extras remain only
```
eric-heiden, line 95: 🟡 **ONNX is added as a core dependency**

This adds `onnx>=1.16.0` to the base `dependencies`, so every Newton install now pulls ONNX plus transitive packages such as protobuf and ml-dtypes even if the user never loads policy networks. The public import in `newton/utils.py` also imports `OnnxRuntime` eagerly, so this is not isolated to examples or actuators. AGENTS.md says: "Avoid new required dependencies. Strongly prefer not adding optional ones — use Warp, NumPy, or stdlib."

**Suggested change:**
Make ONNX an optional extra used by the policy runtime/examples, and lazy-import `onnx` inside `OnnxRuntime` / `load_metadata()` with a clear install error when that optional feature is used.

newton/_src/actuators/controllers/controller_neural_lstm.py
```diff
@@ -3,42 +3,91 @@
...
-    error and velocity error. The network maintains hidden and cell state
-    across timesteps to capture temporal patterns.
+    Uses a pre-trained LSTM (loaded from an ``.onnx`` file) to compute joint
+    effort from position error and velocity error.  Hidden and cell state are
+    maintained across timesteps to capture temporal patterns.
 
-    The network must be callable as::
+    The exported ONNX model must have **three inputs** -- input,
+    initial_hidden, initial_cell -- and **two outputs** -- effort, hidden_out
+    (cell_out is read from the LSTM op directly).  The model's metadata
```
eric-heiden: 🟡 **LSTM docs say two outputs but compute requires three**

The class docstring says the exported ONNX model has "two outputs -- effort, hidden_out" and that `cell_out` is read from the LSTM op directly, but the required metadata includes `cell_out_name` and `compute()` reads `out[self._cell_out_name]`. `OnnxRuntime.__call__()` returns only declared graph outputs, so a model exported according to the two-output contract will raise a `KeyError` when the controller asks for `cell_out_name`.

**Suggested change:**
Either require `cell_out` as a third graph output in the docstring and migration instructions, or change `OnnxRuntime`/the controller to expose the internal LSTM cell output without requiring it as a graph output.

newton/_src/actuators/controllers/controller_neural_mlp.py
```diff
@@ -98,22 +207,47 @@ def __init__(self, model_path: str):
...
+        self._net_output_name = runtime.output_names[0]
+
+        # The compute() copy assumes one effort per actuator, i.e. output shape
+        # exactly ``(num_actuators, 1)``.  A multi-column output would silently
+        # misalign actuator i with row 0, column i, dropping later rows; reject
+        # such models up front instead of producing wrong inferences.
+        out_shape = runtime._shapes[self._net_output_name]
+        if out_shape != (num_actuators, 1):
+            raise ValueError(
+                f"ControllerNeuralMLP: ONNX output '{self._net_output_name}' has shape {out_shape}, "
```
eric-heiden: 🟡 **MLP neural actuators still report non-graphable**

The MLP controller was moved off Torch onto Warp kernels plus `OnnxRuntime`, and its history update is now a Warp kernel, but `is_graphable()` still returns `False`. `Actuator.is_graphable()` delegates directly to this method, so MLP neural actuators remain excluded from CUDA graph capture while the sibling LSTM controller now returns `True`. That looks like a silent missed benefit of this PR unless there is still an MLP-specific capture blocker.

**Suggested change:**
Return `True` for `ControllerNeuralMLP.is_graphable()` if the new path is capture-safe, or leave it false with an explicit comment/test documenting the remaining blocker.

newton/examples/robot/example_robot_anymal_c_walk.py
```diff
@@ -247,36 +234,35 @@ def simulate(self):
...
         )
-        with torch.no_grad():
-            self.act = self.policy(obs)
-            self.rearranged_act = torch.gather(self.act, 1, self.mujoco_to_lab_indices.unsqueeze(0))
-            a = self.joint_pos_initial + 0.5 * self.rearranged_act
-            a_with_zeros = torch.cat([torch.zeros(6, device=self.torch_device, dtype=torch.float32), a.squeeze(0)])
-            a_wp = wp.from_torch(a_with_zeros, dtype=wp.float32, requires_grad=False)
-            wp.copy(
-                self.control.joint_target_pos, a_wp
-            )  # this can actually be optimized by doing  wp.copy(self.solver.mjw_data.ctrl[0], a_wp) and not launching  apply_mjc_control_kernel each step. Typically we update position and velocity targets at the rate of the outer control loop.
```
eric-heiden, line 249: 🟡 **Policy examples force GPU-host-GPU copies each step**

The hot control loop now builds a fresh device input from NumPy, immediately copies the ONNX output back to host with `.numpy()`, then uploads a new Warp array for `joint_target_pos`. The same pattern appears in `example_robot_policy.py` and `example_mpm_anymal.py`. This undercuts the new runtime's "without host round-trips or device allocation" contract for the CUDA examples and adds a synchronization point every policy step.

**Suggested change:**
Keep the policy action and target buffers on device: preallocate input/action/target arrays and use a small Warp kernel to reorder, scale, prepend the six zeros, and write `control.joint_target_pos` directly.

scripts/convert_policies_to_onnx.py
```diff
@@ -0,0 +1,106 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
+
+"""One-shot converter: TorchScript ``.pt`` policies in newton-assets -> ONNX.
+
+This script is intended to be run **once** (e.g. by Newton maintainers) so the
+converted ``.onnx`` files can be committed to the ``newton-assets`` repository.
```
eric-heiden: 🟡 **Conversion script does not embed ONNX metadata — LSTM models unusable**

The conversion script uses `torch.onnx.export()` which produces `.onnx` files with no custom metadata properties. However, `ControllerNeuralLSTM.__init__` (controller_neural_lstm.py:149–157) **requires** eight metadata keys:

```python
for key in (
    "input_name", "hidden_in_name", "cell_in_name",
    "output_name", "hidden_out_name", "cell_out_name",
    "num_layers", "hidden_size",
):
    if key not in metadata:
        raise ValueError(f"ONNX metadata missing required key '{key}'")
```

The script only handles MLP-style single-input/single-output policies. If a maintainer tries to convert an LSTM `.pt` policy with this script, the resulting `.onnx` would load into `OnnxRuntime` but `ControllerNeuralLSTM` would reject it at construction time with a missing-key `ValueError`.

The CHANGELOG migration guidance ("run `scripts/convert_policies_to_onnx.py` once") implies this script handles all policy types. The bundled `.onnx` LSTM files in `newton/examples/assets/rl_policies/` were presumably hand-crafted or converted separately, but the script doesn't document this gap.

**Suggested change:**
Either (a) extend the script to detect LSTM models (they return tuples) and write the required ONNX metadata props, or (b) add a clear note in the script docstring and CHANGELOG that LSTM conversion requires a separate manual step, and provide a helper function or snippet.

newton/_src/utils/onnx_runtime.py
```diff
@@ -0,0 +1,602 @@
...
+    shapes[op.outputs[0]] = (M, N)
+
+
+def _exec_elu(op, tensors, shapes, device):
+    x = tensors[op.inputs[0]]
+    alpha = float(op.attrs.get("alpha", 1.0))
+    out = tensors[op.outputs[0]]
+    shape = shapes[op.inputs[0]]
+    wp.launch(_elu_kernel, dim=shape, inputs=[x, out, alpha], device=device)
+    shapes[op.outputs[0]] = shape
```
eric-heiden: ⚪ **Redundant `shapes` dict writes on every forward call**

In `_exec_gemm` (line 507) and `_exec_elu` (line 517), the shapes dict is overwritten on every forward call:

```python
shapes[op.outputs[0]] = (M, N)    # _exec_gemm
shapes[op.outputs[0]] = shape     # _exec_elu
```

These shapes are already set during `_preallocate_buffers()` and never change (batch size is fixed). These are Python dict writes in the hot path that serve no purpose after the first call. For the current 4-op policy networks this is negligible, but it's unnecessary host-side work that accumulates with deeper models.

**Suggested change:**
Could remove the shapes updates from the exec functions since they're already set during preallocation and the batch size is fixed.

newton/examples/assets/rl_policies/anymal_walking_policy_physx.onnx
nvtw: I'll move them in a follow up MR. Makes sense to have them in the asset repo. For a first version, it was easier to have them directly in the Newton repo which is not to problematic as these files turned out to be really small.

nvtw:
@eric-heiden most of your comments should be addressed now. The onnx files will move to the newton assets repo in a later MR

pyproject.toml
shi-eric, line 224: Do you think onnx should be added to this list?

newton/_src/actuators/controllers/controller_neural_mlp.py
```diff
@@ -98,35 +207,58 @@ def __init__(self, model_path: str):
...
+        runtime, _ = load_checkpoint(self.model_path, device=str(device), batch_size=num_actuators)
+        self._network = runtime
+        self._net_input_name = runtime.input_names[0]
+        self._net_output_name = runtime.output_names[0]
+
+        # The compute() copy assumes one effort per actuator, i.e. output shape
+        # exactly ``(num_actuators, 1)``.  A multi-column output would silently
+        # misalign actuator i with row 0, column i, dropping later rows; reject
+        # such models up front instead of producing wrong inferences.
+        out_shape = runtime._shapes[self._net_output_name]
```
shi-eric, line 258: Context: drafted with GPT-5.5 assistance and verified against PR head `25b97c1`.

Legacy MLP `.pt` / `.pth` checkpoints still fail during `finalize()`.

For torch checkpoints, `load_checkpoint()` returns `_TorchModuleAdapter`, whose `output_names` is `["action"]` but whose `_shapes` dict is empty until the first `__call__()` populates it. This line reads `runtime._shapes[self._net_output_name]` before any inference has happened, so `ControllerNeuralMLP("legacy.pt").finalize(...)` raises `KeyError("action")`.

That means the deprecated loader exists, but existing MLP torch checkpoints still do not have a working deprecation window. Please avoid this ONNX-only shape lookup for `_TorchModuleAdapter`, or add an adapter method that can validate or infer the legacy output shape without requiring `_shapes` to be populated during `finalize()`.

shi-eric (commented):
Adding one verified compatibility finding for legacy MLP checkpoints.

newton/_src/actuators/utils.py
shi-eric: Context: drafted with GPT-5.5 assistance and verified against PR head `25b97c1`.

One concrete compatibility gap remains here for LSTM checkpoints: `ControllerNeuralLSTM.__init__()` now rejects every `.pt` / `.pth` path immediately with `NotImplementedError`.

That still makes the deprecation effectively a hard removal for existing LSTM torch checkpoints in this release. If the intended policy is one release of compatibility, the LSTM path needs to keep the legacy torch-backed behavior with a deprecation warning, or this hard failure should be deferred until after the deprecation window.

newton/_src/actuators/controllers/controller_neural_mlp.py
nvtw: Thanks for catching this. It should be fixed now

pyproject.toml
nvtw: Added onnx to this list

newton/examples/assets/rl_policies/anymal_walking_policy_physx.onnx
shi-eric: I'm still concerned about the irreversible growth of the size of the Newton repo due to the addition of >1 MiB of files (which is compressed). It doesn't seem like much, but can you calculate the growth of the repo size relative to the current size as a result of this change so we understand the impact for something that will be moved to assets?

newton/tests/test_actuators.py
```diff
@@ -48,26 +48,121 @@
...
+_HAS_ONNX = importlib.util.find_spec("onnx") is not None
 _HAS_TORCH = importlib.util.find_spec("torch") is not None
 
-if _HAS_TORCH:
-    import torch as _torch
+if _HAS_ONNX:
+    import onnx as _onnx
+    from onnx import TensorProto as _TensorProto
+    from onnx import helper as _onnx_helper
+    from onnx import numpy_helper as _onnx_np
```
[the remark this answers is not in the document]
nvtw: Should be addressed now

newton/_src/utils/onnx_runtime.py
```diff
@@ -0,0 +1,618 @@
...
+    # The kernel hardcodes ONNX-default activations (sigmoid, tanh, tanh) and
+    # has no peephole / clip / input_forget / sequence_lens support.  Reject
+    # any model that asks for non-default behavior so we never silently
+    # produce wrong inferences.
+    for unsupported in ("activations", "activation_alpha", "activation_beta", "clip", "input_forget"):
+        if unsupported in op.attr_names:
+            raise NotImplementedError(
+                f"OnnxRuntime LSTM: attribute '{unsupported}' is not supported "
+                f"(only default sigmoid/tanh/tanh activations, no clip, no input_forget)"
+            )
```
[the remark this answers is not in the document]
nvtw: Should be addressed now

newton/_src/actuators/controllers/controller_neural_mlp.py
```diff
@@ -98,35 +207,66 @@ def __init__(self, model_path: str):
...
+                f"ControllerNeuralMLP: network output '{self._net_output_name}' has shape {out_shape}, "
+                f"expected {(num_actuators, 1)} (one scalar effort per actuator)"
+            )
 
     def is_stateful(self) -> bool:
         return True
 
     def is_graphable(self) -> bool:
-        return False
+        return True
```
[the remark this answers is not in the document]
nvtw: Should be addressed now

newton/examples/assets/rl_policies/anymal_walking_policy_physx.onnx
eric-heiden: I also think it would be better we first add these onnx files to newton-assets and then merge this PR rather than adding them here and then later moving them out again.

eric-heiden: The torch checkpoints can be moved in a follow-up.

nvtw: Ok, I can move the onnx files now. But honestly their size is not terrible, the largest one is 358KB

nvtw: https://github.com/newton-physics/newton-assets/pull/41

shi-eric: Merged

nvtw: The onnx files were removed from this MR

shi-eric:
GPT-5.5 spotted one possible migration edge case here. Please feel free to dismiss if it seems irrelevant or out of scope for this PR.

For MLP controllers, a common ONNX migration path is:

```python
dummy = torch.zeros(1, feature_count)
torch.onnx.export(model, dummy, "policy.onnx", opset_version=17)
```

That usually exports a fixed leading batch dimension of `1`. If the controller later finalizes with `num_actuators > 1`, `ControllerNeuralMLP` passes that actuator count as `batch_size` and allocates `_net_input` as `(num_actuators, feature_count)`, but `OnnxRuntime` currently preserves the exported fixed `dim_value=1`. Shape inference then keeps the output as `(1, 1)`, so finalize rejects the model even though it is conceptually a valid one-scalar-per-actuator policy.

A possible fix would be to let the controller/runtime explicitly identify which graph-input axis is the runtime batch axis and replace only that axis with the caller-provided `batch_size`. That matters because MLP inputs are shaped like `(N, features)`, while LSTM inputs here are shaped like `(1, N, 2)` and recurrent state inputs like `(num_layers, N, hidden_size)`, so blindly overriding axis 0 in `OnnxRuntime` would be wrong.

If this is worth handling now, I’d suggest a focused regression test with an ONNX MLP exported/built with fixed input shape `(1, features)` and output `(1, 1)`, then finalized through `ControllerNeuralMLP` with `num_actuators > 1`.

nvtw:
> GPT-5.5 spotted one possible migration edge case here. Please feel free to dismiss if it seems irrelevant or out of scope for this PR.
> 
> For MLP controllers, a common ONNX migration path is:
> 
> ```python
> dummy = torch.zeros(1, feature_count)
> torch.onnx.export(model, dummy, "policy.onnx", opset_version=17)
> ```
> 
> That usually exports a fixed leading batch dimension of `1`. If the controller later finalizes with `num_actuators > 1`, `ControllerNeuralMLP` passes that actuator count as `batch_size` and allocates `_net_input` as `(num_actuators, feature_count)`, but `OnnxRuntime` currently preserves the exported fixed `dim_value=1`. Shape inference then keeps the output as `(1, 1)`, so finalize rejects the model even though it is conceptually a valid one-scalar-per-actuator policy.
> 
> A possible fix would be to let the controller/runtime explicitly identify which graph-input axis is the runtime batch axis and replace only that axis with the caller-provided `batch_size`. That matters because MLP inputs are shaped like `(N, features)`, while LSTM inputs here are shaped like `(1, N, 2)` and recurrent state inputs like `(num_layers, N, hidden_size)`, so blindly overriding axis 0 in `OnnxRuntime` would be wrong.
> 
> If this is worth handling now, I’d suggest a focused regression test with an ONNX MLP exported/built with fixed input shape `(1, features)` and output `(1, 1)`, then finalized through `ControllerNeuralMLP` with `num_actuators > 1`.

Done

shi-eric (dismissed):

eric-heiden (dismissed):

adenzler-nvidia (approved):

PR #2670, opened 2026-05-01
VBD cable: Match capsule origin and COM

### LINKED ISSUE #2484: VBD cable: match capsule origin and COM
Currently, capsule origins do not match COM as a default design, but it causes confusion for users/agents.
Match origin and COM for easier tuning.

### LINKED ISSUE #2832: Assign uniform color to cable segments by default.
Assign uniform color to cable segments by default, otherwise, cable has too colorful visualization.

### LINKED ISSUE #2833: Improve mouse picking for cable examples.
Mouse dragging often introduces too large forces into cable system, possibly due to large torque applied to cable segments (capsules).

jumyungc:
## Description

1. Match cable segment origins to COMs
Generated cable capsules previously used the segment start point as the body origin while storing the COM at the segment midpoint. This was easy to misuse because many rigid-body paths expect body origin and COM to coincide.
Cable capsule body origins now sit at the segment midpoint/COM. This changes the local endpoint-anchor convention for custom joints attached to rod segments, but makes generated cable bodies consistent with normal rigid-body expectations.

2. Use uniform cable segment colors
Cable segments previously received per-shape palette colors, making each cable appear as a multicolor chain. Generated rods now use one visual color for all capsule segments in a cable by default. Callers can still pass color= to choose an explicit cable color.

3. Stabilize mouse dragging for cable examples
Mouse dragging cable capsules can apply large off-center torques, which can destabilize long, constrained cable systems. Cable examples now keep viewer drag translation-only for cable bodies by removing mouse-picking torque from the actively picked cable segment while preserving linear drag force.

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **Bug Fixes**
  * Cable segment bodies now use mid-segment origins, fixing geometric/anchor mismatches and attachment correctness.

* **New Features**
  * Segments get a default visual color (overridable).
  * New translation-only mouse-picking for cables to avoid torque during dragging; examples now use this.

* **Behavior Changes**
  * Cable joint anchor conventions updated to midpoint-relative.
  * Solver default descriptions broadened regarding hard-constraint coverage.

* **Tests**
  * Updated tests and added regression verifying segment origin/COM alignment.

* **Documentation**
  * Changelog updated to reflect defaults and migration notes.


newton/_src/sim/builder.py
```diff
@@ -185,6 +185,11 @@ class ModelBuilder:
         (238, 153, 51),  # orange
         (0, 153, 136),  # teal
     )
+    _DEFAULT_ROD_COLOR = (
```
eric-heiden, line 190: Maybe it would be better if rods would also go through the color palette here so by default each rod gets a new color from `_SHAPE_COLOR_PALETTE`? Not critical, it might just be more consistent with how shapes are handled.

newton/examples/cable/_mouse_picking.py
eric-heiden, line 1: Can this be integrated into viewer/picking.py directly?

CHANGELOG.md
```diff
@@ -74,8 +74,11 @@
...
+- Change the cable rod body-frame convention generated by `ModelBuilder.add_rod()` and `add_rod_graph()` so each capsule body origin coincides with its capsule COM. Migration: update custom rod endpoint anchors to use midpoint-relative coordinates.
+- Assign one default visual color to capsule segments generated by `ModelBuilder.add_rod()` or `add_rod_graph()`; pass `color=` to choose an explicit cable color.
+- Improve mouse picking in cable examples by applying viewer drag as translation-only forces on cable capsule bodies, avoiding torque-induced instability while preserving linear dragging.
+- VBD solver uses augmented-Lagrangian hard constraints for body-body contacts and non-cable structural joint slots by default; cable joints remain stiffness-driven. Pass `rigid_contact_hard=False` to `SolverVBD(...)` to restore the previous soft-contact behavior.
```
eric-heiden: There are some unrelated changes here where entries are rewritten that are not part of this PR.

newton/_src/sim/builder.py
jumyungc: At first, I did that, but it gave me eye strain when I looked at the cable pile example, so I changed it again.
I am open to giving different colors, if it is a better way.

newton/examples/cable/_mouse_picking.py
jumyungc: Yes, that is cleaner. I will move it to picking.py (assuming no harm :-))

CHANGELOG.md
jumyungc: I added some migration guidance for recent changes that should have been included in the corresponding PRs. Please let me know if you’d prefer me to revert it.

jumyungc: Reverted, and Cleanly added to new [unreleased]

jumyungc:
I changed this PR to draft because I was not sure whether we should switch the rod body origin to the COM by default, as discussed in Slack.

To be safe, I added an explicit origin option and kept the default as the existing start-node convention, although I think COM is the better long-term convention.

Edited:
After checking the code one more time, I’ll mark it as ready for review.

jumyungc:
Thank you for the review. Please let me know if there’s anything else that need to be changed. I’m also open to adjusting the color convention if needed.

jcarius-nv:

newton/_src/sim/builder.py
eric-heiden: Fine, let's keep this then

eric-heiden (dismissed):
Thanks, LGTM!

adenzler-nvidia (dismissed):
Approving. Deprecation-first handling of the body-frame-origin change is the right call — the legacy `"start"` convention is preserved with a `DeprecationWarning` rather than flipped now, and the only user-visible behavior change (uniform default color) plus the deprecation are both documented in the changelog with migration guidance. No undocumented breaking changes; the internal `apply_picking_force_kernel` signature change has its sole caller updated. Test coverage for both origin conventions and the linear-only picking path is solid.

jumyungc:
Another approval please.

eric-heiden (approved):

PR #2673, opened 2026-05-01
Expose render_megakernel launch args in SensorTiledCamera

rshahid-nvidia:
## Description

Adds `kernel_block_dim` to `SensorTiledCamera.update()` and forwards it through `RenderContext.render()` to the Warp ray-tracer `render_megakernel` launch:

- `kernel_block_dim: int = 64` is forwarded to `wp.launch(..., block_dim=...)`.
- The override is launch-time only; `RenderConfig` and the existing render-kernel cache key are unchanged.

## Checklist

- [x] Existing tests cover these changes

## Test plan

Existing tiled camera sensor tests

## New feature / API change

Callers can tune the render megakernel block dim per update call:

```python
sensor.update(
    state,
    camera_transforms,
    camera_rays,
    color_image=color_image,
    kernel_block_dim=64,
)
```

## Summary by CodeRabbit

* **New Features**
  * Per-render kernel launch tuning added (block/thread and max-blocks controls) with defaults preserving prior behavior; launch config now affects kernel cache keys.
  * New KernelLaunchConfig type introduced and re-exported via the public utilities API.

* **Tests**
  * Added unit tests covering defaults, explicit launch configs, equality/hashing, and cache-key effects.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2673)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - Add linear HDR color output support to `SensorTiledCamera` via `hdr_color_image`.
+- Add an optional `kernel_block_dim` argument to `SensorTiledCamera.update()` for tuning the Warp ray-tracer's `render_megakernel` launch shape.
 - Add composable actuator subsystem with pluggable `Controller` (`ControllerPD`, `ControllerPID`, `ControllerNeuralMLP`, `ControllerNeuralLSTM`), `Clamping` (`ClampingMaxEffort`, `ClampingDCMotor`, `ClampingPositionBased`), and `Delay` components; supports per-DOF delays, CUDA graph capture, and masked environment reset
 - Add heatmap rendering for scalar arrays logged through `ViewerGL.log_array()`
 - Add Blender-style orbit, pan, and dolly controls to the GL viewer using middle-mouse drag combinations
--- a/newton/_src/sensors/sensor_tiled_camera.py
+++ b/newton/_src/sensors/sensor_tiled_camera.py
@@ -207,6 +207,7 @@ class SensorTiledCamera(metaclass=_SensorTiledCameraMeta):
         clear_data: ClearData | None = DEFAULT_CLEAR_DATA,
         refit_bvh: bool | None = None,
         hdr_color_image: wp.array4d[wp.vec3f] | None = None,
+        kernel_block_dim: int = 64,
     ):
         """Render output images for all worlds and cameras.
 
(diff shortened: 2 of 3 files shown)


StafaH:
@rshahid-nvidia This is great! Design-wise I don't think injecting this to be renderer specific is the way to go, modifying block_dim and launch_bounds should be something global to newton, and the configs related to launch should be collected elsewhere, but I'll defer to the newton team's design choice here.

newton/_src/sensors/warp_raytrace/types.py
```diff
@@ -81,6 +85,9 @@ class RenderConfig:
     gaussians_max_num_hits: int = 20
     """Maximum Gaussian hits accumulated per ray."""
 
+    render_kernel_launch_config: KernelLaunchConfig = field(default_factory=KernelLaunchConfig)
```
daniela-hase: Putting the `render_kernel_launch_config` here in the `RenderConfig` class forces a kernel-recompilation, while I guess it's no big deal - I wonder if it'd be better to move it elsewhere?

Maybe just have it as an arg for the `render` method of the RenderContext?

rshahid-nvidia: `KernelLaunchConfig` also contains `launch_bounds`, which probably should also trigger kernel-recompilation. Whereas, the `block_dim` and `max_blocks` need not. Maybe we could split those into `KernelCompileConfig` and `KernelLaunchConfig`?

daniela-hase: Ah yeah, you are right! :)
Maybe we put the `launch_bounds` into the `RenderConfig` and the `block_dim`/`max_blocks` as an arg for the `render()` method?

I personally really like how it is all in one place right now, but I think anyone testing around with the `block_dim` would really appreciate not having to recompile the kernel all the time 🙃

rshahid-nvidia: I am tempted to only expose `block_dim`. That was the only param that actually showed measurable gains on changing. I added `launch_bounds` and `max_blocks` for more configurability, which I am not sure is good design wise. I'll clean this up and only expose `block_dim` through `SensorTiledCamera.update` and `RenderContext.render`.

rshahid-nvidia: Updated this.

daniela-hase: Ah nice, yeah I like this change :)

daniela-hase (approved):

PR #2676, opened 2026-05-03
Fix viewer particle active constant codegen

FahYen:
## Summary

Convert the viewer particle active flag constant through a plain Python integer before wrapping it as a Warp constant.

## Why

With warp-lang 1.12.1, the viewer kernel can emit ParticleFlags.ACTIVE into generated CUDA for build_active_particle_mask. NVRTC then fails compilation because ParticleFlags is not defined in CUDA.

Converting the IntEnum value to int first forces Warp to see a numeric constant and generate a valid literal.

## Validation

- Reproduced locally with python -m newton.examples basic_pendulum on an AWS DCV X display before the fix.
- Verified the patched checkout imports newton._src.viewer.kernels and resolves PARTICLE_ACTIVE to 1.
- Attempted pytest collection locally, but the environment failed during unrelated Kamino test collection imports.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed type conversion issue for particle flag handling to ensure proper constant initialization and improve stability.

### DIFF
--- a/newton/_src/viewer/kernels.py
+++ b/newton/_src/viewer/kernels.py
@@ -784,7 +784,7 @@ def compute_hydro_contact_surface_lines(
     line_colors[tid * 3 + 2] = color
 
 
-PARTICLE_ACTIVE = wp.constant(wp.int32(newton.ParticleFlags.ACTIVE))
+PARTICLE_ACTIVE = wp.constant(wp.int32(int(newton.ParticleFlags.ACTIVE)))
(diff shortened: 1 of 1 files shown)


eric-heiden:
Thanks, but this is not a real codegen bug, we have test function `test_viewer_particle_flags` running successfully on the Warp version 1.13.0rc1 that we require in pyproject.toml.

Issue #2668, opened 2026-05-01
[BUG] Support for USD mesh plane


Milad-Rakhsha-NV:
### Bug Description

MuJoCo solver construction fails on USD assets that contain a colliding mesh plane. This is a common pattern in USD assets: a floor is authored as a finite quad mesh rather than as a MuJoCo/Newton primitive plane.

[example_robot_usd.py](https://github.com/user-attachments/files/27265521/example_robot_usd.py)

Original repro:

```bash
uv run -m newton.examples robot_usd --usd ~/Downloads/Collected_H1/H1.usd
```

Additional local repro from `example_robot_h1_milad.py` with the planar-mesh conversion/proxy disabled:

```bash
uv run --extra examples python newton/examples/robot/example_robot_h1_milad.py \
  --viewer null --num-frames 1 --no-use-mujoco-contacts --quiet
```

This fails during `SolverMuJoCo` construction, before simulation starts.

Shortened console output:

```text
Warning: Non-uniform scaling of spheres is not supported.
Warning: The rigid body at /Environment/h1_ROS/d435_left_imager_link has a possibly invalid inertia tensor ...
Warning: The rigid body at /Environment/h1_ROS/d435_rgb_module_link has a possibly invalid inertia tensor ...
Warning: The rigid body at /Environment/h1_ROS/mid360_link has a possibly invalid inertia tensor ...

QH6154 Qhull precision error: Initial simplex is flat (facet 1 is coplanar with the interior point)
While executing:  | qhull Qt TA60
The input to qhull appears to be less than 3 dimensional
Qhull could not construct a clearly convex simplex from points:
- p3(v4):     5     5     0
- p2(v3):    -5     5     0
- p1(v2):     5    -5     0
- p0(v1):    -5    -5     0
The min and max coordinates for each dimension are:
  0:        -5         5  difference=   10
  1:        -5         5  difference=   10
  2:         0         0  difference=    0

Traceback (most recent call last):
  File "newton/examples/robot/example_robot_h1_milad.py", line 154, in __init__
    self.solver = newton.solvers.SolverMuJoCo(...)
  File "newton/_src/solvers/mujoco/solver_mujoco.py", line 3082, in __init__
    self._convert_to_mjc(...)
  File "newton/_src/solvers/mujoco/solver_mujoco.py", line 5255, in _convert_to_mjc
    self.mj_model = spec.compile()
ValueError: Error: qhull error
Element name '/World/Plane_47', id 0
```

### Current Understanding

[...]

PR #2681, opened 2026-05-04
Parse ground meshes into ground planes during usd loading

nvtw:
## Description

Fix for https://github.com/newton-physics/newton/issues/2668

`ModelBuilder.add_usd` failed to load static ground planes authored as a coplanar `UsdGeomMesh` with `PhysicsCollisionAPI` and `physics:approximation = "none"` (the standard Isaac Sim / PhysX ground-plane pattern). The mesh was added as `GeoType.MESH`, then `SolverMuJoCo` crashed with a Qhull error while building a degenerate convex hull from the flat quad.

The importer now detects axis-aligned coplanar `Mesh` colliders without a rigid body parent and routes them to `add_shape_plane`, with axis correction when the plane normal is along local X or Y. Coplanar meshes attached to a rigid body remain `GeoType.MESH` to preserve inertia accumulation and finite-plane semantics. `physics:approximation = "none"` is also recognised as a valid no-op, removing the spurious "Unknown" warning.

## Checklist

## Test plan

```
uv run --extra dev python -m unittest newton.tests.test_import_usd
uv run --extra dev python -m unittest newton.tests.test_import_usd.TestImportUsdPhysics.test_planar_mesh_collider_imported_as_plane
```

The new regression test covers: Z-coplanar static quad → `PLANE`; Y-coplanar static quad → `PLANE` with axis correction; coplanar quad attached to a rigid body → stays `MESH`; volumetric mesh with `approximation="none"` → stays `MESH`. End-to-end verified by importing the H1 USDA from issue #2668 and constructing `SolverMuJoCo` without error.

## Bug fix

**Steps to reproduce:**

1. Import a USD stage containing a static ground authored as `def Mesh "Plane"` with `PhysicsCollisionAPI`, `PhysicsMeshCollisionAPI`, and `physics:approximation = "none"` (e.g. an Isaac Sim H1 scene).
2. Call `ModelBuilder.add_usd(...)`, finalize, and construct `SolverMuJoCo`.
3. Observe `ValueError: Error: qhull error` from `spec.add_mesh` because the 4-vertex coplanar quad has zero Z-extent.

**Minimal reproduction:**

```python
import warp as wp
import newton
import newton.solvers

builder.add_usd("H1_flattened.usda", xform=wp.transform(wp.vec3(0, 0, 0)))
model = builder.finalize()
newton.solvers.SolverMuJoCo(model)  # ValueError: Error: qhull error (before this PR)
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Static coplanar ground planes are now imported as plane geometry instead of mesh, preventing failures in volumetric solvers. Coplanar meshes attached to rigid bodies remain unchanged.

* **Tests**
  * Added regression test for planar mesh collider import behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,7 @@
 - Fix `State.assign` not copying namespaced extended and custom state attributes 
 - Fix mesh-convex back-face contacts generating inverted normals that trap shapes inside meshes and cause solver divergence (NaN)
 - Fix finite plane geometry 2x too large in collision, bounding sphere, and raytrace sensor
+- Fix `ModelBuilder.add_usd` failing to load static ground planes authored as a coplanar `UsdGeomMesh` with `PhysicsCollisionAPI` (the common Isaac Sim / PhysX `physics:approximation = "none"` ground-plane pattern). Such meshes are now imported as `GeoType.PLANE`, so volumetric solvers like `SolverMuJoCo` no longer fail with a Qhull error when building a degenerate convex hull. Coplanar meshes attached to a rigid body are unaffected and remain `GeoType.MESH`
 - Fix MPR convergence failure on large and extreme-aspect-ratio mesh triangles by projecting the starting point onto the triangle nearest the convex center
 - Fix O(W²·S²) memory explosion in `CollisionPipeline` shape-pair buffer allocation for NXN and SAP broad phase modes by computing per-world pair counts instead of a global N²
 - Fix `SensorRaycast` ignoring `PLANE` geometry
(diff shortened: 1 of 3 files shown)


eric-heiden (changes requested):
Let's wait if this conversion is really needed for Isaac Sim integration. It is not needed when the Newton collision pipeline is used in SolverMuJoCo, maybe this should rather be done in the Isaac Sim backend?

nvtw:
> Let's wait if this conversion is really needed for Isaac Sim integration. It is not needed when the Newton collision pipeline is used in SolverMuJoCo, maybe this should rather be done in the Isaac Sim backend?

I think you are right. I'll close the MR. We can reopen later if needed.

PR #2688, opened 2026-05-04
Fix collapsed URDF articulation starts (#2589)

### LINKED ISSUE #2589: ModelBuilder.finalize() crashes with 3+ articulations

vreutskyy:
## Description

Closes #2589.

`ModelBuilder.finalize()` crashes with `ValueError: articulation_start is not monotonically increasing` when 3 or more URDFs (e.g. ANYmal C) are imported into the same builder with `collapse_fixed_joints=True`.

The root cause is in `ModelBuilder.collapse_fixed_joints()`:

- After remapping joint indices, `self.articulation_start = list(set(...))` was used to drop empty articulations. `set()` is unordered, so for 3+ articulations the starts can come out scrambled (e.g. `[0, 26, 13]`), tripping the monotonicity check in `finalize()`.
- Even when ordering happened to survive, `articulation_label`, `articulation_world`, and `joint_articulation` were not kept aligned with the (potentially shrunken) `articulation_start`, so per-articulation metadata could be silently misattributed.

This PR rebuilds articulation bookkeeping in original order from the retained joints' articulation ownership, drops fully-collapsed empty articulations, and exposes an `articulation_remap` so external callers can update their own per-articulation side data.

This builds on the work started in #2616 by @devshahofficial — that PR is from a fork I cannot push to, so I am opening a fresh one with the same approach plus the two outstanding review comments addressed:

- [Drop the silent dict-only skip in the ARTICULATION custom-attribute remap](https://github.com/newton-physics/newton/pull/2616#discussion_r3164793320) and document that enum-frequency attributes always use dict storage by construction.
- [Remove the dead `articulation_bodies` / `articulation_has_self_collision` remap in `import_usd.py`](https://github.com/newton-physics/newton/pull/2616#discussion_r3164793357) — both maps are consumed during self-collision filter setup, before `collapse_fixed_joints()` runs, so the remap had no user-visible effect.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_import_urdf
uvx pre-commit run -a
```

Manual smoke check against the original issue repro: ANYmal C URDF imported three times with `collapse_fixed_joints=True` now finalizes with `model.articulation_start == [0, 13, 26, 39]` instead of crashing.

## Bug fix

**Steps to reproduce:**

1. Import the same multi-joint URDF (e.g. ANYmal C) three times into one `ModelBuilder` with `collapse_fixed_joints=True`.
2. Call `builder.finalize()`.
3. Without this PR, finalize raises `ValueError: articulation_start is not monotonically increasing: articulation_start[1]=26 > articulation_start[2]=13`.

**Minimal reproduction:**

See `test_repeated_collapse_fixed_joints_preserves_articulation_order` in `newton/tests/test_import_urdf.py` for the inline regression test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a crash when finalizing models with 3+ articulations after simplifying fixed joints; articulation indices and per-articulation metadata are now remapped consistently.
  * Ensures custom per-articulation attributes are preserved and remapped correctly after simplification.

* **Tests**
  * Added regression tests to verify articulation ordering and attribute preservation across repeated simplifications.

### DIFF
--- a/newton/_src/sim/builder.py
+++ b/newton/_src/sim/builder.py
@@ -5131,23 +5131,57 @@ class ModelBuilder:
         # sort joints so they appear in the same order as before
         retained_joints.sort(key=lambda x: x["original_id"])
 
+        original_articulation_start = self.articulation_start[:]
+        original_articulation_label = self.articulation_label[:]
+        original_articulation_world = self.articulation_world[:]
+        original_joint_articulation = self.joint_articulation[:] if self.joint_articulation else []
+
         joint_remap = {}
+        articulation_first_joint: dict[int, int] = {}
         for i, joint in enumerate(retained_joints):
-            joint_remap[joint["original_id"]] = i
-        # update articulation_start
-        for i, old_i in enumerate(self.articulation_start):
-            start_i = old_i
-            while start_i not in joint_remap:
-                start_i += 1
-                if start_i >= self.joint_count:
-                    break
-            self.articulation_start[i] = joint_remap.get(start_i, start_i)
-        # remove empty articulation starts, i.e. where the start and end are the same
-        self.articulation_start = list(set(self.articulation_start))
+            old_joint_idx = joint["original_id"]
+            joint_remap[old_joint_idx] = i
+            if original_joint_articulation and old_joint_idx < len(original_joint_articulation):
+                old_articulation = original_joint_articulation[old_joint_idx]
+                if old_articulation >= 0 and old_articulation not in articulation_first_joint:
+                    articulation_first_joint[old_articulation] = i
+
+        # Update articulation starts from retained joints' original articulation
+        # ownership. This preserves articulation order while dropping empty
+        # articulations whose joints were fully collapsed away.
+        articulation_remap: dict[int, int] = {}
+        new_articulation_start: list[int] = []
+        new_articulation_label: list[str] = []
+        new_articulation_world: list[int] = []
+        for articulation_idx in range(len(original_articulation_start)):
+            if articulation_idx not in articulation_first_joint:
+                continue
+
+            articulation_remap[articulation_idx] = len(new_articulation_start)
+            new_articulation_start.append(articulation_first_joint[articulation_idx])
+            if articulation_idx < len(original_articulation_label):
+                new_articulation_label.append(original_articulation_label[articulation_idx])
+            else:
+                new_articulation_label.append(f"articulation_{articulation_idx}")
+            if articulation_idx < len(original_articulation_world):
+                new_articulation_world.append(original_articulation_world[articulation_idx])
+            else:
+                new_articulation_world.append(self.current_world)
+
+        self.articulation_start = new_articulation_start
+        self.articulation_label = new_articulation_label
+        self.articulation_world = new_articulation_world
+
+        for custom_attr in self.get_custom_attributes_by_frequency([Model.AttributeFrequency.ARTICULATION]):
+            if isinstance(custom_attr.values, dict):
+                custom_attr.values = {
+                    new_idx: custom_attr.values[old_idx]
+                    for old_idx, new_idx in articulation_remap.items()
+                    if old_idx in custom_attr.values
+                }
 
         # save original joint worlds and articulations before clearing
         original_ = self.joint_world[:] if self.joint_world else []
-        original_articulation = self.joint_articulation[:] if self.joint_articulation else []
 
         self.joint_label.clear()
         self.joint_type.clear()
@@ -5200,8 +5234,9 @@ class ModelBuilder:
                 # If no world was assigned, use default -1
                 self.joint_world.append(-1)
             # Rebuild joint articulation assignment
-            if original_articulation and joint["original_id"] < len(original_articulation):
-                self.joint_articulation.append(original_articulation[joint["original_id"]])
+            if original_joint_articulation and joint["original_id"] < len(original_joint_articulation):
+                old_articulation = original_joint_articulation[joint["original_id"]]
+                self.joint_articulation.append(articulation_remap.get(old_articulation, -1))
             else:
                 self.joint_articulation.append(-1)
             for axis in joint["axes"]:
--- a/newton/_src/sim/builder.py
+++ b/newton/_src/sim/builder.py
@@ -5172,6 +5172,17 @@ class ModelBuilder:
         self.articulation_label = new_articulation_label
         self.articulation_world = new_articulation_world
 
+        def remap_articulation_reference(value: Any) -> Any:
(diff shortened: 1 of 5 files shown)


newton/_src/sim/builder.py
```diff
@@ -5131,23 +5131,79 @@ def dfs(parent_body: int, child_body: int, incoming_xform: wp.transform, last_dy
...
+
+        for custom_attr in self.custom_attributes.values():
+            if custom_attr.references != "articulation" or custom_attr.values is None:
+                continue
+            if isinstance(custom_attr.values, dict):
+                custom_attr.values = {
+                    entity_idx: remap_articulation_reference(value) for entity_idx, value in custom_attr.values.items()
+                }
+            else:
+                custom_attr.values = [remap_articulation_reference(value) for value in custom_attr.values]
```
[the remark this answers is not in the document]
vreutskyy: Declining this. No current code path populates articulation-referencing custom attributes with NumPy values — your own grep results in this thread confirm that (`values=...np.array...` against `references="articulation"` returned no hits). The existing test exercises Python `int`/`list`/`tuple`, which are the only types this function sees today. Per the project guidance in `AGENTS.md` ("don't add validation for scenarios that can't happen"), I'd rather not introduce speculative `numpy.integer`/`np.ndarray` branches now and let a real failing case drive the extension if/when one shows up.

newton/_src/sim/builder.py
```diff
@@ -5131,23 +5131,79 @@ def dfs(parent_body: int, child_body: int, incoming_xform: wp.transform, last_dy
...
+                new_articulation_world.append(self.current_world)
+
+        self.articulation_start = new_articulation_start
+        self.articulation_label = new_articulation_label
+        self.articulation_world = new_articulation_world
+
+        def remap_articulation_reference(value: Any) -> Any:
+            if isinstance(value, bool):
+                return value
+            if isinstance(value, int):
```
camevor: Testing for `int` may be too narrow here, as it misses `wp.int` types (especially since `_default_for_dtype` returns `dtype(0)`).

vreutskyy: Good catch — confirmed `wp.int32` instances are not Python ints (`isinstance(wp.int32(0), int)` is `False`), so a custom attribute with `dtype=wp.int32` and `references="articulation"` would skip the remap entirely and keep stale indices.

Fixed in f941ad91: replaced the `isinstance(value, int)` branch with a `try: int(value)` conversion that accepts Python ints, Warp scalar integer types (`wp.int8/16/32/64`, `wp.uint*`), and anything else `int`-castable. Negative values are still left untouched.

Also extended `test_collapse_fixed_joints` with a `wp.int32`-valued articulation reference (`{0: [wp.int32(1), (wp.int32(2), wp.int32(-1))]}`) to lock in the new path.

camevor (approved):
Thanks, lgtm!

PR #2691, opened 2026-05-04
Rework/bvh raycast

### LINKED ISSUE #2629: Deprecate SensorRaycast
Deprecate SensorRaycast with a migration path to SensorTiledCamera that cover the single-camera depth use case ergonomically.

StafaH:
## Description

This PR adds deprecation warnings to SensorRaycast for users to migrate to using the TiledCameraSensor.

## Checklist

## Summary by CodeRabbit

* **Deprecations**
  * SensorRaycast is now deprecated; instantiating it emits a deprecation warning and users are directed to the recommended SensorTiledCamera utilities for single-camera depth workflows.
* **Documentation**
  * Docs updated to mark SensorRaycast as deprecated and provide migration guidance.
* **Tests**
  * Tests updated to assert the deprecation warning and to suppress it where appropriate during existing test runs.


newton/_src/sensors/sensor_raycast.py
```diff
@@ -88,6 +120,12 @@ def __init__(
             height: Image height in pixels
             max_distance: Maximum ray distance; rays beyond this return no hit
         """
+        warnings.warn(
+            "SensorRaycast is deprecated; use SensorTiledCamera instead. "
```
eric-heiden, line 124: 🟡 This deprecation warning is intentional, but the existing `SensorRaycast` tests construct this class many times without suppressing or asserting the warning. That can add warning noise in CI and makes future broader `warnings.catch_warnings(record=True)` checks brittle.

Please add a test-local filter or helper in `newton/tests/test_sensor_raycast.py`, matching the existing deprecated-class test pattern used elsewhere, so the tests explicitly acknowledge this deprecation without emitting warnings on every construction.

eric-heiden:
🟡 `docs/concepts/sensors.rst` still lists `SensorRaycast` beside the active sensor classes without saying it is deprecated. A reader using that overview could still pick `SensorRaycast` as a normal option even though the class now points users to `SensorTiledCamera`.

Please annotate the hand-written concept page entry so the deprecation is visible there too, for example:

```rst
* :class:`~newton.sensors.SensorRaycast` -- *(deprecated)* ray-based depth images from a virtual camera; use :class:`~newton.sensors.SensorTiledCamera` instead.
```

eric-heiden (approved):
Thanks!

PR #2694, opened 2026-05-04
Bind Viser playback to loopback, fix recording replay

eric-heiden:
## Description

Bind the temporary `ViewerViser` notebook recording playback server to `127.0.0.1` instead of all interfaces. This addresses CodeQL alert 41 without changing the live `viser.ViserServer` path used for LAN and cloud viewing.

Closes: https://github.com/newton-physics/newton/security/code-scanning/41

This also fixes generated tutorial playback by removing Newton's bundled Viser web client and using the installed `viser` package as the single source of truth for browser assets. `ViewerViser._serve_viser_recording()` and `docs/conf.py::_copy_viser_client_into_output_static()` now share `ViewerViser._get_viser_client_dir()`, so Jupyter playback and Sphinx/static docs cannot drift to different client builds.

Additional cleanup from the imported docs/viewer work:

- Render `ViewerViser` planes as finite wire grids to avoid flickering artifacts.
- Add notebook Mermaid rendering support for nbsphinx outputs.
- Update docs/notebook dependency extras and `uv.lock` so docs builds use the matching Viser client.

## Checklist

## Test plan

```
uvx pre-commit run -a
uvx pre-commit run --files docs/_static/mermaid-nbsphinx.js
git diff --check
$env:NEWTON_BUILD_NOTEBOOKS='1'; $env:NEWTON_NBSPHINX_EXECUTE='always'; uv run --extra docs --extra sim sphinx-build -E -a -j auto -b html -t notebooks docs docs/_build/html
```

Local docs smoke checks against `http://127.0.0.1:9001/`:

- `/`
- `/guide/tutorials.html`
- `/tutorials/00_introduction.html`
- `/_static/viser/index.html`
- `/_static/mermaid-nbsphinx.js`

## Bug fix

**Steps to reproduce:**

1. Run CodeQL on `newton/_src/viewer/viewer_viser.py` or inspect `_serve_viser_recording()`.
2. Observe the notebook recording playback helper binding a socket to all interfaces via an empty host string.
3. Build the docs tutorial notebook with a `.viser` recording generated by the installed `viser` package.
4. Observe playback fail when the generated docs serve Newton's stale bundled client from `_static/viser`.

**Minimal reproduction:**

```python
from newton.viewer import ViewerViser

ViewerViser._serve_viser_recording("simulation.viser")
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Notebook recording playback now binds only to the loopback interface and reliably serves using a matching local browser client.
  * Playback falls back to bundled client assets when external assets aren’t available.

* **Documentation**
  * Added client-side rendering for Mermaid diagrams and styling to improve docs diagram display.
  * Docs build now tolerates missing viewer assets with a warning instead of failing.

* **Chores**
  * Updated documentation-related dependencies.


preist-nvidia:
In my local docs build, the generated tutorial page does not show the Viser recording because of a client/recording mismatch.

The tutorial recording was generated with installed `viser 1.0.24`, while `_static/viser` was populated from Newton’s older bundled client under `newton/_src/viewer/viser/static`. The iframe shell loads, but playback fails in the browser with:

`TypeError: The compressed data was not valid: incorrect header check.`

I think the most robust direction may be to remove Newton’s bundled Viser web client entirely and rely on the installed `viser` package for the browser client assets. Since the docs extra already depends on `viser`, and `ViewerViser` itself requires `viser`, this would make the installed package the single source of truth for playback compatibility.

If removing the bundle is too broad for this PR, the smaller fix is to make `docs/conf.py::_copy_viser_client_into_output_static()` use the same installed-client-first selection as `_serve_viser_recording()`. Ideally the selection logic should be consolidated so the regular notebook path and Sphinx/static-docs path cannot drift again.

Concretely:
- Preferred: delete `newton/_src/viewer/viser/static` and require `Path(viser.__file__).resolve().parent / "client" / "build"` for playback assets.
- Minimal: update `docs/conf.py` to prefer installed `viser/client/build` before falling back to Newton’s bundled static client.
- Either way: reuse one helper for both `_serve_viser_recording()` and docs asset copying.

I verified locally that after copying the installed `viser 1.0.24` client into `_static/viser`, the generated tutorial iframe logs:

`File loaded! Saved with Viser version: 1.0.24`

and renders successfully.

eric-heiden:
@preist-nvidia thanks for the detailed repro. I merged the ViewerViser fixes from https://github.com/newton-physics/newton/pull/2415 that address your comments:

- Removed Newton's bundled `newton/_src/viewer/viser/static` client assets.
- Added one shared `ViewerViser._get_viser_client_dir()` helper that resolves the installed `viser` package client build.
- Updated both `ViewerViser._serve_viser_recording()` and `docs/conf.py::_copy_viser_client_into_output_static()` to use that helper, so notebook playback and Sphinx/static docs use the same client selection logic.
- Updated docs/notebook extras and the lockfile so the docs build uses the matching installed Viser client.

preist-nvidia (approved):
Approved from my side. I re-tested the latest head (97e5a9f96) locally and the original mismatch is fixed for me.

Validation:
- Rebuilt notebook-enabled docs with Sphinx.
- Served the built docs locally and confirmed the Viser recording plays in the static docs.
- Started JupyterLab from the PR worktree and confirmed the same recording also plays directly in the notebook.

The remaining failed CI jobs look unrelated to this change: Windows failed while fetching google-deepmind/mujoco_menagerie.git with an empty server reply, and macOS lost runner communication. I would rerun those, but I do not see a code-review blocker for this PR.

PR #2696, opened 2026-05-05
Fix floating URDF dummy roots

### LINKED ISSUE #2555: [BUG] urdf stays fixed when using the XPBD solver(even when 'floating' is set to True)
### Bug Description

Image

### Reproduction Script

```python
#URDF
self.robot_xform=wp.transform(p=wp.vec3(0.0, 8.0, drop_z))
builder.add_urdf(
    source="urdf/test_ackermann_drive.xacro.urdf",
    enable_self_collisions=False,
    floating=True,
    xform=self.robot_xform,
)

#add this snippet to the example_basic_shapes.
--------------------------------
#must also call eval_fk(once after build), else the robot links break down
newton.eval_fk(
    self.model,
    self.state_0.joint_q,
    self.state_0.joint_qd,
    self.state_0,
)
```

### System Information

Warp 1.13.0.dev20260422 initialized:
   CUDA Toolkit 12.9, Driver 12.4
   Devices:
     "cpu"      : "x86_64"
     "cuda:0"   : "NVIDIA GeForce RTX 3050 Laptop GPU" (4 GiB, sm_86, mempool enabled)

devshahofficial:
## Description

Closes #2555.

Floating URDF imports can create a free root joint on a massless dummy root link that is fixed to the physical base link. XPBD then sees zero inverse mass for the root body and leaves the articulation immovable even when `floating=True`.

This changes URDF import to collapse that massless fixed root subtree for floating imports, while preserving fixed joints that existed before the import. It also adds a regression that verifies XPBD moves the imported model under gravity.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k floating_massless_fixed_root
uv run --extra dev -m newton.tests -k test_import_urdf.TestImportUrdfBasic
uv run --extra dev -m newton.tests -k test_import_urdf
uvx pre-commit run -a
```

Also verified the targeted regression in a Linux Docker CPU run:

```bash
uv run --extra dev -m newton.tests -k floating_massless_fixed_root
```

## Bug fix

**Steps to reproduce:**

1. Import a URDF whose root link is massless and fixed to the physical base link with `floating=True`.
2. Finalize the model and inspect the root body used by the free joint.
3. Step with `SolverXPBD` under gravity.
4. Without this PR, the root body has zero inverse mass and does not fall.

**Minimal reproduction:**

```python
import newton

builder.add_urdf("robot_with_massless_fixed_root.urdf", floating=True, up_axis="Z")
model = builder.finalize()

root_body = int(model.joint_child.numpy()[0])
print(model.body_inv_mass.numpy()[root_body])
```

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in importer option to collapse massless fixed-root chains while preserving topology, enabling dynamic behavior for previously massless floating roots.

* **Bug Fixes**
  * Importer preserves existing internal and pre-existing fixed joints when collapsing massless fixed-root chains.

* **Tests**
  * Added tests verifying topology preservation and gravity-driven dynamic behavior when the opt-in collapse is used.

* **Documentation**
  * Changelog entry documenting the new opt-in option.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2696?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/newton/_src/sim/builder.py
+++ b/newton/_src/sim/builder.py
@@ -2277,6 +2277,7 @@ class ModelBuilder:
         joint_ordering: Literal["bfs", "dfs"] | None = "dfs",
         bodies_follow_joint_ordering: bool = True,
         collapse_fixed_joints: bool = False,
+        collapse_massless_fixed_root: bool = False,
         mesh_maxhullvert: int | None = None,
         force_position_velocity_actuation: bool = False,
         override_root_xform: bool = False,
@@ -2370,6 +2371,7 @@ class ModelBuilder:
             joint_ordering: The ordering of the joints in the simulation. Can be either "bfs" or "dfs" for breadth-first or depth-first search, or ``None`` to keep joints in the order in which they appear in the URDF. Default is "dfs".
             bodies_follow_joint_ordering: If True, the bodies are added to the builder in the same order as the joints (parent then child body). Otherwise, bodies are added in the order they appear in the URDF. Default is True.
             collapse_fixed_joints: If True, fixed joints are removed and the respective bodies are merged.
+            collapse_massless_fixed_root: If True, collapse only the massless fixed-joint chain below an imported free root body. Ignored when ``collapse_fixed_joints`` is True.
             mesh_maxhullvert: Maximum vertices for convex hull approximation of meshes.
             force_position_velocity_actuation: If True and both position (stiffness) and velocity
                 (damping) gains are non-zero, joints use :attr:`~newton.JointTargetMode.POSITION_VELOCITY` actuation mode.
@@ -2397,6 +2399,7 @@ class ModelBuilder:
             joint_ordering=joint_ordering,
             bodies_follow_joint_ordering=bodies_follow_joint_ordering,
             collapse_fixed_joints=collapse_fixed_joints,
+            collapse_massless_fixed_root=collapse_massless_fixed_root,
             mesh_maxhullvert=mesh_maxhullvert,
             force_position_velocity_actuation=force_position_velocity_actuation,
             override_root_xform=override_root_xform,
@@ -2654,6 +2657,7 @@ class ModelBuilder:
         enable_self_collisions: bool = True,
         ignore_inertial_definitions: bool = False,
         collapse_fixed_joints: bool = False,
+        collapse_massless_fixed_root: bool = False,
         verbose: bool = False,
         skip_equality_constraints: bool = False,
         convert_3d_hinge_to_ball_joints: bool = False,
@@ -2759,6 +2763,7 @@ class ModelBuilder:
             enable_self_collisions: If True, self-collisions are enabled.
             ignore_inertial_definitions: If True, the inertial parameters defined in the MJCF are ignored and the inertia is calculated from the shape geometry.
             collapse_fixed_joints: If True, fixed joints are removed and the respective bodies are merged.
+            collapse_massless_fixed_root: If True, collapse only the massless fixed-joint chain below an imported free root body. Ignored when ``collapse_fixed_joints`` is True.
             verbose: If True, print additional information about parsing the MJCF.
             skip_equality_constraints: Whether <equality> tags should be parsed. If True, equality constraints are ignored.
             convert_3d_hinge_to_ball_joints: If True, series of three hinge joints are converted to a single ball joint. Default is False.
@@ -2799,6 +2804,7 @@ class ModelBuilder:
             enable_self_collisions=enable_self_collisions,
             ignore_inertial_definitions=ignore_inertial_definitions,
             collapse_fixed_joints=collapse_fixed_joints,
+            collapse_massless_fixed_root=collapse_massless_fixed_root,
             verbose=verbose,
             skip_equality_constraints=skip_equality_constraints,
             convert_3d_hinge_to_ball_joints=convert_3d_hinge_to_ball_joints,
--- a/newton/_src/sim/builder.py
+++ b/newton/_src/sim/builder.py
@@ -4838,16 +4838,16 @@ class ModelBuilder:
     def collapse_fixed_joints(
         self,
         verbose: bool = False,
-        joints_to_keep: list[str] | None = None,
-        joint_indices_to_keep: set[int] | None = None,
+        joints_to_keep: Sequence[str | int] | None = None,
     ) -> dict[str, Any]:
         """Removes fixed joints from the model and merges the bodies they connect. This is useful for simplifying the model for faster and more stable simulation.
 
         Args:
             verbose: If True, print additional information about the collapsed joints.
-            joints_to_keep: An optional list of joint labels to be excluded from the collapse process.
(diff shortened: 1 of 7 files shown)


devshahofficial:
Added the companion regression requested in the CodeRabbit quick-win thread in 71f4b6ee.

It verifies that when a pre-existing fixed-joint articulation is already in the builder, the floating URDF auto-collapse only removes the imported massless dummy root fixed joint and preserves those earlier fixed joints.

Validation:
```bash
uv run --extra dev -m newton.tests -k floating_massless_fixed_root
uvx pre-commit run -a
```

devshahofficial:
Pushed a follow-up in devshahofficial/newton@d35e26ec to narrow the floating massless-root auto-collapse path.

The auto-collapse now only removes the fixed joints in the imported massless-root dummy chain, and preserves other fixed joints from the same URDF when `collapse_fixed_joints=False`. I added a regression with an internal fixed sensor link to cover that case.

Validation:
```bash
uv run --extra dev -m newton.tests -k floating_massless_fixed_root
uvx pre-commit run -a
```

devshahofficial:
@eric-heiden could you review this when you have a chance?

I read through the CodeRabbit follow-ups and pushed `a8a50da2` to address the remaining useful test coverage nits:
- the XPBD dynamic regression now runs against both the simple massless fixed-root URDF and the internal-fixed-joint variant;
- the motion assertion now requires a real gravity-scale drop instead of only `end_z < start_z`.

I intentionally did not add a CUDA-only skip because the focused test passes on the CPU Warp backend, and gating it on CUDA would reduce coverage unnecessarily.

Validation:
```bash
uv run --extra dev -m newton.tests -k floating_massless_fixed_root
uvx pre-commit run -a
```

eric-heiden (commented):
Thanks @devshahofficial, but as I commented on https://github.com/newton-physics/newton/issues/2555#issuecomment-4410421979 I don't think we should automatically collapse these dummy joints. Generalized coordinate solvers like MuJoCo or Featherstone can handle them just fine. Users of maximal-coordinate solvers just need to collapse fixed joints, a function we already provide. Removing dummy links/joints may have unintended consequences if users rely on the indexing of joints or bodies etc. so I would rather not merge this PR in its current form.

Perhaps we could make this automatic collapsing of dummy root joints you did here another flag on `add_urdf`? I guess we should also mirror this behavior in the MJCF importer then.

devshahofficial:
Thanks @eric-heiden, I see the concern about silently changing imported topology, especially for users relying on body/joint indices or dummy links.

Would you be comfortable with changing this PR so the default behavior stays topology-preserving, and the dummy-root collapse becomes opt-in on the importer? For example, something like:

```python
builder.add_urdf(..., collapse_massless_fixed_root=True)
```

The intent would be narrower than `collapse_fixed_joints=True`: only collapse the massless fixed chain at the root that prevents maximal-coordinate solvers like XPBD from moving the articulation, while preserving internal fixed joints/sites.

A few questions before I revise it:

1. Should this flag default to `False` to preserve existing behavior?
2. Do you prefer this as a URDF-only fix first, or should I add the matching MJCF flag in the same PR?
3. Is `collapse_massless_fixed_root` the right name, or would you prefer something more explicit like `collapse_fixed_root_joints`?
4. Should the flag collapse only chains where every intermediate body is massless, stopping once it reaches the first massive body?

I can update the tests to cover both default-preserve behavior and opt-in XPBD behavior.

devshahofficial:
Pushed the update, @eric-heiden @preist-nvidia.

This now keeps importer topology unchanged by default. The dummy-root cleanup is opt-in with `collapse_massless_fixed_root=True` on both URDF and MJCF. When it is enabled, it only collapses the massless fixed chain under the free root, so regular/internal fixed joints stay intact.

Added coverage for:
- default topology-preserving behavior;
- XPBD falling with the flag enabled;
- internal fixed-joint preservation for both URDF and MJCF.

Local checks I ran:
- `uv run --extra dev -m newton.tests -k test_import_urdf`
- `uv run --extra dev -m newton.tests -k test_import_mjcf.TestImportMjcfBasic`
- `uv run --extra dev -m newton.tests -k collapse_fixed_joints`
- `uv run --extra dev -m newton.tests -k test_api`
- `uv run --extra dev -m newton.tests -k massless_fixed_root`
- `uvx pre-commit run -a`

eric-heiden (commented):
Thanks @devshahofficial, I like this behavior is now opt-in via `collapse_massless_fixed_root`. A few more comments to follow up before we can merge.

newton/_src/utils/import_utils.py
```diff
@@ -205,6 +206,69 @@ def should_show_collider(
...
+        joint_indices: Joint indices created by the current import.
+        root_joint_indices: Optional subset of ``joint_indices`` that should be
+            considered articulation roots.
+    """
+    imported_joint_indices = set(joint_indices)
+    if root_joint_indices is None:
+        root_joint_indices = [
+            joint_idx
+            for joint_idx in joint_indices
+            if builder.joint_type[joint_idx] == JointType.FREE and builder.joint_parent[joint_idx] == -1
```
eric-heiden, line 227: 🔴 The new auto-detection here is safer than the MJCF caller's current `root_joint_indices` construction, but `import_mjcf.py` still bypasses it by passing the first joint from each root-body slice.

That is fragile because the first joint in a root body's imported slice is not guaranteed to be the actual FREE root. MJCF bodies can contain frames, geom-only bodies, included worldbody content, or other structure that changes joint ordering, so the caller can end up marking a non-root joint as the articulation root. Then the massless fixed-root collapse can remove the wrong subtree or miss the intended FREE-root chain.

Please have the MJCF path use this helper fallback instead of constructing `root_joint_indices` positionally. For example:

```py
elif collapse_massless_fixed_root:
    collapse_massless_fixed_root_joints(builder, joint_indices)
```

That keeps the MJCF behavior aligned with the URDF path and filters roots by `JointType.FREE` with `joint_parent == -1`, which is the property this collapse actually needs.

newton/_src/sim/builder.py
```diff
@@ -4829,12 +4835,18 @@ def shape_type_str(type):
             plt.legend(loc="upper left", fontsize=6)
         plt.show()
 
-    def collapse_fixed_joints(self, verbose: bool = False, joints_to_keep: list[str] | None = None) -> dict[str, Any]:
+    def collapse_fixed_joints(
+        self,
+        verbose: bool = False,
+        joints_to_keep: list[str] | None = None,
+        joint_indices_to_keep: set[int] | None = None,
```
eric-heiden: It would be better to not introduce an extra argument here for joint indices.
Can we just use `joints_to_keep` to accept `set[str]` and `set[int]`, or `set[int | str]`?
Then there is also no contention to check two input arguments for joints that we want to keep.

eric-heiden (approved):
Thanks @devshahofficial for addressing my comments!

PR #2701, opened 2026-05-05
Simplify the hydroelastics broad phase

### LINKED ISSUE #2611: [BUG] Hydroelastic contact surface loses center triangles under deep overlap
### Bug Description

The SDF hydroelastic contact surface can lose the central region of the contact patch under deep interpenetration. This still happens after the degenerate-triangle fix from PR #2424: the failure mode is no longer primarily zero-area triangles, but missing triangles that were never emitted in the center of the contact surface.

Static case, two boxes are overlapped:

Image

Dynamic case, ellipsoid pushed into a box:

Image

### Expected Behavior
The extracted hydroelastic contact surface should cover the pressure-equilibrium surface throughout the overlapping volume, including the center of a symmetric box-box contact patch. If a penetration is outside the valid SDF narrow-band regime, the pipeline should fail or warn in a way that is observable, rather than emitting a plausible-looking annular contact surface with a silent center hole.

### Impact
This is visible in tactile-sensing workflows because raycasts or rasterization over the contact surface produce a nonphysical center trough or missing displacement region. It can also distort contact force distribution because the missing center patch contributes no area.

### Attempts to Debug / Root-Cause Analysis
Parameters swept:
- SDF resolution
- kH stiffness (different ranges + magnitudes of difference between 2 objects)
- narrow band 
- contact gap
- SDF texture format
- geometry type: off-the-shelf high quality mesh (i.e., YCB) + geometry primitives
- scales of objects (mm, cm)

Sanity checks:
- reduce_contacts = False
- contact buffers large enough to accommodate generated contacts

Current root-cause hypothesis, narrowed but not yet proven:
- `texture_sample_sdf()` / `texture_sample_sdf_grad()` clamp query positions to the SDF storage box and add distance outside the box.
- The hydroelastic contact scalar uses the weighted SDF difference when both shapes are inside: `k_eff_a * valA - k_eff_b * valB`.
- The iso-voxel refinement path appears to cull the central region before marching cubes, despite sign-changing texture-SDF samples near the expected center contact plane. The culling seems tied to deep overlap with a narrow interior band, because widening the narrow band restores center iso-voxels for the 10 cm case.
- PR #2424 can prevent collapsed triangles after faces exist, but it cannot recover center faces if the corresponding iso-voxels are absent before marching cubes.
- The dynamic repro shows the same missing-center transition while moving through the collision pipeline, so this is not only an artifact of a pre-overlapped initial pose.

### Ideal Solution
- For the repro below, the 10 cm overlap / 20 mm narrow-band case should either produce a filled center contact surface or emit a clear diagnostic that the requested penetration is outside the supported SDF range.
- The fix should not rely on using a very large narrow band for every shape, because that is memory-expensive and still fails at deeper overlaps (`15 cm` overlap with `50 mm` narrow band still has `center_coverage=0.00` in the sweep).

### Repro Details
In the repro below, two 20 cm hydroelastic boxes are overlapped by 10 cm with a 20 mm narrow band, 20 mm contact gap, equal `kh=1e10`, `sdf_max_resolution=64`, and `reduce_contacts=False`. The contact surface has 47,148 faces and 0 exact zero-area triangles, but no triangle centroids in the central patch. Increasing SDF resolution to 128 and using `float32` SDF textures does not restore the center triangles. Increasing the narrow band to 50 mm does restore the center for this 10 cm overlap case.

[newton-hydroelastic-center-gap_dynamic_repro.py](https://github.com/user-attachments/files/27145597/newton-hydroelastic-center-gap_dynamic_repro.py)
[newton-hydroelastic-center-gap_repro.py](https://github.com/user-attachments/files/27145598/newton-hydroelastic-center-gap_repro.py)

### Reproduction Script

```python

```

### System Information

- Newton: `1.1.0.dev0`
- Newton checkout: clean `main`, commit `b468ab43258f277e8f3438b9ba95da189b2c95d3` (`origin/main` as fetched on 2026-04-27)
- Related fix in history: `e7fba4e5b79a48cb19609503099c3f1d4373ad44` (`Fix degeneracies in hydroelastic contact surfaces (#2424)`)
- Warp: `1.12.0rc1`
- Python: `3.12.3`
- OS: Ubuntu 24.04.4 LTS, Linux `6.17.0-14-generic`
- GPU: NVIDIA RTX 6000 Ada Generation, 49140 MiB
- NVIDIA driver: `570.211.01`
- CUDA reported by `nvidia-smi`: `12.8`
- CUDA Toolkit reported by Warp: `12.9`

nvtw:
## Description
Fix hydroelastic SDF contact patches dropping their central region under deep interpenetration: the broadphase used to skip subgrids whose centers were deeper than the SDF narrow band, leaving a hole in the contact patch when overlap exceeded the narrow-band thickness. The broadphase now visits every subgrid in the SDF coarse grid, deriving block coordinates arithmetically from each shape's coarse-texture dimensions instead of from a precomputed occupancy list. Removes the now-unused `Model.sdf_block_coords`, `Model.sdf_index2blocks`, `SDF.texture_block_coords`, and the `subgrid_occupied` build-pipeline array. Bumps `_sdf_cache.CACHE_FORMAT_VERSION` to `2`; existing on-disk caches are transparently re-cooked. Closes #2611.

## Checklist

## Test plan
```
uv run --extra dev -m newton.tests -k sdf
uv run --extra dev -m newton.tests -k hydro
```

## Bug fix
**Steps to reproduce:**
1. Create two hydroelastic SDF shapes and drive them into deep interpenetration (overlap larger than the SDF narrow-band thickness).
2. Inspect the generated contact patch.
3. Observe a hole in the center of the patch where subgrids whose centers fell outside the narrow band were skipped by the broadphase.

**Minimal reproduction:** see `newton-hydroelastic-center-gap_repro.py` / `newton-hydroelastic-center-gap_dynamic_repro.py` attached to #2611.

## Summary by CodeRabbit

* **Deprecated**
  * texture_block_coords now returns None and emits a DeprecationWarning.
  * Public SDF/texture attributes are aliased to new private underscored fields; deprecated aliases preserved with warnings and lazy legacy reconstruction (removal planned).

* **Changed**
  * Broadphase now visits all subgrids and derives block coordinates from coarse-texture dimensions.
  * On-disk SDF cache format bumped to v2; older caches are transparently re-cooked.
  * Texture‑SDF outputs and sparse-data omit mesh-derived block-coordinates and prior subgrid‑occupancy field.

* **Bug Fixes**
  * Fixed deep-penetration contact-surface central-hole regression.

* **Tests**
  * Updated tests and added CUDA regressions for empty-SDF handling and deep-penetration surfaces.


newton/_src/geometry/sdf_hydroelastic.py
lenroe, line 1: Makes sense to compute blocks algebraically from a user-friendliness perspective because it alleviates needing to set narrow-band ranges according to max expected penetration. The upside of this was that the precomputed blocks arrays were pre-filtering isovoxel checks to only the "interesting" narrow-band region. As I understand now all blocks have to be exhaustively checked in the first isovoxel pruning path. Did you check the performance impact of this?

nvtw: The nut and bolt hydroelastic benchmark that runs on CI only shows an 1% regression. I have not done exhaustive performance testing though. But we do the subgrid pruning, hierarchically, no? So it should not be extremely expensive to include all subgrid cells I assume. Will double check.

nvtw: The reason for the change in this MR is that researchers want to get full iso-pressure-surfaces even when bodies overlap heavily (there was some pruning that prevented that). And we also need to store less data if we include all subgrids in the broadphase, that simplification was another factor to give the current implementation a try.

nvtw: If you can find time to compare performance on a few of your examples, that would be great.

lenroe: @nvtw I tested it on some more scenes and your right, it doesn't seem to have a significant perf impact. So I think this is great!

newton/_src/geometry/sdf_texture.py
```diff
@@ -1421,12 +1411,10 @@ def create_texture_sdf_from_mesh(
             return signature.
 
     Returns:
-        Tuple of ``(texture_sdf, coarse_texture, subgrid_texture, block_coords)``.
+        Tuple of ``(texture_sdf, coarse_texture, subgrid_texture)``.
```
mzamoramora-nvidia, line 1639: Should we update the annotation type of `create_texture_sdf_from_mesh()` now that `block_coords` are not returned?

mzamoramora-nvidia:
Thanks for the fix. A few comments bellow:
- Could we add one small regression test for the fix? Like the one reported in the issue? 
- About the changelog: 
  - This is a bug fix, so the entry could live under `Fixed` rather than `Changed`.
  - `SDF.texture_block_coords` is public API. Should we have a deprecation cycle for that?

CHANGELOG.md
```diff
@@ -103,11 +103,13 @@
 - Deprecate `SensorRaycast` in favor of `SensorTiledCamera`; migrate to `SensorTiledCamera.utils.compute_pinhole_camera_rays()` and `create_depth_image_output()` for single-camera depth rendering — see the `SensorRaycast` class docstring for a complete migration example
 - Deprecate and ignore `rigid_enable_dahl_friction` in `SolverVBD`; Dahl friction is now auto-detected from model attributes (`model.vbd.dahl_eps_max` / `model.vbd.dahl_tau`)
 - Deprecate `newton-actuators` package dependency; all actuator functionality is now built into `newton.actuators`. The dependency is kept for backward compatibility and will be removed in a future release; migrate imports from `newton_actuators` to `newton.actuators`
+- Deprecate `SDF.texture_block_coords`; the attribute now always returns `None` and will be removed in Newton 1.4. The hydroelastic broadphase now derives block coordinates arithmetically from each SDF's coarse-texture dimensions, so callers no longer need to read them off the `SDF` object
```
mzamoramora-nvidia: Deprecation should run for two cycles. So, this should be removed in 1.5.

nvtw: Done

mzamoramora-nvidia (dismissed):
Thanks for fixing this!

adenzler-nvidia (changes requested):
Requesting changes on the deprecation handling.

This is not a real deprecation cycle yet. The PR preserves attribute lookup for `SDF.texture_block_coords`, `Model.sdf_block_coords`, and `Model.sdf_index2blocks`, but it changes their behavior immediately to always return `None`. Existing callers that iterate, index, call `len(...)`, call `.numpy()`, or use the ranges will break in this release.

Per the project guideline, breaking changes require a deprecation first. A compatibility shim should preserve the old behavioral contract during the deprecation window while emitting `DeprecationWarning`, then remove/change the behavior in the scheduled removal release. For this PR, please either:

- keep returning compatible block-coordinate/range data, possibly lazily derived from the coarse texture dimensions, while warning; or
- defer the behavioral removal until after the deprecation period.

The current changelog/docstrings say the attributes are “deprecated and always return `None`”, but that documents the break rather than avoiding it. Please also add actionable migration guidance for users who consumed these values directly, and state whether there is any opt-out or compatibility path for the new all-subgrid hydroelastic broadphase behavior.

nvtw:
> Requesting changes on the deprecation handling.
> 
> This is not a real deprecation cycle yet. The PR preserves attribute lookup for `SDF.texture_block_coords`, `Model.sdf_block_coords`, and `Model.sdf_index2blocks`, but it changes their behavior immediately to always return `None`. Existing callers that iterate, index, call `len(...)`, call `.numpy()`, or use the ranges will break in this release.
> 
> Per the project guideline, breaking changes require a deprecation first. A compatibility shim should preserve the old behavioral contract during the deprecation window while emitting `DeprecationWarning`, then remove/change the behavior in the scheduled removal release. For this PR, please either:
> 
> * keep returning compatible block-coordinate/range data, possibly lazily derived from the coarse texture dimensions, while warning; or
> * defer the behavioral removal until after the deprecation period.
> 
> The current changelog/docstrings say the attributes are “deprecated and always return `None`”, but that documents the break rather than avoiding it. Please also add actionable migration guidance for users who consumed these values directly, and state whether there is any opt-out or compatibility path for the new all-subgrid hydroelastic broadphase behavior.

@adenzler-nvidia  How can we make all SDF stuff private such that it is not public API anymore? This stuff is supposed to be internal but since it is placed on the model or wherever, it leaks into the public API. The data we remove is super specific, I assume less than 1 out of 1000 user accesses it. This completely paralyzes changes for 2 versions and I cannot remove bloat from our kernels in a reasonable amount of time.

Shall I mark SDFs as experimental until everything fully settled?

I'm really strongly against dragging this kind of stuff with us for a long time that ended up in the code because of release pressure. It's clear that every day use API stuff must be stable but this here is just internal stuff leaked into the public API.

adenzler-nvidia:
The new `### Deprecated` block in `CHANGELOG.md` has two entries that look like merge/rebase artifacts and don't belong to this PR:

1. `Deprecate loading .pt / .pth (TorchScript) checkpoints via ControllerNeuralMLP; ...` — verbatim duplicate of the existing entry already in the same `### Deprecated` section (originally added in #2858).
2. `Adjusted the grouping of reset, step, and pause so that they are all grouped together` — duplicates `Adjust grouping of reset, step, and pause controls so they appear together` already in `### Changed`, plus it's in the wrong category (Deprecated vs. Changed) and uses past tense, which violates AGENTS.md's "imperative present tense" rule.

Both lines should be deleted. The PR's own deprecation entries (`SDF.texture_block_coords`, `Model.sdf_block_coords` / `Model.sdf_index2blocks`, and the underscore-prefixed `Model.*` aliases) are fine — just the two unrelated lines need to go.

newton/_src/sim/model.py
```diff
@@ -824,6 +832,223 @@ def __init__(self, device: Devicelike | None = None):
...
+        # Local import keeps the deprecated module out of the normal load path.
+        from ..geometry._deprecated_sdf_block_coords import (  # noqa: PLC0415
+            build_legacy_sdf_block_arrays,
+        )
+
+        subgrid_size = 8
+        if self._texture_sdf_data is not None and len(self._texture_sdf_data) > 0:
+            try:
+                subgrid_size = int(self._texture_sdf_data.numpy()[0]["subgrid_size"])
+            except (KeyError, IndexError, ValueError, TypeError):
```
adenzler-nvidia: Catching four exception types and falling back to a magic constant (`subgrid_size = 8`) with no log/warn is a silent-failure surface: if `_texture_sdf_data` ever changes shape or field name, `Model.sdf_block_coords` is built with the wrong stride and produces wrong block coordinates downstream. Consider narrowing the catch to `KeyError` (the only expected miss) and `logger.warning(...)` the fallback so schema drift surfaces in CI rather than being silently masked.

newton/_src/geometry/_deprecated_sdf_block_coords.py
```diff
@@ -0,0 +1,113 @@
...
+if TYPE_CHECKING:
+    pass
+
+warnings.warn(
+    "newton._src.geometry._deprecated_sdf_block_coords is a compatibility "
+    "shim and will be removed in Newton 1.5. The hydroelastic broadphase "
+    "no longer needs precomputed block coordinates.",
+    DeprecationWarning,
+    stacklevel=2,
+)
```
adenzler-nvidia: Module-level `warnings.warn(...)` fires on every import. `Model._ensure_legacy_sdf_block_arrays` does a lazy `from ..geometry._deprecated_sdf_block_coords import ...` on every cache-miss call, so users who touch `Model.sdf_block_coords` get *two* `DeprecationWarning`s — one from the public property they named and one from the private shim module they didn't. The per-property warnings in `model.py` already cover the user-facing surface; consider dropping this module-level warning.

newton/_src/geometry/sdf_hydroelastic.py
```diff
@@ -953,15 +945,12 @@ def broadphase_collision_pairs_count(
 
     does_collide = sat_box_intersection(centered_transform_a, half_extents_a, centered_transform_b, half_extents_b)
 
-    # Sort shapes so shape with smaller voxel size is shape_b (must match scatter kernel)
-    voxel_radius_a = shape_sdf_data[shape_a].voxel_radius
-    voxel_radius_b = shape_sdf_data[shape_b].voxel_radius
-    if voxel_radius_b > voxel_radius_a:
-        shape_b, shape_a = shape_a, shape_b
+    # Pick whichever SDF has the smaller voxel as shape_b (must match scatter kernel)
```
adenzler-nvidia: The comment promises symmetry with the scatter kernel (`must match scatter kernel`), but the count kernel only reassigns the local `sdf_b` — `shape_b` is never swapped. That's correct (count only needs `dims_b`), but the comment misleads. Suggest: `# Count only needs the smaller-voxel dims, so we overwrite the local sdf_b without swapping shape indices.`

newton/_src/geometry/sdf_hydroelastic.py
```diff
@@ -1006,34 +994,31 @@ def broadphase_collision_pairs_scatter(
...
-        shape_b_idx = shape2blocks[shape_b]
-        shape_b_block_start = shape_b_idx[0]
         block_in_pair = block_tid - block_start_prefix[pair_idx]
 
-        block_broad_collide_shape_pair[block_tid] = wp.vec2i(shape_a, shape_b)
-        block_broad_idx[block_tid] = shape_b_block_start + block_in_pair
-
+        # Decode the (bx, by, bz) subgrid index from the linear block_in_pair.
+        # Layout matches the host-side row-major layout: bz outer, by middle,
+        # bx inner — same as subgrid_start_slots indexing.
```
adenzler-nvidia: `subgrid_start_slots` is a `wp.array3d` indexed `[bx, by, bz]`, so in C-order storage `bz` is the *innermost* axis — the opposite of what this decode does. The decode order actually matches the host-side linearization in `build_sparse_sdf_from_mesh()`, not `subgrid_start_slots` indexing. Suggest dropping the `subgrid_start_slots` clause or replacing it with that reference.

adenzler-nvidia (commented):
A few smaller findings from the same review pass (CHANGELOG entries already covered separately).

newton/_src/geometry/sdf_hydroelastic.py
```diff
@@ -515,26 +516,23 @@ def _from_model(
...
-        shape_sdf_shape2blocks = np.zeros((model.shape_count, 2), dtype=np.int32)
-        for shape_idx in range(model.shape_count):
-            sdf_idx = shape_sdf_index[shape_idx]
-            if sdf_idx >= 0:
-                shape_sdf_shape2blocks[shape_idx] = sdf_index2blocks[sdf_idx]
         for idx in hydroelastic_indices:
-            start_block, end_block = shape_sdf_shape2blocks[idx]
-            num_blocks = end_block - start_block
+            sdf_idx = int(shape_sdf_index[idx])
+            tex = coarse_textures[sdf_idx]
```
adenzler-nvidia, line 547: The first loop at line 502 now validates both `sdf_idx >= 0` *and* `sdf_idx < len(coarse_textures) and coarse_textures[sdf_idx] is not None` — but this second loop skips both checks. Safe today because both loops iterate the same `hydroelastic_indices` and the first one raises before we reach here, but `coarse_textures[sdf_idx]` with `sdf_idx == -1` would silently return the last texture (Python negative-index wrap) and produce a bogus block count under future refactors that split the loops. Worth either hoisting the validation into a small helper used by both loops, or repeating it (cheap).

newton/_src/geometry/_deprecated_sdf_block_coords.py
```diff
@@ -0,0 +1,113 @@
...
+from __future__ import annotations
+
+import warnings
+from typing import TYPE_CHECKING
+
+import numpy as np
+import warp as wp
+
+if TYPE_CHECKING:
+    pass
```
adenzler-nvidia: This `if TYPE_CHECKING: pass` block (and the `from typing import TYPE_CHECKING` on line 25) has no effect — there are no forward references or circular-import-guarded imports in the file. Likely a draft artifact. Both lines can be deleted with zero behavior change.

adenzler-nvidia (commented):
Project recently moved from `.. note:: ... **experimental** ...` to a proper `.. experimental::` admonition (see #2894, which converted the markers in `collide.py`). Consider aligning the three experimental markers this PR adds.

newton/_src/geometry/sdf_hydroelastic.py
```diff
@@ -184,6 +184,10 @@ def mc_calc_face_texture(
 class HydroelasticSDF:
     """Hydroelastic contact generation with SDF-based collision detection.
 
+    .. note::
+        ``HydroelasticSDF`` and the underlying SDF storage on :class:`~newton.Model`
+        are **experimental** and may change without notice.
```
adenzler-nvidia: Suggest switching to the `.. experimental::` directive for consistency with the convention #2894 just established. Something like:

```
.. experimental::

    ``HydroelasticSDF`` and the underlying SDF storage on
    :class:`~newton.Model` may change without notice.
```

newton/_src/geometry/sdf_utils.py
```diff
@@ -175,7 +176,15 @@ def sample_sdf_grad_extrapolated(
...
 class SDF:
-    """Opaque SDF container owning kernel payload and runtime references."""
+    """Opaque SDF container owning kernel payload and runtime references.
+
+    .. note::
+        The ``SDF`` API (including ``newton.SDF``, ``newton.geometry.SDF`` and
+        related helpers such as :meth:`SDF.create_from_mesh`,
+        :meth:`SDF.create_from_points`, :meth:`SDF.create_from_data` and the
+        SDF storage on :class:`~newton.Model`) is **experimental** and may
+        change without notice.
```
adenzler-nvidia: Same migration to the `.. experimental::` directive:

```
.. experimental::

    The ``SDF`` API (including ``newton.SDF``, ``newton.geometry.SDF``
    and related helpers such as :meth:`SDF.create_from_mesh`,
    :meth:`SDF.create_from_points`, :meth:`SDF.create_from_data` and the
    SDF storage on :class:`~newton.Model`) may change without notice.
```

newton/_src/sim/model.py
```diff
@@ -312,23 +313,30 @@ def __init__(self, device: Devicelike | None = None):
...
 
-        # SDF storage (compact table + per-shape index indirection)
-        self.shape_sdf_index: wp.array[wp.int32] | None = None
+        # SDF storage (compact table + per-shape index indirection).
+        # All SDF arrays are private; the public attribute names are exposed
+        # via deprecated property aliases further down for back-compat.
+        #
+        # .. note::
+        #     The SDF storage on ``Model`` is part of the **experimental** SDF
+        #     API (see :class:`~newton.SDF`) and may change without notice.
```
adenzler-nvidia: Same migration to `.. experimental::` — this is a code comment (not a docstring), so the admonition wouldn't render, but matching the wording keeps the codebase consistent:

```
# .. experimental::
#     The SDF storage on ``Model`` is part of the experimental SDF API
#     (see :class:`~newton.SDF`) and may change without notice.
```

adenzler-nvidia:
Overall this looks good — the core kernel rewrite checks out (the linear→`(bx,by,bz)` decode matches the host layout in `build_sparse_sdf_from_mesh`, the voxel-radius sort is consistent between the count and scatter kernels, and the `uint16`/`vec3us` packing carries no new overflow risk).

One thing worth tidying before merge: a few in-repo tests still read the SDF attributes this PR deprecates, so they now emit `DeprecationWarning` on every run:

- `newton/tests/test_sdf_compute.py:826,827,845,847,936,938`
- `newton/tests/test_sdf_texture.py:621,626,627,631,634,635`

These use `model.shape_sdf_index` / `model.texture_sdf_data` / `model.texture_sdf_coarse_textures`. The PR already migrated `viewer.py` and `collide.py` to the `_`-prefixed members, so updating these tests to match would keep the repo off the path it just deprecated. It won't fail the suite today (warnings go to stderr, while `CheckOutput` captures stdout), so this is cleanup rather than a blocker.

nvtw:
> Overall this looks good — the core kernel rewrite checks out (the linear→`(bx,by,bz)` decode matches the host layout in `build_sparse_sdf_from_mesh`, the voxel-radius sort is consistent between the count and scatter kernels, and the `uint16`/`vec3us` packing carries no new overflow risk).
> 
> One thing worth tidying before merge: a few in-repo tests still read the SDF attributes this PR deprecates, so they now emit `DeprecationWarning` on every run:
> 
> * `newton/tests/test_sdf_compute.py:826,827,845,847,936,938`
> * `newton/tests/test_sdf_texture.py:621,626,627,631,634,635`
> 
> These use `model.shape_sdf_index` / `model.texture_sdf_data` / `model.texture_sdf_coarse_textures`. The PR already migrated `viewer.py` and `collide.py` to the `_`-prefixed members, so updating these tests to match would keep the repo off the path it just deprecated. It won't fail the suite today (warnings go to stderr, while `CheckOutput` captures stdout), so this is cleanup rather than a blocker.

Done

adenzler-nvidia:
Re: `Model._ensure_legacy_sdf_block_arrays` (`newton/_src/sim/model.py`, the `subgrid_size` readback).

Following up on my earlier suggestion to narrow this catch to `KeyError` — that advice was wrong, apologies. Indexing a field on a NumPy structured-array record (`_texture_sdf_data.numpy()[0]["subgrid_size"]`) raises `ValueError: no field of name ...` for a missing field, not `KeyError`:

```python
>>> a = np.zeros(1, dtype=np.dtype([("subgrid_size", "<i4")]))
>>> a[0]["nope"]
ValueError: no field of name nope
```

So the `except KeyError` never fires for the "missing `subgrid_size`" case its own log message describes — an unhandled `ValueError` would instead escape this deprecated property getter.

The deeper point: `subgrid_size` is a static field of the `TextureSDFData` `wp.struct`, so it can't actually be absent from valid `_texture_sdf_data`. The `try/except` + `subgrid_size=8` fallback is effectively dead defensive code. I'd just drop it:

```python
subgrid_size = 8
if self._texture_sdf_data is not None and len(self._texture_sdf_data) > 0:
    subgrid_size = int(self._texture_sdf_data.numpy()[0]["subgrid_size"])
```

If you'd rather keep the guard, catch the types that can actually occur so the warning fires instead of escaping:

```python
except (KeyError, ValueError, IndexError):
```

nvtw:
> Re: `Model._ensure_legacy_sdf_block_arrays` (`newton/_src/sim/model.py`, the `subgrid_size` readback).
> 
> Following up on my earlier suggestion to narrow this catch to `KeyError` — that advice was wrong, apologies. Indexing a field on a NumPy structured-array record (`_texture_sdf_data.numpy()[0]["subgrid_size"]`) raises `ValueError: no field of name ...` for a missing field, not `KeyError`:
> 
> ```python
> >>> a = np.zeros(1, dtype=np.dtype([("subgrid_size", "<i4")]))
> >>> a[0]["nope"]
> ValueError: no field of name nope
> ```
> 
> So the `except KeyError` never fires for the "missing `subgrid_size`" case its own log message describes — an unhandled `ValueError` would instead escape this deprecated property getter.
> 
> The deeper point: `subgrid_size` is a static field of the `TextureSDFData` `wp.struct`, so it can't actually be absent from valid `_texture_sdf_data`. The `try/except` + `subgrid_size=8` fallback is effectively dead defensive code. I'd just drop it:
> 
> ```python
> subgrid_size = 8
> if self._texture_sdf_data is not None and len(self._texture_sdf_data) > 0:
>     subgrid_size = int(self._texture_sdf_data.numpy()[0]["subgrid_size"])
> ```
> 
> If you'd rather keep the guard, catch the types that can actually occur so the warning fires instead of escaping:
> 
> ```python
> except (KeyError, ValueError, IndexError):
> ```

Done

CHANGELOG.md
```diff
@@ -175,9 +175,13 @@
...
+- Deprecate the public attribute names `Model.shape_sdf_index`, `Model.texture_sdf_data`, `Model.texture_sdf_coarse_textures`, `Model.texture_sdf_subgrid_textures`, `Model.texture_sdf_subgrid_start_slots` in favor of their underscore-prefixed private counterparts (`Model._shape_sdf_index`, etc.); the public aliases keep working for one release cycle and will be removed in Newton 1.5
 
 ### Fixed
 
+- Fix hydroelastic SDF contact surfaces dropping the central region under deep interpenetration. The broadphase used to skip subgrids whose centers were deeper than the SDF narrow band, leaving a hole in the contact patch when overlap exceeded the narrow-band thickness. Broadphase now visits every subgrid in the SDF coarse grid (block coordinates are derived arithmetically from per-shape SDF coarse-texture dimensions); sampling at far-inside locations is correct because the coarse SDF is dense and accurate everywhere. On-disk SDF caches written by earlier versions are transparently re-cooked on first load (`_sdf_cache.CACHE_FORMAT_VERSION` bumped to `2`)
```
adenzler-nvidia: The changelog entries are in the wrong section now.

adenzler-nvidia (changes requested):
1 final changelog issue

adenzler-nvidia (approved):
Thanks!

PR #2703, opened 2026-05-05
Kamino - FK solver performance improvements

Guirec-Maloisel:
## Description

This improves the performance of the FK solver, both the CPU precomputation (in particular, sparsity pattern generation and symbolic factorization) and the GPU runtime.

For precomputations, utilities are added that group worlds into equivalence classes with respect to discrete FK structure (i.e. worlds with the same joint types between the same bodies are equivalent in terms of matrix sparsity pattern). Then all precomputations that depend on discrete information only are done only once per class, leading to severe performance improvements on large models / large world counts (with speedup basically proportional to the number of identical worlds).

For general runtime, the tile-based kernels have been optimized. The kernels with 2d tiles computing matrix-matrix or matrix-vector products with the jacobians now skip tiles that are known to be zero. The tile sizes and block sizes have also been optimized.

Resolves vastsoun/newton#314

## Checklist

## Test plan

On an example used as benchmark (middle-scale animatronic character on 1024 worlds and randomized poses), this yields an overall 3x speedup for the FK solve time, and a speedup roughly proportional to the number of identical worlds for the precomputations.

## Summary by CodeRabbit

* **New Features**
  * Equivalence-class grouping for worlds and a compute_fk_equivalence_classes helper.
  * New 1D/2D tile kernel factories and a per-class tile_sparsity_pattern to skip zero tiles.
  * capture_sparsity_pattern supports per-equivalence-class ordering and broadcasts results across class members.

* **Improvements**
  * Per-class reuse of sparsity and factorization ordering to reduce redundant work.
  * Removed global TILE_SIZE_CTS/TILE_SIZE_VRS config fields.

* **Tests**
  * Added unit tests for world-equivalence grouping.

### DIFF
--- a/newton/_src/solvers/kamino/_src/linalg/factorize/llt_blocked_semi_sparse.py
+++ b/newton/_src/solvers/kamino/_src/linalg/factorize/llt_blocked_semi_sparse.py
@@ -426,20 +426,30 @@ class SemiSparseBlockCholeskySolverBatched:
         self,
         A: np.ndarray,  # 3D array (batch_size, n, n)
         A_reorder_size: np.ndarray,  # 1D array (batch_size)
+        eq_classes: list[list[int]] | None = None,
     ):
         """
         Captures sparsity pattern and computes fill-reducing ordering for batched matrices.
 
         Args:
-            A: Input SPD matrices of shape (batch_size, n, n), as float arrays or directly as binary 0/1 matrices
-            indicating the sparsity pattern (float arrays will be converted to binary automatically).
-            A_reorder_size: Per-batch size of top-left block to reorder for sparsity
+        A (np.ndarray):
+            Input SPD matrices as float arrays or directly as binary 0/1 matrices indicating the sparsity
+            pattern (float arrays are converted to binary automatically).
+            Shape of (batch_size, n, n); or of (num_classes, n, n) if eq_classes is provided.
+        A_reorder_size (np.ndarray):
+            Size of the active top-left block per matrix, to reorder for sparsity.
+            Shape of (batch_size,); or of (num_classes,) if eq_classes is provided.
+        eq_classes (list[list[int]], optional):
+            List of list of matrix indices (along the batch dimension) that form equivalence classes with
+            the same sparsity pattern.
+            If provided, only one sparsity pattern per class should be provided in `A`.
+            The computed ordering will then be broadcast to all matrices in each class.
 
         Computes Cuthill-McKee ordering on top-left block, analyzes symbolic Cholesky factorization,
         and stores tile-level sparsity patterns. Tiles beyond A_reorder_size are treated as dense.
         """
 
-        batch_size = A.shape[0]
+        batch_size = self.num_batches
 
         # Convert to binary
         A = to_binary_matrix(A)
@@ -450,25 +460,28 @@ class SemiSparseBlockCholeskySolverBatched:
         inverse_orderings = np.zeros((batch_size, self.max_num_equations), dtype=np.int32)
         L_tile_patterns = np.zeros((batch_size, self.num_tiles, self.num_tiles), dtype=np.int32)
 
-        # Process each batch independently
-        for batch_id in range(batch_size):
+        # Process each equivalence class independently
+        if eq_classes is None:
+            eq_classes = [[i] for i in range(batch_size)]
+        assert len(A) == len(eq_classes)
+        assert len(A_reorder_size) == len(eq_classes)
+        for class_id, eq_class in enumerate(eq_classes):
             # Call cuthill_mckee_ordering on the binary version of A and store both orderings
-            reorder_size = A_reorder_size[batch_id]
-            ordering = cuthill_mckee_ordering(A[batch_id, :reorder_size, :reorder_size])
+            reorder_size = A_reorder_size[class_id]
+            ordering = cuthill_mckee_ordering(A[class_id, :reorder_size, :reorder_size])
 
             # Append sequential indices for remaining rows/cols
             remaining_indices = np.arange(reorder_size, self.max_num_equations)
             ordering = np.concatenate([ordering, remaining_indices])
-            orderings[batch_id] = ordering
 
+            # Compute inverse ordering
             inverse_ordering = compute_inverse_ordering(ordering)
-            inverse_orderings[batch_id] = inverse_ordering
 
             # Reorder A and then extract the sparsity patterns
             if self.enable_reordering:
-                A_reordered = A[batch_id][ordering][:, ordering]
+                A_reordered = A[class_id][ordering][:, ordering]
             else:
-                A_reordered = A[batch_id]
+                A_reordered = A[class_id]
 
             L_tile_pattern_np = symbolic_cholesky_dense(A_reordered, self.block_size)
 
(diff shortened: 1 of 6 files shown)


newton/_src/solvers/kamino/_src/solvers/fk/solver.py
```diff
@@ -1287,16 +1329,16 @@ def _solve_for_body_velocities(
         else:
             wp.launch_tiled(
                 self._eval_jacobian_T_jacobian_kernel,
-                dim=(self.num_worlds, self.num_tiles_states, self.num_tiles_states),
-                inputs=[self.jacobian, world_mask, self.lhs],
-                block_dim=64,
+                dim=(self.num_worlds, self.num_tiles_cts_2d, self.num_tiles_cts_2d),
+                inputs=[self.jacobian, self.tile_sparsity_pattern, world_mask, self.lhs],
+                block_dim=32,
```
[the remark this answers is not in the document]
Guirec-Maloisel: Good catch, I pushed a fix as suggested

newton/_src/solvers/kamino/_src/utils/world_equivalence.py
```diff
@@ -0,0 +1,341 @@
...
+    def class_ids_to_classes(class_ids, num_classes):
+        classes = [[] for _ in range(num_classes)]
+        for i in range(len(class_ids)):
+            classes[class_ids[i]].append(i)
+        return classes
+
+    # Dimensions
+    num_worlds = next(iter(signatures)).num_worlds
+    assert all(sig.num_worlds == num_worlds for sig in signatures)
+    device = next(iter(signatures)).data.device
```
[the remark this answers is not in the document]
Guirec-Maloisel: Good catch, I pushed a fix as suggested

newton/_src/solvers/kamino/_src/utils/world_equivalence.py
```diff
@@ -0,0 +1,345 @@
...
+            self.world_size = wp.ones(shape=self.num_worlds, device=self.data.device, dtype=wp.int32)
+        else:
+            assert self.world_size.size == self.num_worlds, "invalid size for world_size"
+
+        if self.world_delta is None:
+            self.world_delta = wp.zeros(shape=self.num_worlds, device=self.data.device, dtype=wp.int32)
+        else:
+            assert self.world_delta.size == self.num_worlds or self.world_delta.size == self.num_worlds + 1, (
+                "invalid size for world_delta"
+            )
```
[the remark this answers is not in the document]
Guirec-Maloisel: I'll leave the code as-is, as this helper module is only there to be used internally, and will typically be called once in init/finalize operations (as in FK). A mix of devices there is therefore an improbably corner case with no clear practical use, so I'd like to keep the code simple and leave it to the caller to not mix devices for now.

newton/_src/solvers/kamino/_src/utils/world_equivalence.py
```diff
@@ -0,0 +1,345 @@
...
+###
+# Kernels
+###
+
+
+@wp.func
+def comparison_value(value: int32, delta: int32, ignore_negative: wp.bool):
+    if ignore_negative and value < 0:
+        return -1
+    return value - delta
```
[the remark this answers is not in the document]
Guirec-Maloisel: The reason for introducing delta is to be able to convert global indices to within-world indices, and only compare them within the world, so indices will always be non-negative there (unless the indices are invalid). I can't think of a use case where there would be both negative relative values and sentinel values currently, so I'd like to keep the code simple, and only complexify it if there is a new type of discrete information that we'd like to group worlds by comparing, and that could not be correctly handled by the current code.

newton/_src/solvers/kamino/_src/utils/world_equivalence.py
vastsoun, line 1: Would it be possible to come up with a single-word name for this since we're trying to avoid multiname python files?

vastsoun (changes requested):
Great work @Guirec-Maloisel. I only have a minor comment, otherwise this seems great. 
I also think it would be meaningful to add unit tests for the `DiscreteSignature` class if possible.

newton/_src/solvers/kamino/_src/utils/world_equivalence.py
Guirec-Maloisel: I wasn't aware of that convention, can I ask why? As usual I'm an advocate of clarity over compactness :) If you have a good single-word suggestion I'm happy to use it though.

vastsoun: What about putting it in `kamino/_src/linalg/utils/equivalences.py`?

Guirec-Maloisel: Hmm I think it's quite a bit more vague, and also not tied directly to linear algebra (I tried to keep it general for other places where some worlds might be equivalent with respect to some operations, e.g. some of the post-processing in the builder)

vastsoun: I see. Ok then let's proceed as is. My second point still stands though, in terms of adding some unit tests for it.

Guirec-Maloisel: Yep that's in the works!

newton/_src/solvers/kamino/_src/linalg/factorize/llt_blocked_semi_sparse.py
```diff
@@ -426,20 +426,30 @@ def capture_sparsity_pattern(
...
+        eq_classes: list[list[int]] | None = None,
     ):
         """
         Captures sparsity pattern and computes fill-reducing ordering for batched matrices.
 
         Args:
-            A: Input SPD matrices of shape (batch_size, n, n), as float arrays or directly as binary 0/1 matrices
-            indicating the sparsity pattern (float arrays will be converted to binary automatically).
-            A_reorder_size: Per-batch size of top-left block to reorder for sparsity
+        A (np.ndarray):
```
chschuma-disney, line 435: As far as I know from the style guidelines, types should not be repeated in the docstring.

newton/_src/solvers/kamino/tests/test_utils_world_equivalence.py
```diff
@@ -0,0 +1,111 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 The Newton Developers
```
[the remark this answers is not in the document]
Guirec-Maloisel: Good catch, I pushed a fix

newton/_src/solvers/kamino/_src/linalg/factorize/llt_blocked_semi_sparse.py
```diff
@@ -450,25 +460,28 @@ def capture_sparsity_pattern(
...
         L_tile_patterns = np.zeros((batch_size, self.num_tiles, self.num_tiles), dtype=np.int32)
 
-        # Process each batch independently
-        for batch_id in range(batch_size):
+        # Process each equivalence class independently
+        if eq_classes is None:
+            eq_classes = [[i] for i in range(batch_size)]
+        assert len(A) == len(eq_classes)
+        assert len(A_reorder_size) == len(eq_classes)
+        for class_id, eq_class in enumerate(eq_classes):
```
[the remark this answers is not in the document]
Guirec-Maloisel: These are some sanity checks to flag early some mis-use of the class, but it is not this class's responsibility to check its input (the requirement that the indices form equivalence classes is listed in the docstring, so it is the caller's responsibility).

vastsoun (approved):
LGTM

chschuma-disney (approved):
Great work, LGTM. Just a small comment, but might not need to be addressed.

newton/_src/solvers/kamino/_src/linalg/factorize/llt_blocked_semi_sparse.py
Guirec-Maloisel: Good point, I'm planning a cleaning pass on the FK solver overall so I think I'll include that then (otherwise each new commit in this PR re-triggers a pass of CI and code rabbit comments).

Issue #2612, opened 2026-04-28
[REQ] Support nonlinear pressure-depth laws for hydroelastic contact


jessicayin:
### Description

I would like Newton's hydroelastic contact model to support nonlinear per-shape pressure-depth laws, while preserving the current linear `kh` behavior as the default.

Today, hydroelastic contact is effectively parameterized by one scalar stiffness per shape:

```text
p_i(d) = k_i * d
```

where `d` is the local compression/depth implied by the SDF field, and the pairwise contact surface is extracted from a linear weighted SDF difference. This is a good default, but it makes it difficult to calibrate hydroelastic contact against real soft materials whose compression response is not well represented by one constant `kh`.

Proposed feature: allow each hydroelastic shape to provide a scalar monotone pressure law:

```text
p_i(d) = f_i(d)
```

and extract the hydroelastic contact surface from pressure equilibrium:

```text
f_a(d_a) - f_b(d_b) = 0
```

instead of only:

```text
k_a * d_a - k_b * d_b = 0
```

This is intentionally narrower than a full nonlinear solid mechanics / FEM request. This request does not ask for tensor stress-strain fields, anisotropy, hysteresis, viscoelasticity, or stateful material history, just a nonlinear scalar hydroelastic pressure model evaluated from SDF depth.

Useful built-in pressure-law options would be:

```text
linear:           p(d) = k * d                         # current behavior
power_law:        p(d) = k * d^n
piecewise_linear: p(d) from calibrated (depth, pressure) samples
table:            p(d) from a sampled monotone lookup table
```

For solver integration, it would also be useful for each law to expose a local tangent:

```text
dp/dd
```

so generated hydroelastic contacts can use a local area-scaled tangent stiffness rather than a single global `area * k_eff`. For example, the effective pair tangent could use the same harmonic-mean idea, but applied to the local slopes of the two pressure laws.

A possible API shape:

```python
cfg = builder.ShapeConfig(
    is_hydroelastic=True,
    sdf_max_resolution=64,
    hydroelastic_pressure_law=newton.HydroelasticPressureLaw.piecewise_linear(
        depths=[0.0, 0.001, 0.003, 0.006],
        pressures=[0.0, 1.0e4, 8.0e4, 3.0e5],
    ),
)
```

or, keeping compatibility with the existing `kh` API:

```python
cfg.configure_sdf(
    max_resolution=64,
    is_hydroelastic=True,
    kh=1.0e8,  # still supported; equivalent to linear law
)
```

[...]

PR #2705, opened 2026-05-05
Add depth to pressure callback wp.func support for hydroelastic contacts

nvtw:
## Description
Should address https://github.com/newton-physics/newton/issues/2612.
Adds a user-supplied pressure-callback API to the hydroelastic SDF contact pipeline. The contact patch is now defined as the iso-pressure surface `p_a == p_b`, where `p = pressure_func(signed_depth, shape_idx, data)` is a Warp `@wp.func` plus a `@wp.struct` of per-shape state supplied through `HydroelasticSDF.Config.pressure_func` / `pressure_data`. The previous hard-coded linear law `pressure = -kh * signed_depth` is kept as the default (`linear_pressure` + `LinearPressureData` backed by `Model.shape_material_kh`), so behavior is unchanged when no callback is provided.
The subblock prune in `count_iso_voxels_block` now uses interval arithmetic on `pressure_func` (relying on monotone-non-increasing-in-`signed_depth` and the 1-Lipschitz spatial bound on `phi`) instead of the linear-stiffness shortcut, and `mc_iterate_voxel_vertices` evaluates `pressure_func` at every corner so corner values stay continuous across the patch boundary and the marching-cubes edge interpolation is well-defined. Both kernels are produced through factory functions specialized per `pressure_func` so the user callback is statically inlined.

## Checklist

## Test plan
Existing hydroelastic regression tests cover the default (linear) pressure path:
```
uv run --extra dev -m newton.tests -k hydro
```
The updated `example_nut_bolt_hydro.py` exercises the user-callback pathway end-to-end by re-implementing the linear law explicitly; running it should produce contact patches identical to the previous default:
```
uv run -m newton.examples nut_bolt_hydro
```

## New feature / API change
```python
import warp as wp
import newton

@wp.struct
class CubicPressureData:
    shape_kh: wp.array[wp.float32]

@wp.func
def cubic_pressure(signed_depth: wp.float32, shape_idx: wp.int32, data: CubicPressureData) -> wp.float32:
    # Stiffer-with-depth response; must be defined for any signed_depth and
    # monotone non-increasing in signed_depth.
    d = -signed_depth
    return data.shape_kh[shape_idx] * d * d * d

model = builder.finalize()

pressure_data = CubicPressureData()
pressure_data.shape_kh = model.shape_material_kh

config = newton.geometry.HydroelasticSDF.Config(
    pressure_func=cubic_pressure,
    pressure_data=pressure_data,
)

pipeline = newton.CollisionPipeline(model, sdf_hydroelastic_config=config)
```

## Summary by CodeRabbit
* **New Features**
  * Hydroelastic contact generation now supports customizable per-shape pressure laws through user-provided callbacks
  * Default linear pressure behavior preserved for backward compatibility
  * Contact force computation driven by pressure equality surface
* **Examples**
  * Updated hydroelastic contact example demonstrating custom pressure law implementation and SDF caching

## Summary by CodeRabbit

* **New Features**
  * Support for user-defined hydroelastic pressure laws that control iso-surface extraction and contact stiffness; default linear law remains when unset.
  * Pressure is threaded through iso-pruning, surface extraction, contact generation, reduction, and export for consistent behavior.

* **Examples**
  * Example updated to persist cooked SDF data across runs and demonstrate a per-shape pressure callback.

* **Tests**
  * Added tests validating custom pressure callbacks and interactions with the default law.

* **Documentation**
  * New "Custom pressure laws" guidance and changelog entry.


docs/concepts/collisions.rst
lenroe, line 1: I think this is a great feature, implementation seems clear (although I did not check in detail yet). 
Just to clarify: The custom pressure laws are valid when considering the isosurface placements, and likely also give correct results for the unreduced contact case. But contact reduction and mujoco stiffness scaling also assume linear-in-depth force laws, which I think would be much more involved to make consistent. Is this feature only meant to be used in the unreduced case?

nvtw: I was not sure how to tackle the code where moment matching and the transfer to mjwarp contact happens. It feels a bit like I would apply the nonlinearity twice when I apply it again on top of the shifted pressure surface. Let me double check. If you have a clear idea about what needs to be changed, please sketch it somewhere in this MR.

nvtw: I tried to apply the user callback more broadly - but I had to adjust some material property combination formulas. Not yet 100% sure if this is all correct.

jessicayin:
@nvtw @mzamoramora-nvidia @lenroe thanks all for your work on this great feature for hydroelastic contact! It is much appreciated and truly creates a lot of value for the hydroelastic model for tactile sim research. From my experience testing this:

### Summary
The nonlinear implementation works well and measurably improves the fit for real2sim, resulting in lower force RMSE across the entire real-world dataset. I tested 4 different pressure laws: power, exponential, quadratic, and piecewise linear and ran a few simple tests. Below I have included some notes from Claude from the testing:

### Clipping to zero in the non-contact region

Our first callback followed the example in the PR:

```python
@wp.func
def power_pressure(signed_depth, shape_idx, data):
    depth = wp.max(-signed_depth, 0.0)          # <-- clips non-contact region to 0
    x = wp.max(depth / data.depth_ref_m, 0.0)
    return data.shape_kh[shape_idx] * data.depth_ref_m * wp.pow(x, data.exponent)
```

At `exponent=1` this should be identical to the built-in linear law, but our calibration
results diverged. The attached repro script (`newton-03-clamp-pitfall_repro.py`) isolates
the issue using a simple box-on-slab scene:

newton-03-clamp-pitfall


[newton-03-clamp-pitfall_repro.py](https://github.com/user-attachments/files/28657628/newton-03-clamp-pitfall_repro.py) - a self-contained box-into-slab scene that reproduces the clipping issue. Compares the built-in linear law, an unclamped custom law at `exponent=1`, and a clipped custom law at `exponent=1`, across equal and 1:10 stiffness ratios. Runs headless on one GPU in ~1–2 minutes and writes `newton-03-clamp-pitfall.png`.

- **When both surfaces have the same stiffness:** the clipped callback matches the
  built-in to within 1% — no problem.
- **When the surfaces have different stiffnesses (e.g. 1:10):** the clipped callback
  produces near-zero contact force at small penetrations where the built-in law already
  gives substantial force, only recovering toward the correct value at larger depths.

**Why this happens:** Newton finds the contact surface by looking for the depth at which
the two pressure functions balance (`p_a = p_b`), evaluated on *both sides* of the
contact boundary — including a thin region just outside contact. The built-in law
continues smoothly (with a negative value) into that outside region, so the balance point
is well-defined. A clipped law is flat at zero there, so the balance point moves or
disappears. When the two surfaces have different stiffnesses, the balance point sits near
the stiffer surface's boundary — exactly where the clip fires.

**Fix:** extend the law linearly into the non-contact region rather than clipping it.
This then reproduces the built-in exactly at `exponent=1`:

```python
@wp.func
def power_pressure(signed_depth, shape_idx, data):
    kh = data.shape_kh[shape_idx]
    if signed_depth >= 0.0:               # not in contact — continue linearly
        return -kh * signed_depth
    depth = -signed_depth
    return kh * data.depth_ref_m * wp.pow(depth / data.depth_ref_m, data.exponent)
```

**Suggested documentation change:** the docstring says the law must be "monotone
non-decreasing," which the clipped version satisfies — but that condition alone is not
enough to avoid this issue. Worth adding a note that the law should be **smooth and
non-zero through the thin region just outside contact**, not clipped to zero there.
Updating the example callback would help avoid this footgun for future users.

## API notes

- **Indexing works cleanly.** Passing `model.shape_material_kh` as `pressure_data` and
  reading it by `shape_idx` in the callback required no remapping — the indices match.
- **Setup ordering issue.** The hydroelastic config is built before `model.finalize()`,
  but `pressure_data` needs stiffness values that only exist after finalization. We worked
  around it, but a post-finalize hook or a note in the docs would smooth this for new
  users.
- **Redundant parameter.** For a power law, `gain` and `kh` are redundant — only their
  product enters the force calculation, so fitting both gives a degenerate result. We
  dropped `gain`. Worth a doc note.

adenzler-nvidia:
Trying to understand the tradeoffs here before it lands — whether the callback is the contract we want, or whether a per-shape table generalizes better.

For context, the issue thread sketched three paths: (1) a per-shape monotone `(depth, pressure)` table, (2) a multi-channel SDF texture, and (3) user callbacks. @mzamoramora-nvidia called (1) "probably the simplest first step" that "keeps the current kh path as the default fast path," and flagged (3) as most flexible but with "runtime/API complexity, especially on Warp/GPU, plus harder validation for monotonicity, finite slopes, units, and solver compatibility." This PR implements (3), and a couple of those flagged downsides seem to surface in the current shape. The framing I'd keep either way is solid — iso-pressure surface `p_a == p_b`, linear law as the default, statically inlined.

Two things I'm trying to square. First, the law is attached globally on `HydroelasticSDF.Config`, but it reads like a per-shape material property — the default already reaches into `Model.shape_material_kh`, which is per-shape from `cfg.kh`. So the linear law is per-shape today, and the nonlinear path lifts that up to a global pipeline setting. With one global `pressure_func` you can vary per-shape coefficients via `data.shape_kh[shape_idx]` but not the functional form: two shapes in a pair with genuinely different laws need the user to hand-write an `if shape_idx ...` dispatch inside their own func (the `cubic_pressure` example applies one form to every shape). It also forces the lifecycle wart where the docs have to tell users to build the config after `finalize()` and wire `shape_material_kh` by hand. Was per-shape attachment (on `ShapeConfig`, next to `kh`) considered and rejected, or just heavier for a first cut?

Second, would a per-shape table give the same flexibility without those downsides? For the scope here — scalar, monotone, depth-only — a sampled `(depth, pressure, slope)` table is the universal representation: power law, piecewise, vendor curve, even an arbitrary user lambda all sample to the same per-shape table at build time. That keeps the "no built-in type zoo" benefit while resolving the exact concerns flagged above — monotonicity, finite slope, units, domain become build-time validations on data, and the `dp/dd` tangent is well-defined by construction. The runtime stays one fixed law-agnostic lookup, no per-law specialization, and the linear default becomes the same mechanism rather than a separate path; the domain also pins naturally to the SDF narrow band (#2611).

The real pro of the callback I can see is exact analytic evaluation — no table resolution or memory. For a monotone compliance curve the approximation seems fine (and `piecewise_linear`/`table` were two of the four forms the issue asked for), but if that's the driving consideration it'd be good to know. Not blocking — mainly trying to understand whether the global callback is what we want to commit to, or whether a per-shape table generalizes better with the callback as an optional build-time way to author one.

nvtw:
You could implement a big switch inside the user callback based on shape index and then return a different pressure law value per shape. Also you can bring arbitrary data arrays into that callback via pressure_data. This allows to implement a look up table by the user if it's their preferred solution. The main motivation for the callback approach was the generality. It's a feature for power users, so I prefer to give them all the flexibility that they can get and do little to no hand holding - it's the users responsibility that the callback is physically meaningful.
The callback also follows the already established pattern of the contact writer callback from the collision pipeline.

adenzler-nvidia:
The limitation of the big switch is usually that it's quite bad for codegen - you will allocate worst-case registers for this in a kernel that is probably already register-starved. But I do see that it's an option. I'm wondering whether we need to make some more detailed experiments to figure out the cost of each option.

I do agree that striving for full generality is cool.

jcarius-nv:

CHANGELOG.md
```diff
@@ -163,6 +163,7 @@
...
 - Add `Viewer.log_image()` for displaying single or batched images in `ViewerGL`; other backends inherit a no-op. Also add `SensorTiledCamera.utils.to_rgba_from_color()`, `to_rgba_from_normal()`, `to_rgba_from_depth()`, and `to_rgba_from_shape_index()` (hash palette or caller-provided RGB lookup) adapters producing output consumable by `log_image()`.
 - Add on-disk caching of cooked texture-based SDFs via the new `cache_dir` argument on `SDF.create_from_mesh` and `Mesh.build_sdf`. Cached entries are content-addressed by mesh and build parameters, written atomically as a single uncompressed `.npz`, and versioned via `CACHE_FORMAT_VERSION` so format changes invalidate stale caches transparently
+- Add user-defined pressure laws to hydroelastic SDF contact via `HydroelasticSDF.Config.pressure_func` (a `@wp.func` mapping `(signed_depth, shape_idx, data) -> pressure`) and `pressure_data` (a `@wp.struct` carrying per-shape state). The contact patch is the iso-pressure surface `p_a == p_b`; the default linear law `pressure = -kh * signed_depth` is preserved when no callback is supplied.
```
adenzler-nvidia: please move the changelog entry to unreleased

adenzler-nvidia (dismissed):
LGTM!

adenzler-nvidia (approved):

PR #2710, opened 2026-05-05
Fix VRAM leak when resetting examples

### LINKED ISSUE #2700: [BUG] Reset feature caused memory leak in diffsim_bear example
### Bug Description

Not sure if it is all examples or only a few, but it looks like reset the `diffsim_bear` examples leads to large increase in memory.

For reference, my RTX4080 is OOM after 2 reset.

Run
```
uv run --extra examples -m newton.examples diffsim_bear
```
Run the command `nvidia-smi`, in my case the process use ~6Gb of memory, use the "Reset" feature on the top left panel. The memory usage jump to 12Gb.

mzamoramora-nvidia:
## Description

Closes #2700.

`_ExampleBrowser.reset` rebuilds the current example by calling its constructor while the previous instance is still referenced — by `run()`'s local and, when the example file is invoked directly, by the script's `__main__` globals. Both old and new live in VRAM during `__init__`, peaking at 2× the example's footprint. For `diffsim_bear` (~6 GB of grad-enabled sim states + captured forward/backward CUDA graph), a single reset hits ~12 GB and the second reset OOMs on a 16 GB GPU.

Two changes:

1. `newton/examples/diffsim/example_diffsim_bear.py`: inline `Example(viewer, args)` into the `newton.examples.run(...)` call, so the example isn't bound at module level.
2. `newton/examples/__init__.py`: in `run()`'s reset branch, set the local `example = None` before calling `_ExampleBrowser.reset(example_class)`, so refcount-based collection actually runs.

The two changes together make the old example's destructors run before the new one starts allocating, so peak VRAM stays at 1×.

## Other examples with the same latent issue

The same `example = Example(viewer, args); newton.examples.run(example, args)` pattern appears in 63 other examples; they have the same latent leak but don't trip it because their per-instance VRAM is small. Happy to include all the other examples in this PR once the fix is confirmed by a reviewer. Distribution:

- `basic/`: 10
- `robot/`: 8
- `mpm/`: 8
- `cloth/`: 6
- `kamino/`: 5
- `contacts/`, `diffsim/`, `ik/`, `selection/`: 4 each
- `cable/`, `sensors/`: 3 each
- `multiphysics/`, `softbody/`: 2 each

The mechanical fix is identical in each: replace
```python
newton.examples.run(example, args)
```
with
```python
newton.examples.run(Example(viewer, args), args)
```

## Checklist

## Test plan

Manual: with the GUI viewer, click Reset multiple times and watch `nvidia-smi`.

```bash
uv run --extra examples -m newton.examples diffsim_bear
```

Before fix: VRAM steps from ~6 GB → ~12 GB on first reset, OOM on a 16 GB GPU on the second. After fix: VRAM stays roughly flat across resets (verified across 5 consecutive resets on an RTX A6000, total drift < 150 MB).

## Bug fix

**Steps to reproduce:**

1. `uv run --extra examples -m newton.examples diffsim_bear`
2. Note initial process VRAM in `nvidia-smi` (~6 GB).
3. Click "Reset" in the top-left panel.
4. Observe VRAM jumps to ~12 GB.
5. Click "Reset" again — OOM on GPUs with ≤16 GB VRAM.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a VRAM memory leak when resetting examples that allocate large GPU state.

* **Documentation**
  * Clarified reset guidance and changelog to instruct dropping old example references before re-creation.

* **Refactor**
  * Simplified many example startup entrypoints to inline instance creation for cleaner wiring; behavior unchanged.

### DIFF
--- a/newton/examples/__init__.py
+++ b/newton/examples/__init__.py
@@ -239,7 +239,11 @@ class _ExampleBrowser:
         return example, type(example)
 
     def reset(self, example_class):
-        """Reset the current example by re-creating it. Returns the new example or None."""
+        """Reset the current example by re-creating it. Returns the new example or None.
+
+        The caller must drop its reference to the old example before calling
+        this method.
+        """
         self._reset_requested = False
         self.viewer.clear_model()
         try:
@@ -280,6 +284,8 @@ def run(example, args):
             continue
 
         if browser is not None and browser._reset_requested:
+            example_class = type(example)
+            example = None
             example = browser.reset(example_class)
             continue
 
--- a/newton/examples/__init__.py
+++ b/newton/examples/__init__.py
@@ -284,7 +284,6 @@ def run(example, args):
             continue
 
         if browser is not None and browser._reset_requested:
-            example_class = type(example)
             example = None
             example = browser.reset(example_class)
             continue
--- a/newton/examples/__init__.py
+++ b/newton/examples/__init__.py
@@ -284,6 +284,8 @@ def run(example, args):
             continue
 
         if browser is not None and browser._reset_requested:
+            # Drop our reference so the old example is freed before reset()
+            # builds the replacement; otherwise both coexist in VRAM.
             example = None
             example = browser.reset(example_class)
             continue
--- a/newton/examples/__init__.py
+++ b/newton/examples/__init__.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import ast
+import gc
 import importlib
 import os
 import warnings
@@ -284,9 +285,12 @@ def run(example, args):
             continue
 
         if browser is not None and browser._reset_requested:
-            # Drop our reference so the old example is freed before reset()
-            # builds the replacement; otherwise both coexist in VRAM.
+            # Drop our reference and force cycle collection so the old
+            # example's destructors finish before reset() enters the new
+            # CUDA graph capture; otherwise late texture/array __del__
+            # calls could fire mid-capture and CUDA rejects them.
             example = None
+            gc.collect()
             example = browser.reset(example_class)
             continue
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,7 @@
 - Fix MPR convergence failure on large and extreme-aspect-ratio mesh triangles by projecting the starting point onto the triangle nearest the convex center
 - Fix O(W²·S²) memory explosion in `CollisionPipeline` shape-pair buffer allocation for NXN and SAP broad phase modes by computing per-world pair counts instead of a global N²
 - Fix `SensorRaycast` ignoring `PLANE` geometry
+- Fix VRAM leak when resetting examples that allocate large GPU state (e.g. `diffsim_bear`)
 - Fix `SensorRaycast` and viewer picking ignoring `HFIELD` (heightfield) geometry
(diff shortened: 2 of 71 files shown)


newton/examples/__init__.py
```diff
@@ -280,6 +284,7 @@ def run(example, args):
             continue
 
         if browser is not None and browser._reset_requested:
+            example = None
```
vreutskyy, line 292: Maybe add a comment? Minor though.

vreutskyy (dismissed):
Gooks good to me

newton/examples/__init__.py
mzamoramora-nvidia: Addressed in https://github.com/newton-physics/newton/pull/2710/commits/117b216ca6a7628f1669390f425b97acf28db431

vreutskyy (dismissed):

vreutskyy (dismissed):

Kenny-Vilella:
@mzamoramora-nvidia  There are a number of examples that you did not changed.
Could you check them as well?
The list:
-   `recording`
-   `cable_bundle_hysteresis`
-   `cloth_hanging`
-   `cloth_rollers`
-   `contacts_rj45_plug`
-   `diffsim_ball`
-   `diffsim_drone`
-   `robot_policy`

I only checked `cable_bundle_hysteresis` and could see the memory leak.

mzamoramora-nvidia:
@Kenny-Vilella thanks for the audit — addressed in 2f0567dd. Summary of how each was missed and what the fix looks like:

| Example | Why missed | Fix |
|---|---|---|
| `cable_bundle_hysteresis` | Multi-line constructor with kwargs | Inlined into `run()` |
| `cloth_hanging` | Multi-line constructor with kwargs | Inlined |
| `cloth_rollers` | `Example(viewer=viewer, args=args)` (kwarg form) | Inlined |
| `contacts_rj45_plug` | `Example(viewer)` (no `args` arg) | Inlined |
| `diffsim_drone` | Multi-line positional constructor | Inlined |
| `diffsim_ball` | `example.check_grad()` call between construction and `run()` | Wrapped in `_build()` helper |
| `robot_policy` | `load_policy_and_setup_tensors(example, ...)` between construction and `run()` | Wrapped in `_build()` helper |

The helper approach for the last two looks like:

```python
def _build():
    ex = Example(...)
    ex.setup_step()
    return ex

newton.examples.run(_build(), args)
```

When `_build()` returns its frame is destroyed, so the only surviving reference is the one passed into `run()` — equivalent to the inline pattern.

`example_recording.py` was intentionally left as-is: it doesn't call `newton.examples.run()` (custom frame loop), so there's no Reset path and no leak surface.

Kenny-Vilella:
Looks fine to me.
I let @vreutskyy  do the final approval.

mzamoramora-nvidia:
`example_robot_policy` has a lot of custom setup stuff that's a bit outside the usage patterns of other examples and needs some more work.

mzamoramora-nvidia:
Follow-up commit refactors `example_robot_policy` to fit the standard `(viewer, args)` constructor convention so reset works for it. Setup that previously lived in `__main__` (robot config lookup, asset download, YAML parse, joint mapping, policy load) is now done inside `Example.__init__`, and the parser definition moves onto `Example.create_parser()` so `_ExampleBrowser.reset()` can find it.

vreutskyy (approved):

PR #2720, opened 2026-05-05
Respect world_mask in SolverKamino joint state writes

aserifi:
## Description

`SolverKamino.reset(...)` accepts a `world_mask` parameter, and most operations inside `_reset_with_fk_solve` follow it. However, the function ended with four unmasked `wp.copy()` calls that overwrote `state_out.q_j`, `state_out.q_j_p`, and `state_out.dq_j` for all worlds regardless of the mask. This silently corrupted state for masked-out worlds, most notably `q_j_p`, the previous-coordinate cache used as the reference for revolute-joint TWOPI angle unwrapping. After a partial-mask reset, `q_j_p == q_j` for every world, which disables the angle correction on the next step and can cause revolute joints to glitch near pi.

This PR replaces the four unmasked copies with a single mask-aware kernel `_set_joint_state_of_select_worlds` that writes `q_j`, `q_j_p`, and (optionally) `dq_j` in one launch, gated per-joint on `world_mask[joint_wid]`. Joints whose world is masked out are left untouched. The kernel and its launcher `_set_joint_state_masked` are private helpers, so no new public API surface is added.

With this fix, `solver.reset(world_mask=zeros)` is a true per-world no-op.

## Checklist

## Test plan

- Extended `test_08_reset_to_joint_state` and `test_09_reset_to_actuator_state` to snapshot the pre-reset state and assert that worlds outside the mask retain their previous `q_j`, `q_j_p`, and `dq_j` values.
- Verified the regression test fails when the per-joint mask guard is removed (`AssertionError: state_out.q_j was modified for an unmasked world`) and passes when restored.
- All five reset tests (`test_05`–`test_09`) pass.
- `uvx pre-commit run -a` is clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Masked resets no longer overwrite joint state in worlds excluded by a partial mask, preserving unmasked joint values and restoring correct revolute-joint angle behavior after partial resets. Masking is now applied consistently during resets.

* **Tests**
  * Added tests that confirm worlds excluded by the mask remain unchanged during masked resets.

* **Changelog**
  * Added an unreleased entry documenting the masked-reset issue and its fix.

### DIFF
--- a/newton/_src/solvers/kamino/_src/kinematics/resets.py
+++ b/newton/_src/solvers/kamino/_src/kinematics/resets.py
@@ -213,7 +213,7 @@ def _reset_joint_state_of_select_worlds(
 @wp.kernel
 def _set_joint_state_of_select_worlds(
     # Inputs:
-    write_velocities: int32,
+    write_velocities: bool,
     world_mask: wp.array[int32],
     model_joint_wid: wp.array[int32],
     model_joint_coords_offset: wp.array[int32],
@@ -244,7 +244,7 @@ def _set_joint_state_of_select_worlds(
         dst_q_p[coords_offset + j] = v
 
     # Optionally write the joint's DoF velocities
-    if write_velocities != 0:
+    if write_velocities:
         dofs_offset = model_joint_dofs_offset[jid]
         num_dofs = model_joint_dofs_offset[jid + 1] - dofs_offset
         for j in range(num_dofs):
@@ -585,7 +585,7 @@ def set_joint_state_masked(
     either is ``None``, the velocity write is skipped and ``src_q`` / ``dst_q`` are
     passed as placeholders to satisfy the kernel signature.
     """
-    write_velocities = wp.int32(1) if (src_u is not None and dst_dq is not None) else wp.int32(0)
+    write_velocities = src_u is not None and dst_dq is not None
     _src_u = src_u if src_u is not None else src_q
     _dst_dq = dst_dq if dst_dq is not None else dst_q
     wp.launch(
--- a/newton/_src/solvers/kamino/_src/kinematics/resets.py
+++ b/newton/_src/solvers/kamino/_src/kinematics/resets.py
@@ -210,6 +210,47 @@ def _reset_joint_state_of_select_worlds(
 
 
 @wp.kernel
+def _set_joint_state_of_select_worlds(
+    # Inputs:
+    world_mask: wp.array[int32],
+    model_joint_wid: wp.array[int32],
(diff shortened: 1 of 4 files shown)


newton/_src/solvers/kamino/_src/kinematics/resets.py
```diff
@@ -209,6 +209,47 @@ def _reset_joint_state_of_select_worlds(
...
+@wp.kernel
+def _set_joint_state_of_select_worlds(
+    # Inputs:
+    world_mask: wp.array[int32],
+    model_joint_wid: wp.array[int32],
+    model_joint_coords_offset: wp.array[int32],
+    model_joint_dofs_offset: wp.array[int32],
+    src_q: wp.array[float32],
+    src_u: wp.array[float32],
+    write_velocities: int32,
```
vastsoun: I would prefer to place scalar constants first, as these we can consider more global than any of the more specific arrays.

newton/_src/solvers/kamino/_src/kinematics/resets.py
```diff
@@ -525,6 +566,48 @@ def reset_joint_constraint_reactions(
     )
 
 
+def _set_joint_state_masked(
```
vastsoun: Please prefer to remove the leading underscore from kernel launcher methods

vastsoun (changes requested):
Great catch @aserifi, and thanks a lot for the fix. Please see my minor review comments.

vastsoun (dismissed):
LGTM now. Thanks @aserifi.

newton/_src/solvers/kamino/_src/kinematics/resets.py
```diff
@@ -209,6 +210,47 @@ def _reset_joint_state_of_select_worlds(
         state_lambda_j[kinematic_cts_offset + j] = 0.0
 
 
+@wp.kernel
+def _set_joint_state_of_select_worlds(
+    # Inputs:
+    write_velocities: int32,
```
chschuma-disney: More out of curiosity: Is there a reason why this is not a boolean?

chschuma-disney (dismissed):
Looks good to me. Just had a small question, no changes necessary.

preist-nvidia:
I assume we need this in 1.2 for a lab integration, so marked 1.2 now.

newton/_src/solvers/kamino/_src/kinematics/resets.py
aserifi: No good reason, I changed it to bool now.

chschuma-disney (approved):

PR #2733, opened 2026-05-06
Fix log_image windows leaking across example browser switches

### LINKED ISSUE #2731: [BUG] The sensor_tiled_camera example does not behave properly with the example browser
### Bug Description

Two unexpected behaviors when using the example browser with the `sensor_tiled_camera` example.

1) The window showing the camera stays when changing example and we have to close it manually.
2) If we go back to the `sensor_tiled_camera` example after closing the window in another example, then the window will not open again.

Image

### Reproduction Script

```python
python -m newton.examples  sensor_tiled_camera
```

vreutskyy:
## Description

`ViewerGL.clear_model()` resets every model-dependent GL resource (objects,
lines, scalar buffers, array textures, side/free UI callbacks, etc.) but never
clears the image logger. The per-name texture/PBO entries and the `_selected`
field carry over between examples in the example browser, producing the two
behaviors reported in #2731:

1. The image window opened by the previous example stays visible after
   switching to one that does not call `log_image`.
2. Closing the window manually (which sets `_selected = None`) and re-entering
   the camera example does not reopen it: `ImageLogger.log` only auto-selects
   when *creating* a new entry, and the surviving `color` / `depth` / etc.
   entries skip that branch, leaving `_selected` at `None`.

Clearing the image logger inside `ViewerGL.clear_model()` discards the stale
GL resources on switch and lets the auto-select branch fire on re-entry.
Guarded with `getattr(...) is not None` because `clear_model()` is called
from `super().__init__()` before the logger is constructed.

Closes #2731

## Checklist

## Test plan

Regression test confirmed to fail without the fix and pass with it:

```
uv run --extra dev -m newton.tests -k test_clear_model_clears_image_logger
```

Full image-logger suite still passes:

```
uv run --extra dev -m newton.tests -k test_viewer_image_logger
```

Example browser smoke test for the camera example (switch + reset):

```
uv run python newton/tests/test_example_browser.py "sensor_tiled_camera" --frames 2
```

## Bug fix

1. `python -m newton.examples sensor_tiled_camera`
2. In the GL viewer's example browser, switch to any non-camera example.
   - Bug: the camera image window remains visible.
3. Close the camera image window via its `X` button, then switch to a
   non-camera example, then back to `sensor_tiled_camera`.
   - Bug: the image window does not reopen.

With this PR, switching examples removes the camera window and re-entering
the camera example reopens it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed image logger windows persisting or failing to re-open when switching examples by ensuring image logger state is cleared.

* **Tests**
  * Added a regression test to verify the image logger is cleared and that re-logging the same image restores selection.

* **Documentation**
  * Added a changelog entry describing the fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,7 @@
 - Fix `SensorTiledCamera` textured albedo output rendering flat colors when color and normal outputs are disabled
 - Fix `contacts_rj45_plug` example crashing on reset
 - Fix `SolverMuJoCo` dependency version-mismatch warning being silently skipped when Newton is installed from a wheel
+- Fix `ViewerGL.log_image()` windows persisting across example-browser switches and failing to re-open on re-entry after manual close, by clearing the image logger in `ViewerGL.clear_model()`
 - Fix `ModelBuilder.add_shape_heightfield` `scale` being ignored by narrow-phase collision and raycast
 - Fix `collision_filter_parent` silently ignored on joints to world (`parent=-1`); now honored for all `add_joint_*` methods, with `add_joint_fixed(parent=-1, ...)` defaulting to filter child shapes against world-static shapes
 - Fix multi-world `qfrc_actuator` conversion using the wrong body center of mass for worlds with `worldid > 0`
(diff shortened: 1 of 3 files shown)


jvonmuralt (approved):
looks good!

PR #2734, opened 2026-05-06
Refactor DelassusOperator attributes

aserifi:
## Description

Small cleanups in `DelassusOperator`:

**Guard `max(self._world_size)` against an empty list**, mirroring the pattern already used at `build()` (`max(self._world_dims) if self._world_dims else 0`). Not reachable today (`ModelBuilderKamino.finalize` rejects `num_worlds == 0`), but cheap consistency.

No public API change.

## Checklist

## Test plan

- All 23 tests in `newton/_src/solvers/kamino/tests/test_dynamics_delassus.py` pass.
- `uvx pre-commit run -a` is clean.

## Summary by CodeRabbit

* **Chores**
  * Refined internal per-world dimension and size bookkeeping and recomputed model totals for more consistent internal metadata; no public API changes.
* **Bug Fixes**
  * Made metadata computations robust for empty or edge-case solver states to prevent incorrect totals and related runtime issues.


Guirec-Maloisel:
Wait so the maxdims[i] * maxdims[i] overflows? Wouldn't that mean that there are 65k constraints per world (2^16 basically)?

Also, even with this fix, if one world already overflows then the size of the full flattened array of all Delassus matrices will definitely be above 2^32 - 1 which is the max size for a warp array (cf vastsoun/newton#322). Or am I misunderstanding something?

aserifi:
> Wait so the maxdims[i] * maxdims[i] overflows? Wouldn't that mean that there are 65k constraints per world (2^16 basically)?
> 
> Also, even with this fix, if one world already overflows then the size of the full flattened array of all Delassus matrices will definitely be above 2^32 - 1 which is the max size for a warp array (cf [vastsoun#322](https://github.com/vastsoun/newton/issues/322)). Or am I misunderstanding something?

Yes sorry you are right, I only want to fix the attribute names and add the guard

Guirec-Maloisel (dismissed):
LGTM!

newton/_src/solvers/kamino/_src/dynamics/delassus.py
```diff
@@ -901,7 +901,7 @@ def finalize(
         self._world_size = [maxdims[i] * maxdims[i] for i in range(self._num_worlds)]
```
vastsoun: @aserifi Since you pointed out issues with naming of `self._world_size`, I noticed that there is in fact a problem. As part of this PR, I would be in your debt if you could fix the following internal size caches:
- Rename `self._world_dims` to `self._world_maxdims`
- Rename `self._world_size` to `self._world_maxsize`

Please check, but it would seem that these slipped by. What were declared in `DelassusOperator.__init__()` were the intended caches, but somehow this slipped by the `DelassusOperator.finalize()` method.

vastsoun (changes requested):

vastsoun (approved):
LGTM, thanks for catching this!

PR #2736, opened 2026-05-06
Fix USD MJC angular limit_ke/kd over-scaling on revolute/D6 joints

### LINKED ISSUE #2536: Investigate USD revolute-joint limit_ke / limit_kd degree-to-radian rescaling
## Summary

`newton/_src/utils/import_usd.py` (lines 948–949 and 1124–1125) divides revolute-joint `limit_ke` / `limit_kd` by `DegreesToRadian` (≈π/180), which assumes the resolved value was authored in per-degree units. It is not clear that this assumption is correct for every USD authoring source we feed into the resolver system.

```python
# import_usd.py:946-949 (revolute branch)
joint_params["limit_lower"] *= DegreesToRadian
joint_params["limit_upper"] *= DegreesToRadian
joint_params["limit_ke"]   /= DegreesToRadian
joint_params["limit_kd"]   /= DegreesToRadian
```

The limit-lower/limit-upper scaling is uncontroversial — `UsdPhysics` authors revolute angle limits in degrees. But whether `limit_ke` / `limit_kd` are authored in per-degree depends on the source schema and is not obvious from USD itself.

## Concrete concern

After #2342 (`SchemaResolverMjc` now reads `mjc:solreflimit` for joint limits), `solref_to_stiffness` / `solref_to_damping` return stiffness and damping in MuJoCo-native per-radian units. Applying the revolute `/= DegreesToRadian` scaling over-scales them by ~57×.

Example: a hinge joint with `mjc:solreflimit = [0.08, 1]` yields `ke = 1/0.08² = 156.25` and `kd = 2/0.08 = 25.0` (per-radian). Newton currently stores these as `limit_ke ≈ 8952.47` and `limit_kd ≈ 1432.39` after the degree→radian scaling — 57.3× too large.

Observed while fixing #2342 but deliberately left out of that PR; the new regression test (`test_solreflimit_parsing`) uses a prismatic joint to sidestep the scaling and isolate the attribute-name fix.

## Questions to answer

1. **For each resolver source** (`SchemaResolverNewton`, `SchemaResolverPhysx`, `SchemaResolverMjc`, any future ones), what are the authored units of `limit_ke` / `limit_kd`?
    - PhysX `physxLimit:angular:stiffness` — per-degree or per-radian?
    - Newton `newton:angular:limitStiffness` — what's the intended unit in the (still-in-development) Newton USD schema?
    - MuJoCo `mjc:solreflimit` → `solref_to_stiffness_damping(...)` — per-radian (MuJoCo convention).
2. Does the current unconditional `/= DegreesToRadian` silently over-scale some sources and under-scale others, or is PhysX/UsdPhysics actually authoring in per-degree?
3. If the authoring unit varies by source, where does the conversion belong? Candidates:
    - Per-resolver `usd_value_transformer` (each resolver owns the conversion to Newton's internal per-radian convention).
    - A flag on `SchemaAttribute` indicating "angular limit scaled" and letting the importer apply the scaling only when set.
    - Drop the scaling in the importer entirely and require every resolver to return already-per-radian values.

Also applies to the D6-joint branch at [import_usd.py:1124–1125](../blob/main/newton/_src/utils/import_usd.py#L1124-L1125).

## Related

- #2342 — renamed `mjc:solref` → `mjc:solreflimit` on joints; surfaced this concern.
- #2504 — a conceptually similar per-degree / per-radian mixup in the MJCF importer (fixed).

vreutskyy:
## Summary

- `SchemaResolverMjc` returns per-radian stiffness/damping from `mjc:solreflimit` because `mjModel` always stores stiffness per-radian regardless of `mjc:compiler:angle`.
- The USD importer's revolute and D6-angular paths divide `limit_ke` / `limit_kd` by `DegreesToRadian`. Correct for `UsdPhysics`-authored gains (per-degree by convention) but over-scales MJC-sourced values by ~57x.
- Pre-multiply the MJC schema's angular `limit_*_ke` / `limit_*_kd` transformer outputs by `pi/180` so the importer's later division cancels and the per-radian value survives. Linear axes unaffected.

## Repro before fix

USD revolute joint with `mjc:solreflimit = [0.08, 1]` was importing as `limit_ke = 8952.47, limit_kd = 1432.39` instead of the per-radian `156.25 / 25.0`. After the fix, Newton stores `156.25 / 25.0` exactly.

Closes #2536.

## Test plan

- [x] New regression `test_solreflimit_parsing_revolute` verifies the per-radian values for `mjc:solreflimit=[0.08, 1]` on a revolute joint
- [x] Existing prismatic counterpart `test_solreflimit_parsing` still passes (linear path unchanged)
- [x] Full USD import suite: `uv run --extra dev -m newton.tests -k test_import_usd` -> `Ran 200 tests in 29s; OK`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect ~57× scaling of joint limit stiffness and damping imported from USD for revolute and D6-angular joints; values are now stored per radian.
* **Tests**
  * Added and updated unit tests to validate the per-radian conversion and absolute stiffness/damping values during USD import.
* **Documentation**
  * Updated changelog to document the USD importer correction.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2736)


newton/tests/test_import_usd.py
```diff
@@ -1937,6 +1937,92 @@ def PhysicsPrismaticJoint "Joint2" (
         self.assertAlmostEqual(float(limit_ke[dof2]), builder.default_joint_cfg.limit_ke, places=4)
         self.assertAlmostEqual(float(limit_kd[dof2]), builder.default_joint_cfg.limit_kd, places=4)
 
+    def test_solreflimit_parsing_revolute(self):
```
camevor, line 1941: This should probably be gated with 
```
```

camevor (dismissed):
Thanks for fixing! One test gate worth adding, otherwise lgtm

newton/tests/test_import_usd.py
vreutskyy: Good catch — added in 3dfe71c2.

camevor (approved):

PR #2745, opened 2026-05-07
Fix bug with nonuniform scaling in mesh vs convex collisions

nvtw:
## Description

Fix triangle-mesh-vs-convex collisions silently dropping all contacts under non-uniform (and even large uniform) mesh scale. The mesh BVH is built over unscaled `mesh.points`, but the AABB query in `mesh_vs_convex_midphase` was performed in scaled mesh-local space. Any convex shape whose unscaled-space AABB lay outside the BVH bounds received 0 triangles and 0 contacts. The query AABB and per-axis contact gap are now converted from scaled to unscaled (BVH) space before the BVH overlap test.

Also adds two small helper funcs in `collision_core.py`:
- `aabb_to_unscaled` — convert an AABB from a component-wise scaled frame to the unscaled frame, swapping per-axis min/max under negative scale.
- `transform_normal_with_scale` — correct normal transform under non-uniform component-wise scale (`R · (n / scale)`, normalized).

## Checklist

## Test plan

New regression class `TestMeshNonUniformScaling` in `newton/tests/test_narrow_phase.py` drops a sphere onto a unit-quad mesh with various scales (uniform, large uniform, pancake `(10,10,1)`, thin `(1,1,0.1)`, extreme `(50,0.5,1)`) and asserts contacts are produced with sane unit-length, roughly +Z normals; a far-separated case asserts 0 contacts (no false positives).

```
uv run --extra dev -m newton.tests -k test_narrow_phase.TestMeshNonUniformScaling
```

## Bug fix

Triangle-mesh-vs-convex collisions silently produce zero contacts whenever the mesh shape uses a non-uniform scale (e.g. `(10, 10, 1)`), and even for sufficiently large uniform scales. Mesh-averaging of the scale caused the AABB query against the BVH (which lives in unscaled mesh-local space) to miss all triangles.

**Steps to reproduce:**

1. Build a triangle mesh and add it as a shape with a non-uniform `scale=(sx, sy, sz)` (e.g. `(10, 10, 1)`).
2. Place a convex shape (sphere/box/capsule/...) so it overlaps the scaled mesh in world space.
3. Run the narrow phase — observe `contact_count == 0`.

**Minimal reproduction:**

```python
import numpy as np
import newton

verts = np.array(
    [[-0.5, -0.5, 0.0], [0.5, -0.5, 0.0], [0.5, 0.5, 0.0], [-0.5, 0.5, 0.0]],
    dtype=np.float32,
)
inds = np.array([0, 1, 2, 0, 2, 3], dtype=np.int32)
mesh = newton.Mesh(verts, inds)

# Non-uniform scale: 10x10 quad in world space, sphere dropped just above it.
# Without this fix: 0 contacts. With this fix: contacts as expected.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed triangle-mesh-vs-convex collisions to correctly handle non-uniform and large uniform mesh scaling, preventing silent loss of contact detection.

* **Tests**
  * Added comprehensive test coverage for mesh-vs-convex collision behavior under various scaling scenarios, including non-uniform and extreme anisotropic scaling configurations.


eric-heiden (approved):
Thanks, LGTM!

PR #2747, opened 2026-05-07
Expose marching-cubes edge clamp; fix nut_bolt_hydro threading

### LINKED ISSUE #2702: [BUG] In the nut_bolt_hydro example, the nuts do not screw into the bolts
### Bug Description

The nut_bolt_hydro seems to be broken as the nuts seems to have lot of difficulty to going down the bolt.
Results after a few minutes of simulation:

Image

### Reproduction Script

```python
uv run --extra examples -m newton.examples nut_bolt_hydro --world-count 2
```

mzamoramora-nvidia:
## Description

Closes #2702.

The threading regression in `nut_bolt_hydro` introduced by #2424 has two independent causes that this PR addresses; a third change opportunistically reclaims performance.

1. **Expose the marching-cubes edge clamp on the public Config.** #2424 hard-coded `MC_EDGE_CLAMP_MIN = 0.02` to keep edge-interpolation vertices off cube corners. The clamp biases contact-surface vertices off the true zero-isosurface, which is invisible at typical contact orientations but measurably damps threading-style contacts. Expose as `HydroelasticSDF.Config.mc_edge_clamp_min` (default `0.02`, recommended close to zero); validated to `[0.0, 0.5]` in `Config.__post_init__`. The example sets it to `0.0`.

2. **Refresh contacts inside the substep loop in the example.** With 5 substeps per frame and threading at ~10 rad/s, the nut rotates ~5° between collision detections, draining helical force on some nuts and pinning them in static friction (3 of 9 stuck on a 9-nut grid). Switch the example to 4 substeps with collision refreshed every 2 substeps under MuJoCo (every substep under XPBD), so contact normals stay aligned with the rotating helix. Also switch contact-buffer setup to the explicit `collision_pipeline.contacts()` + `pipeline.collide(state, contacts)` idiom used by `example_robot_panda_hydro`.

3. **Halve the example's SDF resolution.** With per-substep collision now keeping contacts fresh (item 2), the example relies less on contact-surface fidelity from a high-resolution SDF. Drop `SDF_MAX_RESOLUTION` from 256 to 128: still 0/9 stuck on the test grid, threading omega only ~25% lower, and each collision call is much cheaper. Net effect: the example is faster end-to-end than the pre-PR config despite the new schedule running collision twice per frame instead of once.

Empirical isolation experiments confirmed items 1 and 2 are independently necessary: clamp=0.02 with per-substep collision still produces ~1-2 rad/s creeping; clamp=0.0 with per-frame collision still leaves 3 of 9 stuck. Item 3 is performance-only and has no effect on the stuck-nut behavior.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_hydroelastic
uv run --extra examples -m newton.examples nut_bolt_hydro --world-count 2
uv run --extra examples -m newton.examples nut_bolt_hydro --world-count 2 --solver xpbd
```

New tests in `newton/tests/test_hydroelastic.py`:
- `test_mc_edge_clamp_min_validation`: range/NaN validation in `Config.__post_init__` (device-independent, runs on CPU-only CI).
- `test_mc_edge_clamp_min_changes_contact_surface`: order-invariant check that the value flows through to vertex placement, with a same-args control to rule out atomic-ordering false positives.

The pre-existing `test_no_degenerate_triangles_deep_penetration` from #2424 is preserved and explicitly pins `mc_edge_clamp_min=0.02` so its strict assertions remain meaningful when the default ever evolves.

## Bug fix

**Steps to reproduce:** see #2702.

```bash
uv run --extra examples -m newton.examples nut_bolt_hydro --world-count 2
```

| Branch | stuck nuts (9-grid) | ω peak | descent (3 s) | benchmark `time_simulate` |
|---|---|---|---|---|
| `main` | 3 / 9 | 0–11 rad/s | 12–17 mm | 89.0 ± 0.2 ms |
| this PR | 0 / 9 | 11–17 rad/s | 15–18 mm | **73.4 ± 0.06 ms (0.82×)** |

The example is 18% faster *and* nuts no longer pin in static friction.

## New feature / API change

```python

# Default preserves the previous hard-coded clamp (0.02).
config = HydroelasticSDF.Config()

# Disable the clamp for threading-sensitive contacts (nut_bolt_hydro pattern).
config = HydroelasticSDF.Config(mc_edge_clamp_min=0.0)
```

## Release Notes

* **Bug Fixes**
  * Fixed nut and bolt example threading regression; nuts now thread reliably with both MuJoCo and XPBD solvers.

* **New Features**
  * Added configurable edge-interpolation clamping for hydroelastic contact dynamics (default: 0.02; set to 0.0 to disable).

### DIFF
--- a/newton/examples/contacts/example_nut_bolt_hydro.py
+++ b/newton/examples/contacts/example_nut_bolt_hydro.py
@@ -24,7 +24,7 @@ ASSEMBLY_STR = "m20_loose"
 ISAACGYM_ENVS_REPO_URL = "https://github.com/isaac-sim/IsaacGymEnvs.git"
 ISAACGYM_NUT_BOLT_FOLDER = "assets/factory/mesh/factory_nut_bolt"
 
-SDF_MAX_RESOLUTION = 256
+SDF_MAX_RESOLUTION = 128
 SDF_NARROW_BAND_RANGE = (-0.005, 0.005)
 
 SHAPE_CFG = newton.ModelBuilder.ShapeConfig(
--- a/newton/examples/contacts/example_nut_bolt_hydro.py
+++ b/newton/examples/contacts/example_nut_bolt_hydro.py
@@ -166,7 +166,7 @@ class Example:
         # M20 helix are sensitive to the contact-surface vertex bias the clamp
         # introduces (see #2702).
         sdf_hydroelastic_config = HydroelasticSDF.Config(
-            mc_edge_clamp=0.0,
+            mc_edge_clamp_min=0.0,
         )
 
         self.collision_pipeline = newton.CollisionPipeline(
@@ -207,7 +207,8 @@ class Example:
 
         newton.eval_fk(self.model, self.model.joint_q, self.model.joint_qd, self.state_0)
 
-        self.contacts = self.model.collide(self.state_0, collision_pipeline=self.collision_pipeline)
+        self.contacts = self.collision_pipeline.contacts()
+        self.collision_pipeline.collide(self.state_0, self.contacts)
 
         self.viewer.set_model(self.model)
 
@@ -295,7 +296,7 @@ class Example:
         for _ in range(self.sim_substeps):
             # Re-run collision detection every substep so contact normals
             # stay aligned with the threading rotation (#2702).
-            self.contacts = self.model.collide(self.state_0, collision_pipeline=self.collision_pipeline)
+            self.collision_pipeline.collide(self.state_0, self.contacts)
             self.state_0.clear_forces()
 
             self.viewer.apply_forces(self.state_0)
--- a/newton/examples/contacts/example_nut_bolt_hydro.py
+++ b/newton/examples/contacts/example_nut_bolt_hydro.py
@@ -116,8 +116,8 @@ class Example:
         self.fps = 120
         self.frame_dt = 1.0 / self.fps
         self.sim_time = 0.0
-        self.sim_substeps = 6
-        self.collide_every = 3 if args.solver == "mujoco" else 1  # re-collide every K substeps
+        self.sim_substeps = 4
+        self.collide_every = 2 if args.solver == "mujoco" else 1  # re-collide every K substeps
         self.sim_dt = self.frame_dt / self.sim_substeps
 
         self.world_count = args.world_count
--- a/newton/examples/contacts/example_nut_bolt_hydro.py
+++ b/newton/examples/contacts/example_nut_bolt_hydro.py
@@ -292,8 +292,10 @@ class Example:
             self.graph = None
 
     def simulate(self):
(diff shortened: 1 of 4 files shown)


mzamoramora-nvidia:
## CI: GPU Benchmarks regression — investigated

The failing `Run GPU Benchmarks on AWS EC2` check has exactly one regression, and it's the documented PR trade-off.

**ASV comparison output** (pulled from the failing step's log):

| Benchmark | Before `01502ef6` | After `ac2a36af` | Ratio |
|---|---|---|---|
| `simulation.bench_contacts.FastExampleContactHydroWorkingDefaults.time_simulate` | 87.0 ± 0.1 ms | 134 ± 0.1 ms | **1.53×** |

That benchmark wraps `newton.examples.contacts.example_nut_bolt_hydro` directly — `world_count=20`, MuJoCo, 20 frames (`asv/benchmarks/simulation/bench_contacts.py:82–109`).

**All 26 other benchmarks** in the run are stable (ratio = 1.00 ± noise). No unexpected regressions outside the targeted example.

**Where the 1.53× comes from:**

The example went from `sim_substeps=5` with 1 collision per frame to `sim_substeps=6` with the per-solver `collide_every` heuristic. For MuJoCo (the path the benchmark exercises), that's 2 collisions per frame. So the kernel-launch cost increase is:
- substeps: 5 → 6 ≈ 1.2× more solver work
- collisions/frame: 1 → 2 = 2× more collision-detection work

Hydroelastic SDF collision detection dominates total cost in this example, so the combined effect lands at 1.53×, matching the per-substep collision schedule exactly.

This is the cheapest schedule that eliminates the stuck-nut behavior on a 9-nut grid, verified empirically:

| schedule | collisions/frame | stuck nuts (9-grid) |
|---|---|---|
| 5 substeps, 1 collision (pre-PR) | 1 | 3 / 9 |
| 6 substeps, collide every 3 (this PR, MuJoCo) | 2 | 0 / 9 |
| 6 substeps, collide every 2 | 3 | 0 / 9 |
| 5 substeps, collide every substep | 5 | 0 / 9 |

Going below 2 collisions/frame leaves nuts pinned in static friction. The PR description's "Performance note" already calls this out as the explicit correctness-over-throughput trade for a demonstration example.

newton/examples/contacts/example_nut_bolt_hydro.py
```diff
@@ -161,11 +163,19 @@ def __init__(self, viewer, args):
 
         self.model = main_scene.finalize()
 
+        # Disable the marching-cubes edge clamp — threading dynamics on the
+        # M20 helix are sensitive to the contact-surface vertex bias the clamp
+        # introduces.
+        sdf_hydroelastic_config = HydroelasticSDF.Config(
+            mc_edge_clamp_min=0.0,
```
nvtw, line 170: This is the only API change I assume? I think it's fine

nvtw (approved):

Issue #2749, opened 2026-05-07
[BUG] Ball joint and D6 joint configured as ball joint do not behave the same


gyeomannvidia:
### Bug Description

Repro steps. 
 
1. Create two bodies coupled to the world with a ball joint (add_joint_ball) and a D6 joint configured as a ball joint (add_joint_d6)  
2. For both bodies:  mass = 1, inertia =(1,1,1), com=0, joint frames = identity, armature=0
3. For body bodies: joint_q_init = identity quaternion
4. For body bodies: joint_qd_init = (1, 1, 1)  
5. For both bodies: apply a non-zero force (1,1,1)
6. gravity = 0.0 (really important step to ensure zero gravity compensation force)
7. For both bodies: simulate 1 step 
8. Compute output acceleration.  qddot should have value (force-coriolisForce)/Inertia for each axis.

The equation of motion is 

tau = M * qddot + C * qdot + g(q)

We have purposefully simplified this:
gravity=0 
identity frames and poses
identity inertia

With this simple system we have:
Coriolis force =  C*qdot = qdot X (inertia * qdot) = 0
g(q) = 0
M = inertia = identity

This allows us to rearrange for qddot for each dof:

qddot[i] = tau[i]

We produce the correct qddot with a ball joint but not with a d6 joint configured as a ball joint. 

The minimum repro is that qdot_init has two non-zero values.

### Reproduction Script

```python
class TestMuJoCoSolverBallVsD6Ball(unittest.TestCase):
    """Compare ball joint vs D6-with-3-angular-axes dynamics through SolverMuJoCo."""

    def test_ball_and_d6_ball_torque_response_match(self):
        """A torque applied to a body with a ball joint must produce the same ``qddot`` as the same body with a D6-as-ball joint.

        Two builders are constructed identically -- single body with
        mass = 1, identity inertia, zero CoM -- attached to world by:

            (A) ``add_joint_ball``
            (B) ``add_joint_d6`` with three angular axes (X, Y, Z), no
                linear axes (a "ball-equivalent" D6).

        Both have identity transforms, zero armature, zero gravity, the
        same initial ``joint_qd = omega``, and receive the same applied
        torque ``tau``. With **isotropic inertia**, the Coriolis term
        ``omega x (I * omega)`` collapses to ``omega x (k * omega) = 0``,
        and the manipulator equation reduces to ``M * alpha = tau``.
        With **identity inertia**, this further simplifies to
[...]

PR #2753, opened 2026-05-07
Introduction of inverse dynamics feature

### LINKED ISSUE #2497: [REQ] Articulation Inverse Dynamics - gravity compensation force, Coriolis compensation force, mass matrix
### Description

Given specified joint positions and velocities, use the MuJoCo solver to compute the joint forces required to counteract the Coriolis/centrifugal and gravity forces acting on the articulation. A related goal is to compute the mass matrix of the articulation.  

The forward dynamics equation is as follows:

	tau = M(q) * qdotdot + C(q, qdot) * qdot + g(q)

with M denoting the mass matrix; qdotqdot the joint dof accelerations; qdot the joint dof speeds; q the joint dof positions; the product C(q, qdot) * qdot is the Coriolis (and centrifugal) force; tau denotes the joint forces; and g(q) the joint forces arising from gravity acting on the links of the articulation (at the pose q).

The technical goal is to compute C(q, qdot)*qdot, g(q) and M(q).

### Motivation / Use Case

Controllers make use of the mass matrix and compensation forces to compute joint forces that achieve a desired outcome. Computing the Coriolis/compensation and gravity compensation forces together with the mass matrix allows this to be part of the simulation model. 

### Alternatives Considered

_No response_

gyeomannvidia:
## Description

Introduce inverse dynamics feature.  Closes https://github.com/newton-physics/newton/issues/2497

## Checklist

## Test plan

tests added to test_inverse_dynamics.py 
benchmarks added to bench_inverse_dynamics.py

## New feature / API change

```python
import newton

```
        reference_id, reference_scratch = model.inverse_dynamics()
        newton.eval_inverse_dynamics(model, state, newton.InverseDynamics.EvalType.ALL, reference_id, reference_scratch)
        H = reference_id.mass_matrix.numpy()
        g = reference_id.gravity_compensation_force.numpy()
        c = reference_id.coriolis_compensation_force.numpy()

## Summary by CodeRabbit

* **New Features**
  * Added inverse-dynamics evaluation: compute joint-space mass matrix, gravity and Coriolis compensation, and per-DOF inverse-dynamics forces; model-level helper to allocate required buffers; view-level convenience for scoped evaluation.

* **Documentation**
  * API docs and changelog updated to document the new inverse-dynamics interfaces.

* **Tests/Benchmarks**
  * Added benchmarks and harnesses for inverse-dynamics performance evaluation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2753)


jcarius-nv:

newton/_src/sim/inverse_dynamics.py
```diff
@@ -0,0 +1,517 @@
...
+    All three quantities follow the standard manipulator-equation convention
+    ``tau = M(q)*qddot + C(q,q_dot)*q_dot + g(q)``.
+
+    Requires ``state.body_q`` to be consistent with ``state.joint_q``;
+    callers must invoke :func:`~newton.eval_fk` (or otherwise update
+    ``state.body_q``) before this function.
+
+    Note:
+        Inverse dynamics considers only the kinematic tree. As a consequence,
+        loop-closure joints (``EqType.CONNECT``, ``EqType.WELD``, ``EqType.JOINT``)
```
gyeomannvidia, line 486: Should I refer to EqType here?  We have deprecated Builder.add_equality_constraint() but EqType has not been deprecated.  If the plan is to remove EqType then I should redo this comment.

adenzler-nvidia (commented):
## Review: Inverse Dynamics

Strong, well-engineered contribution. The physics is correct (gravity yields `-m·g`, Coriolis yields pure gyroscopic `ω×(Iω)`, and `tau` round-trips back through `SolverMuJoCo` to recover `qddot` across free/ball/D6/fixed roots) and the full suite passes on CPU + CUDA. No blocking correctness defects found. The comments below are robustness guards, a test-convention alignment, and polish. Inline comments cover the line-anchored items; the cross-cutting test/coverage items are here.

### Test surface (minimality)

The analytical suites are valuable and should stay — the closed-form Coriolis tests, both mass-matrix tests, and the two floating-root Coriolis tests pin *absolute* sign/scale at tight tolerance, which the loose-tolerance (2e-3), self-consistent solver round-trip in `_test_inverse_dynamics_force` cannot catch. But a few items are pure redundancy or over-engineering:

- Consolidate the two "axis ⟂/∥ gravity → zero" gravity tests into one `joint_type`-parametrized test (same invariant, near-identical all-zero expectation).
- Merge `test_..._from_rotated_root` with `test_..._from_rotated_joint_frame` — identical downstream gravity-force physics, differing only in where the rotation is injected.
- Consider dropping the loose `test_gravity_zero_without_gravity` / `test_gravity_nonzero_under_gravity` pair — subsumed by the analytical harness tests (which assert exact zero / exact nonzero values).
- The `_test_two_link_grav_comp_force` harness takes ~10 deeply-nested `[world][articulation][link]` list args and carries three internal shape-consistency `raise`s to guard against mismatched literals — a sign the API shape is too loose for its ~4 callers. Consider inlining model construction like the Coriolis / mass-matrix tests do.

### Coverage gaps worth filling

- An **all-deselected mask** case (the boundary of the zero-output masking convention — confirms the full-zero output and that nothing crashes with zero selected articulations).
- A **prismatic internal DOF** in the force round-trip: every internal joint in `_test_inverse_dynamics_force` is revolute, so the N-vs-N·m unit path for `tau`/`g` on a translational DOF is never round-tripped end-to-end.
- A **direct white-box test of `eval_inverse_dynamics_force`** with hand-chosen `H`/`qddot`/`bias` (currently only exercised via the solver round-trip and `test_coriolis_fixed_root_chain_with_com_offsets`, both of which feed it values from Newton's own pipeline — a self-consistent DOF-slicing bug could pass).
- A finite-only guard for the unsupported **DISTANCE / CABLE** paths and **multi-angular-DOF D6** (the D6 round-trip omission is correctly deferred to #2749, but a NaN regression on those outputs would currently go uncaught).

newton/_src/sim/inverse_dynamics.py
```diff
@@ -0,0 +1,517 @@
...
+        inverse_dynamics: Output container whose buffers are written in place.
+        scratch: Pre-allocated scratch buffers reused across calls.
+        mask: Optional ``wp.array[bool]`` of shape
+            ``(articulation_count,)`` selecting which articulations to
+            compute. Entries belonging to unselected articulations are
+            zero in the output buffers (mirroring
+            :func:`~newton.eval_mass_matrix`'s mask convention). If
+            ``None``, all articulations are computed.
+    """
+    if eval_type & InverseDynamics.EvalType.MASS_MATRIX:
```
adenzler-nvidia, line 511: **Silent no-op on unrecognized `eval_type`.** This is three independent `if eval_type & FLAG:` blocks with no fallthrough, so `eval_type=0` (e.g. a default-initialized `EvalType()`) or any int with no recognized bits returns having computed nothing — leaving the output buffers at their prior values — with no error. A downstream `eval_inverse_dynamics_force` then silently produces `tau=0`. Suggest a guard at the top of the function: `if not (eval_type & InverseDynamics.EvalType.ALL): raise ValueError(...)`.

newton/_src/sim/inverse_dynamics.py
```diff
@@ -0,0 +1,517 @@
...
+            model,
+            state,
+            H=inverse_dynamics.mass_matrix,
+            J=scratch.J,
+            body_I_s=scratch.body_I_s,
+            joint_S_s=scratch.joint_S_s,
+            mask=mask,
+        )
+
+    if eval_type & InverseDynamics.EvalType.GRAVITY_FORCE:
```
adenzler-nvidia, line 534: **Validation asymmetry.** The `MASS_MATRIX` branch above validates `mass_matrix.shape` and raises a clean `ValueError`, but the `GRAVITY_FORCE` / `CORIOLIS_FORCE` branches validate nothing. A caller reusing a stale `InverseDynamics`/`InverseDynamicsScratchBuffer` after rebuilding the model (which the class docstrings warn against but nothing enforces) gets a clean error via `EvalType.ALL` but out-of-bounds kernel access (UB / illegal-memory crash) via `EvalType.GRAVITY_FORCE`. Consider extending the same cheap Python-side shape checks to `gravity_force`/`coriolis_force` (and the scratch sizing) for consistency.

newton/_src/sim/inverse_dynamics.py
```diff
@@ -0,0 +1,517 @@
...
+        """Allocate output buffers for inverse dynamics.
+
+        The mass matrix is stored in the shape expected by
+        :func:`~newton.eval_mass_matrix`:
+        ``(articulation_count, max_dofs_per_articulation, max_dofs_per_articulation)``.
+        ``max_dofs_per_articulation`` already includes any 6 DOFs contributed by a
+        floating-base root joint.
+
+        The ``gravity_force`` and ``coriolis_force`` buffers share the flat
+        joint-space layout that :func:`~newton.Model.joint_qd` uses.
```
adenzler-nvidia: Broken cross-reference: `Model.joint_qd` is a data attribute, not a function, so `:func:` renders broken. Use `:attr:`~newton.Model.joint_qd``.

newton/_src/sim/inverse_dynamics.py
```diff
@@ -0,0 +1,517 @@
...
+
+        self.gravity_force: wp.array[wp.float32] = wp.zeros(jdc, dtype=wp.float32, device=device)
+        """Gravity force ``g(q) = ∂U/∂q`` [N or N·m, depending on joint type], shape (joint_dof_count,), dtype float.
+        The joint-space force a controller must apply to hold the articulation static under gravity."""
+
+        self.coriolis_force: wp.array[wp.float32] = wp.zeros(jdc, dtype=wp.float32, device=device)
+        """Coriolis + centrifugal force ``C(q, q_dot)*q_dot`` [N or N·m, depending on joint type], shape (joint_dof_count,), dtype float.
+        The joint-space force a controller must apply to maintain the current ``joint_qd`` at zero acceleration in the absence of gravity."""
+
+        self.tau: wp.array[wp.float32] = wp.zeros(jdc, dtype=wp.float32, device=device)
```
adenzler-nvidia, line 183: `tau` is allocated here but never written by `eval_inverse_dynamics` — only by `eval_inverse_dynamics_force`, and only when the caller passes this buffer as its output. It's documented, but a user running just `eval_inverse_dynamics(EvalType.ALL)` will find `tau` silently all-zero. Worth a sharper note on the attribute (e.g. "populated only by `eval_inverse_dynamics_force`; remains zero otherwise").

newton/_src/sim/inverse_dynamics.py
```diff
@@ -0,0 +1,517 @@
...
+        self.body_f_s = wp.empty(bc, dtype=wp.spatial_vector, device=device)
+        self.body_a_s = wp.empty(bc, dtype=wp.spatial_vector, device=device)
+        self.body_ft_s = wp.empty(bc, dtype=wp.spatial_vector, device=device)
+
+        # Reused as the eval_mass_matrix Jacobian scratch.
+        self.J = wp.empty((ac, max_links * 6, max_dofs), dtype=wp.float32, device=device)
+
+        # Constant-zero inputs (allocated once, never written).
+        self.zeros_dof = wp.zeros(jdc, dtype=wp.float32, device=device)
+        self.zeros_body = wp.zeros(bc, dtype=wp.spatial_vector, device=device)
```
adenzler-nvidia, line 100: Minor maintainability note: `zeros_body` must stay all-zero. `eval_rigid_tau` writes `body_f_ext[child]` in place, and in the ID path that array IS this buffer — it's safe only because a zero input maps back to zero. A one-line comment here would prevent a future reader from reusing it for a nonzero external force.

newton/_src/sim/articulation.py
```diff
@@ -1378,6 +1378,110 @@ def eval_articulation_mass_matrix(
...
+    dof_count = dof_end - dof_start
+
+    for i in range(dof_count):
+        sum_val = float(0.0)
+        for j in range(dof_count):
+            sum_val += H[art_idx, i, j] * qddot[dof_start + j]
+        tau[dof_start + i] = sum_val + coriolis_force[dof_start + i] + gravity_force[dof_start + i]
+
+
+def eval_inverse_dynamics_force(
```
adenzler-nvidia, line 1469: Unlike `eval_inverse_dynamics` (which checks `mass_matrix.shape`), this consumes `H`/`qddot`/`coriolis_force`/`gravity_force` and produces the user's final `tau` with no shape validation — a mismatched buffer reads out of bounds in-kernel (UB) rather than raising. Acceptable as a hot-path-style no-check, but a single cheap `tau.shape == (model.joint_dof_count,)` assertion would catch the most common caller error (a `tau` not sourced from `InverseDynamics.tau`) and match the validation precedent the feature already sets.

newton/_src/solvers/featherstone/kernels.py
```diff
@@ -1064,9 +1064,142 @@ def convert_free_distance_joint_f_public_to_internal(
...
+        # qd shift and the output-side wrench shift use the same offset vector).
+        X_wpj = joint_X_p[joint_id]
+        if parent >= 0:
+            X_wpj = body_q[parent] * X_wpj
+        q_p = wp.transform_get_rotation(X_wpj)
+        x_anchor_world = wp.transform_get_translation(X_wpj)
+        x_child_com_world = wp.transform_get_translation(body_q_com[child])
+        r_child_com_parent = wp.quat_rotate_inv(q_p, x_child_com_world - x_anchor_world)
+
+        # Velocity-product correction. tau = -f_b_s, so adding to tau is
```
adenzler-nvidia, line 1177: The three-step derivation is also spelled out in full in this kernel's docstring; these inline comments re-derive it, creating dual-maintenance drift risk. Consider trimming the inline comments to one-liners that point back to the numbered steps in the docstring. Separately, the docstring's shorthand `joint_f = -dot(S, body_f_s)` is imprecise — the dotted quantity in `jcalc_tau` is the *total* wrench `f_b + f_t + f_ext`, not the `body_f_s` scratch array (which is passed as `body_fb_s`); a neutral `-dot(S, f_total)` avoids a stale-name trap.

newton/_src/utils/selection.py
```diff
@@ -1670,6 +1681,41 @@ def eval_mass_matrix(self, state: State, H=None, J=None, body_I_s=None, joint_S_
             self.model, state, H, J=J, body_I_s=body_I_s, joint_S_s=joint_S_s, mask=articulation_mask
         )
 
+    def eval_inverse_dynamics(
+        self,
+        state: State,
+        eval_type: "InverseDynamics.EvalType",
```
adenzler-nvidia: `InverseDynamics` / `InverseDynamicsScratchBuffer` are imported at module top level (not under `TYPE_CHECKING`), so these string-quoted hints can be unquoted: `eval_type: InverseDynamics.EvalType`, etc.

asv/benchmarks/simulation/bench_inverse_dynamics.py
```diff
@@ -0,0 +1,145 @@
...
+import os
+import sys
+
+import numpy as np
+import warp as wp
+from asv_runner.benchmarks.mark import SkipNotImplemented, skip_benchmark_if
+
+import newton
+
+wp.config.quiet = True
```
adenzler-nvidia: Every sibling benchmark (`bench_ik.py`, `bench_mujoco.py`, …) uses `wp.config.log_level = wp.LOG_WARNING` here. `wp.config.quiet` is actually the *current* valid key (the siblings set a now-defunct attribute), so this is arguably more correct — but it's inconsistent with the rest of the suite. Either is fine; flagging for awareness. The rest of the benchmark structure matches the `bench_ik.py` conventions cleanly.

newton/tests/test_inverse_dynamics.py
```diff
@@ -0,0 +1,2914 @@
...
+                for dof_idx in range(model.joint_dof_count):
+                    if dof_selected[case_idx, dof_idx]:
+                        self.assertAlmostEqual(float(g[dof_idx]), float(g_ref[dof_idx]), delta=1e-5)
+                        self.assertAlmostEqual(float(c[dof_idx]), float(c_ref[dof_idx]), delta=1e-5)
+                    else:
+                        self.assertAlmostEqual(float(g[dof_idx]), 0.0, delta=1e-6)
+                        self.assertAlmostEqual(float(c[dof_idx]), 0.0, delta=1e-6)
+
+
+class TestGravCompForceCPU(TestGravCompForce, unittest.TestCase):
```
adenzler-nvidia, line 3324: This block hand-duplicates every test class into `…CPU` / `…CUDA` variants (~43 lines of boilerplate). The neighboring `test_kinematics.py` — which tests the sibling `eval_fk`/`eval_ik`/`eval_jacobian`/`eval_mass_matrix` — uses the repo-standard pattern: a single `TestCase`, `devices = get_test_devices()` from `newton.tests.unittest_utils`, and `add_function_test(cls, name, fn, devices=devices)`. Suggest aligning with that convention (it also means writing the tests as `(test, device)` free functions rather than mixin methods reading `self.device`).

newton/tests/test_inverse_dynamics.py
```diff
@@ -0,0 +1,2914 @@
...
+                        M_12 = m * (l2c**2 + L1 * l2c * cos_q2) + Iyy
+                        expected = np.array([[M_11, M_12], [M_12, M_22]])
+
+                        np.testing.assert_allclose(M, expected, atol=1e-3, rtol=1e-5)
+
+
+class TestManipulatorEquation(TestInverseDynamicsBase):
+    """Manipulator-equation tests covering combined inverse-dynamics outputs."""
+
+    def test_eval_all_populates_every_buffer(self):
```
adenzler-nvidia, line 2019: This finite+nonzero smoke test appears fully subsumed: `test_bias_forces_flag_populates_both` asserts finite+nonzero for `g`/`c` *and* is strictly stronger (a `mass_matrix` sentinel proves the flag isolation), while the mass-matrix and round-trip tests cover the rest. The only unique thing here is exercising the literal `EvalType.ALL` in one call — which the masking test's reference run already does. Candidate for removal.

newton/_src/sim/articulation.py
```diff
@@ -1378,6 +1378,116 @@ def eval_articulation_mass_matrix(
...
+    mix of fixed-root (1+ internal DOFs) and floating-root (6 root DOFs +
+    internal DOFs) articulations is handled uniformly.
+    """
+    art_idx = wp.tid()
+
+    if art_idx >= articulation_count:
+        return
+
+    joint_start = articulation_start[art_idx]
+    joint_end = articulation_start[art_idx + 1]
```
adenzler-nvidia: Use `articulation_end[art_idx]` here instead of the next `articulation_start`. The `articulation_start` sentinel intentionally includes converted loop-closing joints, while `H` is sized from the regular-tree range ending at `articulation_end`. For a two-DOF articulation followed by a three-DOF ball loop closure, this computes `dof_count == 5` for a `2 x 2` mass matrix and indexes `H` out of bounds. Please pass `model.articulation_end` into this kernel and add a loop-closure regression test.

newton/_src/solvers/featherstone/kernels.py
```diff
@@ -1064,9 +1064,142 @@ def convert_free_distance_joint_f_public_to_internal(
...
+        ang_correction = body_mass[child] * wp.cross(r_child_com_parent, wp.cross(omega_parent, v_com_parent))
+        joint_f[qd_start + 3] = joint_f[qd_start + 3] + ang_correction[0]
+        joint_f[qd_start + 4] = joint_f[qd_start + 4] + ang_correction[1]
+        joint_f[qd_start + 5] = joint_f[qd_start + 5] + ang_correction[2]
+
+    # Sign flip: ``eval_rigid_tau`` outputs ``-dot(S, body_f_s)`` which is
+    # the negation of the standard manipulator-equation bias. Flip every
+    # per-DOF entry so the buffer stores the standard ``+g(q)`` /
+    # ``+C(q, q_dot)*q_dot`` directly.
+    for i in range(qd_start, qd_end):
```
adenzler-nvidia, line 1234: This conversion leaves FREE/DISTANCE forces in the joint-parent frame, but the new API documentation says `tau` has the same layout as `Control.joint_f`, whose FREE/DISTANCE entries are world-frame CoM wrenches. With a free root whose parent transform is rotated 90 degrees about X, gravity compensation returns approximately `(0, 10, 0)` instead of the controller-ready world wrench `(0, 0, 10)`. Please either transform the exposed FREE/DISTANCE forces, including the `M*qddot` result, into the controller's world frame or document a parent-frame generalized-force contract and remove the controller-ready claims. A rotated-parent regression test would pin this down.

newton/_src/sim/inverse_dynamics.py
```diff
@@ -0,0 +1,540 @@
...
+    callers must invoke :func:`~newton.eval_fk` (or otherwise update
+    ``state.body_q``) before this function.
+
+    Note:
+        Inverse dynamics considers only the kinematic tree. As a consequence,
+        loop-closure joints (``EqType.CONNECT``, ``EqType.WELD``, ``EqType.JOINT``)
+        play no role in the inverse dynamics evaluation.
+
+    Note:
+        A D6 joint with more than one angular DOF uses a stacked
```
adenzler-nvidia: This limitation is stale after the merged D6 axis-transport changes: `jcalc_motion` now uses `transform_2d_rotational_axes` / `transform_3d_rotational_axes` rather than summing raw body-frame axes, and the referenced multi-angular D6 forward-dynamics bug is closed. Please remove or update this note and restore multi-angular D6 coverage so the documented limitation matches the current implementation.

adenzler-nvidia (commented):
Found three actionable issues in the inverse-dynamics implementation and documentation.

newton/_src/sim/inverse_dynamics.py
gyeomannvidia: Added code to report an error and a test that the error is issued.
5694b017ad56e6412dfa611d37328ffc8a2b27fc

newton/_src/sim/inverse_dynamics.py
gyeomannvidia: Add shape validation for gravity_force and coriolis_force and mass matrix buffers.
98f21d929ac03dcac9ed13350a7e85b7c7d81bb8

newton/_src/sim/inverse_dynamics.py
gyeomannvidia: Fixed 
84ec2804cbc27b9085f8787a036f01410dc63428

newton/_src/sim/inverse_dynamics.py
gyeomannvidia: Clarify that InverseDynamics.tau is not written by eval_inverse_dynamics
84ec2804

newton/_src/sim/inverse_dynamics.py
gyeomannvidia: dfa69b3c94fbf29865b5ef7ce70865312534d15f
Document InverseDynamicsScratchBuffer attributes as internal

newton/_src/sim/articulation.py
gyeomannvidia: 1d3f731a39e3d4ce5393d14c4f2c5f1be9561376
Add tau shape validation to eval_inverse_dynamics_force

newton/_src/utils/selection.py
gyeomannvidia: 04ebd9834c14b35f42ddf9927284dfb2b662b3bd
Unquote InverseDynamics type hints in ArticulationView.eval_inverse_dynamics

newton/_src/solvers/featherstone/kernels.py
gyeomannvidia: I'm comfortable with a repetition of the derivation so I will leave it as it is.

The docstring that mentions body_fb_s has been fixed, as suggested.
d5d7cf8c4f00e9bb7fd8b5e4d61eaa87370422fb

asv/benchmarks/simulation/bench_inverse_dynamics.py
gyeomannvidia: 2dc122a85096b7e731bb2cce5a27a35fc96941e9
Align bench_inverse_dynamics warp log level with sibling benchmarks

newton/tests/test_inverse_dynamics.py
gyeomannvidia: I prefer this style.

newton/tests/test_inverse_dynamics.py
gyeomannvidia: The test has some value. I strengthened the test by cross-checking the results of each flag.

fabe10e964b3f7f055aa4c6e5040c221f94e802d
Strengthen test_eval_all_populates_every_buffer with per-flag cross-check

newton/_src/sim/articulation.py
gyeomannvidia: Great catch.  

77e1fbce62938cd8214dfe0c4cc3edf52f78f88a

newton/_src/solvers/featherstone/kernels.py
gyeomannvidia: bc9dcefc5bdb3988b0d970fe8781e2f7ef2af294

newton/_src/sim/inverse_dynamics.py
gyeomannvidia: The disputed note came about because a round trip test was failing with SolverFeatherstone.  I made a mistake as to the reason behind the failure. The real reason behind the failure was a bug in pre-existing Featherstone code used by the inverse dynamics feature.

The bug in the Featherstone code was revealed by a simple angular momentum conservation test with 3 angular dofs with non-zero velocity and no applied forces. This momentum conversation test passes with SolverMuJoCo but fails with SolverFeatherstone.  I have a fix for this now in the commit hash below.

The disputed note has been removed in the commit hash below.

The round trip inverse dynamics test has been amended to include d6_ball and d6_2ang joint configurations. These were failing prior to the fix discussed above.  The test has also been amended to run forward dynamics with SolverFeatherstone and SolverMuJoCo. This is also in the commit hash below.

af83b294dea83e1c0d7f9e8c3a51e125d78fd1f4

newton/_src/sim/articulation.py
```diff
@@ -1378,6 +1378,116 @@ def eval_articulation_mass_matrix(
...
+    inputs follow the standard manipulator-equation sign convention
+    (``+C(q,q_dot)*q_dot`` and ``+g(q) = +∂U/∂q``, the buffers populated by
+    :func:`eval_inverse_dynamics`) and are added directly. Per-articulation
+    DOF counts are recovered from :attr:`Model.joint_qd_start`, so a mix of
+    fixed-root and floating-root articulations across multiple worlds is
+    handled uniformly.
+
+    Args:
+        model: The model containing articulation definitions.
+        H: Joint-space mass matrix [kg, kg·m, or kg·m^2, depending on
```
jcarius-nv: Why do you call it H here but M in the equation?

jcarius-nv (commented):
I only had a chance to look at the main API and documentation, but I like it! Thanks Gordon!

newton/_src/sim/articulation.py
gyeomannvidia: Wow, this is such a mess. 

Featherstone convention is H but the manipulator equation convention is M.

I updated the docstring to make it clear that H and M(q) are the same.

9acd0249ada33d60e60911fc2799642bc969c099

newton/_src/sim/articulation.py
```diff
@@ -1378,6 +1378,116 @@ def eval_articulation_mass_matrix(
...
+    joint_end = articulation_end[art_idx]
+    dof_start = joint_qd_start[joint_start]
+    dof_end = joint_qd_start[joint_end]
+    dof_count = dof_end - dof_start
+
+    for i in range(dof_count):
+        sum_val = float(0.0)
+        for j in range(dof_count):
+            sum_val += H[art_idx, i, j] * qddot[dof_start + j]
+        tau[dof_start + i] = sum_val + coriolis_force[dof_start + i] + gravity_force[dof_start + i]
```
adenzler-nvidia: Could we clarify and fix the FREE/DISTANCE frame contract here? `gravity_force` and `coriolis_force` are converted to the world-frame CoM wrench convention used by `Control.joint_f`, but `H @ qddot` appears to remain in joint-parent coordinates before being added. With a FREE root whose parent frame is rotated 90° about X, mass 2, zero gravity, and parent-frame `qddot = (0, 1, 0, 0, 0, 0)`, this returns `tau ≈ (0, 2, 0, ...)` instead of the documented controller-ready world wrench `(0, 0, 2, ...)`. Would the intended fix be to provide the necessary state/frame data and rotate the completed FREE/DISTANCE contribution into world space, or should all inverse-dynamics terms instead remain in generalized parent-frame coordinates with a separate conversion to `Control.joint_f`? Could we add this nonzero-`qddot` rotated-parent case alongside the existing gravity-only regression?

newton/_src/sim/model.py
```diff
@@ -1673,6 +1674,38 @@ def control(self, requires_grad: bool | None = None, clone_variables: bool = Tru
         )
         return c
 
+    def inverse_dynamics(self) -> tuple[InverseDynamics, InverseDynamicsScratchBuffer]:
```
adenzler-nvidia: Was the decision for `Model.inverse_dynamics()` to return an `(InverseDynamics, InverseDynamicsScratchBuffer)` pair discussed and agreed on as the intended public API? I’m wondering why this does not follow the pattern established by `Model.contacts()`, which returns a single public, model-sized buffer while keeping its supporting `CollisionPipeline` out of the returned interface. Since `InverseDynamicsScratchBuffer` is documented as containing only internal implementation details, could `Model.inverse_dynamics()` instead return one `InverseDynamics` workspace that privately owns its scratch buffers? That would keep independent workspaces possible for reentrancy and CUDA graph capture while avoiding a public scratch type and preventing outputs from being paired with scratch allocated for another model.

newton/_src/sim/articulation.py
```diff
@@ -1378,6 +1378,116 @@ def eval_articulation_mass_matrix(
...
+            Typically :attr:`InverseDynamics.tau`.
+    """
+    if model.articulation_count == 0:
+        return
+
+    expected_shape = (model.joint_dof_count,)
+    if tau.shape != expected_shape:
+        raise ValueError(f"tau has shape {tau.shape}, expected {expected_shape}.")
+
+    tau.zero_()
```
adenzler-nvidia: Could we eliminate the separate `tau.zero_()` launch by having each articulation thread also zero the non-tree DOF range between `articulation_end[art_idx]` and `articulation_start[art_idx + 1]`? Since `articulation_start` includes the final sentinel and bounds the full joint range—including converted loop-closing joints—while `articulation_end` marks the end of the regular tree, the corresponding range `joint_qd_start[articulation_end[art_idx]]:joint_qd_start[articulation_start[art_idx + 1]]` appears to cover exactly the slots the matrix-vector loop intentionally skips. Are there orphan-joint or zero-articulation layouts that would make this unsafe or leave other slots uncovered? If not, could we fold that zeroing into this kernel, remove the extra launch, and strengthen the loop-closure test by prefilling `tau` with a nonzero sentinel and asserting that the loop-closing slot is cleared?

newton/_src/sim/inverse_dynamics.py
adenzler-nvidia: #3296 is introducing a mujoco-scoped enum, but the existing one will be deprecated so depending on what arrives first this might need an update. non-blocking though.

CHANGELOG.md
```diff
@@ -241,6 +243,7 @@
 - Import per-shape authored color from USD stages into `ModelBuilder.shape_color`
 - Add `TRIANGLE_PRISM` support-function type for heightfield triangles, extruding 1 m along the heightfield's local -Z so GJK/MPR naturally resolves shapes on the back side
 - Add `ViewerGL.log_scalar()` for live scalar time-series plots in the viewer
+- Add `newton.utils.OnnxRuntime`, a graph-capturable ONNX inference engine backed solely by Warp kernels (no `onnxruntime` or `torch` runtime dependency); used by `ControllerNeuralMLP` and `ControllerNeuralLSTM` to load `.onnx` policies. To migrate a TorchScript policy, run `torch.onnx.export(model, dummy_input, "policy.onnx", opset_version=17)` once and point the controllers at the resulting `.onnx` file. The `onnx` package is now an optional extra (`pip install newton[onnx]`); install it explicitly to use the ONNX runtime.
```
adenzler-nvidia: there seems to be an unrelated changelog entry that made it in

newton/_src/sim/articulation.py
gyeomannvidia: The free/distance joints cause sooooo many problems and require special code almost everywhere. 

8fc04da8a80c70071b19b2e2dad67607d0d8c7d8

newton/tests/test_inverse_dynamics.py
```diff
@@ -0,0 +1,3139 @@
...
+                lambda b, link: b.add_joint_distance(
+                    parent=-1,
+                    child=link,
+                    parent_xform=identity_xform,
+                    child_xform=identity_xform,
+                ),
+            ),
+            (
+                "cable",
+                lambda b, link: b.add_joint_cable(
```
adenzler-nvidia: Could we remove CABLE from this finite-output test, or define and test an explicit unsupported result instead? The kernel reports `jcalc_motion not implemented for joint type 7`; in two serial runs I saw the CPU CABLE subtest fail once on a non-finite mass matrix and then pass, so the finite assertion appears to depend on undefined output rather than a supported contract.

newton/tests/test_inverse_dynamics.py
```diff
@@ -0,0 +1,3139 @@
...
+                joint_q = state.joint_q.numpy()
+                joint_q[0:7] = [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0]
+                state.joint_q.assign(joint_q)
+                newton.eval_fk(model, state.joint_q, state.joint_qd, state)
+
+                inverse_dynamics, scratch = model.inverse_dynamics()
+                newton.eval_inverse_dynamics(
+                    model=model,
+                    state=state,
+                    eval_type=newton.InverseDynamics.EvalType.GRAVITY_FORCE,
```
adenzler-nvidia, line 1054: Could this regression also cover the nonzero-`qddot` path? It currently requests only `GRAVITY_FORCE`, so it verifies the RNEA bias conversion but not the mass-matrix contribution used by `eval_inverse_dynamics_force`. A rotated parent with gravity disabled and a known `qddot` would pin down whether the exposed `M*qddot` wrench follows the same world-frame contract.

newton/tests/test_inverse_dynamics.py
```diff
@@ -0,0 +1,3139 @@
...
+
+        np.testing.assert_array_equal(inverse_dynamics.mass_matrix.numpy(), sentinel)
+        g = inverse_dynamics.gravity_force.numpy()
+        c = inverse_dynamics.coriolis_force.numpy()
+        self.assertTrue(np.all(np.isfinite(g)))
+        self.assertTrue(np.all(np.isfinite(c)))
+        self.assertGreater(float(np.max(np.abs(g))), 1e-6)
+        self.assertGreater(float(np.max(np.abs(c))), 1e-6)
+
+    def test_eval_inverse_dynamics_raises_on_mass_matrix_shape_mismatch(self):
```
adenzler-nvidia: Could we drop this standalone test and rely on the `mass_matrix` case in `test_eval_inverse_dynamics_raises_on_buffer_shape_mismatch` below? That case supplies the same wrong shape under the same `MASS_MATRIX` flag and checks the same error details, so this appears to exercise the same branch twice.

newton/tests/test_inverse_dynamics.py
```diff
@@ -0,0 +1,3139 @@
...
+
+        # Combined call under test.
+        combined, scratch = model.inverse_dynamics()
+        newton.eval_inverse_dynamics(model, state, EvalType.ALL, combined, scratch)
+
+        np.testing.assert_array_equal(combined.mass_matrix.numpy(), ref_mm.mass_matrix.numpy())
+        np.testing.assert_array_equal(combined.gravity_force.numpy(), ref_gf.gravity_force.numpy())
+        np.testing.assert_array_equal(combined.coriolis_force.numpy(), ref_cf.coriolis_force.numpy())
+
+    def _test_inverse_dynamics_force(self, non_zero_gravity: bool, non_zero_initial_dof_velocities: bool):
```
adenzler-nvidia, line 2075: Would it make sense to trim the combination matrix in `_test_inverse_dynamics_force`? It currently crosses eight root-joint variants, four state configurations, both solvers, four gravity/velocity modes, and both devices. Since the analytical gravity and Coriolis tests already validate the individual bias terms, could we keep one comprehensive joint-type case—perhaps with both biases enabled—plus a smaller baseline or solver-parity case? That should preserve coverage of the force assembly and supported joint layouts without repeating the same behavior across every combination.

newton/_src/sim/model.py
gyeomannvidia: We never discussed the API.  This review was the intended place for discussion.

I like the proposal to make InverseDynamicsScratchBuffer a member of InverseDynamics .

a363a5464072b5b57207d68a62bd8b6238df6811

newton/tests/test_physics_verification.py
```diff
@@ -523,6 +523,79 @@ def compute_momenta(state):
...
+        omega = state.body_qd.numpy()[0, 3:6]
+        R = np.array(wp.quat_to_matrix(wp.quat(*bq[3:7].tolist()))).reshape(3, 3)
+        return (R @ I_body @ R.T) @ omega
+
+    L0 = angular_momentum_world(state_0)
+    quat_0 = state_0.body_q.numpy()[0, 3:7].copy()
+    test.assertGreater(np.linalg.norm(L0), 0.1, "Initial angular momentum should be nonzero")
+
+    sim_dt = 1e-3
+    for _ in range(1000):
```
adenzler-nvidia: Could we reduce the runtime of this regression by using fewer than 1,000 integration steps, possibly with a larger stable time step? The docstring reports roughly 46% momentum drift without the fix over the current interval, so there may be enough separation to use a cheaper configuration that still reliably fails before the fix and passes afterward. Could we find the smallest step count—and, if appropriate, the largest stable `sim_dt`—that preserves that regression signal?

CHANGELOG.md
gyeomannvidia: I asked Claude to resolve a merge conflict.  I am never doing that again!

gyeomannvidia: 1dddea8316d0e516507a14bcfb21889a15635c08

newton/tests/test_inverse_dynamics.py
gyeomannvidia: ca84565410caabf0f2212b352ac87fba6e13b449

newton/tests/test_inverse_dynamics.py
gyeomannvidia: The M*qddot issue is tested in test_inverse_dynamics_force_free_root_rotated_parent
 
A single test could cover both paths but I think it is cleaner to separate them out to separate tests.  They really are testing separate concepts.

newton/tests/test_inverse_dynamics.py
gyeomannvidia: Yes, we can remove this test

2d442a0df8444ced1e2465574ba4bcef523a1512

newton/tests/test_inverse_dynamics.py
gyeomannvidia: It really is useful to run all 8 combinations.  Getting all combinations to work was a significant effort.  Each runs 1 single sim step so the cost is negligible.

newton/tests/test_physics_verification.py
gyeomannvidia: 15cca98bf3cf9b2ba1021fc6bb9ea04b3f6f4dfa

newton/_src/sim/articulation.py
gyeomannvidia: 8fe9c5184720ed84eb5219455fb4601984d9c883

newton/_src/sim/articulation.py
```diff
@@ -1378,6 +1378,175 @@ def eval_articulation_mass_matrix(
...
+    for i in range(dof_count):
+        sum_val = float(0.0)
+        for j in range(dof_count):
+            sum_val += H[art_idx, i, j] * qddot[dof_start + j]
+        tau[dof_start + i] = sum_val
+
+    # Rotate the FREE/DISTANCE root wrench from parent frame to world so it
+    # matches the world-frame bias terms. Uses the same parent-frame-in-world
+    # rotation q_p as convert_free_distance_joint_f_internal_to_public.
+    root_type = joint_type[joint_start]
```
adenzler-nvidia: This rotates only the root block, but `convert_free_distance_joint_f_internal_to_public` runs with `dim=model.joint_count` and converts every FREE/DISTANCE joint's bias entries to the world-frame CoM-wrench convention, including non-root ones (`add_joint_free` documents `parent >= 0`). For a non-root FREE/DISTANCE joint, `tau` therefore sums a parent-frame `H @ qddot` block with world-frame bias terms.

Repro on this branch: two physically identical systems, a floating body whose free-joint parent frame is rotated 90 degrees about X, built once as a root FREE joint and once as a FIXED joint to a rotated parent body with a FREE joint below it. Zero gravity and velocity, so `tau` is purely the `H @ qddot` wrench:

<details><summary>repro script</summary>

```python
import numpy as np
import warp as wp
import newton

qx = wp.quat_from_axis_angle(wp.vec3(1.0, 0.0, 0.0), wp.pi / 2.0)
rot = wp.transform(wp.vec3(0.0, 0.0, 0.0), qx)
identity = wp.transform(wp.vec3(0.0, 0.0, 0.0), wp.quat_identity())
I_local = np.diag([0.3, 0.5, 0.4])
qddot_np = np.array([0.5, -0.3, 0.7, 0.2, 0.4, -0.6], dtype=np.float32)

def tau(nonroot):
    b = newton.ModelBuilder(gravity=0.0, up_axis=newton.Axis.Z)
    link = b.add_link(mass=2.0, com=wp.vec3(0.0, 0.0, 0.0), inertia=wp.mat33(*I_local.flatten().tolist()))
    if nonroot:
        p = b.add_link(xform=rot, mass=1.0, com=wp.vec3(0.0, 0.0, 0.0), inertia=wp.mat33(*np.eye(3).flatten().tolist()))
        joints = [
            b.add_joint_fixed(parent=-1, child=p, parent_xform=rot, child_xform=identity),
            b.add_joint_free(parent=p, child=link, parent_xform=identity, child_xform=identity),
        ]
    else:
        joints = [b.add_joint_free(parent=-1, child=link, parent_xform=rot, child_xform=identity)]
    b.add_articulation(joints)
    model = b.finalize()
    state = model.state()
    inv = model.inverse_dynamics()
    newton.eval_inverse_dynamics(model, state, newton.InverseDynamics.EvalType.ALL, inv)
    newton.eval_inverse_dynamics_force(
        model, state, inv.mass_matrix, wp.array(qddot_np, dtype=wp.float32), inv.coriolis_force, inv.gravity_force, inv.tau
    )
    return inv.tau.numpy()

print("root:    ", tau(False))
print("non-root:", tau(True))
```

</details>

```
root:     [ 1.0 -1.4 -0.6  0.06  0.24  0.2 ]   # world-frame wrench, matches the documented contract
non-root: [ 1.0 -0.6  1.4  0.06  0.2  -0.24]   # unrotated parent-frame wrench
```

Could we rotate each FREE/DISTANCE joint's 6-DOF block by its own `q_p` (looping over `[joint_start, joint_end)` like the converter kernel does), or, if non-root FREE/DISTANCE joints are out of scope for inverse dynamics, reject them in `Model.inverse_dynamics()` like CABLE and document the restriction?

newton/_src/utils/selection.py
```diff
@@ -1710,6 +1720,39 @@ def eval_mass_matrix(self, state: State, H=None, J=None, body_I_s=None, joint_S_
             self.model, state, H, J=J, body_I_s=body_I_s, joint_S_s=joint_S_s, mask=articulation_mask
         )
 
+    def eval_inverse_dynamics(
+        self,
+        state: State,
+        eval_type: InverseDynamics.EvalType,
+        inverse_dynamics: InverseDynamics,
+        mask: wp.array[bool] | None = None,
```
adenzler-nvidia: The docstring documents both the 1-D `[world_count]` and 2-D `[world_count, count_per_world]` mask forms and `get_model_articulation_mask` handles both, but the annotation only admits the 1-D form. `eval_fk` above annotates this as `wp.array[bool] | wp.array2d[bool] | None`.

```suggestion
        mask: wp.array[bool] | wp.array2d[bool] | None = None,
```

gyeomannvidia: The problem here is technical debt in eval_jacobian and eval_mass_matrix.

I cannot change the annotation to wp.array2d in eval_inverse_dynamics because I would need to change the annotations in eval_jacobian and eval_mass_matrix and ensure that they actually support  the new annotation with tests.

This is beyond the scope of the current change and should be implemented in a separate PR that begins with eval_jacobian and eval_mass_matrix before spreading out to functions that call them in turn.  

I created a new issue for this

https://github.com/newton-physics/newton/issues/3355

newton/_src/sim/articulation.py
gyeomannvidia: Well, we can do this but we should figure out first if we should do it.

The key question here is whether we truly support free joints that are not applied to the root link.

The mjcf loader requires that free joints are applied to the root link.
assert parent == -1, "Free joints must have the world body as parent"

USD also does not support free joints applied to non-root links. 

The urdf loader, on the other hand, does support free joints on non-root links.

Does it make sense at all to support this feature?  This is beyond the scope of the current PR. Given that we have no tests for this, it makes little sense to dive in with a test in inverse dynamics. 

I added a ticket for this:

https://github.com/newton-physics/newton/issues/3357

adenzler-nvidia: I agree that the broader question of whether Newton should support non-root FREE joints belongs in #3357. However, Newton's public API currently supports this topology, and the URDF loader can produce it. This PR's public inverse-dynamics API accepts such models but silently returns forces with inconsistent frames.

Please align inverse dynamics with Newton's current behavior by rotating every FREE/DISTANCE joint's `H @ qddot` block using that joint's parent-frame rotation, and add the non-root regression case. If Newton later decides to deprecate this topology, that can be handled as a separate API change.

newton/_src/utils/selection.py
adenzler-nvidia: Is the concern here just consistency with the currently untyped `ArticulationView.eval_jacobian()` and `eval_mass_matrix()` signatures? I may be missing a dependency, but it looks like the 2-D mask is consumed entirely at the view layer: [`get_model_articulation_mask()`](https://github.com/newton-physics/newton/blob/1af1d6d87d439b2869735082351a9af95f2a27dc/newton/_src/utils/selection.py#L1631-L1657) already accepts both forms and converts either one into a flat 1-D model articulation mask. The [existing selection test](https://github.com/newton-physics/newton/blob/1af1d6d87d439b2869735082351a9af95f2a27dc/newton/tests/test_selection.py#L455-L464) also covers this 2-D conversion. `eval_inverse_dynamics()` only forwards the resulting 1-D mask to the lower-level functions, and [`eval_fk()` already annotates the same view-level input](https://github.com/newton-physics/newton/blob/1af1d6d87d439b2869735082351a9af95f2a27dc/newton/_src/utils/selection.py#L1659-L1663) as accepting both forms.

Would it therefore be sufficient here to correct this signature and add a 2-D case to the existing inverse-dynamics view test, while leaving #3355 for consistency and annotation cleanup across the sibling methods? Or is there another path where the 2-D view mask reaches `eval_jacobian()` or `eval_mass_matrix()` directly?

newton/_src/sim/articulation.py
gyeomannvidia: We don't support this across all solvers even with the Newton API.

I created a simple test with two free joints in sequence.  I applied a control force for 1 sim step and compared the velocity output of mujoco, featherstone, xpbd, vbd.

Featherstone completes the sim step and the output looks good.

MuJoCo does not complete the sim step and just throws an error.
ValueError: Error: free joint can only be used on top level
Element name 'body_1', id 2

xpbd completes the sim step and matches Featherstone approximately.

vbd completes the sim step but computes output that is nowhere near the Featherstone output.

I guess my point is that we say that we support this but we really do not and should proceed accordingly.  Adding more code that we can't properly test feels like a wrong step.  I would recommend supporting this in inverse dynamics only after ironing all this out.

adenzler-nvidia: What I would like to avoid is here is adding more to the confusion, and in terms of API direction I think we are looking at a end-state where non-root free joints are supported across all of newton. So I would prefer not to add a limitation here and actually align this feature with what is supported in the newton API, without looking at solvers.

The fact that mujoco does not work is a limitation of MuJoCo, same for VBD. both are fixable. Unless we agree that we would like to disallow non-root free joints completely, we should work on fixing these gaps in a follow up rather than letting this code go stale as well.

I do agree that it's a mess right now and it's unclear what we support and what not.

newton/_src/utils/selection.py
gyeomannvidia: The problem is that eval_jacobian does not expect the mask to be of type array2d.  The annotation is for a 1d array and the docstring states a 1d array. 

Here's the docstring:
        mask: Optional boolean mask to select which articulations to compute.
              Shape [articulation_count]. If None, computes for all articulations.

The kernel eval_articulation_jacobian also expects a 1d array:

    if articulation_mask:
        if not articulation_mask[art_idx]:
            return

eval_mass_matrix has the same pattern.

Supporting array2d is beyond the scope of this PR. We can and should support this but it requires changes way beyond inverse dynamics.

newton/_src/sim/articulation.py
gyeomannvidia: If I believed that we could make MuJoCo support non-root free joints, I would take a different path. However, this is way beyond our control. 

Although we state that we support non-root free joints, we in practice do not.  This makes it difficult to test any code that has free joints applied to non-root links. If we cannot test something, it is better not to add support for it until it can be tested.  I added an issue for this and pointed out that inverse dynamics needs attention.

adenzler-nvidia: MuJoCo cannot compile a non-root `mjJNT_FREE` literally, but this does not prevent `SolverMuJoCo` from supporting the Newton topology. During model conversion, we can split the MuJoCo tree at a non-root FREE joint and attach that child subtree to `worldbody` using a native free joint. At the bridge boundary, we compose the Newton parent-relative pose/twist into world space before stepping and recover the relative values afterward. Joint forces can be mapped to equal-and-opposite wrenches on the original child and parent. The existing bridge already performs analogous frame, CoM, and body-origin conversions for root FREE joints.

That conversion work belongs in #3357, but it means the current MuJoCo limitation is an adapter gap rather than a fundamental inability to support or test this topology. The inverse-dynamics regression is analytical and does not depend on MuJoCo. For this PR, please either handle every non-root FREE/DISTANCE block consistently or reject those models explicitly rather than returning forces with mixed frames.

newton/_src/utils/selection.py
adenzler-nvidia: Sorry for the confusion—my comment was only about the `ArticulationView.eval_inverse_dynamics()` interface, not the module-level `eval_inverse_dynamics()`, `eval_jacobian()`, or `eval_mass_matrix()` interfaces.

The view method accepts a view-local 1-D or 2-D mask and passes it through `get_model_articulation_mask()`, which always returns a flat `wp.array[bool]` over the model’s articulations. Only that flat 1-D mask reaches the lower-level functions and kernels, so they do not need to change.

Could we update only the view-method annotation to match its existing docstring and the established `ArticulationView.eval_fk()` pattern, and add a 2-D case to the inverse-dynamics view test? Issue #3355 can separately cover any broader annotation cleanup.

gyeomannvidia: 3e977847154749445f3e22bdf0ca861fb566ef9a

newton/_src/sim/articulation.py
gyeomannvidia: c10f22c1e2a8e2db8cd866c07160b54831d9e9fc

adenzler-nvidia (approved):
Thanks for the thorough iteration on this. The inverse-dynamics API and frame conventions now look consistent, including non-root FREE/DISTANCE joints and `ArticulationView` masking. I rechecked the latest changes and have no remaining blocking concerns.

PR #2754, opened 2026-05-07
Fix vbd contact offset handling for friction anchors

camevor:
## Description

Fix VBD rigid-rigid friction to use the contact surface anchor (`rigid_contact_point + rigid_contact_offset`) when measuring tangential slip.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_collect_rigid_contact_forces_reports_surface_points
```

## Bug fix

**Steps to reproduce:**

1. Create a VBD capsule lying on its side on the ground.
2. Give it axial angular velocity with no initial linear velocity.
3. Step the simulation with rigid contact friction enabled.
4. Without this fix, tangential slip is evaluated at the skeleton contact point, so axial spin does not dissipate correctly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Rigid-body contacts now carry per-contact surface anchor offsets end-to-end, preserving them across warm-start/snapshots and using anchors for tangential displacement/friction while keeping normals based on geometry.

* **Tests**
  * Added regression tests for contact surface-point reporting and frictional dissipation of axial spin; updated existing contact-history tests to validate offset behavior.

* **Documentation**
  * Changelog updated noting the rigid-rigid friction fix for per-contact offsets.

* **Chores**
  * Increased kinematic gripper substeps for more stable friction tracking.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2754?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -178,6 +178,7 @@
 - Fix `eq_objtype` mismatch for joint equality and mimic constraints in `SolverMuJoCo` so compiled models match native MuJoCo XML behavior
 - Fix implicit-MPM rheology solver launch-dim handling under `warp-lang` 1.13's templated `launch_bounds` (formerly produced out-of-bounds reads)
 - Fix `SolverKamino.reset` clobbering `q_j_p`, `q_j`, and `dq_j` for worlds outside `world_mask` when `joint_q`/`joint_u` targets were provided. The previous unmasked writes broke TWOPI revolute-joint angle unwrapping after partial-mask resets.
+- Fix rigid-rigid friction in `SolverVBD` for contacts with nonzero `rigid_contact_offset0/rigid_contact_offset1`.
 
 ## [1.1.0] - 2026-04-13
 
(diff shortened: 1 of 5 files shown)


adenzler-nvidia:
@camevor could you please fix the tests?

nvtw:
One thing when contact matching is on (SolverVBD(..., rigid_contact_history=True)). AI suggests that the history buffer would need to store the offsets as well to avoid mixing current frame and previous frame (matched) data.

I just noticed that this seems to be the same concern as CodeRabbit raised above.

camevor:
Thanks for the review, @nvtw! 
> the history buffer would need to store the offsets as well to avoid mixing current frame and previous frame (matched) data.

Thanks, good spot! Fixed in the latest commit. I'll implement your suggestion to add a helper function as part of #2904.

nvtw (approved):

PR #2799, opened 2026-05-07
Added cable cross slide sample

felixmey:
## Description

The MR adds a cable driven cross slide table example using the VBD solver based on https://www.mathworks.com/help/sm/ug/cable-driven-xy-table-with-cross-base.html

image

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **New Examples**
  * Added a cable-driven cross-slide table simulation demonstrating multi-pulley cable routing, kinematic pulley-driven rectangular-path motion, and runtime validation of cable bounds and joint travel.

* **Tests**
  * Added an automated test to run the new cable example for extended frames with viewer-enabled validation.

* **Documentation**
  * Updated the examples index to list the new cable cross-slide table example and its run command.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2799)


newton/examples/cable/example_cable_cross_slide_table.py
```diff
@@ -0,0 +1,1121 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
felixmey: Added

jumyungc:
@felixmey Thank you for adding the great example! Can you also please add an image / update README.md?
And I think it is okay not to modify CHANGELOG.md when adding an example.

felixmey:
> @felixmey Thank you for adding the great example! Can you also please add an image / update README.md? And I think it is okay not to modify CHANGELOG.md when adding an example.

Removed the changelog entry and added it to the samples in the Readme

jumyungc (approved):

newton/examples/cable/example_cable_cross_slide_table.py
```diff
@@ -0,0 +1,1121 @@
...
+        table_pos = body_q[self.table_body, 0:3]
+        joint_limit_tolerance = 0.005
+        if not (-0.07 - joint_limit_tolerance <= slide_pos[0] <= 0.07 + joint_limit_tolerance):
+            raise ValueError("Horizontal green carriage moved outside its travel range.")
+        if not (-0.08 - joint_limit_tolerance <= table_pos[1] <= 0.08 + joint_limit_tolerance):
+            raise ValueError("Vertical beige carriage moved outside its travel range.")
+        if not (0.0 <= slide_pos[2] <= 0.04 and 0.0 <= table_pos[2] <= 0.05):
+            raise ValueError("Table bodies left the ground-plane layout.")
+
+        missed_points = np.nonzero(self.table_rect_min_distances > TABLE_RECT_HIT_TOLERANCE)[0]
```
eric-heiden, line 1098: 🔴 This rectangle coverage check is too loose to prove that the table actually traveled the rectangle.

At the origin, the rectangle corner targets are about `0.078 m` away, but `TABLE_RECT_HIT_TOLERANCE` is `0.1 m`. That means `missed_points` is empty even if the table never moves, so `test_final()` can pass without validating the core behavior of the example.

Please tighten the tolerance below the initial corner distance and add an explicit motion or range assertion, or validate expected post-ramp corner visits instead of only checking the minimum distance against this tolerance.

eric-heiden:
🟡 This PR adds a new user-facing example, but `CHANGELOG.md` still does not include an `[Unreleased]` entry for it.

The new `python -m newton.examples cable_cross_slide_table` example and README gallery entry are visible to users, so they should be listed under `### Added`. A suitable entry would be:

```md
- Add cable-driven cross-slide XY table example (`python -m newton.examples cable_cross_slide_table`) modeled on the Simscape Multibody cable-driven XY table.
```

newton/examples/cable/example_cable_cross_slide_table.py
jumyungc: @felixmey Please have a look if it is a valid concern and follow up if needed.

jumyungc:
Merging to keep this PR from going stale. If a CHANGELOG entry is needed, I will follow up.

PR #2808, opened 2026-05-07
Add draw_edges wireframe overlay to GL viewer

AnkaChan:
## Description
image

Split the wireframe edge overlay out of #2300 per @mmacklin's review.

Add `renderer.draw_edges` toggle that redraws mesh geometry as lines on top of solid shading. Uses `GL_POLYGON_OFFSET_LINE` depth bias to eliminate z-fighting between overlay lines and shaded triangles. Ground plane (checker surfaces) is excluded from the overlay via fragment discard.

> The wireframe overlay also has significant z-fighting that should be fixed with `GL_POLYGON_OFFSET` depth bias.
>
> For this PR I would suggest to separate out the addition of the wireframe overlay from the rest of the styling / shading changes, i.e.: just add the `draw_edges = True` feature.
>
> — @mmacklin (#2300)

## Checklist

## Test plan

```bash
uv run --extra examples --extra viewer -m newton.examples cloth_franka
# Toggle "Show Edges" checkbox in the left panel → verify edges render without z-fighting
# Verify ground plane does not show wireframe edges
# Untoggle → verify no visual regression
```

## New feature / API change

```python
import newton

# Enable edge overlay programmatically
viewer.renderer.draw_edges = True

# Or toggle via the "Show Edges" checkbox in the sidebar UI
```

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in wireframe edge overlay. A new "Show Edges" checkbox in the Visualization panel toggles subtle-colored wireframe edges on top of solid geometry to help inspect shapes. The overlay is available for most geometry types and is disabled by default for flat planes.


newton/_src/viewer/gl/shaders.py
```diff
@@ -944,3 +944,50 @@ def update_frame(
...
+{
+    if (Material.z > 0.0)
+        discard;
+    FragColor = edge_color;
+}
+"""
+
+
+class ShaderEdge(ShaderGL):
+    """Flat-color shader for the edge/wireframe overlay pass."""
```
eric-heiden, line 968: 🔴 The edge overlay is coupled to `Material.z`, which is the checker-material flag, not an edge-rendering flag. The new fragment shader discards whenever `Material.z > 0.0`, so the ground plane loses edges, and any caller that enables checker shading on another mesh will silently lose the wireframe overlay too.

That makes `draw_edges` depend on an unrelated public material bit. It also means a future material change can break edge rendering without touching the overlay code.

Please move this decision out of the shader and onto the rendered object, for example with a `draw_edge` flag similar to `cast_shadow`, then filter objects before the edge pass calls `_draw_objects(...)`. If the goal is only to skip the ground plane, that intent should be represented directly instead of inferred from `checker_enable`.

AnkaChan: Thanks @eric-heiden — fixed by decoupling the overlay from the material bit entirely (in `cee4ae66`).

- `edge_fragment_shader` no longer reads `Material`; the `Material.z > 0.0` discard is gone, so the shader is flag-agnostic.
- Added a per-object `draw_edge` flag mirroring `cast_shadow` — the edge pass now filters before drawing, exactly like `shadow_objects` in `_render_shadow_map`:
  ```python
  edge_objects = {k: v for k, v in objects.items() if getattr(v, "draw_edge", True)}
  ...
  self._draw_objects(edge_objects)
  ```
- The "skip the ground plane" intent is now explicit and geometry-driven: plane shape batches set `draw_edge = shapes.geo_type != GeoType.PLANE` in `viewer_gl.py`, with no dependency on `checker_enable`.

Net effect: visually unchanged (planes have no edges, everything else does), but a checker-shaded *non-plane* mesh now correctly keeps its wireframe — the silent-loss case you flagged. The `GL_POLYGON_OFFSET_LINE` z-fighting fix is preserved.

(Also merged current `main` to resolve the conflict; the sidebar "Show Edges" toggle moved with the UI refactor into the new `viewer_gui.py`.)

CHANGELOG.md
```diff
@@ -74,6 +74,7 @@
 - Add linear HDR color output support to `SensorTiledCamera` via `hdr_color_image`.
 - Add composable actuator subsystem with pluggable `Controller` (`ControllerPD`, `ControllerPID`, `ControllerNeuralMLP`, `ControllerNeuralLSTM`), `Clamping` (`ClampingMaxEffort`, `ClampingDCMotor`, `ClampingPositionBased`), and `Delay` components; supports per-DOF delays, CUDA graph capture, and masked environment reset
 - Add heatmap rendering for scalar arrays logged through `ViewerGL.log_array()`
+- Add edge overlay toggle (`renderer.draw_edges`) for wireframe visualization on top of solid geometry
```
eric-heiden: 🔴 This `renderer.draw_edges` changelog entry is under `## [1.2.0] - 2026-05-12`, which is already released. New PR entries need to go under `## [Unreleased]`, in the `### Added` block, so the next release process picks them up and the published `1.2.0` notes are not changed retroactively.

Please move this line into the `### Added` section under `[Unreleased]`.

eric-heiden (approved):

PR #2815, opened 2026-05-08
Improve guard in actuator tests

### LINKED ISSUE #2813: [BUG] Some actuator tests fail on cuda 12.4 due to improper guarding
### Bug Description

Some actuator tests use PyTorch.
These tests are guarded by the presence of PyTorch, but this is not enough.
PyTorch could be present but not compatible, in that case the UTs error out instead of being skipped.

We should rather follow the same pattern as in the example tests:
https://github.com/newton-physics/newton/blob/2f557a63fd182fcfc27f58321f47711e039eb3e4/newton/tests/test_examples.py#L91

### Reproduction Script

```python
Run the actuator tests with cuda 12.4.
```

jvonmuralt:
## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **Tests**
  * Moved PyTorch/CUDA availability checks into per-test setup for more accurate, context-aware skipping and clearer messages
  * Deferred non-PyTorch gating so tests run when prerequisites are truly met, improving reliability
  * Refactored test helpers to lazily create neural-network fixtures, simplifying checkpoint setup and reuse


newton/tests/test_actuators.py
```diff
@@ -48,10 +47,29 @@
...
+    except Exception as exc:
+        return False, None, f"Torch unavailable: {exc}"
+
+    if wp.get_device().is_cuda and not torch.cuda.is_available():
+        return False, None, "Torch not compiled with CUDA support"
+
+    return True, torch, ""
+
+
+_HAS_TORCH, _torch, _TORCH_SKIP_REASON = _get_torch_test_support()
```
Kenny-Vilella: Just wonder, will it cause problem that the check is at the file level instead of being at the test level?

newton/tests/test_actuators.py
```diff
@@ -187,14 +188,24 @@ def _f(vals):
...
+                # Ensure torch has CUDA support
+                self.skipTest("Torch not compiled with CUDA support")
+
+        except Exception as e:
+            self.skipTest(f"{e}")
+
+        self.torch = torch
         self.device = wp.get_device()
-        self._torch_dev = _torch.device(f"cuda:{self.device.ordinal}" if self.device.is_cuda else "cpu")
+        self._torch_dev = torch.device(f"cuda:{self.device.ordinal}" if self.device.is_cuda else "cpu")
```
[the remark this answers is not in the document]
jvonmuralt: Intentionally consistent with the existing pattern in newton/tests/test_examples.py

newton/tests/test_actuators.py
jvonmuralt: dont think it causes a problem, but I switched to a test level check to be safe

Kenny-Vilella (approved):
Tested it and it fixes the issue.
Installed CPU torch and made sure it fail on main but pass in this branch.

PR #2822, opened 2026-05-08
Kamino - Fix SolverKamino body-frame mismatch with Newton model

aserifi:
This now solves the body-frame absorption workaround by making Kamino's joint constraints handle `q_pj != q_cj` natively, so the rotated frame is gone entirely instead of being plumbed through `body_corr` and applied at every `step` / `reset`.

## Description

Newton allows the parent-side (`joint_X_p.rotation`) and follower-side (`joint_X_c.rotation`) of a joint to carry different rotations, and Newton's own `eval_fk` produces `q_F = q_B * q_pj * inv(q_cj)` at `q = 0`. Kamino's joint constraint code was originally written for a single joint frame, so `from_newton` absorbed the offset `q_corr = q_cj * inv(q_pj)` into the child body's local frame, mutating `model.body_com`, `model.body_inertia`, `model.shape_transform`, `model.joint_X_p`, and `model.joint_X_c` in place. The previous patch on this branch stopped the in-place writes into `model.*` and instead carried `body_corr` as a per-body quaternion that was multiplied into `state.body_q` on entry to `solver.step` / `solver.reset` and inverted on exit. That works, but it pushes a transient rotation through `state.body_q` mid-call, ships a quaternion of model state that is identity on every aligned-joint robot, and adds five extra kernel launches per step.

This PR now consumes both joint frames directly in Kamino's joint constraint formulation:

- **Two joint axes matrices per joint instead of one.** `JointsModel.X_j` is replaced with `X_p_j` (parent-side, equal to today's `X_j`) and `X_c_j` (follower-side). For aligned joints the converter writes `X_p_j == X_c_j`, so the data layout is bit-identical for every existing model. The internal builder gains `X_c_j: mat33f | None = None` defaulting to `X_p_j`, so all existing call sites in `models/builders/` keep working with a mechanical `X_j=` to `X_p_j=` rename.
- **Constraint kernels read both sides.** `compute_joint_pose_and_relative_motion`, `compute_joint_relative_quaternion`, the dense and sparse joint Jacobian builders, the FK reset kernels, and the FK constraint kernels all now build `q_Bj = q_B * q_X_p_j` on the parent side and `q_Fj = q_F * q_X_c_j` on the follower side. At `q = 0` Newton's FK gives `q_F = q_B * q_pj * inv(q_cj)`, so `inv(q_Bj) * q_Fj = identity` automatically with no body-frame rotation needed. The FK position-control kernel absorbs the constant `q_X_p_j * inv(q_X_c_j)` offset into `pos_control_transforms` (with an element-wise equality check that skips the multiplication when the matrices are equal, preserving bit-identical behaviour for aligned joints), so all four FK constraint and Jacobian kernels keep using the simple `q_F = q_B * q_control_body` formula.
- **Conversion is now passive.** `convert_entity_local_transforms` collapses to `wp.clone` of Newton's body-local arrays. `joint_conversion_kernel` writes `X_p_j = R(q_pj) @ R_axis` and `X_c_j = R(q_cj) @ R_axis` from the unmodified `model.joint_X_p`, `model.joint_X_c`, and `model.joint_axis`. `entity_local_transform_conversion_kernel` and `shape_transform_conversion_kernel` are deleted.

The new contract: every body-local quantity Kamino exposes (`body_q`, `body_qd`, `body_com`, `body_inertia`, `shape_transform`, `joint_X_p`, `joint_X_c`) stays in Newton's frame end-to-end. `eval_fk`, Newton's collision pipeline, the renderer, and IsaacLab now all see the model as authored. Aligned-joint behaviour is bit-identical to the previous version; non-aligned joints are handled by the constraint formulation rather than by rotating bodies.

## Checklist

## Test plan

New regression suite in `test_solver_kamino_joint_frames.py` with five tests, all passing:
- `test_conversion_does_not_mutate_newton_arrays` confirms `model.body_com`, `model.body_inertia`, `model.shape_transform`, `model.joint_X_p`, `model.joint_X_c` are bit-equal before and after `ModelKamino.from_newton` on a model with non-aligned joint frames.
- `test_aligned_joint_frames_are_bit_identical` and `test_non_aligned_joint_frames_x_p_j_differs_from_x_c_j` lock down the converter's two-side output.
- `test_kamino_joint_X_p_j_matches_newton_joint_frame` verifies the parent-side axis matches `R(q_pj) @ R_axis` from Newton's authored joint frame.
- `test_step_runs_without_nan_on_non_aligned_joint` runs five solver steps under gravity on a non-aligned-joint model and asserts finite poses, unit-quaternion invariant, and that the body actually moved.


newton/_src/solvers/kamino/_src/core/joints.py
```diff
@@ -1085,8 +1085,19 @@ class JointDescriptor(Descriptor):
     """The relative position of the joint in the follower body coordinates."""
 
     # TODO: Remove this when body offsets become transforms
-    X_j: mat33f = field(default_factory=mat33f)
-    """The constant axes matrix of the joint."""
+    X_p_j: mat33f = field(default_factory=mat33f)
+    """The constant axes matrix of the joint expressed in the base body's frame.
```
Guirec-Maloisel: About the naming: we should call this something like X_Bj and X_Fj to match the base-follower convention (we don't talk about parent and child in Kamino). See for instance B_r_Bj and F_r_Fj for the local positions. @vastsoun what would be the correct consistent name here (I know you're using some conventions for local/global frames, so maybe it should be B_X_Bj? not clear to me anymore why there are two B in B_r_Bj)?

Also I'm not sure why the AI's blabbing about R(q_pj) @ R_axis but I'd remove this.

newton/_src/solvers/kamino/_src/core/joints.py
```diff
@@ -1459,6 +1470,11 @@ def __post_init__(self):
                 f"Invalid joint: FIXED joints cannot have implicit PD gains (name={self.name}, uid={self.uid})."
             )
 
+        # Default the follower-side joint frame to the base-side one, which
+        # is the convention for joints with aligned base/follower frames.
+        if self.X_c_j is None:
```
Guirec-Maloisel: Having separate local orientations on the base and follower implies in particular that we are not limited to body orientations being the identity on the initial pose anymore; so the local joint frames may not match the world coordinates joint frames on the initial pose anymore.

So for this code to be general, it should set the missing follower joint frame to align with the base follower joint frame on the initial pose in world coordinates, not in local coordinates. So there's a transform to be applied here based on the initial orientation of the base and follower bodies.

newton/_src/solvers/kamino/_src/core/model.py
```diff
@@ -702,33 +702,11 @@ def from_newton(model: Model) -> ModelKamino:
...
-        # This makes q_base^{-1} * q_follower_new = identity at q = 0, and the joint
-        # rotation axis R(q_pj) * axis is preserved.
-        #
-        # All body-local quantities (CoM, inertia, shapes) are re-expressed in the
-        # rotated frame, and downstream joint_X_p transforms are updated to account
-        # for the parent body's frame change.
-        # ---------------------------------------------------------------------------
+        # Get working copies of Newton's entity-local transforms. These remain in
+        # Newton's frame end-to-end: the joint constraint formulation consumes
+        # ``X_p_j`` and ``X_c_j`` separately, so no body-frame rotation absorption
```
Guirec-Maloisel: Not sure this comment is needed, it makes references to how the code used to be but that may be just confusing to someone reading this code for the first time.

newton/_src/solvers/kamino/_src/solvers/fk/kernels.py
```diff
@@ -171,21 +173,25 @@ def _reset_state_base_q(
...
         x_B = joints_B_r_B[base_jt_id]
         x_F = joints_F_r_F[base_jt_id]
         body_q_F_0 = bodies_q_0[bid_F]
 
-        # Compute pose of the base body (follower of the base joint) given current joint coordinates
-        # Note: the relative transform from base to follower can be written
-        # t_jt = X^T * R_B^T * (c_F + R_F * x_F - c_B - R_B * x_B)
-        # q_jt = X^T * R_B^T * R_F * X
-        # We invert these equations, using R_B = I and c_B = 0 (base body = world)
+        # Compute pose of the base body (follower of the base joint) given current joint coordinates.
```
Guirec-Maloisel: I definitely want to double-check the math and the notation on the whole FK solver (the comments are a bit confusing here to me, the notation is not always consistent). Maybe once the naming has converged I'll take a pass and overwrite the AI comments.

Guirec-Maloisel:
I went through this PR and cleaned up so that it is now in a more ready state. More specifically:
- I double-checked the formulation
- I changed the naming so it is consistent with Kamino notations (X_p_j -> X_Bj, X_c_j -> X_Fj, cf B_r_Bj and F_r_Fj)
- I cleaned up the extra verbose AI comments
- I further simplified the Newton -> Kamino conversions
- I split the joint conversion kernel into two (joint frames vs the rest) so that the joint frames part can also be called in convert_model_joint_transforms() (that code was so far running on CPU and there was a TODO to make a kernel for it)

newton/_src/solvers/kamino/_src/core/joints.py
Guirec-Maloisel: Upon further inspection making this more general would require access to rigid body initial poses, which are not exposed it. That could possibly be done in the builder finalize(), but that would mean keeping the field to None until then. Since this is a mostly a convenience thing anyway, and since we are slowly moving away from using Kamino's builder, I didn't deem it worth the effort in the end.

newton/_src/solvers/kamino/_src/core/joints.py
```diff
@@ -1076,17 +1076,21 @@ class JointDescriptor(Descriptor):
     Defaults to `-1`, indicating that the joint has not been assigned a follower body.
     """
 
-    # TODO: Change this to a transformf
```
Guirec-Maloisel, line 1079: Note: I removed the TODOs here to keep the code clean as we're moving towards 1.0 of Kamino. Is there still a plan at some point to use wp.transformf for these? If yes I can track this as an issue instead.

newton/_src/solvers/kamino/_src/solvers/fk/kernels.py
```diff
@@ -655,33 +668,43 @@ def _eval_position_control_transformations(
...
-                q_x = wp.quat_from_axis_angle(wp.vec3f(X[0, 0], X[1, 0], X[2, 0]), actuators_q[offset_q_j])
-                q_y = wp.quat_from_axis_angle(wp.vec3f(X[0, 1], X[1, 1], X[2, 1]), actuators_q[offset_q_j + 1])
+                q_x = wp.quat_from_axis_angle(wp.vec3f(X_B[:, 0]), actuators_q[offset_q_j])
+                q_y = wp.quat_from_axis_angle(wp.vec3f(X_B[:, 1]), actuators_q[offset_q_j + 1])
                 q = q_x * q_y
             else:
                 assert False, "Unexpected actuator dof type"  # noqa: B011
 
+        # If X_B != X_F, absorb the offset in q_rel so downstream kernels can keep using ``q_F = q_B * q_rel``
+        X_F = joints_X_Fj[jt_id]
```
chschuma-disney: Minor: This read is redundant, since `X_F` has already been read on line 656.

newton/_src/solvers/kamino/_src/core/conversions.py
```diff
@@ -1020,11 +806,11 @@ def convert_rigid_bodies(
...
         inv_m_i=model.body_inv_mass,
-        i_r_com_i=body_com,
-        i_I_i=body_inertia,
-        inv_i_I_i=body_inv_inertia,
+        i_r_com_i=wp.clone(model.body_com, device=model.device),
+        i_I_i=wp.clone(model.body_inertia, device=model.device),
+        inv_i_I_i=wp.clone(model.body_inv_inertia, device=model.device),
         q_i_0=q_i_0,
-        u_i_0=body_qd,
+        u_i_0=wp.clone(model.body_qd, device=model.device),
```
chschuma-disney, line 813: Are these clones necessary, or could we just reuse the Newton model arrays directly?

chschuma-disney (commented):
Fantastic work! Looks good to me, just two minor comments.

newton/_src/solvers/kamino/_src/core/conversions.py
Guirec-Maloisel: Yeah that's a question I expected to come up in the review process, I was hoping possibly to get Vassilios' point of view as well there. Due to the transforms we used to have, the previous behavior was that the converted Kamino model was neither modifying nor aliasing arrays of the Newton model, which I kept and documented explicitly. It could prevent some possible bad surprises in some cases, but also incurs additional copies. Maybe at some point this could be controlled by a flag in the top-level conversion routine.

newton/_src/solvers/kamino/_src/solvers/fk/kernels.py
Guirec-Maloisel: Good point, I pushed a fix.

newton/_src/solvers/kamino/_src/core/conversions.py
chschuma-disney: The cloning is a negligible operation, I think, so I'm happy to leave this as is. But it might be good to revisit at some point, since we do alias some other Newton model arrays (body mass, for example).

chschuma-disney (approved):

PR #2823, opened 2026-05-08
Fix device for gravity data allocation in Kamino

chschuma-disney:
## Description

When converting gravity from a Newton model to Kamino's gravity data structure, a specific code path failed to allocate new warp arrays on the same device the model data lives, instead allocating it on the default device. This specific case generally happens when an existing gravity data structure is updated, e.g., as a reaction to the solver being notified about an update to the gravity values.

At the same time, an issue with the size check of the existing array was addressed. The issue lead to a new array being allocated with every change, even if the size was compatible with the new gravity data.

To simplify the testing and in anticipation of the test migration, I added Kamino to the existing Newton gravity tests, instead of creating a new test in Kamino.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_runtime_gravity
```

## Bug fix

**Steps to reproduce:**

1. Add Kamino to `test_runtime_gravity.py`
2. Run `uv run --extra dev -m newton.tests -k test_runtime_gravity`
3. The CPU-based test for Kamino will fail

## Summary by CodeRabbit

* **Tests**
  * Gravity runtime tests now include the Kamino solver, increasing solver coverage.

* **Bug Fixes**
  * Fixed device-aware allocation for gravity model data to ensure correct behavior across compute devices.

### DIFF
--- a/newton/_src/solvers/kamino/_src/core/gravity.py
+++ b/newton/_src/solvers/kamino/_src/core/gravity.py
@@ -231,14 +231,14 @@ def convert_model_gravity(model_in: Model, gravity_out: GravityModel | None = No
     # correct shape and type, and copy the converted data into them.
     else:
         # Ensure that the output GravityModel has allocated arrays of the correct shape and type
-        if gravity_out.g_dir_acc is None or gravity_out.g_dir_acc.shape != (model_in.world_count, 4):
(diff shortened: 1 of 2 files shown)


vastsoun (approved):
LGTM, thanks @chschuma-disney

PR #2837, opened 2026-05-11
Add support for negative shape scale to collision pipeline

### LINKED ISSUE #2802: [REQ] Negative mesh scale is silently dropped in Newton's USD import pipeline.
### Description

In USD (and MuJoCo — see the [3.7.0 changelog](https://mujoco.readthedocs.io/en/stable/changelog.html#version-3-7-0-april-14-2026) for a related compiler fix on the MuJoCo side), an `xformOp:scale `with a negative-determinant value (e.g. (-1, 1, 1)) is legal and means the geometry is mirrored. The USD spec requires winding order to be auto-corrected so face normals continue to point outward.

A few related observations that are likley related to this
-`newton._src.usd.utils.get_scale extracts` scale via wp.transform_decompose, which returns an unsigned scale and discards the reflection.
-`shape_scale` in the builder is always non-negative for USD-imported shapes. Solvers (including the MuJoCo backend) never see the original negative scale, so any solver-side handling — even MuJoCo's recent fix — is moot for Newton-driven scenes.
-`ModelBuilder.add_shape_mesh` accepts a Vec3 scale with no sign validation. 

### Motivation / Use Case

This bridges the gap for a lot of USD/OV assets 

### Alternatives Considered

_No response_

nvtw:
## Description

Note: This MR does not address negative scale in usd or anywhere else besides the collision pipeline.

Add support for negative shape scale in the collision pipeline. Mesh, convex hull, and SDF shapes now accept signed (mirrored) scales natively, with triangle winding, AABB conversion, SDF gradient remapping, inertia, soft-contact margins, and viewer normals all corrected for `det(scale) < 0`. Symmetric primitives (sphere, box, capsule, cylinder, ellipsoid, plane, gaussian) silently absorb sign flips via `abs()` at `ModelBuilder.add_shape*` time. Cones and heightfields reject negative components (cones would flip the apex/base; heightfield mirroring is not yet implemented).

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_negative_scaling
uv run --extra dev -m newton.tests -k test_viewer_geometry_batching
uv run -m newton.examples basic_shapes_neg_scale
```

## New feature / API change

```python
import newton
import warp as wp

mesh = newton.Mesh(vertices, indices)

# Share one Mesh across shapes with different signed scales (mirror on X).
builder.add_shape(type=newton.GeoType.MESH, src=mesh, scale=(1.0, 1.0, 1.0))
builder.add_shape(type=newton.GeoType.MESH, src=mesh, scale=(-1.0, 1.0, 1.0))

# Symmetric primitives silently normalize negative components.
builder.add_shape_box(scale=(-0.5, 0.5, 0.5))  # treated as (0.5, 0.5, 0.5)

# Cones / heightfields reject negative scale.
builder.add_shape_cone(scale=(-1.0, 1.0, 1.0))  # ValueError
```

## Summary by CodeRabbit

* **New Features**
  * Added support for negative (mirrored) scales on mesh, convex hull, and SDF shapes, allowing meshes to be reused across shapes with different signed scales without re-baking.

* **Bug Fixes**
  * Fixed physics, collision detection, and rendering for mirrored geometry.
  * Improved scale validation in shape creation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2837?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


newton/_src/geometry/sdf_hydroelastic.py
```diff
@@ -509,7 +509,9 @@ def _from_model(
                 raise ValueError(f"Hydroelastic shape {idx} requires SDF data but has no attached/generated SDF.")
             if not texture_sdf_data[sdf_idx]["scale_baked"]:
                 sx, sy, sz = shape_scale[idx]
-                if not (np.isclose(sx, 1.0) and np.isclose(sy, 1.0) and np.isclose(sz, 1.0)):
+                # Allow mirrored unit scale (e.g. (-1, 1, 1)) since the SDF gradient
```
camevor: It seems like unbaked SDFs with negative scale still don't apply the scale in the hydroelastic path, so this guard should not be relaxed.

newton/_src/sim/builder.py
```diff
@@ -5505,6 +5510,37 @@ def add_shape(
...
+            GeoType.BOX,
+            GeoType.CAPSULE,
+            GeoType.CYLINDER,
+            GeoType.ELLIPSOID,
+            GeoType.PLANE,
+            GeoType.GAUSSIAN,
+        ):
+            scale = (abs(float(scale[0])), abs(float(scale[1])), abs(float(scale[2])))
+        elif type == GeoType.CONE:
+            if any(float(s) < 0.0 for s in scale):
```
camevor: nit: This seems overly strict given most other shapes apply or silently invert negative scaling. While negative `scale[1]` would require a rotation, negative `scale[0]` seems trivial.

camevor (dismissed):
Thanks! Two minor comments, looks good otherwise!

newton/_src/geometry/sdf_hydroelastic.py
nvtw: Done

newton/_src/sim/builder.py
nvtw: Done

camevor (approved):

Issue #1448, opened 2026-01-23
Solver Coupling VBD <> MuJoCo


preist-nvidia:
This epic collects issues related to working toward two-way coupling between robots simulated with the MuJoCo solver and deformables simulated by VBD.

### Goal
The primary goal is enabling robotic interaction with flexible assets, prioritized as follows:
- MJC + VBD Cables (P0): Enabling robots to manipulate tubes, wires, and cables.
- MJC + VBD Cloth (P1): Enabling manipulation of thin-shell deformables (e.g., bags, fabrics).
- MJC + VBD Solids (P2): Once we have solids support, unlikely in 1.0 timeline

 ### Integration considerations
We should enable asset-authoring users and define coupling parameters in USD schema. This is lower priority than the technical work toward getting coupling to work on a robotic manipulation workflow.

Also lower priority for the 1.0 milestone, we should understand the architecure for integrating coupling into IsaacLab for RL and other robotics simulation workflows.

jumyungc:
MJC + VBD Cables (P0): this is followed separately, https://github.com/newton-physics/newton/issues/1709.

PR #2848, opened 2026-05-12
Experimental coupled solver framework

### LINKED ISSUE #1709: Solver coupling: MJC Robot + VBD Cables
Support for Mujoco robot and VBD cable solver coupling.
This should naturally extend to handle general Mujoco and VBD solver coupling regardless of the types.

gdaviet:
## Description

Adds an experimental multi-solver coupling framework for Newton (#1448).

This PR introduces:

- `SolverCoupled`, the shared wrapper for partitioning a model across solver
  entries, distributing input state, stepping sub-solvers, and reconciling
  owned state.
- `ModelView`, a view-local model proxy used to expose solver-specific bodies,
  joints, particles, shapes, mass/inertia overrides, proxy flags, and contact
  filtering without mutating the parent model.
- `SolverProxyCoupled`, a lagged/staggered proxy-coupling scheme with
  body/particle proxies, virtual inertia, proxy-local contact preparation,
  velocity rewind, and force/impulse harvesting.
- `SolverAdmmCoupled`, a linearized ADMM coupling scheme over cross-solver
  model joints, body-particle attachments, and explicitly enabled
  cross-solver contact pairs.
- Experimental solver coupling hooks for effective mass queries, private
  input-state updates, proxy contact preparation, velocity rewind, and proxy
  feedback harvesting.
- Multiphysics examples covering MuJoCo/Kamino, MuJoCo/VBD, XPBD/VBD,
  XPBD/MPM, VBD/MPM, MuJoCo/MPM, ADMM contacts, proxy convergence, and the
  cable robot proxy path.

The public API is intentionally experimental and is exported under:

```python
from newton.solvers.coupled_experimental import (
    CouplingInterface,
    ModelView,
    SolverAdmmCoupled,
    SolverCoupled,
    SolverProxyCoupled,
)
```

Design / status docs:

- Coupling notes PDF: https://gdaviet.github.io/newton/coupled_examples/2026-04-23-admm-coupling.pdf
- Coupling concepts page: [`docs/concepts/coupling.rst`](docs/concepts/coupling.rst)
- Coupled example cards: https://gdaviet.github.io/newton/coupled_examples/

### Proxy Coupling

`SolverProxyCoupled` is the recommended entry point for virtual proxy coupling.
It maps source bodies or particles into destination proxy bodies/particles and
iterates lagged or staggered transfer plus destination-side feedback harvest.

```python
from newton.solvers import SolverVBD, SolverXPBD
from newton.solvers.coupled_experimental import SolverCoupled, SolverProxyCoupled

solver = SolverProxyCoupled(
    model=model,
    entries=[
        SolverCoupled.Entry(
            name="vbd",
            solver=lambda view: SolverVBD(model=view, iterations=8),
            particles=vbd_particles,
        ),
        SolverCoupled.Entry(
            name="xpbd",
            solver=lambda view: SolverXPBD(model=view, iterations=8),
            particles=xpbd_particles,
        ),
    ],
    coupling=SolverProxyCoupled.Config(
        proxies=[
            SolverProxyCoupled.Proxy(
                source="vbd",
                destination="xpbd",
                particles=interface_particles,
                mass_scale=0.5,
                mode="lagged",
            )
        ],
        iterations=3,
    ),
)
```

Proxy configuration supports:

- body and particle proxy maps through `bodies`, `proxy_bodies`, `particles`,
  and `proxy_particles`;
- virtual inertia scaling through `mass_scale`;
- lagged and staggered proxy transfer modes;
- optional proxy-local collision pipeline factories and collision refresh
  intervals;
- repeated proxy relaxation through `SolverProxyCoupled.Config.iterations`.

### ADMM Coupling

`SolverAdmmCoupled` implements linearized ADMM coupling over model-derived
constraints and explicitly enabled contact pairs.

```python
from newton.solvers import SolverMuJoCo, SolverVBD
from newton.solvers.coupled_experimental import SolverAdmmCoupled, SolverCoupled

SolverAdmmCoupled.add_body_particle_attachment(
    builder,
    body=ball_body,
    particle=cloth_particle,
    body_point=(0.0, 0.0, 0.08),
    stiffness=1.0e3,
    damping=1.0,
)

entries = [
    SolverCoupled.Entry(
        name="rigid",
        solver=lambda view: SolverMuJoCo(model=view, iterations=8),
        bodies=rigid_bodies,
        joints=rigid_joints,
    ),
    SolverCoupled.Entry(
        name="soft",
        solver=lambda view: SolverVBD(model=view, iterations=8),
        bodies=soft_bodies,
        particles=soft_particles,
    ),
]

solver = SolverAdmmCoupled(
    model=model,
    entries=entries,
    coupling=SolverAdmmCoupled.Config(
        iterations=4,
        rho=50.0,
        gamma=0.1,
        baumgarte=0.01,
        contact_pairs=SolverAdmmCoupled.auto_detect_contact_pairs(
            entries,
            detection_margin=0.02,
        ),
    ),
)
```

The ADMM path supports:

- cross-solver model `BALL`, `FIXED`, and `REVOLUTE` joints;
- custom `coupling:body_particle_attachment` model attributes;
- explicitly enabled rigid-rigid, rigid-particle, and particle-particle
  contact interfaces through `SolverAdmmCoupled.ContactPair`;
- Coulomb contact friction via a maximum-dissipation local solve;
- ball/revolute joint box friction;
- proximal mass scaling and proximal velocity targets through the shared
  `ModelView` and state-update machinery.

### Solver Coupling Hooks

Solvers that participate in coupled simulations inherit the experimental
`CouplingInterface` contract and override only the hooks they need:

- `coupling_eval_effective_mass()`
- `coupling_eval_effective_mass_block()`
- `coupling_notify_input_state_update()`
- `coupling_prepare_proxy_contacts()`
- `coupling_rewind_proxy_body_velocity()`
- `coupling_rewind_proxy_particle_velocity()`
- `coupling_harvest_proxy_wrenches()`
- `coupling_harvest_proxy_particle_forces()`

Missing hooks use generic wrapper fallbacks. Solvers can declare unsupported
hooks through `CouplingInterface.coupling_unsupported`. Instance attributes set
to `None` are treated as missing, which lets a solver disable a hook when
construction-time state is unavailable.

Force injection and model-view inertia/mass overrides are not custom hooks:
couplers write `state.body_f` and `state.particle_f` directly, mutate
`ModelView` mass/inertia arrays, then notify solvers through the normal
`notify_model_changed()` path when required.

### VBD

VBD provides custom proxy contact preparation, body-proxy harvesting, and
input-state update handling when it owns rigid AVBD state. The input-state hook
keeps `body_q_prev` consistent when the coupler synchronizes proxy poses or
restarts a coupling iteration. The proxy contact hook filters invalid proxy
contacts and controls rigid-history refresh cadence, while the harvest hook
reduces final rigid and body-particle contact forces onto proxy bodies.

### Implicit MPM

Implicit MPM gained `collider_particle_ids` support so mesh colliders can be
tied to finite-mass model particles and coupled as deformable proxy colliders.
It separates transfer and material particle flags: pure proxy particles remain
active for P2G/G2P transfers while material and stress contributions are
suppressed through `material_particle_flags`. The solver provides custom
body/particle proxy rewind and harvest hooks to sync collider caches, subtract
lagged gravity/feedback, and harvest grid or collider impulses.

### XPBD

XPBD particle contact kernels understand proxy particles and proxy bodies.
Owned particles can collide with destination proxy particles, while proxy-proxy,
proxy-static, and proxy-particle-versus-proxy-body contacts are filtered so the
destination solve does not create feedback between two proxy endpoints or
against immovable particles.

### MuJoCo

MuJoCo provides custom diagonal and block effective-mass hooks on the GPU path.
These hooks query MuJoCo Warp `body_invweight0` data and map it back to Newton
body ids, giving proxy virtual inertia and ADMM endpoint weights an articulated
effective-mass estimate instead of a raw body mass fallback. Multi-world
coupling supports non-contiguous Newton body partitions by using
`model.body_world` plus the existing `mjc_body_to_newton` map.

### SemiImplicit

SemiImplicit reserves and rebuilds the model particle hash grid before particle
contact evaluation, matching XPBD's solver-owned particle-contact broad-phase
pattern. Particle contact forces accumulate into the existing `particle_f`
buffer instead of overwriting it, preserving externally injected or coupled
particle forces.

### Kamino

Kamino model construction accepts `ModelView` instances by duck-typing the
required model fields instead of requiring an actual `newton.Model` instance.
This lets coupled examples pass a rigid-only Kamino view that hides unsupported
soft topology while still using the normal Kamino solver construction path.

## Current Status

The branch keeps the internal source layout under
`newton/_src/solvers/coupled/` and exposes the experimental public API as
`newton.solvers.coupled_experimental`.

The coupling concepts documentation describes model partitioning,
`ModelView`, proxy coupling, ADMM coupling, solver hooks, contact handling,
examples, limitations, and expected future direction.

Known follow-up:

- `ModelView` currently masks and prefix-compacts where possible; it does not
  perform full body/joint compaction. A future cleanup could build truly
  compact sub-model views with remapped body, joint, shape, state, world, and
  custom-attribute indices.
- `SolverMuJoCo` still builds from selected world/topology arrays rather than
  fully honoring `joint_enabled` during conversion. This is manageable with
  the current masking behavior, but full compaction would be the cleaner
  long-term fix.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Experimental multi-solver coupling: ADMM contact coupling (with Coulomb friction), proxy coupling (lagged/staggered), ModelView overlays, and deformable-collider support.
  * New solver implementations for coupled, proxy-coupled, and ADMM-coupled workflows.
  * Many new end-to-end multiphysics examples demonstrating coupled usage and convergence.

* **Bug Fixes**
  * Corrected particle contact force accumulation and soft-contact generation guard behavior.

* **Documentation**
  * Added coupling framework concept and API docs.

* **Tests**
  * Added broad ADMM and coupling unit/regression tests and example integration tests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2848?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


docs/conf.py
```diff
@@ -114,6 +114,16 @@
     "sphinx-env",
     "**/site-packages/**",
     "**/lib/**",
+    "plans/2026-03-23-coupling-sdd-review.md",
+    "plans/2026-03-24-coupling-implementation-status.md",
+    "plans/2026-04-24-admm-implementation-status.md",
```
shi-eric: Seems odd to add these very specific files to the ignore list, but how about just adding a `"plans"` to this list?

gdaviet: Huh, sorry I did not notice this change. It looks like the whole file could be reverted

adenzler-nvidia:
Thanks for pushing this through. This is a substantial piece of work, and the overall direction looks strong: keeping the API under an experimental namespace, making ownership explicit through solver entries, and adding both proxy and ADMM paths gives us a useful foundation without pretending the numerical/API surface is fully settled yet.

I took a principal-architect pass and found a few issues I think should be addressed before merging:

1. **High: `ModelView` can freeze copied inertial arrays across runtime model updates.**

   The view overlay is currently a mix of live parent reads and copy-on-write cloned arrays. During coupled construction, non-owned body masking calls `view.disable_body_dynamics(...)`, which clones arrays such as `body_inv_mass` and `body_inv_inertia` into the view. After that, `SolverCoupled.notify_model_changed()` only forwards flags to sub-solvers; it does not refresh or rebuild those view overrides.

   I reproduced this locally: after changing `model.body_inv_mass` and calling `notify_model_changed(SolverNotifyFlags.BODY_INERTIAL_PROPERTIES)`, `s.view("a").body_inv_mass` still contained the old value. That means domain randomization or any runtime mass/inertia edit can leave coupled sub-solvers using stale inertial data.

   Relevant code:
   - `newton/_src/solvers/coupled/model_view.py` around `_cow_array()` / `disable_body_dynamics()`
   - `newton/_src/solvers/coupled/solver_coupled.py` around entry construction masking and `notify_model_changed()`

   Architecturally, I think `ModelView` needs an explicit refresh path for overrides derived from parent arrays, or those overlays need to be represented as masks/transforms that are recomputed from the current parent data when model-change notifications arrive.

2. **High: the concepts docs describe an older `Entry` API.**

   `docs/concepts/coupling.rst` still shows:

   ```python
   SolverCoupled.Entry(
       name="soft",
       solver=SolverVBD,
       ...
       solver_kwargs={"iterations": 20},
   )
   ```

   and says construction calls `solver(model=view, **solver_kwargs)`. The actual API now takes a pre-bound `Callable[[ModelView], SolverBase]`, for example `lambda v: SolverVBD(model=v, iterations=20)`. The same section also mentions `zero_body_mass()`, which is not a `ModelView` method.

   Since this is an experimental public API, the concepts page is likely to be the first thing early adopters copy from. It should match the implementation exactly.

3. **Medium: the changelog entry landed under the released `1.2.0` section instead of `[Unreleased]`.**

   The coupled-solver entry is currently in the `1.2.0` Added section. Per the repo PR instructions, user-facing behavior for this PR should be recorded in `[Unreleased]`, at a random position in the correct category.

4. **Medium: new examples are not fully wired into the example registry/docs expectations.**

   This PR adds many runnable multiphysics examples, but `README.md` was not updated with the corresponding `python -m newton.examples <name>` entries and screenshots. `test_examples.py` also only registers two of the new examples. If some examples are intentionally not in regular CI because they are expensive, it would help to document that split explicitly, but the README/screenshot requirement still looks unmet for public examples.

One operational note: GitHub currently reports the PR as not mergeable, and a local merge-tree check shows conflicts in `CHANGELOG.md`, `newton/tests/test_examples.py`, and `newton/tests/test_implicit_mpm.py`. The CI on the current head is green, so this looks like a rebase/merge cleanup rather than a branch-health issue.

docs/api/newton_solvers.rst
```diff
@@ -406,10 +406,12 @@ See the `DiffSim examples on GitHub`_ for the current reference workflows.
 .. toctree::
    :hidden:
 
+   newton_solvers_coupled_experimental
```
mmacklin: I would maybe just put this under `newton.solvers.experimental` to avoid a proliferation of experimental namespaces.

docs/api/newton_solvers_experimental_coupled.rst
```diff
@@ -0,0 +1,21 @@
...
+
+newton.solvers.coupled_experimental
+===================================
+
+.. py:module:: newton.solvers.coupled_experimental
+.. currentmodule:: newton.solvers.coupled_experimental
+
+.. rubric:: Classes
+
+.. autoclass:: CouplingInterface
```
mmacklin, line 16: I would consider just making this interface part of `SolverBase` (with some way to signal level of support).

docs/concepts/coupling.rst
```diff
@@ -0,0 +1,364 @@
...
+factory is supplied, the destination solve uses contacts passed to the outer
+``step()`` call.
+
+The generic proxy loop currently supports at most two solver entries. Within
+that limit, body and particle mappings are grouped by ``(source,
+destination)``. One source step and one destination step are performed for each
+solver pair and proxy iteration, so a single proxy declaration can carry both
+body and particle mappings around the same destination solve.
+
+ADMM Coupling
```
mmacklin, line 220: Does ADMM coupling also require solvers to be "proxy" body aware? If not it definitely seems like the more elegant / less invasive option.

newton/_src/solvers/coupled/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+from .interface import CouplingInterface
+from .model_view import ModelView
+from .solver_admm_coupled import SolverAdmmCoupled
+from .solver_coupled import SolverCoupled
+from .solver_proxy_coupled import SolverProxyCoupled
+
+__all__ = [
+    "CouplingInterface",
+    "ModelView",
+    "SolverAdmmCoupled",
```
mmacklin: If we were being strict on our common prefix-first convention then I think these would be:

```
SolverCoupled
SolverCoupledAdmm
SolverCoupledProxy
```

Which would nicely group them also :)

newton/_src/solvers/coupled/interface.py
```diff
@@ -0,0 +1,148 @@
...
+    - List a ``CouplingInterface.Hook`` value in
+      :attr:`coupling_unsupported` to declare that no fallback can produce a
+      meaningful result for this solver. The wrapper raises
+      :class:`NotImplementedError` rather than silently using a fallback.
+
+    The nested ``Hook``, ``InputStateFlags``, and ``EndpointKind`` enums keep
+    the public coupling namespace compact.
+    """
+
+    class Hook(IntFlag):
```
mmacklin: Could maybe have a more descriptive name for this class, feels like it's more like a `CouplingFlags` or something similar? For bitmasks I think we typically use the `Flags` suffix, e.g.: `ShapeFlags`.

newton/examples/multiphysics/example_admm_contact_solver.py
```diff
@@ -0,0 +1,443 @@
...
+                    name="tray",
+                    solver=lambda v: SolverSemiImplicit(
+                        model=v,
+                        **{"enable_tri_contact": False, "joint_attach_ke": 2.5e4, "joint_attach_kd": 4.0e2},
+                    ),
+                    bodies=[self.tray_body],
+                    joints=[self.tray_joint],
+                ),
+            ],
+            coupling=SolverAdmmCoupled.Config(
```
mmacklin: Is there a reason for this to be a separate `Config` class instead of just constructor parameters?

mmacklin:
Amazing work @gdaviet !

docs/api/newton_solvers.rst
gdaviet: So would use suggest `newton.solvers.experimental.coupled`, or just  `newton.solvers.experimental` ? 
Regarding the eventual stable namespace I would imagine `newton.solvers.coupled` so that the utilities like `ModelView`, `CoupledInterface` do not end up directly on `newton.solvers`

(switched to `newton.solvers.experimental.coupled` in  https://github.com/newton-physics/newton/pull/2848/commits/38cadfdac3ef6f1bec3cb27b693053cc84172327)

docs/api/newton_solvers_experimental_coupled.rst
gdaviet: There are a few reasons why I moved this away from `SolverBase`, though I don't have a very strong:
 - signaling that the solver adheres to the coupling framework contract. Coupling hooks are opt-out rather than opt-in for brevity, so without an explicit support signal one could get wrong behavior silently
 - make the minimum amount of changes to SolverBase while we're still in the experimental phase
 - act as a namespace for some subtypes like 'CouplingInterface.Hook'

docs/concepts/coupling.rst
gdaviet: ADMM is definitely less invasive. However, there are also downsides:
  - It requires multiple coupling iterations per substep, in particular for transient constraints. The proxy scheme should generally behave better with a single iteration
  - The coupling constraints need to be implemented on the ADMM solver, so it would need to support every joint type, contact attributes, etc -- currently there's only a minimum set that is implemented. Though we could probably share the implementation with other solvers

Having a more implicit update for the \lambda could possibly reduce iteration count, would be interesting to think about

newton/_src/solvers/coupled/__init__.py
gdaviet: Agree!

(done in  https://github.com/newton-physics/newton/pull/2848/commits/38cadfdac3ef6f1bec3cb27b693053cc84172327)

newton/_src/solvers/coupled/interface.py
gdaviet: Thinking about it those should probably be Enum rather than Flag. They're not really meant to be combined, mostly to serve as dict keys

(switched to Enum in https://github.com/newton-physics/newton/pull/2848/commits/38cadfdac3ef6f1bec3cb27b693053cc84172327))

newton/examples/multiphysics/example_admm_contact_solver.py
gdaviet: No really --  there are precedents for both styles in the code base, but I don;t have a strong opinion here either

gdaviet:
> Thanks for pushing this through. This is a substantial piece of work, and the overall direction looks strong: keeping the API under an experimental namespace, making ownership explicit through solver entries, and adding both proxy and ADMM paths gives us a useful foundation without pretending the numerical/API surface is fully settled yet.

Thanks @adenzler-nvidia ! 1 to 3 should be addressed now. For 4, this is mostly intentional -- avoid marketing the feature too much and reduce CI burden while it is still experimental. I can add a few more small ones to test_examples though

newton/_src/solvers/coupled/solver_coupled_proxy.py
```diff
@@ -0,0 +1,961 @@
...
+            src_ids = [int(i) for i in proxy.bodies]
+            if not src_ids:
+                continue
+            proxy_local_ids = [int(i) for i in (proxy.proxy_bodies if proxy.proxy_bodies is not None else proxy.bodies)]
+            if len(src_ids) != len(proxy_local_ids):
+                raise ValueError("Proxy source bodies and proxy_bodies must have the same length")
+            self._validate_proxy_ids("Proxy source body", src_ids, model.body_count)
+            self._validate_proxy_ids("Proxy destination body", proxy_local_ids, model.body_count)
+            self._validate_unique_proxy_ids("source body", src_ids)
+            self._validate_unique_proxy_ids("proxy body", proxy_local_ids)
```
Kenny-Vilella: I wonder whether we also need to check that the source and destination bodies lives in the same world.
It could create problem in multi world setup.

gdaviet: Addressed in https://github.com/newton-physics/newton/pull/2848/commits/13d73a10c02214b8df7e06d804a2037f2b19c3f3

newton/_src/solvers/coupled/solver_coupled.py
```diff
@@ -0,0 +1,1050 @@
...
+        view.shape_contact_pairs = wp.array(filtered, dtype=wp.vec2i, device=self.model.device)
+        view.shape_contact_pair_count = len(keep)
+
+    @staticmethod
+    def _prefix_length(indices: set[int]) -> int | None:
+        """Return the dense prefix length represented by *indices*, or None."""
+        if not indices:
+            return 0
+        prefix_len = max(indices) + 1
+        return prefix_len if indices == set(range(prefix_len)) else None
```
Kenny-Vilella: There is actually an hidden limitation here.
When we couple two solvers that each own rigid bodies, for example cable + mjwarp, we do something like that:
```
  builder.add_solver_a_bodies()   # bodies [0..N)
  builder.add_solver_b_bodies()   # bodies [N..M)
  
  SolverCoupledProxy(model, entries=[
      Entry(name="A", solver=..., bodies=range(0, N)),
      Entry(name="B", solver=..., bodies=range(N, M)),
  ])
```

For the B solver we will not have a range starting by 0 meaning that it will take the full model.
This  may have a large perf impact depending on the solver.

Supporting any range would be great, but not sure how much work is required and what are the implication.

For clarity, I don't ask for any modification, mainly want to flag this limitation and see if I miss anything.

Kenny-Vilella: Another thing about this mechanism is that it doesn't work at all for multi worlds environment right?
Or did I miss something?

gdaviet: The prefix thing is just an optimization, in the general case where this is not applicable it will keep  all bodies/joints in the views but disable them. Ideally we should remove them altogether but there are a few subtle thins around handling of dependent arrays, this is marked as future work for now: 


Known follow-up:

    ModelView currently masks and prefix-compacts where possible; it does not
    perform full body/joint compaction. A future cleanup could build truly
    compact sub-model views with remapped body, joint, shape, state, world, and
    custom-attribute indices.

gdaviet: Yes, the fact the API use global body ids is by design (so we can still map between bodies of the two solvers), but under the hood `SolverCoupled` is responsible for adjusting the model views so that each solver sees/simulates only what it needs.

Kenny-Vilella: >  but under the hood SolverCoupled is responsible for adjusting the model views so that each solver sees/simulates only what it needs.

Humm so maybe I am missing something.
I am curious, do you have an example somewhere with Kamino (or another solver) as the source solver, mjwarp as the destination solver and in which mjwarp does not see the full model? In particular does not see the DoF of the first solver.

gdaviet: In the current implementation the other solver's bodies are still visible by mujoco but disabled -- actually removing the bodies from the views is what's marked as future work. But that should be transparent to the caller

Kenny-Vilella: My concern is that this has a important perf implication for mjwarp, especially that, as you mentioned, there are a few subtle things that make the resolution complicated.

It would be particularly impactful for multi env setup.
Any coupling with mjwarp where the other solver adds a significant amount of "shadow" DoFs (let's say > 30) would tank perf of mjwarp step (my quick test on a simple example shows a 3 time slowdown).
From my experience, typical run with 8k envs will be impossible if the other solver bring more than 100 "shadow "DoFs.

But yes this concern should not block the merge of this PR.
This is great work that is enabling a lot of use cases.

gdaviet: Yes, I fully agree with that, the current implementation will not scale to large RL workloads, and that definitively should be solved before moving out of the experimental status

docs/api/newton_solvers_experimental_coupled.rst
mmacklin: Resolving for this PR but would still consider making this a standard part of the `SolverBase` to simplify the number of types the user has to understand/discover.

newton/_src/solvers/coupled/interface.py
mmacklin: My main point was more that "Hook" maybe a little of a generic term to use - could we use something more specific like "CouplingType", "CouplingMode", "CouplingScheme" or similar? I think Type and Mode are typically what we use elsewhere for enums.

newton/_src/solvers/coupled/solver_coupled_admm.py
```diff
@@ -0,0 +1,3041 @@
...
+
+    owner_a: str
+    owner_b: str
+    particles_a: tuple[int, ...] | None = None
+    particles_b: tuple[int, ...] | None = None
+    contact_distance: float | None = None
+    detection_margin: float | None = None
+
+
+class SolverCoupledAdmm(SolverCoupled):
```
eric-heiden: I think we should probably spell `SolverCoupledAdmm` as  `SolverCoupledADMM`? Existing Newton public solver and config names preserve acronym casing, such as `SolverXPBD`, `SolverVBD`, `SolverImplicitMPM`, and `PADMMSolverConfig`.

CHANGELOG.md
```diff
@@ -99,6 +104,11 @@
 - Add per-mesh `color` override to `ViewerBase.log_mesh()` for tinting individual meshes without authoring per-vertex colors
 - Add per-mesh `roughness` and `metallic` PBR overrides to `ViewerBase.log_mesh()`
 - Support negative (mirrored) scale on mesh, convex hull, and SDF shapes, so a single `Mesh` instance can be shared across shapes with different signed scales without re-baking
+- Add an experimental coupled solver framework:
+  - Introduce `newton.solvers.coupled_experimental` with `SolverCoupled`, `SolverProxyCoupled`, `SolverAdmmCoupled`, and `ModelView` for multi-solver ownership, state mapping, and view-local model overrides.
```
eric-heiden: 🔴 This duplicate changelog entry still points users at symbols that do not exist. The earlier `[Unreleased]` entry already documents the coupled solver under `newton.solvers.experimental.coupled`, but this entry says `newton.solvers.coupled_experimental` and uses `SolverProxyCoupled` / `SolverAdmmCoupled`. Please remove this duplicate entry so the changelog does not advertise the wrong module path or class names.

newton/_src/solvers/coupled/solver_coupled_proxy.py
```diff
@@ -0,0 +1,985 @@
...
+    ) -> list[_ProxyBodyMapping]:
+        mappings = []
+        device = model.device
+        for proxy in coupling.proxies:
+            src_ids = [int(i) for i in proxy.bodies]
+            if not src_ids:
+                continue
+            proxy_local_ids = [int(i) for i in (proxy.proxy_bodies if proxy.proxy_bodies is not None else proxy.bodies)]
+            if len(src_ids) != len(proxy_local_ids):
+                raise ValueError("Proxy source bodies and proxy_bodies must have the same length")
```
eric-heiden: Not sure if this Codex finding is actually relevant to practical use cases, feel free to ignore:
> The proxy id validation should also reject destination ids that are owned by the destination entry. Today a proxy can declare `proxy_bodies=[1]` while the destination entry owns body `1`, so the same real body is left dynamic, marked as proxy, and then gets source-scaled inertia during proxy mass setup. The particle path has the same issue. Please reject proxy body and particle ids that overlap the destination entry's owned endpoint sets, or make aliasing explicit and skip proxy flags and mass overrides for owned endpoints.

newton/_src/solvers/coupled/solver_coupled_admm.py
```diff
@@ -0,0 +1,3041 @@
...
+        if shape_flags is None:
+            return False
+        return bool(int(shape_flags[shape]) & int(flag))
+
+    def _rigid_particle_detection_margin(self) -> float:
+        """Return the soft-contact margin to apply on the shared collision pipeline.
+
+        The collision pipeline is shared across all rigid-particle ADMM pairs,
+        so it must use a single margin. We pick the max of all
+        rigid-particle ``ContactPair.detection_margin`` values (defaulting to
```
eric-heiden: 🔴 Explicit rigid-particle `detection_margin` values below `_DEFAULT_DETECTION_MARGIN` are ignored here. Since `margin` starts at the default and each value is combined with `max()`, `ContactPair(detection_margin=0.0)` still uses the 1 cm default envelope. That differs from the particle-particle path, where the default is used only when the field is `None`. Please aggregate from the actual spec values and apply `_DEFAULT_DETECTION_MARGIN` only to specs with `detection_margin is None`, or reject smaller margins with a clear error.

newton/_src/solvers/semi_implicit/kernels_contact.py
```diff
@@ -81,7 +81,7 @@ def eval_particle_contact(
 
                 f += particle_force(n, vrel, err, k_contact, k_damp, k_friction, k_mu)
 
-    particle_f[i] = f
+    particle_f[i] = particle_f[i] + f
```
eric-heiden: 🔴 This change fixes accumulation for `eval_particle_contact`, but it also changes standalone `SolverSemiImplicit` behavior because particle contacts now add to forces that may already include spring, triangle, bending, or FEM contributions. That is the right behavior, but it is a silent user-visible change. Please add a `CHANGELOG.md` `Fixed` entry documenting that particle-particle contact forces now accumulate instead of overwriting `particle_f`.

newton/_src/solvers/__init__.py
```diff
@@ -1,6 +1,13 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 The Newton Developers
 # SPDX-License-Identifier: Apache-2.0
 
+from .coupled import (
```
eric-heiden: 🟡 These imports make the experimental coupled types show up as attributes on `newton.solvers`. Although `newton/solvers.py` leaves them out of `__all__`, importing them through `newton._src.solvers` means `newton.solvers.SolverCoupledAdmm` is still discoverable by `dir()` and IDE completion, alongside the intended `newton.solvers.experimental.coupled` path. If the experimental namespace is the only public path, please avoid re-exporting these coupled symbols from `newton/_src/solvers/__init__.py` and import the coupled module directly where `newton/solvers.py` builds the experimental namespace.

newton/_src/solvers/coupled/admm_utils.py
```diff
@@ -0,0 +1,1365 @@
...
+    any warm-started contact force.
+    """
+    i = wp.tid()
+    W_i = W[i]
+    p = Jv[i]
+    denom = rho * W_i
+    if denom > 0.0:
+        p = p - lambda_k[i] / denom
+
+    u_min_i = u_min[i]
```
eric-heiden, line 833: 🟡 The inactive `u_min` convention is split between two values. This reader treats values below `-1.0e7` as inactive, while the writers store `-1.0e8`; that works today but makes the sentinel fragile if either side changes independently. Please define one constant, for example `_INACTIVE_U_MIN`, and use it for both writer assignments and this reader check, or document why the reader intentionally uses a looser threshold.

newton/_src/solvers/coupled/solver_coupled.py
```diff
@@ -0,0 +1,1050 @@
...
+        coupling: Optional algorithm configuration reserved for derived
+            couplers. The base class steps entries independently and
+            reconciles owned state.
+    """
+
+    @dataclass(frozen=True)
+    class Entry:
+        """Public configuration for one sub-solver.
+
+        Each entry names a solver factory, the global model ids owned by that
```
eric-heiden, line 249: 🟡 `Entry` is public configuration, but this docstring only describes it in prose and does not document each dataclass field. The sibling coupled config dataclasses use Google-style `Args:` blocks, so generated API docs for `Entry` will miss the semantics for `bodies`, `particles`, `joints`, `shapes`, `configure_view`, `substeps`, and `in_place`. Please add an `Args:` section or inline comments for the attributes like we have in `ModelBuilder.JointConfig` etc. with one short description per field.

newton/_src/solvers/coupled/proxy_utils.py
```diff
@@ -0,0 +1,321 @@
...
+    f = coupling_forces[global_id] + dst_body_f[local_id]
+
+    inv_m = dst_body_inv_mass[local_id]
+    r = wp.transform_get_rotation(dst_body_q[local_id])
+    inv_I = dst_body_inv_inertia[local_id]
+
+    delta_v = dt * inv_m * wp.spatial_top(f)
+    delta_w = dt * wp.quat_rotate(r, inv_I * wp.quat_rotate_inv(r, wp.spatial_bottom(f)))
+
+    # Subtract gravity (driving solver already applied it)
```
eric-heiden: 🔴 The rewind path uses the destination proxy effective mass after proxy mass scaling, not the source body's effective mass. With a non-unit `mass_scale`, subtracting the lagged source feedback through `dst_body_inv_mass` leaves a velocity residual, so a proxy with `mass_scale=0.25` or `4.0` does not rewind back to the same source-side state. Please rewind this feedback using the source effective mass, or scale the force and inverse mass consistently so non-unit proxy mass scales produce zero residual.

newton/_src/solvers/coupled/interface.py
```diff
@@ -0,0 +1,148 @@
...
+        NOTIFY_INPUT_STATE_UPDATE = auto()
+        PROXY_CONTACT_PREPARE = auto()
+
+    class EndpointKind(IntEnum):
+        """Kinds of model endpoints addressed by coupling hooks."""
+
+        BODY = 0
+        PARTICLE = 1
+
+    class InputStateFlags(IntFlag):
```
eric-heiden: Can we wrap this into `StateFlags` that I'm introducing in #2657?

newton/_src/solvers/coupled/model_view.py
```diff
@@ -0,0 +1,1011 @@
...
+
+        particle_flags = self._cow_array("particle_flags")
+        wp.launch(
+            _mark_particle_flag_kernel,
+            dim=particle_indices.shape[0],
+            inputs=[particle_indices, int(ParticleFlags.PROXY), particle_flags],
+            device=parent.device,
+        )
+
+    def deactivate_particles(self, particle_indices: wp.array[int]) -> None:
```
eric-heiden: Since the other methods start with "disable", maybe name this one also `disable_particles`?

newton/_src/solvers/coupled/model_view.py
```diff
@@ -0,0 +1,1011 @@
...
+        inv_mass = self._cow_array("particle_inv_mass")
+        mass = self._cow_array("particle_mass")
+        wp.launch(
+            _zero_particle_mass_kernel,
+            dim=particle_indices.shape[0],
+            inputs=[particle_indices, inv_mass, mass],
+            device=parent.device,
+        )
+
+    def scale_particle_mass(self, factor: float, particle_indices: wp.array[int] | None = None) -> None:
```
eric-heiden: Nit: I would put the particle_indices first before the factor, in case we want to add more arguments later.

newton/_src/solvers/coupled/interface.py
gdaviet: Yes those look similar in intent, I'll make the switch once the other PR is merged

newton/_src/solvers/coupled/proxy_utils.py
gdaviet: I don't think this one is correct, rewinding using `dst_body_inv_mass` is intended (since we want to match back the original one after integration trough `dst`)

newton/_src/solvers/coupled/model_view.py
gdaviet: Addressed in https://github.com/newton-physics/newton/pull/2848/commits/b8e155d32e97bb13fdad145c1b9e54a236b15a8a

newton/_src/solvers/coupled/model_view.py
gdaviet: Addressed in https://github.com/newton-physics/newton/pull/2848/commits/b8e155d32e97bb13fdad145c1b9e54a236b15a8a

newton/_src/solvers/coupled/solver_coupled.py
gdaviet: Addressed in https://github.com/newton-physics/newton/pull/2848/commits/b8e155d32e97bb13fdad145c1b9e54a236b15a8a

newton/_src/solvers/coupled/admm_utils.py
gdaviet: Addressed in https://github.com/newton-physics/newton/pull/2848/commits/b8e155d32e97bb13fdad145c1b9e54a236b15a8a

newton/_src/solvers/__init__.py
gdaviet: Addressed in https://github.com/newton-physics/newton/pull/2848/commits/b8e155d32e97bb13fdad145c1b9e54a236b15a8a

newton/_src/solvers/semi_implicit/kernels_contact.py
gdaviet: Addressed in https://github.com/newton-physics/newton/pull/2848/commits/b8e155d32e97bb13fdad145c1b9e54a236b15a8a

newton/_src/solvers/coupled/solver_coupled_admm.py
gdaviet: Addressed in https://github.com/newton-physics/newton/pull/2848/commits/b8e155d32e97bb13fdad145c1b9e54a236b15a8a

newton/_src/solvers/coupled/solver_coupled_proxy.py
gdaviet: Addressed in https://github.com/newton-physics/newton/pull/2848/commits/b8e155d32e97bb13fdad145c1b9e54a236b15a8a

CHANGELOG.md
gdaviet: Addressed in https://github.com/newton-physics/newton/pull/2848/commits/b8e155d32e97bb13fdad145c1b9e54a236b15a8a

newton/_src/solvers/coupled/solver_coupled_admm.py
gdaviet: Addressed in https://github.com/newton-physics/newton/pull/2848/commits/b8e155d32e97bb13fdad145c1b9e54a236b15a8a

newton/_src/solvers/coupled/solver_coupled.py
gdaviet: The latest revision now implements view compression (for bodies and joints; shapes are still global by default, though you can pass `preserve_shape_ids=False` to `Entry` to compress them as well)

On a franka + cable scene scaling is encouraging:


```  ┌────────┬──────────┬──────────────┬──────────────┬────────┐
  │ Worlds │ ms/frame │ env frames/s │ ms/env-frame │ Status │
  ├────────┼──────────┼──────────────┼──────────────┼────────┤
  │      1 │   59.717 │       16.746 │       59.717 │ ok     │
  │      2 │   60.916 │       32.832 │       30.458 │ ok     │
  │      4 │   61.887 │       64.634 │       15.472 │ ok     │
  │      8 │   62.028 │      128.974 │        7.754 │ ok     │
  │     16 │   63.374 │      252.468 │        3.961 │ ok     │
  │     32 │   69.151 │      462.758 │        2.161 │ ok     │
  │     64 │   74.529 │      858.726 │        1.165 │ ok     │
  │    128 │   81.093 │     1578.440 │        0.634 │ ok     │
  │    256 │   93.115 │     2749.286 │        0.364 │ ok     │
  └────────┴──────────┴──────────────┴──────────────┴────────┘
```

newton/_src/solvers/coupled/interface.py
gdaviet: I got rid of this Hook enum; now CouplingInterface provide the default implementations, and solvers may override or raise. A little more verbose, but probably cleaner

jcarius-nv:

Kenny-Vilella:
FYI, after the latest change, I could manage to run multi worlds on my experimental branch.
Everything works perfectly and smoothy.

Is there anything holding this PR to be merged?

newton/_src/solvers/coupled/interface.py
gdaviet: https://github.com/newton-physics/newton/pull/2848/commits/00528f0d2c39d3836490d60db86e3add81273cf3 switched to using `StateFlags`. That meant adding BODY_F JOINT_F PARTICLE_F to StateFlags, that should not break anything so I simply added a Changelog entry

eric-heiden:
@mmichelis can you please sign the CLA? This is needed before we can merge.

newton/_src/sim/model.py
```diff
@@ -882,6 +930,76 @@ def __init__(self, device: Devicelike | None = None):
...
+        """Return the frequency domain addressed by a builder reference declaration."""
+        if references is None:
+            return None
+        built_in = {
+            "body": Model.AttributeFrequency.BODY,
+            "shape": Model.AttributeFrequency.SHAPE,
+            "joint": Model.AttributeFrequency.JOINT,
+            "joint_dof": Model.AttributeFrequency.JOINT_DOF,
+            "joint_coord": Model.AttributeFrequency.JOINT_COORD,
+            "joint_constraint": Model.AttributeFrequency.JOINT_CONSTRAINT,
```
eric-heiden, line 1186: `ModelBuilder.CustomAttribute.references` still documents `"equality_constraint"` as a supported reference domain, but this normalization map omits it. Since `finalize()` now forwards `custom_attr.references` into `Model.add_attribute()`, existing builders using `references="equality_constraint"` fail with `ValueError: Unknown custom attribute reference frequency 'equality_constraint'`.

eric-heiden (commented):
One inline API compatibility issue in the new reference-metadata path.

newton/_src/sim/model.py
eric-heiden: @adenzler-nvidia instead of handling this equality constraint domain here, can we remove this for 1.4?
I have a PR for it: https://github.com/newton-physics/newton/pull/3296

gdaviet: I have added the equality constraint frequency handling, it does not hurt in the meantime

docs/concepts/coupling.rst
```diff
@@ -0,0 +1,386 @@
+.. SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+.. SPDX-License-Identifier: CC-BY-4.0
+
+.. currentmodule:: newton
+
+Coupled Solvers
+===============
+
+.. warning::
```
adenzler-nvidia: Please replace this warning with the standard `.. experimental::` directive. The developer guide defines that directive as the user-facing compatibility marker and says not to duplicate the status with an inline prose block. The namespace explanation is already repeated below.

newton/solvers.py
```diff
@@ -451,10 +454,20 @@ def loss_kernel(particle_q: wp.array[wp.vec3], target: wp.vec3, loss: wp.array[f
     SolverXPBD,
     style3d,
 )
+from ._src.solvers import coupled as _coupled
 
 # solver flags
 from ._src.solvers.flags import SolverNotifyFlags
 
+experimental = ModuleType(f"{__name__}.experimental")
+experimental.__doc__ = "Experimental solver namespaces."
```
adenzler-nvidia: Please include the bare `.. experimental::` directive in this module docstring so the generated `newton.solvers.experimental` landing page carries Newton's standard compatibility marker. `docs/generate_api.py` copies this docstring into the generated RST, so this is the source location to change.

newton/_src/solvers/coupled/__init__.py
```diff
@@ -0,0 +1,16 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
+
+from .interface import CouplingInterface
```
adenzler-nvidia, line 9: Please add a module docstring containing `.. experimental::` here. The generated `newton.solvers.experimental.coupled` page currently has no compatibility marker, and generation reads the package docstring; adding it here prevents regeneration from dropping a manual RST edit.

newton/_src/geometry/flags.py
```diff
@@ -13,6 +13,9 @@ class ParticleFlags(IntEnum):
     ACTIVE = 1 << 0
     """Indicates that the particle is active."""
 
+    PROXY = 1 << 1
+    """Indicates that the particle is a solver-coupling proxy."""
```
adenzler-nvidia: `ParticleFlags` is otherwise stable public API. Should `PROXY` get a scoped `.. experimental::` marker so users know this member belongs to the experimental coupled-solver contract?

newton/_src/sim/enums.py
```diff
@@ -100,8 +126,11 @@ class BodyFlags(IntEnum):
     KINEMATIC = 1 << 1
     """User-prescribed body that does not respond to applied forces."""
 
-    ALL = DYNAMIC | KINEMATIC
-    """Filter bitmask selecting both dynamic and kinematic bodies."""
+    PROXY = 1 << 2
+    """View-local proxy body marker for coupled simulations."""
```
adenzler-nvidia, line 143: `BodyFlags` is otherwise stable public API. Should its class docstring carry a scoped `.. experimental::` notice covering `PROXY` and the fact that `ALL` now includes it?

newton/_src/sim/enums.py
```diff
@@ -54,13 +54,17 @@ class ModelFlags(IntEnum):
...
 class StateFlags(IntEnum):
-    """Flags indicating which state attributes should be reset.
+    """Flags indicating which state attributes were updated or should be reset.
 
-    These flags are used with :meth:`~newton.solvers.SolverBase.reset` to control
-    which parts of the simulation state are reset, allowing the solver to
-    efficiently update only the necessary components.
+    These flags are used with :meth:`~newton.solvers.SolverBase.reset` to
+    control which parts of the simulation state are reset, and with
+    :meth:`~newton.solvers.experimental.coupled.CouplingInterface.coupling_notify_input_state_update`
```
adenzler-nvidia, line 62: Should this class docstring add a scoped `.. experimental::` notice for use with `CouplingInterface`? The reset contract can remain stable; the marker could name only the coupling-specific interpretation of these flags.

newton/_src/solvers/coupled/interface.py
```diff
@@ -0,0 +1,577 @@
...
+)
+
+if TYPE_CHECKING:
+    from ...sim import Contacts, State
+
+__all__ = ["CouplingInterface"]
+
+
+class CouplingInterface:
+    """Marker mixin for solvers that participate in coupled simulations.
```
adenzler-nvidia, line 108: Since existing solver classes inherit these public coupling hooks and Sphinx includes inherited members, should `CouplingInterface` itself carry `.. experimental::` in its class docstring as well as the module marker? That would keep the hook contract marked when linked directly.

adenzler-nvidia (commented):
Focused review of experimental compatibility markers.

CHANGELOG.md
```diff
@@ -63,6 +70,7 @@
 - Fix `ModelBuilder.collapse_fixed_joints()` crashing with `IndexError` when a `mujoco:equality_constraint` row omits optional fields (`anchor`, `relpose`) that carry defaults. (#3054)
 - Fix `ViewerGL.set_model()` resetting headless/interactive camera and wind state when switching between models that use the same up-axis. (#2658)
 - Fix bend force calculation error in Style3D solver
+- Fix `SolverSemiImplicit` particle-particle contact forces accumulating into `particle_f` instead of overwriting existing forces.
```
adenzler-nvidia: just for verification, is this drive-by included in the PR, or is this a stray changelog entry?

gdaviet: It is an actual fix in the PR --  IIRC `SolverSemiImplicit` would not get correct coupling forces without that

adenzler-nvidia: makes sense, thanks!

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -771,6 +785,269 @@ def notify_model_changed(self, flags: ModelFlags | int) -> None:
...
+        if (
+            contacts is not None
+            and contacts.soft_contact_max > 0
+            and contacts.soft_contact_count is not None
+            and contacts.soft_contact_particle is not None
+            and contacts.soft_contact_shape is not None
+            and self.body_particle_contact_penalty_k.shape[0] >= contacts.soft_contact_max
+        ):
+            body_q_for_particles = state.body_q
+            body_q_prev_for_particles = self._coupling_body_q_prev_snapshot if self.model.body_count > 0 else None
```
eric-heiden: **Bug (confirmed crash path):** this reads `self._coupling_body_q_prev_snapshot` whenever `model.body_count > 0`, but the attribute is only assigned in `_init_rigid_system` under `if not self.integrate_with_external_rigid_solver and model.body_count > 0:` (line 645) — there is no default and no `getattr` guard. With `SolverVBD(model=view, integrate_with_external_rigid_solver=True)` on a model with bodies, used as a proxy destination with soft contacts, this line raises `AttributeError` before reaching the `integrate_with_external_rigid_solver` override two lines below (whose existence shows this configuration is meant to be reachable). The other read sites (817/886) are guarded by `_coupling_has_rigid_avbd_state`; this one isn't. Suggest initializing the attribute to `None` unconditionally in `__init__`, or restructuring so the external-solver branch is evaluated first.

newton/_src/solvers/semi_implicit/solver_semi_implicit.py
```diff
@@ -161,6 +166,10 @@ def step(
                 eval_muscle_forces(model, state_in, control, body_f)
 
             # particle-particle interactions
+            if model.particle_count > 1 and model.particle_grid is not None:
+                search_radius = model.particle_max_radius * 2.0 + model.particle_cohesion
+                with wp.ScopedDevice(model.device):
+                    model.particle_grid.build(state_in.particle_q, radius=search_radius)
```
eric-heiden, line 172: **Silent behavior change for existing non-coupled users:** before this PR, nothing ever built `model.particle_grid` for `SolverSemiImplicit` (only XPBD built it), so `eval_particle_contact` early-returned on `hash_grid_point_id == -1` and was effectively inert. `ModelBuilder.finalize()` auto-creates the grid whenever `particle_count > 1 and particle_max_radius > 0`, so after this change every existing cloth/soft-body scene silently gains particle-particle contact forces plus a per-step grid build. The docstring suggests this was always the intended behavior, so activating it is defensible — but it changes trajectories and per-step cost for existing scenes and deserves an explicit CHANGELOG entry (the current `Fixed` entry only mentions the accumulate-vs-overwrite flip, and with inverted wording — see the CHANGELOG comment).

CHANGELOG.md
```diff
@@ -64,6 +71,7 @@
...
```
eric-heiden: **Inverted wording:** the shipped change is `particle_f[i] = f` → `particle_f[i] += f`, i.e. *overwriting* was the bug and *accumulation* is the fix — but this entry reads as if accumulation was the bug being removed. Under the changelog's own "Fix <buggy behavior>" convention a user would audit for the opposite regression. Suggest something like: "Fix `SolverSemiImplicit` particle-particle contact forces overwriting previously accumulated forces in `particle_f` instead of adding to them." Consider also mentioning that particle-particle contacts are now actually active every step (the solver now builds `model.particle_grid` itself), since that's the larger user-visible change.

docs/concepts/coupling.rst
```diff
@@ -0,0 +1,382 @@
...
+
+XPBD understands proxy particles and proxy bodies in particle contact kernels.
+Owned particles may collide with destination proxy particles, but proxy-proxy,
+proxy-static, and proxy-particle versus proxy-body contacts are filtered so the
+destination solve does not create feedback between two proxy endpoints or
+against immovable particles.
+
+MuJoCo provides GPU effective-mass hooks from MuJoCo Warp data so proxy virtual
+inertia and ADMM endpoint weights can use articulated mass estimates rather than
+raw body mass. MuJoCo currently marks proxy impulse harvesting as unsupported
```
eric-heiden: **Doc/code mismatch (here and the Kamino paragraph at line 358):** this says MuJoCo (and Kamino) "marks proxy impulse harvesting as unsupported", but neither solver overrides `coupling_harvest_proxy_wrenches`/`coupling_harvest_proxy_particle_forces`, and the `CouplingInterface` defaults don't raise — they run the generic momentum-harvest kernels, and `SolverCoupledProxy` calls the hooks unguarded. So both solvers *silently inherit the generic momentum fallback*, contradicting this page's own contract at line ~106 ("A solver that cannot support a hook raises :class:`NotImplementedError` from that hook instead of silently using an invalid path"). A reader coupling MuJoCo/Kamino as a proxy destination expects an explicit error and instead gets momentum-difference feedback that ignores articulation constraint impulses — and will misdiagnose the resulting forces. Either add raising overrides in the two solvers or correct this text to say they fall back to the generic momentum harvest.

newton/tests/test_examples.py
```diff
@@ -868,6 +876,42 @@ class TestMultiphysicsExamples(unittest.TestCase):
...
+add_example_test(
+    TestMultiphysicsExamples,
+    name="multiphysics.example_kamino_mujoco_admm_solver",
+    devices=["cpu"],
+    test_options={"num-frames": 30, "world-count": 4, "graph-capture": False},
+    use_viewer=True,
+)
+add_example_test(
+    TestMultiphysicsExamples,
+    name="multiphysics.example_xpbd_vbd_coupled_solver",
```
eric-heiden, line 986: **Coverage gap:** of the 12 new multiphysics examples, only 4 are registered here — 8 have no `add_example_test` entry: `example_franka_cable_ik_pick_place`, `example_mujoco_franka_vbd_cable_admm_solver`, `example_mujoco_mpm_coupled_solver`, `example_mujoco_vbd_coupled_solver`, `example_mujoco_xpbd_coupled_solver`, `example_proxy_joint_gripper`, `example_vbd_mpm_coupled_solver`, `example_xpbd_mpm_coupled_solver`. Any API change that breaks their imports or coupling configuration ships unnoticed.

newton/_src/sim/model.py
```diff
@@ -122,6 +122,52 @@ class AttributeFrequency(IntEnum):
         WORLD = 15
         """Attribute frequency follows the number of worlds (see :attr:`~newton.Model.world_count`)."""
 
+    _CORE_ATTRIBUTE_REFERENCES: ClassVar[dict[str, AttributeFrequency]] = {
```
eric-heiden: **Maintainability (silent-failure contract):** these hand-maintained shadow tables (`_CORE_ATTRIBUTE_REFERENCES`, `_ATTRIBUTE_ROW_WIDTHS`, `_ATTRIBUTES_REQUIRING_EMPTY_SENTINEL`, `_ATTRIBUTE_FREQUENCY_COUNT_ATTRS`) plus the `_infer_attribute_frequency` name-prefix heuristic are consumed only by the coupled module, and both projection paths fail *silently* when an entry is missing: `_build_attribute_projections` does `if frequency is None: continue` (attribute silently excluded from index remapping), and `ModelView._count_limited_attribute` passes the full-length parent array through unsliced when the frequency can't be resolved. A future core attribute added without a table entry — especially an index-valued one missing from `_CORE_ATTRIBUTE_REFERENCES` or a multi-value row missing from `_ATTRIBUTE_ROW_WIDTHS` — produces wrong physics in coupled runs with no error. Even for experimental code, consider declaring frequency/references/row-width at attribute-definition time (the `ModelBuilder.CustomAttribute` mechanism already has a `references` field to extend to core attributes) and failing loudly for unregistered attributes, instead of parallel string tables + prefix inference. Related: `ModelView` duck-types via unbounded `__getattr__` passthrough (no nameable contract for what a view guarantees), and `kamino/_src/core/model.py` now imports `ModelView` from the experimental package just to widen an `isinstance` check — a stable-solver → experimental-module dependency that a `Model`-like protocol (or ABC registration) at the sim layer would avoid.

newton/_src/solvers/coupled/solver_coupled_proxy.py
```diff
@@ -0,0 +1,1721 @@
...
+                    src_name=proxy.source,
+                    dst_name=proxy.destination,
+                    src_joint_ids=wp.array(src_ids, dtype=int, device=device),
+                    proxy_joint_ids_global=wp.array(proxy_local_ids, dtype=int, device=device),
+                )
+            )
+
+        return mappings
+
+    def _build_proxy_mappings(
```
eric-heiden, line 556: **Duplication cluster (will drift):** the body and particle proxy paths are ~200-line near-clones — `_build_proxy_mappings` here vs `_build_proxy_particle_mappings` (identical validation/mapping logic modulo body↔particle), `_ProxyBodyMapping` vs `_ProxyParticleMapping` (field-for-field mirrors), and the Aitken kernel pairs (`_accumulate_aitken_body/particle_stats_kernel`, `_blend_aitken_body/particle_forces_kernel`) are byte-identical except `wp.spatial_vector` vs `wp.vec3`. A fix to proxy-id validation or the Aitken update must land in two places or they silently diverge. Related duplications in the same PR: `_positive_integer` is verbatim-duplicated between `SolverCoupledProxy` and `SolverCoupledADMM` (both subclass `SolverCoupled`, where it could live once); `_interface_weight` exists as both a `wp.func` in `admm_utils.py` and a host staticmethod in `solver_coupled_admm.py`; `solver_coupled_admm.py` hand-rolls four NumPy quaternion helpers including anti-parallel handling that `quat_between_vectors_robust` in `newton/_src/math/spatial.py` already provides; and `proxy_utils.smooth_proxy_teleportation_kernel` is dead code (never imported or launched) whose math lives on only as `_update_vbd_body_input_state_kernel` in `vbd_coupling_kernels.py` — the documented shared kernel would rot while the live copy hides in VBD. Suggest a single parameterized mapping builder + generic kernels, shared validators on the base class, and deleting or actually using the proxy_utils kernel.

newton/_src/solvers/coupled/solver_coupled_admm.py
```diff
@@ -0,0 +1,4324 @@
...
+        buf = self._admm_buffers.get(entry.name)
+        if buf is not None and buf.body_proximal_mass is not None and buf.body_proximal_inertia is not None:
+            entry.view.add_body_lumped_inertia(buf.body_proximal_mass, buf.body_proximal_inertia)
+        if entry.body_dynamics_disabled_local_indices.shape[0] > 0:
+            entry.view.disable_body_dynamics(entry.body_dynamics_disabled_local_indices)
+
+    def _setup_admm(self, coupling: SolverCoupledADMM.Config) -> None:
+        for entry in self._entries.values():
+            buf = _AdmmBuffers()
+            buf.supports_dynamic_inertial_refresh = bool(entry.solver.coupling_supports_inertial_property_refresh())
```
eric-heiden, line 1026: **Error quality:** this calls `entry.solver.coupling_supports_inertial_property_refresh()` without the `_require_supports_coupling` guard used on the runtime paths, and it's the first coupling touch during construction. `SolverStyle3D` is the one shipped solver that doesn't inherit `CouplingInterface`, so a Style3D entry crashes here with a raw `AttributeError` instead of the framework's explanatory TypeError ("...inherit CouplingInterface and override the hook methods it needs"). Suggest validating entries once at registration (in `SolverCoupled.__init__`) so all couplers get the clear error. More broadly, hook capability discovery is split between two mechanisms: this explicit predicate vs. call-and-catch `NotImplementedError` for the effective-mass hooks — the broad `except NotImplementedError` around a full hook body can also misclassify an incidental `NotImplementedError` raised *inside* a working hook as "unsupported", and the populate functions re-probe on every `notify_model_changed` rather than consulting `_admm_effective_mass_unsupported`. The explicit-predicate pattern used here would be the uniform fix.

eric-heiden (commented):
Some final comments from Fable 5 follow, might be worth to check before we merge.

newton/_src/solvers/vbd/solver_vbd.py
gdaviet: > Fixed by selecting the external-rigid-solver state before accessing _coupling_body_q_prev_snapshot. The snapshot is now read only on the internal rigid-solver
  > path. I also added regression coverage for harvesting soft-contact forces with an external rigid solver.

newton/_src/solvers/semi_implicit/solver_semi_implicit.py
gdaviet: > Agreed that this is a separate user-visible behavior change. I added an explicit changelog entry stating that SolverSemiImplicit now rebuilds
  > Model.particle_grid, enabling particle-particle contacts that were previously skipped.

CHANGELOG.md
gdaviet: > Fixed. The changelog now says that particle-particle contact forces previously overwrote accumulated particle_f values instead of adding to them. Grid
  > rebuilding and force accumulation are documented as separate fixes.

docs/concepts/coupling.rst
gdaviet: > Agreed—the documentation was stale, not the implementation. I removed the MuJoCo and Kamino paragraphs claiming that proxy harvesting was unsupported. The
  > general limitations section still identifies generic momentum harvesting as a fallback and recommends solver-specific harvest hooks where appropriate.

newton/tests/test_examples.py
gdaviet: > Added CI registrations for the other seven examples listed here. I intentionally left example_franka_cable_ik_pick_place out: I ran its default configuration
  > with the default segment count, and it succeeds visually, but its final numerical target check was not reliable enough for CI. Since we are not adding it to
  > CI, I reverted the experimental tolerance change rather than weakening the example’s assertion or introducing a flaky test.

newton/_src/sim/model.py
gdaviet: > Addressed the primary metadata issue by introducing Model.AttributeSpec. Core attributes retain their normal Python definitions and have one corresponding
  > specification map containing frequency, assignment, references, row width, empty-sentinel behavior, and the experimental compaction policy. This replaces the
  > parallel shadow tables and prefix inference. Tests check core metadata coverage, late registrations, non-prefix references, and misaligned storage failures.
  >
  > I am not introducing a Model ABC/protocol in this PR. That would establish a broader public model contract primarily to support an experimental view type.
  > ModelView remains experimental, and we can revisit that abstraction if it is promoted into a stable, generally supported model-view API.

newton/_src/solvers/coupled/solver_coupled_proxy.py
gdaviet: > Addressed. Body and particle proxies now share one mapping type, parameterized mapping/remapping logic, feedback-state initialization, fixed-relaxation
  > kernels, and generic Aitken kernels. Positive-integer validation lives on SolverCoupled; the ADMM quaternion code uses quat_between_vectors_robust; and the
  > unused smooth-teleportation kernel was removed.
  >
  > _interface_weight now has one Warp implementation. Static and candidate weights are computed in a batched kernel from device-side effective-mass buffers,
  > avoiding both duplicate host math and device-to-host mass reads. Type-specific sync, rewind, and harvest kernels remain separate because their state and
  > mechanics genuinely differ.

newton/_src/solvers/coupled/solver_coupled_admm.py
gdaviet: > Fixed the concrete error-quality issue by validating every entry solver against CouplingInterface immediately after its factory constructs it. Unsupported
  > solvers now fail during coupled-solver construction with the explanatory TypeError; a regression test covers this path.
  >
  > I have kept effective-mass support detection on the existing CouplingInterface contract: generic implementations are provided by the mixin, and a solver that
  > cannot support a hook raises NotImplementedError. Adding parallel support predicates for every optional hook would expand the experimental interface and
  > duplicate capability state.

adenzler-nvidia (dismissed):
Approving — the framework is cleanly contained in the experimental package, and apart from the points below I didn't find unintended changes to existing code paths. Before this merges, could you confirm the following behavior changes for non-coupled users are intended, and add/extend changelog entries for them?

1. `SolverSemiImplicit` now builds `model.particle_grid` every substep, and `eval_particle_contact` accumulates into `particle_f` instead of overwriting it. Existing multi-particle scenes where neighbors sit closer than `2 * particle_max_radius + particle_cohesion` will pick up particle-particle contact forces that were previously inert (the grid was never built for pure semi-implicit runs), so tuned scenes can change trajectories after upgrading. The two entries under `Fixed` don't convey this — please add migration guidance (e.g. the `model.particle_grid = None` opt-out).

2. `SolverFeatherstone` shares `eval_particle_contact_forces` but didn't get the grid build, so standalone Featherstone still silently skips particle-particle contacts. Intentional, or should it get the same treatment as `SolverSemiImplicit`?

3. Inactive (`~ACTIVE`) particles are now excluded from MPM grid mass/velocity transfers (`implicit_mpm_solver_kernels.py`). This matches the pre-existing comment's stated intent, but it changes results for sims that used frozen particles as obstacles — please add a changelog entry for it.

Two registration questions:

4. Only 2 of the 12 new multiphysics examples are registered in `README.md` — is that a deliberate curation, or should the rest be added with commands and screenshots?

5. `example_franka_cable_ik_pick_place` is the only new example without a `test_examples.py` registration, so its `test_final()` never runs in CI — intentionally excluded?

gdaviet:
> Approving — the framework is cleanly contained in the experimental package, and apart from the points below I didn't find unintended changes to existing code paths. Before this merges, could you confirm the following behavior changes for non-coupled users are intended, and add/extend changelog entries for them?

CHANGELOG entries haven been added/extended. I confirm that the example registration is intended -- do not add all the coupling behaviors to the README yet (may be revisited after experimental status is removed). Franka test criterion is flaky even if looking visually ok, until we have a deterministic solver path it is probably better to skip it on CI

adenzler-nvidia (approved):

vastsoun (approved):
LGTM, no comments from my side

PR #2851, opened 2026-05-12
[DRAFT] Rigid body VBD contact determinism

jumyungc:
## Description

Simple determinism was added to vbd rigid bodies, when collision pipeline gives deterministic contact pairs.

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **Bug Fixes**
  * Rigid body–body contact accumulation is now deterministic, enabling bit-identical replay with a deterministic collision pipeline.
  * Rigid-body iteration is more robust when there are no rigid contacts.

* **Tests**
  * Added determinism tests covering per-body contact ordering, solve-order determinism, and bit-identical rigid-contact replay.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2851)


newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -757,6 +758,9 @@ def _init_rigid_system(
         self._rigid_contact_point1_world = wp.zeros(0, dtype=wp.vec3, device=self.device)
         self._rigid_contact_zero_count = wp.zeros(1, dtype=wp.int32, device=self.device)
         self._rigid_contact_zero_force = wp.zeros(0, dtype=wp.vec3, device=self.device)
+        self._rigid_contact_empty_shape = wp.empty(0, dtype=wp.int32, device=self.device)
```
AnkaChan: I dont think we should have those dummy arrays. When those arrays are not available, just pass `None` and they will be pass as `null` in the compiled kernel. Then you can use:
```
if rigid_contact_empty_shape:
    ...

```

newton/_src/solvers/vbd/rigid_vbd_kernels.py
```diff
@@ -3065,6 +2945,107 @@ def solve_rigid_body(
         ext_h_ll[2, 2] + inertial_coeff,
     )
 
+    # Accumulate body-body contact forces in deterministic sorted row order.
```
AnkaChan, line 2948: This is becoming a mega kernel that solves everything. I think it's probably still advisable to separate the collision handling and rigid solve kernel, for better structure.

Also you changed from multi thread processing to a single loop. You dont have to do that for determinism. A better way would be a tiled implementation.

Some thing like:  

```
    tid = wp.tid()
    block_idx = tid // TILE_SIZE_BODY_CONTACT
    thread_idx = tid % TILE_SIZE_BODY_CONTACT

    body_id = color_group[block_idx]
    contact_count = wp.min(body_contact_counts[body_id], body_contact_buffer_pre_alloc)
    row_offset = body_id * body_contact_buffer_pre_alloc

    f = wp.vec3(0.0)
    tau = wp.vec3(0.0)
    h_ll = wp.mat33(0.0)
    h_al = wp.mat33(0.0)
    h_aa = wp.mat33(0.0)

    # Contacts are already sorted by contact index. A fixed tile reduction over
    # a fixed sorted row gives deterministic accumulation without atomics.
    i = thread_idx
    while i < contact_count:
        contact_idx = body_contact_indices[row_offset + i]

        # Evaluate this contact's contribution for body_id.
        # Details omitted: load contact endpoints/materials, compute C_eff,
        # call evaluate_rigid_contact_from_collision(), then select the b0/b1
        # contribution matching body_id.
        f_i, tau_i, h_ll_i, h_al_i, h_aa_i = evaluate_contact_for_body(
            body_id,
            contact_idx,
            dt,
            body_q_prev,
            body_q_contact,
            # ...
        )

        f += f_i
        tau += tau_i
        h_ll += h_ll_i
        h_al += h_al_i
        h_aa += h_aa_i

        i += TILE_SIZE_BODY_CONTACT

    f = wp.tile_reduce(wp.add, wp.tile(f, preserve_type=True))[0]
    tau = wp.tile_reduce(wp.add, wp.tile(tau, preserve_type=True))[0]
    h_ll = wp.tile_reduce(wp.add, wp.tile(h_ll, preserve_type=True))[0]
    h_al = wp.tile_reduce(wp.add, wp.tile(h_al, preserve_type=True))[0]
    h_aa = wp.tile_reduce(wp.add, wp.tile(h_aa, preserve_type=True))[0]

    if thread_idx == 0:
        body_forces[body_id] = f
        body_torques[body_id] = tau
        body_hessian_ll[body_id] = h_ll
        body_hessian_al[body_id] = h_al
        body_hessian_aa[body_id] = h_aa

```

newton/tests/test_solver_vbd.py
```diff
@@ -579,6 +582,262 @@ def _rigid_contact_stick_flags_require_cone_and_small_residual(test, device):
...
 
+def _rigid_body_contact_indices_are_sorted_for_deterministic_traversal(test, device):
+    """Body-centric rigid contact traversal should not depend on atomic insertion order."""
+    body_count = 4
+    contact_capacity = 6
+    with wp.ScopedDevice(device):
+        contact_counts = wp.array([5, 4, 0, 8], dtype=wp.int32, device=device)
+        contact_indices = wp.array(
+            [
+                8,
```
AnkaChan: use #fmt off and #fmt on to fold them

AnkaChan (commented):
I think it's better to add determinism as an option globally (I can add the soft part later)

newton/_src/solvers/vbd/rigid_vbd_kernels.py
jumyungc: I agree that separating contact accumulation from the rigid solve would look cleaner structurally, but I would prefer not to do that in this determinism PR. 
I already tested a separate-kernel version, and it was noticeably slower on the current examples. 

I plan to come back to rigid-contact performance separately; if a split-kernel design is faster or comparable there, I will separate the kernels in that follow-up.

AnkaChan: When I was designing the particle solvers, @mmacklin mentioned sometimes it is advisable to choose simplicity over speed. I do notice separating kernels is easier to understand & maintain, and it makes the components decoupled.

jumyungc: No objection in principle, but it was 20% slower or more in my exmples.

AnkaChan: yeah maybe there are just more throughputs for the rigid force, torque and hessian

AnkaChan: Maybe we should ask for Miles' opinion

jumyungc: Thanks Anka. 
Clarifying my earlier comments:

The fused kernel seems much faster than simply separating the contact and rigid solve kernels.

However, with tiled contact accumulation, the separate structure can become faster when each body has a large number of body-body contacts. This shows up in cases much denser than the cable pile example.

Maybe I will run more extensive tests (with different approaches) and bring this up again.

jcarius-nv:

johnnynunez:
how is it going this?

jumyungc:
This will be tracked separately. Please don't worry about this draft PR.

PR #2856, opened 2026-05-12
Auto-scale viewer debug helpers by scene size

### LINKED ISSUE #2359: [REQ] Scaling the visualization help with the scene size
### Description

The visualization helpers in the UI, such as the joints, contact, etc., should scale with the scene size for improving the user experience.

### Motivation / Use Case

In the `contacts_rj45_plug` example, it is difficult to see the contacts and other helper because of the small scale of the scene.

Image

### Alternatives Considered

_No response_

vreutskyy:
## Description

Closes #2359.

Visualization helpers in `ViewerGL` (contact arrows, joint axes, COM markers) were drawn at hardcoded world-space sizes (`0.1 m` for arrows/axes, `0.05 m` for COM). This made them invisible on small-scale scenes like `contacts_rj45_plug` and overwhelming on scenes whose AABB is dominated by large environment shapes like `basic_heightfield`.

Changes:

- Add `Viewer.scene_scale`, computed once in `set_model()` as `median(shape_collision_radius)` over body-attached shapes (`shape_body >= 0`). Static, world-attached environment shapes (heightfields, ground planes) are excluded so the scale tracks the bodies that actually move.
- Multiply each helper's previous hardcoded size by `scene_scale * <user_multiplier>`. Defaults preserve the previous look on typical ~10 cm body scenes.
- Split the contact slider into **Contact Length** (world-space, on `RendererGL.arrow_length_scale`) and **Contact Width** (screen-space, on `RendererGL.arrow_scale` — original meaning, line width + arrowhead pixels). Add **Joint Scale** and **COM Scale** sliders.
- Override points `_arrow_scale()` / `_joint_scale()` / `_com_scale()` on `ViewerBase` default to `1.0`, so non-GL backends (USD, Rerun, Null) get the auto-scaling without needing renderer-specific knobs.

## Checklist

## Test plan

```
```

Manually compared `contacts_rj45_plug`, `basic_heightfield`, and `basic_pendulum`:

| Example       | scene_scale | Default contact length | Before |
| ------------- | ----------- | ---------------------- | ------ |
| rj45_plug     | 8 mm        | 8 mm                   | 100 mm |
| heightfield   | 300 mm      | 300 mm                 | 100 mm |
| pendulum      | 1.0 m       | 1.0 m                  | 100 mm |

The pendulum case is the lossy one — `shape_collision_radius` captures the bounding-sphere radius, so for elongated capsules it overestimates thickness. Users can drag the Contact Length slider down (slider range 0.25–5.0×). Splitting length from width keeps the arrows legible when shortened.

## New feature / API change

```python
viewer = newton.ViewerGL()
viewer.set_model(model)

# Auto-computed once at set_model:
print(viewer.scene_scale)  # ~median body size in metres

# Customize per helper at runtime:
viewer.renderer.arrow_length_scale = 0.5   # shorter contact arrows
viewer.renderer.arrow_scale = 2.0          # but wider/thicker
viewer.renderer.joint_scale = 1.5
viewer.renderer.com_scale = 0.8
```

## Summary by CodeRabbit

* **New Features**
  * Visualization helpers (contact arrows, joint axes, COM markers) now auto-scale to scene size for consistent visibility.
  * Contact visualization split into independent controls for arrow length and arrow width.
  * Added Joint Scale and COM Scale sliders for fine-tuned control.
  * Visualization UI updated so controls appear contextually when relevant options are enabled.

* **Documentation**
  * Changelog updated to note the new auto-scaling behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2856)


Kenny-Vilella:
It is now much better, overall it looks neat.
Did not go through all the examples, but I can see that the COM for the cloth_franka example does not make any sense.
It may be a problem in the example itself.

Kenny-Vilella:
The contact for brick_stacking seems also to not be located properly:
Screenshot from 2026-05-13 09-55-45

vreutskyy:
Thanks for taking a look @Kenny-Vilella! Both findings, I think, are pre-existing and separate from this PR.

**`cloth_franka` COMs.** The example header explains it: the simulation runs in **centimeter** scale internally for VBD numerical stability, and a `vis_state` converts back to meters for rendering. The `render()` method does that conversion for `particle_q`, `body_q` (translation), `shape_transform`, and `shape_scale` — but it does *not* convert `body_com` or `shape_collision_radius`. So:

- The base viewer's `_log_com` reads `model.body_com` (in cm) and transforms it by `viz_state.body_q` (in m). That mixes units and scatters COM markers at "centimeter-offsets-treated-as-meter-offsets" away from each body.
- `_estimate_scene_scale` reads `shape_collision_radius` (in cm) and gets a value ~100× larger than the real Franka, which on `main` (and on the first revision of this PR) inflated the COM marker radius to ~3 m. After moving scene_scale to a more robust metric in this PR, the COM markers are now sized correctly but their *positions* are still wrong — which exposes the underlying example bug.

Easy local fix: pre-compute `viz_body_com = body_com * 0.01` and swap `self.model.body_com` in/out around `log_state()` the same way the example already swaps `shape_transform`. I'd file a follow-up rather than fold it into this PR — happy to do that if you want.

**`brick_stacking` contacts.** I don't think the contact *positions* are wrong; the screenshot looks like a typical speculative/predictive-contact dust cloud over a stud-rich Lego baseplate. With many small features (~3 mm studs) and the default contact distance, the broad-phase legitimately generates a lot of candidate contacts, and the viewer draws all of them — including ones whose normal extends along +Z and visually overlaps. This isn't something this PR changed (only the line length is multiplied by scene_scale now; start positions are unchanged from `main`). If anything, the smaller scene_scale here makes the situation a little better than the previous fixed `0.1 m` arrows.

Let me know if you'd like me to file the cloth_franka fix as a separate issue or PR.

Kenny-Vilella:
For the cloth_franka it's fine to me, you can create a ticket and do a follow up PR.
For the brick_stacking, can you please check again?
When I run on main I did not see anything similar.

If you confirm that it is normal, then I will trust you.
But please double check by comparing with main.

vreutskyy:
What exactly looks sus to you? Because I hardly can see anything at all on main.

image

But then again, this PR only changes the length of the normals. Contact positions shouldn't change.

Kenny-Vilella (approved):
Ok I got it my bad.
My concern was about the floating contact points on the left side of the franka arm.
But you are right, there are real contacts that are generated because the margin is so large.

Kenny-Vilella:
Created a ticket for the brick_stacking example: #2878

PR #2857, opened 2026-05-12
Clamp warp-lang below 1.14

preist-nvidia:
## Description

Clamp `warp-lang` to `>=1.13.0,<1.14` for the 1.2 release and add a changelog entry for the dependency range.

Warp 1.14 removes the previously deprecated `quadrature` and `domain` arguments from `warp.fem.interpolate()`. Newton's implicit MPM grain rendering path still passes `quadrature=...`, so fresh installs could resolve Warp 1.14 and fail with:

```text
TypeError: interpolate() got an unexpected keyword argument 'quadrature'
```

This keeps 1.2 on the Warp 1.13 line until the MPM rendering code is migrated to the newer `at=` API.


camevor (approved):
Thanks, lgtm!

PR #2864, opened 2026-05-13
Expose BVH constructor in ModelBuilder

rshahid-nvidia:
## Description

Adds `ModelBuilder.BvhConfig` so callers can choose Warp BVH constructors during model finalization. The new `builder.default_bvh_cfg` controls:

- `mesh_constructor`: forwarded to `wp.Mesh(..., bvh_constructor=...)` for mesh geometry, including meshes created from heightfields.
- `gaussian_constructor`: forwarded to `wp.Bvh(..., constructor=...)` when finalizing Gaussian geometry.
- `shape_constructor`: forwarded to `Model.bvh_build_shapes(..., bvh_constructor=...)` for the model shape broadphase BVH.

All fields default to `None`, so existing behavior continues to use Warp's default constructors unless a caller opts in.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k TestModelBuilderBvhConstructor
git diff --check
```

## New feature / API change

```python
import newton

# Leave as None to preserve Warp's default constructor for that BVH.
builder.default_bvh_cfg.mesh_constructor = "cubql"
builder.default_bvh_cfg.gaussian_constructor = "sah"
builder.default_bvh_cfg.shape_constructor = "lbvh"

model = builder.finalize()
```

## Summary by CodeRabbit

* **New Features**
  * Added configurable BVH construction backends during model finalization, with separate options for mesh, Gaussian, and shape BVHs.
  * `Mesh.finalize` and `Gaussian.finalize` now support an optional BVH constructor override, and `requires_grad` is keyword-only for `Mesh`.
* **Tests**
  * Added coverage to verify BVH constructor settings are correctly forwarded through model finalization, including direct Warp BVH creation.
* **Documentation**
  * Updated `CHANGELOG.md` with the new BVH configuration option.
* **Chores**
  * Bumped the Warp dependency to a newer dev build.

### DIFF
--- a/newton/_src/geometry/types.py
+++ b/newton/_src/geometry/types.py
@@ -1433,13 +1433,19 @@ class Mesh:
         self._cached_hash = None
 
     # construct simulation ready buffers from points
-    def finalize(self, device: Devicelike = None, requires_grad: bool = False) -> wp.uint64:
+    def finalize(
+        self,
+        device: Devicelike = None,
+        requires_grad: bool = False,
+        bvh_constructor: str | None = None,
+    ) -> wp.uint64:
         """
         Construct a simulation-ready Warp Mesh object from the mesh data and return its ID.
 
         Args:
             device: Device on which to allocate mesh buffers.
             requires_grad: If True, mesh points and velocities are allocated with gradient tracking.
+            bvh_constructor: Optional Warp mesh BVH constructor backend. If ``None``, Warp's default is used.
 
         Returns:
             The ID of the simulation-ready Warp Mesh.
@@ -1449,7 +1455,7 @@ class Mesh:
             vel = wp.zeros_like(pos)
             indices = wp.array(self.indices, dtype=wp.int32)
 
-            self.mesh = wp.Mesh(points=pos, velocities=vel, indices=indices)
+            self.mesh = wp.Mesh(points=pos, velocities=vel, indices=indices, bvh_constructor=bvh_constructor)
             return self.mesh.id
 
     def compute_convex_hull(self, replace: bool = False) -> "Mesh":
@@ -2379,11 +2385,12 @@ class Gaussian:
 
     # ---- Finalize (GPU upload) -----------------------------------------------
 
-    def finalize(self, device: Devicelike = None) -> Data:
+    def finalize(self, device: Devicelike = None, bvh_constructor: str | None = None) -> Data:
         """Upload Gaussian data to the GPU as Warp arrays.
 
         Args:
             device: Device on which to allocate buffers.
+            bvh_constructor: Optional Warp BVH constructor backend. If ``None``, Warp's default is used.
 
         Returns:
             Gaussian.Data struct containing the Warp arrays.
@@ -2412,7 +2419,7 @@ class Gaussian:
                 inputs=[self.warp_data, lowers, uppers],
             )
 
-            self.warp_bvh = wp.Bvh(lowers, uppers)
+            self.warp_bvh = wp.Bvh(lowers, uppers, constructor=bvh_constructor)
             self.warp_data.bvh_id = self.warp_bvh.id
         return self.warp_data
 
--- a/newton/_src/geometry/types.py
+++ b/newton/_src/geometry/types.py
@@ -12,6 +12,7 @@ import numpy as np
 import warp as wp
 
 from ..core.types import Axis, Devicelike, Vec2, Vec3, override
+from ..utils.deprecation import deprecate_nonkeyword_arguments
 from ..utils.texture import compute_texture_hash
 
 if TYPE_CHECKING:
@@ -1433,9 +1434,11 @@ class Mesh:
         self._cached_hash = None
 
     # construct simulation ready buffers from points
+    @deprecate_nonkeyword_arguments
     def finalize(
         self,
         device: Devicelike = None,
+        *,
         requires_grad: bool = False,
         bvh_constructor: str | None = None,
     ) -> wp.uint64:
@@ -2385,7 +2388,7 @@ class Gaussian:
 
     # ---- Finalize (GPU upload) -----------------------------------------------
 
-    def finalize(self, device: Devicelike = None, bvh_constructor: str | None = None) -> Data:
+    def finalize(self, device: Devicelike = None, *, bvh_constructor: str | None = None) -> Data:
         """Upload Gaussian data to the GPU as Warp arrays.
 
         Args:
--- a/newton/_src/sim/builder.py
+++ b/newton/_src/sim/builder.py
@@ -257,6 +257,19 @@ class ModelBuilder:
         delay_args: list[dict[str, Any]]  # Per-actuator delay params (empty if no delay)
         clamping_args: list[list[dict[str, Any]]]  # Per-actuator per-clamping array params
 
+    @dataclass
+    class BvhConfig:
+        """Default BVH construction settings used during model finalization."""
+
+        mesh_constructor: str | None = None
+        """Warp mesh BVH constructor backend. If ``None``, Warp's default is used."""
+
+        gaussian_constructor: str | None = None
+        """Warp Gaussian BVH constructor backend. If ``None``, Warp's default is used."""
+
+        shape_constructor: str | None = None
+        """Warp model shape BVH constructor backend. If ``None``, Warp's default is used."""
+
     @dataclass(kw_only=True)
     class ShapeConfig:
         """
@@ -867,6 +880,9 @@ class ModelBuilder:
         """Number of worlds accumulated for :attr:`Model.world_count`."""
 
         # region defaults
+        self.default_bvh_cfg = ModelBuilder.BvhConfig()
+        """Default BVH construction configuration used during model finalization."""
+
         self.default_shape_cfg = ModelBuilder.ShapeConfig()
         """Default shape configuration used when shape-creation methods are called with ``cfg=None``.
         Update this object before adding shapes to set default contact/material properties."""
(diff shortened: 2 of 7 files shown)


newton/_src/sim/builder.py
```diff
@@ -227,6 +227,13 @@ class ActuatorEntry:
         delay_args: list[dict[str, Any]]  # Per-actuator delay params (empty if no delay)
         clamping_args: list[list[dict[str, Any]]]  # Per-actuator per-clamping array params
 
+    @dataclass
+    class BvhConfig:
```
mmacklin, line 261: Not sure this needs to be a whole class?

I think we can just have a string for the constructor type.

rshahid-nvidia: I created a `BvhConfig` class because we might want to use this for specifying the bvh constructor of Guassians in the future, for which I believe the cubql backend is not yet supported. But, we don't need it right now.

daniela-hase: We could actually add the field for the Gaussian BVH already and just have it default to `None` as that is what the current default for it is.

This way we don't need to update anything further in the future once cuBQL support lands - aside from possibly changing the default value. And any downstream user could update their default regardless of Newton ahead of time if wanted :)

AnkaChan:
Agree that we should have a class `BvhConfig` for the parameters, let's keep the newton's philosophy of being simple and flat. It only add one to many builder's argument anyway.

rshahid-nvidia:
Seems like updating the default mesh bvh constructor to "cubql" in `ModelBuilder` is not safe. Some of the unit tests seem to be hitting unsupported mesh query paths for cuBQL meshes. A safer change would probably be to create render-owned meshes inside RenderContext with "cubql" as the bvh constructor.

newton/_src/sensors/warp_raytrace/render_context.py
```diff
@@ -480,6 +487,45 @@ def __load_texture_and_mesh_data(self, model: Model, load_textures: bool):
         self.mesh_data = wp.array(self.__mesh_data, dtype=MeshData, device=self.device)
         self.shape_mesh_data_ids = wp.array(mesh_data_ids, dtype=wp.int32, device=self.device)
 
+    def __create_render_shape_source_ptr(self, model: Model) -> wp.array[wp.uint64]:
+        """Return source pointers for rendering, creating renderer-owned mesh handles only when needed."""
+        constructor = self.config.mesh_bvh_constructor
+        if constructor == getattr(model, "mesh_bvh_constructor", None):
+            return model.shape_source_ptr
+
+        self.__validate_mesh_bvh_constructor(constructor)
```
daniela-hase: I think this is uneccessary, because the `wp.Mesh` call does the same check anyway.
We'd also always have to update the list of valid constructors here whenever Warp adds a new one .. so I'd remove this here.

newton/_src/sensors/warp_raytrace/render_context.py
```diff
@@ -480,6 +487,45 @@ def __load_texture_and_mesh_data(self, model: Model, load_textures: bool):
...
+                if not isinstance(shape, Mesh):
+                    continue
+
+                if shape_source_ptr is None:
+                    shape_source_ptr = model.shape_source_ptr.numpy().astype(np.uint64, copy=True)
+
+                mesh_hash = hash(shape)
+                render_mesh = self.__render_meshes.get(mesh_hash)
+                if render_mesh is None:
+                    points = wp.array(shape.vertices, dtype=wp.vec3, device=self.device)
```
daniela-hase: We don't need to create new arrays for the vertices/indices here. We can reuse the ones from the original Meshes from the model :)

newton/_src/sim/builder.py
```diff
@@ -9885,8 +9904,10 @@ def finalize(
 
             m.shape_type = wp.array(self.shape_type, dtype=wp.int32)
             m.shape_source_ptr = wp.array(geo_sources, dtype=wp.uint64)
+            m.mesh_bvh_constructor = self.default_bvh_cfg.mesh_constructor
```
daniela-hase: This and the below `gaussian_bvh_constructor` could also use the `BvhConfig`, couldn't it?

daniela-hase (commented):
I am uncertain if I like the idea of the renderer creating it's own BVH per mesh or not .. I can see the benefit of it, especially currently when cuBQL does have missing features in it's warp implementation.

cuBQL is a big performance gain for the renderer, so it makes sense to do this. But at the same time it feels kinda bad to do .. 🤔 

@mmacklin what's your opinion on this? Yay or Ney?

daniela-hase (commented):
I just updated Warp to support cuBQL in the missing mesh queries (https://github.com/NVIDIA/warp/issues/1467). So once this lands we should be fine to use `bvh_constructor="cubql"` for all `wp.Mesh` in Newton and there will be no need to create separate ones just for the renderer anymore :)

adenzler-nvidia:
moving to 1.4 because of warp 1.15 dependency

jcarius-nv:

daniela-hase:
The cuBQL stuff has been merged into Warp today and will be part of the 1.15 release! :D

It'd be great if we could get this stuff here merged as well. But we can get rid of the whole `render_meshes` and just give an option to choose what constructor to use for the Newton Meshes. No changes to `warp_raytrace` should be needed.

rshahid-nvidia:
> The cuBQL stuff has been merged into Warp today and will be part of the 1.15 release! :D
> 
> It'd be great if we could get this stuff here merged as well. But we can get rid of the whole `render_meshes` and just give an option to choose what constructor to use for the Newton Meshes. No changes to `warp_raytrace` should be needed.

Nice! I will clean up the branch and update the PR.

daniela-hase:
> Nice! I will clean up the branch and update the PR.

This all looks good to me now, thanks a lot! :D

daniela-hase (dismissed):

adenzler-nvidia:
**Heads-up: `cubql` on the Bvh-bound fields needs a Warp nightly bump**

`mesh_constructor` is fine — it routes to `wp.Mesh(bvh_constructor=...)`, where `cubql` has been available since Warp 1.13.0.

But `gaussian_constructor` and `shape_constructor` route to `wp.Bvh(constructor=...)`, and `cubql` support there only landed upstream in commit `57f7a778` ("Added cuBQL as a BVH constructor", GH-1467) on 2026-06-24. The current floor (`warp-lang>=1.14.0`) and even the locked nightly (`1.15.0.dev20260612`) predate it, so a user setting either of those to `"cubql"` hits `ValueError: CUBQL constructor is not available for wp.Bvh` deep inside `finalize()`. Verified:

```
warp 1.15.0.dev20260612 (current lock): wp.Bvh(constructor="cubql") -> FAIL
warp 1.15.0.dev20260624:                wp.Bvh(constructor="cubql") -> FAIL
warp 1.15.0.dev20260625:                wp.Bvh(constructor="cubql") -> OK   (first nightly with GH-1467)
```

To make all three fields usable at the declared minimum, please bump Warp in all three places that pin it:

1. **`pyproject.toml`** — `warp-lang>=1.14.0` → `warp-lang>=1.15.0.dev20260626` (first build with the feature is `dev20260625`; pinning today's `dev20260626` matches how #2840/#2528 pinned a recent dated nightly).
2. **`uv.lock`** — re-lock; it's currently at `1.15.0.dev20260612`, which lacks the feature, so CI would otherwise stay green on a build where `wp.Bvh` cubql doesn't exist.
3. **`asv.conf.json`** — line 19 pins `warp-lang==1.15.0.dev20260612`; bump it to match so the benchmarks run against the same Warp.

daniela-hase:
@adenzler-nvidia this PR only gives the user the option to choose the BVH constructor. It has nothing to do with cubql :)

I don't think we need to bump the warp version for this PR.

adenzler-nvidia:
Makes sense — you're best placed to judge whether anyone's actually passing `cubql` to the gaussian/shape BVHs right now, and I agree that's marginal for this PR on its own.

The reason I'd still lean toward bumping here: we're moving to Warp 1.15 for the next Newton release anyway, so pulling the nightly forward (`pyproject.toml`, `uv.lock`, `asv.conf.json`) in this PR is basically free and gets the newer Warp exercised in CI early — with `cubql` becoming valid across all three fields as a side benefit. If you'd rather keep this PR tightly scoped, equally fine to land it as-is and let the floor move in a separate bump.

adenzler-nvidia:
CI note: the `check-licenses` failure here is just because this branch predates #3245 ("Allow release-assets host in license check egress"). Merging latest `main` will pull in that workflow fix and unblock the check.

daniela-hase:
> The reason I'd still lean toward bumping here: we're moving to Warp 1.15 for the next Newton release anyway, so pulling the nightly forward (`pyproject.toml`, `uv.lock`, `asv.conf.json`) in this PR is basically free and gets the newer Warp exercised in CI early — with `cubql` becoming valid across all three fields as a side benefit. If you'd rather keep this PR tightly scoped, equally fine to land it as-is and let the floor move in a separate bump.

Oh I see, yeah I don't mind including it in this PR - as long as it doesn't derail this PR 🙃

@rshahid-nvidia Do you mind making that change? :D

rshahid-nvidia:
> > The reason I'd still lean toward bumping here: we're moving to Warp 1.15 for the next Newton release anyway, so pulling the nightly forward (`pyproject.toml`, `uv.lock`, `asv.conf.json`) in this PR is basically free and gets the newer Warp exercised in CI early — with `cubql` becoming valid across all three fields as a side benefit. If you'd rather keep this PR tightly scoped, equally fine to land it as-is and let the floor move in a separate bump.
> 
> Oh I see, yeah I don't mind including it in this PR - as long as it doesn't derail this PR 🙃
> 
> @rshahid-nvidia Do you mind making that change? :D

Yeah, I’ll update this PR later today.

newton/_src/geometry/types.py
```diff
@@ -1433,13 +1433,19 @@ def metallic(self, value: float | None):
         self._cached_hash = None
 
     # construct simulation ready buffers from points
-    def finalize(self, device: Devicelike = None, requires_grad: bool = False) -> wp.uint64:
+    def finalize(
+        self,
+        device: Devicelike = None,
```
eric-heiden, line 1440: Actually, while you're at it, can you enforce the arguments to be kwarg-only after the device argument?
```py
from ..utils.deprecation import deprecate_nonkeyword_arguments

@deprecate_nonkeyword_arguments
def finalize(
        self,
        device: Devicelike = None,
        *,
        requires_grad: bool = False,
        bvh_constructor: str | None = None,
    ) -> wp.uint64:
```

newton/_src/geometry/types.py
```diff
@@ -2379,11 +2385,12 @@ def _find_sh_degree(self) -> int:
 
     # ---- Finalize (GPU upload) -----------------------------------------------
 
-    def finalize(self, device: Devicelike = None) -> Data:
+    def finalize(self, device: Devicelike = None, bvh_constructor: str | None = None) -> Data:
```
eric-heiden: Same here, would be good to force `bvh_constructor` to be a keyword argument.

```py
def finalize(self, device: Devicelike = None, *, bvh_constructor: str | None = None) -> Data:
```

eric-heiden (dismissed):
LGTM! Just small nits for improving API consistency.

daniela-hase (approved):

adenzler-nvidia (approved):
Thank you!

PR #2870, opened 2026-05-13
Update MuJoCo lock pins to 3.8.1

adenzler-nvidia:
## Description

Update the locked MuJoCo packages to the latest 3.8 patch release and align ASV's direct benchmark install pins:
- `mujoco==3.8.1`
- `mujoco-warp==3.8.1`

`pyproject.toml` remains unchanged because the existing `~=3.8.0` policy already allows 3.8.1 and blocks 3.9+.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_mujoco_version_check
uv lock --locked
jq empty asv.conf.json
uvx pre-commit run -a
```

## Summary by CodeRabbit

* **Chores**
  * Updated benchmark environment configuration with newer MuJoCo packages: core MuJoCo to version 3.8.1 and mujoco-warp to simplified 3.8.1 versioning.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2870)

### DIFF
--- a/asv.conf.json
+++ b/asv.conf.json
@@ -17,8 +17,8 @@
   "install_command": [
     "python -m pip install -U numpy",
     "python -m pip install -U --pre warp-lang==1.14.0.dev20260511 --index-url=https://pypi.nvidia.com/",
-    "python -m pip install -U mujoco==3.8.0",
-    "python -m pip install -U mujoco-warp==3.8.0.3",
+    "python -m pip install -U mujoco==3.8.1",
+    "python -m pip install -U mujoco-warp==3.8.1",
     "python -m pip install -U torch==2.10.0+cu130 --index-url https://download.pytorch.org/whl/cu130",
     "in-dir={env_dir} python -m pip install {wheel_file}[dev]"
   ]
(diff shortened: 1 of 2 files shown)


adenzler-nvidia:
Note on the non-MuJoCo `uv.lock` diff: the `<3.14` -> `<3.13` marker changes are Open3D transitive dependency normalization. `open3d` is already declared only for Python `<3.13`, but its package block in the existing lock still had stale child markers up to `<3.14`; a clean `uv lock -P mujoco -P mujoco-warp` reproduces this rewrite.

I also checked reverse deps: the affected packages are Open3D-only in the lock, except `nbformat`, which remains reachable through docs/notebook extras. One separate follow-up is that some lazy plotting paths import `matplotlib`; on Python 3.13 they will no longer get it indirectly via Open3D.

We are not sure why this was not properly updated in `uv.lock` before, but the current diff is from an autogenerated `uv.lock` and should be correct.

Kenny-Vilella (approved):
One question about the pyproject file, we are not updating the MjWarp version there because there is no change important enough for us to force the user to update?

adenzler-nvidia:
Thanks. I think we should not raise the version floor if there were no breaking changes, and the ~= pin should be good. New pip installs will resolve to 3.8.1 anyway like this.

PR #2871, opened 2026-05-13
Fix SolverMuJoCo FREE joint parent-frame contract

### LINKED ISSUE #2704: [BUG] Gravity applied to a free body produces a state.joint_qd that is in the wrong frame.
### Bug Description

The docstring for State.joint_qd is as follows:
"For FREE and DISTANCE joints, the linear entries are child-COM velocity in the joint parent frame and the angular entries are angular velocity in that same frame."

The following scenario tests whether the spec is implemented as expected:
1) single free body with m=1 and inertia = (1,1,1)
2) Free joint with non-identity parent frame (vec3(0,0,0), from_axis_angle(unit_x, Pi/2))
3) gravity = (0, 0, -10) [any non-zero value along Z will reveal the issue]
4) Run for 1 sim step dt with SolverMuJoCo and inspect state_out.joint_qd
5) Observe that state.joint_qd = (0, 0, gravity*dt, 0, 0, 0) 

State.joint_qd in step 5) above is in the world frame.  This is incorrect.  It  should be in the joint parent frame so should have value (0, gravity*dt, 0, 0, 0, 0)

The test code added here tests the effect of gravitational force on State.joint_qd with SolverMuJoCo. We need to further test:

1) other solvers eg Feathestone 
2) if the issue extends to forces applied through Control.joint_f
3) distance joints

### Reproduction Script

```python
class TestMuJoCoSolverFreeJointParentFrame(unittest.TestCase):
    """Verify free joint state.joint_qd / control.joint_f honor the parent-frame convention."""

    def test_free_joint_qd_is_in_parent_frame_under_gravity(self):
        """``state.joint_qd[0:3]`` for a free joint must be reported in the joint parent frame.

        Per :class:`~newton.State` (``self.joint_qd`` docstring), for FREE
        joints the linear entries of ``joint_qd`` are child-COM velocity
        in the joint parent frame. The test exposes a bridge bug where
        the simulator reports them in world frame instead, by stepping
        the simplest configuration that distinguishes the two:

        - Single unit-mass body, identity inertia, zero CoM.
        - Free joint with ``parent_xform.rotation = R(x, 90 deg)``,
          identity translation, identity ``child_xform``.
        - World gravity ``(0, 0, -10)``.
        - ``joint_q = identity``, ``joint_qd = 0``, ``control.joint_f = 0``.

        After one ``dt`` step under gravity alone, the body's CoM linear
        acceleration in world frame is ``(0, 0, -10)``; in the parent
        frame it is ``g_parent = R^T * g_world = (0, -10, 0)``. So per
        the docstring contract:

            joint_qd_after[0:3] == g_parent * dt == (0, -10*dt, 0)

        The test currently fails: observed
        ``joint_qd_after[0:3] == (0, 0, -10*dt)``, which is
        ``g_world * dt`` -- the velocity is being reported in world
        frame, contradicting the documented parent-frame convention.
        """
        identity_xform = wp.transform(wp.vec3(0.0, 0.0, 0.0), wp.quat_identity())
        parent_quat = wp.quat_from_axis_angle(wp.vec3(1.0, 0.0, 0.0), float(np.pi / 2.0))
        parent_xform = wp.transform(wp.vec3(0.0, 0.0, 0.0), parent_quat)

        builder = newton.ModelBuilder(gravity=-10.0, up_axis=newton.Axis.Z)
        body = builder.add_link(
            xform=identity_xform,
            mass=1.0,
            inertia=wp.mat33(1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0),
            com=wp.vec3(0.0, 0.0, 0.0),
        )
        j_root = builder.add_joint_free(
            parent=-1,
            child=body,
            parent_xform=parent_xform,
            child_xform=identity_xform,
        )
        builder.add_articulation([j_root], label="floating_body")

        model = builder.finalize()
        state = model.state()
        state_next = model.state()
        control = model.control()
        contacts = model.contacts()
        solver = SolverMuJoCo(model)
        dt = 1e-2

        joint_q = state.joint_q.numpy()
        joint_q[:] = (0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0)
        state.joint_q.assign(joint_q)
        # joint_qd starts at zero; control.joint_f starts at zero. Body
        # falls under world gravity alone for one step.
        solver.step(state, state_next, control, contacts, dt)

        qd_after = state_next.joint_qd.numpy()
        # Expected v_com in parent frame: R^T * g_world * dt = (0, -10, 0) * dt.
        expected = np.zeros(6, dtype=np.float64)
        expected[1] = -10.0 * dt
        np.testing.assert_allclose(qd_after, expected, atol=1e-6, rtol=1e-3)
```

### System Information

_No response_

adenzler-nvidia:
## Description

`SolverMuJoCo` returned `State.joint_q` / `State.joint_qd` for root `FREE` joints
in world frame, contradicting the parent-frame contract documented on
`State.joint_qd` (and `docs/concepts/articulations.rst`). With a non-identity
`parent_xform`, the post-step FK kernel then composed the world-frame values
with `X_wpj` and produced doubly-rotated `body_q` / `body_qd`.
`SolverFeatherstone` already honored the documented contract.

This PR makes the MuJoCo bridge produce parent-frame state and fixes a
parallel bug in `ModelBuilder.add_joint_free` that stored the world-frame
`body_q[child]` directly into `joint_q[0:7]` without rotating into the parent
joint frame. The fix is a no-op when both `parent_xform` and `child_xform`
are identity, which is the default and what every importer currently produces.

- `convert_mj_coords_to_warp_kernel`: for FREE joints, solve
  `X_j = inv(X_wpj) * body_world * X_cj` and rotate the COM/angular twist into
  the parent joint frame.
- `convert_warp_coords_to_mj_kernel`: for FREE joints, compose
  `body_world = X_wpj * X_j * inv(X_cj)` and rotate the parent-frame twist
  into world before MuJoCo's CoM→origin and world→body conversions.
- `ModelBuilder.add_joint_free`: replace the `joint_q = body_q[child]`
  initializer with `X_j = inv(X_wpj) * body_q[child] * X_cj`.

`Control.joint_f` for FREE/DISTANCE stays world-frame at the COM by design
(see `docs/concepts/conventions.rst`): both solvers route it through
`xfrc_applied` / `body_f_ext` as a world wrench, sidestepping the
generalized-force basis. Not touched.

BALL joints share the same parent-frame contract for `joint_q[0:4]` /
`joint_qd[0:3]` and the MuJoCo bridge passes them through without rotation —
this PR does not address that path (no reported issue, and MuJoCo's BALL
convention with a rotated joint anchor needs separate verification).

Closes #2704.

## Checklist

## Test plan

```bash
uv run --extra sim --extra dev -m newton.tests -k "test_root_free_joint_under_rotated_xforms_uses_parent_frame_qd"
uv run --extra sim --extra dev -m newton.tests -k "test_collapse_fixed_joints"
uv run --extra sim --extra dev -m newton.tests -k "test_body_force"
uv run --extra sim --extra dev -m newton.tests -k "test_kinematics"
uv run --extra sim --extra dev -m newton.tests -k "test_mujoco_solver"
uv run --extra sim --extra dev -m newton.tests -k "test_ik"
uv run --extra sim --extra dev -m newton.tests -k "test_jacobian_mass_matrix"
```

All green locally on Featherstone + MuJoCo (CPU + Warp) across CPU/CUDA.

## Bug fix

**Steps to reproduce:**

1. Build a model with a single body, root `FREE` joint, and non-identity
   `parent_xform` (rotation).
2. Add world gravity along `-Z`.
3. Run one `SolverMuJoCo.step()` and inspect `state.joint_qd[0:3]`.
4. Observed: `(0, 0, -g*dt)` — world frame, contradicting the docs.
   Expected: `(0, -g*dt, 0)` for a 90° rotation about X — parent frame.

**Minimal reproduction:**

```python
import numpy as np
import warp as wp
import newton

parent_xform = wp.transform(wp.vec3(), wp.quat_from_axis_angle(wp.vec3(1, 0, 0), np.pi / 2))
builder = newton.ModelBuilder(gravity=-10.0, up_axis=newton.Axis.Z)
body = builder.add_link(mass=1.0, inertia=wp.mat33(1, 0, 0, 0, 1, 0, 0, 0, 1))
joint = builder.add_joint_free(parent=-1, child=body, parent_xform=parent_xform)

model = builder.finalize()
solver = newton.solvers.SolverMuJoCo(model, use_mujoco_cpu=True, disable_contacts=True)
newton.eval_fk(model, state_0.joint_q, state_0.joint_qd, state_0)
solver.step(state_0, state_1, model.control(), model.contacts(), 1e-2)

# Before this PR: (0., 0., -0.1, 0., 0., 0.) (world frame, wrong)
# After this PR:  (0., -0.1, 0., 0., 0., 0.) (parent frame, matches Featherstone)
print(state_1.joint_qd.numpy())
```

## Release Notes

* **Bug Fixes**
  * Fixed `SolverMuJoCo` to correctly report joint velocities in the parent frame for FREE joints with rotated parent transforms, preventing potential corruption of derived body velocity calculations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2871)


newton/_src/solvers/mujoco/kernels.py
```diff
@@ -568,52 +570,51 @@ def convert_mj_coords_to_warp_kernel(
...
-        # Get angular velocity in body frame from MuJoCo and convert to world frame
+        pos_body_world = wp.vec3(qpos[worldid, q_i + 0], qpos[worldid, q_i + 1], qpos[worldid, q_i + 2])
+        body_world = wp.transform(pos_body_world, rot_body_world)
+
+        # Recover X_j = inv(X_wpj) * body_world * X_cj. Q_p is the parent joint
+        # frame rotation, used below for velocity conversion.
+        X_j = wp.transform_inverse(X_pj) * body_world * X_cj
+        q_p = wp.transform_get_rotation(X_pj)
+        for i in range(7):
+            joint_q[wq_i + i] = X_j[i]
```
gyeomannvidia: I think we have a documentation hole for State.joint_q. 

We now compute joint_q as the child joint anchor in the joint parent frame.  We used to compute joint_q as the pose of the body origin in the world frame.  Which is correct?  We have a clear spec for State.joint_qd when we have a free joint but I don't see anything for State.joint_q.

Here's the spec for State.joint_q:

Generalized joint position coordinates [m or rad, depending on joint type], shape (joint_coord_count,), dtype float.

and here's the spec for State.joint_qd:

For FREE and DISTANCE joints, the linear entries are child-COM velocity in the joint parent frame and the angular entries are angular velocity in that same frame.

Should we be conservative with this PR and sort out State.joint_q for free and distance joints in a separate PR?

newton/_src/solvers/mujoco/kernels.py
```diff
@@ -568,52 +570,51 @@ def convert_mj_coords_to_warp_kernel(
         return
 
     if type == JointType.FREE:
```
gyeomannvidia, line 571: From the spec for State.joint_qd, I believe we need to extend this if statement to include JointType.DISTANCE. 

We should test  JointType.DISTANCE as well.

Edit: MuJoCo does not have a distance joint so we don't need that here. Resolving comment.

newton/_src/solvers/mujoco/kernels.py
adenzler-nvidia: Yeah you're right, I was assuming these frames are matching but we actually don't state anything.

There's also this PR: https://github.com/newton-physics/newton/pull/2559.

Happy to de-scope to qd, but it feels like something we need to address soon.

adenzler-nvidia: Done in 6b9d68cbf — descoped to joint_qd only (FREE joint_q stays world-frame in the bridge as before). State.joint_q docstring update for FREE/DISTANCE can happen in a separate PR alongside #2559.

newton/_src/solvers/mujoco/kernels.py
adenzler-nvidia: Confirmed — DISTANCE isn't routed through this kernel since SolverMuJoCo doesn't support DISTANCE joints. Dropped the JointType.DISTANCE branch I had briefly added in 6b9d68cbf to keep the change minimal.

newton/_src/solvers/mujoco/kernels.py
```diff
@@ -586,34 +593,27 @@ def convert_mj_coords_to_warp_kernel(
...
-
-        # Angular velocity: convert from body frame (MuJoCo) to world frame (Newton)
-        joint_qd[wqd_i + 3] = w_world[0]
-        joint_qd[wqd_i + 4] = w_world[1]
-        joint_qd[wqd_i + 5] = w_world[2]
+        com_world = wp.quat_rotate(rot, body_com[child])
+        v_origin_world = wp.vec3(qvel[worldid, qd_i + 0], qvel[worldid, qd_i + 1], qvel[worldid, qd_i + 2])
+        v_com_world = v_origin_world + wp.cross(w_world, com_world)
+
+        v_com_parent = wp.quat_rotate_inv(q_p, v_com_world)
```
gyeomannvidia, line 608: Looking at this code revealed something else that is independent of this PR.

We assume here that the parent body of a free joint is always the world with identity pose.  MuJoCo requires this so the assumption here that the parent body has identity pose is correct. 

The problem is that builder.add_joint_free() allows the parent body to be an actual body with non-identity pose.

No action required for this PR due to MuJoCo's restrictions.  However, I don't believe we have managed this restriction well in SolverMuJoCo.  I added an issue for this. https://github.com/newton-physics/newton/issues/2875

gyeomannvidia (approved):

PR #2874, opened 2026-05-13
Add strict output test base

adenzler-nvidia:
## Description

Add an opt-in `NewtonTestCase` base class with a strict stdout/stderr output contract.

Tests inheriting from `NewtonTestCase` now fail on unexpected stdout or stderr unless the test explicitly declares expected or allowed output with regex helpers. The output contract covers `setUp`, the test body, `tearDown`, `addCleanup()` callbacks, and subprocess output passed through `assertSubprocessSuccess()`.

This PR also converts `TestBasicExamples` to the new base class and annotates the currently expected output for those examples, including the PXR thread-limit banner, the basic plotting diagnostic message, and the known Warp SDF compiler warning.

## Checklist

- [x] `CHANGELOG.md` has been updated (if user-facing change; N/A, test-only change)

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_unittest_utils
uv run --extra dev -m newton.tests -k test_basic.example_basic_pendulum_cpu
uv run --extra dev -m newton.tests -k TestBasicExamples
git diff --check
uvx pre-commit run ruff --files newton/tests/unittest_utils.py newton/tests/test_examples.py newton/tests/test_unittest_utils.py
uvx pre-commit run ruff-format --files newton/tests/unittest_utils.py newton/tests/test_examples.py newton/tests/test_unittest_utils.py
uvx pre-commit run typos --files newton/tests/unittest_utils.py newton/tests/test_examples.py newton/tests/test_unittest_utils.py
```

## New feature / API change

```python
from newton.tests.unittest_utils import NewtonTestCase


class TestExample(NewtonTestCase):
    def test_expected_output(self):
        self.expectOutputRegex(r"done\n", stream="stdout")
        print("done")

    def test_optional_output(self):
        self.allowOutputRegex(r"optional warning", stream="stderr")

    def test_subprocess(self):
        result = subprocess.run(..., capture_output=True, text=True, check=False)
        self.assertSubprocessSuccess(result, command=[...])
```

## Summary by CodeRabbit

* **Tests**
  * Added pattern-based assertions for example subprocess output; plotting example tolerates known font-cache/diagnostic noise.
  * Refactored example tests to use a shared helper and consistent allowed-output handling.
  * Introduced a strict test base enforcing per-test stdout/stderr contracts with expect/allow regex APIs.
  * Improved subprocess failure reporting (includes command and captured outputs) and expanded coverage across setup/teardown/cleanup, synchronized output, and runner interactions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2874?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


adenzler-nvidia:
Ready for review now. @eric-heiden @shi-eric, please take a look when you have a chance.

eric-heiden (dismissed):
Looks good to me!

shi-eric:
Confirmed the plotting fallback regex can mask unexpected trailing stdout.

Minimal repro against the current `_BASIC_PLOTTING_OUTPUT_RE`:

```bash
uv run python - <<'PY'
import re

pattern = (
    r"(?s)(?:"
    r"Diagnostics plot saved to solver_convergence\.png\n?"
    r"|"
    r"\n?Simulation diagnostics summary \(\d+ steps\):\n"
    r"  Iterations:   mean=.*\n"
    r"  Kinetic E:    final=.*\n"
    r"  Potential E:  final=.*\n"
    r"  Constraints:  mean=.*\n?"
    r")"
)

stdout = (
    "Simulation diagnostics summary (3 steps):\n"
    "  Iterations:   mean=1.0\n"
    "  Kinetic E:    final=2.0\n"
    "  Potential E:  final=3.0\n"
    "  Constraints:  mean=4.0\n"
    "UNEXPECTED EXTRA STDOUT\n"
)

unmatched = re.sub(pattern, "", stdout, flags=re.MULTILINE)
print("matched:", bool(re.search(pattern, stdout, flags=re.MULTILINE)))
print("unmatched_repr:", repr(unmatched))
print("strict_checker_would_fail:", bool(unmatched.strip()))
PY
```

Current output:

```text
matched: True
unmatched_repr: ''
strict_checker_would_fail: False
```

So the strict-output checker has no unmatched output left to reject. Replacing the per-line `.*` fields with `[^\n]*` leaves `UNEXPECTED EXTRA STDOUT\n` unmatched and makes the same check fail as intended.

shi-eric:
Non-blocking cleanup idea: the output regexes in `test_examples.py` may become hard to maintain as more dependency/runtime noise shows up. The strict output contract is useful, but the dependency-noise allowances feel like a separate concern from example registration.

One possible direction would be to move those patterns into a discoverable shared module, e.g. `newton/tests/unittest_output_noise.py`, and expose named profiles instead of raw regexes:

```python
allow_output_profiles=[
    OutputNoiseProfile.PXR,
    OutputNoiseProfile.WARP_RUNTIME,
    OutputNoiseProfile.MATPLOTLIB_STARTUP,
]
```

Then `test_examples.py` can stay focused on each example's intentional output, while PXR/Warp/matplotlib environment noise is maintained in one place. This is not a blocker for this PR, but it may keep the strict-output contract easier to extend without turning example registrations into a growing list of dependency-specific regexes.

newton/tests/unittest_utils.py
```diff
@@ -221,6 +223,186 @@ def __exit__(self, exc_type, exc_value, traceback):
                 self.test.fail(f"Unexpected output:\n'{s.rstrip()}'")
 
 
+@dataclasses.dataclass
+class _OutputRegex:
```
shi-eric, line 239: Non-blocking: could you add class docstrings for the new output-contract helper classes here? `_OutputRegex` and `_OutputCapture` are internal, but they encode the core model for required vs. allowed output and how captured streams are checked. A short docstring on each would make this easier to understand when extending the strict-output machinery later.

newton/tests/test_examples.py
```diff
@@ -27,12 +28,43 @@
...
+)
+_WARP_SDF_CONSTANT_CONVERSION_WARNING_RE = (
+    r"(?m)"
+    r"(?:^.*wp_sdf_contact_write_contact_to_reducer_[^\n]*\.cpp:\d+:\d+: warning: "
+    r"implicit conversion from 'long' to 'const wp::int32'.*\n"
+    r"^.*\n"
+    r"^.*\n"
+    r")+"
+    r"^\d+ warnings? generated\.\n?"
+)
```
jcarius-nv, line 65: some of these are very specific. Did you intentionally choose that and accept that we'll just update them if the maintainers of warp, pxr, etc change the wording ever so slightly?

jcarius-nv (approved):
Made a review pass, the PR is sound and good to go from my side.

What is the suggested pathway to converting more tests to the stricter checking? I think it'd be great to roll this out widely to make sure Newton is rather silent unless debug info is explicitly requested or there are actual errors to surface to the user.

newton/tests/test_examples.py
adenzler-nvidia: yes, that was intentional to avoid overengineering things. Ideally we fix the source of these problems anyway, the exception should not be the norm.

If the maintenance churn starts to be problematic, we can address this for sure.

adenzler-nvidia:
I was deferring an actual plan until this first set is merged. I'm not sure how mandatory I want to make this, but at least the examples should use this base class, and then we can see whether it makes sense to enforce it at some point.

for the other side of verbosity, there is also the logging change (#2918) which ties into this and will also come in incrementally.

PR #2876, opened 2026-05-13
Add support for layers in the opengl viewer

nvtw:
## Description
Allows to do overlay rendering of different solvers simulating the same scene - useful for solver comparisons and requested by Disney.
Adds a viewer "layer" system: call `Viewer.activate(layer_id)` to route subsequent `set_model` / `log_state` / `log_*` calls into a named layer (object names are auto-prefixed with `/layers/<layer_id>` so layers do not collide). Each layer keeps its own model, shape batches, contact/joint/COM caches, visibility toggles, and CUDA graph. Use `Viewer.set_layer_visible()` (or the new "Layers" group in the `ViewerGL` sidebar) to toggle layers independently.
The default (unnamed) code path is unchanged: when no user layer is activated, names remain unprefixed, so existing examples, tests, USD layers, and Rerun entity paths keep working as-is.
## Checklist
## Test plan
New unit tests in `newton/tests/test_viewer_layers.py` cover:
- default layer keeps legacy unprefixed names
- two activated layers emit names under distinct `/layers/<id>/` namespaces
- `activate()` preserves and restores per-layer state across switches
- `set_layer_visible(False)` forces `hidden=True` on all shape instance calls
- `remove_layer()` drops a layer and falls back to the default
- the internal default layer cannot be removed
- `activate("")` is rejected
```
uv run --extra dev -m newton.tests -k test_basic.example_basic_multi_solver_overlay
```
Manual verification with the new example (XPBD / Featherstone / MuJoCo Warp pendulums overlaid in one window, toggled via the sidebar):
```
python -m newton.examples basic_multi_solver_overlay
```
## New feature / API change
```python
import newton
viewer = ViewerGL()
viewer.activate("xpbd")
viewer.set_model(model_xpbd)
viewer.activate("featherstone")
viewer.set_model(model_featherstone)
while viewer.is_running():
    viewer.begin_frame(t)
    viewer.activate("xpbd")
    viewer.log_state(state_xpbd)
    viewer.activate("featherstone")
    viewer.log_state(state_featherstone)
    viewer.end_frame()
viewer.set_layer_visible("featherstone", False)
```

## Release Notes

* **New Features**
  * Multi-layer viewer overlay system enabling independent layer activation, per-layer visibility toggling, and layer-specific transforms
  * Layer-scoped object namespacing prevents naming collisions across concurrent visualization layers
  * Example demonstrating simultaneous multi-solver overlay visualization

* **Documentation**
  * Updated API documentation for the Layer class

* **Tests**
  * Comprehensive test coverage for layer system behavior


newton/_src/viewer/recording.py
```diff
@@ -0,0 +1,394 @@
...
+                stdin=subprocess.PIPE,
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+            )
+        except Exception:
+            logger.exception("Failed to launch ffmpeg for recording.")
+            self._proc = None
+            return False
+
+        self._queue = queue.Queue(maxsize=8)
```
camevor: A queue size of 8 frames seems quite small, should we bump this?

newton/_src/viewer/recording.py
```diff
@@ -0,0 +1,394 @@
...
+        if frame.dtype != np.uint8 or frame.ndim != 3 or frame.shape[2] != 3:
+            return
+        if frame.shape[0] != self._height or frame.shape[1] != self._width:
+            return
+        if not frame.flags["C_CONTIGUOUS"]:
+            frame = np.ascontiguousarray(frame)
+        try:
+            self._queue.put_nowait(frame.tobytes())
+        except queue.Full:
+            # Keep renderer real-time by dropping frames under sustained pressure.
```
camevor: Emitting a warning here would be nice.

newton/_src/viewer/recording.py
```diff
@@ -0,0 +1,394 @@
...
+                    break
+                try:
+                    self._proc.stdin.write(item)
+                except BrokenPipeError:
+                    unexpected_exit = True
+                    break
+        finally:
+            try:
+                self._proc.stdin.close()
+            except Exception:
```
camevor: Surfacing the error stream from ffmpeg (e.g. as a warning) would be useful as a diagnostic.

newton/_src/viewer/viewer.py
```diff
@@ -26,6 +26,117 @@
...
 
+#: Sentinel layer id used when no user-defined layer has been activated.
+#: Preserves the legacy behavior of unprefixed object names so that existing
+#: examples, tests, and viewer backends keep working unchanged.
+_DEFAULT_LAYER_ID = "__default__"
+
+#: Per-model attribute names that are part of "layer state". Used by
+#: :meth:`ViewerBase.activate` to snapshot/restore everything that belongs
+#: to one model when switching layers. These are exactly the attributes
+#: initialized by :meth:`ViewerBase.clear_model`.
```
camevor: This list is missing `scene_scale`.

newton/_src/viewer/viewer_gl.py
```diff
@@ -1606,6 +1724,94 @@ def _update(self):
...
+            return
+
+        record_dt = 1.0 / float(self._record_fps)
+        # Nothing to emit yet: sim has not advanced past the next slot.
+        if self.time + 0.5 * record_dt < self._record_next_sim_t:
+            return
+
+        w = self.renderer._screen_width
+        h = self.renderer._screen_height
+        if self._record_frame_gpu is None or self._record_frame_gpu.shape != (h, w, 3):
```
camevor: nit: some kind of message here would be nice

newton/_src/viewer/viewer.py
```diff
@@ -26,6 +26,117 @@
...
+    "_sdf_isomesh_instances",
+    "_sdf_isomesh_populated",
+    "_shape_sdf_index_host",
+    "_sdf_margin_mesh_cache",
+    "_sdf_margin_vdata_cache",
+    "_sdf_margin_edge_caches",
+)
+
+
+class ViewerLayer:
```
camevor: nit: `ViewerLayer` could be mistaken for an actual Viewer alongside `ViewerGL` etc., `Layer` would also work.

newton/_src/viewer/viewer.py
```diff
@@ -46,9 +157,214 @@ def __init__(self):
...
+
         # All model-dependent state is initialized by clear_model()
         self.clear_model()
 
+    # ------------------------------------------------------------------
+    # Layer management
+    # ------------------------------------------------------------------
+
+    @property
+    def layer(self) -> ViewerLayer:
```
camevor: nit: These properties have quite inconsistent naming, consider simplifying them before merging.

pyproject.toml
```diff
@@ -89,6 +89,7 @@ examples = [
     "pyyaml>=6.0.2",
     "cbor2>=5.7.0",         # for binary recording format (.bin files) - more efficient than JSON
     "Pillow>=11.3.0",
+    "imageio-ffmpeg>=0.5.1", # optional: bundled ffmpeg for ViewerGL live MP4 recording
```
camevor: Question: is the `imageio-ffmpeg` dependency introduced only for the bundled ffmpeg binary?

newton/_src/viewer/viewer_gl.py
```diff
@@ -1606,6 +1724,94 @@ def _update(self):
...
+
+    def _stop_recording(self) -> Path | None:
+        """Stop live MP4 recording and return the output path (if any)."""
+        path = self._recorder.stop()
+        self._record_frame_gpu = None
+        self._record_next_sim_t = None
+        self._record_last_frame_bytes = None
+        return path
+
+    def _record_frame_if_needed(self):
```
camevor: Question: is recording duplicate frames useful, rather than tying the recording frame rate more closely to the simulation frame rate?

newton/examples/basic/example_basic_multi_solver_overlay.py
```diff
@@ -0,0 +1,213 @@
...
+        for layer in self.layers:
+            self.viewer.activate(layer.layer_id)
+            self.viewer.set_model(layer.model)
+
+    def step(self):
+        for layer in self.layers:
+            layer.step()
+        self.sim_time += self.frame_dt
+
+    def test_final(self):
```
camevor, line 207: This example should be registered in `test_examples.py` to be included in the test suite.

camevor (commented):
Looks good to me overall, added a few minor comments.

newton/_src/viewer/viewer_gl.py
nvtw: That's a good question. I assume video compression is so good that it does not harm much. The intent was to have the video at standard FPS (30, 60 or 120) since these FPS are usually requested by marketing. We could also just use the simulation target fps but that might then be 50FPS or anything. I don't know what the best choice is.

pyproject.toml
nvtw: Yes. And access to the nvidia video encoder. Are there better packages?

camevor: My concern is that it is a relatively heavy dependency that is included with fairly low granularity, when users might already have a system-wide `ffmpeg` installation or not need the functionality at all. Maybe it would make sense to use a separate extra for this?

newton/_src/viewer/viewer_gl.py
camevor: Agreed, the compression would probably handle this fine, but the motion might be jerky.
This ties into the question whether we should be dropping frames that don't make the queue or block the simulation when the queue is full. To give this recording feature a clear advantage over a standard third-party screen recording, I think we should aim for perfect frame capture at the native simulation frame rate. It would make for a lightweight and quick tool to record in good quality without requiring external rendering enabled by `ViewerUSD`. We could also show a warning when using non-standard frame rates.

chschuma-disney: Based on our use cases, we'd probably also want to be able to capture at the native simulation rate, ideally with the option to set a frame rate manually if required. The most common use case would likely be that we run the simulation at a multiple of the video frame rate, e.g., simulate at 120 Hz and capture at 24 or 30 Hz.
For this, it'd probably be easiest if the video frame rate is a text box prefilled with the simulation rate.

adenzler-nvidia (changes requested):
I think the MP4 recording changes should be split into a separate PR.

They are orthogonal to the viewer layer work and add a lot of separate surface area: ffmpeg discovery, encoder selection, subprocess/thread lifecycle, queue/drop policy, filesystem defaults, UI controls, and recording FPS semantics. That makes this PR harder to review as a layer-system change.

I would also avoid adding `imageio-ffmpeg` here. If video recording lands later, it should initially detect a system `ffmpeg` at runtime and gracefully disable/fail when unavailable, without introducing a new dependency.

API-wise, recording should probably follow the existing viewer/export pattern, similar to `ViewerUSD` / `ViewerFile`: an explicit recording viewer or GL-backed recording wrapper with an output path and FPS, driven by the normal `set_model()` / `begin_frame()` / `log_state()` / `end_frame()` / `close()` loop. The `ViewerGL` sidebar controls could come later as UI on top of that programmatic API.

adenzler-nvidia:
I am not fully convinced that `Viewer.activate(layer_id)` is the right public abstraction here.

The underlying requirement seems to be “render several independent model/state views in one scene,” usually for solver comparison. Other visualization systems tend to model this as scene roots, display groups, topics, or render streams, not as a global mutable active layer. For example, USD/Rerun-style path namespaces or RViz-style display groups map naturally to this problem.

Could we make the public API express that more directly?

```python
xpbd = viewer.add_overlay("XPBD", model_xpbd)
featherstone = viewer.add_overlay("Featherstone", model_featherstone)

viewer.begin_frame(t)
xpbd.log_state(state_xpbd)
featherstone.log_state(state_featherstone)
viewer.end_frame()
```

or for the common same-topology solver-comparison case:

```python
viewer.set_model(model)
viewer.log_states({
    "XPBD": state_xpbd,
    "Featherstone": state_featherstone,
    "MuJoCo Warp": state_mujoco,
})
```

Internally this can still use path namespacing like `/overlays/<name>/...`, per-overlay visibility, per-overlay transforms, and per-overlay render caches. But that state would live in explicit overlay/render-context objects instead of being snapshotted/restored through `_LAYER_STATE_ATTRS` on the viewer. The current implementation feels fragile because every future model-bound viewer cache has to be added to the layer-state list or layer switching can silently regress.

@eric-heiden, curious to get your take on the public API here.

eric-heiden (commented):
Thanks, this is a nice feature @nvtw!

The API seems to be inline with libraries like imgui, which is not bad actually I think.

It would be good if this also worked with the other viewers if possible. Particularly with the USD/RTX viewers since those are often used to generate camera-ready renderings.

adenzler-nvidia:
Following up now that recording's been split out and the other-viewer support landed — thanks, the recording/dependency concern from my earlier review is resolved.

On the API: happy to drop the overlay-object suggestion. Per @eric-heiden's point the imgui-style `activate()` surface is a reasonable, well-precedented idiom and keeps the public surface small — no objection to the paradigm.

One thing I'd still treat as blocking on best-practice grounds, independent of the paradigm: per-layer state lives on the viewer and is copied in/out of `layer.state` on every `activate()` via `_snapshot_layer`/`_restore_layer` over the ~50-entry `_LAYER_STATE_ATTRS` list (plus each backend's `_extra_layer_state_attrs()` override). That's a hand-maintained parallel registry: any future model-bound cache added to the viewer or a subclass but not registered there will silently bleed across layers — no error, just wrong rendering, which is painful to track down later.

The `Layer` class already documents the right model ("the layer owns the model reference, all shape-instance batches, contact/joint/COM caches, ...") but the implementation snapshots/restores instead. Could we make the data match the docstring: have `Layer` own those fields and route the viewer's reads/writes through `self.layer.<field>`, so `activate()` just rebinds the active id with no snapshot/restore? That removes `_LAYER_STATE_ATTRS`, `_extra_layer_state_attrs()`, and `_snapshot_layer`/`_restore_layer` entirely, keeps the `activate()` API exactly as-is, and makes "add a per-layer cache" a local change to `Layer` that can't silently regress.

jcarius-nv:

newton/_src/viewer/viewer.py
```diff
@@ -46,9 +104,234 @@ def __init__(self):
...
+            layer_id: Identifier of the layer to toggle.
+            visible: ``True`` to show the layer, ``False`` to hide it.
+        """
+        if layer_id not in self._layers:
+            raise KeyError(f"Unknown layer: {layer_id}")
+        self._layers[layer_id].visible = bool(visible)
+        # Force a re-emit of all owned objects on the next frame so the
+        # hidden flag is propagated through the cached log_* calls.
+        if layer_id == self._active_layer_id:
+            self.model_changed = True
```
adenzler-nvidia: `model_changed = True` here doesn't actually drive visibility. In `log_state` the `hidden` flag is recomputed every frame from `self.layer.visible` and re-emitted unconditionally, while `model_changed` only gates the colors/materials re-send (see `viewer_gl.py` ~1559-1560). So this line is a no-op for the stated purpose and the comment is inaccurate.

The practical effect: a visibility change only takes effect when the affected layer is next re-logged via `log_state`. That works in the per-frame re-log loop (including the new example), but a `set_layer_visible` / GUI-panel toggle won't update a layer that isn't being re-logged that frame (e.g. when paused). Either fix the comment to describe what the flag really does, or — if toggling should update cached geometry without a re-log — force the affected layer's geometry to be re-emitted. Note also there's no test covering visibility toggling of an *inactive* layer, which is the core multi-layer case.

newton/_src/viewer/viewer_gl.py
```diff
@@ -467,22 +483,56 @@ def clear_model(self):
...
+                if owns(k):
+                    if hasattr(v, "destroy"):
+                        v.destroy()
+                else:
+                    kept[k] = v
+            return kept
+
+        self.objects = _filter_destroy(getattr(self, "objects", {}))
+        self.lines = _filter_destroy(getattr(self, "lines", {}))
+        self.arrows = _filter_destroy(getattr(self, "arrows", {}))
```
adenzler-nvidia, line 511: Geometry dicts are correctly filtered by `_is_layer_owned_path` here, but further down in this same method `_scalar_buffers`, `_scalar_arrays`, `_array_buffers`, and the image logger are cleared unconditionally (the `self._scalar_buffers.clear()` / `_image_logger.clear()` block). Because `remove_layer()` runs `clear_model()` under the removed layer's context, removing one layer (or re-`set_model` on one layer) wipes the scalar plots / logged arrays / image windows of every *other* live layer. Those keys are already layer-qualified via `_qualify`, so they should be filtered by ownership like the geometry dicts rather than cleared wholesale. `ViewerViser.clear_model` has the same gap for `_plot_handles` / scalar state.

adenzler-nvidia (commented):
Re-review of the layer system. The earlier CodeRabbit threads are all resolved and the core design is sound — the kernel transform math (left-multiply `layer_xform` after the per-world offset) is correct and consistent, and the default single-layer geometry path is genuinely preserved. The inline comments below cover three items worth addressing before merge plus a few minor points.

Two behavioral notes that don't fit on a changed line, so flagging here instead: the "default path unchanged" claim isn't literally true for two backends. Rerun's `log_lines`/`log_points` now emit `rr.Clear` on hidden/None input where they previously returned silently, and USD instance visibility is now time-sampled per-frame instead of written once statically. Both look intentional and are covered by the new tests, but they change the default `.rrd`/USD output — worth a line in the PR description. The branch also currently shows merge conflicts with `main`.

newton/_src/viewer/viewer.py
```diff
@@ -46,9 +104,234 @@ def __init__(self):
...
+
+    def set_layer_transform(
+        self,
+        layer_id: str,
+        xform: wp.transform | tuple[float, float, float] | list[float] | wp.vec3,
+    ) -> None:
+        """Set a per-layer rendering transform.
+
+        The transform is applied to every drawn position/orientation in the
```
adenzler-nvidia, line 326: This says the transform is applied to *every* drawn position/orientation, but `_log_particles` and `_log_triangles` (viewer.py ~2644-2695) pass `state.particle_q` straight through with only name-qualification and the hidden flag — no `layer_xform`. So in a side-by-side layer layout a cloth/particle solver's geometry stays at the origin while the rigid shapes shift. Either apply the layer transform to the particle/triangle paths, or soften this sentence to match the enumeration that follows it (which already omits particles/cloth).

Minor: per AGENTS.md the `xform` translation arg should carry an SI unit in the `Args:` entry, e.g. "a translation [m] as a tuple, list, or `wp.vec3`".

newton/_src/viewer/viewer.py
```diff
@@ -46,9 +104,234 @@ def __init__(self):
...
+            raise KeyError(f"Unknown layer: {layer_id}")
+        if isinstance(xform, (list, tuple)):
+            xform = wp.transform(
+                wp.vec3(float(xform[0]), float(xform[1]), float(xform[2])),
+                wp.quat_identity(),
+            )
+        elif isinstance(xform, wp.vec3):
+            xform = wp.transform(xform, wp.quat_identity())
+        elif not isinstance(xform, wp.transform):
+            xform = wp.transform(*xform)
```
adenzler-nvidia: For an unsupported type (e.g. a numpy array, a dict, or a 7-element sequence that slipped past the `list`/`tuple` branch) this `wp.transform(*xform)` surfaces a low-level unpacking/Warp error rather than a clear message. A `TypeError` naming the accepted types (`wp.transform`, `wp.vec3`, or a 3-element translation) would be more actionable.

newton/_src/viewer/viewer.py
```diff
@@ -46,9 +104,234 @@ def __init__(self):
...
+        rendering immediately and no GPU resources leak. If the removed
+        layer is currently active, the default layer is re-activated. The
+        internal default layer cannot be removed.
+
+        Args:
+            layer_id: Identifier of the layer to remove.
+        """
+        if layer_id == _DEFAULT_LAYER_ID:
+            raise ValueError("Cannot remove the default layer")
+        if layer_id not in self._layers:
```
adenzler-nvidia, line 274: `remove_layer` on an unknown id silently returns here, whereas `set_layer_visible` and `set_layer_transform` raise `KeyError` for the same condition. Either raise `KeyError` here too for consistency, or document the deliberate idempotency. Relatedly, `activate()` silently creates a layer on a typo'd id — worth a one-line doc caution since there's no strict variant.

newton/_src/viewer/viewer.py
```diff
@@ -46,9 +104,234 @@ def __init__(self):
...
+        # empty name prefix to keep backward compatibility for code that
+        # never calls activate().
+        self._layers: dict[str, Layer] = {}
+        self._active_layer_id: str = _DEFAULT_LAYER_ID
+        self._layers[_DEFAULT_LAYER_ID] = Layer(_DEFAULT_LAYER_ID)
+
         # All model-dependent state is initialized by clear_model()
         self.clear_model()
 
+    def __getattribute__(self, name: str) -> Any:
```
adenzler-nvidia: This intercepts every non-dunder attribute access (a `__dict__` lookup plus `hasattr(layer, name)`) so it runs on hot-loop accesses like `self.model`, `self.world_offsets`, and `self._shape_instances`. It's functionally correct — no name collisions between `Layer` and viewer-level attributes — but worth being aware of the per-access overhead in the render loop. A short docstring on `__getattribute__`/`__setattr__` explaining the routing and the `__`-prefix escape hatch would also help future maintainers, since attribute interception is surprising.

newton/_src/viewer/viewer_rtx.py
```diff
@@ -724,7 +727,11 @@ def _init_ovrtx(self):
             except Exception as e:
                 raise RuntimeError(f"Failed to create window: {e}") from e
 
-        self._build_flat_shape_arrays()
+        self._use_layered_transform_updates = any(layer_id != _DEFAULT_LAYER_ID for layer_id in self._layers)
```
adenzler-nvidia, line 740: `_use_layered_transform_updates` is latched once at the build→render transition from the current `self._layers`. If a layer is first activated during the render phase, this stays `False`, the flat-shape-array fast path (built from the default layer only) keeps being used, and the newly added layer's shapes won't receive transform updates via that path. The documented usage activates layers before rendering, so this is an edge case, but a guard or recompute when a new layer appears would be more robust.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Add `newton.use_coord_layout_targets` opt-in flag exposing `Model.joint_target_q` / `Control.joint_target_q` shaped `(joint_coord_count,)` (matching `joint_q`) and `joint_target_qd` shaped `(joint_dof_count,)` (matching `joint_qd`); solvers, the actuator library, and `ModelBuilder.finalize()` honor the flag. Defaults to `False` for backwards compatibility; will flip in a future release.
 - Add opt-in `validate_mesh` parameter to `ModelBuilder.add_cloth_mesh()`, `ModelBuilder.add_soft_mesh()`, and `style3d.add_cloth_mesh()` that warns on degenerate geometry; add public `newton.utils.validate_triangle_mesh()` and `newton.utils.validate_tet_mesh()` utilities
+- Add viewer layer system to overlay multiple solvers/models in supported rendering viewers; call `Viewer.activate(layer_id)` to route subsequent `set_model` / `log_state` / `log_*` calls into a named layer, and `Viewer.set_layer_visible()` to toggle layers independently. See `example_basic_multi_solver_overlay.py`
```
adenzler-nvidia: There's no public `Viewer` symbol — these methods live on `ViewerBase` (the exported classes are `ViewerBase`, `ViewerGL`, `ViewerUSD`, …). Reference `ViewerBase.activate(...)` / `ViewerBase.set_layer_visible(...)` instead. Also consider mentioning `set_layer_transform`, since per-layer positioning is a key part of the feature.

CHANGELOG.md
```diff
@@ -53,6 +55,12 @@
 - Add `StateFlags` enum to control which state attributes are reset. (#2657)
 - Add `ModelFlags` as the canonical name for model-change notification flags. (#2657)
 - Add opt-in `validate_mesh` parameter to `ModelBuilder.add_cloth_mesh()`, `ModelBuilder.add_soft_mesh()`, and `style3d.add_cloth_mesh()` that warns on degenerate geometry; add public `newton.utils.validate_triangle_mesh()` and `newton.utils.validate_tet_mesh()` utilities
+- Warn from `SolverMuJoCo` when a `JointType.FREE` joint has a non-world parent; MuJoCo requires free joints to attach directly to the world.
```
adenzler-nvidia: again some changelog issues here

newton/_src/viewer/viewer.py
```diff
@@ -90,86 +429,121 @@ def clear_model(self) -> None:
...
 
         # Characteristic body size in world units, used to auto-scale
         # visualization helpers (contact arrows, joint axes, COM markers).
         # Set in :meth:`set_model` from :meth:`_estimate_scene_scale`; falls
         # back to 1.0 when no dynamic shapes are present.
-        self.scene_scale: float = 1.0
+        layer.scene_scale: float = 1.0
 
         # Picking
         self.picking_enabled = True
```
adenzler-nvidia: `picking_enabled` is written to `self`, not `layer`, inside `_init_layer_state` — whose contract is to initialize fields on the passed-in `layer`. Two consequences:

1. Every `activate()` of a *new* layer runs this and silently re-enables picking even if the user had disabled it.
2. It's the one `self.`-write in a method otherwise full of `layer.`-writes, so a future edit that "fixes" it to `layer.picking_enabled` would turn it into a genuine per-layer field and create a viewer/layer desync.

Suggest moving the `picking_enabled` default into `__init__` so `_init_layer_state` only ever touches `layer`.

newton/_src/viewer/viewer.py
```diff
@@ -46,9 +105,289 @@ def __init__(self):
...
+
+        Args:
+            layer_id: Stable identifier for the layer. Re-activates an
+                existing layer when the id is already known.
+
+        Returns:
+            Layer: The activated layer object.
+        """
+        if not isinstance(layer_id, str) or not layer_id:
+            raise ValueError("layer_id must be a non-empty string")
```
adenzler-nvidia, line 240: `activate()` rejects empty/non-string ids but not the internal sentinel `"__default__"`. `activate("__default__")` re-activates the internal default layer, and since `Layer.name_prefix` special-cases that id to `""`, a user layer with that name would silently emit *unprefixed* names and collide with the legacy namespace — while `remove_layer("__default__")` raises, so it can be created-by-activate but never removed.

Suggest rejecting `layer_id == _DEFAULT_LAYER_ID` here with a clear error, or using a non-string sentinel that user ids (always strings) can never equal.

newton/_src/viewer/viewer.py
```diff
@@ -90,86 +429,121 @@ def clear_model(self) -> None:
...
+            name: Backend object name (path).
+
+        Returns:
+            bool: True if the object belongs to the active layer.
+        """
+        prefix = self.layer.name_prefix
+        if prefix:
+            return name.startswith(prefix + "/") or name == prefix
+        # Default layer: own everything that is NOT inside a "/layers/" namespace.
+        return not name.startswith("/layers/")
```
adenzler-nvidia: `_is_layer_owned_path` and `_qualify` disagree about names starting with `/layers/` when the **default** layer is active. `_qualify` (line 383) returns a `/layers/...` name unchanged on the default layer (empty prefix), but `_is_layer_owned_path` here returns `not name.startswith("/layers/")` → `False` for it. So an object logged under a `/layers/...` path while on the default layer is owned by no layer and is never destroyed by `clear_model()` / `remove_layer()` — a backend resource leak.

Unusual input, but the two predicates should agree: either reserve the `/layers/` namespace (reject/warn on the default path) or have the default layer own any `/layers/` name that no live layer claims.

newton/_src/viewer/viewer_rerun.py
```diff
@@ -245,6 +245,8 @@ def log_mesh(
             metallic: Metallicity in ``[0, 1]``. ``0`` is dielectric, ``1``
                 is metal.
         """
+        name = self._qualify(name)
```
adenzler-nvidia, line 278: `ViewerRerun` has no `clear_model()` override, so `remove_layer()` runs only the base (Python-cache) cleanup and never emits an `rr.Clear` for the removed layer's `/layers/<id>` entity subtree. The removed layer's geometry therefore stays in the Rerun recording/stream after removal, unlike `ViewerUSD`/`ViewerGL` which destroy the active layer's backend objects.

Consider adding a `clear_model()` override (or removal hook) that clears the active layer's entity subtree.

adenzler-nvidia (commented):
A few targeted findings on the layer system (the core feature looks good and the default no-`activate` path is preserved). These are robustness issues worth addressing; see inline comments.

newton/_src/viewer/viewer.py
```diff
@@ -46,9 +105,289 @@ def __init__(self):
...
+        raise AttributeError(f"{type(self).__name__!s} object has no attribute {name!r}")
+
+    def __setattr__(self, name: str, value: Any) -> None:
+        """Keep active layer-owned fields synchronized on writes."""
+        object.__setattr__(self, name, value)
+        layers = self.__dict__.get("_layers")
+        active_layer_id = self.__dict__.get("_active_layer_id")
+        if layers is not None and active_layer_id in layers:
+            layer = layers[active_layer_id]
+            if not name.startswith("__") and name in layer.__dict__:
```
adenzler-nvidia: Question about how the set of layer-owned fields is defined here. The mirror keys off `name in layer.__dict__`, and `_is_layer_runtime_field` just below is a denylist of the three config fields — so "which fields are layer-owned" is effectively expressed in three places: this `__dict__` check, the denylist, and the actual `layer.x = ...` writes in `_init_layer_state`. Two consequences I wanted to check are intentional:

1. A write to a layer-owned field that happens *before* that field has been seeded into `layer.__dict__` isn't mirrored (the guard is false), so it lives only on the viewer and is overwritten on the next `activate()`. That's safe today only because `_init_layer_state` pre-seeds everything — is that invariant something worth pinning explicitly?
2. The denylist fails open: a future per-model field added via a plain `self.x = ...` (instead of in `_init_layer_state`) is silently treated as runtime and can desync across layers with no error.

Would it make sense to snapshot the layer field names once after the default layer is initialized — e.g. `frozenset(default_layer.__dict__) - {"layer_id", "visible", "xform"}` — and have `__setattr__`/save/load route against that allowlist instead? It keeps the current fast-read mirror exactly as-is (no perf change vs. the `__getattribute__`/property-descriptor approaches you tried earlier), but makes the field set a single fail-closed source of truth rather than three implied ones. Curious whether you'd considered that.

newton/_src/viewer/viewer_rtx.py
```diff
@@ -1994,6 +2044,7 @@ def clear_model(self) -> None:
         self._flat_shape_scales = None
         self._flat_total_shapes = 0
         self._flat_mat44_offset = 0
+        self._use_layered_transform_updates = False
```
adenzler-nvidia, line 2053: **RTX `clear_model` isn't layer-aware — `remove_layer` / re-`set_model` corrupts other layers**

This PR makes the other backends' `clear_model` (GL, USD, Rerun, Viser) filter their per-name caches by `_is_layer_owned_path` so clearing one layer leaves the others intact, but RTX's `clear_model` still resets every build dict wholesale. Those dicts are keyed by the qualified name — e.g. `_mesh_prim_paths[name]` / `_instance_prim_paths[name]` at lines 1454/1499 — so they hold every layer's prims under `/layers/<id>/...`.

Because `clear_model` runs under the active layer's context whenever `remove_layer()` is called, or whenever `set_model()` is called a second time on an already-modeled layer (`self.model is not None`), removing or re-modeling one RTX layer silently drops every *other* layer's prims, meshes, and batches — the exact cross-layer wipe this PR fixes elsewhere. The build-then-render overlay path itself is fine (each new layer is activated with `self.model is None`, so setup never hits `clear_model`), and the activate-before-render guard doesn't cover this since it only blocks *new* layers after render starts, not removal/re-model of existing ones.

One wrinkle: RTX `clear_model` also does genuinely global renderer teardown (`self._rtx = None`, drains the async pipeline, unbinds transforms, resets phase/camera) that can't be ownership-filtered. So the clean fix mirrors the USD split — when other user layers remain, remove only the active layer's prims and skip the global teardown; do the full teardown only when the last layer is gone (cf. `_has_user_layers()` / `_remove_active_layer_prims()` in `viewer_usd.py`).

If full per-layer RTX teardown is out of scope here, at minimum raise a clear "not supported" error when `clear_model` would run with other live user layers, so this fails loudly instead of silently corrupting them.

adenzler-nvidia (commented):
One blocking item on the RTX backend's layer support, inline below. It's the only backend whose `clear_model` wasn't made layer-aware in this PR; the rest of my earlier feedback looks addressed.

adenzler-nvidia (approved):
Re-reviewed at `28a60403a` — looks like everything from my earlier reviews is addressed, including the RTX `clear_model` fail-loud fix. Thanks for working through all of it, @nvtw!

PR #2877, opened 2026-05-13
Use absolute VBD damping

AnkaChan:
## Description

Switch VBD damping coefficients from stiffness-relative factors to absolute
physical damping units across particle elasticity, springs, contacts, and rigid
joints.

This PR also changes collision damping to use relative normal contact motion
instead of absolute displacement, so uniform stencil motion and tangential
sliding are not damped by normal collision damping.

Key changes:

- Use absolute damping for VBD Neo-Hookean membrane/tet elasticity, springs,
  bending, joint constraints, joint drives/limits, rigid contacts, and
  body-particle contacts.
- Use objective metric-based damping for Neo-Hookean membrane/tet elasticity,
  based on the rate of `C = F^T F`, so rigid rotations do not add damping force.
- Compute self-contact damping from relative normal gap rate.
- Preserve AVBD contact ramp behavior by scaling damping with the same current
  penalty ratio used for stiffness.
- Convert examples and tests from the old relative `kd` convention to absolute
  coefficients.
- Update solver docs, model/builder parameter docs, and `CHANGELOG.md`.

Side-by-side comparison (left: before, right: after, covering most of the impacted examples):

https://github.com/user-attachments/assets/c8e05080-3c80-4879-a521-1bc29067e99f

[![Side-by-side VBD damping comparison: before on the left, after on the right]

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k damping
uv run --extra dev -m newton.tests -k test_cable
uv run --extra dev -m newton.tests -k test_cloth
uv run --extra dev -m newton.tests -k softbody
uv run --extra dev -m newton.tests -k test_rigid_contact
uvx --with virtualenv pre-commit run -a
git diff --check
```

After the final dead-code cleanup, re-ran:

```bash
uvx pre-commit run -a
git diff --check
```

Recorded 4-second before/after MP4 captures for all affected examples at
1280x720 and 30 fps. All 32 MP4s and posters were generated successfully and
poster-frame pixel checks were nonblank.

## Bug fix

**Steps to reproduce:**

1. Configure VBD contact damping or self-contact damping on a moving contact
   stencil.
2. Move all vertices in the contact stencil uniformly, or slide tangentially
   along the contact.
3. Observe damping from absolute displacement even when the contact gap is not
   closing.

With this PR, collision damping is based on relative normal gap rate and is
only applied for approaching contacts.

**Minimal reproduction:**

The new tests in `newton/tests/test_solver_vbd.py` cover:

- body-particle contact damping using absolute coefficients, independent of
  contact stiffness;
- self-contact damping using relative gap rate, including uniform contact
  stencil motion.

## New feature / API change

VBD now interprets damping parameters as absolute damping coefficients instead
of dimensionless stiffness-relative multipliers. Migrate old VBD damping
settings per constraint family with:

```python
new_damping = old_relative_damping * stiffness_previously_scaled
```

Use the stiffness or penalty term that the old damping value was paired with:
elastic elements, edge bending, springs, cable stretch/bend joints, joint
targets/limits, rigid-joint penalties, and contact penalties. If a value was
not used by VBD or was already absolute for another solver, do not convert it
just because it is named `kd`.

## Summary by CodeRabbit

* **Changed**
  * Damping parameters are now interpreted as absolute physical coefficients (e.g., Pa·s, N·s/m); many defaults adjusted upward for consistent solver behavior.

* **Bug Fixes**
  * Contact damping now uses relative normal gap rate to avoid artificial normal damping during uniform contact-stencil motion and tangential sliding.

* **Documentation**
  * Clarified damping parameter descriptions and units across solver/API docs.

* **Examples**
  * Updated example/demo damping defaults to match new absolute-coefficient semantics.

* **Tests**
  * Added tests verifying absolute damping behavior and gap-rate invariance.

* **Chores**
  * Runtime migration warning added to inform about legacy kd conversion.


jumyungc:
Thank you for changing the damping convention. @AnkaChan 

1. I see why damping_scale was introduced for ramped AVBD contacts, but I would still prefer to treat kd as an absolute damping coefficient with no additional scaling. That seems cleaner and more transparent: if a user sets kd, the solver applies that kd.
Last time I checked, none of the examples seemed to rely on kd scaling, and penalty ramping is not encouraged. Can we use kd without internal scaling?

2. This is a user-facing breaking change so I am wondering whether we should follow the breaking-change/deprecation rule.

gdaviet:
I am a bit concerned that this mixes the parameter change (relative -> absolute) with damping energy changes.

 - The contact damping change looks great -- damping the positive normal response is what is expected.
 - The changes to the elastic energies are less straightforward. This now damps |\dot{F}|, which is not frame-invariant -- so you will get damping pour pure rigid rotations. Would damping the NH modes instead make sense?
 - The spring damping currently include geometric stiffness, which damps tangential motion. Should be dropped at least in compression (this introduces negative values), but I believe it would also be more acceptable to simply damp along the spring direction (similar to contacts, but bidirectional)

newton/_src/solvers/vbd/particle_vbd_kernels.py
```diff
@@ -2439,6 +2233,7 @@ def accumulate_contact_force_and_hessian_no_self_contact(
         if particle_colors[particle_idx] == current_color:
             # Read per-contact AVBD penalty and material properties shared with the rigid side
             contact_ke = body_particle_contact_penalty_k[t_id]
+            contact_damping_scale = contact_ke / wp.max(body_particle_contact_material_ke[t_id], 1.0e-12)
```
eric-heiden: 🔴 This changes body-contact damping for `SolverStyle3D` too, not just VBD. `newton/_src/solvers/style3d/collision/kernels.py` still calls `evaluate_body_particle_contact()`, and that wrapper now passes `damping_scale=1.0` into `_eval_body_particle_contact()`. Since `_compute_body_particle_contact_force()` now uses `kd * damping_scale`, Style3D loses the previous `kd * ke` behavior for these contacts.

That is a silent behavior change outside the VBD migration described by this PR. Please either preserve the old wrapper behavior for non-VBD callers, for example by passing `body_particle_contact_ke` as the damping scale from `evaluate_body_particle_contact()`, or explicitly migrate and document the Style3D damping change with matching tests/examples.

jcarius-nv:

newton/_src/solvers/vbd/particle_vbd_kernels.py
AnkaChan: I chose to migrate the Style3D body-particle contact path to the same absolute-`kd` convention instead of duplicating the shared contact kernel to preserve the old `kd * ke` behavior.

Style3D has its own internal cloth damping model, and this shared body-soft contact damping is only one secondary damping path for that solver. Keeping a special legacy wrapper just for this contact path did not seem worth the extra duplicated kernel/API surface, especially because it would leave one shared contact entry point where `kd` still means stiffness-relative damping while the rest of this PR makes `kd` absolute.

To avoid silently changing the examples, I retuned the affected Style3D example values to preserve the old effective damping scale:
- `example_cloth_style3d.py`: `soft_contact_kd` from `1.0e-6` to `1.0e-5` because `soft_contact_ke = 1.0e1`
- `example_cloth_hanging.py`: Style3D branch uses `1.0e2` because `soft_contact_ke = 1.0e2`

I also ran `uv run --extra dev -m newton.tests -k test_solver_style3d`.

AnkaChan:
@jumyungc Addressed both items from https://github.com/newton-physics/newton/pull/2877#issuecomment-4492146515.

Contact damping now treats `kd` as an absolute coefficient with no extra scaling by contact/penalty stiffness. The body-particle and body-body contact damping tests check that changing the penalty ramp does not change the damping contribution.

For the migration side, `SolverVBD` now emits a one-time `UserWarning` explaining the convention change and the legacy conversion:
`kd_new = kd_old * k`, where `kd_old` is the previous relative damping value and `k` is the stiffness/penalty coefficient for the same constraint, contact, or material.

AnkaChan:
@gdaviet Addressed the damping-energy concerns from https://github.com/newton-physics/newton/pull/2877#issuecomment-4506121796 by changing the formulations rather than keeping `|F_dot|` damping.

Membrane/tet damping now uses an objective metric-rate formulation based on `C = F^T F`, so pure rigid rotations do not produce damping force. Spring damping is now axial: it damps only edge length change along the spring direction, with an axial Hessian, so tangential rigid rotation is not damped by spring damping. Bending damping uses dihedral angle rate, and contact damping remains relative normal gap-rate based.

I also added fixed-seed rigid-motion tests over 100 random SO(3) + translation transforms covering spring, membrane, tet, bending, and rigid-rigid / rigid-soft / soft-soft contact damping.

eric-heiden (dismissed):

gdaviet (dismissed):
Thanks for making the damping changes !
I would still recommend mentioning the model changes in the CHANGELOG; there is an entry for the contact damping model change, but it should mention the elastic and spring changes as well

adenzler-nvidia (changes requested):
Thanks for this — moving VBD to absolute damping coefficients is the right call. I'm explicitly fine taking this as an **immediate, hard break**: `SolverVBD` is marked experimental, so it doesn't need the deprecation cycle a stable API would, and a units/semantics change like this can't be deprecated per-call anyway — an old relative `kd=0.5` and a new absolute `kd=0.5` are indistinguishable at the value level, so there's nothing to attach a targeted `DeprecationWarning` to.

My one ask before merge is that `CHANGELOG.md` make the break **unmistakable and complete**. Right now a reader scanning `### Changed` gets no signal that their tuned `kd` values are silently invalidated. Three points:

**1. The entries are in the wrong section.** Both new lines currently sit inside the already-released `## [1.2.0] - 2026-05-12` section. The branch is also behind `main` (1.3.0 shipped 2026-06-11), so please merge `main` and move both entries into `## [Unreleased]` — this ships in 1.4.0.

**2. The runtime warning hardcodes the wrong version.** `solver_vbd.py` says "changed in Newton 1.2.0" and "after the Newton 1.2 migration window". This lands in 1.4.0, so both should say 1.4.0.

**3. Mark it as breaking and enumerate everything.** Could we replace the `Changed` entry with an explicit breaking note, e.g.:

> - **Breaking change (experimental `SolverVBD`):** VBD now interprets all damping coefficients as absolute physical units instead of dimensionless stiffness-relative (Rayleigh) multipliers (`D = kd · ke`). Existing `kd`-family values will produce different damping. Affected parameters: tetrahedral `k_damp` [Pa·s], `tri_kd`, spring `kd` [N·s/m], cable `stretch_damping` [N·s/m] and `bend_damping` [N·m·s/rad] in `add_joint_cable()`/`add_rod()`/`add_rod_graph()`, `joint_target_kd` and `joint_limit_kd` (including `JointDofConfig.limit_kd`), shape contact `kd`/`shape_material_kd` and `soft_contact_kd` [N·s/m], and `SolverVBD(rigid_joint_linear_kd=…, rigid_joint_angular_kd=…)`. To preserve previous behavior, set `kd_new = kd_old · k`, where `k` is the stiffness or penalty coefficient the value was previously paired with, and pass the product to the same field.

Keep the contact-damping fix in `Fixed`, and please also call out the membrane/tet damping model change so all three behavior changes are documented:

> - Change VBD Neo-Hookean membrane/tet damping to an objective metric based on the rate of `C = FᵀF`, so rigid-body rotations no longer generate damping force.

To be clear, I'm OK with the immediate break given VBD's experimental status — this is purely about making the changelog and the runtime warning state it loudly, completely, and against the right version.

AnkaChan:
Thanks @adenzler-nvidia — agreed on all three, and good to confirm we're treating this as an immediate hard break given `SolverVBD`'s experimental status. Pushed in 83ac6eaf:

1. Merged `main` and moved both VBD entries into `[Unreleased]` (ships in 1.4.0).
2. Updated the `solver_vbd.py` runtime warning to say 1.4.0 / "1.4 migration window".
3. Replaced the `Changed` line with the explicit breaking note enumerating every affected `kd`-family parameter (your wording), kept the gap-rate fix under `Fixed`, and added entries for the objective `C = FᵀF` membrane/tet damping model change and the now-axial-only spring damping (per @gdaviet's request to also call out the elastic/spring changes).

PTAL when you have a moment.

adenzler-nvidia (approved):
Looking good now, thank you!

PR #2880, opened 2026-05-14
Added rigid body vbd migration instructions

jumyungc:
## Description

Added rigid body vbd migration instructions/warnings:
Following this instruction from the group:

`Adding warnings when constructing VBD solver. Telling users the behavior has changed, and what parameter to changes to if users want the old behavior
`

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **Improvements**
  * Solver now honors model-level joint hard/soft settings at initialization and keeps runtime per-joint updates working.

* **Documentation**
  * Clarified guidance on authoring per-slot hard/soft joint modes and selecting structural slots (linear/angular and cable aliases).

* **Bug Fixes**
  * Added a one-time Newton 1.2.0 migration warning when rigid bodies are present with internal rigid integration.

* **Examples**
  * Updated example to register VBD attributes before build, set joint hard/soft via vbd:joint_is_hard, disable Dahl cable friction, and construct the solver after model configuration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2880?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


newton/examples/contacts/example_contacts_rj45_plug.py
```diff
@@ -381,14 +382,14 @@ def __init__(self, viewer, args=None):
 
         self._initial_body_q = self.state_0.body_q.numpy().copy()
 
+        self.model.vbd.joint_is_hard = wp.zeros_like(self.model.vbd.joint_is_hard)
```
adenzler-nvidia: Can we author this VBD mode when the joints are created instead of finalizing with the default `joint_is_hard=1` and then replacing the finalized model array? `SolverVBD.register_custom_attributes()` declares `vbd:joint_is_hard` as a JOINT custom attribute, and `add_joint_d6()` / `add_joint_revolute()` already accept `custom_attributes`, so these two joints can be created with `custom_attributes={"vbd:joint_is_hard": 0}`. That keeps the model correct from construction time, avoids a broad post-finalize overwrite, and leaves the rod/cable joints alone since VBD already initializes cable stretch/bend as soft.

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -376,6 +382,25 @@ def __init__(
         rigid_avbd_linear_beta = rigid_avbd_linear_beta if rigid_avbd_linear_beta is not None else rigid_avbd_beta
         rigid_avbd_angular_beta = rigid_avbd_angular_beta if rigid_avbd_angular_beta is not None else rigid_avbd_beta
 
+        if model.body_count > 0 and not integrate_with_external_rigid_solver and not _vbd_migration_warned["value"]:
+            warnings.warn(
+                "SolverVBD rigid-body defaults changed in Newton 1.2.0:\n"
+                "- Non-cable structural joints are now hard augmented-Lagrangian constraints by default. "
+                "To restore soft penalty behavior, set the relevant entries of model.vbd.joint_is_hard to 0 "
```
adenzler-nvidia: Can we make this migration guidance self-contained? Existing models that did not call `SolverVBD.register_custom_attributes(builder)` will not have `model.vbd.joint_is_hard`, so following this text literally can fail with `AttributeError` or be impossible after finalization. The warning should spell out the required setup, e.g. register VBD custom attributes before building/finalizing, then author `custom_attributes={"vbd:joint_is_hard": 0}` on the relevant joints or set `model.vbd.joint_is_hard.fill_(0)` before constructing `SolverVBD`. Also, please link a cleanup issue or add a `TODO(#...)` / target release for removing this migration warning so it does not become permanent runtime noise.

newton/examples/contacts/example_contacts_rj45_plug.py
```diff
@@ -239,6 +239,7 @@ def __init__(self, viewer, args=None):
         latch_mesh, lc = _load_mesh(stage, "/World/Latch")
 
         builder = newton.ModelBuilder(gravity=-9.81)
+        SolverVBD.register_custom_attributes(builder)
```
adenzler-nvidia: This registration has a second side effect beyond making `joint_is_hard` available: `SolverVBD.register_custom_attributes()` also registers `vbd:dahl_eps_max` / `vbd:dahl_tau` with nonzero defaults, and `SolverVBD` auto-enables Dahl cable friction whenever those attributes exist. This example has rod/cable joints, so adding this line can change cable bending behavior even if the only intended migration is to keep the non-cable joints soft. Please either explicitly disable Dahl for this example, e.g. set `model.vbd.dahl_eps_max` to zero before constructing the solver, or provide/use a narrower way to author only `vbd:joint_is_hard`.

newton/_src/solvers/vbd/solver_vbd.py
jumyungc: Agreed. I made it explicit that SolverVBD.register_custom_attributes(builder) must be called before finalizing the model.

newton/examples/contacts/example_contacts_rj45_plug.py
jumyungc: Thank you for the good catch. I will follow up on this separately shortly:
https://github.com/newton-physics/newton/issues/2903

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -376,6 +382,25 @@ def __init__(
...
+                "- Non-cable joint stiffness ceilings now default to 1.0e5 instead of 1.0e9 to make the "
+                "fixed-stiffness path (rigid_avbd_beta=0.0) safer. To restore the previous ceilings, pass "
+                "rigid_joint_linear_ke=1.0e9 and rigid_joint_angular_ke=1.0e9.\n"
+                "- Cable rod stretch/bend stiffness inputs are now interpreted as direct per-joint values; "
+                "the previous length normalization in rod helpers has been removed. To recover the previous "
+                "behavior, divide material stiffnesses by segment length before building the model.",
+                UserWarning,
+                stacklevel=2,
+            )
+            _vbd_migration_warned["value"] = True
```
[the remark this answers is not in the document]
jumyungc: This PR is, adding migration guide, not actual changes.

newton/examples/contacts/example_contacts_rj45_plug.py
jumyungc: Reverted unrelated changes that were left in the migration guide branch.

adenzler-nvidia:
Could you rebase this on top of #2921 once it lands (or coordinate ordering)? Two spots need updating:

1. The migration warning's "soft structural joints" workaround currently advises calling `model.vbd.dahl_eps_max.fill_(0.0)` to avoid the implicit Dahl side-effect. With #2921 this becomes `SolverVBD.register_custom_attributes(builder, dahl_defaults_enabled=False)` — much cleaner.
2. The `set_joint_constraint_mode` docstring change here overlaps with the equivalent change in #2921, so that hunk will conflict / become redundant.

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -380,6 +383,31 @@ def __init__(
...
+                "  - Non-cable structural joints default to hard augmented-Lagrangian constraints.\n"
+                "  - Body-body contacts default to hard augmented-Lagrangian constraints.\n"
+                "  - AVBD penalty ramping is off by default (rigid_avbd_beta=0.0).\n"
+                "  - Non-cable joint stiffness ceilings default to 1.0e5 instead of 1.0e9.\n"
+                "  - Cable rod stretch/bend stiffness inputs are direct per-joint values.\n"
+                "To restore previous behavior:\n"
+                "  - Soft structural joints: call "
+                "SolverVBD.register_custom_attributes(builder, dahl_defaults_enabled=False) before adding "
+                "joints, then pass custom_attributes={'vbd:joint_is_hard': 0} to the relevant add_joint_* "
+                "calls. Rebuild models finalized without VBD custom attributes.\n"
```
adenzler-nvidia: This migration step frames `dahl_defaults_enabled=False` as part of restoring soft structural joints, but that flag has no effect on joint hardness — it only controls the Dahl friction defaults (`dahl_eps_default`/`dahl_tau_default`). The `vbd:joint_is_hard` attribute defaults to `1` regardless of `dahl_defaults_enabled`, so the only steps actually required to restore soft joints are registering the custom attributes and authoring `vbd:joint_is_hard=0`.

The `dahl_defaults_enabled=False` is really only carried over from the canonical example to suppress the *separate* Dahl deprecation warning. As written, a user could reasonably conclude the flag governs joint hardness. Suggest either dropping `dahl_defaults_enabled=False` from this line, or adding a short clause noting it only suppresses the unrelated Dahl warning.

jumyungc:
Now that we have https://github.com/newton-physics/newton/pull/3068, do you think we still need this PR? All resolved anyway. @adenzler-nvidia

adenzler-nvidia:
let's close this and focus on doing better in the future - even with experimental status it's nice to give users a good upgrade path. But I agree this is so outdated now it does not make a lot of sense to add this noise.

PR #2894, opened 2026-05-18
Improve sdf contact preprocessing and contact generation

nvtw:
## Description
Improve SDF-mesh contact preprocessing and contact generation. Adds a dihedral-angle pre-filter for the precomputed mesh edges iterated by the SDF-mesh narrow phase, an opt-in box-absorption pass that drops manifold edges fully covered by another edge's oriented box, and a hardware-filtered SDF sampling path used by the SDF-mesh narrow phase. Brent's tolerance floor on the edge-vs-SDF search is now scaled by the caller's contact precision so very short edges exit in zero iterations while long edges still converge. The redundant SDF value sample after Brent is skipped by reusing Brent's converged value (the gradient — and therefore the contact normal — is still freshly sampled, just without re-sampling the value at the same point). All edge-simplification configuration lives on `Mesh.build_sdf()`; the simplified edge set is cached on the mesh and `ModelBuilder.finalize()` consumes it transparently.
## Checklist
## Test plan
```
uv run --extra dev -m newton.tests -k test_edge_redundancy
uv run --extra dev -m newton.tests -k test_mesh_edge_angle_filter
uv run --extra dev -m newton.tests -k test_sdf
```
## New feature / API change
```python
import math
import newton
mesh = newton.Mesh(vertices, indices)
mesh.build_sdf(
    edge_lower_angle_threshold_rad=math.radians(0.1),  # default; set 0 to keep every manifold edge
    edge_upper_angle_threshold_rad=math.radians(10.0),  # default; only used with edge_box_absorption
    edge_box_absorption=True,                           # opt-in; default False
    # Mutually exclusive absolute (metres) / relative (fraction of AABB diagonal) box half-extents.
    # "half_normal" is along the edge normal; "half_lateral" is in-plane across the edge and as per-end overhang along it:
    edge_box_half_normal=None,       # default None -> use edge_box_half_normal_rel
    edge_box_half_normal_rel=1e-3,   # default 1e-3
    edge_box_half_lateral=None,      # default None -> use edge_box_half_lateral_rel
    edge_box_half_lateral_rel=5e-3,  # default 5e-3
)
body = builder.add_body()
builder.add_shape_mesh(body=body, mesh=mesh)
model = builder.finalize()
```

## Summary by CodeRabbit

* **New Features**
  * Mesh SDF edge-simplification with dihedral thresholds and optional oriented-box absorption (absolute or relative extents), cached collision-edge usage, and a precision-target-aware SDF contact search that uses hardware-filtered sampling with a faster gradient-only path; opt-out restores prior behavior.

* **Tests**
  * Added unit and integration tests for edge redundancy, box absorption, removal resolution, dihedral filtering, and cache/lifecycle semantics.

* **Documentation**
  * CHANGELOG and hydroelastic contact docs updated; hydroelastic handling marked experimental.

* **Chores**
  * Benchmark adjusted to avoid interference from edge-simplification during timing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2894?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


adenzler-nvidia (commented):
The CHANGELOG only documents the new edge-filter API; the SDF-mesh narrow phase changes (HW texture sampling, centred-difference gradients, Brent precision floor) are user-visible behavior changes too and should be called out. Could you add two `Changed` entries:

```markdown
### Changed
- **Breaking:** Drop internal coplanar mesh edges from SDF-mesh contact
  generation by default (dihedral angle < 0.1°). Set
  `ModelBuilder.mesh_edge_lower_angle_threshold_rad = 0.0` to restore the
  previous unfiltered behavior. Boundary and non-manifold edges are always
  kept.

- **Breaking:** Switch the SDF-mesh narrow phase to hardware-filtered SDF
  texture sampling and centred-difference gradients, with a precision-aware
  Brent search for edge-vs-SDF contacts. Contact distances may shift by up
  to ~1/256 of the SDF voxel size and contact normals by a few milliradians;
  the error scales with voxel size, so users running coarse SDFs (≤64³) on
  large objects may want to increase resolution if sub-mm contact precision
  matters. Hydroelastic SDF is unaffected.
```

The second entry introduces a breaking numerical change with no user-facing opt-out — we should discuss this before merge. Could you share more detail on:

- What workflows you'd expect to break, if any (regression suites, hydroelastic-adjacent code, anything with tight contact tolerances)?
- What concrete migration guidance you'd give a user who hits a regression?
- Whether the changes were validated against any reference simulations / golden contact data, and if so what the observed deltas looked like in practice?

nvtw:
> The CHANGELOG only documents the new edge-filter API; the SDF-mesh narrow phase changes (HW texture sampling, centred-difference gradients, Brent precision floor) are user-visible behavior changes too and should be called out. Could you add two `Changed` entries:
> 
> ```
> ### Changed
> - **Breaking:** Drop internal coplanar mesh edges from SDF-mesh contact
>   generation by default (dihedral angle < 0.1°). Set
>   `ModelBuilder.mesh_edge_lower_angle_threshold_rad = 0.0` to restore the
>   previous unfiltered behavior. Boundary and non-manifold edges are always
>   kept.
> 
> - **Breaking:** Switch the SDF-mesh narrow phase to hardware-filtered SDF
>   texture sampling and centred-difference gradients, with a precision-aware
>   Brent search for edge-vs-SDF contacts. Contact distances may shift by up
>   to ~1/256 of the SDF voxel size and contact normals by a few milliradians;
>   the error scales with voxel size, so users running coarse SDFs (≤64³) on
>   large objects may want to increase resolution if sub-mm contact precision
>   matters. Hydroelastic SDF is unaffected.
> ```
> 
> The second entry introduces a breaking numerical change with no user-facing opt-out — we should discuss this before merge. Could you share more detail on:
> 
> * What workflows you'd expect to break, if any (regression suites, hydroelastic-adjacent code, anything with tight contact tolerances)?
> * What concrete migration guidance you'd give a user who hits a regression?
> * Whether the changes were validated against any reference simulations / golden contact data, and if so what the observed deltas looked like in practice?

I don't expect the changes to lead to significant different contacts. We never promised that the contacts will appear at exactly the same spot, especially since most simulators don't support determinism anyway. I added the changelog entries but the changes from this MR are no really breaking.

adenzler-nvidia:
Thanks for the info - I'm ok with treating the second one as non-breaking, but the first one is, right? and we have good migration guidance/opt-out options?

The main thing I want to avoid is this flying below the radar, but I think we're good on that given the discussion.we are having right now.

nvtw:
> Thanks for the info - I'm ok with treating the second one as non-breaking, but the first one is, right? and we have good migration guidance/opt-out options?
> 
> The main thing I want to avoid is this flying below the radar, but I think we're good on that given the discussion.we are having right now.

By default only edges connecting triangles with angles (between the triangle normals) smaller than 0.1 degrees get skipped. These edges are not contributing significant contacts - they are basically waste. And the second, heavier edge trimming is opt in, it is deactivated by default. So I would dare to classify this as non breaking. I can make it more clear that a negative value for the lower threshold angle leads to a complete opt out (so the irrelevant edges are kept). But users doing that basically waste cycles on these irrelevant edges.

But to be honest, it is theoretically possible that I did not think about all possible edge cases.

EDIT: The negative value trick is already in the changelog

adenzler-nvidia:
ok, but we can still call it a bug if we encounter something we didn't think about.

I would say for the first changelog entry, it would be great to state which values for those new options restore the old behavior though, even if it's functionally almost the same.

nvtw:
> ok, but we can still call it a bug if we encounter something we didn't think about.
> 
> I would say for the first changelog entry, it would be great to state which values for those new options restore the old behavior though, even if it's functionally almost the same.

Done

adenzler-nvidia (commented):
**Behavioral review findings**

- High: `newton/_src/geometry/types.py` (`clear_sdf`) - `clear_sdf()` leaves `_collision_edges` populated. Since `ModelBuilder.finalize()` prefers that cache over `mesh.edges`, a mesh that once built an SDF can keep simplified contact edges after the SDF is cleared, changing later mesh-mesh contact generation even on the no-SDF/BVH path. Clear the edge cache here too, or provide an explicit separate lifecycle.

- High: `newton/_src/geometry/types.py` (`Mesh.copy`) - `Mesh.copy()` copies `sdf` but not `_collision_edges`. A copied mesh with an attached SDF silently falls back to full edges, so original and copy can produce different contact counts and performance. Copy the collision-edge cache when copying the SDF-backed mesh.

- Medium: `newton/_src/geometry/types.py` (`build_sdf`) - `build_sdf()` assigns `self.sdf` before validating/applying edge options. If `_build_collision_edges()` raises, the public call fails but leaves the SDF attached, so a corrected retry hits `Mesh already has an SDF` unless the user manually clears it. Validate edge options before cooking, or roll back `self.sdf` on failure.

asv/benchmarks/setup/bench_sdf.py
```diff
@@ -11,6 +13,33 @@
...
+# runs this benchmark against an older base commit (where the kwarg does not
+# exist yet), forwarding it unconditionally raises ``TypeError`` from
+# ``setup`` and the entire benchmark is marked ``failed`` on the base side,
+# which in turn makes ``asv continuous`` exit non-zero on this PR.  Detect
+# the kwarg once at import time and only forward it where supported; on
+# older commits this just measures the SDF cook + the (then-absent) edge
+# pass, matching pre-change behaviour.
+_BUILD_SDF_SUPPORTS_EDGE_THRESHOLD = (
+    "edge_lower_angle_threshold_rad" in inspect.signature(newton.Mesh.build_sdf).parameters
+)
```
adenzler-nvidia, line 27: we could also just force the PR in and remove this, future runs will run against this one?

docs/concepts/collisions.rst
```diff
@@ -1493,6 +1493,13 @@ pipeline. When ``contacts`` is omitted from ``collide``, a buffer is allocated a
 Hydroelastic Contacts
 ---------------------
 
+.. warning::
+   Hydroelastic contact handling is **experimental**.  The configuration
+   surface (``ShapeConfig.is_hydroelastic``, ``ShapeConfig.kh``, and
+   :class:`~newton.geometry.HydroelasticSDF.Config`) and the numerical
+   behaviour may change without notice.  Validate accuracy and stability
+   on your scene before relying on it in production simulations.
```
adenzler-nvidia: Does it make sense to move this to a separate PR, along with the in-code tags? Feels like it's convoluting this PR

newton/_src/geometry/edge_redundancy.py
```diff
@@ -0,0 +1,787 @@
...
+    v0 = vertices[e[0]]
+    v1 = vertices[e[1]]
+
+    edge_vec = v1 - v0
+    edge_len = wp.length(edge_vec)
+    n = avg_normals[i]
+    n_len = wp.length(n)
+
+    # Degenerate edge or missing normal (NaN-filled) -> mark as no-box.
+    if edge_len <= 1.0e-12 or n_len <= 1.0e-12 or wp.isnan(n[0]):
```
adenzler-nvidia: can we use a minval constant here? we already have that defined somewhere, right?

newton/_src/geometry/edge_redundancy.py
```diff
@@ -0,0 +1,787 @@
...
+    rnor = box_axis_normal[i]
+
+    # World half-extents via |R| * h (R has axes as columns).
+    hx = wp.abs(rdir[0]) * h[0] + wp.abs(rtan[0]) * h[1] + wp.abs(rnor[0]) * h[2]
+    hy = wp.abs(rdir[1]) * h[0] + wp.abs(rtan[1]) * h[1] + wp.abs(rnor[1]) * h[2]
+    hz = wp.abs(rdir[2]) * h[0] + wp.abs(rtan[2]) * h[1] + wp.abs(rnor[2]) * h[2]
+    world_half = wp.vec3(hx, hy, hz)
+
+    aabb_lower[i] = c - world_half
+    aabb_upper[i] = c + world_half
```
adenzler-nvidia, line 183: we didn't have this function before?

adenzler-nvidia (commented):
Reviewed the latest head with the PR review toolkit, focusing on behavioral changes, simplification, duplication, and ease of understanding.

One non-inline docs note: docs/concepts/collisions.rst currently says the optional `Mesh.build_sdf()` kwargs have defaults shown, but the new edge-simplification kwargs and the opt-out are omitted. Since the default now changes the mesh-edge set used for SDF contact generation, this should be visible in the main SDF docs, not only in the changelog/docstring.

newton/_src/geometry/sdf_contact.py
```diff
@@ -550,6 +564,13 @@ def do_edge_sdf_collision_func(
         """
         golden = 0.3819660112501051  # (3 - sqrt(5)) / 2
         edge_dir = v1 - v0
+        edge_length = wp.length(edge_dir)
+
+        # Parametric tolerance floor: skip Brent for edges where the
+        # midpoint already meets ``precision_target``. ``+ 1e-12`` keeps
+        # zero-length edges from dividing by zero (they trivially meet
+        # any positive precision).
+        tol_floor = 0.5 * precision_target / (edge_length + 1.0e-12)
```
adenzler-nvidia: The adaptive Brent shortcut can drop short-edge contacts. When `edge_length <= 2 * contact_threshold_unscaled`, the first Brent iteration can break at the midpoint (`x = 0.5`), then the endpoint check below is skipped because `x` is not near a boundary. However the culling pass accepts edges with midpoint SDF up to `contact_threshold + radius`; in that accepted set, an endpoint can be inside the contact threshold while the midpoint is outside. The later strict check (`dist_approx < contact_threshold`) then discards a contact that the old fixed-search path would have found.

Could we add a short-edge endpoint regression, or still check endpoints when the tolerance floor causes this early exit?

newton/_src/geometry/types.py
```diff
@@ -714,6 +745,8 @@ def copy(
             m.com = self.com
             m.has_inertia = self.has_inertia
         m.sdf = self.sdf
+        if self._collision_edges is not None:
```
adenzler-nvidia: `Mesh.copy(vertices=..., indices=...)` now copies `_collision_edges` even when the copy replaces the geometry. This cache is indexed against the original topology and `ModelBuilder.finalize()` prefers it over `mesh.edges`, so a geometry-changing copy can carry stale or out-of-range edge indices into contact generation. I locally reproduced a 3-vertex copy retaining an original box edge like `[[20, 22]]`.

Please only copy `_collision_edges` when both `vertices` and `indices` are unchanged, or clear/recompute it when either override is supplied. The same stale-copy question may also apply to copying `sdf`, but this new cache copy makes the failure mode immediate for contacts.

newton/_src/geometry/types.py
```diff
@@ -795,15 +851,132 @@ def build_sdf(
             texture_format=texture_format,
             cache_dir=cache_dir,
         )
+
+        try:
```
adenzler-nvidia, line 880: The edge-option validation still happens after the expensive SDF cook. The rollback below fixes the stale `self.sdf`, but invalid edge options can still spend time cooking and can populate `cache_dir` before the public call raises.

Please validate `edge_lower_angle_threshold_rad`, abs/rel exclusivity, and half-extent ranges before `SDF.create_from_mesh()` so invalid edge options fail before doing work or writing cache entries.

newton/_src/geometry/sdf_texture.py
```diff
@@ -951,6 +1082,117 @@ def texture_sample_sdf_grad(
     return sdf_val, grad
 
 
+@wp.func
+def texture_sample_sdf_grad_hw(
```
adenzler-nvidia: Simplification: `texture_sample_sdf_grad_hw()` and `texture_sample_sdf_grad_only_hw()` duplicate the clamp/out-of-box handling and the six-sample finite-difference gradient. A shared private `wp.func` for the gradient-only part would make these two paths less likely to drift.

newton/_src/geometry/types.py
```diff
@@ -864,6 +1044,226 @@ def edges(self) -> np.ndarray:
             self._edges = orig_edges[first_idx]
         return self._edges
 
+    def _filter_edges_by_dihedral_angle(
```
adenzler-nvidia, line 1192: Simplification: `_filter_edges_by_dihedral_angle()` and `_compute_edge_dihedral_diagnostics()` repeat canonical edge slot construction, sorting, face normal computation, and pair diagnostics. Extracting shared topology/diagnostic preparation would make this easier to audit and reduce the risk of the two implementations drifting.

newton/_src/geometry/types.py
```diff
@@ -795,15 +851,132 @@ def build_sdf(
...
+            diagonal = 0.0
+
+        half_normal = _resolve_relative_or_absolute(
+            half_normal_abs, half_normal_rel, default_rel=1.0e-3, name="edge_box_half_normal", diagonal=diagonal
+        )
+        half_lateral = _resolve_relative_or_absolute(
+            half_lateral_abs, half_lateral_rel, default_rel=5.0e-3, name="edge_box_half_lateral", diagonal=diagonal
+        )
+
+        full_edges = self._filter_edges_by_dihedral_angle(lower_angle_threshold_rad)
```
adenzler-nvidia: Simplification/perf: with `edge_box_absorption=True`, `_build_collision_edges()` first computes `full_edges = _filter_edges_by_dihedral_angle(...)`, then `find_redundant_edges()` recomputes the same filtered edge set and diagnostics. It would be cleaner and cheaper to pass the already-computed diagnostics into the absorption path, or split the absorption helper so this preprocessing happens once.

newton/_src/geometry/sdf_texture.py
```diff
@@ -894,15 +923,117 @@ def texture_sample_sdf(
     return sdf_val + diff_mag
 
 
+@wp.func
+def texture_sample_sdf_hw(
```
adenzler-nvidia, line 944: Could we avoid adding a second cell-location implementation here? `texture_sample_sdf_hw()` repeats the same clamp/fine-cell/base-slot logic as `texture_sample_sdf()` and `_read_cell_corners()`, with only the final fetch/blend differing. A private `wp.func` that returns the clamped position/diff, fine cell indices/fractions, coarse base, and start slot would let the software and hardware samplers share the tricky boundary and `SLOT_LINEAR` handling. That seems worth doing here because future fixes to edge clamping/subgrid addressing otherwise have to be applied in multiple places.

newton/_src/geometry/edge_redundancy.py
```diff
@@ -0,0 +1,787 @@
...
+        aabb_upper[i] = wp.vec3(-1.0e30, -1.0e30, -1.0e30)
+        return
+
+    c = box_center[i]
+    h = box_half_extents[i]
+    rdir = box_axis_dir[i]
+    rtan = box_axis_tang[i]
+    rnor = box_axis_normal[i]
+
+    # World half-extents via |R| * h (R has axes as columns).
```
adenzler-nvidia: This OBB-to-AABB projection is another local implementation of the `|R| * half_extents` pattern already used in `sim/collide.py` and the heightfield overlap code. Can we move this to a shared private `wp.func` and call it here too? That would keep the axis/half-extent math in one place and reduce the chance that future margin/sign/rotation fixes diverge across broad phase, heightfields, and edge absorption.

nvtw:
@adenzler-nvidia most of your comments should be addressed

asv/benchmarks/setup/bench_sdf.py
nvtw: I kept it in for now - it's easy to remove it later in another MR (would even be fine as drive by fix since it is not a delicate change)

newton/_src/geometry/sdf_contact.py
```diff
@@ -550,6 +564,21 @@ def do_edge_sdf_collision_func(
...
+
+        # On very short edges ``tol_floor`` can be >= 0.5 so Brent breaks
+        # at the midpoint on iter 0 with ``x = 0.5``. The interior-only
+        # endpoint check below (``x < 0.2 or x > 0.8``) is then skipped,
+        # which can drop a vertex contact when an endpoint lies inside
+        # the contact threshold while the midpoint sits just outside.
+        # Track this case so we still evaluate both endpoints below.
+        force_endpoint_check = tol_floor >= 0.5
```
adenzler-nvidia: This still misses some short-edge endpoint contacts. The new fallback is gated on `tol_floor >= 0.5`, but the initial Brent stop condition at line 600 exits with zero iterations already when `tol_floor >= 0.25` (`abs(x - m) == 0`, `b - a == 1`). For edges where `precision_target < edge_length <= 2 * precision_target`, we can return the midpoint without contracting the bracket or checking either endpoint. If the midpoint is just outside the threshold while an endpoint is inside, the later `dist_approx < contact_threshold` check drops the contact entirely. Can we trigger the endpoint fallback whenever Brent exits before contracting the bracket, or at least use the actual zero-iteration threshold here?

adenzler-nvidia:
Several review threads look addressed by the latest commits but are still open in the GitHub UI (for example the `Mesh.copy()` stale `_collision_edges` issue, edge-option validation before SDF cook, and some duplicate-helper/refactor comments). Could you resolve the threads that are addressed so the remaining open items reflect only work still needed?

adenzler-nvidia (dismissed):
Reviewed latest head 10b5ba7. No blocking issues from my pass.

newton/_src/geometry/types.py
```diff
@@ -864,6 +1099,237 @@ def edges(self) -> np.ndarray:
...
+        angles_pair = np.where(valid, np.arccos(cos_ab), np.nan)
+        with np.errstate(invalid="ignore", divide="ignore"):
+            unit_a = n_a / np.where(norm_a[:, None] > 0.0, norm_a[:, None], 1.0)
+            unit_b = n_b / np.where(norm_b[:, None] > 0.0, norm_b[:, None], 1.0)
+        avg = unit_a + unit_b
+        avg_norm = np.linalg.norm(avg, axis=1)
+        with np.errstate(invalid="ignore", divide="ignore"):
+            unit_avg = avg / np.where(avg_norm[:, None] > 0.0, avg_norm[:, None], 1.0)
+        # Opposing normals (avg ~= 0) or degenerate triangle -> NaN.
+        unit_avg[(~valid) | (avg_norm == 0.0)] = np.nan
```
adenzler-nvidia: The `avg_norm == 0.0` exact float-equality check misses near-antiparallel adjacent faces.

For two adjacent triangles whose normals are *almost* but not exactly opposite (a tiny perturbation away from 180°), `avg_norm` can be e.g. `1e-12` — not zero. Under the `np.errstate` block on line 1182, the divide on line 1183 goes through and produces a `unit_avg` of magnitude ~1e12.

That value flows into `_build_collision_edges` → `_build_edge_box_kernel` as the edge's `avg_normal`. The kernel's degeneracy guard at `edge_redundancy.py:143-145` checks `n_len = wp.length(n)` against `MINVAL = 1e-15` — `n_len ≈ 1e12` passes — and the kernel constructs a perfectly-valid-looking but geometrically meaningless box. That edge can then silently absorb or be absorbed by the wrong neighbours.

Cleanest fix: drop the NaN sentinel for this case and collapse opposing-normal / degenerate pairs to a zero-length `avg_normal`, which the existing Warp guard already treats as "no valid box":

```python
# Opposing normals or degenerate triangle -> zero-length avg_normal, which
# _build_edge_box_kernel's `n_len <= MINVAL` guard treats as "no valid box".
# Epsilon (not exact zero) so near-antiparallel pairs don't slip through.
EPS = 1.0e-6
safe = valid & (avg_norm > EPS)
denom = np.where(safe, avg_norm, 1.0)[:, None]
unit_avg = np.where(safe[:, None], avg / denom, 0.0)
```

This also removes the `np.errstate` / `unit_avg[...] = np.nan` two-step. The non-pair "no data" NaN sentinel in `_compute_edge_dihedral_diagnostics` is unaffected — it still legitimately marks "this edge has no adjacent pair to report a normal for", which is distinct from "this pair has a bad normal."

nvtw: Done

adenzler-nvidia (approved):
LGTM. The avg_norm epsilon fix in 346dbba3 addresses the last remaining concern from my pass.

Issue #2362, opened 2026-04-08
Resolve warnings from some examples


Kenny-Vilella:
It is not critical but it could be good to resolve all the warnings in the examples.
Below a list of all the warnings I noted:

`basic_conveyor`:
```
/home/horde/.venv/lib/python3.13/site-packages/newton/examples/basic/example_basic_conveyor.py:326: UserWarning: Inertia validation corrected 1 bodies. Set validate_inertia_detailed=True for detailed per-body warnings.
  self.model = builder.finalize()
```

`cloth_h1` and `cloth_style3d`
```
/home/horde/.venv/lib/python3.13/site-packages/newton/_src/usd/utils.py:1601: UserWarning: Metallic texture inputs are not yet supported; using scalar fallback.
  properties = _extract_preview_surface_properties(shader, prim)
/home/horde/.venv/lib/python3.13/site-packages/newton/_src/usd/utils.py:1601: UserWarning: Roughness texture inputs are not yet supported; using scalar fallback.
  properties = _extract_preview_surface_properties(shader, prim)
```

`softbody_gift` and `softbody_franka`:
```
/home/horde/.venv/lib/python3.13/site-packages/newton/_src/utils/mesh.py:231: UserWarning: Detected non-manifold edge
  self.add_edge(tri[0], tri[1], tri[2], index)
/home/horde/.venv/lib/python3.13/site-packages/newton/_src/utils/mesh.py:233: UserWarning: Detected non-manifold edge
  self.add_edge(tri[2], tri[0], tri[1], index)
/home/horde/.venv/lib/python3.13/site-packages/newton/_src/utils/mesh.py:232: UserWarning: Detected non-manifold edge
  self.add_edge(tri[1], tri[2], tri[0], index)
```

`robot_allegro_hand`:
```
Warning: The rigid body at /World/envs/env_0/object/DexCube has a possibly invalid inertia tensor of {1.0, 1.0, 1.0}, small sphere approximated inertia was used. Either specify correct values in the mass properties, or add collider(s) to any shape(s) that you wish to automatically compute mass properties for. If you do not want the objects to collide, add colliders regardless then disable the 'enable collision' property. [python3.13]
/home/horde/.venv/lib/python3.13/site-packages/newton/_src/sim/builder.py:2412: UserWarning: Body /World/envs/env_0/object/DexCube: authored mass and density without authored diagonalInertia. Ignoring body-level density.
  return parse_usd(
```

`robot_ur10`:
```
[...]

PR #2897, opened 2026-05-18
Fix basic_conveyor inertia validation warning

### LINKED ISSUE #2635: basic_conveyor: resolve inertia validation warning
Sub-task of #2362.

`example_basic_conveyor` emits the following warning at `builder.finalize()`:

```
example_basic_conveyor.py:326: UserWarning: Inertia validation corrected 1 bodies. Set validate_inertia_detailed=True for detailed per-body warnings.
```

### Source
The warning is emitted by the inertia validation kernel inside `ModelBuilder.finalize()` (`newton/_src/sim/builder.py`). The triggering body is the kinematic conveyor belt body created in `example_basic_conveyor.py` (around line 237), which has `is_kinematic=True` and a mass of 15 kg but no authored inertia tensor.

### Fix
Author a valid inertia tensor explicitly on the kinematic belt body in the example.

Do **not** silence the validator for kinematic bodies: kinematic bodies can be switched to dynamic at runtime, so a valid inertia must always be available.

### Verification
- Run `uv run -m newton.examples basic_conveyor` and confirm no `UserWarning` from the inertia validator.
- Run the example test (`uv run --extra dev -m newton.tests -k test_basic.example_basic_conveyor` or equivalent) and confirm it still passes.

vreutskyy:
## Description

The `basic_conveyor` example emits a spurious warning at `builder.finalize()`:

```
UserWarning: Inertia validation corrected 1 bodies. Set validate_inertia_detailed=True for detailed per-body warnings.
```

With detailed validation enabled, the warning resolves to:

```
UserWarning: Inertia matrix for body 'conveyor_belt' is not symmetric, making it symmetric
```

The kinematic conveyor belt was created with `mass=15.0` and no authored inertia; the mesh-shape integrator then accumulated a tensor with sub-tolerance asymmetry onto the body, which the validator caught at finalize.

Fix: author an analytical annular-ring inertia tensor on `add_link()` (about the belt's COM, ring axis along Z) and set `density=0.0` on `belt_cfg` so the mesh shape no longer accumulates mass/inertia. The body now carries a clean, symmetric inertia, and the authored mass is exactly 15 kg instead of `15 + mesh_volume * density`.

Closes #2635 (sub-task of epic #2362).

## Checklist

## Test plan

- `uv run --extra examples -m newton.examples basic_conveyor --viewer null --num-frames 5 --test` — exits 0, no `Inertia validation` warning. Re-run with `builder.validate_inertia_detailed = True` patched in locally — also silent (no per-body warning).
- GPU smoke test: `uv run --extra examples -m newton.examples basic_conveyor` on RTX 3080 — belt spins, bags rest on it, no warnings.
- `uvx pre-commit run --files newton/examples/basic/example_basic_conveyor.py CHANGELOG.md` — passes.

## Bug fix

1. `uv run --extra examples -m newton.examples basic_conveyor`
2. Observe `UserWarning: Inertia validation corrected 1 bodies.` in stderr at startup.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed spurious inertia validation warning emitted during finalization of the basic conveyor example.

* **Tests**
  * Added an example test that runs the conveyor demo with the viewer for extended frames to ensure it remains warning-free.

* **Documentation**
  * Updated changelog with an entry noting the fix for the basic conveyor example.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2897?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 ### Fixed
 
 - Fix `SolverMuJoCo` returning `State.joint_qd` in world frame for root `FREE` joints with non-identity `parent_xform`, violating the documented parent-frame contract and corrupting derived `body_qd`.
+- Fix `basic_conveyor` example emitting a spurious inertia validation warning at finalize.
 
 ## [1.2.0] - 2026-05-12
 
--- a/newton/examples/basic/example_basic_conveyor.py
+++ b/newton/examples/basic/example_basic_conveyor.py
@@ -182,6 +182,7 @@ class Example:
         )
 
         belt_cfg = newton.ModelBuilder.ShapeConfig(
+            density=0.0,  # mass and inertia are authored explicitly on the belt body below
             mu=1.2,
             ke=1.0e5,  # vbd only
             kd=0.0,  # vbd only
@@ -234,8 +235,24 @@ class Example:
             metallic=0.9,
         )
 
+        # Annular-ring inertia about the belt's COM (ring axis along Z).
+        belt_mass = 15.0
+        belt_radii_sum_sq = belt_inner_radius**2 + belt_outer_radius**2
+        belt_i_transverse = belt_mass / 12.0 * (3.0 * belt_radii_sum_sq + (2.0 * BELT_HALF_THICKNESS) ** 2)
+        belt_i_axial = 0.5 * belt_mass * belt_radii_sum_sq
         self.belt_body = builder.add_link(
-            mass=15.0,
+            mass=belt_mass,
+            inertia=wp.mat33(
+                belt_i_transverse,
+                0.0,
+                0.0,
+                0.0,
+                belt_i_transverse,
+                0.0,
+                0.0,
+                0.0,
+                belt_i_axial,
+            ),
             is_kinematic=True,
             label="conveyor_belt",
         )
--- a/newton/tests/test_examples.py
+++ b/newton/tests/test_examples.py
@@ -251,6 +251,14 @@ add_example_test(
     test_options={"num-frames": 150},
 )
 
+add_example_test(
+    TestBasicExamples,
+    name="basic.example_basic_conveyor",
+    devices=test_devices,
+    use_viewer=True,
+    test_options={"num-frames": 100},
+)
+
 
 class TestCableExamples(unittest.TestCase):
     pass


adenzler-nvidia:
Please register `basic.example_basic_conveyor` in `newton/tests/test_examples.py`. The example already has `test_final()` coverage, but it is not currently run by the example test suite, so this regression would not be caught by CI. Consider enabling output checking for this entry as well so the inertia warning fails the test.

Kenny-Vilella:
Except @adenzler-nvidia 's comment, everything is fine and could be merged.

vreutskyy:
@adenzler-nvidia — I registered `basic.example_basic_conveyor` locally and tried turning on the existing output-checking by flipping `check_output=False` → `True` in `add_example_test` (the hardcoded value at [test_examples.py:194](https://github.com/newton-physics/newton/blob/main/newton/tests/test_examples.py#L194), which currently disables it for every example).

The test fails, but not on the inertia warning — it fails on the OpenUSD banner that the test runner emits when [unittest_parallel.py:26](https://github.com/newton-physics/newton/blob/main/newton/tests/thirdparty/unittest_parallel.py#L26) sets `PXR_WORK_THREAD_LIMIT=1`:

```
AssertionError: Unexpected output:
'##################################################################
#  PXR_WORK_THREAD_LIMIT is overridden to '1'.  Default is '0'.  #
##################################################################'
```

The subprocess inherits that env var, libusd re-emits the 3-line banner on stderr, the example-test runner does `print(result.stderr)` ([test_examples.py:176-177](https://github.com/newton-physics/newton/blob/main/newton/tests/test_examples.py#L176-L177)), and `CheckOutput`'s only filter is for `"Module xxx load on device"` lines ([unittest_utils.py:216-218](https://github.com/newton-physics/newton/blob/main/newton/tests/unittest_utils.py#L216-L218)), so the banner trips it. Any other example test would hit the same thing if we turned this on.

Options I can think of:

1. Extend `CheckOutput`'s filter to also skip the PXR banner block.
2. Pop `PXR_WORK_THREAD_LIMIT` from the subprocess env in `add_example_test` so the banner never fires (works fine for examples that don't load USD; could re-introduce the documented OpenUSD thread-safety crash for USD-loading examples).
3. Some other mechanism I'm missing.

Do you have a preferred direction? Happy to take this further but want to confirm the approach before changing shared test infra.

adenzler-nvidia:
I'd suggest to just forget about the output testing for now - https://github.com/newton-physics/newton/pull/2874 this base class should be merged soon and then we can move the tests over to this. That way, we can catch all output properly.

vreutskyy:
Thanks @adenzler-nvidia — done in 83a077fb. Registered `basic.example_basic_conveyor` as a plain smoke-test (no `check_output`); we'll wire up proper output checking after [#2874](https://github.com/newton-physics/newton/pull/2874) lands. Locally 2/2 pass on CPU + CUDA.

adenzler-nvidia (approved):

PR #2898, opened 2026-05-18
Document loop joints / excludeFromArticulation in concepts/articulations

### LINKED ISSUE #2650: Document loop joints / `excludeFromArticulation` in concepts/articulations
Loop joints (joints with `joint_articulation == -1`) are a cross-cutting concept — `SolverMuJoCo` uses them to synthesise equality constraints (see `docs/integrations/mujoco.rst` "Loop closures") — but the broader docs don't cover them. Two surfaces produce a loop joint:

- **USD**: `PhysicsJoint.excludeFromArticulation`, handled in `newton/_src/utils/import_usd.py:1899`
- **Manual builder**: any joint not included in a `ModelBuilder.add_articulation([...])` call (default at `newton/_src/sim/builder.py:3790`)

Neither `docs/concepts/articulations.rst` nor `docs/concepts/usd_parsing.rst` mentions either path.

**Ask:** add a short "Loop joints" subsection to `docs/concepts/articulations.rst` covering both surfaces, then cross-link from `docs/integrations/mujoco.rst` (the "Loop closures" paragraph under `Equality constraints`) and any other solver doc that synthesises equality constraints from `joint_articulation == -1`.

gyeomannvidia:
## Description

Document loop closure joints and equality joints in articulations.rst

Closes https://github.com/newton-physics/newton/issues/2650

## Checklist

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive "Loop closure" section distinguishing orphan vs loop-closing joints, with an end-to-end example and USD import behavior.
  * Clarified solver-specific behavior and limitations (maximal- vs generalized-coordinate differences, MuJoCo handling, unsupported joint types/features, and excluded DOFs).
  * Noted that loop-closing joints are omitted from FK/IK/ArticulationView traversal.
  * Added stable cross-link anchors and updated changelog.


adenzler-nvidia:
A lot of the MuJoCo-specific equality/loop-closure behavior here already lives in `docs/integrations/mujoco.rst` under "Equality constraints" / "Loop closures". Could we keep this concepts page focused on the general modeling pattern and link out to the MuJoCo section for solver-specific mappings and limitations?

The missing piece seems to be the user-facing explanation of loop joints / `excludeFromArticulation`, not duplicating the MuJoCo equality-constraint details.

preist-nvidia:
@jcarius-nv for an additional review.

gyeomannvidia:
I think we have this the wrong way round.  mujoco.rst should only discuss specifics of the mujoco implementation eg the mapping between newton equality joint type and mujoco equality joint type; and the limited way that we implement orphan joints.  The general points should be kept in articulations.rst with links to specifics discussed in mujoco.rst.

I didn't want to start removing text from mujoco.rst but maybe that is what we should do.

adenzler-nvidia:
Yes I think this is the right way to go - note the "newton equality constraints" are going to be deprecated and remain as mujoco-specific attributes, as discussed earlier this month.

gyeomannvidia:
Thanks for the review.  I'll have another pass at this on Monday.

docs/concepts/articulations.rst
```diff
@@ -712,3 +712,242 @@ To proceed with orphan joints, skip this validation:
...
+form a closed kinematic loop (for example a four-bar linkage or a parallel
+mechanism). Closed loops must instead be expressed by declaring the topology
+as a tree and adding a separate constraint that re-couples the open end.
+
+There are two ways to add a loop-closing constraint. The first is to call
+one of the typed equality-constraint methods —
+:meth:`~newton.ModelBuilder.add_equality_constraint_connect`,
+:meth:`~newton.ModelBuilder.add_equality_constraint_weld`, or
+:meth:`~newton.ModelBuilder.add_equality_constraint_joint` (or the generic
+:meth:`~newton.ModelBuilder.add_equality_constraint`) — which introduces the
```
adenzler-nvidia: https://github.com/newton-physics/newton/pull/2961 deprecates these functions - so I don't think we should mention this in the docs.

They are a mujoco-specific extension in the future, so it's still possible, but maybe should move to the mujoco-specific docs?

gyeomannvidia: Will that make it in for 1.3?

docs/concepts/articulations.rst
```diff
@@ -712,3 +712,114 @@ To proceed with orphan joints, skip this validation:
...
+Newton's :meth:`~newton.ModelBuilder.add_joint_*` methods author **kinematic
+trees**: each body has at most one parent joint, so the joints alone cannot
+form a closed kinematic loop (for example a four-bar linkage or a parallel
+mechanism). Closed loops must instead be expressed by declaring the topology
+as a tree and adding a separate joint that re-couples the open end.
+
+To close a loop, create the loop-closing joint with
+:meth:`~newton.ModelBuilder.add_joint_*` but **omit it from the
+``joint_list`` passed to** :meth:`~newton.ModelBuilder.add_articulation`,
+so the articulation graph remains a tree and the closing joint becomes an
```
adenzler-nvidia: A loop-closing joint isn't an orphan joint in Newton's validation sense — `_validate_joints` explicitly exempts joints whose child is already reachable through the tree (its docstring even names them "loop joints"). Equating the two is what leads to the `skip_validation_joints` claim below. Suggest distinguishing them:

```suggestion
so the articulation graph remains a tree. The closing joint stays outside
any articulation as a **loop-closing joint** (its two bodies are both
already in the tree, which distinguishes it from an
:ref:`orphan joint <Orphan joints>`, whose child has no articulated path):
```

adenzler-nvidia (commented):
Thanks for documenting this — the per-solver breakdown is accurate and I verified the MuJoCo equality-constraint behavior (FIXED→6, REVOLUTE→5 via two `mjEQ_CONNECT`, BALL→3), the Featherstone "silently not enforced" claim, the ignored drive/limits/armature/friction, and the USD `excludeFromArticulation` import path all against the source. Those are correct.

One conceptual issue runs through the new section: it treats a **loop-closing joint** as an **orphan joint**, but Newton's `finalize()` validation deliberately distinguishes them (`ModelBuilder._validate_joints`). An *orphan joint* is a non-articulated joint whose child body is unreachable through any tree; a *loop-closing joint* is non-articulated but its child is *already* in a tree. Only the former trips validation. This affects three places (inline) — most importantly the claim that you must pass `skip_validation_joints=True`, which the section's own runnable example contradicts by calling `finalize()` with no flag. One D6 detail is also slightly off.

docs/concepts/articulations.rst
```diff
@@ -712,3 +712,114 @@ To proceed with orphan joints, skip this validation:
...
+leave it out of the surrounding :meth:`~newton.ModelBuilder.add_articulation`
+call — producing exactly the orphan-joint topology shown above. This is how
+a USD asset can author a four-bar linkage or other parallel mechanism.
+
+.. note::
+
+   Omitting a joint from :meth:`~newton.ModelBuilder.add_articulation` is
+   accepted by every Newton solver (after the default ``finalize()``
+   orphan-joint check is disabled with ``skip_validation_joints=True``), but
+   each solver handles the resulting orphan joint differently:
```
adenzler-nvidia: This says the topology is only accepted after disabling the check with `skip_validation_joints=True`, but a genuine loop-closing joint passes `finalize()` by default — its child is already reachable (`_validate_joints`, builder.py:9244-9250). The runnable example just above calls `builder.finalize()` with no flag and works, so the note contradicts it. `skip_validation_joints=True` is only needed for *true* orphans (unreachable child). Suggest:

```suggestion
   A loop-closing joint passes :meth:`~newton.ModelBuilder.finalize`
   validation by default: unlike a true orphan joint, its child body is
   already reachable through the tree, so ``skip_validation_joints=True`` is
   not required. Each solver handles the loop-closing joint differently:
```

Follow-through: the rest of this note still says "the orphan joint" / "the resulting orphan joint" in a couple of spots — those should become "the loop-closing joint" for consistency.

docs/concepts/articulations.rst
```diff
@@ -712,3 +712,114 @@ To proceed with orphan joints, skip this validation:
 
 Only maximal-coordinate solvers (:class:`~newton.solvers.SolverXPBD`, :class:`~newton.solvers.SolverSemiImplicit`) support orphan joints.
 Generalized-coordinate solvers (:class:`~newton.solvers.SolverFeatherstone`, :class:`~newton.solvers.SolverMuJoCo`) require every joint to belong to an articulation.
```
adenzler-nvidia, line 731: This statement is correct for *orphan* joints (a body with no articulated path has no generalized coordinates, so Featherstone/MuJoCo can't simulate it). But once the new section refers to loop-closing joints as "orphan joints" and states MuJoCo handles them via equality constraints, this line reads as a direct contradiction. After the terminology fix above, a pointer here would help:

```suggestion
Generalized-coordinate solvers (:class:`~newton.solvers.SolverFeatherstone`, :class:`~newton.solvers.SolverMuJoCo`) require every joint to belong to an articulation.
(Loop-closing joints, whose bodies are both already in a tree, are a separate case — see :ref:`Loop closure`; :class:`~newton.solvers.SolverMuJoCo` supports these via equality constraints.)
```

docs/integrations/mujoco.rst
```diff
@@ -199,21 +201,35 @@ array; slot layout depends on the constraint type.
...
+- :attr:`~newton.JointType.REVOLUTE` — 5 DOFs constrained; one rotational
+  DOF about the hinge axis remains free.
+- :attr:`~newton.JointType.BALL` — the 3 translational DOFs constrained;
+  all 3 rotational DOFs remain free.
+
+Other joint types used as loop closures
+(:attr:`~newton.JointType.PRISMATIC`, :attr:`~newton.JointType.FREE`,
+:attr:`~newton.JointType.DISTANCE`, :attr:`~newton.JointType.D6`,
+:attr:`~newton.JointType.CABLE`) emit a warning and are silently skipped —
+the loop is *not* closed.
```
adenzler-nvidia: The loop-closure synthesis dispatches on DOF dimensionality `(lin_count, ang_count)`, not `JointType` (`solver_mujoco.py:5300,5338`), so a `D6` isn't unconditionally skipped: one angular axis → revolute closure, three → ball. The old text ("other joint types … produce a warning") was vaguer but didn't have this issue. Suggest:

```suggestion
Other joint types used as loop closures
(:attr:`~newton.JointType.PRISMATIC`, :attr:`~newton.JointType.FREE`,
:attr:`~newton.JointType.DISTANCE`, :attr:`~newton.JointType.CABLE`) emit a
warning and are silently skipped — the loop is *not* closed. A
:attr:`~newton.JointType.D6` is dispatched by its degrees of freedom: one
angular axis behaves as a revolute closure and three as a ball closure;
any other configuration is skipped.
```

docs/concepts/articulations.rst
gyeomannvidia: I'm very glad you caught that.

docs/concepts/articulations.rst
```diff
@@ -728,3 +729,119 @@ To proceed with orphan joints, skip this validation:
...
+(Loop-closing joints are not orphan joints and are handled separately — see :ref:`Loop closure`.)
+
+.. _Loop closure:
+
+Loop closure
+------------
+
+Newton's :meth:`~newton.ModelBuilder.add_joint_*` methods author **kinematic
```
adenzler-nvidia, line 739: `:meth:`~newton.ModelBuilder.add_joint_*`` (here and at line 746) is used as a Sphinx cross-reference role, but `add_joint_*` is not a real method, so this is a dangling xref. It won't fail the `-W` build (`nitpicky = False`), but it renders as a non-clickable code span rather than a link. Every other wildcard method family in the docs uses a double-backtick literal — and this same section already does so for `add_joint_*()` further down (e.g. in the USD paragraph). Suggest using ``` ``add_joint_*()`` ``` here and at line 746 for consistency.

adenzler-nvidia (approved):
LGTM

PR #2901, opened 2026-05-18
Add VBD volumetric Neo-Hookean correctness tests

AnkaChan:
## Description

Add 6 volumetric (tetrahedral soft-body) correctness tests for the Stable Neo-Hookean material in the VBD solver.

**Tests added:**

| # | Test | What it checks |
|---|------|----------------|
| 1 | Beam extension | Gravity-loaded hanging beam; tip displacement ±15% of empirical baseline |
| 2 | Uniaxial stretch | Beam stretched to 2×; volume preservation \|V/V₀ − 1\| ≤ 10% |
| 3 | 360° twist | Full-rotation twist; no NaN + volume preservation |
| 4 | Compression + recovery | Cube compressed to 50% height then released; recovery > 50% |
| 5 | Convergence refinement | Same extension at 3 mesh resolutions; max/min displacement < 3× |
| 6 | Sliver elements | 10:1 aspect-ratio tets; no NaN, bounded velocities |

**Quantitative report with videos:** https://ankachan.github.io/reports/vbd-volumetric-neohookean.html

### Screenshots

**Beam extension (gravity hanging)**

extension

**Uniaxial stretch (2× along X)**

stretch

**360° twist**

twist

**Compression + recovery**

compression

## Checklist

## Test plan

```bash
# Run all volumetric tests
uv run --extra dev -m newton.tests -k test_VBD.Volumetric

# Run individual tests
uv run --extra dev -m newton.tests -k test_VBD.Volumetric.example_soft_beam_extension
uv run --extra dev -m newton.tests -k test_VBD.Volumetric.example_soft_beam_stretch
uv run --extra dev -m newton.tests -k test_VBD.Volumetric.example_soft_beam_twist
uv run --extra dev -m newton.tests -k test_VBD.Volumetric.example_soft_cube_compression
uv run --extra dev -m newton.tests -k test_VBD.Volumetric.example_soft_convergence_refinement
uv run --extra dev -m newton.tests -k test_VBD.Volumetric.example_soft_sliver_elements
```

All 6 tests pass on L40 GPU. Each test's `test_final()` validates regression metrics (volume preservation, displacement bounds, stability).

## Summary by CodeRabbit

* **New Features**
  * Added runnable volumetric soft-body examples for beam extension, stretching, twisting, cube compression, mesh refinement, and sliver elements.
  * Added runtime checks for stability, volume preservation, deformation accuracy, mesh quality, and recovery behavior.
  * Added viewer loading-splash controls.

* **Documentation**
  * Documented mesh validation options and public volumetric mesh validation utilities.

* **Tests**
  * Added automated coverage for the new volumetric simulation examples on supported devices.


newton/examples/vbd/example_soft_beam_extension.py
```diff
@@ -0,0 +1,163 @@
...
+    DIM_Z = 20
+    CELL = 0.05
+    DENSITY = 1000.0
+    K_MU = 5.0e4
+    K_LAMBDA = 5.0e4
+    K_DAMP = 0.1
+
+    # Empirical equilibrium extension of the bottom layer (m), calibrated on
+    # L40 with VBD iterations=20, substeps=5, 300 frames.
+    EXPECTED_DELTA = 0.10
```
jumyungc: If we compute expected delta from material parameters, would it still work?

newton/examples/vbd/example_soft_cube_compression.py
```diff
@@ -0,0 +1,187 @@
...
+
+###########################################################################
+# Example Soft Cube Compression — Extreme Compression + Recovery
+#
+# A tetrahedral cube with its bottom face pinned. The top face is driven
+# downward to 50% of the rest height over 150 frames, then released.
+# The stable Neo-Hookean material should survive the compression
+# without NaN or inversion, and recover toward the rest height after
+# release.
+#
```
jumyungc, line 12: Could we make this example exercise robustness even more directly? e.g, see near flat or inverted cubes come back?

CHANGELOG.md
```diff
@@ -36,6 +36,11 @@
...
+- Add `cloth_stiff_material_hanging` example: stiff square cloth (`tri_ke = tri_ka = 1e8`) pinned along one edge under gravity, exercising the new Neo-Hookean triangle material at parameters where the previous StVK form would explode. `test_final` checks for non-explosion via a per-particle velocity bound in addition to the existing position-volume checks
+- Add `cloth_stiff_material_stretch` example: five cloth sheets pinned at +/-X edges and stretched, each with a different effective Poisson ratio set via `tri_ka / tri_ke`. `test_final` checks bulk area against the closed-form StableNH 2D-membrane prediction within 10%, regression-guarding the new triangle material model
+- Add `cloth_stiff_material_hanging` and `cloth_stiff_material_stretch` examples regression-guarding the new Neo-Hookean triangle material (stability under gravity at extreme stiffness, and bulk area-preservation across a Poisson-ratio sweep)
```
jumyungc: Some changelog entries seem unrelated.

jumyungc:
Thank you for the correctness test.
For volume tests, would it make sense to use a simple scene with small deformation and verify volume changes with varying poisson ratios?

CHANGELOG.md
```diff
@@ -36,6 +36,11 @@
...
+- Add `soft_beam_extension`, `soft_beam_stretch`, `soft_beam_twist`, `soft_cube_compression`, `soft_convergence_refinement`, and `soft_sliver_elements` volumetric examples regression-guarding Neo-Hookean tet-mesh elasticity (extension under gravity, volume preservation under stretch and twist, recovery from extreme compression, mesh-convergence sanity, and sliver-element robustness)
```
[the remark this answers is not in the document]
AnkaChan: Removed these entries rather than rewording: they describe cloth examples (`cloth_stiff_material_*`) and a `validate_mesh` line that are not part of this PR (the cloth files do not exist on this branch, and this PR registers volumetric tests only). The CHANGELOG now keeps only the volumetric entry. (ec030753)

newton/examples/vbd/example_soft_beam_twist.py
```diff
@@ -0,0 +1,186 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
+
+###########################################################################
+# Example Soft Beam Twist — 180° Twist Stability
+#
+# A vertical tetrahedral beam with its bottom face pinned and top face
+# twisted 180° around the Z-axis over a linear ramp. The stable
+# Neo-Hookean material handles this extreme rotation without inversion
+# (unlike StVK). The test validates volume preservation and stability.
```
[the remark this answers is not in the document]
AnkaChan: Updated the banner to 360° to match `TWIST_ANGLE = 2.0 * np.pi`. (31ab1cfc)

newton/examples/vbd/example_soft_convergence_refinement.py
```diff
@@ -0,0 +1,202 @@
...
+        with wp.ScopedCapture() as capture:
+            for _ in range(5):
+                s0.clear_forces()
+                model.collide(s0, contacts)
+                solver.step(s0, s1, ctrl, contacts, dt)
+                s0, s1 = s1, s0
+        graph = capture.graph
+        for _ in range(n_frames):
+            wp.capture_launch(graph)
+    else:
```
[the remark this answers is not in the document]
AnkaChan: Fixed by using an even substep count (6) in the captured examples (this one, `beam_extension`, `sliver_elements`) so the graph replays from the correct start buffer instead of the stale parity. Separately reworked this test to run each resolution to convergence with iterations scaled to resolution (`iters = 5*dim_z`) and to assert monotone convergence toward the continuum solution `delta = rho*g*L^2/(2E)` plus shrinking refinement steps. At the old fixed 20 iterations the fine mesh was badly under-converged (delta ~0.24 m vs ~0.056 m converged), which had masked the parity issue. (31ab1cfc, a513d446)

newton/examples/vbd/example_soft_cube_compression.py
```diff
@@ -0,0 +1,187 @@
...
+        for _ in range(self.sim_substeps):
+            self.state_0.clear_forces()
+            self.viewer.apply_forces(self.state_0)
+            self.model.collide(self.state_0, self.contacts)
+            self.solver.step(self.state_0, self.state_1, self.control, self.contacts, self.sim_dt)
+            self.state_0, self.state_1 = self.state_1, self.state_0
+
+    def _apply_compression(self):
+        if self._frame_index >= self.RELEASE_FRAME:
+            if not self._released:
```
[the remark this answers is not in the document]
AnkaChan: Fixed with `denom = max(COMPRESS_FRAMES-1, 1); t = min(idx/denom, 1.0)` so the ramp reaches exactly the target ratio at the last compressed frame. Re-ran the test; the recovery/volume thresholds still hold. (32db91fd)

newton/examples/vbd/example_soft_beam_extension.py
AnkaChan: Done. The expected tip extension is now derived from the material parameters: the linear-elastic self-weight solution `delta = M*g*L/(2*A*E)` with `E = mu(3*lambda+2*mu)/(lambda+mu)`. The stable Neo-Hookean material is calibrated to the linear Lame parameters (`mu_nh = mu`, `lambda_nh = lambda + mu`; Smith et al. 2018), so that E is the correct target. The mass `M` is read from the model because `add_soft_grid` lumps a full cell mass onto every grid vertex. Measured matches the analytical value to ~4% (solver iterations raised to 100 so the stiff beam converges to its static equilibrium). (6aaa7eae)

newton/examples/vbd/example_soft_cube_compression.py
AnkaChan: Added a near-flat (90%) compression case that inverts elements, then checks the cube springs back to its rest height and volume once released, exercising inversion recovery directly. The stable Neo-Hookean material recovers fully (recovery ~1.0, no NaN) even at 95% squash; the test asserts recovery >= 0.90 of both height and volume. (32db91fd)

CHANGELOG.md
AnkaChan: Removed them. The `validate_mesh` line duplicated an existing entry, and the three `cloth_stiff_material_*` lines describe cloth examples that are not part of this PR. The CHANGELOG now keeps only the volumetric entry. (ec030753)

AnkaChan:
Addressed the remaining CodeRabbit nitpicks (compute contacts once per frame): `model.collide()` is now called once per frame and reused across substeps in all six volumetric examples.

Summary of this review pass:
- **CHANGELOG**: removed four unrelated/duplicate entries (a `validate_mesh` duplicate + three `cloth_stiff_material_*` lines for examples not in this PR); kept only the volumetric entry.
- **CUDA-graph parity**: use an even substep count (6) in the three examples that capture a graph (`beam_extension`, `sliver_elements`, `convergence_refinement`) so the state ping-pong replays from the correct start buffer.
- **beam_extension**: the expected tip displacement is now derived analytically from the material parameters (`delta = M*g*L/(2*A*E)`, `E = mu(3*lambda+2*mu)/(lambda+mu)`), matching the simulation to ~4%; iterations raised to 100 so the stiff beam converges.
- **cube_compression**: fixed the ramp off-by-one; added a near-flat (90%) compression case that inverts elements and verifies recovery.
- **convergence_refinement**: reworked to solve each resolution to convergence (iterations scaled to resolution) and assert monotone convergence to the continuum solution with shrinking refinement steps.

All six example tests pass on an L40.

AnkaChan:
@jumyungc I address the all the issues except the varying poison ratio one. Can we defer that to a future PR?

newton/examples/vbd/example_soft_beam_twist.py
```diff
@@ -0,0 +1,186 @@
...
+        )
+
+        q = self.state_0.particle_q.numpy()
+
+        # No NaN
+        if np.any(np.isnan(q)):
+            raise ValueError("NaN detected in particle positions")
+
+        # Volume preservation
+        final_volume = _compute_tet_volume(q, self.tet_indices)
```
eric-heiden: 🟡 This final volume check runs after the driven face has returned to the rest orientation. `TWIST_ANGLE` is `2 * np.pi`, so once `_frame_index >= RAMP_FRAMES`, `_apply_twist_ramp()` places the top face at the same X/Y coordinates as the rest pose. The registered test runs 400 frames with a 200-frame ramp, so `test_final()` mostly validates the relaxed boundary condition rather than the claimed extreme twist. Consider sampling volume and signed-tet validity during the ramp, or use a non-periodic final angle such as 180 degrees if the check is meant to live in `test_final()`.

newton/examples/vbd/example_soft_cube_compression.py
```diff
@@ -0,0 +1,295 @@
...
+            self.state_0,
+            "particle velocities do not explode",
+            lambda q, qd: wp.length(qd) < 100.0,
+        )
+
+        # Height recovery: top particles should recover to >50% of rest height
+        top_z_final = float(np.mean(q[self.top_indices, 2]))
+        recovered_height = top_z_final - self.base_z
+        recovery_ratio = recovered_height / self.cube_size
+        if recovery_ratio < 0.50:
```
eric-heiden: 🟡 This recovery check can pass a cube that never recovers from the 50% compression. At release, a no-recovery state has `recovery_ratio == 0.50`, and this condition only fails values below `0.50`. The file header says this scenario should recover toward rest height after release, so the registered run should assert meaningful recovery in the main example state, for example by requiring the top face to return near `1.0` after the settle frames or by checking a measured improvement after release.

newton/examples/vbd/example_soft_cube_compression.py
```diff
@@ -0,0 +1,295 @@
...
+def _compute_tet_volume(q: np.ndarray, tet_indices: np.ndarray) -> float:
+    v0 = q[tet_indices[:, 0]]
+    v1 = q[tet_indices[:, 1]]
+    v2 = q[tet_indices[:, 2]]
+    v3 = q[tet_indices[:, 3]]
+    d1 = v1 - v0
+    d2 = v2 - v0
+    d3 = v3 - v0
+    volumes = np.einsum("ij,ij->i", d1, np.cross(d2, d3)) / 6.0
+    return float(np.sum(np.abs(volumes)))
```
eric-heiden: 🟡 `np.abs(volumes)` makes this volume check insensitive to inverted tetrahedra. This example now claims the near-flat case inverts elements and recovers, but a state with negative signed tet volumes can still produce a healthy absolute total volume here. The same helper pattern is used by the stretch and twist examples, so those volume-preservation checks can also miss inversion. Consider checking signed volumes separately, for example requiring each tet to stay positive within a small tolerance while still using absolute volume for the bulk-volume ratio.

jumyungc:
> @jumyungc I address the all the issues except the varying poison ratio one. Can we defer that to a future PR?

Thank you for following up with the issues.
I am okay with deferring the volume correctness tests to a follow-up PR.
Also for a recovery test, it would be great if we can have a classic test like this (this can also be a follow up):
image

jcarius-nv:

adenzler-nvidia:
Hi @AnkaChan, are you still planning to update this PR? There are a few unresolved review comments, and the branch now needs adjustment for the newer VBD damping semantics on `main`. If not, should we close it for now?

jcarius-nv:
Closing, please re-open when you get back to this.

eric-heiden:
Hi @AnkaChan, can you update this PR to use Towncrier changelog fragments? I think the examples also need t o be updated, they still use the old `Model.collide()` method which we deprecated.

PR #2908, opened 2026-05-19
[Kamino]: Fix contact filtering and constraint stabilization

vastsoun:
## Description

This PR resolves several issues in Kamino that effect contact generation/propagation and constraint stabilization.

Specifically,
- ContactsKamino can now be instantiated with a model  or per-world contact capacities
- Contact conversions to/from Newton contacts now better handle edge cases where contacts indices between the representations may be different
- Contact forces/wrenches are now consistent with the contact interface contract and are filled in using the correct contact index mapping
- Contact margins are now generated from the associated shapes and propagated to necessary internal computations
- Contact filtering both in Newton->Kamino conversions and internal CD pipeline can now optionally either filter-out or allow positive contact distances (i.e. contacts generated between the gap and margin regions)
- Contact constraint stabilization can now handle both positive-distances and allow for a configurable penetration deadzone to stabilize contacts and reduce flickering
- Additionally, `LimitsKamino.detect()` was changed to directly accept joint coords arrays (this was bundled to prepare for more upcoming changes to SolverKamino internals and physics performance metrics)
- A new internal example of a sphere-on-floor was added to be used for debugging and validation of the changes to contact filtering and constraint stabilization

## Checklist

## Release Notes

* **New Features**
  * Added example demonstrating basic Kamino solver usage with physics simulation.
  * Extended test utilities with reusable scene builders (sphere, stacked boxes, hinged objects).
  * Enhanced contact sensor integration with Kamino solver for force aggregation validation.

* **Bug Fixes**
  * Improved contact stabilization algorithm and restitution gating behavior.
  * Better margin and gap handling in collision detection.

* **Tests**
  * Expanded contact conversion and force validation test coverage.


newton/_src/solvers/kamino/_src/geometry/unified.py
```diff
@@ -99,125 +102,143 @@ class ContactWriterDataKamino:
...
+            wid = wid_b
+        world_max_contacts = writer_data.world_max_contacts[wid]
+
+        # Always allocate from the model-level counter so the active count
+        # stays accurate regardless of whether the narrowphase pre-allocated
+        # an output_index (primitive kernel) or left it to the writer (-1).
+        mcid = wp.atomic_add(writer_data.contacts_model_num_active, 0, 1)
+        wcid = wp.atomic_add(writer_data.contacts_world_num_active, wid, 1)
+        if mcid >= writer_data.model_max_contacts or wcid >= world_max_contacts:
+            # TODO: Are these roll-backs correct? Can't they cause overwriting of other contacts?
```
chschuma-disney: As far as I can tell, if the roll backs are not split, they might cause overwriting. The rollbacks work when you can be sure that if the current thread gets an index that exceeds the limit, any other thread will also get an index that exceeds the limit. If you have two indices with separate checks, it gets a bit tricky.

Here's a scenario where it fails if you get both indices at the same time:
- Thread 0 working on world A gets `mcid` and `wcid_A`.
- Thread 1 working on world B gets `mcid + 1` and `wcid_B`.
- Thread 0 checks its indices, and sees that `wcid_A` exceeds the limit, so it rolls back.
- Thread 1 checks its indices, everything is fine, and it writes the contact to `mcid + 1`.
- Thread 2 working on world C now gets `mcid + 1` (since Thread 0 rolled back), and might overwrite the contact from Thread 1.

I think we can build on the observation that no contacts will be written if `contacts_model_num_active` is full, but contacts can still be written if a specific `contacts_world_num_active` is full. So this should work:
- First get `wcid`, and check whether it is valid. If not, roll back and return.
- Then get `mcid`, and check whether it is valid. If not, roll back *both* indices and return.

newton/_src/solvers/kamino/examples/newton/example_kamino_basic_sphere.py
```diff
@@ -0,0 +1,175 @@
...
+        self.viewer = viewer
+        self.device = wp.get_device()
+
+        # Create a single-robot model builder and register the Kamino-specific custom attributes
+        scene_builder = newton.ModelBuilder(up_axis=newton.Axis.Z)
+        newton.solvers.SolverKamino.register_custom_attributes(scene_builder)
+        scene_builder.default_shape_cfg.margin = 1e-3
+        scene_builder.default_shape_cfg.gap = 0.1
+
+        # Load the DR Legs USD and add it to the builder
```
chschuma-disney: Wrong comment :)

newton/_src/solvers/kamino/_src/geometry/contacts.py
```diff
@@ -399,10 +413,32 @@ class ContactsKamino:
...
+            capacity:
+                The maximum number of contacts to allocate.\n
+                If not provided, the contact allocation meta-data of the model will be used.\n
+                If an integer is provided, it specifies the capacity for a single world.\n
+                If a list of integers is provided, it specifies the capacity for each world.
+            default_max_contacts:
+                The default maximum number of contacts per world.\n
+                If `None`, uses the default value of 1000.
+            device:
+                The device on which to allocate the contacts data.
```
Guirec-Maloisel, line 451: Missing an entry for the new `remappable` argument

newton/_src/solvers/kamino/_src/geometry/contacts.py
```diff
@@ -399,10 +413,32 @@ class ContactsKamino:
...
+                The model container holding the time-invariant data of the system being simulated.\n
+                If provided, the contacts will be finalized using the contact allocation meta-data of the model.\n
+                If `None`, the contacts will be created empty without allocating data, and
+                can be finalized later by providing a model to the `finalize` method.\n
+            capacity:
+                The maximum number of contacts to allocate.\n
+                If not provided, the contact allocation meta-data of the model will be used.\n
+                If an integer is provided, it specifies the capacity for a single world.\n
+                If a list of integers is provided, it specifies the capacity for each world.
+            default_max_contacts:
```
Guirec-Maloisel, line 446: By reading the docstring, it is not clear what the distinction between `capacity` and `default_max_contacts` is (when providing a single integer as `capacity`)

chschuma-disney (commented):
Looks mostly good to me. I wrote a small comment on the rollbacks for the contact writer, happy to look at this in person.

newton/_src/solvers/kamino/_src/geometry/contacts.py
```diff
@@ -638,22 +686,64 @@ def mode(self) -> wp.array:
...
+                num_worlds = model.size.num_worlds
+                world_max_contacts = [model_max_contacts // num_worlds] * num_worlds
+            capacity = world_max_contacts
+            self._device = model.device
+        else:
+            self._device = device
+
         # The memory allocation requires the total number of contacts (over multiple worlds)
         # as well as the contacts capacities for each world. Corresponding sizes are defaulted to 0 (empty).
         model_max_contacts = 0
```
Guirec-Maloisel: These lines will overwrite the changes just above (and even worse, enforce a single world!)

Guirec-Maloisel (commented):
I'm consistently getting failures in the PADMM unit tests with this branch. If I change the tolerances from 1e-6 to 1e-5 in the unit tests, that's enough to make them pass again. So this might need investigation.

Guirec-Maloisel:
> I'm consistently getting failures in the PADMM unit tests with this branch. If I change the tolerances from 1e-6 to 1e-5 in the unit tests, that's enough to make them pass again. So this might need investigation.

I tracked down the changes causing the test failures, turns out it is the line that makes the delta "dead zone" for contact stabilization one-sided only. ~~I changed it back to two-sided.~~

**Edit:** The change back to two-sided causes flickering on the box-on-plane example due to discontinuities in the stabilization at +/- delta. I changed the formulation to be symmetric and continuous (by shifting distances outside the dead-zone by delta, so that the stabilization target is the closest limit of the dead-zone rather than zero). Unit tests are passing, and there is no visible flickering on the box-on-plane test.

newton/_src/solvers/kamino/_src/geometry/contacts.py
```diff
@@ -746,202 +834,407 @@ def _assert_has_data(self):
...
+
+    # Retrieve the contact index for the Newton format and ensure that
+    # it has a valid mapping to the target Newton contacts container.
+    cid_out = kamino_remap[cid]
+
+    # Skip conversion if this contact index exceeds the number of active
+    # contacts or it has no mapping to the target Newton contacts container.
+    if cid >= num_active or cid_out < 0:
+        newton_force[cid] = wp.spatial_vectorf()
+        return
```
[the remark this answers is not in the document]
Guirec-Maloisel: Good point, I fixed as suggested.

newton/_src/solvers/kamino/_src/geometry/unified.py
```diff
@@ -99,125 +102,143 @@ class ContactWriterDataKamino:
...
+            # wp.atomic_sub(writer_data.contacts_model_num_active, 0, 1)
+            # wp.atomic_sub(writer_data.contacts_world_num_active, wid, 1)
             return
```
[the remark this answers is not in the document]
Guirec-Maloisel: This is a valid point, but the commented-out rollbacks that undo the atomic add are not fully thread-safe either. This needs a dedicated fix in a separate PR (the issue is new in this file but already present in narrowphase.py)

newton/_src/solvers/kamino/examples/newton/example_kamino_basic_sphere.py
```diff
@@ -0,0 +1,175 @@
...
+#
+###########################################################################
+
+import numpy as np
+import warp as wp
+
+import newton
+import newton.examples
+from newton._src.solvers.kamino._src.utils import logger as msg
+from newton.tests.utils import basics
```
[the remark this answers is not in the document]
Guirec-Maloisel: This example is not exposed at the newton top-level, but Kamino-internal only. This type of clean-up will happen when (and if) migrating it into newton examples at the top level.

newton/_src/solvers/kamino/solver_kamino.py
```diff
@@ -651,19 +670,26 @@ def update_contacts(self, contacts: Contacts, state: State | None = None) -> Non
...
 
         # Ensure the output contacts containers has sufficient size to hold the contact data from Kamino
-        if self._contacts_kamino._data.model_max_contacts_host > contacts.rigid_contact_max:
-            raise ValueError(
+        if self._contacts_kamino.model_max_contacts_host > contacts.rigid_contact_max:
+            raise RuntimeError(
                 f"Contacts container has insufficient capacity for Kamino contacts: "
-                f"model_max_contacts={self._contacts_kamino._data.model_max_contacts_host} > "
+                f"model_max_contacts={self._contacts_kamino.model_max_contacts_host} > "
                 f"contacts.rigid_contact_max={contacts.rigid_contact_max}"
```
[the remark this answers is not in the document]
Guirec-Maloisel: This is a valid issue but such a comparison is not possible within graph capture. Therefore it is up to the user to ensure that capacities are matched on the Newton and Kamino side upon seeing this error. Finding ways to automatically propagate capacities to make it easier is left as future work.

newton/_src/solvers/kamino/tests/test_geometry_contacts.py
```diff
@@ -461,407 +656,665 @@ def _setup_newton_scene(self):
...
             np.testing.assert_allclose(
-                p1w,
-                kamino_pos_B[i],
-                atol=1e-4,
-                err_msg=f"Contact {i}: round-trip point1 world mismatch",
+                force[cid_out, 3:].astype(np.float64),
+                expected_torque[cid_out],
+                atol=1e-5,
+                err_msg=f"Contact cid={cid_out}: torque must equal (r_pt - r_com) x f_world",
             )
```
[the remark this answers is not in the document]
Guirec-Maloisel: In this test the capacity is larger on the Kamino side (nc + 16) so there are actually no unmapped contacts here.

newton/examples/kamino/example_kamino_robot_dr_legs.py
```diff
@@ -51,6 +51,9 @@ def __init__(self, viewer: newton.viewer.ViewerBase, args=None):
         # Create the multi-world model by duplicating the single-robot
         # builder for the specified number of worlds
         builder = newton.ModelBuilder(up_axis=newton.Axis.Z)
+        builder.request_contact_attributes("force")
+        builder.default_shape_cfg.margin = 1e-6
+        builder.default_shape_cfg.gap = 1e-2
```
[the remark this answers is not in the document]
Guirec-Maloisel: These are minor adjustments that don't change the overall behavior at a visible level for users.

newton/_src/solvers/kamino/_src/dynamics/dual.py
```diff
@@ -622,36 +622,37 @@ def _build_free_velocity_bias_contacts(
...
+    # NOTE: We still write zeros to overwrite previous values
+    problem_v_b[ccio_k] = 0.0
+    problem_v_b[ccio_k + 1] = 0.0
+    problem_v_b[ccio_k + 2] = alpha * xi_relaxed
 
     # Initialize the restitutive Newton-type impact model term
+    # NOTE: We still write zeros to overwrite previous values
     problem_v_i[ccio_k] = 0.0
     problem_v_i[ccio_k + 1] = 0.0
     problem_v_i[ccio_k + 2] = epsilon_k
```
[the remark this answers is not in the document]
Guirec-Maloisel: This is a deeper issue tied to the formulation of contacts and more specifically of stabilization and restitution through additive biases. Your proposed fix would prevent bouncing at a distance, but also remove correct restitution for objects incoming sufficiently fast (practically requiring penetration before any restitution, and thus delaying restitution by one time step; and even then this delayed restitution will be based on the incoming velocity at the next step, so the bouncing amount might not be correct). For now we are living with this approximation (the bouncing at a distance is limited by the collision detection gap being small) and leave it as future work to find a better handling to this case.

newton/_src/solvers/kamino/_src/geometry/contacts.py
Guirec-Maloisel: Added.

newton/_src/solvers/kamino/_src/geometry/contacts.py
Guirec-Maloisel: Clarified in the docstring.

newton/_src/solvers/kamino/solver_kamino.py
```diff
@@ -424,7 +424,20 @@ def __init__(
...
+                remappable=True,
+            )
+
+        # Declare an internal reference cache to be able to detect if
+        # a Kamino-internal collision detector was used at runtime.
+        # NOTE: This is used to determine whether to clear the output
+        # contacts and populate them with only active contacts or fill
+        # in solver-specific contact attributes for existing contacts.
+        # TODO: Do we need this additional indirection or is there a better way to do this?
+        self._detector = None
```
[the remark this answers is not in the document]
Guirec-Maloisel: I added a note that the contacts are expected to be the same as during step(), otherwise the behavior is undefined (there is not usecase for that case currently so we don't want to support it)

newton/tests/utils/basics.py
```diff
@@ -550,13 +666,141 @@ def build_cartpole(
...
+    )
+
+    # Add top body
+    bid_t = _builder.add_body(
+        label="top_box",
+        mass=1.0,
+        inertia=I_t,
+        xform=xform_t,
+        lock_inertia=True,
+    )
```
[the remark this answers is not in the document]
Guirec-Maloisel: Good catch, I fixed it as suggested

newton/tests/utils/basics.py
```diff
@@ -89,11 +89,106 @@ def _add_ground_box(builder: ModelBuilder) -> None:
...
+    custom_shape_cfg = (
+        ModelBuilder.ShapeConfig(
+            gap=0.01,
+            margin=1e-6,
+            mu=friction if friction is not None else _builder.default_shape_cfg.mu,
+            restitution=restitution if restitution is not None else _builder.default_shape_cfg.restitution,
+        )
+        if use_custom_shape_cfg
+        else _builder.default_shape_cfg
+    )
```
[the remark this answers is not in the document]
Guirec-Maloisel: Good point, I fixed as suggested

newton/_src/solvers/kamino/_src/geometry/unified.py
Guirec-Maloisel: I double checked manually + with Claude, your solution seems to work cleanly. If the model counter is saturated you don't get the same issue after decreasing back the world index, because the model counter is saturated for all threads in all worlds anyway. I applied the fix also to similar code in the primitive pipeline (primitive/narrowphase.py)

newton/_src/solvers/kamino/_src/geometry/unified.py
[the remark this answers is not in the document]
Guirec-Maloisel: Actually the fix suggested by @chschuma-disney below is safe after double-checking, so I applied it directly in this PR

newton/_src/solvers/kamino/_src/geometry/primitive/narrowphase.py
```diff
@@ -382,12 +382,23 @@ def add_multiple_contacts(
...
+            wp.atomic_sub(contact_world_num, wid, num_contacts)
+            return
 
-        # Retrieve the maximum number of contacts that can be stored for this geom pair
+        # Determine the maximum number of contacts that can be stored for this geom pair
+        # (may be < num_contacts if this thread saturated the world/model counter)
         max_num_contacts = wp.min(wp.min(model_max_contacts - mcio, world_max_contacts - wcio), num_contacts)
+        if max_num_contacts < num_contacts:  # Correct counters if partial contacts were written
+            wp.atomic_sub(contact_model_num, 0, num_contacts - max_num_contacts)
+            wp.atomic_sub(contact_world_num, wid, num_contacts - max_num_contacts)
```
chschuma-disney: I think these updates might lead to the possibility of two contacts being written to the same index. The decrements should follow a similar logic like in the case of one contact, i.e., check for saturation of the world (and updating the number of contacts) before incrementing the number of model contacts.

newton/_src/solvers/kamino/_src/geometry/contacts.py
```diff
@@ -640,27 +701,72 @@ def mode(self) -> wp.array:
...
-        Finalizes the contacts data allocations based on the specified capacity.
+        Finalizes the contacts data allocations based on the specified model or capacity.
 
         Args:
-            capacity (int | list[int]):
-                The maximum number of contacts to allocate.\n
+            model:
+                The model container holding the time-invariant data of the system being simulated.\n
+                If provided, the contacts will be finalized using the contact allocation meta-data of the model,
+                taking precedence over `capacity` and `default_max_contacts`.\n
```
chschuma-disney: Could be clarified that `model` doesn't take precedence over `capacity`, having both specified will raise an error.

newton/_src/solvers/kamino/_src/geometry/contacts.py
```diff
@@ -750,208 +854,393 @@ def _assert_has_data(self):
...
-    rest = float32(0.5) * (shape_restitution[s0] + shape_restitution[s1])
+    # Store the contact data in the Kamino format
+    gapfunc = wp.vec4f(normal[0], normal[1], normal[2], distance)
+    q_frame = wp.quat_from_matrix(make_contact_frame_znorm(normal))
 
+    # Increment the number of active contacts in the model and world
     mcid = wp.atomic_add(kamino_model_active, 0, 1)
     wcid = wp.atomic_add(kamino_world_active, wid, 1)
+    if mcid >= max_converted_contacts:
+        return
```
chschuma-disney, line 1016: Here, we're no longer checking for the maximum contacts per world. Is that intended?

newton/_src/solvers/kamino/examples/newton/example_kamino_basic_sphere.py
```diff
@@ -0,0 +1,175 @@
...
+        self.graph = None
+        if self.device.is_cuda:
+            with wp.ScopedCapture() as capture:
+                self.simulate()
+            self.graph = capture.graph
+
+    # simulate() performs one frame's worth of updates
+    def simulate(self):
+        for _ in range(self.sim_substeps):
+            print("\n\n--------------------------------------------------------------------")
```
chschuma-disney: Could we clean this up so it only outputs anything to the console if a `verbose` flag is set?

newton/_src/solvers/kamino/tests/test_geometry_contacts.py
```diff
@@ -907,22 +1253,133 @@ def test_06_newton_to_kamino_surface_anchors_strip_shape_margin(self):
...
 
+    ###
+    # Optional Contacts.force handling
+    ###
 
-###
-# Helpers
-###
-
+    def test_06_force_optional(self):
```
chschuma-disney: Nit: `test_06_...` already exists above.

chschuma-disney (commented):
Looks mostly good to me, some small comments.

newton/_src/solvers/kamino/_src/geometry/primitive/narrowphase.py
Guirec-Maloisel: You're right, allocated slots in the global counter could be released again because a world is saturated.

I updated the code to first check saturation on the world side, and only then try to allocate as many slots as allowed by the world on the global counter. So it should now be correct following the same logic as in other places (we only allocate slots in the global counter if we commit to using them, and decrease that counter only if it saturates; if that is the case, the decrease to the world counter is safe because any threads in any world will see the same saturation of the global counter so the order doesn't matter).

newton/_src/solvers/kamino/_src/geometry/contacts.py
Guirec-Maloisel: Good point, I updated as suggested

newton/_src/solvers/kamino/_src/geometry/contacts.py
Guirec-Maloisel: I checked in the history of this file and there was never such a check. But it would make sense to add it, so I changed the code to follow the same safe counter increment pattern here as well (passing the needed extra arguments to the kernel). I also removed instead the existing argument and check (max_converted_contacts) because that is already the number of threads that is launched.

newton/_src/solvers/kamino/examples/newton/example_kamino_basic_sphere.py
Guirec-Maloisel: I set all printing lines to use the "debug" log level

newton/_src/solvers/kamino/tests/test_geometry_contacts.py
Guirec-Maloisel: Good point, I shifted the numbers.

chschuma-disney (approved):

PR #2916, opened 2026-05-20
Fix mesh-convex margin candidate query

adenzler-nvidia:
## Description

Fix mesh-vs-convex and heightfield-vs-convex midphase candidate generation to query candidates using the full `margin + gap` contact envelope, not just `gap`.

This keeps the mesh BVH query and heightfield grid-cell lookup consistent with broad-phase AABB expansion and final contact acceptance. Without this, valid margin-shell contacts can be culled before narrow phase processing.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k Midphase
uv run --extra dev ruff check newton/_src/geometry/collision_core.py newton/_src/geometry/narrow_phase.py newton/_src/utils/heightfield.py newton/tests/test_collision_pipeline.py
git diff --check
```

## Bug fix

**Steps to reproduce:**

1. Create a mesh-vs-sphere or heightfield-vs-sphere scene with per-shape `margin=0.02`, `gap=0.005`.
2. Place the sphere with true surface separation `0.03`.
3. Run the Newton collision pipeline.
4. Without this PR, no contact is generated even though `0.03 <= margin_a + margin_b + gap_a + gap_b`.

The heightfield regression covers the lateral edge case where grid-cell candidate lookup can cull the contact before narrow phase. Vertical margin-shell contacts above the heightfield already reached narrow phase because the grid lookup is driven by XY cell overlap.

**Minimal reproduction:**

```python
# Covered by:
# uv run --extra dev -m newton.tests -k Midphase
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed missing mesh–convex and heightfield–convex contacts when shapes are separated by margin but still lie within the contact envelope; contact thresholds now include per-shape margin.
* **Tests**
  * Added midphase tests verifying contacts are produced for mesh/heightfield vs convex cases when margins/gaps match.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2916?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 ### Fixed
 
 - Fix MJCF `xyaxes` parsing to treat the second vector as Y and derive Z from X cross Y.
+- Fix mesh-convex contacts missing when shapes are separated by margin but still within the contact envelope.
 - Fix `SolverMuJoCo` returning `State.joint_qd` in world frame for root `FREE` joints with non-identity `parent_xform`, violating the documented parent-frame contract and corrupting derived `body_qd`.
 - Fix `basic_conveyor` example emitting a spurious inertia validation warning at finalize.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 ### Fixed
 
 - Fix MJCF `xyaxes` parsing to treat the second vector as Y and derive Z from X cross Y.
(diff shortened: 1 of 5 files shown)


nvtw:
I think the bugfix is correct. But can you double check - it's very likely that the heightfield code has the same problem.

nvtw (approved):

PR #2921, opened 2026-05-21
Fix vbd custom attribute registration scopes

### LINKED ISSUE #2903: VBD: Make Dahl cable friction opt-in when registering custom attributes
SolverVBD.register_custom_attributes(builder) currently registers both the joint hard/soft mode attribute (vbd:joint_is_hard) and the Dahl cable-friction attributes (vbd:dahl_eps_max, vbd:dahl_tau).

If users only want to author joint modes, they currently have to explicitly set the Dahl parameters to zero to avoid enabling Dahl cable friction, which is less convenient.

jumyungc:
## Description

This PR addresses unintended Dahl cable-friction enablement when registering VBD custom attributes.

`SolverVBD.register_custom_attributes(builder)` remains the single VBD custom-attribute registration hook. 
To preserve existing behavior during the transition, it keeps implicit positive Dahl defaults by default, but that path is now deprecated.

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented Dahl cable-friction from being enabled unintentionally when only joint-mode attributes are present.

* **Behavior Change**
  * Dahl friction is disabled by default and now activates only when positive Dahl parameters are explicitly authored per joint.

* **New Features**
  * Joint hard/soft mode is initialized at build time from authored attributes, with runtime changes still supported.

* **Deprecated**
  * Deprecation warnings clarify changed defaults and opt-in for Dahl friction.

* **Examples & Tests**
  * Examples and tests updated to show explicit opt-in and new defaults.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2921?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


adenzler-nvidia:
I agree with the issue this PR is addressing, but I don't think the right API shape is to add feature-specific public registration methods.

My understanding is that each solver should expose a single `register_custom_attributes(builder)` hook. The current problem is not that users need a narrower public hook, but that registering VBD attributes has a non-inert side effect: `register_custom_attributes()` declares `vbd:dahl_eps_max` and `vbd:dahl_tau`, `ModelBuilder.finalize()` materializes them with positive defaults (`0.5` / `1.0`), and `SolverVBD(model)` treats the presence of both arrays as opting into Dahl friction.

So a user registering VBD attributes only to author `vbd:joint_is_hard` can unintentionally enable Dahl cable friction.

Could we instead keep the single `SolverVBD.register_custom_attributes(builder)` API and make registration inert by default? For example, register the Dahl attributes with a disabled default such as `dahl_eps_max=0.0`, and require examples/users that want Dahl friction to author positive values. That preserves the solver registration contract and fixes the behavioral side effect at the source.

jumyungc:
> I agree with the issue this PR is addressing, but I don't think the right API shape is to add feature-specific public registration methods.
> 
> My understanding is that each solver should expose a single `register_custom_attributes(builder)` hook. The current problem is not that users need a narrower public hook, but that registering VBD attributes has a non-inert side effect: `register_custom_attributes()` declares `vbd:dahl_eps_max` and `vbd:dahl_tau`, `ModelBuilder.finalize()` materializes them with positive defaults (`0.5` / `1.0`), and `SolverVBD(model)` treats the presence of both arrays as opting into Dahl friction.
> 
> So a user registering VBD attributes only to author `vbd:joint_is_hard` can unintentionally enable Dahl cable friction.
> 
> Could we instead keep the single `SolverVBD.register_custom_attributes(builder)` API and make registration inert by default? For example, register the Dahl attributes with a disabled default such as `dahl_eps_max=0.0`, and require examples/users that want Dahl friction to author positive values. That preserves the solver registration contract and fixes the behavioral side effect at the source.

Thank you @adenzler-nvidia for the review and for clarifying the preferred API direction. I was weighing both options: the cleaner single-register approach versus keeping the existing default, and I now see the clearer direction. I updated the code and changelog accordingly.

adenzler-nvidia:
Thanks for updating this toward the single `SolverVBD.register_custom_attributes(builder)` API shape. I think that is the right long-term direction, and it addresses the concern that users should not need feature-specific public registration methods just to author `vbd:joint_is_hard`.

Looking at this again, though, I think the specific default change here is still breaking. In `v1.2.0`, `rigid_enable_dahl_friction` was deprecated/ignored and Dahl enablement moved to the registered model attributes. That means existing users may now rely on `SolverVBD.register_custom_attributes(builder)` materializing `vbd:dahl_eps_max=0.5` / `vbd:dahl_tau=1.0` and enabling Dahl. Changing `dahl_eps_max` to `0.0` immediately would silently disable Dahl for those users.

Could we make this a compatibility transition instead? For example:

```python
SolverVBD.register_custom_attributes(builder, dahl_defaults_enabled=True)
```

where `True` preserves the current released behavior, and call sites that only need `vbd:joint_is_hard` can pass:

```python
SolverVBD.register_custom_attributes(builder, dahl_defaults_enabled=False)
```

That would let the RJ45 example avoid enabling Dahl without adding feature-specific registration methods, while still avoiding a behavior break for existing users.

I think this should also come with a deprecation warning / changelog entry for the implicit Dahl enablement path: registering VBD custom attributes with positive Dahl defaults should be deprecated, and users who want Dahl should explicitly author positive `model.vbd.dahl_eps_max` and `model.vbd.dahl_tau`.

Can you confirm the intended final behavior? My understanding is that the eventual target is for `register_custom_attributes()` to be inert by default, with `dahl_eps_max=0.0`, and Dahl enabled only when positive Dahl parameters are explicitly authored.

adenzler-nvidia (approved):
Bug fix is sound: registering `vbd:joint_is_hard` no longer implicitly enables Dahl friction. Back-compat preserved via `dahl_defaults_enabled=True` default with a deprecation warning, and the migration path (`dahl_defaults_enabled=False` + explicit per-joint authoring) is clear. Tests cover the core contract and in-repo examples are updated.

PR #2926, opened 2026-05-21
[BUG] Make ViewerGL GUI DPI-aware and movable

### LINKED ISSUE #2674: Fix GUI scaling in ViewerGL
The GUI needs to be aware of high-dpi monitors to scale appropriately.

Image

johnnynunez-nv:
Apply ImGui style and font scaling derived from pyglet's `window.scale` (falling back to the framebuffer/window-size ratio for X11/Win32) so the left control panel, FPS overlay, scalar plots window, and image logger render at their intended physical size on Retina/4K displays instead of half size. The detection prefers `window.scale` because on macOS pyglet's cocoa backend reports both `get_size()` and `get_framebuffer_size()` as physical pixels, so the ratio alone is always 1.0 and never triggers HiDPI scaling.

Also switch the left panel from an always-anchored fixed-size window to one with `Cond_.first_use_ever` position/size and no `no_resize` flag, so users can drag it by the title bar, resize via the bottom-right corner (clamped between 220x160 dpi-scaled and the display size), and scroll with the mouse wheel or trackpad gestures when contents overflow. The initial dock-on-left layout is preserved.

Closes #2674.

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **New Features**
  * Left control panel is movable (drag title bar), resizable (drag bottom-right corner), preserves initial left dock, and shows an automatic vertical scrollbar with mouse-wheel and two-finger trackpad scrolling.

* **Bug Fixes**
  * HiDPI/Retina improvements: overall UI, fonts, sidebar, plots, image/log windows and grid spacing now scale correctly and update at runtime when display DPI changes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2926?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


adenzler-nvidia (commented):
Principal engineer review notes from the DPI/layout pass.

newton/_src/viewer/gl/gui.py
```diff
@@ -319,6 +380,13 @@ def resize(self, width, height):
         # Use framebuffer size for display size
         self.io.display_size = fb_width, fb_height
 
+        # Reapply style/font scaling only when the DPI actually changes —
+        # e.g. the window moved to a display with different scaling.
+        new_dpi_scale = self._detect_dpi_scale()
```
adenzler-nvidia, line 414: Runtime DPI changes can bypass this path. pyglet 2.1 dispatches `on_scale` when the window moves between displays with different DPI, and that event does not necessarily go through `on_resize`. In that case `window.scale` changes, but `dpi_scale`, the scaled ImGui style/font state, and the image logger sidebar hint stay stale until the user manually resizes the window. Please hook `on_scale` as well, or centralize/poll the DPI refresh so the viewer reacts to scale-only changes.

newton/_src/viewer/viewer_gl.py
```diff
@@ -240,7 +243,7 @@ def __init__(
 
         self.renderer = RendererGL(vsync=vsync, screen_width=width, screen_height=height, headless=headless)
         self.renderer.set_title("Newton Viewer")
-        self._image_logger = ImageLogger(device=self.device, sidebar_width_px=_SIDEBAR_WIDTH_PX)
+        self._image_logger = ImageLogger(device=self.device, sidebar_width_px=self._sidebar_width_fb_px())
```
adenzler-nvidia: This only scales the image logger's sidebar-avoidance offset. The image logger still opens and lays out its own window using unscaled constants in `newton/_src/viewer/gl/image_logger.py` (`_INITIAL_TILE_PX`, `_INITIAL_WINDOW_MAX_W`, `_INITIAL_WINDOW_MAX_H`, `_INITIAL_WINDOW_PAD_*`, and `_TILE_SPACING_PX`). On a 2x display, logged image windows will still appear at roughly half the intended physical tile/window size, even though the PR description says the image logger is fixed too. Please pass the DPI scale into `ImageLogger` and apply it to those initial sizing and spacing constants as well.

newton/_src/viewer/gl/gui.py
johnnynunez-nv: Done

newton/_src/viewer/viewer_gl.py
johnnynunez-nv: done

johnnynunez-nv:
image
@preist-nvidia now it scales the other windows too! thank you

newton/_src/viewer/viewer_gl.py
```diff
@@ -2189,6 +2208,54 @@ def on_resize(self, width: int, height: int):
...
+        self._refresh_dpi_state()
+
+    def _on_window_scale(self, scale: float, dpi: int) -> None:
+        """Refresh DPI-dependent layout when pyglet reports a display change.
+
+        pyglet 2.1 dispatches ``on_scale`` whenever the window crosses to a
+        display with a different ``backingScaleFactor`` / DPI. The window
+        size need not change, so ``on_resize`` isn't always fired.
+        """
+        self._refresh_dpi_state()
```
adenzler-nvidia: `ViewerGL._on_window_scale()` can still propagate the old DPI value here. pyglet stores pushed handlers as a stack, and this handler is registered after `UI(self.renderer.window)`, so it runs before `UI._on_window_scale()` updates `ui.dpi_scale`. `_refresh_dpi_state()` then calls `_dpi_scale()`, which returns `ui.dpi_scale` when the UI is available, so the image logger/sidebar hint can be left at the previous scale on a scale-only display move. Please either compute the scale from the window directly in this path, force the UI DPI refresh before propagating, or otherwise make the handler ordering irrelevant.

johnnynunez-nv: You were right about the handler order: the ViewerGL `on_scale` handler can run before the UI one, so it could read the old `ui.dpi_scale`.
I changed the flow so `ViewerGL._refresh_dpi_state()` explicitly calls `self.ui.refresh_dpi()` first, before updating the image logger/sidebar values. That way the UI scale is refreshed from the current pyglet window state even if the handlers fire in the “wrong” order.
I also narrowed the `on_scale` exception handling while I was there.

I am really trying to understand debugging pyglet behavior

newton/_src/viewer/gl/gui.py
```diff
@@ -56,16 +60,119 @@ def on_mouse_scroll(x, y, scroll_x, scroll_y):
...
+            return
 
         self._setup_dark_style()
+        style = self.imgui.get_style()
+        if self.dpi_scale != 1.0:
+            style.scale_all_sizes(self.dpi_scale)
+        # ImGui >= 1.92 exposes ``font_scale_dpi`` for the dynamic font system;
+        # older builds use the deprecated ``io.font_global_scale``. Set whichever
+        # is available so fonts scale to DPI on both versions.
+        if hasattr(style, "font_scale_dpi"):
```
adenzler-nvidia: `pyproject.toml` now requires `imgui_bundle` 1.92.x or newer for examples, and the 1.92 API uses `style.font_scale_dpi`; the old `io.font_global_scale` fallback is for an unsupported version and is absent in the locked 1.92.601 environment. Keeping this version probe/comment makes the supported API surface unclear. Please set `style.font_scale_dpi` directly, or intentionally lower/document the imgui minimum if older builds are meant to work.

newton/_src/viewer/gl/gui.py
```diff
@@ -56,16 +60,119 @@ def on_mouse_scroll(x, y, scroll_x, scroll_y):
...
 
-        # Set up proper DPI scaling for high-DPI displays
+        # pyglet 2.1 dispatches ``on_scale`` when the window moves between
+        # displays with different DPI scaling. Such transitions don't always
+        # fire ``on_resize``, so we'd otherwise miss the DPI change. Older
+        # pyglet builds don't register the event type at all; AttributeError
+        # / KeyError from ``push_handlers`` is the expected outcome there.
+        try:
+            self.window.push_handlers(on_scale=self._on_window_scale)
+        except (AttributeError, KeyError) as exc:
```
adenzler-nvidia: Given the declared `pyglet` minimum, `Window.on_scale` should be part of the supported dependency range, so this older-pyglet fallback is likely dead. If we keep defensive handling, unknown pyglet event names raise `pyglet.event.EventException`, not `KeyError`; catching `KeyError` makes the comment inaccurate and misses the actual no-event case. Please either remove the fallback or catch the pyglet exception explicitly. Same pattern appears in `ViewerGL.__init__`.

newton/_src/viewer/gl/gui.py
```diff
@@ -56,16 +60,119 @@ def on_mouse_scroll(x, y, scroll_x, scroll_y):
...
+            if window_scale > 1.0:
+                scale = window_scale
+        except Exception:
+            pass
+        try:
+            ww, wh = self.window.get_size()
+            fw, fh = self.window.get_framebuffer_size()
+            if ww > 0 and wh > 0:
+                scale = max(scale, fw / ww, fh / wh)
+        except Exception:
```
adenzler-nvidia: These DPI probes should not swallow every exception. Falling back for a missing or non-numeric `window.scale`, or for an unavailable size API, is reasonable; hiding arbitrary backend/window failures makes the viewer silently behave like a 1x display and makes real integration bugs hard to diagnose. Please narrow these catches to the expected cases or surface a warning. Same broad silent pattern appears in `ViewerGL._dpi_scale()`.

newton/_src/viewer/gl/gui.py
johnnynunez-nv: Thanks, you were right that resize-only refresh was too narrow here. I added the `on_scale` hook so scale-only display moves refresh the cached DPI state, ImGui style/font scaling, and the image logger/sidebar layout without waiting for a manual resize.

newton/_src/viewer/viewer_gl.py
johnnynunez-nv: Good catch. The logger now gets the current DPI scale from `ViewerGL` and applies it to the initial tile size, max window size, padding, and spacing values, so logged image windows open at the intended physical size on 2x displays too.

newton/_src/viewer/viewer_gl.py
johnnynunez-nv: Nice catch on pyglet handler ordering. I changed the viewer-side `on_scale` path to pass the event scale directly into `_refresh_dpi_state()` and coerce it there, so the image logger/sidebar update no longer depends on `ui.dpi_scale` having already been refreshed by another handler.

newton/_src/viewer/gl/gui.py
johnnynunez-nv: Agreed. Since the examples dependency already requires `imgui_bundle >= 1.92`, I removed the old `io.font_global_scale` fallback and now set `style.font_scale_dpi` directly.

newton/_src/viewer/gl/gui.py
johnnynunez-nv: Yep, that fallback was stale for the supported pyglet range. I removed the defensive `push_handlers(on_scale=...)` fallback in both `UI` and `ViewerGL` instead of keeping an inaccurate `KeyError` catch.

newton/_src/viewer/gl/gui.py
johnnynunez-nv: Agreed. I narrowed the DPI fallback handling to the expected missing/non-numeric scale cases, and stopped swallowing arbitrary failures from the size/framebuffer-size calls. The same cleanup is applied in the viewer DPI path as well.

johnnynunez-nv (commented):
Addressed the DPI review comments in the latest push. Thanks again for the careful pass.

newton/_src/viewer/viewer_gl.py
eric-heiden: 🟡 The scalar heatmap sizing still has a few unscaled pixel constants, so this path can remain visibly out of scale on HiDPI displays.

`window_width` is now scaled with `s`, but the margin passed to `_render_array_heatmap` is still `40.0` framebuffer pixels. At `s=2`, that cuts the visual gap in half in logical pixels and makes the heatmap wider relative to the rest of the sidebar. The helper also still uses fixed framebuffer-pixel values for `120.0`, the `80.0` to `220.0` height clamp, and `_heatmap_min_cell_pixels`.

Please scale these values with `self.ui.dpi_scale`, for example `window_width - 40.0 * s`, and apply the same scale to the heatmap minimum width, height clamp, and minimum cell size.

newton/_src/viewer/viewer_gl.py
```diff
@@ -290,6 +297,9 @@ def __init__(
         # Only create UI in non-headless mode to avoid OpenGL context dependency
         if not headless:
             self.ui = UI(self.renderer.window)
+            # pyglet fires ``on_scale`` when the window crosses to a display
+            # with a different DPI without necessarily firing ``on_resize``.
+            self.renderer.window.push_handlers(on_scale=self._on_window_scale)
```
eric-heiden, line 271: 🟡 `ViewerGL` and `UI` both register `on_scale` handlers on the same pyglet window. Since `push_handlers` dispatches through the handler stack, each scale event now refreshes DPI twice, and the two paths use different scale sources: the UI re-detects using the framebuffer ratio, while the image logger gets pyglet's raw `scale` argument.

That can make `ui.dpi_scale` and `image_logger.dpi_scale` diverge on fractional-scaling setups where the framebuffer ratio is larger than `window.scale`, so image-logger windows can render smaller than the rest of the UI.

Please make one component own `on_scale`, then feed the same resolved scale value into both `UI` and `ImageLogger`.

eric-heiden:
Thanks @johnnynunez-nv for addressing part of my comments. There still appear to be some unscaled constants in the heatmap plotting path. The PR should be good to merge once addressed and the merge conflicts are resolved.

newton/_src/viewer/viewer_gl.py
johnnynunez-nv: Good catch, addressed in 5e4e512. `_render_array_heatmap` now takes a `dpi_scale` argument and scales the four constants (`120.0` minimum width, the `80.0`–`220.0` height clamp, `_heatmap_min_cell_pixels`) by it. The `viewer_gui._render_scalar_plots` caller now passes both the scaled sidebar margin (`window_width - 40.0 * s`) and `dpi_scale=s`. While merging upstream, I also propagated `self.ui.dpi_scale` through `viewer_gui` to the left panel, plots window, stats overlay, and the selection-panel scroll height, and restored the movable/resizable left panel behavior. Merge with main is resolved.

adenzler-nvidia:
Looks like the automated changelog merge messed up the changelog - could you please fix it and remove the git stuff? Sorry about that. Otherwise this looks good to go.

johnnynunez-nv:
> Looks like the automated changelog merge messed up the changelog - could you please fix it and remove the git stuff? Sorry about that. Otherwise this looks good to go.

done

adenzler-nvidia (approved):
Thank you very much!

Issue #2751, opened 2026-05-07
[BUG] Bumping mujoco_warp to 3.8.0.2 cause a slowdown on macos runner


Kenny-Vilella:
### Bug Description

https://github.com/newton-physics/newton/pull/2744

The slowdown seems to be reproducible on the runner.
It does not show up on my local mac (test suite run takes 5min).
It looks like the slowdown appears mid run when compared to a successful run.

Kenny-Vilella:
I suspect that one of the 3 threads has a hang so that we run at 2/3 the speed.
Need some testing for confirmation.

preist-nvidia:
@eric-heiden  to disable macOS UTs for now until this is resolved so we can accept the version bump on main already.

Re-Enable when resolving this issue.

Kenny-Vilella:
Close with 3.8.0.3 (#2744)

PR #2927, opened 2026-05-21
Add CPU regression benchmark suite

### LINKED ISSUE #2830: Add benchmark to detect CPU regression
In order to prevent and better detect potential issue with the CPU codepath, we should add a benchmark measuring CPU perf.

This is to address the problem encountered in #2751.

vreutskyy:
## Description

Adds [asv/benchmarks/simulation/bench_cpu_regression.py](asv/benchmarks/simulation/bench_cpu_regression.py) — a minimal but broad benchmark suite that exercises Newton's CPU codepath, so regressions like #2751 (a mujoco_warp bump that silently slowed down the macOS CI runner) are detected directly rather than incidentally via test timeouts.

Three benchmarks, each pinned to ``wp.ScopedDevice("cpu")`` and ``use_cuda_graph=False``:

| Benchmark | Subsystem |
|---|---|
| ``CpuMuJoCoCartpole`` | mujoco_warp + Newton glue |
| ``CpuXPBDQuadruped`` | XPBD solver, rigid contacts, articulations |
| ``CpuIKFranka`` | IK solver, Jacobian path |

Each runs in well under a second of wall-clock on a warm Warp cache; total measured time across all three is ~1.6 s locally.

Closes #2830.

## Checklist

- [ ] ``CHANGELOG.md`` has been updated (if user-facing change)

## Test plan

\`\`\`
uv run --extra dev python asv/benchmarks/simulation/bench_cpu_regression.py
\`\`\`

Validated locally on Linux (Warp CPU backend, no CUDA driver). Sample run:

\`\`\`
CpuMuJoCoCartpole.time_simulate (): 0.961890
CpuXPBDQuadruped.time_simulate ():  0.325665
CpuIKFranka.time_solve ():          0.354647
\`\`\`

## Summary by CodeRabbit

* **New Features**
  * Added a CPU-only regression benchmark suite with three scenarios: MuJoCo Ant, XPBD Quadruped, and IK Franka.
  * Added a command-line option to select and run one or more specific benchmarks for performance testing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2927?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Kenny-Vilella:
It looks OK, provide consistent results, two minor things:
- what was the reasoning for choosing these three benchmarks?
- Could you rename the file to bench_cpu.py? Technically all benchmarks are here to catch regression.

vreutskyy:
Thanks for the review @Kenny-Vilella\!

> Could you rename the file to bench_cpu.py?

Done in 2e4e6721 — fair point, the suffix was redundant.

> what was the reasoning for choosing these three benchmarks?

Picked one minimal workload from each major, non-overlapping CPU codepath, so a regression anywhere broad gets caught and ASV attributes it to a specific subsystem:

- `CpuMuJoCoCartpole` — mujoco_warp on the Warp CPU backend plus Newton's mujoco solver glue. This is the codepath that #2751 actually regressed on.
- `CpuXPBDQuadruped` — XPBD solver, rigid-body contacts, articulations. Entirely separate solver/contact path.
- `CpuIKFranka` — IK solver and the analytic Jacobian. Self-contained module.

Workload sizes are tuned so each runs in well under a second on a warm Warp cache (~1.6 s total measured), keeping the suite cheap.

Other CPU codepaths (VBD/cloth, Featherstone, implicit MPM) are not yet covered. Happy to add any of them if you'd like broader coverage — I left them out only to keep the initial suite minimal per the issue framing.

Kenny-Vilella:
Code looks good to me, benchmark seems fast enough and consistent enough.
They are not running on every PR, but I think it is fair to request every `warp` or `mujoco_warp` update PR to test all the benchmark suite.

For the choice of benchmark I can maybe let @adenzler-nvidia make a decision.

adenzler-nvidia:
Thanks all - I would argue that cartpole is likely not a good stress-test for MuJoCo, as it is quite pure forward kinematics without any solver action. So my vote would be to exchange for a slightly more complex model - a quadruped, ant, or even the mujoco humanoid. Ant is probably enough, and still simple enough to be worthwhile.

vreutskyy:
Thanks @adenzler-nvidia — fair point, cartpole has no contacts and no real solver work.

Swapped to ant in 4a030334. Compared to cartpole, ant has 9 bodies, 14 joint DOFs, a ground plane, and up to 15 max contacts per world — so the constraint solver and contact path actually run. Measured on CPU (warm Warp cache):

- `CpuMuJoCoAnt.time_simulate`: 1.70 s
- `CpuXPBDQuadruped.time_simulate`: 0.36 s
- `CpuIKFranka.time_solve`: 0.37 s

Still cheap. Let me know if you'd prefer humanoid for more solver pressure.

adenzler-nvidia (approved):
The MuJoCo benchmark change looks sound to me. Swapping cartpole for ant gives this CPU benchmark actual contact/constraint solver coverage while keeping the workload small, and the current checks are green. Approving; the Fast/Cpu naming question can be handled as follow-up if we decide these should be selected by the existing PR ASV filter.

PR #2934, opened 2026-05-21
Kamino: accept wp.bool world_mask in SolverKamino.reset

### LINKED ISSUE #2932: SolverKamino.reset world_mask: kernels are strict wp.array[int32] but docstring advertises wp.int8 | wp.bool
## Summary

`SolverKamino.reset(world_mask=...)`'s docstring at [`solver_kamino.py:462`](https://github.com/newton-physics/newton/blob/main/newton/_src/solvers/kamino/solver_kamino.py#L462) advertises:

> `world_mask: ... Shape of (num_worlds,) and type wp.int8 | wp.bool`

but every internal mask-consumer kernel is declared `wp.array[int32]`. Warp's launch-time dtype check is strict, so a caller obeying the documented contract crashes immediately:

```
RuntimeError: Error launching kernel '_extract_actuators_state_from_joints',
  argument 'world_mask' expects an array with dtype=<class 'warp._src.types.int32'>
  but passed array has dtype=<class 'warp._src.types.bool'>.
```

The same failure occurs for `wp.uint8` and `wp.int8`. The view trick (`mask.view(dtype=wp.uint8)`) is fine to construct but Warp still rejects it at launch against an `int32`-typed parameter.

## Where the strict `wp.array[int32]` sites are

13 kernel signatures across:

- `newton/_src/solvers/kamino/_src/kinematics/joints.py:897, 944`
- `newton/_src/solvers/kamino/_src/kinematics/resets.py:48, 68, 102, 165, 217, 320, 342`
- `newton/_src/solvers/kamino/_src/core/bodies.py:432, 459`
- `newton/_src/solvers/kamino/_src/dynamics/delassus.py:619`
- `newton/_src/solvers/kamino/_src/solvers/padmm/kernels.py:83`

Their bodies use the mask only as `world_mask[wid] == 0` or `if not world_mask[wid]`, so widening the type annotation to `wp.array[uint8]` (or `wp.array[bool]`) is a signature-only change — no semantic edits to the kernel bodies, no callers to migrate beyond the `_world_reset_mask` allocation in IsaacLab.

## Why this matters

We are aligning the per-world reset bitmask plumbing across solvers in IsaacLab (the MJWarp NaN-recovery work for https://github.com/newton-physics/newton/issues/1266). `mujoco_warp.reset_data(reset=mask)` strictly requires `wp.array[bool]`:

https://github.com/google-deepmind/mujoco_warp/blob/main/mujoco_warp/_src/io.py#L1471

So MJWarp wants bool natively. With Kamino's current contract, IsaacLab now keeps a Kamino-side `wp.int32` mirror of the base `wp.bool` reset mask and copies bool→int32 every step via an extra kernel launch. That's wasted bandwidth (1B → 4B per world, every step) plus an extra kernel launch and an extra allocation, all to satisfy a dtype mismatch between Kamino's docstring and its own kernels.

It also blocks the cleanest integration with [`wp.capture_while`](https://nvidia.github.io/warp/stable/api_reference/_generated/warp.capture_while.html#warp.capture_while). The natural pattern for "skip the entire per-world reset region inside a captured graph when no world is flagged" is a single `wp.capture_while(any_pending)` wrapper with a 1-byte predicate array. A bool/uint8 mask plugs directly into that pattern — an int32 mask forces an additional reduction-plus-conversion step before the predicate even exists.

## Proposed fix

Widen the 13 strict `wp.array[int32]` kernel signatures to `wp.array[uint8]` (or `wp.array[bool]`, whichever the team prefers for the public mask contract). The fix is mechanical and signature-only.

After this lands, IsaacLab can drop the int32 mirror buffer and pass its `_world_reset_mask` through to `SolverKamino.reset` zero-copy, matching what `mujoco_warp.reset_data` already accepts.

## Repro (IsaacLab side, for reference)

In the IsaacLab repo:

```bash
./isaaclab.sh -p scripts/reinforcement_learning/train.py \
    --rl_library rsl_rl --task Isaac-Cartpole-Direct-v0 \
    physics=newton_kamino --headless --num_envs 256 --max_iterations 5
```

with `_world_reset_mask` allocated as `wp.bool` and passed through to `solver.reset(world_mask=...)` — fails on the first env reset with the kernel-launch dtype error shown above.

cc @vastsoun @adenzler-nvidia

AntoineRichard:
## Description

`SolverKamino.reset(world_mask=...)`'s docstring advertises `wp.int8 | wp.bool` but every per-world reset kernel is declared `wp.array[int32]`, so a caller obeying the documented contract crashes at kernel launch. This PR widens the kamino reset path to `wp.bool`, matching what `mujoco_warp.reset_data` already accepts and unblocking IsaacLab (issue #1266) from keeping an int32 mirror of its `_world_reset_mask`.

The 13 strict kernel signatures listed in the issue (plus the 2 sibling `mask:` reset kernels in `resets.py`) and `_all_worlds_mask` switch to `wp.bool`. The FK solver entry points (`run_fk_solve` / `solve_fk`) accept `wp.array[bool]` and cast into the internal int32 `newton_mask` via a small `_cast_world_mask_to_int32` kernel, leaving the FK solver internals and shared linalg ops untouched. Example, test, and simulator docstrings/allocations follow.

Closes #2932

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -s newton/_src/solvers/kamino/tests/                                       # 429 tests pass
uv run --extra dev -m newton.tests -s newton/_src/solvers/kamino/tests/ -p test_solver_kamino.py              # 14 pass, includes the reset-with-world-mask cases
uv run --extra dev -m newton.tests -s newton/_src/solvers/kamino/tests/ -p test_solvers_forward_kinematics.py # 6 pass
uvx pre-commit run --files <changed files>                                                                    # clean
```

Manual: `example_sim_basics_box_on_plane --headless --num-steps 50` runs cleanly with the bool-typed `_all_worlds_mask`.

## Summary by CodeRabbit

* **Breaking Changes**
  * Solver APIs now require world_mask arrays of boolean dtype instead of integer masks. Migrate callers to pass boolean masks (e.g., wp.ones(..., dtype=wp.bool) or wp.array([...], dtype=wp.bool)).

* **Documentation**
  * Solver and simulator docstrings updated to reflect boolean world_mask semantics.

* **Examples & Tests**
  * Examples and unit tests updated to use boolean world masks for selective resets.

* **Behavior**
  * Forward kinematics and related solver entry points accept boolean masks and propagate boolean masking through internal solver paths.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2934?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


newton/_src/solvers/kamino/_src/solvers/fk/kernels.py
```diff
@@ -99,6 +100,29 @@ def read_quat_from_array(array: wp.array[wp.float32], offset: int, normalize: bo
...
+    to the int32 mask buffer consumed by the internal velocity-solve kernels and the
+    linalg ``world_active`` parameter.
+
+    Inputs:
+        world_mask: Per-world bool flag (True = process this world).
+    Outputs:
+        out: Per-world int32 flag (1 = process, 0 = skip).
+    """
+    wid = wp.tid()
+    out[wid] = wp.int32(world_mask[wid])
```
adenzler-nvidia: I think it's ok to just change the internal kernels as well? or is there a more fundamental issue?

adenzler-nvidia:
otherwise this looks ready, can we move it out of draft?

AntoineRichard:
No, no good reasons, we can modfiy all the Kamino kernels to take bools instead. I just wanted to limit the scope of the changes, but it doesn't look that bad in the end.

newton/_src/solvers/kamino/_src/solvers/fk/kernels.py
Guirec-Maloisel: I'd be fine with the internal FK kernels being updated as well, or would even prefer it for consistency. I can help with this if needed (I'm the main author of that part of the code)

AntoineRichard: I did these changes a couple of days ago. @Guirec-Maloisel feel free to take a look. I kept the kernels permissive in some areas with `Any` as an input so they can ingest either int32 or bool

newton/_src/solvers/kamino/_src/solvers/fk/solver.py
```diff
@@ -1984,7 +1990,7 @@ def run_fk_solve(
             self._update_gradient(bodies_q, self.newton_mask)
         self._eval_max_residual(self.constraints, self.grad, self.max_residual)
         self.newton_loop_condition.zero_()
-        wp.copy(self.line_search_success, self.newton_mask)  # To disregard line search success in initial Newton check
+        self.line_search_success.fill_(1)  # To disregard line search success in initial Newton check
```
Guirec-Maloisel: Please revert this change, as `newton_mask` contains the information on which worlds in which there is no need to solve FK at all (as it is initialized from `world_mask`). It is important to propagate this information to `line_search_success` as well, as this is used later to filter worlds in which to apply the line search step. Filling this with 1 could lead to worlds being modified that shouldn't.

newton/_src/solvers/kamino/_src/linalg/factorize/llt_blocked_semi_sparse.py
```diff
@@ -75,12 +76,12 @@ def reorder_rows_kernel(
     ordering: wp.array2d[int],
     n_rows_arr: wp.array[int],
     n_cols_arr: wp.array[int],
-    batch_mask: wp.array[int],
+    batch_mask: wp.array[Any],
```
Guirec-Maloisel: I was wondering if there was a reason for using Any instead of bool, here and in a couple other places? Also, I noticed usage of `bool` in some places vs `wp.bool` elsewhere. Shouldn't we just use `wp.bool` everywhere?

Guirec-Maloisel: I just saw your message above about keeping the kernels more permissive with Any in some places. But I think the question remains (are there specific things that break without keeping the kernels permissive? what motivated keeping these specific kernels permissive?)

newton/_src/solvers/kamino/_src/solvers/fk/solver.py
AntoineRichard: Ha right sorry, we'll need to launch a kernel to do the update, I try to be frugal on kernel launches but I forget everything is graphed here. Should be addressed now.

Guirec-Maloisel: I think the additional kernel could be avoided by making also `line_search_success` a boolean array. This one as well as `newton_success`, `line_search_loop_condition` and `newton_loop_condition` are all 0-1 arrays. The last two must stay integer arrays because capture_while() requires integer arrays for the condition; but I think both `line_search_success` and  `newton_success` can be changed to boolean arrays to avoid the extra kernels and array.

(if this doesn't work out I'd be fine merging in the current state, and having a look myself later)

Guirec-Maloisel (dismissed):
Looks good to me now!

AntoineRichard:
> * #### [pre-commit.ci - pr](https://results.pre-commit.ci/run/github/970467647/1780575882.7Szf5MhjTwquVml7bUY8FQ) — error during mergeable check

I'll address the conflicts and it should be good to go. Thanks @Guirec-Maloisel

AntoineRichard:
While resolving the latest `main` merge, the FK solver conflict came from two independent changes: `main` renamed the internal FK transform buffer/path from `pos_control_transforms` to `target_rel_transforms`, while this PR changes the per-world masks to bool arrays.

The small test fixture update is only to mirror that upstream rename. `test_initial_line_search_success_honors_world_mask` constructs a `ForwardKinematicsSolver` with `__new__` and bypasses `finalize()`, so it has to provide the preallocated `target_rel_transforms` buffer and stub `_eval_target_relative_transformations()` explicitly. That keeps the test focused on the original regression: skipped worlds must not look like successful line-search candidates.

Guirec-Maloisel (approved):

PR #2944, opened 2026-05-22
Preserve API SPDX years

adenzler-nvidia:
## Description

Preserve correct SPDX copyright years when regenerating API docs.

`docs/generate_api.py` previously hard-coded `2025` into every generated
`docs/api/*.rst` file. That is wrong for newly generated API pages created in
later years, and it can also overwrite the original creation year for existing
generated files.

This change snapshots existing generated API page SPDX lines before deleting
`docs/api/`, reuses those lines for regenerated files, and uses the current year
only for genuinely new generated pages.

## Checklist

- [x] `CHANGELOG.md` has been updated (if user-facing change; not applicable)

## Test plan

```bash
uvx pre-commit run -a
```

Verified that rerunning `docs/generate_api.py` does not produce unrelated
`docs/api` diffs.

## Bug fix

**Steps to reproduce:**

1. Add a new public API module or symbol that causes `docs/generate_api.py` to
   create a new `docs/api/*.rst` file in 2026 or later.
2. Run:

   ```bash
   ```

3. Observe that the new generated page receives a hard-coded 2025 SPDX year
   instead of the file creation year.

**Minimal reproduction:**

```python
from pathlib import Path

source = Path("docs/generate_api.py").read_text(encoding="utf-8")
assert '".. SPDX-FileCopyrightText: Copyright (c) 2025 The Newton Developers"' in source
```

## Summary by CodeRabbit

* **Chores**
  * API documentation generation now preserves existing SPDX copyright lines when regenerating files and falls back to the current year for newly generated pages.

* **Documentation**
  * API docs headers updated to reflect the current copyright year.

* **Tests**
  * Added tests covering copyright preservation and year-fallback behavior during API documentation generation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2944?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/api/_toctree.rst
+++ b/docs/api/_toctree.rst
@@ -1,4 +1,4 @@
-.. SPDX-FileCopyrightText: Copyright (c) 2025 The Newton Developers
+.. SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
 .. SPDX-License-Identifier: CC-BY-4.0
 
 .. toctree::
--- a/docs/api/newton_actuators.rst
+++ b/docs/api/newton_actuators.rst
@@ -1,4 +1,4 @@
-.. SPDX-FileCopyrightText: Copyright (c) 2025 The Newton Developers
+.. SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
 .. SPDX-License-Identifier: CC-BY-4.0
 
 newton.actuators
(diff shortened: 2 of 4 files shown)


preist-nvidia (dismissed):
This seems a bit overkill for a couple of generated and tracked doc pages, but it IS correct according to our copyright year policy and tested, so I'm ok with the change.

If we are doing this, it may make sense to double-check year correctness of the currently tracked files - actuators is wrong for sure. But not critical so approving without

eric-heiden (approved):

PR #2949, opened 2026-05-22
safe warp int32 conversion in Kamino

rubengrandia:
## Description

Conversions from Numpy int64 to a warp array of in32 silently overflows. This PR introduces 2 helpers that check that the conversion is done safely and raises an Error in case an overflow would occur. 

Concretely, this catches the otherwise hard-to-debug case of the Jacobian offset overflowing due to the way contact buffers are assigned per world.

## Checklist

- [x] The documentation is up to date with these changes. (No changes required)

## Test plan

Tests that show the helpers work:
     ```
     uv run --extra dev python -m unittest -v newton._src.solvers.kamino.tests.test_core_types
     ``` -->


## Summary by CodeRabbit

* **Bug Fixes**
  * SolverKamino now raises an OverflowError when integer inputs exceed the supported 32-bit range during model construction, preventing silent truncation and ensuring incorrect inputs are rejected.

* **Tests**
  * Added comprehensive unit tests validating integer-array conversion and assignment behavior, including range checks and type error handling.

<!-- review_stack_entry_start -->

[![Review Change Stack](https://storage.googleapis.com/coderabbit_public_assets/review-stack-in-coderabbit-ui.svg)](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2949?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

<!-- review_stack_entry_end -->


Guirec-Maloisel (dismissed):
Looks good to me! please just fix the issues flagged by code rabbit (for the one in animation.py we should just add an explicit device to the array allocation, seems like it was forgotten before)

Guirec-Maloisel (dismissed):

chschuma-disney (commented):
Thanks for pushing a fix for this. This looks good to me, given the intended target.

Long term, I think it would be great to have the model dimensions checked before assembling all data structures, which would also let us display an appropriate error message that users understand.

Also, to get the checks to pass, you might have to set your profile to `public` here: https://github.com/orgs/newton-physics/people

A small note: I don't think we technically need a changelog entry. We've avoided them in the past with the justification that Kamino is in beta.

Guirec-Maloisel (approved):

PR #2950, opened 2026-05-22
Allow ViewerUsd to properly show particles

### LINKED ISSUE #917: Issue with `show_particles` in `ViewerUSD`
The cloth particles appear to be some amorphous blob. In the old USDRenderer, the particles would be individually shown as spheres.

This is with `python -m newton.examples cloth_franka --viewer usd` on the latest commit f0b6dbf035ca50a660b8d47cf9d3679c5c862920 and the following change:

```python
        self.viewer.show_particles = True
```
now:
Image

before:
Image

aneangel:
## Description

closes #917 

## Checklist

## Test plan

`uv run --extra dev -m newton.tests -k test_viewer_usd`
`uv run --extra dev -m newton.tests -k test_viewer_particle_flags`

## Bug fix

**Steps to reproduce:**

1. Run python -m newton.examples cloth_franka with a ViewerUSD or GL viewer that has show_particles enabled
2. Observe particles rendered as ~0.8-metre spheres instead of ~0.008-metre spheres; 100× too large because the example simulates in centimetre scale but model.particle_radius was passed unscaled to the viewer
3. Separately: run any simulation in ViewerUSD where particles transition from active to all-inactive; observe that sphere instances from the previous active frame remain visible instead of being hidden

**Minimal reproduction:**

```
import tempfile
import warp as wp
from newton.viewer import ViewerUSD

# ViewerUSD leaves stale particle spheres visible when all particles become inactive
with tempfile.NamedTemporaryFile(suffix=".usda") as f:
    viewer = ViewerUSD(output_path=f.name, num_frames=2)
    points = wp.array([[0.0, 0.0, 0.0], [0.1, 0.0, 0.0]], dtype=wp.vec3)

    viewer.begin_frame(0.0)
    viewer.log_points("/particles", points, radii=0.01, as_spheres=True)
    # All particles become inactive — should hide the instancer:
    viewer.log_points("/particles", None, as_spheres=True, hidden=True)
    # Before fix: PointInstancer visibility remains "inherited" (still visible)
    # After fix: PointInstancer visibility is set to "invisible"

```

## Summary by CodeRabbit

* **New Features**
  * Added an `as_spheres` option to `log_points` to request 3D sphere-instance rendering in supported viewers for better particle visuals.

* **Bug Fixes**
  * Active-particle rendering is now filtered to avoid logging inactive particles.
  * `ViewerUSD` hides the instancer when there are zero active particles (prevents stale visible spheres).
  * Corrected particle visualization scale in the cloth example to match meter-based viewer sizing.

* **Tests**
  * Extended viewer particle-flag tests and added a USD instancer visibility test for the `points=None` path.

* **Examples**
  * Cloth example now computes visualization-only particle radii and restores simulation values afterward.


aneangel:
@coderabbitai review

adenzler-nvidia:
Thanks for picking this up. I'd like to suggest restructuring before this lands.

**Split into smaller, independent PRs.** This PR bundles three unrelated changes:

1. `cloth_franka` particle radius scaling (cm → m)
2. `ViewerUSD` not hiding stale spheres when active particle count drops to 0
3. New `as_spheres` API on `Viewer.log_points`

(1) and (2) are clean, well-scoped bug fixes that can land on their own immediately. (3) is the part that needs discussion. Splitting also lets the bug fixes get backported cleanly if needed.

**On the `as_spheres` API:** I'd push back on adding it to the abstract `Viewer.log_points`. Only `ViewerUSD` uses it — the other six backends accept and ignore it, with docstrings literally saying *"Ignored — …"*. Two cleaner options:

- **A.** Always render points as instanced spheres in `ViewerUSD` (matches what every other backend already does); deprecate the flat-splat path.
- **B.** Move the toggle to a `ViewerUSD` constructor option (`ViewerUSD(..., points_as_spheres=True)`) so it's a per-viewer rendering preference, not a cross-backend API parameter.

Either keeps the abstract API clean and avoids future backends inheriting an obligation to ignore the flag.

**Two smaller things** if (3) returns in a follow-up:

- The new `-> Sdf.Path | None` return on `ViewerUSD.log_points` leaks a USD type through the abstract API (no other backend returns anything). The new test relies on it; it could verify state via `viewer.stage` / `viewer._instancers` directly instead.
- The `points=None` branch unconditionally hides the prim regardless of the `hidden` argument, and the two `points=None` sub-branches are asymmetric. Cleanest is to handle the empty-particle case at the call site in `_log_particles` (pass an empty array with `hidden=True`) rather than overloading `log_points(points=None)`.

Happy to chat through any of this.

aneangel:
@adenzler-nvidia I agree that option A makes sense the most sense regarding the API, I think I was trying too much to address the issue without having to break the APIs current struct. I will go ahead and split this PR into three that way each of the bundled issues have their own PR that can be tracked.

aneangel:
@adenzler-nvidia I went ahead and split them into smaller PRs for each. Let me know if something doesn't look right in any one of the three PRs, thanks!

jcarius-nv:

newton/_src/viewer/viewer.py
```diff
@@ -1222,6 +1222,7 @@ def log_points(
         radii: wp.array[wp.float32] | float | None = None,
         colors: (wp.array[wp.vec3] | wp.array[wp.float32] | tuple[float, float, float] | list[float] | None) = None,
         hidden: bool = False,
+        as_spheres: bool = False,
```
eric-heiden, line 1675: Let's get https://github.com/newton-physics/newton/pull/2990 merged first, then those changes don't land in this PR.

adenzler-nvidia:
Closing as superseded: all of this PR's changes have landed through the split-off PRs #2990 (`points_as_spheres` option for `ViewerUSD`), #2992 (stale particle geom fix), and #2991 (cloth_franka particle radius scale fix). After merging the latest main, this branch has no remaining diff. Thanks @aneangel for splitting these up!

PR #2951, opened 2026-05-23
Add USD parsing for NewtonMassAPI

andrewkaufman:
## Description

Add USD import support for `NewtonMassAPI` attributes: shell mass model (`newton:massModel`), shell thickness (`newton:shellThickness`), and compact 6-element inertia tensor (`newton:inertia`).

Requires `newton-usd-schemas>=0.3.0b1` which ships the final `NewtonMassAPI` schema.

### Shell mass model

When `newton:massModel` is set to `"shell"`, the importer computes shell (hollow) inertia for collider shapes instead of solid inertia. Shell thickness is sourced from `newton:shellThickness` when authored, falling back to the collision margin. The `_build_mass_info_from_shape_geometry` helper is now shell-aware so the density-derived mass computer path also produces correct shell inertia.

### Compact inertia tensor

`newton:inertia` accepts a 6-element array `[Ixx, Iyy, Izz, Ixy, Ixz, Iyz]` and takes precedence over `physics:diagonalInertia` + `physics:principalAxes`. Validation rejects tensors with wrong length, non-finite values, or negative diagonal elements, emitting warnings.

### SchemaResolver integration

`mass_model` and `shell_thickness` are surfaced through the `SchemaResolver` system so that MuJoCo-exported USD files with `mjc:shellinertia` are automatically resolved to the shell mass model without requiring `NewtonMassAPI` to be authored.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_newton_mass_api_parsing
uv run --extra dev -m newton.tests -k test_newton_inertia_tensor_validation
uv run --extra dev -m newton.tests -k test_mass_model
```

- `test_newton_mass_api_parsing` — 7 bodies covering all combinations of mass source (authored / density-derived) x inertia source (explicit tensor / diagonal / accumulated) x mass model (solid / shell with thickness / shell with margin fallback)
- `test_newton_inertia_tensor_validation` — 3 bodies with malformed `newton:inertia` (wrong length, non-finite values, negative diagonal) verifying warnings are emitted and tensors are ignored
- `test_mass_model` — SchemaResolver unit tests for `newton:massModel` and `mjc:shellinertia` resolution, priority ordering, and full-import integration verifying `is_solid` on the resulting shape

## New feature / API change

```python
import newton

# Shell mass model via USD attribute
# On a collider prim: newton:massModel = "shell", newton:shellThickness = 0.01
# The importer will compute hollow-shell inertia instead of solid inertia.

# Compact inertia tensor via USD attribute
# On a body prim: newton:inertia = [Ixx, Iyy, Izz, Ixy, Ixz, Iyz]
# Takes precedence over physics:diagonalInertia + physics:principalAxes.

# MuJoCo interop: mjc:shellinertia = true on a geom prim is automatically
# resolved to shell mass model through the SchemaResolver system.
```

### DIFF
--- a/docs/concepts/usd_parsing.rst
+++ b/docs/concepts/usd_parsing.rst
@@ -78,6 +78,8 @@ For rigid bodies with ``UsdPhysics.MassAPI`` applied, Newton resolves each inert
    is used.
 3. When ``physics:mass`` is authored but inertia is not, the inertia
    accumulated from collision shapes is scaled by ``authored_mass / accumulated_mass``.
+   Shell colliders (``newton:massModel = "shell"``) contribute shell-derived inertia to the
+   accumulation before this scaling is applied.
 4. For any remaining unresolved properties, Newton falls back to
    ``UsdPhysics.RigidBodyAPI.ComputeMassProperties(...)``.
    In this fallback path, collider contributions use a two-level precedence:
--- a/docs/concepts/usd_parsing.rst
+++ b/docs/concepts/usd_parsing.rst
@@ -70,18 +70,26 @@ For rigid bodies with ``UsdPhysics.MassAPI`` applied, Newton resolves each inert
 (mass, inertia, center of mass) independently.  Authored attributes take precedence;
 ``UsdPhysics.RigidBodyAPI.ComputeMassProperties(...)`` provides baseline values for the rest.
 
-1. Authored ``physics:mass``, ``physics:diagonalInertia``, and ``physics:centerOfMass`` are
+1. ``newton:inertia`` (from ``NewtonMassAPI``) is a compact 6-element symmetric tensor
+   ``[Ixx, Iyy, Izz, Ixy, Ixz, Iyz]`` already in the body frame.  When authored, it
+   overrides ``physics:diagonalInertia`` and ``physics:principalAxes``.
+2. Authored ``physics:mass``, ``physics:diagonalInertia``, and ``physics:centerOfMass`` are
    applied directly when present.  If ``physics:principalAxes`` is missing, identity rotation
    is used.
-2. When ``physics:mass`` is authored but ``physics:diagonalInertia`` is not, the inertia
+3. When ``physics:mass`` is authored but inertia is not, the inertia
    accumulated from collision shapes is scaled by ``authored_mass / accumulated_mass``.
-3. For any remaining unresolved properties, Newton falls back to
+4. For any remaining unresolved properties, Newton falls back to
    ``UsdPhysics.RigidBodyAPI.ComputeMassProperties(...)``.
    In this fallback path, collider contributions use a two-level precedence:
 
    a. If collider ``UsdPhysics.MassAPI`` has authored ``mass`` and ``diagonalInertia``, those
       authored values are converted to unit-density collider mass information.
-   b. Otherwise, Newton derives unit-density collider mass information from collider geometry.
+   b. Otherwise, Newton derives unit-density collider mass information from collider
+      geometry.  When ``NewtonMassAPI`` is applied to the collider, ``newton:massModel``
+      controls whether inertia is derived from the full volume (``"solid"``, default) or a
+      thin shell at the surface (``"shell"``).  For shell shapes,
+      ``newton:shellThickness`` sets the wall thickness [m] measured inward from the outer
+      surface; the sentinel ``-inf`` (default) falls back to ``newton:contactMargin``.
 
    A collider is skipped (with warning) only if neither path provides usable collider mass
    information.
--- a/newton/_src/usd/schemas.py
+++ b/newton/_src/usd/schemas.py
@@ -87,6 +87,9 @@ class SchemaResolverNewton(SchemaResolver):
             # Contact stiffness/damping
             "ke": SchemaAttribute("newton:contact_ke", None),
             "kd": SchemaAttribute("newton:contact_kd", None),
+            # Mass model
+            "mass_model": SchemaAttribute("newton:massModel", "solid"),
+            "shell_thickness": SchemaAttribute("newton:shellThickness"),
         },
         PrimType.BODY: {},
         PrimType.ARTICULATION: {
@@ -346,6 +349,8 @@ class SchemaResolverMjc(SchemaResolver):
             # Contact stiffness/damping from per-geom solref
             "ke": SchemaAttribute("mjc:solref", [0.02, 1.0], solref_to_stiffness),
             "kd": SchemaAttribute("mjc:solref", [0.02, 1.0], solref_to_damping),
+            # Mass model: mjc:shellinertia (bool) → "shell" / "solid"
+            "mass_model": SchemaAttribute("mjc:shellinertia", False, lambda v: "shell" if v else "solid"),
         },
         PrimType.MATERIAL: {
             # Materials
--- a/newton/_src/usd/schemas.py
+++ b/newton/_src/usd/schemas.py
@@ -89,7 +89,7 @@ class SchemaResolverNewton(SchemaResolver):
             "kd": SchemaAttribute("newton:contact_kd", None),
             # Mass model
             "mass_model": SchemaAttribute("newton:massModel", "solid"),
-            "shell_thickness": SchemaAttribute("newton:shellThickness"),
+            "shell_thickness": SchemaAttribute("newton:shellThickness", float("-inf")),
         },
         PrimType.BODY: {},
         PrimType.ARTICULATION: {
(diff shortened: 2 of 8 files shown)


newton/_src/utils/import_usd.py
```diff
@@ -2651,6 +2654,25 @@ def _build_mass_info_from_shape_geometry(
...
 
                 shape_color = material_props.get("color")
+
+                # Mass model and shell thickness (resolved across Newton / MuJoCo schemas)
+                mass_model = R.get_value(prim, PrimType.SHAPE, "mass_model", default="solid")
+                shape_is_solid = mass_model != "shell"
+                shell_thickness_val = R.get_value(prim, PrimType.SHAPE, "shell_thickness")
+                # When shell thickness is authored, pass it as margin so compute_inertia_shape
+                # uses the correct thickness. The real collision margin is restored after add_shape.
+                if shell_thickness_val is not None and math.isfinite(float(shell_thickness_val)):
```
eric-heiden, line 2689: 🟡 The tests should cover the common `newton:inertia` path without also authoring `physics:diagonalInertia`.

Right now the non-singular tensor case also sets `physics:diagonalInertia`, so it only proves that `newton:inertia` wins over the diagonal inertia value. Please add a case in `newton/tests/test_import_usd.py` with `NewtonMassAPI`, `PhysicsMassAPI`, authored `physics:mass`, and `newton:inertia = [1.0, 2.0, 3.0, 0.1, 0.2, 0.3]`, but no `physics:diagonalInertia`. The assertion should verify that the imported inertia matches the authored tensor exactly and is not scaled by mass.

docs/concepts/usd_parsing.rst
```diff
@@ -70,18 +70,26 @@ For rigid bodies with ``UsdPhysics.MassAPI`` applied, Newton resolves each inert
...
 
    a. If collider ``UsdPhysics.MassAPI`` has authored ``mass`` and ``diagonalInertia``, those
       authored values are converted to unit-density collider mass information.
-   b. Otherwise, Newton derives unit-density collider mass information from collider geometry.
+   b. Otherwise, Newton derives unit-density collider mass information from collider
+      geometry.  When ``NewtonMassAPI`` is applied to the collider, ``newton:massModel``
+      controls whether inertia is derived from the full volume (``"solid"``, default) or a
+      thin shell at the surface (``"shell"``).  For shell shapes,
+      ``newton:shellThickness`` sets the wall thickness [m] measured inward from the outer
+      surface; the sentinel ``-inf`` (default) falls back to ``newton:contactMargin``.
```
eric-heiden, line 94: 🟡 This documents `-inf` as the default sentinel for `newton:shellThickness`, but the resolver entry in `newton/_src/usd/schemas.py` currently registers `SchemaAttribute("newton:shellThickness")` without that default.

That means `SchemaResolverManager.get_value(..., "shell_thickness")` returns `None` for unauthored attributes, while the schema and docs say the default is `-inf`. Please set the resolver default to `float("-inf")` and keep the importer fallback for non-finite values.

newton/tests/test_import_usd.py
```diff
@@ -3711,6 +3711,427 @@ def Cube "Site2" (
...
+    prepend apiSchemas = ["PhysicsRigidBodyAPI", "PhysicsMassAPI"]
+)
+{
+    double3 xformOp:translate = (2, 0, 1)
+    uniform token[] xformOpOrder = ["xformOp:translate"]
+
+    float physics:mass = 10.0
+
+    def Sphere "Collider" (
+        prepend apiSchemas = ["PhysicsCollisionAPI", "NewtonMassAPI"]
```
eric-heiden: 🟡 This fixture authors `newton:contactMargin`, but the collider does not apply `NewtonCollisionAPI`.

The resolver can still read the raw attribute by name, so the test passes, but the USD fixture is schema-inconsistent and can mislead users about which API schema is needed for the margin fallback path.

```suggestion
        prepend apiSchemas = ["PhysicsCollisionAPI", "NewtonCollisionAPI", "NewtonMassAPI"]
```

docs/concepts/usd_parsing.rst
```diff
@@ -70,18 +70,26 @@ For rigid bodies with ``UsdPhysics.MassAPI`` applied, Newton resolves each inert
...
+2. Authored ``physics:mass``, ``physics:diagonalInertia``, and ``physics:centerOfMass`` are
    applied directly when present.  If ``physics:principalAxes`` is missing, identity rotation
    is used.
-2. When ``physics:mass`` is authored but ``physics:diagonalInertia`` is not, the inertia
+3. When ``physics:mass`` is authored but inertia is not, the inertia
    accumulated from collision shapes is scaled by ``authored_mass / accumulated_mass``.
-3. For any remaining unresolved properties, Newton falls back to
+4. For any remaining unresolved properties, Newton falls back to
    ``UsdPhysics.RigidBodyAPI.ComputeMassProperties(...)``.
    In this fallback path, collider contributions use a two-level precedence:
```
eric-heiden, line 85: 🟡 The shell behavior is only described under the `ComputeMassProperties` fallback, but it also affects the preceding `physics:mass` path.

When `physics:mass` is authored and inertia is not, Newton scales the inertia accumulated from collision shapes, and that accumulated inertia has already used each collider's `newton:massModel` and `newton:shellThickness`. Please add a sentence to path 3 clarifying that shell colliders produce shell-derived accumulated inertia before the `authored_mass / accumulated_mass` scaling.

docs/concepts/usd_parsing.rst
andrewkaufman: `-inf` is just the USD version of `None` though. I can do this but we shouldn't really think of them as being different things.

pyproject.toml
```diff
@@ -69,7 +69,7 @@ importers = [
     "usd-exchange>=2.2.0 ; platform_machine == 'aarch64' and python_version < '3.13'",
 
     # Newton USD Schemas
-    "newton-usd-schemas>=0.2.0",
+    "newton-usd-schemas>=0.3.0b1",
```
eric-heiden: @andrewkaufman can you check this Codex comment - just double-checking whether we really want to depend on a beta release here:

🔴 This raises `newton-usd-schemas` to a beta pre-release without an upper bound.

`>=0.3.0b1` opts this dependency into later beta builds, so an install without `uv.lock` can silently pick up `0.3.0b2` or the eventual `0.3.0` final. If the schema changes any of the exact names or tensor ordering this importer now depends on, users can get broken mass parsing from a normal install.

Please either cap the beta range until the final schema is available, or wait for the stable `0.3.0` release before merging this dependency bump.

```suggestion
    "newton-usd-schemas>=0.3.0b1,<0.3.0b2",
```

newton/_src/usd/schemas.py
```diff
@@ -87,6 +87,9 @@ class SchemaResolverNewton(SchemaResolver):
             # Contact stiffness/damping
             "ke": SchemaAttribute("newton:contact_ke", None),
             "kd": SchemaAttribute("newton:contact_kd", None),
+            # Mass model
+            "mass_model": SchemaAttribute("newton:massModel", "solid"),
+            "shell_thickness": SchemaAttribute("newton:shellThickness", float("-inf")),
```
eric-heiden, line 92: ⚪ This default makes `shell_thickness_val is not None` true for every collider when the Newton resolver is active.

That means the restore block in `newton/_src/utils/import_usd.py:2810-2814` writes `builder.shape_margin[shape_id] = margin_val` even when no shell thickness was substituted. The write is harmless, but it makes the restore comment misleading and hides the intended condition.

pyproject.toml
andrewkaufman: > or wait for the stable 0.3.0 release before merging this dependency bump.

I don't think that is practical. We need to land 3 conflicting PRs in 2 working days. I would suggested we accept the beta bump, merge these 2 now, and acknowledge that this will be fixed (0.3.0 will be released) before 1.3 releases.

>  so an install without uv.lock can silently pick up 0.3.0b2 or the eventual 0.3.0 final.

Also, that is desirable behaviour. I don't know what codex is complaining about here. If a user didn't want that they'd set `==` for the dependency.

newton/_src/usd/schemas.py
andrewkaufman: Added in [546a433](https://github.com/newton-physics/newton/pull/2951/commits/546a43349d76d854521abcd4a127a19e0894399b)

pyproject.toml
andrewkaufman: 0.3.0 is released now so I updated it here

eric-heiden (approved):

PR #2959, opened 2026-05-27
Convert MJC equality constraints to joints on import by default

### LINKED ISSUE #2665: Import MJC equality constraints as joints
Deprecate the `ModelBuilder.add_equality_constraint*()` methods - these are only needed for MuJoCo and therefore should be treated as a solver-specific concept to be handled by custom attributes and frequencies.

To allow any Newton solver to handle MJC equality constraints, the MJCF and USD importers should add a flag like `convert_mjc_equality_constraints` (defaults to True) that will trigger the following conversion:
* MjcEqualityJointAPI → Newton’s mimic constraint
* MjcEqualityWeldAPI → Fixed joint
* MjcEqualityConnectAPI → Ball joint

Note that tendon equality constraints are not yet supported in MJWarp, so we ignore them for now.

The converted joints should be added after the regular joints in each articulation. They will be treated like the loop-closing joints from UsdPhysics that have the `excludeFromArticulation` flag set to true. We may consider adding a sorting method to the ModelBuilder at a later time that will move such loop-closing joints after the regular joints for each articulation, but for now we just enforce this convention in the importers.
If this `convert_mjc_equality_constraints` flag is false we keep the mujoco-specific equality constraints as before without converting them to joints by using the equality constraint custom frequency defined by `SolverMuJoCo`. This is meant as an option that ensures the MuJoCo models are imported with exactly the same number of joints in MuJoCo and Newton for expert users who need this.

Introduce `Model(Builder).articulation_end` to denote the index of the last regular joint of an articulation.
That means the regular joints appear in indices `articulation_start[i]:articulation_end[i]` where `i` is the index of the articulation. Loop-closing joints appear at indices `articulation_end[i]:articulation_start[i+1]`.

`ArticulationView` by default should not include joints that have `joint_articulation == -1` (loop-closing joints). This should be another setting on `ArticulationView`.

eric-heiden:
## Description

Closes #2665.

This PR converts MuJoCo equality constraints imported from MJCF and USD into Newton loop joints or mimic constraints by default, while preserving the MuJoCo equality metadata needed for `SolverMuJoCo` to recreate the compiled equality rows without losing data. It adds `convert_mjc_equality_constraints=False` as the escape hatch for the legacy `ModelBuilder.add_equality_constraint*()` storage path and keeps those methods intact.

It also adds `ModelBuilder.articulation_end` / `Model.articulation_end` so regular articulation traversal excludes converted loop-closing joints, and adds `ArticulationView(..., include_loop_closing_joints=False)` so loop joints remain opt-in for selection.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_kinematics
uv run --extra dev -m newton.tests -k test_mjc_equality_conversion_roundtrips_to_mujoco
uvx pre-commit run -a
```

## Bug fix

Without this PR, converting imported MuJoCo equality constraints through Newton joints/mimics loses authored MuJoCo data such as higher-order joint equality coefficients, solver parameters, weld relposes, torque scale, and disabled state. `SolverMuJoCo` therefore cannot reconstruct equivalent MuJoCo equality constraints from the converted Newton representation.

**Steps to reproduce:**

1. Import an MJCF or USD model containing MuJoCo connect, weld, and joint equalities with non-default `polycoef`, `solref`, `solimp`, weld `relpose`, `torquescale`, and disabled equalities.
2. Convert equality constraints to Newton loop joints or mimic constraints.
3. Instantiate `SolverMuJoCo`.
4. Compare the compiled MuJoCo equality rows against the legacy equality-array import path.

**Minimal reproduction:**

```python
import newton

builder.add_mjcf(mjcf_with_mujoco_equalities)
model = builder.finalize()
solver = SolverMuJoCo(model)

# solver.mj_model.eq_data, eq_solref, eq_solimp, active flags, and refs
# now match the legacy equality-array import path.
```

## New feature / API change

```python
import newton

builder.add_mjcf(mjcf_text)  # converts MuJoCo equalities by default

legacy_builder = newton.ModelBuilder()
legacy_builder.add_mjcf(mjcf_text, convert_mjc_equality_constraints=False)
```

## Summary by CodeRabbit

* **New Features**
  * MuJoCo equality constraints now convert by default into native loop joints and mimic constraints during MJCF/USD import while preserving MuJoCo metadata for faithful round-trips.
  * ArticulationView gains an opt-in option to include loop-closing joints in selections.

* **Changed**
  * MJCF and USD importers add a new convert_mjc_equality_constraints flag (default: True) to opt out and preserve legacy equality behavior.
  * Articulation indexing now uses explicit per-articulation end bounds.

* **Tests**
  * New and updated tests validate equality-conversion round-trips, preserved mimic runtime behavior, and selection opt-in.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2959?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


newton/_src/sim/builder.py
```diff
@@ -2430,6 +2433,7 @@ def add_usd(
         mesh_maxhullvert: int | None = None,
         schema_resolvers: list[SchemaResolver] | None = None,
         force_position_velocity_actuation: bool = False,
+        convert_mjc_equality_constraints: bool = True,
```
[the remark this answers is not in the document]
eric-heiden: At some point we have to switch over to this new behavior, so this breaking change is inevitable. We prefer to do this now.

newton/_src/utils/import_mjcf.py
```diff
@@ -2056,17 +2207,27 @@ def get_site_body_and_anchor(site_name: str) -> tuple[int, wp.vec3] | None:
...
+                    builder.add_constraint_mimic(
+                        joint0=joint1_idx,
+                        joint1=joint2_idx,
+                        coef0=polycoef_values[0],
+                        coef1=polycoef_values[1],
+                        label=equality_label(common),
+                        enabled=common["active"],
+                        custom_attributes=mimic_eq_custom_attrs(polycoef_values, custom_attrs),
+                    )
+                else:
```
vreutskyy: MuJoCo `joint` equalities support a degree-4 polycoef, but only `coef0`/`coef1` reach the runtime mimic here. The full polycoef is preserved for `SolverMuJoCo` via `mujoco:mimic_eq_polycoef`, but Featherstone/XPBD/etc. would only see the linear part. Should we `warnings.warn` (or at least set a `verbose` log) when `polycoef[2:5]` is non-zero so the loss isn't silent for non-MuJoCo backends? Same question for the USD path.

newton/_src/sim/builder.py
```diff
@@ -8843,9 +8863,23 @@ def _finalize_imported_articulation(
...
+                    if joint_idx not in imported_joints and self.joint_articulation[joint_idx] == -1:
+                        joint_name = (
+                            self.joint_label[joint_idx] if joint_idx < len(self.joint_label) else f"#{joint_idx}"
+                        )
+                        raise ValueError(
+                            f"Cannot attach imported joints to articulation #{parent_articulation}: "
+                            f"loop-closing joint '{joint_name}' at index {joint_idx} lies between the existing "
+                            "regular joints and the imported joints."
+                        )
                 # Mark all new joints as belonging to the parent's articulation
```
vreutskyy, line 8941: Should this also reject joints in `[old_end, new_end)` that belong to a different articulation, not just those with `joint_articulation == -1`? `_validate_structure` does catch the resulting overlap at `finalize()` via the `articulation_end > next_start` check, but only with a generic message ("articulation_end[i] exceeds the next articulation start"). Dropping the `== -1` clause and naming the owning articulation here would localize the failure to the cause:

```python
for joint_idx in range(old_end, new_end):
    if joint_idx not in imported_joints:
        owner = self.joint_articulation[joint_idx]
        kind = "loop-closing joint" if owner == -1 else f"joint owned by articulation #{owner}"
        raise ValueError(...)
```

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -5275,6 +5422,41 @@ def get_body_name(body_idx: int) -> str:
...
+                        cns_relpose = wp.transform(*joint_eq_relpose[j])
+                        eq.data[3:6] = wp.transform_get_translation(cns_relpose)
+                        eq.data[6:10] = quat_to_mjc(wp.transform_get_rotation(cns_relpose))
+                    if joint_eq_torquescale is not None:
+                        eq.data[10] = float(joint_eq_torquescale[j])
+                if joint_eq_solref is not None:
+                    eq.solref = joint_eq_solref[j]
+                if joint_eq_solimp is not None:
+                    eq.solimp = joint_eq_solimp[j]
+                continue
```
vreutskyy: For converted CONNECT/WELD the metadata is read once into the template `MjSpec` here, and unlike the mimic path there's no runtime update kernel analogous to `update_mimic_eq_data_and_active_kernel`. With `separate_worlds=True`, would per-world differences (`joint_enabled`, `joint_eq_solref`, `joint_eq_solimp`, etc.) silently collapse to world 0's values after `_expand_model_fields()` tiles them? `test_preserved_mimic_runtime_update_keeps_mujoco_data` covers the mimic side — worth an analogous multi-world test (and possibly a kernel) for the loop-joint side?

newton/_src/utils/import_mjcf.py
```diff
@@ -1905,6 +1909,113 @@ def get_site_body_and_anchor(site_name: str) -> tuple[int, wp.vec3] | None:
...
+                parent_anchor = world_anchor
+            child_anchor = wp.transform_point(wp.transform_inverse(child_xform_world), world_anchor)
+            return (
+                parent,
+                child,
+                wp.transform(parent_anchor, wp.quat_identity()),
+                wp.transform(child_anchor, wp.quat_identity()),
+            )
+
+        def add_converted_loop_joint(
```
vreutskyy, line 2019: `import_usd.py::_parse_mjc_equality_constraints` auto-registers `SolverMuJoCo.register_custom_attributes(builder)` if `mujoco:joint_eq_type` is missing on the builder. I don't see the equivalent fallback in the MJCF path — does this rely on SolverMuJoCo attributes being registered upstream by the caller? If so, worth a docstring note on `add_mjcf`; otherwise mirroring the USD fallback here would keep the two importers symmetric.

newton/_src/utils/import_usd.py
```diff
@@ -3167,6 +3171,115 @@ def _get_collision_mass_information(collider_prim: Usd.Prim):
...
+                "mujoco:joint_eq_body1": body1,
+                "mujoco:joint_eq_body2": body2,
+                "mujoco:joint_eq_anchor": anchor,
+                "mujoco:joint_eq_relpose": relpose or wp.transform_identity(),
+                "mujoco:joint_eq_torquescale": torquescale,
+                "mujoco:joint_eq_solref": eq_solref(custom_attrs),
+                "mujoco:joint_eq_solimp": eq_solimp(custom_attrs),
+            }
+
+        def loop_joint_xforms(body1: int, body2: int, anchor: wp.vec3) -> tuple[int, int, wp.transform, wp.transform]:
```
vreutskyy: `loop_joint_xforms`, `joint_eq_custom_attrs`, `add_converted_loop_joint`, `mimic_eq_custom_attrs`, `eq_solref`, `eq_solimp` look essentially identical between `import_mjcf.py` and `import_usd.py`. Worth extracting them to a shared helper module (e.g. `newton/_src/utils/_mjc_equality.py`) so the two importers can't drift over time?

newton/_src/sim/model.py
```diff
@@ -583,7 +583,9 @@ def __init__(self, device: Devicelike | None = None):
         """
 
         self.articulation_start: wp.array[wp.int32] | None = None
-        """Articulation start index, shape [articulation_count], int."""
+        """Articulation start index plus sentinel, shape [articulation_count + 1], int."""
+        self.articulation_end: wp.array[wp.int32] | None = None
+        """Exclusive end index of regular tree joints per articulation, shape [articulation_count], int."""
```
vreutskyy, line 593: Now that `articulation_end` is explicit, is the trailing sentinel in `articulation_start` (`[articulation_count + 1]`) still serving a purpose? The asymmetry between the two shapes is a bit of a footgun for future callers. If every traversal has been migrated to `articulation_end`, dropping the sentinel (or at least flagging it as planned cleanup) would make the contract cleaner. Is anything still relying on `articulation_start[i+1]` directly?

newton/tests/test_import_mjcf.py
```diff
@@ -1735,7 +1735,7 @@ def test_single_mujoco_fixed_tendon_defaults(self):
     </fixed>
   </tendon>
 </mujoco>
-"""
+        """
```
vreutskyy: Looks like the closing `"""` of this MJCF string picked up indentation in this PR — adds 8 spaces of trailing whitespace inside the string literal. Intentional, or stray reformat?

vreutskyy (commented):
Read through the diff — really like the round-trip preservation story and the explicit `articulation_end` split. A handful of questions inline, mostly around behavior for non-MuJoCo solvers, multi-world preservation of the converted CONNECT/WELD path, and a couple of small consistency / cleanup nits. Nothing blocking from my side.

newton/_src/sim/model.py
eric-heiden: Good point. The sentinel is still intentional: `articulation_start[i]:articulation_end[i]` bounds the regular tree joints, while `articulation_end[i]:articulation_start[i + 1]` bounds the converted loop-closing joints. `ArticulationView(include_loop_closing_joints=True)` still uses `articulation_start[i + 1]` as that full-range upper bound, and the `Model.articulation_start` docstring now documents this contract. So I’m keeping the sentinel shape in this PR.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -5705,15 +5917,21 @@ def get_body_name(body_idx: int) -> str:
...
                 template_eq = newton_eq % eq_constraints_per_world if eq_constraints_per_world > 0 else newton_eq
                 for w in range(nworld):
                     mjc_eq_to_newton_eq_np[w, mjc_eq] = w * eq_constraints_per_world + template_eq
             for mjc_eq, newton_jnt in mjc_eq_to_newton_jnt.items():
                 for w in range(nworld):
                     mjc_eq_to_newton_jnt_np[w, mjc_eq] = w * joints_per_world + newton_jnt
+            for mjc_eq, newton_jnt in mjc_eq_to_newton_preserved_jnt.items():
+                template_jnt = newton_jnt % joints_per_world if joints_per_world > 0 else newton_jnt
+                for w in range(nworld):
+                    mjc_eq_to_newton_preserved_jnt_np[w, mjc_eq] = w * joints_per_world + template_jnt
```
vreutskyy: Small inconsistency between the two adjacent per-world expansions: this new one applies `template_jnt = newton_jnt % joints_per_world` before tiling, while the existing `mjc_eq_to_newton_jnt_np` loop right above (5925-5927) tiles `newton_jnt` directly without the modulo. Both populate their source dicts from the same `joints_loop = selected_joints[joint_articulation[selected_joints] == -1]`, so unless I'm missing something either the modulo here is defensive-but-unnecessary, or the existing loop is silently buggy when `first_world != 0`. Worth aligning the two?

vreutskyy (dismissed):
Re-reviewed against db77163. All seven of my earlier points addressed:

- Polycoef higher-order warning added in both importers
- `_finalize_imported_articulation` guard now names the owning articulation in the error
- New `update_preserved_jnt_eq_data_and_active_kernel` + `_update_preserved_jnt_eq_properties` + multi-world `test_preserved_loop_equality_runtime_update_keeps_per_world_mujoco_data` cover the converted CONNECT/WELD runtime sync path
- MJCF importer now auto-registers `SolverMuJoCo` custom attributes symmetric to USD
- Shared helpers extracted to `newton/_src/solvers/mujoco/utils.py`
- Sentinel kept intentional — the expanded `articulation_start` docstring documents the contract clearly
- Test docstring whitespace reverted

Nice work on the round-trip kernels and the multi-world test. One small consistency question inline at the new per-world expansion — non-blocking. Approving.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -810,7 +818,133 @@ def register_custom_attributes(cls, builder: ModelBuilder) -> None:
...
+        builder.add_custom_attribute(
+            ModelBuilder.CustomAttribute(
+                name="mimic_eq_solimp",
+                frequency=AttributeFrequency.CONSTRAINT_MIMIC,
+                assignment=AttributeAssignment.MODEL,
+                dtype=vec5,
+                default=vec5(0.9, 0.95, 0.001, 0.5, 2.0),
+                namespace="mujoco",
+            )
+        )
```
adenzler-nvidia, line 902: I think we need to be careful here. With the newton-level equality constraint deprecation coming in as well, we need to figure out what custom attributes to expose correctly, to avoid cluttering this and adding multiple sources of truth.

I went with a mujoco-namespaced copy of the existing equality constraint attributes in that PR but I think we should go back on it and actually just expose the pure mujoco attributes. Converted joints could then write into that, as well as the legacy fallback during import or any explicit equality constraint that is authored through custom attributes. Then, we just need another array to mark what was converted to joint what is a pure equality constraint?

Does that make sense?

adenzler-nvidia: Following up with a concrete proposal after going through both PRs.

Instead of the parallel `joint_eq_*` / `mimic_eq_*` sets, I think we keep a **single** equality-constraint custom-attribute table (`solref`/`solimp` already live at that frequency) and add just **three** new columns:

- `equality_constraint_target_kind` — `NONE` / `JOINT` / `MIMIC`; what the row was projected to (`NONE` = pure equality).
- `equality_constraint_target` — the joint or mimic index it was converted to (`-1` for pure rows).
- `equality_constraint_objtype` — `mjtObj` discriminator so the table can also hold site/tendon refs later.

Everything you currently preserve on the joint/mimic (`type`, `body1/2`, `anchor`, `relpose`, `torquescale`, full `polycoef`, `solref`, `solimp`) is already a column on that table, so it doesn't need a second home: converted joints, the legacy import fallback, and explicitly-authored equalities all write the same fields. `mimic_eq_preserve` then becomes `target_kind == MIMIC`, and the "skip in native MuJoCo sim" exclusion is derived from these columns at build time (joints via `articulation_end`, mimics from `target_kind`). That keeps one source of truth and drops the redundancy where `joint_eq_anchor` / `joint_eq_body*` duplicate the loop joint's own geometry.

The Newton-level deprecation (#2961) would expose exactly these namespaced attributes and forward the old `Model.equality_constraint_*` arrays onto them.

eric-heiden: Implemented this in 3dc96410d. The parallel `joint_eq_*` / `mimic_eq_*` preservation attrs are removed now, and converted CONNECT/WELD/JOINT constraints all keep one authoritative equality row.

The row has the three projection columns from your proposal:
- `mujoco:equality_constraint_target_kind`
- `mujoco:equality_constraint_target`
- `mujoco:equality_constraint_objtype`

CONNECT/WELD imports add a projected loop joint plus an equality row with `target_kind == JOINT`; JOINT imports add a mimic constraint plus an equality row with `target_kind == MIMIC`. Full MuJoCo data stays on the equality table: type/body refs/anchor/relpose/torquescale/polycoef/solref/solimp. The MuJoCo conversion path derives the skip lists from those target columns, and the builder remaps the polymorphic target through add_builder/collapse_fixed_joints.

newton/_src/solvers/mujoco/solver_mujoco.py
eric-heiden: Good catch. Addressed in a666b81f0 by normalizing the loop-joint equality mapping to the per-world template joint index before tiling:

```python
template_jnt = newton_jnt % joints_per_world if joints_per_world > 0 else newton_jnt
```

That makes `mjc_eq_to_newton_jnt` consistent with the adjacent equality and mimic mappings. The separate preserved-joint mapping that originally prompted the comparison is gone after the equality-table refactor, but this existing path still benefited from the same alignment.

Validation: `ruff-format`, `ruff`, `compileall`, `git diff --check`, and `uv run --extra dev -m newton.tests -k test_mixed_loop_joints_and_equality_constraints`.

newton/_src/solvers/mujoco/solver_mujoco.py
adenzler-nvidia: Confirmed implemented in 3dc96410 + a666b81f: the parallel `joint_eq_*` / `mimic_eq_*` sets are gone, and converted CONNECT/WELD/JOINT rows keep a single authoritative equality row with the `equality_constraint_target_kind` / `target` / `objtype` projection columns. The remaining overlap with #2961 (duplicate registration of those columns + `eq_solref`/`eq_solimp`, and consolidating `MjcEqualityTargetKind`/`MJC_OBJ_*` into `EqTarget`/`EqObjType`) I will reconcile on the #2961 side once this lands. Resolving.

adenzler-nvidia (dismissed):
Approving. The equality-conversion design is sound and consistent across the MJCF and USD importers, the `articulation_start[i+1]` → `articulation_end[i]` migration is applied consistently (with the loop-closing-joint range contract guarded by a fail-fast `ValueError` in `add_articulation`), and the per-world expansion path is correct. CI is green across all platforms.

Two non-blocking follow-ups worth considering:

- The new index-remap plumbing is under-tested: `add_builder` / `collapse_fixed_joints` remapping of `equality_constraint_target` (including the drop → `NONE` invalidation), and the multi-world `mjc_eq_to_newton_jnt` `template_jnt % joints_per_world` fix for generic CONNECT loop joints, have no direct coverage. They work today but a regression wouldn't be caught.
- WELD `relpose`/`torquescale` live only on the equality row (the synthesized loop joint is metadata-only) — fine for current consumers, worth a brief comment.

One thing before merge: the branch currently conflicts with `main` (needs an `upstream/main` merge + resolve).

eric-heiden:
Followed up on the approval notes from https://github.com/newton-physics/newton/pull/2959#pullrequestreview-4392592539 in ee09e397d:

- Merged `origin/main` into this branch and resolved the import conflicts from the new MuJoCo constants module.
- Added direct coverage for `mujoco:equality_constraint_target` remapping through `add_builder` for both `JOINT` and `MIMIC` target kinds.
- Added coverage for `collapse_fixed_joints()` invalidating a removed projected joint target back to `target_kind == NONE` / `target == -1`.
- Added coverage for generic CONNECT loop-joint `mjc_eq_to_newton_jnt` mapping across worlds.
- Added a brief comment clarifying that WELD `relpose` / `torquescale` stay on the equality row while the loop joint is only the projected Newton simulation object.

adenzler-nvidia (approved):

PR #2960, opened 2026-05-27
Allow specifying recording ID for ViewerRerun

breuerpeter:
## Description

Allows specifying the recording ID for the Rerun Viewer. This is needed for when a different process is also logging to Rerun and you want data from both to appear in the same recording, ie be on one timeline.

Example application: I'm using Newton as a simulator with the PX4 Autopilot. I have a PX4 module that uses Rerun to log  data (eg state estimate). If I can specify the same application ID (already possible) AND recording ID (this PR), then the Rerun viewer shows data from both sources, allowing for a great debug/visualization experience.

I think this is a feature that is valuable for all devs that run other software that uses Rerun for logging alongside Newton.

## Checklist

## Test plan

2 new tests that are run as part of `uv run --extra dev -m newton.tests -k test_viewer_rerun_init_args`:

- `test_custom_rec_id_used` — verifies a custom rec_id is stored on the viewer and forwarded to rr.init as recording_id
- `test_default_rec_id_is_none` — verifies rec_id defaults to None when not provided

## New feature / API change

```python
from newton.viewer import ViewerRerun

viewer = ViewerRerun(app_id="my-app", rec_id="shared-recording")
```


eric-heiden (approved):
Thanks!

Issue #1675, opened 2026-02-17
Improve `ModelBuilder.add_builder` performance


eric-heiden:

jcarius-nv:
@camevor if this makes sense in the context of IsaacLab startup performance work, let me know and we can assign to a suitable release again

eric-heiden:
@jcarius-nv this came up again and is now very important for IsaacLab. Since I've done some prototyping with this, I think I can tackle this for 1.4 if you agree?

jcarius-nv:
agree! Is this something you want/can take? Otherwise I'm thinking @camevor ?

PR #2961, opened 2026-05-27
Deprecate model equality arrays

adenzler-nvidia:
## Description

Deprecates the top-level `Model.equality_constraint_*` arrays and `Model.equality_constraint_count` in favor of the MuJoCo namespaced model extension fields at `model.mujoco.equality_constraint_*`.

The builder and importers keep creating equality constraints through the existing entry points, while finalized models now populate the MuJoCo namespace as the source of truth. Deprecated top-level properties forward to the namespaced fields during the compatibility window.

Refs #2665

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_equality_constraint_model_fields
uv run --extra dev -m newton.tests -k test_equality_constraints
uv run --extra dev -m newton.tests -k TestMuJoCoSolverEqualityConstraintProperties
uv run --extra dev -m unittest newton._src.solvers.kamino.tests.test_solver_kamino.TestSolverKaminoCompatibility.test_validate_model_compatibility_allows_model_without_mujoco_namespace
uvx pre-commit run -a
```

## New feature / API change

```python
model = builder.finalize()
model.mujoco.equality_constraint_count
model.mujoco.equality_constraint_body1
```

Deprecated compatibility properties remain available during the deprecation window and forward to the MuJoCo namespace:

```python
model.equality_constraint_count
model.equality_constraint_body1
```

## Summary by CodeRabbit

* **Deprecation**
  * Top-level equality-constraint fields are deprecated; use model.mujoco.equality_constraint_* and model.mujoco.equality_constraint_count. Legacy getters warn and forward; legacy builder writes are disallowed during the deprecation window.

* **Documentation**
  * Docs updated to reference namespaced equality-constraint fields and explain deprecation and compatibility behavior.

* **Behavior**
  * Builder/model auto-register and materialize namespaced equality-constraint attributes so existing workflows remain compatible.

* **Tests**
  * Tests adjusted and new compatibility tests added to validate namespaced fields and forwarding.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2961?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


chschuma-disney (dismissed):
Changes look good from Kamino's side.

eric-heiden:
Thanks for moving the equality constraint storage into the MuJoCo custom-attribute namespace.

The main issues are that the deprecation path currently removes some old behavior immediately, and a few public docs and tests still need to match the new API shape.
None of these are blocking, just something to double-check.

newton/_src/sim/model.py
```diff
@@ -834,6 +799,221 @@ def __init__(self, device: Devicelike | None = None):
...
+        """Deprecated in Newton 1.3; will be removed in a future release. Use ``model.mujoco.equality_constraint_world``."""
+        return self._get_deprecated_equality_constraint_model_field("equality_constraint_world")
+
+    @equality_constraint_world.setter
+    def equality_constraint_world(self, value: wp.array[wp.int32] | None) -> None:
+        self._set_deprecated_equality_constraint_model_field("equality_constraint_world", value)
+
+    @property
+    def equality_constraint_world_start(self) -> wp.array[wp.int32] | None:
+        """Deprecated in Newton 1.3; will be removed in a future release. Equality constraints have no per-world start array.
```
eric-heiden: 🔴 `Model.equality_constraint_world_start` now warns and returns `None`, but this API used to expose the finalized Warp start array. Existing callers that do `model.equality_constraint_world_start.numpy()` or compute per-world bounds from it will fail immediately in the same release that only marks the field deprecated.

I guess nobody has been using this attribute anyway, so we might as well remove it entirely? Or add it back if it doesn't cause too much trouble.

docs/integrations/mujoco.rst
```diff
@@ -254,6 +254,12 @@ array; slot layout depends on the constraint type.
        :attr:`~newton.JointType.REVOLUTE` and
        :attr:`~newton.JointType.PRISMATIC` joints are supported.
 
+Newton's core API does not expose equality constraints as a dedicated
+builder call. Construct them through the MuJoCo
+:ref:`custom-attribute namespace <mujoco-custom-attributes>` with
```
eric-heiden, line 304: 🟡 This new migration text points users to `mujoco:equality_constraint_*`, but the public importer docstrings still point the other way. `ModelBuilder.add_usd()` still says `convert_mjc_equality_constraints=False` stores constraints in "the legacy equality constraint arrays", and `add_mjcf()` repeats that wording.

Please update both docstrings to say preserved MuJoCo equality rows live in the `mujoco:equality_constraint` custom-attribute namespace and finalize under `model.mujoco.equality_constraint_*`.

newton/_src/sim/builder.py
```diff
@@ -1258,6 +1230,138 @@ def __init__(self, up_axis: AxisType = Axis.Z, gravity: float = -9.81):
...
+            f"ModelBuilder.{name} is deprecated in Newton 1.3 and is scheduled for removal in "
+            f"a future release. Populate equality constraints via "
+            f'add_custom_values(**{{"mujoco:{name}": ...}}) and read finalized values from '
+            f"model.mujoco.{name}.",
+            DeprecationWarning,
+            stacklevel=3,
+        )
+        return self._eq_list(name)
+
+    def _warn_deprecated_builder_add_equality_constraint(self, name: str) -> None:
```
eric-heiden, line 1324: 🔴 This warning is currently triggered by existing solver tests without being asserted or suppressed. In `newton/tests/test_mujoco_solver.py`, the tests still call `add_equality_constraint_connect`, `add_equality_constraint_weld`, and `add_equality_constraint_joint` directly at several sites, plus `model.equality_constraint_enabled.assign(...)`, so the suite emits deprecation noise and would fail under `-W error::DeprecationWarning` without showing whether the deprecation behavior is intentional.

Please either migrate those tests to the documented `add_custom_values` path, or wrap each deprecated API call with `self.assertWarns(DeprecationWarning)` so the warning is part of the tested contract.

eric-heiden:
Please ignore this note, my agent just did some benchmarking and found the equality row add/finalize path has a measurable regression. This because of the slowness of `ModelBuilder.add_custom_values()` which is slightly related to https://github.com/newton-physics/newton/issues/1675, will investigate more.

In a 5,000-row benchmark averaged over three 10-run invocations, this branch takes 0.223s vs 0.160s on `origin/main`, which is about 40% slower. Empty builder finalization is effectively flat at 1.136s vs 1.124s. 

| Benchmark | `origin/main` | This branch | Delta |
| --- | ---: | ---: | ---: |
| Add and finalize 5,000 equality rows | 0.160s | 0.223s | +40% |
| Finalize 500 empty builders | 1.124s | 1.136s | +1% |

<details>

<summary>Minimal benchmark code</summary>

```py
import warp as wp

import newton
from newton.utils import run_benchmark

wp.config.log_level = wp.LOG_WARNING

try:
    from newton._src.solvers.mujoco.equality import _add_equality_constraint as add_eq
except Exception:
    add_eq = None


def add_equality_row(builder, kind, body1, body2):
    if add_eq is not None:
        return add_eq(
            builder,
            constraint_type=kind,
            body1=body1,
            body2=body2,
            anchor=wp.vec3(0.1, 0.2, 0.3) if kind == newton.EqType.CONNECT else None,
        )

    if kind == newton.EqType.CONNECT:
        return builder.add_equality_constraint_connect(
            body1=body1,
            body2=body2,
            anchor=wp.vec3(0.1, 0.2, 0.3),
        )
    return builder.add_equality_constraint_weld(body1=body1, body2=body2)


def equality_count(model):
    return getattr(getattr(model, "mujoco", model), "equality_constraint_count")


class BenchEqualityBuilder:
    empty_builders = 500
    equality_rows = 5000

    def setup(self):
        wp.init()

    def time_finalize_empty_builders(self):
        for _ in range(self.empty_builders):
            newton.ModelBuilder().finalize(device="cpu", skip_all_validations=True)

    def time_add_finalize_equality_rows(self):
        body1 = builder.add_body()
        body2 = builder.add_body()
        for i in range(self.equality_rows):
            kind = newton.EqType.CONNECT if i % 2 == 0 else newton.EqType.WELD
            add_equality_row(builder, kind, body1, body2)
        model = builder.finalize(device="cpu", skip_all_validations=True)
        assert equality_count(model) == self.equality_rows


run_benchmark(BenchEqualityBuilder, number=10)
```

</details>

eric-heiden (dismissed):

chschuma-disney (dismissed):

newton/_src/sim/builder.py
```diff
@@ -9660,51 +9876,47 @@ def _validate_structure(self) -> None:
...
+        # Validate equality constraint body/joint references
+        equality_count = self._equality_constraint_count
         if equality_count > 0:
-            eq_body1 = np.array(self.equality_constraint_body1, dtype=np.int32)
+            eq_labels = self._eq_list("equality_constraint_label")
+
+            def _eq_label(idx: int) -> str:
+                return eq_labels[idx] or f"equality_constraint_{idx}"
+
+            eq_body1 = np.array(self._eq_list("equality_constraint_body1"), dtype=np.int32)
```
eric-heiden: 🟡 This validation path now calls `_eq_list()` for each equality field, and each call builds a fresh dense Python list over every equality constraint before NumPy converts it again. On large models, `finalize()` pays that allocation cost several times just to validate body and joint indices, while the old path read the backing lists directly.

Please use the raw custom-attribute values for internal validation and collapse paths, for example through a small helper like `_eq_values_raw(name)` that returns `_eq_attr(name).values or []`. The existing `_at()` and NumPy coercion logic can still apply defaults without materializing every equality row up front.

adenzler-nvidia: Done in a3097994f. `_validate_structure` no longer routes through `_eq_list()`: `_eq_values_raw()` reads the backing custom-attribute values without default-filling, and `_eq_index_array()` coerces them straight into the int32 array via `np.fromiter`, applying the attribute default only for missing/`None` rows — one allocation per field, no intermediate dense list.

newton/_src/sim/builder.py
adenzler-nvidia: Done in a3097994f. The deprecated `add_equality_constraint_*` builder calls were already migrated to the `add_custom_values`-backed `_add_equality_constraint` helper in an earlier commit; this one migrates the remaining top-level `model.equality_constraint_enabled` / `equality_constraint_polycoef` accessors in `test_mujoco_solver.py` to the `model.mujoco.*` namespace, so the suite no longer emits deprecation warnings.

jcarius-nv:
Principal-engineer review notes. Posting as a top-level comment so this does not carry an approval verdict.

1. `newton/_src/sim/builder.py`, around `add_builder()` string-frequency merge (`merged.values.extend(new_values)`): this loses row alignment for sparse string-frequency attributes. `index_offset` is computed above, but optional equality attrs such as `mujoco:eq_solref` / `mujoco:eq_solimp` are appended without padding to that offset.

   I reproduced this with two builders: builder A has one equality with default `mujoco:eq_solref`; builder B has one equality with custom `mujoco:eq_solref=(9, 9)`. After `main.add_builder(A); main.add_builder(B); finalize()`, the materialized `model.mujoco.eq_solref` is:

   ```text
   [[9.   9.  ]
    [0.02 1.  ]]
   ```

   The custom value should be on row 1, not row 0. Please pad string-frequency `merged.values` to `index_offset` before appending/copying incoming values, preserving `None` gaps.

2. `newton/_src/sim/builder.py`, `ModelBuilder._deprecated_eq_list()`: the deprecated builder accumulator lists now return mutable snapshots. Historically `builder.equality_constraint_type.append(...)` mutated the builder. Now that call still succeeds on the returned list object, but the builder state is unchanged, so downstream code silently drops the constraint. For a deprecation window, this should either remain functional through a backing/proxy object or fail loudly on mutation. Silent data loss is worse than a clear deprecation error.

3. `newton/_src/sim/builder.py`, custom-attribute materialization skip for `count == 0`: finalized models expose `model.mujoco.equality_constraint_count` and `model.mujoco.equality_constraint_world_start`, but none of the replacement `model.mujoco.equality_constraint_*` arrays/lists are materialized when there are zero equality rows. I checked an empty finalized builder and `hasattr(model.mujoco, "equality_constraint_type")` is `False`. Since the migration message tells callers to use the namespaced finalized fields, the replacement namespace should be shape-stable even when there are zero rows, likely as empty arrays/lists or otherwise documented `None` values.

4. Test cleanup: a few tests still read deprecated top-level model equality fields, for example `converted_model.equality_constraint_count`, `converted_model.equality_constraint_type`, and `converted_model.equality_constraint_polycoef` in `newton/tests/test_import_mjcf.py` / `newton/tests/test_import_usd.py`. Since the PR makes those accessors emit `DeprecationWarning`, these tests should either move to `model.mujoco.*` or wrap the deprecated access as an explicit deprecation-contract assertion; otherwise strict warning runs such as `PYTHONWARNINGS=error::DeprecationWarning` will fail for the wrong reason.

jcarius-nv (dismissed):
looks good apart from the notes I posted above. My agent failed to attach it to specific lines, sorry about that :/

I had a look on my own and found no further issues beyond what was already mentioned.

eric-heiden (approved):
Thanks! My comments and it looks like @jcarius-nv's comments have been addressed.

adenzler-nvidia:
All four addressed:

1. **Sparse attribute alignment** (`d625ac57`) — `add_builder`'s string-frequency merge now left-pads `merged.values` to `index_offset` with `None` (which resolves to the attribute default at finalize) before appending, in both the first-declaration and extend paths. Your A/B reproducer now yields `[[0.02, 1.0], [9.0, 9.0]]`.

2. **Silent mutation** (`9fea7457`) — the deprecated `ModelBuilder.equality_constraint_*` accessors now return a read-only `list` view that raises `TypeError` on any in-place mutation, and the snapshot detaches mutable inner values (polycoef lists and the shared default) so element mutation can't reach the backing store either. Went with fail-loud rather than a functional proxy because appending to a single field list can't construct a valid constraint row — all fields must stay in sync, which is why `add_equality_constraint_*()` wrote them atomically.

3. **Zero-row materialization** (`cefd0d42`) — finalize now exempts the `mujoco:equality_constraint` frequency from the zero-count skip, so every registered equality field materializes as a shape-stable empty array even with no constraints, driven by the registration rather than a hardcoded field list. `model.mujoco.equality_constraint_*` and the deprecated top-level getter now both return empty arrays at zero rows instead of raising `AttributeError` / returning `None`.

4. **Test cleanup** (`1b6461a4`) — `test_import_mjcf.py` / `test_import_usd.py` moved off the deprecated top-level model accessors to `model.mujoco.*`.

newton/_src/sim/model.py
adenzler-nvidia: `cefd0d42` makes the whole equality namespace shape-stable at zero rows, and `equality_constraint_world_start` is materialized at finalize unconditionally, so the deprecated `model.equality_constraint_world_start` forwards to the real Warp array rather than `None` — it returns `[0, 0, 0]` for a constraint-free 3-world model and `[0, 1, 2, 2]` for a 2-world model with one constraint each. Kept it functional rather than removing it, to stay consistent with the other deprecated fields during the window.

docs/integrations/mujoco.rst
adenzler-nvidia: `add_mjcf()` and `add_usd()` docstrings now state that preserved MuJoCo equality rows live in the `mujoco:equality_constraint` custom-attribute namespace and finalize under `model.mujoco.equality_constraint_*` (commits `2305903c` / `c8540434`).

chschuma-disney (approved):

PR #2964, opened 2026-05-27
Fix MuJoCo shape_material force-space scaling (#2009)

vreutskyy:
## Summary

Finishes issue #2009 for shape-material contacts. Now rebased on top of merged #2610.

Newton's documented force-space `shape_material_ke` / `shape_material_kd` now produce the requested contact stiffness when `SolverMuJoCo` runs with `use_mujoco_contacts=False` (the default Newton-contacts path).

## Design

- New per-shape custom attributes `mujoco.solref` (vec2) and `mujoco.solref_mode` (int32), mirroring the joint-limit pattern from #2610.
- Modes: `SOLREF_MODE_FORCE_SPACE` (Newton ke/kd, scaled per-contact), `SOLREF_MODE_RAW` (authored `mjc:solref` forwarded unscaled), `SOLREF_MODE_MJCF_DEFAULT` (MJCF/USD geom without authored `mjc:solref` — preserves MuJoCo's compiled default until Newton ke/kd are edited).
- `convert_newton_contacts_to_mjwarp_kernel` gains a force-space override: when both shapes in a contact are `FORCE_SPACE` it bypasses per-geom `solref` averaging and writes `contact.solref = (-ke·factor, -kd·factor)` directly using MuJoCo's negative-direct-mode convention, with `factor = (body_invweight0[A] + body_invweight0[B]) · (1 − dmax)`.
- Auto-promote: `MJCF_DEFAULT` shapes flip to `FORCE_SPACE` on the first edit to Newton ke/kd. `BODY_INERTIAL_PROPERTIES` already invalidates the contact fast path, so mass randomization picks up refreshed `body_invweight0` automatically.
- Importers preserve authored `mjc:solref` verbatim into `mujoco.solref` (with `SOLREF_MODE_RAW`); the legacy `solref_to_stiffness_damping` round-trip into Newton `shape_material_ke` / `kd` is kept for back-compat so that imported MuJoCo defaults land at `DEFAULT_SHAPE_KE` / `DEFAULT_SHAPE_KD` and the solver picks `mujoco.solref` over `shape_material_ke` whenever `solref_mode == SOLREF_MODE_RAW`.

## Limitations

- `use_mujoco_contacts=True` and the MuJoCo CPU backend remain unsupported for `FORCE_SPACE` (MuJoCo's internal `contact_params` averages per-geom `solref` and cannot reproduce a per-contact two-body inverse-mass sum). Shapes in `FORCE_SPACE` fall back to the legacy `convert_solref(ke, kd, 1, 1)` approximation on those paths.
- `condim > 1` friction-cone scaling and the new `diagexact` mode are out of scope (Alain confirmed acceptable in the issue thread).

## Test plan

- [x] `uv run --extra dev -m newton.tests -k TestMuJoCoSolverForceSpaceContactSolref` (6/6 pass)
- [x] `uv run --extra dev -m newton.tests -k TestMuJoCoSolverInvweightScaledSolref` (20/20 pass — no regression in #2610's coverage)
- [x] `uv run --extra dev -m newton.tests -k test_import_mjcf` (211/211 pass)
- [x] `uvx pre-commit run -a` clean

Refs: #2009, #2610.

## Release Notes

* **Bug Fixes**
  * Fixed contact dynamics handling to properly honor authored force-space parameters when `use_mujoco_contacts=False`
  * Contact configuration values are now correctly preserved and applied per-shape


newton/_src/solvers/mujoco/kernels.py
```diff
@@ -2313,10 +2362,45 @@ def update_geom_properties_kernel(
...
+    #   * ``SOLREF_MODE_MJCF_DEFAULT``: same legacy ``convert_solref(ke,
+    #     kd, 1, 1)`` write as ``FORCE_SPACE`` so example/tuning code that
+    #     relies on the historical ``shape_material_ke``/``kd`` →
+    #     ``geom_solref`` round-trip (e.g. ``example_brick_stacking``'s
+    #     ``BRICK_KE = mg/penetration``) keeps working out of the box.
+    #     The per-contact ``body_invweight0`` override in
+    #     ``convert_newton_contacts_to_mjwarp_kernel`` only fires for
+    #     ``FORCE_SPACE`` shapes; ``MJCF_DEFAULT`` shapes therefore land
+    #     at the pre-PR contact dynamics (acceleration-space ``solref``
+    #     averaged across the contact pair).
```
adenzler-nvidia: A lot of this PR is just comments. Could you shorten these, maybe find a good place to document the actual mechanism and then refer to that? I realize that there is precedent from the limit conversion, but given that that is the same conversion, maybe we can have a guide section that explains the generic mechanism and then remove a lot of the comment noise?

newton/tests/test_mujoco_solver.py
```diff
@@ -10639,5 +10640,268 @@ def test_parse_vec_unknown_shorthand_warns(self):
...
+    def test_force_space_contact_normal_force_matches_mg(self):
+        """``shape_material_ke`` in force space: resting contact force equals ``m * g``.
+
+        Only the Newton-contacts (Warp) pipeline is exercised — the CPU
+        backend uses the legacy ``convert_solref(ke, kd, 1, 1)``
+        approximation that ignores both ``body_invweight0`` and ``dmax``.
+        """
+        mass = 5.0
+        model, _ = self._build_box_on_plane(mass=mass, ke=1.0e4, kd=100.0)
+        solver = SolverMuJoCo(model, njmax=20, iterations=10)
```
adenzler-nvidia: **Critical: this test (and the other two `test_force_space_contact_*`) never runs the force-space override.**

`SolverMuJoCo(model, njmax=20, iterations=10)` uses the default `use_mujoco_contacts=True` (`solver_mujoco.py:3170`), and `step()` is called with `contacts=None`. The override lives in `convert_newton_contacts_to_mjwarp_kernel`, which only runs on the `use_mujoco_contacts=False` branch — so with `True` it's dead code here.

The assertion `F ≈ m·g` holds from static equilibrium under *any* sufficiently stiff `solref`: the test passes identically with `MJCF_DEFAULT` and would still pass if the entire override block (`kernels.py:415-443`) were deleted. The mass-independence test is a tautology (`F = m·g` is mass-dependent by construction).

Please either construct with `use_mujoco_contacts=False` and feed a real `CollisionPipeline.collide()` → `Contacts` into `step()`, or — mirroring #2610's joint suite (`test_mujoco_solver.py:9911`) — read back the per-contact `solref` and assert it equals `-ke·(invw_a+invw_b)·(1-dmax)` directly. Confirm the new test fails without the kernel change.

newton/tests/test_mujoco_solver.py
```diff
@@ -10639,5 +10640,268 @@ def test_parse_vec_unknown_shorthand_warns(self):
...
+    Issue #2009: MJCF/USD importers used to convert ``mjc:solref``
+    (acceleration space) directly into Newton ``shape_material_ke`` /
+    ``shape_material_kd`` (force space), corrupting the documented
+    semantics. The fix stores raw ``mjc:solref`` in the
+    ``mujoco.solref`` custom attribute and derives the per-contact
+    ``solref`` from Newton ke/kd plus ``body_invweight0[A] +
+    body_invweight0[B]`` in ``convert_newton_contacts_to_mjwarp_kernel``.
+
+    These tests cover the Newton-contacts pipeline
+    (``use_mujoco_contacts=False``, the default). ``use_mujoco_contacts=
```
adenzler-nvidia: This docstring is incorrect on two counts: `use_mujoco_contacts` defaults to `True` (`solver_mujoco.py:3170`), not `False`, and these tests don't set it to `False`. So they actually run MuJoCo's own contact solver, not the Newton-contacts pipeline this class claims to cover. See the critical comment below.

newton/_src/solvers/mujoco/constants.py
```diff
@@ -32,6 +32,24 @@
 DEFAULT_LIMIT_SOLREF_TIMECONST = DEFAULT_LIMIT_SOLREF[0]
 """MuJoCo's implicit default joint-limit solref time-constant component."""
 
+DEFAULT_SHAPE_GAIN_ATOL = 1.0e-4
```
adenzler-nvidia: These six new constants (`DEFAULT_SHAPE_GAIN_ATOL`, `DEFAULT_SHAPE_KD`, `DEFAULT_SHAPE_KE`, `DEFAULT_SHAPE_SOLREF`, `DEFAULT_SHAPE_SOLREF_DAMPRATIO`, `DEFAULT_SHAPE_SOLREF_TIMECONST`) are referenced **only in comments** — a grep finds no code use. They mirror the live `DEFAULT_LIMIT_*` constants that #2610 uses for joint auto-promotion, but shape auto-promotion was intentionally dropped (`solver_mujoco.py:7019-7032`), so nothing consumes them. Recommend deleting all six (and trimming the comments that reference `DEFAULT_SHAPE_KE`/`KD`), or adding a note that they're intentional scaffolding for a future feature.

newton/_src/usd/schemas.py
```diff
@@ -346,11 +346,25 @@ class SchemaResolverMjc(SchemaResolver):
...
+            # ``parse_solref_mode_usd`` transformer), while these mappings
+            # also populate Newton ``shape_material_ke`` / ``kd`` via the
+            # lossy ``solref_to_stiffness_damping`` conversion so that
+            # imported MuJoCo defaults land at ``DEFAULT_SHAPE_KE`` /
+            # ``DEFAULT_SHAPE_KD``. The MuJoCo solver picks ``mujoco.solref``
+            # over ``shape_material_ke`` whenever ``solref_mode`` is
+            # ``SOLREF_MODE_RAW``, so the (possibly acceleration-space)
+            # value sitting in ``shape_material_ke`` does not affect
+            # simulation until the user explicitly opts into force-space
+            # scaling (see ``SolverMuJoCo`` auto-promote logic on
```
adenzler-nvidia: This points readers to "`SolverMuJoCo` auto-promote logic on `SOLREF_MODE_MJCF_DEFAULT`" for shapes, but no such logic exists — the NOTE at `solver_mujoco.py:7019-7032` explicitly states shape auto-promotion was omitted and force-space is strictly opt-in. Please remove/reword this cross-reference so it doesn't contradict the authoritative NOTE.

newton/_src/usd/schemas.py
```diff
@@ -359,6 +373,13 @@ class SchemaResolverMjc(SchemaResolver):
             # Contact models
             "priority": SchemaAttribute("mjc:priority", 0),
             "weight": SchemaAttribute("mjc:solmix", 1.0),
+            # See PrimType.SHAPE comment: ``mjc:solref`` is mirrored into
+            # both ``mujoco.solref`` (raw, authoritative for the solver
+            # under ``SOLREF_MODE_RAW``) and Newton's force-space
+            # ``stiffness`` / ``damping`` (so the auto-promote check in
```
adenzler-nvidia: Same issue: "the auto-promote check in `SolverMuJoCo`" does not exist for shapes/materials (nothing compares `shape_material_ke` against `DEFAULT_SHAPE_KE`). Drop the auto-promote rationale; the only true statement is that the raw value is authoritative under `SOLREF_MODE_RAW`.

newton/_src/utils/import_mjcf.py
```diff
@@ -711,8 +713,20 @@ def parse_shapes(
...
+            # MJCF ``solref`` is stored two ways for back-compat with the
+            # joint-limit pattern (PR #2610):
+            #   * Raw ``(timeconst, dampratio)`` lands in ``mujoco.solref``
+            #     via the registered ``mjcf_attribute_name="solref"``;
+            #     ``mujoco.solref_mode`` is set explicitly below.
+            #   * The lossy ``solref_to_stiffness_damping`` conversion still
+            #     populates Newton's ``shape_material_ke`` / ``kd`` so that
+            #     ``SOLREF_MODE_MJCF_DEFAULT`` shapes start at
+            #     ``DEFAULT_SHAPE_KE`` / ``DEFAULT_SHAPE_KD`` (which the
+            #     ``SolverMuJoCo`` auto-promote check uses to decide whether
```
adenzler-nvidia: "(which the `SolverMuJoCo` auto-promote check uses to decide whether the user has edited the gains)" describes joint-limit behavior that has no shape equivalent. Recommend dropping that clause — the factual part (the conversion still populates `shape_material_ke`/`kd` for back-compat) stands on its own.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -6925,6 +7013,24 @@ def _update_geom_properties(self):
         mujoco_attrs = getattr(self.model, "mujoco", None)
         shape_geom_solimp = getattr(mujoco_attrs, "geom_solimp", None) if mujoco_attrs is not None else None
         shape_geom_solmix = getattr(mujoco_attrs, "geom_solmix", None) if mujoco_attrs is not None else None
+        shape_mjc_solref = getattr(mujoco_attrs, "solref", None) if mujoco_attrs is not None else None
+        shape_mjc_solref_mode = getattr(mujoco_attrs, "solref_mode", None) if mujoco_attrs is not None else None
+
+        # NOTE on auto-promote (issue #2009): the joint-limit fix in
```
adenzler-nvidia: The decision not to auto-promote here is well reasoned. Related runtime gap: when a user sets `SOLREF_MODE_FORCE_SPACE` but runs `use_mujoco_contacts=True` or the MuJoCo CPU backend, the per-contact override never fires and the shape silently falls back to the lossy `convert_solref(ke,kd,1,1)` — i.e. the exact #2009 bug, with no signal. This file already uses `warnings.warn(...)` for unsupported features in several places; a single startup warning when any shape is FORCE_SPACE on an unsupported backend would surface it instead of leaving it only in the CHANGELOG/comments.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -687,6 +703,47 @@ def parse_solreflimit_mode_usd(_value: Any, context: dict[str, Any]) -> int | No
         # endregion geom attributes
 
         # region body and joint attributes
+        builder.add_custom_attribute(
+            ModelBuilder.CustomAttribute(
+                name="solref",
```
adenzler-nvidia, line 698: Two minor notes on these two new SHAPE-frequency attributes:

1. They're registered under `# region body and joint attributes` (line 705), just after `# endregion geom attributes` (703) — but the sibling geom attrs `geom_solimp`/`geom_solmix` sit above the endregion marker. Move these up so the region comments stay accurate.
2. Naming: adjacent geom attrs use a `geom_` prefix (`geom_solimp`, `geom_solmix`), while this is bare `solref`/`solref_mode`. Since `model.mujoco.solref` is user-visible, `geom_solref` would be more consistent and avoids a future breaking rename. (Defensible either way — #2610's joint attr is bare `solreflimit`.)

adenzler-nvidia (commented):
## Review summary — force-space `shape_material` contact scaling (#2009)

Overall the design is sound and mirrors the #2610 joint-limit pattern closely (negative-direct-mode `solref = (-ke·factor, -kd·factor)`, `factor = M_inv·(1-dmax)` with `M_inv = invweight0[A]+invweight0[B]`), kernel indexing is correct, and the attribute-absent / `MJCF_DEFAULT` fallbacks preserve pre-PR behavior for builder-API models. The deliberate divergence from #2610 (no shape auto-promote) is well reasoned and documented.

Two things should be addressed before merge — see inline comments:

- **Critical:** the three `test_force_space_contact_*` tests don't exercise the new override (they run with the default `use_mujoco_contacts=True`, where `convert_newton_contacts_to_mjwarp_kernel` never executes, and pass `contacts=None`). They pass purely from static equilibrium and would still pass with the override deleted.
- **Important:** six new `DEFAULT_SHAPE_*` constants are unused (referenced only in comments), and several comments point to a shape "auto-promote check" that does not exist — contradicting the authoritative NOTE in `_update_geom_properties`.

A few non-blocking consistency notes that don't map cleanly to a diff line:

- `SOLREF_MODE_*` docstrings in `constants.py:65-72` still say "joint-limit" only, but the constants are now shared with shape contacts — generalize them.
- The same multi-paragraph "stored two ways" explanation is duplicated across `schemas.py` (SHAPE + MATERIAL), `import_mjcf.py`, and the `solref_mode` registration; pick one canonical location and point the others at it (the MATERIAL block already does "See PrimType.SHAPE comment").
- Missing test coverage for the genuinely new logic: a mixed-mode contact (FORCE_SPACE vs RAW/MJCF_DEFAULT, where the override must *not* fire), the `m_inv == 0` / `dmax >= 1.0` guard boundaries, and the `mix` blend of heterogeneous FORCE_SPACE `ke`/`kd`. #2610's joint suite asserts directly on the computed `jnt_solref`; an equivalent direct assertion on the per-contact `solref` would be the highest-value addition.

newton/_src/solvers/mujoco/kernels.py
```diff
@@ -395,6 +400,48 @@ def convert_newton_contacts_to_mjwarp_kernel(
...
+        # two-body contact. MuJoCo's solver uses an effective inverse mass
+        # ``M_inv = body_invweight0[A] + body_invweight0[B]`` along the
+        # constraint, but ``contact_params``' averaging cannot reproduce that
+        # sum from per-geom values. Compute ``solref`` directly from Newton's
+        # force-space ``shape_material_ke`` / ``shape_material_kd`` using
+        # MuJoCo's negative-direct-mode convention so the per-contact
+        # ``stiffness * x = ke * x`` holds independent of the partner body's
+        # mass. Mixed-mode contacts (RAW or MJCF_DEFAULT on either side) fall
+        # through to the existing ``geom_solref`` mixing path.
+        if shape_mjc_solref_mode:
```
adenzler-nvidia, line 407: This new override is the core of the PR but is the least-tested part. Three coverage gaps worth closing:

- **Mixed-mode** (FORCE_SPACE vs RAW/MJCF_DEFAULT): the override only fires when *both* shapes are FORCE_SPACE (line 418); the fall-through to `geom_solref` mixing is untested, yet it's the common real config (force-space box on imported MJCF floor).
- **Guard boundaries** `m_inv == 0` and `dmax >= 1.0` (line 441): untested; #2610 has the analogous `dmax==1` fallback test.
- **`mix` blend** of heterogeneous `ke`/`kd` (lines 426-427): brand-new logic (the `mix` return was added by this PR), no test pairs differing materials/solmix.

Also, the comment above (403-414) reads as if the computed `solref` is final, but the `if rigid_contact_stiffness:` block just below (449-462) overwrites it when per-contact stiffness is supplied — worth one sentence noting that precedence.

CHANGELOG.md
```diff
@@ -206,6 +206,7 @@
 - Fix MJCF importer creating finite planes from MuJoCo visual half-sizes instead of infinite planes
 - Honor authored `mujoco.solreflimit_mode` even when a non-zero `mujoco.solreflimit` is also present, so the explicit mode (force-space or raw) is authoritative
 - Fix `SolverMuJoCo` CPU backend overwriting `mj_model.body_iquat` with Newton's eigendecomposition result on every `BODY_INERTIAL_PROPERTIES` notify; the compiled principal-axes basis is now preserved, fixing single-contact box equilibrium (incorrect normal force) and stiff WELD-loop instabilities (`Nan, Inf or huge value in QACC`) caused by basis ambiguity on repeated eigenvalues
+- Fix `SolverMuJoCo` honoring force-space `shape_material_ke` / `shape_material_kd` for contacts (`use_mujoco_contacts=False`); authored `mjc:solref` is preserved via new `mujoco.solref` / `mujoco.solref_mode` per-shape custom attributes. Force-space scaling is unsupported on `use_mujoco_contacts=True` and the MuJoCo CPU backend.
```
adenzler-nvidia, line 209: Flagging the simulation-behavior change for the "breaking changes" question: for MJCF/USD assets with *authored* `solref`, RAW mode now forwards the authored `(timeconst, dampratio)` verbatim into `geom_solref` instead of the prior lossy `ke→convert_solref` round-trip — including on the default `use_mujoco_contacts=True` path. This shifts contact dynamics for existing imported assets. It's the intended fix (not a regression), but it is a behavior change; the entry correctly lives under Fixed.

newton/_src/solvers/mujoco/kernels.py
vreutskyy: Done in `c3a2019e`. Added a new `Shape-material contact stiffness and damping` subsection to `docs/integrations/mujoco.rst` covering the generic mechanism (force-space gains, per-contact factor, three modes, unsupported-backend caveat) and trimmed the long block comments in `kernels.py`, `solver_mujoco.py`, `schemas.py`, `import_mjcf.py` down to one-line cross-refs. Net `-51` lines of code comments for `+45` lines of doc.

newton/tests/test_mujoco_solver.py
vreutskyy: Caught — thanks. Rewritten in `8f8f9450` to mirror the joint-limit suite pattern (`test_joint_limit_solref_uses_dof_invweight0`). All three tests now construct with `use_mujoco_contacts=False`, step the model until contacts are generated, and read `mjw_data.contact.solref` back directly against `(-ke·factor, -kd·factor)` with `factor = (body_invweight0[A] + body_invweight0[B]) · (1 − dmax)`. Sign sanity assertions explicitly forbid the legacy `convert_solref(ke, kd, 1, 1)` positive `(timeconst, dampratio)` fallback. Coverage extended per A9 below.

newton/tests/test_mujoco_solver.py
vreutskyy: Fixed in `8f8f9450`. The class docstring now correctly states the tests pin `use_mujoco_contacts=False` and read `mjw_data.contact.solref` back directly, matching the new test bodies.

newton/_src/solvers/mujoco/constants.py
vreutskyy: All six constants deleted in `8f8f9450`. The stale "auto-promote" references in `solver_mujoco.py:6988`, `import_mjcf.py`, and `schemas.py` were trimmed at the same time so nothing in the tree points to the removed symbols.

newton/_src/usd/schemas.py
vreutskyy: Fixed. The comment-trim pass in `c3a2019e` replaced this block with a one-line cross-ref to the docs section, and the docs section itself says "force-space scaling is opt-in" without claiming any auto-promote mechanism — the stale rationale is gone.

newton/_src/usd/schemas.py
vreutskyy: Fixed in `c3a2019e`: the `PrimType.MATERIAL` comment was shortened to "See `PrimType.SHAPE` above for the `mjc:solref` → stiffness/damping back-compat mirror." The shape-side comment itself no longer claims any auto-promote check exists.

newton/_src/utils/import_mjcf.py
vreutskyy: Dropped the auto-promote clause across both passes (`c3a2019e` trimmed the block to a cross-ref; `8f8f9450` finalised the wording to "for back-compat with the legacy `convert_solref(ke, kd, 1, 1)` round-trip"). Same wording is now used in `schemas.py` for consistency.

newton/_src/solvers/mujoco/solver_mujoco.py
vreutskyy: Added the startup warn in `8f8f9450`. When any shape carries `SOLREF_MODE_FORCE_SPACE` and the solver is constructed with `use_mujoco_contacts=True` or `use_mujoco_cpu=True`, `SolverMuJoCo.__init__` fires one `warnings.warn(..., stacklevel=2)` listing the affected shape count, the unsupported backend(s), and pointing at the docs section. Two regression tests cover the warn / no-warn paths.

newton/_src/solvers/mujoco/solver_mujoco.py
vreutskyy: Re-grouped in `8f8f9450`: `mujoco.solref` and `mujoco.solref_mode` are now registered *inside* the `# region geom attributes` block (next to `geom_solimp` / `geom_solmix`), so the region comments stay accurate.

On naming: left the bare `solref` / `solref_mode` to match the joint-limit precedent from #2610 (`solreflimit`, not `joint_solreflimit`) — a rename to `geom_solref` / `geom_solref_mode` would split the convention across the two PRs. Happy to take a follow-up if you'd rather rename both `solreflimit` and `solref` to their prefixed forms in a single sweep.

newton/_src/solvers/mujoco/kernels.py
vreutskyy: All three gaps covered in `8f8f9450`:

- **Mixed-mode** (`test_force_space_contact_mixed_mode_falls_through_to_geom_solref`): FORCE_SPACE box on a MJCF_DEFAULT plane; the override is bypassed and the resulting `contact.solref` is positive (`convert_solref` fallback), proving the both-FORCE_SPACE guard at `kernels.py:418` is honoured.
- **dmax guard** (`test_force_space_contact_solref_dmax_one_disables_override`): pin `geom_solimp[:, 1] = 1.0` so the mixed contact dmax forces `factor = m_inv·(1−dmax) = 0`; verifies the kernel discards the override and keeps the per-geom mixed `(timeconst, dampratio)`. (`m_inv == 0` is the static-vs-static case, which the immovable-pair guard already discards before reaching the override; happy to wire a third test if you'd like an explicit assertion.)
- **Heterogeneous mix** (`test_force_space_contact_solref_heterogeneous_mix`): two FORCE_SPACE shapes with `(ke_a, kd_a) = (5e3, 50)` and `(ke_b, kd_b) = (2e4, 200)`; asserts the resulting `contact.solref` matches the `mix=0.5` blend and is strictly between the two single-shape endpoints, proving the `mix` reuse from `contact_params`.

Also added the precedence note above the `rigid_contact_stiffness` block at `kernels.py:431` — per-contact stiffness still wins over the shape-material override.

CHANGELOG.md
vreutskyy: Acknowledged — yes, this is the intended behaviour change: RAW shapes now forward their authored `(timeconst, dampratio)` verbatim into `geom_solref` on both contact paths (including `use_mujoco_contacts=True`), where the previous code re-derived it via the lossy `ke→convert_solref` round-trip. The entry sits under `Fixed` because the round-trip was the #2009 bug, but flagging it on the breaking-changes question is the right call; happy to move it or add a "Changed" entry if you prefer.

eric-heiden (approved):
Thanks, this looks good to me now!

PR #2965, opened 2026-05-27
Change Control.joint_target_pos layout from joint_dof_count to joint_coord_count

### LINKED ISSUE #2556: Change Control.joint_target_pos layout from joint_dof_count to joint_coord_count
Control.joint_target_pos is currently shaped (joint_dof_count,) — the same layout as joint_qd (velocities). However, position targets semantically correspond to positions (joint_q), which has a different layout when the articulation contains free joints (7 coords: 3 translation + 4 quaternion) or ball joints (4 coords: quaternion) vs their velocity DOFs (6 and 3 respectively).

The actuator library uses pos_indices (derived from joint_q_start, i.e. joint_q layout) to index into both state.joint_q and control.joint_target_pos. When an articulation contains any free or ball joint, all downstream joint_q_start values diverge from joint_qd_start — so every actuated revolute/prismatic joint further along the chain will have mismatched pos_index vs the joint_dof_count-sized joint_target_pos array. This means the PD position term reads from the wrong index or goes out of bounds.
Proposed fix:

Add a flag (e.g. coord_layout_targets: bool) to Model / Control.
When True, joint_target_pos is allocated as (joint_coord_count,) matching joint_q.
Default to False (current behavior) with a DeprecationWarning.
Update all solver paths that read joint_target_pos to respect the flag.
In a subsequent release, flip the default to True, then remove the flag.

jvonmuralt:
## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **New Features**
  * Opt-in flag to enable coordinate-layout joint targets; new primary joint-target buffers joint_target_q and joint_target_qd.

* **Deprecations**
  * Legacy joint_target_pos / joint_target_vel deprecated — emit warnings or become unavailable when coord-layout is enabled; builder-level positional aliases removed.

* **Documentation**
  * API, migration guide, concepts and tutorials updated with indexing, layout semantics, and migration steps.

* **Examples & Benchmarks**
  * Examples and benchmarks updated to use coord-layout targets when enabled.

* **Tests**
  * Tests expanded to validate both layouts, conversions, and deprecation behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2965?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


adenzler-nvidia:
The `joint_target_q` / `joint_target_qd` rename and the layout fix are the right direction. I'd suggest inverting the rollout, though: rather than shipping the flag default-`False` (legacy preserved, opt *in* to the fix, flip later, remove later — two future transitions), **flip the default to `True` now and mark `use_coord_layout_targets` as deprecated from day one.** This is a bugfix, so the correct layout should be the default; the flag's only job is a temporary, opt-*out* backward-compat hatch, and being born deprecated means its whole lifetime is "remove me," so removal is clean (and a module-attr setter degrades to a harmless no-op anyway). One transition instead of two.

Three things to make that work:

1. **Scope the legacy accessor by the model — don't raise unconditionally under `True`.** As written, `joint_target_pos` raises `AttributeError` whenever the snapshot is `True`, so flipping the default would break *all* position-target code, including fixed-base models where `coord == dof` and nothing actually changed. Instead:
   - `coord == dof`: keep `joint_target_pos` as a soft-deprecated alias of `joint_target_q` (identical shape, just a `DeprecationWarning`) — the majority keeps working and migrates by rename.
   - free/ball present: raise `AttributeError` pointing at `joint_target_q` (shapes genuinely diverge; aliasing would silently corrupt dof-indexed code).

   So the flip is non-breaking for everyone except the population the bug affected, and loud for them rather than silent.

2. **Deprecate the `ModelBuilder` attributes too — don't remove them outright.** `ModelBuilder.joint_target_pos` / `joint_target_vel` are deleted with no shim, while the `Model`/`Control` versions get a deprecation path. That's inconsistent and conflicts with the deprecate-before-removing guideline. Keep them one release as warning-emitting aliases forwarding to `joint_target_q` / `joint_target_qd` (`joint_target_vel` → `qd` is a same-shape alias; `joint_target_pos` can follow the same scoped rule), unless there's a concrete reason they can't be.

3. **Keep the default path warning-free.** Only the legacy names and an explicit `use_coord_layout_targets = False` should warn — an untouched default-`True` user should see nothing.

Migration guidance can then be: rename `pos → q` / `vel → qd` (including on `ModelBuilder`); if you need the old layout temporarily, set `newton.use_coord_layout_targets = False` (deprecated, slated for removal next release).

Minor: the snapshot is stored publicly on `Model` (`self.use_coord_layout_targets`) but privately on `Control` — make it private in both. And this reads as `Fixed` + `Deprecated` rather than `Added`.

jvonmuralt:
@adenzler-nvidia  Thanks! Not sure about flipping the default to True,  users with a free or ball joint in their articulation will get an error in their code which touches joint_target_pos. With the current version their code would keep working but emit a deprecation warning. If we set the flag to True there is no need for the flag in general, or let me know if I am missing something.

adenzler-nvidia:
Fair point - I have been going back and forth about it. Here's my reasoning: whoever has been doing this right now had buggy code anyway, so it's really a bugfix. In that sense, fixing it and giving existing users a good way to back out of it sounds reasonable and will accelerate the deprecation timeline.

But I'm also ok with just having it default to False for now and flipping it. We can discuss this in the wider group, I feel like breaking change is ok if there is a way to back out in this case.

jvonmuralt:
@adenzler-nvidia  Why is this a bugfix though? It's mainly a layout/convention switch. The only real bug fixed along the way is in the mujoco solver, users setting Euler-angle targets on ball/free joints would have gotten slightly wrong behaviour because mujoco internally interpreted those 3 slots as axis-angle. Otherwise the PR is just layout switch: joint_target_pos (DOF-shaped) -> joint_target_q (coord-shaped) so it matches joint_q.
For revolute/prismatic users nothing was broken.

adenzler-nvidia:
Fair enough — let's go with `use_coord_layout_targets = False` as the default and treat this as a convention switch rather than a bugfix. Deprecation warning now, flip the default later. That's a clean one-way migration.

One thing to fix before merge: `ModelBuilder.joint_target_pos` / `joint_target_vel` are hard-removed (they raise `AttributeError` unconditionally, regardless of the flag), while the `Model`/`Control` versions get a proper deprecation path. These are public API, so they need a deprecation cycle regardless of how heavily they're used — removing them outright breaks existing builder code immediately even under the default `False`, which is exactly what we're avoiding here and conflicts with the deprecate-before-removing guideline.

I realize a writable DOF alias is awkward since the builder now stores `joint_target_q` coord-shaped and only projects to DOF at `finalize()`. But let's make them read-shims that return the projected DOF list and emit a `DeprecationWarning` under `False` (raising only under `True`), matching `Model`/`Control`. Also worth fixing the docstring while you're there — it says "deprecated" but the message says "was removed."

newton/_src/sim/model.py
```diff
@@ -896,8 +1015,10 @@ def control(self, requires_grad: bool | None = None, clone_variables: bool = Tru
             requires_grad = self.requires_grad
         if clone_variables:
             if self.joint_count:
-                c.joint_target_pos = wp.clone(self.joint_target_pos, requires_grad=requires_grad)
-                c.joint_target_vel = wp.clone(self.joint_target_vel, requires_grad=requires_grad)
+                if self.joint_target_q is not None:
```
eric-heiden, line 1006: 🔴 `Model.control()` still lets `Control()` snapshot the current module-global layout flag instead of the layout stored on this `Model`.

That breaks the contract documented on `Model`: once a model is finalized, later changes to `newton.use_coord_layout_targets` should not change how that model's control arrays are interpreted. With the current code, a legacy-layout model can return a `Control` whose deprecated aliases raise after the global is toggled to coord layout, and a coord-layout model can return a `Control` whose `joint_target_pos` alias exposes the coord-shaped quaternion buffer after the global is toggled back.

Please copy the model snapshot onto the control object before returning it, for example by passing `self.use_coord_layout_targets` into `Control` or assigning `c._use_coord_layout_targets = self.use_coord_layout_targets` in `Model.control()`.

newton/_src/solvers/kamino/_src/core/control.py
```diff
@@ -123,67 +123,73 @@ def finalize(self, model: ModelKamino, device: wp.DeviceLike = None) -> None:
...
-            self._q_j_ref_coords_space = wp.zeros(
-                shape=model.size.sum_of_num_joint_coords,
-                dtype=float32,
-                device=device,
-            )
-        else:
-            self._needs_coord_conversion = False
-            self._q_j_ref_coords_space = None
+        self._needs_coord_conversion = (
+            not newton.use_coord_layout_targets
```
eric-heiden: 🔴 `ControlKamino.finalize()` should not read `newton.use_coord_layout_targets` here.

The underlying Newton `Model` has already snapshotted the target layout at finalize time, so a later global toggle can make Kamino disagree with the model it is wrapping. For example, if the model was built with coord-layout targets and the global is later set to `False`, Kamino enables Euler-to-quaternion conversion even though `control.joint_target_q` is already coord-shaped. In the reverse case, Kamino skips conversion while the control buffer is still DOF-shaped.

Please read the layout snapshot from the underlying Newton model, such as `model.use_coord_layout_targets` if exposed through `ModelKamino`, instead of consulting the mutable global flag.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -3099,6 +3107,9 @@ def __init__(
         For CTRL_DIRECT: MJCF-order index into control.mujoco.ctrl array
 
         Shape [nu], dtype int32."""
+        self.mjc_actuator_to_newton_target_q_idx: wp.array[wp.int32] | None = None
+        """Coord-layout index into :attr:`Control.joint_target_q` for position actuators
```
eric-heiden: 🔴 `mjc_actuator_to_target_q_axis_idx` is used later by the solver but is never declared in `SolverMuJoCo.__init__`.

That means paths that reach the kernel launch before `_build_mjc_model()` assigns this attribute can raise `AttributeError` instead of seeing the normal `None` sentinel used by the neighboring mappings. This should be initialized next to `mjc_actuator_to_newton_target_q_idx`.

The docstring here also says this is a coord-layout index, but under the legacy flag the index follows `model.joint_target_q_start`, which is DOF-layout. Please make the wording layout-aware instead of coord-specific.

newton/_src/actuators/actuator.py
```diff
@@ -145,8 +159,19 @@ def __init__(
 
         self.state_pos_attr = state_pos_attr
         self.state_vel_attr = state_vel_attr
-        self.control_target_pos_attr = control_target_pos_attr
-        self.control_target_vel_attr = control_target_vel_attr
+        if control_target_pos_attr is None or control_target_vel_attr is None:
```
eric-heiden, line 162: 🔴 The default actuator path still chooses deprecated control attribute names when `newton.use_coord_layout_targets` is `False`.

`ModelBuilder.finalize()` passes these attributes as `None`, so builder-created actuators default to `joint_target_pos` and `joint_target_vel`. Then `Actuator.step()` reads those properties every frame, which emits `DeprecationWarning` from Newton's own internal actuator path even though `joint_target_q` and `joint_target_qd` exist in both layouts.

Please default new actuators to the canonical `joint_target_q` and `joint_target_qd` attributes in both layouts. Only the target index mapping needs to vary between DOF and coord indexing.

newton/_src/sim/control.py
```diff
@@ -67,10 +152,10 @@ def clear(self) -> None:
             self.tet_activations.zero_()
         if self.muscle_activations is not None:
             self.muscle_activations.zero_()
-        if self.joint_target_pos is not None:
-            self.joint_target_pos.zero_()
-        if self.joint_target_vel is not None:
-            self.joint_target_vel.zero_()
+        if self.joint_target_q is not None:
```
eric-heiden, line 152: 🔴 `Control.clear()` now zeroes `joint_target_q`, which leaves coord-layout quaternion targets invalid.

For FREE and BALL joints in coord layout, the target slices include quaternions. After `clear()`, those slots become `(0, 0, 0, 0)` instead of identity quaternions, so later simulation code can consume invalid rotation targets.

I think we can assign (copy) `Model.joint_target_q` here since this is what we use to initialize the `Control` object.

newton/_src/actuators/actuator.py
jvonmuralt: Didn't make the default unconditionally canonical because that would silently break callers passing custom Control objects under flag=False. Instead:

ModelBuilder.finalize() constructs builder-side actuators with explicit "joint_target_q" / "joint_target_qd" — Newton's own actuators never touch the deprecated property in either layout.
The Actuator.__init__ default stays None -> flag-conditional, with a DeprecationWarning under False telling users to pass "joint_target_q" explicitly. Keeps custom-Control users working today while signalling the upcoming flip.

eric-heiden: Sounds good!

eric-heiden (approved):
Thanks for addressing my comments!

Guirec-Maloisel (approved):
The changes look good from the Kamino side!

PR #2971, opened 2026-05-28
Add intersect_ray function that uses BVH accelerated raycasting, Remove SensorRaycast

### LINKED ISSUE #2465: [REQ] BVH accelerated raycast sensor
### Description

Raycasting in newton currently uses a brute force approach. This limits performance for raycast sensors. BVH accelearted raycast logic is already being used in the warp raytracer and can be extended to traditional raycasts as well. To accomplish this we would need:

- BVH compatible raycast functions that accept a scene level BVH
- Integration with raycast sensors similar to camera sensor

### Motivation / Use Case

BVH-Accelerated raycast sensors would allow for downstream locomotion tasks to be trained an order of magnitude faster

### Alternatives Considered

_No response_

StafaH:
# Add `intersect_ray`, heightfield mesh raycast, remove `SensorRaycast`

### `newton.intersect_ray()` — new public raycast API

Adds `newton.intersect_ray()`, a functional helper for casting batched rays against all shapes in a model across multiple worlds. It uses the shape BVH for broad-phase culling and dispatches to per-shape intersection routines (primitives, meshes, and heightfields).

```python
newton.intersect_ray(
    model,           # Model — must have a built BVH (build_bvh_shape)
    ray_origins,     # wp.array2d[wp.vec3]  shape [world_count, ray_count]
    ray_directions,  # wp.array2d[wp.vec3]  shape [world_count, ray_count]
    out_dist,        # wp.array2d[float]    output hit distances (-1 = miss)
    out_shape_id,    # wp.array2d[wp.int32] output hit shape indices
    out_normal,      # wp.array2d[wp.vec3]  output surface normals at hit
)
```

This is the recommended building block for custom raycast sensors, lidar simulations, and picking — replacing ad-hoc kernel launches against internal kernel symbols.

### Heightfield raycast via `wp.Mesh`

Previously `GeoType.HFIELD` shapes used a per-thread 2D DDA grid traversal in the raycast kernel, requiring three extra arrays (`shape_heightfield_index`, `heightfield_data`, `heightfield_elevations`) threaded through every raycast kernel signature and a factory pattern (`_make_raycast_funcs`) to generate two compiled variants.

Now `ModelBuilder.finalize()` builds a `wp.Mesh` from each heightfield (same CCW triangulation as the collision path) and stores its ID in `model.shape_source_ptr`, exactly like `MESH` and `CONVEX_MESH` shapes. `ray_intersect_geom` handles `HFIELD` in the same branch as meshes.

Benefits:
- Raycast kernel signatures shrink by 3 parameters; no more two-variant factory
- BVH-accelerated ray-triangle intersection replaces per-thread grid walking
- Same `Heightfield` object reuses one `wp.Mesh` across instances; per-instance scale is applied at query time
- `model.heightfield_meshes` keeps the `wp.Mesh` objects alive; `heightfield_data`/`heightfield_elevations` are unchanged (still used by collision kernels)

### Remove `SensorRaycast`

`SensorRaycast` was deprecated in 1.2 in favour of `SensorTiledCamera`. This PR removes it along with the `sensor_raycast_kernel`, `sensor_raycast_particles_kernel`, and `sensor_raycast_kernel_no_hfield` that existed solely to support it. `ray_intersect_particle_sphere` is retained — still used by `warp_raytrace`.

Migration: use `SensorTiledCamera` with `SensorTiledCamera.utils.compute_pinhole_camera_rays()` and `create_depth_image_output()`.

Closes https://github.com/newton-physics/newton/issues/2465.

## Checklist

## Test plan

```bash
# Raycast tests (28 tests — all heightfield cases now use wp.Mesh path)

# Full suite — no regressions
```

## New feature / API change

```python
import newton
import numpy as np
import warp as wp

builder.begin_world()
body = builder.add_body(xform=wp.transform(wp.vec3(0.0, 0.0, 0.0), wp.quat_identity()))
builder.add_shape_sphere(body=body, radius=0.5)

elevation = np.random.rand(32, 32).astype(np.float32) * 2.0
hf = newton.Heightfield(data=elevation, nrow=32, ncol=32, hx=5.0, hy=5.0)
builder.add_shape_heightfield(heightfield=hf)
builder.end_world()

model = builder.finalize()
state = model.state()
newton.geometry.build_bvh_shape(model, state)

# Cast two rays in one call: one toward the sphere, one toward the terrain.
origins = wp.array([[[- 2.0, 0.0, 0.0], [0.0, 0.0, 8.0]]],  dtype=wp.vec3)
directions = wp.array([[[1.0, 0.0, 0.0], [0.0, 0.0, -1.0]]], dtype=wp.vec3)

out_dist     = wp.empty(shape=(1, 2), dtype=float)
out_shape_id = wp.empty(shape=(1, 2), dtype=wp.int32)
out_normal   = wp.empty(shape=(1, 2), dtype=wp.vec3)

newton.intersect_ray(model, origins, directions, out_dist, out_shape_id, out_normal)

print(out_dist.numpy())     # [[1.5, <terrain_dist>]]
print(out_shape_id.numpy()) # [[0, 1]]
print(out_normal.numpy())   # [[-1, 0, 0], <terrain_normal>]
```

## Summary by CodeRabbit

* **New Features**
  * Added intersect_ray() API for custom raycast queries.
* **Improvements**
  * Heightfield raycasting now uses mesh-based BVH; heightfields are converted to retained meshes during model finalization.
* **Documentation**
  * API docs updated to include intersect_ray and remove SensorRaycast entries.
* **Deprecations / Removals**
  * SensorRaycast removed from public exports; use SensorTiledCamera for pinhole ray/depth output.
* **Tests**
  * Raycast tests updated for new interfaces; legacy SensorRaycast tests removed.


newton/_src/sim/builder.py
```diff
@@ -9905,13 +9946,20 @@ def finalize(
...
             finalized_geos = {}  # do not duplicate geometry
             gaussians = []
+            heightfield_meshes = []
             for geo in self.shape_source:
                 geo_hash = hash(geo)  # avoid repeated hash computations
-                if geo and not isinstance(geo, Heightfield):
+                if isinstance(geo, Heightfield):
+                    if geo_hash not in finalized_geos:
+                        hf_mesh = _build_heightfield_warp_mesh(geo, device=device)
+                        heightfield_meshes.append(hf_mesh)
```
daniela-hase: I think we could add the `_build_heightfield_warp_mesh` function to the `Heightfield` class and have it own the `wp.Mesh`, right? Then there is no need for the separate `heightfield_meshes` array here, nor in the model :)

StafaH: Good call, after checking Heightfield, I realized there's already a static function for building the hfield mesh. Going to use that function and remove the one I made. For mesh reference, we will have to keep it with the model, the original Heightfield data can be owned by other Models (there is a test that fails if we try to attach the mesh to the original Heightfield class).

daniela-hase: Oh I see, I wonder if we should do the same for the Meshes and Gaussians then 🤔 Because they aren't owned by the Model either.

But, that's got nothing to do with this PR - so all good :)

daniela-hase:
This looks all good to me! :D

Only question I'd have, do we still really need that raycast kernel? Because it really doesn't do anything that the render kernel doesn't do as well, does it?

It is only used for the viewer-picking as far as I can tell, right?

While I personally would probably be in favor of using the render-kernel for this (assuming the performance would be the same), I don't have a super uber strong opinion on this and can see reasons for keeping it separated as well.

@mmacklin @eric-heiden do you guys have an opinion?

daniela-hase (dismissed):
Approving, but lets wait for @mmacklin / @eric-heiden to comment on my above before we merge please :)

newton/_src/geometry/raycast.py
```diff
@@ -692,489 +691,252 @@ def ray_intersect_mesh_no_transform(
...
+
+
+def intersect_ray(
+    model,
+    ray_origins: wp.array2d[wp.vec3],
+    ray_directions: wp.array2d[wp.vec3],
+    out_dist: wp.array2d[float],
+    out_shape_id: wp.array2d[wp.int32],
+    out_normal: wp.array2d[wp.vec3],
+) -> None:
```
eric-heiden: There is an issue when we assume 2D arrays for these inputs and outputs.
It is not possible to query global shapes with this because they are in [world -1](https://newton-physics.github.io/newton/stable/concepts/worlds.html#world-assignment).
If a caller wants to only query shapes in world `k` they have to allocate `k+1` inputs and outputs, and ignore the rows that come before the world they are interested in.

Perhaps a better interface is to explicitly let the user pass in `ray_worlds: wp.array[wp.int32]` and then the other arrays are one-dimensional?

Another API nit is that it might be better for future-proofing to use kwargs for most inputs as I'm trying to do in https://github.com/newton-physics/newton/pull/2993, so maybe everything after the `model` argument should be a kwarg. Could the outputs default to None so that the user can ignore outputs they don't need?

@mmacklin for feedback on some of these API questions.

newton/_src/geometry/raycast.py
```diff
@@ -692,489 +691,252 @@ def ray_intersect_mesh_no_transform(
...
+    if min_shape_id < 0:
+        out_dist[worldid, rayid] = -1.0
+    else:
+        out_dist[worldid, rayid] = min_dist
+    out_shape_id[worldid, rayid] = min_shape_id
+    out_normal[worldid, rayid] = min_normal
+
+
+def intersect_ray(
+    model,
```
eric-heiden: `model` needs a type hint.

newton/__init__.py
```diff
@@ -30,6 +30,7 @@
     ParticleFlags,
     ShapeFlags,
     TetMesh,
+    intersect_ray,
```
eric-heiden, line 33: 🟡 `intersect_ray` is now exported from both `newton.intersect_ray` and `newton.geometry.intersect_ray`, and the docs list both paths. That gives the same public helper two discoverable import locations, which makes it unclear whether ray queries are part of the top-level API or the geometry query API.

Please pick one canonical public path, i.e. `newton.intersect_ray`.

CHANGELOG.md
```diff
@@ -41,6 +43,7 @@
 
 ### Removed
 
+- Remove `SensorRaycast` (deprecated in 1.2); use `SensorTiledCamera` with `SensorTiledCamera.utils.compute_pinhole_camera_rays()` and `create_depth_image_output()` instead — see the removed class docstring for a complete migration example
```
eric-heiden: 🟡 This removal note points users to the removed `SensorRaycast` class docstring for the complete migration example, but this PR deletes that class and removes it from the generated API docs. Users upgrading from the released package will not have a stable place to read that example.

Please move the migration example into retained docs or into this changelog entry, then link or reference that stable location instead of the deleted docstring.

newton/_src/geometry/raycast.py
```diff
@@ -692,489 +691,252 @@ def ray_intersect_mesh_no_transform(
...
+        old_min = wp.atomic_min(min_dist, 0, t)
+        if t <= old_min:
+            min_index[0] = shape_idx
+            min_body_index[0] = b
+        _spinlock_release(lock)
 
 
-def _make_raycast_funcs(enable_heightfields: bool):
-    """Generate ``ray_intersect_geom`` + raycast kernels with HFIELD support gated.
+# Alias for backward compatibility.
```
eric-heiden: 🔴 This alias keeps `raycast_kernel_no_hfield` importable, but it does not preserve the old kernel launch signature. Existing callers that import `raycast_kernel_no_hfield` and launch it with the previous positional arguments, including `shape_heightfield_index`, `heightfield_data`, and `heightfield_elevations`, will still import successfully and then fail later at `wp.launch` with an argument-count mismatch.

That is a harder failure mode to diagnose than an immediate `ImportError`. Please either remove the alias, or keep it and add an explicit changelog note that `raycast_kernel_no_hfield` is now an alias for `raycast_kernel` and no longer accepts the old heightfield arguments.

newton/_src/sim/model.py
```diff
@@ -298,13 +298,15 @@ def __init__(self, device: Devicelike | None = None):
...
-        """True iff the model contains at least one ``GeoType.HFIELD`` shape. Lets launch sites pick lean kernel variants."""
+        """True iff the model contains at least one ``GeoType.HFIELD`` shape."""
         self.shape_heightfield_index: wp.array[wp.int32] | None = None
         """Per-shape heightfield index, shape [shape_count]. -1 means shape has no heightfield."""
         self.heightfield_data: wp.array[HeightfieldData] | None = None
         """Compact array of HeightfieldData structs, one per actual heightfield shape."""
         self.heightfield_elevations: wp.array[wp.float32] | None = None
         """Concatenated 1D elevation array for all heightfields. Kernels index via HeightfieldData.data_offset."""
+        self.heightfield_meshes: list = []
+        """wp.Mesh objects built from heightfield shapes, kept alive for the model's lifetime."""
```
eric-heiden, line 309: ⚪ The docstring says this stores `wp.Mesh` objects, so the element type should be in the annotation instead of only in prose. Nearby fields already use concrete annotations such as `wp.array[HeightfieldData] | None`.

Please change the declaration above this docstring to `self.heightfield_meshes: list[wp.Mesh] = []`.

docs/concepts/sensors.rst
```diff
@@ -103,8 +103,6 @@ attributes, and usage examples.
   with optional per-counterpart breakdown.
 * :class:`~newton.sensors.SensorFrameTransform` -- relative transforms of shapes/sites with respect to reference sites.
 * :class:`~newton.sensors.SensorIMU` -- linear acceleration and angular velocity at site frames.
-* :class:`~newton.sensors.SensorRaycast` -- *(deprecated)* ray-based depth images from a virtual camera; use
```
eric-heiden, line 106: 🟡 Removing the `SensorRaycast` bullet leaves this concepts section with four listed sensor types, but the paragraph above still says Newton provides five sensor types. That makes the public docs stale in the same PR that removes the public sensor.

Please update the count to four, or otherwise adjust the sentence and list so they match.

newton/_src/sim/builder.py
```diff
@@ -10025,13 +10025,30 @@ def finalize(
...
             gaussians = []
+            heightfield_meshes = []
             for geo in self.shape_source:
                 geo_hash = hash(geo)  # avoid repeated hash computations
-                if geo and not isinstance(geo, Heightfield):
+                if isinstance(geo, Heightfield):
+                    if geo_hash not in finalized_geos:
+                        # Transpose: create_heightfield uses ij-indexing (i=X, j=Y)
+                        # while Heightfield stores row-major data (row=Y, col=X).
+                        actual_heights = geo.min_z + geo.data * (geo.max_z - geo.min_z)
```
eric-heiden, line 10082: 🟡 The existing comment on `HeightfieldData` extent baking still says raycast reads from `HeightfieldData` and refers to the raycast DDA, but this PR moves raycasting to the `wp.Mesh` path created here. After this change, `HeightfieldData` is only relevant to the collision kernels.

Please update the `HeightfieldData` comment to mention narrow-phase collision only, and drop the raycast references. A short note pointing to this mesh-backed path for raycasting would also make the new split easier to follow.

docs/concepts/sensors.rst
StafaH: modified five -> four

CHANGELOG.md
StafaH: Removing the comment about docstring, the changelog should be sufficient on figuring out how to migrate

newton/_src/geometry/raycast.py
StafaH: Removed the alias and noted the breaking change, so that used can just use the correct name kernel

newton/_src/geometry/raycast.py
mmacklin: Agree that specifying world_ids in an array would be better, also because worlds might want to cast different numbers of rays per-batch.

Also would be good to decide on what the right way to handle the background / global world is. I would suggest a top-level flag on this function is enough, e.g.: `newton.intersect_ray(..., shape_flags)` (not necessarily per-ray). I imagine users will also want some way of determining which types of shapes to hit so a bitmask of `SHAPE_FLAGS` and some way of enabling global environment would probably make sense.

StafaH: Thanks @eric-heiden and @mmacklin, I've made the kernel be compatible with 1D inputs, and also enabled global shapes through an option. Also enforced keyword args.

For the shape filtering I'll leave that out for now, I can tackle that in the next round of changes, since I'll want to match and reduce code duplication with the tiled renderer, so I'll find a joint solution.

eric-heiden: Thanks @StafaH, this looks good to me! I see the issue with `enable_global_world`. But this option should be true by default I think because this is what happens in collision detection and (I think?) also in raytracing, so we might not even need this option.

StafaH: I defaulted to True. Thanks!

eric-heiden (dismissed):

eric-heiden (approved):

PR #2974, opened 2026-05-28
fix usd import of an orphaned fixed joint to world

rubengrandia:
## Description

This PR fixes the transform of orphaned fixed joints when importing a model with an environment offset. This shows up when using Kamino USD models with fixed joints (but without ArticulationRootAPI) in IsaacLab, with Isaaclab giving them an environment offset. 

The new unit test illustrates well what went wrong: The importer was ignoring the environment offset when determining the world location of the fixed joint. The changed importer will derive the fixed joint location from the child transform instead.

## Checklist

## Test plan

A minimal unit test was added
 ```
 uv run --extra dev -m newton.tests -k test_orphan_world_fixed_joint_respects_env_offset_and_xform
 ``` 

## Bug fix

**Steps to reproduce:**

The new unit tests fails with the old code. I will have a `joint_X_p = (5.0, 0.0, 0.0)`, which excludes the environment offset.

## Summary by CodeRabbit

* **Bug Fixes**
  * USD import now correctly resolves forward kinematics for orphaned body-to-world fixed joints by accounting for ancestor transform and spawn offsets, ensuring pinned bodies reach the intended world pose.
* **Tests**
  * Added a test verifying orphaned world-fixed joints respect environment offsets and spawn transforms.
* **Documentation**
  * Changelog updated to document the fix.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/2974?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,7 @@
 - Fix USD import losing authored negative scales on shape and parent xforms, so mirrored primitives and meshes are now imported with the correct signed scale
 - Fix rigid-rigid friction in `SolverVBD` for contacts with nonzero `rigid_contact_offset0/rigid_contact_offset1`.
 - Respect USD color-space metadata for scalar material colors and convert linear-authored USD color textures to display space when loading them
+- Fix USD import of orphaned body-to-world fixed joints not accounting for ancestor xform offsets, so pinned bodies now FK to the correct world pose (env origin + spawn xform)
 
 ## [1.2.0] - 2026-05-12
 
--- a/newton/_src/utils/import_usd.py
+++ b/newton/_src/utils/import_usd.py
@@ -2356,34 +2356,32 @@ def parse_usd(
         free_joints_auto_inserted = not (no_articulations and has_joints)
         if is_body_to_world and free_joints_auto_inserted and not is_fixed_joint:
             continue
-        # For body-to-world joints, `body0` is empty so there is no world-side
-        # prim to inherit a frame from. Authoring tools commonly write
-        # `localPose0` relative to the joint prim's USD ancestor Xform rather
-        # than in world coords, leaving the joint anchor offset by that
-        # ancestor's world transform. Recover the missing world-side frame
-        # from the child body's world pose and the joint's local poses so the
-        # joint chain FK reproduces the imported child world pose:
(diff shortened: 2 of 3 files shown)


adenzler-nvidia:
The math works out — I verified `parent_tf_o` algebraically cancels inside `parse_joint` (the `inv(parent_tf_o)` in `world_body_xform_o` is undone by `parse_joint`'s `parent_tf = incoming_xform * parent_tf` at L918), so the final anchor is `incoming_world_xform * child_world * child_tf` — `localPose0` doesn't enter the result. The framing in the comment is helpful; one small suggestion is to say that explicitly ("`localPose0` cancels; effectively we anchor at the USD-authored body pose, sidestepping however the world-side pose was authored") so a future reader doesn't have to rederive the cancellation.

One concrete behavior issue worth fixing: the new `resolve_joint_parent_child` call at L2358-2360 sits **outside** the existing `try:` at L2386. Previously the equivalent call lived inside `parse_joint`, inside that try, so its `ValueError("Unable to parse joint ... both bodies unresolved")` was caught at L2395 and logged in verbose mode only. After this change, that `ValueError` escapes and aborts the whole `parse_usd`. Moving the new block inside the `try:` (or wrapping the resolve call in its own try) preserves prior behavior on malformed joints.

Two follow-up suggestions that don't need to block this PR:
- Test variants with swapped `body0`/`body1` order (currently uncovered) and with non-identity `localPose0` / `localPose1` (the identity case in the new test hides the formula — any left-vs-right multiply or `inv` placement bug would still pass).
- A separate refactor that pushes the `parent_id == -1` anchoring rule into `parse_joint` itself, so any body-to-world joint that ends up in the articulation path also gets the fix.

rubengrandia:
@adenzler-nvidia , moved code into the `try:`, clarified the comment in the code, and extended the test to cover both `body0`, `body1` cases.

adenzler-nvidia (approved):
Thanks for the quick turnaround! All three points are addressed: the orphan-frame computation now lives inside the `try:` so a `ValueError` from `resolve_joint_parent_child` is caught and logged in verbose mode as before, the comment spells out the `localPose` cancellation, and the test now covers both `body0`/`body1` orientations with non-identity local poses (which is what exercises the formula). Verified the test fails without the fix and passes with it. LGTM.

PR #2980, opened 2026-05-28
Upgrade to mujoco_warp 3.10.0.1 and align margin/gap semantics

adenzler-nvidia:
## Description

Upgrade Newton's `mujoco` dependency to `~=3.10.0` and `mujoco-warp` to `~=3.10.0,>=3.10.0.1`. The `3.10.0.1` minimum includes the upstream sparse elliptic-cone fix from [google-deepmind/mujoco_warp#1456](https://github.com/google-deepmind/mujoco_warp/pull/1456).

This upgrade also adopts the margin/gap semantics introduced in MuJoCo 3.9. They now align with Newton's `shape_margin` and `shape_gap`: `margin` is the force-generation threshold and `gap` is an additional detection cushion. Newton's compensating translation layers are therefore removed.

Default behavior changes:

- Propagate `Model.shape_gap` into `mjw_model.geom_gap` and `pair_gap` at solver initialization and after `notify_model_changed`. MuJoCo's broadphase now honors the authored gap; CCD-handled contacts in `[margin, margin + gap)` are reported as sensor contacts without generating constraint forces.
- Set `contact.includemargin = margin` in `write_contact` instead of `margin - gap`, matching upstream's writer under the new gap semantics.
- Map MJCF and USD `margin` directly to `Model.shape_margin` instead of subtracting `gap`. Pass `legacy_margin_gap=True` to `ModelBuilder.add_mjcf()` or `ModelBuilder.add_usd()` to restore the previous translation for files authored against MuJoCo 3.8 or older.

MuJoCo 3.10 compatibility changes:

- Treat the deprecated `SolverMuJoCo(ls_parallel=...)` argument as a warning-only no-op because mujoco-warp removed the corresponding option in 3.9.1.
- Skip removed solver-option properties when copying options in the menagerie tests.
- Skip the new derived `dof_length` field in menagerie model comparisons, alongside the existing sparse-structure fields whose values depend on body ordering and inertia re-diagonalization.
- Require mujoco-warp 3.10.0.1 so the sparse Newton solver receives enough cone-Hessian support pairs for elliptic contacts.

Newton continues to zero `geom_margin` and `pair_margin` in the MJCF spec when using NATIVECCD or MULTICCD because mujoco-warp rejects non-zero margins for those paths.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k "mujoco"
uv run --extra dev -m newton.tests -k "import"
uv run --extra dev -m newton.tests -k "menagerie"
uvx pre-commit run -a
```

Dependency-update validation:

```bash
uv lock --check
uvx pre-commit run --files pyproject.toml uv.lock
```

Regression coverage includes:

- `test_includemargin_uses_margin_only`
- `test_geom_gap_forwarded_from_shape_gap`
- `test_pair_gap_not_zeroed_under_nativeccd`
- `test_mjcf_legacy_margin_gap_subtracts`
- `test_usd_margin_gap_identity_import`

## New feature / API change

```python
import newton

# Default: MuJoCo 3.9+ identity mapping for margin and gap.
builder.add_mjcf("model.xml")

# Restore the pre-3.9 translation: newton_margin = mj_margin - mj_gap.
builder.add_mjcf("model.xml", legacy_margin_gap=True)
builder.add_usd("scene.usda", legacy_margin_gap=True)
```


jcarius-nv:

newton/_src/solvers/mujoco/kernels.py
```diff
@@ -2309,7 +2309,7 @@ def update_geom_properties_kernel(
     if shape_geom_solmix:
         geom_solmix[world, geom_idx] = shape_geom_solmix[shape_idx]
 
-    geom_gap[world, geom_idx] = 0.0
+    geom_gap[world, geom_idx] = shape_gap[shape_idx]
```
vreutskyy, line 2471: [P1] Please account for the larger contact set before forwarding `shape_gap` here. With Newton's default `rigid_gap=0.1`, `nconmax=None` is still sized from the initial contact count, and `basic_plotting` now repeatedly reports broadphase overflow (49–56 contacts / 196–224 atomic contacts) on Linux, ARM, Windows, CPU, and GPU. Overflow clips contacts, so this is a simulation correctness issue rather than harmless output. Please make capacity/default-gap handling consistent with the expanded detection shell (or explicitly provision sufficient capacity) and cover the dynamic-contact case.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -5474,12 +5474,11 @@ def add_body_from_joint(j: int, *, mocap: bool | None):
...
-                body_kwargs["ipos"] = body_com[child, :]
-                # Use diaginertia when off-diagonals are exactly zero to preserve
-                # MuJoCo's sameframe optimization (body_simple=1).  fullinertia
-                # triggers eigendecomposition that reorders eigenvalues and applies
-                # a permutation rotation, setting body_simple=0 even for diagonal
-                # matrices whose entries are not in descending order.
+                body_ipos = body_com[child, :].copy()
+                compile_ipos = body_ipos.copy()
+                compile_ipos[0] += 1.0e-3 if compile_ipos[0] >= 0.0 else -1.0e-3
+                # A temporary COM offset forces qM storage that remains valid after inertia edits.
```
vreutskyy, line 5606: [P1] Please avoid forcing general `qM` storage for every positive-mass body. This disables the simple-body layout even when callers never randomize inertia, making the optional runtime-edit capability a cost paid by all simulations. The current GPU benchmark gate reports `FastAllegro.time_simulate` regressing from 814 ms to 1.14 s (1.40x) and G1 Newton overhead from 17.90 to 22.42 (1.25x). Gate this behind an explicit capability/opt-in or preserve the simple fast path for ordinary models.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -6204,7 +6205,13 @@ def add_body_equality(i: int):
             self.mjc_actuator_to_newton_ball_jnt = None
 
         self.mj_model = spec.compile()
+        # Keep the compiled qM layout, but restore the physical COM and derived constants.
+        for body_id, body, body_ipos in full_inertia_bodies:
+            body.ipos = body_ipos
+            self.mj_model.body_ipos[body_id] = body_ipos
+            self.mj_model.body_sameframe[body_id] = mujoco.mjtSameFrame.mjSAMEFRAME_NONE
```
vreutskyy, line 6349: [P2] Please keep compiler-derived actuator constants consistent after restoring the physical COM. On the current head, `TestActuatorDampratio.test_mjw_model_matches_compiled` fails on every completed platform: `mjw_model.actuator_biasprm[2]` is about `-0.577350`, while the restored CPU model contains `-0.577523`. That means the two execution paths resolve `dampratio` from different effective inertial constants. Recompute or synchronize the same physical constants through `put_model` / later `set_const_0` updates rather than leaving the models divergent.

newton/_src/utils/import_usd.py
```diff
@@ -2874,6 +2879,21 @@ def _build_mass_info_from_shape_geometry(
                 )
                 if gap_val == float("-inf"):
                     gap_val = builder.default_shape_cfg.gap
+                if legacy_margin_gap:
```
vreutskyy: [P2] Please preserve schema-resolver precedence in legacy mode. `margin_val` was already selected by `SchemaResolverManager` according to the configured resolver order, but this block reads `mjc:margin` directly and overwrites it whenever the attribute exists—even when `SchemaResolverNewton` has higher priority or `SchemaResolverMjc` was not configured. Apply the legacy subtraction inside the MuJoCo resolver (or only when that resolver actually won), and add coverage for both mixed-resolver orderings.

vreutskyy (changes requested):
Requesting changes for four issues that should be resolved before merge: contact-capacity overflow after gap propagation, the unconditional general-qM performance regression, inconsistent actuator dampratio constants between MuJoCo and MJWarp, and USD resolver precedence in legacy mode. The current head has five failing checks; details and suggested directions are inline.

newton/_src/solvers/mujoco/solver_mujoco.py
adenzler-nvidia: Thanks for flagging the benchmark gate. I agree the regressions need to be resolved, but isolation does not support attributing them to general `qM` storage.

G1 remains fully general when the temporary COM offset is removed (`qLD_simple_dofs=0`, `nC=589`), so its reported overhead percentage cannot result from losing simple storage. Its absolute CI runtime improves from 324 ms to 260 ms across the dependency upgrade.

Allegro recovers six simple DOFs when the offset is removed, but this produced no measurable speedup. The actual regression came from the benchmark's gap configuration: only `default_shape_cfg.gap` was zeroed, leaving imported visual shapes and the separately-added ground at the 0.1 m `rigid_gap`. Commit f625271c7 sets the incoming rigid-gap default to zero and propagates it to the outer builder; in the same 32-world diagnostic this reduced runtime from approximately 1.35 s to 0.64 s.

The general layout is required for Newton's existing runtime inertia-update contract, not speculative optional setup. `body_mass`, `body_com`, and `body_inertia` are writable arrays, and `notify_model_changed(BODY_INERTIAL_PROPERTIES)` supports arbitrary changes. A compiled simple layout omits entries that can become nonzero after COM or inertial-frame changes and cannot expand at runtime. We could separately add an opt-in `assume_static_inertia` mode that preserves simple storage and rejects incompatible notifications, but the cited benchmarks do not establish a need for it.

newton/_src/utils/import_usd.py
adenzler-nvidia: Fixed in `221265597`. `SchemaResolverManager` now returns the resolver that supplied the selected authored value, and legacy margin-gap subtraction runs only when `SchemaResolverMjc` actually won. The existing legacy conversion test now covers both mixed resolver orderings.

newton/_src/solvers/mujoco/solver_mujoco.py
adenzler-nvidia: Fixed in `186203aa0`. The first compile consumed the positive `dampratio` placeholder; after restoring the physical COM, `mj_setConst` saw a negative explicit damping coefficient and did not recompute it. We now preserve the placeholder from `MjSpec` and restore it before the post-restoration `mj_setConst`, so both CPU and MuJoCo Warp derive damping from the physical inertia. The regression test now compares both representations against a native MuJoCo compile.

newton/_src/solvers/mujoco/kernels.py
adenzler-nvidia: Fixed in `ba6435cf4`. `basic_plotting` does not use detection-only contacts, so it now sets `rigid_gap = 0.0` on the incoming humanoid builder and propagates that default to the scene builder before adding the ground plane. This keeps its dynamic contact set within the automatically sized capacity instead of provisioning for an unused sensing shell. The registered CPU and CUDA example variants both complete without overflow. I also scanned the complete ARM, Ubuntu, macOS, Windows, GPU, and benchmark logs; every actual broadphase or `nefc` overflow was from this example, with no other contact or narrowphase overflow.

newton/_src/solvers/mujoco/solver_mujoco.py
adenzler-nvidia: Correction after rerunning the actual ASV comparison: the Allegro regression does reproduce on an NVIDIA RTX PRO 6000 Blackwell Workstation Edition. Running `asv continuous` with the same baseline and PR revisions, interleaved rounds, and benchmark configuration as CI measured 774 ms before and 1.07 s after, a 1.38x slowdown. My earlier standalone diagnostic was not representative of the ASV benchmark despite using the same world and frame counts.

The benchmark does not compare different Warp builds: `asv.conf.json` pins `warp-lang==1.15.0.dev20260626`, ASV uses one virtual environment for both revisions, and both Newton wheels accept that same version. The local ASV environment confirms that exact Warp build. Historical CI places the first regression exactly at the MuJoCo Warp 3.10 dependency bump, before the full-qM workaround. This is therefore reproducible across the CI `g7e.2xlarge` and local workstation Blackwell systems, and is not an L40S-versus-Blackwell split or a baseline-versus-PR Warp version mismatch.

vreutskyy (dismissed):
All requested code changes are addressed and all required checks pass. The remaining optional Allegro benchmark regression has been isolated to the MuJoCo Warp 3.10 dependency change rather than the reviewed `qM` implementation and can be tracked independently.

adenzler-nvidia:
I tracked the Allegro regression to solver iterations rather than slower kernels. Nsight showed that the 3.10 path invokes the line-search and gradient-update kernels roughly three to four times as often, while their individual timings are generally unchanged or faster.

Bisecting MuJoCo Warp identified google-deepmind/mujoco_warp@47c63618 (google-deepmind/mujoco_warp#1374, “Fix line search cost precision”) as the first commit exposing the behavior. Before that change, subtracting large float32 absolute costs could round a small improvement to zero and terminate early. The corrected calculation reveals that this Allegro setup is poorly damped and can reach the 100-iteration cap.

The benchmark applies one fixed target (`0.8365207`) to every actuated finger joint with `kp=150` and `kd=5`. I tested tolerance changes, the Menagerie-derived asset, target scaling, limit clamping, and drive gains. Loosening tolerance hides the issue, changing the asset changes the physical workload, and clamping targets made the contact workload worse. Increasing only the Allegro derivative gain to `kd=20` preserves the asset, target, stiffness, contacts, and solver tolerances while stabilizing convergence.

In the exact interleaved ASV comparison, `kd=20` changes FastAllegro from 717±9 ms on the 3.8.1 baseline to 686±1 ms on this PR (0.96×, not significant). The original `kd=5` setup measured approximately 774 ms versus 1.07 s. The damping adjustment is in bab46ac54.

vreutskyy (approved):
Re-reviewed the post-approval changes at this head. No blocking findings, and all required checks pass.

jcarius-nv:
Approving, but worth to double-check why that MJ benchmark got 25% slower:

| +        |              17.8356 |             22.4081 |    1.26 | simulation.bench_mujoco.FastNewtonOverheadG1.track_simulate(256) |

jcarius-nv (approved):

adenzler-nvidia:
The overhead benchmark is showing us how much time we spend in newton compared to mujoco, so it's always correlated with a speedup mujoco warp. The mujoco side is getting much faster with this update, so the overhead is higher.

We definitely need to address the fact that we now have more than 20% time spend in newton.

PR #2981, opened 2026-05-28
Fix ball joint conversion for `SolverMuJoCo`

camevor:
## Description

Fix `SolverMuJoCo` ball-joint conversion under non-identity `child_xform` rotation. MuJoCo's ball `qvel`/`qfrc` live in the post-`qpos` child body frame, not the rest-body frame, so the four bridge kernels (`convert_{mj_coords_to_warp,warp_coords_to_mj}_kernel`, `apply_mjc_qfrc_kernel`, `convert_qfrc_actuator_from_mj_kernel`) now compose `joint_X_c.q` with the current ball `joint_q` (`r = X_cj.q^{-1} * qpos * X_cj.q` and duals). Newton's `body_q`/`body_qd` and applied/actuator torques now round-trip through MuJoCo for arbitrary ball poses.

Folded in while touching these kernels:

- `apply_mjc_qfrc_kernel` was reading `joint_type` / `joint_dof_dim` at the per-world template index instead of the global joint index — silent on homogeneous worlds, wrong on heterogeneous multi-world models.
- Added the KINEMATIC-child early-return to `convert_qfrc_actuator_from_mj_kernel` BALL to match `apply_mjc_qfrc_kernel`.

## Checklist

## Test plan

```
```

New tests in `TestMuJoCoArticulationConversion` covering FK, qvel, applied torque (identity and non-identity `joint_q`), and the actuator readback branch. Each fails without the corresponding fix. Full suite: 202 pass / 7 skipped.

## Bug fix

**Steps to reproduce:**

1. Build a `SolverMuJoCo` model with a ball joint whose `child_xform` has a non-identity rotation.
2. Set non-identity `joint_qd` (and optionally `joint_q`), step, or call `eval_fk`.
3. Compare Newton's `body_qd` against `mujoco.mj_objectVelocity` — they diverge for any rotation that does not commute with `child_xform.q`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed ball-joint conversions so position, velocity, and applied/actuator torques are correctly transformed when a joint’s child anchor has a non‑identity rotation; actuator targets are now rotated into the solver’s expected child-frame.

* **Tests**
  * Added tests validating ball-joint kinematics, applied-torque behavior, actuator readback, and controller targets for rotated child anchors.


adenzler-nvidia:
**Comment verbosity**

Each BALL branch in `kernels.py` independently re-derives the same two facts — `r = X_cj⁻¹ · q_mj · X_cj` and "MuJoCo qvel/qfrc live in the post-qpos child body frame" — that the new `ball_rotate_*` helpers already encode in their names. Suggest hosting the derivation once in the helper docstrings and trimming each call site to a one-liner that names the frame.

Spots that read as restatement rather than load-bearing context:

- `convert_mj_coords_to_warp_kernel` BALL qpos (the "Other joint types (D6/REVOLUTE/PRISMATIC) absorb `X_cj.rot`..." paragraph): this is about `_convert_to_mjc` axis pre-rotation, not the BALL branch itself. It belongs over there, with one line here: `# BALL has no axis hook to absorb X_cj.rot, so we conjugate explicitly.`
- Both qvel comment blocks (mj→warp and warp→mj): the `mju_quatIntegrate` citation is load-bearing (justifies "current child body frame"). The "simplifies to..." sentence is restatement — the named `q_inv_cj_q_mj` local and the `ball_rotate_newton_to_mj` call already carry it.
- `apply_mjc_qfrc_kernel` BALL and `convert_qfrc_actuator_from_mj_kernel` BALL: keep "forces are dual to velocities" (load-bearing), drop the frame preamble.

The test docstrings have the same pattern. The non-`_with_joint_q` tests re-narrate the full bridge math, but their actual purpose is regression coverage for the common identity-`joint_q` path — that's worth saying in one sentence. The `_with_joint_q` variants are the ones that earn their length because they carry the "this test exists because the simpler variant cannot detect a missing `r⁻¹`" rationale.

**Ambiguous notation**

Two of the comments use `X_cj.rot⁻¹ * q_mj` where the code is `wp.quat_rotate(q_inv_cj_q_mj, vec)`, i.e. a rotation acting on a vector — not a quaternion product. Reads as ambiguous since the surrounding text mixes quaternion products and rotated-vector results.

- `convert_mj_coords_to_warp_kernel` BALL qvel: "Equating world omegas gives `w = X_cj.rot⁻¹ * qpos * qvel`, which simplifies to `X_cj.rot⁻¹ * q_mj` since..." → suggest `w = R(X_cj.rot⁻¹ · q_mj) · omega_mj`.
- `convert_qfrc_actuator_from_mj_kernel` BALL: "The Newton-side rotation is `r * X_cj.rot⁻¹`, which simplifies to `X_cj.rot⁻¹ * q_mj`..." → suggest `tau = R(X_cj.rot⁻¹ · q_mj) · tau_mj` (dual of `apply_mjc_qfrc_kernel` BALL).

The other two blocks already use the `R(·) · v` convention.

newton/_src/solvers/mujoco/kernels.py
```diff
@@ -622,22 +623,26 @@ def convert_mj_coords_to_warp_kernel(
...
-        # change quaternion order from wxyz to xyzw
-        rot = quat_wxyz_to_xyzw(
-            wp.quat(
-                qpos[worldid, q_i],
-                qpos[worldid, q_i + 1],
-                qpos[worldid, q_i + 2],
-                qpos[worldid, q_i + 3],
-            )
+        # Newton uses the parent anchor frame for both qpos and qvel.
+        # MuJoCo splits them: qpos in the child rest frame, qvel/qfrc in the current (post-qpos) body frame.
```
eric-heiden, line 669: Can you double-check the conventions we have for Newton (I'm not sure if we have ball joint targets implemented/documented anywhere so far):

This frame conversion is not applied to target controls. `apply_mjc_control_kernel` still copies `JOINT_TARGET` values straight into `mj_ctrl`, but the ball actuators in `solver_mujoco.py` use MuJoCo joint transmissions with per-axis gear. For ball joints, MuJoCo compares velocity targets against `qvel` in the current child-body frame, while Newton targets are in the parent-anchor frame. Position targets have the same mismatch because the actuator length comes from MuJoCo's `axis_angle(qpos)` basis.

Please make the control path ball-aware too: pass the joint type/start, `joint_X_c`, and current `joint_q` into `apply_mjc_control_kernel`, rotate velocity targets with the same mapping used for qvel writeback, and either convert position targets into MuJoCo's ball actuator basis or explicitly reject/document rotated-child ball target controls.

camevor: Good spot, thanks! This should be addressed now. Rebased after #2965 landed and applied the transformation to the ball joint target positions and velocities.

newton/_src/solvers/mujoco/kernels.py
```diff
@@ -1413,23 +1430,59 @@ def apply_mjc_control_kernel(
...
+                        mj_ctrl[world, actuator] = aa_mj[axis_idx]
+                else:
+                    if world_target_q + 2 < joint_target_q.shape[0]:
+                        aa_newton = wp.vec3(
+                            joint_target_q[world_target_q + 0],
+                            joint_target_q[world_target_q + 1],
+                            joint_target_q[world_target_q + 2],
+                        )
+                        aa_mj = wp.quat_rotate(mjc_actuator_q_cj[actuator], aa_newton)
+                        mj_ctrl[world, actuator] = aa_mj[axis_idx]
```
adenzler-nvidia: The DOF-layout branch rotates the 3 stored target scalars as if they were an axis-angle rotation vector, but under DOF layout those are **extrinsic ZYX Euler angles**, not a rotation vector.

`builder.py` stores them raw (`self.joint_target_q[quat_offset + i] = dim.target_pos`, with slot 3 set to `1.0`), and `_quat_from_axis_targets` documents the convention as extrinsic ZYX / `wp.quat_from_euler(vec3(t_x, t_y, t_z), 2, 1, 0)`. Euler angles don't transform as a vector under rotation, so `R(q_cj) * (euler triple)` is not a meaningful operation for non-identity `q_cj`.

The coord-layout branch above is correct because it log-maps the target quaternion first (`_target_quat_to_axis_angle`), yielding a genuine rotation vector that does rotate by `R(q_cj)` under the conjugation `q_cj * r * q_cj^{-1}`.

Since `newton.use_coord_layout_targets` defaults to `False`, this DOF path is the default, so ball-joint position targets under a non-identity `child_xform` rotation are still wrong after this PR — it replaces the old passthrough with a different incorrect value. The identity-`child_xform` case is unaffected (`quat_rotate(identity, v) == v`), which is why existing tests pass.

Suggested fix: in the DOF branch, compose the Euler triple into a quaternion (same `quat_from_euler(..., 2, 1, 0)` convention as the builder), apply the `q_cj` conjugation on the quaternion, then log-map — rather than rotating the raw 3-vector. A DOF-layout variant of `test_ball_controller_coord_layout_rotated_anchor` (non-identity `child_xform`) would cover this; the existing rotated-anchor test only runs under coord layout.

The comment just above ("DOF layout stores the 3 axis-angle components directly") is also inaccurate — they are Euler angles.

camevor: Thanks for spotting this!

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -5337,16 +5356,12 @@ def add_geoms(newton_body_id: int):
...
-                            template_target_q_axis = i
-                        else:
-                            template_target_q = tq_start + i
-                            template_target_q_axis = -1
+                        # Base + axis encoding under both layouts. The kernel applies the q_cj
+                        # basis change either way; the layout flag tells it whether the slot
+                        # holds a 4-float quat (coord) or 3 axis-angle scalars (DOF).
+                        template_target_q = tq_start
+                        template_target_q_axis = i
+                        q_cj_entry = (joint_rot[0], joint_rot[1], joint_rot[2], joint_rot[3])
```
eric-heiden: 🔴 **Ball target basis is frozen to world-0 template at construction time** (high)

The new ball-target-control conversion snapshots `q_cj_entry = (joint_rot[0], ..., joint_rot[3])` from `child_xform.q` at construction time (line 5364) and stores it in a 1-D per-actuator array `self.mjc_actuator_q_cj`. The `apply_mjc_control_kernel` then uses `mjc_actuator_q_cj[actuator]` for every world when rotating position and velocity targets (kernels.py:1464 and 1494-1495).

Issue in **`notify_model_changed(JOINT_PROPERTIES)`**: `_update_joint_properties()` at solver_mujoco.py:6757-6808 updates the MuJoCo model's joint positions and axes from the current `model.joint_X_c`, but `self.mjc_actuator_q_cj` is never rebuilt. If a user updates `model.joint_X_c.q` at runtime and notifies the solver, the coordinate bridges will pick up the new rotation but the control kernel will still use the stale snapshot.

Recommendation: add a `self.mjc_actuator_q_cj` rebuild inside `_update_joint_properties()` or add a note/TODO that `q_cj` is assumed immutable for the solver's lifetime.

camevor: Thanks! Fixed in a48cc6c

newton/_src/solvers/mujoco/kernels.py
```diff
@@ -1420,23 +1440,72 @@ def apply_mjc_control_kernel(
...
-                # of the axis-angle 3-vector (matches mjc actuator_length units).
-                if world_target_q + 3 < joint_target_q.shape[0]:
-                    aa = _target_quat_to_axis_angle(
+                # Ball-joint position target
+                # Coord layout stores a 4-float quat (needs log-map); DOF layout stores
+                # extrinsic ZYX Euler target angles directly at the qd_start base.
+                last_elem = world_target_q + wp.where(use_coord_layout_targets, 3, 2)  # check size
+                assert last_elem < joint_target_q.shape[0]
+                if not last_elem < joint_target_q.shape[0]:
+                    return
```
adenzler-nvidia, line 1447: This `assert` is a no-op in normal runs: Warp compiles release builds with `-DNDEBUG` and the suite doesn't enable debug mode, so only the `if not last_elem < ...: return` below actually guards the read. Suggest keeping just the `if` (the bound is correct for both layouts) and dropping the `assert` plus the `# check size` trailing comment, so it doesn't read as an enforced check that never fires.

newton/_src/solvers/mujoco/kernels.py
```diff
@@ -1381,27 +1393,35 @@ def apply_mjc_control_kernel(
...
 
-    For ball-joint position actuators under the coord layout, the target stored
-    in ``joint_target_q`` is a quaternion. ``mjc_actuator_to_target_q_axis_idx``
-    selects axis 0/1/2 of the axis-angle representation; a value of -1 means
-    "scalar passthrough" (the legacy DOF-layout and all non-ball actuators).
+    For ball-joint actuators, ``axis_idx >= 0`` selects the angular component to feed MuJoCo.
+    Position targets are rotated by the per-world child anchor ``q_cj`` (``joint_X_c`` indexed by
+    ``mjc_actuator_to_newton_ball_jnt`` and the current world). Velocity targets read the current
+    quaternion start from ``mjc_actuator_to_newton_target_q_idx`` and rotate by ``q_cj * r^{-1}``
+    (mirroring the qpos / qvel bridges in :func:`convert_mj_coords_to_warp_kernel` BALL). The
```
adenzler-nvidia: This cross-reference points the wrong direction. This kernel converts Newton→MuJoCo (it writes `mj_ctrl`), and the `q_cj * r^{-1}` operator used here is the writeback defined in `convert_warp_coords_to_mj_kernel` (and `apply_mjc_qfrc_kernel`). `convert_mj_coords_to_warp_kernel` is the inverse direction and uses `q_cj^{-1} * q_mj`. Suggest referencing `convert_warp_coords_to_mj_kernel` instead.

newton/_src/solvers/mujoco/kernels.py
```diff
@@ -1420,23 +1440,72 @@ def apply_mjc_control_kernel(
                 if world_target_q < joint_target_q.shape[0]:
                     mj_ctrl[world, actuator] = joint_target_q[world_target_q]
             else:
-                # Ball-joint quaternion target: feed MuJoCo the matching component
-                # of the axis-angle 3-vector (matches mjc actuator_length units).
-                if world_target_q + 3 < joint_target_q.shape[0]:
-                    aa = _target_quat_to_axis_angle(
+                # Ball-joint position target
+                # Coord layout stores a 4-float quat (needs log-map); DOF layout stores
+                # extrinsic ZYX Euler target angles directly at the qd_start base.
```
adenzler-nvidia: Minor wording: the DOF-layout target isn't read at a `qd_start` base — it's indexed via `world_target_q`, i.e. the joint's target-q base (`joint_target_q_start`). Calling it the `qd_start` base implies a velocity/DOF index. Suggest "at the joint's target-q base".

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -4800,6 +4822,7 @@ def joint_has_raw_limit_solref(dof_idx: int) -> bool:
         # holds a quaternion; this list selects which axis-angle component (0/1/2)
         # to feed MuJoCo. -1 means "scalar passthrough" (all other actuators).
```
adenzler-nvidia: This comment block is now stale. After this PR the axis selector is set for ball-joint position *and* velocity actuators (the velocity branch now appends `i`), and under both coord and DOF layouts (the `if use_coord_layout_targets` gate was removed). It's no longer specific to "position actuators under the coord layout". Worth updating to match the `mjc_actuator_to_target_q_axis_idx` attribute docstring, which already says "position and velocity actuators".

adenzler-nvidia (commented):
A few documentation and cleanup notes on the ball-joint conversion path — all non-blocking. The frame-conversion math and the actuator wiring look correct to me; these are just about a redundant debug `assert`, a stale comment, and two docstring/comment wording fixes. Details inline.

newton/_src/solvers/mujoco/kernels.py
camevor: Fixed!

newton/_src/solvers/mujoco/kernels.py
camevor: Fixed

newton/_src/solvers/mujoco/kernels.py
camevor: This is intentional: If the guard is needed, we should assert to catch invariant invalidation in debug mode. Otherwise, we should just remove the guard.

newton/_src/solvers/mujoco/solver_mujoco.py
camevor: Fixed

camevor:
Thanks for the review, @adenzler-nvidia! Fixed the remaining items (except for the assert). Reasoning:
The assert documents and trips an invariant, not a bounds check that is expected trigger occasionally. The return still protects the following indexed reads in environments/modes where the assert is not fatal.

adenzler-nvidia (approved):

PR #2990, opened 2026-05-28
Add opt-in sphere rendering for points in ViewerUSD

aneangel:
## Description

Split from #2950 per reviewer feedback. Adds an opt-in `points_as_spheres` constructor flag on `ViewerUSD`. When `True`, `log_points` writes each point as an instance of a `UsdGeom.Sphere` prototype under a `UsdGeom.PointInstancer`, producing individually-visible 3D spheres scaled by `radii`. When `False` (default) the existing flat `UsdGeom.Points` splat behavior is preserved — no behavior change for callers that do not opt in.

Keeping the toggle on the constructor (rather than as a parameter on the abstract `Viewer.log_points`) avoids forcing every other backend to accept and ignore a USD-specific rendering preference. Considered always-spheres + deprecation cycle (option A from the review thread); chose the opt-in flag (option B) to ship the capability now without a behavior change for existing users.

Also normalizes the interpolation tokens returned by `_normalize_point_colors` to `UsdGeom.Tokens.constant` / `UsdGeom.Tokens.vertex` instead of string literals, matching how the new sphere path creates its `displayColor` primvar.

## Checklist

## Test plan

```
uv run --extra dev -m unittest newton.tests.test_viewer_usd newton.tests.test_viewer_particle_flags -v
```

Specifically the new regression: `test_log_points_renders_as_point_instancer_when_points_as_spheres_set` — verifies that constructing `ViewerUSD(..., points_as_spheres=True)` makes `log_points` produce a `UsdGeom.PointInstancer` with a sphere prototype (and *not* a flat `UsdGeom.Points` prim) and that scalar `radii` broadcast correctly to the `(N, 3)` `scales` array.

## New feature / API change

```python
import newton

# Default: particles render as flat UsdGeom.Points splats (unchanged behavior)
viewer = newton.viewer.ViewerUSD("out.usda")

# Opt in: particles render as a UsdGeom.PointInstancer of sphere prototypes,
# producing individually-visible 3D spheres scaled by the per-point radii.
viewer = newton.viewer.ViewerUSD("out.usda", points_as_spheres=True)
```

## Summary by CodeRabbit

* **New Features**
  * Added `points_as_spheres` option to `ViewerUSD` (enabled by default) to render logged points as instanced spheres, with scalar/per-point radius control; disable for flat point splats.
* **Bug Fixes**
  * Improved point color interpolation to follow the selected per-point color style.
* **Tests**
  * Added tests covering instanced sphere rendering and scalar-to-per-point radius scaling.
* **Documentation**
  * Updated the changelog to describe the new behavior.


jcarius-nv:

newton/_src/viewer/viewer_usd.py
```diff
@@ -81,6 +81,7 @@ def __init__(
         up_axis: str = "Z",
         num_frames: int | None = 100,
         scaling: float = 1.0,
+        points_as_spheres: bool = False,
```
eric-heiden: I think we could make the default `True` here? To me the behavior described in https://github.com/newton-physics/newton/issues/917 looks wrong so we better change the default.

aneangel: I went ahead and changed the default so that it is now True

CHANGELOG.md
```diff
@@ -4,6 +4,21 @@
...
+- Add robotics tutorial notebook covering ModelBuilder, solvers, CUDA graphs, IK, and pick-and-place
+- Add `newton.utils.OnnxRuntime`, a graph-capturable ONNX inference engine backed solely by Warp kernels (no `onnxruntime` or `torch` runtime dependency); used by `ControllerNeuralMLP` and `ControllerNeuralLSTM` to load `.onnx` policies. To migrate a TorchScript policy, run `torch.onnx.export(model, dummy_input, "policy.onnx", opset_version=17)` once and point the controllers at the resulting `.onnx` file. The `onnx` package is now an optional extra (`pip install newton[onnx]`); install it explicitly to use the ONNX runtime.
+- Add USD parsing for `NewtonSiteAPI` to mark shapes as sites.
+- Add `ViewerRTX`, a real-time ray-traced viewer powered by NVIDIA OVRTX.
+- Support negative (mirrored) scale on mesh, convex hull, and SDF shapes, so a single `Mesh` instance can be shared across shapes with different signed scales without re-baking
+- Add `newton.utils.ColorSpace`, `color_srgb_to_linear()`, `color_linear_to_srgb()`, and `SensorTiledCamera.RenderConfig.output_color_space` for color-space boundaries
```
jcarius-nv: changelog entries look a bit off here, is this from a bad merge?

aneangel: mm, I am unsure. It could be possible since the last merge from main that I did was quite big since I had not updated the branch with the upstream changes for a while since I was on leave

eric-heiden: @aneangel can you please fix this changelog and merge the latest from main? Then we can merge the PR.

aneangel: @eric-heiden I went and head resolved this issue

adenzler-nvidia:
Updated the branch with current `upstream/main` and addressed the remaining items: `log_points(..., points=None)` now hides `PointInstancer`-backed sphere clouds, the docs reflect that sphere rendering is the default, the sphere prototype explicitly authors radius 1.0, and the ViewerUSD regression test covers default instancing and hiding. The ViewerUSD test suite and pre-commit checks pass.

jcarius-nv (approved):

PR #1765, opened 2026-02-24
Franka cloth demo improvement

AnkaChan:
Fixes #1624

- Switch the simulation from meter scale to centimeter scale for better numerical stability of the VBD solver, which resolves the degraded cloth-robot interaction behavior. This is really necessary for reaching high quality in this demo. The old visualization, which looks better is simulated in centimeter scale.
- Add a `viz_state` that converts positions back to meter scale for rendering, with GPU kernels (`scale_positions`, `scale_body_transforms`) to avoid per-frame allocations
- Render the table box manually via `log_shapes` at meter scale, working around the GL viewer baking primitive dimensions into geometry meshes
- Tune cloth/contact stiffness parameters for cm-scale stability
- Clamp to the last key pose after the sequence ends instead of looping
image

## Release Notes

* **Refactor**
  * Enhanced cloth simulation examples with visualization scaling capabilities to support centimeter-to-meter unit conversions without affecting core physics simulation
  * Simplified rolling cloth example constructor by removing externally configurable parameters and using predefined default values
  * Improved ground plane configuration in multiphysics examples with explicit control over contact stiffness, damping, and friction properties

PR #2991, opened 2026-05-28
Fix cloth_franka example to render the particles in proper scale

aneangel:
## Description

Split from #2950 per reviewer feedback. The `cloth_franka` example simulates in cm but renders in meters, with `shape_transform` and `shape_scale` already swapped to viewer-scale copies before each render. Particle radii were left at cm scale, making particles appear ~100x larger than intended when `show_particles` is enabled.

Mirrors the existing `shape_transform` / `shape_scale` pattern for `particle_radius`: builds a viewer-scaled copy at finalize time, swaps it into the model before rendering, restores the simulation value after.

Also wraps the swap / render / restore block in `try/finally` so a viewer-scale `particle_radius` cannot leak permanently into the model if `begin_frame`, `log_state`, or `end_frame` raises.

## Checklist

## Test plan

This is a visualization fix in an example — no meaningful automated assertion is possible for the rendered output. Verified manually by running the example and observing that particles render at the same scale as the cloth geometry:

```
uv run -m newton.examples cloth_franka
```

The existing example test suite continues to pass:

```
uv run --extra dev -m newton.tests -k cloth_franka
```

## Bug fix

**Steps to reproduce:**

1. Check out `main` (without this PR applied)
2. Run `uv run -m newton.examples cloth_franka`
3. Observe that the cloth particles render with radii ~100x larger than expected — they appear as oversized spheres dwarfing the cloth and Franka arm

The root cause is in `example_cloth_franka.py`: the example computes `viz_shape_transform` and `viz_shape_scale` as meter-scaled copies of the simulation cm values and swaps them in for rendering, but no equivalent treatment was applied to `particle_radius`.

**Minimal reproduction:**

```python
# The cloth_franka example is the minimal repro for this rendering bug:
#   uv run -m newton.examples cloth_franka
#
# Without this PR: particles render ~100x larger than the cloth geometry.
# With this PR: particles render at the same scale as the cloth.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected cloth/robot visualization particle sizing to use simulation scale (centimeters) rather than viewer scale (meters).
  * Ensured rendering scale changes (including particle radius adjustments) are reverted after each frame so simulation-scale settings remain consistent.


jcarius-nv:

adenzler-nvidia:
Stepping back from this specific fix: the whole viz-swap chain here (`viz_state`, `viz_shape_transform`/`viz_shape_scale`, and now `viz_particle_radius`) exists only because #1765 moved the simulation to cm for VBD stability but kept rendering in meters. Every Model/State field the viewer reads needs a mirrored meter-scale copy, and each new field is a chance to miss one — which is exactly the bug this PR fixes.

I think the better direction is to make the viewer render a cm-scale scene just as well natively and drop the swap ceremony entirely. The render path itself is unit-agnostic and the example already sets its own camera; the remaining meter-scale assumptions are small and fixable: camera fly speed (`_camera_speed`), `pick_stiffness`, and `ViewerUSD` hardcoding `metersPerUnit = 1.0`. Making those scale-aware would benefit any non-meter scene (related: #917) and would let us delete this pattern instead of extending it.

@AnkaChan you introduced the cm-scale setup in #1765 — what do you think?

adenzler-nvidia:
Looks good as a workaround - I merged latest main and resolved the conflicts.

eric-heiden (approved):
Thanks!

PR #2992, opened 2026-05-28
Fix ViewerUSD leaving stale particle geom

aneangel:
## Description

Split from #2950 per reviewer feedback. Fixes `ViewerUSD` leaving stale particle geometry visible after the active particle count drops to zero.

On `main`, `_log_particles` signaled the "no active particles" case by calling `log_points(name, points=None, hidden=True)`. `ViewerUSD.log_points` treated `points=None` as a no-op (early return), so any `UsdGeom.Points` prim authored on prior frames remained visible — stale spheres at last-known positions.

The original implementation in #2950 fixed it by overloading `log_points` to hide the prim on `points=None`, but that approach was rejected in review because it:
- Made `log_points` ignore the `hidden` argument in that branch (asymmetric)
- Required formalizing `-> Sdf.Path | None` as the return type, leaking a USD-specific type through the abstract API

This PR implements the reviewer's redesign: fix at the call site instead. `_log_particles` now allocates an empty `wp.array` and passes it with `hidden=True`. The existing `log_points` already handles empty arrays correctly (positions set to empty, visibility set to `"invisible"`), so no backend needs to change.

## Checklist

## Test plan

```
uv run --extra dev -m unittest newton.tests.test_viewer_usd newton.tests.test_viewer_particle_flags -v
```

Specifically:
- `test_log_points_hides_existing_prim_with_empty_points_and_hidden` (new regression in `test_viewer_usd.py`) — verifies a previously-visible `UsdGeom.Points` prim becomes invisible after `log_points` is called with an empty array and `hidden=True`, asserting via `viewer.stage` directly (no reliance on return-type information).
- `test_all_inactive_clears_particles` (updated assertion in `test_viewer_particle_flags.py`) — confirms `_log_particles` now passes an empty array (not `None`) when all particles are inactive.

## Bug fix

**Steps to reproduce:**

1. Check out `main` (without this PR applied)
2. Create a `ViewerUSD` scene with a model containing particles
3. Render a few frames with all particles flagged `ParticleFlags.ACTIVE`; the particle `Points` prim appears in the USD output
4. Clear the `ACTIVE` flag on every particle so the active count drops to zero
5. Render more frames
6. Open the resulting `.usda` in `usdview`: the particle `Points` prim from the earlier frames stays visible at last-known positions instead of disappearing

**Minimal reproduction:**

```python
# Easiest way to reproduce the on-main bug end-to-end is to run the
# regression test on a pre-fix tree:
#
#   uv run --extra dev -m unittest \
#     newton.tests.test_viewer_particle_flags.TestViewerParticleFlags.test_all_inactive_clears_particles
#
# On main (without the viewer.py change in this PR), the assertion
#   self.assertEqual(len(viewer.logged_points), 0)
# fails because _log_particles passes None instead of an empty array.

import warp as wp

import newton
from newton.viewer import ViewerUSD

viewer = ViewerUSD("out.usda", num_frames=2)
active = int(newton.ParticleFlags.ACTIVE)
for i in range(3):
    builder.add_particle(pos=(float(i), 0.0, 0.0), vel=(0.0, 0.0, 0.0), mass=1.0, radius=0.1, flags=active)
state = model.state()
viewer.set_model(model)

# Frame 0: particles render normally.
viewer.begin_frame(0.0)
viewer._log_particles(state)
viewer.end_frame()

# Frame 1: clear all ACTIVE flags so active_count drops to 0.
model.particle_flags.fill_(0)
viewer.begin_frame(1.0 / 60.0)
viewer._log_particles(state)   # main: silent no-op, prim stays visible
                               # this PR: prim hidden as expected
viewer.end_frame()
viewer.close()
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed particle rendering so stale particle geometry is cleared when the active particle count drops to zero.
  * Fixed USD point handling so logging with empty point data and `hidden=True` correctly hides existing `UsdGeom.Points` prims on subsequent frames.

* **Tests**
  * Updated particle-flag tests to expect an explicit empty points payload (rather than no data) when no particles are active.
  * Added a USD test covering per-frame visibility transitions for empty hidden point logs.


adenzler-nvidia:
Before this lands I'd like to pin down the intended contract for `log_points`, because the root cause here is that `points` and `hidden` mean different things in different backends — this PR fixes the symptom for the particle path but the underlying ambiguity remains.

Today `points=None` is interpreted three different ways:

| Backend | `points=None` |
|---|---|
| `ViewerGL` | hides the existing prim |
| `ViewerUSD` | no-op (this is the stale-geometry bug) |
| `ViewerRerun` / `ViewerViser` | no-op |
| `ViewerFile` | no-op (doesn't render points) |

So `None` is simultaneously trying to mean "clear/hide this" (GL) and "no update" (everyone else), and the two diverged silently. That's what produced the bug.

I think there are really two orthogonal facts a call should state: **what the point set is** (the `points` data, where empty = zero points) and **whether to draw it** (`hidden`). The COM path already uses exactly this model — it logs real positions every frame and just toggles `hidden=not self.show_com` (`viewer.py:2268`). The particle-clearing path was the only place reaching for `None`, which is what this PR corrects.

Proposed contract, to make explicit:
- `points` = the current data; an empty array means "zero points."
- `hidden` = visibility only; doesn't change the data and is cheaply reversible.
- `None` = strict no-op ("no update") in *all* backends — which means removing `ViewerGL`'s `None`-as-hide branch so it stops being the outlier. (Longer term `points` could drop `| None`, but that's a public-API breaking change needing a deprecation cycle.)

This PR is consistent with this direction, so I'm happy to land it. The follow-up would be (1) documenting this contract on the abstract `Viewer.log_points` and (2) the small `ViewerGL` cleanup.

@eric-heiden @mmacklin — does this match your mental model of `hidden` vs `points=None`, or is there a use case for `None`-as-hide I'm missing? Want to lock the semantics down before more callers depend on either reading.

jcarius-nv:

adenzler-nvidia (dismissed):

adenzler-nvidia (dismissed):

jcarius-nv (dismissed):

jcarius-nv (approved):

PR #2994, opened 2026-05-29
Add per-shape contact margin to VBD particle-rigid contacts

AnkaChan:
## Summary
- Thread `model.shape_margin` through VBD soft contact detection and body-particle contact force evaluation, so each shape's margin independently inflates the contact shell
- Previously only the global `soft_contact_margin` and `particle_radius` were used, causing thin mesh shapes (e.g. bear mesh legs) to slip through cloth
- Add three VBD contact examples: rigid-rigid, soft-rigid, and mixed (cloth bag containing rigid bodies)

## Test plan
- [x] Existing VBD tests pass (`uv run --extra dev -m newton.tests -k test_vbd`)
- [x] Smoke test: sphere with `margin=0.01` on cloth runs 10 frames without crash, `shape_margin` array correctly populated
- [ ] Run `example_vbd_soft_rigid_contact` — rigid shapes should rest on cloth without penetrating
- [ ] Run `example_vbd_soft_rigid_mix_contact` — rigid shapes should stay inside bag during lift
- [ ] Run `example_vbd_rigid_rigid_contact` — rigid shapes should bounce in box

## Release Notes

* **New Features**
  * Introduced per-shape contact margins for soft-contact detection and particle–rigid interactions, providing independent collision threshold configuration for each shape.

* **Documentation**
  * Added three new VBD simulation examples: rigid–rigid contact demonstration, soft–rigid contact with cloth “bag” enclosures, and a soft–rigid mixed contact scenario.

* **Tests**
  * Updated VBD and collision pipeline test coverage to account for per-shape margins.


WenchaoHuang (commented):
The `example_vbd_soft_rigid_contact.py` and `example_vbd_soft_rigid_mix_contact.py` examples are currently not working correctly. The cloth immediately blows up during the simulation. It’s possible that some changes were not committed.

newton/_src/solvers/vbd/rigid_vbd_kernels.py
```diff
@@ -799,7 +800,8 @@ def _eval_body_particle_contact(
     bx = wp.transform_point(X_wb, contact_body_pos[contact_index])
     n = contact_normal[contact_index]
 
-    penetration_depth = -(wp.dot(n, particle_pos - bx) - particle_radius[particle_index])
+    margin = shape_margin[shape_index] if shape_margin.shape[0] > 0 else 0.0
```
eric-heiden, line 817: 🟡 This adds user-visible support for `ShapeConfig.margin` in VBD contacts, but `CHANGELOG.md` still needs an `[Unreleased]` entry.

Users who track release notes will miss that shapes with nonzero `margin` now affect contact detection and penetration depth. Please add an entry under `### Added` or `### Changed`, depending on whether this is considered new support or a behavior change for existing margins.

newton/_src/geometry/kernels.py
```diff
@@ -1125,7 +1133,7 @@ def create_soft_contacts(
         hfd = heightfield_data[shape_heightfield_index[shape_index]]
         d, n = sample_sdf_grad_heightfield(hfd, heightfield_elevations, x_local)
 
-    if d < margin + radius:
+    if d < margin + s_margin + radius:
```
eric-heiden, line 1128: 🟡 `body_pos` is still stored on the raw shape surface while `s_margin` is applied later during force evaluation.

When a contact is detected, the contact point is stored as:

```py
body_pos = wp.transform_point(X_bs, x_local - n * d)
```

This computes the closest point on the actual surface of the shape (at SDF distance d along normal n). However, with per-shape margin, the effective contact surface should be at `d - s_margin` from the shape center (the margin inflates the shape). The stored `body_pos` represents the raw geometry surface, not the margin-inflated surface.

Downstream consumers like `_eval_body_particle_contact` then compute:

```py
penetration_depth = -(wp.dot(n, particle_pos - bx) - radius - margin)
where bx = wp.transform_point(X_wb, contact_body_pos[...]) — this uses the un-inflated surface point and adds margin analytically. So the two-stage approach (store raw point, add margin at force evaluation) is internally consistent.
```

But is this intentional? It means the contact normal `n` is computed at the raw surface, not the inflated surface. For convex shapes this is fine (normal doesn't change with offset), but for concave mesh shapes with sharp features, the normal at `d=0` vs `d=margin` could differ. Worth a comment documenting the design choice.

newton/examples/vbd/example_vbd_rigid_rigid_contact.py
```diff
@@ -0,0 +1,293 @@
...
+
+###########################################################################
+# Example VBD Rigid Drop
+#
+# Rigid bodies of all supported shape types (sphere, box, capsule,
+# cylinder, cone, mesh bear) are dropped into a static rigid box
+# container and settle under gravity.  This exercises pure rigid-rigid
+# contact handling in the VBD solver at a small (centimeter) scale.
+#
+# Command: python -m newton.examples vbd_rigid_drop
```
eric-heiden: 🟡 These documented example names do not match the names discovered from the filenames.

`get_examples()` strips `example_` and `.py`, so these commands should use `vbd_rigid_rigid_contact`, `vbd_soft_rigid_contact`, and `vbd_soft_rigid_mix_contact`. As written, users following the headers will get an example-not-found error.

newton/examples/vbd/example_vbd_soft_rigid_contact.py
```diff
@@ -0,0 +1,387 @@
...
+                        get_or_add_vertex(x0, y0 + dy, z0),
+                        get_or_add_vertex(x0, y0, z0 + dz),
+                        get_or_add_vertex(x0, y0 + dy, z0 + dz),
+                    )
+
+    return np.array(vertices, dtype=np.float32), faces
+
+
+def _load_bear_mesh(target_size):
+    bear_path = os.path.join(newton.examples.get_asset_directory(), "bear.usd")
```
eric-heiden, line 145: ⚪ `_generate_box_bag()` and `_load_bear_mesh()` are duplicated across the new VBD examples.

This is example code, so it is not urgent, but future fixes to bag mesh generation or bear mesh loading will need to be made in multiple places. A small shared helper such as `newton/examples/vbd/utils.py` would keep these examples from drifting.

eric-heiden:
Thanks for addressing my comments! The remaining blocker that fails the CI is now that `create_soft_contacts()` requires `shape_margin`, but `newton/tests/test_collision_pipeline.py` still launches it with the old argument list. This fails that direct kernel-launch test, so please pass a shape-margin array there before merging.

AnkaChan:
@coderabbitai review

Pushed fixes for the open review comments:
- Added an `[Unreleased]` CHANGELOG entry for the three new VBD contact examples (`vbd_rigid_rigid_contact`, `vbd_soft_rigid_contact`, `vbd_soft_rigid_mix_contact`).
- Added a stability assertion to `example_vbd_soft_rigid_contact.py`'s `test_final` (it previously only printed).

newton/_src/geometry/kernels.py
AnkaChan: This design aligns with the current implementation of rigid-rigid contact scheme where it keeps a support point per shape plus a separate per-shape effective thickness, and reconstructs the inflated-surface gap analytically.

As to the normal on the original surface vs on the inflated surface: I think they are the same. The contact normal is the direction from the closest surface point to the particle (the SDF gradient at the query point). Inflating by the margin is just SDF - margin, which leaves that gradient unchanged and only slides the closest point outward by margin along the same n, so the raw-point + analytic-margin split is geometrically exact for convex and concave shapes.

eric-heiden (approved):

eric-heiden:
@WenchaoHuang could you approve if this PR looks good? We need your approval to merge it. 🙂

WenchaoHuang (approved):

PR #3005, opened 2026-05-29
Parse contact response attributes from NewtonMaterialAPI

andrewkaufman:
## Description

Read contact response attributes (`ke`, `kd`, `kf`, `ka`) from `NewtonMaterialAPI` on bound materials during USD import instead of per-shape custom attributes. The schema resolver resolves these across Newton (`newton:contactStiffness`/`contactDamping`/`contactFrictionStiffness`/`contactAdhesion`), PhysX (`physxMaterial:compliantContactStiffness`/`Damping`), and MuJoCo (`mjc:solref`) at the `MATERIAL` level, with configurable resolver priority.

Legacy `newton:contact_ke`/`kd` custom attributes on shape prims are retained as a lowest-priority fallback when no material value is authored, emitting a `DeprecationWarning` at resolve time. `mjc:solref` on material prims is similarly deprecated. Explicitly authoring `-inf` on a material attr ("use engine default") blocks legacy fallback.

Requires newton-usd-schemas>=0.3.0.

## Test plan

```
uv run --extra dev -m newton.tests -k test_contact_response
uv run --extra dev -m newton.tests -k test_contact_response_attrs
uv run --extra dev -m newton.tests -k test_contact_response_legacy
```

## New feature / API change

```usda
def Material "RubberMaterial" (
    prepend apiSchemas = ["NewtonMaterialAPI", "PhysicsMaterialAPI"]
)
{
    float newton:contactStiffness = 5000.0
    float newton:contactDamping = 200.0
    float newton:contactFrictionStiffness = 800.0
    float newton:contactAdhesion = 0.01
}

def Cube "Collider" (
    prepend apiSchemas = ["PhysicsCollisionAPI"]
)
{
    rel material:binding:physics = </RubberMaterial>
}
```

## Summary by CodeRabbit

* **New Features**
  * Added support for USD parsing of Newton material contact-response attributes (stiffness, damping, friction, adhesion) from NewtonMaterialAPI on bound materials.

* **Deprecated**
  * Per-shape contact custom attributes (newton:contact_ke/newton:contact_kd) are now deprecated. Authors should use newton:contactStiffness/newton:contactDamping on bound NewtonMaterialAPI materials instead.

* **Documentation**
  * Updated collision documentation with Newton USD schema APIs for collision, material, and mesh collision handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/3005?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


newton/_src/utils/import_usd.py
```diff
@@ -2632,22 +2647,22 @@ def _build_mass_info_from_shape_geometry(
...
+
+                for attr_key in ("ke", "kd"):
+                    if getattr(material, attr_key) is not None:
+                        continue
+                    per_shape_val = R.get_value(prim, prim_type=PrimType.SHAPE, key=attr_key, verbose=verbose)
+                    if per_shape_val is not None and math.isfinite(float(per_shape_val)):
+                        if attr_key == "ke":
+                            shape_ke = float(per_shape_val)
+                        else:
+                            shape_kd = float(per_shape_val)
```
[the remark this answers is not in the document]
andrewkaufman: `kf`/`ka` are not in the `PrimType.SHAPE` dict for any resolver, therefor there is no legacy per-shape kf/ka codepath. Users authoring those would not have seen them propogate through to the builder previously, so there is no behavoiur change.

newton/_src/usd/schemas.py
```diff
@@ -95,6 +139,10 @@ class SchemaResolverNewton(SchemaResolver):
         PrimType.MATERIAL: {
             "mu_torsional": SchemaAttribute("newton:torsionalFriction", 0.25),
             "mu_rolling": SchemaAttribute("newton:rollingFriction", 0.0005),
+            "ke": SchemaAttribute("newton:contactStiffness", None),
+            "kd": SchemaAttribute("newton:contactDamping", None),
+            "kf": SchemaAttribute("newton:contactFrictionStiffness", None),
```
andrewkaufman: There is ongoing talk of renaming `newton:contactFrictionStiffness`. I'd prefer to hold off merging this until the name settles, rather than need yet another legacy fallback.

andrewkaufman: It was renamed to `newton:contactFrictionGain` in v0.3.1 of the schemas. Backwards compatibility should not be necessary as no newton runtime was released with v0.3.0 schemas

newton/_src/usd/schemas.py
```diff
@@ -39,6 +39,42 @@ def _physx_gap_from_prim(prim: Usd.Prim) -> float | None:
...
+    """Return a getter that reads a legacy contact custom attr with a deprecation warning."""
+
+    def _getter(prim: Usd.Prim) -> float | None:
+        value = usd.get_attribute(prim, legacy_name)
+        if value is not None:
+            warnings.warn(
+                f"'{legacy_name}' on shape prim is deprecated; "
+                f"author '{material_attr}' on the bound NewtonMaterialAPI material instead.",
+                DeprecationWarning,
+                stacklevel=4,
```
eric-heiden, line 52: The deprecation of `mjc:solref` should also go into the `### Deprecated` section of the changelog.

andrewkaufman: Do you have a suggestion for wording? Its a bit of an odd one -- no solref on materials ever existed, neither in newton nor mjcPhysics schemas, & the converter never produced one. I'm not clear who would have been leveraging this.

solref on shapes is in mjcPhysics schemas, the converter does producet that, & its still supported here.

Milad-Rakhsha-NV: We should have used contact compliance parameters on shapes only for our internal assets, I believe (nothing we put out there officially). Not sure if we should have a deprecation path. I assume someone could have made their asset using this parsing support already but I am not really certain anyone has.

newton/_src/usd/schemas.py
```diff
@@ -358,8 +406,8 @@ class SchemaResolverMjc(SchemaResolver):
             ),
             "gap": SchemaAttribute("mjc:gap", 0.0),
             # Contact stiffness/damping from per-geom solref
-            "ke": SchemaAttribute("mjc:solref", [0.02, 1.0], solref_to_stiffness),
-            "kd": SchemaAttribute("mjc:solref", [0.02, 1.0], solref_to_damping),
+            "ke": SchemaAttribute("mjc:solref", None, solref_to_stiffness),
```
Milad-Rakhsha-NV: what was the reasoning behind removing the schema defaults here?

newton/_src/usd/schemas.py
```diff
@@ -370,8 +418,16 @@ class SchemaResolverMjc(SchemaResolver):
             # Contact models
             "priority": SchemaAttribute("mjc:priority", 0),
             "weight": SchemaAttribute("mjc:solmix", 1.0),
-            "stiffness": SchemaAttribute("mjc:solref", [0.02, 1.0], solref_to_stiffness),
-            "damping": SchemaAttribute("mjc:solref", [0.02, 1.0], solref_to_damping),
+            "ke": SchemaAttribute(
+                "mjc:solref",
+                None,
+                usd_value_getter=_mjc_legacy_material_solref(solref_to_stiffness, "newton:contactStiffness"),
```
Milad-Rakhsha-NV, line 441: - I think the logging here should be a bit more informative than what you currently have in `_mjc_legacy_material_solref`. We need to inform users that conversion from solref/imp to stiffness/damping can be lossy (@eric-heiden could you please comment). 

- not sure what the behavior is if someone decides to author both stiffness/damping and solref/impd. I think we should have safeguards against use cases like that?

newton/_src/usd/schemas.py
andrewkaufman: Added to the Deprecated section:

> Deprecate `mjc:solref` on material prims for contact stiffness/damping; author `newton:contactStiffness`/`newton:contactDamping` on the bound `NewtonMaterialAPI` material instead. Per-shape `mjc:solref` (from `MjcGeomAPI`) is unchanged

Kept it explicit that per-shape solref is unaffected, since that's the one people are actually using.

newton/_src/usd/schemas.py
andrewkaufman: > not sure what the behavior is if someone decides to author both stiffness/damping and solref/impd. I think we should have safeguards against use cases like that?

I _think_ that should be covered here, if mujoco resolver was given preference, the result would prefer the solref values, otherwise the result would prefer the ke/kd values.

@jcarius-nv did explicitly request that the mujoco-usd-converter does output both attrs, so I think this will come up a bunch.

newton/_src/usd/schemas.py
andrewkaufman: With the old default of `[0.02, 1.0]`, every MuJoCo shape *without* an authored `mjc:solref` would return a converted ke/kd from that default. In `import_usd.py` the per-shape resolver value is checked when the material has no contact attrs authored — so the default solref would leak through as a per-shape override on every MuJoCo shape, even when the user never authored one.

Changing the default to `None` means "not authored" — the resolver returns `None` for shapes without an explicit `mjc:solref`, and only shapes where the user actually authored a per-geom solref get the converted value. The MuJoCo engine default still applies at simulation time via the builder default; this just prevents the *mapping* default from masquerading as an authored value.

newton/_src/usd/schemas.py
andrewkaufman: @Milad-Rakhsha-NV  I agree its unlikely to have been used in the wild & we're kind of advertising it for the first time via the Deprecated log now... but I guess we have to since its technically a behaviour change.

andrewkaufman: @eric-heiden I added the extra deprecation line

newton/_src/usd/schemas.py
andrewkaufman: @Milad-Rakhsha-NV if you think that approach exposes a bug though, please do point it out :pray:

newton/_src/usd/schemas.py
jcarius-nv: Right, I think we need to do this because in the mujoco-usd-converter we don't know yet which solver will be used in Newton afterwards.

I'm open to be convinced otherwise but right now I don't see a way around authoring both.

newton/_src/utils/import_usd.py
```diff
@@ -3073,12 +3087,8 @@ def _build_mass_info_from_shape_geometry(
                     "cfg": ModelBuilder.ShapeConfig(
                         ke=shape_ke,
                         kd=shape_kd,
-                        kf=usd.get_float_with_fallback(
-                            prim_and_scene, "newton:contact_kf", builder.default_shape_cfg.kf
-                        ),
-                        ka=usd.get_float_with_fallback(
-                            prim_and_scene, "newton:contact_ka", builder.default_shape_cfg.ka
-                        ),
+                        kf=shape_kf,
```
jcarius-nv, line 3104: This drops the previous shape/scene fallback for `newton:contact_kf` / `newton:contact_ka`. Before this change, `ShapeConfig.kf` / `ka` were populated via `usd.get_float_with_fallback((prim, physics_scene_prim), "newton:contact_kf" / "newton:contact_ka", ...)`, and those attributes were documented in `docs/concepts/collisions.rst`. Moving the canonical path to `NewtonMaterialAPI` is fine, but removing these without a deprecation path is a breaking change for existing authored USD. Please keep them as lowest-priority legacy fallbacks, emit deprecation warnings, and add regression coverage, including material `-inf` blocking fallback if that is the intended behavior.

newton/_src/utils/import_usd.py
```diff
@@ -2876,22 +2890,22 @@ def _build_mass_info_from_shape_geometry(
...
-                if shape_kd is None:
-                    shape_kd = builder.default_shape_cfg.kd
+                # Contact response: material (if authored) > per-shape resolver > builder default.
+                _default = builder.default_shape_cfg
+                shape_ke = material.ke if material.ke is not None and math.isfinite(material.ke) else _default.ke
+                shape_kd = material.kd if material.kd is not None and math.isfinite(material.kd) else _default.kd
+                shape_kf = material.kf if material.kf is not None and math.isfinite(material.kf) else _default.kf
+                shape_ka = material.ka if material.ka is not None and math.isfinite(material.ka) else _default.ka
+
+                for attr_key in ("ke", "kd"):
```
jcarius-nv: This skips shape-level resolution whenever the bound material has any `ke` / `kd` value, which means resolver priority no longer controls the common dual-authored MuJoCo case. For example, with `schema_resolvers=[SchemaResolverMjc(), SchemaResolverNewton()]`, a geom that has `mjc:solref` but is bound to a Newton material with `newton:contactStiffness` will take the material stiffness before the MJC shape resolver ever gets a chance. That conflicts with the resolver-priority model and with the changelog statement that per-shape `mjc:solref` is unchanged. Please make the precedence policy explicit and add a regression test for dual-authored Newton material + MJC geom assets.

newton/_src/utils/import_usd.py
```diff
@@ -2876,22 +2890,22 @@ def _build_mass_info_from_shape_geometry(
...
+                    if getattr(material, attr_key) is not None:
```
jcarius-nv: A smaller migration concern: when deprecated shape attrs are present but shadowed by material attrs, this code skips the shape lookup entirely, so the user gets no warning that the legacy attrs are now ignored. We should consider warning in that case as well; otherwise assets can carry stale `newton:contact_ke` / `newton:contact_kd` values indefinitely with no migration signal.

newton/_src/usd/schemas.py
jcarius-nv, line 441: I agree with the existing thread that the `mjc:solref` conversion warning should be more explicit. Since the conversion from MuJoCo `solref` / `solimp` semantics to Newton stiffness/damping is not generally lossless, the deprecation warning should say that directly so users understand why authored Newton material attrs are preferred.

jcarius-nv (dismissed):
Thanks for moving the canonical contact-response representation onto materials; that is the right direction. I think there are two compatibility/precedence issues to resolve before merging, plus a couple of tests that should pin down the intended behavior.

newton/_src/usd/schemas.py
eric-heiden: Now I'm also not sure if we should deprecate `mjc:solref` actually. The conversion we have here is indeed lossy, see #2964 that handles the `ke/kd`->`solref` conversion for MuJoCo which requires the `body_invweight0` to be known. So it will be a bit tricky in practice to tune `ke/kd` exactly to get the right `mjc:solref` behavior, so as @jcarius-nv said we should just allow both values. MuJoCo can directly use `mjc:solref` which is probably what users expect. In other cases we have conversions in place that are lossy at the importer level.

andrewkaufman: > Now I'm also not sure if we should deprecate mjc:solref actually.

Just to clarify, the code here does not deprecate `mjc:solref`, it keeps it as before (on the shape) & removes a fake mjc:solref (on the material) that never should have existed (mujoco schemas do not support mjc:solref on the material at all)

eric-heiden: Oh I see but then the deprecation message should maybe call that out, i.e. to author `mjc:solref` on the shape instead of or in addition to this?
> `f"'mjc:solref' on material prim is deprecated; "
                f"author '{material_attr}' on the bound NewtonMaterialAPI material instead.",`

newton/_src/usd/schemas.py
Milad-Rakhsha-NV: Got it, don't we need the same thing for other attributes as well (unrelated to this MR)

newton/_src/utils/import_usd.py
andrewkaufman: Good catch — added `newton:contact_kf` and `newton:contact_ka` as legacy SHAPE entries in the Newton resolver with the same deprecation-warning getter pattern as ke/kd. They now fire a `DeprecationWarning` pointing users to `newton:contactFrictionGain` / `newton:contactAdhesion` on `NewtonMaterialAPI`.

newton/_src/utils/import_usd.py
andrewkaufman: Fixed — the per-shape resolver is now checked first for all four attrs (ke/kd/kf/ka). Precedence is:

1. Per-shape resolver (MuJoCo `mjc:solref`, Newton legacy `newton:contact_*`)
2. Material (`NewtonMaterialAPI` contact attrs)
3. Builder default

So a geom with `mjc:solref` now correctly beats the bound material's `newton:contactStiffness` regardless of which resolver is listed first, since the per-shape lookup runs before the material check.

newton/_src/utils/import_usd.py
andrewkaufman: Addressed by the precedence fix — the per-shape resolver now always runs (it's checked before the material), so legacy attrs fire their deprecation warning regardless of whether a material value is also present. Users always see the migration signal when stale `newton:contact_*` attrs exist on their shapes.

newton/_src/usd/schemas.py
andrewkaufman: Updated the warning to recommend both paths:

> `mjc:solref` on material prim is deprecated; author `newton:contactStiffness` on the bound NewtonMaterialAPI material, or use per-shape `mjc:solref` (MjcGeomAPI) instead.

andrewkaufman: Updated the deprecation message to recommend both alternatives — `NewtonMaterialAPI` for material-level contact attrs, or per-shape `mjc:solref` (`MjcGeomAPI`) which remains supported and unchanged:

> `mjc:solref` on material prim is deprecated; author `newton:contactStiffness` on the bound NewtonMaterialAPI material, or use per-shape `mjc:solref` (MjcGeomAPI) instead.

eric-heiden: The new wording looks good to me!

eric-heiden (dismissed):

andrewkaufman (commented):
Address review feedback on precedence, legacy kf/ka, and solref warning

Milad-Rakhsha-NV (approved):
LGTM

jcarius-nv (approved):

PR #3008, opened 2026-05-30
Fix MuJoCo CPU inertia frame sync

Langford-H:
## Summary

Fix `SolverMuJoCo(use_mujoco_cpu=True)` syncing `mj_model.body_inertia` from Newton's eigendecomposition while preserving the compiled `mj_model.body_iquat`.

The previous behavior mixed two different inertia representations: sorted principal moments from Newton/MJWarp with the old compiled inertia-frame quaternion from MuJoCo C. For MJCF bodies whose `diaginertia` is valid but not descending, this permuted the effective body-frame inertia axes in the live CPU model.

This patch keeps the compiled CPU inertia frame when it already represents the Newton tensor, updates only the diagonal moments when that same frame still diagonalizes the updated tensor, and falls back to copying both the MJWarp principal moments and `body_iquat` only when the principal axes actually need to rotate.

## Regression covered

Adds a minimal one-body MJCF repro:

- one hinge joint around Z
- one primitive geom
- explicit `diaginertia="0.115841 0.082211 0.142951"`
- velocity actuator with `kv=1000000`, `ctrlrange=-4 4`, `forcerange=-600000 600000`

The test verifies that native MuJoCo and Newton's live MuJoCo CPU model reconstruct the same inertia matrix and produce the same one-step actuator response in `ctrl_direct=True` mode.

## Prior-art search

I searched the repository issues/discussions for `body_iquat`, `mj_model.body_inertia`, `principal axes`, and MuJoCo CPU inertia sync. I did not find an issue or discussion describing this exact runtime CPU permutation. The nearest related change is #2610, which intentionally preserved the compiled `body_iquat` to avoid repeated-eigenvalue instability; this PR fixes the complementary case where preserving `body_iquat` while still overwriting `body_inertia` made the pair inconsistent.

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_mujoco_cpu_preserves_mjcf_diagonal_inertia_axes
uv run --extra dev -m newton.tests -k body_inertia
uv run --extra dev -m newton.tests -k TestMuJoCoSolverMassProperties
uvx ruff check newton/_src/solvers/mujoco/solver_mujoco.py newton/tests/test_mujoco_solver.py
uvx ruff format --check newton/_src/solvers/mujoco/solver_mujoco.py newton/tests/test_mujoco_solver.py
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed CPU backend issue where diagonal inertia axes from MJCF files were incorrectly permuted in MuJoCo solver. Inertia frame orientation is now preserved when possible.

* **Tests**
  * Added regression test to verify correct inertia axis preservation and matching behavior with native MuJoCo in CPU mode.

[Review Change Stack](https://app.coderabbit.ai/change-stack/newton-physics/newton/pull/3008?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Langford-H:
Closing for now at the author's request; the fix will be reworked or resubmitted later if needed.

PR #3018, opened 2026-06-01
Fix MuJoCo CPU inertia frame sync

### LINKED ISSUE #3017: [BUG] SolverMuJoCo CPU backend produces incorrect diagonal inertia for asymmetric MJCF inertia
### Bug Description

`SolverMuJoCo(use_mujoco_cpu=True)` can produce incorrect dynamics for an MJCF body with asymmetric `diaginertia`.

The same MJCF loaded directly by native MuJoCo gives the expected result. The mismatch appears after Newton/MJWarp data is synchronized back into MuJoCo CPU.

On upstream `main` at `429312d5e505825fd5ae76994e2b52f9cb0d5587`, the final MuJoCo CPU model has reordered diagonal inertia but keeps the old inertial-frame orientation:

```text
SolverMuJoCo(use_mujoco_cpu=True):
  body_inertia=[0.142951 0.115841 0.082211]
  body_iquat=[1. 0. 0. 0.]
  mjw_body_inertia=[0.142951 0.115841 0.082211]
  mjw_body_iquat=[ 0.5 -0.5 -0.5 -0.5]
```

This makes the final MuJoCo CPU inertia tensor inconsistent. In a one-step hinge test, native MuJoCo and `SolverMuJoCo(use_mujoco_cpu=True)` produce different velocities:

```text
Native MuJoCo:
  effective hinge inertia=0.143951
  qfrc=600000 qvel=4168.08497 inferred inertia=0.143951

SolverMuJoCo(use_mujoco_cpu=True):
  effective hinge inertia=0.0832110028
  qfrc=600000 qvel=7210.5849 inferred inertia=0.0832110028

Comparison:
  qfrc ratio=1
  qvel ratio=1.72995151
  model inertia ratio=0.578050884
  inferred inertia ratio=0.578050884
```

Expected behavior:

`SolverMuJoCo(use_mujoco_cpu=True)` should match native MuJoCo for the same MJCF. If MJWarp re-expresses the inertia with a different principal-axis frame, the MuJoCo CPU backend should not copy only the diagonal inertia values while leaving the inertial-frame orientation stale.

[mujoco_cpu_inertia_repro.py](https://github.com/user-attachments/files/28462708/mujoco_cpu_inertia_repro.py)

### Reproduction Script

Please see the attached file:

```python

mujoco_cpu_inertia_repro.py

```

Run it from an unfixed Newton checkout:

```python

uv run --extra dev python mujoco_cpu_inertia_repro.py

```

The script compares native MuJoCo against `SolverMuJoCo(use_mujoco_cpu=True)` on a one-body MJCF with asymmetric `diaginertia`.

### System Information

- Newton: upstream `main` at `429312d5e505825fd5ae76994e2b52f9cb0d5587`
- Warp: `1.14.0rc2`
- Python: `3.12.13`
- OS: macOS `15.5` arm64
- GPU: no NVIDIA GPU on this machine; `nvidia-smi` is not available

Langford-H:
## Description

Fixes #3017

This fixes incorrect dynamics in `SolverMuJoCo(use_mujoco_cpu=True)` for MJCF bodies whose asymmetric `diaginertia` is re-expressed by MJWarp using a different principal-axis frame.

Before this change, the MuJoCo CPU backend could copy `mjw_model.body_inertia` back into the existing MuJoCo CPU `MjModel` without also synchronizing the matching `body_iquat`. This could leave the final MuJoCo CPU model with reordered diagonal inertia values but the old inertial-frame orientation.

The fix synchronizes the complete inertial representation from MJWarp to MuJoCo CPU:

- `body_inertia`
- `body_iquat`
- MuJoCo CPU simple-path metadata derived from the inertial frame

When any body inertia or inertial frame changes, the CPU backend invalidates `body_sameframe`, `body_simple`, and `dof_simplenum` before the existing `mj_setConst()` call recomputes derived constants.

This makes `SolverMuJoCo(use_mujoco_cpu=True)` match native MuJoCo for the same MJCF in the added regression test.

**Note on the removed `body_iquat` comment:**

The original code comment intentionally avoided overwriting `mj_model.body_iquat`, because MuJoCo's compiler chooses a stable principal-axis basis and Newton/MJWarp may choose an arbitrary valid basis for repeated eigenvalues. That concern is valid when the principal-axis frame is unchanged, such as pure mass or density scaling.

The bug here is different: the CPU path already copies `mjw_model.body_inertia`. If MJWarp re-expresses inertia with reordered principal moments, then `body_inertia` and `body_iquat` must be synchronized together. Copying only `body_inertia` leaves MuJoCo CPU with diagonal values from one frame and `body_iquat` from another frame.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_mujoco_cpu_inertia_frame_sync
uvx pre-commit run -a
git diff --check origin/main..HEAD
```

All commands pass locally.

## Bug fix

**Steps to reproduce:**

See the linked issue for the failing output from upstream `main`.

The bug can be reproduced by loading an MJCF model with one hinge body and asymmetric `diaginertia`, then comparing native MuJoCo against `SolverMuJoCo(use_mujoco_cpu=True)` after one actuator step.

**Minimal reproduction:**

The linked issue includes the standalone reproduction script as an attachment.

After this fix, the reproduction matches native MuJoCo:

```text
Native MuJoCo model:
  body_inertia=[0.115841 0.082211 0.142951]
  body_iquat=[1. 0. 0. 0.]
  effective hinge inertia=0.143951
  qfrc=600000 qvel=4168.08497 inferred inertia=0.143951

SolverMuJoCo(use_mujoco_cpu=True):
  body_inertia=[0.142951 0.115841 0.082211]
  body_iquat=[ 0.5 -0.5 -0.5 -0.5]
  mjw_body_inertia=[0.142951 0.115841 0.082211]
  mjw_body_iquat=[ 0.5 -0.5 -0.5 -0.5]
  body_sameframe=0
  body_simple=0
  dof_simplenum=[0]
  effective hinge inertia=0.143950997
  qfrc=600000 qvel=4168.08507 inferred inertia=0.143950997

Comparison:
  qfrc ratio=1
  qvel ratio=1.00000002
  model inertia ratio=0.999999978
  inferred inertia ratio=0.999999978
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed CPU backend dynamics to correctly handle asymmetric diagonal inertia models where principal moments could be reordered, ensuring consistent inertial-frame data.

* **Tests**
  * Added a regression test to verify inertia and inertial-frame synchronization between implementations and ensure matched dynamics (actuator forces and joint velocities).


vreutskyy (dismissed):
LGTM. The diagnosis is correct and the fix complements — rather than reverts — the CPU `body_iquat` change from #2610.

Quick reasoning, since this revisits a previously-removed write: my prior comment was added because Eric's first iteration unconditionally overwrote `mj_model.body_iquat` on every `BODY_INERTIAL_PROPERTIES` notify, which destabilised stiff WELD-loop tests (`Nan, Inf or huge value in QACC`) and broke single-contact equilibrium (`test_parallel_scenario` showed `352.81 N` vs the expected `441.45 N`). Both regressions traced to `wp.eig3()` picking an arbitrary valid basis for repeated eigenvalues and the overwrite rotating MuJoCo's compiled stable basis underneath the C solver.

Two changes have happened since that make the conditional sync in this PR safe:

1. `431a04d6` ("Keep axis-aligned inertia frames stable") taught `update_body_inertia_kernel` to short-circuit to identity `body_iquat` for already-diagonal inputs, so the degenerate-eigenvalue case never produces a spurious basis flip in MJWarp.
2. This PR's `~np.isclose(...)` guard in `_sync_mjw_inertias_to_mjc_cpu` makes the write a no-op whenever MJWarp's iquat already matches MuJoCo's compiled one. With (1) in place, that's true for every axis-aligned case I was worried about; the helper only triggers when MJWarp genuinely picked a different principal-axes frame (the asymmetric-`diaginertia` case this PR fixes).

I ran the original regression cases plus the broader notify path locally on this commit, and all pass: `test_parallel_scenario`, `test_fixed_loop_joint` (both `test_fixed_loop_joint_mujoco_cpu_*` variants), `test_randomize_body_{mass,inertia,com}`, `test_isaaclab_mass_randomization_loop`, and the new `test_mujoco_cpu_inertia_frame_sync`.

One thing worth noting since it's not obvious from the diff: invalidating `body_sameframe` / `body_simple` / `dof_simplenum` is load-bearing — `mj_setConst()` does recompute `dof_M0` / `dof_invweight0`, but it does **not** recompute the simple-path metadata, so without this invalidation the runtime would keep using a stale optimization path that assumes the old iquat. Good catch.

ruziniuuuuu:
🤣

Langford-H:
I just resolved the conflict in `CHANGELOG.md`; there was no additional source-code alteration. It seems I should not make further changes here without maintainer direction. My bad, sorry.

vreutskyy (approved):

PR #3020, opened 2026-06-01
Point install/actuator docs at ONNX, not torch

adenzler-nvidia:
## Description

RL policy inference now runs on Newton's Warp-backed ONNX runtime, so the `torch` extras are no longer required to run the examples. This brings the docs in line with that change so PyTorch is mentioned only where it is still needed (Kamino RL training and the deprecated TorchScript checkpoint path):

- `docs/guide/installation.rst`: drop the `torch-cu12` install from the RL-inference quick start (the `examples` extra already pulls in ONNX support) and reframe the `torch-cu12`/`torch-cu13` extras-table rows as Kamino RL training scripts. Also removes a now-orphaned label.
- `docs/concepts/actuators.rst`: correct the claim that the neural-network controllers require PyTorch and are not graphable — the ONNX-backed controllers are graphable; only the deprecated TorchScript checkpoint path requires PyTorch.
- `docs/guide/development.rst`: reframe the torch-gated tests around the deprecated TorchScript checkpoint path rather than generic RL inference, and shorten the paragraph.

## Checklist

- [ ] New or existing tests cover these changes (N/A — docs only)
- [ ] `CHANGELOG.md` has been updated (not needed: docs-only copy fix)

## Test plan

```bash
uvx pre-commit run --files docs/guide/installation.rst docs/guide/development.rst docs/concepts/actuators.rst
# Confirm no remaining torch in the getting-started flow and no orphaned label refs:
grep -rn 'running-examples' docs --include='*.rst'
```

## Summary by CodeRabbit

* **Documentation**
  * Clarified that built-in controllers are graphable and differentiable, with ONNX-based controllers supported (TorchScript deprecated path excluded)
  * Updated neural controller descriptions to reflect ONNX-based implementations and state/buffer behaviors
  * Specified PyTorch is only required for a deprecated checkpoint path and for training scripts/examples, not for ONNX inference
  * Simplified RL inference example and streamlined installation/testing guidance

### DIFF
--- a/docs/guide/installation.rst
+++ b/docs/guide/installation.rst
@@ -110,8 +110,6 @@ with other packages:
     ``examples`` extra). If you encounter installation errors, we recommend upgrading to a later
     Python version, or follow the :doc:`development` guide to install Newton using ``uv``.
 
-.. _running-examples:
-
 Running Examples
 ^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


docs/guide/installation.rst
```diff
@@ -122,27 +120,14 @@ After installing Newton with the ``examples`` extra, launch the default
 
     python -m newton.examples
 
-Run an example that runs RL policy inference. Choose the extra matching your
-NVIDIA driver's CUDA support (``torch-cu12`` for CUDA 12.x, ``torch-cu13`` for
-CUDA 13.x) and the corresponding pytorch wheel (e.g, ``128`` for CUDA 12.8); run ``nvidia-smi``
-to check the supported CUDA version (shown in the top-right corner of the output):
+Run an example that runs RL policy inference. RL policies ship as ONNX files and
+run on Newton's Warp-backed runtime, so the ``examples`` extra is all you need —
+no PyTorch installation is required:
```
shi-eric: This text as it is written has been biased by the historical page, so it is reading as if it has to explain that PyTorch is not required. But a reader to the page isn't necessarily aware about the historical PyTorch dependency in Newton, so it might seem odd to them that we state `no PyTorch installation is required`.

adenzler-nvidia: Good point — a reader here has no context for the historical PyTorch dependency, so contrasting against it is more confusing than helpful. I dropped the explanation entirely and reduced it to just running the example.

shi-eric (approved):

PR #3022, opened 2026-06-01
Use region fallback in AWS workflows

shi-eric:
## Description

Update the remaining EC2-backed GPU workflows to use the dynamic AWS runner discovery path already used by `aws_gpu_tests.yml`.

This removes fixed `us-east-2` AMI/subnet/security-group runner setup from:

- `.github/workflows/aws_gpu_benchmarks.yml`
- `.github/workflows/minimum_deps_tests.yml`
- `.github/workflows/warp_nightly_tests.yml`
- `.github/workflows/mujoco_warp_tests.yml`

The workflows now discover tagged runner networking across the shared ordered region list, pass the generated `availability-zones-config` to `machulav/ec2-github-runner`, and use the selected launch region during teardown. The discovery helper docstring now reflects that it is shared by EC2-backed GPU workflows.

## Checklist

## Test plan

```bash
uv run python -m py_compile scripts/ci/discover_aws_runner_config.py
uvx pre-commit run -a
```

Also ran a local workflow audit confirming the upgraded workflows no longer contain the fixed security group, hardcoded subnet IDs, or local Deep Learning AMI lookup, and that each workflow includes runner discovery, selected region output, launch-region validation, and selected-region teardown.

## Bug fix

**Steps to reproduce:**

1. Trigger one of the remaining EC2-backed GPU workflows when `us-east-2` lacks suitable GPU capacity.
2. Observe the workflow only has fixed `us-east-2` runner networking available.
3. The runner launch can fail even when another configured AWS region has capacity.

**Minimal reproduction:**

```python
# CI-only runner provisioning issue; no Python reproduction applies.
```

## New feature / API change

```python
# No public API change.
```

## Summary by CodeRabbit

* **Chores**
  * Added a prioritized region-candidates input and ordered region preference for selecting AWS regions.
  * Start job now advertises the discovered launch region; runner startup uses dynamic discovery of availability-zone and network configuration instead of hardcoded values.
  * Hardened teardown: skip stop when start was skipped, validate discovered region when an instance ID exists, configure credentials with the discovered region, and only stop when an instance exists.

### DIFF
--- a/.github/workflows/aws_gpu_benchmarks.yml
+++ b/.github/workflows/aws_gpu_benchmarks.yml
@@ -7,12 +7,14 @@ name: GPU Benchmarks on AWS EC2 (Reusable)
 # Workflow configuration variables
 env:
   AWS_REGION: us-east-2
+  # Prefer the validated US capacity order before Asia-Pacific fallback.
+  AWS_REGION_CANDIDATES: ${{ inputs.region-candidates || 'us-east-1 us-east-2 us-west-2 ap-northeast-1 ap-northeast-2' }}
   AWS_INSTANCE_TYPE: g7e.2xlarge
   AWS_VOLUME_SIZE: 92
   AWS_VOLUME_TYPE: gp3
-  AWS_SECURITY_GROUP_IDS: sg-07807c44e7f2a368a
   AWS_ROLE_ARN: arn:aws:iam::968945269301:role/newton-physics-newton-ec2-github-runner-role
   AWS_ROLE_DURATION: 7200
+  AWS_RUNNER_RESOURCE_TAG: newton-github-runner
   HOME: /actions-runner
 
 on:
@@ -26,6 +28,11 @@ on:
       base_ref:
         description: 'Base ref to compare against (for ASV)'
         required: true
+        type: string
+        default: ''
+      region-candidates:
+        description: 'Space-separated AWS regions to try'
+        required: false
         type: string
         default: ''
     secrets:
@@ -43,11 +50,17 @@ jobs:
     outputs:
       label: ${{ steps.start-ec2-runner.outputs.label }}
       ec2-instance-id: ${{ steps.start-ec2-runner.outputs.ec2-instance-id }}
+      region: ${{ steps.start-ec2-runner.outputs.region }}
     steps:
       - name: Harden the runner (Audit all outbound calls)
         uses: step-security/harden-runner@fa2e9d605c4eeb9fcad4c99c224cee0c6c7f3594  # v2.16.0
         with:
           egress-policy: audit
+
+      - name: Checkout repository
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # v6.0.1
+        with:
+          persist-credentials: false
 
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@8df5847569e6427dd6c4fb1cf565c83acfa8afa7  # v6.0.0
@@ -55,20 +68,11 @@ jobs:
           aws-region: ${{ env.AWS_REGION }}
           role-to-assume: ${{ env.AWS_ROLE_ARN }}
           role-duration-seconds: ${{ env.AWS_ROLE_DURATION }}
-      - name: Get the latest AWS Deep Learning Base GPU AMI
-        run: |
-          echo "Finding the latest AWS Deep Learning Base GPU AMI..."
-          LATEST_AMI_ID=$(aws ec2 describe-images --region ${{ env.AWS_REGION }} \
-            --owners amazon \
-            --filters 'Name=name,Values=Deep Learning Base AMI with Single CUDA (Ubuntu 22.04) ????????' 'Name=state,Values=available' \
-            --query 'reverse(sort_by(Images, &CreationDate))[:1].ImageId' \
-            --output text)
-          if [[ -z "$LATEST_AMI_ID" ]]; then
-            echo "❌ No AMI ID found. Exiting."
-            exit 1
-          fi
-          echo "Latest AMI ID found: $LATEST_AMI_ID"
-          echo "LATEST_AMI_ID=$LATEST_AMI_ID" >> "$GITHUB_ENV"
+
+      - name: Discover AWS runner networking
+        id: discover-runner-config
+        run: python scripts/ci/discover_aws_runner_config.py
+
       - name: Start EC2 runner
         id: start-ec2-runner
         uses: machulav/ec2-github-runner@343a1b2ae682e681c3cec9a235d882da17ff04ef  # v2.6.1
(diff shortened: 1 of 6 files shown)


.github/workflows/aws_gpu_benchmarks.yml
```diff
@@ -204,21 +204,29 @@ jobs:
...
       - gpu-benchmarks
-    if: always() && github.repository == 'newton-physics/newton'
+    if: always() && needs.start-runner.result != 'skipped' && github.repository == 'newton-physics/newton'
     steps:
       - name: Harden the runner (Audit all outbound calls)
         uses: step-security/harden-runner@fa2e9d605c4eeb9fcad4c99c224cee0c6c7f3594  # v2.16.0
         with:
           egress-policy: audit
 
+      - name: Validate launch region
```
adenzler-nvidia, line 214: **`Validate launch region` can't actually prevent a wrong-region teardown.**

Because a custom `if:` replaces GitHub's implicit `if: success()`, the `Configure AWS credentials` and `Stop EC2 runner` steps below — guarded only by `ec2-instance-id != ''` — will still run after this step exits 1. So if the instance launched but the `region` output is empty:

1. this step fails (job goes red), but
2. `Configure AWS credentials` runs with `region || env.AWS_REGION` → falls back to `us-east-2`, and
3. `Stop EC2 runner` targets `us-east-2` for an instance that may live in `us-east-1`/`ap-northeast-*` → `InvalidInstanceID.NotFound` → the instance is never terminated (cost leak).

This is the change that makes it reachable: previously everything was single-region `us-east-2`, so the fallback could never resolve to the wrong region. Now multi-region launches are enabled.

Question before acting: can `machulav/ec2-github-runner@v2.6.1` set `ec2-instance-id` without also setting `region`? If it always sets both atomically, this is just unreachable defensive code. If not, suggest gating teardown on a real region instead of relying on this step to block:

```yaml
- name: Configure AWS credentials
  if: ${{ needs.start-runner.outputs.ec2-instance-id != '' && needs.start-runner.outputs.region != '' }}
  with:
    aws-region: ${{ needs.start-runner.outputs.region }}   # drop the || env.AWS_REGION fallback
- name: Stop EC2 runner
  if: ${{ needs.start-runner.outputs.ec2-instance-id != '' && needs.start-runner.outputs.region != '' }}
- name: Fail with manual-cleanup instructions
  if: ${{ needs.start-runner.outputs.ec2-instance-id != '' && needs.start-runner.outputs.region == '' }}
  run: |
    echo "::error::Instance ${{ needs.start-runner.outputs.ec2-instance-id }} launched but region is empty; terminate manually across: ${{ env.AWS_REGION_CANDIDATES }}."
    exit 1
```

This applies identically to the same step in `minimum_deps_tests.yml`, `mujoco_warp_tests.yml`, and `warp_nightly_tests.yml`, and is inherited from `aws_gpu_tests.yml`. Since the flaw is inherited, a follow-up fixing all five workflows together may be cleaner — but flagging it here since this PR is what makes it live.

.github/workflows/minimum_deps_tests.yml
```diff
@@ -6,12 +6,14 @@ name: Minimum Dependency Version Tests on AWS EC2 (Reusable)
 
 env:
   AWS_REGION: us-east-2
+  # Prefer the validated US capacity order before Asia-Pacific fallback.
+  AWS_REGION_CANDIDATES: us-east-1 us-east-2 us-west-2 ap-northeast-1 ap-northeast-2
```
adenzler-nvidia: Minor consistency nit: `aws_gpu_benchmarks.yml` and the reference `aws_gpu_tests.yml` use the parameterizable form `${{ inputs.region-candidates || '...' }}` (and add a `region-candidates` input), but this file (and `mujoco_warp_tests.yml`, `warp_nightly_tests.yml`) hardcode the literal list and add no input, despite having a `workflow_dispatch` trigger.

Not a functional issue — the discover script reads `AWS_REGION_CANDIDATES` from the env either way, the list is identical, and no caller overrides it today. But if matching the reference is the intent, consider switching these three to the `inputs.region-candidates ||` form and adding the matching input. Otherwise feel free to resolve — possibly intentional.

adenzler-nvidia (commented):
Reviewed via the pr-review-toolkit (code review + silent-failure + comment analysis). This is a clean, faithful propagation of the dynamic-runner-discovery pattern from `aws_gpu_tests.yml` — consistency with the reference is excellent and the docstring/comment changes are accurate. Two things worth a look before merge, left inline: a teardown-region concern (also present in the reference, but this PR is what makes it reachable) and a small `AWS_REGION_CANDIDATES` consistency nit.

.github/workflows/aws_gpu_benchmarks.yml
shi-eric: Addressed the teardown side of this in `84da8947` by removing the `|| env.AWS_REGION` fallback from all five EC2 runner workflows. Teardown now uses only `needs.start-runner.outputs.region`, so it cannot silently guess the workflow default region for a launched instance.

I do not think the exact `Validate launch region` failure path is reachable as described: GitHub still applies the implicit `success()` status check to step `if:` expressions unless they include a status function, and the pinned action sets `ec2-instance-id` and `region` from the same successful launch result. But dropping the fallback makes the multi-region invariant explicit and avoids wrong-region teardown if that invariant is ever broken.

.github/workflows/minimum_deps_tests.yml
shi-eric: Leaving these three fixed intentionally. No current caller needs to override the candidate region list, so adding `region-candidates` inputs here would expand the workflow interface for unused configurability. The env list remains identical to the reference workflow, and we can add an input later if a caller actually needs to vary it.

adenzler-nvidia (approved):
Both points from the previous review are handled.

The teardown `Configure AWS credentials` steps no longer fall back to `env.AWS_REGION` (now `aws-region: ${{ needs.start-runner.outputs.region }}`), and each workflow gained a `Validate launch region` step. Since `configure-aws-credentials` treats `aws-region` as required, an empty launch region now fails loudly instead of silently tearing down against `us-east-2` — that closes the wrong-region teardown path I was worried about.

On the consistency note: the `region-candidates` override input stays on `aws_gpu_tests.yml` and `aws_gpu_benchmarks.yml` while the other three keep the hardcoded candidate list. That's the leave-as-is option I'd flagged as fine, so no objection.

LGTM.

shi-eric (commented):
@coderabbitai I checked this against the current PR head and do not think this finding is valid.

The `Validate launch region` step failing does not let the later teardown steps run: GitHub Actions applies the implicit `success()` status check to step `if:` expressions unless the expression includes a status function. The later `Configure AWS credentials` / `Stop EC2 runner` steps do not include one, so they are skipped after validation fails.

The state is also not expected from the pinned action in this workflow. `discover_aws_runner_config.py` puts `region` on every candidate, and `machulav/ec2-github-runner@343a1b2...` returns `ec2InstanceId` and `region` from the same successful launch result before setting both outputs.

The suggested broad tag sweep would be risky and is not keyed correctly: `newton-github-runner` is the tag value used to discover networking resources, while launched instances are tagged `Name=ec2-github-runner`. Sweeping by broad runner tags could terminate a concurrent active runner. If we add orphan sweeping later, it should first add a run-specific instance tag such as `GitHub-Run-ID=${{ github.run_id }}` and sweep only that run.

PR #3026, opened 2026-06-01
Add fisheye and USD pinhole camera ray helpers to SensorTiledCamera

daniela-hase:
## Description

Add fisheye and aperture-based camera ray helpers to `SensorTiledCamera.utils`, plus USD pinhole camera ray and transform utilities.

**New `SensorTiledCamera.utils` methods:**
- `compute_fisheye_camera_rays_opencv()` — OpenCV `(fx, fy, cx, cy, k1–k4)` fisheye model
- `compute_fisheye_camera_rays_ftheta()` — F-theta polynomial model
- `compute_fisheye_camera_rays_kannala_brandt()` — Kannala-Brandt `k0–k3` model
- `compute_usd_pinhole_camera_rays()` — reads standard `UsdGeom.Camera` perspective attributes directly
- `compute_usd_camera_transforms()` — builds `(camera_count, world_count)` transform arrays from USD prims, with automatic stage-to-model up-axis conversion

**Extended `compute_pinhole_camera_rays()`** to accept `focal_length`, `horizontal_aperture`, `vertical_aperture`, and aperture offset parameters (in addition to the existing FOV path). All ray
helpers support an `out_rays`/`camera_index` pair for writing into a preallocated shared buffer.

**Bug fixes:**
- Zero-length rays (e.g. from masked fisheye pixels outside `max_fov`) now write the clear-value sentinel in `render_megakernel` instead of producing undefined geometry hits.
- `convert_ray_depth_to_forward_depth_kernel` preserves the depth sentinel for zero-direction rays.

## Checklist

## Test plan

uv run --extra dev -m newton.tests -k "test_sensor_usd_camera or test_sensor_tiled_camera_forward_depth"

New `newton/tests/test_sensor_usd_camera.py` covers:
- Mathematical round-trips for all three fisheye models (known-angle → ray direction)
- Aperture/FOV equivalence for pinhole
- Aperture offset principal-ray shift
- USD attribute round-trip (`UsdGeom.Camera` prim and schema both accepted)
- `out_rays`/`camera_index` preallocated buffer writes
- `max_fov` masking producing zero rays
- Zero-ray rendering preserving clear color and depth sentinels

## New feature / API change

```python
from pxr import Usd, UsdGeom

stage = Usd.Stage.Open("scene.usda")
usd_camera = stage.GetPrimAtPath("/World/Camera")

sensor = SensorTiledCamera(model)

# Pinhole rays directly from a USD camera prim
camera_rays = sensor.utils.compute_usd_pinhole_camera_rays(640, 480, usd_camera)
camera_transforms = sensor.utils.compute_usd_camera_transforms(usd_camera)

# Or fisheye rays from calibration values
camera_rays = sensor.utils.compute_fisheye_camera_rays_opencv(
    640, 480, fx=500.0, fy=500.0, cx=320.0, cy=240.0, k1=0.1, k2=-0.05
)

color = sensor.utils.create_color_image_output(640, 480)
sensor.update(state, camera_transforms, camera_rays, color_image=color)
```

## Summary by CodeRabbit

* **New Features**
  * USD camera support in tiled cameras: derive render inputs from USD camera prims with multiple lens models (pinhole, OpenCV fisheye, F‑theta, Kannala‑Brandt K3) and optional preallocated ray outputs.

* **Documentation**
  * Added guide showing how to compute camera rays/transforms from USD and use fisheye helpers.

* **Behavioral**
  * Rendering now consistently writes clear outputs for invalid or zero-direction rays.

* **Tests**
  * New tests for USD camera rays, lens models, transforms, and clear-output behavior.


jcarius-nv:

docs/concepts/sensors.rst
```diff
@@ -107,6 +107,38 @@ attributes, and usage examples.
...
+       state,
+       camera_transforms,
+       camera_rays,
+       color_image=color,
+   )
+
+For fisheye cameras, extract the calibration values from your chosen USD attributes and call one of
+``compute_fisheye_camera_rays_opencv()``, ``compute_fisheye_camera_rays_ftheta()``, or
+``compute_fisheye_camera_rays_kannala_brandt()``. Each fisheye helper builds rays for one camera; pass
+``out_rays`` and ``camera_index`` to fill a shared ray buffer.
```
eric-heiden: These function names would be good to have as links to the actual documentation of these functions.

daniela-hase: Good call, done! :)

newton/_src/sensors/warp_raytrace/utils.py
```diff
@@ -380,45 +368,432 @@ def create_hdr_color_image_output(self, width: int, height: int, camera_count: i
         )
 
     def compute_pinhole_camera_rays(
```
eric-heiden: I realize we already merged some of this before but generally we follow the common-prefix convention in Newton, so can we rename the functions to the following?
* `compute_camera_rays_pinhole()`
* `compute_camera_rays_usd_pinhole()`
* `compute_camera_rays_fisheye_opencv()`
* ...

Of course this needs to be properly deprecated. @mmacklin for viz

newton/_src/sensors/warp_raytrace/utils.py
```diff
@@ -380,45 +368,432 @@ def create_hdr_color_image_output(self, width: int, height: int, camera_count: i
...
+                out_rays,
+            ],
+            device=self.__render_context.device,
+        )
+
+        return out_rays
+
+    def compute_usd_camera_transforms(
+        self,
+        cameras: Any | list[Any] | tuple[Any, ...] | list[list[Any]] | tuple[tuple[Any, ...], ...],
```
eric-heiden: Can we simplify these ugly-looking type hints?

newton/_src/sensors/warp_raytrace/utils.py
daniela-hase: I renamed the functions to `compute_camera_rays_*` :)

newton/_src/sensors/warp_raytrace/utils.py
daniela-hase: I've changed them, take a look if you like it more this way :)

I am still not super sure I like them, but .. there is little that can be done realistically I think, the function just accepts a lot of possibilities

eric-heiden:
Sorry for the delay reviewing this PR! The USD and fisheye camera helpers are a useful addition. A few API consistency and transform-contract issues need tightening before this is ready.

newton/_src/sensors/warp_raytrace/utils.py
```diff
@@ -379,46 +367,497 @@ def create_hdr_color_image_output(self, width: int, height: int, camera_count: i
             device=self.__render_context.device,
         )
 
-    def compute_pinhole_camera_rays(
-        self, width: int, height: int, camera_fovs: float | list[float] | np.ndarray | wp.array[wp.float32]
+    def compute_camera_rays_pinhole(
+        self,
+        width: int,
+        height: int,
+        camera_fovs: float | list[float] | np.ndarray | wp.array[wp.float32] | None = None,
```
eric-heiden: ⚪ `camera_fovs` defaults to `None` and the docstring says aperture parameters can replace it, so the public type needs to allow `None` too. Otherwise generated type info advertises a non-optional argument with a `None` default.

Please keep the annotation aligned with the runtime contract, for example `float | list[float] | np.ndarray | wp.array[wp.float32] | None = None`.

newton/_src/sensors/warp_raytrace/camera_utils.py
```diff
@@ -0,0 +1,538 @@
...
+    if isinstance(param, list | tuple | np.ndarray):
+        values = np.asarray(param, dtype=np.float32).reshape(-1)
+        if values.size == camera_count:
+            return wp.array(values, dtype=wp.float32, device=device)
+        if values.size == 1:
+            return wp.full((camera_count,), value=float(values[0]), dtype=wp.float32, device=device)
+        raise ValueError(f"{name} must have length 1 or {camera_count}.")
+
+    if param.size != camera_count:
+        raise ValueError(f"{name} must have length {camera_count}.")
```
eric-heiden: 🔴 A length-1 `wp.array` is rejected here, while a scalar, list, or NumPy array of length 1 broadcasts to `camera_count`. That makes equivalent intrinsic inputs behave differently depending on whether the caller already has the values in a Warp array.

Please mirror the list/NumPy path for `wp.array` inputs, for example by accepting `param.size == 1` and returning a filled array on the target device.

newton/_src/sensors/warp_raytrace/utils.py
```diff
@@ -379,46 +367,497 @@ def create_hdr_color_image_output(self, width: int, height: int, camera_count: i
...
+        view values or aperture/focal-length values.
 
         Args:
             width: Image width [px].
             height: Image height [px].
-            camera_fovs: Vertical FOV angles [rad], shape ``(camera_count,)``.
+            camera_fovs: Vertical FOV angles [rad], shape
+                ``(camera_count,)``. Required unless *focal_length*,
+                *horizontal_aperture*, and *vertical_aperture* are supplied.
+            focal_length: Focal length in the same units as the apertures.
```
eric-heiden: 🟡 These public physical camera parameters do not state their units. Users matching USD cameras need to know whether to pass USD-style millimeters or converted SI values, even though the math mostly uses ratios.

Please add units to `focal_length`, the apertures, and aperture offsets, such as `[mm or any consistent unit]`, and note that the unit must match across these values.

newton/_src/sensors/warp_raytrace/camera_utils.py
```diff
@@ -0,0 +1,538 @@
...
+    k2: wp.float32,
+    k3: wp.float32,
+    k4: wp.float32,
+    max_theta: wp.float32,
+):
+    if radius <= 1.0e-7:
+        return wp.float32(0.0)
+
+    # When k0 != 0 the polynomial has a nonzero floor at theta=0 (r(0) = k0).
+    # Pixels inside that central circle are undefined by the model; return theta=0 (forward).
```
eric-heiden, line 339: 🔴 This range check assumes the f-theta radius polynomial is monotonic on `[0, max_theta]`. If the polynomial has an interior maximum and then falls below the requested `radius` at `max_theta`, this returns `-1.0` even though a valid theta exists.

Please either validate or document that the distortion polynomial must be monotonically increasing over the supported FOV. The same monotonicity assumption should be called out for the other fisheye inverse solvers too.

newton/_src/sensors/warp_raytrace/camera_utils.py
```diff
@@ -0,0 +1,538 @@
...
+    from ...core import Axis, quat_between_axes  # noqa: PLC0415
+    from ...usd.utils import get_transform  # noqa: PLC0415
+
+    time_code = _coerce_usd_time(time)
+    xform_cache = UsdGeom.XformCache(time_code)
+
+    def world_transform(usd_camera: Any) -> wp.transformf:
+        transform = get_transform(usd_camera.GetPrim(), local=False, xform_cache=xform_cache)
+        if target_up_axis is not None:
+            stage_up_axis = Axis.from_string(str(UsdGeom.GetStageUpAxis(usd_camera.GetPrim().GetStage())))
```
eric-heiden, line 220: 🔴 `world_transform()` only composes the USD stage transform with the up-axis conversion. It does not include the scene placement transform passed to `ModelBuilder.add_usd(path, xform=...)`, while imported geometry is placed under that transform.

That means a camera returned here can be offset from the imported scene by the user's `add_usd` transform. Please either accept a scene transform to compose into the result, or document clearly that callers must manually compose the returned camera transforms with their import placement transform.

newton/_src/sensors/warp_raytrace/utils.py
```diff
@@ -379,46 +367,497 @@ def create_hdr_color_image_output(self, width: int, height: int, camera_count: i
...
+    ) -> wp.array4d[wp.vec3f]:
+        """Compute camera-space ray directions for USD pinhole cameras.
+
+        Reads standard ``UsdGeom.Camera`` perspective attributes and forwards
+        them to :meth:`compute_camera_rays_pinhole`.
+
+        Args:
+            width: Image width [px].
+            height: Image height [px].
+            cameras: USD camera prim, ``UsdGeom.Camera``, or a flat sequence
```
eric-heiden, line 622: 🟡 This ray helper only accepts a flat camera sequence, but `compute_usd_camera_transforms` accepts a 2D per-world camera layout. A caller can compute distinct per-world poses, but cannot compute matching per-world USD intrinsics through this helper without looping.

Please document the constraint on `compute_usd_camera_transforms` or make the input contracts consistent. Since `camera_rays` has no world axis, the 2D transform layout is only safe when all worlds share the same camera intrinsics.

newton/_src/sensors/warp_raytrace/utils.py
```diff
@@ -379,46 +367,497 @@ def create_hdr_color_image_output(self, width: int, height: int, camera_count: i
...
+        """Compute camera-space ray directions for OpenCV fisheye cameras.
+
+        Args:
+            width: Output image width [px].
+            height: Output image height [px].
+            fx: Horizontal focal length [px].
+            fy: Vertical focal length [px].
+            cx: Principal point x-coordinate [px].
+            cy: Principal point y-coordinate [px].
+            image_width: Calibration image width [px]. If ``None``, uses
```
eric-heiden, line 681: 🟡 The fisheye helpers use different public names for the same calibration image size concept: OpenCV uses `image_width` and `image_height`, while the f-theta and Kannala-Brandt helpers use `nominal_width` and `nominal_height`. A caller switching models has to rename kwargs even though the values serve the same purpose.

Please use one kwarg pair across these public helpers, or accept aliases and document the model-specific terminology.

newton/_src/sensors/warp_raytrace/utils.py
daniela-hase: I am not sure I understand what the request here is, the type hint seems to correctly include `None` already, doesn't it?

newton/_src/sensors/warp_raytrace/camera_utils.py
daniela-hase: Addressed this in my [latest commit](https://github.com/newton-physics/newton/pull/3026/commits/e3db6f8a1b1854a38f1c235f09c45d18be58fc35)

newton/_src/sensors/warp_raytrace/utils.py
daniela-hase: Addressed this in my [latest commit](https://github.com/newton-physics/newton/pull/3026/commits/e3db6f8a1b1854a38f1c235f09c45d18be58fc35)

newton/_src/sensors/warp_raytrace/camera_utils.py
daniela-hase: Addressed this in my [latest commit](https://github.com/newton-physics/newton/pull/3026/commits/e3db6f8a1b1854a38f1c235f09c45d18be58fc35)

newton/_src/sensors/warp_raytrace/camera_utils.py
daniela-hase: Addressed this in my [latest commit](https://github.com/newton-physics/newton/pull/3026/commits/e3db6f8a1b1854a38f1c235f09c45d18be58fc35)

newton/_src/sensors/warp_raytrace/utils.py
daniela-hase: Addressed this in my [latest commit](https://github.com/newton-physics/newton/pull/3026/commits/e3db6f8a1b1854a38f1c235f09c45d18be58fc35)

newton/_src/sensors/warp_raytrace/utils.py
daniela-hase: Addressed this in my [latest commit](https://github.com/newton-physics/newton/pull/3026/commits/e3db6f8a1b1854a38f1c235f09c45d18be58fc35)

newton/_src/sensors/warp_raytrace/utils.py
```diff
@@ -380,46 +383,548 @@ def create_hdr_color_image_output(self, width: int, height: int, camera_count: i
             device=self.__render_context.device,
```
eric-heiden, line 380: 🔵 `camera_fovs` defaults to `None`, and aperture-only calls are supported, but the annotation excludes `None`. This gives type checkers and generated documentation a contract that disagrees with the runtime behavior. Please add `| None` to the annotation.

eric-heiden:
🟠 `CHANGELOG.md` needs two consistency fixes. The `[Unreleased]` section has two `### Deprecated` headings, with VBD damping changes incorrectly placed under the first one. Please consolidate the headings and move those entries under `### Changed`. Also update the `SensorRaycast` removal guidance to recommend `compute_camera_rays_pinhole()` instead of the helper deprecated by this PR.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Added
 
+- Add scalar value-based OpenCV, F-theta, and Kannala-Brandt fisheye camera ray helpers to `SensorTiledCamera.utils`, plus pinhole aperture/focal-length parameters, `compute_camera_rays_usd_pinhole()`, and optional preallocated ray output writes.
```
eric-heiden: ⚪ This entry omits the new public `compute_usd_camera_transforms()` helper. It also does not record that forward-depth conversion now preserves the clear-depth sentinel for zero-direction rays and non-positive depths. Please mention the transform helper under `Added`, using its final API name, and add the sentinel behavior under `Fixed`.

newton/_src/sensors/warp_raytrace/camera_utils.py
```diff
@@ -0,0 +1,548 @@
...
+    lo = wp.float32(0.0)
+    hi = max_theta
+    for _i in range(24):
+        mid = (lo + hi) * 0.5
+        if _kannala_brandt_k3_radius(mid, k0, k1, k2, k3) < radius:
+            lo = mid
+        else:
+            hi = mid
+    return (lo + hi) * 0.5
+
```
eric-heiden, line 386: 🟠 The fisheye kernels write the result of `_fisheye_direction_from_theta()` directly, while the pinhole kernels normalize explicitly. The fisheye result is unit-length by construction, but that invariant is not documented. Please add a brief comment to `_fisheye_direction_from_theta()` explaining this and noting that the zero vector is reserved for invalid rays.

newton/_src/sensors/warp_raytrace/utils.py
```diff
@@ -380,46 +383,548 @@ def create_hdr_color_image_output(self, width: int, height: int, camera_count: i
...
+
+        return out_rays
+
+    def compute_usd_camera_transforms(
+        self,
+        cameras: camera_utils.UsdCameraGridInput,
+        *,
+        time: camera_utils.UsdTime | None = None,
+        xform: Any | None = None,
+    ) -> wp.array2d[wp.transformf]:
```
eric-heiden: 🔵 This name sorts separately from the new `compute_camera_*` helper family, which makes the related transform helper harder to discover through autocomplete. Please rename it to `compute_camera_transforms_usd()` before release and update its documentation, changelog entry, and call sites.

newton/_src/sensors/warp_raytrace/camera_utils.py
```diff
@@ -0,0 +1,548 @@
...
+        if hasattr(camera, "GetProjectionAttr") and hasattr(camera, "GetPrim"):
+            usd_camera = camera
+        elif hasattr(camera, "IsA"):
+            usd_camera = UsdGeom.Camera(camera)
+        else:
+            raise TypeError("Expected a UsdGeom.Camera or Usd.Prim.")
+
+        prim = usd_camera.GetPrim()
+        if not prim.IsValid() or not prim.IsA(UsdGeom.Camera):
+            raise TypeError(f"Expected a UsdGeom.Camera prim, got {prim.GetPath()!r}.")
```
eric-heiden: 🟠 When `prim` is invalid, this branch tries to include `prim.GetPath()` in the error message. Accessing the path of a null prim raises `RuntimeError`, so callers never receive the intended validation error. Please handle `not prim.IsValid()` separately without accessing the path, then use `GetPath()` only after validity has been established.

CHANGELOG.md
daniela-hase: Addressed in https://github.com/newton-physics/newton/pull/3026/changes/fe1d72fa82fa2d0c9bb0a3b26d43f92ef8dfc89a

newton/_src/sensors/warp_raytrace/camera_utils.py
daniela-hase: Addressed in https://github.com/newton-physics/newton/pull/3026/changes/fe1d72fa82fa2d0c9bb0a3b26d43f92ef8dfc89a

newton/_src/sensors/warp_raytrace/utils.py
daniela-hase: Addressed in https://github.com/newton-physics/newton/pull/3026/changes/fe1d72fa82fa2d0c9bb0a3b26d43f92ef8dfc89a

newton/_src/sensors/warp_raytrace/camera_utils.py
daniela-hase: Addressed in https://github.com/newton-physics/newton/pull/3026/changes/fe1d72fa82fa2d0c9bb0a3b26d43f92ef8dfc89a

newton/_src/sensors/warp_raytrace/utils.py
daniela-hase: Similar to my comment above about the same issue - not sure what this is about. It seems all fine and correct to me 🤔

newton/_src/sensors/warp_raytrace/utils.py
eric-heiden: Hmm this is not the right format, we typically write `[m]` or `[rad]` explicitly. Now the units are referring to other arguments but not of them spell out explicitly what the unit is.

eric-heiden: See the ModelBuilder method docstrings for reference.

newton/_src/sensors/warp_raytrace/utils.py
eric-heiden: Sorry this comment got messed up. I think what I meant originally here is that we could make `camera_fovs` a keyword-only argument, so move the `*` in front of this argument if you think that makes sense?

eric-heiden: Could you also document what happens when these arguments are left `None`, i.e. what is the default behavior/value here?

daniela-hase: Changed in https://github.com/newton-physics/newton/pull/3026/changes/7a0cdc57a1eced92ad7837c7f1c196dd51ed28c7

newton/_src/sensors/warp_raytrace/utils.py
daniela-hase: The docstring says:

```
            focal_length: Focal length [mm or any consistent unit]. If
                ``None`` and the other aperture parameters are also ``None``,
                uses *camera_fovs*.
```
Is this not correct?

PR #3030, opened 2026-06-02
Fix #2492: MJCF body euler convention + <compiler> include-merge

### LINKED ISSUE #2492: Aloha MJCF import has multiple issues (dof_damping, equality constraints, ngeom)
The Aloha bimanual robot has multiple MJCF import issues that prevent dynamics and FK tests from passing.

## Known issues

- **dof_damping**: values differ from native (possibly due to nested default class inheritance not being resolved fully)
- **eq_ / neq**: equality constraints (finger joint coupling) not fully imported
- **ngeom**: Newton creates 71 geoms vs native's 69 (visual geom handling)

## Symptoms

- Model comparison requires skipping `dof_damping`, `eq_`, `neq`, `ngeom`, and `jnt_` fields
- FK test fails with xpos diff of 0.14m
- Dynamics test not enabled (blocked by above issues)

## Context

Discovered during Epic #1401 menagerie testing. Aloha has unique characteristics (bimanual, finger coupling) that expose multiple import paths that work fine for simpler robots.

Test is disabled via `num_steps = 0` in `TestMenagerie_Aloha` with broad `model_skip_fields`.

vreutskyy:
## Description

Closes most of #2492.

Two interacting MJCF import bugs prevented Aloha (and any model that chains `<include>`s past a `<compiler>` element) from importing correctly:

1. **`quat_from_euler_mjcf` used extrinsic rotation** where MuJoCo's default `eulerseq=xyz` is intrinsic. The two agreed when at most one input angle was non-zero, so the bug stayed invisible until a body like Aloha's `gripper_base` used `euler="0 1.57 -1.57"`.
2. **`parse_mjcf` only read the first `<compiler>` element** via `root.find("compiler")`. After `<include>` expansion the root has one compiler per source file; MuJoCo merges them globally with later overriding earlier. Aloha loads via `scene.xml` (`<compiler/>`, defaults `angle="degree"`) and includes `aloha.xml` (`<compiler angle="radian"/>`). Newton's first-only read treated aloha's radian-authored eulers as degrees, dwarfing the convention bug.

Both fixes verified empirically against MuJoCo (regression tests included). The two layouts — inner-compiler-after-outer and outer-compiler-after-inner — are both checked: in each case, the latest-in-document-order compiler wins for ALL bodies, regardless of which file authored them. Newton now matches.

Also enables `TestMenagerie_Aloha` FK and dynamics: same `_align_models` integrator-pin pattern as FR3v2/Cassie (#2970); tolerance backed by a 15-trial native-vs-native characterization (0 variance under IMPLICITFAST, newton-vs-native max 1.43e-6).

The remaining `geom_group` flattening (Newton imports 2 extra type-7 geoms vs native and collapses all to `group=0`) is a separate concern left in #2492 as the only follow-up.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k TestMenagerie_Aloha
```

All `test_import_mjcf` cases pass (213 incl. 2 new regression tests for body `euler` and compiler-merge across includes). `TestMenagerie_Aloha` runs FK and dynamics under the integrator pin.

## Bug fix

1. Temporarily flip `num_steps = 0 → 20` and `fk_enabled = False → True` on `TestMenagerie_Aloha`.
2. Run `uv run --extra dev -m newton.tests -k TestMenagerie_Aloha`.
3. Observe: FK fails on the four finger bodies (xpos diff ~5e-2 m). Their parent `gripper_base` body has `euler="0 1.57 -1.57"` in MJCF; Newton imports as near-identity quaternion because the included `<compiler angle="radian"/>` is lost and the eulers are then treated as degrees.

**Fix:** `quat_from_euler_mjcf` rewritten to compose intrinsic rotations; `parse_mjcf` merges all `<compiler>` elements (later overrides earlier).

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected MJCF Euler-angle conversion to match MuJoCo intrinsic rotation semantics.
  * Fixed merging of MJCF compiler attributes across included files (document-order, later wins).
  * Ensured static worldbody geoms honor visual/collider filtering so visuals can be dropped as configured.

* **Tests**
  * Added regression tests validating Euler conversion and compiler-merge behavior against MuJoCo.
  * Enabled dynamics/forward-kinematics in compatibility tests and tightened comparison tolerances.

### DIFF
--- a/newton/_src/utils/import_mjcf.py
+++ b/newton/_src/utils/import_mjcf.py
@@ -2353,14 +2353,15 @@ def parse_mjcf(
             )
 
         # -----------------
-        # add static geoms
+        # add static geoms — partition by class so `parse_visuals=False` /
+        # `parse_visuals_as_colliders=True` apply uniformly to worldbody
+        # geoms too (not just geoms inside bodies).
 
-        parse_shapes(
+        _process_body_geoms(
+            geoms=world.findall("geom"),
             defaults=world_defaults,
             body_name="world",
             link=-1,
-            geoms=world.findall("geom"),
-            density=default_shape_density,
             incoming_xform=xform,
             label_prefix=root_label_path,
         )
--- a/newton/_src/utils/import_mjcf.py
+++ b/newton/_src/utils/import_mjcf.py
@@ -327,7 +327,7 @@ def parse_mjcf(
     mjcf_dirname = base_dir or "."  # Backward compatible fallback for mesh paths
 
     use_degrees = True  # angles are in degrees by default
-    euler_seq = [0, 1, 2]  # XYZ by default
+    eulerseq = "xyz"  # default sequence (lowercase = body-frame axes per MuJoCo's actual behavior)
 
     # load joint defaults
     default_joint_limit_lower = builder.default_joint_cfg.limit_lower
@@ -374,11 +374,14 @@ def parse_mjcf(
         compiler_attribs.update(c.attrib)
     if compiler_attribs:
         use_degrees = compiler_attribs.get("angle", "degree").lower() == "degree"
-        euler_seq = ["xyz".index(c) for c in compiler_attribs.get("eulerseq", "xyz").lower()]
+        # Per-character case carries the intrinsic/extrinsic axis convention
+        # (lowercase = body-frame axis, uppercase = world-frame axis); keep it.
+        eulerseq = compiler_attribs.get("eulerseq", "xyz")
         mesh_dir = compiler_attribs.get("meshdir", ".")
         texture_dir = compiler_attribs.get("texturedir", mesh_dir)
         fitaabb = compiler_attribs.get("fitaabb", "false").lower() == "true"
     else:
+        eulerseq = "xyz"
         mesh_dir = "."
         texture_dir = "."
         fitaabb = False
@@ -578,20 +581,17 @@ def parse_mjcf(
 
         return wp.types.vector(length, wp.float32)(out)
 
-    def quat_from_euler_mjcf(e: wp.vec3, i: int, j: int, k: int) -> wp.quat:
-        """Convert MJCF euler to intrinsic-rotation quaternion (matches MuJoCo).
+    def quat_from_euler_mjcf(e: wp.vec3, seq: str) -> wp.quat:
+        """Convert MJCF euler to quaternion respecting per-character ``eulerseq`` case.
 
-        Composes ``q_axis[i](e[0]) * q_axis[j](e[1]) * q_axis[k](e[2])``.
+        For each character, lowercase composes around the body-frame axis
+        (right-multiply) and uppercase composes around the world-frame axis
+        (left-multiply). The default ``"xyz"`` yields ``qx*qy*qz``; ``"XYZ"``
+        yields ``qz*qy*qx``; mixed cases yield the corresponding hybrid.
         """
         half = np.asarray([float(e[0]), float(e[1]), float(e[2])]) * 0.5
         c = np.cos(half)
         s = np.sin(half)
-
-        def axis_quat(axis_idx: int, idx: int) -> np.ndarray:
-            q = np.zeros(4)
-            q[0] = c[idx]
-            q[1 + axis_idx] = s[idx]
-            return q
 
         def qmul(a: np.ndarray, b: np.ndarray) -> np.ndarray:
             aw, ax, ay, az = a
@@ -605,7 +605,14 @@ def parse_mjcf(
                 ]
             )
 
-        result = qmul(qmul(axis_quat(i, 0), axis_quat(j, 1)), axis_quat(k, 2))
+        result = np.array([1.0, 0.0, 0.0, 0.0])  # identity (w, x, y, z)
+        for n, ch in enumerate(seq):
+            axis_idx = "xyz".index(ch.lower())
+            q = np.zeros(4)
+            q[0] = c[n]
+            q[1 + axis_idx] = s[n]
+            result = qmul(result, q) if ch.islower() else qmul(q, result)
+
         return wp.quat(float(result[1]), float(result[2]), float(result[3]), float(result[0]))
 
     def parse_orientation(attrib) -> wp.quat:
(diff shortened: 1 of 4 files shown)


newton/_src/utils/import_mjcf.py
```diff
@@ -368,13 +368,16 @@ def parse_mjcf(
...
-        mesh_dir = compiler.attrib.get("meshdir", ".")
-        texture_dir = compiler.attrib.get("texturedir", mesh_dir)
-        fitaabb = compiler.attrib.get("fitaabb", "false").lower() == "true"
+    # Merge all <compiler> elements (document order, later wins) — matches MuJoCo.
+    compiler_attribs: dict[str, str] = {}
+    for c in root.iter("compiler"):
+        compiler_attribs.update(c.attrib)
+    if compiler_attribs:
+        use_degrees = compiler_attribs.get("angle", "degree").lower() == "degree"
+        euler_seq = ["xyz".index(c) for c in compiler_attribs.get("eulerseq", "xyz").lower()]
```
camevor: `compiler_attribs.get("eulerseq", "xyz").lower()` discards the information about whether an extrinsic or intrinsic rotation is requested. From the [MuJoCo reference](https://mujoco.readthedocs.io/en/stable/XMLreference.html#compiler-eulerseq):
> This must be a string with exactly 3 characters from the set {x, y, z, X, Y, Z}. The character at position n determines the axis around which the n-th rotation is performed. Lower case letters denote axes that rotate with the frame (intrinsic), while upper case letters denote axes that remain fixed in the parent frame (extrinsic).

While this PR corrects the default to being intrinsic, the distinction from extrinsic frames should be honored. If this remains unsupported, it should raise on extrinsic axes.

newton/_src/utils/import_mjcf.py
```diff
@@ -2338,14 +2353,15 @@ def add_converted_loop_joint(
...
 
         # -----------------
-        # add static geoms
+        # add static geoms — partition by class so `parse_visuals=False` /
+        # `parse_visuals_as_colliders=True` apply uniformly to worldbody
+        # geoms too (not just geoms inside bodies).
 
-        parse_shapes(
+        _process_body_geoms(
+            geoms=world.findall("geom"),
```
camevor, line 2370: This change makes sense, but highlights an issue in `_process_body_geoms`:
Geoms with `contype == 0` and `conaffinity == 0` are treated as visual-only (`ShapeFlags.COLLIDE_SHAPES = False`), preventing them from colliding even when an explicit pair is defined:
```xml
<contact>
  <pair geom1="floor" geom2="ball"/>
</contact>
```

newton/tests/test_import_mjcf.py
```diff
@@ -311,6 +311,122 @@ def test_site_euler_sequence_matches_mujoco(self):
...
+    def test_compiler_merge_across_includes(self):
+        """``<compiler>`` attributes merge globally across ``<include>``-expanded
+        files (document order, later wins, scope is not file-local).
+
+        Both layouts (inner-compiler-after-outer and outer-compiler-after-inner)
+        check that the latest compiler wins for ALL bodies regardless of which
+        file authored them.
+        """
+        import os  # noqa: PLC0415
+        import tempfile  # noqa: PLC0415
```
camevor: Redundant imports

newton/_src/utils/import_mjcf.py
vreutskyy: Good catch — applied in [af16f78b](https://github.com/newton-physics/newton/pull/3030/commits/af16f78b). `quat_from_euler_mjcf` now walks the raw sequence and dispatches per character (lowercase → right-multiply / intrinsic, uppercase → left-multiply / extrinsic). The body-euler regression test now sweeps all 6³ = 216 combinations of {x,y,z,X,Y,Z} and diffs against MuJoCo's body_quat (every one accepted; every one matches at float32 precision).

FYI I empirically confirmed your reading of the docs: lowercase = intrinsic axis (rotates with the body), uppercase = extrinsic axis (fixed in the parent frame). Earlier in the investigation I had pulled what looked like an older/different snippet that flipped those meanings, but the current XMLreference text matches the actual behavior.

newton/_src/utils/import_mjcf.py
vreutskyy: Good observation. The `contype==0 && conaffinity==0 → visual-only` classification is pre-existing in `_process_body_geoms`; this PR only extends it to worldbody static geoms (so the issue you describe now affects worldbody geoms too if they're targeted by a `<contact><pair>` rule). Treating the existing-explicit-pair case correctly is a separate concern that I'd rather not bolt onto this PR. Filed as #2492's last remaining follow-up — I'll add it to the issue description so it doesn't fall through the cracks.

newton/tests/test_import_mjcf.py
vreutskyy: Removed in [0e8b4a77](https://github.com/newton-physics/newton/pull/3030/commits/0e8b4a77). `os` and `tempfile` are already imported at the top of the file.

newton/_src/utils/import_mjcf.py
```diff
@@ -368,14 +368,20 @@ def parse_mjcf(
...
-        texture_dir = compiler.attrib.get("texturedir", mesh_dir)
-        fitaabb = compiler.attrib.get("fitaabb", "false").lower() == "true"
+    # Merge all <compiler> elements (document order, later wins) — matches MuJoCo.
+    compiler_attribs: dict[str, str] = {}
+    for c in root.iter("compiler"):
+        compiler_attribs.update(c.attrib)
+    if compiler_attribs:
+        use_degrees = compiler_attribs.get("angle", "degree").lower() == "degree"
+        # Per-character case carries the intrinsic/extrinsic axis convention
+        # (lowercase = body-frame axis, uppercase = world-frame axis); keep it.
```
camevor: The implementation appears to preserve the intrinsic/extrinsic distinction, but this comment says uppercase axes are “world-frame” axes. MuJoCo specifies uppercase axes as fixed in the parent frame.

newton/_src/utils/import_mjcf.py
camevor: Thanks, looks good! Only includes one misleading comment, see below.

newton/_src/utils/import_mjcf.py
vreutskyy: Fixed in [97a139ea](https://github.com/newton-physics/newton/pull/3030/commits/97a139ea) — reworded all three comments to say "fixed in the parent frame" (extrinsic) instead of "world-frame", matching MuJoCo's wording. The implementation was already correct; only the comments were imprecise.

camevor: Thanks!

camevor (approved):

PR #3039, opened 2026-06-03
API Cleanup

eric-heiden:
## Description

This PR moves BVH lifecycle management from module-level geometry functions onto `Model`, so the acceleration structures owned by a model are built and refit through that model's API. `ModelBuilder.finalize()` now also builds the initial shape and particle BVHs automatically from the finalized model's initial `body_q` / `particle_q` data when supported geometry or particles are present.

The old flow required callers to reach into `newton.geometry` for global helpers:

```python
newton.geometry.build_bvh_shape(model, state)
newton.geometry.refit_bvh_shape(model, state)
newton.geometry.build_bvh_particle(model, state)
newton.geometry.refit_bvh_particle(model, state)
```

The new `Model` methods are:

```python
model.bvh_build_shapes(state, *, bvh_constructor: str | None = None)
model.bvh_refit_shapes(state)
model.bvh_build_particles(state, *, bvh_constructor: str | None = None)
model.bvh_refit_particles(state)
```

For models created with `ModelBuilder.finalize()`, callers no longer need to build BVHs before the first BVH-backed ray query or sensor render. After body or particle state changes, callers should still explicitly refit the BVHs with `model.bvh_refit_shapes(state)` and/or `model.bvh_refit_particles(state)`. The explicit build methods remain available for manually populated models, structural changes, and custom `bvh_constructor` choices.

The implementation moves the shape and particle BVH build/refit logic into `Model` methods while leaving the low-level Warp launch helpers in `newton._src.geometry.bvh`. The previous `newton.geometry.build_bvh_*` and `newton.geometry.refit_bvh_*` functions remain available as deprecated aliases that forward to the corresponding `Model` methods, preserving compatibility while steering users toward the new API.

This also adds `Model.heightfield_count` as explicit model metadata and deprecates the boolean `Model.has_heightfields` alias. Call sites that only need a boolean now use `heightfield_count > 0`, while tests cover the new count behavior and compatibility path.

Updated areas include tiled camera setup, raycast/BVH error messaging, sensor render context BVH construction, heightfield metadata construction, examples, benchmarks, tests, recorder serialization of derived BVH data, and the changelog entries for the public API changes.

## Checklist

## Test plan

```bash
uvx pre-commit run -a
```

## New feature / API change

```python
import newton

# ... populate builder ...
model = builder.finalize()

# finalize() builds the initial BVHs automatically when applicable.

# Refit after simulation state changes.
model.bvh_refit_shapes(state)
model.bvh_refit_particles(state)

# Rebuild explicitly for manually populated models, structural changes,
# or custom construction algorithms.
model.bvh_build_shapes(state, bvh_constructor="sah")
model.bvh_build_particles(state)
```

## Summary by CodeRabbit

* **New Features**
  * Model adds built-in BVH build/refit APIs for shapes and particles; Model.heightfield_count exposes heightfield support; new color-space utilities and camera output color-space option; USD import recognizes additional mass/material attributes.

* **Deprecations**
  * Legacy geometry-level BVH helpers and boolean has_heightfields deprecated in favor of Model APIs and heightfield_count.

* **Documentation**
  * Changelog and sensor/camera docs updated to reference Model-scoped BVH workflows and finalize behavior.

* **Tests**
  * Added/updated tests for heightfield aliasing, BVH build/refit behavior, and sensor BVH usage.


adenzler-nvidia (dismissed):
Approving, conditional on this being the API direction agreed between the author and @mmacklin (moving BVH lifecycle off `newton.geometry.*` onto `Model`, and replacing the boolean `Model.has_heightfields` with `Model.heightfield_count`).

The deprecations themselves are all correct:

- The four `newton.geometry.build_bvh_*` / `refit_bvh_*` functions stay re-exported from the public module and forward to the new `Model` methods, emit `DeprecationWarning` (stacklevel=2), preserve the old positional signature, and `build_*` correctly forwards the new `bvh_constructor` kwarg while `refit_*` omits it.
- `Model.has_heightfields` becomes a deprecated getter/setter over `heightfield_count`; no internal code reads it anymore (builder sets `heightfield_count`, collide uses `heightfield_count > 0`), so nothing trips its own warning.
- `.. deprecated:: 1.3` matches the `1.3.0.dev0` version. The moved method bodies are line-for-line identical to the originals aside from `self` and the added `constructor=` kwarg, and CHANGELOG/cross-refs/error messages are consistent.

Non-blocking suggestions for a follow-up:

- The deprecated-particle-alias test runs on a zero-particle scene, so it only exercises the no-op early return and asserts the warning rather than that the alias actually forwards and builds a particle BVH. Worth running it on a particle-bearing model (and passing `bvh_constructor` through the particle alias) to match the shape-path assertion. The refit-before-build `RuntimeError` guards and the `has_heightfields` false branches are also currently untested.
- Minor docstring nit: `bvh_shape_world_transforms` still says "computed during shape BVH refit" while its siblings were updated to "build"; it's populated during both, so "build/refit" reads more consistently.

mmacklin:
Thanks Eric, regarding naming:

Instead of:

```
model.build_bvh_shape(state, bvh_constructor=sah)
model.refit_bvh_shape(state)
model.build_bvh_particle(state)
model.refit_bvh_particle(state)
```

I think we should consider:

```
model.bvh_refit_shapes(state, bvh_constructor=sah)
model.bvh_refit_particles(state)
model.bvh_build_shapes(state)
model.bvh_build_particles(state)
```

Also should we just always build the BVH in finalize? I feel like users shouldn't have to remember to call this so either do it lazily in some bvh property accessor or just always do it in finalize().

eric-heiden:
Thanks @mmacklin, I've renamed the methods now to share the common prefix.

> Also should we just always build the BVH in finalize? I feel like users shouldn't have to remember to call this so either do it lazily in some bvh property accessor or just always do it in finalize().

`SensorTiledCamera` handles this all automatically (building + refitting the BVH during `sensor.update()`), but `newton.intersect_ray()` doesn't do any of that. We could also simply add the `state` as input to `newton.intersect_ray()` to also perform this kind of build + refit step on every call? Even with an optional flag like `refit_bvh: bool = True` so the user has a way to skip the refitting if it was done already for the current shape transforms.

Edit: I've now added the BVH building to the `ModelBuilder.finalize()` step which adds only minimal overhead.

mmacklin (approved):

PR #3041, opened 2026-06-03
Cap newton-actuators to <0.1.1

adenzler-nvidia:
`newton-actuators` 0.1.1 emits a `DeprecationWarning` at import time. The example test suite runs example subprocesses with `PYTHONWARNINGS=error::DeprecationWarning`, so that warning becomes a fatal import error and every non-opted-out example test fails on a fresh install.

CI doesn't catch this because it resolves from the lockfile, which pins 0.1.0; only fresh resolutions pick up 0.1.1. This caps the dependency (deprecated, kept only for backward compatibility) below 0.1.1 so fresh installs match the already-locked behavior — the lock already resolves to 0.1.0, so this only tightens the published constraint.

## Summary by CodeRabbit

* **Chores**
  * Updated a dependency version constraint to ensure improved compatibility and stability.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ sim = [
     # MuJoCo simulation
     "mujoco-warp>=3.8.0.3,~=3.8.0",  # MuJoCo, warp-accelerated (3.8.0 was broken)
     "mujoco~=3.8.0",        # MuJoCo physics engine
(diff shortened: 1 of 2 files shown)


jvonmuralt (approved):

jcarius-nv:
you'll cherry-pick this to 1.2.1 as well, right?

jcarius-nv (approved):

PR #3044, opened 2026-06-03
Add an opt-in --strict-warnings flag to the test runner

adenzler-nvidia:
## Problem

`newton.tests` ships inside the package, so consumers run `python -m newton.tests` to verify their installation. But `test_examples.py` always launched example subprocesses with `PYTHONWARNINGS=error::DeprecationWarning`, so any deprecation — including one from a dependency the user cannot act on — turned the suite red. A working install then looks broken.

## Change

Add a `--strict-warnings` flag to the test runner, **off by default**. When set, the suite treats the warnings a developer can act on as errors:

- **all `DeprecationWarning`s** (from Newton or its dependencies), and
- **any warning of any category attributed to a `newton.*` module**.

Dependency-owned *non-deprecation* warnings (e.g. a `UserWarning` from `usd`/`torch`) stay warnings — a consumer cannot act on them, so they should not fail an install check. When the flag is unset the whole suite is lenient.

The in-process worker applies both filters directly. Example subprocesses get the deprecation axis via `-W error::DeprecationWarning`; `newton.examples` now defers to any explicit `-W`/`PYTHONWARNINGS` policy (via `sys.warnoptions`) instead of installing its own lenient `default` filter for `newton.*` deprecations (this also fixes that shadowing for end users running `python -W error::DeprecationWarning -m newton.examples.<name>`). The policy reaches test modules through `newton.tests.unittest_utils.strict_warnings`, mirroring `coverage_enabled`.

## Dependencies

Broadening strict mode surfaces pre-existing warnings in the in-process tests, fixed in separate PRs that should merge first:

- #3058 — capture expected warnings in inertia-validation tests
- #3059 — migrate tests off deprecated ellipsoid/SDF aliases
- #3060 — drop a false-positive solref-limit warning
- #3061 — handle expected warnings in mujoco/softbody/usd tests

(#3056, the joint-target/collider migration, has already merged.) Until these land and this branch is rebased, CI here is expected to be red.

## CI

Because this flips the default, the CI invocations opt in via `--strict-warnings` so warning debt keeps surfacing: `ci.yml`, `aws_gpu_tests.yml`, `minimum_deps_tests.yml`, `mujoco_warp_tests.yml`, `warp_nightly_tests.yml`. A plain `python -m newton.tests` is lenient.

## Release Notes

* **Tests**
  * Enhanced deprecation warning handling in test suite—warnings are now treated as fatal errors only in CI environments when strict-warnings mode is enabled
  * Improved subprocess test execution for better warning policy enforcement

* **Chores**
  * Updated CI/CD pipelines to enforce strict warning handling during automated test runs


shi-eric:
It seems like this branch picked up a bunch of unrelated changes?

adenzler-nvidia:
Not unrelated — they're a prerequisite for this PR. It turns on `--deprecations-as-errors` in CI, which escalates `DeprecationWarning` to a hard failure. `joint_target_pos`/`joint_target_vel` and the old collider velocity-mode strings are deprecated aliases now, so any test still using them would fail under the strict flag. The "noisy" diff is the migration off those aliases — without it, flipping CI to strict just turns the suite red.

The branch keeps infra and fixes separate:

- `ad9950733` — add the `--deprecations-as-errors` flag (off by default), wired through the worker/example paths
- `cd07e5761` — docstring follow-up
- `e41df1d81` — enable the flag in CI (`.github/` workflow YAMLs only)
- `e05c46970` — gate `test_lazy_init` on the flag instead of hard-coding strict
- `9f78f21d2` — migrate the 8 test files off the deprecated aliases to the canonical `joint_target_q`/`joint_target_qd` names

So commits 1–4 are the mechanism (none touch test bodies) and commit 5 is the entire alias fallout in one isolated, find-and-replace commit. Happy to pull `9f78f21d2` into its own PR that merges first if you'd prefer the strict-CI flip to land on an already-clean suite.

shi-eric:
Since commits are squashed on merge, I think the test migration changes should be merged first, in a separate pull request to keep this PR more focused.

adenzler-nvidia:
fair point - let me do that.

newton/tests/test_examples.py
```diff
@@ -103,20 +104,26 @@ def run(test, device):
...
         if warp_cache_path is not None:
             env_vars["WARP_CACHE_PATH"] = os.path.dirname(warp_cache_path)
-        if not allow_deprecation_warnings:
-            env_vars["PYTHONWARNINGS"] = "error::DeprecationWarning"
-        else:
-            env_vars.pop("PYTHONWARNINGS", None)
+        # Govern the subprocess's deprecation policy solely through warning_args
+        # below; drop any ambient PYTHONWARNINGS so a stray error::DeprecationWarning
+        # in the caller's environment cannot turn a lenient run strict.
+        env_vars.pop("PYTHONWARNINGS", None)
```
shi-eric, line 126: This should preserve/set `PYTHONWARNINGS` for strict example runs instead of removing it and relying only on `-W`.

The example subprocess imports `newton.examples`, whose `_enable_example_deprecation_warnings()` helper only skips installing its own `default` deprecation filter when `PYTHONWARNINGS` is present. With this env var removed, the helper adds a `default` filter for `newton.*` ahead of the interpreter's `error` filter, so Newton-attributed `DeprecationWarning`s in examples are printed but do not fail the subprocess.

I reproduced this locally: `python -W error::DeprecationWarning` still survived a warning emitted from a `newton.*` module after the example helper ran, while the same warning raised once `PYTHONWARNINGS=error::DeprecationWarning` was present.

newton/tests/thirdparty/unittest_parallel.py
```diff
@@ -93,6 +94,13 @@ def main(argv=None):
     parser.add_argument(
         "--junit-report-xml", metavar="FILE", help="Generate JUnit report format XML file"
     )  # NVIDIA Modification
+    parser.add_argument(
+        "--deprecations-as-errors",
```
shi-eric: Since the name is already long, what about `--deprecation-warnings-as-errors` ?

adenzler-nvidia: Good call on the length. The flag now escalates more than deprecations — it also turns any warning attributed to a `newton.*` module into an error — so I went with `--strict-warnings`: shorter than `--deprecation-warnings-as-errors` and accurate for the broadened scope. Done in 29178af6e.

newton/tests/test_examples.py
adenzler-nvidia: Fixed in 7906c06c6 — thanks for catching this. Root cause: `newton.examples._enable_example_deprecation_warnings()` only deferred when `PYTHONWARNINGS` was set, so a plain `-W error::DeprecationWarning` was shadowed by the helper's own `default` filter (installed after startup, it took precedence). It now defers on `sys.warnoptions`, which covers both `-W` and `PYTHONWARNINGS`, and the runner drives strict example runs with `-W`. This also fixes the same shadowing for anyone running `python -W error::DeprecationWarning -m newton.examples.<name>` directly.

shi-eric:
It seems like the enablement of treating all warnings as failures makes the unit test suite even stricter than before. What is your current plan to get this in a mergeable state?

adenzler-nvidia:
yes, I have separate PRs for all the issues in review: #3061 #3058 #3059 #3060 so this should become green after these are merged.

shi-eric:
Non-blocking suggestion: `--strict-warnings` currently installs the warning filters after test discovery, so import-time `DeprecationWarning`s from discovered test modules are not covered. If we want the flag to mean “strict for the whole test run,” we could install the same policy before `TestLoader.discover()` and in the worker initializer before any test module imports.

I’d treat this as a minor follow-up rather than a blocker, since it only affects import-time warnings.

newton/tests/test_lazy_init.py
```diff
@@ -11,8 +11,15 @@
 
 class TestLazyInit(unittest.TestCase):
     def test_import_newton_does_not_init_warp(self):
+        import newton.tests.unittest_utils  # noqa: PLC0415
```
shi-eric: Is this deferred import strictly necessary? Just want to make sure we're not abusing the use of PLC0415 exclusions.

shi-eric (dismissed):

shi-eric (dismissed):

nvtw (approved):

shi-eric (approved):

PR #3045, opened 2026-06-03
Add USD root scale option to importer

Viru97:
## Description

Add a `scale` parameter to USD import so callers can scale the imported USD root before parsing physics bodies, joints, and shapes. This enables loading USD assets at a different size without manually editing or exporting a scaled USD file first.

## Checklist

## Test plan

```bash
python3 -m py_compile newton/_src/utils/import_usd.py newton/_src/sim/builder.py newton/tests/test_import_usd.py
git diff --check
```
## New feature/API

```bash
import newton

builder.add_usd(
    "robot.usda",
    scale=(0.05, 0.05, 0.05),
)

## Summary by CodeRabbit

* **New Features**
  * Added a `scale` parameter to USD import to scale the selected USD root prim using three axis values.

* **Bug Fixes**
  * Improved MuJoCo geometry property handling by normalizing geometry data indexing for consistent runtime processing.

* **Tests**
  * Added coverage to verify root scaling affects imported body transforms and collision shape scales.

### CODE REVIEW

**Comment by adenzler-nvidia** [comment]
Thanks for the contribution! A few points after review.

**Unrelated `solver_mujoco.py` change:** the `geom_dataid` normalization can't trigger on a supported install. `update_geom_properties_kernel` has taken a 2D `geom_dataid` since the mujoco-warp 3.7 bump (#2455), and the pinned mujoco-warp (`>=3.10.0.2`) always provides a batched `(*, ngeom)` array. If you saw a 1D array locally, that points to a stale `mujoco_warp` in your environment rather than a bug here. Please drop this change — and in general, PRs are easier to review when filed from a dedicated feature branch rather than your fork's `main`.

**On the `scale` parameter:** we don't want to take this in its current form. The implementation writes `xformOp:scale` onto the root prim of the input stage, which:

- permanently mutates a caller-provided `Usd.Stage` as a silent side effect,
- overwrites any authored root scale instead of composing with it (and if the attribute exists but isn't in `xformOpOrder`, does nothing),
- with `root_path="/"` and no `defaultPrim`, applies the scale to an arbitrary first root prim (possibly the `PhysicsScene`, which isn't Xformable) and leaves sibling root prims unscaled,
- leaves authored masses/inertias, linear joint limits, and drive gains unscaled, so the imported model is physically inconsistent.

Since USD already provides a native, composable way to do this, the parameter doesn't add capability over two lines of user code before the import:

```python
UsdGeom.Xformable(stage.GetPrimAtPath("/World")).AddScaleOp().Set(Gf.Vec3d(0.05))
builder.add_usd(stage)
```

(or a session-layer override, or `metersPerUnit` if the goal is unit conversion).

If import-time scaling matters to you for parity with `add_urdf`/`add_mjcf` (which take `scale: float`), the right shape would be a uniform float applied inside the import pipeline — with an explicit decision on how masses and limits rescale — rather than editing the stage. Happy to discuss that in an issue first.

newton/_src/sim/builder.py
```diff
@@ -2711,6 +2711,7 @@ def add_usd(
         force_position_velocity_actuation: bool = False,
         convert_mjc_equality_constraints: bool = True,
         override_root_xform: bool = False,
+        scale: Sequence[float] | None = None,
```
andrewkaufman, line 2714: Followup on the coderabbit comment, why are we supporting a Sequence[float] here? Do you really need non-uniform scale or can this just be a scalar instead like in add_urdf/add_mjcf?

jcarius-nv:
Closing due to inactivity

PR #3053, opened 2026-06-03
Add opacity support to shapes and viewers

eric-heiden:
## Description

Add display opacity for rigid shapes, cloth, and soft bodies across Newton's builders, importers, recorders, and rendering viewers.

- Add `opacity=` to rigid-shape construction, with `Mesh.opacity` and `Model.shape_opacity` carrying authored values.
- Add `Model.tri_color` and `Model.tri_opacity` for cloth and soft-body surface triangles. Multi-triangle builder helpers accept either one broadcast color/opacity or per-triangle arrays.
- Preserve authored appearance from USD, URDF, and MJCF assets.
- Support mesh and instance opacity in ViewerGL, ViewerUSD, ViewerRTX, ViewerRerun, and ViewerViser; ViewerFile records it and ViewerNull accepts the common API.
- Render ViewerGL transparency through weighted-blended OIT while keeping opaque geometry in the resolved MSAA, depth, and shadow passes.
- Honor `backface_culling=False` in USD/RTX output by authoring double-sided geometry.
- Use opacity in the RJ45, Panda, cable-table, MPM viscous, and softbody-Franka examples.
- Add checker coloring to the cloth rollers and stiff-material stretch examples, orange stripes to the rollers, and shape-palette colors to the poker cards.

Transparent RJ45 example
Transparent Panda example
Transparent MPM example
Transparent soft-body example

## Checklist

## Test plan

- `uv run --extra dev -m newton.tests -k test_shape_colors` (39 tests)
- `uv run -m newton.examples cloth_stiff_material_stretch --viewer null --test --num-frames 360`
- `uv run -m newton.examples cloth_rollers --viewer null --num-frames 1`
- `uv run -m newton.examples cloth_poker_cards --viewer null --test --num-frames 30`
- `uv run --frozen --extra dev -m newton.tests -k test_viewer_get_frame -k test_viewer_log_shapes -k test_viewer_geometry_batching -k test_viewer_layers -k test_viewer_usd -k test_viewer_rerun_hidden -k test_recorder -k test_import_urdf -k test_import_mjcf -k test_import_usd` (736 tests, 1 skipped)
- `uv run --frozen docs/generate_api.py`
- `uvx pre-commit run -a`
- Interactive ViewerGL validation of the cloth rollers, stiff-material stretch, and poker-card examples
- Headless ViewerGL render of a transparent shape through the OIT accumulation and resolve passes

## New feature / API change

```python
import newton

builder.add_cloth_grid(
    pos=(0.0, 0.0, 0.0),
    rot=(0.0, 0.0, 0.0, 1.0),
    vel=(0.0, 0.0, 0.0),
    dim_x=1,
    dim_y=1,
    cell_x=0.1,
    cell_y=0.1,
    mass=1.0,
    color=[(0.1, 0.2, 0.8), (0.9, 0.9, 0.9)],
    opacity=[0.5, 1.0],
)
model = builder.finalize()
```


newton/_src/viewer/viewer_usd.py
```diff
@@ -639,6 +670,21 @@ def _promote_colors_to_array(self, colors, num_items):
...
+        if isinstance(opacities, wp.array):
+            opacities = opacities.numpy()
+        elif isinstance(opacities, list | tuple) and len(opacities) == 1 and np.isscalar(opacities[0]):
+            opacities = np.tile(float(opacities[0]), num_items)
+        elif np.isscalar(opacities):
+            opacities = np.tile(float(opacities), num_items)
+        opacities_np = np.asarray(opacities, dtype=np.float32).reshape(-1)
+        if len(opacities_np) == 1 and num_items > 1:
+            opacities_np = np.tile(float(opacities_np[0]), num_items)
+        return np.clip(opacities_np, 0.0, 1.0)
```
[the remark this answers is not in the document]
eric-heiden: Fixed in 8cc477fda: USD opacity promotion now rejects arrays that are not scalar/length-1 or exactly the instance count, and regular instance logging uses the same validator. Added a regression test for mismatched opacity counts.

newton/_src/viewer/viewer_file.py
```diff
@@ -576,6 +576,8 @@ def callback(x, path):
                 "is_solid": x.is_solid,
                 "has_inertia": x.has_inertia,
                 "maxhullvert": x.maxhullvert,
+                "color": x.color,
+                "opacity": x.opacity,
```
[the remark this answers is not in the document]
eric-heiden: Fixed in 8cc477fda: recorder mesh dedup now keys on the Mesh object, so different Mesh instances that share buffers but differ in color/opacity serialize separately while repeated references to the same Mesh still restore as references. Added a regression test.

newton/_src/sim/builder.py
```diff
@@ -5284,11 +5289,18 @@ def add_shape(
...
+            resolved_opacity = getattr(src, "opacity", None)
+        if resolved_opacity is None:
+            resolved_opacity = 1.0
+
         self.shape_flags.append(shape_flags)
         self.shape_type.append(type)
         self.shape_scale.append((float(scale[0]), float(scale[1]), float(scale[2])))
         self.shape_source.append(src)
         self.shape_color.append(resolved_color)
+        self.shape_opacity.append(float(resolved_opacity))
```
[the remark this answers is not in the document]
eric-heiden: Fixed in 8cc477fda: shape opacity is converted to float and validated as finite in [0, 1] before any shape lists are mutated, with a regression test covering invalid values.

jcarius-nv:

newton/_src/viewer/gl/shaders.py
```diff
@@ -382,7 +403,22 @@
     // gamma correction (sRGB)
     color = pow(color, vec3(1.0 / 2.2));
 
+#ifdef ENABLE_TRANSPARENCY
+    if (transparent_pass)
+    {
+        float alpha = clamp(Opacity, 0.0, 1.0);
+        float depth_weight = clamp(pow(1.0 - gl_FragCoord.z, 2.0) * 8.0 + 0.01, 0.01, 10.0);
```
mmacklin: This looks suspicious like it would be quite unit / scale dependent?

newton/_src/viewer/viewer.py
```diff
@@ -2259,15 +2353,85 @@ def _log_com(self, state: newton.State):
 
         self.log_points("/model/com", self._com_positions, com_radius, self._com_colors, hidden=not self.show_com)
 
+    def _get_triangle_opacity_groups(
```
mmacklin: I wonder if a simpler approach here would just be to do a sorted back-to-front render pass for transparent triangles instead of doing multiple groups?

newton/_src/viewer/gl/shaders.py
```diff
@@ -382,7 +404,23 @@
     // gamma correction (sRGB)
     color = pow(color, vec3(1.0 / 2.2));
 
+#ifdef ENABLE_TRANSPARENCY
+    if (transparent_pass)
+    {
+        float alpha = clamp(Opacity, 0.0, 1.0);
+        // Avoid depth-derived weights here: clip-space depth varies with camera range and scene scale.
+        float accum_weight = max(alpha, 0.01);
+        FragColor = vec4(color * alpha * accum_weight, alpha * accum_weight);
```
adenzler-nvidia: This resolves the scale dependence, but by removing depth from the weight entirely this is no longer McGuire-Bavoil weighted-blended OIT — it is the depth-oblivious weighted-average method (Meshkin / Bavoil-Myers) that the depth weights were designed to improve on. Concretely: the depth order of transparent layers has no effect on the result (red glass in front of blue renders identically to blue in front of red); only the background attenuation via revealage is order-aware.

If we want depth-aware OIT, the principled version of this fix is to weight by linearized view-space depth normalized by the camera range, which is unit-free by construction: `d = (z_view - near) / (far - near)`, `w = alpha * clamp(a / (1e-5 + pow(d, k)), w_min, w_max)` (McGuire's equation 10 is the clip-space variant of the same idea). If we deliberately stay with alpha-only weights, the pass should be named/documented as weighted-average transparency, and it strengthens the case for making the sorted path the default (see the MSAA comment in opengl.py).

newton/_src/usd/utils.py
```diff
@@ -1663,6 +1668,48 @@ def _coerce_float(value: Any) -> float | None:
...
+    properties = _empty_material_properties()
+    primvars_api = UsdGeom.PrimvarsAPI(prim)
+
+    display_color = primvars_api.GetPrimvar("displayColor")
+    if display_color:
+        color_value = display_color.Get()
+        color = _coerce_color(color_value)
+        if color is not None:
+            properties["color"] = _color_to_display_space(color, display_color.GetAttr())
+        properties["opacity"] = _coerce_opacity(color_value)
```
adenzler-nvidia: Blocking: `displayColor` is a `color3f[]` primvar — an array of RGB triples, never RGBA. For any prim with 2+ entries (per-face/per-vertex colors, common in DCC exports), the flattened array has size >= 6, so `_coerce_opacity` takes the `size >= 4` branch and returns element [3] — the red channel of the second color — as the shape's opacity. A vertex-colored mesh without `displayOpacity` then imports semi-transparent with no warning, and because opacity now counts as a visual cue in `_has_visual_material_properties`, a collision-only mesh carrying display colors can additionally be promoted to visible. Suggest dropping this line (the `displayOpacity` primvar/attribute lookups below cover the real sources) or restricting the RGBA interpretation to a single 4-component value.

Related, line 1701: an N-entry `displayOpacity` with N >= 4 also collapses to element [3] (the 4th vertex's value — neither first nor mean), and N = 2 or 3 silently returns None. A deterministic representative (first or mean) plus a one-time warning when varying data is discarded would be better.

CHANGELOG.md
```diff
@@ -340,6 +349,8 @@
 - Add `newton.geometry.compute_offset_mesh()` for extracting offset surface meshes from any collision shape, and a viewer toggle to visualize gap + margin wireframes in the GL viewer
 - Add differentiable rigid contacts (experimental) with respect to body poses via `CollisionPipeline` when `requires_grad=True`
 - Add per-shape display colors via `ModelBuilder.shape_color`, `Model.shape_color`, and `color=` on `ModelBuilder.add_shape_*`; mesh shapes fall back to `Mesh.color` when available and viewers honor runtime `Model.shape_color` updates
+- Add per-shape display opacity via `ModelBuilder.shape_opacity`, `Model.shape_opacity`, `Mesh.opacity`, and `opacity=` on `ModelBuilder.add_shape_*`; viewers and URDF/MJCF/USD importers preserve authored alpha where supported
```
adenzler-nvidia: These two entries (and the example entry at line 375) are inside the released `[1.1.0] - 2026-04-13` section — merging as-is rewrites shipped release notes. They need to move to `[Unreleased]`. The line-375 entry could also mention the RJ45 socket and softbody-Franka duck opacity changes this PR makes.

CHANGELOG.md
```diff
@@ -62,6 +62,15 @@
 
 ### Changed
 
+- Use a `wp.Mesh` BVH for `GeoType.HFIELD` raycasting (built during `ModelBuilder.finalize()`) instead of per-thread DDA grid traversal. Migrate custom `raycast_kernel` launches by removing the old `shape_heightfield_index`, `heightfield_data`, and `heightfield_elevations` arguments; pass `model.shape_source_ptr` with `model.shape_type` and `model.shape_scale` like other mesh geometry. Keep the heightfield arrays only for collision kernels.
```
adenzler-nvidia: Merge artifact: lines 65-73 restate entries that already exist at lines 74-79 in different wording (the HFIELD BVH raycasting change appears three times: 65, 68, 74; the mesh-SDF writeback, convex-hull ValueError, auto-BVH-build, and USD SDF entries are each duplicated). Similarly further down: lines 367-370 duplicate shipped 1.2.0 entries, and line 371 credits this PR with adding the cable cross-slide example, which already ships in 1.3.0 (line 59) — this PR only makes its pulleys transparent. Please rebuild the changelog from upstream main plus this PR's own entries.

newton/_src/utils/import_urdf.py
```diff
@@ -319,18 +319,21 @@ def resolve_urdf_asset(filename: str | None) -> tuple[str | None, tempfile.Named
...
             rgba = color_el.get("rgba")
             if rgba:
-                parts = rgba.split()
-                if len(parts) >= 3:
-                    color = (float(parts[0]), float(parts[1]), float(parts[2]))
+                values = np.fromstring(rgba, sep=" ", dtype=np.float32)
+                if len(values) >= 3:
+                    color = (float(values[0]), float(values[1]), float(values[2]))
+                if len(values) >= 4:
+                    opacity = float(values[3])
```
adenzler-nvidia: This forwards the raw alpha into the builder's new [0, 1] validation, so a URDF with e.g. `rgba="1 0 0 1.5"` — which imported fine before this PR (alpha ignored) — now aborts the entire import over a display-only attribute. MJCF clamps for exactly this reason (import_mjcf.py:816), and the USD path has the same problem via `_coerce_opacity`. Suggest clamping (optionally with a warning) in all three importers and keeping the `ValueError` for the direct builder API.

newton/_src/viewer/gl/opengl.py
```diff
@@ -1254,13 +1438,19 @@ def render(self, camera, objects, lines=None, wireframe_shapes=None, arrows=None
...
         gl.glViewport(0, 0, self._screen_width, self._screen_height)
 
+        use_weighted_oit = self._can_use_weighted_transparency(scene_has_transparency)
+
         # select target framebuffer (MSAA or regular) for scene rendering
-        target_fbo = self._frame_msaa_fbo if getattr(self, "msaa_samples", 0) > 0 else self._frame_fbo
+        target_fbo = (
+            self._frame_fbo
+            if use_weighted_oit
+            else (self._frame_msaa_fbo if getattr(self, "msaa_samples", 0) > 0 else self._frame_fbo)
```
adenzler-nvidia: When the weighted path engages, the opaque pass renders into the single-sample `_frame_fbo` and the MSAA resolve at line 1484 is skipped — so making a single shape transparent silently disables anti-aliasing for the entire scene, while the sorted fallback keeps MSAA. Either render the opaque pass into the MSAA FBO and resolve color+depth into `_frame_fbo` before the OIT accumulation pass, or document the tradeoff. This is also an argument for defaulting to the sorted path (see the shaders.py comment).

newton/_src/geometry/types.py
```diff
@@ -1581,7 +1594,7 @@ class TetMesh:
             tet_mesh = newton.TetMesh(vertices, tet_indices)
     """
 
-    _RESERVED_ATTR_KEYS = frozenset({"vertices", "tet_indices", "k_mu", "k_lambda", "k_damp", "density"})
+    _RESERVED_ATTR_KEYS = frozenset({"vertices", "tet_indices", "k_mu", "k_lambda", "k_damp", "density", "opacity"})
```
adenzler-nvidia: Two file-IO regressions here: (1) `opacity` is now part of `TetMesh` state, but `save()` doesn't write it and `create_from_file()` doesn't restore it, so a `TetMesh` with opacity round-trips to fully opaque with no signal; (2) adding `"opacity"` to `_RESERVED_ATTR_KEYS` means a third-party `.vtk`/`.msh` file whose cell/point data happens to contain an array named `opacity` now raises `ValueError` in `__init__` where it previously loaded fine. The USD tet path handles both correctly; the file path needs the same treatment.

newton/_src/viewer/viewer.py
```diff
@@ -1664,6 +1746,10 @@ def _populate_shapes(self):
             flags = shape_flags[s]
             parent = shape_body[s]
             static = parent == -1
+            opacity = float(shape_display_opacity[s]) if shape_display_opacity is not None else 1.0
+            opacity = max(0.0, min(1.0, opacity))
+            transparent = opacity < 0.999
```
adenzler-nvidia: The `< 0.999` transparency threshold is duplicated across five files (here, viewer_gl.py:172 and :756, opengl.py:389 and :1026, viewer_rtx.py:1393) with nothing pinning them together, and no test asserts the opaque/transparent classification at all — a regression that drops transparency from `_hash_shape` or drifts one constant would pass the current suite while rendering transparent shapes opaque. Suggest a single shared constant and a ViewerNull/`_shape_instances`-based test that two identical shapes with opacity 1.0 vs 0.5 land in separate batches with `transparent` set.

adenzler-nvidia (dismissed):
Thanks — this is a substantial feature and most of it holds up well: the weighted-transparency mechanics (accumulation/revealage buffers, resolve, shared depth) are implemented correctly, the opacity dirty-tracking chain traces out sound, and the importer/recorder tests are genuinely behavioral. Requesting changes for the inline items; the two hard blockers are the USD `displayColor` opacity misread and the CHANGELOG placement.

On the transparency design, following up on Miles's two comments: switching the OIT weight to `max(alpha, 0.01)` does remove the scale dependence, but it also removes depth from the weighting entirely, which demotes weighted-blended OIT to depth-oblivious weighted-average transparency — flipping the depth order of two overlapping transparent surfaces no longer changes the image at all; only the background attenuation via revealage is order-aware. Details and the principled alternative in the shaders.py comment. Combined with the fact that the weighted path currently disables MSAA scene-wide as soon as one shape is transparent (opengl.py comment), I think we should either make the sorted path the default — it is exact for the non-intersecting showcase cases (socket, cups, pulleys) and keeps MSAA — with the weighted path opt-in for self-overlapping cloth/soft bodies, or keep weighted as the default but with a normalized view-space depth weight and an MSAA-resolved opaque pass.

Smaller items, collected here rather than inline:

- Per-triangle opacity is never validated (NaN or 5.0 is accepted at build time and clamped at render), while shape opacity raises `ValueError` — these should behave the same.
- `_get_triangle_opacity_groups` creates one mesh group per unique opacity value; continuous per-triangle opacity would explode into thousands of groups/draw calls. Worth a cap or bucketing (also relevant to Miles's grouping question).
- A wrong-length `model.tri_opacity` silently resets all triangles to fully opaque (viewer.py:2376) — deserves a warning.
- Viser: `batched_opacities` is silently filtered out by `_call_scene_method` for `add_batched_meshes_trimesh` (the pinned viser only accepts it on `add_batched_meshes_simple`), so a one-shot `log_instances` on a textured mesh stays opaque with no signal. A warning when a requested appearance kwarg is dropped would surface this class of issue.
- Stale docs: `MeshGL.update()` documents a nonexistent `scale` parameter; the `ViewerGL.log_state()` docstring still says "2 kernel launches + 1 D2H copy" (the steady state now also copies the opacity flags and conditionally launches the repack kernel); "for this prototype" in the shadow-pass exclusion comment reads as leftover.
- Coverage: most of the uncovered lines are genuinely GL-bound, but the dirty-tracking chain (`_shape_opacity_groups_changed`, `_rebuild_shape_batches_for_opacity_groups`, the flag logic in `_compute_shape_vbo_xforms`) and the triangle-grouping logic are testable without a GL context (module-level kernel, plus the existing `ViewerGL.__new__` pattern) and are the most regression-prone code in the PR.

Also: the branch is currently conflicting with main.

newton/_src/viewer/gl/shaders.py
eric-heiden: Fixed in c136a0ca1. The weight is now computed on a unit-free depth: view depth divided by a per-frame reference distance derived from the transparent content itself (mean camera-space depth of the transparent objects, taken from the already-cached batch centers; camera pivot distance as fallback). Verified that the same two-layer scene renders identically when authored at 0.01x vs 1x scale (mean abs diff 0.03/255), while depth order still affects the blend.

newton/_src/viewer/gl/shaders.py
eric-heiden: Implemented along these lines in c136a0ca1, with one deviation: normalizing by the camera range still degenerates for scenes much smaller than the far plane — with the default near/far of 0.01/1000, a meter-scale scene lands at d ~ 0.001 and the weight saturates at its clamp for every fragment, which is depth-oblivious again in practice. Instead the shader normalizes view depth by a per-frame reference distance derived from the transparent content itself (mean camera-space depth of the transparent objects, with the camera pivot distance as fallback), then applies the McGuire-style curve `w = clamp(10 / (1e-5 + (2d)^2 + (0.6d)^6), 0.01, 300)` on the unit-free `d`. This is scale-independent by construction and keeps depth discrimination centered on the content the camera is actually looking at.

Validated with a two-layer scene: swapping the depth order of the transparent layers changes the blend (it did not affect the overlap region before), and the same scene authored at 0.01x vs 1x scale renders identically (mean abs diff 0.03/255).

newton/_src/usd/utils.py
eric-heiden: Fixed in d717eaa57: `displayColor` is no longer used as an opacity source. Opacity is read from the `displayOpacity` primvar/attribute and material inputs, and RGBA-alpha extraction is restricted to a single 4-component color value (`_coerce_color_opacity` returns `None` for anything else), so per-face/per-vertex `displayColor` arrays can no longer be misread as opacity or promote collision-only meshes to visible.

CHANGELOG.md
eric-heiden: Fixed in d717eaa57: the changelog was rebuilt from upstream main, and this PR's entries now live under `[Unreleased]` — the opacity feature under Added, the transparent example visuals (including the RJ45 socket and softbody-Franka duck) under Changed, and the double-sided USD/RTX output fix under Fixed. Released sections are untouched.

CHANGELOG.md
eric-heiden: Fixed: the changelog was rebuilt from upstream main in d717eaa57, and a last merge-duplicated `soft_contact_max` entry was removed in 5b2559542. The PR's changelog delta vs main is now exactly its own three entries, and the example entry credits only the pulley transparency (the cross-slide example itself shipped in 1.3.0).

newton/_src/utils/import_urdf.py
eric-heiden: Fixed in d717eaa57: importers clamp display-only opacity instead of raising. URDF goes through `_clamp_imported_opacity` (warns and clamps out-of-range rgba alpha, ignores non-finite values), USD uses the same semantics in `_coerce_opacity`, and MJCF keeps its existing clamp. The `ValueError` remains only for the direct builder API. `test_visual_material_rgba_clamps_invalid_opacity` covers the `rgba="0.1 0.2 0.8 1.5"` case.

newton/_src/viewer/gl/opengl.py
eric-heiden: Fixed in d717eaa57: the opaque pass always renders into the MSAA FBO. When the weighted path engages, color and depth are resolved into the single-sample frame FBO mid-frame (`_resolve_msaa_frame`, a `glBlitFramebuffer` of both buffers), and the OIT accumulation FBO shares that resolved depth texture, so transparent fragments still depth-test against the opaque scene and enabling transparency no longer disables anti-aliasing.

newton/_src/geometry/types.py
eric-heiden: Fixed in d717eaa57: (1) `TetMesh.save()` now writes opacity under the dunder key `__opacity__` and `create_from_file()` restores it; the meshio path round-trips it as `newton_opacity` cell data. (2) `"opacity"` was removed from `_RESERVED_ATTR_KEYS`, so third-party `.vtk`/`.msh` files carrying an attribute named `opacity` load as before.

newton/_src/viewer/viewer.py
eric-heiden: Fixed in d717eaa57: the threshold now lives in one place (`OPAQUE_OPACITY_THRESHOLD` in `newton/_src/viewer/utils.py`) and all sites import it. `test_opaque_and_transparent_shapes_use_separate_batches` covers the classification: two identical shapes with opacity 1.0 vs 0.5 land in separate batches with `transparent` set on the latter.

newton/_src/viewer/viewer.py
eric-heiden: The sorted back-to-front pass exists and is the automatic fallback (and can be forced via `renderer.enable_weighted_transparency = False`), but it sorts at object/instance granularity — per-triangle opacity would need a per-frame index-buffer re-sort by camera depth to fold into it, and it still mis-renders self-overlapping cloth where no triangle order is correct. The opacity groups exist for a narrower reason: a mesh draw carries one opacity per draw, so distinct per-triangle opacity values need distinct index ranges regardless of which transparency pass runs. To keep that bounded, continuous per-triangle opacity is quantized to at most 32 groups (d717eaa57), and group recomputation is now gated behind a device-side change check so the common static case costs one small kernel per frame (c136a0ca1).

nvtw (changes requested):
AI findings (feel free to ignore - it does not sound super problematic):

  - [P2] Mixed-opacity ViewerGL batches render incorrectly: MeshInstancerGL.has_transparency()
    (https://github.com/newton-physics/newton/blob/c4301c5008be1594caf59dcdb36f6250221918e0/newton/_src/viewer/gl/opengl.py#L1000-L1002) classifies
    the entire instancer as transparent if any instance is translucent. Consequently, opacity-1 instances in that batch:
      - are excluded from the shadow pass
        (https://github.com/newton-physics/newton/blob/c4301c5008be1594caf59dcdb36f6250221918e0/newton/_src/viewer/gl/opengl.py#L2052-L2059);

      - do not write depth;
      - can blend geometry behind them through OIT.

    Internal model shapes are separated into opaque/transparent batches, but the public ViewerGL.log_instances() API accepts arbitrary per-instance
    opacities without performing that split. The GL backend should split mixed batches or render filtered opaque and translucent subsets.
    
    
Will approve as soon as the merge conflicts are resolved

eric-heiden:
@nvtw Addressed the mixed-opacity ViewerGL batch issue in 25816e2a5. ViewerGL.log_instances() now partitions a mixed batch into homogeneous opaque and translucent instancers, so opaque instances remain in the depth and shadow passes while only the translucent subset enters the transparency pass. Homogeneous batches keep the existing fast path. A regression test verifies both the data partition and renderer pass classification. Validation passed: all 32 test_shape_colors tests, the headless GL shadow-render test, the RJ45 transparency example, and the full pre-commit suite. Please take another look when convenient.

nvtw (dismissed):
Looks good now, thanks for the improvements

nvtw (dismissed):

nvtw (dismissed):

eric-heiden:
I've added `Model.tri_color` so that cloths can now also store color information for triangles. Then I improved rendering of some of the cloth examples, like the cloth rollers demo.
image
@AnkaChan for viz

nvtw (approved):
Message from the approval bot: Still looking good :)

PR #3058, opened 2026-06-04
Capture expected warnings in inertia validation tests

adenzler-nvidia:
## Description

The inertia-validation tests wrap warning-emitting calls in `warnings.catch_warnings(record=True)` without `warnings.simplefilter("always")`. `catch_warnings` copies the ambient filter list rather than resetting it, so these tests silently depend on the process-wide policy showing `UserWarning`s. Under any policy that escalates warnings to errors — e.g. `python -W error::UserWarning`, or a consumer running the suite strictly — the expected correction warning is raised instead of recorded, and the test errors before its assertions run.

This makes warning capture independent of the ambient filter:

- blocks that expect a specific warning now use the unittest-native `assertWarns` / `assertWarnsRegex` (which reset to `simplefilter("always")` internally);
- record-and-count blocks (`len(w) == N`, including the "no warnings" cases) keep `catch_warnings(record=True)` but add `simplefilter("always")`;
- three blocks that previously swallowed an expected warning now assert it instead;
- the shared parity helper, which exercises many inputs and only checks numeric parity between the two validation paths, records (does not raise) the warnings it deliberately doesn't assert on.

No production code changes; behavior of `verify_and_correct_inertia` and `ModelBuilder.finalize()` is unchanged.

## Checklist

- [ ] `CHANGELOG.md` has been updated (not a user-facing change)

## Test plan

```
uv run --extra dev python -W error::UserWarning -m unittest newton.tests.test_inertia_validation
uv run --extra dev -m newton.tests -k test_inertia_validation
```

## Bug fix

**Steps to reproduce (without this PR, on `main`):**

Run the suite under a warnings-as-errors policy using a stock interpreter flag — no special test infrastructure required:

```
uv run --extra dev python -W error::UserWarning -m unittest newton.tests.test_inertia_validation
```

Observe `FAILED (errors=29)`: expected `UserWarning`s (e.g. "Negative mass … setting to 0", "Mass … is below bound") are raised inside `catch_warnings(record=True)` blocks that lack `simplefilter("always")`, so they propagate instead of being recorded. With this PR the same command reports `OK`.

## Summary by CodeRabbit

* **Tests**
  * Standardized warning assertions across inertia validation and integration tests for consistent capture and absence checks
  * Added explicit checks for corrected mass/inertia values (including NaN/Inf, zero-with-bound, and singular-inertia repair)
  * Strengthened integration tests to verify bounds and triangle-inequality after corrections
  * Ensured parity tests record warnings without affecting comparison and preserved builder state checks

### DIFF
--- a/newton/tests/test_inertia_validation.py
+++ b/newton/tests/test_inertia_validation.py
@@ -65,7 +65,7 @@ class TestInertiaValidation(unittest.TestCase):
         bound_inertia = 1.0
         inertia = wp.mat33([[0.1, 0.0, 0.0], [0.0, 2.0, 0.0], [0.0, 0.0, 0.5]])
 
-        with self.assertWarns(UserWarning):
+        with self.assertWarnsRegex(UserWarning, r"Minimum eigenvalue .* is below bound"):
             corrected_mass, corrected_inertia, was_corrected = verify_and_correct_inertia(
                 mass, inertia, bound_inertia=bound_inertia
             )
--- a/newton/tests/test_inertia_validation.py
+++ b/newton/tests/test_inertia_validation.py
@@ -21,15 +21,13 @@ class TestInertiaValidation(unittest.TestCase):
         mass = -10.0
         inertia = wp.mat33([[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]])
 
-        with warnings.catch_warnings(record=True) as w:
+        with self.assertWarnsRegex(UserWarning, "Negative mass"):
             corrected_mass, corrected_inertia, was_corrected = verify_and_correct_inertia(mass, inertia)
 
-            self.assertTrue(was_corrected)
-            self.assertEqual(corrected_mass, 0.0)
-            # Zero mass should have zero inertia
-            self.assertTrue(np.allclose(np.array(corrected_inertia), 0.0))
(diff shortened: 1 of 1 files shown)


vreutskyy (dismissed):
LGTM

newton/tests/test_inertia_validation.py
```diff
@@ -21,130 +21,118 @@ def test_negative_mass_correction(self):
...
+        self.assertTrue(inertia_array[2, 2] >= 0)
 
     def test_inertia_bound(self):
         """Test that inertia diagonal elements below bound are clamped."""
         mass = 1.0
         bound_inertia = 1.0
         inertia = wp.mat33([[0.1, 0.0, 0.0], [0.0, 2.0, 0.0], [0.0, 0.0, 0.5]])
 
-        with warnings.catch_warnings(record=True) as w:
+        with self.assertWarns(UserWarning):
```
eric-heiden: 🟡 `test_inertia_bound` should assert the bound-specific warning, not just any `UserWarning`. This fixture can also trigger triangle-inequality warnings, so a regression that drops the inertia-bound warning could still satisfy this assertion.

```suggestion
        with self.assertWarnsRegex(UserWarning, r"Minimum eigenvalue .* is below bound"):
```

eric-heiden (approved):

PR #3061, opened 2026-06-04
Handle expected warnings in tests under strict mode

adenzler-nvidia:
## Description

Several tests emit `newton.*`-attributed `UserWarning`s on legitimate input, which fail under a warnings-as-errors policy. Where the warning had a clean cause it is removed; otherwise the specific advisory is tolerated while all other warnings stay strict.

- **test_softbody** — `add_surface_mesh_edges=False`: the tet adjacency/coloring tests don't use surface bending edges, whose non-manifold-boundary warning was the only cause.
- **test_mujoco_solver** — `njmax` 30 → 32 (MjWarp's minimum), so the solver no longer bumps it and warns; the test never inspects it.
- **test_import_usd** — drop the gratuitous `MassAPI` from the gravcomp fixture bodies; nothing consumes it and it triggered a zero-mass warning.
- **test_menagerie_mujoco / test_robot_composer** — ignore the specific advisory (margin zeroing #2106; massless tool-flange parent) emitted by legitimate real-robot assets, leaving other warnings strict.

## Checklist

- [ ] `CHANGELOG.md` has been updated (test-only change)

## Test plan

```
uv run --extra dev -m newton.tests -k test_softbody -k test_import_usd -k test_robot_composer
uv run --extra dev -m newton.tests -k test_menagerie_mujoco.TestMenagerie_UnitreeGo2
uv run --extra dev -m newton.tests -k test_mujoco_solver.TestMuJoCoSolverForceSpaceContactSolref
```

## Summary by CodeRabbit

* **Tests**
  * Modified USD gravcomp import test to change how gravcomp values are produced and verified.
  * Added targeted warning suppression around solver initialization and robot-assembly steps to reduce expected advisory noise.
  * Increased solver capacity in one test to accommodate larger joint/contact arrays.
  * Wrapped soft-mesh construction in tests to ignore non-manifold advisory warnings during mesh setup.


newton/tests/test_menagerie_mujoco.py
```diff
@@ -1669,7 +1670,13 @@ def _ensure_models(self):
         if self.solver_integrator is not None:
             solver_kwargs["integrator"] = self.solver_integrator
 
-        cls._newton_solver = SolverMuJoCo(cls._newton_model, **solver_kwargs)
+        # Some real MJCFs (e.g. Apollo, Go2) author geom margins that the
+        # native-CCD path zeroes (#2106); the field comparison below already
+        # mirrors that zeroing, so tolerate the per-geom advisory rather than
+        # failing under strict warnings. Other warnings still surface.
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", message=r"Geom .* zeroed for NATIVECCD")
```
eric-heiden: 🔴 This filter only matches NATIVECCD zeroing warnings that start with `Geom`, but `SolverMuJoCo` can also emit the same advisory for contact pairs as `Pair (...)`. Since `warnings.filterwarnings` uses `re.match`, the pair variant falls through and strict-warning runs can still fail as soon as a menagerie MJCF includes a tuned `<pair>` with `margin` or `gap`.

Please widen the regex so both intentional NATIVECCD zeroing advisories are ignored, and update the nearby comment so it does not describe this as only per-geom.

```suggestion
            warnings.filterwarnings("ignore", message=r"(Geom|Pair).* zeroed for NATIVECCD")
```

newton/tests/test_softbody.py
```diff
@@ -251,6 +251,9 @@ def _build_model_with_soft_mesh(vertices: list[tuple[float, float, float]], tets
         k_mu=1.0,
         k_lambda=1.0,
         k_damp=0.0,
+        # These tests verify tet adjacency/coloring only; skip the surface
+        # bending edges, whose non-manifold-boundary warning is unrelated noise.
+        add_surface_mesh_edges=False,
```
eric-heiden: 🟡 I think this is suppressing more than the warning. With the default `add_surface_mesh_edges=True`, `add_soft_mesh()` builds surface triangles, runs `MeshAdjacency`, creates surface edges, and later `SolverVBD._compute_particle_force_element_adjacency()` includes vertex-edge adjacency. Setting this to `False` skips that whole path for the pyramid tests; in this mesh the default path creates 57 surface edges while emitting the expected `Detected non-manifold edge` warning.

Could we keep the default path covered and narrow the warning handling instead? Either fix the pyramid mesh if the non-manifold boundary is unintended, or wrap the `add_soft_mesh()` calls with `warnings.catch_warnings()` and ignore only `Detected non-manifold edge`, including the direct call in `test_tet_graph_coloring_is_valid` below.

newton/tests/test_menagerie_mujoco.py
adenzler-nvidia: Widened to `(Geom|Pair).* zeroed for NATIVECCD` and updated the comment to drop "per-geom" — confirmed both message variants match under `re.match` (the `Pair (...)` one did not before). Fixed in 88ce369b0.

newton/tests/test_softbody.py
adenzler-nvidia: Reverted the `add_surface_mesh_edges=False` change in both call sites and instead wrapped each `add_soft_mesh()` in `warnings.catch_warnings()` ignoring only `Detected non-manifold edge`, so the default surface-edge path (MeshAdjacency + VBD vertex-edge adjacency) stays covered. The pyramid surface is non-manifold by construction — verified the advisory still fires on the default path (57 surface edges). Fixed in 88ce369b0.

eric-heiden (approved):

PR #3062, opened 2026-06-04
Implement SolverMuJoCo.reset()

### LINKED ISSUE #2552: [REQ] Implement `SolverMuJoCo.reset()` method
### Description

Based on the agreed signature in https://github.com/newton-physics/newton/issues/1266 we need to implement a reset method that allows the resetting of a subset of world in SolverMuJoCo.

### Motivation / Use Case

Isaac Lab encountered an issue when resetting environments that leaves some part of the internal MuJoCo state/model stale, so a proper solver-specific reset mechanism is needed.
See https://github.com/isaac-sim/IsaacLab/issues/5359 for more details.

### Alternatives Considered

_No response_

adenzler-nvidia:
## Description

Implements `SolverMuJoCo.reset()` on top of the unified `SolverBase.reset()` API (#2657 / #1266), so a subset of worlds can be reset in place — e.g. for per-world NaN recovery during RL training, where Isaac Lab resets the Newton-facing state but MuJoCo-owned solver buffers can remain NaN and warm-start the next step from stale data.

Compared against `mujoco_warp.reset_data` to keep the cleared set minimal: buffers that MuJoCo recomputes every step (mass matrix, contacts, constraint data, kinematics) are left untouched.

For the selected worlds, `reset()`:

- Resets `State.joint_q` / `joint_qd` to the model defaults, gated by `StateFlags` (`JOINT_Q` / `JOINT_QD`). Because MuJoCo is a reduced-coordinate solver, `body_q` / `body_qd` are FK outputs recomputed on the next step, so the `BODY_*` and particle flags are not actionable and are ignored (no FK is run — it would be redundant and costly).
- Unconditionally clears the persistent internal buffers `qacc_warmstart`, `qfrc_applied`, `xfrc_applied`, `act`, and `ctrl`. `qacc` is intentionally not cleared: the solver overwrites it from `qacc_warmstart` at the start of every step.

`qpos` / `qvel` are normally re-synced from the reset `state` on the next `step()`. When `update_data_interval != 1` that per-step sync is disabled or sparse, so the reset coordinates are pushed into `qpos` / `qvel` immediately.

Closes #2552.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_mujoco_reset
```

Covers per-world masking, all-worlds reset, joint_q/joint_qd reset to model defaults, `flags=0` (buffer-only), body flags ignored, NaN warm-start recovery, and the `update_data_interval` sync behavior.

## New feature / API change

```python
import newton
from newton import StateFlags

solver = SolverMuJoCo(model)
# Reset worlds 0 and 3 (boolean per-world mask) to model defaults and
# scrub MuJoCo's stale internal buffers before the next step.
solver.reset(state, world_mask=mask, flags=StateFlags.JOINT_Q | StateFlags.JOINT_QD)
```

## Summary by CodeRabbit

* **New Features**
  * Added selective world reset: restore joint coordinates/velocities to model defaults, clear per-world simulation buffers, and optionally target worlds via a mask. Resets support recovery from numerical divergence (e.g., NaNs) and respect reset flags.

* **Behavior**
  * Validates mask length and honors flag-driven semantics; can immediately sync reset state to simulator depending on sync interval.

* **Tests**
  * Added comprehensive tests covering masked/all-world resets, buffer clearing, NaN recovery, flag behavior, and sync timing.


eric-heiden:
Thanks, just a small nit.

I also ran a RL-style reproduction with a 4-world cartpole model using `cartpole.usda`.

Result: the behavior matches the requested expectation.

- Applied a `NaN` joint force only to world 2.
- After that step, only world 2 became non-finite in public state and MuJoCo internals: `qpos`, `qvel`, `qacc`, `qacc_warmstart`, `qfrc_applied`.
- With `solver.reset(state, world_mask=[False, False, True, False])`, the next regular-control step recovered fully: all worlds were finite in public state and MuJoCo internals.
- With manual repair only, by assigning public `state.joint_q` / `state.joint_qd` back to finite defaults and running FK, MuJoCo internals still retained NaNs. The next regular-control step re-poisoned world 2.

newton/tests/test_mujoco_reset.py
```diff
@@ -0,0 +1,191 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 The Newton Developers
```
eric-heiden: 🔴 `newton/tests/test_mujoco_reset.py` is a new file, but its SPDX copyright year is `2025`. New files should use the current year, so this should be `2026`.

```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
```

eric-heiden (approved):

Issue #2914, opened 2026-05-20
[BUG] Selection API link names contains duplicated body names when articulation has close loop


ooctipus:
### Bug Description

When same body 1 appears under multiple joints
selection api's link_name will duplicate them number of joint times, and link_names it becomes unreliable source to reference the order/indices of the links

Though interestly enough `frequency_layouts[self.root_view.model.AttributeFrequency.BODY]` will still be accurate source to get de-duplicated link names that follows right link data order.

Here is reproducing script,

```
# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
# SPDX-License-Identifier: Apache-2.0

"""Reproduce duplicate leaf link names in Newton's articulation selection.

This is a minimal Newton-only reproduction of the Digit/Lab failure mode where
an articulation view resolves multiple distinct bodies to the same leaf link
name.  The real Digit asset has closed-loop / mimic structure that produces
three ``*_leg_toe_roll`` bodies per side in Newton, while Lab's contact sensor
body names are deduplicated to one per side.
"""

from __future__ import annotations

from collections import Counter

import warp as wp

import newton
from newton.selection import ArticulationView


TOE_ROLL_BODY_LABELS = [
    "World/Digit/left/main/left_leg_toe_roll",
    "World/Digit/left/loop_a/left_leg_toe_roll",
    "World/Digit/left/loop_b/left_leg_toe_roll",
    "World/Digit/right/main/right_leg_toe_roll",
    "World/Digit/right/loop_a/right_leg_toe_roll",
    "World/Digit/right/loop_b/right_leg_toe_roll",
]


def _leaf_name(label: str) -> str:
    return label.rsplit("/", 1)[-1]


def _add_digit_like_articulation() -> newton.Model:
    builder = newton.ModelBuilder()
    body_ids = []
    joint_ids = []
    inertia = wp.mat33(0.2, 0.0, 0.0, 0.0, 0.2, 0.0, 0.0, 0.0, 0.2)

    for index, label in enumerate(TOE_ROLL_BODY_LABELS):
        body_id = builder.add_link(label=label, mass=1.0, inertia=inertia)
        body_ids.append(body_id)

        if index == 0:
            joint_id = builder.add_joint_free(parent=-1, child=body_id, label="root")
        else:
            joint_id = builder.add_joint_revolute(
                parent=body_ids[index - 1],
                child=body_id,
                axis=wp.vec3(0.0, 0.0, 1.0),
[...]

PR #3071, opened 2026-06-04
Add opt-in leaf-or-label matching for ArticulationView and sensors

adenzler-nvidia:
## Description

Lets `ArticulationView` and the contact/IMU/frame-transform sensors match label patterns against a label's **leaf** (final `/`-delimited component) in addition to the full label. This makes the `*_labels` attributes (renamed in #3057) usable as selection keys: a path pattern like `"*/left/fingertip"` can target one of several entries that share a leaf, and a slash-bearing `exclude_*` pattern can carve a single colliding entry out of a broad leaf-based include. A pattern containing `/` can never match a slash-free leaf, so such patterns stay precise.

This is **opt-in** to avoid changing behavior. `match_labels()` gains an internal `match_leaf` flag (default `False`), and `ArticulationView`, `SensorContact`, `SensorIMU`, and `SensorFrameTransform` gain a `match_full_labels: bool | None = None` parameter:

- `None` (default) keeps the current matching and emits a `DeprecationWarning` **only** when enabling leaf matching would change the result, so unaffected callers (flat labels, exact patterns, integer indices) are never warned.
- `True` adopts leaf-or-label matching now.
- `False` pins the current behavior.

The default will become `True` in a future release. This mirrors the `newton.use_coord_layout_targets` transition pattern (opt-in flag + `DeprecationWarning`, default flips later).

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_sensor_contact
uv run --extra dev -m newton.tests -k test_sensor_imu
uv run --extra dev -m newton.tests -k test_sensor_frame_transform
```

## New feature / API change

```python
import newton

# Gripper with two fingertips sharing the leaf name "fingertip":
#   palm/left/fingertip, palm/right/fingertip
view = newton.selection.ArticulationView(
    model, "gripper", include_links="*/left/fingertip", match_full_labels=True
)
view.link_labels  # ['palm/left/fingertip'] — one of the colliding pair, targeted by path

# Without match_full_labels, today's behavior (leaf-only) is unchanged; a
# DeprecationWarning is emitted only if the upcoming default would select differently.
```

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in match_full_labels option for articulation views and sensors so patterns can match full labels or leaf names; current behavior preserved by default.
* **Bug Fixes / Behavior**
  * Emits a deprecation warning when selections would change under the upcoming default.
* **Documentation**
  * Updated changelog and constructor docs to describe the new option and warning.
* **Tests**
  * Added tests covering matching modes and deprecation-warning cases.

### DIFF
--- a/newton/_src/utils/selection.py
+++ b/newton/_src/utils/selection.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import functools
+import warnings
 from fnmatch import fnmatch
 from types import NoneType
 from typing import TYPE_CHECKING, Any
@@ -394,7 +395,7 @@ def find_matching_ids(pattern: str, labels: list[str], world_ids, world_count: i
     return grouped_ids, global_ids
 
 
-def match_labels(labels: list[str], pattern: str | list[str] | list[int]) -> list[int]:
+def match_labels(labels: list[str], pattern: str | list[str] | list[int], match_leaf: bool = False) -> list[int]:
     """Find indices of elements in ``labels`` that match ``pattern``.
 
     See :ref:`label-matching` for the pattern syntax accepted across Newton APIs.
@@ -405,6 +406,8 @@ def match_labels(labels: list[str], pattern: str | list[str] | list[int]) -> lis
             A ``list[str]`` matches any pattern.
             A ``list[int]`` is returned as-is (indices used directly).
             Mixing ``str`` and ``int`` in the same list is not allowed.
+        match_leaf: If ``True``, also match a pattern against each label's leaf
+            (final ``/``-delimited component). Slash-bearing patterns stay precise.
 
     Returns:
         Unique list of matching indices, or ``pattern`` itself for ``list[int]``.
@@ -412,8 +415,12 @@ def match_labels(labels: list[str], pattern: str | list[str] | list[int]) -> lis
     Raises:
         TypeError: If list elements are not all ``str`` or all ``int``.
     """
+
+    def matches(label: str, pat: str) -> bool:
+        return fnmatch(label, pat) or (match_leaf and fnmatch(get_name_from_label(label), pat))
+
     if isinstance(pattern, str):
-        return [idx for idx, label in enumerate(labels) if fnmatch(label, pattern)]
+        return [idx for idx, label in enumerate(labels) if matches(label, pattern)]
 
     if not isinstance(pattern, list):
         raise TypeError(f"Expected a list of str patterns or a list of int indices, got: {type(pattern)}")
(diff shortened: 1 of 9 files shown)


newton/_src/utils/selection.py
```diff
@@ -498,6 +555,11 @@ class ArticulationView:
         exclude_joint_types (list[int] | None): List of joint types to exclude.
         include_loop_closing_joints (bool): If True, include converted loop-closing joints.
         verbose (bool | None): If True, prints selection summary.
+        match_full_labels (bool): If ``True``, ``include_*``/``exclude_*`` patterns match full
```
eric-heiden, line 558: Nit: it would be nice if we could remove the type annotations from the docstring here and only use type hints like we do in the rest of the code.

eric-heiden (approved):
Looks good to me! It's nice you added the deprecation warning when the behavior would change in the future.

nvlukasz (commented):
The original intent of `ArticulationView.<thing>_name` was to hold names of robot parts. Filtering was based on the name of the parts, so you could match "torso", "head", "left_finger", which seems intuitive.

When loading from MJCF or URDF, Newton labels are just the robot part names, so that works. But when loading from USD, the labels are USD stage paths, so there's a prefix that encodes the robot location in the stage, like `/world/envs/env0/`). You now have to match with "*torso", "*head", etc. This difference feels wrong to me.

IMO, the right approach is to strip the leading part of the path if it's present, which drops the location in the stage and just leaves the unique robot part name (which may still contain path separators, e.g., "left/finger" or "right/finger". We wouldn't need to add any boolean flags for matching against leaf names and/or full labels, we would just match against the names, but we change how the names are derived from labels a bit (strip common leading path component). I think that would be a simpler change.

Given these labels from the original GH-2914:
```python
TOE_ROLL_BODY_LABELS = [
    "World/Digit/left/main/left_leg_toe_roll",
    "World/Digit/left/loop_a/left_leg_toe_roll",
    "World/Digit/left/loop_b/left_leg_toe_roll",
    "World/Digit/right/main/right_leg_toe_roll",
    "World/Digit/right/loop_a/right_leg_toe_roll",
    "World/Digit/right/loop_b/right_leg_toe_roll",
]
```
The leading "World/Digit/" prefix would be stripped and `view.link_names` would be:
```python
[
    left/main/left_leg_toe_roll",
    left/loop_a/left_leg_toe_roll",
    left/loop_b/left_leg_toe_roll",
    right/main/right_leg_toe_roll",
    right/loop_a/right_leg_toe_roll",
    right/loop_b/right_leg_toe_roll",
]
```
Filtering/matching would be done against those.

If the overall consensus is to just use full labels for matching, I'm not going to block it and I can approve this change. I just wanted to reiterate my take on it for the record, because I think is simpler and more intuitive.

newton/_src/utils/selection.py
```diff
@@ -432,12 +439,62 @@ def match_labels(labels: list[str], pattern: str | list[str] | list[int]) -> lis
...
+) -> tuple[list[int], bool]:
+    """Resolve ``pattern`` for callers that currently match full labels only.
+
+    Returns ``(indices, differs)``; ``differs`` is ``True`` only when ``match_full_labels``
+    is False and enabling leaf matching would change the result.
+    """
+    if match_full_labels:
+        return match_labels(labels, pattern, match_leaf=True), False
+    legacy = match_labels(labels, pattern)
+    new = match_labels(labels, pattern, match_leaf=True)
```
camevor, line 465: Matching with `fnmatch` introduces a nontrivial amount of overhead when applied to many labels, such as for the sensors, so doubling this could negatively affect startup times.

adenzler-nvidia:
@nvlukasz thanks — two things.

On the globbing: I think the `*torso` concern assumes full-label-*only* matching. What's implemented is leaf-*or*-full, so a bare part name still matches the leaf in every mode — `include_links="torso"` matches a USD `.../torso` with no `*`, today and after the default flips. The path form is only needed to disambiguate parts whose *leaves collide* (the GH-2914 case). So intuitive part-name matching doesn't regress.

On deriving cleaner names by stripping the prefix: I like it for making `*_names` unique/location-independent, but two practical snags. (1) It can't key off the articulation root — per the UsdPhysics schema, floating articulations put `ArticulationRootAPI` on the *root body*, which isn't a path ancestor of its sibling links, so we'd have to compute the longest common path-component prefix of the member labels instead. That's doable and degrades gracefully, but (2) it changes `*_names` from the leaf to a relative path for every hierarchical model, which breaks bare `joint_names.index("elbow")` / `include_links="elbow"` for the non-colliding majority — and the relative names still need a glob to match a leaf (`"*toe_roll"`), so it removes the env-location noise but not globbing. So it's really a separate, opt-in change to name derivation rather than a flag-free simplification of this one.

That said, there's no real urgency here — this wasn't driven by a concrete request, so I'd rather we converge on the design we actually want than rush anything in. If the consensus leans toward prefix-stripped names, I'm happy to go that way instead.

newton/_src/utils/selection.py
```diff
@@ -394,7 +395,7 @@ def find_matching_ids(pattern: str, labels: list[str], world_ids, world_count: i
     return grouped_ids, global_ids
```
AntoineRichard, line 395: Could this use the new method instead of a bare fnmatch? Currently this is preventing IsaacLab from matching rigid object collections properly.

AntoineRichard (changes requested):

jcarius-nv:

jcarius-nv:
Closing after discussion with @adenzler-nvidia

PR #3073, opened 2026-06-04
Add versions to deprecation directives; fix changelog wording

adenzler-nvidia:
## Description

Small documentation/changelog consistency fixes for existing deprecations — no behavior change.

**Annotate three `.. deprecated::` directives with their release version** (the project convention is `.. deprecated:: X.Y`; these omitted it):

- `SolverNotifyFlags` (`solvers/flags.py`) → `1.3`
- `max_worlds` on `Viewer.set_model()` (`viewer/viewer.py`) → `1.1`
- legacy `add_actuator(actuator_class=...)` signature (`sim/builder.py`) → `1.2`

**Fix an inaccurate CHANGELOG entry.** The `[Unreleased]` entry claimed `ModelBuilder.joint_target_pos` / `joint_target_vel` "are removed", but only the *writable* attributes were removed — the names remain as read-only deprecated aliases that emit a `DeprecationWarning`. The wording now says so.

**Add a missing CHANGELOG entry.** The `Actuator` constructor warns that its implicit default for `control_target_pos_attr` / `control_target_vel_attr` (resolving to legacy `joint_target_pos` / `joint_target_vel` under `use_coord_layout_targets=False`) will switch to `joint_target_q` / `joint_target_qd` in a future release, but this deprecation had no `### Deprecated` entry. Added one with migration guidance.

## Checklist

- [x] New or existing tests cover these changes
- [x] The documentation is up to date with these changes
- [x] `CHANGELOG.md` has been updated (if user-facing change)

## Test plan

Docstring/changelog-only change. Verified the `ruff`, `ruff format`, typos, and warp-array-syntax pre-commit hooks pass on the edited files, and that no bare `.. deprecated::` directives (without a version) remain under `newton/`.

PR #3074, opened 2026-06-04
Add deprecation directives to warned-only API symbols

adenzler-nvidia:
## Description

Documentation-only change: add `.. deprecated:: X.Y` directives to deprecated public API symbols that warn at runtime and are already listed in `CHANGELOG.md`, but whose docstrings carried no directive — so they did not appear in the rendered API deprecation index. The directives are annotated with the release in which each symbol was deprecated:

- `Control` / `Model` / `ModelBuilder` `joint_target_pos` / `joint_target_vel` → `1.3`
- `ViewerBase.update_shape_colors()` → `1.1`
- `add_shape_ellipsoid()` `a` / `b` / `c` parameter names → `1.1`
- `add_shape_gaussian()` positional `gaussian` argument → `1.1`

The last two follow the existing precedent of the `armature`-parameter deprecation on `add_link()` / `add_body()` (method-level directive whose text scopes it to the parameter).

**Intentionally left untouched** (no clean directive home, or the owning method is still supported and already annotated in its `Args:`): the USD-data names (`newton:contact_*`, `mjc:solref`, `newton:armature`), `Model.mujoco.dof_passive_damping`, and the `SolverVBD` parameter-level deprecations (`rigid_enable_dahl_friction`, `register_custom_attributes(dahl_defaults_enabled=...)`).

No behavior change; the deprecation warnings and changelog entries are unchanged.

## Checklist

## Test plan

Docstring-only change. Verified `py_compile` and the `ruff`, `ruff format`, typos, and warp-array-syntax pre-commit hooks pass on the four edited files.

## Summary by CodeRabbit

* **Documentation**
  * Enhanced deprecation notices across the simulation builder and control modules with explicit version numbers and migration paths to recommended alternatives.
  * Improved guidance for geometric shape creation methods, clarifying parameter naming conventions.
  * Updated shape color management deprecation documentation with clear migration instructions.

### DIFF
--- a/newton/_src/sim/builder.py
+++ b/newton/_src/sim/builder.py
@@ -2153,6 +2153,9 @@ class ModelBuilder:
         slot is dropped; other joints copy verbatim. Mutations do not
         propagate back. Raises :class:`AttributeError` under
         :data:`newton.use_coord_layout_targets` ``True``.
+
+        .. deprecated:: 1.3
+            Use :attr:`joint_target_q` instead.
         """
         import newton  # noqa: PLC0415
 
@@ -2176,6 +2179,9 @@ class ModelBuilder:
         fresh copy — mutations do not propagate back. Raises
         :class:`AttributeError` under
         :data:`newton.use_coord_layout_targets` ``True``.
+
+        .. deprecated:: 1.3
+            Use :attr:`joint_target_qd` instead.
         """
         import newton  # noqa: PLC0415
 
@@ -6090,6 +6096,10 @@ class ModelBuilder:
         The ellipsoid is centered at its local origin as defined by `xform`, with semi-axes
         `rx`, `ry`, `rz` along the local X, Y, Z axes respectively.
 
+        .. deprecated:: 1.1
+            The ``a``, ``b``, ``c`` parameter names are deprecated; use
+            ``rx``, ``ry``, ``rz`` instead.
+
         Note:
             Ellipsoid collision is handled by the GJK/MPR collision pipeline,
             which provides accurate collision detection for all convex shape pairs.
@@ -6520,6 +6530,10 @@ class ModelBuilder:
 
         The Gaussian is attached as a ``GeoType.GAUSSIAN`` shape for rendering.
         Collision is handled separately via *collision_proxy*.
+
+        .. deprecated:: 1.1
+            Passing ``gaussian`` as the second positional argument is
+            deprecated; pass it via the ``gaussian=`` keyword instead.
 
         Args:
             body: The index of the parent body this shape belongs to.
--- a/newton/_src/sim/control.py
+++ b/newton/_src/sim/control.py
@@ -96,6 +96,9 @@ class Control:
         """Deprecated alias for :attr:`joint_target_q` (DOF-shape only).
         Raises :class:`AttributeError` under
         :data:`newton.use_coord_layout_targets` ``True``.
+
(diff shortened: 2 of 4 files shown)


eric-heiden:
⚪ This sweep makes the deprecation docs easier to read. A couple of version and status mismatches still need to be made consistent before the docs are internally aligned.

newton/_src/sim/builder.py
```diff
@@ -6090,6 +6096,10 @@ def add_shape_ellipsoid(
         The ellipsoid is centered at its local origin as defined by `xform`, with semi-axes
         `rx`, `ry`, `rz` along the local X, Y, Z axes respectively.
 
+        .. deprecated:: 1.1
+            The ``a``, ``b``, ``c`` parameter names are deprecated; use
```
eric-heiden, line 6100: 🟡 This PR adds versioned `.. deprecated::` directives, but a few existing directives in the same touched API area still use the versionless form. That leaves the rendered docs inconsistent, with some admonitions showing `Deprecated since version 1.3` and others only showing `Deprecated`.

Could we also update these existing directives to `.. deprecated:: 1.3`?

| File | API |
| --- | --- |
| `newton/_src/sim/builder.py` | legacy `add_actuator()` `newton_actuators` signature |
| `newton/_src/viewer/viewer.py` | `set_model(max_worlds=...)` parameter |
| `newton/_src/solvers/flags.py` | `SolverNotifyFlags` |

newton/_src/sim/builder.py
```diff
@@ -2153,6 +2153,9 @@ def joint_target_pos(self) -> list[float]:
         slot is dropped; other joints copy verbatim. Mutations do not
         propagate back. Raises :class:`AttributeError` under
         :data:`newton.use_coord_layout_targets` ``True``.
+
+        .. deprecated:: 1.3
+            Use :attr:`joint_target_q` instead.
```
eric-heiden, line 2158: 🟡 This new `.. deprecated:: 1.3` directive conflicts with the `[Unreleased]` CHANGELOG wording for `ModelBuilder.joint_target_pos` and `ModelBuilder.joint_target_vel`, which says those properties are removed. The code still exposes warned read-only aliases, so users can read the CHANGELOG as saying the API is gone while the docs say it is deprecated.

The changelog should say these attributes are deprecated, not yet removed.

newton/_src/sim/builder.py
adenzler-nvidia: Good catch — these three are addressed in a companion PR, #3073, which versions exactly those directives. One correction: they should not all be `1.3`. Per their CHANGELOG sections the correct versions are `SolverNotifyFlags` → 1.3, `max_worlds` → 1.1, and the legacy `add_actuator()` signature → 1.2, and that is what #3073 uses. I scoped the directive *additions* into this PR and the directive *version fixes* into #3073; once both land the rendered docs are consistent.

newton/_src/sim/builder.py
adenzler-nvidia: This is also handled in #3073, which corrects the `[Unreleased]` CHANGELOG wording from "are removed" to "are now read-only deprecated aliases (the writable attributes are removed)", matching the `.. deprecated:: 1.3` directive added here.

eric-heiden (approved):
Thanks, I see https://github.com/newton-physics/newton/pull/3073 will add these versions to deprecation directives.

PR #3081, opened 2026-06-05
Add public API review detection

### LINKED ISSUE #3051: Add GitHub action to label PRs that have API changes
If a PR introduces API changes it should automatically receive a label.

eric-heiden:
## Description

Adds a bounded, static AST check that requests API review when a pull request changes Newton's supported public interface or when analysis becomes inconclusive.

The detector intentionally covers a narrow contract:

- direct imports, simple `__all__` declarations, and Newton's known lazy solver exports;
- functions, classes, overloads, properties, dataclasses, NamedTuples, constants, decorators, and method bindings;
- Newton solver custom attributes and frequencies, including registrations delegated to solver helper functions;
- three outcomes (`changed`, `unchanged`, and conservative `unknown`) with a capped `api-changes` comment.

The workflow runs trusted base-branch code under `pull_request_target` and treats both source trees as inert data. It never imports or executes code from an untrusted pull request.

Closes #3051

## Checklist

- [x] `CHANGELOG.md` is not required for this repository-CI-only change

## Test plan

```text
uv run --no-project -m unittest discover -s scripts/ci/tests -v
uvx pre-commit run -a
```

The detector was also replayed from each merge base against currently open pull requests:

| PR | Expected surface | Result |
| --- | --- | --- |
| #3772 | Internal XPBD fix | Unchanged |
| #3755 | Documentation only | Unchanged |
| #3775 | Public `Contacts` field | 1 addition |
| #3752 | MuJoCo custom registrations | 2 custom-attribute additions |
| #3623 | Deprecated MuJoCo API removal | Removals detected |
| #3598 | New public controllers module | Additions detected; report capped |

### DIFF
--- a/.github/workflows/pr_api_changes.yml
+++ b/.github/workflows/pr_api_changes.yml
@@ -1,0 +1,152 @@
+name: Public API Change Detection
+permissions: {}
+
+on:
+  pull_request_target:
+    branches:
+      - main
+      - "release-*"
+
+concurrency:
+  group: ${{ github.workflow }}-pr-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  # Safe under pull_request_target only because the trusted base-branch
+  # detector parses the head statically; never execute head code here.
+  detect-api-changes:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      issues: write
+    if: github.repository == 'newton-physics/newton'
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@fa2e9d605c4eeb9fcad4c99c224cee0c6c7f3594  # v2.16.0
+        with:
+          egress-policy: block
+          allowed-endpoints: >
+            api.github.com:443
+            github.com:443
+            release-assets.githubusercontent.com:443
+
+      - name: Checkout PR head
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # v6.0.1
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.event.pull_request.head.sha }}
+          path: head
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Checkout base branch history
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # v6.0.1
+        with:
+          ref: ${{ github.event.pull_request.base.sha }}
+          path: base
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Set up Python
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6.2.0
+        with:
+          python-version-file: "base/.python-version"
+
+      - name: Checkout PR merge base
+        id: merge_base
+        run: |
+          script_path="${RUNNER_TEMP}/detect_api_changes.py"
+          git -C base show "${{ github.event.pull_request.base.sha }}:scripts/ci/detect_api_changes.py" > "$script_path"
+
+          git -C base fetch --no-tags ../head HEAD:refs/remotes/pr/head
+          merge_base=$(git -C base merge-base HEAD refs/remotes/pr/head)
+          git -C base checkout --detach "$merge_base"
+
+          echo "script_path=$script_path" >> "$GITHUB_OUTPUT"
+
+      - name: Detect API changes
+        id: detect
+        env:
+          API_COMMENT_PATH: ${{ runner.temp }}/api-comment.md
+        run: |
+          output=$(python "${{ steps.merge_base.outputs.script_path }}" base head)
+          needs_review=$(printf '%s' "$output" | python -c "import sys, json; print(str(json.load(sys.stdin)['needs_review']).lower())")
+          echo "needs_review=$needs_review" >> "$GITHUB_OUTPUT"
+
+          printf '%s' "$output" | python -c "
+          import sys, json
+          data = json.load(sys.stdin)
+          print(data['comment'])
+          " > "$API_COMMENT_PATH"
+
+      - name: Sync API review
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd  # v8.0.0
+        env:
+          API_COMMENT_PATH: ${{ runner.temp }}/api-comment.md
+          NEEDS_REVIEW: ${{ steps.detect.outputs.needs_review }}
+        with:
+          script: |
+            const fs = require('fs');
+            const needsReview = process.env.NEEDS_REVIEW === 'true';
+            const repo = {
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+            };
+            const issue = {
+              ...repo,
+              issue_number: context.issue.number,
+            };
+
+            if (needsReview) {
+              await github.rest.issues.addLabels({
+                ...issue,
+                labels: ['api-changes'],
+              });
+            } else {
+              try {
+                await github.rest.issues.removeLabel({
+                  ...issue,
+                  name: 'api-changes',
+                });
+              } catch (error) {
+                // The current state is already correct when the label is absent.
+                if (error.status !== 404) throw error;
+              }
+            }
+
+            const marker = '<!-- newton-api-changes -->';
+            const comments = await github.paginate(github.rest.issues.listComments, {
+              ...issue,
+              per_page: 100,
+            });
+            const existing = comments.find((comment) =>
+              comment.user?.login === 'github-actions[bot]' &&
+              comment.user?.type === 'Bot' &&
+              comment.body?.startsWith(marker)
+            );
+
+            if (!needsReview) {
+              if (existing) {
+                await github.rest.issues.deleteComment({
+                  ...repo,
+                  comment_id: existing.id,
+                });
+              }
+              return;
+            }
+
+            const body = fs.readFileSync(process.env.API_COMMENT_PATH, 'utf8').trim();
+            if (!body) return;
+            const fullBody = `${marker}\n${body}`;
+            if (existing) {
+              await github.rest.issues.updateComment({
+                ...repo,
+                comment_id: existing.id,
+                body: fullBody,
+              });
+            } else {
+              await github.rest.issues.createComment({
+                ...issue,
+                body: fullBody,
+              });
+            }
--- a/scripts/ci/detect_api_changes.py
+++ b/scripts/ci/detect_api_changes.py
@@ -1,0 +1,809 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
+
+"""Conservatively detect Newton API changes with a small static AST model.
+
+Security invariant - DO NOT BREAK:
+    This script runs under ``pull_request_target`` while inspecting an
+    untrusted checkout. Treat both revisions as inert data: never import,
+    execute, evaluate, or spawn code from either tree. Only bounded file reads
+    and stdlib ``ast`` parsing are allowed.
+
+The result is advisory. Exact records cover ordinary declarations and the
+simple export forms used by Newton. Changed constructs outside that grammar
+produce ``unknown`` and still request API review.
+"""
+
+from __future__ import annotations
+
+import ast
+import hashlib
+import html
+import importlib.util
+import json
+import stat
+import sys
+from pathlib import Path
+
+PACKAGE = "newton"
+EXCLUDED = ("newton.examples", "newton.tests")
+KNOWN_FORWARD_ALL = {"newton.solvers": {"_solvers": "newton._src.solvers"}}
+KNOWN_LAZY_MAPS = {"newton._src.solvers": "_LAZY_IMPORTS"}
+VIRTUAL_MODULES = {"newton.solvers.experimental.coupled": "newton._src.solvers.coupled"}
+
+MAX_FILES = 2_000
+MAX_FILE_BYTES = 2 * 1024 * 1024
+MAX_TOTAL_BYTES = 32 * 1024 * 1024
+MAX_AST_NODES = 200_000
+MAX_SYMBOLS = 20_000
+MAX_EXPR_CHARS = 2_000
+MAX_REPORT_ITEMS = 20
+MAX_COMMENT_CHARS = 40_000
+
+
+class Definition:
+    def __init__(self, symbol: dict, module: str, node: ast.AST, members: dict[str, Definition] | None = None):
+        self.symbol = symbol
+        self.module = module
+        self.node = node
+        self.members = members or {}
+
+
+class Module:
+    def __init__(self, name: str, tree: ast.Module, digest: str, is_package: bool):
+        self.name = name
+        self.tree = tree
+        self.digest = digest
+        self.is_package = is_package
+        self.definitions: dict[str, Definition] = {}
+        self.imports: dict[str, tuple[str, str | None]] = {}
+        self.exports: list[str] | None = None
+        self.forwarded_exports: list[str] = []
+        self.lazy: dict[str, tuple[str, str | None]] = {}
+
+
+def _digest(value: str | bytes) -> str:
+    value = value.encode("utf-8", "backslashreplace") if isinstance(value, str) else value
+    return hashlib.sha256(value).hexdigest()
+
+
+def _note(notes: dict[str, dict[str, str]], key: str, reason: str, fingerprint: str | bytes) -> None:
+    notes[key] = {"reason": reason[:500], "fingerprint": _digest(fingerprint)}
+
+
+def _expr(node: ast.AST | None) -> str | None:
+    if node is None:
+        return None
+    try:
+        value = ast.unparse(node)
+    except Exception:
+        return "?"
+    return value if len(value) <= MAX_EXPR_CHARS else value[:MAX_EXPR_CHARS] + "..."
+
+
+def _tail(node: ast.AST) -> str:
+    if isinstance(node, ast.Name):
+        return node.id
+    if isinstance(node, ast.Attribute):
+        return node.attr
+    if isinstance(node, ast.Call):
+        return _tail(node.func)
+    return ""
+
+
+def _public_module(name: str) -> bool:
+    excluded = any(name == prefix or name.startswith(prefix + ".") for prefix in EXCLUDED)
+    return not excluded and all(not part.startswith("_") for part in name.split(".")[1:])
+
+
+def _absolute_import(module: Module, imported: str | None, level: int) -> str:
+    if not level:
+        return imported or ""
+    package = module.name if module.is_package else module.name.rpartition(".")[0]
+    return importlib.util.resolve_name("." * level + (imported or ""), package)
+
+
+def _signature(node: ast.FunctionDef | ast.AsyncFunctionDef) -> str:
+    prefix = "async " if isinstance(node, ast.AsyncFunctionDef) else ""
+    returns = f" -> {_expr(node.returns)}" if node.returns is not None else ""
+    return f"{prefix}({_expr(node.args)}){returns}"
+
+
+def _callable_definition(
+    definitions: dict[str, Definition],
+    node: ast.FunctionDef | ast.AsyncFunctionDef,
+    module: str,
+    *,
+    method: bool,
+) -> None:
+    decorators = [_expr(item) or "?" for item in node.decorator_list]
+    tails = {_tail(item) for item in node.decorator_list}
+    accessor = None
+    if "property" in tails:
+        accessor = "getter"
+    else:
+        for item in node.decorator_list:
+            if isinstance(item, ast.Attribute) and item.value and _tail(item.value) == node.name:
+                if item.attr in {"getter", "setter", "deleter"}:
+                    accessor = item.attr
+
+    existing = definitions.get(node.name)
+    if accessor:
+        symbol = dict(existing.symbol) if existing and existing.symbol["kind"] == "property" else {"kind": "property"}
+        symbol.setdefault("accessors", {})[accessor] = _signature(node)
+        symbol["decorators"] = sorted(set(symbol.get("decorators", [])) | set(decorators))
+        definitions[node.name] = Definition(symbol, module, node)
+        return
+
+    binding = "function"
+    if method:
+        binding = "classmethod" if "classmethod" in tails else "staticmethod" if "staticmethod" in tails else "instance"
+    kind = "method" if method else "function"
+    if existing and existing.symbol.get("kind") == kind:
+        symbol = dict(existing.symbol)
+    else:
+        symbol = {"kind": kind, "binding": binding, "decorators": decorators, "overloads": []}
+    if "overload" in tails:
+        symbol.setdefault("overloads", []).append(_signature(node))
+    else:
+        symbol["signature"] = _signature(node)
+        symbol["binding"] = binding
+        symbol["decorators"] = decorators
+    definitions[node.name] = Definition(symbol, module, node)
+
+
+def _namedtuple_assignment(node: ast.Assign | ast.AnnAssign, module: str) -> tuple[str, Definition] | None:
+    target = node.target if isinstance(node, ast.AnnAssign) else node.targets[0] if len(node.targets) == 1 else None
+    value = node.value
+    if not isinstance(target, ast.Name) or not isinstance(value, ast.Call) or _tail(value.func) != "NamedTuple":
+        return None
+    if len(value.args) < 2 or not isinstance(value.args[1], (ast.List, ast.Tuple)):
+        return None
+    fields = []
+    for item in value.args[1].elts:
+        if not isinstance(item, (ast.List, ast.Tuple)) or len(item.elts) != 2:
+            return None
+        name = item.elts[0].value if isinstance(item.elts[0], ast.Constant) else None
+        if not isinstance(name, str):
+            return None
+        fields.append({"name": name, "annotation": _expr(item.elts[1])})
+    return target.id, Definition({"kind": "namedtuple", "fields": fields}, module, node)
+
+
+def _assignment(definitions: dict[str, Definition], node: ast.Assign | ast.AnnAssign, module: str) -> None:
+    namedtuple = _namedtuple_assignment(node, module)
+    if namedtuple:
+        definitions[namedtuple[0]] = namedtuple[1]
+        return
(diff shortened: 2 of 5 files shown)


.github/workflows/pr_api_changes.yml
```diff
@@ -0,0 +1,146 @@
...
+          script: |
+            const fs = require('fs');
+            const body = fs.readFileSync('/tmp/api-comment.md', 'utf8').trim();
+            if (!body) return;
+
+            const marker = '<!-- newton-api-changes -->';
+            const fullBody = `${marker}\n${body}`;
+
+            // Find existing comment from this action
+            const comments = await github.rest.issues.listComments({
```
adenzler-nvidia: `github.rest.issues.listComments({...})` returns only the first page (30 comments), so on a PR with more than 30 comments the `<!-- newton-api-changes -->` comment won't be found and this step posts a duplicate instead of updating. The delete step at L134 has the same problem (stale comment never removed). Suggest paginating:

```js
const comments = await github.paginate(github.rest.issues.listComments, {
  owner: context.repo.owner,
  repo: context.repo.repo,
  issue_number: context.issue.number,
  per_page: 100,
});
const existing = comments.find(c => c.body && c.body.startsWith(marker));
```

(`github.paginate` returns the flat array, so the `.data` goes away.)

.github/scripts/detect_api_changes.py
```diff
@@ -0,0 +1,900 @@
...
+    for file_path in sorted(package_root.rglob("*.py")):
+        if "__pycache__" in file_path.parts:
+            continue
+        module_name = _module_name_from_path(workspace_root, file_path)
+        if _is_excluded(module_name):
+            continue
+        try:
+            source = file_path.read_text(encoding="utf-8")
+            tree = ast.parse(source, filename=str(file_path.relative_to(workspace_root)))
+        except (OSError, SyntaxError):
```
adenzler-nvidia: This catch (and the matching one at L244) swallows parse failures with no logging, which has two failure modes: (1) if a public module parses on base but not head, it's dropped from one side and `compare_symbols` reports every symbol it exported as removed — a confidently-wrong diff posted to the PR with nothing in the logs to explain it; (2) a head file with a null byte or invalid UTF-8 raises `ValueError`/`UnicodeDecodeError`, which isn't caught and aborts the job. Suggest broadening the catch and logging the skip:

```python
except (OSError, SyntaxError, ValueError, RecursionError) as exc:
    print(f"WARNING: skipping {file_path}: {exc}", file=sys.stderr)
    continue
```

Ideally also surface skipped modules in the output so a dropped module can't masquerade as a clean "removed" diff.

.github/workflows/pr_api_changes.yml
```diff
@@ -0,0 +1,146 @@
...
+    branches:
+      - main
+      - "release-*"
+
+concurrency:
+  group: ${{ github.workflow }}-pr-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  detect-api-changes:
```
adenzler-nvidia, line 17: Most of our workflows guard the job with `if: github.repository == 'newton-physics/newton'` (the aws_gpu and nightly ones) so they don't run on forks. Without it, this labels and comments on forks' own PRs. Suggest adding the same guard to this job.

.github/scripts/detect_api_changes.py
```diff
@@ -0,0 +1,900 @@
...
+        "summary": {
+            "added_count": len(added),
+            "removed_count": len(removed),
+            "changed_count": len(changed),
+            "total": len(added) + len(removed) + len(changed),
+        },
+    }
+
+
+def format_comment(diff: dict) -> str:
```
adenzler-nvidia: Signatures and values here come from `ast.unparse` of head source and are interpolated into the comment without escaping, so a crafted symbol value/annotation could embed markdown or HTML (`</details>`, ``, …) into a comment posted by the `github-actions` bot. Not code execution — it's passed as a `body` param and the embedded marker can't hijack the `startsWith(marker)` lookup — but worth closing by wrapping the interpolated fields in a small escaper or rendering values inside fenced code blocks.

adenzler-nvidia (commented):
Nice approach — running the analyzer from the trusted `base/` checkout and restricting it to `ast` parsing (never importing or executing head code) is the right way to handle untrusted PR content under `pull_request_target`, and the `persist-credentials: false` + egress `block` hardening is tighter than our other PRT workflows. Left a few inline suggestions: comment pagination, parse-failure handling, a fork guard, and escaping the comment body.

One non-blocking thought on the trigger: applying the label does need a write token (a fork `pull_request` run only gets a read-only one), so `pull_request_target` is justified. The part I'd flag is that this is the only PRT workflow that runs a full checkout + parse of head code in the *same* job that holds the write token — `actions/labeler` avoids that by only reading the changed-file list, and our aws_gpu workflows keep the privileged step away from head code. If we want to keep the symbol-level detail, the usual shape is a `pull_request` job that parses head and uploads an artifact plus a small `workflow_run` job that labels/comments from it — same feature, token never next to head code. Not blocking; fine to keep the current shape with an explicit "never execute head code" comment if you'd rather. Curious what you think.

.github/workflows/pr_api_changes.yml
eric-heiden: Addressed in e28166185. Both the create/update path and the delete path now use github.paginate(..., per_page: 100), and the marker lookup is scoped to the github-actions bot before updating or deleting the comment.

.github/workflows/pr_api_changes.yml
eric-heiden: Addressed in e28166185. The privileged detect-api-changes job now has the repository guard, so the label/comment path only runs for newton-physics/newton. The new author-membership and manual-approval gates sit in front of that job as well.

adenzler-nvidia (dismissed):
LGTM

scripts/ci/tests/test_detect_api_changes.py
shi-eric, line 1: Are you sure it's a good idea to have a unit test in `newton/tests` that is testing CI/CD tooling outside the `/newton` directory rather than `newton` library functionality? I see the meta-test `newton/tests/test_generate_api.py` already slipped through the cracks, but I would have pushed back against that as well.

scripts/ci/detect_api_changes.py
shi-eric, line 1: Can this be placed in the existing `/scripts/ci` directory instead?

shi-eric (changes requested):

.github/scripts/detect_api_changes.py
```diff
@@ -0,0 +1,1397 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
+
+"""Detect public API changes between two revisions using static AST analysis.
+
+This script performs purely static analysis (no code execution) to extract the
+public API surface from the Newton package and compare two revisions. It outputs
+a JSON report describing added, removed, and modified public symbols.
+
+Safety: only stdlib ``ast`` parsing is used; no user code is imported or executed.
```
shi-eric: The `Safety:` note here describes current behavior, but this file is what the entire `pull_request_target` workflow's security model rests on. It's worth a comment that tells a future editor what *not* to do and why, especially if the membership/approval gates get removed (see my comment on `pr_api_changes.yml`).

The workflow runs under `pull_request_target` with a privileged `GITHUB_TOKEN` (`issues: write`) and checks out untrusted fork-PR code for this script to inspect. That's only safe because this script treats both revisions as inert data, reading files as text and parsing with stdlib `ast`, never importing or executing them. If someone later switches to `importlib`/runtime introspection of the target package, attacker-controlled code runs with the privileged token, which is an RCE.

Suggest expanding the note into an explicit invariant, e.g.:

```python
Security invariant - DO NOT BREAK:
    Runs under ``pull_request_target`` with a privileged token against
    untrusted fork PRs. Safe ONLY because both revisions are treated as
    inert data (read as text, parsed with stdlib ``ast``). Never import,
    exec, eval, runpy, or subprocess the analyzed packages, and never
    import the target tree to introspect it at runtime. Doing so executes
    attacker-controlled code with the privileged token (RCE).
```

.github/workflows/pr_api_changes.yml
```diff
@@ -0,0 +1,227 @@
...
+      name: external-pr-approval
+      url: ${{ github.event.pull_request.html_url }}
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@fa2e9d605c4eeb9fcad4c99c224cee0c6c7f3594  # v2.16.0
+        with:
+          egress-policy: audit
+
+      - name: Approval granted
+        run: echo "Manual approval granted for external PR"
```
shi-eric: I think the `check-author-membership` and `require-approval` jobs can be removed, leaving just `detect-api-changes` gated on the repository guard.

The usual reason to gate a `pull_request_target` workflow behind membership/approval is to avoid either (a) burning expensive runners on untrusted PRs, or (b) executing attacker-controlled code with the privileged token. Neither applies here:

- **No paid runners.** All jobs run on `ubuntu-latest` (GitHub-hosted standard tier), so there's no AWS/GPU cost to protect against.
- **No untrusted code execution.** The job runs the *base-branch* copy of the detector (extracted via `git show ${base.sha}:...`), and that script only statically parses the head tree with stdlib `ast`. It never imports or executes head code. The only privilege in scope is `GITHUB_TOKEN` with `issues: write` + `contents: read`, with harden-runner egress already blocked to an allowlist.

So the approval gate adds friction for every external contributor (a maintainer has to click approve before the API-changes label/comment appears) without protecting against a real threat.

Worth keeping: the `github.repository == 'newton-physics/newton'` guard and the harden-runner `egress: block` on the detect job.

One caveat if these gates come out: the safety then rests entirely on two invariants holding, namely that the job runs the base-branch script and that the script only statically parses head. I left a comment on `detect_api_changes.py` suggesting we document the second one in the script itself. A one-line pointer here would help too, e.g.:

```yaml
# Safe under pull_request_target only because detect_api_changes.py parses
# head statically and is run from the base branch; never execute head code here.
```

.github/workflows/pr_api_changes.yml
```diff
@@ -0,0 +1,227 @@
...
+              per_page: 100,
+            });
+            const existing = comments.find(isApiChangesComment);
+            if (existing) {
+              await github.rest.issues.deleteComment({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                comment_id: existing.id,
+              });
+            }
```
shi-eric: Both directions are necessary (a PR can gain API changes, or revert them after a comment was already posted), but these two steps are really one operation: reconcile the comment to match current state. They duplicate the marker, the `isApiChangesComment` predicate, and the `listComments` lookup, and each makes its own `listComments` call.

Suggest collapsing them into a single step that lists once and branches on `has_changes`. Instead of deleting when there are no changes, update the existing comment to a "no changes" state, and only touch a comment that already exists so PRs that never had API changes stay comment-free:

```yaml
- name: Sync API change comment
  uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd  # v8.0.0
  env:
    API_COMMENT_PATH: ${{ runner.temp }}/api-comment.md
    HAS_CHANGES: ${{ steps.detect.outputs.has_changes }}
  with:
    script: |
      const fs = require('fs');
      const marker = '<!-- newton-api-changes -->';
      const isApiChangesComment = (c) =>
        c.user?.login === 'github-actions[bot]' &&
        c.user?.type === 'Bot' &&
        c.body?.startsWith(marker);

      const comments = await github.paginate(github.rest.issues.listComments, {
        owner: context.repo.owner, repo: context.repo.repo,
        issue_number: context.issue.number, per_page: 100,
      });
      const existing = comments.find(isApiChangesComment);
      const params = { owner: context.repo.owner, repo: context.repo.repo };

      let body;
      if (process.env.HAS_CHANGES === 'true') {
        body = fs.readFileSync(process.env.API_COMMENT_PATH, 'utf8').trim();
        if (!body) return;
      } else {
        // No changes: flip an existing comment to a "no changes" state,
        // but don't create one on PRs that never touched the API.
        if (!existing) return;
        body = 'No public API changes detected.';
      }

      const fullBody = `${marker}\n${body}`;
      if (existing) {
        await github.rest.issues.updateComment({ ...params, comment_id: existing.id, body: fullBody });
      } else {
        await github.rest.issues.createComment({ ...params, issue_number: context.issue.number, body: fullBody });
      }
```

Net: two steps become one, a single `listComments` call, and the marker/predicate defined once. The comment now stays as a stable anchor and flips to "no changes" rather than disappearing.

The label pair just above ("Add api-changes label" / "Remove api-changes label if no changes") has the identical conditional-pair shape and can be merged the same way for consistency.

scripts/ci/tests/test_detect_api_changes.py
adenzler-nvidia: I didn't pay enough attention there, and that test already didn't work on a pure pip install so I added a workaround. But yes I guess we should either move those to a separate test directory, or just remove them again.

.github/workflows/pr_api_changes.yml
shi-eric: Please wire analysis warnings into the PR comment path.

The detector emits `has_analysis_warnings` and can generate a warning-only comment when modules are skipped. Today the workflow only exposes `diff.has_changes`, so a parse/decode warning with no API diff sets `has_changes=false`. That skips `Post or update comment` and can delete the existing bot comment.

Requested behavior:
- promote `has_analysis_warnings` to a detect-step output
- post/update the API comment when either `has_changes` or `has_analysis_warnings` is true
- delete or switch to the no-change state only when both are false
- keep the `api-changes` label tied to `has_changes` only

scripts/ci/tests/test_detect_api_changes.py
eric-heiden: Done. I moved these tests to `scripts/ci/tests` and added them to the CI workflow.

scripts/ci/detect_api_changes.py
eric-heiden: Done. The detector now lives at `scripts/ci/detect_api_changes.py`.

.github/scripts/detect_api_changes.py
eric-heiden: Done. I added an explicit security invariant that forbids importing or executing PR code.

.github/workflows/pr_api_changes.yml
eric-heiden: Done. I removed the membership and approval gates. The repository guard and hardened runner remain.

.github/workflows/pr_api_changes.yml
eric-heiden: Done. Labels and comments are now reconciled in single steps. Warning-only results keep the comment, while the label still follows API changes only.

scripts/ci/detect_api_changes.py
```diff
@@ -0,0 +1,2086 @@
...
+    skipped_modules: list[dict],
+    skipped_module_names: set[str],
+) -> ast.Module | None:
+    module_name = _module_name_from_path(workspace_root, file_path)
+    if module_name in skipped_module_names:
+        return None
+
+    relative_path = file_path.relative_to(workspace_root)
+    try:
+        source = file_path.read_text(encoding="utf-8")
```
shi-eric: A reproduction against the current PR head shows that a `newton/escaped.py` symlink to a Python file outside the checkout is picked up by `rglob()` and parsed as `newton.escaped`. A `newton/zero.py` link to `/dev/zero` is also selected, and `read_text()` would follow it without a bound. Since `head` comes from an untrusted fork in a `pull_request_target` job, an untrusted PR author could hang or OOM the privileged job.

The detector should reject symlinks and non-regular files, verify that resolved paths remain under the workspace, and cap file sizes before reading. A regression test can use a finite external target and a device-file candidate without attempting the unbounded read.

scripts/ci/detect_api_changes.py
```diff
@@ -0,0 +1,2086 @@
...
+    return symbols
+
+
+def _collect_definitions(tree: ast.Module, module_name: str, defs: dict[str, dict]) -> None:
+    """Collect all top-level definitions from a module (for re-export resolution)."""
+    for node in tree.body:
+        if isinstance(node, ast.ClassDef):
+            _collect_class_defs(node, node.name, defs)
+        elif isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef)):
+            defs[node.name] = _make_symbol("function", _format_callable_signature(node))
```
shi-eric: A focused reproduction confirms this false negative. The collector stores every definition under `defs[node.name]`, so each overload declaration is overwritten and only the implementation signature survives. The current snapshot for `newton.utils.create_cable_stiffness_from_elastic_moduli` contains the broad implementation signature but none of its three overloads. Changing an overload in a minimal fixture leaves `has_changes` false.

The callable metadata should retain and compare the ordered overload signatures. The same issue affects overloads exposed through `newton.usd`.

scripts/ci/detect_api_changes.py
```diff
@@ -0,0 +1,2086 @@
...
+        bases=bases,
+        class_name=class_name,
+    )
+
+    for child in node.body:
+        if isinstance(child, (ast.FunctionDef, ast.AsyncFunctionDef)):
+            if not _is_public_name(child.name) and child.name != "__init__":
+                continue
+            method_key = f"{class_name}.{child.name}"
+            defs[method_key] = _make_symbol(
```
shi-eric: A focused reproduction confirms the same false negative for properties. The getter and setter both use `Class.name` as their key, so the later setter replaces the getter. In the current snapshot, `newton.Mesh.color` is recorded only as `color(self: Mesh, value: Vec3 | None)`. Changing the getter return type alone leaves `has_changes` false.

The snapshot should store getter, setter, and deleter signatures separately, or represent them together as one property symbol. `Mesh.texture` and `ViewerBase.camera_speed` have the same structure.

shi-eric (commented):
Three confirmed P1 findings are noted inline.

shi-eric:
`detect_api_changes.py` is now 2,086 lines with 101 top-level helper functions. The first version was 642 lines. Much of the file now evaluates exports, reconstructs re-export graphs, emulates dynamic modules, and reconciles inherited or delegated attributes. That seems out of proportion to the original goal of applying an advisory `api-changes` label.

The remaining gaps suggest that the current approach will keep accumulating special cases. Overloads, read/write properties, NamedTuples, binding decorators, referenced defaults, and deprecated dynamic exports all need additional handling. Continuing in this direction means maintaining a partial Python static analyzer without being able to model every valid API pattern reliably.

Please consider narrowing the detector's requirements instead of adding more special cases:

- Follow direct imports, simple `__all__` declarations, and a small set of known Newton export maps.
- Record common declaration forms such as functions, classes, overloads, properties, dataclasses, and NamedTuples.
- When dynamic exports, delegation, or inheritance cannot be resolved, apply the label conservatively instead of emulating more Python behavior.
- Produce a short, capped report rather than reconstructing and rendering the entire API surface.

The repository's `.claude/skills/release-audit` skill already provides a place for detailed release-wide API and changelog review, including signature comparisons and deprecation checks. The PR check only needs to draw reviewer attention, so some false positives seem like a reasonable tradeoff for a much smaller deterministic implementation.

None of this is intended as a hard requirement or a blocker for merging this PR. The suggestion is to reconsider the detector's requirements and goals before investing further in special-case coverage. Please consider simplifying the design and accepting some conservative false positives in exchange for a smaller and more maintainable implementation.

jcarius-nv (commented):
I agree with Eric Shi's broader concern that the current detector is becoming a partial Python static analyzer for what is ultimately an advisory label. I do not think a broad changed-file rule is sufficient either: labeling every PR that touches production Python would produce enough false positives to make the label unhelpful.

A useful middle ground may be to make Newton's interface an explicit generated artifact. `docs/generate_api.py` already owns public API discovery, and the PR docs job already rejects stale generated output. It could emit a deterministic checked-in inventory containing qualified names, kinds, bindings, normalized signatures, overloads, and source/provider ownership. That would create a reviewable interface/implementation seam without immediately splitting every public class across source files. I wrote up the design and CI flow here: https://gist.github.com/jcarius-nv/a2bf97d95199430bda63ec0a057d5aa3

Some false positives are acceptable if the label means "API review needed" rather than "API change proven." Exact additions, removals, compatibility breaks, deprecations, and migration guidance can remain part of `release-audit`, as Eric suggested.

I also researched [Griffe](https://mkdocstrings.github.io/griffe/) as an alternative to owning all of the AST machinery here. It is actively maintained and models aliases, `__all__`, binding decorators, overloads, property accessors, signatures, and dataclass constructors with runtime inspection disabled. Pinned Griffe 2.1.0 correctly modeled `SolverBase.register_custom_attributes` as a `classmethod`, including its decorator and `cls` parameter.

Griffe is not a drop-in replacement: [`griffe check`](https://mkdocstrings.github.io/griffe/guide/users/checking/) focuses on breaking changes, its checker does not compare every decorator/binding or overload-set change, and the static loader did not expand Newton's dynamic `newton.solvers.__all__` facade. However, a small Newton normalization adapter over Griffe could populate the same inventory contract, outsource most of the Python semantic model, and conservatively label unresolved cases.

Would you be open to comparing two narrowed prototypes against a representative set of historical Newton PRs: (1) a generated interface inventory extending `docs/generate_api.py`, and (2) the same inventory populated by a small Griffe-backed adapter? The comparison should measure false positives and false negatives, especially for lazy exports, overloads, properties, dataclasses, defaults, and method-binding changes, before committing to the current custom analyzer.

shi-eric:
Note: In the future things like this I _hope_ are more easily maintained as a set of instructions describing a workflow for a Claude code action.

eric-heiden:
I compared three approaches: Griffe, Jan’s generated inventory, and a small custom AST detector.
Griffe handles standard Python APIs well, but still needs Newton-specific logic and produced false positives for lazy solver exports. The generated inventory was accurate but added a large checked-in file (ca. 1.28 MB/42,000-line checked-in JSON artifact) and more CI complexity.
I chose the lean AST approach. It has no new dependencies, handles common Python APIs, and detects solver custom attributes and frequencies—including registrations in helper functions. Unsupported patterns conservatively request API review.
Testing against historical and open PRs correctly identified public API changes, including MuJoCo custom registrations, while ignoring documentation-only and internal implementation changes.
The PR now has 1,274 lines, down from 3,879 lines.

scripts/ci/detect_api_changes.py
```diff
@@ -0,0 +1,736 @@
...
+    symbols[path] = definition.symbol
+    if definition.symbol["kind"] in {"class", "namedtuple"}:
+        for base in definition.symbol.get("bases", []):
+            if base in {"object", "Enum", "IntEnum", "StrEnum", "Flag", "IntFlag", "NamedTuple"}:
+                continue
+            found = _lookup(modules, definition.module, base) if base.isidentifier() else None
+            fingerprint = modules[found[1].module].digest if found and found[0] == "definition" else base
+            _note(notes, f"inheritance:{path}:{base}", f"inherited members of {path} are not expanded", fingerprint)
+    for name, member in definition.members.items():
+        if not name.startswith("_"):
```
jcarius-nv, line 483: **P1 — Public constructor changes are missed.** `_class_definition()` records `__init__` and `__new__` as members, but this filter then discards them with every other underscore-prefixed name. I reproduced changing `__init__(self, value: int)` to `__init__(self, value: str, mode: bool = False)` and the detector returned `unchanged`. This affects common public classes such as `newton.ModelBuilder`. Could we record `__init__`/`__new__` signatures on the class and add a regression test proving that a constructor change sets `needs_review=True`?

scripts/ci/detect_api_changes.py
```diff
@@ -0,0 +1,736 @@
...
+            target, value, append = node.target.id, node.value, True
+        if target != "__all__" or value is None:
+            continue
+        parsed = _literal_exports(value, module)
+        if parsed is None:
+            _note(
+                notes,
+                f"dynamic-export:{module.name}",
+                f"{module.name} uses an unsupported __all__ expression",
+                ast.dump(value),
```
jcarius-nv: **P2 — Unsupported indirection is not always classified as `unknown`.** This fingerprints only the unsupported expression itself. With `_EXPORTS = ["Foo"]; __all__ = _EXPORTS` changed to `_EXPORTS = ["Bar"]`, the fingerprint remains `Name('_EXPORTS')`; my base/head test returned `unchanged` with no analysis warning. Unsupported registration helpers have an analogous gap when they contain no recognized record. Since some false positives are acceptable for this advisory label, could we fingerprint the relevant containing module/provider (or traced dependencies) so changes outside the supported grammar conservatively produce `unknown`?

jcarius-nv (commented):
Eric, thank you for reconsidering the approach and substantially simplifying it. The narrower, bounded detector with conservative `unknown` handling addresses the architectural concern I raised. I found two remaining false-negative paths that seem worth closing before approval.

scripts/ci/detect_api_changes.py
eric-heiden: Fixed. The detector now checks `__init__` and `__new__` signatures, so constructor changes request API review. I also added a regression test.

scripts/ci/detect_api_changes.py
eric-heiden: Fixed. Unsupported `__all__` expressions now track changes in the module, and delegated registration calls track their helper definitions. These changes now return `unknown`. I added regression tests for both cases.

jcarius-nv (commented):
One remaining concern before approval: `_load_modules()` calls `package.is_dir()` and then `package.rglob("*.py")` before validating the individual paths. Because `is_dir()` follows symlinks, an untrusted fork can replace the top-level `newton/` directory with a symlink to a large or pathological runner directory. The per-file checks prevent out-of-tree reads, but only after traversal has already happened, so the `pull_request_target` job can still be exhausted or timed out before the configured limits apply.

I reproduced the detector traversing a directory outside the checkout through a top-level `newton/` symlink. Could we `lstat()` the package path itself and reject a symlink, non-directory, or out-of-root path before calling `rglob()`? A focused regression test for a symlinked package directory would close this.

Otherwise, the narrowed design and the latest fixes look good, and this is ready to go from my side once this concern is addressed.

eric-heiden:
Fixed. The detector now checks the top-level `newton/` directory before traversal and rejects symlinks, non-directories, and paths outside the checkout. I added a regression test that confirms traversal never starts for a symlinked package.

jcarius-nv (approved):
Re-reviewed at `09a9887f`. The package-root symlink concern is fixed: validation now happens before `rglob()`, and the regression test proves traversal does not start for a symlinked package. The focused suite passes 19/19, all review threads are resolved, and all checks are green. I have no remaining technical, Spec, or Newton Fit concerns. Ready to go from my side.

shi-eric (approved):

PR #3083, opened 2026-06-05
Skip generate_api test outside source checkouts

adenzler-nvidia:
## What

`test_generate_api` imports `docs.generate_api`, but the `docs` package lives at the repository root and is not shipped in installed/wheel builds. Running the bundled test suite from an installed package (e.g. `python -m unittest` against an installed `newton`) fails at import time with `ModuleNotFoundError: No module named 'docs'`.

## Change

Guard the import and skip `TestGenerateApiCopyright` when the top-level `docs` package is unavailable, following the optional-import pattern already used in the test suite (e.g. `test_download_assets.py`). Import errors other than a missing `docs` package are re-raised, so a genuinely broken `generate_api` is still surfaced rather than silently skipped.

Verified that an installed/non-source run now reports these tests as skipped instead of erroring, while a source checkout continues to run and pass them.

## Summary by CodeRabbit

* **Tests**
  * Improved test resilience by implementing conditional import handling. Tests now gracefully skip when optional dependencies are unavailable, reducing failures in non-source installation contexts while still catching genuine import errors.

### DIFF
--- a/newton/tests/test_generate_api.py
+++ b/newton/tests/test_generate_api.py
@@ -7,9 +7,19 @@ from pathlib import Path
 from types import SimpleNamespace
 from unittest import mock
 
-from docs import generate_api
+try:
+    from docs import generate_api
+except ModuleNotFoundError as exc:
+    # The ``docs`` package lives at the repository root and is not included in
+    # installed/wheel builds, so these tests only run from a source checkout.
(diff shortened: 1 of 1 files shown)


camevor (approved):
Thanks, lgtm!

jcarius-nv (approved):

PR #3088, opened 2026-06-05
Render tiled-camera heightfields via the mesh path

adenzler-nvidia:
## Summary

`SensorTiledCamera` did not render heightfield (`HFIELD`) shapes, and #2971 introduced a tiled-camera render-performance regression. Both stem from the same gap.

#2971 added an `HFIELD` ray-intersection branch to the three tiled-camera render kernels, but heightfields were never wired into the render BVH: `is_supported_shape_type`, `compute_shape_local_bounds`, and `compute_shape_bvh_bounds` (all in `geometry/bvh.py`) omitted `HFIELD`, so heightfields never entered the visible-shape BVH and never reached the renderer. The added branch was therefore unreachable, yet it still enlarged the color render megakernel's register footprint, dropping it below full GPU occupancy and regressing tiled-camera rendering.

## Fix

Heightfields are already triangulated `wp.Mesh` shapes — their mesh is stored in `shape_source_ptr` and they carry no per-vertex normals — so they can be rendered through the existing mesh path rather than a dedicated branch:

- Add `HFIELD` to the render-BVH inclusion (`is_supported_shape_type`) and to both the local- and world-bounds kernels, so heightfields enter the visible-shape BVH with correct AABBs.
- `RenderContext` presents heightfields to the render kernels as `MESH` (via a per-shape render-type array passed to the kernel). `model.shape_type` is left as `HFIELD` for collision and BVH bounds.
- The `HFIELD` branch is removed from the render kernels entirely. The render kernel is now identical to its pre-#2971 form, so it regains full occupancy while gaining heightfield support, with no per-scene kernel specialization.

## Tests

Adds `newton/tests/test_sensor_tiled_camera_heightfield.py`, which renders a flat heightfield from above and verifies the depth is geometrically correct. The existing tiled-camera, heightfield-collision, and raycast suites continue to pass.

## Release Notes

* **Bug Fixes**
  * Fixed SensorTiledCamera rendering for heightfield geometry so heightfields now appear correctly.
  * Restored tiled-camera render performance to previous levels, eliminating a regression.
  * Improved depth-hit reliability and accuracy when rendering heightfield surfaces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,7 @@
 
 ### Fixed
 
+- Fix `SensorTiledCamera` not rendering heightfield (`HFIELD`) shapes, which were missing from the render BVH. Heightfields are now rendered through the existing mesh path (they are triangulated `wp.Mesh` shapes), which also resolves a tiled-camera render-performance regression caused by the unused heightfield branch lowering the render kernel's GPU occupancy.
 - Fix `eval_fk()` overwriting VBD-simulated `JointType.CABLE` body poses.
 - Fix hydroelastic SDF contact surfaces dropping the central region under deep interpenetration. The broadphase used to skip subgrids whose centers were deeper than the SDF narrow band, leaving a hole in the contact patch when overlap exceeded the narrow-band thickness. Broadphase now visits every subgrid in the SDF coarse grid (block coordinates are derived arithmetically from per-shape SDF coarse-texture dimensions); sampling at far-inside locations is correct because the coarse SDF is dense and accurate everywhere. On-disk SDF caches written by earlier versions are transparently re-cooked on first load (`_sdf_cache.CACHE_FORMAT_VERSION` bumped to `2`)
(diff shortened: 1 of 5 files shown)


adenzler-nvidia:
[benchmarks](https://newton-physics.github.io/newton-asv/#simulation.bench_sensor_tiled_camera.SensorTiledCameraBenchmark.time_rendering_pixel_priority_color_only?arch=AMD64&cpu=Intel(R)%20Core(TM)%20Ultra%209%20285K%20(24%20physical%20cores%2C%2024%20logical)&gpu=NVIDIA%20NVIDIA%20RTX%20PRO%206000%20Blackwell%20Workstation%20Edition&machine=newton-benchmark-machine-d773f8a4&num_cpu=24&os=Windows%2011%20version%2010.0.26200&ram=64GB&python=3.13&p-resolution=64&p-world_count=4096&p-iterations=50&commits=a8979dd8-bf5f3fbd) one instance of the regression

StafaH (approved):

PR #3093, opened 2026-06-05
Fix CPU ViewerGL frame capture

mehdiataei:
## Description

Fix `ViewerGL.get_frame()` crashing when `ViewerGL` follows a CPU model while a CUDA context is active.

The previous implementation always registered the OpenGL pixel buffer object (PBO) with CUDA. When the viewer device was CPU, this could expose a CUDA pointer as a CPU Warp array. The subsequent CPU kernel then dereferenced that GPU pointer and caused a segmentation fault.

This change avoids CUDA-OpenGL registration when `ViewerGL` follows a CPU model. CPU viewers read the PBO into host memory with `glGetBufferSubData()`, while CUDA viewers retain the existing zero-copy CUDA-OpenGL path.

## Checklist

## Test plan

The regression test was verified to fail on `origin/main` and pass with this fix.

```bash
uv run --extra dev -m newton.tests -k test_viewer_get_frame
uvx pre-commit run -a
```

The full test suite passed:

```text
Ran 4629 tests
OK (skipped=140)
```

## Bug fix

**Steps to reproduce:**

1. Create a Newton model on the CPU.
2. Attach a headless `ViewerGL` to the model.
3. Make a CUDA context current.
4. Render a frame and call `ViewerGL.get_frame()`.
5. Without this fix, the process terminates with a segmentation fault.

**Minimal reproduction:**

```python
import newton
import newton.viewer
import warp as wp

builder.add_particle(pos=(0.0, 0.0, 0.0), vel=(0.0, 0.0, 0.0), mass=1.0)

viewer = newton.viewer.ViewerGL(width=64, height=64, headless=True)
try:
    viewer.set_model(model)

    # Make a CUDA context current while the viewer follows a CPU model.
    wp.zeros(1, dtype=wp.float32, device="cuda:0")

    viewer.begin_frame(0.0)
    viewer.log_state(model.state())
    viewer.end_frame()
    frame = viewer.get_frame()
    print(frame.device, frame.shape)
finally:
    viewer.close()
```

Run with:

```bash
uv run --extra examples python repro.py
```

With this fix, the frame remains on the viewer's CPU device:

```text
cpu (64, 64, 3)
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented a crash during headless frame capture when rendering CPU-based models with an active CUDA context, improving reliability across CPU/CUDA combinations.

* **Documentation**
  * Updated headless frame-capture guidance to clarify that `ViewerGL.get_frame()` returns a Warp array on the viewer device.

* **Tests**
  * Expanded frame-capture coverage to validate CPU and CUDA behaviors, including correct output content and writing into provided target tensors.


newton/tests/test_viewer_get_frame.py
eric-heiden, line 1: Could we make this a small headless `ViewerGL` regression instead of only testing a fake GL interface? The bug depends on `ViewerGL.__init__`, a real framebuffer/PBO, `set_model()` switching the viewer to a CPU model while a CUDA context may already exist, and `get_frame()` writing either a new image or a caller-provided `target_image`. Consider building a tiny CPU model, rendering one 64x48 headless frame with the real renderer, asserting the returned `wp.uint8` CPU image has the expected shape and nontrivial pixels, then calling `get_frame(target_image=...)` and repeating after `_invalidate_pbo()` or a real resize hook.

Sketch:

```python
if wp.is_cuda_available():
    wp.zeros(1, device="cuda:0")  # make CUDA context active

try:
    viewer = newton.viewer.ViewerGL(width=64, height=48, headless=True)
except Exception as exc:
    self.skipTest(f"ViewerGL not available: {exc}")

try:
    body = builder.add_body(xform=wp.transform(wp.vec3(0, 0, 0), wp.quat_identity()))
    builder.add_shape_box(body, hx=0.25, hy=0.25, hz=0.25, color=(1.0, 0.0, 0.0))

    viewer.set_model(model)
    self.assertEqual(viewer.device, wp.get_device("cpu"))

    viewer.set_camera(pos=wp.vec3(2, -3, 2), pitch=-25.0, yaw=35.0)
    viewer.begin_frame(0.0)
    viewer.log_state(model.state())
    viewer.end_frame()

    frame = viewer.get_frame()
    self.assertEqual(frame.shape, (48, 64, 3))
    self.assertEqual(frame.dtype, wp.uint8)
    self.assertEqual(frame.device, wp.get_device("cpu"))
    self.assertGreater(np.ptp(frame.numpy()), 0)

    target = wp.empty(shape=(48, 64, 3), dtype=wp.uint8, device="cpu")
    self.assertIs(viewer.get_frame(target_image=target), target)

    viewer._invalidate_pbo()
    self.assertEqual(viewer.get_frame().shape, (48, 64, 3))
finally:
    viewer.close()
```

newton/_src/viewer/viewer_gl.py
```diff
@@ -1658,12 +1662,27 @@ def get_frame(self, target_image: wp.array | None = None, render_ui: bool = Fals
...
+            )
+
         gl.glBindBuffer(gl.GL_PIXEL_PACK_BUFFER, 0)
         gl.glBindFramebuffer(gl.GL_FRAMEBUFFER, 0)
 
-        # Map PBO buffer and copy using RGB kernel.
-        assert self._wp_pbo is not None
-        buf = self._wp_pbo.map(dtype=wp.uint8, shape=(w * h * 3,))
+        if self.device.is_cuda:
+            assert self._wp_pbo is not None
```
eric-heiden, line 1808: 🔴 This assert can fail after the viewer changes device because the PBO resources are only registered inside the allocation block. A CPU capture leaves `_pbo` non-null while `_wp_pbo` is `None`; if `set_model()` later moves `self.device` to CUDA, `get_frame()` skips allocation and reaches this branch with no CUDA registration. The same lifetime issue can also reuse a `RegisteredGLBuffer` from a previous CUDA device.

Please tie `_pbo`, `_wp_pbo`, and `_pbo_host_buffer` to the current viewer device, either by invalidating them from `set_model()` when `self.device` changes or by lazily registering/re-registering the existing GL buffer before mapping it here. A regression test that captures once on CPU, switches to a CUDA-backed model path, and captures again would cover the failure mode.

jcarius-nv:

adenzler-nvidia:
Could you please update this PR with the latest `main` and resolve the merge conflicts?

mehdiataei:
@adenzler-nvidia Done

eric-heiden (approved):
Thanks!

PR #3095, opened 2026-06-08
Add contact force handling in Kamino to Newton conversion

aserifi:
## Description

`SolverKamino.update_contacts()` converted Kamino's internal contacts to a `newton.Contacts` object but dropped the per-contact forces, so `newton.sensors.SensorContact` reported zero `total_force` under `SolverKamino` (it only worked for MuJoCo). This fills in the forces so the contact sensor works identically across backends.

### Changes
- Populate `Contacts.force` in `convert_contacts_kamino_to_newton`: each contact reaction (local frame) is rotated to world frame and written as the force on `shape0`, matching `SensorContact`'s sign convention.
- Inactive contacts contribute zero; the write is guarded so it's a no-op when `Contacts.force` isn't allocated.
- The reaction is the PADMM impulse and is forwarded verbatim (no `1/dt` rescaling), so values match the existing per-body contact aggregation.
- Add `TestSensorContactKamino` (box-on-plane) asserting `total_force` is non-zero and equals `ContactAggregation.body_net_force` under `SolverKamino`.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Contact-force output is now available for Kamino solver contacts, with automatic world-frame transformation based on contact geometry.

* **Tests**
  * Added regression test to verify contact forces are correctly converted and numerically match solver aggregation results.

### DIFF
--- a/newton/_src/solvers/kamino/_src/geometry/contacts.py
+++ b/newton/_src/solvers/kamino/_src/geometry/contacts.py
@@ -897,6 +897,9 @@ def _convert_contacts_kamino_to_newton(
     kamino_position_A: wp.array[vec3f],
     kamino_position_B: wp.array[vec3f],
     kamino_gapfunc: wp.array[vec4f],
+    kamino_frame: wp.array[quatf],
+    kamino_reaction: wp.array[vec3f],
+    kamino_mode: wp.array[int32],
     shape_body: wp.array[int32],
     body_q: wp.array[wp.transformf],
     # outputs
@@ -906,6 +909,7 @@ def _convert_contacts_kamino_to_newton(
     rigid_contact_point0: wp.array[vec3f],
     rigid_contact_point1: wp.array[vec3f],
     rigid_contact_normal: wp.array[vec3f],
+    rigid_contact_force: wp.array[wp.spatial_vector],
 ):
     """Converts Kamino's internal contact representation to Newton's Contacts format."""
     # Retrieve the contact index for this thread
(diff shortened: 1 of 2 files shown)


newton/_src/solvers/kamino/_src/geometry/contacts.py
```diff
@@ -953,6 +957,13 @@ def _convert_contacts_kamino_to_newton(
     rigid_contact_point0[cid] = wp.transform_point(X_inv_a, position_A)
     rigid_contact_point1[cid] = wp.transform_point(X_inv_b, position_B)
 
+    # Force on shape0 (world frame); zero for inactive contacts.
+    if rigid_contact_force:
+        force_world = vec3f(0.0, 0.0, 0.0)
+        if kamino_mode[cid] != ContactMode.INACTIVE:
+            force_world = wp.quat_rotate(kamino_frame[cid], kamino_reaction[cid])
+        rigid_contact_force[cid] = wp.spatial_vector(-force_world, vec3f(0.0, 0.0, 0.0))
```
Guirec-Maloisel, line 965: This looks wrong to me, the contact force in Newton is expressed relative to the CoM, and in Kamino it is expressed in the contact frame. So the torque component of the contact force in Newton cannot be always zero (would mean that all contact frames are always at the center of mass). Probably for your use case only the linear part was needed and not the application point?

newton/tests/test_sensor_contact.py
```diff
@@ -762,5 +764,72 @@ def test_parallel_scenario(self):
...
 
 
+class TestSensorContactKamino(unittest.TestCase):
+    """End-to-end contact-sensor tests using the Kamino solver.
+
+    Regression coverage that the Kamino->Newton conversion populates ``Contacts.force`` so
+    ``SensorContact.total_force`` matches ``ContactAggregation``.
+    """
+
+    def test_box_on_plane_total_force_matches_aggregation(self):
```
Guirec-Maloisel, line 774: Would be good to test here more generally what is added by the PR (Kamino -> Newton conversion of contact forces) rather than the single use case that motivated adding this conversion (at least something that would recover also the application point based on the force and torque component in Newton). The Newton -> Kamino conversion of the force is being added in parallel by PR #2908 so we should also add a round trip unit test (either here or in the other based on what is merged first)

Guirec-Maloisel: Actually I just realized that the Kamino -> Newton force conversion is already handled (and taking into account the application point) by PR #2908, so I'll close this one and focus on reviewing the other one.

Guirec-Maloisel:
This functionality is already part of PR #2908 currently under review, so I'm closing this one for now. We could maybe add the unit test for sensor forces to the other PR to ensure that use case is covered.

PR #3096, opened 2026-06-08
Deprecate ls_parallel option in SolverMuJoCo

adenzler-nvidia:
## Description

Parallel line search is being removed from `mujoco_warp`, so the `ls_parallel` argument of `SolverMuJoCo` is deprecated. Its default becomes `None`; passing any value emits a `DeprecationWarning` but is still applied while `mujoco_warp` supports it, so callers that omit the argument see unchanged behavior. The bundled ANYmal example, the MuJoCo benchmark, and the robotics tutorial no longer pass `ls_parallel`.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_ls_parallel_deprecated
```

## New feature / API change

```python
import newton

# Passing ls_parallel now emits a DeprecationWarning (still applied for now):
solver = newton.solvers.SolverMuJoCo(model, ls_parallel=True)

# Recommended: drop the argument entirely
```

## Summary by CodeRabbit

* **Deprecations**
  * The `ls_parallel` parameter in the MuJoCo solver is now deprecated and emits a warning when used. Users should remove this parameter from their solver initialization.

* **Documentation**
  * Updated examples and tutorials to reflect removal of the deprecated parameter.


vreutskyy (approved):
LGTM

jcarius-nv (approved):
looks good, thanks!

One thing to consider, though I'd be fine with just removing it directly:

asv/benchmarks/benchmark_mujoco.py:288, :478, :568-580 removes ls_parallel from the benchmark Python API and CLI outright. Because the CLI uses parse_known_args(), existing invocations like --ls-parallel or --no-ls-parallel are now silently ignored instead of warning. That undercuts the PR’s deprecation story: SolverMuJoCo deprecates gracefully, but this command surface removes behavior immediately. I’d keep the argument for one release, pass it through to SolverMuJoCo so the same DeprecationWarning fires, or explicitly warn and ignore it.

adenzler-nvidia:
Fair point, I was removing it because my understanding is that the benchmark code is similar to tests, and considered internal code. The only reason it has a command-line interface is because it makes it easier for us to quickly run and test them.

PR #3097, opened 2026-06-08
Reinstate SensorRaycast

adenzler-nvidia:
## Description

#2971 removed `SensorRaycast`, which had been deprecated in 1.2. The single-cycle deprecation policy had not been announced when the sensor was deprecated, so the removal was premature. This PR reinstates the sensor (still deprecated in favor of `SensorTiledCamera`) and its test suite.

Because #2971 also rewrote `raycast.py` — `GeoType.HFIELD` now raycasts through a `wp.Mesh` BVH instead of the per-thread DDA grid, and `ray_intersect_geom`/`raycast_kernel` became standalone — this is not a plain revert. The kernels the sensor relied on are restored on top of the new `ray_intersect_geom`: a single `sensor_raycast_kernel` (the gated no-HFIELD variant is gone), plus `sensor_raycast_particles_kernel` and `ray_for_pixel`. HFIELD shapes raycast via their `wp.Mesh` id in `shape_source_ptr`, so heightfield support is retained without the removed heightfield-array kernel plumbing. The `intersect_ray` helper and BVH-based HFIELD raycasting introduced by #2971 are left intact.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_sensor_raycast
uv run --extra dev -m newton.tests -k test_picking
```

## Summary by CodeRabbit

* **New Features**
  * Added a deprecated SensorRaycast depth sensor (perspective raycasting with optional particle intersections).

* **Documentation**
  * API and concept docs updated to include SensorRaycast, list five sensors, and mark Raycast as deprecated with TiledCamera as the recommended replacement.

* **Tests**
  * Added comprehensive tests validating SensorRaycast behavior, particle interactions, and deprecation warnings.

* **Changelog**
  * Fixed Unreleased notes to no longer state SensorRaycast was removed.


jcarius-nv (dismissed):
Thanks for the detailed PR context. I reviewed this as a targeted restoration of a prematurely removed deprecated API, not as a plain revert.

  The approach looks sound to me. The restored `SensorRaycast` implementation preserves the prior API shape while adapting the geometry path to the newer
  standalone `ray_intersect_geom` flow. I also checked the HFIELD behavior called out in the description: the mesh-backed `shape_source_ptr` path works as expected
  in a manual smoke test.

  Validation I ran locally from PR head `a20046ac`:

  - `uv run --extra dev -m newton.tests -k test_sensor_raycast`
  - `uv run --extra dev -m newton.tests -k test_raycast`
  - `uv run --extra dev -m newton.tests -k test_picking`
  - `uv run --extra dev python docs/generate_api.py`

  All passed, and `docs/generate_api.py` produced no API-doc drift.

  No blocking findings from me. The only small follow-up I’d consider is adding a dedicated `SensorRaycast` HFIELD regression test, since the PR specifically
  preserves that behavior, but I don’t think that should block this change.

camevor (dismissed):
Thanks, agree that this looks good!

adenzler-nvidia:
added a regression test. Thanks for the reviews!

newton/tests/test_sensor_raycast.py
```diff
@@ -513,6 +513,58 @@ def test_sensor_raycast_ellipsoid(test: unittest.TestCase, device: str):
...
+        max_distance=100.0,
+    )
+
+    sensor.update(state)
+    depth = sensor.get_depth_image_numpy()
+
+    hits = depth[depth > 0.0]
+    # The terrain fills the frame, but ~10-15% of rays miss along triangle
+    # edges (non-watertight mesh_query_ray), so require "most" pixels rather
+    # than all -- matching test_sensor_tiled_camera_heightfield.
```
camevor: This seems a bit suspicious. Would be interesting to learn whether this is an artifact of the precise camera orientation being pathological. Not a blocker for this PR, but probably worth following up on.

camevor (dismissed):
The height field regression test looks good.
No blockers , but since the tiled camera renders height fields through the mesh query path, we should confirm whether the high miss rate translates to the preferred tiled camera sensor.

newton/tests/test_sensor_raycast.py
adenzler-nvidia: Good catch -- dug into it, and it's not the camera orientation. SensorRaycast hits all 256 pixels here (verified on CPU and CUDA, and robust to small camera offsets/tilts), so the 80% threshold and "~10-15% miss" comment were just wrong -- I'd copied them over from `test_sensor_tiled_camera_heightfield`. Tightened the test to require every ray to hit, with tight depth bounds (478578f27).

The reason that other test legitimately needs ~80%: both sensors intersect the same heightfield `wp.Mesh` via `wp.mesh_query_ray`, but SensorTiledCamera enables backface culling by default and SensorRaycast doesn't. With culling forced off the tiled camera also hits 256/256 -- the ~25 it drops are grazing hits on the heightfield's non-watertight internal edges that the culling test rejects.

jcarius-nv (approved):
approving based on prior approval

PR #3104, opened 2026-06-08
Fix MJCF collision mask filtering

### LINKED ISSUE #3014: [BUG] contype/conaffinity collision filtering not working
### Bug Description

**The contype/conaffinity collision filtering is not working**

In this minimal example, both spheres are configured with `contype="1", conaffinity="0"`.
And the floor is configured with `contype="0" conaffinity="1"`

This means that both spheres should collide with the floor but not with each other.

```xml
<mujoco model="minexample">

    <default>
        <geom contype="0" conaffinity="0" condim="3"/>
    </default>

    <worldbody>
        <geom name="floor" type="plane" size="5 5 0.01"
              contype="0" conaffinity="1" rgba="0.7 0.7 0.7 1"/>

        <body name="sphere_a" pos="0 -0.06 0.6">
            <freejoint/>
            <geom name="sphere_a" type="sphere" size="0.1"
                  contype="1" conaffinity="0" rgba="1 0.2 0.2 1" mass="1"/>
        </body>

        <body name="sphere_b" pos="0 0.06 0.6">
            <freejoint/>
            <geom name="sphere_b" type="sphere" size="0.1"
                  contype="1" conaffinity="0" rgba="0.2 0.4 1 1" mass="1"/>
        </body>
    </worldbody>
</mujoco>

```
However, in Newton, they do collide with each other, but the same example in Mujoco works as expected.
I provided example scripts to test in both cases.

### Reproduction Script

```python
from pathlib import Path
import numpy as np
import warp as wp
import newton

MODEL_PATH      = Path(__file__).parent / "minimal_robot.xml"
DT              = 0.002
STEPS_PER_CYCLE = 200

builder.add_mjcf(str(MODEL_PATH))
model = builder.finalize()

solver  = newton.solvers.SolverMuJoCo(model)
control = model.control()

q0  = model.state().joint_q.numpy().copy()
qd0 = np.zeros(model.state().joint_qd.numpy().shape, dtype=np.float32)

def make_state():
    s = model.state()
    wp.copy(s.joint_q,  wp.array(q0,  dtype=wp.float32))
    wp.copy(s.joint_qd, wp.array(qd0, dtype=wp.float32))
    return s

viewer.set_model(model, max_worlds=1)

state_in  = make_state()
state_out = model.state()
contacts  = model.contacts()
sim_t     = 0.0
step      = 0

while viewer.is_running():
    if step > 0 and step % STEPS_PER_CYCLE == 0:
        state_in = make_state()
        sim_t    = 0.0

    solver.step(state_in, state_out, control, contacts, dt=DT)

    viewer.begin_frame(sim_t)
    viewer.log_state(state_in)
    viewer.end_frame()

    sim_t += DT
    step  += 1

viewer.close()


###################################################################
#  Same example in Mujoco
###################################################################

from pathlib import Path
import mujoco
import mujoco.viewer
import time

MODEL_PATH      = Path(__file__).parent / "minimal_robot.xml"
STEPS_PER_CYCLE = 300

model = mujoco.MjModel.from_xml_path(str(MODEL_PATH))
data  = mujoco.MjData(model)

q0  = data.qpos.copy()
qd0 = data.qvel.copy()

def reset():
    data.qpos[:] = q0
    data.qvel[:] = qd0
    mujoco.mj_forward(model, data)

step = 0

with mujoco.viewer.launch_passive(model, data) as viewer:
    while viewer.is_running():
        if step > 0 and step % STEPS_PER_CYCLE == 0:
            reset()

        mujoco.mj_step(model, data)
        viewer.sync()
        time.sleep(0.01)

        step += 1
```

### System Information

Tested on two machines:

Machine 1 (CPU only)
- Newton: 1.2.0
- Warp: 1.13.0
- Python: 3.12.0
- OS: Ubuntu 24.04.4
- GPU: no

Machine 2 (Nvidia GPU)
- Newton: 1.2.0
- Warp: 1.13.0
- Python: 3.12.13
- OS: Ubuntu 24.04.4
- GPU: Nvidia RTX 5090

m-abr:
## Description

Fixes #3014.

This PR implements MuJoCo-style `contype` / `conaffinity` collision filtering in Newton contact generation.

The change adds `collision_type` and `collision_affinity` masks to shape configuration/model data, maps MJCF `contype` and `conaffinity` directly to those masks, and resolves existing `collision_group` filtering into the same runtime mask representation during model finalization.

The broad phase and explicit contact-pair generation now use the same mask rule:

```text
(type_a & affinity_b) != 0 or (type_b & affinity_a) != 0
```

This fixes the reported case where two spheres with `contype=1`, `conaffinity=0` should each collide with a floor using `contype=0`, `conaffinity=1`, but should not collide with each other.

A full pair matrix would be very expressive, but it would scale poorly in memory and is not ideal for broad-phase kernels. This PR keeps the existing collision mechanisms rather than replacing them:

- `collision_group` remains the simple group-level API. It is useful for common cases and existing models, and is resolved into masks at finalization, so it does not add per-step cost.
- `collision_type` / `collision_affinity` provide the more expressive category-level filtering needed for MJCF-style masks.
- `shape_collision_filter_pairs` remains the sparse exact-pair exclusion layer for same-body, parent-child, USD `filteredPairs`, disabled collision shapes, articulation self-collision filtering, and other one-off pair exclusions.

The runtime path therefore stays uniform and fast: broad phase and solver code use resolved type/affinity masks, while pairwise filters handle sparse exceptions.

## Checklist

## Test plan

```
WARP_CACHE_PATH=/tmp/warp-cache python -m unittest newton.tests.test_broad_phase.TestCollisionTypeAffinityBroadPhase
WARP_CACHE_PATH=/tmp/warp-cache python -m unittest newton.tests.test_import_mjcf.TestContypeConaffinityZero.test_contype_conaffinity_masks_filter_contacts
python -m py_compile newton/_src/sim/builder.py newton/_src/sim/collide.py newton/_src/sim/model.py newton/_src/geometry/broad_phase_common.py newton/_src/geometry/broad_phase_nxn.py newton/_src/geometry/broad_phase_sap.py newton/_src/solvers/mujoco/solver_mujoco.py newton/_src/utils/import_mjcf.py newton/tests/test_broad_phase.py newton/tests/test_import_mjcf.py
```

## Bug fix

**Steps to reproduce:**

1. Import an MJCF model with:
    - floor: contype="0" conaffinity="1"
    - two spheres: contype="1" conaffinity="0"
2. Finalize the Newton model and inspect generated shape contact pairs.
3. Without this fix, the sphere-sphere pair is generated even though MuJoCo filtering excludes it.

**Minimal reproduction:**

```
<mujoco model="minexample">

    <default>
        <geom contype="0" conaffinity="0" condim="3"/>
    </default>

    <worldbody>
        <geom name="floor" type="plane" size="5 5 0.01"
              contype="0" conaffinity="1"/>

        <body name="sphere_a" pos="0 -0.06 0.6">
            <freejoint/>
            <geom name="sphere_a" type="sphere" size="0.1"
                  contype="1" conaffinity="0" mass="1"/>
        </body>

        <body name="sphere_b" pos="0 0.06 0.6">
            <freejoint/>
            <geom name="sphere_b" type="sphere" size="0.1"
                  contype="1" conaffinity="0" mass="1"/>
        </body>
    </worldbody>
</mujoco>
```

## New feature / API change

This PR adds mask-based collision filtering through `ModelBuilder.ShapeConfig`:

```python
floor_cfg = newton.ModelBuilder.ShapeConfig(
    collision_type=0,
    collision_affinity=1,
)
sphere_cfg = newton.ModelBuilder.ShapeConfig(
    collision_type=1,
    collision_affinity=0,
)
```

`collision_group` remains available for group-level filtering. During finalization, Newton resolves either group-derived or explicit type/affinity masks into `Model.shape_collision_type` and `Model.shape_collision_affinity`, which are used by the runtime collision pipeline.

## Summary by CodeRabbit

* **New Features**
  * Added per-shape `collision_type` / `collision_affinity` masks for model building and runtime collision filtering, used by both NXN and SAP broad-phase/contact pair generation (with automatic derivation from `collision_group` when masks are not authored).
* **Bug Fixes**
  * Fixed MJCF `contype` / `conaffinity` mask collision filtering and improved contact-generation stability across multiple solver and example scenarios.
* **Documentation**
  * Documented collision type/affinity semantics and added a matching test example.
* **Tests**
  * Added broad-phase, MJCF import, solver export, and model-builder regression tests for mask-based filtering.


jcarius-nv:

m-abr:
Thank @jcarius-nv. I think all the suggestions are now addressed

adenzler-nvidia:
hi @m-abr - sorry for the late response. How critical is this feature for you? we are still undecided on our side if we want to adopt a full mujoco-style collision filtering in the newton core API and what the implications are.

m-abr:
Hi @adenzler-nvidia, it's not super critical for me, and I am also not fully happy with my PR as-is, because I would prefer Newton to have one coherent collision filtering method rather than a mix of methods.

Also, I think Newton's collision groups plus excluded pairs are expressive enough for MJCF collisions. Unless I'm missing something, contype/conaffinity can be converted by assigning unique negative groups and excluding every pair that the Mujoco mask rule would reject. This would not be as efficient as the approach in this PR, but it would be cleaner, and MJCF collisions would at least be respected if you prefer to keep the current mechanism.

Regarding efficiency, I think masks might sometimes be faster since they can avoid some excluded-pair lookups, but they are also limited by the fixed bit width, which is usually not a problem anyway. My impression is that the two approaches are probably in a similar range in terms of performance, and I am also not sure what the best long-term direction is.

han-xudong:
Hi @adenzler-nvidia and @m-abr, I wanted to share some context from recent MJCF integration work where this limitation has a practical impact.
Some MJCF models require selective self-collision within the same articulation: certain internal geometry pairs must collide, while others must remain filtered. Treating all nonzero contype/conaffinity values alike can introduce unintended internal contacts, while disabling self-collision entirely also removes required contacts. This affects both simulation correctness and collision-processing cost.
Given the preference for keeping one coherent collision-filtering API in Newton, the conversion approach discussed above seems particularly suitable. The importer could evaluate geom-pair compatibility using the MuJoCo mask rule and compile rejected pairs into Newton collision groups and shape_collision_filter_pairs, potentially using groups to reduce the number of explicit exclusions.
A regression test comparing eligible geom pairs before and after MJCF import would also help verify semantic preservation. I appreciate the work already put into this PR and the careful consideration of the long-term API design.

eric-heiden:
Hi @han-xudong and @m-abr, I'm prototyping a solution to come up with a conversion from contype/conaffinity to Newton's collision groups and pairwise filters here: https://github.com/newton-physics/newton/pull/3714.
Can you give this a try? There are limitations going from collision groups/filters to MuJoCo's 32-bit contype/conaffinity values but at least using `SolverMuJoCo(..., use_mujoco_contacts=False)` should work with Newton's CollisionPipeline to simulate these contact exclusions from MJCF.
I tried this on the Menagerie assets which seems to work so far. I would be interested to see if your use cases are also covered by this. Ideally we can maintain just one convention here, which is the collision group/filter pairs approach from UsdPhysics that we implement in Newton.

han-xudong:
Hi @eric-heiden, thank you for putting this together. I tested #3714 with our MJCF use case. The imported collision relationships match the original contype/conaffinity configuration, and SolverMuJoCo(..., use_mujoco_contacts=False) runs correctly with the expected contact exclusions. This approach covers our use case well.

eric-heiden:
Closing in favor of #3714.

PR #3105, opened 2026-06-08
Handle deprecation warnings in actuator and sensor tests

adenzler-nvidia:
## Description

These tests trip deprecation warnings that the upcoming opt-in `--strict-warnings` runner flag (#3044) escalates to errors. They are pre-existing warning debt that accumulated in `main` after the earlier cleanup PRs (#3058–#3061), so this continues that work and should merge before #3044.

- **`test_actuators`** — three tests write controller targets through the deprecated `Control.joint_target_pos` alias incidentally; migrate them to the canonical `joint_target_q` (the affected models are revolute-only, so the DOF and coord layouts are identical).
- **`test_actuators`** — `test_target_pos_read_from_dof_index_not_coord_index` deliberately exercises the legacy DOF-index target layout via the default attribute resolution, which is deprecated and warns. Assert the expected `DeprecationWarning` instead of suppressing it.
- **`test_sensor_tiled_camera_heightfield`** — migrate off the deprecated `newton.geometry.build_bvh_shape`/`build_bvh_particle` helpers to `model.bvh_build_shapes`/`model.bvh_build_particles`.

## Checklist

Test-only change with no user-facing behavior change, so no `CHANGELOG.md` entry.

## Test plan

The flag does not exist on `main` yet, so strict mode's deprecation axis is reproduced directly by escalating `DeprecationWarning` at the interpreter level:

```
PYTHONWARNINGS=error::DeprecationWarning uv run --extra dev -m newton.tests -k test_actuators
PYTHONWARNINGS=error::DeprecationWarning uv run --extra dev -m newton.tests -k test_sensor_tiled_camera_heightfield
```

Both modules pass with these changes; without them the five tests error on the escalated `DeprecationWarning`s.

## Tests
* Updated actuator pipeline integration tests to verify joint target handling across different execution paths.
* Updated sensor camera heightfield test setup for BVH construction verification.

### DIFF
--- a/newton/tests/test_actuators.py
+++ b/newton/tests/test_actuators.py
@@ -671,7 +671,7 @@ class TestActuatorStep(unittest.TestCase):
         control = model.control()
         for step_i in range(5):
             tgt = target_schedule[step_i]
-            _write_dof_values(model, control.joint_target_pos, dofs, [tgt] * n)
+            _write_dof_values(model, control.joint_target_q, dofs, [tgt] * n)
             written_targets.append(tgt)
 
             control.joint_f.zero_()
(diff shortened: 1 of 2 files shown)


vreutskyy (approved):
LGTM

PR #3106, opened 2026-06-08
Add workflow to auto-assign PR creator when no assignee is set.

dsmither-lgtm:
Uses pull_request_target so fork PRs can be assigned in newton-physics/newton.

## Description
Adds a GitHub Actions workflow (.github/workflows/pr_auto_assign_creator.yml) that automatically assigns the PR author as Assignee when a pull request is opened without one.

This helps keep PR ownership visible in the sidebar and in tooling that reads assignee data (for example, internal PR dashboards). Many Newton PRs are opened without an assignee today.

Behavior:

Triggers on pull_request_target for opened, reopened, and ready_for_review
Skips PRs that already have assignees
Skips draft PRs and bot-authored PRs (*[bot])
Uses pull_request_target (not pull_request) so the workflow can assign authors on fork PRs, where the default pull_request token is read-only
Does not check out PR code — only calls the Issues API via actions/github-script
Logs a warning (non-failing) if the author cannot be assigned (e.g. external contributors who are not assignable in this repo)
Note: pull_request_target is used here intentionally and safely: the job performs a single privileged API call and does not execute untrusted code from the PR branch.

## Checklist

- [ ] New or existing tests cover these changes — N/A (CI workflow only; no application code changed)
- [ ] The documentation is up to date with these changes — N/A
- [ ] CHANGELOG.md has been updated (if user-facing change) — N/A (maintainer/CI-only change)

## Test plan

Verified on a fork before opening this PR:

Added the workflow to main on dsmither-lgtm/ci-auto-assign-pr-creator (required for pull_request_target, which reads workflows from the default branch).
Opened [dsmither-lgtm/ci-auto-assign-pr-creator#1](https://github.com/dsmither-lgtm/ci-auto-assign-pr-creator/pull/1) without setting an assignee.
Confirmed the Auto-assign PR creator workflow completed successfully.
Confirmed the PR author (dsmither-lgtm) was added under Assignees.
After merge to newton-physics/newton:

Open a new PR without selecting an assignee → author should appear under Assignees.
Open a PR with an assignee already set → workflow should no-op (logged as "already has assignee(s)").
For an external fork PR where the author is not assignable → workflow should log a warning and not fail CI.

## Summary by CodeRabbit

* **Chores**
  * Automated pull request assignment: PRs are now automatically assigned to their creators when opened, reopened, or marked as ready for review, with graceful error handling for assignment issues.
  * Added a GitHub Actions workflow to handle assignment checks and apply an additional runner hardening step for eligible non-draft pull requests.


.github/workflows/pr_auto_assign_creator.yml
```diff
@@ -0,0 +1,56 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 3
+    steps:
+      - name: Harden the runner (Audit all outbound calls)
+        uses: step-security/harden-runner@fa2e9d605c4eeb9fcad4c99c224cee0c6c7f3594  # v2.16.0
+        with:
+          egress-policy: audit
+
+      - name: Assign PR creator when no assignee
+        uses: actions/github-script@v7
```
jcarius-nv: Blocking: this is a privileged `pull_request_target` workflow with write permission, so `actions/github-script@v7` should not be tag-pinned. The repo policy is `action@<sha>  # vX.Y.Z`, and existing workflows already follow that pattern, including `actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd  # v8.0.0` in `scheduled_nightly.yml`. Please pin this action by commit SHA before merging.

.github/workflows/pr_auto_assign_creator.yml
```diff
@@ -0,0 +1,56 @@
...
+            }
+
+            try {
+              await github.rest.issues.addAssignees({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: pr.number,
+                assignees: [author],
+              });
+              core.info(`Assigned PR #${pr.number} to ${author}`);
```
jcarius-nv, line 83: Medium: `addAssignees` can return successfully while silently ignoring users who are not assignable, so this log can say the PR was assigned even when the author was not actually added. That contradicts the PR description’s external-contributor behavior and makes the workflow hard to trust operationally. Please verify the response contains `author` in `data.assignees`, or preflight with the assignee check endpoint, and log a warning/no-op when the author is not assignable.

.github/workflows/pr_auto_assign_creator.yml
```diff
@@ -0,0 +1,56 @@
...
+        with:
+          egress-policy: audit
+
+      - name: Assign PR creator when no assignee
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const pr = context.payload.pull_request;
+
+            if (pr.assignees?.length > 0) {
```
jcarius-nv: Low: this uses the event payload rather than current issue state. If a maintainer or another automation assigns someone after the event fires but before this step runs, the workflow can still append the author. If the invariant is strictly “assign only when currently unassigned,” fetch the current issue/PR assignees immediately before mutating.

jcarius-nv (commented):
Principal-engineer pass: I found three issues worth addressing before merge. The first is blocking because this is a privileged `pull_request_target` workflow; the second affects whether the workflow can be trusted to do what it reports; the third is a smaller race around the “only when unassigned” invariant.

.github/workflows/pr_auto_assign_creator.yml
dsmither-lgtm: Addressed in c97e32b: replaced actions/github-script@v7 with the same SHA-pinned ref used elsewhere in this repo (scheduled_nightly.yml): actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0

.github/workflows/pr_auto_assign_creator.yml
dsmither-lgtm: Addressed in c97e32b with two checks:

1. **Preflight:** call `issues.checkUserCanBeAssigned()` before `addAssignees`. A 404 response logs a warning and returns without attempting assignment.
2. **Post-flight:** after `addAssignees`, verify `data.assignees` contains the author login. If not, log a warning instead of reporting success.

This avoids the false-positive success log when GitHub silently ignores unassignable users.

.github/workflows/pr_auto_assign_creator.yml
dsmither-lgtm: Addressed in c97e32b: removed the early check against `context.payload.pull_request.assignees` and instead fetch live state immediately before mutating:

```javascript
const { data: issue } = await github.rest.issues.get({ owner, repo, issue_number: pr.number });
if (issue.assignees?.length > 0) { /* no-op */ }
```

This preserves the "assign only when currently unassigned" invariant even if another automation or maintainer assigns someone between the event firing and this job running.

jcarius-nv (approved):

PR #3110, opened 2026-06-09
Add release process audit helpers

jcarius-nv:
## Description

Add release-process skill updates for Newton release work:

- Add a `release_audit` dependency/license helper that compares `pyproject.toml`, `uv.lock`, project license metadata, and in-tree license notice files across release refs.
- Wire the license audit output into the release-audit workflow and report template.
- Add the release-notes skill used for drafting/reviewing release notes.

## Checklist

## Test plan

```bash
python3 -B .claude/skills/release-audit/scripts/license_audit.py --base v1.2.1 --head v1.3.0rc1 --skip-pypi
git diff --check -- .claude/skills/release-audit
```

## Bug fix

Not a bug fix.

## New feature / API change

Not a Newton public API change. This updates release-process skills only.

## Release Notes

* **Documentation**
  * Updated the release-audit procedure and references to include a dedicated **Dependency & License Audit** section, expanded recognized inputs (e.g., lockfiles and license/notice files), and clarified rendering/selection behavior for pre-release/RC vs retrospective modes.
  * Added standardized release-notes drafting and acknowledgment guidance, templates, and formatting rules.

* **Chores**
  * Added an audit tool that compares two git refs and generates a Markdown dependency-and-license delta report (including optional license metadata review and in-tree notice changes).

* **Tests**
  * Added automated coverage for parsing, diffing, license-review behavior, and `--skip-pypi` output.


.claude/skills/release-audit/scripts/license_audit.py
```diff
@@ -0,0 +1,685 @@
...
+            [
+                "project.license-files",
+                "; ".join(base_project.license_files),
+                "; ".join(head_project.license_files),
+            ]
+        )
+
+    license_pathspecs = _license_file_pathspecs(base_project, head_project)
+    license_diff = (
+        _git(repo, "diff", "--name-status", f"{base}..{head}", "--", *license_pathspecs, required=False)
```
adenzler-nvidia: **`license-files` pathspecs are passed to `git diff` as plain pathspecs, so `**` globs silently match nothing.**

The declared `project.license-files` entries are PEP 639 globs, but they're handed straight to git:

```python
license_diff = _git(repo, "diff", "--name-status", f"{base}..{head}", "--", *license_pathspecs, required=False)
```

Git's default pathspec semantics differ from PEP 639: a plain `**/*.txt` requires a literal intermediate `/`, so it won't match a file sitting directly in the globbed dir. Newton's current declaration is `license-files = ["LICENSE.md", "newton/licenses/**/*.txt"]`, and the three notice files live directly under `newton/licenses/`:

```
$ git ls-files 'newton/licenses/**/*.txt'         # 0
$ git ls-files ':(glob)newton/licenses/**/*.txt'  # 3
```

Over the commit that added `newton/licenses/CC-BY-4.0.txt`, the plain pathspec reports nothing while `:(glob)` reports the `A`. So the "In-Tree License Notice File Changes" section is a false negative for the exact config shipped today: a notice file could be added/removed/modified during a release and the audit would still say "0 changes."

Fix: prefix each pathspec with `:(glob)` before passing to git (safe for the literal `LICENSE.md` case too). Worth a regression test that drops a `*.txt` directly under a `**`-globbed dir, since the current tests don't exercise a real `license-files` glob.

adenzler-nvidia:
Took a look and the approach seems reasonable — the script is well-structured, and assuming the release tooling is exercised in practice and useful, it looks fine. I ran the new tests (pass on 3.11/3.12) and a real `v1.2.1..main` invocation; the dependency/version/extra/license-metadata diffs came out correct.

One thing to resolve before this is good to go: the in-tree license-notice detection (see inline comment) misses files for Newton's current `license-files` glob. Not approving yet pending that, but no broader concerns.

adenzler-nvidia (approved):
LGTM

PR #3117, opened 2026-06-09
Require registered Newton USD schemas for add_usd collision API parsing

maxkra15:
## Description

  `ModelBuilder.add_usd()` relies on registered Newton USD schemas for both API detection and schema-declared fallback/default values. Previously, `newton/_src/usd/__init__.py` swallowed `ImportError` from `newton_usd_schemas`, so `add_usd()` could continue when the schema plugin was missing or too old. In that state, `Usd.Prim.HasAPI()` /  `GetAppliedSchemas()` omit unregistered applied APIs, causing authored
  `NewtonSDFCollisionAPI` / `NewtonMeshCollisionAPI` configuration to be  silently ignored.

  This change makes USD import fail fast when `newton-usd-schemas` is  unavailable or does not register the Newton collision schemas required by  the importer. Detection stays on the normal registered-schema path via
  `Prim.HasAPI()`, preserving schema defaults needed for SDF and hydroelastic  behavior.

  ## Checklist

  - [x] New or existing tests cover these changes (`test_sdf_usd` covers the  registered schema path and schema-default behavior)
  - [x] The documentation is up to date with these changes (no doc change  needed)

  ## Summary by CodeRabbit

  * **Bug Fixes**
    * Fixed a case where USD import could silently ignore Newton SDF / mesh    collision schemas when `newton-usd-schemas` was missing or outdated.    Import now fails clearly instead of proceeding without required registered schema support.

## Summary by CodeRabbit

* **Bug Fixes**
  * USD import now fails with a clear, descriptive error when Newton USD schema support is unavailable or incompatible, instead of silently continuing with missing schema defaults. This prevents generation of invalid SDF and hydroelastic configurations and makes missing schema issues easier to diagnose.


andrewkaufman (commented):
I would argue the bug should be fixed the other way around -- newton_usd_schemas are not optional if you've opted-in to USD itself. We should make that conditional import fail instead. There is a minimum newton-usd-schemas package associated with each newton release.

Simply detecting the APIs via workarounds is not safe -- you also need the default/fallback values for the attributes, which can only be achieved with the plugin registered properly

maxkra15:
@andrewkaufman  Good point. I changed the fix to fail fast when newton-usd-schemas is missing or too old, instead of reading raw apiSchemas metadata. That keeps detection on Prim.HasAPI() and preserves schema-provided fallback/default values for SDF and hydroelastic attributes.

newton/_src/usd/__init__.py
```diff
@@ -24,6 +26,31 @@
...
+        schema_name
+        for schema_name in ("NewtonSDFCollisionAPI", "NewtonMeshCollisionAPI")
+        if Usd.SchemaRegistry().FindAppliedAPIPrimDefinition(schema_name) is None
+    ]
+    if missing_schemas:
+        raise ImportError(
+            "Newton USD support requires a newer newton-usd-schemas package; missing registered schemas: "
+            f"{', '.join(missing_schemas)}. Install the USD importer dependencies with "
+            "`pip install 'newton[importers]'`."
+        )
```
andrewkaufman: Why is this necessary? Doesn't the pyproject.toml & uv.lock already guarantee this?

`NewtonMeshCollisionAPI` has been available since the first release of the schemas. Only `NewtonSDFCollisionAPI` is new, but even that is guaranteed by the dependency resolution.

newton/_src/usd/__init__.py
```diff
@@ -24,6 +26,31 @@
...
+def require_newton_usd_schemas(Usd=None) -> None:
+    """Raise if Newton USD schema support is unavailable or too old."""
+    if _newton_usd_schemas_import_error is not None:
+        raise ImportError(
+            "Newton USD support requires newton-usd-schemas. Install the USD importer dependencies with "
+            "`pip install 'newton[importers]'`."
+        ) from _newton_usd_schemas_import_error
+
+    if Usd is None:
+        return
```
andrewkaufman, line 33: This should come first, right? If USD itself is not available then we don't require the schemas either.

CHANGELOG.md
```diff
@@ -130,6 +130,7 @@
 - Fix `SolverMuJoCo` honoring force-space `shape_material_ke` / `shape_material_kd` for contacts (`use_mujoco_contacts=False`); authored `mjc:solref` is preserved via new `mujoco.solref` / `mujoco.solref_mode` per-shape custom attributes. Force-space scaling is unsupported on `use_mujoco_contacts=True` and the MuJoCo CPU backend.
 - Fix MJCF importer rejecting MuJoCo's one-value `solreflimit`/`solref` shorthand emitted by `mujoco.MjSpec.to_xml()` for default-valued joints, which broke `SolverMuJoCo(save_to_mjcf=...)` round-trips
 - Fix ground grid clipping in the Viser renderer
+- Fix `ModelBuilder.add_usd()` silently ignoring authored Newton USD collision schemas when `newton-usd-schemas` is unavailable or too old. USD import now fails clearly instead of continuing without the registered schema defaults required for SDF and hydroelastic configuration.
```
andrewkaufman: Needs to move to an unreleased section, I'm assuming this is the tail end of the 1.3.0 section

jcarius-nv:

CHANGELOG.md
maxkra15: this entry is under ## [Unreleased] / ### Fixed, just before the 1.2.1 section. Seems to be correct.

andrewkaufman: I think that will not be the case after you rebase on main

maxkra15:
@jcarius-nv @andrewkaufman sorry, didn't see that, I will wait and move it after #3145 lands

maxkra15:
@andrewkaufman Updated: rebased on latest main, moved the changelogg entry to the new top-level Unreleased section, removed the explicit schema registry check, and now only fail when USD import is used and newton_usd_schemas failed to import

andrewkaufman (dismissed):

adenzler-nvidia (approved):
approving based on previous approval.

jcarius-nv (approved):
Approving again :)

PR #3119, opened 2026-06-10
Fix USD orphan world joints

### LINKED ISSUE #3092: USD importer creates articulation without ArticulationRootAPI
### Summary

A USD stage with no `PhysicsArticulationRootAPI` can still import with `model.articulation_count == 1` if it contains a body-to-world `PhysicsFixedJoint`.

That makes `ArticulationView` selectable even though the USD did not author an articulation root.

### Minimal repro

```python

import newton
from newton.selection import ArticulationView

body = UsdGeom.Xform.Define(stage, "/World/Base")
mass_api = UsdPhysics.MassAPI.Apply(body.GetPrim())
mass_api.GetMassAttr().Set(1.0)
mass_api.GetCenterOfMassAttr().Set(Gf.Vec3f(0.0, 0.0, 0.0))
mass_api.GetDiagonalInertiaAttr().Set(Gf.Vec3f(1.0, 1.0, 1.0))

joint = UsdPhysics.FixedJoint.Define(stage, "/World/root_joint")
joint.CreateBody1Rel().SetTargets(["/World/Base"])
joint.CreateLocalPos0Attr().Set(Gf.Vec3f(0.0, 0.0, 0.0))
joint.CreateLocalPos1Attr().Set(Gf.Vec3f(0.0, 0.0, 0.0))
joint.CreateLocalRot0Attr().Set(Gf.Quatf(1.0, 0.0, 0.0, 0.0))
joint.CreateLocalRot1Attr().Set(Gf.Quatf(1.0, 0.0, 0.0, 0.0))

builder.add_usd(stage, load_visual_shapes=False)
model = builder.finalize()

print(model.articulation_count)
print(model.articulation_label)
print(model.joint_articulation.numpy().tolist())

view = ArticulationView(model, "/World/Base")
print(view.joint_labels)
```

### Actual result

```text
1
['/World/Base']
[0]
['/World/root_joint']
```

### Expected result

No Newton articulation should be synthesized from a plain USD joint graph unless the USD authors `PhysicsArticulationRootAPI`.

The importer should either leave this as a no-articulation joint configuration, or otherwise require an explicit articulation root before creating `model.articulation_*` metadata.

### Why this matters

For USD assets that intentionally do not author articulations, Newton can create partial or misleading articulations "out of nowhere". This is especially confusing when debugging closed-loop assets, because `ArticulationView` may appear to work even though the USD did not define an articulation.

eric-heiden:
## Description

Fix USD import and MuJoCo conversion for jointed assets that do not author `PhysicsArticulationRootAPI`, including the Apptronik Apollo USD produced by the MJCF-to-USD converter. This addresses #3092.

The root cause was a mismatch between the authored USD topology and Newton's assumptions: the importer could synthesize partial articulation metadata, while SolverMuJoCo treated every joint outside an articulation as a loop closure and therefore did not instantiate standalone bodies attached directly to world.

This PR now:

- Keeps jointed mechanisms without `PhysicsArticulationRootAPI` consistently outside articulation metadata instead of synthesizing a partial articulation.
- Keeps strict `ModelBuilder.finalize()` validation for non-root orphan joints, while documenting and allowing standalone world-root joints (`parent == -1`) without articulation metadata. _This is needed for some assets, like the MJCF-converted Apollo model from the Menagerie._
- Removes orphan-articulation warnings from USD import; solver-specific support and diagnostics are handled by the selected solver.
- Makes SolverMuJoCo instantiate supported standalone world-root joints directly, with fixed roots represented as static or mocap bodies and dynamic roots represented as MuJoCo joints.
- Emits a SolverMuJoCo warning when the standalone-root fallback is used and raises a clear error for unsupported rootless mechanisms such as `boxes_fourbar.usda`.
- Recognizes the MJCF converter's non-rigid default prim as world for weld/connect equality constraints.
- Updates the articulations guide and changelog with the validation exception and solver limitations.

Standalone rigid bodies with no enabled mechanism joints continue to receive their normal single-body articulations. Users of unsupported non-root orphan topologies must still pass `skip_validation_joints=True` to finalize before handing the model to a compatible solver.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k TestImportUsdArticulation --strict-warnings --no-cache-clear
uv run --extra dev -m newton.tests -k articulation_validation --strict-warnings --no-cache-clear
uv run --extra dev -m newton.tests -k TestMuJoCoArticulationConversion --strict-warnings --no-cache-clear
uv run --extra dev -m newton.tests -k mjc_equality --strict-warnings --no-cache-clear
uv run --extra dev -m newton.tests -k TestMenagerieUSD_ApptronikApollo --strict-warnings --no-cache-clear
uvx pre-commit run -a
```

## Bug fix

**Steps to reproduce:**

1. Import a jointed USD stage that has no `PhysicsArticulationRootAPI`, or import the converted Apptronik Apollo USD containing a standalone `world_link` joint.
2. Inspect `builder.joint_articulation` and finalize the model.
3. Construct SolverMuJoCo.

Previously, rootless mechanisms could receive partial importer-created articulation metadata, and a valid standalone world-root body could be omitted during MuJoCo conversion. General rootless mechanisms also failed later with an indirect body-count mismatch.

**Minimal reproductions:**

- `test_rootless_mechanism_root_and_child_joints_stay_orphan` verifies importer metadata for a rootless mechanism.
- `test_orphan_world_fixed_body_is_exported_static` verifies SolverMuJoCo's standalone fixed-root conversion.
- `test_rootless_usd_mechanism_is_rejected` verifies the explicit error for an unsupported rootless four-bar mechanism.
- `TestMenagerieUSD_ApptronikApollo` verifies model parity, initial transforms, forward kinematics, and dynamics against native MuJoCo.


newton/_src/utils/import_usd.py
```diff
@@ -2616,13 +2619,12 @@ def _resolve_contact_attr(key, _prim=prim):
...
-            # Creates an articulation for the body-to-world FIXED joint (consistent with MuJoCo approach)
-            if joint_index is not None and is_body_to_world and is_fixed_joint:
-                child_body = builder.joint_child[joint_index]
-                builder.add_articulation([joint_index], label=builder.body_label[child_body])
-            else:
-                orphan_joints.append(joint_path)
+            if joint_index is not None:
+                if is_body_to_world and is_fixed_joint and not no_articulations:
+                    child_body = builder.joint_child[joint_index]
+                    builder.add_articulation([joint_index], label=builder.body_label[child_body])
```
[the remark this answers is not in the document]
eric-heiden: Addressed in the follow-up commits: the importer no longer synthesizes a one-joint articulation for body-to-world fixed joints. These joints keep `joint_articulation == -1`; model validation allows only the fixed-to-world pin case, and SolverMuJoCo exports standalone fixed roots without changing articulation metadata.

jcarius-nv:

newton/tests/test_import_usd.py
```diff
@@ -171,6 +171,94 @@ def test_import_non_articulated_joints(self):
...
+        self.assertEqual(builder.joint_count, 1)
+        self.assertEqual(builder.joint_label, ["/World/root_joint"])
+        self.assertEqual(builder.joint_articulation, [-1])
+
+        model = builder.finalize()
+        self.assertEqual(model.articulation_count, 0)
+        self.assertEqual(model.joint_articulation.numpy().tolist(), [-1])
+
+    @unittest.skipUnless(USD_AVAILABLE, "Requires usd-core")
+    def test_rootless_mechanism_root_and_child_joints_stay_orphan(self):
```
jcarius-nv, line 233: Minor: doesn't this test build on the previous one and could just be tested in one go? Or use two test cases so we don't duplicate 90% of the code?

newton/tests/test_import_usd.py
```diff
@@ -210,10 +298,70 @@ def define_body(path):
         self.assertEqual(builder.body_count, 2)
         self.assertEqual(builder.joint_count, 2)
         self.assertEqual(builder.joint_type.count(newton.JointType.FREE), 2)
-        # Each floating body should get its own single-joint articulation.
+        # With no enabled mechanism joints, each standalone body gets its own articulation.
```
jcarius-nv: wording could be clearer: Does it mean "with disabled mechanism joints" or "without any enabled mechanism joint"

newton/tests/test_model.py
```diff
@@ -1656,10 +1656,11 @@ def test_joint_world_validation(self):
...
         """Test that joints not belonging to an articulation raise an error on finalize."""
         builder = ModelBuilder()
-        body = builder.add_link()
+        parent = builder.add_link()
+        child = builder.add_link()
 
-        # Add joint but do NOT add it to an articulation
-        builder.add_joint_revolute(parent=-1, child=body, label="orphan_joint")
+        # Add a non-root joint but do NOT add it to an articulation.
+        builder.add_joint_revolute(parent=parent, child=child, label="orphan_joint")
```
jcarius-nv, line 1852: can you explain why this test was changed? Why did previously joints not belonging to an articulation raise an error?

newton/tests/test_menagerie_usd_mujoco.py
```diff
@@ -1795,6 +1795,10 @@ class TestMenagerieUSD_ApptronikApollo(TestMenagerieUSD):
     robot_xml = "apptronik_apollo.xml"
     usd_asset_folder = "apptronik_apollo"
     usd_scene_file = "usd_structured/apptronik_apollo.usda"
+    skip_reason = (
```
jcarius-nv: This disables the test, right? This feels like a symptom that the change may break more assets, at least we need to mention this in the changelog. Was the USD invalid before anyways and we were just permissive, or are we tightening what we can parse?

newton/_src/sim/builder.py
```diff
@@ -9891,7 +9855,13 @@ def _validate_joints(self):
             orphan_joints = []
             for i, art in enumerate(self.joint_articulation):
                 if art < 0:  # Joint is not in an articulation
+                    parent = self.joint_parent[i]
                     child = self.joint_child[i]
+                    if parent == -1:
```
jcarius-nv, line 9879: The new parent == -1 exception affects every ModelBuilder caller. However, the articulations guide
     (https://github.com/newton-physics/newton/blob/a60ee0b92755a222707296209d5baa24f3412d3b/docs/concepts/articulations.rst#L716-L744) still says all orphan joints fail finalization by default and generalized-coordinate solvers require articulation membership.

this warrants an additional Changed changelog entry. Suggested wording:

     > Allow world-root joints to remain outside articulation metadata during ModelBuilder.finalize(); use a maximal-coordinate solver for such joints, or add them to an articulation for generalized-coordinate solvers.

The guide and finalize() documentation should be updated with the same solver limitations. If existing MuJoCo support is intentionally removed, that is a breaking change requiring explicit migration guidance.

newton/tests/test_import_usd.py
jcarius-nv, line 1: Test the claimed solver behavior, not only metadata.

The new tests verify joint_articulation == -1 and successful finalization, but never simulate the orphan fixed root. Add an XPBD or SemiImplicit regression that applies gravity/forces and confirms the world-fixed body remains pinned. This directly validates thencomment that maximal-coordinate solvers consume the joint correctly.

newton/tests/test_import_usd.py
eric-heiden: Good point. I consolidated the duplicated stage setup into `_make_rootless_fixed_stage(with_child_joint=...)`. The two tests now share the same fixture and separately assert the standalone-root and root-plus-child cases.

newton/tests/test_import_usd.py
eric-heiden: Clarified to: `Because the stage has no enabled mechanism joints, each body is treated as standalone and receives its own articulation.` This is specifically the “without any enabled mechanism joint” case.

newton/tests/test_model.py
eric-heiden: The old test used `parent == -1`, but this PR intentionally allows standalone world-root joints without articulation metadata, so it no longer exercised the failure path. I changed it to a non-root body-to-body joint, added a separate positive test for the world-root exception, and documented the exception in `_validate_joints()` and `finalize()`.

newton/tests/test_menagerie_usd_mujoco.py
eric-heiden: The Apollo tests are re-enabled now. The USD is not invalid: its main robot has authored articulation metadata, while the converter-produced `world_link` is a standalone fixed world root. SolverMuJoCo now converts that root through a warned fallback. The changelog documents this, and the Apollo model, initial-transform, FK, and dynamics parity tests all pass.

newton/_src/sim/builder.py
eric-heiden: Added the Changed changelog entry and updated the articulations guide, `ModelBuilder.finalize()` documentation, `_validate_joints()` docstring, and the inline `parent == -1` exception. MuJoCo support is retained through a warned standalone-root fallback; non-root orphan validation remains strict by default.

newton/tests/test_import_usd.py
eric-heiden: I kept `test_import_usd.py` scoped to importer topology, metadata, and finalization rather than adding an XPBD/SemiImplicit simulation. This PR does not change those solvers. The solver behavior introduced here is covered in `test_mujoco_solver.py` (supported standalone world roots and explicit rejection of unsupported rootless mechanisms) and by the Apollo USD/MuJoCo integration tests. This keeps solver behavior in the solver-owned suites while covering the behavior changed by this PR.

eric-heiden:
Thanks @jcarius-nv for your review!
This got really hairy, I kept going back and forth between the importer, SolverMujoco, and even the MJCF2USD converter to investigate who should handle these orphan joints to make the conversion from MJCF to USD to MjSpec lossless. 
To recap, the Apollo MJCF looks like this:
```xml
<mujoco model="apollo_world_link">
    <worldbody>
        <body name="robot_base">
            <freejoint/>
            <geom type="sphere" size="0.1" mass="1"/>
            <body name="robot_link" pos="0 0 0.25">
                <joint type="hinge"/>
                <geom type="sphere" size="0.05" mass="0.1"/>
            </body>
        </body>

        <!-- Matches Apollo's empty top-level body. -->
        <body name="world_link"/>
    </worldbody>
</mujoco>
```
The resulting USD from the converter has this layout:
```
/apptronik_apollo/Geometry
├── base_link                 PhysicsArticulationRootAPI
│   └── entire Apollo robot   articulation 0
└── world_link                outside articulation
    └── PhysicsFixedJoint     world → world_link, articulation -1
```
So `world_link` has no PhysicsArticulationRootAPI, it just has a fixed joint connecting it to the world. Since it is outside an articulation, we call it an "orphan joint". Before this PR, our importer would just create an articulation out of nowhere, which caused issues in other assets that have descent joints below these orphan joints (only the first joint received an articulation while the others were kept with articulation -1).

With this PR, the USD importer no longer invents articulations - it just parses joints, and maps them to articulations if they are under an ArticulationRootAPI prim. The MuJoCo conversion now has a special treatment of these orphan root joints so that we can cleanly go from MJCF -> USD -> MjSpec to recover the original structure (although it prints a warning that we have this orphan root joint - see `test_orphan_world_fixed_body_is_exported_static()` and other tests, for more complicated scenarios with trees of orphan joints we raise an error). I have the PR description updated with the rules that we follow, I hope it is more consistent now.

jcarius-nv (dismissed):
Thanks for the changes, looks good to me now!

jcarius-nv (approved):
re-approving based on earlier approval

PR #3129, opened 2026-06-10
Renamed SensorTiledCameraBenchmark to FastSensorTiledCamera so it gets picked up by CI ASV runs

daniela-hase:
## Summary

Rename `SensorTiledCameraBenchmark` to `FastSensorTiledCamera` in `asv/benchmarks/simulation/bench_sensor_tiled_camera.py` so the benchmark is picked up by CI ASV runs.

## Background

ASV discovers benchmark classes by scanning for names that match a specific pattern. The previous name `SensorTiledCameraBenchmark` did not match the expected naming convention, causing the tiled
camera benchmark to be silently skipped during CI runs. Renaming to `FastSensorTiledCamera` ensures it is included.

## Changes

- Renamed class `SensorTiledCameraBenchmark` → `FastSensorTiledCamera`
- Updated the `__main__` benchmark registry to reference the new class name

## Testing

Benchmark can be run locally with:
```bash
uvx --with virtualenv asv run --launch-method spawn main^!
```

## Summary by CodeRabbit

* **New Features**
  * Added a CLI option to enable additional tiled render-order benchmarks so they are discoverable and runnable.

* **Chores**
  * Renamed an internal benchmark for clearer identification in listings.
  * Benchmark setup now rebuilds scene acceleration data and clears model collision settings for more consistent, reliable results.
* **Behavior**
  * Tiled-render benchmark methods are hidden by default and only exposed when the tiled benchmark option is enabled.

### DIFF
--- a/asv/benchmarks/simulation/bench_sensor_tiled_camera.py
+++ b/asv/benchmarks/simulation/bench_sensor_tiled_camera.py
@@ -10,6 +10,7 @@ wp.config.log_level = wp.LOG_WARNING
 import math
 
 import newton
+from newton import ShapeFlags
 from newton.sensors import SensorTiledCamera
 
 NICE_NAMES = {}
@@ -46,6 +47,9 @@ class FastSensorTiledCamera:
             newton.utils.download_asset("franka_emika_panda") / "urdf/fr3_franka_hand.urdf",
             floating=False,
         )
+        COLLIDE = int(ShapeFlags.COLLIDE_SHAPES) | int(ShapeFlags.COLLIDE_PARTICLES)
+        franka.shape_flags = [int(f) & ~COLLIDE for f in franka.shape_flags]
+        franka.shape_collision_filter_pairs = []
 
         scene = newton.ModelBuilder()
         scene.replicate(franka, world_count)
--- a/asv/benchmarks/simulation/bench_sensor_tiled_camera.py
+++ b/asv/benchmarks/simulation/bench_sensor_tiled_camera.py
@@ -8,12 +8,23 @@ wp.config.enable_backward = False
 wp.config.log_level = wp.LOG_WARNING
 
 import math
+import os
 
 import newton
 from newton import ShapeFlags
 from newton.sensors import SensorTiledCamera
 
 NICE_NAMES = {}
+ASV_RUN_TILED_CAMERA_BENCHMARKS_ENV_VAR = "NEWTON_RUN_TILED_CAMERA_BENCHMARKS"
+TILED_BENCHMARK_METHODS = {
+    "time_rendering_tiled_color_depth",
+    "time_rendering_tiled_color_only",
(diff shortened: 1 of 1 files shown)


adenzler-nvidia:
I agree having coverage for pull requests is a good idea here. How does this affect the total CI time?

daniela-hase:
> I agree having coverage for pull requests is a good idea here. How does this affect the total CI time?

It should only add like 10-20 seconds since the kernels are built for the unit tests already anyway - but we'll see for sure once the CI does a full run through here :)

daniela-hase:
Interestingly the benchmark ran about 10min longer than on other current PRs 🤨 I am not super sure why it runs so long, the benchmark I added takes like 20s to run locally 🤔

adenzler-nvidia:
I dug into the "10 min in CI vs ~20s locally" puzzle and I think almost all of it is in model `setup()`, not the render — wanted to run a hypothesis by you.

Two things compound:

- Locally (`run_benchmark` / `__main__`) `setup()` runs **once**. Under asv it runs once per (timed method × round × commit): with 6 `time_*` methods, the default `rounds=2`, and `asv continuous` covering base+head, that's ~24 full rebuilds of the 4096-world model.
- Each `setup()` is dominated by `finalize()`, and most of *that* is collision bookkeeping: one fr3 carries ~2,074 collision filter pairs, so `replicate(4096)` produces ~8.5M pairs and `finalize()` churns through every one of them.

The catch is that this is a pure rendering benchmark — no solver, no `step()`, just `eval_fk` once + repeated `sensor.update()`. The tiled camera only reads `VISIBLE` shapes + geometry + transforms + BVH, so none of that collision data is ever touched.

So my question: would it make sense to build the franka as render-only and drop the collision metadata before replicating?

```python
COLLIDE = int(ShapeFlags.COLLIDE_SHAPES) | int(ShapeFlags.COLLIDE_PARTICLES)
franka.shape_flags = [int(f) & ~COLLIDE for f in franka.shape_flags]
franka.shape_collision_filter_pairs = []
```

In a local repro at 4096 worlds this cut `finalize()` from ~5s to ~1.8s (~3x) with bit-identical output (color and depth max-|diff| both 0). Multiplied across the ~24 setups that's a large share of the extra wall-clock, without changing what the benchmark measures.

One caveat: this is only safe because the benchmark never steps physics — if it ever grows into a dynamics benchmark the collision data matters again, so it'd be worth a comment in the code. (FWIW `enable_self_collisions=False` goes the wrong way — it adds *more* filter pairs.)

Does that seem reasonable, or is there a reason to keep the full collision setup here?

daniela-hase:
> Does that seem reasonable, or is there a reason to keep the full collision setup here?

Oh yeah, that's a good idea! I've made the change :D

adenzler-nvidia (approved):
Thank you!

PR #3134, opened 2026-06-11
Route Joint Torque

david-cao-mueller:
## Description

Fix SolverKamino so Control.joint_f is routed to Kamino's `tau_j_ref` instead of `tau_j`. Add a minimal armature of `1e-6` in order to avoid division by zero.

## Checklist

- [x ] New or existing tests cover these changes

## Test plan

```
uv run --extra dev -m newton._src.solvers.kamino.tests.test_core_model TestModelConversions.test_30_control_conversions
```

## Summary by CodeRabbit

* **New Features**
  * Added `force_implicit_actuator_dynamics`/`force_implicit_joint_dynamics` option to force revolute/prismatic joints to use implicit dynamic modeling across the Kamino pipeline (solver, model conversion, USD import, and joint building).

* **Improvements**
  * Implicit dynamics now include per-joint external load torques in the resulting response.
  * Updated Newton↔Kamino control mapping to consistently use joint forces.
  * Joint wrench computation now skips joints whose dynamics are modeled elsewhere.

* **Tests**
  * Updated unit tests for the revised implicit-joint configuration and control comparison behavior.


chschuma-disney:
Adding my previous notes on the latest changes here:
- `joint_f` is again routed to `tau_j`, but `tau_j` is now routed to the dynamics constraint if there is one. This way, the model doesn't require a dynamics constraint to make `joint_f` have an effect, though it might be worth checking what the difference is between applying an external torque to the joint dynamics or the kinematic constraint.
- I replaced setting the minimum armature in the conversion to a minimum joint dynamics mass enforced directly in the dynamics kernel. This way, there is no way to create issues by overwriting the non-zero armature with a zero armature after the conversion.
- I introduced a flag in the Kamino USD importer and in `SolverKamino` (to be passed to the conversion) that can be set to force the creation of the dynamics constraint for actuators, instead of creating them always (even for passive joints). This way, we can support the previous behavior, and Isaac Lab can force dynamic joints if it needs them. I hope that could be a good compromise, let me know if that might work.

newton/_src/solvers/kamino/_src/core/control.py
```diff
@@ -95,6 +95,21 @@ def device(self) -> wp.DeviceLike:
     # Operations
     ###
 
+    @classmethod
+    def create_newton_wrapper(cls, model: ModelKamino, device: wp.DeviceLike = None) -> ControlKamino:
```
Guirec-Maloisel: Could you comment on why this additional wrapper function is needed, rather than adding the tau_j_ref initialization in the existing from_newton, or in finalize?

newton/_src/solvers/kamino/_src/core/conversions.py
```diff
@@ -823,6 +826,7 @@ def convert_joints(
     model: Model,
     model_size: SizeKamino,
     model_info: ModelKaminoInfo,
+    force_joint_dynamics: bool = False,
```
Guirec-Maloisel: The new flag makes sense to me, but maybe we could improve the naming? "force" can be a bit ambiguous in this context (especially for `force_dynamic` below). So maybe `use_dynamic_joints` or `enforce_dynamic_joints`? (dynamic_joints makes also more sense to me than joint_dynamics, but that could be just me).

newton/_src/solvers/kamino/_src/core/model.py
```diff
@@ -684,13 +684,20 @@ def control(self, requires_grad: bool = False, device: wp.DeviceLike = None) ->
         return control
 
     @staticmethod
-    def from_newton(model: Model) -> ModelKamino:
+    def from_newton(
+        model: Model,
+        force_joint_dynamics: bool = False,
```
Guirec-Maloisel: It looks like the new argument is not used yet in this function

newton/_src/solvers/kamino/_src/core/control.py
chschuma-disney: With the current routing, `from_newton()` doesn't assign anything to `tau_j_ref`, so to be able to have a complete `ControlKamino` object, we need to initialize that somewhere. Adding a memory allocation to `from_newton()` is not ideal, since we want to graph-capture that function. Adding it to `finalize()` is also not ideal since we might want to assign an existing array to it later, so we'd have allocated that memory for nothing. Plus since the dof-coords conversion will be unnecessary at some point, `finalize()` might disappear.

My ideal solution would be to just leave `tau_j_ref` as `None` and handle that in the solver, but I haven't gotten around to properly thinking this through and testing it.

Guirec-Maloisel: Thanks for clarifying! Two points:
- I realized this only recently, but allocating within a graph capture is actually fine (see e.g. wp.clone() within reset() in SolverKamino for caching values and restoring them after - I flagged this in the PR review and learned at that occasion that this was actually handled well by warp). There are probably a few conditions (like the array size should stay fixed) but in this case I think it should be ok.
- Passing None to a warp kernel instead of a wp.array is also supported (you have to leave the type annotation just as wp.array though, not add | None; and you should check then in the kernel if the array is set, e.g. if my_array:). See e.g. set_floating_base in resets.py in my reset API PR. So I think the route of leaving it as None by default would also be fine.
So all in all, if it's not too much to ask, I think it would be cleaner to handle this one of these two ways rather than adding another conversion function.

newton/_src/solvers/kamino/_src/core/conversions.py
chschuma-disney: I see how the naming can be ambiguous, though I'm not sure changing the verb would be the right way. `force_*` follows the convention of, e.g., `add_usd()` (`force_show_colliders`, `force_position_velocity_actuation`). `use_dynamic_joints` would suggest dynamic joints don't work without this set to `True`, which is not the intention. `enforce_dynamic_joints` sounds a bit like it enforces a specific rule/check.

Maybe we need to be a bit more verbose? `force_implicit_joint_dynamics` perhaps? Or some variant of `force_implicit_actuation`/`force_dynamic_actuators`/`force_implicit_actuator_dynamics`, if we want to be a bit more precise.

Guirec-Maloisel: Hmm I didn't see that there were already similar flags in add_usd. Maybe it's fine as-is then, the consistency helps to interpret this correctly.

newton/_src/solvers/kamino/_src/core/model.py
chschuma-disney: Good catch, this should be fixed now.

newton/_src/solvers/kamino/_src/core/conversions.py
chschuma-disney: I adjusted it to `force_implicit_actuator_dynamics` for now, but further discussion is needed.

newton/_src/solvers/kamino/_src/core/control.py
chschuma-disney: I went with the second option, to avoid the extra allocation and some uncertainty with the allocation (since the allocation would be graph-captured, but not the deallocation).

newton/_src/solvers/kamino/_src/core/builder.py
```diff
@@ -400,6 +400,7 @@ def add_joint(
         name: str | None = None,
         uid: str | None = None,
         world_index: int = 0,
+        force_dynamic: bool = False,
```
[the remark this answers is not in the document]
chschuma-disney: This should be fixed now.

newton/_src/solvers/kamino/_src/utils/io/usd.py
```diff
@@ -1294,6 +1295,11 @@ def _parse_joint(
             b_j=b_j,
             k_p_j=k_p_j,
             k_d_j=k_d_j,
+            force_dynamic=(
+                force_dynamic
+                and act_type != JointActuationType.PASSIVE
+                and (dof_type in (JointDoFType.REVOLUTE, JointDoFType.PRISMATIC))
+            ),
```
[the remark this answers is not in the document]
chschuma-disney: The behavior has been adjusted and the parameter name updated to better reflect its purpose.

newton/_src/solvers/kamino/_src/core/conversions.py
```diff
@@ -233,16 +234,22 @@ def joint_conversion_kernel(
...
+    else:
+        # Infer if the joint requires dynamic constraints
+        for dof_id in range(ndofs_j):
+            a_j = model_joint_armature[dofs_start_j + dof_id]
+            b_j = model_joint_friction[dofs_start_j + dof_id]
+            ke_j = model_joint_target_ke[dofs_start_j + dof_id]
+            kd_j = model_joint_target_kd[dofs_start_j + dof_id]
+            is_dynamic_j = is_dynamic_j or (a_j > 0.0) or (b_j > 0.0) or (ke_j > 0.0) or (kd_j > 0.0)
 
-    # Set joint dimensions
```
Guirec-Maloisel, line 245: Minor point, but I don't think this comment should be removed?

newton/_src/solvers/kamino/_src/dynamics/wrenches.py
```diff
@@ -54,6 +55,10 @@ def _compute_joint_dof_body_wrenches_dense(
     # Retrieve the thread index as the joint index
     jid = wp.tid()
 
+    # If the joint dynamics are modeled, `tau_j` will be taken care there
```
Guirec-Maloisel: Minor point, but should be "will be taken care of there"

newton/_src/solvers/kamino/_src/dynamics/wrenches.py
```diff
@@ -120,6 +125,10 @@ def _compute_joint_dof_body_wrenches_sparse(
...
```
Guirec-Maloisel: Same comment

newton/_src/solvers/kamino/_src/kinematics/joints.py
```diff
@@ -722,6 +722,7 @@ def compute_and_write_joint_implicit_dynamics(
     model_joint_k_d_j: wp.array[float32],
     data_joint_q_j: wp.array[float32],
     data_joint_dq_j: wp.array[float32],
+    data_joint_tau_j: wp.array[float32],
```
Guirec-Maloisel, line 725: Maybe add a comment to this line to clarify that this may be None (although unfortunately the annotation `wp.array[float32] | None` isn't supported here)

Edit: sorry wrong line, I meant the `data_joint_tau_j_ref` line.

newton/_src/solvers/kamino/tests/test_core_model.py
```diff
@@ -288,7 +288,51 @@ def test_01_model_conversions_fourbar_from_usd(self):
...
+    def test_02_model_conversion_fourbar_forced_implicit_actuator_dynamics(self):
+        """
+        Test that force_implicit_actuator_dynamics produces matching Kamino models
+        via Newton conversion and direct Kamino USD import.
+        """
+        # Define the path to the USD file for the fourbar model
+        asset_file = get_kamino_basics_asset("boxes_fourbar.usda")
+
+        # Create a fourbar using Newton's ModelBuilder
+        builder_0: ModelBuilder = ModelBuilder()
```
Guirec-Maloisel: Minor point, but maybe we could use something else than the _0, _1, _2 suffixes for clarity (e.g. builder_newton, builder_kamino etc). Especially as they're not even in the right order, model_2 comes before model_1...

newton/_src/solvers/kamino/tests/test_core_model.py
```diff
@@ -1181,23 +1225,26 @@ def test_30_control_conversions(self):
...
-        control_2: ControlKamino = ControlKamino()
+        control_2 = ControlKamino()
         control_2.finalize(model_1)
         control_2.from_newton(control_0, model_1)
         self.assertIsInstance(control_2.tau_j, wp.array)
         self.assertIs(control_2.tau_j, control_0.joint_f)
         self.assertEqual(control_2.tau_j.size, model_0.joint_dof_count)
-        test_util_checks.assert_control_equal(self, control_2, control_1)
+        self.assertIsNone(control_2.tau_j_ref)
+        test_util_checks.assert_control_equal(self, control_2, control_1, ["tau_j_ref"])
```
Guirec-Maloisel: Minor point, but for clarity we could write the `excluded=` keyword explicitly (so it doesn't look that tau_j_ref is the one thing we compare).

Guirec-Maloisel: I'd also have the comment about the _0 etc suffixes here, but up to you since this is not a new test.

newton/_src/solvers/kamino/tests/test_kinematics_joints.py
```diff
@@ -451,6 +471,30 @@ def test_04_multiple_dynamic_revolute_joints(self):
         np.testing.assert_almost_equal(inv_m_j_np, inv_m_j_expected)
         np.testing.assert_almost_equal(dq_b_j_np, dq_b_j_expected)
 
+    def test_05_implicit_dynamics_minimum_mass(self):
+        # Construct the model description with forced implicit actuator dynamics
+        builder = build_unary_revolute_joint_test(
```
Guirec-Maloisel, line 476: Maybe add a comment that a_j, b_j etc are all zero for this model, so that only the minimal mass kicking in makes it non-zero. Or possibly even assert this as well on the builder or model (in case we one day change the `build_unary_revolute_joint_test` code, since then we wouldn't be testing this anymore).

newton/_src/solvers/kamino/solver_kamino.py
```diff
@@ -646,7 +653,6 @@ def __init__(
             config=self._config,
         )
 
-        # Initialize the internal Kamino control wrapper
```
Guirec-Maloisel, line 649: Same comment here, not sure why this comment is being remove?

newton/_src/solvers/kamino/tests/test_core_model.py
```diff
@@ -288,7 +288,51 @@ def test_01_model_conversions_fourbar_from_usd(self):
         # test_util_checks.assert_model_equal(self, model_2, model_1)
         # test_util_checks.assert_data_equal(self, data_2, data_1)
 
-    def test_02_model_conversions_dr_testmech_from_usd(self):
+    def test_02_model_conversion_fourbar_forced_implicit_actuator_dynamics(self):
+        """
+        Test that force_implicit_actuator_dynamics produces matching Kamino models
```
Guirec-Maloisel: Should we also add a check in this test case that the forced dynamics were indeed applied (e.g. checking that the number of dynamic constraints is indeed positive)?

Guirec-Maloisel (changes requested):
Looks good overall, nice work! I had a bunch of comments but they're almost all minor.

newton/_src/solvers/kamino/_src/core/conversions.py
```diff
@@ -233,16 +234,22 @@ def joint_conversion_kernel(
...
     is_dynamic_j = bool(False)
-    for dof_id in range(ndofs_j):
-        a_j = model_joint_armature[dofs_start_j + dof_id]
-        b_j = model_joint_friction[dofs_start_j + dof_id]
-        ke_j = model_joint_target_ke[dofs_start_j + dof_id]
-        kd_j = model_joint_target_kd[dofs_start_j + dof_id]
-        is_dynamic_j = is_dynamic_j or (a_j > 0.0) or (b_j > 0.0) or (ke_j > 0.0) or (kd_j > 0.0)
+    if (
+        force_implicit_actuator_dynamics
+        and act_type_j != JointActuationType.PASSIVE
```
rubengrandia: Why have a condition on passive joints? Passive joints might have armature and in implicit friction, so could also require a dynamic constraint.

We can also open an issue to support this later if there are other blockers

rubengrandia:
"force_implicit_actuator_dynamics" seems like a very wasteful thing to do if this is just about resolving IsaacLab setting the parameters after model creation. 

I would also worry about the numerics of these extra constraints that are set to almost zero. Did we benchmark that?

Can we not keep the minimal set based on the actual parameters, and then redo this part of the model when "notify_solver_changes" is called.

newton/_src/solvers/kamino/_src/dynamics/wrenches.py
chschuma-disney: Good point, I rewrite it a bit differently

newton/_src/solvers/kamino/_src/kinematics/joints.py
chschuma-disney: Good point, we should make such a comment the convention

newton/_src/solvers/kamino/tests/test_core_model.py
chschuma-disney: Good point. I copied these from the other tests, so adjusted the naming in the whole test file.

newton/_src/solvers/kamino/tests/test_core_model.py
chschuma-disney: I added the keyword. We could think about enforcing keywords more widely in such cases.

newton/_src/solvers/kamino/tests/test_kinematics_joints.py
chschuma-disney: I added a comment and check.

newton/_src/solvers/kamino/solver_kamino.py
chschuma-disney: No point, just left over from going back and forth with changes. I restored the comment.

newton/_src/solvers/kamino/_src/core/conversions.py
chschuma-disney: This should be resolved with the flag being dropped.

Guirec-Maloisel (approved):
Looks good to me now!

PR #3132, opened 2026-06-10
Use positive MuJoCo solref for joint limits

eric-heiden:
## Description

Closes #3109.

`SolverMuJoCo` now keeps the existing Newton force-space joint-limit scaling by `dof_invweight0 * (1 - dmax)`, but converts the scaled direct stiffness/damping pair to MuJoCo's positive `(timeconst, dampratio)` `jnt_solref` convention instead of writing negative direct-mode values. This preserves the intended unclamped force-space response while allowing MuJoCo's timestep safety clamp to soften limits that are too stiff for the step size.

The CPU mirror now uses the same `MJ_MINVAL` clamp as the Warp kernel, the focused joint-limit tests expect positive solrefs, and the MuJoCo integration docs describe the runtime conversion and export semantics. The existing Unreleased changelog entry was amended rather than duplicating the same fix.

## Checklist

- [x] New or existing tests cover these changes
- [x] The documentation is up to date with these changes
- [x] `CHANGELOG.md` has been updated (if user-facing change)

## Test plan

```
uv run --extra dev -m newton.tests -k test_joint_limit_solref_conversion
uv run --extra dev -m newton.tests -k TestMuJoCoSolverInvweightScaledSolref
uv run --extra dev python newton\examples\repro_issue_3109.py
uvx pre-commit run -a
```

Also ran a local multi-mechanism issue-3109 diagnostic harness over Allegro hand, Inspire RH56, Unitree Dex5, Robotiq 2F85, Franka Panda, and Shadow Hand without tendons for 120 steps; all finished with `status=ok` and `first_nonfinite=None`.

## Bug fix

**Steps to reproduce:**

1. Run the issue #3109 Allegro hand repro using `SolverMuJoCo` with Newton default joint-limit gains.
2. Drop/contact a sphere against the low-inertia fingers.
3. On `main`, the local repro reaches non-finite constraint forces around step 98; with this PR, the same repro remains finite for the tested run.

**Minimal reproduction:**

```python
# See issue #3109 and the local repro_issue_3109.py script:
# load the Allegro hand USD, run SolverMuJoCo, drop a sphere onto
# the fingers, and track first_nonfinite / max_qfrc_constraint.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected joint-limit stiffness and damping conversion in the MuJoCo solver to use the proper positive convention, ensuring timestep safety clamping remains active.

* **Documentation**
[...]

Issue #3109, opened 2026-06-09
SolverMuJoCo MuJoCo-Warp contact response becomes non-finite for Allegro contact scene


AntoineRichard:
## Summary

We are seeing MuJoCo-Warp contact dynamics blow up to very large and then non-finite values for an Allegro hand contact scene. This came up while working on IsaacLab PR #5730, but the repro below is Newton-only: it imports `newton`, `warp`, `numpy`, and Python standard-library modules only.

The repro loads the public Isaac Sim Allegro USD, applies the same spawn rotation and soft actuator settings used by IsaacLab's Allegro config, adds one free falling sphere per world, and steps `SolverMuJoCo` with `use_mujoco_contacts=True`.

The solver settings in the repro below are matched to the IsaacLab contact sensor test configuration:

- `solver="newton"`
- `integrator="implicitfast"`
- `iterations=100`
- `ls_iterations=20`
- `nconmax=None`
- `njmax=200`
- `cone="elliptic"`
- `impratio=1.0`
- `dt=1 / 120`
- `num_substeps=1`

## Environment where reproduced

- Newton: `1.4.0.dev0`, installed from commit `c5f7902d289af78e10cd2875ee89e335cf2f4777`
- Warp: `1.14.0`
- MuJoCo-Warp: `3.8.1`
- GPU: NVIDIA RTX 5000 Ada Generation Laptop GPU
- CUDA device: `cuda:0`

## Newton-only repro

Run:

```bash
python newton_allegro_mjwarp_blowup_min_isaaclab_cfg.py
```

Script:

```python
#!/usr/bin/env python3
"""Newton-only repro for MuJoCo-Warp contact blow-up with Allegro."""

import math
import urllib.request
from pathlib import Path

import numpy as np
import warp as wp

import newton

ModelFlags = getattr(newton, "ModelFlags", None)
from newton import JointTargetMode, JointType
from newton.solvers import SolverMuJoCo

BASE_URL = "https://omniverse-content-production.s3-us-west-2.amazonaws.com/Assets/Isaac/6.0/Isaac/Robots/WonikRobotics/AllegroHand"
CACHE = Path("/tmp/isaac_allegro_usd_cache")
SIM_DT = 1.0 / 120.0
WORLD_COUNT = 4
STEPS = 120
DROP_VELOCITY = -1.5

FINGERTIP_OFFSETS_FROM_ROOT = [
    (-0.052, -0.252, 0.052),  # index
[...]

PR #3139, opened 2026-06-11
Use positive MuJoCo solref for force-space contacts

adenzler-nvidia:
## Description

Follow-up to #3132 (joint limits), found while reviewing it.

The shape-material force-space contact override in `convert_newton_contacts_to_mjwarp_kernel` writes per-contact `solref` in MuJoCo's negative direct-mode convention `(-ke * factor, -kd * factor)`. Direct mode bypasses MuJoCo's `refsafe` clamp (`timeconst >= 2*dt`) — the exact same latent failure mode #3109 hit for joint limits. Confirmed in `mujoco_warp`'s `constraint.py`: the shared `_update_efc_row` applies `refsafe` only to the positive `(timeconst, dampratio)` branch; the negative direct-mode branch discards the clamped `timeconst`. Contacts flow through that same function via `d.contact.solref`.

This PR converts the scaled stiffness/damping pair to the positive convention via the existing `convert_solref` helper. The conversion reproduces the same unclamped force-space response while letting `refsafe` soften contacts too stiff for the step.

Scope/severity notes:

- This path is **opt-in**: shapes default to `SOLREF_MODE_MJCF_DEFAULT`, and the override only fires when both shapes in a contact are explicitly `SOLREF_MODE_FORCE_SPACE`. (Joint limits, which default to `FORCE_SPACE`, were the broad #3109 trigger.) The blow-up here is therefore **latent** — not independently reproduced as non-finite, unlike #3109.
- Marked **draft** pending team review of two open points: (1) whether to also add a finiteness regression test exercising the `refsafe`-clamping regime (same gap noted on #3132), and (2) the separate question of whether joint limits should keep defaulting to `FORCE_SPACE` at all.

## Checklist

- [ ] `CHANGELOG.md` has been updated (if user-facing change) — N/A: internal convention fix to the already-listed (unreleased) force-space contact entry

## Test plan

```
uv run --extra dev -m newton.tests -k "test_mujoco_solver.TestMuJoCoSolverForceSpaceContactSolref"
```

All 11 tests in the class pass on GPU. The sign-based discriminators in `mixed_mode` / `dmax_one` were replaced with value-based checks, since both the override and the per-geom fall-through now emit positive `solref`.

## Bug fix

**Steps to reproduce (latent — same class as #3109):**

1. Build a model with two shapes both set to `mujoco.solref_mode == SOLREF_MODE_FORCE_SPACE` and large `shape_material_ke`, run with `SolverMuJoCo(use_mujoco_contacts=False)`.
2. The per-contact `solref` is written negative (direct mode), so MuJoCo's `refsafe` timestep clamp never engages.
3. A contact stiff enough for the timestep can diverge, mirroring #3109's joint-limit blow-up.

## Summary by CodeRabbit

* **Documentation**
  * Clarified SolverMuJoCo contact-material docs to explain positive stiffness/damping (time-constant, dampratio) convention and per-contact solref calculation.

* **Bug Fixes**
  * Ensured per-contact solref uses clamped positive stiffness/damping values and the proper positive (timeconstant, dampratio) mapping for consistent solver behavior.

* **Tests**
  * Updated and tightened contact-solver tests and assertions to validate the positive solref convention, bounds, and mixed-material behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -128,7 +128,7 @@
 - Honor authored `mujoco.solreflimit_mode` even when a non-zero `mujoco.solreflimit` is also present, so the explicit mode (force-space or raw) is authoritative
 - Fix `SolverMuJoCo` CPU backend overwriting `mj_model.body_iquat` with Newton's eigendecomposition result on every `BODY_INERTIAL_PROPERTIES` notify; the compiled principal-axes basis is now preserved, fixing single-contact box equilibrium (incorrect normal force) and stiff WELD-loop instabilities (`Nan, Inf or huge value in QACC`) caused by basis ambiguity on repeated eigenvalues
 - Fix `SolverMuJoCo` CPU backend dynamics for asymmetric MJCF `diaginertia` models whose principal moments are reordered during Newton/MJWarp synchronization
-- Fix `SolverMuJoCo` honoring force-space `shape_material_ke` / `shape_material_kd` for contacts (`use_mujoco_contacts=False`) using MuJoCo's positive solref convention so timestep safety clamping stays active; authored `mjc:solref` is preserved via new `mujoco.solref` / `mujoco.solref_mode` per-shape custom attributes. Force-space scaling is unsupported on `use_mujoco_contacts=True` and the MuJoCo CPU backend.
+- Fix `SolverMuJoCo` honoring force-space `shape_material_ke` / `shape_material_kd` for contacts (`use_mujoco_contacts=False`); authored `mjc:solref` is preserved via new `mujoco.solref` / `mujoco.solref_mode` per-shape custom attributes. Force-space scaling is unsupported on `use_mujoco_contacts=True` and the MuJoCo CPU backend.
 - Fix MJCF importer rejecting MuJoCo's one-value `solreflimit`/`solref` shorthand emitted by `mujoco.MjSpec.to_xml()` for default-valued joints, which broke `SolverMuJoCo(save_to_mjcf=...)` round-trips
 - Fix ground grid clipping in the Viser renderer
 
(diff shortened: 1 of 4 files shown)


jcarius-nv:
Review PR #3139 in the context of the MuJoCo stiffness/clamping issue.

  - The change makes sense for 1.3: it fixes the same refsafe bypass pattern as #3132, but for shape-side force-space contact stiffness/damping instead of joint
    limits.

  - Release risk is low-to-moderate. It only affects the opt-in path where both shapes use force-space contact solrefs, but behavior can change: very stiff
    shape_material_ke / shape_material_kd contacts may now be softened by MuJoCo’s safety clamp instead of preserving the raw direct-mode response.

  - The main validation gap is that the test appears to check the written solref conversion, but not a stiff-contact/high-timestep stability scenario. Short before
    release, I’d want at least the targeted test plus one quick smoke validation for stiff contacts.

  - The changelog point is right. We should add a similar but separate entry for shape-side ke/kd, scoped to shape_material_ke / shape_material_kd,
    use_mujoco_contacts=False, and SOLREF_MODE_FORCE_SPACE. See https://github.com/newton-physics/newton/pull/3140/changes for wording, for example
> Change `SolverMuJoCo` force-space contact handling for `shape_material_ke` / `shape_material_kd` to use MuJoCo’s positive `solref` convention so `refsafe`
  clamping can soften too-stiff contacts at large timesteps; users with explicitly authored force-space contact stiffness may need to retune values. This applies to
  `use_mujoco_contacts=False` when both shapes use `SOLREF_MODE_FORCE_SPACE`. (#3139)

  - Recommendation: include #3139 if #3132 is also going into 1.3. It is a consistency and safety follow-up, but I would avoid cherry-picking it alone without the
    joint-limit fix context.

adenzler-nvidia:
Thanks @jcarius-nv — agree on release framing (1.3, with #3132). On the validation gap: added `test_force_space_contact_stiff_contact_stays_finite` — a low-inertia body with a very stiff force-space contact at a coarse timestep (written timeconst below `2·dt`), stepped through contact and asserting `qvel`/`qfrc_constraint`/body velocity stay finite and bounded. So we now cover the `refsafe`-clamped regime, not just the written `solref`.

On the changelog: both #2610 and #2009 are unreleased, so I'd frame the difference as **default vs opt-in** rather than changed-vs-new:

- Joints: `solreflimit_mode` defaults to `FORCE_SPACE`, and released 1.2.1 used `jnt_solref = (-ke, -kd)` flat (no inertia scaling). #2610 added the `dof_invweight0·(1-dmax)` scaling, so existing assets' limit dynamics change *by default* (and can hit the safety clamp — #3109). That's why #3140's retune guidance fits.
- Contacts: `solref_mode` defaults to `MJCF_DEFAULT`, which is `convert_solref(ke,kd,1,1)` — identical to released 1.2.1 contact behavior. The two-body FORCE_SPACE override is new and opt-in, so existing assets are unaffected unless they explicitly set `SOLREF_MODE_FORCE_SPACE`. Nothing to retune by default.

The negative convention this PR replaces only existed on `[Unreleased]`/1.3.0-RC, so #3139 just makes the new opt-in feature ship in its correct positive form. The existing entry already documents the addition (*"Fix `SolverMuJoCo` honoring force-space `shape_material_ke`/`kd` for contacts (`use_mujoco_contacts=False`)…"*).

jcarius-nv (approved):
Aside of my limited Mujoco knowledge, this looks good now.

PR #3148, opened 2026-06-12
Bump warp-lang to 1.15.0 dev nightly

adenzler-nvidia:
## Description

Track upstream warp development on the 1.15.x line ahead of a 1.15 release. Now that a stable 1.14.0 is published on the nvidia index, default resolution prefers it over the nightly, so `prerelease = "allow"` is set in `[tool.uv]` and only `warp-lang` is locked to the latest nightly (`uv lock -P warp-lang`) — no other dependency resolves to a prerelease. The published `warp-lang>=1.14.0` floor is left unchanged, keeping prerelease tracking in the dev lockfile rather than forcing it onto Newton installs. `asv.conf.json` is pinned to the same nightly.

Setting the mode in `[tool.uv]` rather than only on the command line keeps it consistent with the lockfile so the `check-lockfile` job's `uv lock --check` stays green.

## Checklist

## Test plan

Rely on CI.

## Summary by CodeRabbit

* **Chores**
  * Updated pre-release package to the latest development version.
  * Enabled pre-release package versions for dependency resolution.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -34,9 +34,6 @@ conflicts = [[
     { package = "newton", extra = "torch-cu13" },
 ]]
 
-[options]
-prerelease-mode = "allow"
-
 [[package]]
 name = "absl-py"
 version = "2.4.0"
--- a/uv.lock
+++ b/uv.lock
@@ -34,6 +34,9 @@ conflicts = [[
     { package = "newton", extra = "torch-cu13" },
 ]]
 
+[options]
+prerelease-mode = "allow"
+
 [[package]]
 name = "absl-py"
 version = "2.4.0"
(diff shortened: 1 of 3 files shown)


nvtw (dismissed):

uv.lock
```diff
@@ -34,6 +34,9 @@ conflicts = [[
     { package = "newton", extra = "torch-cu13" },
 ]]
 
+[options]
+prerelease-mode = "allow"
+
```
shi-eric: To avoid having to constantly remove and add this option, have you considered just remove this, and upgrading warp by doing something like `uv lock -P warp-lang==1.15.0.dev20260612` instead?

adenzler-nvidia: Good call — done. Dropped `prerelease = "allow"` and re-locked with `uv lock -P warp-lang==1.15.0.dev20260612`. uv keeps the locked nightly as a preference, so `uv lock --check` stays green without the global option.

shi-eric (approved):

PR #3153, opened 2026-06-12
Remove deprecated body armature

adenzler-nvidia:
## Description

Remove body armature, deprecated since 1.1.0. This drops `ModelBuilder.default_body_armature`, the `armature` parameter of `ModelBuilder.add_link()` / `add_body()`, and the body-level USD-authored `newton:armature` read. Add any isotropic artificial inertia directly to `inertia` instead.

The remaining `add_link()` / `add_body()` parameters after `xform` are now keyword-only. Joint armature and the joint-level `newton:armature` schema mapping are unaffected.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_body_armature_removed
uv run --extra dev -m newton.tests -k test_import_body_newton_armature_ignored
uv run --extra dev -m newton.tests -k test_joint_controllers
```

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed deprecated “armature” support from rigid-body creation; provide `mass` and `inertia` explicitly.
  * Removed the `default_body_armature` property.
  * After the initial transform, body configuration parameters are now keyword-only.
  * USD-authored `newton:armature` at the body level is now ignored silently (no inertia contribution); use `inertia` instead.

* **Tests**
  * Updated USD import and model/joint controller tests to stop using `armature`, and adjusted inertia expectations accordingly.

### DIFF
--- a/newton/_src/sim/builder.py
+++ b/newton/_src/sim/builder.py
@@ -3860,8 +3860,8 @@ class ModelBuilder:
 
     def add_link(
         self,
-        xform: Transform | None = None,
         *,
+        xform: Transform | None = None,
         com: Vec3 | None = None,
         inertia: Mat33 | None = None,
         mass: float = 0.0,
@@ -3945,8 +3945,8 @@ class ModelBuilder:
 
     def add_body(
         self,
-        xform: Transform | None = None,
         *,
+        xform: Transform | None = None,
         com: Vec3 | None = None,
         inertia: Mat33 | None = None,
         mass: float = 0.0,
--- a/newton/_src/sim/builder.py
+++ b/newton/_src/sim/builder.py
@@ -7,7 +7,6 @@ from __future__ import annotations
 
 import copy
 import ctypes
-import inspect
 import math
 import warnings
 from collections import Counter, deque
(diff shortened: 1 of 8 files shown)


newton/_src/sim/builder.py
```diff
@@ -4104,7 +4034,7 @@ def add_link(
     def add_body(
         self,
         xform: Transform | None = None,
-        armature: float | None = None,
+        *,
```
eric-heiden, line 3965: Should we make `xform` also a kwarg-only argument?

newton/_src/sim/builder.py
```diff
@@ -3957,57 +3946,10 @@ def _coerce_mat33(value: Any) -> wp.mat33:
...
-    @default_body_armature.setter
-    def default_body_armature(self, value: float) -> None:
-        self._warn_default_body_armature_deprecated()
-        self._default_body_armature = value
-
     def add_link(
         self,
         xform: Transform | None = None,
-        armature: float | None = None,
+        *,
```
eric-heiden, line 3880: Same here, not sure if enforcing all args to be keywords arguments is better.

newton/_src/sim/builder.py
adenzler-nvidia: Done — `xform` is now keyword-only on both `add_link()` and `add_body()`, so the full signature is keyword-only (matching `add_shape()`). Updated the two positional call sites in the tests.

newton/_src/sim/builder.py
adenzler-nvidia: Went with all-keyword-only now — `xform` included — so the whole signature is keyword-only and consistent with `add_shape()` and the importer methods.

jcarius-nv (approved):

PR #3176, opened 2026-06-12
Adds a newton.controllers module

jeff-hough:
## Description

Please see the temporary `DESIGN_DOC.md` in the file diffs. This PR is for early review of design ideas with some sample controllers implemented. DO NOT MERGE.

## Checklist

## Release Notes
* **New Features**
  * Added a new controllers API with a shared `Controller` interface and per-step input/output struct lifecycle.
  * Introduced `ControllerPID` (stateful, with anti-windup integral clamping).
  * Introduced `ControllerDifferentialDrive` and `ControllerDifferentialKinematics` (one-step IK with DLS or transpose).
  * Added a public controllers entry point and new controller examples (diff-drive swarm, Diff IK Panda, PID pendulum).
* **Documentation**
  * Added a temporary design document for the controllers architecture.
* **Tests**
  * Added unit tests (including gradient checks) and example-run integration tests for the new examples.

### DIFF
--- a/newton/_src/controllers/DESIGN_DOC.md
+++ b/newton/_src/controllers/DESIGN_DOC.md
@@ -1,0 +1,381 @@
+# Newton Controllers — Design Doc
+
+**NOTE** This is a temporary document for design review purposes only. It should be deleted
+and replaced by proper documentation before merging.
+
+## Background
+
+There are many controllers implemented across Isaac Lab/Sim. The goal of this module is to centralize them in Newton, re-implementing each so that they are completely vectorized and CUDA-graphable.
+
+Below is a table summarizing existing Isaac Sim/Lab controllers, where their implementation currently lives, and whether there is a replacement already completed in this module.
+
+| Controller | Current Repo | Completed? |
+| --- | --- | --- |
+| Differential IK | Isaac Lab | :heavy_check_mark: |
+| PID | — | :heavy_check_mark: |
+| Operational Space | Isaac Lab | :x:  |
+| Joint Impedance | Isaac Lab | :x: |
+| Differential Drive | Isaac Sim | :heavy_check_mark: |
+| Holonomic Drive | Isaac Sim | :x: |
+| Ackermann | Isaac Sim | :x: |
+
+An open question is whether to broaden this "control toolbox" beyond the controllers in the table above to include general-purpose algorithms common in robotics. For example, a linear-filter class could implement:
+
+- low-pass
+- band-pass
+- notch
+---
+
+## Architecture
+
+The module exposes a single abstract base class — `Controller(ABC)` — that every concrete control law (`ControllerPID`, `ControllerDifferentialKinematics`, `ControllerDifferentialDrive`, …) subclasses directly. There is **no framework-level composition**: callers wanting to combine multiple control laws invoke each one's `compute()` in sequence themselves.
+
+`Controller` declares the surface every law implements:
+
+- `is_graphable() -> bool` — CUDA-graph compatibility predicate.
+- `is_stateful() -> bool` — whether the law carries state between steps.
+- `state()` — allocate a fresh per-step `State` (subclass-specific `@dataclass`), or `None` for stateless laws.
+- `input_struct()` — allocate a fresh, auto-generated dataclass holding one `wp.array` field per input port.
+- `output_struct()` — same idea for write ports.
+- `compute(input_struct, output_struct, controller_state_now, controller_state_next, time_step) -> None`.
+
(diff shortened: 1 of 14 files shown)


newton/_src/controllers/impl/controller_diff_ik.py
```diff
@@ -0,0 +1,623 @@
...
+        """Jacobian inverse strategy."""
+
+        DAMPED_LEAST_SQUARES = 0
+        TRANSPOSE = 1
+
+    DEFAULT_SOLVER_DAMPING: float = 0.05
+
+    def __init__(
+        self,
+        model_builder: ModelBuilder,
```
hujc7, line 290: Is the goal to provide a general controller even for different backends? I think we might need something general to replace this so it can also be used by other backends. In general, I think it needs to assume the caller side has checked assumptions, e.g. fresh_fk.

newton/_src/controllers/controller.py
```diff
@@ -0,0 +1,96 @@
...
+        construction. Each field is a freshly :func:`wp.zeros`-allocated
+        array of the right dtype and size. The user typically reassigns
+        these fields to point at live data buffers.
+        """
+
+    @abstractmethod
+    def output_struct(self) -> Any:
+        """Allocate a fresh output container with one field per output port."""
+
+    @abstractmethod
```
hujc7, line 69: I think a reset function might be needed in general for env resets

newton/_src/controllers/DESIGN_DOC.md
```diff
@@ -0,0 +1,381 @@
...
+## Architecture
+
+The module exposes a single abstract base class — `Controller(ABC)` — that every concrete control law (`ControllerPID`, `ControllerDifferentialKinematics`, `ControllerDifferentialDrive`, …) subclasses directly. There is **no framework-level composition**: callers wanting to combine multiple control laws invoke each one's `compute()` in sequence themselves.
+
+`Controller` declares the surface every law implements:
+
+- `is_graphable() -> bool` — CUDA-graph compatibility predicate.
+- `is_stateful() -> bool` — whether the law carries state between steps.
+- `state()` — allocate a fresh per-step `State` (subclass-specific `@dataclass`), or `None` for stateless laws.
+- `input_struct()` — allocate a fresh, auto-generated dataclass holding one `wp.array` field per input port.
```
hujc7, line 38: what if caller side already allocates those? directly reading and writing to callers owned buffers could be supported?

hujc7:
For the graphable part, is it supported such that if caller puts the call also in a graphed path?

newton/_src/controllers/DESIGN_DOC.md
```diff
@@ -0,0 +1,381 @@
...
+- notch
+---
+
+## Architecture
+
+The module exposes a single abstract base class — `Controller(ABC)` — that every concrete control law (`ControllerPID`, `ControllerDifferentialKinematics`, `ControllerDifferentialDrive`, …) subclasses directly. There is **no framework-level composition**: callers wanting to combine multiple control laws invoke each one's `compute()` in sequence themselves.
+
+`Controller` declares the surface every law implements:
+
+- `is_graphable() -> bool` — CUDA-graph compatibility predicate.
```
hujc7, line 35: are there cases the controller might not be graphable? is it possible to avoid that?

newton/_src/controllers/DESIGN_DOC.md
jeff-hough: Yes, that is fully supported! The `input_struct` and `output_struct` functions are only convenience functions.

newton/_src/controllers/controller.py
jeff-hough: Does that need to be in this library? The state is managed external to this class. To me, `reset` is just the user creating whatever `state` object they want as their initial condition, and passing that to the `compute` function - or am I missing something else that is needed?

newton/_src/controllers/impl/controller_diff_ik.py
jeff-hough: Yes, the goal is to work with any backend. We read the data from anywhere, pass it to these controllers, and then the output can go anywhere.

In the controllers which require a robot model, I assume that the user passes a newton model. However, all controllers will also have `ModelFree` variants if they just wish to pass computed data (i.e. computed Jacobian, mass matrix, gravity vector, etc).

>In general, I think it needs to assume the caller side has checked assumptions, e.g. fresh_fk.
Could you expand on this? I am not sure I understand what you mean exactly.

jeff-hough:
This PR is out of date.

Issue #2726, opened 2026-05-06
Reduce `gh-pages` Bloat From Documentation Deploys


christophercrouzet:
The `gh-pages` branch grows by tens of megabytes per push because the Sphinx output is copied wholesale, including artifacts that no user agent loads but which regenerate with non-deterministic bytes every build. Git keeps every delta forever. See https://github.com/newton-physics/newton/issues/2352.

Main offenders:
- `.doctrees/` — Sphinx's pickled doctree cache. `env.pickle` alone is ~10 MB and embeds absolute build-host paths, so the bytes differ every build.
- `__pycache__/`, `*.pyc` — appear when `conf.py` or extensions import Python modules at build time.
- Object reprs with memory addresses — autodoc renders default values via `repr()`, so any default whose class lacks a stable `__repr__` produces `<Foo object at 0x7f...>` in the HTML, with the address changing every build (ASLR). Set `autodoc_preserve_defaults = True` in `conf.py`, or give those classes a deterministic `__repr__`.

preist-nvidia:
Thank you for tackling this @christophercrouzet much appreciated!

jcarius-nv:
@christophercrouzet is this issue fixed with the PR #3186 or is there more work to do?

christophercrouzet:
Yes, it's fine to close. The remaining bits that are still in the way to make the doc 100% deterministic are the Viser recordings that are generated by running (non-deterministic) simulations from the notebooks, but that's a different piece of work.

PR #3186, opened 2026-06-15
Stop deploying Sphinx's .doctrees cache to gh-pages

christophercrouzet:
## Description

Sphinx writes its pickled doctree cache to `<outdir>/.doctrees` by default, so the documentation deploy workflows were publishing the whole ~25 MB cache to the `gh-pages` branch on every build — build-only state that no browser ever loads and that regenerates with non-deterministic bytes each run.

This passes `-d docs/_build/doctrees` to `sphinx-build` in both the `docs-dev` and `docs-release` workflows, so the cache is written as a sibling of `docs/_build/html` and is no longer deployed. The published tree drops from ~50 MB to ~26 MB (excluding the Viser recordings), and two of the three files that previously churned on every build disappear.

It also makes the build a little more deterministic:
- the robotics tutorial's CUDA-graph image now has explicit `alt` text, so nbsphinx no longer renders a random per-build `uuid4` substitution name as the image's `alt` attribute;
- a `strip_repr_addresses()` `doctree-resolved` hook is added as defense-in-depth against autodoc embedding ASLR memory addresses (`<… object at 0x…>`) in default reprs (a no-op on today's API).

Addresses #2726. This is a focused first step that intentionally keeps the existing `gh-pages` orphan-reset + force-push deploy model, so the branch stays bounded exactly as it is today. The remaining per-build churn — most notably the non-deterministic Viser `.viser` recordings (~13 MB/push) regenerated by executing the tutorial notebooks — is left for a follow-up, since removing it requires either committing the recordings or moving the deploy to a commit-on-top model backed by a fully deterministic build.

## Checklist

- [ ] New or existing tests cover these changes — docs deployment has no automated tests; verified manually (see Test plan).
- [x] The documentation is up to date with these changes — CI/build-only change; no documentation content affected.
- [ ] `CHANGELOG.md` has been updated (if user-facing change) — intentionally omitted; internal build/CI reproducibility, not a user-facing change.

## Test plan

Built the docs twice from the same commit and confirmed the doctrees cache is gone and the infra-level churn is eliminated:

```bash
uv run --extra docs --extra sim sphinx-build -j auto -d /tmp/dt1 -b html docs /tmp/h1
uv run --extra docs --extra sim sphinx-build -j auto -d /tmp/dt2 -b html docs /tmp/h2
test ! -d /tmp/h1/.doctrees           # .doctrees no longer in the published tree
du -sh /tmp/h1                         # ~26 MB (was ~50 MB)
diff -r /tmp/h1 /tmp/h2                # byte-identical with nbsphinx_execute=never
```

- `tutorials/01_robotics.html` renders the stable `alt="CUDA graph capture of a simulation step"` and no random 32-hex `alt`.
- `uvx pre-commit run --files …` passes on all changed files.

## Release Notes

* **Documentation**
  * Enhanced documentation consistency and output quality across builds.
  * Improved tutorial accessibility by adding descriptive alternative text for images, supporting screen readers and users with visual impairments.
  * Optimized documentation build and publication workflow for cleaner artifacts.

### DIFF
--- a/.github/workflows/docs-dev.yml
+++ b/.github/workflows/docs-dev.yml
@@ -41,7 +41,10 @@ jobs:
         uses: pandoc/actions/setup@86321b6dd4675f5014c611e05088e10d4939e09e  # v1.1.1
 
       - name: Build Sphinx documentation
-        run: uv run --extra docs --extra sim sphinx-build -j auto -b html docs docs/_build/html
+        # `-d docs/_build/doctrees` keeps Sphinx's pickled doctree cache out of
+        # the published html/ tree (GH-2726): it is build-only state, ~25 MB of
+        # non-deterministic bytes that no browser loads.
+        run: uv run --extra docs --extra sim sphinx-build -j auto -d docs/_build/doctrees -b html docs docs/_build/html
         env:
           NEWTON_REQUIRE_PANDOC: "1"
 
--- a/.github/workflows/docs-release.yml
+++ b/.github/workflows/docs-release.yml
@@ -74,7 +74,10 @@ jobs:
 
       - name: Build Sphinx documentation
         if: steps.version.outputs.SHOULD_DEPLOY == 'true'
-        run: uv run --extra docs --extra sim sphinx-build -j auto -b html docs docs/_build/html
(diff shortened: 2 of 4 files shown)


eric-heiden (approved):
Thanks, LGTM!

PR #3187, opened 2026-06-15
Optimize ModelBuilder collision filter replication

### LINKED ISSUE #1675: Improve `ModelBuilder.add_builder` performance

eric-heiden:
## Description

Fixes #1675.

This PR reduces `ModelBuilder` construction overhead for large replicated robot scenes. Replicated shape collision filter pairs are kept in a compact block representation instead of eagerly expanding duplicate Python pair lists, and `finalize()` builds a canonical filter store as a sorted packed integer pair-code array (replication is a broadcast offset add, membership is a `searchsorted`). Contact-pair generation replays per-world templates with NumPy broadcasting instead of building millions of Python tuples.

`Model.shape_collision_filter_pairs` remains a plain `set` view that materializes on access; mutating it after `finalize()` is deprecated. Solver integrations query the canonical store through `Model.shape_collision_filter_contains()` (scalar), `Model.shape_collision_filter_mask()` (bulk), and `Model.shape_collision_filter_pairs_array()` (the sorted `[N, 2]` exclusion array consumed by the collision pipeline and Kamino).

Residual explicit filters, such as global ground filters added after replication, participate in the compact contact-pair template path, as do filter blocks from sub-builders added outside a world context. The fast path falls back to the general layout-agnostic path whenever its assumptions do not hold (world starts not built, world layout mismatch with skipped validation, or a model filter set mutated after finalize).

The PR also avoids large Python default-value lists for empty numeric custom attributes by using `wp.full()`, and caches mesh content hashes with a framed SHA digest so repeated mesh deduplication does not repeatedly flatten and tuple-convert mesh arrays.

## Benchmark summary

Benchmarks were run locally on `cuda:0` against `origin/main` at `c336b7ae6054`, with both sides re-measured under identical conditions after the final review pass. `finalize()` time at 8192 worlds:

| Example | Baseline ms | Current ms | Delta |
| --- | ---: | ---: | ---: |
| `robot_allegro_hand` | 9828.84 | 4156.16 | -5672.68 ms (-57.7%) |
| `robot_anymal_d` | 14597.63 | 5140.24 | -9457.39 ms (-64.8%) |
| `robot_cartpole` | 717.55 | 513.35 | -204.20 ms (-28.5%) |
| `robot_g1` | 59173.86 | 15997.64 | -43176.22 ms (-73.0%) |
| `robot_h1` | 18411.58 | 6933.83 | -11477.76 ms (-62.3%) |
| `robot_panda_hydro` | 14443.43 | 5818.44 | -8624.99 ms (-59.7%) |
| `robot_ur10` | 8582.54 | 3377.27 | -5205.27 ms (-60.6%) |

Full world-count sweep tables (1 to 8192 worlds) are in the latest PR comment. The compact fast path is verified bit-identical against the general path on a 2.8M contact-pair scene.

More results on env creation in Isaac Lab based on this PR:
https://reports.eric-heiden.com/modelbuilder-perf/

## Checklist

- [x] The documentation is up to date with these changes (collision docs updated for the filter mutation deprecation)
- [x] `CHANGELOG.md` has been updated (Added, Deprecated, and Fixed entries)

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_empty_numeric_custom_attribute_uses_wp_full_default -k test_empty_vector_custom_attribute_uses_wp_full_default -k test_mesh_hash_uses_cached_sha_digest -k test_finalize_deduplicates_equal_mesh_content -k test_large_replicated_collision_filter_pairs_deprecate_mutation_and_preserve_contacts -k test_compact_replicated_collision_filters_allow_residual_filters -k test_collision_filter_array_queries_match_set -k test_mixed_replicated_and_global_builder_filters_preserve_contacts
uvx ruff check newton\_src\sim\builder.py newton\_src\sim\model.py newton\tests\test_model.py
```

Full sweep after the array-native store landed:

```bash
uv run --extra examples --extra torch-cu12 python reports\modelbuilder_robot_benchmark.py sweep-finalize --repo . --branch-label array-native-review --output reports\modelbuilder_robot_finalize_world_counts_array_native.json --repeats 1 --device cuda:0 --timeout-s 900
```

## Bug fix

**Steps to reproduce:**

1. Build a robot template with shape collision filter pairs.
2. Replicate it across many worlds with `ModelBuilder.replicate()` or repeated `add_world()` calls.
3. Add a small number of explicit residual filters, such as global ground filters, after replication.
4. Call `ModelBuilder.finalize()`.
5. Observe construction time and host memory scale poorly if compact replicated filters are forced to materialize.

**Minimal reproduction:**

```python
import newton

robot = newton.ModelBuilder()
body0 = robot.add_body()
shape0 = robot.add_shape_box(body=body0, hx=0.5, hy=0.5, hz=0.5)
body1 = robot.add_body()
shape1 = robot.add_shape_box(body=body1, hx=0.5, hy=0.5, hz=0.5)
robot.add_shape_collision_filter_pair(shape0, shape1)

ground = builder.add_ground_plane()
builder.replicate(robot, 8192)

for world in range(8192):
    builder.add_shape_collision_filter_pair(ground, 1 + world * robot.shape_count)

model = builder.finalize()
```

Note: use `add_shape_collision_filter_pair()` to add filters. Reading the `builder.shape_collision_filter_pairs` list property materializes the compact block storage into a plain list, which keeps working but gives up the compact representation.

## Summary by CodeRabbit
* **Performance & Optimization**
  * Reduced memory usage in large replicated scenes by keeping collision-filter pairs compact/lazy and speeding up contact-pair generation via metadata-aware reconstruction.
  * Improved model build efficiency by deduplicating finalized geometry sources by object identity and using deterministic mesh content hashing for cache reuse.
  * Optimized custom-attribute array materialization for “all-default” cases.
* **Bug Fixes**
  * Fixed performance/memory regressions during `add_builder()` / `finalize()` with many replicated worlds.
* **Tests**
  * Added coverage for mesh hash caching/deduplication, identity-based sharing, and compact collision-filter/contact-pair behavior under replication and mutation.


eric-heiden:
@coderabbitai review

newton/_src/sim/builder.py
```diff
@@ -10540,9 +10787,24 @@ def finalize(
...
+            explicit_filter_pairs = tuple(self.shape_collision_filter_pairs.explicit_pairs)
+            compact_filter_blocks = self.shape_collision_filter_pairs.compact_contact_blocks
+            if (
+                len(self.shape_collision_filter_pairs) > self._SHAPE_COLLISION_FILTER_PAIR_SET_LIMIT
+                and compact_filter_blocks
+            ):
+                # For very large replicated scenes, avoid building a second
+                # Python set of all filter tuples. The contact-pair fast path
+                # below can consume compact blocks plus explicit residual filters.
+                m.shape_collision_filter_pairs = _ShapeCollisionFilterPairs(
```
camevor: nit: The type annotation for `model.shape_collision_filter_pairs` should be updated (e.g. `Collection[tuple[int, int]]`) to reflect this new behavior.

adenzler-nvidia: This changes the public shape of `Model.shape_collision_filter_pairs` for large compact scenes. `Model` still documents/annotates this field as `set[tuple[int, int]]`, but this path assigns `_ShapeCollisionFilterPairs`, which only supports a subset of set behavior.

That makes behavior threshold-dependent: small scenes get a real set, large compact scenes get a custom iterable. User/downstream code doing `.add()`, `.discard()`, set algebra, or `isinstance(..., set)` can break only once the compact threshold is crossed. It also does not preserve set cardinality when duplicates exist: compact finalization can report `len(filters)` including duplicate explicit/block pairs, while the materialized fallback deduplicates them.

Could we keep the public field as a real set, or move the compact representation to a separate private/internal field used by contact generation? If the wrapper stays public, I think it needs to preserve set semantics, including deduplication and either supported mutation or a clearly updated public contract.

newton/_src/sim/builder.py
```diff
@@ -11568,8 +11830,172 @@ def find_shape_contact_pairs(self, model: Model):
...
+                # cache can then be keyed by local layout rather than absolute shape
+                # ids, which lets homogeneous replicated worlds share one result.
+                blocks_by_world.setdefault(world, []).append(
+                    (shape_start - world_start, shape_count, local_filter_pairs)
+                )
+
+            if use_filter_blocks:
+                for filter_a, filter_b in explicit_filter_pairs:
+                    s1, s2 = (filter_a, filter_b) if filter_a <= filter_b else (filter_b, filter_a)
+                    world1 = self.shape_world[s1]
```
adenzler-nvidia: The compact path indexes `self.shape_world[s1]` / `self.shape_world[s2]` while classifying explicit residual filters. That creates compact-only behavior for out-of-range filter pairs: with compact blocks present, `finalize()` raises a raw `IndexError`; with blocks materialized/fallback, the same invalid pair is just stored in the finalized set and does not fail during contact-pair generation.

If invalid shape filter pairs are unsupported, we should validate them consistently before either path and raise a clear `ValueError`. If they are currently tolerated, the compact path should preserve that behavior. The current threshold-dependent failure mode seems hard to diagnose.

adenzler-nvidia (commented):
Focused behavioral review of the compact collision-filter path.

newton/_src/sim/builder.py
```diff
@@ -11568,8 +11830,172 @@ def find_shape_contact_pairs(self, model: Model):
             - Sets `model.shape_contact_pairs` to a wp.array of shape pairs (wp.vec2i).
             - Sets `model.shape_contact_pair_count` to the number of contact pairs found.
         """
-        filters: set[tuple[int, int]] = model.shape_collision_filter_pairs
-        contact_pairs: list[tuple[int, int]] = []
+        explicit_filter_pairs = tuple(self.shape_collision_filter_pairs.explicit_pairs)
+        compact_filter_blocks = self.shape_collision_filter_pairs.compact_contact_blocks
+        if compact_filter_blocks:
```
adenzler-nvidia: Non-blocking simplification thought: this compact path is doing three distinct jobs inline: validating compact block layout, classifying explicit residual filters, and building/replaying cached per-world contact templates. Factoring those into small private helpers would make the invariants easier to audit and would make future changes to residual-filter handling less risky.

newton/_src/sim/builder.py
```diff
@@ -77,6 +77,209 @@
...
+class _ShapeCollisionFilterBlock:
+    """Compact replicated collision-filter block."""
+
+    shape_start: int
+    local_pairs: tuple[tuple[int, int], ...]
+    world: int | None = None
+    shape_count: int = 0
+
+
+class _ShapeCollisionFilterPairs:
```
adenzler-nvidia: Naming nit: `_ShapeCollisionFilterPairs` reads like the general finalized filter-pair representation, but this type is specifically a lazy/block-backed set-like snapshot for the large compact path. A more explicit name, or keeping this behind a separate internal field, would make the distinction from the public/model `shape_collision_filter_pairs` contract clearer.

newton/_src/sim/builder.py
```diff
@@ -77,6 +77,209 @@
...
+                yield (block.shape_start + s1, block.shape_start + s2)
+
+    def __contains__(self, pair: object) -> bool:
+        try:
+            s1, s2 = pair  # type: ignore[misc]
+        except (TypeError, ValueError):
+            return False
+
+        canonical = (s1, s2) if s1 <= s2 else (s2, s1)
+        if self._set is None and len(self) <= self._materialize_limit:
```
adenzler-nvidia: This small-scene materialization branch is a little surprising because the wrapper is only constructed in normal flow when `len(self.shape_collision_filter_pairs) > _SHAPE_COLLISION_FILTER_PAIR_SET_LIMIT`. I think this exists mostly for forced-threshold tests/defensiveness, but it adds some conceptual noise to the large-scene representation. If the type stays, it may be clearer to either remove this branch or document why the lazy view supports a small-scene mode.

adenzler-nvidia (commented):
A few non-blocking simplification notes.

newton/_src/sim/builder.py
camevor: Agreed; gating on the length here seems counter-productive.

This is mostly a dead branch as the way the builder uses this class never allows this condition to be true. Also, it makes sense to avoid materializing the large set if it is not needed, but if membership does turn out to be used, we should probably lazily create the set here.

newton/_src/sim/builder.py
eric-heiden: Addressed in d1eb82734: `Model.shape_collision_filter_pairs` is now a read-only `frozenset[tuple[int, int]]`, and the docs/changelog were updated to describe the immutable finalized-model contract.

eric-heiden: Addressed in d1eb82734: I removed the lazy `_ShapeCollisionFilterPairs` model-facing wrapper entirely. `Model.shape_collision_filter_pairs` now returns an eager `frozenset[tuple[int, int]]` for all scenes, while compact block metadata stays builder-side for contact-pair generation. That removes the threshold-dependent public behavior and gives consistent deduplicated immutable-set semantics.

newton/_src/sim/builder.py
eric-heiden: Addressed in d1eb82734: filter pairs are validated before finalization/contact-pair generation now, including explicit residual filters used with compact blocks. Invalid shape indices raise a clear `ValueError`, and I added regression coverage for both normal explicit filters and compact residual filters.

newton/_src/sim/builder.py
eric-heiden: Addressed in d1eb82734 with a smaller split: compact-block validation and explicit residual validation are factored out, while residual classification plus template build/replay stayed inline with comments. That keeps the implementation simple per follow-up discussion while making the validation invariants auditable.

newton/_src/sim/builder.py
eric-heiden: Addressed in d1eb82734 by removing `_ShapeCollisionFilterPairs` entirely. The only remaining compact representation is builder-side (`_BuilderShapeCollisionFilterPairs` / compact blocks), and the finalized `Model` always exposes the eager read-only snapshot.

newton/_src/sim/builder.py
eric-heiden: Addressed in d1eb82734 by removing the lazy wrapper and its small-scene materialization branch altogether. `Model` now always receives an eager `frozenset`, and compact blocks are only used internally for builder contact-pair generation.

newton/_src/sim/builder.py
adenzler-nvidia: Stepping back: the change to `Model.shape_collision_filter_pairs` is a **breaking API change** — it was a public, mutable `set`, and this makes it a read-only `AbstractSet` (no `.add()`, no reassignment) with no deprecation cycle, whereas AGENTS.md asks breaking changes to be deprecated first. So before it lands, I want to check whether the break is actually necessary.

The `finalize()` win this PR targets comes from *not materializing the filter set at finalize* — which doesn't require `Model.shape_collision_filter_pairs` to be compact, only that finalize doesn't build it. The compact model-side view (`_CompactShapeCollisionFilterPairs`) is what forces the read-only change, and it also drops the filter set on viewer-file record→playback (the deserializer only restores native `set`/`list`/`dict`, not a custom container or a bare `frozenset`).

Would it be acceptable to keep the model field a plain (lazily-materialized) `set` and let solver-side consumers — e.g. the MuJoCo `in`-tests — materialize it on first access, as they did before this PR? That keeps the builder-side compaction and the contact-pair fast path (the bulk of the speedup) and avoids the API break and the deprecation question entirely. The only thing traded away is compact membership during solver setup, which could be a separate optimization later if it shows up in a profile.

Have you measured whether solver-side membership at high world counts is actually a bottleneck? If not, keeping it a native, editable `set` looks like the lower-risk path — same finalize numbers, no breaking change, and record→playback keeps working.

newton/_src/sim/builder.py
```diff
@@ -190,6 +419,15 @@ class ModelBuilder:
...
     )
+    _BODY_ARMATURE_ARG_DEPRECATION_MESSAGE = (
+        "ModelBuilder.add_link(..., armature=...) and ModelBuilder.add_body(..., armature=...) "
+        "are deprecated and will be removed in a future release. "
+        "Add any isotropic artificial inertia directly to 'inertia' instead."
+    )
+    _DEFAULT_BODY_ARMATURE_DEPRECATION_MESSAGE = (
+        "ModelBuilder.default_body_armature is deprecated and will be removed in a future release. "
+        "Add any isotropic artificial inertia directly to 'inertia' instead."
+    )
```
adenzler-nvidia: These two constants look like dead code that is also now inaccurate. `_BODY_ARMATURE_ARG_DEPRECATION_MESSAGE` and `_DEFAULT_BODY_ARMATURE_DEPRECATION_MESSAGE` aren't referenced anywhere in the tree, and they describe `add_link(..., armature=...)` / `add_body(..., armature=...)` / `default_body_armature` — API that this branch already removed (#3153), so the 'is deprecated and will be removed' wording no longer matches reality. They look like a leftover from the `origin/main` merge and are unrelated to this PR; safe to delete.

newton/_src/sim/builder.py
```diff
@@ -1005,7 +1264,7 @@ def __init__(self, up_axis: AxisType = Axis.Z, gravity: float = -9.81):
         # Mesh SDF storage (texture SDF arrays created at finalize)
 
         # filtering to ignore certain collision pairs
-        self.shape_collision_filter_pairs: list[tuple[int, int]] = []
+        self.shape_collision_filter_pairs = _BuilderShapeCollisionFilterPairs()
```
adenzler-nvidia: This also changes the public `ModelBuilder.shape_collision_filter_pairs` from `list` to a custom `MutableSequence`, which is a smaller breaking change in its own right: assigning a plain list (`builder.shape_collision_filter_pairs = [...]`) and then calling `finalize()` now raises `AttributeError` on `.compact_contact_blocks`, and `+` / `.sort()` / list-slice semantics change. Could `finalize()` tolerate a reassigned plain list (treat it as all-explicit pairs) so existing code that assigns a list keeps working?

adenzler-nvidia (commented):
Two additional findings, flagged inline.

newton/_src/sim/builder.py
eric-heiden: Addressed in 118ab1154. I kept the finalized-model API set-compatible during the deprecation cycle: Model.shape_collision_filter_pairs is an isinstance(..., set)-compatible mutable set, and existing mutation/reassignment still works while emitting DeprecationWarning. Its compact source remains private and is materialized only on public set access or mutation.

SolverMuJoCo now uses the public Model.shape_collision_filter_contains() API, so its membership tests use compact metadata without materializing the set. NxN/SAP use Model.shape_collision_filter_pairs_for_broad_phase(), which expands directly into the sorted exclusion list those broad phases require, without first materializing the public set. ViewerFile restores the property explicitly, and the regression test now starts from an unmaterialized compact model and verifies record/playback preserves the filters.

I also rechecked the G1 finalize path after merging current main: 1.819 s at 1,024 worlds and 14.431 s at 8,192 worlds, versus roughly 2.095 s and 15.15 s in the prior run.

newton/_src/sim/builder.py
eric-heiden: Addressed in 118ab1154. I removed both stale armature deprecation constants. I also audited the other private helpers added by this PR and removed the unnecessary MutableSequence facade and its list-emulation methods; the remaining helpers are all exercised by compact storage, contact generation, serialization, or solver/broad-phase consumers.

newton/_src/sim/builder.py
eric-heiden: Addressed in 118ab1154. ModelBuilder.shape_collision_filter_pairs is again an actual list at the public boundary. Compact replicated blocks live only in private builder storage; public access materializes that storage to a list, and assigning a list preserves that exact list and its normal ordering, copy, concatenation, slice, sort, and mutation behavior.

add_builder(), finalize(), and contact-pair discovery now handle either private compact storage or a user-reassigned plain list. I added regressions covering isinstance(..., list), ordering, copy/concatenation, source and destination list reassignment, add_builder(), and finalize().

newton/_src/geometry/types.py
```diff
@@ -1502,16 +1503,19 @@ def __hash__(self) -> int:
...
-                (
-                    tuple(np.array(self.vertices).flatten()),
-                    tuple(np.array(self.indices).flatten()),
-                    self.is_solid,
-                    self._compute_texture_hash(),
-                    self._roughness,
-                    self._metallic,
-                )
+            digest = hashlib.sha256()
+            digest.update(self._vertices.tobytes())
```
adenzler-nvidia: These updates hash the vertex and index buffers as one unframed byte stream. Because the digest does not include either array's shape or byte length, distinct meshes can produce exactly the same SHA input without finding a cryptographic collision: moving three `int32` indices across the boundary and viewing those 12 bytes as one additional `float32` vertex preserves `vertices.tobytes() + indices.tobytes()`.

I reproduced a two-triangle mesh and a different one-triangle mesh receiving the same hash; `ModelBuilder.finalize()` then assigned both shapes the same geometry pointer because `finalized_geos` is keyed by `hash(geo)`. Could we include type/shape or byte-length metadata before each component and add a regression test ensuring these layouts do not deduplicate?

adenzler-nvidia (commented):
Otherwise, this looks good to me.

newton/_src/geometry/types.py
eric-heiden: Addressed in 5205aa49f. Each mesh array is now framed with its field name, dtype, rank, and dimensions before its contents are added to the digest, so the vertex/index boundary is unambiguous.

The regression reproduces the reported construction: the two-triangle and one-triangle meshes have identical unframed vertices.tobytes() + indices.tobytes() streams, but now hash differently, both call finalize(), and receive distinct geometry pointers. I also verified that cached hashing and deduplication of genuinely equal mesh content still pass; the full pre-commit suite is clean.

adenzler-nvidia (dismissed):
**Approving — this is careful work and the fuzzing/tests hold up.** I verified the compact path against the eager path on randomized replicated scenes (worlds, groups, flags, residual global/local filters) with no mismatches, and all downstream consumers (MuJoCo, Kamino, collide, viewer recording) handle the new representation correctly.

Before or after merging, I'd like us to consider an alternative representation: making a packed integer array the canonical store for filter pairs (replication = broadcast offset add, membership = `searchsorted`/`isin`, canonicalization = vectorized min/max), with the public builder list and model set as materialized-on-access compat views. The motivation is that at the PR's own benchmark scale (8192 worlds, UR10-ish robot), the two remaining hot paths still do Python-tuple work that dominates what the filter laziness saves:

- The compact fast path in `find_shape_contact_pairs` builds the ~3M contact pairs as Python tuples via `contact_pairs.extend(...)` before `np.asarray`: 791 ms vs 35 ms for a broadcast replay of the same template — ~750 ms per finalize, which is larger than the ~28% the lazy filter set saves.
- `shape_collision_filter_pairs_for_broad_phase()` does `sorted(set(iter_pairs()))`, so nxn/sap pipelines still materialize all 524k pairs as tuples transiently: 128 ms vs 34 ms packed.

The array-native version would also delete most of the compat machinery this PR needs (the lazy source type, the read-path overrides on the set subclass, the dual compact/fallback paths — roughly 350–400 lines), keep one eager code path with no staleness protocol, and the canonical store is exactly the sorted `wp.vec2i` exclusion array that collide and Kamino already want. The mutation-deprecation wrapper on the model set would stay as-is. Trade-offs: scalar `contains` is ~30× slower than a set lookup, so the SolverMuJoCo per-pair loops would need to become bulk `isin` queries (they arguably should anyway), and importers appending pair-by-pair would buffer into a list and flush vectorized.

Note: this alternative is what Claude (Fable) suggested during review, including the measurements above — please double-check whether it actually makes sense for this code before investing in it.

If you'd rather land this as-is and treat the array-native store as a follow-up, that works too — the deprecation window this PR opens is the natural place to point the end state at an array-backed attribute, and the two vectorization items above are localized changes that don't require abandoning the block design.

chschuma-disney (dismissed):
Looks good from Kamino's side.

eric-heiden:
Thanks, Alain, for the thorough verification and for the array-native proposal. I decided to land it in this PR rather than defer it. Your measurements reproduced on my machine: the tuple replay in `find_shape_contact_pairs` went from 1293 ms to 93 ms with a broadcast replay of the per-world templates (bit-identical output), and the broad-phase exclusion list is now free because the canonical store is the array itself.

What changed (commits `abbc76a3b` to `d87d93f62`):

- The canonical store is a sorted packed `int64` pair-code array built vectorized in `finalize()`. The lazy source type, the read-path overrides on the set subclass, and the dual representations are gone (about 350 lines, as you estimated). `Model.shape_collision_filter_pairs` stays a materialize-on-access set view with the mutation deprecation unchanged, so record/playback and existing reads keep working.
- Query API: `shape_collision_filter_contains()` (scalar), `shape_collision_filter_mask()` (bulk), and `shape_collision_filter_pairs_array()` (the `[N, 2]` array consumed by collide and Kamino).
- The trade-offs you flagged are handled: the SolverMuJoCo per-pair loops became bulk mask queries, scalar `contains` normalizes NumPy ints (`np.int32 << 32` silently wraps), and importers needed no change since the builder keeps the compact block representation.
- The fast path now falls back to the general path when world starts are not built, when the world layout does not match, or when the model filter set was mutated after finalize. Blocks without a world assignment fold into residual pairs instead of disabling the fast path.

One gotcha worth flagging: `np.unique` on 1-D integers uses a hash-based path in NumPy 2.3+ that degrades badly on bit-packed pair codes (3.7 s alone at UR10-8192 scale). Sort plus mask dedup fixed it; the sorted order is needed for `searchsorted` anyway.

Fresh sweep against `origin/main`, both sides re-measured under identical conditions. `finalize()` at 8192 worlds:

| Example | Baseline ms | Current ms | Delta |
| --- | ---: | ---: | ---: |
| `robot_allegro_hand` | 9828.84 | 4156.16 | -5672.68 ms (-57.7%) |
| `robot_anymal_d` | 14597.63 | 5140.24 | -9457.39 ms (-64.8%) |
| `robot_cartpole` | 717.55 | 513.35 | -204.20 ms (-28.5%) |
| `robot_g1` | 59173.86 | 15997.64 | -43176.22 ms (-73.0%) |
| `robot_h1` | 18411.58 | 6933.83 | -11477.76 ms (-62.3%) |
| `robot_panda_hydro` | 14443.43 | 5818.44 | -8624.99 ms (-59.7%) |
| `robot_ur10` | 8582.54 | 3377.27 | -5205.27 ms (-60.6%) |

Every example matches or beats the numbers this PR originally claimed, now with one code path and no staleness protocol. The broadcast fast path is verified bit-identical against the general path at 2.8M pairs.

newton/_src/sim/builder.py
```diff
@@ -11732,7 +11949,223 @@ def find_shape_contact_pairs(self, model: Model):
...
+            )
+
+        # The fast path replays builder-side filters against the finalized
+        # world layout; require built world starts that match shape_world (the
+        # layout validation can be skipped in finalize) and a pristine model
+        # filter store (a mutated store must be honored by the general path).
+        use_filter_blocks = bool(world_filter_blocks)
+        if use_filter_blocks:
+            model_filters = model._shape_collision_filter_store()  # pyright: ignore[reportPrivateUsage]
+            use_filter_blocks = model_filters is not None and model_filters.packed_pairs() is not None
```
adenzler-nvidia: Question about the filter source here: the fast path replays the builder's current filters, while the general path below (and this method before the PR) reads them from `model`. During `finalize()` the two are identical, but since this method is public, mutating the builder's filters after finalize and re-running `find_shape_contact_pairs(model)` yields different contact sets depending on which path runs — and the path choice hinges on hidden state (e.g. whether a property read de-compacted the builder store earlier). The same question applies when `model` wasn't produced by this builder.

Is that intended? If not, would it work to extend this gate to also require that the builder's filters still match the model store, e.g.

```python
packed = model_filters.packed_pairs()
use_filter_blocks = packed is not None and np.array_equal(
    self._build_shape_collision_filter_packed(), packed
)
```

so the fast path only runs when replaying builder filters is provably equivalent to the model store, and everything else falls back to the general path? The repack is one vectorized sort — tens of ms at the 8192-world benchmark scale, well under 1% of finalize — and if even that matters, `finalize()` could pass its already-built array down so the hot-path check becomes an identity test.

newton/_src/sim/model.py
```diff
@@ -32,6 +33,269 @@
...
+        super().intersection_update(*canonical_others)
+
+    @override
+    def symmetric_difference_update(self, other: Iterable[tuple[int, int]]) -> None:
+        self._prepare_mutation()
+        super().symmetric_difference_update(self._canonical_pair(shape_a, shape_b) for shape_a, shape_b in other)
+
+    @override
+    def __ior__(self, other: Iterable[tuple[int, int]]):
+        self.update(other)
```
adenzler-nvidia: The in-place operators add a frame between the caller and `_prepare_mutation`'s `stacklevel=3`, so warnings from `model.shape_collision_filter_pairs |= {...}` get attributed to this line in model.py instead of the user's mutation site — and the default once-per-location warning filter then collapses all `|=` call sites into a single warning. Direct method calls and reassignment are attributed correctly. Giving `_prepare_mutation` a `stacklevel` parameter and having the operators call it with 4 (then delegating to the `super()` mutator directly) would fix the attribution.

newton/tests/test_model.py
```diff
@@ -700,6 +854,309 @@ def test_collision_filter_pairs_canonical_order(self):
...
+        """Source-builder filter cache should invalidate when pair contents change."""
+
+        source = ModelBuilder()
+        body0 = source.add_body()
+        shape0 = source.add_shape_box(body=body0, hx=0.5, hy=0.5, hz=0.5)
+        body1 = source.add_body()
+        shape1 = source.add_shape_box(body=body1, hx=0.5, hy=0.5, hz=0.5)
+        body2 = source.add_body()
+        shape2 = source.add_shape_box(body=body2, hx=0.5, hy=0.5, hz=0.5)
+        source.shape_collision_filter_pairs.append((shape0, shape1))
```
adenzler-nvidia: This test doesn't exercise what its name describes: `source.shape_collision_filter_pairs.append(...)` goes through the property getter, which de-compacts the source store to a plain list before the first `add_builder`, so the template cache is never populated or invalidated — the test covers the list-source path. Mutating via `source.add_shape_collision_filter_pair(...)` between the two `add_builder` calls keeps the source compact and exercises the cache invalidation the name refers to.

newton/_src/sim/model.py
```diff
@@ -32,6 +33,269 @@
...
+    def packed_pairs(self) -> np.ndarray | None:
+        """Sorted unique packed pair codes, or ``None`` after mutation."""
+        return self._packed
+
+    def contains_pair(self, shape_a: int, shape_b: int) -> bool:
+        """Return membership of a shape pair in any argument order."""
+        try:
+            # Normalize to Python ints: NumPy int32 inputs would overflow the
+            # 32-bit shift in the packed code and give wrong membership.
+            shape_a, shape_b = int(shape_a), int(shape_b)
```
adenzler-nvidia: Since this backs the new public query methods, the `int()` coercion sets a lenient contract: `shape_collision_filter_contains("0", "1")` returns True, and floats truncate, so `contains(0.9, 1.2)` reports the `(0, 1)` filter as a match; `mask_pairs`' `astype(np.int64)` does the same for float arrays. Suggest rejecting non-integers while the contract is still fresh: `operator.index()` here (drop the `try/except`), and a dtype-kind check in `mask_pairs` instead of the coercing `astype`. That keeps all real integer types including NumPy ints working, makes garbage raise `TypeError` instead of silently matching, and costs nothing — integer-array inputs skip the copy either way, and `operator.index` is no slower than `int()` for the scalar path. (Open question whether unsigned dtypes should stay accepted via an explicit cast.)

adenzler-nvidia:
Thanks for the update — the array-native store is a nice simplification over the previous revision. I did another full pass over the redesign and didn't find any correctness issues: the pair-code packing and `searchsorted` membership check out at the boundaries, the rewritten MuJoCo body-exclusion and coloring paths preserve the old semantics (including the vacuous-exclusion edge for bodies without colliding shapes), and post-finalize mutation is now handled consistently across all query paths. The new tests also pass locally on GPU.

I left a few small comments — one question about which filter source the fast path should honor after finalize, plus minor nits on the new query APIs' input handling and one test. Nothing blocking; looks good otherwise.

newton/tests/test_model.py
eric-heiden: Addressed in 4b74320ff. The test now uses `source.add_shape_collision_filter_pair()` before the first `add_builder()` and again between copies, so the source remains compact, the first copy populates the template cache, and the second mutation invalidates it through the compact store. The second copy asserts that both the original and newly added filter pairs are present.

newton/_src/sim/builder.py
eric-heiden: Addressed in 4b74320ff. `ModelBuilder.find_shape_contact_pairs()` is now deprecated for Newton 1.4 because contact pairs are generated by `finalize()`. Finalization calls the private compact implementation with the already-built filter store, while deprecated public calls always disable builder-side compact blocks and use the model-backed general path. This keeps calls correct after builder mutation or with a model from another builder without repacking filters during `finalize()`. The regression covers both builder/model divergence and mutation of the finalized model, and separately asserts that `finalize()` builds the packed filter array only once.

newton/_src/sim/model.py
eric-heiden: Addressed in 4b74320ff. Scalar queries now use `operator.index()` and accept `SupportsIndex`, preserving Python and NumPy integer inputs while rejecting strings and floats with `TypeError`. Bulk queries reject non-integer dtypes, accept signed integer arrays directly, and explicitly range-check and cast unsigned arrays to `int64`. Tests cover NumPy integer scalars, unsigned arrays, floats, and strings.

newton/_src/sim/model.py
eric-heiden: Addressed in 4b74320ff. The in-place operators now call `_prepare_mutation()` directly and delegate to the corresponding `super()` mutator, removing the extra frame. With that call chain, `stacklevel=3` points to the user mutation site. The descriptor setter also ignores the same object returned by augmented assignment, so each operation emits exactly one warning and preserves set identity. A regression checks `|=`, `&=`, `-=`, and `^=` for warning count, filename, line number, and identity.

adenzler-nvidia (dismissed):
All review comments addressed at 343c6faef — thanks! Verified the digest framing, the model-backed deprecated path, the warning stacklevels, and the strict query-input validation against the branch.

adenzler-nvidia (approved):
Re-approving after the main merge (9b9312ef4) — only change is the merge itself with a hand-resolved CHANGELOG (drops entries duplicated from main's Changed section, folds the stray second Deprecated heading from the #3026 merge into the existing section).

chschuma-disney (approved):

PR #3190, opened 2026-06-16
Fix ViewerFile running state after close

### LINKED ISSUE #3094: [BUG] ViewerFile.is_running() returns True after viewer.close() -- because it's not implemented
### Bug Description

When using `Newton.Viewer.ViewerGL`, calling `viewer.close()` changes some internal flag such that `viewer.is_running()` returns False afterwards, as would be suggested by the function names. However, this is not the case with `ViewerFile`: even after `viewer.close()` is called, `viewer.is_running()` continues to return True.

Looking at the code, this is because`ViewerFile` does not implement `is_running()`, and in `ViewerBase` the default behavior is to return True. Thus, it seems like an easy fix could be to just add a flag to `ViewerFile` to track whether `ViewerFile.close()` has been called.

This inconsistency/bug causes an issue when a main simulation loop is implemented as:
```
while viewer.is_running(): 
    ...
```
This while loop can be terminated by `viewer.close()` when `viewer` is an instance of `ViewerGL`, but this termination no longer works if `viewer` is an instance of `ViewerFile`. Being able to easily swap between viewer types while using this pattern seems like something that is worth supporting...?

I have attached a reproduction script for reproducing this issue. It is a modified version of `example_recording.py`. If you set `HEADLESS` on line 129 to True, then Newton will run with `ViewerFile`, and you will see that although the log file is saved when the max frame count is reached and thus `viewer.close()`is called, the simulation loop will continue, and frames will continue to be saved. If you set `HEADLESS` to False, then Newton will correctly terminate after the maximum frame count has been exceeded.

I can submit a PR for this soon, if this is a problem deemed worth fixing and a contribution is welcome. Just let me know!

### Reproduction Script
This is a modified version of `example_recording.py` made to show how the current lack of a `viewer.is_running()` implementation for `ViewerFile` prevents workflows that have easy switching between a GUI render mode and a headless log-file mode.

Changes were only made near the end of the file. There is a comment indicating where changes were made.

```python

###########################################################################
# Example Recording
#
# Shows how to record a simulation using ViewerFile for automatic recording
# of model structure and state data during simulation.
#
# Recording happens automatically - ViewerFile captures all logged states
# and saves them when the viewer is closed.
#
# Command: python -m newton.examples recording
#
###########################################################################

import numpy as np
import warp as wp

import newton
import newton.examples


class Example:
    def __init__(self, viewer, _args):
        # Setup simulation parameters
        self.fps = 60
        self.frame_dt = 1.0 / self.fps
        self.sim_time = 0.0
        self.sim_substeps = 10

        self.viewer = viewer
        self.world_count = 100

        # Set numpy random seed for reproducibility
        self.seed = 123
        self.rng = np.random.default_rng(self.seed)

        start_rot = wp.quat_from_axis_angle(wp.normalize(wp.vec3(*self.rng.uniform(-1.0, 1.0, size=3))), -wp.pi * 0.5)

        mjcf_filename = newton.examples.get_asset("nv_humanoid.xml")

        articulation_builder = newton.ModelBuilder()
        articulation_builder.add_mjcf(
            mjcf_filename,
            ignore_names=["floor", "ground"],
            up_axis="Z",
        )

        # Joint initial positions
        articulation_builder.joint_q[:7] = [0.0, 0.0, 1.5, *start_rot]

        for _i in range(self.world_count):
            articulation_builder.joint_q[7:] = self.rng.uniform(
                -1.0, 1.0, size=(len(articulation_builder.joint_q) - 7,)
            ).tolist()
            builder.add_world(articulation_builder)
        builder.add_ground_plane()

        # Finalize model

        self.solver = newton.solvers.SolverMuJoCo(
            self.model,
            use_mujoco_cpu=False,
            solver="newton",
            integrator="euler",
            iterations=10,
            ls_iterations=5,
            njmax=100,
        )

        self.state_0, self.state_1 = self.model.state(), self.model.state()

        self.use_cuda_graph = wp.get_device().is_cuda

        if self.use_cuda_graph:
                self.simulate()
            self.graph = capture.graph

        # Set model in viewer (ViewerFile will automatically record it)

    def simulate(self):
        for _ in range(self.sim_substeps):
            self.state_0.clear_forces()
            self.solver.step(self.state_0, self.state_1, self.control, None, self.sim_dt)

    def step(self):
        with wp.ScopedTimer("step", active=False):
            if self.use_cuda_graph:
                wp.capture_launch(self.graph)
            else:
                self.simulate()

    def render(self):
        self.viewer.log_state(self.state_0)  # ViewerFile automatically records this
        self.viewer.end_frame()

    def test_final(self):
        pass


if __name__ == "__main__":
    # Create ViewerFile for automatic recording
    import sys

    recording_file = "humanoid_recording.bin"

    # Check if user wants a different filename from command line
    if len(sys.argv) > 1 and not sys.argv[1].startswith("-"):
        recording_file = sys.argv[1]

    print(f"Recording simulation to: {recording_file}")
    print("ViewerFile will automatically save when the simulation ends.")

    parser = newton.examples.create_parser()
    args = parser.parse_args()

    # -------------------------- BELOW IS THE CHANGED SECTION TO SHOW BUG --------------------------
    HEADLESS = True #CHANGE THIS TO TOGGLE BEHAVIOR
    if HEADLESS:
        viewer = newton.viewer.ViewerFile(recording_file, auto_save=False)
    else:
        viewer = newton.viewer.ViewerGL(headless=False)

    # Create example

    # Run for a reasonable number of frames
    max_frames = 1000
    frame_count = 0

    # while frame_count < max_frames:
    while viewer.is_running():
        example.step()
        example.render()
        frame_count += 1

        if frame_count % 100 == 0:
            print(f"Frame {frame_count}/{max_frames}")
        
        if frame_count >= max_frames:
            viewer.close()


    # Close viewer (automatically saves recording)
    # viewer.close()
    print(f"Recording completed: {recording_file}")
```

### System Information

- Newton version: commit a046d613
- Python version: 3.11.14
- OS: Ubuntu 24.04.3
- Kernel: 6.17.0-29 generic
- GPU: RTX4070 mobile
- Nvidia driver: 580.159.03
- CUDA version: 13.0

Langford-H:
## Description

Fixes #3094.

`ViewerFile.close()` already saves recorded snapshots, but `ViewerFile` did not implement `is_running()`, so it inherited `ViewerBase.is_running()` and continued to report `True` after close.

This change adds a `ViewerFile` running-state flag, makes `is_running()` report that flag, and marks the recorder as no longer running after `close()` saves the final recording.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests.test_recorder -k test_viewer_file
```

Verified:
- `test_viewer_file_is_running_reflects_close`
- `test_viewer_file_playback_cpu`

## Bug fix

**Steps to reproduce:**

1. Create a `ViewerFile`.
2. Check `viewer.is_running()`.
3. Call `viewer.close()`.
4. Check `viewer.is_running()` again.
5. Without this PR, it still returns `True`.

**Minimal reproduction:**

```python
import tempfile

import newton

with tempfile.NamedTemporaryFile(suffix=".json") as tmp:
    viewer = newton.viewer.ViewerFile(tmp.name, auto_save=False)

    print(viewer.is_running())  # True

    viewer.close()

    print(viewer.is_running())  # True before this PR, False after this PR
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed headless recording loops so they reliably stop: the viewer file now reports a stopped running state immediately after it is closed.

* **Tests**
  * Added a test to verify `is_running()` returns `False` immediately following `close()`.
  * Updated the recorder test flow to close the viewer file and assert the stopped state before validating outputs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 
 ### Fixed
 
+- Fix `ViewerFile.is_running()` to return `False` after `ViewerFile.close()` so headless recording loops can terminate like interactive viewers. (#3094)
 - Fix `SolverVBD` rigid contact injecting kinetic energy for yawed finite-radius contacts (e.g. small-radius cables blowing up). The normal response now acts at the geometric skeleton point rather than the rotating surface anchor, which was non-conservative under reorientation; friction still uses the surface anchor to preserve finite-radius slip. (#3125)
 - Fix `SolverKamino` contact filtering and constraint stabilization so gap/margin contacts are handled consistently, positive-distance contacts can be filtered as configured, and converted contact forces/wrenches populate matching Newton contact slots for `SensorContact`. (#2908)
 - Fix `newton.eval_jacobian`, `SolverFeatherstone`, and the IK analytic Jacobian building `JointType.D6` angular motion-subspace columns from raw axes, so `J @ joint_qd` now matches `State.body_qd` for two- or three-angular-DOF joints at non-identity configurations.
--- a/newton/_src/viewer/viewer_file.py
+++ b/newton/_src/viewer/viewer_file.py
@@ -1120,6 +1120,7 @@ class ViewerFile(ViewerBase):
 
         self._frame_count = 0
         self._model_recorded = False
+        self._running = True
 
     @override
     def set_model(self, model: Model | None):
@@ -1150,6 +1151,15 @@ class ViewerFile(ViewerBase):
         # Auto-save if enabled
         if self.auto_save and self._frame_count % self.save_interval == 0:
             self._save_recording()
+
+    @override
+    def is_running(self) -> bool:
+        """Report whether the file viewer should continue recording.
+
+        Returns:
+            bool: False after :meth:`close` has been called.
+        """
+        return self._running
 
     def save_recording(self, file_path: str | None = None, verbose: bool = False):
         """Save the recorded data to file.
@@ -1426,6 +1436,7 @@ class ViewerFile(ViewerBase):
         """Save final recording and cleanup."""
         if self._frame_count > 0:
             self._save_recording()
+        self._running = False
         print(f"ViewerFile closed. Total frames recorded: {self._frame_count}")
 
     def load_recording(self, file_path: str | None = None, verbose: bool = False):
--- a/newton/tests/test_recorder.py
+++ b/newton/tests/test_recorder.py
@@ -22,7 +22,15 @@ from newton.viewer import ViewerFile
 
 
 class TestRecorder(unittest.TestCase):
-    pass
+    def test_viewer_file_is_running_reflects_close(self):
+        """ViewerFile loop lifecycle matches interactive viewers."""
+        with tempfile.NamedTemporaryFile(suffix=".json", delete=True) as tmp:
+            viewer_file = ViewerFile(tmp.name, auto_save=False)
+
+            self.assertTrue(viewer_file.is_running())
+
+            viewer_file.close()
+            self.assertFalse(viewer_file.is_running())
 
 
 def test_ringbuffer_basic(test: TestRecorder, device):
@@ -225,8 +233,9 @@ def test_viewer_file_playback(test: TestRecorder, device):
         viewer_file_record.set_model(model)
         for state in states:
             viewer_file_record.log_state(state)
-        viewer_file_record.save_recording()
+
         viewer_file_record.close()
+        test.assertFalse(viewer_file_record.is_running())
 
         # Playback via ViewerFile
         viewer_file_play = ViewerFile(file_path)


adenzler-nvidia:
Nice catch on the fix. Minor, non-blocking: `test_viewer_file_is_running_reflects_close` doesn't use `device` (it just flips a Python flag), so registering it via `add_function_test(..., devices=devices)` runs it once per device for no benefit. You're following the existing convention here — `test_ringbuffer_basic`, `test_ringbuffer_edge_cases`, and a couple of the recorder tests do the same — so this isn't on you.

The established pattern for device-independent viewer logic is a plain `unittest.TestCase` method (see most of `test_viewer_*.py`, e.g. `test_viewer_image_logger_class.py`), which also reads more honestly than a CPU-only device list. For this PR, simplest is to make just the new test a plain method on `TestRecorder`:

```python
def test_viewer_file_is_running_reflects_close(self):
    """ViewerFile loop lifecycle matches interactive viewers."""
    with tempfile.NamedTemporaryFile(suffix=".json", delete=True) as tmp:
        viewer_file = ViewerFile(tmp.name, auto_save=False)
        self.assertTrue(viewer_file.is_running())
        viewer_file.close()
        self.assertFalse(viewer_file.is_running())
```

Converting the other device-independent recorder tests the same way would be a tidy follow-up, but no need to fold that into this bugfix.

Langford-H:
Thanks for the review! That makes sense. I updated the test to be a plain TestRecorder method and removed the device-based registration. The focused recorder tests still pass.

adenzler-nvidia (approved):
Looking good, thank you!

wfan19:
Thanks for fixing this! I'm glad my bug report was helpful.

PR #3194, opened 2026-06-16
Unify mesh adjacency on a single refactored MeshAdjacency

AnkaChan:
## Description

Newton had two parallel soft-mesh adjacency representations: the dict-based
`newton.utils.MeshAdjacency` (used by the builder and Style3D) and the VBD
solver's own `ParticleForceElementAdjacencyInfo`. This PR unifies both onto a
single refactored `MeshAdjacency` host class that holds edge topology
(`edge_indices`, `edge_tri_indices`, `tri_edge_indices`) and CSR vertex
adjacency (`v_adj_tris` / `v_adj_edges` / `v_adj_springs` / `v_adj_tets`),
deployed to kernels as a compact `MeshAdjacencyDeviceData` Warp struct via `to(device)`.

- `ModelBuilder.finalize()` builds it into the new `Model.soft_mesh_adjacency`.
- The VBD solver builds the CSR fields on demand via `init_vertex_adjacency`
  (idempotent) and reads them through the per-vertex accessor functions in
  `mesh.py`; `ParticleForceElementAdjacencyInfo` is removed.
- Style3D's bending-data path constructs a `MeshAdjacency` from triangle indices.
- `newton.utils.MeshAdjacency` is the unified, supported public class —
  construction and the `edge_indices` / `edge_tri_indices` / `tri_edge_indices`
  arrays emit no warning; only its legacy `.edges` dict accessor is deprecated
  in favor of those arrays.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_collision_cloth
uv run --extra dev -m newton.tests -k test_softbody
```

All pass (full suite green, including under `--strict-warnings`). `pre-commit run -a` is clean.

## New feature / API change

```python
import newton

builder.add_cloth_mesh(...)
model = builder.finalize()
adjacency = model.soft_mesh_adjacency  # MeshAdjacency: edge topology + CSR vertex adjacency

# MeshAdjacency is also a supported public class (no warning on import/use):
from newton.utils import MeshAdjacency
adj = MeshAdjacency(tri_indices)        # edge_indices / edge_tri_indices / tri_edge_indices
adj.edges                               # the only deprecated bit: emits DeprecationWarning; use the arrays
```

## Summary by CodeRabbit

* **New Features**
  * Model instances now expose `soft_mesh_adjacency` after `finalize()`, providing soft-mesh topology/adjacency data for simulations.

* **Deprecations**
  * The `MeshAdjacency.edges` dict accessor is deprecated; use the `edge_indices` / `edge_tri_indices` arrays instead.

* **Improvements**
  * Cloth bending-edge handling and VBD particle adjacency now rely on the unified soft-mesh adjacency data.
  * Topological collision filtering can be derived from mesh adjacency, with support for externally provided adjacency data.

* **Tests**
  * Updated collision and model adjacency tests to match the new filtering/adjoining sources.


newton/_src/utils/mesh.py
```diff
@@ -179,90 +178,565 @@ def smooth_vertex_normals_by_position(
...
+    v_adj_hinges_offsets: wp.array[wp.int32]
+    v_adj_springs: wp.array[wp.int32]
+    v_adj_springs_offsets: wp.array[wp.int32]
+    v_adj_tets: wp.array[wp.int32]
+    v_adj_tets_offsets: wp.array[wp.int32]
+
+    def to(self, device):
+        if device == self.edge_indices.device:
+            return self
+
```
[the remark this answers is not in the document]
AnkaChan: Fixed in f3efa6b8. `MeshAdjacency.to()` now normalizes the argument with `wp.get_device(device)` before comparing it against `edge_indices.device`, so string devices such as `"cuda:0"` match correctly. This also aligns with the device-normalization convention already used elsewhere in `mesh.py` (e.g. lines 92 and 131).

Verified on an L40: both string and `wp.Device` inputs short-circuit when the device already matches, and CPU↔CUDA transfer is unaffected.

newton/_src/sim/builder.py
```diff
@@ -7754,6 +7776,87 @@ def init_if_none(arr, defaultValue):
...
+        self.edge_tri_indices = np.concatenate((self.edge_tri_indices, missing_rows), axis=0)
+
+    @staticmethod
+    def _expand_edge_parameter(values: float | Sequence[float] | np.ndarray | None, count: int):
+        """Broadcast scalar edge parameters while preserving sequences and None."""
+        if values is None:
+            return None
+        if np.isscalar(values):
+            return [float(values)] * count
+        return values
```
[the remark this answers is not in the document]
AnkaChan: Done in 3ecc2a51. `_expand_edge_parameter` now normalizes via `np.asarray`: a 0-D array broadcasts like a scalar (no longer "not iterable"), and a sequence whose length doesn't match the generated edge count raises `ValueError` instead of silently desyncing `edge_bending_properties` from `edge_indices` through the `strict=False` zip in `add_edges()`.

Added a regression test (`test_expand_edge_parameter`) covering the scalar / 0-D / matched-sequence / None / length-mismatch cases. Note: in practice this path is only reached by Style3D, which already passes length-matched arrays (it and the builder both derive edges from the same `MeshAdjacency.compute_edge_adjacency`), so this is defensive hardening. Thanks for the catch!

WenchaoHuang (dismissed):
LGTM!

newton/_src/utils/mesh.py
```diff
@@ -180,90 +180,682 @@ def smooth_vertex_normals_by_position(
...
         o1: int
-        """Index of the vertex opposite to the edge in the second adjacent triangle, or -1 if boundary."""
         f0: int
-        """Index of the first adjacent triangle."""
         f1: int
-        """Index of the second adjacent triangle, or -1 if boundary edge."""
 
-    def __init__(self, indices: Sequence[Sequence[int]] | np.ndarray):
-        """Build edge adjacency from triangle indices.
+    def __init__(
```
eric-heiden, line 300: 🔴 `MeshAdjacency` is public through `newton.utils`, and the old constructor accepted `indices` as the triangle input. Renaming that keyword to `tri_indices` makes existing calls like `MeshAdjacency(indices=my_tris)` fail with `TypeError`, so this needs a deprecation path before the keyword is removed.

Please keep `indices` as a deprecated alias for `tri_indices` for at least one release, and raise clearly if both names are passed with conflicting values.

newton/_src/utils/mesh.py
```diff
@@ -180,90 +180,682 @@ def smooth_vertex_normals_by_position(
...
+        particle_count: int,
+        *,
+        edge_indices: "wp.array[wp.int32] | np.ndarray | None" = None,
+        tri_indices: "wp.array[wp.int32] | np.ndarray | None" = None,
+        spring_indices: "wp.array[wp.int32] | np.ndarray | None" = None,
+        tet_indices: "wp.array[wp.int32] | np.ndarray | None" = None,
+    ) -> "MeshAdjacency":
+        """Compute and store the vertex-to-element CSR tables (edges/tris/springs/tets) as NumPy.
+
+        Bending-edge adjacency uses ``edge_indices`` (defaulting to :attr:`edge_indices`);
```
eric-heiden: 🔴 `init_vertex_adjacency()` drops triangle topology by default even when the object was constructed with triangles. A public call path like `adj = MeshAdjacency(my_tris); adj.init_vertex_adjacency(particle_count=N); adj.to(device)` will build empty `v_adj_tris` and `v_adj_tris_offsets`, so later kernels that read `v_adj_tris_offsets[vertex + 1]` can see wrong adjacency or hit an out-of-bounds access.

Please default `tri_indices` from the stored constructor topology, for example `self.indices`, when the caller does not pass an explicit value. It would also be safer to reject a second initialization with different topology instead of silently returning stale CSR tables.

newton/_src/utils/mesh.py
```diff
@@ -180,90 +180,682 @@ def smooth_vertex_normals_by_position(
 default_num_segments = 32
 
 
+@wp.struct
+class MeshAdjacencyDeviceData:
```
eric-heiden: This should be importable somehow if you decide to keep `MeshAdjacency.to()` as a method so we can clearly document/expose the return type and make sure user code can use this struct in custom kernels.

newton/_src/utils/mesh.py
```diff
@@ -180,90 +180,682 @@ def smooth_vertex_normals_by_position(
...
+        tris = np.asarray(indices, dtype=np.int32).reshape(-1, 3)
+        tri_count = tris.shape[0]
+        if tri_count == 0:
+            return (
+                np.empty((0, 4), dtype=np.int32),
+                np.empty((0, 2), dtype=np.int32),
+                np.empty((0, 3), dtype=np.int32),
+            )
 
-    def add_edge(self, i0: int, i1: int, o: int, f: int):
```
eric-heiden, line 236: We should retain this method and handle deprecation behavior here rather than removing it outright.

newton/_src/utils/mesh.py
eric-heiden: Maybe naming it `MeshAdjacencyData` is better, to keep it shorter?

newton/_src/utils/mesh.py
AnkaChan: Good catch — kept `indices` as a deprecated alias for `tri_indices`: it emits a `DeprecationWarning` and forwards through, and raises `ValueError` if both are passed with conflicting values.

newton/_src/utils/mesh.py
AnkaChan: Addressed at the root: the element indices (edges/tris/springs/tets) are now stored as members on `MeshAdjacency` at construction, so `init_vertex_adjacency(particle_count)` takes no topology arguments — it always builds from the constructed topology. That removes the empty-`v_adj_tris` path, and since there's no topology to re-pass, the stale-CSR-on-conflicting-re-init case can't occur; the `vertex_adjacency_initialized` flag now just memoizes the one-time build.

newton/_src/utils/mesh.py
AnkaChan: Done — it's now public as `newton.utils.MeshAdjacencyData` (renamed per your other comment), so `MeshAdjacency.to()`'s return type is importable, documented, and usable in custom kernels. Regenerated the API docs.

newton/_src/utils/mesh.py
AnkaChan: Restored `add_edge` as a deprecated shim — emits a `DeprecationWarning` (steering to `edge_indices`) and keeps the old incremental semantics (first call fills `o0`/`f0`, second `o1`/`f1`, third warns non-manifold), updating `edge_indices`/`edge_tri_indices` and invalidating the vertex-adjacency CSR. One caveat — now in the docstring and the warning: it does **not** update `tri_edge_indices` (the `(i0,i1,o,f)` args don't carry the per-triangle local-edge slot), so an edge added this way won't appear in the per-triangle map; users can reconstruct via the constructor if they need it. Cold-path O(edge_count) shim.

newton/_src/utils/mesh.py
AnkaChan: Good call — renamed `MeshAdjacencyDeviceData` → `MeshAdjacencyData` everywhere (struct, `to()` return type, the `get_vertex_*` accessors, and the VBD kernels).

newton/_src/utils/mesh.py
```diff
@@ -180,90 +180,735 @@ def smooth_vertex_normals_by_position(
...
-                return
+        Uploads the edge/triangle topology maps and the vertex-adjacency CSR; this is the
+        only place the host NumPy tables become Warp arrays. :meth:`init_vertex_adjacency`
+        must have populated the CSR tables first.
+        """
+        device = wp.get_device(device)
+        data = MeshAdjacencyData()
+        data.edge_tri_indices = wp.array(self.edge_tri_indices, dtype=wp.int32, device=device)
+        data.tri_edge_indices = wp.array(self.tri_edge_indices, dtype=wp.int32, device=device)
+        data.v_adj_tris = wp.array(self.v_adj_tris, dtype=wp.int32, device=device)
```
eric-heiden, line 454: 🔴 `MeshAdjacency.to()` still accepts an object whose vertex CSR was never initialized, then passes `None` into the `wp.array` uploads. This is a public API path (`MeshAdjacency(tris).to(device)`) and the docstring says `init_vertex_adjacency()` is required, but the method does not enforce that precondition, so callers get a low-level upload failure or an unusable `MeshAdjacencyData` instead of a clear error.

Please guard `vertex_adjacency_initialized` here and raise a `ValueError` that tells callers to run `init_vertex_adjacency(particle_count)` first.

newton/_src/sim/builder.py
```diff
@@ -11180,6 +11204,22 @@ def _to_wp_array(data, dtype, requires_grad):
...
+            # derive the edge/triangle maps against the final triangles. Vertex adjacency stays
+            # unset until the solver builds it via init_vertex_adjacency; kernels get a device copy
+            # from MeshAdjacency.to.
+            edge_indices = (
+                np.array(self.edge_indices, dtype=np.int32).reshape(-1, 4)
+                if self.edge_indices
+                else np.empty((0, 4), dtype=np.int32)
+            )
+            m.soft_mesh_adjacency = MeshAdjacency(
+                tri_indices=self.tri_indices,
```
eric-heiden, line 11260: 🔴 This constructs the finalized model adjacency with `self.tri_indices`, `self.spring_indices`, and `self.tet_indices` still pointing at the mutable builder lists. `m.tri_indices` and the other model arrays are already frozen Warp copies, but `m.soft_mesh_adjacency` can later build or reuse vertex CSR from builder state if the builder is modified after `finalize()`, which makes the finalized model topology inconsistent.

Please copy these inputs into owned NumPy arrays before storing them in `MeshAdjacency`, either here or inside the constructor, so the finalized model adjacency is detached from the builder.

newton/_src/solvers/vbd/tri_mesh_collision.py
```diff
@@ -128,9 +273,28 @@ def __init__(
         self.edge_filtering_list_offsets = edge_filtering_list_offsets
 
         self.edge_edge_parallel_epsilon = edge_edge_parallel_epsilon
+        # The soft-mesh adjacency comes from the model; ensure its vertex-adjacency CSR is built.
+        # init_vertex_adjacency is idempotent (vertex_adjacency_initialized flag), so this is a no-op
+        # once the solver has built it.
+        self.mesh_adjacency = model.soft_mesh_adjacency.init_vertex_adjacency(model.particle_count)
```
eric-heiden, line 281: 🔴 This dereferences `model.soft_mesh_adjacency` without the guard used by `SolverVBD._compute_particle_force_element_adjacency()`. A caller that constructs `TriMeshCollisionDetector` directly, or passes a manually built `Model`, gets an `AttributeError` when `soft_mesh_adjacency` is `None` instead of a clear validation error.

Please add the same `None` check here, or make the detector build a temporary `MeshAdjacency` from `model.tri_indices` and `model.edge_indices` when the model does not already carry one.

newton/_src/solvers/vbd/tri_mesh_collision.py
```diff
@@ -128,9 +273,28 @@ def __init__(
...
 
         self.collision_detection_block_size = collision_detection_block_size
 
+        if (
+            vertex_triangle_filtering_list is None
```
eric-heiden: 🔴 The filter build is skipped whenever either explicit filter list is supplied, even if the other side still needs generated filters. For example, passing a custom vertex-triangle list plus `external_edge_edge_filtering_map` leaves `edge_filtering_list` unset because this `if` requires both existing lists to be `None`.

Please build or merge each filter family independently: keep an explicit list for the side the caller provided, but still generate the missing vertex-triangle or edge-edge filters when threshold or external inputs request them.

newton/_src/utils/mesh.py
AnkaChan: Good point — fixed the `None`-into-`wp.array` failure. Rather than hard-raise, I went a touch more permissive: `to()` always uploads `edge_tri_indices` and `tri_edge_indices` (they're computed at construction), and when `vertex_adjacency_initialized` is `False` it emits a `UserWarning` and leaves the eight `v_adj_*` fields as `None` (no `wp.array(None)`). Rationale: a caller can legitimately want only `edge_tri_indices` (each edge's two adjacent triangles) and `tri_edge_indices` (each triangle's three edges) without building the vertex-adjacency CSR — so returning those plus a clear warning felt friendlier than refusing outright. Happy to make it a hard `ValueError` if you'd prefer strictness.

newton/_src/sim/builder.py
AnkaChan: Fixed — the constructor now stores owned `int32` NumPy copies of the tri/spring/tet indices (not aliases of the builder's mutable lists), so the finalized `model.soft_mesh_adjacency` is detached from builder state and can't drift if the builder is modified after `finalize()`.

newton/_src/solvers/vbd/tri_mesh_collision.py
AnkaChan: Fixed — `TriMeshCollisionDetector` now raises the same `ValueError` the VBD solver uses when `model.soft_mesh_adjacency` is `None`, so a manually-built `Model` gets a clear validation error instead of an `AttributeError`.

newton/_src/solvers/vbd/tri_mesh_collision.py
AnkaChan: Fixed — vertex-triangle and edge-edge filters are now built independently by separate per-family methods, `_build_vertex_triangle_filter` and `_build_edge_edge_filter`, sharing a `_extract_filter_adjacency` helper (so the vertex-adjacency extraction still runs once). Each side builds when its own list wasn't supplied and a source requests it — VT when `vertex_triangle_filtering_list is None and (threshold >= 2 or external_vt_map)`, EE likewise — so an explicit list on one side no longer suppresses generated filters on the other.

newton/_src/utils/mesh.py
```diff
@@ -180,90 +180,746 @@ def smooth_vertex_normals_by_position(
...
+
+    return wp.array(_numpy_int_rows(data, width), dtype=wp.int32, device="cpu")
+
+
+def _as_cpu_int_array1d(data) -> wp.array:
+    """Return topology data as a flat CPU Warp int array."""
+    if isinstance(data, wp.array):
+        if data.ndim == 1:
+            return data.to("cpu")
+        return wp.array(data.numpy().reshape(-1), dtype=wp.int32, device="cpu")
```
gdaviet, line 888: Warp arrays have `flatten()`/`reshape()` functions, can we use that instead of delegating to numpy then rebuilding the warp array?

newton/_src/utils/mesh.py
```diff
@@ -180,90 +180,746 @@ def smooth_vertex_normals_by_position(
...
+    ``MeshAdjacency`` stays free of Warp arrays (``to`` re-uploads on demand).
+    """
+    with wp.ScopedDevice("cpu"):
+        counts = wp.zeros(shape=(particle_count,), dtype=wp.int32, device="cpu")
+        wp.launch(count_kernel, inputs=[topology, counts], dim=1, device="cpu")
+
+        counts_np = counts.numpy()
+        offsets_np = np.empty(shape=(particle_count + 1,), dtype=np.int32)
+        offsets_np[0] = 0
+        offsets_np[1:] = np.cumsum(values_per_entry * counts_np)[:]
```
gdaviet, line 918: `warp.utils` has `array_scan` if we want to get rid of numpy here. but since the result is returned on host that probably does not matter

gdaviet (commented):
Looks good to me !

Device-side adjacency query look very helpful.
Regarding build there's a lot that could be converted to warp/device kernels if required, but since that's just done once that would be low priority

newton/_src/utils/mesh.py
AnkaChan: Good call — switched the `wp.array` path in `_as_cpu_int_array1d`/`_as_cpu_int_array2d` to Warp's `flatten()`/`reshape()` and dropped the numpy round-trip. Kept a numpy fallback for the non-contiguous case, since Warp's `flatten()` requires contiguous input.

newton/_src/utils/mesh.py
AnkaChan: Kept `np.cumsum` here — as you noted the offsets come back on host, and they're stored as NumPy on `MeshAdjacency` (which is meant to stay directly Python-readable), so `array_scan` wouldn't buy anything. Happy to switch if you'd rather it be Warp-native throughout.

AnkaChan:
Thanks Gilles! Two reasons we keep the build host-side: (1) `MeshAdjacency` is meant to be read directly from Python — per-element indexing over its arrays, which `wp.array` doesn't support — so the host copies stay NumPy; the `@wp.struct` `MeshAdjacencyData` is the dedicated device-side view kernels get via `to()`. (2) The topological computation is already largely vectorized in NumPy (the CSR fill uses Warp count/fill kernels on CPU), so switching the rest to device kernels wouldn't gain much, and it runs once at `finalize()`. Happy to revisit if profiling ever says otherwise.

eric-heiden (approved):
Thanks, LGTM!

PR #3196, opened 2026-06-17
Align CoACD threshold default

### LINKED ISSUE #3195: [BUG] CoACD convex decomposition uses overly coarse default threshold and is hard to configure
### Bug Description

Newton maps imported USD/PhysX convexDecomposition collision meshes to CoACD, but the CoACD settings are currently hardcoded in ModelBuilder.approximate_meshes(). In particular, Newton uses:

threshold = 0.5
while CoACD’s default threshold is much smaller, around 0.05. This can produce an extremely coarse convex decomposition, sometimes collapsing the collision approximation to too few hulls. For some assets, the resulting collision geometry no longer matches the intended shape well enough and the simulation fails or behaves incorrectly.

This is difficult to work around when loading assets because the CoACD parameters are not exposed through the USD/import path. Users currently need to modify Newton source code to change the threshold or try a different convex decomposition method.

### Reproduction Script

```python

```

### System Information

_No response_

jumyungc:
Use CoACD's upstream threshold for default convex decomposition so imported meshes are less likely to produce overly coarse hulls. Document how to preserve the previous threshold.

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the default CoACD convex decomposition threshold from 0.5 to 0.05 to align with CoACD's standard behavior. Previous behavior can be preserved through explicit configuration.

* **Documentation**
  * Updated changelog documenting the threshold change.


eric-heiden (approved):
Thanks!

PR #3199, opened 2026-06-17
Fixes `RenderContext` triangle mesh construction by removing the unsupported `device=` keyword from `wp.Mesh(...)`

huidongc:
## Description

Fixes `RenderContext` triangle mesh construction by removing the unsupported `device=` keyword from `wp.Mesh(...)`.

`wp.Mesh` does not accept a `device` argument; it uses the devices of the provided points and indices arrays. Since `triangle_points` and `triangle_indices` are already allocated on the intended device, passing `device=self.device` is unnecessary and causes mesh initialization to fail.

## Checklist

## Test plan

```sh
uv run --extra dev -m newton.tests -k test_cloth_renders_via_triangle_mesh_construction
```

Verified the test fails with the pre-fix `wp.Mesh(..., device=self.device)` call and passes with the fix applied.

## Bug fix

**Minimal reproduction:**

```python
import math
import warp as wp
import newton
builder.add_cloth_grid(
    pos=wp.vec3(0.0, 0.0, 0.0),
    rot=wp.quat_identity(),
    vel=wp.vec3(0.0),
    dim_x=2,
    dim_y=2,
    cell_x=0.1,
    cell_y=0.1,
    mass=0.1,
    fix_top=True,
)
sensor = SensorTiledCamera(model=model)
width, height = 8, 8
camera_transforms = wp.array(
    [[wp.transformf(wp.vec3f(0.0, 0.0, 0.5), wp.quatf(0.0, 0.0, 0.0, 1.0))]],
    dtype=wp.transformf,
    device="cpu",
)
camera_rays = sensor.utils.compute_pinhole_camera_rays(width, height, math.radians(60.0))
depth_image = sensor.utils.create_depth_image_output(width, height)
sensor.update(model.state(), camera_transforms, camera_rays, depth_image=depth_image)
```

The following Python type error would pop up:
```log
if has_shapes or has_particles or self.has_triangle_mesh or self.has_gaussians:
            if self.has_triangle_mesh:
                if self.triangle_mesh is None:
>                   self.triangle_mesh = wp.Mesh(self.triangle_points, self.triangle_indices, device=self.device)
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E                   TypeError: Mesh.__init__() got an unexpected keyword argument 'device'

env_isaaclab/lib/python3.12/site-packages/newton/_src/sensors/warp_raytrace/render_context.py:222: TypeError
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved triangle-mesh rendering setup to use supported mesh construction behavior, improving consistency of where mesh data lives during rendering.
  * Fixed state assignment to properly retain namespaced extended/custom attributes.
* **Tests**
  * Added coverage to confirm cloth rendering uses the triangle-mesh path and produces expected depth hits.
* **Documentation**
  * Updated release notes with a deprecation notice for `SensorContact.shape`, recommending `total_force.shape` / `force_matrix.shape` instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - Fix `SolverKamino` contact filtering and constraint stabilization so gap/margin contacts are handled consistently, positive-distance contacts can be filtered as configured, and converted contact forces/wrenches populate matching Newton contact slots for `SensorContact`. (#2908)
 - Fix mesh inertia computation to produce deterministic results across repeated CUDA runs. (#3136)
 - Fix VBD collision damping to use relative normal gap rate so uniform contact-stencil motion and tangential sliding do not create artificial normal damping.
+- Fix `RenderContext` triangle mesh construction by removing the unsupported `device=` keyword from `wp.Mesh(...)`.
 - Fix MJCF `euler` producing wrong orientations for multi-component angles by treating angles as intrinsic rotations. (#3030)
 - Fix MJCF parsing so attributes from multiple `<compiler>` elements, including `<include>`-expanded children, are merged in document order. (#3030)
 - Fix MJCF worldbody static geoms bypassing the visual/collider class filter, so `parse_visuals=False` drops visual-class geoms attached directly to `<worldbody>` too. (#3030)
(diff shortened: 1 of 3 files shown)


adenzler-nvidia (approved):
Thanks for the focused fix. The new cloth/triangle-mesh camera test is a clear red/green regression, and the change matches Warp's Mesh API by relying on the input arrays' devices.

PR #3200, opened 2026-06-17
Generalize Rod joints to q7/qd6 with separate rest poses

### LINKED ISSUE #3183: Support separate cable rest and initial poses
Newton cable setup currently makes it awkward to build a cable whose model/rest pose differs from its simulation initial pose. This is useful for workflows where the authored rest shape and the desired initial world-space placement are intentionally different, such as starting a straight-rest cable from a curved pose or starting a curved-rest cable from a straight pose.

jumyungc:
## Description

Generalize experimental Rod joints from four SolverVBD-specific material
coordinates to an ordinary q7 relative pose and qd6 relative twist:

- `joint_q`: translation XYZ followed by quaternion XYZW
- `joint_qd`: relative linear XYZ followed by angular XYZ

This gives Rod joints the same kinematic representation as free and distance
joints, including FK, IK, and six-column motion-subspace support.

### Structural rest

Use the existing Model-owned `joint_target_q` storage for Rod structural rest
instead of adding Rod-specific Model arrays.

At structural rest, the parent and child anchor points coincide:

- the three linear target entries are zero;
- the rotation stores structural-rest bend and twist;
- coordinate layout stores a quaternion;
- legacy layout stores extrinsic ZYX angles.

SolverVBD derives and caches its Rod bend/twist rest values from
`Model.joint_target_q` when constructed. Control targets do not override this
structural rest.

### Builder support

Add independent initial and structural-rest inputs:

- `add_joint_rod(rest_rotation=...)`
- `add_rod(rest_positions=..., rest_quaternions=...)`
- `add_rod_graph(rest_node_positions=..., rest_quaternions=...)`

Rest edge lengths determine capsule geometry, mass properties, and anchor
offsets. Initial positions and frames determine the initial body poses. This
allows a rod to begin stretched, bent, or twisted relative to its structural
rest pose.

`add_rod(rest_straight=True)` is a convenience for zero intrinsic bend and
twist. It keeps the authored initial pose, segment lengths, and anchor geometry
unchanged while setting each generated Rod rest rotation to identity

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

- **New Features**
  - Experimental rod and cable joints now support six degrees of freedom with improved forward/inverse kinematics.
  - Added separate initial and structural-rest poses, including rest rotations and straight-rest configuration for rods and cables.
  - VBD simulations now support rod wrenches, material behavior, and control forces more consistently.
  - Added an example demonstrating cables with different initial and structural-rest poses.

- **Documentation**
  - Updated rod-joint conventions, migration guidance, limitations, and usage examples.

- **Bug Fixes**
  - Improved rod pose, velocity, stiffness, anchor, and closed-loop handling.


newton/_src/sim/ik/ik_lbfgs_optimizer.py
```diff
@@ -141,6 +145,9 @@ def __new__(
         *a: Any,
         **kw: Any,
     ) -> IKOptimizerLBFGS:
+        if model._has_rod_joints and JointType.ROD in model.joint_type.numpy():  # pyright: ignore[reportPrivateUsage]
+            raise ValueError("IKOptimizerLBFGS does not support JointType.ROD joints.")
+
```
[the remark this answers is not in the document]
jumyungc: _has_rod_joints is maintained for supported construction and ModelView.disable_joints() updates it. Direct joint_type mutation is an unsupported structural change, and removing the cache from only these three guards would remain inconsistent with inverse dynamics. Existing tests cover partial and fully disabled Rod views.

newton/examples/cable/example_cable_bundle_hysteresis.py
```diff
@@ -437,8 +439,8 @@ def test_final(self):
 
         if self.with_dahl and (np.min(straightness) < 0.5 or np.max(straightness) > 0.9):
             raise ValueError(f"Dahl cable bundle did not retain plausible curvature: {metrics}")
-        if not self.with_dahl and np.min(straightness) < 0.9:
-            raise ValueError(f"Elastic cable bundle did not recover: {metrics}")
+        if not self.with_dahl and np.min(straightness) < 0.75:
+            raise ValueError(f"Elastic cable bundle did not recover plausibly: {metrics}")
```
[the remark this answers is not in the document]
jumyungc: straightness is already the residual-curvature metric, and np.min enforces it independently for every cable. The no-Dahl case still includes contact friction, so requiring every cable to reach 0.9 is overly restrictive. The 0.75 threshold is intentionally a collapse-regression gate.

eric-heiden:
Thanks for this, the q7/qd6 generalization and the separate rest pose make the rod joint conventions much easier to follow, and the docs updates help a lot.

A few things are worth a look before merge: a stray decorator on `add_joint_free` that undoes an earlier keyword-only change already on `main`, the loss of the runtime refresh path for the rod rest pose, a measurable slowdown in `add_rod()` authoring, and two small documentation gaps.

newton/_src/sim/builder.py
```diff
@@ -5575,6 +5649,21 @@ def add_joint_fixed(
...
+            child: Child body index.
+        """
+        q_start = self.joint_q_start[joint_id]
+        parent_body_xform = wp.transform_identity() if parent == -1 else self.body_q[parent]
+        parent_anchor_world = parent_body_xform * self.joint_X_p[joint_id]
+        joint_q = wp.transform_inverse(parent_anchor_world) * self.body_q[child] * self.joint_X_c[joint_id]
+        self.joint_q[q_start : q_start + 7] = list(joint_q)
+
+    @deprecate_nonkeyword_arguments
     def add_joint_free(
```
eric-heiden, line 5667: 🟠 `@deprecate_nonkeyword_arguments` is back on `add_joint_free`, but `main` removed it in #4070 ("Enforce mature keyword-only API signatures"). This branch already contains that removal, so this line re-adds the decorator on top of it. It looks like it came along with one of the `Merge branch 'main'` commits rather than from the rod work.

The effect is a public API regression. The wrapper in `newton/_src/utils/deprecation.py` rebinds extra positional arguments onto the keyword-only names and emits only a `DeprecationWarning`, so `builder.add_joint_free(child, xform_p, xform_c)` keeps working on this branch while it raises `TypeError` on `main`. `CODING_GUIDELINES.rst:210-212` is the rule #4070 was implementing, and this line effectively restarts that deprecation process.

The signature itself is unchanged and nothing in this PR calls it positionally, so removing the line should be safe. The intended change in this hunk, extracting `_init_joint_q_from_body_poses` so `add_joint_rod` can reuse it, is unaffected.

```suggestion
    def add_joint_free(
```

For reference, the `add_joint_cable` decorator further down is legitimate, it is already on `main` as a deprecated alias. Only this one is new.

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -1194,8 +1199,6 @@ def notify_model_changed(self, flags: ModelFlags | int) -> None:
             self._refresh_joint_material_params()
         if refresh_structural_k:
             self._refresh_structural_k()
-        if flags & (ModelFlags.JOINT_PROPERTIES | ModelFlags.BODY_PROPERTIES):
-            self._refresh_rod_rest_bend_twist_cache()
 
     @override
     def coupling_supports_inertial_property_refresh(self) -> bool:
```
eric-heiden, line 1201: 🔵 Dropping this refresh makes the rod rest pose immutable once the solver is constructed.

The gap is wider than the rename, because the rest source also moved from `model.body_q` to `model.joint_target_q`, and no `ModelFlags` value covers `joint_target_q`: `JOINT_PROPERTIES` covers `joint_q` / `X_p` / `X_c`, and `JOINT_DOF_PROPERTIES` covers the DOF material params. So the previously working workflow of mutating `model.body_q` and calling `notify_model_changed(ModelFlags.BODY_PROPERTIES)` now silently keeps a stale rest, with no flag that could fix it. Rod rest ends up being the one rod property that cannot be retuned live, while every rod material property still can.

The `SolverVBD` class docstring does state the new contract of constructing a new solver after changing these rest targets, so this is documented rather than hidden.
But I think we must offer a way to refresh the rest positions via `notify_model_changed` gated on a flag that covers `joint_target_q` (please consider adding one if nothing fits).

newton/_src/sim/builder.py
```diff
@@ -717,6 +717,29 @@ def _resolve_rod_body_frame_origin(
...
+        if not math.isfinite(norm) or norm == 0.0:
+            raise ValueError(f"{method_name}: {argument_name} must be a finite, nonzero quaternion")
+        q = wp.quat(*(component / norm for component in components))
+
+        local_z_world = wp.quat_rotate(q, wp.vec3(0.0, 0.0, 1.0))
+        alignment = float(wp.dot(segment_direction, local_z_world))
+        if not math.isfinite(alignment) or alignment < 0.999:
+            raise ValueError(f"{method_name}: {argument_name} must align local +Z with the segment direction")
+
+        cross = wp.cross(local_z_world, segment_direction)
```
eric-heiden, line 739: 🔵 The per-frame alignment math added here makes rod authoring noticeably slower. Building 8 rods of 150 segments (1192 rod joints) against the current merge base:

| Measurement | This branch | `main` | Delta |
|---|---|---|---|
| `add_rod()` min, run 1 / run 2 | 1535.3 / 1474.1 ms | 1268.2 / 1195.7 ms | **+21.1% / +23.3%** |
| `add_rod()` median, run 1 / run 2 | 1644.7 / 1478.2 ms | 1309.9 / 1196.4 ms | **+25.6% / +23.6%** |
| `finalize()` min, run 1 / run 2 | 361.5 / 358.5 ms | 374.6 / 354.2 ms | -3.5% / +1.2% (noise) |

That lands around +23%, roughly 0.23 ms of one-time build cost per rod joint. The cost shows up in Warp's Python-side builtin dispatch: `call_builtin_from_desc` goes from 52,754 to 73,058 calls (+38%, about 17 extra per rod joint) and `warp/_src/types.py __getitem__` from 83,760 to 177,984 (+112%). The source is the new per-segment/per-joint scalar Warp math: `_normalize_and_align_rod_quaternion` runs a `quat_rotate`, a `dot`, a `cross`, and two `normalize` per frame, and it runs for both the initial and the rest frame, plus `_edge_rest_rotation` / `_closed_rest_quaternion` and `_init_joint_q_from_body_poses`.

Please take a look at the internal `transform_mul()` function we have in builder.py that shows how to explicitly shortcut Warp's overload resolution to offer much better performance when calling Warp-native functions from a Python context.

newton/_src/sim/builder.py
```diff
@@ -5828,29 +5917,45 @@ def add_joint_rod(
...
+                [unitless quaternion]. If None, it uses the initial relative anchor rotation at creation.
 
         Returns:
             The index of the added joint.
 
+        Raises:
+            ValueError: If ``rest_rotation`` contains non-finite values or has zero
+                norm, or if any material stiffness is negative.
+
+        .. experimental::
```
eric-heiden, line 5930: ⚪ Good to see the marker on `add_joint_rod`. Could `add_rod` and `add_rod_graph` in this same file get one as well? They are the entry points users actually reach (the README example table and every `example_cable_*.py` go through them), and they pick up the new `rest_positions` / `rest_node_positions` / `rest_quaternions` / `rest_straight` parameters whose conventions this PR is explicitly reserving the right to change.

Right now `newton/_src/sim/builder.py` carries `.. experimental::` only on lines 4041, 4061, and this one, so someone landing on `add_rod` through autodoc or `help()` gets no signal at all. `docs/concepts/articulations.rst` does name the parameters as experimental, but the docstrings are where most readers will look. Reusing the wording from this block, scoped to the rod rest-pose parameters and conventions, would cover it.

newton/__init__.py
```diff
@@ -20,8 +20,9 @@
 - ``True`` (the default): shape ``(joint_coord_count,)``, matching
   :attr:`~newton.State.joint_q`.
 - ``False``: legacy shape ``(joint_dof_count,)``, which is misaligned with
-  :attr:`~newton.State.joint_q` whenever an articulation contains a free,
-  ball, or distance joint upstream of a position-controlled DOF.
+   :attr:`~newton.State.joint_q` whenever an articulation contains a free,
+   ball, or distance joint upstream of a position-controlled DOF, or contains
+   a rod joint.
```
eric-heiden, line 25: ⚪ The continuation lines of this bullet are now indented 3 spaces, while the `True` bullet above uses 2. For a `- ` marker docutils expects the continuation at 2 spaces; at 3 it reads as unexpected indentation inside the list item and renders the remainder as a nested block quote, so the two sibling bullets stop rendering the same way.

`docs/api/newton.rst` lists these constants in a manual `list-table` with name and value only, so the docs build is not broken. It does affect `help(newton)` and would affect any future autodata directive.

```suggestion
- ``False``: legacy shape ``(joint_dof_count,)``, which is misaligned with
  :attr:`~newton.State.joint_q` whenever an articulation contains a free,
  ball, or distance joint upstream of a position-controlled DOF, or contains
  a rod joint.
```

newton/_src/sim/builder.py
```diff
@@ -5771,12 +5855,19 @@ def add_joint_rod(
         collision_filter_parent: bool | None = None,
         enabled: bool = True,
         custom_attributes: dict[str, Any] | None = None,
+        rest_rotation: Quat | None = None,
```
eric-heiden, line 5858: Can you move this up in the order of arguments, somewhere after/in-between the other rod-related arguments? Same for the other methods you are modifying in this PR.

PR #3201, opened 2026-06-17
Fix hydroelastic reduction gating units after pressure callback

### LINKED ISSUE #3202: example_nut_bolt_hydro stopped threading after #2705
## Summary

`example_nut_bolt_hydro` stopped threading after #2705. Instead of advancing down the bolt, the nut rotates only ~30° and stalls — it threaded (~80° rotation) before that commit.

## Cause

#2705 moved the contact-reduction aggregate force into physical pressure units (`sum(area * pressure_func(depth) * normal)`) but left the `EPS_LARGE` reliability gates (aggregate-direction, normal-matching, anchor) tuned for the previous depth-volume aggregate. At `kh ~ 1e10` those gates become effectively always-true, which corrupts the reduced contact normals/anchors and stalls thread engagement.

## Reproduction

```
python -m newton.examples.contacts.example_nut_bolt_hydro --viewer null --test
```

Before #2705 the nut threads down the bolt (~80° rotation); after #2705 it rotates only ~30° and fails to thread. The example's original `test_final` rotation threshold (5.7°) was too loose to catch the regression.

mzamoramora-nvidia:
## Description

Closes #3202.

Fixes the hydroelastic contact-reduction reliability gates that regressed after #2705.

#2705 moved the reduction's aggregate force into physical pressure units (`sum(area * pressure_func(depth) * normal)`) but left the `EPS_LARGE` reliability gates tuned for the old depth-volume aggregate. At `kh ~ 1e10` those gates become effectively always-true, corrupting the reduced contact normals/anchors and stalling nut-bolt thread engagement.

The fix rescales the gated magnitude by shape_b's pressure coefficient before the `EPS_LARGE` comparison (restoring the original units for the default linear law), and applies the same rescale to the moment-accumulation gate so it stays consistent with the export kernel.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k reduced_vs_unreduced
uv run --extra dev -m newton.tests -k custom_pressure_func
```

Both pass (5/5 each, including the cubic nonlinear case).

## Bug fix

On `main`, `python -m newton.examples.contacts.example_nut_bolt_hydro --viewer null` rotates the nut only ~30° and stalls instead of threading down the bolt (it reached ~80° before #2705). The example's existing `test_final` rotation threshold (5.7°) was too loose to flag this, so `--test` still passed.

This PR restores threading (the nut reaches ~80°) and tightens `test_final` to ≥60° rotation / ≥5 mm descent so the regression is caught going forward.

## Summary by CodeRabbit

* **New Features**
  * Improved hydroelastic contact reduction reliability for more stable results in offset or uneven contact scenarios.
  * Added support for custom pressure laws in hydroelastic contact handling.
  * Updated the nut-and-bolt example to use stricter, more realistic engagement checks.

* **Bug Fixes**
  * Reduced contact-direction errors by using geometry-based depth information alongside force data.
  * Improved force preservation when using very stiff contact settings.


nvtw:
AI review found a complaint we should check:
_pressure_to_depth_scale(shape_material_k_hydro[shape_b]) assumes the pressure magnitude is governed by shape_material_k_hydro. That is only exact for the default linear pressure law. This module explicitly supports custom pressure_func, and those callbacks may ignore shape_material_k_hydro or use a nonlinear scale. For custom pressure laws, the direction-reliability gate can become arbitrarily wrong.

nvtw:
And one more:
The new scaled gate is inconsistent with _compute_normal_matching_rotation(). The MR may set has_reliable_agg_direction / nbin_direction_mag true after dividing by shape_material_k_hydro, but _compute_normal_matching_rotation() still rejects the same force using the unscaled agg_force_mag > EPS_LARGE check. In the low-stiffness regime this change is meant to fix, normal matching can silently remain disabled while anchor/export logic treats the direction as reliable. The helper should either accept the scaled reliability decision or use a similarly scaled threshold internally.

nvtw (dismissed):
Looks good now. Something small still showed up but it might not need adjustments:

One small safety question: since reliability is now gated by agg_depth_volume, should anchor/normal direction use also require agg_force_mag > EPS_SMALL before normalizing/using agg_force_vec? This would protect custom pressure laws that produce tiny or degenerate aggregate force while geometric depth-volume is nonzero.

nvtw (approved):

PR #3207, opened 2026-06-22
Warn when adding duplicate joints

### LINKED ISSUE #2803: Raise warning when adding a joint to link created from `add_body`
ModelBuilder.add_body() internally calls add_link() and creates a 6-DOF FREE joint (type 4). When an explicit add_joint_revolute() is then called on the same body, XPBD receives two conflicting world-anchored constraints. The FREE joint overrides the REVOLUTE drive torque, causing zero motion regardless of target_vel or target_kd settings.
Reproduction:

Minimum reproduction (Python):
```py
  # add_body path → 0 rpm after 200 steps
  b.add_body(xform=..., mass=0.1)
  b.add_joint_revolute(parent=-1, child=0, target_vel=2*pi, target_kd=500)

  # add_link path → immediately spinning at target rpm
  lid = b.add_link(xform=..., mass=0.1, label='Sun')
  jid = b.add_joint_revolute(parent=-1, child=lid, ...)
  b.add_articulation([jid])
```
Expected behavior: add_body() + add_joint_revolute() should create a single revolute-constrained body that responds to velocity drive targets.
Observed behavior: Body remains stationary. Joint actuator torque (joint_act) is applied but canceled by the implicit FREE joint constraint.

We should raise a warning when a joint is added here to such a link that has been created with `add_body()`.

camevor:
## Description

`ModelBuilder.add_joint()` now warns when a new joint connects a pair of bodies that are already directly connected by another joint. Such parallel joints are ambiguous and may be handled differently by different solvers, and previously nothing flagged the mistake.

The most common way to hit this is calling `add_body()` — which implicitly attaches the body to the world with a FREE joint — and then adding another joint (e.g. a revolute) from the world to the same body to "anchor" it. That leaves two parallel joints between the world and the body
. The fix for that case is to build the body with `add_link()` (no implicit joint) and add the intended joint yourself.

By default, users are warned about the solver-defined semantics/applicability of parallel joints, whereas the common case of a free joint combined with any other joint has a stronger warning as it is always inconsistent.

Closes #2803.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k TestParallelJointWarning
```

## New feature / API change

```python
import newton

body = builder.add_body(mass=1.0)  # implicitly attaches `body` to the world with a FREE joint

# Adding a second joint between the world and the same body now warns:
builder.add_joint_revolute(parent=-1, child=body)
# UserWarning: Adding a REVOLUTE joint between parent -1 and child 0 (label: 'body_0'),
# but a FREE joint already connects these bodies. A FREE joint parallel to another joint
# is inconsistent. Use add_link() with the appropriate joint type instead of add_body().

# Intended pattern when anchoring a body to the world with a non-free joint:
builder2 = newton.ModelBuilder()
link = builder2.add_link(mass=1.0)   # no implicit joint
builder2.add_joint_revolute(parent=-1, child=link)  # no warning
```

## Tests modified to avoid the new warning
Many tests were triggering the warning for different reasons. These now either avoid creating parallel joints where it is not justified (e.g. using `add_link` instead of `add_body` where appropriate) or suppress the warning where this is justified.

- Corrected accidental implicit FREE joints by replacing add_body() with add_link() where fixtures explicitly create their own joints:
    - newton/examples/multiphysics/example_admm_contact_solver.py
    - newton/tests/test_builder_replicate.py
    - newton/tests/test_collision_pipeline.py
    - newton/tests/test_equality_constraints.py
    - newton/tests/test_joint_limits.py
    - newton/tests/test_model.py
    - newton/tests/test_sdf_compute.py
    - newton/tests/test_shapes_no_bounce.py

- Reworked two-node parallel or anti-parallel “loops” into genuine multi-body loop topologies:
    - newton/tests/test_coupled_solver.py
    - newton/tests/test_equality_connect_constraint_with_sim_step.py
    - newton/tests/test_mujoco_solver.py
    - newton/tests/test_selection.py

- Asserted the warning where parallel topology is intentionally under test:
    - newton/tests/test_import_usd_deformable_cable.py
    - newton/tests/test_model.py
    - newton/tests/test_selection.py

- Narrowly suppressed the expected warning where a parallel joint intentionally represents a MuJoCo equality constraint and the test targets conversion, parsing, or equality behavior:
    - newton/tests/test_import_usd.py
    - newton/tests/test_mujoco_solver.py
    - newton/tests/test_physics_verification.py

Focused validation passes with --strict-warnings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added `UserWarning`s when adding parallel joints between the same two connected bodies, highlighting ambiguous/undefined semantics and flagging inconsistent cases involving `FREE` joints.
* **Tests**
  * Added dedicated warning coverage for parallel-joint scenarios.
  * Updated multiple test setups to build articulated elements using `add_link()` and to register articulations more explicitly.
  * Suppressed the new warning where it is expected to avoid noise in unrelated assertions.
* **Documentation**
  * Updated **Unreleased** changelog entry to describe the new parallel-joint warning behavior.


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
...
 - Change `SolverKamino.reset(world_mask=...)` to accept `wp.bool` arrays instead of `wp.int32`; callers passing `wp.int32` masks must switch to `wp.bool` (e.g. `wp.array([False, True, False], dtype=wp.bool)` or `wp.ones((num_worlds,), dtype=wp.bool)`). (#2934)
 - Assign one default visual color to capsule segments generated by `ModelBuilder.add_rod()` or `add_rod_graph()`; pass `color=` to choose an explicit cable color.
+- Warn from `ModelBuilder.add_joint()` when adding a joint between a pair of bodies that are already directly connected, since parallel joints between the same bodies are ambiguous and may behave differently across solvers.
```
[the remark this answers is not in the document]
camevor: This is not a behavioral change, only a new warning that contains guidance to fix the issue.

newton/tests/test_model.py
```diff
@@ -221,6 +221,107 @@ def test_model_builder_joint_target_pos_vel_setters_warn_and_forward(self):
...
+        sub = ModelBuilder()
+        sub.add_body(mass=1.0)  # carries an implicit FREE joint from world
+
+        main = ModelBuilder()
+        main.add_link(mass=1.0)  # occupy index 0 so the merged body lands at a nonzero offset
+        merged_body = main.body_count  # index the sub-builder's body will occupy after merge
+        main.add_builder(sub)
+
+        with self.assertWarnsRegex(UserWarning, r"FREE.*inconsistent"):
+            main.add_joint_revolute(parent=-1, child=merged_body)
```
adenzler-nvidia: I think the test coverage here is a bit excessive, I think we could reduce it a bit?

newton/_src/sim/builder.py
```diff
@@ -4098,6 +4104,35 @@ def add_joint(
...
+                )
+            else:
+                warnings.warn(
+                    f"Adding a {joint_type.name} joint between parent {parent} and "
+                    f"child {child} (label: {self.body_label[child]!r}), but another joint already connects these "
+                    f"bodies. Parallel joints between the same pair of bodies have undefined semantics and may not "
+                    f"behave as expected.",
+                    UserWarning,
+                    stacklevel=2,
+                )
```
adenzler-nvidia, line 4609: this warns, but it still creates the joint. Is this the intended behavior? In many cases, this setup is ill-defined anyway.

eric-heiden (commented):
Thanks, I think this looks good overall! There are now some existing tests that emit duplicate-joint warnings outside the new warning-specific tests, perhaps you can fix them in this PR?

| Test | Warnings |
|---|---:|
| `test_model.py::TestModelJoints.test_mimic_constraint_programmatic` | 3 |
| `test_model.py::TestModelJoints.test_add_base_joint_fixed_to_parent` | 1 |
| `test_model.py::TestModelValidation.test_control_clear` | 1 |
| `test_joint_limits.py::TestJointLimits.test_robustness_of_limit_comparisons` | 1 |
| `test_selection.py::TestSelection.test_duplicate_joint_child_is_one_link` | 1 |
| `test_sdf_compute.py::TestSDFNonUniformScaleBrickPyramid.test_brick_pyramid_stability` | 6 |
| `test_collision_pipeline.py::TestMeshConvexMidphase.test_mesh_convex_midphase_queries_margin_shell` | 1 |
| `test_collision_pipeline.py::TestHeightfieldConvexMidphase.test_heightfield_convex_midphase_queries_margin_shell_at_lateral_edge` | 1 |
| `test_collision_pipeline.py::TestDeterministicPipeline.test_deterministic_pipeline_500_steps` | 64 |
| `test_collision_pipeline.py::TestDeterministicPipeline.test_deterministic_pipeline_sticky_500_steps` | 64 |
| `test_shapes_no_bounce.py::TestShapesNoBounce.test_shapes_never_exceed_initial_z` | 128 |

Most of these should probably use `add_link()` when the test intends to add its own explicit joint. The intentional duplicate-topology case in `test_selection.py` should instead assert or filter the new warning.

newton/tests/test_model.py
camevor: Agree; reduced to the genuinely useful ones.

newton/_src/sim/builder.py
camevor: Agree in general, although there may be situations in which two joints may be valid, think a swing having two distance joints between a body and world.

Since we don't currently document the semantics of parallel joints, this is left up to the solver. A complication is that e.g. MuJoCo interprets parallel joints as each introducing degrees of freedom (although this behavior is currently not supported when used via Newton), whereas many other solvers interpret parallel joints as introducing constraints.

So the approach here is to warn, but not raise unless we are certain that the setup is bogus, which is arguably the case for the free + parallel joint, where this PR only introduces a stronger warning.

newton/_src/sim/builder.py
```diff
@@ -3971,6 +3977,35 @@ def add_joint(
                 f"but current world is {self.current_world}"
             )
 
+        has_parallel_joint = False
+        parallel_free = False
+        for existing_parent, existing_joint_idx in self.joint_parents.get(child, ()):
```
adenzler-nvidia: Should this also catch duplicates added in the reverse direction? Since only `joint_parents.get(child)` is scanned, adding a joint A->B and then B->A leaves the pair doubly connected without a warning, even though the semantics are just as ambiguous.

camevor: That's a good point, fixed to warn for those similarly. This does raise the question whether the articulation validation should raise on parallel joints (parallel or anti-parallel, neither of which it does today)?

```python
import newton

def check_pair(*, reversed_second: bool) -> None:
    body_a = builder.add_link(label="A")
    body_b = builder.add_link(label="B")

    joint_0 = builder.add_joint_revolute(parent=body_a, child=body_b)
    if reversed_second:
        joint_1 = builder.add_joint_prismatic(parent=body_b, child=body_a)
    else:
        joint_1 = builder.add_joint_prismatic(parent=body_a, child=body_b)

    builder.add_articulation([joint_0, joint_1])


for name, reversed_second in (("parallel", False), ("anti-parallel", True)):
    check_pair(reversed_second=reversed_second)
    print(f"{name}: accepted by add_articulation()")
```

camevor: The articulation validation does not actually validate that there are no loops, only that there are no links that are children of more than one distinct parent. Since reduced coordinate solvers require some root joint, pure cycles like above will result in a downstream error. On the other hand, e.g. `SolverXPBD` will happily accept and simulate this.

CHANGELOG.md
```diff
@@ -162,6 +162,20 @@
...
+- Change `SolverFeatherstone` FREE/DISTANCE `joint_qd` to use the public Newton convention: callers must pass six values per joint as child-COM linear velocity followed by angular velocity, both expressed in the joint parent frame; floating-root FREE/DISTANCE articulations now use a root-COM-centered internal solve frame for improved stability far from the world origin.
+- Assign one default visual color to capsule segments generated by `ModelBuilder.add_rod()` or `add_rod_graph()`; pass `color=` to choose an explicit cable color.
+- Warn from `ModelBuilder.add_joint()` when adding a joint between a pair of bodies that are already directly connected, since parallel joints between the same bodies are ambiguous and may behave differently across solvers.
+- Exclude particles without `ParticleFlags.ACTIVE` from `SolverImplicitMPM` grid transfers, including mass and velocity. Simulations that used inactive particles as fixed obstacles must keep those particles active and set their mass to `0` to retain kinematic boundary behavior.
+- Rework the API of the `reset()` operation in `SolverKamino` to use an explicit `ResetConfig` instead of many keyword arguments.
```
eric-heiden: Changelog needs to be cleaned up.

jcarius-nv: Cleaned up at final head `eb750af65cf32250ea8acb0aaa395f4f6bace860`. After merging current `main`, I removed the 13 unrelated carryover bullets and kept a single entry for this PR, including migration guidance to use `add_link()` before adding the intended joint. Focused strict-warning tests and the full pre-commit suite pass.

jcarius-nv:
@eric-heiden I took over this PR from Chris, ready for re-approval and merging now.

jcarius-nv (approved):

mzamoramora-nvidia (approved):

PR #3209, opened 2026-06-22
Fix IndexError in collapse_fixed_joints for sparse equality constraint attrs

### LINKED ISSUE #3054: Handle sparse custom-attribute defaults during fixed-joint collapse
## Summary

`ModelBuilder.collapse_fixed_joints()` can mutate custom-attribute backing storage as if optional/defaulted fields have already been materialized to the full row count. Custom attributes intentionally apply defaults lazily during `CustomAttribute.build_array(count)`, so an omitted optional field can have `attr.values == []` or a short/`None`-padded list even though the custom frequency has rows.

This came up while reviewing PR #2961, where MuJoCo equality constraints are moved into the `mujoco:equality_constraint` custom-attribute namespace. The equality-specific crash is one example of the invariant mismatch: code that reads custom attributes can use defaults lazily, but code that writes transformed values during collapse needs to materialize/pad the backing storage before assignment.

## Minimal repro

Run from a checkout where the namespaced MuJoCo equality row attributes are available:

```bash
uv run python repro_sparse_custom_attr_collapse.py
```

```python
import traceback
import warnings

import warp as wp
import newton

warnings.filterwarnings("ignore", category=DeprecationWarning)

shape_cfg = newton.ModelBuilder.ShapeConfig(density=1.0)

# Root body: survives collapse.
root = builder.add_link()
builder.add_shape_box(body=root, hx=0.5, hy=0.5, hz=0.5, cfg=shape_cfg)
root_joint = builder.add_joint_revolute(parent=-1, child=root, axis=wp.vec3(0.0, 1.0, 0.0))

# Fixed child: will be merged into root by collapse_fixed_joints().
merged = builder.add_link()
builder.add_shape_box(body=merged, hx=0.2, hy=0.2, hz=0.2, cfg=shape_cfg)
fixed_joint = builder.add_joint_fixed(parent=root, child=merged)

# Another surviving body referenced by the equality row.
target = builder.add_link()
builder.add_shape_box(body=target, hx=0.5, hy=0.5, hz=0.5, cfg=shape_cfg)
target_joint = builder.add_joint_revolute(parent=root, child=target, axis=wp.vec3(0.0, 0.0, 1.0))

builder.add_articulation([root_joint, fixed_joint, target_joint])

# Add an equality row through the namespaced custom-attribute path, but omit
# optional fields whose defaults should be supplied by CustomAttribute.build_array().
builder.add_custom_values(
    **{
        "mujoco:equality_constraint_type": int(newton.EqType.WELD),
        "mujoco:equality_constraint_body1": merged,
        "mujoco:equality_constraint_body2": target,
        "mujoco:equality_constraint_enabled": True,
        "mujoco:equality_constraint_world": 0,
        # Intentionally omit:
        # "mujoco:equality_constraint_anchor"
        # "mujoco:equality_constraint_relpose"
    }
)

try:
    builder.collapse_fixed_joints(verbose=True)
except Exception:
    traceback.print_exc()
    raise
```

Current output includes:

```text
IndexError: list assignment index out of range
  File "newton/_src/sim/builder.py", line 5725, in collapse_fixed_joints
    relpose_values[i] = merge_xform * relpose
```

Depending on which referenced body is merged and which equality type is used, the same pattern can also hit `anchor_values[i] = ...`.

## Expected behavior

`collapse_fixed_joints()` should either avoid mutating sparse custom-attribute value lists directly, or materialize the row before writing through it. Missing optional custom-attribute entries should behave the same as they do at finalize time: read as `attr.default`, then write the transformed value into backing storage that has been padded to the row index.

## Suggested fix direction

Add a small helper for mutation sites, for example:

```python
def _materialized_custom_value_list(attr, count):
    if attr.values is None:
        attr.values = []
    while len(attr.values) < count:
        attr.values.append(None)
    return attr.values
```

Then use it before assignment in `collapse_fixed_joints()` for attributes such as `mujoco:equality_constraint_anchor` and `mujoco:equality_constraint_relpose`. Reads can still preserve lazy default behavior via the existing `idx >= len(values)` / `None` fallback logic.

vreutskyy:
## Description

Closes #3054

`ModelBuilder.collapse_fixed_joints()` crashed with `IndexError: list assignment index out of range` when an equality constraint row omitted optional fields (`anchor`, `relpose`) that carry defaults via `CustomAttribute.default`.

Two bugs combined:

1. `anchor_values = anchor_attr.values or []` created a disconnected local list when `attr.values` is `None`, so indexed assignments never reached the backing storage.
2. Even with a real but short list, `values[i] = ...` raises `IndexError` when `i >= len(values)`.

The fix mirrors the existing `enabled_attr` pattern: initialize `attr.values` to `[]` when `None`, alias the local to the real storage, and pad with `None` before each write.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_collapse_fixed_joints
```

Both `test_collapse_fixed_joints_with_equality_constraints` and the new `test_collapse_fixed_joints_sparse_optional_fields` pass. The new test was verified to fail on the unfixed code with the exact `IndexError` from the issue.

## Bug fix

**Steps to reproduce:**

1. Register MuJoCo custom attributes on a builder.
2. Add a WELD equality constraint via `add_custom_values`, omitting `anchor` and `relpose`.
3. Add a fixed joint so that one of the referenced bodies gets merged.
4. Call `collapse_fixed_joints()`.

**Minimal reproduction:**

```python
import warp as wp
import newton

shape_cfg = newton.ModelBuilder.ShapeConfig(density=1.0)

root = builder.add_link()
builder.add_shape_box(body=root, hx=0.5, hy=0.5, hz=0.5, cfg=shape_cfg)
root_joint = builder.add_joint_revolute(parent=-1, child=root, axis=wp.vec3(0.0, 1.0, 0.0))

merged = builder.add_link()
builder.add_shape_box(body=merged, hx=0.2, hy=0.2, hz=0.2, cfg=shape_cfg)
fixed_joint = builder.add_joint_fixed(parent=root, child=merged)

target = builder.add_link()
builder.add_shape_box(body=target, hx=0.5, hy=0.5, hz=0.5, cfg=shape_cfg)
target_joint = builder.add_joint_revolute(parent=root, child=target, axis=wp.vec3(0.0, 0.0, 1.0))

builder.add_articulation([root_joint, fixed_joint, target_joint])
builder.add_custom_values(**{
    "mujoco:equality_constraint_type": int(newton.EqType.WELD),
    "mujoco:equality_constraint_body1": merged,
    "mujoco:equality_constraint_body2": target,
    "mujoco:equality_constraint_enabled": True,
    "mujoco:equality_constraint_world": 0,
    # anchor and relpose intentionally omitted
})

builder.collapse_fixed_joints()  # IndexError before fix
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a crash when collapsing fixed joints on MuJoCo models where equality-constraint rows omit optional anchor/relpose data. The model builder now safely handles sparse constraint specifications using proper defaulting, preventing index-related failures.
* **Tests**
  * Added a regression test covering equality constraints with missing optional fields to ensure joint-collapsing remains stable and the constraint count is preserved.

### DIFF
--- a/newton/_src/sim/builder.py
+++ b/newton/_src/sim/builder.py
@@ -5742,10 +5742,11 @@ class ModelBuilder:
         self.joint_constraint_count = len(self.joint_cts)
 
         # Remap equality constraint body/joint indices and transform anchors for merged bodies.
-        # Each ``*_values`` is the dense ``list`` backing the string-frequency CustomAttribute;
-        # the ``mujoco:equality_constraint`` ``add_custom_values`` path populates all twelve in
-        # lockstep so indexed access is safe for every ``i`` in
-        # ``range(self._equality_constraint_count)``.
+        # Each ``*_values`` is the ``list`` backing the string-frequency CustomAttribute. These
+        # lists are sparse: ``add_custom_values`` only populates the fields that were supplied,
+        # so an omitted optional field can be ``None``, shorter than the row count, or absent
+        # entirely. Reads go through ``_at`` (default fallback), and writes pad the list to the
+        # row index before assignment.
         body1_attr = self._eq_attr("equality_constraint_body1")
         body2_attr = self._eq_attr("equality_constraint_body2")
         type_attr = self._eq_attr("equality_constraint_type")
--- a/newton/_src/sim/builder.py
+++ b/newton/_src/sim/builder.py
@@ -5757,8 +5757,12 @@ class ModelBuilder:
         body1_values = body1_attr.values or []
         body2_values = body2_attr.values or []
         type_values = type_attr.values or []
-        anchor_values = anchor_attr.values or []
-        relpose_values = relpose_attr.values or []
+        if anchor_attr.values is None:
+            anchor_attr.values = []
+        anchor_values = anchor_attr.values
+        if relpose_attr.values is None:
+            relpose_attr.values = []
+        relpose_values = relpose_attr.values
         joint1_values = joint1_attr.values or []
         joint2_values = joint2_attr.values or []
         if enabled_attr.values is None:
@@ -5787,15 +5791,23 @@ class ModelBuilder:
                 merge_xform = body_merged_transform[old_body1]
                 if constraint_type == EqType.CONNECT:
                     anchor = axis_to_vec3(_at(anchor_values, i, anchor_attr.default))
+                    while len(anchor_values) <= i:
+                        anchor_values.append(None)
                     anchor_values[i] = wp.transform_point(merge_xform, anchor)
                 if constraint_type == EqType.WELD:
                     relpose = _at(relpose_values, i, relpose_attr.default)
+                    while len(relpose_values) <= i:
+                        relpose_values.append(None)
                     relpose_values[i] = merge_xform * relpose
 
             if body2_was_merged and constraint_type == EqType.WELD:
                 merge_xform = body_merged_transform[old_body2]
                 anchor = axis_to_vec3(_at(anchor_values, i, anchor_attr.default))
                 relpose = _at(relpose_values, i, relpose_attr.default)
+                while len(anchor_values) <= i:
+                    anchor_values.append(None)
+                while len(relpose_values) <= i:
+                    relpose_values.append(None)
                 anchor_values[i] = wp.transform_point(merge_xform, anchor)
                 relpose_values[i] = relpose * wp.transform_inverse(merge_xform)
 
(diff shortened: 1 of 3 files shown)


newton/_src/sim/builder.py
camevor: This comment is stale as `add_custom_values` does not necessarily populate all arrays, would be good to fix in the same go.

camevor (commented):
One comment that is stale. otherwise, looks good to me!

newton/_src/sim/builder.py
vreutskyy: Good catch — rewrote the comment in e97602e00 to describe the lists as sparse and document the read (`_at` default fallback) / write (pad-then-assign) handling, since the lockstep claim was exactly the invariant this PR removes.

camevor (approved):

PR #3214, opened 2026-06-22
Add USD mesh loading support

### LINKED ISSUE #2667: [REQ] Implement method for logging USD in Viewers
### Description

Requesting a Newton API which takes in a USD filepath or handle and outputs a Newton mesh, which can be consumed by Newton based viewers.

There is an existing method to convert simple USD files to Newton meshes, see below for an example
```
usd_stage = Usd.Stage.Open(newton.examples.get_asset("bunny.usd"))
usd_geom = UsdGeom.Mesh(usd_stage.GetPrimAtPath("/root/bunny"))
mesh_vertices = np.array(usd_geom.GetPointsAttr().Get())
mesh_indices = np.array(usd_geom.GetFaceVertexIndicesAttr().Get())
self.bunny_mesh = newton.Mesh(mesh_vertices, mesh_indices)
self.bunny_mesh.finalize()
```

However, for more complex USDs with dependencies (including USD that depend on assets that are on a URL, so need to be downloaded), a few more steps are reqyured. It would be great to have a general API which can convert to meshes for any input USD.

### Motivation / Use Case

We would like to support "visualization markers" in IsaacLab for Newton  based viewers which can show lightweight (instaceable and not part of newton model) general monitor/debugging shapes, like arrows, lines, or any input USD model, in training environments. 

We currently have this capability in Kit Visualizer (based on Isaac Sim).

see this example below for Visualization Markers using arrow meshes above anymal D robots
Image

### Alternatives Considered

_No response_

devshahofficial:
## Description

Related to #2667.

This PR is now scoped to the missing USD-to-Newton mesh loading path only. It does not add new viewer marker APIs; marker placement stays on the existing public `viewer.log_mesh()` + `viewer.log_instances()` flow.

What changed:

- Extend `newton.usd.get_mesh()` so it can load from a USD mesh prim, stage handle, file path, or HTTPS URL.
- Merge descendant `UsdGeom.Mesh` prims under `root_path` into one `newton.Mesh`, applying root-relative transforms, authored stage units, mirrored winding correction, and normal transforms.
- Resolve HTTPS USD references into the local cache, including redirects and absolute HTTPS dependencies, while rejecting cleartext HTTP and path traversal.
- Preserve the existing public `prim=` keyword for `newton.usd.get_mesh()` and `Mesh.create_from_usd()` while keeping `source` as the broader input name.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_usd_mesh_loading
uv run --extra dev -m newton.tests -k TestResolveUsdFromUrl
uv run --extra dev -m newton.tests -k test_viewer_log_shapes
uv run --extra dev docs/generate_api.py
uvx pre-commit run -a
```

## New feature / API change

```python
import newton

# Load a single mesh prim, preserving the existing prim= keyword.
mesh = newton.usd.get_mesh(prim=usd_mesh_prim, compute_inertia=False)

# Or load and merge mesh prims from a stage/path/HTTPS URL.
mesh = newton.usd.get_mesh("marker.usda", root_path="/Marker", compute_inertia=False)

viewer.log_mesh(
    "/markers/arrow",
    mesh.vertices,
    mesh.indices,
    normals=mesh.normals,
    uvs=mesh.uvs,
    color=mesh.color,
    roughness=mesh.roughness,
    metallic=mesh.metallic,
    hidden=True,
)
viewer.log_instances("/markers/arrows", "/markers/arrow", xforms, scales, colors, materials)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
   - Add `--coupled-view` to coupled multiphysics examples and expose `SolverCoupled` entry view/state helpers for rendering individual sub-solver views.
 - Add `BODY_F`, `PARTICLE_F`, and `JOINT_F` to `StateFlags`.
 - Add USD mesh marker logging via `newton.usd.get_mesh()` support for USD stages, file paths, and URLs plus `ViewerBase.log_usd()` and `ViewerBase.log_mesh_instances()`.
+- Add `newton.usd.get_mesh()` support for USD stages, file paths, and URLs.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,6 @@
   - Add standalone multiphysics examples and regression coverage for MuJoCo/Kamino, VBD, XPBD, MPM, and ADMM contacts.
   - Add `--coupled-view` to coupled multiphysics examples and expose `SolverCoupled` entry view/state helpers for rendering individual sub-solver views.
 - Add `BODY_F`, `PARTICLE_F`, and `JOINT_F` to `StateFlags`.
-- Add USD mesh marker logging via `newton.usd.get_mesh()` support for USD stages, file paths, and URLs plus `ViewerBase.log_usd()` and `ViewerBase.log_mesh_instances()`.
 - Add `newton.usd.get_mesh()` support for USD stages, file paths, and URLs.
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
   - Add standalone multiphysics examples and regression coverage for MuJoCo/Kamino, VBD, XPBD, MPM, and ADMM contacts.
   - Add `--coupled-view` to coupled multiphysics examples and expose `SolverCoupled` entry view/state helpers for rendering individual sub-solver views.
 - Add `BODY_F`, `PARTICLE_F`, and `JOINT_F` to `StateFlags`.
+- Add USD mesh marker logging via `newton.usd.get_mesh()` support for USD stages, file paths, and URLs plus `ViewerBase.log_usd()` and `ViewerBase.log_mesh_instances()`.
 
 ### Changed
 
--- a/newton/_src/geometry/types.py
+++ b/newton/_src/geometry/types.py
@@ -1517,23 +1517,24 @@ class Mesh:
     # ---- Factory methods ---------------------------------------------------
 
     @staticmethod
-    def create_from_usd(prim, **kwargs) -> "Mesh":
-        """Load a Mesh from a USD prim with the ``UsdGeom.Mesh`` schema.
+    def create_from_usd(source, **kwargs) -> "Mesh":
+        """Load a Mesh from a USD mesh prim, stage, file path, or URL.
 
         This is a convenience wrapper around :func:`newton.usd.get_mesh`.
         See that function for full documentation.
 
         Args:
-            prim: The USD prim to load the mesh from.
+            source: USD mesh prim, stage, file path, or URL to load the mesh
+                from.
             **kwargs: Additional arguments passed to :func:`newton.usd.get_mesh`
-                (e.g. ``load_normals``, ``load_uvs``).
+                (e.g. ``root_path``, ``load_normals``, ``load_uvs``).
 
         Returns:
             Mesh: A new Mesh instance.
         """
         from ..usd.utils import get_mesh  # noqa: PLC0415
 
-        result = get_mesh(prim, **kwargs)
+        result = get_mesh(source, **kwargs)
         if isinstance(result, tuple):
             return result[0]
         return result
--- a/newton/_src/geometry/types.py
+++ b/newton/_src/geometry/types.py
@@ -1517,7 +1517,7 @@ class Mesh:
     # ---- Factory methods ---------------------------------------------------
 
     @staticmethod
-    def create_from_usd(source, **kwargs) -> "Mesh":
+    def create_from_usd(source=None, *, prim=None, **kwargs) -> "Mesh":
         """Load a Mesh from a USD mesh prim, stage, file path, or URL.
 
         This is a convenience wrapper around :func:`newton.usd.get_mesh`.
@@ -1526,6 +1526,7 @@ class Mesh:
         Args:
             source: USD mesh prim, stage, file path, or URL to load the mesh
                 from.
+            prim: Legacy keyword alias for ``source`` when loading a USD prim.
             **kwargs: Additional arguments passed to :func:`newton.usd.get_mesh`
                 (e.g. ``root_path``, ``load_normals``, ``load_uvs``).
 
@@ -1533,6 +1534,11 @@ class Mesh:
             Mesh: A new Mesh instance.
         """
         from ..usd.utils import get_mesh  # noqa: PLC0415
+
+        if prim is not None:
+            if source is not None:
+                raise TypeError("Mesh.create_from_usd() received both 'source' and legacy 'prim'; pass only one.")
+            source = prim
 
         result = get_mesh(source, **kwargs)
         if isinstance(result, tuple):
--- a/newton/_src/usd/utils.py
+++ b/newton/_src/usd/utils.py
@@ -1040,7 +1040,7 @@ def _get_mesh_from_source(
 
 @overload
 def get_mesh(
-    source: Usd.Prim,
+    source: Usd.Prim | Usd.Stage | str | os.PathLike[str],
     load_normals: bool = False,
     load_uvs: bool = False,
     maxhullvert: int | None = None,
@@ -1058,7 +1058,7 @@ def get_mesh(
 
 @overload
 def get_mesh(
-    source: Usd.Prim | Usd.Stage | str | os.PathLike[str],
+    source: Usd.Prim,
     load_normals: bool = False,
     load_uvs: bool = False,
     maxhullvert: int | None = None,
--- a/newton/_src/usd/utils.py
+++ b/newton/_src/usd/utils.py
@@ -850,9 +850,197 @@ def _triangulate_face_varying_indices(counts: Sequence[int], flip_winding: bool)
     return corner_faces.reshape(-1)
 
 
+def _is_usd_url(source: str) -> bool:
+    """Return True when ``source`` is an HTTPS USD asset URL."""
+    return source.startswith("https://")
+
+
+def _open_usd_stage(source: str | os.PathLike[str]):
+    """Open a USD stage from a local path or resolved URL."""
+    if Usd is None:
+        raise ImportError("Failed to import pxr. Please install USD (e.g. via `pip install usd-core`).")
+
+    source_path = os.fspath(source)
+    if source_path.startswith("http://"):
+        raise ValueError("HTTP USD URLs are not supported; use HTTPS or download the asset explicitly.")
+    if _is_usd_url(source_path):
+        from ..utils.import_usd import resolve_usd_from_url  # noqa: PLC0415
+
+        source_path = resolve_usd_from_url(source_path)
+
+    stage = Usd.Stage.Open(source_path, Usd.Stage.LoadAll)
+    if stage is None:
+        raise FileNotFoundError(f"Unable to open USD stage: {source_path}")
+    return stage
+
+
+def _get_root_prim(stage, root_path: str | None):
+    """Return the merge root prim for a stage-backed mesh load."""
+    if root_path is None or root_path == "/":
+        return stage.GetPseudoRoot()
+
+    root = stage.GetPrimAtPath(root_path)
+    if not root or not root.IsValid():
+        raise ValueError(f"USD root path '{root_path}' does not exist.")
+    return root
+
+
+def _iter_mesh_prims(root) -> list[Any]:
+    """Return all mesh prims under ``root`` in traversal order."""
+    return [prim for prim in Usd.PrimRange(root) if prim.IsA(UsdGeom.Mesh)]
+
+
+def _matrix_to_numpy(matrix) -> np.ndarray:
+    """Convert a USD matrix value into a NumPy row-vector matrix."""
+    return np.array(matrix, dtype=np.float64)
+
+
+def _relative_transform_matrix(prim, root, xform_cache) -> np.ndarray:
+    """Return ``prim`` transform relative to the selected merge ``root``."""
+    prim_world = _matrix_to_numpy(xform_cache.GetLocalToWorldTransform(prim))
+    if root.IsPseudoRoot():
+        return prim_world
+
+    root_world = _matrix_to_numpy(xform_cache.GetLocalToWorldTransform(root))
+    return prim_world @ np.linalg.inv(root_world)
+
+
+def _transform_mesh_data(mesh: Mesh, matrix: np.ndarray) -> tuple[np.ndarray, np.ndarray, np.ndarray | None]:
+    """Apply a row-vector USD transform to mesh vertices, winding, and normals."""
+    vertices = np.asarray(mesh.vertices, dtype=np.float64)
+    vertices_h = np.concatenate((vertices, np.ones((len(vertices), 1), dtype=np.float64)), axis=1)
+    transformed_vertices = (vertices_h @ matrix)[:, :3].astype(np.float32)
+
+    indices = np.asarray(mesh.indices, dtype=np.int32).copy()
+    linear = matrix[:3, :3]
+    if np.linalg.det(linear) < 0.0:
+        indices = indices.reshape(-1, 3)
+        indices[:, [1, 2]] = indices[:, [2, 1]]
+        indices = indices.reshape(-1)
+
+    normals = None
+    if mesh.normals is not None:
+        try:
+            normal_matrix = np.linalg.inv(linear).T
+        except np.linalg.LinAlgError:
+            normal_matrix = None
+        if normal_matrix is not None:
+            normals = np.asarray(mesh.normals, dtype=np.float64) @ normal_matrix
+            lengths = np.linalg.norm(normals, axis=1, keepdims=True)
+            lengths[lengths < 1e-20] = 1.0
+            normals = (normals / lengths).astype(np.float32)
+
+    return transformed_vertices, indices, normals
+
+
+def _get_mesh_from_source(
+    source,
+    *,
+    root_path: str | None,
+    load_normals: bool,
+    load_uvs: bool,
+    maxhullvert: int | None,
+    face_varying_normal_conversion: Literal["vertex_averaging", "angle_weighted", "vertex_splitting"],
+    vertex_splitting_angle_threshold_deg: float,
+    preserve_facevarying_uvs: bool,
+    compute_inertia: bool,
+    apply_stage_units: bool,
+) -> Mesh:
+    """Load and merge mesh prims from a USD stage, path, URL, or prim subtree."""
+    if Usd is None or UsdGeom is None:
+        raise ImportError("Failed to import pxr. Please install USD (e.g. via `pip install usd-core`).")
+    if preserve_facevarying_uvs:
+        raise ValueError(
+            "preserve_facevarying_uvs is not supported for merged USD sources; "
+            "load a single mesh prim with return_uv_indices=True or use preserve_facevarying_uvs=False."
+        )
+
+    if isinstance(source, str | os.PathLike):
+        stage = _open_usd_stage(source)
+        root = _get_root_prim(stage, root_path)
+    elif isinstance(source, Usd.Stage):
+        stage = source
+        root = _get_root_prim(stage, root_path)
+    elif isinstance(source, Usd.Prim):
+        stage = source.GetStage()
+        root = source if root_path is None else _get_root_prim(stage, root_path)
+    else:
+        raise TypeError(
+            f"get_mesh() expected a USD prim, USD stage, filesystem path, or URL; received {type(source).__name__}."
+        )
+
+    mesh_prims = _iter_mesh_prims(root)
+    if not mesh_prims:
+        raise ValueError(f"No UsdGeom.Mesh prims found under '{root.GetPath()}'.")
+
+    linear_unit = 1.0
+    if apply_stage_units and UsdGeom.StageHasAuthoredMetersPerUnit(stage):
+        linear_unit = float(UsdGeom.GetStageMetersPerUnit(stage))
+
+    xform_cache = UsdGeom.XformCache(Usd.TimeCode.Default())
+    vertices_parts = []
+    indices_parts = []
+    normals_parts = []
+    uvs_parts = []
+    vertex_offset = 0
+    all_have_normals = True
+    all_have_uvs = True
+    source_meshes = []
+
+    for prim in mesh_prims:
+        source_mesh = get_mesh(
+            prim,
+            load_normals=load_normals,
+            load_uvs=load_uvs,
+            maxhullvert=maxhullvert,
+            face_varying_normal_conversion=face_varying_normal_conversion,
+            vertex_splitting_angle_threshold_deg=vertex_splitting_angle_threshold_deg,
+            preserve_facevarying_uvs=preserve_facevarying_uvs,
+            compute_inertia=False,
+        )
+        source_meshes.append(source_mesh)
+        matrix = _relative_transform_matrix(prim, root, xform_cache)
+        vertices, indices, normals = _transform_mesh_data(source_mesh, matrix)
+
+        vertices_parts.append(vertices)
+        indices_parts.append(indices + vertex_offset)
+        vertex_offset += len(vertices)
+
+        if normals is None:
+            all_have_normals = False
+        else:
+            normals_parts.append(normals)
+
+        if source_mesh.uvs is None:
+            all_have_uvs = False
+        else:
+            uvs_parts.append(np.asarray(source_mesh.uvs, dtype=np.float32))
+
+    vertices = np.concatenate(vertices_parts, axis=0)
+    if linear_unit != 1.0:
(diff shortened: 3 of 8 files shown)


devshahofficial:
@vreutskyy this is ready for a first pass when you have a chance.

I kept the scope to the USD-to-Newton mesh helper plus lightweight viewer logging API discussed in #2667. The PR covers stage/file/URL sources, root-relative mesh merging, stage-unit handling, mirrored transform winding, viewer mesh instance logging, and basic caching.

newton/_src/usd/utils.py
```diff
@@ -850,9 +850,182 @@ def _triangulate_face_varying_indices(counts: Sequence[int], flip_winding: bool)
...
+    linear = matrix[:3, :3]
+    if np.linalg.det(linear) < 0.0:
+        indices = indices.reshape(-1, 3)
+        indices[:, [1, 2]] = indices[:, [2, 1]]
+        indices = indices.reshape(-1)
+
+    normals = None
+    if mesh.normals is not None:
+        try:
+            normal_matrix = np.linalg.inv(linear)
```
vreutskyy: The normal transform is missing a transpose. Vertices use USD's row-vector convention here (`v_h @ matrix`), so normals must transform as `n @ (M^-1)^T`, i.e.:

```python
normal_matrix = np.linalg.inv(linear).T
```

I verified this empirically against this commit: a mesh with a +90° Z rotation and an authored +X normal. The vertex correctly rotates +X→+Y, but the normal comes out **−Y** (the inverse rotation). So this is wrong for *any* rotation, not only non-uniform scale. Adding `.T` makes the normal come out +Y as expected.

The existing tests don't catch this because they only exercise translation and a mirror (negative scale, where the winding flip dominates). Could you add the `.T` and a small regression test with an authored rotation + `load_normals=True` asserting the rotated normal direction? (CodeRabbit flagged this too.)

newton/_src/viewer/viewer.py
```diff
@@ -1314,6 +1315,162 @@ def _ensure_vec4_array(arr, default):
...
+        materials = _ensure_vec4_array(materials)
+
+        mesh_path = self._populate_geometry(
+            int(newton.GeoType.MESH),
+            (1.0, 1.0, 1.0),
+            0.0,
+            bool(mesh.is_solid),
+            geo_src=mesh,
+        )
+        self.log_instances(name, mesh_path, xforms, scales, colors, materials, hidden=hidden)
```
vreutskyy: `name` needs to be qualified before dispatch:

```python
name = self._qualify(name)
mesh_path = self._populate_geometry(...)
self.log_instances(name, mesh_path, ...)
```

Every other public `log_*` method routes the name through `self._qualify()` first (`log_arrows`, `log_geo`, the base `log_instances` wrapper at ~1585), and `log_instances`' own docstring requires it. Without it, instances logged under a non-default layer aren't namespaced and two layers logging the same path collide.

Verified on this commit: after `viewer.activate("markers")`, `log_mesh_instances("/debug/marker", ...)` forwards `/debug/marker` to the backend instead of `/layers/markers/debug/marker`. Adding the `_qualify` line fixes it. `log_usd` inherits the same issue through this method. (CodeRabbit flagged this too.)

vreutskyy (changes requested):
Thanks for this, @devshahofficial — nicely scoped to what #2667 asked for, and the test module is a good start. Two correctness issues to fix before merge (both inline, and I verified each empirically by running the PR code on this commit):

1. **Normal transform is missing a transpose** in `_transform_mesh_data` — normals come out under the *inverse* rotation, so any rotated USD marker shades incorrectly (not just non-uniform scale). One-character fix plus a rotation+normals regression test.
2. **`log_mesh_instances` doesn't qualify the instance name** to the active layer, so multi-layer logging collides. `log_usd` inherits it. One-line fix.

A couple of non-blocking notes:
- The CI docstring-coverage check is failing (~24% vs 80%) — the new private helpers (`_get_mesh_from_source`, `_transform_mesh_data`, `_open_usd_stage`, etc.) just need short docstrings.
- Nice catch forwarding `color`/`roughness`/`metallic` from `geo_src` in `log_geo` — that was a real gap.
- Dropping material for merged multi-mesh sources is fine as a documented follow-up.

Happy to re-review quickly once the two fixes are in.

adenzler-nvidia (changes requested):
Thanks for taking this on. After going through the viewer internals, I think the marker capability from #2667 is already achievable with the public API we ship today, without dedicated USD-loading code.

**The marker flow with existing public API.** #2667 is a *dynamic instanced overlay* — a prototype mesh drawn at N caller-supplied transforms that change each frame, with a variable count, explicitly not part of the simulated `Model`. The viewer's register-then-instance path already does exactly that:

```python
from pxr import Usd
import newton.usd

# USD -> newton.Mesh (existing; carries color/roughness/metallic/normals/uvs)
prim = Usd.Stage.Open("marker.usda").GetPrimAtPath("/Marker/mesh")
mesh = newton.usd.get_mesh(prim, load_normals=True)

# register the prototype once (hidden so the prototype itself isn't drawn)
viewer.log_mesh("/markers/arrow", mesh.vertices, mesh.indices,
                normals=mesh.normals, uvs=mesh.uvs,
                color=mesh.color, roughness=mesh.roughness, metallic=mesh.metallic,
                hidden=True)

# each frame — placement is the caller's, nothing enters the Model:
viewer.log_instances("/markers/arrows", "/markers/arrow", xforms, scales, colors, materials)
```

`log_instances` resolves the prototype purely by the name passed to `log_mesh`, so no private API is involved. Placement stays with the caller via `xforms`, and the mesh never touches `ModelBuilder`/`Model`.

@matthewtrepte — since you filed #2667: does the existing-API flow above cover what you need for IsaacLab visualization markers (load the asset once, then re-`log_instances` with your own per-frame transforms and count)? Or is there a requirement it misses — multi-mesh/multi-material assets, instancer prototypes, something else — that would justify dedicated loading API? Want to make sure we're evaluating this against the real need.

devshahofficial:
Thanks @vreutskyy @adenzler-nvidia. I pushed a rebased pass with the concrete fixes from review:

- fixed USD normal transforms with the row-vector inverse-transpose path and added a rotated-normal regression
- qualified `log_mesh_instances()` names through the active layer and added a layer regression
- added short docstrings for the new USD helper path

Validated locally:
- `uv run --extra dev -m newton.tests -k test_usd_mesh_viewer_logging`
- `uv run --extra dev -m newton.tests -k test_import_usd.TestImportUsdMeshNormals`
- `uv run --extra dev -m newton.tests -k test_viewer_layers`
- `uvx pre-commit run -a`

On the broader API question: I’ll keep this scoped and won’t expand it. If the existing `log_mesh` + `log_instances` flow fully covers #2667, I’m happy to trim or close this and keep only whichever mesh-helper piece maintainers still want.

newton/_src/viewer/viewer.py
```diff
@@ -1314,6 +1315,163 @@ def _ensure_vec4_array(arr, default):
...
+            source_key = os.fspath(source)
+            if not source_key.startswith(("http://", "https://")):
+                try:
+                    stat = os.stat(source_key)
+                except OSError:
+                    stat = None
+                if stat is not None:
+                    source_key = (os.path.abspath(source_key), stat.st_mtime_ns, stat.st_size)
+        else:
+            source_key = (type(source).__name__, id(source))
```
vreutskyy: **Use a stable cache key for USD handles.** `id(source)` is only unique while the Python wrapper is alive, but this cache retains only the integer. I reproduced this through `log_usd()` with transient prim handles: after alternating `/A` and `/B` and allowing the wrappers to be collected, `/B` returned `/A`'s cached mesh on the sixth call. Key prims by a stable stage/root-layer identifier plus prim path (with a policy for mutable stages), or avoid caching handle sources by default.

newton/_src/viewer/viewer.py
```diff
@@ -1314,6 +1315,163 @@ def _ensure_vec4_array(arr, default):
...
+                if len(arr_np) == 1 and num_instances > 1:
+                    arr_np = np.tile(arr_np, (num_instances, 1))
+            return wp.array([wp.vec4(*value) for value in arr_np], dtype=wp.vec4, device=self.device)
+
+        scales = _ensure_vec3_array(scales, wp.vec3(1.0, 1.0, 1.0))
+        colors = None if colors is None else _ensure_vec3_array(colors, wp.vec3(1.0, 1.0, 1.0))
+        materials = _ensure_vec4_array(materials)
+
+        name = self._qualify(name)
+        mesh_path = self._populate_geometry(
```
vreutskyy: **Account for mirrored per-instance scales when selecting the prototype.** This call always leaves `mirror=False`, so `scales=(-1, 1, 1)` registers the original winding; I confirmed the uploaded indices remain `[0, 1, 2]`. The existing model batching path explicitly uses a winding-flipped prototype when the scale determinant is negative because default back-face culling otherwise reverses the visible faces. Partition mixed-sign batches and select the corresponding prototype, or reject/document negative determinant scales.

vreutskyy (changes requested):
Thanks for the updated pass. I reviewed the current head and ran the focused `test_usd_mesh_viewer_logging` suite (11/11 passed), then exercised the new edge cases with targeted in-memory reproductions.

I found two additional correctness issues, noted inline: transient USD handle cache keys can collide and return the wrong mesh, and negative per-instance scales bypass the viewer's mirrored-winding path.

I also independently reproduced and agree with the existing open threads on merged face-varying UVs, per-instance array length validation, and prototype color/mirrored material handling. Requesting changes until these correctness issues are addressed.

devshahofficial:
Thanks for the re-review. I pushed a rebased pass on upstream/main and addressed the remaining correctness items:

- reject cleartext HTTP USD URLs; HTTPS still uses the resolver path
- reject `preserve_facevarying_uvs=True` for merged USD-source loads until UV indices can be preserved
- validate per-instance vec3/vec4 array lengths after broadcast
- stop caching live USD handle sources so transient wrapper IDs cannot collide
- reject negative-determinant `log_mesh_instances` scales rather than partitioning mixed-sign batches in this PR
- include mesh color/roughness/metallic in geometry cache keys and mirrored prototype uploads

Validated:
- `uv run --extra dev -m newton.tests -k test_usd_mesh_viewer_logging`
- `uv run --extra dev -m newton.tests -k test_import_usd.TestImportUsdMeshNormals`
- `uv run --extra dev -m newton.tests -k test_viewer_layers`
- `uv run --extra dev -m newton.tests -k test_viewer_geometry_batching`
- `uv run --extra dev -m newton.tests -k test_viewer_log_shapes`
- `uv run --extra dev -m newton.tests -k test_viewer_usd`
- `uvx pre-commit run -a`

All passed locally on macOS CPU. The `uv` pyproject warning is still the existing local uv parser warning, not a test failure.

newton/_src/usd/utils.py
```diff
@@ -862,12 +1050,15 @@ def get_mesh(
...
+    root_path: str | None = None,
+    compute_inertia: bool = True,
+    apply_stage_units: bool = True,
 ) -> Mesh: ...
 
 
 @overload
 def get_mesh(
-    prim: Usd.Prim,
+    source: Usd.Prim | Usd.Stage | str | os.PathLike[str],
```
vreutskyy: **Narrow the tuple-return overload to inputs supported at runtime.** I confirmed that this overload advertises `Usd.Stage`, path, and URL sources for `return_uv_indices=True`, while each of those calls reaches the explicit `ValueError("return_uv_indices is only supported when loading a single UsdGeom.Mesh prim.")`. Please restrict this overload to `Usd.Prim` with `root_path=None` (and clarify that constraint in the parameter documentation) so statically valid calls do not deterministically fail at runtime.

newton/_src/viewer/viewer.py
```diff
@@ -1314,6 +1315,177 @@ def _ensure_vec4_array(arr, default):
...
+                arr_np = arr_np.reshape(-1, 4)
+                if len(arr_np) == 1 and num_instances > 1:
+                    arr_np = np.tile(arr_np, (num_instances, 1))
+            _validate_instance_count(len(arr_np), "vec4")
+            return wp.array([wp.vec4(*value) for value in arr_np], dtype=wp.vec4, device=self.device)
+
+        scales = _ensure_vec3_array(scales, wp.vec3(1.0, 1.0, 1.0))
+        colors = None if colors is None else _ensure_vec3_array(colors, wp.vec3(1.0, 1.0, 1.0))
+        materials = _ensure_vec4_array(materials)
+        scale_values = scales.numpy()
```
vreutskyy: **Non-blocking design question:** this unconditionally calls `.numpy()` in a per-frame marker path, which synchronizes CUDA even when `scales` was omitted and the generated unit scales are known positive. ViewerUSD/RTX/Rerun already need host data, but ViewerGL’s CUDA-interoperability path otherwise keeps transforms/scales on-device, so this defeats that path. The synchronization itself is certain; I’m less certain whether the preferred tradeoff is host-only/best-effort validation, an explicit batch `mirror` flag, or opt-in debug validation. @adenzler-nvidia, could you give a second opinion on whether runtime value validation belongs here?

vreutskyy (changes requested):
Thanks for addressing the previous round—the original six correctness issues are fixed, and the updated focused suite passes 18/18 locally.

I’m requesting changes for two remaining confirmed issues:

1. I independently reproduced the HTTPS-policy bypass reported in the open CodeRabbit thread. An accepted `https://` top-level layer containing an `http://` reference caused `resolve_usd_from_url()` to call `requests.get()` for that cleartext dependency. Automatic redirects can downgrade similarly because the resolver follows them without validating each destination.
2. The `return_uv_indices=True` overload still advertises stage/path/URL inputs even though those combinations always raise at runtime; details inline.

I also left a non-blocking design question inline about the unconditional CUDA-to-host synchronization in the per-frame scale path and tagged @adenzler-nvidia for a second opinion. That synchronization is definite; the preferred API tradeoff is the part where additional input would help.

devshahofficial:
Pushed follow-up commit `2eb5494e` addressing the latest concrete review items:

- Enforce HTTPS throughout `resolve_usd_from_url()`, including the top-level URL, redirect targets, and nested/reference URLs.
- Add resolver regressions for cleartext top-level URLs, cleartext references, cleartext redirects, and allowed HTTPS redirects.
- Narrow the `get_mesh(..., return_uv_indices=True)` overload to the runtime-supported single `Usd.Prim` / `root_path=None` path and clarify that in the docs.
- Add docstrings for the new USD mesh viewer logging test/probe helpers to address the CodeRabbit docstring coverage warning.

The branch already includes the earlier fixes for layer path qualification, normal transforms, live handle caching, negative-determinant `log_mesh_instances()` scales, and mesh material cache keys. I left the non-blocking `.numpy()` design question open for @adenzler-nvidia's input since that thread explicitly asks for a second opinion on the tradeoff.

Validation:

- Regression check without the resolver implementation fix: `uv run --extra dev -m newton.tests -k TestResolveUsdFromUrl` failed as expected in a temporary worktree with only the new tests applied.
- `uv run --extra dev -m newton.tests -k test_usd_mesh_viewer_logging`
- `uv run --extra dev -m newton.tests -k TestResolveUsdFromUrl`
- `uvx pre-commit run -a`

All fixed-tree checks passed locally on macOS CPU. The local `uv` pyproject parser warning still appears, but it did not fail the test or pre-commit runs.

newton/_src/utils/import_usd.py
```diff
@@ -4417,10 +4442,11 @@ def _extract_references(layer_str, parent_url_folder, parent_local_folder):
             continue
         downloaded_urls.add(ref_url)
         try:
-            response = requests.get(ref_url, allow_redirects=True, timeout=request_timeout_s)
+            response, resolved_ref_url = _download_https_url(ref_url)
```
vreutskyy, line 4594: **Handle absolute HTTPS references before downloading them.** `_extract_references()` derives `local_path` directly from `raw_ref`; for `references = @https://cdn.example.com/child.usd@`, Windows normalizes that to an invalid `https:\\...` filesystem path. I reproduced this on the current head: both HTTPS requests occurred, then the broad exception handler printed `Failed to download`, and only `scene.usd` existed in the cache. On POSIX the derived `https:/...` path is also not a usable replacement for the absolute asset reference. Please map absolute URLs to a safe cache-relative path and rewrite the layer reference accordingly, or reject them explicitly instead of silently returning a stage with a missing dependency.

newton/_src/usd/utils.py
```diff
@@ -850,9 +850,197 @@ def _triangulate_face_varying_indices(counts: Sequence[int], flip_winding: bool)
...
+        texture=material_source.texture if material_source is not None else None,
+        metallic=material_source.metallic if material_source is not None else None,
+        roughness=material_source.roughness if material_source is not None else None,
+    )
+
+
 @overload
 def get_mesh(
-    prim: Usd.Prim,
+    source: Usd.Prim,
```
vreutskyy, line 1061: **The overload narrowing is applied to the wrong overload.** This `return_uv_indices=False` overload must retain the broad `Usd.Prim | Usd.Stage | str | os.PathLike[str]` source type because default path/stage calls are valid at runtime. Conversely, the `return_uv_indices=True` overload below still has that broad union even though those calls deterministically raise. Please restore the broad source type here and narrow the tuple-return overload to `Usd.Prim` with `root_path: None`.

vreutskyy (changes requested):
Thanks for the follow-up. I re-reviewed the latest head and verified that HTTPS is now enforced across top-level downloads, redirects, and nested references. The focused USD mesh/viewer suite passes 18/18 and the resolver suite passes 11/11 locally.

I’m requesting changes for two remaining confirmed issues, both noted inline:

1. Allowed absolute HTTPS references are fetched but cannot be cached/resolved correctly because the raw URL is used as a filesystem-relative path; I reproduced the dependency request followed by a swallowed write failure, leaving only the root layer on disk.
2. The overload narrowing was applied to the wrong overload: the valid default-return path was narrowed, while the tuple-return overload still advertises invalid stage/path/URL calls.

The `.numpy()` synchronization question remains non-blocking pending @adenzler-nvidia’s second opinion. I do not consider the general HTTPS-host SSRF hardening suggestion a blocker for this local library.

vreutskyy:
Thanks for working through the review feedback, and apologies that this scope decision is coming late. After revisiting #2667 and discussing the API surface with @adenzler-nvidia, we’d like to narrow this PR to the genuinely missing USD-loading capability:

- Keep `newton.usd.get_mesh()` support for loading a USD path, stage, or URL into a `newton.Mesh`, including the resolver robustness and focused tests needed for that path.
- Drop or defer `ViewerBase.log_mesh_instances()` and `ViewerBase.log_usd()` and their supporting viewer caching, broadcasting, scale/winding, and material-handling changes.

The dynamic marker flow is already available through the existing public `log_mesh()` + `log_instances()` API. If IsaacLab identifies a concrete requirement that flow cannot cover, we can evaluate a separate focused viewer API with that requirement in hand.

Please still address the remaining loader-specific correctness comments (the overload signatures and absolute HTTPS reference handling) within the narrowed scope.

devshahofficial:
Thanks, that scope makes sense. I’ll narrow this PR to the loader path and drop the new `ViewerBase.log_mesh_instances()` / `ViewerBase.log_usd()` API plus the supporting viewer cache/broadcast/material changes.

For the absolute HTTPS reference issue, I suggest keeping support rather than rejecting it: cache absolute HTTPS dependencies under a deterministic safe relative path like `_external_usd/<url-hash>/<basename>` and rewrite the layer reference to that local path. That keeps URL-loaded USDs with CDN-hosted dependencies usable from the local cache while avoiding URL-shaped filesystem paths.

For the marker flow, I’ll leave it on the existing public `log_mesh()` + `log_instances()` path and keep only `newton.usd.get_mesh()` as the helper for constructing the prototype mesh.

devshahofficial:
Pushed rebased follow-up `116e3fa2` on current `upstream/main`.

What changed:

- Narrowed the PR to USD mesh loading only.
- Removed the proposed `ViewerBase.log_mesh_instances()` / `ViewerBase.log_usd()` APIs and supporting viewer cache/broadcast/material changes.
- Kept marker usage on the existing public `log_mesh()` + `log_instances()` path.
- Restored the default-return `get_mesh()` overload to accept `Usd.Prim | Usd.Stage | str | os.PathLike[str]` and narrowed only the tuple-return overload to `Usd.Prim` with `root_path=None`.
- Fixed absolute HTTPS USD references by caching them under deterministic safe relative paths (`_external_usd/<hash>/<basename>`) and rewriting layer references to those local cache paths.
- Renamed the new focused test module to `test_usd_mesh_loading.py` and updated the changelog entry to remove the deferred viewer APIs.

Validation:

- Tests-only regression check against the previous pushed head failed as expected for `test_absolute_https_reference_cached_safely`.
- `uv run --extra dev -m newton.tests -k test_usd_mesh_loading`
- `uv run --extra dev -m newton.tests -k TestResolveUsdFromUrl`
- `uv run --extra dev -m newton.tests -k test_viewer_log_shapes`
- `uvx pre-commit run -a`

All fixed-tree checks passed locally on macOS CPU. After the rebase, GitHub now reports the PR as mergeable; CodeRabbit/pre-commit.ci are pending on the new head.

newton/_src/usd/utils.py
```diff
@@ -877,11 +1068,14 @@ def get_mesh(
...
     return_uv_indices: Literal[True] = True,
+    root_path: None = None,
+    compute_inertia: bool = True,
+    apply_stage_units: bool = True,
 ) -> tuple[Mesh, np.ndarray | None]: ...
 
 
 def get_mesh(
-    prim: Usd.Prim,
+    source: Usd.Prim | Usd.Stage | str | os.PathLike[str],
```
vreutskyy: Please preserve `prim` as the public keyword here, even though the accepted value is now broader, or add a deprecation-compatible alias. `newton.usd.get_mesh(prim=prim)` worked before this PR and now raises `TypeError: get_mesh() got an unexpected keyword argument 'prim'`. Public parameter renames require a deprecation period.

newton/_src/geometry/types.py
```diff
@@ -1517,23 +1517,24 @@ def __hash__(self) -> int:
     # ---- Factory methods ---------------------------------------------------
 
     @staticmethod
-    def create_from_usd(prim, **kwargs) -> "Mesh":
-        """Load a Mesh from a USD prim with the ``UsdGeom.Mesh`` schema.
+    def create_from_usd(source, **kwargs) -> "Mesh":
```
vreutskyy: The same compatibility issue applies to this public factory: `Mesh.create_from_usd(prim=prim)` previously worked and now fails because `prim` was renamed to `source`. Please retain the old keyword or support it through a deprecation-compatible alias.

CHANGELOG.md
```diff
@@ -26,6 +26,8 @@
   - Add standalone multiphysics examples and regression coverage for MuJoCo/Kamino, VBD, XPBD, MPM, and ADMM contacts.
   - Add `--coupled-view` to coupled multiphysics examples and expose `SolverCoupled` entry view/state helpers for rendering individual sub-solver views.
 - Add `BODY_F`, `PARTICLE_F`, and `JOINT_F` to `StateFlags`.
+- Add USD mesh marker logging via `newton.usd.get_mesh()` support for USD stages, file paths, and URLs plus `ViewerBase.log_usd()` and `ViewerBase.log_mesh_instances()`.
```
vreutskyy: Please remove this stale entry. `ViewerBase.log_usd()` and `ViewerBase.log_mesh_instances()` were removed when the PR scope was narrowed, and the loader-only entry immediately below already describes the remaining change.

vreutskyy (changes requested):
Thanks for narrowing the PR to the USD-to-`newton.Mesh` loader. Two changes remain before this is ready:

1. Preserve compatibility with the existing public `prim=` keyword in both `newton.usd.get_mesh()` and `Mesh.create_from_usd()` (or introduce a deprecation-compatible alias).
2. Remove the stale viewer-API changelog entry and update the PR title, description, test plan, and example so they describe the loader-only scope.

I reran the focused loader/resolver tests and existing mesh regressions, and also exercised an absolute HTTPS reference through the real public loader; those parts look good.

devshahofficial:
@vreutskyy pushed a rebased pass on current `upstream/main`. Kept this one loader-only.

Fixed:
- kept `prim=` working for both `newton.usd.get_mesh()` and `Mesh.create_from_usd()`
- added small conflict checks for ambiguous `source` + `prim` calls
- removed the stale viewer-API changelog entry
- updated the PR title/body/test plan/example so it no longer advertises the deferred viewer APIs

Small take: I did not add a deprecation warning for `prim=`. Since it was already public API and is still a solid name for prim-only calls, warning users here felt like noise. `source` now just covers the broader stage/path/URL cases.

Validated on the rebased head:
- `uv run --extra dev -m newton.tests -k test_usd_mesh_loading`
- `uv run --extra dev -m newton.tests -k TestResolveUsdFromUrl`
- `uv run --extra dev -m newton.tests -k test_viewer_log_shapes`
- `uv run --extra dev docs/generate_api.py`
- `uvx pre-commit run -a`

The local `uv` run still prints the existing `[tool.uv.build-backend]` settings warning, but the commands completed cleanly. Should be ready for another pass 🙂

vreutskyy (approved):
Re-reviewed the latest head. The requested public `prim=` compatibility and loader-only scope/documentation changes are addressed, the focused local suites pass, and the full standard CI workflow is green. Approved; the required AWS GPU test approval remains with the maintainers.

devshahofficial:
Thanks @vreutskyy. Standard CI, pre-commit.ci, CodeRabbit, EasyCLA, docs, and license checks are green on the latest head.

The only remaining blockers I see are the AWS GPU test/benchmark workflows waiting on manual approval for the external PR. Could a maintainer approve those when convenient?

adenzler-nvidia (approved):

PR #3218, opened 2026-06-23
Fix add_usd stripping articulations from unrelated floating bodies

### LINKED ISSUE #3002: [BUG] `add_usd`: a stray joint strips auto-created articulations from unrelated floating rigid bodies in the same parse range
## Summary

`ModelBuilder.add_usd` (via `parse_usd`) auto-wraps each loose rigid body into its own articulation by adding a base/free joint. That step is gated by a **global** check over the whole parsed range:

```python
no_articulations = UsdPhysics.ObjectType.Articulation not in ret_dict
has_joints       = any(<joint types in ret_dict>)
if not (no_articulations and has_joints):
    builder._add_base_joints_to_floating_bodies(...)
```

Because the condition is evaluated over the **entire range** and applied to **all** floating bodies, a *single* authored joint that lives under no articulation root — e.g. a leftover `FixedJoint` on an otherwise-static prop — makes `has_joints=True` and `no_articulations=True`, which **silently suppresses base-joint (→ articulation) creation for every unrelated free rigid body in the same `add_usd` call**. Those bodies then have no articulation, so `ArticulationView('/path/to/body')` raises `KeyError: "No articulations matching pattern ..."`.

### Expected behavior

`/World/Box` is an independent free rigid body and should always be wrapped into its own articulation, regardless of an unrelated authored joint elsewhere in the parsed range.

### Actual behavior

```
stray_joint=False  articulations=['/World/Box']  ->  Box: ['/World/Box']
stray_joint=True   articulations=[]              ->  Box: MISSING
```

Adding the unrelated `FixedJoint` on `/World/StaticProp` removes `/World/Box`'s articulation entirely. No error or warning is raised at parse time; the failure only surfaces later when something looks the body up by prim path (e.g. `ArticulationView`).

## Impact

This is easy to hit in any pipeline that imports a whole scene in one `add_usd` call (e.g. Isaac Lab's Newton cloner, which parses `/World/envs/env_0` containing the robot, props, and free objects together). A single static prop carrying a leftover joint — common after "bake to static collider" conversions that strip `RigidBodyAPI`/`ArticulationRootAPI` but not the joint prims — makes **every** free rigid body in that environment fail `ArticulationView` lookup with `KeyError: "No articulations matching pattern ..."`. The cause is non-obvious because the offending prim is unrelated to the bodies that break.

## Suggested fix

Scope the decision rather than applying it globally:
- Decide `_add_base_joints_to_floating_bodies` per **connected component** (a body with no incident authored joint should still get a free base joint even if other, disjoint bodies have joints), or
- Restrict the `has_joints`/`no_articulations` heuristic to the subtree/component a given floating body actually belongs to, and/or ignore joints under excluded/static subtrees.

At minimum, emit a warning when floating rigid bodies are left without a base joint because of this heuristic, so the silent "missing articulation" is diagnosable.

## Workaround

Ensure static-collider assets contain **no** joint prims (delete `*Joint` prims, not just the `RigidBodyAPI`/`ArticulationRootAPI` schemas) so `has_joints` stays `False` for the range.

### Reproduction Script

```python
import newton


def build_stage(with_stray_joint: bool) -> Usd.Stage:
    world = UsdGeom.Xform.Define(stage, "/World")
    stage.SetDefaultPrim(world.GetPrim())

    # Free dynamic rigid body -- should be wrapped into its own articulation.
    box = UsdGeom.Cube.Define(stage, "/World/Box")
    box.GetSizeAttr().Set(0.1)
    box.AddTranslateOp().Set(Gf.Vec3d(0.0, 0.0, 1.0))
    UsdPhysics.CollisionAPI.Apply(box.GetPrim())
    UsdPhysics.RigidBodyAPI.Apply(box.GetPrim())
    UsdPhysics.MassAPI.Apply(box.GetPrim()).CreateMassAttr(1.0)

    # Static collider -- no rigid body, no articulation root.
    prop = UsdGeom.Cube.Define(stage, "/World/StaticProp")
    prop.GetSizeAttr().Set(0.2)
    UsdPhysics.CollisionAPI.Apply(prop.GetPrim())

    if with_stray_joint:
        # A stray FixedJoint with no articulation root anywhere.
        joint = UsdPhysics.FixedJoint.Define(stage, "/World/StaticProp/StrayFixedJoint")
        joint.CreateBody0Rel().SetTargets(["/World/StaticProp"])

    return stage


def box_articulation(with_stray_joint: bool) -> list[str]:
    builder.add_usd(build_stage(with_stray_joint), root_path="/World")
    arts = list(getattr(builder, "articulation_key", None) or getattr(builder, "articulation_label", []))
    box = [a for a in arts if "Box" in a]
    print(f"stray_joint={str(with_stray_joint):5}  articulations={arts}  ->  Box: {box or 'MISSING'}")
    return box


assert box_articulation(False), "control: Box should get an articulation"
assert not box_articulation(True), "bug: stray joint stripped Box's articulation"
print("REPRO CONFIRMED")
```

### System Information

- **newton**: `1.3.0.dev0`
- **warp**: `1.15.0.dev20260529` (CUDA 12.9, driver 13.1)
- GPU: NVIDIA RTX A6000
- File: `newton/_src/utils/import_usd.py`, function `parse_usd` (guard around the `_add_base_joints_to_floating_bodies(...)` call; `no_articulations`/`has_joints` defined just above).

jpatel3:
## Description

`ModelBuilder.add_usd` auto-wraps each loose rigid body into its own articulation by adding a base joint. That step was gated by a **global** check over the whole parsed range (`import_usd.py`):

```python
no_articulations = UsdPhysics.ObjectType.Articulation not in ret_dict
has_joints       = any(<joint types in ret_dict>)
...
if not (no_articulations and has_joints):
    builder._add_base_joints_to_floating_bodies(...)
```

Because the condition is evaluated over the entire range and applied to all floating bodies, a *single* authored joint that lives under no articulation root — e.g. a leftover `FixedJoint` on an otherwise-static prop — makes `has_joints=True` while `no_articulations=True`, which **silently suppresses base-joint (→ articulation) creation for every unrelated free rigid body in the same `add_usd` call**. Those bodies are then left with no joint and no articulation.

This PR decides per body instead. When the USD authored a joint graph but no articulation root, only bodies that already belong to a joint (as parent *or* child) are skipped; rigid bodies connected to no joint at all still receive a base joint. The authored orphan-joint graph is left untouched, so existing non-articulated-joint behavior (e.g. `test_import_non_articulated_joints`) is preserved.

Closes #3002.

## Checklist

## Test plan

New regression test (fails before this change, passes after):

```
uv run --extra dev --extra sim -m newton.tests -k test_stray_joint_does_not_strip_unrelated_floating_bodies
```

Verified no regressions in the surrounding import suites:

```
uv run --extra dev --extra sim -m newton.tests -k test_import_usd          # 217 passed
uv run --extra dev --extra sim -m newton.tests -k test_import_usd_multi_dof # 3 passed
uv run --extra dev --extra sim -m newton.tests -k test_sites_usd_import     # 5 passed
uv run --extra dev --extra sim -m newton.tests -k test_import_mjcf          # 219 passed
```

## Bug fix

**Steps to reproduce:**

1. Build a USD stage with no `PhysicsArticulationRootAPI`.
2. Add one free-floating rigid body (no joint).
3. Add a separate authored `PhysicsFixedJoint` between two other props.
4. `ModelBuilder().add_usd(stage)` — the free-floating body receives no base joint and no articulation.

**Minimal reproduction:**

```python
import newton

# Unrelated free-floating body (should get a FREE base joint + articulation).
free = UsdGeom.Cube.Define(stage, "/World/FreeBody")
UsdPhysics.RigidBodyAPI.Apply(free.GetPrim())
UsdPhysics.CollisionAPI.Apply(free.GetPrim())

# A stray authored fixed joint between two props, under no articulation root.
a = UsdGeom.Cube.Define(stage, "/World/PropA"); UsdPhysics.RigidBodyAPI.Apply(a.GetPrim()); UsdPhysics.CollisionAPI.Apply(a.GetPrim())
b = UsdGeom.Cube.Define(stage, "/World/PropB"); UsdPhysics.RigidBodyAPI.Apply(b.GetPrim()); UsdPhysics.CollisionAPI.Apply(b.GetPrim())
fj = UsdPhysics.FixedJoint.Define(stage, "/World/StrayFixedJoint")
fj.CreateBody0Rel().SetTargets([a.GetPath()]); fj.CreateBody1Rel().SetTargets([b.GetPath()])

builder.add_usd(stage)

free_idx = builder.body_label.index("/World/FreeBody")
print("FreeBody has a base joint?", free_idx in set(builder.joint_child))
# Before this PR: False  -> after: True
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed USD import so an authored “stray” joint outside any articulation root no longer prevents base (articulation) joint creation for unrelated floating rigid bodies.

* **Tests**
  * Added a regression test covering stray orphan fixed joints: verifies the unrelated free body still gets the expected base joint with valid articulation association, while the orphan joint is imported as fixed with no articulation association (and emits a warning).


adenzler-nvidia:
Thanks for the fix—this looks functionally correct, and the regression test covers the reported failure well. Please trim the added comments and test docstring to the non-obvious rationale. Most currently restate the code or test setup, particularly the six-line orphan-joint explanation and the comments describing `FreeBody` and the expected joint types. A one-line production comment such as `# Preserve authored orphan-joint graphs while articulating unrelated bodies.` is sufficient. In the test, keep only the useful explanation that `CollisionAPI` gives the bodies positive mass and, if needed, that the authored joint must remain orphaned.

jpatel3:
Thanks for the review! Trimmed as suggested:

- Production comment collapsed to a single line: `# Preserve authored orphan-joint graphs while still articulating unrelated bodies (#3002).`
- Test docstring shortened, and the comments that restated the setup/expected joint types removed. Kept the note that `CollisionAPI` gives the bodies positive mass and a one-liner that the authored joint must remain orphaned.

Also rebased onto the latest `main` merge on the branch. Regression test still passes.

adenzler-nvidia (approved):
Thanks!

jcarius-nv (approved):
Approving based on previous approval

PR #3219, opened 2026-06-23
Warn on unsupported USD stage units

eric-heiden:
﻿## Description

USD stages with non-unit `metersPerUnit` or `kilogramsPerUnit` now emit a `UserWarning` as soon as the stage units are parsed. Newton currently does not support importing USD stages authored in non-meter or non-kilogram units, so both unsupported unit declarations are reported directly instead of only surfacing through gravity behavior.

`PhysicsScene` `gravityMagnitude` is still imported as authored instead of being scaled by `metersPerUnit`. Authored mass and inertia values are also imported as authored; non-unit `kilogramsPerUnit` now warns rather than implying conversion support that does not exist.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_scene_gravity_non_unit_linear_unit
uv run --extra dev -m newton.tests -k test_material_density_mass_properties_with_stage_linear_scale
uv run --extra dev -m newton.tests -k test_massapi_authored_mass_with_non_unit_mass_unit_warns
uvx pre-commit run -a
```

Also verified the gravity regression test fails without the gravity fix by temporarily restoring the previous scaled gravity assignment.

## Bug fix

**Steps to reproduce:**

1. Import a USD stage with `metersPerUnit = 0.01` or `kilogramsPerUnit = 0.001`.
2. Add authored physics values such as `PhysicsScene.gravityMagnitude` or `PhysicsMassAPI.mass`.
3. Observe that Newton does not fully convert the stage's non-unit USD units.

**Minimal reproduction:**

```python
import newton

UsdGeom.SetStageMetersPerUnit(stage, 0.01)
UsdPhysics.SetStageKilogramsPerUnit(stage, 0.001)
scene = UsdPhysics.Scene.Define(stage, "/physicsScene")
scene.CreateGravityMagnitudeAttr().Set(12.34)

builder.add_usd(stage)  # warns for both unsupported unit declarations

print(builder.gravity)  # should be -12.34
```

## Bug Fixes
* Updated USD import behavior to validate authored unit metadata:
  * Warns when USD uses non-standard **mass** units (`kilogramsPerUnit` not effectively 1.0).
  * Warns when USD uses non-standard **linear** units (`metersPerUnit` not effectively 1.0).
* **PhysicsScene gravity** is now imported **as authored** and is no longer scaled by the stage’s unit scale (while still respecting gravity enablement).

### DIFF
--- a/newton/_src/utils/import_usd.py
+++ b/newton/_src/utils/import_usd.py
@@ -354,6 +354,12 @@ def parse_usd(
     except Exception as e:
         if verbose:
             print(f"Failed to get linear unit: {e}")
+    if not math.isclose(linear_unit, 1.0):
+        warnings.warn(
+            "USD stages with non-unit linear units are not supported. "
+            f"Set metersPerUnit to 1.0 before import. Found metersPerUnit={linear_unit}.",
+            stacklevel=_external_stacklevel(),
+        )
 
     non_regex_ignore_paths = [path for path in ignore_paths if ".*" not in path]
     ret_dict = UsdPhysics.LoadUsdPhysicsFromRange(stage, [root_path], excludePaths=non_regex_ignore_paths)
@@ -1852,7 +1858,7 @@ def parse_usd(
             print("Found PhysicsScene:", path)
             print("Gravity direction:", scene_desc.gravityDirection)
             print("Gravity magnitude:", scene_desc.gravityMagnitude)
-        builder.gravity = -scene_desc.gravityMagnitude * linear_unit
+        builder.gravity = -scene_desc.gravityMagnitude
 
         # Storing Physics Scene attributes
         physics_scene_prim = stage.GetPrimAtPath(path)
--- a/newton/_src/utils/import_usd.py
+++ b/newton/_src/utils/import_usd.py
@@ -347,6 +347,12 @@ def parse_usd(
     except Exception as e:
         if verbose:
             print(f"Failed to get mass unit: {e}")
+    if not math.isclose(mass_unit, 1.0):
+        warnings.warn(
+            "USD stages with non-unit mass units are not supported. "
+            f"Set kilogramsPerUnit to 1.0 before import. Found kilogramsPerUnit={mass_unit}.",
+            stacklevel=_external_stacklevel(),
+        )
     linear_unit = 1.0
     try:
         if UsdGeom.StageHasAuthoredMetersPerUnit(stage):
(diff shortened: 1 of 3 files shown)


newton/_src/utils/import_usd.py
```diff
@@ -354,6 +354,12 @@ class PhysicsMaterial:
     except Exception as e:
         if verbose:
             print(f"Failed to get linear unit: {e}")
+    if not math.isclose(linear_unit, 1.0):
+        warnings.warn(
```
andrewkaufman, line 364: @eric-heiden should we be doing the same for `mass_unit` above? I suspect we are not properly scaling things by non-kg mass_unit either

eric-heiden: Good call. Yes, `mass_unit` is only parsed and returned today; I do not see importer handling that converts non-kg mass, density, or inertia values. I added the matching unsupported-unit warning for non-unit `kilogramsPerUnit`, plus a regression test that locks in the current authored-mass-as-authored behavior.

newton/tests/test_import_usd.py
```diff
@@ -6770,6 +6803,28 @@ def test_scene_gravity_enabled_parsing(self):
         # Gravity should be disabled (zero)
         self.assertEqual(builder2.gravity, 0.0)
 
+    @unittest.skipUnless(USD_AVAILABLE, "Requires usd-core")
+    def test_scene_gravity_non_unit_linear_unit(self):
+        """Test non-unit linear unit warning and unscaled PhysicsScene gravity."""
+        from pxr import Usd, UsdGeom, UsdPhysics
+
+        stage = Usd.Stage.CreateInMemory()
+        UsdGeom.SetStageUpAxis(stage, UsdGeom.Tokens.z)
```
andrewkaufman, line 6813: Just realizing we probably also need one for non-Z up axis? Or do we handle y-up properly?

eric-heiden: Y-up is handled by the existing stage up-axis path. By default we rotate the imported stage so its up axis aligns with the builder up axis, and with apply_up_axis_from_stage=True the builder adopts the stage up axis. I also checked a Y-up stage locally: default import finalizes gravity along -Z, while apply_up_axis_from_stage=True finalizes it along -Y. So I don’t think we need a non-Z-up warning here.

andrewkaufman (approved):

PR #3221, opened 2026-06-23
Bump jupyter-server floor

jcarius-nv:
## Description

Bump the `notebook` extra's `jupyter-server` floor to `>=2.20.0` and refresh `uv.lock` to `jupyter-server` 2.20.0. This avoids the vulnerable `<=2.19.0` range reported by Dependabot alert 37 for CVE-2026-44727 / GHSA-fcw5-x6j4-ccmp.

Dependabot should clear the alert after the branch is merged and rescanned.

## Checklist

## Test plan

```bash
uv --cache-dir /tmp/uv-cache lock --check
uv --cache-dir /tmp/uv-cache run --extra dev --extra notebook -m newton.tests -p test_viewer_layers.py
uv --cache-dir /tmp/uv-cache run --extra dev --extra notebook -m newton.tests -p test_viewer_rerun_init_args.py
uv --cache-dir /tmp/uv-cache run --extra notebook python -c "import jupyter_server; print(jupyter_server.__version__)"
UV_CACHE_DIR=/tmp/uv-cache UV_TOOL_DIR=/tmp/uv-tools PRE_COMMIT_HOME=/tmp/pre-commit-cache uvx --python 3.12 pre-commit run -a
```

Also ran an independent subagent critical review; it found no actionable issues.

## Bug fix

**Steps to reproduce:**

1. Inspect Dependabot alert 37 or the previous `uv.lock`.
2. Observe `jupyter-server` locked at `2.19.0`, which is in the vulnerable `<=2.19.0` range.

**Minimal reproduction:**

```bash
rg -n 'name = "jupyter-server"|version = "2.19.0"' uv.lock
```

## Summary by CodeRabbit

* **Chores**
  * Updated the minimum required `jupyter-server` version to **>= 2.20.0** for the notebook extra, pulling in a security fix (CVE-2026-44727).
  * Refreshed the changelog entry to reflect the updated requirement.


pyproject.toml
```diff
@@ -129,6 +129,7 @@ docs = [
 notebook = [
     "newton[examples]",
     "rerun-sdk[notebook]>=0.27.1; python_version < '3.14'",
+    "jupyter-server>=2.20.0",  # >=2.20.0 for GHSA-fcw5-x6j4-ccmp / CVE-2026-44727
```
shi-eric: For bumps like this, a uv.lock is sufficient. We shouldn’t raise the floor in pyproject.toml unless a version is geniuely incompatible.

jcarius-nv: I see, thanks for pointing this out.

So the reasoning is that the lockfile pins the version we use as developers but for consumers of the wheel we leave it to the package managers of the user to take the latest compatible one?

shi-eric: yeah, basically leave it to other libraries or their own projects to constrain the dependencies if something else has a more opinionated position.

jcarius-nv: Good point. I updated the branch so this is lockfile-only now: the `pyproject.toml` floor and changelog entry are both removed, and the final diff is just `uv.lock` moving `jupyter-server` from 2.19.0 to 2.20.0.

shi-eric (approved):

PR #3227, opened 2026-06-24
Reduce usage of pytorch and add warp-nn dependency

nvtw:
## Description
Add ONNX/Warp-NN inference for neural MLP/LSTM controllers and robot policy examples so exported `.onnx` policies can run without PyTorch at inference time. This moves policy execution onto Warp-backed buffers, makes neural controllers graphable when using ONNX, validates ONNX metadata and policy I/O shapes, and updates examples to use pre-exported ONNX assets. Legacy TorchScript `.pt`/`.pth` checkpoints remain loadable during a deprecation window with guidance to export to ONNX.
## Checklist
## Test plan
Not run while preparing this MR text. Relevant coverage includes actuator controller tests for ONNX metadata, invalid metadata, fixed batch shapes, LSTM state updates, and legacy TorchScript compatibility, plus example tests gated on ONNX/Warp-NN availability:
```bash
uv run --extra dev -m newton.tests -k test_examples
```
## New feature / API change
Neural controllers can now load `.onnx` checkpoints directly, and RL examples install ONNX inference support through `newton[examples]` via the new `newton[onnx]` extra:
```python
import newton

controller = newton.ControllerNeuralMLP(model_path="policy.onnx")
```

## Summary by CodeRabbit

* **New Features**
  * Added ONNX/Warp-NN execution for neural controllers (MLP/LSTM) and migrated robot examples to run pre-exported ONNX policies without PyTorch.
  * Added installation extras (`onnx`, included via `examples`) plus ONNX policy I/O shape validation.
* **Deprecations**
  * Deprecated TorchScript `.pt/.pth` neural-controller checkpoints in favor of ONNX (accepted during a transition window).
* **Documentation**
  * Updated installation and controller concepts; removed PyTorch/CUDA-specific example guidance.
* **Tests**
  * Refreshed fixtures and gating to prefer ONNX; added coverage for corrupted/invalid metadata and legacy TorchScript regression tests.


docs/guide/development.rst
```diff
@@ -174,10 +174,9 @@ Pass ``--help`` to either run method below to see all available flags.
             # run tests
             python -m newton.tests
             
-Most tests run when the ``dev`` extras are installed. The tests using PyTorch
-to run inference on an RL policy are skipped if the ``torch`` dependency is
-not installed. In order to run these tests, include the ``torch-cu12`` or
-``torch-cu13`` extras matching your NVIDIA driver's CUDA support:
+Most tests run when the ``dev`` extras are installed. RL policy inference tests
+use ONNX checkpoints through Warp-NN and are included in ``dev`` via the
+``examples`` extra:
```
adenzler-nvidia: I don't think we need to state this explicitly, dev just downloads everything?

docs/concepts/actuators.rst
```diff
@@ -221,10 +223,10 @@ Controllers
...
 * :class:`ControllerPID` — proportional-integral-derivative control law
   (stateful: integral accumulator with anti-windup clamp).
-* :class:`ControllerNeuralMLP` — MLP neural-network controller (requires
-  PyTorch, stateful: position/velocity history buffers).
-* :class:`ControllerNeuralLSTM` — LSTM neural-network controller (requires
-  PyTorch, stateful: hidden/cell state).
+* :class:`ControllerNeuralMLP` — MLP neural-network controller (ONNX/Warp-NN,
+  stateful: position/velocity history buffers).
+* :class:`ControllerNeuralLSTM` — LSTM neural-network controller (ONNX/Warp-NN,
+  stateful: hidden/cell state).
```
adenzler-nvidia: not sure if we need the bracket annotations here if the dependency is pulled anyway

docs/guide/development.rst
```diff
@@ -188,27 +187,22 @@ not installed. In order to run these tests, include the ``torch-cu12`` or
...
 
-    The ``torch-cu12`` extra requires PyTorch built against CUDA 12.8. If your
-    driver only supports CUDA 12.4 or 12.5 (check with ``nvidia-smi``), install
-    PyTorch 2.6.0 manually instead of using the ``torch-cu12`` extra:
-
-    .. code-block:: console
-
-        pip install torch==2.6.0 --extra-index-url https://download.pytorch.org/whl/cu124
+    The ``torch-cu12`` and ``torch-cu13`` extras remain available for workflows
+    that explicitly need PyTorch, such as training or exporting policies.
```
adenzler-nvidia: same here, I think we can just drop this.

docs/guide/installation.rst
```diff
@@ -122,26 +122,19 @@ After installing Newton with the ``examples`` extra, launch the default
 
     python -m newton.examples
 
-Run an example that runs RL policy inference. Choose the extra matching your
-NVIDIA driver's CUDA support (``torch-cu12`` for CUDA 12.x, ``torch-cu13`` for
-CUDA 13.x) and the corresponding pytorch wheel (e.g, ``128`` for CUDA 12.8); run ``nvidia-smi``
-to check the supported CUDA version (shown in the top-right corner of the output):
+Run an example that performs RL policy inference. The ``examples`` extra
+includes ``newton[onnx]``, which installs Warp-NN's ONNX runtime and the ONNX
+parser; PyTorch is not required for these pre-exported policies:
```
adenzler-nvidia: drop the "pytorch is not required". Is there a reason we have [onnx], it could also just be part of dev/examples?

docs/guide/installation.rst
```diff
@@ -122,26 +122,19 @@ After installing Newton with the ``examples`` extra, launch the default
...
-    The ``torch-cu12`` extra installs PyTorch built against CUDA 12.8. If your
-    driver only supports CUDA 12.4 or 12.5 (check with ``nvidia-smi``), you
-    need to install PyTorch 2.6.0 manually instead:
-
-    .. code-block:: console
-
-        pip install "newton[examples]"
-        pip install torch==2.6.0 --extra-index-url https://download.pytorch.org/whl/cu124
+    The ``torch-cu12`` and ``torch-cu13`` extras remain available for workflows
+    that explicitly use PyTorch, such as training or exporting policies.
```
adenzler-nvidia: drop this

pyproject.toml
```diff
@@ -40,6 +40,13 @@ sim = [
     "mujoco~=3.8.0",        # MuJoCo physics engine
 ]
 
+# Optional ONNX policy inference for neural actuators and RL policy examples.
+onnx = [
+    # OnnxRuntime currently lives on warp-nn's develop branch; main does not
+    # expose warp_nn.runtime yet.
+    "warp-nn[onnx] @ git+https://github.com/NVIDIA/warp-nn.git@develop",
+]
+
```
adenzler-nvidia, line 47: do you have an ETA on when we will get a warp-nn release such that we don't need to rely on a source checkout?

newton/_src/actuators/utils.py
```diff
@@ -4,19 +4,137 @@
...
+    runtime = OnnxRuntime(path, device=device, batch_size=batch_size, input_batch_axes=input_batch_axes)
+    return runtime, metadata
+
+
+def _extract_metadata(model) -> dict[str, Any]:
+    props = {p.key: p.value for p in model.metadata_props}
+    if "metadata" in props:
+        try:
+            return json.loads(props["metadata"])
+        except json.JSONDecodeError:
```
adenzler-nvidia: If a `metadata` property exists but its JSON is corrupt/truncated, this `except JSONDecodeError: pass` falls through to the per-key loop below, which re-fails on the same blob and stores the raw JSON string under `parsed["metadata"]`. The controller then sees none of its expected keys — for the MLP that means it silently runs with all default scales (`pos/vel/effort=1.0`, `input_idx=[0]`): wrong behavior, no error.

Suggest: when a `metadata` key is present, let a decode error propagate (with the path) rather than swallowing it; only use the per-key fallback when there is no single `metadata` property.

newton/_src/actuators/controllers/controller_neural_mlp.py
```diff
@@ -98,35 +181,61 @@ def __init__(self, model_path: str):
...
+        feat = 2 * len(self.input_idx)
+        self._net_input = wp.zeros((num_actuators, feat), dtype=wp.float32, device=device)
+        self._pos_error = wp.zeros(num_actuators, dtype=wp.float32, device=device)
+        self._vel_error = wp.zeros(num_actuators, dtype=wp.float32, device=device)
+        self._input_idx_wp = wp.array(self.input_idx, dtype=wp.int32, device=device)
+
+        if self._net_output_name not in runtime._shapes:
+            runtime({self._net_input_name: self._net_input})
+
+        out_shape = runtime._shapes[self._net_output_name]
```
adenzler-nvidia: This reads `runtime._shapes`, a private attribute of warp-nn's `OnnxRuntime` (only `input_names` / `output_names` are public). The same private access recurs in `controller_neural_lstm.finalize` and in `example_robot_policy.load_policy_and_setup_arrays`. With warp-nn pinned to `@develop`, an upstream rename of `_shapes` would break this in several places with no type-checker warning.

Suggest centralizing shape access in one helper (e.g. return shapes alongside the runtime), or requesting a public shape accessor from warp-nn.

newton/_src/actuators/controllers/controller_neural_lstm.py
```diff
@@ -73,72 +114,117 @@ def resolve_arguments(cls, args: dict[str, Any]) -> dict[str, Any]:
...
 
-        self._torch_device = torch.device(f"cuda:{device.ordinal}" if device.is_cuda else "cpu")
-        self.network = self.network.to(self._torch_device)
-        self._torch_sequential_indices = torch.arange(num_actuators, dtype=torch.long, device=self._torch_device)
+    def finalize(self, device: wp.Device, num_actuators: int) -> None:
+        self._device = device
+        self._num_actuators = num_actuators
+
+        if self._legacy_adapter is not None:
+            self._legacy_adapter._device = device
```
adenzler-nvidia: `finalize` reaches into three private fields of the adapter here (`_device`, `_torch_device`, and `_model` via `.to(...)`). That's the adapter's encapsulation leaking back into the controller.

Cleaner: give `_LegacyLstmTorchAdapter` a `to(device)` method that owns this dance, or defer adapter construction to `finalize` (where the device is known) instead of constructing in `__init__` and rebinding here.

newton/tests/test_actuators.py
```diff
@@ -404,6 +516,115 @@ def test_metadata_scales(self):
...
+                y, hc_new = self.lstm(x, hc)
+                effort = self.fc(y[:, -1, :])
+                return effort, hc_new
+
+        model = _LegacyLSTM(hidden_size).eval()
+        scripted = torch.jit.script(model)
+        extra_files = {"metadata.json": json.dumps(metadata or {})}
+        scripted.save(path, _extra_files=extra_files)
+
+    def test_load_emits_deprecation_warning(self):
```
adenzler-nvidia: This test asserts that a `DeprecationWarning` fires. Prefer testing behavior and silencing the warning in setup — the sibling legacy tests (`test_finalize_and_compute`, `test_synthesizes_metadata_from_torch_module`) already cover the `.pt` path, so this assertion is redundant and can be dropped.

newton/_src/actuators/controllers/controller_neural_mlp.py
```diff
@@ -98,35 +181,61 @@ def __init__(self, model_path: str):
             raise ValueError(f"input_idx must contain non-negative integers; got {self.input_idx}")
         self.history_length = max(self.input_idx) + 1
 
-        self.pos_scale = metadata.get("pos_scale", 1.0)
-        self.vel_scale = metadata.get("vel_scale", 1.0)
-        self.effort_scale = metadata.get("effort_scale", metadata.get("torque_scale", 1.0))
+        self.pos_scale = float(metadata.get("pos_scale", 1.0))
+        self.vel_scale = float(metadata.get("vel_scale", 1.0))
+        self.effort_scale = float(metadata.get("effort_scale", metadata.get("torque_scale", 1.0)))
```
adenzler-nvidia: `float(metadata.get("effort_scale", metadata.get("torque_scale", 1.0)))` (same pattern in `controller_neural_lstm`): an explicit `effort_scale: 0.0` is silently accepted and yields a dead actuator (zero effort every step), and `effort_scale: null` makes `float(None)` raise an opaque `TypeError` far from the cause.

Suggest parsing each scale through a small helper that rejects non-finite values and re-raises with the key name + `model_path` on failure.

newton/examples/robot/example_robot_policy.py
```diff
@@ -39,145 +38,144 @@ class RobotConfig:
...
-    q_w = q[..., 3]  # w component is at index 3 for XYZW format
-    q_vec = q[..., :3]  # xyz components are at indices 0, 1, 2
-    a = v * (2.0 * q_w**2 - 1.0).unsqueeze(-1)
-    b = torch.cross(q_vec, v, dim=-1) * q_w.unsqueeze(-1) * 2.0
-    # for two-dimensional tensors, bmm is faster than einsum
-    if q_vec.dim() == 2:
-        c = q_vec * torch.bmm(q_vec.view(q.shape[0], 1, 3), v.view(q.shape[0], 3, 1)).squeeze(-1) * 2.0
+def _load_onnx_runtime(path: str, device: wp.Device):
+    try:
+        from warp_nn.runtime import OnnxRuntime  # noqa: PLC0415
```
adenzler-nvidia: This graceful-fallback wrapper isn't really needed in the examples. Example discovery lists modules by filename (`get_examples()` uses `os.listdir`) and only imports the one you actually run, and an ONNX example can't do anything useful without ONNX — so a missing dependency failing loudly is fine here. The old torch examples just did a top-level `import torch` and let it fail; following that pattern (a top-level `from warp_nn.runtime import OnnxRuntime`) lets you drop this helper.

It's also currently triplicated — verbatim in both robot examples, plus `_require_warp_nn_runtime` in `_src/actuators/utils.py`. (The deferred/guarded import does still make sense in the library controllers, which are pulled in by `import newton`.)

adenzler-nvidia (commented):
A few structural / error-handling notes on the ONNX migration (the torch→Warp kernel translations themselves check out as faithful). The `_extract_metadata` fallback is the only load-bearing item — it can silently mis-configure a controller; the rest are cleanups.

adenzler-nvidia:
@jvonmuralt, could you weigh in on the intended direction here? We should decide whether removing PyTorch from the actuator implementation is actually desirable, rather than treating it as an implementation detail of the ONNX work.

This PR adds ONNX/Warp-NN support, but it also effectively replaces the existing Torch neural-controller path: the public ControllerNeuralMLP/ControllerNeuralLSTM state fields change from Torch tensors to Warp arrays, the public network attribute disappears, accepted output shapes narrow, and legacy .pt/.pth execution now performs synchronous host round-trips on CUDA. Those controllers shipped in 1.3, so deprecating the checkpoint format alone does not cover these interface and behavior changes.

I think we need an explicit product/API decision before merging:

- If Torch remains a supported actuator use case, ONNX should be an additive backend and .pt/.pth should continue through the existing Torch implementation.
- If the goal is to remove Torch from actuators, the existing Torch behavior and interface should still remain intact for the deprecation window, with a documented removal release and a tested conversion tool.

At the moment there is no converter in this PR, only guidance to call torch.onnx.export directly. Conversion is not entirely mechanical because the exported input/state signatures, dynamic axes, Newton metadata, and Warp-NN operator restrictions all need to match.

My preference would be to keep two explicit backend paths for now: preserve the current Torch implementation for .pt/.pth and use the new Warp-NN path for .onnx. That keeps ONNX additive, avoids hidden breaking changes, and gives us time to decide whether the Torch backend should eventually be removed.

jvonmuralt:
I'd prefer to keep onnx additive for now instead of replacing the torch path as Alain suggests. 

I'd also wait with the deprecation warning. I think we would first need to verify if we can successfully convert existing lab .pt files to onnx and compare actual performance improvement before making decision to not support pt.

nvtw:
> I'd prefer to keep onnx additive for now instead of replacing the torch path as Alain suggests.
> 
> I'd also wait with the deprecation warning. I think we would first need to verify if we can successfully convert existing lab .pt files to onnx and compare actual performance improvement before making decision to not support pt.

Ok, I'll adjust the MR accordingly and will keep pytorch supported. The goal is not speedup but avoiding the 2.5GB pytorch dependency whenever possible, so as long as we do not get a slow down, the onnx path should be a god alternative.

docs/concepts/actuators.rst
```diff
@@ -201,7 +201,9 @@ Differentiability and Graph Capture
 Whether an actuator supports differentiability and CUDA graph capture depends on
 its controller.  :class:`ControllerPD` and :class:`ControllerPID` are fully
 graphable.  Neural-network controllers (:class:`ControllerNeuralMLP`,
-:class:`ControllerNeuralLSTM`) require PyTorch and are not graphable due to
+:class:`ControllerNeuralLSTM`) load ONNX checkpoints through Warp-NN's
+Warp-backed runtime and are graphable.  TorchScript checkpoints (``.pt`` /
+``.pth``) use the Torch backend, require PyTorch, and are not graphable due to
```
adenzler-nvidia: I think this should be formulated as an either/or now, right? both options are possible?

newton/_src/utils/download_assets.py
```diff
@@ -20,7 +20,7 @@
 # commit.  Update these SHAs when assets change upstream and the new versions
 # have been validated against Newton's test suite.
 NEWTON_ASSETS_URL = "https://github.com/newton-physics/newton-assets.git"
-NEWTON_ASSETS_REF = "8e8df07d2e4829442d3d3d3aeecee1857f9951d7"
+NEWTON_ASSETS_REF = "d69b2e04bc1fc246c415f0549e5f02d8aae1ef31"
```
adenzler-nvidia: Is advancing the asset pin through `d69b2e04` intentional here? The ONNX policies land in the preceding `261cd1f4` commit, while `d69b2e04` also adds articulation roots and changes joint membership/topology for the existing Disney Research models. That changes model parsing and simulation behavior outside this PR's stated scope. If those changes are not intended, could we pin to `261cd1f429619d8ef4f546bd788ab9dea906b5e1` instead? If they are intended, should the behavior change be covered and documented separately?

newton/tests/test_actuators.py
```diff
@@ -198,8 +291,9 @@ def _f(vals):
             self.assertAlmostEqual(forces.numpy()[0], expected, places=4, msg=f"step {step_i}")
 
 
+@unittest.skipUnless(_HAS_ONNX and _HAS_WARP_NN, "onnx or warp-nn not installed")
```
adenzler-nvidia, line 294: These are ONNX-only tests now, but `setUp()` still imports Torch and skips without it. In a `newton[dev]` environment without Torch, all new MLP controller tests skip; the duplicated setup in the LSTM and USD-parsing classes also skips those ONNX tests. This leaves the advertised no-PyTorch path without regression coverage in that configuration. Please remove the Torch setup from the three ONNX test classes and keep Torch gating only in the legacy checkpoint classes.

adenzler-nvidia (commented):
Two compatibility questions on the latest ONNX changes.

newton/examples/robot/example_robot_policy.py
```diff
@@ -39,145 +40,136 @@ class RobotConfig:
...
+    obs[0, 5] = avel_b[2]
+    obs[0, 6] = grav_b[0]
+    obs[0, 7] = grav_b[1]
+    obs[0, 8] = grav_b[2]
+    obs[0, 9] = command[0]
+    obs[0, 10] = command[1]
+    obs[0, 11] = command[2]
+
+    for k in range(num_dofs):
+        idx = physx_to_mjc_idx[k]
```
adenzler-nvidia, line 109: `find_physx_mjwarp_mapping` silently skips joints missing from either name list, and this kernel (like `_build_joint_target_q_kernel`) indexes the mapping for all `k < num_dofs` without a bounds check — a short mapping reads past the end of the array, which is unchecked in release Warp and yields garbage observations instead of an error. The old torch path failed loudly at `index_select` in that case. Suggest asserting both mappings have length `num_dofs` right after `find_physx_mjwarp_mapping`.

newton/_src/actuators/controllers/controller_neural_mlp.py
```diff
@@ -9,30 +9,123 @@
...
-    Uses a pre-trained MLP to compute joint effort from position error
-    and velocity error history.
-
-    The network receives concatenated, scaled position-error and
-    velocity-error history as input.  The output is multiplied by
+    Uses a pre-trained MLP to compute joint effort from concatenated, scaled
+    position-error and velocity-error history. The output is multiplied by
     ``effort_scale`` to convert from network units to physical effort
-    [N or N·m].  All three scale factors default to ``1.0`` (no scaling).
+    [N or N*m].
```
adenzler-nvidia: `[N or N*m]` → `[N or N·m]` to match the unit notation elsewhere (e.g. `controller_pid.py`, `delay.py`).

pyproject.toml
nvtw: It's now using the 0.3 release that just came out

adenzler-nvidia (approved):
LGTM

PR #3238, opened 2026-06-24
Kamino: Fused Conjugate Residual Kernel

camevor:
## Description

Work in progress: Adds a single-kernel sparse Conjugate Residual solver (`CRF`) for Kamino's matrix-free
Delassus operator, plus two supporting fixes in the PADMM and USD-import paths.

- **`ConjugateResidualSolverFused` (`CRF`)** — runs the entire CR iteration for
  `A = P·J·M⁻¹·Jᵀ·P + diag(eta)` in one Warp tile-block per world, reading the raw constraint
  Jacobian directly and applying `P` and `M⁻¹` on the fly. No `P·J·M⁻¹` / `(P·J)ᵀ` value copies
  are materialized — the only auxiliary data are `int32` index arrays. The matrix-free operator
  now skips assembling those copies when the solver sets `uses_raw_jacobian`.

- **PADMM adaptive-penalty graph fix + inexact-ADMM scheduling** — splits the operator's
  dirty-flag handling so an eta-only change refreshes the combined regularization without
  rebuilding/re-sorting the Jacobian index (wasted work per iteration, and unsafe inside
  `wp.capture_while`). Adds optional scheduling of the inner linear-solver tolerance from the
  primal residual (`linear_solver_tolerance_ratio`).

- **USD importer** — stop parsing `PhysicsArticulationRootAPI`. Kamino is a maximal-coordinate,
  constraint-based solver and needs no spanning-tree articulation; honoring it routed loop-closure
  models into the tree-ordering path and failed with a spurious cycle error.

## Checklist

## Test plan

```
uv run --no-sync --extra dev -m newton.tests -k test_linalg_conjugate_fused
```

Covers the fused kernel against numpy's dense solve (with/without preconditioner) and an
end-to-end PADMM contact solve checking `CRF` matches the multi-launch `CR`.

## New feature / API change

```python
config.solver.constraints.linear_solver_type = "CRF"  # single-kernel sparse CR
config.solver.padmm.linear_solver_tolerance_ratio = 0.1  # inexact-ADMM (0 = fixed inner tol)
```

## Summary by CodeRabbit

* **New Features**
  * Added the **CRF** matrix-free sparse linear solver for Kamino PADMM (single-kernel fused Conjugate Residual per world).
  * You can now select **`linear_solver_type="CRF"`**, including optional inexact-ADMM inner-tolerance scheduling.
  * Extended the DR legs example with **`--sparse`** and **`--asset`** flags.

* **Bug Fixes**
  * Improved sparse/adaptive-penalty regularization and linear-solver tolerance propagation for accelerated/CUDA-graph runs.

* **Tests**
  * Added CUDA-focused unit tests and PADMM integration tests validating **CRF vs CR**, including preconditioning.


newton/_src/solvers/kamino/examples/sim/example_sim_dr_legs.py
```diff
@@ -230,10 +232,16 @@ def __init__(
...
         config.solver.padmm.eta = 1e-5
-        config.solver.padmm.rho_0 = 0.02  # try 0.02 for Balanced update
-        config.solver.padmm.rho_min = 0.05
-        config.solver.padmm.penalty_update_method = "fixed"  # try "balanced"
         config.solver.padmm.use_acceleration = True
+        # CRF converges only with the adaptive "balanced" penalty; rho_min/alpha/tau are no-ops
+        # under the fixed penalty the direct solvers use.
+        crf = linear_solver == "CRF"
+        config.solver.padmm.penalty_update_method = "balanced" if crf else "fixed"
+        config.solver.padmm.rho_0 = 0.5 if crf else 0.02
```
chschuma-disney: Minor point, but these CRF-specific settings don't work for me, and the usual values seem to work fine.

Guirec-Maloisel:
Looks very promising from testing! I tried this on Iron Man (version with fixed hands) on my machine (single-world, tracking an animation), I'm getting 9.5 fps with CR and 21 fps with the new CRF solver!

For reference I'm getting 20 fps with the LLTBRCM solver (with sparse Jacobian but dense solve), so we're marginally better. I'd expect a larger difference on the model with the detailed hand mechanisms.

newton/_src/solvers/kamino/_src/utils/io/usd.py
```diff
@@ -1905,30 +1905,18 @@ def import_from(
         # Articulations
         ###
 
-        # Construct a list of articulation root bodies to create FREE
-        # joints if not already defined explicitly in the USD file
+        # Articulation roots are intentionally not parsed. Kamino is a maximal-coordinate,
```
chschuma-disney: I think it might make sense to put this into its own PR. I have another PR in the pipeline (though blocked by another issue) that ran into the same problem. I fixed it differently (parsing `excludeFromArticulation` and building the articulation without that), but you're right that we don't really need to parse articulations in the first place.

chschuma-disney (commented):
Fantastic work. I've seen performance similar to what Guirec reported, so obviously very exciting. I still have to go more into the details on the engineering, but I'm not seeing any visual discrepancies compared to CR.

newton/_src/solvers/kamino/examples/sim/example_sim_dr_legs.py
camevor: Thanks, yes these often lead to tiny rho, causing the dual residual to become very small and seemingly converge, but with unphysical solutions.

newton/_src/solvers/kamino/_src/utils/io/usd.py
camevor: Agreed, I'll open a PR with this. I assume we'll stop maintaining the custom importer mid-term, so this seems like a viable stop-gap.

newton/_src/solvers/kamino/_src/dynamics/delassus.py
```diff
@@ -1378,36 +1393,42 @@ def finalize(
...
-        # preconditioning directly to the Jacobian.
-        if self._col_major_jacobian is None and self._transpose_op_matrix is None:
-            self._transpose_op_matrix = copy.copy(jacobians._J_cts.bsm)
-            self._transpose_op_matrix.nzb_values = wp.empty_like(self.constraint_jacobian.nzb_values)
+        # Solvers that read the raw constraint Jacobian directly (e.g. the fused single-kernel CR)
+        # set ``uses_raw_jacobian = True``; for those we skip assembling the row-major P·J·M⁻¹ copy
+        # and the column-major transpose copy entirely, avoiding the matrix-value duplication.
+        self._assemble = not bool(getattr(solver, "uses_raw_jacobian", False))
+
+        if self._assemble:
```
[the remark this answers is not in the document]
camevor: Fixed in 2e7c76bc0

newton/_src/solvers/kamino/_src/dynamics/delassus.py
```diff
@@ -1251,6 +1251,21 @@ def __init__(
...
+        self._assemble: bool = True
+
+        # Dirty flags for raw-Jacobian solvers (e.g. the fused CR), cleared by the solver once it
+        # has acted on them. ``_raw_jacobian_dirty`` means the Jacobian *structure* changed and the
+        # per-step index structures (row index + segmented transpose sort) must be rebuilt.
+        # ``_regularization_dirty`` means only the diagonal regularization (eta) changed -- the
+        # sparsity is unchanged, so the solver refreshes its combined-regularization copy but skips
+        # the index rebuild. Keeping these separate avoids re-sorting on eta-only updates (adaptive
+        # penalty, or the transient set_regularization(None)/restore around info collection).
+        self._raw_jacobian_dirty: bool = True
```
Guirec-Maloisel: Possibly subjective, but could we use something else than "dirty" as a suffix? e.g. "_needs_update" (would also be consistent with the existing flag)

newton/_src/solvers/kamino/_src/dynamics/delassus.py
```diff
@@ -1251,6 +1251,21 @@ def __init__(
         # Flag to indicate that the operator needs an update to its data structure
         self._needs_update: bool = False
 
+        # Whether to assemble the row-major (P·J·M⁻¹) and column-major (P·J)ᵀ Jacobian copies used
+        # by the multi-launch matvec. Set False in finalize() when the solver reads the raw
+        # Jacobian directly (e.g. the fused single-kernel CR), to avoid duplicating matrix values.
```
Guirec-Maloisel: Comment could be made a bit more compact, e.g. "Flag set automatically based on solver in finalize(), indicating whether the row-major (P·J·M⁻¹) and column-major (P·J)ᵀ Jacobian copies are required"

And the flag name could be a bit more specific, e.g. "_assemble_preconditioned_jacobians"

newton/_src/solvers/kamino/_src/dynamics/delassus.py
```diff
@@ -1412,6 +1433,18 @@ def set_needs_update(self):
         Flags the operator as needing to update its data structure.
         """
         self._needs_update = True
+        self._raw_jacobian_dirty = True
+
+    def set_regularization_needs_update(self):
+        """Flag that only the diagonal regularization (eta) changed, not the Jacobian structure.
+
+        Used by the PADMM proximal/adaptive-penalty update: it rewrites ``_eta`` but leaves the
```
Guirec-Maloisel: This second paragraph could be removed (redundant with the comment explaining the "dirty" flags above).

newton/_src/solvers/kamino/_src/dynamics/delassus.py
```diff
@@ -1547,7 +1585,12 @@ def set_regularization(self, eta: wp.array[wp.float32] | None):
                 Shape of ``(sum_of_max_total_cts,)``.
         """
         self._eta = eta
-        self.set_needs_update()
+        # Regularization-only change: the Jacobian sparsity is unchanged, so flag the lightweight
```
Guirec-Maloisel: Same here, this comment is a bit redundant.

newton/_src/solvers/kamino/_src/linalg/linear.py
```diff
@@ -955,23 +965,327 @@ def _update_preconditioner(self):
...
+
+    # Signals the matrix-free operator to skip assembling its P·J·M⁻¹ / column-major copies, since
+    # this solver consumes the raw Jacobian directly (see BlockSparseMatrixFreeDelassusOperator).
+    uses_raw_jacobian: bool = True
+
+    def __init__(self, block_dim: int = 128, **kwargs: dict[str, Any]):
+        self._delassus_op = None
+        if int(block_dim) <= 0:
+            raise ValueError("ConjugateResidualSolverFused requires block_dim > 0.")
+        # Threads per block (one block solves one world). 128 (four warps) is the measured sweet
```
Guirec-Maloisel: This comment could be made less specific (should probably not mention DR Legs here), while still pointing out the same trade-off.

newton/_src/solvers/kamino/_src/linalg/linear.py
```diff
@@ -955,23 +965,327 @@ def _update_preconditioner(self):
         )
 
 
+class ConjugateResidualSolverFused(IterativeSolver):
```
Guirec-Maloisel: It would be nice to add type annotations to this class like other solvers (e.g. IterativeSolver[ScalarType, IndexType] on this line, and then adding bracket types to the wp.arrays as function arguments below)

Guirec-Maloisel: Actually it's probably IterativeSolver[wp.float32, wp.int32] as it looks like the kernel in conjugate_fused.py is directly specified for these concrete types rather than generics.

newton/_src/solvers/kamino/_src/linalg/linear.py
```diff
@@ -955,23 +965,327 @@ def _update_preconditioner(self):
...
+                outputs=[op._combined_regularization],
+                device=device,
+            )
+        return op._combined_regularization
+
+    def prepare_step(self) -> None:
+        """Rebuild the per-step index structures and combined regularization, if the operator was
+        marked changed since the last build.
+
+        The index structures and combined regularization depend only on the Jacobian sparsity and
```
Guirec-Maloisel, line 1151: I'm a bit confused here about the statement that the index structures etc are rebuilt outside of the CUDA graph, usually we capture the entire time step (or even set of multiple simulation steps per viewer frame) in the graph, including any precomputations in PADMM before its internal loop...

chschuma-disney (approved):

newton/_src/solvers/kamino/_src/linalg/linear.py
chschuma-disney: I revised the comment, should hopefully be clearer now.

newton/_src/solvers/kamino/_src/linalg/linear.py
chschuma-disney: Updated

rubengrandia (approved):

PR #3240, opened 2026-06-25
Fix TetMesh._infer_frequency for length-1 attribute arrays

### LINKED ISSUE #3228: [BUG] USD Parse for tetMesh can not handle unknown attribute frequency.
### Bug Description

I am looking at `_infer_frequency in geometry/types.py.` It seems like is missing the ablility to infer constant attribute frequencies. If the attribute array has length of 1, the function will error out, instead of return `Model.AttributeFrequency.ONCE `.
I am encountering this error when trying to enable deformable simulation with newton in isaac sim. PhysX will add a `deformablePose:default:omniphysics:purposes` custom attribute on the tet mesh prim. Though the attribute is irrelevant to newton's simulation, the usd parser will fail from the error mentioned above, because the custom attribute has a length of 1 and Newton can not infer its frequency from its length.

there is also `physxVolumeDeformableSim:simMeshHexCrc` that has a length of 16. Newton should probably ignore unknown attributes with undetermined frequency that come from other places.

### Reproduction Script

```python
To reproduce:

1. Launch isaac-sim.newton.sh 
2. open Window->Examples->Physics Examples
3. load Physics Demo Scenes->Deformables->Volume Deformable
4. switch simulator to PhysX, and run the simulation once. (this will populate the tetmesh and create the rogue attributes)
5. switch simulator to Newton, and run the simulation. (The simulation will error out due to unknown attribute frequency).
```

### System Information

Newton - 1.2.1

eric-heiden:
## Description

Prevent unrelated USD attributes from aborting TetMesh and volume-deformable imports.

This PR:

- infers an unambiguous length-1 TetMesh custom attribute as `AttributeFrequency.ONCE`, while preserving the ambiguity error when a geometry count is also one;
- makes direct `get_tetmesh()` loading warn and omit custom arrays whose frequency cannot be inferred instead of rejecting the entire TetMesh;
- makes `add_usd()` load only builder-registered TetMesh custom attributes, using the registered frequency as authoritative;
- validates registered attribute lengths and skips malformed or unsupported values without dropping the soft body; and
- flattens indexed USD primvars before importing their values.

This fixes PhysX-authored volume deformables containing irrelevant metadata such as `deformablePose:default:omniphysics:purposes = ["bindPose"]` or `physxVolumeDeformableSim:simMeshHexCrc`. These attributes are no longer normalized eagerly and cannot prevent the simulation TetMesh from loading.

Closes #3228

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_import_usd_deformable_volume
# Ran 23 tests: OK

# Ran 373 tests: OK (skipped=1)

uvx pre-commit run -a
# All hooks passed
```

Also verified manually with the reported OmniPhysics representation:

```usda
uniform token[] deformablePose:default:omniphysics:purposes = ["bindPose"]
```

Direct `get_tetmesh()` loading classifies this as `ONCE` on an unambiguous multi-tet mesh, while `ModelBuilder.add_usd()` imports the soft body without warning about or retaining the unregistered metadata.

## Bug fix

**Steps to reproduce:**

1. Author or cook a `UsdGeom.TetMesh` with more than one tetrahedron.
2. Author the PhysX/OmniPhysics length-1 `deformablePose:default:omniphysics:purposes` token array on the simulation TetMesh.
3. Import the stage using `ModelBuilder.add_usd()`.
4. Without this fix, custom-attribute frequency inference raises before `add_usd()` can discard the unregistered attribute.

**Minimal reproduction:**

```python
import newton
from pxr import Sdf, Usd, UsdGeom

tet = UsdGeom.TetMesh.Define(stage, "/Soft")
tet.CreatePointsAttr(
    [
        (0.0, 0.0, 0.0),
        (1.0, 0.0, 0.0),
        (0.0, 1.0, 0.0),
        (0.0, 0.0, 1.0),
        (1.0, 1.0, 1.0),
    ]
)
tet.CreateTetVertexIndicesAttr([(0, 1, 2, 3), (1, 2, 3, 4)])
tet.GetPrim().CreateAttribute(
    "deformablePose:default:omniphysics:purposes",
    Sdf.ValueTypeNames.TokenArray,
).Set(["bindPose"])

builder.add_usd(stage)

assert builder.particle_count == 5
assert builder.tet_count == 2
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved TetMesh USD loading for custom attributes, including reliable handling of length-one arrays and indexed primvars.
  - Unambiguous custom attributes are now imported with the correct frequency.
  - Ambiguous or invalid attributes are skipped with warnings without preventing the soft body from loading.
  - `ModelBuilder.add_usd()` now imports only registered TetMesh custom attributes with their declared frequencies.


haixiangl-nv:
Is it possible to skip attributes with frequencies that are unable to be inferred at all, instead of erroring out?

If user has some custom attributes with weird size, but are irrelevant to the newton solver, usd parsing will error out even though the attributes are not relevant to Newton. It would be nice for Newton to print a warning regarding the bad attributes, instead of erroring out.

huidongc:
> Is it possible to skip attributes with frequencies that are unable to be inferred at all, instead of erroring out?
> 
> If user has some custom attributes with weird size, but are irrelevant to the newton solver, usd parsing will error out even though the attributes are not relevant to Newton. It would be nice for Newton to print a warning regarding the bad attributes, instead of erroring out.

Agreed. I am hitting the same issue in IL. 

`parse_usd()` later filters `tetmesh.custom_attributes` down to only attributes registered on the `ModelBuilder`. The offending  attribute would be dropped anyway, but the failure occurs earlier during `TetMesh` construction.

Newton should not hard-fail on an unregistered custom attribute that will be discarded.

huidongc:
@eric-heiden Do you plan to rework and fix the issue? We hit a bug related to this - NVBUG#6445358.

mzamoramora-nvidia:
@eric-heiden @huidongc @mmichelis I have a few follow up commits regarding this PR on this branch: https://github.com/mzamoramora-nvidia/newton/tree/mzamoramora/pr-3240-merge-ready 
Would that solve the issue for you?

huidongc:
@mzamoramora-nvidia yes, that will fix the error.

vreutskyy (approved):

PR #3243, opened 2026-06-25
Fix memory growth in the Style3D solver when CUDA Graph capture is disabled

haoran-whynot:
## Description

This PR moves the Style3D contact Hessian-vector multiply Warp kernel out
of `Collision.hessian_multiply()` and defines it at module scope.

`hessian_multiply()` is called from the PCG solve when collision is enabled.
Defining `@wp.kernel` inside that method constructs a new kernel object on
each call, adding Python/Warp registration and cache lookup overhead in a
hot path.

The change keeps numerical behavior unchanged: it still multiplies the
contact Hessian diagonal by the input vector and writes to the same output
buffer.

This also makes the collision solving path more friendly to non-CUDA-capture scenarios.

## Checklist

## Test plan

```bash
uvx pre-commit run -a

## Summary by CodeRabbit

## Summary

* **Bug Fixes**
  * Fixed memory growth in the Style3D solver when CUDA Graph capture is disabled.

* **Performance / Reliability**
  * Prevented repeated kernel rebuild behavior during Style3D collision Hessian-vector computations to improve performance and consistency.

* **Refactor**
  * Streamlined Style3D collision Hessian-vector product computation by routing it through a dedicated kernel, keeping output the same.

### DIFF
--- a/newton/_src/solvers/style3d/collision/collision.py
+++ b/newton/_src/solvers/style3d/collision/collision.py
@@ -9,6 +9,7 @@ from newton._src.solvers.style3d.collision.kernels import (
     eval_body_contact_kernel,
     handle_edge_edge_contacts_kernel,
     handle_vertex_triangle_contacts_kernel,
+    hessian_multiply_kernel,
     solve_untangling_kernel,
 )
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: newton/_src/solvers/style3d/collision/collision.py [hunk 1]
```diff
@@ -17,6 +17,17 @@
 ########################################################################################################################
 
 
+@wp.kernel
```
WenchaoHuang: Thanks! I wasn't aware that this could cause an issus before. You're right, this kernel should be moved outside of the function. Please move it to `./collision/kernels.py` for better code organization.

WenchaoHuang (changes requested):

WenchaoHuang (changes requested):
Please update `CHANGELOG.md`

newton/_src/solvers/style3d/collision/collision.py
haoran-whynot: Thanks, updated. I moved the kernel to collision/kernels.py and added a CHANGELOG entry under Fixed.

WenchaoHuang (approved):

adenzler-nvidia:
Thanks - this one actually needs a warp fix, the leak is real and fixable. Still a good idea to move it to module-scope though.

adenzler-nvidia (approved):

PR #3245, opened 2026-06-25
Allow release-assets host in license check egress

adenzler-nvidia:
## Description

The `check-licenses` job started failing with `connect ECONNREFUSED` during `actions/setup-go`, which is run by the `apache/skywalking-eyes/header` composite action (it builds `license-eye` from Go source on every run). The Go toolchain tarball is downloaded from `release-assets.githubusercontent.com`, which was missing from this job's harden-runner allowlist.

While that host resolved into GitHub's Fastly CDN range it was implicitly trusted, but release assets are now also served from AWS addresses that harden-runner drops — so the download fails before any license header is checked. This is why the job is flaky: it only fails when the host resolves to an AWS address.

This adds `release-assets.githubusercontent.com:443` to `allowed-endpoints`, matching `ci.yml` and `pr.yml` which already list it.

## Checklist

## Test plan

CI-only change; verified by the `License Check` workflow passing on this PR. The `ci.yml` and `pr.yml` workflows already allowlist this host and pass.

## Bug fix

**Steps to reproduce:**

1. Run the `License Check` workflow on a runner where `release-assets.githubusercontent.com` resolves to an AWS address (e.g. `54.185.253.63`).
2. `actions/setup-go` attempts to download the Go 1.23 toolchain from `https://github.com/actions/go-versions/releases/download/...`, which redirects to `release-assets.githubusercontent.com`.
3. harden-runner drops the connection (`domain not allowed`), `setup-go` fails with `connect ECONNREFUSED`, and the job errors before checking any license header.

## Summary by CodeRabbit

* **Chores**
  * Updated a pull request check workflow to allow access to an additional required endpoint, helping the check run successfully and reliably.

### DIFF
--- a/.github/workflows/pr_license_check.yml
+++ b/.github/workflows/pr_license_check.yml
@@ -28,6 +28,7 @@ jobs:
             api.github.com:443
             github.com:443
             proxy.golang.org:443
+            release-assets.githubusercontent.com:443
 
       - name: Checkout repository
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # v6.0.1


.github/workflows/pr_license_check.yml
```diff
@@ -28,6 +28,7 @@ jobs:
             api.github.com:443
             github.com:443
             proxy.golang.org:443
+            release-assets.githubusercontent.com:443
```
jcarius-nv, line 31: this would mean we allow downloading from any github content, basically any binary that someone puts in their release assets, right? That almost defeats the point of hardening the runners, it seems.

Kenny-Vilella: To my knowledge, this is download only. So even if an attacker would pull an affected binary, what are the consequence for us?
As long as you cannot push out it should be OK.

jcarius-nv: Well, couldn't a malicious PR pull a malicious binary and then also execute it in their code if this job runs before any human review?

Together with being able to write to github, that would mean they could exfiltrate anything and upload it to a repo of their choice again, no?

adenzler-nvidia: there might be an alternate solution, I see github also changed the versions of golang that are cached, we might just need to request a different version. I'm not sure how much we can rely on the caching though.

One thing to note is that this URL is already allowed for the PR and CI workflows.

Kenny-Vilella: But the job has read permission only no?
Would it be possible to write anything to github?

jcarius-nv: I see, ok, yeah with read-only it's ok, especially if we already allow this in the PR and CI workflows anyways

adenzler-nvidia: ok my alternate solution does not work, so I think we should go with this. I can confirm the action is read-only.

jcarius-nv (approved):

PR #3248, opened 2026-06-26
Improve mpr/gjk for very non-uniformly scaled boxes

nvtw:
## Description
Fix MPR/GJK contact initialization for small convex/primitive shapes touching large thin boxes, such as objects resting near the edge of a table. The box seed now stays laterally aligned with the partner shape so contacts resolve to the near face with a shallow distance instead of a side-face normal and spuriously deep penetration.
## Checklist
## Test plan
```bash
uv run --extra dev -m newton.tests -k test_thin_box_convex_contacts
```
## Bug fix
Small shapes contacting a large thin box away from the box center could report a side-face normal and excessive penetration instead of the expected top-face contact.
**Steps to reproduce:**
1. Place a small convex/primitive shape near the edge of a thin box, for example a table with half-extents `(8.0, 6.0, 0.1)`.
2. Run collision detection with the shape separated, touching, or shallowly penetrating the top face.
3. Observe the reported contact normal pointing sideways or the signed distance being far too deep.
**Minimal reproduction:**
```python
import newton

# See newton/tests/test_thin_box_convex_contacts.py for the focused regression.
```

## Summary

* **Bug Fixes**
  * Fixed MPR/GJK contact reporting for small-shape interactions with large thin boxes, correcting side-face normals and eliminating spurious deep penetration.
  * Improved thin-box edge and overhang contacts so contact normals remain aligned with the expected near face (+Z) at shallow-to-penetrating separations.

* **New Features**
  * Enhanced MPR/GJK initial contact resolution for extreme “slab” box setups to improve stability.

* **Tests**
  * Added regression tests for thin-table box/convex contacts across many shape types.
  * Added dynamic stability tests for box piles on thin tables and ground (no energy injection/sinking).


mzamoramora-nvidia:
@nvtw  Nice work! Just a few small things to note:

1. **There's some leftover code that isn't used.** `closest_point_on_box_surface` (`mpr.py:43-62`) doesn't get called anywhere anymore — looks like the recent rewrite stopped using it. Its docstring even mentions "callers" that no longer exist. Probably safe to just delete it.

2. **The box-vs-box seed reuses an already-adjusted value.** In `mpr.py:303-317`, the box-A block tweaks `center_a`, and then the box-B block reads that tweaked value instead of A's original center. For two very flat boxes overhanging each other, this makes the result depend on which box is "A." It doesn't change the final contact (it's just the starting guess), but it'd be cleaner to grab `center_a` before the box-A block changes it. CodeRabbit flagged this one too.

3. **A few test docstrings are out of date.** In `test_thin_box_convex_contacts.py` (lines 14-18, 167-169, 192-194), they mention an "analytical box-box path" that this pipeline doesn't actually use (box-box goes through GJK/MPR — see `narrow_phase.py:622`). Two of them also disagree with each other about whether the overhang case is handled, and the docstring says we seed from "the closest point on the box" when the code actually projects onto the face. Worth a quick refresh so the docs match the code.

4. **One comment promises a bit more than the code delivers.** The note in `mpr.py:295` says the 1% blend keeps `v0` off the face, but that isn't quite true when the partner hangs past the edge of the box. A proposal for rewording would be: 

    ```
    # The 1% blend nudges the seed inward along the thin axis while preserving
    # a face-aligned initial direction for edge overhang cases.
    ```

nvtw:
Let's close this as it is not really a clean fix and the problem this tries to solve turned out to be a solver bug.

Issue #3242, opened 2026-06-25
[BUG] Collision detection uses `mesh_query_point_sign_parity` -- silently broken for non-watertight meshes


WenchaoHuang:
### Bug Description

The mesh collision detection code in `geometry/sdf_contact.py` and `geometry/kernels.py` uses `wp.mesh_query_point_sign_parity()` to compute signed distances. This function requires watertight (closed, manifold) meshes. However, most rigid-body meshes in practice are not watertight (e.g newton-assets/unitree_h1).

Afther  PR #3004 is merged,  users with non-watertight mesh cases may encounter collision detection failures — for example `examples/cloth/example_cloth_style3d.py` and `examples/cloth/example_cloth_h1.py`. I'd like to preserve the previous code path while adding an option such as `watertight_mesh = False` for backward compatibility with existing projects. In practice, most production meshes are not watertight — they can be processed to become watertight, but it's not that convenient.

<img width="1909" height="1099" alt="Image" src="https://github.com/user-attachments/assets/4d800d0e-aef4-4651-9bcb-51fae5afbcab" />
<img width="1907" height="1095" alt="Image" src="https://github.com/user-attachments/assets/a7226778-5fef-45eb-8b51-52128e914f28" />

### Reproduction Script

```python
uv run python .\newton\examples\cloth\example_cloth_style3d.py
uv run python .\newton\examples\cloth\example_cloth_h1.py
```

### System Information

- Newton: v1.4.0
- Python 3.12.1
- Windows 11

nvtw:
Hm, this is strange - create_from_mesh has a parameter called sign_method which defaults to "auto". Auto means the mesh is checked for watertightness and for non-watertight meshes, winding numbers (slower) are used. I'll have a look.

Although I checked on latest Newton main - maybe this SDF improvement was added after Newton 1.3.0?

WenchaoHuang:
@nvtw correction v1.4.0, commit 87b6d5de48bbd378c76b2a4696a72f1115d95965

nvtw:
@AnkaChan the MR https://github.com/newton-physics/newton/pull/3004 mentioned above was merged by you. Do you have more insights about reasons behind this change and why it could cause failures as mentioned in this issue?

AnkaChan:
[...]

PR #3249, opened 2026-06-26
Fix handling of non watertight meshes

nvtw:
## Description
Fix mesh collision sign queries so parity is used only for watertight meshes, while open/non-watertight meshes use the normal-based query. Should address https://github.com/newton-physics/newton/issues/3242
## Checklist
## Test plan
```bash
uv run --extra dev -m newton.tests -k test_mesh_query_type_tracks_watertight
```
## Bug fix
Non-watertight meshes could use parity sign queries, producing incorrect inside/outside signs for collision and SDF-backed contact generation.
**Steps to reproduce:**
1. Add an open mesh, such as a single square plane, as collision geometry.
2. Query mesh SDF/contact signs near the open surface.
3. Observe parity-based signs treating the open mesh as if it were closed.
**Minimal reproduction:**
```python
import newton

# Create an open mesh and finalize a model; its mesh query type should be normal-based.
```

## Summary by CodeRabbit

* **New Features**
  * Added per-shape mesh sign-query selection with automatic watertight vs non-watertight behavior, and manual overrides via new mesh sign flags.
  * Soft-contact generation and mesh-SDF sampling now follow the selected sign strategy consistently.
* **Bug Fixes**
  * Mesh sign queries now use parity for watertight meshes and a normal-based method for non-watertight meshes.
  * Closest-point mesh queries were updated to use signless mesh queries.
* **Tests**
  * Expanded coverage for watertightness tracking, sign-method resolution/overrides, and invalid sign-flag combinations.


AnkaChan:
Thank you Tobi for doing this! Really appreciate it! I wonder if it makes sense to allow the user to specify the type of sdf query method when calling the `add_shape_**` method?

newton/_src/geometry/sdf_contact.py
```diff
@@ -529,12 +535,12 @@ def _sample_sdf_at_t(
             if sdf_is_heightfield:
                 return sample_sdf_heightfield(hfd_sdf, elevation_data, pp)
             elif use_bvh_for_sdf:
-                return sample_sdf_using_mesh(sdf_mesh_id, pp)
+                return sample_sdf_using_mesh(sdf_mesh_id, pp, 1000.0, sdf_mesh_query_type)
```
WenchaoHuang: I noticed a few places where `1000.0` is hard-coded. Would it make sense to replace it with a named constant, such as `DEFAULT_MAX_SEARCH_DIST`, `INIT_SEARCH_RADIUS` or something similar?

I also wonder whether this should simply be `wp.inf` instead. Since the physical meaning of 1000 depends on the simulation scale, a fixed limit may be appropriate for some solvers but overly restrictive for others. For example, in the VBD solver, 1000 corresponds to only 10 m. Using `wp.inf` (or at least a much larger value) might help avoid subtle issues in edge cases.

newton/_src/geometry/kernels.py
```diff
@@ -11,6 +11,9 @@
     GeoType,
 )
 
+MESH_SIGN_QUERY_NORMAL = 0
```
WenchaoHuang: Nitpick: It might be slightly cleaner to represent this constant as an `enum`, similar to `GeoType`.  😃

AnkaChan:
Thanks for the fix, Tobi! One thing worth catching before this lands:

**`mesh_sdf` (and the public `sdf_mesh`) now uses the normal sign query for *all* meshes.** In `kernels.py`, `mesh_sdf` was changed from `wp.mesh_query_point_sign_parity` to `mesh_sdf_with_sign_query(..., MESH_SIGN_QUERY_NORMAL)`. Unlike the contact paths, this one isn't gated by `is_watertight` — it's a hard-coded default — so watertight meshes lose parity here. On the thin-gap fixture from this PR's own `test_parity_sign_accuracy_exceeds_normal_query`, the normal query is ~0% sign-correct vs ~100% for parity, so this is a real regression for watertight meshes. It also affects the public `newton.geometry.sdf_mesh` and `examples/diffsim/example_diffsim_drone.py`.

FWIW the accuracy test was updated to call `mesh_sdf_with_sign_query(..., MESH_SIGN_QUERY_PARITY)` instead of `mesh_sdf`, so the default-path regression isn't caught by CI.

Simplest fix: default `mesh_sdf` back to `MESH_SIGN_QUERY_PARITY` (its prior behavior). If normal is intended, could you add a CHANGELOG note + a test that pins the default?

Minor: the non-watertight fallback here uses `normal`, whereas the SDF-construction path (`SDF.create_from_mesh(sign_method="auto")`) and the `is_watertight` docstring use `winding` for non-watertight. Fine for open planes — just worth a one-line comment so it doesn't read as inconsistent.

newton/_src/geometry/kernels.py
```diff
@@ -12,6 +14,13 @@
 )
 
 
+class MeshSignQuery(enum.IntEnum):
+    """Mesh sign-query strategies used by collision kernels."""
+
+    NORMAL = 0
+    PARITY = 1
```
eric-heiden, line 21: Need inline docstrings, maybe come up with better names here.

newton/tests/test_collision_pipeline.py
```diff
@@ -771,6 +803,122 @@ def test_parity_sign_accuracy_exceeds_normal_query(test, device):
...
+def test_shape_config_mesh_sign_flags(test, device):
+    cfg = newton.ModelBuilder.ShapeConfig()
+    cfg.flags |= ShapeFlags.MESH_SIGN_NORMAL
+    test.assertTrue(cfg.flags & ShapeFlags.MESH_SIGN_NORMAL)
+
+    cfg.flags |= ShapeFlags.MESH_SIGN_PARITY
+    with test.assertRaisesRegex(ValueError, "Invalid mesh sign method"):
+        cfg.validate(GeoType.MESH)
+
+    cfg.flags = 1 << 7
```
mmacklin, line 910: Shouldn't this use a named enum value rather than a numeric literal? Not clear what the 7 is here.

newton/_src/geometry/flags.py
```diff
@@ -35,6 +35,15 @@ class ShapeFlags(IntEnum):
     HYDROELASTIC = 1 << 4
     """Indicates that the shape uses hydroelastic collision."""
 
+    MESH_SIGN_NORMAL = 1 << 5
```
mmacklin, line 38: Would be good to document which routines these flags have an effect. Is it just in the CollisionPipeline? Under which circumstances? Are they respected during `build_sdf()`?

When would a user use one or the other?

Do they have an impact on performance, etc.

newton/_src/geometry/sdf_contact.py
```diff
@@ -19,6 +19,9 @@
     GlobalContactReducerData,
     export_and_reduce_contact_centered_two_spatial_depths,
 )
+from .kernels import MeshSignMethod, mesh_query_point_sign, resolve_mesh_sign_method
+
+_MESH_QUERY_MAX_DIST = 1000.0
```
mmacklin, line 24: Should probably be either, larger, or somehow tied to global units.

newton/_src/geometry/sdf_contact.py
```diff
@@ -1505,7 +1535,12 @@ def mesh_sdf_collision_global_reduce_kernel(
                                 midpoint_sdf = sample_sdf_heightfield(hfd_sdf, heightfield_elevations, bsphere_center)
                                 add_edge = midpoint_sdf <= threshold
                             elif use_bvh_for_sdf:
-                                midpoint_sdf = sample_sdf_using_mesh(mesh_id_sdf, bsphere_center, 1.01 * threshold)
+                                midpoint_sdf = sample_sdf_using_mesh(
+                                    mesh_id_sdf,
+                                    bsphere_center,
+                                    1.01 * threshold,
```
mmacklin, line 1541: Can we remove this magic number?

newton/_src/geometry/sdf_contact.py
AnkaChan: **[1/7]** Addressed in #3403: the constant is named (`_MESH_QUERY_MAX_DIST`) and raised to `1e11` — effectively unbounded for any asset under any unit convention (the queries run in mesh-local coordinates, with per-shape scale divided out, so there's no simulation scale to tie it to).

On `wp.inf` / float32-max: Warp's mesh queries square `max_dist` internally (`native/mesh.h`: `min_dist_sq = max_dist * max_dist`), and the no-hit branches return `max_dist` as a sentinel that contact math multiplies by per-shape scale. Budgeting scale factors up to `1e6` for extreme unit conversions, `(1e11 * 1e6)^2 = 1e34` clears `FLT_MAX` (~3.4e38) by four orders of magnitude, whereas values near `sqrt(FLT_MAX) ~ 1.8e19` would overflow to `inf` and propagate into contacts.

newton/_src/geometry/kernels.py
AnkaChan: **[2/7]** Done — `MeshSignMethod(IntEnum)` and `MeshProperties(IntFlag)`. In #3403 the values additionally match the `ShapeFlags` field encoding (`AUTO=0, NORMAL=1, PARITY=2`; flag = `method << 5`), so the flag field and the enum are the same number.

newton/_src/geometry/kernels.py
AnkaChan: **[3/7]** Member docstrings added in #3403. On names: kept `MeshSignMethod` / `MeshProperties` — after the flag-field redesign they line up 1:1 with `ShapeFlags.MESH_SIGN_*`, which reads consistently in context. If you had specific alternatives in mind, happy to rename there.

newton/tests/test_collision_pipeline.py
AnkaChan: **[4/7]** Root cause fixed in #3403 rather than just naming the literal: the mask claimed three bits (`0b111 << 5`) while only two one-hot method bits existed, so `1 << 7` was a phantom "invalid" bit. The field is now a continuous 3-bit value (`MESH_SIGN_AUTO=0`, `NORMAL=1`, `PARITY=2`, remaining values reserved for e.g. a future winding mode), and the tests use a named `_RESERVED_MESH_SIGN_FLAGS` constant for the invalid-encoding cases.

newton/_src/geometry/flags.py
AnkaChan: **[5/7]** Documented in the `ShapeFlags.MESH_SIGN_*` docstrings in #3403 — summary below.

**Which routines honor these flags?**
1. **Runtime mesh sign queries** — everywhere a kernel asks "is this point inside the mesh": soft/particle contacts (`create_soft_contacts`) and the mesh-mesh SDF narrow phase driven by `CollisionPipeline`.
2. **SDFs baked during `ModelBuilder.finalize()`** — the deferred bake maps the shape's flag to the corresponding bake `sign_method`.

**Are they respected during `build_sdf()`?**
Only for the finalize-time bakes above. A direct `mesh.build_sdf()` call runs before any shape (and its flags) exists, so it can't see them — it now takes its own `sign_method` argument (`"auto" | "parity" | "winding" | "normal"`) for that.

**What does each value do, and when would a user set it?**

| flag | runtime queries | baked SDF | use when |
|---|---|---|---|
| `MESH_SIGN_AUTO` (= 0, default) | parity if watertight, else pseudo-normal | parity if watertight, else winding | default — topology decides |
| `MESH_SIGN_PARITY` | parity | parity | a closed mesh is misdetected as open (e.g. unwelded vertices) and you want the fast path back |
| `MESH_SIGN_NORMAL` | pseudo-normal | pseudo-normal | open sheets/planes, where "inside" is undefined and you want side-of-surface |

Background on the three methods: **parity** counts ray crossings — cheap, correct only for closed meshes; **pseudo-normal** signs against the angle-weighted normal of the closest feature — meaningful for open surfaces; **winding** uses the generalized winding number — robust for messy solids, but the most expensive.

**Performance impact?**
The flags only select which Warp query runs — no extra memory or per-contact bookkeeping. Parity is the cheapest, pseudo-normal comparable, winding the most expensive and the only mode requiring winding-number acceleration on the mesh (now built only when winding is actually used).

newton/_src/geometry/sdf_contact.py
AnkaChan: **[6/7]** Raised to `1e11` in #3403 — effectively unbounded (~8x Earth's diameter in millimeter units). It can't be tied to global units: these queries run in mesh-local coordinates with the per-shape scale divided out. The ceiling for going larger is float32: Warp's mesh queries square `max_dist` internally, and the no-hit sentinel is multiplied by per-shape scale in contact math — budgeting scale up to `1e6`, `(1e11 * 1e6)^2 = 1e34` still clears `FLT_MAX` by four orders, while anything near `sqrt(FLT_MAX) ~ 1.8e19` would overflow. It's never binding in practice: culling passes tight `slack * threshold` radii and the remaining queries run at points the narrow phase already placed near the surface.

newton/_src/geometry/sdf_contact.py
AnkaChan: **[7/7]** Named in #3403: `_SDF_QUERY_RADIUS_SLACK = 1.01`, with a comment — the search radius slightly exceeds the culling threshold so points right at the threshold are classified by a real closest-point query rather than the no-hit sentinel. Applied to both kernel variants.

AnkaChan:
Superseded by #3403.

PR #3252, opened 2026-06-26
Warn on MuJoCo velocity limit drops

jcarius-nv:
## Description

`SolverMuJoCo` now warns when a model contains non-default finite `Model.joint_velocity_limit` values, because MuJoCo has no equivalent joint velocity clamp and those limits are otherwise silently ignored. The MuJoCo integration docs now call out the behavior and the recommended alternatives: effort limits, damping/controller limits, or a custom actuator.

## Checklist

## Test plan

```bash
UV_CACHE_DIR=/tmp/uv-cache UV_TOOL_DIR=/tmp/uv-tools PRE_COMMIT_HOME=/tmp/pre-commit-cache uvx --python 3.12 pre-commit run ruff-format --files newton/_src/solvers/mujoco/solver_mujoco.py newton/tests/test_mujoco_solver.py
UV_CACHE_DIR=/tmp/uv-cache UV_TOOL_DIR=/tmp/uv-tools PRE_COMMIT_HOME=/tmp/pre-commit-cache uvx --python 3.12 pre-commit run ruff --files newton/_src/solvers/mujoco/solver_mujoco.py newton/tests/test_mujoco_solver.py
UV_CACHE_DIR=/tmp/uv-cache uv run --extra dev -m newton.tests -k test_warn_unsupported_joint_velocity_limits
UV_CACHE_DIR=/tmp/uv-cache uv run --extra dev -m newton.tests -k test_joint_attributes_registration_and_updates
git diff --check
```

## Bug fix

**Steps to reproduce:**

1. Build or import an articulated model with a non-default finite `Model.joint_velocity_limit`.
2. Construct `SolverMuJoCo(model)`.
3. Observe that MuJoCo has no corresponding velocity-limit field, so the authored limit is dropped without feedback.

**Minimal reproduction:**

```python
import newton

joint = builder.add_joint_revolute(-1, body, velocity_limit=1.0)
model = builder.finalize()
SolverMuJoCo(model)
```

## Summary by CodeRabbit

* **Bug Fixes**
  * `SolverMuJoCo` now emits a `RuntimeWarning` when authored, finite, non-default `joint_velocity_limit` values can’t be enforced by MuJoCo (warning includes affected DOF indices when applicable).
* **Documentation**
  * Updated MuJoCo integration docs to clarify there’s no MuJoCo equivalent for `Model.joint_velocity_limit`, and that the warning is emitted at `SolverMuJoCo` construction time.
  * Added guidance for achieving hard speed limiting via controllers/custom actuators (and/or effort limits and damping).
* **Tests**
  * Added targeted tests for the warning logic and updated solver initialization tests to expect the warning.


newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -112,6 +113,42 @@
...
+
+    finite_limits = np.isfinite(joint_velocity_limit)
+    non_default_limits = finite_limits & (joint_velocity_limit != _DEFAULT_JOINT_VELOCITY_LIMIT)
+    if not np.any(non_default_limits):
+        return
+
+    dof_indices = np.flatnonzero(non_default_limits)
+    shown = dof_indices[:8].tolist()
+    suffix = "" if len(dof_indices) <= len(shown) else f" and {len(dof_indices) - len(shown)} more"
+    warnings.warn(
```
jcarius-nv, line 144: maybe info-level print is enough here? The user may not have done anything wrong if the usd converter authored the limit but they want to go with the mujoco solver nonetheless, aware that the limit won't be respected.

jcarius-nv:
Not sure if it's even worth a warning here, happy to hear opinions!

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -112,6 +113,42 @@
     "Model.mujoco.dof_passive_damping is deprecated and will be removed in a future release. "
     "Use Model.joint_damping instead."
 )
+_DEFAULT_JOINT_VELOCITY_LIMIT = ModelBuilder.JointDofConfig().velocity_limit
```
adenzler-nvidia, line 116: **Blocking — this initializes Warp at `import newton` and breaks `test_lazy_init`.**

`solver_mujoco` is imported eagerly by `import newton`, so this module-level assignment runs at import time. `JointDofConfig.__init__` calls `wp.normalize(...)`, which triggers `wp.init()` — so `import newton` now initializes the Warp runtime and `test_lazy_init.test_import_newton_does_not_init_warp` fails on every platform (this is what's failing the GPU + all four CPU unittest jobs, not the new velocity-limit tests, which pass).

Note that #2631 fixed exactly this class of module-level eager-init (module-level `wp.sin`/`wp.cos`/`wp.pi` → stdlib `math.*`) and added that test as the regression guard, because lazy init is what lets `WARP_CACHE_ROOT` and other Warp config be set after import but before init.

Suggest computing the default lazily inside `_warn_unsupported_joint_velocity_limits` (Warp is already initialized by the time a solver is constructed), or reading it without constructing a config: `inspect.signature(ModelBuilder.JointDofConfig.__init__).parameters["velocity_limit"].default`.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -112,6 +113,42 @@
...
+        stacklevel += 1
+    return 2
+
+
+def _warn_unsupported_joint_velocity_limits(joint_velocity_limit: np.ndarray | None) -> None:
+    if joint_velocity_limit is None or joint_velocity_limit.size == 0:
+        return
+
+    finite_limits = np.isfinite(joint_velocity_limit)
+    non_default_limits = finite_limits & (joint_velocity_limit != _DEFAULT_JOINT_VELOCITY_LIMIT)
```
adenzler-nvidia, line 137: **Non-blocking / follow-up (codebase consistency, not for this PR).**

This treats a limit as "real" when it's finite and `!= 1e6` (the `JointDofConfig` default), but the codebase's canonical "unlimited" sentinel is `MAXVAL = 1e10` — used and documented for `limit_lower`/`limit_upper`, and already detected for position limits in this same solver via `lower <= -MAXVAL and upper >= MAXVAL`. So a user who follows that convention and sets `velocity_limit = MAXVAL` to mean "unlimited" would get a spurious warning here.

Not this PR's job to fix — `velocity_limit` doesn't adopt the `MAXVAL` convention today — but flagging it as a follow-up to unify "unlimited" handling across position/velocity/effort limits.

PR #3254, opened 2026-06-26
Silence MJWarp contact overflow spam

jcarius-nv:
## Description

Newton-to-MJWarp contact conversion currently prints from inside the Warp kernel every full contact-conversion pass when the Newton contact count exceeds MJWarp `nconmax`. That can flood stdout for steady overflow cases. This PR removes the device-side `wp.printf`; conversion behavior is unchanged and still clips contacts to `nconmax`.

## Checklist

## Test plan

```bash
UV_CACHE_DIR=/tmp/uv-cache UV_TOOL_DIR=/tmp/uv-tools PRE_COMMIT_HOME=/tmp/pre-commit-cache uvx --python 3.12 pre-commit run ruff-format --files newton/_src/solvers/mujoco/kernels.py
UV_CACHE_DIR=/tmp/uv-cache UV_TOOL_DIR=/tmp/uv-tools PRE_COMMIT_HOME=/tmp/pre-commit-cache uvx --python 3.12 pre-commit run ruff --files newton/_src/solvers/mujoco/kernels.py
rg -n "wp\\.printf\\(|Number of Newton contacts \\(%d\\) exceeded MJWarp limit" newton/_src/solvers/mujoco/kernels.py
git diff --check
```

The `rg` command exits with status `1`, confirming the device print and message string are absent.

## Bug fix

**Steps to reproduce:**

1. Run `SolverMuJoCo` with `use_mujoco_contacts=False` and a scene where generated Newton contacts exceed MJWarp `nconmax`.
2. Step the simulation through repeated full contact conversion passes.
3. Observe repeated device-side stdout lines from `wp.printf`.

**Minimal reproduction:**

```python
# Any dense Newton-contact scene whose contact count exceeds SolverMuJoCo(..., nconmax=small_value)
# triggers the repeated device-side overflow print during contact conversion.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced noisy device logging during Newton→MJWarp contact conversion by emitting a warning only once per run when overflow occurs.
  * Added overflow diagnostics for Newton→MJWarp contact conversion: generated vs. clipped counts are now tracked and exposed via `get_newton_contact_overflow_counts()` (and clipped count via `get_newton_contact_overflow_count()`).
* **Tests**
  * Added/extended regression coverage to verify the warn-once behavior and that overflow counts remain stable across repeated conversions with tight `nconmax`/`njmax`.
* **Documentation**
  * Documented the clipping/overflow caveat and how to avoid it by increasing `nconmax`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,7 @@
 - Fix `newton.eval_fk` / `newton.eval_ik` producing wrong rotations and joint velocities for `JointType.D6` with three angular DOFs whose axes form a left-handed orthonormal basis.
 - Fix MJCF parsing so attributes from multiple `<compiler>` elements, including `<include>`-expanded children, are merged in document order. (#3030)
 - Fix MJCF worldbody static geoms bypassing the visual/collider class filter, so `parse_visuals=False` drops visual-class geoms attached directly to `<worldbody>` too. (#3030)
-- Fix `SolverMuJoCo` Newton-contact conversion spamming device prints when the Newton contact count exceeds MJWarp `nconmax`; call `get_newton_contact_overflow_count()` to inspect the latest clipped count. Increase `nconmax` to avoid contact clipping.
+- Fix `SolverMuJoCo` Newton-contact conversion spamming device prints when the Newton contact count exceeds MJWarp `nconmax`; it now emits one device warning per solver and exposes generated/clipped counts through `get_newton_contact_overflow_counts()`. Increase `nconmax` to avoid contact clipping.
 - Fix `cable_cross_slide_table` example stability so the cable-driven table reliably tracks its rectangular path and catches drift during regression runs.
 - Fix URDF `package://` mesh fallback resolution without `resolve-robotics-uri-py` so package names only match full path components instead of unrelated directory-name substrings
 - Fix `ModelBuilder.collapse_fixed_joints()` crashing with `IndexError` when a `mujoco:equality_constraint` row omits optional fields (`anchor`, `relpose`) that carry defaults. (#3054)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@
 - Fix `newton.eval_fk` / `newton.eval_ik` producing wrong rotations and joint velocities for `JointType.D6` with three angular DOFs whose axes form a left-handed orthonormal basis.
 - Fix MJCF parsing so attributes from multiple `<compiler>` elements, including `<include>`-expanded children, are merged in document order. (#3030)
 - Fix MJCF worldbody static geoms bypassing the visual/collider class filter, so `parse_visuals=False` drops visual-class geoms attached directly to `<worldbody>` too. (#3030)
+- Fix `SolverMuJoCo` Newton-contact conversion spamming device prints when the Newton contact count exceeds MJWarp `nconmax`; increase `nconmax` to avoid contact clipping.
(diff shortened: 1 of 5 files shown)


adenzler-nvidia (changes requested):
Removing this print silences the only signal that contacts are being clipped (`count = naconmax`), which silently drops contacts and produces wrong physics. Under CUDA graph capture we can't read the count back on the host to warn there, so this device-side print is the only diagnostic that survives — deleting it makes the overflow undetectable in exactly the case where it matters.

Two questions:

1. Where did this actually bite you — a steady-state scene where the Newton contact count sits above `nconmax` every step? Curious what the setup was.
2. Wouldn't it make more sense to fix the counts on the user side instead — i.e. ensure `nconmax`/`naconmax >= rigid_contact_max` so the clip is unreachable? Then there's nothing to print and the overflow simply can't happen, rather than being silenced.

jcarius-nv:
updated to be a print-once style warning, but I'm still unsure whether it's worth it at all. Seems to introduce a bit of complexity and I also need to check benchmarks still for comparison.

adenzler-nvidia:
Rather than detect-and-report clipping, it may be cleaner to make the conversion lossless by construction: require `naconmax >= contacts.rigid_contact_max` and raise otherwise, mirroring the existing guard in `update_contacts` (which already raises for the opposite mismatch, `naconmax > rigid_contact_max`, at `solver_mujoco.py:4389`).

Concretely, at the top of `_convert_contacts_to_mjwarp`, where `contacts.rigid_contact_max` is authoritative:

```python
if naconmax < contacts.rigid_contact_max:
    raise ValueError(
        f"MJWarp nconmax ({naconmax}) is smaller than contacts.rigid_contact_max "
        f"({contacts.rigid_contact_max}); Newton contacts would be silently clipped. "
        f"Increase nconmax to at least {contacts.rigid_contact_max}."
    )
```

With that invariant, MJWarp's buffer is always at least as large as Newton's, so conversion can never drop a contact that made it into the Newton buffer. The only way `count` can exceed `naconmax` is if it also exceeded `rigid_contact_max` — i.e. the Newton buffer itself overflowed — which `NarrowPhase`'s `verify_buffers` (on by default) already reports. So the device-side detection isn't needed: the `wp.printf`, the `contact_overflow_counts` / `contact_overflow_warned` arrays and kernel params, and both getters can all come out, and the PR reduces to this guard plus the docs/CHANGELOG note.

I'd use `>=` rather than `==` — there's no harm in `naconmax` being larger (it just over-allocates MJWarp's side), and exact equality would reject setups that legitimately size `nconmax` generously.

One caveat: this is a behavioral change — a too-small `nconmax` currently clips silently, and this makes it a hard error — so it'd want a `Changed` CHANGELOG entry with migration guidance ("increase `nconmax` to at least `rigid_contact_max`").

jcarius-nv:
Closing this draft in favor of #3702.

The underlying failure is valid, but this branch is not the right solution shape. Deduplicating the device print and adding conversion-specific getters does not establish a coherent capacity-overflow contract, and clipping could still be too easy to miss. #3702 now provides a concrete NaN reproduction and covers the related `nconmax`, `njmax`, and `CollisionPipeline` observability concerns.

A replacement should define graph-capture-safe overflow reporting and reset semantics, and investigate integrating Newton-to-MJWarp clipping with MuJoCo Warp’s existing overflow state rather than introducing parallel diagnostic state.

Thanks for the valid review feedback here.

PR #3261, opened 2026-06-26
Split disconnected mesh components before decomposition

eric-heiden:
## Description

Split disconnected mesh components before running convex decomposition through CoACD or V-HACD. This prevents a unified multi-part collision mesh from being decomposed as one connected input and merged into a single convex hull.

The component scan uses a NumPy-only vectorized union pass and avoids constructing temporary `Mesh` objects, keeping the common connected-mesh path lightweight.

## Checklist

No docs update: this preserves intended collision geometry during existing mesh approximation behavior.

## Test plan

```bash
uvx pre-commit run -a
uv run --extra dev --extra importers python -m unittest -v newton.tests.test_model.TestModelMesh.test_mesh_approximation newton.tests.test_model.TestModelMesh.test_mesh_approximation_coacd_uses_builder_default_cfg newton.tests.test_model.TestModelMesh.test_mesh_approximation_coacd_unavailable_falls_back_to_convex_hull newton.tests.test_model.TestModelMesh.test_mesh_approximation_ignores_non_mesh_shapes newton.tests.test_model.TestModelMesh.test_mesh_approximation_convex_hull_failure_falls_back_to_bounding_box newton.tests.test_model.TestModelMesh.test_mesh_approximation_convex_decomposition_preserves_visual_properties newton.tests.test_model.TestModelMesh.test_mesh_approximation_convex_decomposition_splits_disconnected_components newton.tests.test_model.TestModelMesh.test_approximate_meshes_collision_filter_child_bodies
uv run --extra dev --extra importers python -m unittest -v newton.tests.test_import_urdf.TestImportUrdfBasic.test_mesh_urdf
```

The new regression test was also run against `origin/main` with only the test applied and failed as expected: the backend received one `(16 vertices, 24 faces)` mesh instead of two `(8 vertices, 12 faces)` components.

## Performance

Median connectivity-scan timings on connected icospheres:

| Faces | Original Python walk | NumPy union pass | Speedup |
| ---: | ---: | ---: | ---: |
| 320 | 1.238 ms | 0.127 ms | 9.8x |
| 1,280 | 5.563 ms | 0.315 ms | 17.7x |
| 5,120 | 21.739 ms | 1.310 ms | 16.6x |
| 20,480 | 88.503 ms | 5.023 ms | 17.6x |

For comparison, real CoACD runs on the first three representative meshes took approximately 935 ms, 1,071 ms, and 1,210 ms. The optimized scan is therefore about 0.01%, 0.03%, and 0.11% of CoACD runtime, respectively. Fully disconnected synthetic meshes also improved: 1,000 triangles scanned in 0.721 ms versus 4.035 ms, and 5,000 in 5.439 ms versus 37.041 ms.

The optimized topology labels matched the original graph walk for 1,000 randomized meshes.

## Bug fix

**Steps to reproduce:**

1. Create one triangle mesh containing two disconnected convex parts.
2. Add it as a mesh shape.
3. Run `builder.approximate_meshes(method="coacd")`.
4. Observe that the decomposition backend may receive the full disconnected mesh as one input and return one merged hull.

**Minimal reproduction:**

```python
import numpy as np
import newton

vertices = np.array([
    [0, 0, 0], [1, 0, 0], [1, 1, 0], [0, 1, 0],
    [0, 0, 1], [1, 0, 1], [1, 1, 1], [0, 1, 1],
    [3, 0, 0], [4, 0, 0], [4, 1, 0], [3, 1, 0],
    [3, 0, 1], [4, 0, 1], [4, 1, 1], [3, 1, 1],
], dtype=np.float32)

cube_faces = np.array([
    [0, 1, 2], [0, 2, 3], [4, 6, 5], [4, 7, 6],
    [1, 5, 6], [1, 6, 2], [0, 3, 7], [0, 7, 4],
    [3, 2, 6], [3, 6, 7], [0, 4, 5], [0, 5, 1],
], dtype=np.int32)
faces = np.concatenate((cube_faces, cube_faces + 8), axis=0)

shape = builder.add_shape_mesh(-1, mesh=newton.Mesh(vertices, faces.flatten()))
builder.approximate_meshes(method="coacd", shape_indices=[shape])
print(builder.shape_count)
```

With the fix, the disconnected cubes are decomposed separately, so the builder gets separate convex mesh shapes for each component.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved convex decomposition for meshes with disconnected components by processing each component independently.
  * Corrected collision-filter relationships and settings for generated convex parts.
  * Ensured replacement meshes clear normals and UVs when convex parts are substituted.
* **Tests**
  * Added coverage for disconnected components, seam connectivity, and shape and collision settings.
* **Documentation**
  * Updated the 1.4.0 changelog with the corrected decomposition behavior.


newton/_src/sim/builder.py
```diff
@@ -7393,6 +7393,49 @@ def get_shape_custom_attributes(shape: int) -> dict[str, Any] | None:
...
             return custom_attributes or None
 
+        # Decomposition backends may merge disconnected convex parts into one hull.
+        def split_mesh_components(mesh: Mesh) -> list[tuple[np.ndarray, np.ndarray]]:
+            """Split mesh faces into vertex-connected components."""
+            faces = np.asarray(mesh.indices, dtype=np.int32).reshape(-1, 3)
+            if len(faces) <= 1:
+                return [(mesh.vertices, faces)]
+
+            parents = np.arange(len(mesh.vertices), dtype=np.int32)
```
mzamoramora-nvidia: Could we build connectivity from geometrically coincident vertices instead of raw vertex indices? Newton's `Mesh.create_box()` duplicates vertices per face by default, as do many meshes with UV seams or hard normals.

I tested a default Newton box on this branch. It was sent to both real backends as six separate 4-vertex, 2-triangle components, and both methods produced six shapes from one connected box. Could we use the mesh's canonical vertex IDs here and add a default `Mesh.create_box()` regression test?

newton/_src/sim/builder.py
```diff
@@ -7476,26 +7519,30 @@ def get_shape_custom_attributes(shape: int) -> dict[str, Any] | None:
...
-                            decomposition = coacd.run_coacd(cmesh, **coacd_settings)
-                        else:
-                            tmesh = trimesh.Trimesh(mesh.vertices, mesh.indices.reshape(-1, 3))
-                            vhacd_settings = {
-                                "maxNumVerticesPerCH": mesh.maxhullvert,
-                            }
-                            vhacd_settings.update(remeshing_kwargs)
-                            decomposition = trimesh.decomposition.convex_decomposition(tmesh, **vhacd_settings)
-                            decomposition = [(d["vertices"], d["faces"]) for d in decomposition]
+                        decomposition = []
```
mzamoramora-nvidia, line 7490: Could we preserve all shape settings when these extra component results are added below? Assigning `cfg.flags` in that path resets `density` to 1000 and `collision_group` to 1. The config also falls back to the builder's `gap` and `collision_filter_parent` values.

In a local dynamic-body test, decomposition changed body mass from `4.0` to about `1004.0`; the added component changed gap from `0.123` to `0.1` and collision group from `7` to `1`. It also gained a parent filter that the original shape disabled. Could we restore these values after setting flags and add a regression test with non-default shape settings?

newton/_src/sim/builder.py
```diff
@@ -7476,26 +7519,30 @@ def get_shape_custom_attributes(shape: int) -> dict[str, Any] | None:
...
+                        for component_vertices, component_faces in split_mesh_components(mesh):
```
mzamoramora-nvidia, line 7492: Could we clear or rebuild normals and UVs when the first component result is installed below? `Mesh.copy()` preserves the old arrays when replacements are not supplied.

I reproduced a result with 8 vertices but 16 normals and 16 UV coordinates after splitting a two-cube mesh. A test that starts with normals and UVs would help catch this mismatch.

newton/_src/sim/builder.py
```diff
@@ -7393,6 +7393,49 @@ def get_shape_custom_attributes(shape: int) -> dict[str, Any] | None:
             }
             return custom_attributes or None
 
+        # Decomposition backends may merge disconnected convex parts into one hull.
```
mzamoramora-nvidia: Nice fix. Once the connectivity behavior is settled, could we also describe it in the public `approximate_meshes()` docstring? Splitting by component can change the number and order of generated shapes, so it is useful for callers to know.

A short note may be enough: “For `coacd` and `vhacd`, each connected component is decomposed separately and may produce one or more convex shapes.”

newton/_src/sim/builder.py
eric-heiden: Fixed. The splitter now treats coincident vertices as connected and welds duplicate seam vertices before decomposition. I added a default Mesh.create_box() regression test for both CoACD and V-HACD.

newton/_src/sim/builder.py
eric-heiden: Fixed. Extra convex parts now keep the original flags, gap, collision group, materials, and collision filters while adding no mass. The regression test covers non-default settings and parent filtering.

newton/_src/sim/builder.py
eric-heiden: Fixed. Replacement convex meshes now clear normals and UVs because the backend does not return remapped attributes. The disconnected-mesh test starts with both arrays and checks that they are cleared.

newton/_src/sim/builder.py
eric-heiden: Added. The public approximate_meshes() docstring now says that CoACD and V-HACD process each connected component separately and may create one or more convex shapes.

mzamoramora-nvidia (dismissed):
Thanks for the updates. I reran the PR test plan locally, and the fixes for component connectivity, shape properties, normals and UVs, collision filters, and documentation look good.

I found one remaining concern: extra convex parts do not copy shape_force_sdf. With force_sdf=True, a two-part decomposition currently produces [True, False], so the parts can have different collision behavior. Could we copy this setting and cover it in the existing shape-settings test?

The performance comment is a non-blocking suggestion and could be handled separately.

newton/_src/sim/builder.py
```diff
@@ -7523,13 +7544,15 @@ def get_shape_custom_attributes(shape: int) -> dict[str, Any] | None:
...
                             margin=self.shape_margin[shape],
                             is_solid=self.shape_is_solid[shape],
-                            collision_group=self.shape_collision_group[shape],
-                            collision_filter_parent=self.default_shape_cfg.collision_filter_parent,
                         )
                         cfg.flags = self.shape_flags[shape]
+                        cfg.density = 0.0  # do not add extra mass / inertia
+                        cfg.gap = self.shape_gap[shape]
+                        cfg.collision_group = self.shape_collision_group[shape]
+                        cfg.collision_filter_parent = False
```
mzamoramora-nvidia, line 7556: Could we also set force_sdf=self.shape_force_sdf[shape]?

I tested a source mesh with force_sdf=True and a two-part decomposition. The resulting values were [True, False], so only the first part keeps the setting. Could we add this case to test_mesh_approximation_convex_decomposition_preserves_shape_settings?

newton/_src/sim/builder.py
```diff
@@ -7476,42 +7478,61 @@ def get_shape_custom_attributes(shape: int) -> dict[str, Any] | None:
...
+                    self.shape_source[shape] = replacement_mesh
                     # mark as convex mesh type
                     self.shape_type[shape] = GeoType.CONVEX_MESH
                     if len(decomposition) > 1:
                         body = self.shape_body[shape]
                         xform = self.shape_transform[shape]
                         color = self.shape_color[shape]
                         custom_attributes = get_shape_custom_attributes(shape)
+                        filtered_shapes = []
+                        for shape_a, shape_b in self.shape_collision_filter_pairs:
```
mzamoramora-nvidia: Non-blocking: could we build a lookup of filtered shapes once before this loop? The current code scans every collision-filter pair for each decomposed shape and can also expand the builder compact filter storage into a list. A follow-up issue would also be fine.

newton/_src/sim/builder.py
eric-heiden: Fixed in 0ae93977. Extra convex parts now copy `force_sdf` through `ShapeConfig`, and the existing shape-settings regression test sets it to `True` and verifies every generated part preserves it.

newton/_src/sim/builder.py
eric-heiden: Implemented in 0ae93977. The decomposition path now builds one reverse index directly from the compact backing store, limited to requested source shapes, so it neither rescans per shape nor materializes the store. Generated parts are tracked separately to retain all-to-all filters when both source shapes decompose. A 50k-pair/100-shape microbenchmark improved from 480.8 ms to 6.24 ms, and the regression test verifies the compact store remains unmaterialized.

newton/tests/test_model.py
```diff
@@ -936,6 +936,235 @@ def __init__(self, vertices, faces):
...
+                extra_shape = shape + 1
+                self.assertAlmostEqual(builder.body_mass[child], body_mass)
+                self.assertEqual(builder.shape_flags[extra_shape], builder.shape_flags[shape])
+                self.assertEqual(builder.shape_gap[extra_shape], builder.shape_gap[shape])
+                self.assertEqual(builder.shape_collision_group[extra_shape], builder.shape_collision_group[shape])
+                self.assertEqual(builder.shape_material_ke[extra_shape], builder.shape_material_ke[shape])
+                self.assertEqual(builder.shape_material_kd[extra_shape], builder.shape_material_kd[shape])
+                self.assertEqual(builder.shape_material_mu[extra_shape], builder.shape_material_mu[shape])
+                self.assertEqual(builder.shape_force_sdf[extra_shape], builder.shape_force_sdf[shape])
+                self.assertNotIsInstance(builder._shape_collision_filter_pairs, list)
```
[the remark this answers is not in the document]
eric-heiden: Fixed in 63062acd. The private test access now carries the file's established `pyright: ignore[reportPrivateUsage]` suppression, and the full pre-commit suite passes.

mzamoramora-nvidia (approved):
Thanks!

PR #3262, opened 2026-06-26
Add water-tight rigid-soft SDF contacts (detection + VBD integration)

AnkaChan:
## Description

Adds an SDF-based back-end that generates **full-surface** rigid–soft contacts so a soft body cannot tunnel through a rigid shape *between* its vertices, and feeds the new contacts into the VBD solver (both the soft side and the two-way rigid reaction). "Full-surface" = contacts over the whole soft-mesh surface (edges and triangle interiors), not only its vertices.

**Detection** (`newton/_src/geometry/soft_contacts_sdf.py`): three additive passes — the legacy particle pass, a soft-**edge** pass (1-D golden-section), and a soft-**face** pass (Frank-Wolfe over the barycentric simplex), per Macklin et al. 2020 ("Local Optimization for Robust SDF Collision"). Each minimizes the rigid shape's signed-distance field over its feature and emits a contact if within margin. All three passes write **one unified, self-describing record type** through a **single soft counter** (`counter_increment` with static per-pass thread-id offsets, so the tape-recorded backward pass replays identical indices). Gated by `enable_rigid_soft_full_surface_contact` — **off ⇒ byte-for-byte legacy behavior** (`soft_contact_count` and the record layout are then bit-identical to `main`).

**Unified contact records.** A soft contact is described by `Contacts.soft_contact_indices` (a `vec3i` of the soft-side particle ids, `-1` padded) plus `Contacts.soft_contact_barycentric`:
- particle contact `(p, -1, -1)`, barycentric `(1, 0, 0)`
- edge contact `(v0, v1, -1)`, barycentric `(u, 1-u, 0)`
- face contact `(v0, v1, v2)`, barycentric `(w0, w1, w2)`

The number of non-negative index slots gives the feature kind, so there are **no per-kind count ranges and no separate feature-id field** — `soft_contact_count` stays a single total (a 1-element view, unchanged from `main`). `soft_contact_particle` is kept as the particle-only view (particle id, `-1` for edge/face) for solvers that consume particle contacts exclusively. Detection **no longer needs mesh adjacency** — the edge pass reads endpoints straight from `edge_indices`.

**New API — `ModelBuilder.enable_rigid_mesh_sdfs()`:** an opt-in builder step (mirrors `color()`). It flags that the rigid **mesh/convex** shapes participating in soft contact need a volume SDF; `finalize()` then builds them, because finalize owns the device and volume-SDF construction is expensive (so it is opt-in, not automatic). Analytic primitives (box, sphere, capsule, …) use their closed-form SDF and need nothing. **Call it before `finalize()`.** If `CollisionPipeline` is constructed with `enable_rigid_soft_full_surface_contact=True` but a participating mesh has no SDF, it raises `ValueError` — the fix is to call `enable_rigid_mesh_sdfs()` first. The flag lives on `CollisionPipeline` (construction time; it sizes the contact buffers), not on `collide()`.

**VBD integration**: the particle-side and body-side contact kernels each consume the unified stream in a **single launch**, dispatching per record on the `-1` padding — a particle record uses the single-vertex path (bit-identical to before), an edge/face record distributes the contact at `x = Σ bary_i · v_i` as `bary_i · F` (force) and `bary_i² · H` (Hessian) to its soft particles. The per-body kernel applies the equal-and-opposite reaction to dynamic rigid bodies (two-way). All paths reuse the existing penalty force law with the same `penetration > 0` gate and share the AVBD penalty ramp and pre-mixed per-contact material.

**Solver support.** Full-surface contacts are consumed by **`SolverVBD` only**. `CollisionPipeline` is solver-agnostic, so a private capability flag on `Contacts` records that full-surface generation was enabled, and the other soft-contact solvers (XPBD, semi-implicit, Style3D) **raise** if handed such contacts rather than silently misreading edge/face records as particle contacts. With the flag off those solvers are byte-for-byte unchanged from `main`.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_collision_pipeline   # detection (TestFullSurfaceSoftContact): schema, SDF optimizers vs brute-force, edge/face passes, per-shape margin, mesh-SDF provisioning, multi-world pairs, end-to-end no-false-pos/neg, backward-compat, graph-capture
uv run --extra dev -m newton.tests -k test_solver_vbd           # VBD consumption (TestVBDFullSurfaceContact): edge/face pushout, two-way body reaction, barycentric distribution, per-shape material mixing, shape margin, flag-off no-op
uv run --extra examples -m newton.examples vbd_gripper_soft_triangle --viewer null --test   # gripper demo (also vbd_gripper_soft_grid); both registered in test_examples
```

All green on CPU and CUDA (NVIDIA L40); the broader `test_collision_pipeline` (136) + `test_solver_vbd` (73) + `test_softbody` (21) suites also pass (`enable_rigid_soft_full_surface_contact` off leaves rigid/legacy paths byte-for-byte unchanged).

## New feature / API change

```python
import newton

# ... add a soft body (cloth/triangles) and a rigid shape ...

# Opt in to volume SDFs for the rigid meshes that will collide with soft geometry.
# Like color(): it only sets a flag; finalize() builds the SDFs (it owns the device).
# Analytic primitives (box/sphere/...) use their closed-form SDF and are skipped.
builder.enable_rigid_mesh_sdfs()

model = builder.finalize()  # builds the volume SDFs for the participating meshes

# The full-surface flag is set at construction time: it sizes the contact buffers and
# validates that every participating mesh has an SDF (raises ValueError otherwise).
# Consumed by SolverVBD; other solvers raise if given full-surface contacts.
pipeline = newton.CollisionPipeline(model, enable_rigid_soft_full_surface_contact=True)

# Soft EDGES and FACES (not just vertices) now generate contacts, so a soft body
# whose vertices straddle a thin rigid feature can no longer tunnel through it.
solver.step(state_in, state_out, control, contacts, dt)
```

## Summary by CodeRabbit

* **New Features**
  * Added opt-in full-surface rigid–soft EDGE/FACE contact generation with unified feature indices and per-record barycentric weights.
  * Updated collision contact data to expose `soft_contact_indices` and `soft_contact_barycentric` (and richer replay sizing).
  * Extended VBD to consume water-tight EDGE/FACE contacts.
  * Added new VBD gripper examples (EDGE and FACE).
* **Documentation**
  * Updated collision/contact-data docs to describe the unified soft-contact schema.
* **Bug Fixes**
  * Added validation for full-surface mode: unsupported/unsafe configurations now warn or raise, and incompatible solvers fail fast.
* **Tests**
  * Added end-to-end and SDF optimizer regression coverage for EDGE/FACE contact behavior and schema invariants.


AnkaChan:
Thanks for the review — addressed the findings across commits `b840495b..90fd22df`:

- **Per-call `enable_water_tight_rigid_soft_contact` overflow (`collide.py`)** — removed the per-call `collide()` override entirely. The flag is now construction-time only (buffer headroom is sized where the flag lives), and SDF provisioning is an explicit `ModelBuilder.build_rigid_mesh_sdfs()` step; `CollisionPipeline` now raises if a participating mesh lacks an SDF instead of silently dropping records. The sizing/override mismatch can no longer occur. (`71543863`, `90fd22df`)
- **Face cull uses `circumradius` (`soft_contacts_sdf.py`)** — fixed to the max centroid-to-vertex reach (the true 1-Lipschitz bound); removed the now-unused `circumradius()` and added `test_face_cull_uses_max_vertex_reach` (verified it fails against the old cull). (`b9f19f26`)
- **SDF cache keyed only on `id(src)` (`builder.py`)** — now keyed on `(id(src), margin, narrow_band, voxel_size, resolution, texture_format)`, mirroring the rigid SDF cache; scale stays out since the SDF is built `scale_baked=False`. (`89dd5ccb`, `90fd22df`)
- **Document the `finalize` water-tight arg** — the argument was removed from `finalize()` and replaced by the explicit `ModelBuilder.build_rigid_mesh_sdfs()` step, so this is moot; the new method and flag are documented. (`90fd22df`)
- **`test_final` independent of `python -O` (both gripper examples)** — now `raise AssertionError`. (`358fb4f4`)
- **Shape-sized zero-margin baseline in the margin test** — done. (`358fb4f4`)

Not addressed here (out of scope): `newton/_src/utils/mesh.py:795` `spring_indices.shape[0] / 2` → `// 2` — that line is inherited from PR #3194 (git blame → its MeshAdjacency commit) and only appears in this PR's diff because #3194 isn't merged yet. I'll fix it in the #3194 branch.

@coderabbitai review

AnkaChan:
@coderabbitai your summary is out-dated

newton/_src/sim/builder.py
```diff
@@ -9762,6 +9789,24 @@ def color(
...
+        (they need the finalize device) in unscaled mesh space, so scale is applied at query time and
+        one SDF serves all scales of a shared :class:`~newton.Mesh`. They are general-purpose distance
+        fields — any SDF query can use them — but their primary consumer is the water-tight rigid-soft
+        edge/face contact passes.
+
+        Call before :meth:`finalize <ModelBuilder.finalize>` when constructing a
+        :class:`~newton.CollisionPipeline` with ``enable_water_tight_rigid_soft_contact=True``;
+        :meth:`finalize <ModelBuilder.finalize>` does not build these implicitly (mirroring
+        :meth:`color`). Texture SDFs are CUDA-only, so on CPU (or on any per-mesh failure) the shape
+        falls back to the legacy per-particle soft-contact path.
```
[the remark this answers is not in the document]
AnkaChan: Fixed in 387a7292 — reworded both the `enable_rigid_mesh_sdfs` docstring and the finalize SDF-build comment to say the shape's SDF is left unprovisioned and a water-tight `CollisionPipeline` then raises (fail-loud), rather than degrading to the per-particle path. Thanks for the catch!

eric-heiden:
The water-tight contact path is a valuable addition. The main concerns are API compatibility, incomplete deprecations, and correctness of contact generation and differentiation.

newton/_src/sim/contacts.py
```diff
@@ -316,9 +321,23 @@ def __init__(
...
+            # index space, and pack contiguously by kind into three consecutive ranges:
+            #   particle  [0, c0)                              -- legacy vertex-vs-surface pass
+            #   edge      [c0, c0 + n_edge)                    -- water-tight soft-edge pass
+            #   face      [c0 + n_edge, c0 + n_edge + n_face)  -- water-tight soft-face pass
+            # where (c0, n_edge, n_face) = soft_contact_count. A record's kind is implied by which
+            # range its index falls in; there is no separate per-record kind flag.
+            # Soft feature id: particle id in the particle range, soft-triangle id in the
+            # edge/face ranges (renamed from soft_contact_particle).
+            self.soft_contact_primitive = wp.full(soft_contact_max, -1, dtype=int)
+            self.soft_contact_barycentric = wp.zeros(soft_contact_max, dtype=wp.vec3, requires_grad=requires_grad)
```
eric-heiden: 🟠 Changing `soft_contact_count` from the legacy one-element counter to a three-element `[particle, edge, face]` buffer breaks its existing public contract, including code that copies, assigns, or validates its shape. Please keep `soft_contact_count` as the one-element legacy view and introduce a separate field for the three range counts, or provide a staged compatibility alias and deprecation period.

newton/_src/geometry/soft_contacts_sdf.py
```diff
@@ -0,0 +1,529 @@
...
+
+    bary, x, phi, grad = optimize_face_sdf(
+        geo, scale, a_s, b_s, c_s, sdf_idx, texture_sdf_table, sdf_face_iters, sdf_ls_iters
+    )
+    if phi < threshold:
+        y = x - phi * grad
+        base = soft_contact_count[0] + soft_contact_count[1]  # particle + edge counts (both final)
+        _emit_soft_ef_contact(
+            base,
+            2,
```
eric-heiden: 🟠 The new FACE and EDGE detection passes use only `margin + radius` as their threshold, so contacts within the additional per-shape `shape_margin` are culled before VBD can evaluate them. Please pass `model.shape_margin` into both kernels and include `shape_margin[shape_index]` in the conservative culls and final emission checks, matching the legacy particle contact path.

newton/_src/geometry/soft_contacts_sdf.py
```diff
@@ -0,0 +1,529 @@
...
+    (``shape_count * (particle_count + tri_count + edge_count)``), so this only matters if a caller
+    supplies a smaller ``soft_contact_max`` — in which case the overflow flag trips."""
+    idx = base + wp.atomic_add(soft_contact_count, slot, 1)
+    if idx < soft_contact_max:
+        soft_contact_primitive[idx] = prim_id
+        soft_contact_barycentric[idx] = bary
+        soft_contact_shape[idx] = shape_index
+        soft_contact_body_pos[idx] = body_pos
+        soft_contact_body_vel[idx] = body_vel
+        soft_contact_normal[idx] = normal
```
eric-heiden: 🟠 This raw `wp.atomic_add()` does not record the chosen output index for autodiff replay. These kernels are tape-recorded, but unlike the legacy path they neither use `counter_increment()` nor write per-thread counter values, so backward replay can use different dynamic indices or fail. Please add replayable counter storage for the EDGE and FACE passes, or explicitly reject water-tight contacts when differentiable collision is enabled until that backward path is supported.

newton/_src/sim/collide.py
```diff
@@ -580,6 +622,7 @@ def __init__(
         include_static_kinematic_pairs: bool = True,
         soft_contact_max: int | None = None,
         soft_contact_margin: float = 0.01,
+        enable_water_tight_rigid_soft_contact: bool = False,
         requires_grad: bool | None = None,
         broad_phase: Literal["nxn", "sap", "explicit"]
         | BroadPhaseAllPairs
```
eric-heiden: ⚪ Please add `enable_water_tight_rigid_soft_contact` to the constructor's `Args:` block. The entry should explain that this enables the opt-in EDGE/FACE SDF path, that mesh and convex shapes require SDFs provisioned through `ModelBuilder.enable_rigid_mesh_sdfs()`, and that the current solver integration is VBD-specific.

newton/_src/sim/contacts.py
```diff
@@ -415,3 +434,18 @@ def device(self):
...
         """
         return self.rigid_contact_count.device
+
+    @property
+    def soft_contact_particle(self) -> wp.array:
+        """Deprecated alias for :attr:`soft_contact_primitive`.
+
+        .. deprecated::
+            Use :attr:`soft_contact_primitive`, which holds the soft feature id: a particle id in the
+            particle range and a soft-triangle id in the edge/face ranges.
```
eric-heiden: ⚪ Newton's compatibility documentation requires `.. deprecated:: X.Y` to identify when a deprecation began. Please include the current deprecation version here so Sphinx and users have the expected version context.

```suggestion
        .. deprecated:: 1.4
```

newton/_src/geometry/soft_contacts_sdf.py
```diff
@@ -0,0 +1,529 @@
...
+    sdf_sphere,
+    sdf_sphere_grad,
+)
+from .sdf_texture import TextureSDFData, texture_sample_sdf_grad
+from .types import Axis, GeoType
+
+# Fixed iteration counts -> data-independent loops -> CUDA-graph-capturable. Passed as kernel args
+# (uniform across threads/launches). Tuned against a brute-force grid reference
+# (tests/test_water_tight_soft_contact.py): edge golden-section is accurate (~1e-4); the face
+# Frank-Wolfe tail is ~O(1/iters) (~3e-3 at 24 iters), sufficient for contact within margin.
```
eric-heiden, line 41: ⚪ `tests/test_water_tight_soft_contact.py` does not exist. The relevant coverage is in `newton/tests/test_collision_pipeline.py`, under `TestWaterTightSoftContact`; please update this pointer so readers can locate the referenced tests.

eric-heiden:
This PR states to deprecate `MeshAdjacency` but there is no `DeprecationWarning` thrown when using the `MeshAdjacency`. Please add this deprecation handling in code.

mmacklin:
A general comment - can we not call this "watertight" collision detection. That has a lot of overloaded meaning, and this method should even work for planar shapes (which are not watertight).

newton/_src/sim/builder.py
```diff
@@ -1038,6 +1038,9 @@ def __init__(self, up_axis: AxisType = Axis.Z, gravity: float = -9.81):
         self.shape_sdf_padding: list[float | None] = []
         """Per-shape SDF generation margins [m] retained until :meth:`finalize <ModelBuilder.finalize>`.
         When ``None``, :attr:`shape_gap` is used for primitive texture SDF generation."""
+        self._enable_rigid_mesh_sdfs: bool = False
```
mmacklin: @nvtw @eric-heiden does it make sense to do it this way?

We don't have this kind of pattern elsewhere in the builder. It could make sense, but I feel like we have a lot of other ways to control SDF creation already?

Perhaps we can remove this from ths PR and revisit, doesn't seem necessary as long as users create SDFs for collision shapes via some other method?

newton/_src/sim/collide.py
```diff
@@ -580,6 +622,7 @@ def __init__(
         include_static_kinematic_pairs: bool = True,
         soft_contact_max: int | None = None,
         soft_contact_margin: float = 0.01,
+        enable_water_tight_rigid_soft_contact: bool = False,
```
mmacklin: Let's rename this - we shouldn't call this "watertight".

I would just expose options for each type of feature we will generate, e.g.:

```
enable_rigid_soft_particle_contact
enable_rigid_soft_edge_contact
```

newton/_src/sim/contacts.py
mmacklin: Hmm, one way to avoid having the 3 separate counts and ranges would just be to write the particle indices involved directly, e.g.:

```
**particle contact:** 
soft_contact_indices.append(particle_id, -1, -1)
soft_contact_barycentric.append(1.0, 0.0, 0.0)

**edge contact:** 
soft_contact_indices.append(p0, p1, -1)
soft_contact_barycentric.append(u, v, 0.0)

**face contact:** 
soft_contact_indices.append(p0, p1, p2)
soft_contact_barycentric.append(u, v, w)
```

The worst case (face contact) it would write out 2 more values than the current implementation, but would probably be a relatively small overhead.

This way we can keep 1 counter and have things quite unified so there are no separate ranges etc. Just a single set of soft contacts.

mmacklin: Just to summarize, I am proposing that instead of writing `self.soft_contact_primitive` and then indirectly looking up the edge/triangle indices we just store out the indices directly to this structure.

newton/_src/sim/model.py
```diff
@@ -1765,11 +1776,29 @@ def collide(
             contacts: The contacts buffer to populate (will be cleared first). If None, a new
                 contacts buffer is allocated via :meth:`contacts`.
             collision_pipeline: Optional collision pipeline override.
+            enable_water_tight_rigid_soft_contact: When ``True``, additionally run the triangle-driven
```
mmacklin: As mentioned previously, let's just expose flags for the set of features that users want generated and avoid calling this "watertight".

newton/_src/sim/contacts.py
mmacklin: I guess this will have the side-effect that `soft_contact_particle` is no longer valid. In that case I would deprecate it and just write it redundantly from the particle / rigid contact kernel until it's removed while we transition to the `soft_contact_indices`.

AnkaChan: @mmacklin  Just writing the particle ids of the soft side won't work unfortunately. Because it will make digesting the collision information very hard. The solver doesn't just need the ids of vertices on the soft primitives but but also the id of the primitive (edge or triangles) so the adjacency information can be used. This is crucial particularly important for VBD solver. primitive_id -> vertex_id is trivial but the other way around is hard. If we only store `vertex_id` we will lose some information.

AnkaChan: @eric-heiden this is the best way I can think of to make it backward compatible. `soft_contact_count[0]` will still have the same meaning as before, and all previous kernel that handles collision will work without requiring any change. A alternative option will be adding a `soft_edge_contact_count` and `soft_triangle_contact_count`. I am okay with that, we will probably also add a `soft_particle_contact_count` and make `soft_contact_count` alias of it.

newton/_src/geometry/soft_contacts_sdf.py
AnkaChan: good catch

newton/_src/sim/builder.py
AnkaChan: The SDF constructs has to be deferred to `finalize()` stage, since it relies on some information that's only available in the finalization stage. So we have to keep a flag to tell `finalize()` whether to construct SDFs.

I make this design choice to avoid putting a duplicated `enable_water_tight_rigid_soft_contact` parameter to  `finalize()` since it's already in `CollisionPipeline.__init__()`.

This design choice is to mimic `VBDSolver`'s strategy: to use VBD you have to call `Builder.color()`. We explicitly name the function by what it does (coloring), rather than 'prepare_for_vbd_solver()'. Same here, we name this function by what it does thus it's decoupled from a specific subsequent operation.

newton/_src/sim/collide.py
AnkaChan: We also need to add a `enable_rigid_soft_triangle_contact` if we are going to this direction. I cannot think of a reason why the user wants to use edge contact with out triangle or the other way around. To be honest the only reason we keep the legacy path is probably for the compatibility with other solvers.

What about we change this to a flag instead a binary input? We can call it `rigid_soft_contact_mode`, then we have macros: _Rigid_soft_particle_only_contact=0
_Rigid_soft_surface_contact=1
this also opens the door for edge-only of triangle only contact if it's ever needed in the future.

mmacklin: Yeah, I think one option to enable surface also works, e.g.: `enable_rigid_soft_surface_contact` - Maya calls this "full surface" for nCloth.

newton/_src/sim/contacts.py
mmacklin: Shouldn't we access adjacency during the contact generation itself, and not inside the solver? Ultimately the contact energy just depends on particles right?

mmacklin: The other advantage of writing out particle indices directly is that we can use the same data structure for 1D collisions e.g.: hair / strand-based systems without the solver having to have special handling for triangles.

AnkaChan: @mmacklin I agree changing to vertex-id based recording can be more unified. However the 3-separate-counts recording is still crucial for backward compatibility: when collision pipeline is called with `enable_rigid_soft_full_surface_contact`, the `soft_contact_count[0]` will be longer than the particle range. Every existing solver except VBD (XPBD / semi-implicit / Style3D) sizes its loop as min(contact_max, soft_contact_count[0]) and treats each record as a particle contact, with a single total counter they'd silently misread edge/face records whenever the surface flag is on.

To keep a single count, we need to change all the contact digesting code of the existing solvers the least. However having 3 separate counts will make all other solvers backward compatible without any change.

mmacklin: I would just say that `enable_rigid_soft_full_surface_contact` is only supported by VBD, that seems reasonable for a new feature.

mmacklin (dismissed):

newton/_src/geometry/soft_contacts_sdf.py
```diff
@@ -0,0 +1,535 @@
...
+    if geo == GeoType.ELLIPSOID:
+        return sdf_ellipsoid(x_local, scale), sdf_ellipsoid_grad(x_local, scale)
+    if geo == GeoType.PLANE:
+        return sdf_plane(x_local, scale[0] * 0.5, scale[1] * 0.5), wp.vec3(0.0, 0.0, 1.0)
+
+    # Volume SDF (mesh / convex / other). Honor the descriptor's scale_baked flag: if the shape
+    # scale was baked into the grid (e.g. hydroelastic primitives), query directly in shape-local
+    # (= scaled) space; otherwise apply query-time scaling (mirrors mesh_sdf_collision_kernel +
+    # scale_sdf_result_to_world).
+    tex = texture_sdf_table[shape_sdf_index]
```
eric-heiden, line 108: 🔴 A nonnegative `shape_sdf_index` does not guarantee that this descriptor contains valid textures. The builder can assign an index and append `create_empty_texture_sdf_data()`, while pipeline validation checks only for an index below zero; sampling such a descriptor reproduced CUDA error 700. Please keep the shape index unset for empty descriptors or validate the descriptor contents before this lookup and fail on the host.

newton/_src/geometry/soft_contacts_sdf.py
```diff
@@ -0,0 +1,535 @@
...
+    Uses :func:`counter_increment` on the shared soft counter so the chosen index is recorded per
+    thread (in ``soft_contact_tids``) for differentiable backward replay, matching the legacy
+    particle pass. ``tid_base`` offsets this pass's thread ids into the shared tids array so the
+    three passes (particle / edge / face) never alias the same tids slot: particle uses ``[0, ...)``,
+    edge ``[n_particle_pairs, ...)``, face ``[n_particle_pairs + n_edge_pairs, ...)``. The offsets
+    are static (pair counts fixed at pipeline init), so this stays CUDA-graph-capturable.
+
+    The counter is incremented even when the record overflows ``soft_contact_max`` (the write is
+    guarded); ``counter_increment`` returns -1 in that case."""
+    idx = counter_increment(soft_contact_count, 0, soft_contact_tids, tid + tid_base, soft_contact_max)
```
eric-heiden, line 273: 🟠 `soft_contact_tids` has length `soft_contact_max`, but `tid + tid_base` ranges over the full particle, edge, and face candidate-pair space. With a custom capacity smaller than that space, an accepted edge/face thread can address beyond the replay array; with `soft_contact_max=1`, a generated face record retained `soft_contact_tids[0] == -1`. Please size replay storage for the candidate space, use separate per-pass replay arrays, or reject this differentiable configuration.

newton/_src/geometry/soft_contacts_sdf.py
```diff
@@ -0,0 +1,535 @@
...
+
+    # Volume SDF (mesh / convex / other). Honor the descriptor's scale_baked flag: if the shape
+    # scale was baked into the grid (e.g. hydroelastic primitives), query directly in shape-local
+    # (= scaled) space; otherwise apply query-time scaling (mirrors mesh_sdf_collision_kernel +
+    # scale_sdf_result_to_world).
+    tex = texture_sdf_table[shape_sdf_index]
+    if tex.scale_baked:
+        return texture_sample_sdf_grad(tex, x_local)
+    inv_scale = wp.vec3(1.0 / scale[0], 1.0 / scale[1], 1.0 / scale[2])
+    min_scale = wp.min(scale)
```
eric-heiden: 🟠 `wp.min(scale)` becomes negative for a mirrored mesh, so it flips the sampled SDF sign: with scale `(-1, 1, 1)`, an outside distance of `+0.2` becomes `-0.2`. Newton supports negative mesh scales, and inverted mesh winding can reverse the generated SDF sign as well. Please normalize distance magnitude and gradient orientation for these supported mesh configurations, and add mirrored and inverted-mesh regressions.

newton/_src/solvers/vbd/vbd_coupling_kernels.py
```diff
@@ -302,6 +302,8 @@ def _harvest_vbd_proxy_particle_body_contact_forces_kernel(
         return
 
     particle_idx = body_particle_contact_particle[contact_idx]
+    if particle_idx < 0:  # edge/face record: no single particle id (harvested via the per-body path)
```
eric-heiden, line 305: 🟠 This return drops edge/face force feedback for proxy particles. The comment says the per-body path harvests those records, but the coupled soft-contact filter still rejects `soft_contact_particle < 0` and copies neither `soft_contact_indices` nor `soft_contact_barycentric`, so the records are gone before that path can use them. Please update both the coupled filter and proxy harvesting to preserve and consume the unified records instead of silently losing full-surface contacts.

newton/_src/geometry/soft_contacts_sdf.py
```diff
@@ -0,0 +1,535 @@
...
+            sdf_cylinder_grad(x_local, scale[0], scale[1], int(Axis.Z)),
+        )
+    if geo == GeoType.CONE:
+        return (
+            sdf_cone(x_local, scale[0], scale[1], int(Axis.Z)),
+            sdf_cone_grad(x_local, scale[0], scale[1], int(Axis.Z)),
+        )
+    if geo == GeoType.ELLIPSOID:
+        return sdf_ellipsoid(x_local, scale), sdf_ellipsoid_grad(x_local, scale)
+    if geo == GeoType.PLANE:
```
eric-heiden, line 100: 🟠 Full-surface shape coverage is incomplete here. Heightfields have no dispatch case and therefore emit no edge/face contacts, while finite planes use the bounded distance but always return a `+Z` normal, which produced off-quad contacts with the wrong normal; meshes enabled through runtime flags can also reach the passes without a provisioned SDF and be skipped. Please implement these cases correctly or exclude and fail loudly for unsupported shape configurations when building or validating candidate pairs.

newton/_src/sim/collide.py
```diff
@@ -1381,3 +1476,19 @@ def collide(
                 ],
                 device=self.device,
             )
+
+        # Full-surface EDGE/FACE passes (opt-in, set at construction): add the soft edge/face contacts
+        # the per-particle path cannot detect. Run after the legacy particle launch on the same stream;
+        # the particle records therefore occupy [0, particle_count) and the edge/face records append.
+        # The flag is fixed at construction because soft_contact_max headroom is sized there.
+        if self.enable_rigid_soft_full_surface_contact and state.particle_q:
```
eric-heiden, line 1597: 🟠 This launches full-surface detection into any caller-provided `Contacts`, but the capability marker is set only by `CollisionPipeline.contacts()`. Reusing a buffer created by a flag-off pipeline here can populate edge/face records while leaving `_enable_rigid_soft_full_surface_contact` false, so XPBD and other particle-only solvers do not raise and silently ignore those contacts. Please set or validate the marker on every `collide()` call, or reject an incompatible external buffer before launching these passes.

AnkaChan:
Thanks for the thorough review, @eric-heiden — all six items addressed. Each has a regression test that fails without the fix. The fixes landed in `b21c9ae1`; the branch is now rebased onto current main.

### 1. Empty SDF descriptor -> CUDA 700 (`soft_contacts_sdf.py:103`)
The construction guard only rejected `shape_sdf_index < 0`, but the mesh-mesh path can point a nonnegative index at an empty placeholder descriptor (a BVH fallback). The guard now also treats a descriptor whose coarse texture is `None` as unprovisioned and fails on the host before any launch. Test: `test_full_surface_empty_sdf_descriptor_rejected`.

### 2. Replay array undersized (`soft_contacts_sdf.py:256`)
`soft_contact_tids` was sized to `soft_contact_max` but is indexed by the global launch-thread id over the full particle+edge+face candidate space, so a smaller custom `soft_contact_max` dropped a thread's replay slot. Added a `soft_contact_tids_size` argument on `Contacts`; the pipeline sizes the replay array to the candidate count, decoupled from capacity. This also closes the same latent issue on the legacy particle-only path. Tests: `test_full_surface_replay_spans_candidate_space`, `test_soft_contact_tids_decoupled_from_capacity`.

### 3. Mirrored-mesh sign flip (`soft_contacts_sdf.py:107`)
`wp.min(scale)` went negative for a mirrored scale and flipped the SDF sign. Changed to `wp.min(wp.abs(scale))` for the magnitude; the mirror itself is already applied by the query-time division and by `inv_scale` on the gradient. Test: `test_eval_shape_sdf_mirrored_mesh_scale_preserves_sign`. (Inverted mesh winding is a build-time sign concern in `build_sdf`, out of scope here.)

### 4. Incomplete shape coverage (`soft_contacts_sdf.py:96`)
Took the fail-loud route rather than partial support: the construction guard now rejects participating heightfields (no SDF dispatch) and finite planes (the `+Z` normal is correct only for an infinite plane) with a clear `NotImplementedError`; infinite planes still work. Tests: `test_full_surface_rejects_finite_plane`, `test_full_surface_allows_infinite_plane`. The mesh-without-SDF case is covered by the empty-descriptor guard in (1).

### 5. Proxy coupling drops edge/face (`vbd_coupling_kernels.py:305`)
Deferred behind a loud guard: `coupling_prepare_proxy_contacts` now raises `NotImplementedError` when the contacts buffer carries full-surface records, so the cross-solver wrench harvest can't silently drop edge/face feedback. Standalone `SolverVBD` is unaffected — the per-body reaction (`accumulate_body_particle_contacts_per_body` -> `_eval_soft_ef_contact`) already consumes edge/face in both standalone and coupled runs; only the multi-solver `SolverCoupledProxy` feedback path is missing it. Full proxy-harvest support is a scoped follow-up PR. Test: `test_full_surface_rejected_by_vbd_proxy_coupling`.

### 6. Capability marker not set on external buffer (`collide.py:1484`)
`collide()` now sets `contacts._enable_rigid_soft_full_surface_contact` to match the pipeline on every call, so a buffer created elsewhere (or by a flag-off pipeline) can't receive edge/face records while the marker stays `False` and particle-only solvers fail to refuse it. Test: `test_collide_syncs_full_surface_marker`.

Separately, I rebased onto current main (which picks up the Featherstone fix that clears an unrelated flaky GPU test) and added particle + full-surface gap-differentiability tests in `test_differentiable_contacts.py`.

newton/_src/geometry/soft_contacts_sdf.py
AnkaChan: Update on this, revising what I said in the summary: rather than a hard reject when an unsupported shape participates, unsupported shape *types* now **warn and fall back to per-particle soft contact** — the shape is excluded from the edge/face candidate pairs, so full-surface keeps working for the rest of the scene instead of one shape disabling the whole pass. This covers heightfields (they expose only a per-cell local-plane distance, discontinuous across cells and unsuitable for the edge/face SDF optimizers), finite planes (the `+Z` normal is correct only for an infinite plane), and Gaussian splats.

A participating mesh/convex *without* an SDF still raises a `ValueError`, as you asked — that's a fixable provisioning mistake (`configure_sdf(force_sdf=True)`), not an inherent shape limitation. (3c0b613d)

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -2475,14 +2492,13 @@ def _solve_particle_iteration(
                         # body-particle contact
                         self.friction_epsilon,
                         model.particle_radius,
-                        contacts.soft_contact_particle,
+                        contacts.soft_contact_indices,
```
eric-heiden, line 2495: 🟠 This makes coupled VBD regress even when full-surface contact is disabled. `SolverCoupled._filter_contacts_for_solver()` still copies only `soft_contact_particle`, leaving `soft_contact_indices` and `soft_contact_barycentric` at their sentinel values, but VBD now consumes the unified fields here; in a particle-only probe, the retained contact reached VBD as `(-1, -1, -1)`, and an edge/face row was dropped. Please copy the unified fields and preserve generalized records in the coupled filter, or reconstruct a valid unified particle record before launching this kernel.

newton/_src/geometry/soft_contacts_sdf.py
```diff
@@ -0,0 +1,538 @@
...
+    # (= scaled) space; otherwise apply query-time scaling (mirrors mesh_sdf_collision_kernel +
+    # scale_sdf_result_to_world).
+    tex = texture_sdf_table[shape_sdf_index]
+    if tex.scale_baked:
+        return texture_sample_sdf_grad(tex, x_local)
+    inv_scale = wp.vec3(1.0 / scale[0], 1.0 / scale[1], 1.0 / scale[2])
+    # Magnitude of the smallest scale factor. Use abs() so a mirrored/negative mesh scale (e.g.
+    # (-1, 1, 1)) does not flip the SDF sign; the mirror itself is already applied by the query-time
+    # cw_div below and by inv_scale on the gradient.
+    min_scale = wp.min(wp.abs(scale))
```
eric-heiden: 🟠 `min_scale` is only a conservative distance bound under nonuniform scaling, not the true world-space SDF, so using it for both the contact threshold and `y = x - phi * grad` creates ghost contacts and off-surface points. With mesh scale `(2, 1, 1)`, a triangle `0.08 m` outside a box with a `0.06 m` margin still emitted an edge/face contact and projected to `x = 1.04` instead of the actual surface at `x = 1.0`. Please compute an exact scaled distance and closest point for nonuniform transforms, or reject nonuniform mesh scales for this full-surface path until that calculation is supported.

newton/_src/solvers/vbd/solver_vbd.py
AnkaChan: Fixed in `4b9bfcd4`. `_filter_soft_contacts_global_shape_ids_kernel` now threads `soft_contact_indices` and `soft_contact_barycentric` through the coupled filter and copies them at the compacted slot, alongside the fields it already carried. The particle-only path writes `(p, -1, -1)` + `(1, 0, 0)`, so a retained contact now reaches VBD as a valid record instead of the `(-1, -1, -1)` sentinel — restoring coupled VBD with full-surface off. Regression: `test_soft_contact_filter_preserves_unified_fields` (drops to `(-1, -1, -1)` without the copy).

newton/_src/geometry/soft_contacts_sdf.py
AnkaChan: Fixed in `7cb8201c`. Rather than rejecting nonuniform scales, `eval_shape_sdf` now converts the shape-local distance to world by the stretch *along the surface normal*, `dist * |(sx*gx, sy*gy, sz*gz)|` — exact for an axis-aligned face, first-order near the surface where contacts live. Your `(2, 1, 1)` probe now reports the true 0.08 m gap (no contact under the 0.06 m margin) and projects onto `x = 1.0`, not 1.04.

`min_scale` stays as a conservative lower bound driving the candidate cull and the optimizer's internal search (an under-estimate can never drop a real contact); the accurate distance is used only for the `phi < threshold` accept and the `x - phi*grad` projection — conservative in the query, exact in the narrow phase — so a nonuniform mesh keeps full-surface contact instead of being rejected. `eval_shape_sdf` now returns `(phi_lower, phi, grad)` accordingly. Regression: `test_full_surface_nonuniform_mesh_accurate_distance` (min_scale under-reports 0.04 → ghost, and projects to ~1.015).

eric-heiden (approved):
Thanks!

PR #3268, opened 2026-06-29
[Kamino] Small fix to PADMM Unit Tests

vastsoun:
## Description

Fixes broken PADMM UT when verbose output and plotting are enabled.

## Summary by CodeRabbit

* **Tests**
  * Renamed the PADMM single-contact unit test and updated the saved output filename accordingly.
  * Updated PADMM test setup to collect solver information only when saving is enabled.
  * Tightened solver plot generation to run only when solver info exists and saving is turned on.

### DIFF
--- a/newton/_src/solvers/kamino/tests/test_solvers_padmm.py
+++ b/newton/_src/solvers/kamino/tests/test_solvers_padmm.py
@@ -401,7 +401,7 @@ class TestPADMMSolver(unittest.TestCase):
             config=config,
             warmstart=PADMMWarmStartMode.NONE,
             use_acceleration=False,
-            collect_info=True,
+            collect_info=self.savefig,
         )
 
         # Solve the test problem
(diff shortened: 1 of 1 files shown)


newton/_src/solvers/kamino/tests/test_solvers_padmm.py
```diff
@@ -666,7 +666,7 @@ def test_10_padmm_solve_single_contact(self):
         test = TestSetup(builder_fn=basics.build_box_on_plane, max_world_contacts=1, device=self.default_device)
 
         # Create the PADMM solver
-        solver = PADMMSolver(model=test.model)
+        solver = PADMMSolver(model=test.model, collect_info=True)
```
Guirec-Maloisel: Is `collect_info` needed in all cases, or just when one tries to create a plot? In other words, should this be conditioned e.g. on `self.savefig` instead of always `True`?

vastsoun: Fair point. Fixed in latest commit.

Guirec-Maloisel (approved):
LGTM!

chschuma-disney (approved):

PR #3270, opened 2026-06-29
Exclude kinematic armature from mean inertia

adenzler-nvidia:
## Description

`SolverMuJoCo` represents kinematic bodies using a large synthetic DOF armature. MuJoCo and MJWarp include this armature when calculating `meaninertia`, causing the locking value to dominate the statistic and loosen solver convergence tolerances for the entire model.

Calculate `meaninertia` using the authored armature, then restore the kinematic locking armature and recompute the remaining model constants. Preserve `actuator_biasprm` between the two passes because constant calculation resolves dampratio placeholders in place.

This applies during initial model conversion and subsequent model-property updates in both the MJWarp and native MuJoCo execution paths.

## Checklist

## Test plan

```bash
.venv/bin/python -m newton.tests -k TestMuJoCoSolverKinematicBodyProperties
.venv/bin/python -m newton.tests -k TestActuatorDampratio
uvx pre-commit run -a
```

## Bug fix

**Steps to reproduce:**

1. Create an articulated model containing a kinematic body with mapped DOFs.
2. Construct `SolverMuJoCo`.
3. Inspect `mjw_model.stat.meaninertia`.
4. Observe that the synthetic locking armature dominates the statistic instead of reflecting the model's physical generalized inertia.

The regression coverage includes initial conversion, runtime kinematic transitions, body-mass and authored-armature updates, native MuJoCo and MJWarp execution, and preservation of dampratio resolution.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `SolverMuJoCo` convergence-related tolerance/scaling issues for models containing kinematic bodies.
  * Improved kinematic-body armature handling so kinematic DOFs use the correct armature behavior during body-property updates.
  * Recalculated MuJoCo constants using physical mean inertia excluding kinematic-locking armature, preserving solver statistics and actuator biasprm resolution across backends.
* **Tests**
  * Added regression tests covering kinematic mean-inertia stability and correct actuator parameter preservation across both supported MuJoCo solver backends.

### DIFF
--- a/newton/_src/solvers/mujoco/solver_mujoco.py
+++ b/newton/_src/solvers/mujoco/solver_mujoco.py
@@ -3878,7 +3878,7 @@ class SolverMuJoCo(SolverBase):
                 self.mj_model.jnt_range[:] = self.mjw_model.jnt_range.numpy()[0]
                 self.mj_model.jnt_actfrcrange[:] = self.mjw_model.jnt_actfrcrange.numpy()[0]
             if need_length_range or need_const_fixed or need_const_0:
-                self._mujoco.mj_setConst(self.mj_model, self.mj_data)
+                self._set_const_0_with_physical_meaninertia()
             if need_solref_update:
                 # ``mj_setConst`` refreshes the derived ``dof_invweight0``
                 # factors; ``jnt_solimp`` was already written by
@@ -3909,7 +3909,7 @@ class SolverMuJoCo(SolverBase):
                     if need_const_fixed:
                         self._mujoco_warp.set_const_fixed(self.mjw_model, self.mjw_data)
                     if need_const_0:
-                        self._mujoco_warp.set_const_0(self.mjw_model, self.mjw_data)
+                        self._set_const_0_with_physical_meaninertia()
                     if need_solref_update:
                         # ``set_const_0`` refreshes ``dof_invweight0`` and
                         # ``jnt_solimp`` was already written by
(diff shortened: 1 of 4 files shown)


newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -6648,7 +6648,41 @@ def _update_model_inertial_properties(self):
...
+            self.mjw_model.stat.meaninertia.fill_(physical_meaninertia)
+        else:
+            actuator_biasprm = wp.clone(self.mjw_model.actuator_biasprm)
+            self._mujoco_warp.set_const_0(self.mjw_model, self.mjw_data)
+            physical_meaninertia = wp.clone(self.mjw_model.stat.meaninertia)
+
+            self._update_body_properties()
+            wp.copy(self.mjw_model.actuator_biasprm, actuator_biasprm)
+            self._mujoco_warp.set_const_0(self.mjw_model, self.mjw_data)
+            wp.copy(self.mjw_model.stat.meaninertia, physical_meaninertia)
```
vreutskyy, line 6682: `mj_model.stat.meaninertia` remains stale on the MJWarp path. This branch restores the physical statistic only to `mjw_model`; after a runtime `BODY_INERTIAL_PROPERTIES` update, my reproduction changed the MJWarp value from approximately `1.716` to `2.573`, while the native model stayed at `1.716`. Please mirror `physical_meaninertia` back to `self.mj_model.stat.meaninertia` here and add a cross-model assertion to the regression test.

vreutskyy (commented):
One actionable consistency issue found; the targeted regression suites otherwise pass.

newton/_src/solvers/mujoco/solver_mujoco.py
adenzler-nvidia: Thanks for checking. Cross-model equality is not an invariant we maintain here. Runtime updates are backend-local: on the MJWarp path, fields such as `mj_model.body_mass`, `body_inertia`, and `dof_armature` already remain unchanged, while the corresponding `mjw_model` fields are updated. Copying only the derived `meaninertia` value would create a partial mirror that is otherwise inconsistent with the runtime update design.

`step()` uses only the selected backend, and the MJWarp path does not read `mj_model.stat.meaninertia`. I pushed `dc8ace4ed` to make this explicit in both directions: the CPU branch now restores the statistic only on `mj_model`, while the MJWarp branch restores it only on `mjw_model`. The regression tests intentionally check the active backend rather than asserting cross-model equality.

vreutskyy: Thanks, that makes sense. I agree that the models are backend-local and synchronizing only `meaninertia` would create a misleading partial mirror. The follow-up commit makes the policy consistent.

vreutskyy (approved):
Looks good. The backend-local mean-inertia handling is consistent, and the targeted regression suites and CI pass.

PR #3275, opened 2026-06-29
feat(usd): import NewtonJointAPI per-joint attributes

### LINKED ISSUE #3271: Add Newton Importer Support for USD NewtonJointAPI
The newton importer needs to support the new NewtonJointAPI USD schema https://github.com/newton-physics/newton-usd-schemas/pull/72 

If a specific solver does not support some values or configuration in the schema, the user should receive an appropriate warning.

andrewkaufman:
## Summary

Import the six `NewtonJointAPI` schema attributes (from `newton-usd-schemas` v0.4.0) and broadcast them uniformly across all DOFs of revolute, prismatic, and D6 joints.

| Schema Attribute | Resolved Key | Notes |
|---|---|---|
| `newton:armature` | `armature` | Resolved via schema resolver priority order |
| `newton:damping` | `damping` | Passive joint damping (Newton-specific) |
| `newton:friction` | `friction` | Resolved via schema resolver priority order |
| `newton:limitStiffness` | `limit_ke` | `inf` = hard limit (1e8), `-inf` = builder default |
| `newton:limitDamping` | `limit_kd` | Ignored when `limitStiffness = inf`; `-inf` = builder default |
| `newton:velocityLimit` | `velocity_limit` | `inf` = builder default |

## Changes

**`pyproject.toml` + `uv.lock`**
- Bump `newton-usd-schemas` dependency to `>=0.4.0` (official release with `NewtonJointAPI` / `NewtonArticulationRootAPI` schemas)

**`newton/_src/usd/schemas.py`**
- Register six `NewtonJointAPI` attrs in the Newton schema resolver (JOINT prim type)
- Add `_newton_legacy_joint_limit_attr()` deprecation getter for 10 legacy per-DOF limit attrs (`limitStiffness`/`limitDamping` per axis → replaced by broadcast schema attrs)
- Add `_newton_non_schema_joint_state_attr()` warning getter for 10 per-DOF state attrs (non-schema, warns users to file an issue)

**`newton/_src/utils/import_usd.py`**
- Read broadcast `armature`/`damping`/`friction`/`limit_ke`/`limit_kd`/`velocity_limit` from the schema resolver
- Add `_resolve_newton_limit_ke()` and `_resolve_newton_limit_kd()` helpers for sentinel handling (+inf → hard limit, -inf → engine default)
- Apply to all three joint-parsing paths (revolute/prismatic, D6 linear, D6 angular) and the MJCF sibling-joint proxy path
- Newton broadcast values layer on top of per-DOF fallbacks while preserving solref source tracking

**`docs/concepts/usd_parsing.rst`**
- Document NewtonJointAPI attribute remapping table

**Tests**
- `test_import_usd.py`: 5 new tests (basic revolute, prismatic, D6 mixed, hard-limit, velocity_limit=inf)
- `test_schema_resolver.py`: comprehensive cross-resolver ordering tests (all 6 permutations of Newton/PhysX/MuJoCo), deprecation warning tests

## Closes

Closes #3271

## Summary by CodeRabbit

* **New Features**
  * Improved USD joint import support for more joint settings, including limits, damping, friction, armature, and velocity limits.
  * Added support for articulation joint model settings during import.
  * Joint values are now applied consistently across all degrees of freedom.

* **Bug Fixes**
  * Better handling of unlimited velocity limits and extreme limit values during import.
  * Legacy joint settings now map more reliably, with clearer warnings for outdated or non-schema values.

* **Documentation**
  * Expanded USD parsing docs with joint attribute remapping details and import behavior notes.


newton/tests/test_schema_resolver.py
```diff
@@ -862,11 +862,13 @@ def test_humanoid_d6_joint_state_initialization(self):
...
-            schema_resolvers=[SchemaResolverNewton(), SchemaResolverPhysx()],
-            verbose=False,
-        )
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", UserWarning)
+            builder.add_usd(
+                source=str(humanoid_path),
+                schema_resolvers=[SchemaResolverNewton(), SchemaResolverPhysx()],
+                verbose=False,
+            )
```
andrewkaufman, line 871: @Milad-Rakhsha-NV I want to understand this more. This test is changed because I added a `UserWarning` for those multi-apply joint state attributes that have no associated schema. I didn't know why they exist in the importer, but it seems we do have assets (this humanoid at least) using them... Do we need a JointState schema? Or was this an internal experiment?

andrewkaufman:
@coderabbitai re-review please

newton/_src/utils/import_usd.py
```diff
@@ -1312,6 +1367,7 @@ def parse_joint(
                 joint_params["limit_upper"] *= DegreesToRadian
                 joint_params["limit_ke"] /= DegreesToRadian
                 joint_params["limit_kd"] /= DegreesToRadian
+                joint_params["damping"] /= DegreesToRadian
```
adenzler-nvidia: The fallback here is `builder.default_joint_cfg.damping`, which is already stored in per-radian units, but it is divided by `DegreesToRadian` unconditionally. As a result, a revolute joint with no authored `newton:damping` changes a customized default of `3.0` to `171.887...`. The same unit regression occurs for D6 angular DOFs and merged revolute joints. Please distinguish authored per-degree damping from the builder fallback (or pre-scale the angular fallback before this conversion), and add a regression test with a nonzero customized default.

newton/_src/usd/schemas.py
```diff
@@ -160,6 +285,7 @@ class SchemaResolverNewton(SchemaResolver):
         PrimType.BODY: {},
         PrimType.ARTICULATION: {
             "self_collision_enabled": SchemaAttribute("newton:selfCollisionEnabled", True),
+            "joint_model": SchemaAttribute("newton:jointModel", "maximal"),
```
adenzler-nvidia: Can this `joint_model` mapping and the corresponding changelog mention be removed? `newton-usd-schemas` 0.4.0 does not define `newton:jointModel`; newton-physics/newton-usd-schemas#72 replaced it before release with the boolean `newton:jointsAddMobility`. The token was replaced because `"maximal"`/`"reduced"` conflated authored joint semantics with a solver's internal representation, while `jointsAddMobility` directly describes the intended behavior. I also cannot find anywhere in the importer that resolves `joint_model`, so this mapping currently appears unused. If articulation mobility support is intended in this PR, should it instead consume `newton:jointsAddMobility`?

andrewkaufman: Good catch, and bad on my agent for not updating this. However, afaik `newton:jointsAddMobility` isn't actually needed here. The parser already handles the case it is intended to identify (it was added in 1.3 iirc). The flag exists in the schema so engine's that cannot handle this case correctly yet (i.e. physx) can warn about it.

I'll just drop the line entirely for now.

andrewkaufman: Dropped in afc9deca — removed the mapping and the changelog mention entirely.

newton/_src/utils/import_usd.py
andrewkaufman: Fixed in c15cf081 — damping is now resolved with `default=None` first; only converted via `/ DegreesToRadian` when the value was actually authored from USD. When falling back to `builder.default_joint_cfg.damping` it's used as-is (already per-radian). Regression test added.

newton/_src/usd/schemas.py
```diff
@@ -91,29 +127,118 @@ class SchemaResolverNewton(SchemaResolver):
...
+            "limit_rotY_kd": SchemaAttribute(
+                "newton:rotY:limitDamping",
+                None,
+                usd_value_getter=_newton_legacy_joint_limit_attr("newton:rotY:limitDamping", "newton:limitDamping"),
+            ),
+            "limit_rotZ_kd": SchemaAttribute(
+                "newton:rotZ:limitDamping",
+                None,
+                usd_value_getter=_newton_legacy_joint_limit_attr("newton:rotZ:limitDamping", "newton:limitDamping"),
+            ),
```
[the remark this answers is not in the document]
andrewkaufman: Not adding these — `limit_transX/Y/Z` are resolved through the PhysX/MuJoCo resolvers (which already define them), and no asset authors a `newton:transX:*` namespace. A Newton-side alias here would be dead code that nothing reads. Reverted.

newton/tests/test_schema_resolver.py
Milad-Rakhsha-NV: Are you referring to the PhysicsJointStateAPI ? That is a physX schema 
https://docs.omniverse.nvidia.com/kit/docs/omni_physics/107.2/dev_guide/schemas/physxschema.html

andrewkaufman: No, this is about changes on `SchemaResolverNewton`

Milad-Rakhsha-NV: I see, yes the humanoid and other assets do have these physx schemas. They can be used more as initial conditioning of the robot through joint position (so doing forward kinematics would give the link transforms), but they aren't used in newton atm, I believe, so I am not worried about `newton:*:position` but what I do wonder about is how we use the ke/kd terms for multi-DOF joints?

andrewkaufman: > how we use the ke/kd terms for multi-DOF joints?

That is explained in the [schema PR](https://github.com/newton-physics/newton-usd-schemas/pull/72) (search for "Design choices" for rationalle). For multi-DOF joints you can use `NewtonJointAPI` directly if you don't need per-DOF variance. If you do need per-DOF variance you need to "break apart" the joint into a stack of 1-DOF joints & you should set `NewtonArticulationRootAPI.jointsAddMobility` on articulations that do this.

>  I am not worried about newton:*:position

Ok, so should I remove those from this humanoid test case or keep the test supressing these `UserWarning`?

newton/_src/utils/import_usd.py
```diff
@@ -56,6 +56,45 @@
...
+        return _HARD_LIMIT_KE, "force"
+    return limit_ke, "force"
+
+
+def _resolve_newton_limit_kd(
+    limit_ke: float | None, limit_kd: float | None, fallback: float, fallback_source: str
+) -> tuple[float, str]:
+    """Resolve a NewtonJointAPI ``newton:limitDamping`` value.
+
+    Hard limits (``limit_ke`` or ``limit_kd`` == ``+inf``) have no damping. Otherwise
```
eric-heiden: 🟠 An authored `newton:limitStiffness = -inf` or `newton:limitDamping = -inf` must select the engine default, but this path treats it the same as a missing attribute and returns `fallback`. Because that fallback may already contain a PhysX or MuJoCo per-DOF gain, a lower-priority schema can override the explicit Newton sentinel. Please preserve whether each Newton attribute was authored and select the builder default for an authored `-inf`, with precedence regression tests for both stiffness and damping.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Add `CollisionPipeline.soft_rigid_contact_pair_count` for the number of precomputed soft-rigid (particle-shape) candidate pairs, filtered to compatible worlds, launched for soft-contact generation; this is the default capacity for `soft_contact_max`
 - Add user-defined pressure laws to hydroelastic SDF contact via `HydroelasticSDF.Config.pressure_func` (a `@wp.func` mapping `(signed_depth, shape_idx, data) -> pressure`) and `pressure_data` (a `@wp.struct` carrying per-shape state). The contact patch is the iso-pressure surface `p_a == p_b`; the default linear law `pressure = -kh * signed_depth` is preserved when no callback is supplied.
 - Add `SensorTiledCamera.utils.assign_checkerboard_material(shape_indices=...)` for applying the checkerboard texture to selected shapes.
+- Add USD import support for `NewtonJointAPI` (`newton:armature`, `newton:damping`, `newton:friction`, `newton:velocityLimit`, `newton:limitStiffness`, `newton:limitDamping`). Attributes broadcast uniformly to every DOF on the joint. If per DOF variance is required, recommendation is to break apart into 1-DOF (i.e. revolute & prismatic) joints instead.
```
eric-heiden, line 19: ⚪ This changelog entry does not document the new warnings for attributes that were previously resolved silently. Please add a `Deprecated` entry directing users from per-DOF `newton:{axis}:limitStiffness` and `limitDamping` attributes to the broadcast `newton:limitStiffness` and `newton:limitDamping` attributes, including the multi-DOF migration guidance. Also add a `Changed` note that unsupported `newton:{axis}:position` and `velocity` attributes now emit a `UserWarning`.

newton/_src/utils/import_usd.py
andrewkaufman: Fixed in `43c9c1b1`. Changed the schema mapping default for `limit_ke`/`limit_kd` from `-inf` to `None`, so `get_value()` now returns:
- `None` when not authored → falls through to per-DOF fallback from lower-priority schemas
- `-inf` only when explicitly authored → selects builder default (engine default)
- `+inf` → hard limit
- finite value → used directly

No sentinel object needed — `None` naturally distinguishes "unset" from "authored `-inf`".

Precedence regression tests:
- `test_newton_limit_sentinel_precedence_over_mjc`: authored `-inf` with MuJoCo `solreflimit` present → builder defaults win.
- `test_newton_limit_unset_falls_through_to_mjc`: no Newton attrs authored → MuJoCo gains flow through.

CHANGELOG.md
andrewkaufman: Added in `43c9c1b1`:

**Deprecated:**
- Per-DOF `newton:{axis}:limitStiffness` and `newton:{axis}:limitDamping` attrs (where axis = linear/angular/rotX/rotY/rotZ) → use broadcast `newton:limitStiffness` / `newton:limitDamping` from NewtonJointAPI. Migration guidance: the broadcast value applies uniformly; for per-DOF variance, split into 1-DOF joints.

**Changed:**
- Non-schema per-DOF initial state attrs (`newton:{axis}:position`, `newton:{axis}:velocity`) now emit a `UserWarning` when encountered during import.

jcarius-nv:
@Milad-Rakhsha-NV @eric-heiden any more concerns or issues here? Otherwise I suggest we get this in by end of day so it's on main before the codefreeze cutoff tomorrow.

newton/_src/utils/import_usd.py
```diff
@@ -56,6 +56,72 @@
...
+    Hard limits (``limit_ke`` or ``limit_kd`` == ``+inf``) have no damping.
+    An authored ``-inf`` selects the builder default (engine default), taking
+    precedence over per-DOF fallbacks from lower-priority schemas.
+    When neither Newton attribute is authored (``None``), the per-DOF ``fallback``
+    from other resolvers is used.
+
+    Returns (resolved_value, source) where source is ``"force"`` when Newton
+    broadcast values are used, or the original ``fallback_source`` otherwise.
+    """
+    # Hard-limit stiffness or damping → zero damping.
```
Milad-Rakhsha-NV: `Hard-limit stiffness or damping → zero damping.` 
The intention isn't super clear to me here, or maybe a better way to write this is just to say `hard-limit > zero damping`?

Milad-Rakhsha-NV:
> @Milad-Rakhsha-NV @eric-heiden any more concerns or issues here? Otherwise I suggest we get this in by end of day so it's on main before the codefreeze cutoff tomorrow.

I talked to Andrew about this offline; I wish we could get rid of things like this to make things simpler 
`
                usd_value_getter=_newton_legacy_joint_limit_attr(
                    "newton:linear:limitStiffness", "newton:limitStiffness"
                ),
`
since the newton:linear:limitStiffness were not associated with a formal schema, so perhaps being backward compatible isn't super critical here, but I leave this to Eric/Andrew to decide.

Milad-Rakhsha-NV (dismissed):

andrewkaufman:
Addressed @Milad-Rakhsha-NV's comment — reworded the hard-limit damping comment to: `# Hard (rigid) limit: infinite ke or kd means no dissipation is needed.` which explains *why* damping is zero rather than just restating the code.

adenzler-nvidia (approved):
Re-reviewed at b59844f94. All open threads are addressed in code: the damping unit conversion is now gated on authored values in all three joint paths (revolute, D6 angular, merged) with regression tests, the `None` mapping defaults let an authored `-inf` select the builder default while unauthored attrs still fall through to PhysX/MuJoCo per-DOF gains (tested in both directions), and the unused `joint_model` mapping is gone. LGTM.

PR #3276, opened 2026-06-30
Add SensorBatchedCamera and deprecate SensorTiledCamera

daniela-hase:
## Description

Add `SensorBatchedCamera` as the new public raytraced camera sensor and deprecate `SensorTiledCamera` in favor of explicit per-view batching.

- Expose `newton.sensors.SensorBatchedCamera` with `camera_indices` mapping rendered views to `(world, ray bundle)` pairs.
- Updated batched output helpers for color, depth, shape index, normal, albedo, HDR color, RGBA conversion, and forward-depth conversion.
- Move shared raytrace internals under `batched_camera_renderer` while keeping legacy tiled-camera compatibility paths.
- Update the sensor example, README entry, API docs, concepts docs, visualization guide, benchmark, and changelog.
- Keep `SensorTiledCamera` available with a deprecation warning and updated compatibility tests.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_sensor_batched_camera
uv run --extra dev -m newton.tests -k test_sensor_to_rgba
uv run --extra dev -m newton.tests -k test_sensors.example_sensor_batched_camera
```

## New feature / API change

```python
import math
import numpy as np
import warp as wp

from newton.sensors import SensorBatchedCamera

width = 128
height = 128
selected_worlds = np.array([0, 1], dtype=np.int32)

sensor = SensorBatchedCamera(model)
camera_rays = sensor.utils.compute_pinhole_camera_rays(width, height, math.radians(60.0))
camera_indices = sensor.utils.create_camera_indices(selected_worlds, ray_bundle_indices=0)

camera_transforms = wp.array(
    [wp.transformf(wp.vec3f(0.0, 0.0, 2.0), wp.quatf(0.0, 0.0, 0.0, 1.0))] * len(selected_worlds),
    dtype=wp.transformf,
    device=model.device,
)
color_image = sensor.utils.create_color_image_output(len(selected_worlds), width, height)

model.bvh_refit_shapes(state)
model.bvh_refit_particles(state)
sensor.update(state, camera_transforms, camera_rays, camera_indices, color_image=color_image)

rgba = sensor.utils.to_rgba_from_color(color_image)
```

## Summary by CodeRabbit

* **New Features**
  * Added `SensorBatchedCamera` for batched, multi-view raytraced rendering with outputs for color, depth, normals, albedo, HDR color, and shape indices.
  * Added a runnable `SensorBatchedCamera` example with viewer logging and GUI controls.
* **Deprecation**
  * Deprecated `SensorTiledCamera` in favor of `SensorBatchedCamera`.
* **Documentation**
  * Updated API docs, sensor concepts, conventions, visualization guidance, and README sensor examples to reference batched camera usage.
* **Tests**
  * Added/expanded automated coverage for batched rendering, HDR, forward-depth conversion, heightfields, multiworld particles, and packed color-to-RGBA.
* **Performance/Benchmarks**
  * Replaced tiled camera benchmarks with batched-camera equivalents.


docs/api/newton_sensors.rst
```diff
@@ -13,7 +13,8 @@ newton.sensors
    :toctree: _generated
    :nosignatures:
 
+   SensorBatchedCamera
    SensorContact
    SensorFrameTransform
    SensorIMU
-   SensorTiledCamera
```
StafaH, line 19: Formatting?

newton/_src/sensors/batched_camera_renderer/tiling.py
```diff
@@ -0,0 +1,75 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 The Newton Developers
```
StafaH, line 1: Can we merge this file with the other files (maybe render and utils), instead of having a seperate file?

newton/_src/sensors/batched_camera_renderer/render.py
```diff
@@ -0,0 +1,294 @@
...
+from ...geometry import Gaussian, GeoType
+from ...utils.color import ColorSpace, color_srgb_to_linear, linear_to_srgb_wp, srgb_to_linear_wp
+from . import lighting, raytrace, textures, tiling
+from .types import ClearData, MeshData, RenderOrder, TextureData
+
+if TYPE_CHECKING:
+    from .render_context import RenderContext
+
+
+def _srgb_packed_rgba_to_linear(packed: int) -> int:
```
StafaH: Let's collect these util ops in one place (either utils.py or elsewhere)

newton/_src/sensors/sensor_batched_camera.py
```diff
@@ -0,0 +1,218 @@
...
+        ::
+
+            sensor = SensorBatchedCamera(model)
+            rays = sensor.utils.compute_pinhole_camera_rays(width, height, fov)
+            color = sensor.utils.create_color_image_output(1, width, height)
+            camera_transforms = wp.array(
+                [wp.transformf(wp.vec3f(0.0, 0.0, 1.0), wp.quat_identity())],
+                dtype=wp.transformf,
+            )
+            camera_indices = sensor.utils.create_camera_indices(0)
```
StafaH, line 55: The usage of create_camera_indices is a bit confusing to me in the first pass, I had to dig a bit deeper to find out what is happening, so it might be unclear to the end-user as well.

Also is there a need for camera indices to exist outside the batched sensor or render context?

newton/_src/sensors/batched_camera_renderer/utils.py
```diff
@@ -0,0 +1,887 @@
...
+
+        Returns:
+            Array of shape ``(camera_count, height, width)``, dtype ``vec3f``.
+        """
+        return wp.zeros((camera_count, height, width), dtype=wp.vec3f, device=self.__render_context.device)
+
+    def create_camera_indices(
+        self,
+        world_indices: int | Sequence[int] | np.ndarray,
+        ray_bundle_indices: int | Sequence[int] | np.ndarray | None = None,
```
StafaH: ray bundle naming seems a bit odd, can we use ray_adr (start adr for rays for a specific rendered view).

StafaH (commented):
Doing a first pass over the PR, the implementation looks solid, just have some thoughts on the user side API ergonomics. Do we expect camera indices to change? It could be something we attach to the sensor itself and do as part of initialization of the sensor?

daniela-hase:
> Doing a first pass over the PR, the implementation looks solid, just have some thoughts on the user side API ergonomics. Do we expect camera indices to change? It could be something we attach to the sensor itself and do as part of initialization of the sensor?

Yeah, they could change - the original request was to disable cameras during simulation once they don't need further training anymore :)

docs/api/newton_sensors.rst
daniela-hase: I am not sure I understand '^^

newton/_src/sensors/batched_camera_renderer/tiling.py
daniela-hase: We could, yeah. But .. is there a benefit to it?
Actually, `pack_rgba_to_uint32` sneaked in there - that should not live in there. But the rest I feel like can stay there - keeps it cleanly separated

newton/_src/sensors/batched_camera_renderer/render.py
daniela-hase: So far `utils.py` is only for literal utility functions for the user, none that are actually used by the renderer.

Maybe I rename `tiling.py` to `render_utils.py` and put it all in there 🤔

newton/_src/sensors/batched_camera_renderer/utils.py
daniela-hase: Yeah, true - bundle isn't the best name. How about just `rays_indices`? 🤔

docs/api/newton_sensors.rst
StafaH: Sorry, it was showing a newline but now I see it's fine

newton/_src/sensors/batched_camera_renderer/tiling.py
StafaH: Moving it is optional, it's just a small nit since it's only used in one place (render.py), they could live there as well.

newton/_src/sensors/batched_camera_renderer/render.py
StafaH: Yea that's better

newton/_src/sensors/batched_camera_renderer/utils.py
StafaH: Yea that would be better.

daniela-hase: Renamed it to `rays_indices` :)

newton/_src/sensors/batched_camera_renderer/tiling.py
daniela-hase: I move them to `render_utils.py` 😄

newton/_src/sensors/sensor_batched_camera.py
```diff
@@ -0,0 +1,218 @@
...
+    - **normal** -- surface normal at hit point (``vec3f``).
+    - **albedo** -- unshaded surface color (``uint32``).
+    - **shape_index** -- shape id per pixel (``uint32``).
+
+    Each rendered view is selected by a row in ``camera_indices`` passed to
+    :meth:`update`: ``camera_indices[view_id, 0]`` is the world index, and
+    ``camera_indices[view_id, 1]`` is the ray bundle index into
+    ``camera_rays``. Output arrays have shape ``(view_count, height, width)``.
+    Use the ``flatten_*`` helpers to arrange them into a single RGBA grid, or
+    the ``to_rgba_from_*`` helpers for batched viewer display.
```
eric-heiden, line 37: Similar to what @StafaH commented, I'm also confused by this.
Can you clarify here what is a view and how do camera_count, view_count, and ray bundles fit together. How do you go from `camera_indices` to final view index, for view_index `i` how do I know which camera it belongs to?

Perhaps a simpler design would be to only have cameras assigned to world. There can only be one camera in one world, or "global" cameras which have world index == -1. This is the same as how we define `Model.body_world`, `shape_world` and other world index attributes. Then you have separate arrays for storing these ray bundles. At the end the sensor output is always `(camera, y, x)` with a 1:1 mapping from this camera array to the final output. Does that make sense?

daniela-hase: I agree it is a bit confusing, I should probably clarify it in the docstring better.

A view is one rendered image (a tile). To render that we need the camera transform, the precomputed camera rays and what world this view should render.

We already have the `camera_transforms` array containing one transform per camera, as well as the `camera_rays` array containing the rays that can be shared between cameras since typically all cameras across worlds share the same lens. This PR introduces the `camera_indices` array, containing two indices per camera. One pointing to which world to render, and the other telling which camera rays to use.

Maybe a more fitting name would be `camera_world_and_rays_indices`.

The output of the sensor is `(camera_idx, y, x)` where `camera_idx` maps 1:1 to `camera_transforms` / `camera_indices`.

eric-heiden: I think if we want to keep this design it's better we store these indices separately. I.e., we have
* `camera_world` - which world the camera should render
* `camera_ray_start` - start index in `camera_rays` for this camera, with an additional sentinel at the end of the array where we have the total number of rays (similar to how `Model.joint_q_start` and similar arrays work)

daniela-hase: `camera_ray_start` implies that the `camera_rays` would turn into a one dimensional array?
How about `camera_world_index` and `camera_ray_index` if we really want to keep them separate?

However, I kept them together like this because I didn't see a reason to separate them and figured having them like this is better for cache usage - having both values live side by side in memory, since they are used in the kernel at the same time.

PR #3283, opened 2026-06-30
Preserve MJCF geom groups

vreutskyy:
## Description

Preserve MuJoCo geom visualization groups when importing MJCF or USD and constructing a `SolverMuJoCo` model.

`SolverMuJoCo.register_custom_attributes()` now declares the per-shape `model.mujoco.geom_group` attribute, mapped from MJCF `group` and USD `mjc:group`. The MJCF importer uses its existing MuJoCo-attribute registration path. USD uses the existing generic custom-attribute pipeline: callers must follow the documented flow and call `SolverMuJoCo.register_custom_attributes(builder)` before `builder.add_usd(...)`. `import_usd.py` itself is unchanged.

The solver forwards the stored value to the MuJoCo geom spec instead of allowing every geom to fall back to group 0. This keeps Newton-generated MuJoCo and MJWarp models aligned without conflating visualization groups with Newton collision groups.

This addresses the remaining `geom_group` portion of #2492. The separate zero-mask explicit-pair classification edge is tracked in #3284 and fixed independently by #3285.

## Checklist

- [x] New or existing tests cover these changes
- [x] The documentation is up to date with these changes
- [x] `CHANGELOG.md` has been updated (if user-facing change)

## Test plan

```text
uv run --no-sync -m unittest newton.tests.test_import_mjcf.TestImportMjcfSolverParams.test_geom_group_parsing -v
uv run --no-sync -m unittest newton.tests.test_mujoco_solver.TestMuJoCoSolverGeomProperties.test_geom_group_conversion -v
uv run --no-sync -m unittest newton.tests.test_import_usd.TestImportSampleAssetsBasic.test_geom_group_parsing_and_conversion -v
uv run --no-sync -m unittest newton.tests.test_mujoco_solver.TestMuJoCoSolverGeomProperties.test_geom_priority_conversion -v
uv run --no-sync -m unittest newton.tests.test_menagerie_mujoco.TestMenagerie_Aloha -v
uvx pre-commit run -a
```

The Aloha regression covers model comparison, forward kinematics, and dynamics. The automated USD regression authors `mjc:group = 3` after pre-registering the MuJoCo custom attributes, then verifies the value in the Newton model, native MuJoCo model, and MJWarp model.

## Bug fix

**Steps to reproduce:**

1. Import an MJCF model containing geoms assigned to nonzero visualization groups.
2. Finalize the Newton model and construct `SolverMuJoCo`.
[...]

PR #3285, opened 2026-06-30
Preserve explicit-pair MJCF geoms

### LINKED ISSUE #2492: Aloha MJCF import has multiple issues (dof_damping, equality constraints, ngeom)

### LINKED ISSUE #3284: MJCF import drops zero-mask geoms referenced by explicit contact pairs
## Description

MJCF geoms with `contype="0" conaffinity="0"` are classified as visual-only when they do not have an explicitly matched collider class. With `parse_visuals=False`, those geoms are dropped even when a `<contact><pair>` explicitly references them.

MuJoCo keeps pair-referenced geoms: zero contact masks disable automatic contacts, but an explicit pair still enables contact between the referenced geoms.

This was identified while resolving #2492 and was previously noted during #3030. It is separate from MJCF `geom_group` preservation, which is handled by #3283.

## Steps to reproduce

```python
import newton

mjcf = """<mujoco>
    <default><geom contype="0" conaffinity="0"/></default>
    <worldbody>
        <geom name="floor" type="plane" size="1 1 0.1"/>
        <body name="ball">
            <freejoint/>
            <geom name="ball" type="sphere" size="0.1"/>
        </body>
    </worldbody>
    <contact><pair geom1="floor" geom2="ball"/></contact>
</mujoco>"""

builder.add_mjcf(mjcf, parse_visuals=False)
print(builder.shape_count)
```

Native MuJoCo reports two geoms and one explicit pair. Newton currently reports zero shapes, so the pair cannot be preserved.

## Root cause

`_process_body_geoms()` uses `contype`/`conaffinity` to decide whether an unclassified geom is a collider, without considering whether `<contact><pair>` references that geom. Pair parsing happens later, after visual-only geoms may already have been discarded.

## Expected behavior

Pre-collect names referenced by `<contact><pair>` and retain those geoms as colliders during body/worldbody/frame processing. They should keep automatic collisions disabled while remaining available to explicit pairs.

vreutskyy:
## Description

Preserve MJCF geoms referenced by explicit `<contact><pair>` elements when visual-only geoms are filtered during import.

MuJoCo allows explicit pairs to generate contact even when both geoms use `contype="0" conaffinity="0"`. Newton previously classified unclassified zero-mask geoms as visual-only before parsing contact pairs, so `parse_visuals=False` discarded the shapes and the pair could not be exported.

The importer now pre-collects pair-referenced geom names and routes those geoms through the collider path. Their Newton collision group remains 0, which continues to disable automatic contacts while allowing the explicit pair to work. Repeated top-level `<contact>` sections are handled consistently when collecting geom references and parsing both `<pair>` and `<exclude>` elements. Early retention uses the same exact-or-suffix matching as pair lookup, so full hierarchical Newton geom labels are preserved too.

Closes #3284.

This is the final follow-up for #2492 and should merge after #3283, which preserves MJCF `geom_group` metadata. Closes #2492.

## Checklist

## Test plan

```text
uv run --no-sync -m unittest newton.tests.test_import_mjcf.TestContypeConaffinityZero -v
uv run --no-sync -m unittest newton.tests.test_import_mjcf -q
uvx pre-commit run -a
```

The original regression fails on upstream `main` with `shape_count == 0`; it passes with two retained group-0 shapes, one exported pair, and a generated contact. The repeated-contact regression fails before the follow-up fix with two of three geoms retained; it now verifies three retained group-0 shapes, two exported pairs, and an exclude from the second contact section. The hierarchical-label regression fails before the review fix with no retained shapes; it now verifies two retained group-0 shapes and one exported pair. The complete MJCF importer suite passes all 222 tests.

## Bug fix

**Steps to reproduce:**

1. Define unclassified MJCF geoms with `contype="0" conaffinity="0"`.
2. Reference them from `<contact><pair>`.
3. Import with `parse_visuals=False`.
4. Observe that Newton drops both geoms before parsing the pair.

**Minimal reproduction:**

```python
import newton

mjcf = """<mujoco>
    <default><geom contype="0" conaffinity="0"/></default>
    <worldbody>
        <geom name="floor" type="plane" size="1 1 0.1"/>
        <body name="ball">
            <freejoint/>
            <geom name="ball" type="sphere" size="0.1"/>
        </body>
    </worldbody>
    <contact><pair geom1="floor" geom2="ball"/></contact>
</mujoco>"""

builder.add_mjcf(mjcf, parse_visuals=False)
print(builder.shape_count)
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved MJCF geoms referenced only by explicit contact pairs during import when visual parsing is disabled, preventing them from being filtered out from collision shapes.
  * Improved explicit contact-pair parsing to correctly handle multiple top-level contact sections, including related exclusions and hierarchical geom references.
* **Tests**
  * Added regression coverage for non-visual collision geoms from explicit contact pairs, including multi-section scenarios, MuJoCo pair compilation checks, and verification that contacts are generated.

### DIFF
--- a/newton/_src/utils/import_mjcf.py
+++ b/newton/_src/utils/import_mjcf.py
@@ -331,9 +331,9 @@ def parse_mjcf(
     root, base_dir = _load_and_expand_mjcf(source, path_resolver)
     mjcf_dirname = base_dir or "."  # Backward compatible fallback for mesh paths
 
-    contact = root.find("contact")
+    contact_sections = root.findall("contact")
     explicit_pair_geom_names: set[str] = set()
-    if contact is not None:
+    for contact in contact_sections:
         for pair in contact.findall("pair"):
             for geom_key in ("geom1", "geom2"):
                 geom_name = pair.attrib.get(geom_key)
@@ -2465,9 +2465,10 @@ def parse_mjcf(
                 return idx
         return None
 
-    if contact is not None and has_pair_attrs:
+    if has_pair_attrs:
         # Parse <pair> elements - explicit contact pairs with custom properties
-        for pair in contact.findall("pair"):
+        pairs = (pair for contact in contact_sections for pair in contact.findall("pair"))
+        for pair in pairs:
             geom1_name = pair.attrib.get("geom1")
             geom2_name = pair.attrib.get("geom2")
 
@@ -2507,7 +2508,7 @@ def parse_mjcf(
                 print(f"Parsed contact pair: {geom1_name} ({geom1_idx}) <-> {geom2_name} ({geom2_idx})")
 
     # Parse <exclude> elements - body pairs to exclude from collision detection
-    if contact is not None:
+    for contact in contact_sections:
         for exclude in contact.findall("exclude"):
             body1_name = exclude.attrib.get("body1")
             body2_name = exclude.attrib.get("body2")
--- a/newton/_src/utils/import_mjcf.py
+++ b/newton/_src/utils/import_mjcf.py
@@ -1340,7 +1340,11 @@ def parse_mjcf(
             collides_with_anything = not (int(contype) == 0 and int(conaffinity) == 0)
 
             # Explicit pairs override contact masks, so their geoms must survive visual filtering.
-            if geom_name in explicit_pair_geom_names:
+            geom_label = f"{label_prefix}/{geom_name}" if label_prefix else geom_name
+            is_explicit_pair_geom = any(
+                geom_label == name or geom_label.endswith(f"/{name}") for name in explicit_pair_geom_names
+            )
+            if is_explicit_pair_geom:
                 required_colliders.append(geom)
             elif geom_class is not None:
                 neither_visual_nor_collider = True
--- a/newton/_src/utils/import_mjcf.py
+++ b/newton/_src/utils/import_mjcf.py
@@ -331,6 +331,15 @@ def parse_mjcf(
     root, base_dir = _load_and_expand_mjcf(source, path_resolver)
     mjcf_dirname = base_dir or "."  # Backward compatible fallback for mesh paths
 
+    contact = root.find("contact")
+    explicit_pair_geom_names: set[str] = set()
+    if contact is not None:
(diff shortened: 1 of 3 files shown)


newton/_src/utils/import_mjcf.py
```diff
@@ -1329,7 +1339,10 @@ def _process_body_geoms(
             conaffinity = geom_attrib.get("conaffinity", 1)
             collides_with_anything = not (int(contype) == 0 and int(conaffinity) == 0)
 
-            if geom_class is not None:
+            # Explicit pairs override contact masks, so their geoms must survive visual filtering.
+            if geom_name in explicit_pair_geom_names:
```
camevor: This does not match hierarchical labels, unlike `_find_shape_idx`, which is narrower, but can lead to the same kind of bug addressed here.

vreutskyy: Fixed in 819a41b3. Early geom retention now builds the prospective hierarchical shape label and uses the same exact-or-suffix semantics as `_find_shape_idx`. The new regression uses full hierarchical pair references with `parse_visuals=False` and verifies both zero-mask geoms and the exported pair are preserved.

camevor (approved):
Thanks, lgtm!

PR #3287, opened 2026-06-30
Fix XPBD restitution thickness gate

### LINKED ISSUE #3034: XPBD rigid restitution double-counts contact thickness in separation check
`apply_rigid_restitution()` computes the restitution activation separation from offset contact points and then subtracts `contact_thickness0 +
contact_thickness1` again.

The contact offsets already move the raw support/skeleton points to the effective surface anchors:

```python
bx_a = contact_surface_point(X_wb_a_prev, contact_point0[tid], contact_offset0[tid])
bx_b = contact_surface_point(X_wb_b_prev, contact_point1[tid], contact_offset1[tid])
d = contact_surface_separation(bx_a, bx_b, n, contact_thickness0[tid], contact_thickness1[tid])
```

Since `contact_surface_separation()` expects raw support points and subtracts the margins/thickness itself, this makes d too negative. As a result, restitution can be applied to contacts that are merely touching or less deeply penetrating than reported.

han-xudong:
## Description

Fix XPBD rigid restitution so the activation gate does not double-count contact thickness after contact offsets have already produced effective surface anchors.

`apply_rigid_restitution()` was computing `bx_a` / `bx_b` with `contact_surface_point()`, then passing `contact_thickness0/1` into `contact_surface_separation()` again. This could make already-separated surface anchors look penetrating and apply a spurious restitution impulse.

Closes #3034.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_rigid_restitution_surface_gate_does_not_double_count_thickness
uv run --extra dev -m newton.tests -k restitution
uvx pre-commit run -a
```

## Bug fix

**Steps to reproduce:**

1. Construct an XPBD rigid contact whose stored support point plus contact offset already places the effective surface anchor above the other surface.
2. Run `apply_rigid_restitution()` with nonzero `contact_thickness0`.
3. Without this PR, the restitution gate subtracts thickness again and writes a nonzero velocity delta even though the effective surface anchors are separated.
4. With this PR, the gate returns early and leaves the restitution delta at zero.

**Minimal reproduction:**

```python
import numpy as np
import warp as wp

from newton._src.solvers.xpbd.kernels import apply_rigid_restitution

# See test_rigid_restitution_surface_gate_does_not_double_count_thickness
# for the complete kernel-level reproduction.
```

The added regression test `test_rigid_restitution_surface_gate_does_not_double_count_thickness` directly launches `apply_rigid_restitution()` with surface anchors separated by `0.01 m`. Before the fix, it fails with `deltas[0].linear.y == -2.0`; after the fix, `deltas` remains zero.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected XPBD rigid-body restitution “surface gate” logic to prevent double-counting contact thickness after applying contact offsets.
  * Adjusted the restitution contact gating to use signed separation projection, which changes which contacts receive restitution impulses.
* **Tests**
  * Added a regression test covering the double-counting scenario for rigid restitution to ensure the expected no-delta outcome.
* **Documentation**
  * Updated the **Unreleased → Fixed** changelog entry to describe the corrected restitution behavior.


newton/_src/solvers/xpbd/kernels.py
```diff
@@ -2634,7 +2634,7 @@ def apply_rigid_restitution(
     bx_b = contact_surface_point(X_wb_b_prev, contact_point1[tid], contact_offset1[tid])
 
     n = contact_normal[tid]
-    d = contact_surface_separation(bx_a, bx_b, n, contact_thickness0[tid], contact_thickness1[tid])
+    d = contact_surface_separation(bx_a, bx_b, n, 0.0, 0.0)
```
adenzler-nvidia: Now that `bx_a` and `bx_b` are effective-surface anchors, `contact_thickness0/1` are unused by this kernel. Could we remove them from the kernel signature and launch sites, and compute this as `d = wp.dot(n, bx_b - bx_a)`? `contact_surface_separation()` represents raw support points plus margins, so passing already-offset points with zero margins obscures which representation is being used and leaves two dead kernel arguments.

adenzler-nvidia (changes requested):
The restitution gate fix looks correct—thanks for adding the focused regression coverage. I have one requested cleanup to make the contact representation explicit and remove the now-unused kernel inputs.

newton/_src/solvers/xpbd/kernels.py
han-xudong: @adenzler-nvidia, addressed in f4e79ad1. The restitution gate now computes separation directly with wp.dot(n, bx_b - bx_a), and the unused contact_thickness0/1 inputs have been removed from the kernel signature and both launch sites. The branch is also updated to the latest upstream/main; all 54 XPBD tests and pre-commit checks pass. Please take another look when convenient.

adenzler-nvidia (approved):
Thank you!

PR #3289, opened 2026-06-30
Publish generated viewer textures atomically

### LINKED ISSUE #3288: [BUG] Generated viewer textures can be read before PNG writes complete
## Bug Description

When `ViewerUSD` exports an in-memory texture array to a PNG file, it currently writes directly to the final texture path. `ViewerRTX` uses the USD stage generated by `ViewerUSD`, so resetting/recreating an example can rewrite the same generated texture path while the renderer-side texture processor may still be reading it.

This can produce repeated RTX/Omni texture loading errors such as:

```text
[Error] [omni.rtx] Couldn't process /tmp/_tex__geometry_mesh_14.png, it might not have written completely. Reason: Failed to load image: PNG not supported: unknown PNG chunk type
[Error] [omni.ujitso] TextureProcessor : Failed to process texture /tmp/_tex__geometry_mesh_14.png : 6
```

Expected behavior: generated texture files should only become visible at the final path once the PNG is completely written, so readers never observe a partially-written file.

## Reproduction Script

This was observed with an RTX viewer example that imports a URDF containing textured visual meshes and then uses the viewer Reset button to recreate the example. A representative command for the example is:

```bash
uv run python examples/import/import_urdf.py --viewer rtx
```

The issue is in the generated texture export path rather than in a specific robot asset: any `ViewerUSD.log_mesh(..., texture=<numpy array>)` path can rewrite `_tex_<mesh>.png` directly on reset/rebuild.

## System Information

- Newton: 1.4.0.dev0
- Newton commit: 49ca73b9
- Warp: 1.15.0.dev20260626
- Python: 3.12.3
- OS: Linux
- GPU: NVIDIA GeForce RTX 5090 D v2
- NVIDIA driver: 590.48.01

ruziniuuuuu:
## Description

Fixes #3288.

`ViewerUSD` materializes in-memory texture arrays as PNG files before USD/RTX consumers load them. Those generated texture paths are intentionally stable across rebuilds, but reset can rewrite the same `_tex_<mesh>.png` while renderer-side texture processing is still reading it.

This change keeps the stable texture path but writes the PNG to a sibling temporary file first, then publishes it with `os.replace()`. Readers therefore see either the previous complete PNG or the new complete PNG, never a partially written file.

The regression test now exercises that guarantee directly: after publishing a complete PNG, it simulates the next save crashing after writing half of the bytes and verifies the previously published PNG remains intact and no temporary sibling leaks.

## Changes

- Publish generated PNGs through a same-directory temporary file and `os.replace()`.
- Clean up temporary files when saving or replacement fails.
- Isolate ViewerUSD test output in a per-test temporary directory.
- Add a discriminating partial-write regression test that fails with the legacy direct writer.
- Keep the changelog entry in the active Unreleased section after merging current `main`.

## Checklist

## Validation

- `uvx pre-commit run --files CHANGELOG.md newton/_src/viewer/viewer_usd.py newton/tests/test_viewer_usd.py`
- `python -m unittest newton.tests.test_viewer_usd` — 11 passed
- Replaced `_save_texture_atomic` with the legacy direct writer while running the new regression test; it failed by reading a truncated PNG, confirming the test distinguishes the fix.
- Previously validated a headless RTX import and a Galbot reset-path diagnostic with stable generated texture paths.

## Bug reproduction

1. Run an RTX viewer scene whose mesh textures are held as in-memory arrays and exported by `ViewerUSD`.
2. Reset or recreate the example so the same generated texture path is rewritten.
3. Without atomic publication, renderer-side texture processing can observe an incomplete PNG and report errors such as an unknown PNG chunk type.

## Not run

- Interactive RTX reset after the latest merge with `main`.
- Maintainer-gated AWS GPU workflows.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated texture PNG exports to be written atomically, reducing the chance of partially written or corrupted images being detected.
  * Regenerating textured models after `clear_model()` keeps the generated texture asset path stable.
  * Improved tests by using an isolated per-test working directory and validating there are no leaked temporary `*.tmp` files after regeneration.


ruziniuuuuu:
Before:

https://github.com/user-attachments/assets/1f72df9c-7c46-49c3-9630-696a9ed0522b

After:

https://github.com/user-attachments/assets/b11ae75c-ea31-4fb6-a7aa-c49515d04361

newton/tests/test_viewer_usd.py
```diff
@@ -101,6 +114,55 @@ def test_reuses_existing_layer_for_same_output_path(self):
         self.assertFalse(prim_after.IsValid())
         self.assertTrue(os.path.exists(temp_file.name))
 
+    def test_generated_texture_path_stays_stable_after_clear_model(self):
```
adenzler-nvidia: This test passes without the fix: the old code also uses a deterministic path (path stability holds), the file exists, and it never creates `.tmp` files (no-leak holds). So it doesn't discriminate the atomic write.

A discriminating version goes through `log_mesh` and simulates a crash mid-write on the second export, then asserts the published PNG still holds the first write's complete bytes. Fails on main (truncated bytes land at the final path), passes with this PR:

```python
def test_partial_texture_rewrite_keeps_published_png_intact(self):
    """A crash mid-write must not corrupt the previously published PNG."""
    viewer = self._make_viewer()
    mesh_name = "/textured_mesh"
    # ... points/indices/uvs/texture as in the current test ...

    viewer.begin_frame(0.0)
    viewer.log_mesh(mesh_name, points, indices, uvs=uvs, texture=texture)
    tex_path = self._logged_texture_path(viewer, mesh_name)
    with open(tex_path, "rb") as f:
        published = f.read()

    real_save = PIL.Image.Image.save

    def partial_save(img, fp, *args, **kwargs):
        buf = io.BytesIO()
        real_save(img, buf, *args, **kwargs)
        with open(fp, "wb") as f:
            f.write(buf.getvalue()[: buf.tell() // 2])
        raise OSError("simulated crash mid-write")

    viewer.clear_model()
    viewer.begin_frame(0.0)
    with mock.patch("PIL.Image.Image.save", partial_save), self.assertWarns(UserWarning):
        viewer.log_mesh(mesh_name, points, indices, uvs=uvs, texture=texture)

    with open(tex_path, "rb") as f:
        self.assertEqual(f.read(), published)
    leaked = [n for n in os.listdir(viewer._test_work_dir) if n.endswith(".tmp")]
    self.assertEqual(leaked, [])
```

This also subsumes the `.tmp`-leak assertion, and the `assertWarns` keeps the swallowed-exception warning from tripping the strict-warnings test config. The path-stability assertion from the current test is fine to keep if you want it, but as a separate concern.

CHANGELOG.md
```diff
@@ -55,6 +55,7 @@
 - Fix URDF `package://` mesh fallback resolution without `resolve-robotics-uri-py` so package names only match full path components instead of unrelated directory-name substrings
 - Fix `ModelBuilder.collapse_fixed_joints()` crashing with `IndexError` when a `mujoco:equality_constraint` row omits optional fields (`anchor`, `relpose`) that carry defaults. (#3054)
 - Fix `ViewerGL.set_model()` resetting headless/interactive camera and wind state when switching between models that use the same up-axis. (#2658)
+- Fix `ViewerUSD` exposing partial PNGs to texture consumers by writing generated textures to a sibling temp file and publishing them with `os.replace`, so `clear_model` cannot leave a renderer reading a truncated image while the new PNG is being written.
```
adenzler-nvidia: This describes the implementation rather than the user-facing behavior, and is missing the issue ref:

```suggestion
- Fix `ViewerUSD` texture consumers observing partially written PNGs by publishing generated textures atomically (#3288)
```

adenzler-nvidia (changes requested):
The fix itself looks right — same-directory temp + `os.replace` is the standard idiom for a path a renderer may re-read at any time, and the unique temp name plus failure cleanup are both worth their lines (stray `.tmp` files would otherwise accumulate next to the user's output USD, since the caller warns and continues on failure).

Requesting changes on one thing: the headline test passes without the fix, so the PR currently ships with no regression test for the bug it fixes. Details inline.

Not for this PR, but for context: the trigger in #3288 is reset rewriting byte-identical PNGs to the same path. Skipping the rewrite when content is unchanged (a viewer-held path→digest map that survives `clear_model`) would remove the race trigger in the common case and avoid RTX texture-reload churn on every reset. That composes with this fix rather than replacing it — first writes and genuinely changed textures still need the atomic publish.

newton/tests/test_viewer_usd.py
ruziniuuuuu: Addressed in 6d905fe6. The test now simulates the second PNG export failing after writing half its bytes, then verifies the previously published PNG is unchanged and no .tmp sibling remains. I also verified the same test fails with the legacy direct writer, merged current main while preserving points_as_spheres=False in the helper, and moved the changelog entry back to Unreleased.

ruziniuuuuu:
@adenzler-nvidia, the requested discriminating regression test is now in 6d905fe6 and the branch has been merged with current main. The inline thread has the implementation and red/green details; please take another look when convenient.

adenzler-nvidia (approved):
Thanks for the quick turnaround. The new `test_partial_texture_rewrite_keeps_published_png_intact` now discriminates the fix — with the legacy direct writer the truncated bytes land at the published path, with the atomic publish the prior PNG survives — and the `os.replace`-failure test covers the remaining cleanup branch. CHANGELOG wording and placement look good after the main merge. LGTM.

PR #3290, opened 2026-06-30
Add regex selection patterns

### LINKED ISSUE #3182: [REQ] Selection API support for list of regexes rather than just a single regex.
### Description

The selection API would support a list of regexes as inputs instead of the current single regex support. 

```python
    def __init__(
        self,
        model: Model,
        pattern: list[str], # Use to be str
        include_joints: list[str] | list[int] | None = None,
        exclude_joints: list[str] | list[int] | None = None,
        include_links: list[str] | list[int] | None = None,
        exclude_links: list[str] | list[int] | None = None,
        include_joint_types: list[int] | None = None,
        exclude_joint_types: list[int] | None = None,
        include_loop_closing_joints: bool = False,
        verbose: bool | None = None,
        match_full_labels: bool = False,
    ):
```

### Motivation / Use Case

That would enable support for ObjectCollections and ArticulationCollections:
https://github.com/isaac-sim/IsaacLab/issues/6147
https://github.com/isaac-sim/IsaacLab/pull/6140

### Alternatives Considered

Use another matching function than fnmatch. Like a plain regex match from `re`.

### LINKED ISSUE #3272: [REQ] Full regex support for Selection APIs
### Description

The Selection APIs currently use fnmatch for pattern matching when selecting prims, which supports only a subset of allowed regex expressions in comparison to full support from regex (re). It would be very useful for the selection APIs to also allow any valid regex expression when filtering prims.

### Motivation / Use Case

In Isaac Lab, we may want to filter specific environments (especially for multi-task scenarios), which may also have multiple objects and we only want to select a subset of objects:

- /World/envs/env_[0-9]+/Object_(A|B|C)

Sometimes, we would also like to match specific bodies or sensors by name:

- /World/envs/env_[0-9]+/Robot/(LF|RF)_FOOT
- /World/envs/env_[0-9]+/Robot/(?!base$).*

Exclude certain bodies or objects from the list:

- /World/envs/env_[0-9]+/(?!Object_B$)Object_.*

### Alternatives Considered

_No response_

han-xudong:
## Description

Add opt-in regex pattern matching to Newton selection APIs while preserving existing glob matching by default.

This introduces `pattern_syntax="regex"` for `match_labels()`, `find_matching_ids()`, and `ArticulationView`. Regex patterns use `re.fullmatch`, and list-of-string patterns match when any regex matches. Existing calls continue to use `fnmatch` glob matching unless `pattern_syntax` is explicitly set.

Closes #3272.
Closes #3182.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k regex
uv run --extra dev -m newton.tests -k test_match_labels
uvx pre-commit run -a
```

## New feature / API change

```python
import newton

view = newton.selection.ArticulationView(
    model,
    pattern=r"env_[0-9]+/Robot/(LF|RF)_FOOT",
    pattern_syntax="regex",
)

feet = newton.selection.ArticulationView(
    model,
    pattern=r"env_[0-9]+/Robot",
    include_links=[r".*/LF_FOOT", r".*/RF_FOOT"],
    pattern_syntax="regex",
)
```

Calls without `pattern_syntax` keep the existing glob behavior, such as `pattern="robot*"` or `include_links="foot_*"`.

## Summary by CodeRabbit

* **New Features**
  * Added optional regex-based label matching for selection; glob matching remains the default.
  * Selection filters now accept single strings in addition to lists for more flexible pattern-based selection.

* **Documentation**
  * Updated label matching docs to explain glob vs regex behavior, including full-string regex matching semantics.

* **Tests**
  * Added unit and integration coverage for regex selection, including de-duplication and correct articulation/link selection results.


camevor:
Thank you for your contribution, @han-xudong. The list-of-patterns requirement from #3182 was addressed by #3294. The remaining native regex functionality proposed here is not currently planned and is being discussed in #3272.

On that basis, I’m closing this PR for now. We can revisit the approach depending on the outcome of that discussion.

PR #3292, opened 2026-06-30
Fix imported collider visibility fallback

### LINKED ISSUE #3291: Imported colliders become visible when a body has no visual geometry
## Bug report

Imported collision shapes are marked as visual geometry by default when their owning body has no visual-only shape. This makes collision proxies render as normal visuals in viewers even though the source asset authored them as collision geometry.

## Reproduction

A minimal MJCF pattern is:

```xml
<mujoco model="collision_visibility">
  <default>
    <default class="visual">
      <geom contype="0" conaffinity="0"/>
    </default>
    <default class="collision">
      <geom contype="1" conaffinity="1"/>
    </default>
  </default>
  <worldbody>
    <body name="wheel">
      <geom name="wheel_visual" class="visual" type="box" size="0.1 0.1 0.1"/>
      <body name="roller">
        <geom name="roller_collision" class="collision" type="sphere" size="0.05"/>
      </body>
    </body>
  </worldbody>
</mujoco>
```

After `ModelBuilder.add_mjcf(...)`, `roller_collision` has both `COLLIDE_SHAPES` and `VISIBLE`. In a larger robot with passive wheel roller collision bodies, this causes the collision spheres to appear in both the GL and RTX viewers as regular visual geometry.

## Expected behavior

Collision shapes should remain collision-only by default. They should only receive the `VISIBLE` flag when callers explicitly request collider visualization, for example with `force_show_colliders=True`, or when visuals are intentionally parsed as colliders.

## Actual behavior

The shared import visibility helper returns `True` for collision shapes whenever the owning body has no visual-only geometry. That fallback makes collision-only child bodies visible by default.

## Why this fallback seems problematic

The visibility of a collision shape should follow the authored geometry role, not whether a visual sibling exists on the same body. Body-local fallback visibility leaks an importer heuristic into normal rendering and is especially surprising for articulated assets where collision proxies are often placed on child bodies.

## Environment

Observed on current Newton `main` at commit `2064e3b7` in a local checkout. CUDA is not required to reproduce the flag state; the issue is visible by inspecting `builder.shape_flags` after import.

ruziniuuuuu:
## Summary

Fixes #3291.

Scope the collider-visibility fallback to a single `add_mjcf()`, `add_urdf()`, or `add_usd()` call instead of evaluating each body independently:

- When an import contains visual geometry, collision-only bodies stay hidden by default and remain available through the viewer's **Show Collision** toggle.
- When an import is collision-only, its colliders retain `VISIBLE` so the asset does not render as an empty viewport.
- `force_show_colliders=True` and `parse_visuals_as_colliders=True` continue to opt into visible colliders explicitly.

The per-import scope is deliberate: if one scene file authors visual geometry anywhere, collision-only props elsewhere in that same importer call are treated as colliders rather than ordinary visuals.

## Root cause

`should_show_collider()` used a per-body `has_visual_shapes` decision. This made collision proxies on child bodies render as ordinary visuals even when the asset authored a visual/collision split. Removing the fallback entirely fixed that case but also made collision-only MJCF, URDF, and USD assets invisible.

## Changes

- Rename the shared policy input to `model_has_visual_shapes` and preserve fallback visibility only for collision-only imports.
- Pre-scan URDF links for authored visual geometry.
- Use the set of effectively visible USD visual shapes for the import-wide decision while preserving the existing per-body `hide_collision_shapes` override.
- Defer MJCF fallback visibility until all geoms in the import have been classified.
- Add MJCF, URDF, and USD regression coverage for mixed visual/collision assets and collision-only assets.
- Update the Unreleased changelog entry to cover all three importers.
- Merge current `main`.

## Validation

- `uvx pre-commit run --files CHANGELOG.md newton/_src/utils/import_utils.py newton/_src/utils/import_urdf.py newton/_src/utils/import_mjcf.py newton/_src/utils/import_usd.py newton/tests/test_import_mjcf.py newton/tests/test_import_urdf.py newton/tests/test_import_usd.py`
- `python -m unittest newton.tests.test_import_mjcf.TestImportMjcfBasic` — 22 passed
- `python -m unittest newton.tests.test_import_urdf.TestImportUrdfBasic` — 22 passed
- `python -m unittest newton.tests.test_import_usd.TestImportSampleAssetsComposition` — 36 passed
- Confirmed the four new collision-only/invisible-visual assertions fail against the previous PR implementation and pass with the per-import policy.

## Not run

- Full Newton test suite.
- Maintainer-gated AWS GPU workflows.

## Summary by CodeRabbit

* **Bug Fixes**
  * Collision shapes no longer become visible by default when the imported model/body/link has no visible visual geometry.
  * Import behavior is now consistent across USD, MJCF, and URDF, keeping colliders hidden unless explicitly enabled (such as forced-collider behavior or treating visuals as colliders).
* **Tests**
  * Updated USD collision/visibility test expectations.
  * Added an MJCF test covering default collider/visual visibility and the effect of forcing colliders.


ruziniuuuuu:
Before (in gl viewer):

Image

Before (in rtx viewer):

Image

After (in gl viewer):

Image

After (in rtx viewer):

Image

adenzler-nvidia (changes requested):
Thanks for the report and the PR — the underlying issue in #3291 is real: the per-body fallback leaks an importer heuristic into rendering, and collision proxies on child bodies shouldn't render when the asset authors an explicit visual/collision split.

Removing the fallback entirely over-corrects, though. Assets that author *only* collision geometry — classic MJCF without visual/collision classes (`no_class_as_colliders=True` routes every geom to the colliders list), physics-only USD, URDF links without `<visual>` — now import with zero `VISIBLE` shapes and render an empty viewport, since the viewer's "Show Collision" toggle defaults to off. Several of Newton's own bundled assets hit this:

- `nv_ant.xml` → 0/14 shapes visible (was 14/14)
- `nv_humanoid.xml` → 0/42 (was 20/42)
- `tabletop.xml` → 0/23, `cartpole.usda` → 0/4, `quadruped.urdf` → 0/13

which blanks out `example_basic_urdf`, `example_robot_cartpole`, the selection examples, and others. Note the change also applies to URDF via the shared helper, which the PR description, docs, and CHANGELOG entry don't mention.

Suggestion: keep the fallback but change its scope from per-body to per-import. Hide colliders by default only when the import authored any visible visual geometry anywhere; if the asset is collision-only, keep showing colliders (nothing else would render). This fixes the wheel/roller case from #3291 (the model authors a visual on the wheel) with no behavior change for any bundled asset, and no migration note needed. Concretely:

- `should_show_collider()`: keep the three-way logic, rename `has_visual_shapes` → `model_has_visual_shapes`. This also removes the "retained for API compatibility" dead parameter — the helper is internal, so there's no compatibility to preserve.
- URDF: pre-scan before the link loop: `model_has_visual_shapes = any(len(link.findall("visual")) > 0 for link in urdf_links)`.
- USD: visual shapes are parsed before colliders, so pass `load_visual_shapes and len(bodies_with_visual_shapes) > 0`. Since that set only records effectively visible visuals, the invisible-visual edge case keeps its current behavior for free.
- MJCF: geom classification is interleaved with body traversal, so defer the decision: create colliders hidden (unless `force_show_colliders`/`parse_visuals_as_colliders`), collect their shape indices, and at the end of `parse_mjcf` (before `collapse_fixed_joints()`) OR the `VISIBLE` flag back in if no visual geoms were parsed.

On tests: your new MJCF regression test and the flipped `BodyWithoutVisuals` expectation in `test_collision_shape_visibility_flags` hold as-is under this design; the `load_visual_shapes=False` and invisible-visual-sibling flips revert to the current expectations. Please also add a test pinning that a collision-only asset still imports visible (that's the gap that let this change pass locally), and a URDF counterpart of the MJCF test since that path currently has no visibility coverage.

The CHANGELOG entry then becomes a plain fix, e.g. "Fix imported collision-only bodies rendering as visuals when the asset authors visual geometry elsewhere (#3291)", noting it applies to MJCF/URDF/USD imports.

One trade-off worth stating in the PR description: the scope is a single `add_mjcf`/`add_urdf`/`add_usd` call, so a scene file that authors visuals anywhere will hide collision-only props elsewhere in the same file. That's deliberate and matches the issue's own framing — visibility follows the authored geometry role.

The branch also needs a merge with main (CHANGELOG conflict).

ruziniuuuuu:
Implemented the requested per-import collider visibility policy in ad246902: collision-only MJCF/URDF/USD imports remain visible, while collision-only bodies stay hidden when the same importer call has visual geometry. I also added collision-only and URDF coverage, restored the USD fallback cases, updated the changelog/description, and merged current main. @adenzler-nvidia, please take another look.

adenzler-nvidia:
Thanks for the update — the per-import implementation looks good. I verified locally that visibility is unchanged vs main across all bundled example assets (MJCF/URDF/USD), and the new regression tests correctly fail against main's importer.

One small cleanup before this can go in: the merge with main duplicated a changelog entry. `Fix ViewerFile.is_running()... (#3094)` now appears under `[Unreleased]`, but that fix already shipped in 1.4.0 (listed there as `#3190; fixes #3094`) — the release merge-back restructured the changelog after this branch was created, and the union-style merging re-added the old line without a conflict. Please delete the `[Unreleased]` copy; your own #3291 entry is correctly placed and can stay as is.

ruziniuuuuu:
@adenzler-nvidia Removed the duplicate `[Unreleased]` `ViewerFile.is_running()` entry in 1359b061 while keeping the #3291 entry. I also merged the latest `main` (which now includes #3289). Validation: pre-commit passed for `CHANGELOG.md`, and the MJCF/URDF/USD importer suites passed 80/80. Please take another look when convenient.

adenzler-nvidia (approved):
Thanks for the quick turnaround, approved!

PR #3294, opened 2026-06-30
Support integer-indexed articulation selection

camevor:
## Description
ArticulationView was the remaining label-selection API that only accepted a single glob. 

Widen `ArticulationView`'s `pattern` argument to accept, in addition to a single glob string, a list of patterns or a list of explicit articulation indices. Explicit indices are sorted into model order (so layouts match string-pattern selection), checked for duplicates, and bounds-checked.

Allows callers to perform custom matching, e.g. for #3272.

## Checklist

## Test plan

```
```

## New feature / API change

```python
import newton

# Multiple patterns
view = newton.ArticulationView(model, pattern=["robot_l_*", "robot_r_*"])

# Explicit articulation indices
view = newton.ArticulationView(model, pattern=[0, 2, 3, 5])
```

## Summary of Changes

- **New Features**
  - Enhanced articulation selection to accept label patterns (single or list) and explicit integer index lists.
- **Bug Fixes**
  - Added stricter validation for explicit indices, including duplicate detection and clearer errors for out-of-range selections.
- **Documentation**
  - Updated the changelog to reflect the expanded selector support for articulation views.
- **Tests**
  - Added tests covering string patterns, index-list behavior (including order), duplicates, and out-of-range indices.


adenzler-nvidia:
This looks good to me and seems like a useful additive improvement to bring `ArticulationView` in line with the other selection APIs. @nvlukasz, do you have an opinion on this approach?

eric-heiden:
🔵 Integer-indexed articulation selection is a useful addition. The main concern is that explicit indices are silently reordered, which conflicts with the documented selection semantics and can misalign caller-provided values.

newton/_src/utils/selection.py
```diff
@@ -382,18 +382,31 @@ def get_name_from_label(label: str):
...
 
 
-def find_matching_ids(pattern: str, labels: list[str], world_ids, world_count: int):
+def find_matching_ids(
+    pattern: str | list[str] | list[int], labels: list[str], world_ids, world_count: int
+) -> tuple[list[list[int]], list[int]]:
+    matching_ids = match_labels(labels, pattern)
+
+    if isinstance(pattern, list) and pattern and isinstance(pattern[0], int):
+        # ArticulationView derives its layouts from model order. String patterns
```
eric-heiden: 🟠 Sorting `matching_ids` silently changes the caller's requested order. The shared selection documentation says integer indices select directly by index, and callers may prepare getter or setter values in that same order. For example, `[2, 0]` becomes `[0, 2]`, so values intended for articulation 2 can be applied to articulation 0.

Since `ArticulationView` requires model order for its stride-derived layouts, please either reject unsorted integer selectors with a clear `ValueError` or explicitly document that this API normalizes integer selectors to model order and differs from APIs that preserve caller order.

camevor: Good point, thanks! Fixed in 6e3100871 by raising when indices are unsorted.

nvlukasz (approved):
LGTM, I like that we now consistently match articulation/body/joint labels.

eric-heiden (approved):

PR #3297, opened 2026-07-01
Guard USD import against NaN mass values

### LINKED ISSUE #3024: Consider adding guards against NaN attribute values in USD importer
The USD importer in `newton/_src/utils/import_usd.py` reads numeric attributes (mass, density, inertia, friction, contact margin/gap, shell thickness, joint limits, mesh-collision parameters, SDF / hydroelastic fields, etc.) and validates many of them with `<= 0` / `< 0` / `>= 0` range checks. None of these checks reject `NaN`, so a `NaN`-valued attribute on any prim propagates unfiltered through the parser into `finalize()` and downstream simulation state (`NaN` AABB extents, `NaN` inertia, `NaN` limits) where the symptoms are non-obvious. Consider an importer-wide audit to add `NaN `guards.

han-xudong:
## Description

Guard USD import against authored NaN mass, density, and diagonal inertia values so they warn and fall back instead of propagating non-finite values into Newton model state.

This keeps the change narrow to USD mass-property import paths:

- body `MassAPI.mass` must be finite and non-negative before it overrides computed mass
- collider-authored `MassAPI.mass` / `diagonalInertia` must be finite before deriving mass information
- authored collider/material density must be finite, otherwise the importer uses the default shape density
- non-finite `ComputeMassProperties()` mass falls back to builder-accumulated mass properties

Closes #3024.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k import_usd_nan
uvx pre-commit run -a
```

## Bug fix

**Steps to reproduce:**

1. Create a USD rigid body or collider with `UsdPhysics.MassAPI` authored numeric fields set to `NaN`, such as body `mass`, collider `mass`, collider `density`, or diagonal inertia.
3. Without this PR, some NaN values can pass through authored-property checks and reach imported mass or inertia state without a clear warning.
4. With this PR, the importer warns and falls back to finite computed/default mass properties.

**Minimal reproduction:**

```python
import newton

body = UsdGeom.Xform.Define(stage, "/World/Body").GetPrim()
UsdPhysics.RigidBodyAPI.Apply(body)
UsdPhysics.MassAPI.Apply(body).CreateMassAttr().Set(float("nan"))

sphere = UsdGeom.Sphere.Define(stage, "/World/Body/Sphere")
sphere.CreateRadiusAttr().Set(0.1)
UsdPhysics.CollisionAPI.Apply(sphere.GetPrim())

builder.add_usd(stage)
```

The added tests cover body NaN mass, collider NaN mass-derived density, and collider NaN density fallback paths.

## Summary by CodeRabbit

* **Bug Fixes**
  * USD import now detects non-finite (NaN/inf) authored mass, density, and inertia values, warns, and falls back to safe defaults/computed properties instead of propagating invalid values into the model.
  * Added stricter validation for collider and rigid-body mass/inertia inputs to ensure resulting mass and inertia remain finite.

* **Tests**
  * Added regression coverage for NaN authored body mass, collider mass/density, and verified warnings and finite imported mass/inertia.

* **Documentation**
  * Updated the changelog with an Unreleased “Fixed” entry for the NaN-handling behavior.


adenzler-nvidia:
Thanks @han-xudong for the contribution and for tackling #3024 — the NaN-blindness of the existing sign checks was a real hole (`NaN <= 0` is false, so a NaN mass was accepted as a valid authored value), and your tests caught genuine NaN propagation into `body_mass`.

In the meantime, #3418 has reworked the same MassAPI code paths to follow the UsdPhysics value semantics: non-finite, non-positive, or blocked attributes are treated as unspecified and fall back accordingly. That subsumes the guards here (mass, diagonal inertia, and the `ComputeMassProperties` result) and additionally covers center of mass and body-level density. We've also folded the remaining pieces of this PR into that branch — the warning on non-finite material density and NaN test coverage for the mass paths.

Closing in favor of #3418. The broader importer-wide audit requested in #3024 stays open; the current thinking there is a bulk finiteness validation over the builder arrays rather than per-attribute guards. Thanks again!

Issue #2479, opened 2026-04-16
Determinism in Newton


eric-heiden:
Enable Newton to produce deterministic results, i.e. same-code, same-scene, same-machine reproducibility.

eric-heiden:
Dependency / limitation note for the Newton determinism work:

- Warp dependency: [NVIDIA/warp#1355](https://github.com/NVIDIA/warp/pull/1355) by @mmacklin is needed here. It adds the `wp.config.deterministic` support for deterministic handling of supported atomic patterns, including the scatter/sort/reduce path for accumulation atomics and the two-pass count/scan/execute path for counter-style atomics. The local Newton/Warp work depends on those changes.

- `wp.capture_while` limitation: MuJoCo Warp's default `m.opt.graph_conditional = True` path enters `wp.capture_while(...)` for solver iterations. With Warp deterministic mode enabled, deterministic launches currently allocate scratch buffers during the captured body: Python-side scatter/counter/overflow buffers, native CUB sort/reduce temporary buffers, and scan temporary buffers. A normal warm-up step before capture does not solve this; it can warm kernels/mempools, but the allocation/free calls still occur while capturing the conditional body, so Warp rejects the conditional graph with `Conditional body graph contains an unsupported operation (memory allocation)`.

  For now we need to avoid that conditional body in deterministic runs by setting MuJoCo Warp's `m.opt.graph_conditional = False` before graph capture, or by disabling graph capture for these runs. Setting it to `False` uses the fixed `for range(m.opt.iterations)` path instead of `wp.capture_while`. One caveat: with the default 100 solver iterations, fully unrolled capture can be quite large/slow, so a Newton-side no-capture path for MuJoCo + Warp deterministic mode may still be the more practical test-harness option.

Relevant MuJoCo Warp determinism PRs I found:

- [google-deepmind/mujoco_warp#1281](https://github.com/google-deepmind/mujoco_warp/pull/1281) - "Determinism support 1/N": adds opt-in `opt.deterministic` and sorts contacts after narrowphase to remove contact-index permutation from atomic counters.
- [google-deepmind/mujoco_warp#1300](https://github.com/google-deepmind/mujoco_warp/pull/1300) - "Determinism support 2/N": extends `opt.deterministic` to deterministic constraint row allocation via count -> scan -> emit and persisted scratch buffers. Its notes also call out graph capture with `opt.deterministic=True` as not yet supported.

jvonmuralt:
[...]

PR #3300, opened 2026-07-01
Add deterministic solver configuration

eric-heiden:
## Description

Add an optional `deterministic` constructor argument to `SolverXPBD`,
`SolverSemiImplicit`, `SolverFeatherstone`, `SolverVBD`, and `SolverMuJoCo`.
Each solver applies the resolved `wp.DeterministicMode` to the kernel modules
it uses and derives any required deterministic atomic record capacity from its
model and solver configuration. No record-capacity tuning is exposed through
the public API.

`deterministic=None` inherits the current `wp.config.deterministic` value at
construction. Concrete solvers retain their generic module options and reapply
them before `step()` and other public kernel-launching hooks, so sequential live
solver instances cannot leak modes or record bounds into one another.

Add bit-exact repeated-rollout coverage under `newton/tests/determinism`, plus
focused API, live-instance, module-reset, MuJoCo scoping, and record-bound tests.
Update the Warp nightly requirement to `warp-lang>=1.15.0.dev20260626` and keep
ASV on the compatible `1.15.0.dev20260626` build, both of which include the
required deterministic code generation. Use Warp's public marching-cubes
corner table to keep hydroelastic benchmarks compatible with the nightly.

Existing solver behavior remains unchanged when the effective mode is
`NOT_GUARANTEED`.

Addresses #2479.

## Checklist

## Test plan

```powershell
uv venv .venv-pr --python 3.12
$env:UV_PROJECT_ENVIRONMENT = ".venv-pr"
uv sync --extra dev --frozen
uvx pre-commit run -a
uv run --frozen -m newton.tests -k test_solver_determinism
uv run --frozen -m unittest -v newton.tests.determinism.test_solver_determinism newton.tests.test_api newton.tests.test_mujoco_version_check
```

The CUDA rollout tests pass bit-exactly on an RTX 4090. The XPBD regression
test fails on `origin/main` because the baseline constructor rejects
`deterministic`. A built Newton wheel also installs successfully with ordinary
`pip` in a disposable `uv venv` after ASV's pinned Warp nightly is installed.

## New feature / API change

```python
import warp as wp

from newton.solvers import SolverXPBD

solver = SolverXPBD(
    model,
    deterministic=wp.DeterministicMode.RUN_TO_RUN,
)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -151,6 +151,7 @@
 - Add `ViewerRTX`, a real-time ray-traced viewer powered by NVIDIA OVRTX. (#1861)
 - Add edge overlay toggle (`renderer.draw_edges`) for wireframe visualization on top of solid geometry
 - Add `newton.utils.ColorSpace`, `color_srgb_to_linear()`, `color_linear_to_srgb()`, and `SensorTiledCamera.RenderConfig.output_color_space` for color-space boundaries. (#2411)
+- Add a `deterministic` constructor argument to `SolverXPBD`, `SolverSemiImplicit`, `SolverFeatherstone`, `SolverVBD`, and `SolverMuJoCo` to configure solver kernel modules for Warp's deterministic codegen
 - Add `cable_cross_slide_table` example demonstrating a cable-driven XY table
 - Add robotics tutorial notebook covering ModelBuilder, solvers, CUDA graphs, IK, and pick-and-place
 - Add rigid-soft contact example covering a rigid sphere dropping onto an XPBD tetrahedral soft grid
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -151,7 +151,7 @@
 - Add `ViewerRTX`, a real-time ray-traced viewer powered by NVIDIA OVRTX. (#1861)
 - Add edge overlay toggle (`renderer.draw_edges`) for wireframe visualization on top of solid geometry
 - Add `newton.utils.ColorSpace`, `color_srgb_to_linear()`, `color_linear_to_srgb()`, and `SensorTiledCamera.RenderConfig.output_color_space` for color-space boundaries. (#2411)
-- Add a `deterministic` constructor argument to `SolverXPBD`, `SolverSemiImplicit`, `SolverFeatherstone`, `SolverVBD`, and `SolverMuJoCo` to configure solver kernel modules for Warp's deterministic codegen
+- Add a `deterministic` constructor argument to `SolverXPBD`, `SolverSemiImplicit`, `SolverFeatherstone`, `SolverVBD`, and `SolverMuJoCo` to opt into deterministic solver execution.
 - Add `cable_cross_slide_table` example demonstrating a cable-driven XY table
 - Add robotics tutorial notebook covering ModelBuilder, solvers, CUDA graphs, IK, and pick-and-place
 - Add rigid-soft contact example covering a rigid sphere dropping onto an XPBD tetrahedral soft grid
--- a/newton/_src/solvers/mujoco/solver_mujoco.py
+++ b/newton/_src/solvers/mujoco/solver_mujoco.py
@@ -571,6 +571,15 @@ class SolverMuJoCo(SolverBase, CouplingInterface):
         warp_util._KERNEL_CACHE.clear()
         SolverMuJoCo._generated_kernel_deterministic_options = options
 
+    def _set_mujoco_warp_module_options(self) -> None:
+        """Configure loaded shared modules without overriding code-generated bounds."""
+        options = {
+            "deterministic": self._deterministic,
+            "deterministic_max_records": 0,
+        }
+        for module in [*_mujoco_warp_deterministic_modules(), kernels]:
+            self._set_module_options(options, module=module)
+
     @staticmethod
     def _parse_integrator(value: str | int, context: dict[str, Any] | None = None) -> int:
         """Parse integrator option: Euler=0, RK4=1, implicit=2, implicitfast=3."""
@@ -3343,12 +3352,7 @@ class SolverMuJoCo(SolverBase, CouplingInterface):
             # path and compacts contacts with atomics. Keep the resolved mode
             # as the single source of truth instead of relying on global Warp
             # determinism during module import.
-            options = {
-                "deterministic": self._deterministic,
-                "deterministic_max_records": self._deterministic_max_records,
-            }
-            for module in [*_mujoco_warp_deterministic_modules(), kernels]:
-                self._set_module_options(options, module=module)
+            self._set_mujoco_warp_module_options()
 
         # Deferred from module scope: wp.static() in this kernel imports mujoco_warp.
         if SolverMuJoCo._convert_mjw_contacts_to_newton_kernel is None:
@@ -3604,6 +3608,14 @@ class SolverMuJoCo(SolverBase, CouplingInterface):
             wp.config.deterministic = original_mode
             wp.config.deterministic_max_records = original_max_records
 
+    @contextmanager
+    def _scoped_mujoco_warp_execution(self):
+        """Prepare and apply all solver-local MJWarp compilation options."""
+        self._apply_module_options()
+        self._prepare_generated_kernels()
+        with self._scoped_deterministic_config():
+            yield
+
     @event_scope
     def _mujoco_warp_step(self):
         self._mujoco_warp.step(self.mjw_model, self.mjw_data)
@@ -3611,9 +3623,6 @@ class SolverMuJoCo(SolverBase, CouplingInterface):
     @event_scope
     @override
     def step(self, state_in: State, state_out: State, control: Control, contacts: Contacts, dt: float) -> None:
-        self._apply_module_options()
-        if not self.use_mujoco_cpu:
-            self._prepare_generated_kernels()
         if self.use_mujoco_cpu:
             self._apply_mjc_control(self.model, state_in, control, self.mj_data)
             if self.update_data_interval > 0 and self._step % self.update_data_interval == 0:
@@ -3623,17 +3632,16 @@ class SolverMuJoCo(SolverBase, CouplingInterface):
             self._mujoco.mj_step(self.mj_model, self.mj_data)
             self._update_newton_state(self.model, state_out, self.mj_data, state_prev=state_in)
         else:
-            self._enable_rne_postconstraint(state_out)
-            self._apply_mjc_control(self.model, state_in, control, self.mjw_data)
-            if self.update_data_interval > 0 and self._step % self.update_data_interval == 0:
-                self._update_mjc_data(self.mjw_data, self.model, state_in)
-            self.mjw_model.opt.timestep.fill_(dt)
-            with wp.ScopedDevice(self.model.device), self._scoped_deterministic_config():
+            with wp.ScopedDevice(self.model.device), self._scoped_mujoco_warp_execution():
+                self._enable_rne_postconstraint(state_out)
+                self._apply_mjc_control(self.model, state_in, control, self.mjw_data)
+                if self.update_data_interval > 0 and self._step % self.update_data_interval == 0:
+                    self._update_mjc_data(self.mjw_data, self.model, state_in)
+                self.mjw_model.opt.timestep.fill_(dt)
                 if not self.mjw_model.opt.run_collision_detection:
                     self._convert_contacts_to_mjwarp(self.model, state_in, contacts)
                 self._mujoco_warp_step()
-
-            self._update_newton_state(self.model, state_out, self.mjw_data, state_prev=state_in)
+                self._update_newton_state(self.model, state_out, self.mjw_data, state_prev=state_in)
         self._step += 1
 
     @override
@@ -4061,7 +4069,13 @@ class SolverMuJoCo(SolverBase, CouplingInterface):
 
     @override
     def notify_model_changed(self, flags: ModelFlags | int) -> None:
-        self._apply_module_options()
+        if self.use_mujoco_cpu:
+            self._notify_model_changed(flags)
+        else:
+            with self._scoped_mujoco_warp_execution():
+                self._notify_model_changed(flags)
+
+    def _notify_model_changed(self, flags: ModelFlags | int) -> None:
         need_const_fixed = False
         need_const_0 = False
         need_length_range = False
@@ -6793,12 +6807,7 @@ class SolverMuJoCo(SolverBase, CouplingInterface):
                     self._deterministic_max_records = _mujoco_warp_deterministic_max_records(
                         self.mj_model, self.mjw_data
                     )
-                options = {
-                    "deterministic": self._deterministic,
-                    "deterministic_max_records": self._deterministic_max_records,
-                }
-                for module in [*_mujoco_warp_deterministic_modules(), kernels]:
-                    self._set_module_options(options, module=module)
+                self._set_mujoco_warp_module_options()
                 self._prepare_generated_kernels()
 
             # expand model fields that can be expanded:
--- a/newton/_src/solvers/mujoco/solver_mujoco.py
+++ b/newton/_src/solvers/mujoco/solver_mujoco.py
@@ -252,6 +252,9 @@ def _mujoco_warp_deterministic_modules() -> list[Any]:
     ]
 
 
+_MUJOCO_WARP_DYNAMIC_RECORD_MODULES = frozenset({"mujoco_warp._src.smooth"})
+
+
 def _mujoco_warp_max_constraint_row_width(mj_model: MjModel) -> int:
     """Return a model-derived upper bound for sparse constraint row width."""
     nv = int(mj_model.nv)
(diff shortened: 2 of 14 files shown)


newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -372,6 +379,31 @@ def __init__(
...
+                max_buffer_size + NUM_THREADS_PER_COLLISION_PRIMITIVE - 1
+            ) // NUM_THREADS_PER_COLLISION_PRIMITIVE
+        if model.particle_count > 0:
+            wp.set_module_options(
+                {
+                    "deterministic": effective_deterministic,
+                    "deterministic_max_records": particle_deterministic_max_records,
+                },
+                module=particle_vbd_kernels,
+            )
```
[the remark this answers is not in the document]
eric-heiden: This is covered by the code-generated bound in the required Warp nightly. In that API, `deterministic_max_records=0` means "use the static lower bound," not zero record capacity. `accumulate_particle_body_contact_force_and_hessian()` has no data-dependent loop and its atomic sites are statically visible, as are the spring atomics, so Warp derives their bound automatically. The explicit VBD override is only needed for the self-contact kernels' data-dependent buffer loops; that model-derived bound also includes the additional soft-contact record when those paths are combined.

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -346,6 +349,10 @@ def __init__(
                 Negative values are clamped to 0.
             rigid_joint_angular_kd: Damping coefficient for non-cable angular joint constraints [N·m·s/rad].
                 Negative values are clamped to 0.
+            deterministic: Opt-in determinism for this solver's atomic-emitting
```
jvonmuralt, line 352: I see that vbd_coupling_kernels is not conifgured to be run deterministically, so maybe it is worth stating explicitly in the docstring that coupled simulations are not covered by this flag.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -3492,13 +3587,33 @@ def __init__(
         if self.mjw_model is not None:
             self.mjw_model.opt.run_collision_detection = use_mujoco_contacts
 
+    @contextmanager
+    def _scoped_deterministic_config(self):
```
jvonmuralt, line 3601: I think there is a gap here: `_scoped_deterministic_config` only wraps `step()`,
but `__init__` also creates MJWarp kernels. It calls `notify_model_changed(ModelFlags.ALL)`
after `_prepare_generated_kernels()`, so those kernels are compiled with the global
config, not the solver's mode. Concrete case: for a model whose kinematic tree has more than 64 dofs
(`M_BLOCK_DENSE_MAX`), `set_const_0` → `solve_LD` generates `_solve_LD_sparse_fused`
during `__init__`. That kernel accumulates with `wp.atomic_add`, so with
`deterministic=RUN_TO_RUN` it still gets compiled as `NOT_GUARANTEED` and the
guarantee is silently broken.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -242,6 +245,65 @@ def _version_lt(left: tuple[int, ...], right: tuple[int, ...]) -> bool:
...
+    tendon_pair_width = 2 * max((int(width) for width in mj_model.ten_J_rownnz), default=0)
+    flex_width = max((int(width) for width in mj_model.flexedge_J_rownnz), default=0)
+
+    # Actuator moment rows are state-dependent for some transmission types, so
+    # retain the full-DOF bound whenever the model contains actuators.
+    actuator_width = nv if int(mj_model.nu) > 0 else 0
+    return min(nv, max(1, body_pair_width, tendon_pair_width, flex_width, actuator_width))
+
+
+def _mujoco_warp_deterministic_max_records(mj_model: MjModel, mjw_data: MjWarpData) -> int:
```
jvonmuralt, line 290: I think there is a problem with `deterministic_max_records` for batched simulations.

In Warp this option is a floor for every atomic in the module, not a fallback —
`records_per_thread = max(static_bound, max_records)`. Here the value is
`max(njmax, ...)` and it is applied to all mujoco_warp modules. So kernel that runs
over `(nworld, njmax)` grid and needs 1 record per thread will allocate
`nworld * njmax * njmax` records.


Maybe better to not set module-wide bound at all and put the big value only in
`wp.config.deterministic_max_records` inside `_scoped_deterministic_config`, so it
reaches only the generated kernels with dynamic loops.

newton/tests/determinism/test_solver_determinism.py
```diff
@@ -0,0 +1,355 @@
...
+
+    for _ in range(20):
+        state_0.clear_forces()
+        solver.step(state_0, state_1, control, contacts, 1.0 / 240.0)
+        state_0, state_1 = state_1, state_0
+
+    return _snapshot(state_0, ("particle_q", "particle_qd"))
+
+
+def test_particle_determinism(test, device, solver_name):
```
jvonmuralt, line 89: Each test compares two RUN_TO_RUN rollouts for bit-equality, but nothing establishes that the same rollout is non-deterministic under NOT_GUARANTEED. For small single-world models on a fixed GPU, NOT_GUARANTEED atomics are frequently already run-to-run stable, so these assertions can pass even if the deterministic flag is a no-op. Could we add a baseline that shows the workload actually diverges without the flag (e.g. assert two NOT_GUARANTEED rollouts differ, then that two RUN_TO_RUN rollouts match)?

jvonmuralt (commented):
It looks good, I think only mujoco problems are worth checking before merging.

jvonmuralt (dismissed):

newton/_src/solvers/vbd/solver_vbd.py
jcarius-nv: Given the public SolverVBD determinism contract, I think this should be fixed rather than only documented. vbd_coupling_kernels contains atomic accumulations in the proxy wrench and particle-force harvest paths, and the public coupling_harvest_proxy_wrenches() / coupling_harvest_proxy_particle_forces() hooks launch them without applying the solver instance module options. A RUN_TO_RUN VBD solver can therefore execute those coupled paths as NOT_GUARANTEED, and another live solver can leave the module in its own mode. Please register vbd_coupling_kernels with the instance options and reapply them before the coupling hooks, or explicitly exclude coupled simulations from the supported determinism contract and add coverage.

newton/_src/solvers/mujoco/solver_mujoco.py
jcarius-nv: There is a concrete cache sequence worth covering in the fix: _prepare_generated_kernels() clears the MJWarp factory cache, then constructor-time notify_model_changed(ModelFlags.ALL) calls set_const_0() outside _scoped_deterministic_config(). _solve_LD_sparse_fused() is created and cached under the global mode. Later, entering the scoped mode in step() returns that same cached kernel, so its module stays NOT_GUARANTEED. I reproduced this by creating the unique kernel under NOT_GUARANTEED, switching the global mode to RUN_TO_RUN, and observing the cached kernel remain NOT_GUARANTEED until the factory cache was cleared and recreated. The fix should ensure constructor- and notify-time MJWarp kernel factories run under the solver-local config, with a greater-than-64-DOF regression that checks the generated unique-kernel mode.

newton/tests/determinism/test_solver_determinism.py
jcarius-nv: The current rollout can indeed pass if the flag is a no-op, but I would avoid asserting that two NOT_GUARANTEED runs must differ. NOT_GUARANTEED permits stable results, so that assertion would be workload-, driver-, and hardware-dependent. A stronger non-flaky regression is to verify a deterministic sentinel at the configuration and compilation boundary: for example, assert the generated greater-than-64-DOF MuJoCo unique kernel and the VBD coupling module actually carry RUN_TO_RUN, then retain the bit-exact rollout as the end-to-end check.

newton/_src/solvers/xpbd/solver_xpbd.py
```diff
@@ -111,8 +112,18 @@ def __init__(
         rigid_contact_con_weighting: bool = True,
         angular_damping: float = 0.0,
         enable_restitution: bool = False,
+        deterministic: wp.DeterministicMode | None = None,
```
jcarius-nv, line 115: The new deterministic parameter is public, but unlike the other four solver constructors it has no Args documentation explaining the accepted DeterministicMode values, that None inherits wp.config.deterministic at construction, or the default behavior. Please add Google-style API documentation for it, consistent with the other solver constructors.

jcarius-nv:
@eric-heiden I tried my hand at implementing the raised issues in https://github.com/eric-heiden/newton/pull/7 . Feel free to merge this in here, then I'd be ok with getting this MR #3300 merged.

pyproject.toml
```diff
@@ -28,7 +28,7 @@ description = "A GPU-accelerated physics engine for robotics simulation"
 readme = "README.md"
 requires-python = ">=3.10"
 dependencies = [
-    "warp-lang>=1.15.0.dev20260626",
+    "warp-lang>=1.16.0.dev20260627",
```
adenzler-nvidia: can we avoid going up to 1.16 here, and use a 1.15 version that has the determinism support until we cut the release branch?

jcarius-nv:
I merged my changes into this branch after discussions with @adenzler-nvidia and confirmation from @jvonmuralt so we see how CI turns out.

pyproject.toml
eric-heiden: I updated Warp to `1.15.0.dev20260626` which should have the determinism support.

CHANGELOG.md
```diff
@@ -151,6 +151,7 @@
 - Add `ViewerRTX`, a real-time ray-traced viewer powered by NVIDIA OVRTX. (#1861)
 - Add edge overlay toggle (`renderer.draw_edges`) for wireframe visualization on top of solid geometry
 - Add `newton.utils.ColorSpace`, `color_srgb_to_linear()`, `color_linear_to_srgb()`, and `SensorTiledCamera.RenderConfig.output_color_space` for color-space boundaries. (#2411)
+- Add a `deterministic` constructor argument to `SolverXPBD`, `SolverSemiImplicit`, `SolverFeatherstone`, `SolverVBD`, and `SolverMuJoCo` to configure solver kernel modules for Warp's deterministic codegen
```
jcarius-nv: Could we phrase this in terms of the user-facing outcome rather than the implementation details (“kernel modules” and “Warp’s deterministic codegen”)? For example: `Add a deterministic constructor argument to SolverXPBD, SolverSemiImplicit, SolverFeatherstone, SolverVBD, and SolverMuJoCo to opt into deterministic solver execution.`

jcarius-nv (dismissed):
Approving based on the current implementation and green CI, conditional on the outstanding review threads being addressed and resolved. Once the changelog wording and PR description are updated and the existing threads are closed, this is ready to merge from my side.

jcarius-nv:
The Warp version in the PR description is stale after `93e76d09`: it still says `warp-lang>=1.16.0.dev20260627` and ASV `1.16.0.dev20260627`, while the current branch uses `warp-lang>=1.15.0.dev20260626` and ASV `1.15.0.dev20260626`. Please update the description so it matches the current dependency and ASV configuration.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -495,6 +558,28 @@ def import_mujoco(cls):
...
+        from mujoco_warp._src import warp_util
+
+        warp_util._KERNEL_CACHE.clear()
+        SolverMuJoCo._generated_kernel_deterministic_options = options
+
+    def _set_mujoco_warp_module_options(self) -> None:
+        """Configure loaded shared modules without overriding code-generated bounds."""
+        options = {
+            "deterministic": self._deterministic,
+            "deterministic_max_records": 0,
```
adenzler-nvidia: `deterministic_max_records=0` here overrides the scoped model-derived bound for MJWarp source modules. This still breaks deterministic construction for sparse articulations: a 70-revolute-joint chain with `SolverMuJoCo(..., disable_contacts=True, deterministic=wp.DeterministicMode.RUN_TO_RUN)` fails in constructor-time `set_const_0()` with `Deterministic scatter buffer overflow in kernel '_M'`.

I instrumented this path and confirmed that Newton computes `_deterministic_max_records=2485`, while `mujoco_warp._src.smooth` remains configured with `deterministic_max_records=0`. Applying the computed bound to that module makes the same model construct successfully; the identical model also succeeds under `NOT_GUARANTEED`.

Could we provide a targeted bound for affected MJWarp source kernels/modules without reintroducing the previous blanket allocation, and add a >64-DOF constructor regression? The current five-DOF rollout does not exercise this path.

newton/_src/solvers/mujoco/solver_mujoco.py
jcarius-nv: Addressed in `681f86f7`: GPU `step()` and `notify_model_changed()` now run inside `_scoped_mujoco_warp_execution()`, which reapplies instance options, refreshes generated-kernel cache state, and activates the solver-local Warp configuration while factory kernels are created. `test_notify_model_changed_uses_solver_config_for_mjwarp` covers the notify-time scope. The separate source-module record-bound overflow is tracked in https://github.com/newton-physics/newton/pull/3300#discussion_r3549741569. Resolving this thread.

CHANGELOG.md
jcarius-nv: Addressed in `cc86a40d`: the changelog now describes the user-facing outcome as opting into deterministic solver execution, without referring to kernel modules or code generation internals. Resolving this thread.

newton/_src/solvers/vbd/solver_vbd.py
jcarius-nv: Addressed in `681f86f7`: `vbd_coupling_kernels` is registered with the solver-local deterministic options, and the public coupling hooks reapply those options before launching. `test_vbd_coupling_hook_reapplies_deterministic_options` covers cross-instance state. Resolving this thread.

newton/_src/solvers/mujoco/solver_mujoco.py
jcarius-nv: Addressed in `681f86f7`: shared MJWarp modules keep `deterministic_max_records=0`, while the model-derived bound is applied through the scoped configuration for dynamically generated kernels. `test_loaded_modules_keep_codegen_record_bound` covers the allocation behavior. The separate sparse source-module under-allocation is tracked in https://github.com/newton-physics/newton/pull/3300#discussion_r3549741569. Resolving this original blanket-allocation concern.

newton/_src/solvers/xpbd/solver_xpbd.py
jcarius-nv: Addressed in `681f86f7`: the XPBD constructor now documents the `deterministic` argument, accepted `DeterministicMode`, and `None` inheritance behavior. Broader completion of the pre-existing constructor argument documentation is postponed. Resolving this thread.

pyproject.toml
jcarius-nv: Addressed in `93e76d09`: `pyproject.toml` and ASV now use Warp `1.15.0.dev20260626`, and the lockfile was refreshed accordingly. The PR description has also been updated to match. Resolving this thread.

newton/tests/determinism/test_solver_determinism.py
jcarius-nv: Addressed in `681f86f7`: instead of requiring `NOT_GUARANTEED` runs to diverge, the tests now assert configuration/compilation boundaries for VBD coupling, MJWarp cache/scoped configuration, and module record bounds, while retaining the bit-exact rollout checks. The new >64-DOF constructor regression requested for the separate overflow remains tracked in https://github.com/newton-physics/newton/pull/3300#discussion_r3549741569. Resolving this original flaky-baseline request.

newton/_src/solvers/mujoco/solver_mujoco.py
jcarius-nv: After discussion: For now we'll re-apply th large record bound to the MJWarp smooth module, alongside a regression test. There's no extra API risk for this change, so we'll get it into this release.

@eric-heiden  Longer-term we do need to have a discussion though on how to handle this more generally or automatically, since changes in a kernel can easily exceed the bound again.

jcarius-nv: Addressed in `03a82083`: the model-derived record bound is now applied only to `mujoco_warp._src.smooth`; generated and unrelated MJWarp modules remain at `deterministic_max_records=0`, so this does not restore the previous blanket allocation.

Added `test_mujoco_sparse_articulation_construction`, which builds the 70-revolute-joint chain through the public API, constructs a GPU `RUN_TO_RUN` `SolverMuJoCo`, and executes one step. It reproduced the `_M` deterministic scatter-buffer overflow before the fix and passes afterward with Warp `1.15.0.dev20260626` on an NVIDIA L4. The complete 14-test solver determinism suite also passes.

adenzler-nvidia (approved):

newton/_src/solvers/mujoco/solver_mujoco.py
jcarius-nv: #3414

PR #3304, opened 2026-07-01
Fix XPBD particle contact NaN guards

### LINKED ISSUE #1562: xpbd: reduce nans in solve_particle_particle_contacts
1. d = wp.length(n) can be 0.0 and then n = n / d will produce NaNs.  this can happen if you teleport two particles to the same location.

2. delta_f = wp.normalize(vt) * lambda_f will produce NaNs if vt is near-zero.

han-xudong:
## Description

Fix XPBD particle-particle contacts so degenerate contact inputs do not write non-finite particle state.

- Skip exact-overlap particle pairs when no well-defined contact normal exists.
- Avoid normalizing zero-length tangential slip vectors in the friction correction.
- Add CPU/CUDA regressions for exact-overlap and zero-tangent cohesive particle contacts.

Closes #1562.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k nan_guard
uv run --extra dev -m newton.tests -k test_particle_particle_friction
uvx pre-commit run -a
```

## Bug fix

**Steps to reproduce:**

1. Build an XPBD model with two active particles at the same position and nonzero particle radius.
2. Step `SolverXPBD` so `solve_particle_particle_contacts()` evaluates the exact-overlap pair.
3. Without this PR, the contact normal divides by zero and can write non-finite particle positions/velocities.
4. A second degenerate friction path can occur when the tangential relative velocity is zero and the friction direction is normalized.

**Minimal reproduction:**

```python
import numpy as np
import warp as wp
import newton

builder = newton.ModelBuilder(up_axis="Y")
builder.add_particles(
    pos=[wp.vec3(0.0, 0.0, 0.0), wp.vec3(0.0, 0.0, 0.0)],
    vel=[wp.vec3(0.0), wp.vec3(0.0)],
    mass=[1.0, 1.0],
    radius=[0.5, 0.5],
)
model = builder.finalize()
model.set_gravity((0.0, 0.0, 0.0))

solver = newton.solvers.SolverXPBD(model=model, iterations=1)
state0 = model.state()
state1 = model.state()
solver.step(state0, state1, model.control(), model.contacts(), 1.0 / 60.0)

assert np.all(np.isfinite(state1.particle_q.numpy()))
assert np.all(np.isfinite(state1.particle_qd.numpy()))
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed XPBD particle–particle cohesive contact handling to prevent non-finite particle states when particles are exactly overlapping (including cases where the contact normal calculation would otherwise divide by zero).

* **Tests**
  * Added regression coverage for exact-overlap and tiny-separation contact stability, asserting particle positions/velocities remain finite and that contact resolution stays meaningfully active.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
 ### Fixed
 
 - Fix `ViewerFile.is_running()` to return `False` after `ViewerFile.close()` so headless recording loops can terminate like interactive viewers. (#3094)
+- Fix XPBD particle-particle contacts to avoid non-finite particle state for exact-overlap contacts and zero-tangent friction directions. (#1562)
 - Fix `SolverVBD` rigid contact injecting kinetic energy for yawed finite-radius contacts (e.g. small-radius cables blowing up). The normal response now acts at the geometric skeleton point rather than the rotating surface anchor, which was non-conservative under reorientation; friction still uses the surface anchor to preserve finite-radius slip. (#3125)
 - Fix `SolverKamino` contact filtering and constraint stabilization so gap/margin contacts are handled consistently, positive-distance contacts can be filtered as configured, and converted contact forces/wrenches populate matching Newton contact slots for `SensorContact`. (#2908)
 - Fix memory growth in the Style3D solver when CUDA Graph capture is disabled
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@
 ### Fixed
 
 - Fix `ViewerFile.is_running()` to return `False` after `ViewerFile.close()` so headless recording loops can terminate like interactive viewers. (#3094)
-- Fix XPBD particle-particle contacts to avoid non-finite particle state for exact-overlap contacts and zero-tangent friction directions. (#1562)
+- Fix XPBD particle-particle contacts to avoid non-finite particle state for exact-overlap contacts. (#1562)
 - Fix `SolverVBD` rigid contact injecting kinetic energy for yawed finite-radius contacts (e.g. small-radius cables blowing up). The normal response now acts at the geometric skeleton point rather than the rotating surface anchor, which was non-conservative under reorientation; friction still uses the surface anchor to preserve finite-radius slip. (#3125)
 - Fix `SolverKamino` contact filtering and constraint stabilization so gap/margin contacts are handled consistently, positive-distance contacts can be filtered as configured, and converted contact forces/wrenches populate matching Newton contact slots for `SensorContact`. (#2908)
 - Fix memory growth in the Style3D solver when CUDA Graph capture is disabled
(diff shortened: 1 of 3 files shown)


adenzler-nvidia (changes requested):
Thanks for taking this on—the exact-overlap failure is real and worth fixing. I don’t think the current patch is ready to merge yet, because it changes behavior beyond the demonstrated singularity and one of the new tests does not fail without the corresponding change.

The `d > 1.0e-8` condition needs reconsideration. This is not a general float32 safeguard: float32 precision is relative to the magnitude of the coordinates, rather than an absolute `1e-8` resolution. A separation of `5e-9` near the origin is representable and produces a finite normal, but this patch skips that valid contact. At larger coordinates, differences greater than `1e-8` may already round away. If the computed norm is exactly zero, including through underflow or loss of separation during subtraction, `d > 0.0` safely excludes the division. Unless there is a separate, scale-aware reason to ignore near-coincident contacts, the guard should therefore match the actual singularity: `d > 0.0`.

The friction rewrite also does not currently have a demonstrated regression. I ran the new zero-tangent test against the parent commit, and it already passes on CPU and CUDA. Newton’s minimum supported Warp version defines `wp.normalize(0)` to return the zero vector, so the existing expression is already safe for the tested case. The new `1.0e-8` velocity threshold consequently changes behavior without a test showing why that change is needed.

Could you please revise the patch to:

- Use `d > 0.0` for the undefined-normal case, or explain and test a scale-aware alternative.
- Keep `wp.normalize(vt)` unless there is a failing reproduction with a supported Warp version.
- Confirm each regression test fails against the parent commit for the reason it claims to cover.
- Add a small, nonzero-separation case that verifies the contact remains active and is resolved.

The underlying fix should remain small, and I’d be happy to review an updated version.

han-xudong:
@adenzler-nvidia Thank you for the careful review and for checking the tests against the parent commit.

I revised the patch to keep the behavior change limited to the demonstrated exact-overlap singularity:
- changed the guard to `d > 0.0`
- restored the existing `wp.normalize(vt)` friction path
- replaced the zero-tangent cohesion test with a tiny nonzero-separation test to ensure valid contacts remain active
- narrowed the changelog entry accordingly

I also reran:
- `uv run --extra dev -m newton.tests -k test_xpbd_particle_particle`
- `uv run --extra dev -m newton.tests -k test_solver_xpbd`
- `uvx pre-commit run -a`

adenzler-nvidia (approved):
Thanks for the contribution, and for the careful follow-up on the review feedback! The guard now matches the actual singularity, the friction path is left untouched, and the new tiny-separation test pins the behavior-preservation requirement nicely. I verified the exact-overlap regression test fails on the parent commit and passes with the fix on both CPU and CUDA.

PR #3305, opened 2026-07-01
Use analytic primitive SDFs for hydro texture generation

### LINKED ISSUE #3239: Hydroelastic primitive texture SDFs use fixed tessellated meshes instead of analytic primitive SDFs
Hydroelastic primitive texture SDF generation currently appears to convert primitives to temporary meshes, then calls `create_texture_sdf_from_mesh()`.

Relevant path:

- `ModelBuilder.finalize()` defines `_create_primitive_mesh()`
- primitives are converted with `Mesh.create_sphere()`, `create_capsule()`, `create_cylinder()`, etc.
- the result is passed to `create_texture_sdf_from_mesh()`

Because the temporary primitive meshes use fixed default tessellation, e.g. sphere/ellipsoid `32x32`, capsule/cylinder/cone `segments=32`, increasing `sdf_max_resolution` refines the SDF over that fixed mesh rather than converging to the analytic primitive.

This means high-resolution hydroelastic primitive SDFs can hit a hidden tessellation error floor. For boxes this is mostly fine because the mesh is exact, but curved primitives retain the approximation error from the temporary mesh.

Newton already has analytic primitive SDF kernels for sphere/box/capsule/cylinder/cone/ellipsoid used by the sparse SDF path, so it may be worth using those directly for primitive texture SDF generation instead of going through a temporary mesh.

han-xudong:
## Description

Build hydroelastic primitive texture SDFs by sampling the existing analytical primitive SDF functions instead of first converting primitives to temporary tessellated meshes. This removes the fixed primitive-mesh tessellation error floor for curved primitives while keeping mesh and convex-mesh SDF generation on the existing mesh path.

Closes #3239.

## Checklist

## Test plan

```bash
uvx pre-commit run -a
uv run --extra dev -m newton.tests -k test_sdf_texture
uv run --extra dev -m newton.tests -k test_stacked_primitive_cubes_hydroelastic_xpbd_no_reduction
```

## Bug fix

**Steps to reproduce:**

1. Create a hydroelastic primitive sphere with `sdf_max_resolution=64` and `sdf_texture_format="float32"`.
2. Finalize the model and sample the generated texture SDF near the sphere surface.
3. Compare sampled distances with the analytic sphere distance `length(p) - radius`.
4. Without this PR, the texture SDF follows the temporary tessellated sphere mesh and retains a fixed mesh-approximation error even as the texture SDF is refined.

**Minimal reproduction:**

```python
import numpy as np
import warp as wp
import newton

radius = 0.5
builder = newton.ModelBuilder(gravity=0.0)
body = builder.add_body()
cfg = newton.ModelBuilder.ShapeConfig(
    is_hydroelastic=True,
    sdf_max_resolution=64,
    sdf_texture_format="float32",
)
shape = builder.add_shape_sphere(body, radius=radius, cfg=cfg)
sdf_idx = int(model._shape_sdf_index.numpy()[shape])

# Sample the texture SDF via a Warp kernel and compare against
# np.linalg.norm(points, axis=1) - radius.
```

## Summary by CodeRabbit

* **New Features**
  * Added public APIs to build sparse and texture SDFs directly from analytic primitives (in addition to mesh inputs).
  * Updated model building to generate hydroelastic primitive texture SDFs via the new primitive-based path.
* **Bug Fixes**
  * Primitive texture SDFs now sample analytic primitive signed distances (no intermediate tessellated meshes).
  * Unsupported primitives produce empty SDF results; other failures safely fall back without breaking compaction bookkeeping.
* **Tests**
  * Added CUDA tests validating analytic sphere SDF accuracy and primitive input validation.
* **Documentation**
  * Updated **[Unreleased] → Fixed** in the changelog to note the analytic sampling improvement.


mzamoramora-nvidia (commented):
Thanks for this change. The analytic primitive path looks like a clear improvement. I measured better accuracy and faster SDF construction, and the sphere regression is a good high-signal test.

Could you update the branch with current `main`?

I left a few inline suggestions to reduce duplicated construction logic and make the generalized source flow easier to follow.

newton/_src/geometry/sdf_texture.py
```diff
@@ -1708,6 +1948,94 @@ def create_texture_sdf_from_mesh(
...
+        Tuple of ``(texture_sdf, coarse_texture, subgrid_texture)``.
+        Caller must keep texture references alive to prevent GC.
+    """
+    shape_type, shape_scale = _validate_primitive_sdf_inputs(shape_type, shape_scale)
+
+    min_prim, max_prim = get_primitive_extents(shape_type, shape_scale)
+    min_ext = np.asarray(min_prim, dtype=float) - margin
+    max_ext = np.asarray(max_prim, dtype=float) + margin
+
+    ext = max_ext - min_ext
```
mzamoramora-nvidia: Could we extract the construction steps shared with `create_texture_sdf_from_mesh()`? Resolution checks, grid sizing, narrow-band setup, texture creation, and `scale_baked` finalization are nearly identical.

One option is a private `_create_texture_sdf_from_source()` helper. It could accept the source extents and a bound sparse-data builder. The mesh and primitive entry points would only compute their extents, bind their respective builder, and delegate the shared steps.

newton/_src/geometry/sdf_texture.py
```diff
@@ -353,7 +435,10 @@ def _apply_subgrid_linearity_kernel(
 
 @wp.kernel
 def _build_coarse_sdf_from_mesh_kernel(
+    source_kind: wp.int32,
```
mzamoramora-nvidia: This kernel now supports both mesh and primitive sources. Could we rename it to `_build_coarse_sdf_kernel`?

newton/_src/geometry/sdf_texture.py
```diff
@@ -309,7 +391,7 @@ def _accumulate_subgrid_linearity_error_kernel(
         float(gy) * cell_size[1],
         float(gz) * cell_size[2],
     )
-    mesh_val = _query_mesh_sdf(mesh, pos, 10000.0, winding_threshold, use_parity)
+    mesh_val = _query_sdf(source_kind, mesh, shape_type, shape_scale, pos, 10000.0, winding_threshold, use_parity)
```
mzamoramora-nvidia: Could we use `mesh_id` for the optional mesh handle and `sdf_value` for the sampled result? Those names make the source dispatch easier to follow without renaming the genuinely mesh-specific helpers.

newton/_src/geometry/sdf_texture.py
```diff
@@ -222,7 +296,10 @@ def _write_subgrid_slot(
 
 @wp.kernel
 def _check_subgrid_occupied_kernel(
+    source_kind: wp.int32,
```
mzamoramora-nvidia: These kernels now query either a mesh or a primitive. Could we replace references to “mesh SDF” with “source SDF” in the generalized docstrings? Mesh-only helpers such as `_query_mesh_sdf()` can keep their current wording.

han-xudong: Updated the generalized kernel and module docstrings to use source-neutral SDF wording in `edd69855`; mesh-only helpers retain their existing terminology.

newton/_src/geometry/sdf_texture.py
han-xudong: Renamed the generalized kernel to `_build_coarse_sdf_kernel` and updated its launch site in `edd69855`.

newton/_src/geometry/sdf_texture.py
han-xudong: Extracted the shared resolution checks, grid sizing, narrow-band setup, texture creation, and `scale_baked` finalization into `_create_texture_sdf_from_source()` in `edd69855`. The mesh and primitive entry points now compute extents, bind their sparse-data builders, and delegate to the common helper.

newton/_src/geometry/sdf_texture.py
han-xudong: Updated the generalized source-dispatch parameters to `mesh_id` and the sampled linearity/subgrid values to `sdf_value` in `edd69855`. The genuinely mesh-specific helper retains its mesh naming.

han-xudong:
Hi @mzamoramora-nvidia, I merged the latest `upstream/main` and addressed all four inline suggestions in `edd69855`: the mesh and primitive entry points now delegate to a shared `_create_texture_sdf_from_source()` helper, and the generalized kernels use source-neutral names, variables, and docstrings. I also moved the changelog entry to `[Unreleased]` -> `Fixed`. The complete `TestTextureSDF` suite passes (28 CUDA tests), and `pre-commit` passes.

mzamoramora-nvidia (commented):
Thanks for the update. The shared texture-SDF construction helper and the source-neutral names make the implementation much easier to follow. The focused texture-SDF tests and stacked-cubes regression also pass locally.

I found one issue that needs to be fixed before merging. `ModelBuilder` now imports `create_texture_sdf_from_primitive`, but the full-surface mesh path still calls `create_texture_sdf_from_mesh`. Could you keep both imports? Without the mesh import, the `NameError` is caught by the fallback handler and seven full-surface mesh tests fail.

As a small optional cleanup, a few comments and docstrings still refer only to mesh samples even though the code is now shared by mesh and primitive sources.

Otherwise, the updated refactor looks good to me.

han-xudong:
Hi @mzamoramora-nvidia, I restored `create_texture_sdf_from_mesh` alongside the primitive helper and generalized the remaining shared-path comments/docstring to use source-neutral wording. I reproduced the regression before the fix: the full-surface suite hit the `NameError` fallback and failed two tests. After the fix, all 46 `TestFullSurfaceSoftContact` tests and all 28 `TestTextureSDF` tests pass, as does `pre-commit`. I also merged the latest `upstream/main`. Could you please take another look when convenient?

mzamoramora-nvidia (dismissed):
Thanks for the contribution!

mzamoramora-nvidia:
Hi @han-xudong, the GPU test job is failing in the new sphere texture SDF test.

`ModelBuilder(gravity=0.0)` emits a deprecation warning, and CI treats warnings as errors.

Could you use `gravity=(0.0, 0.0, 0.0)` instead? Since the test does not run a simulation, omitting the `gravity` argument would also work.

This appears to be the only test failure.

han-xudong:
@mzamoramora-nvidia Thank you for catching this. I updated the sphere texture SDF test to use an explicit zero gravity vector and merged the latest upstream/main. The merge also preserves upstream specialized mesh sign-mode kernels while adding source-specialized primitive kernels. Primitive accuracy, default mesh construction, open-mesh normal-sign, input validation, and pre-commit all pass. Could you please take another look when convenient?

mzamoramora-nvidia:
Hi @han-xudong, the latest `main` merge looks functionally good.

It seems to have reintroduced some mesh-specific wording in the generalized texture SDF kernel docstrings. Could you switch those references back from “mesh SDF” to “source SDF”? This affects the module summary and the occupancy, coarse-grid, and subgrid-population kernel docstrings.

The kernel names and implementation still look clean. This is just a small documentation cleanup after the merge.

han-xudong:
Hi @mzamoramora-nvidia, I switched the module summary and the generalized occupancy, linearity, coarse-grid, and subgrid-population documentation back to source-neutral wording in ee9b1fb7. I also updated the cache rationale to refer to the source SDF while retaining the mesh benchmark context. Pre-commit passes.

newton/_src/geometry/sdf_texture.py
mzamoramora-nvidia, line 1: Thanks @han-xudong for addressing all the comments and for the quick turnaround!

One small nitpick: `create_sparse_sdf_textures()` now accepts sparse data from both the mesh and primitive builders. Could we make the `sparse_data` description source-neutral?

> `sparse_data: Dictionary produced by a sparse SDF builder.`

han-xudong: Updated in 64729401. The sparse_data argument now uses the source-neutral description "Dictionary produced by a sparse SDF builder." Pre-commit passes.

mzamoramora-nvidia (approved):

PR #3310, opened 2026-07-02
Fix hollow shell thickness validation

### LINKED ISSUE #3313: [BUG] Invalid hollow primitive shell thickness is accepted
## Bug Description

Hollow primitive inertia accepts non-positive or oversized shell thickness values. For hollow primitives, the importer/builder uses `ShapeConfig.margin` as the shell thickness when computing inertia. If the thickness is equal to or larger than the primitive's inner dimension, inertia computation can subtract invalid inner dimensions and produce physically invalid mass properties instead of rejecting the input.

Related PR: #3310

## Reproduction Script

```python
import newton

cfg = newton.ModelBuilder.ShapeConfig(is_solid=False, margin=0.6)
builder.add_shape_sphere(-1, radius=0.5, cfg=cfg)
```

Expected: the invalid hollow shell thickness is rejected before inertia computation.

Actual: the invalid thickness is accepted.

## System Information

Not platform-specific.

han-xudong:
## Description

Reject invalid hollow primitive shell thickness before inertia computation subtracts inner dimensions. This keeps hollow sphere, box, capsule, cylinder, cone, and ellipsoid inertia from accepting non-positive or too-large shell thickness values that cannot fit inside the shape.

Closes #3313.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_inertia
uvx pre-commit run -a
```

## Bug fix

**Steps to reproduce:**

1. Add a hollow primitive with shell thickness equal to or larger than its smallest inner dimension.
2. Observe that the invalid thickness is accepted and can produce physically invalid mass properties.

**Minimal reproduction:**

```python
import newton

cfg = newton.ModelBuilder.ShapeConfig(is_solid=False, margin=0.6)
builder.add_shape_sphere(-1, radius=0.5, cfg=cfg)
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Hollow primitive inertia calculations now validate shell thickness up-front, rejecting invalid values before any inertia math is performed.
  * Thickness must be a finite real number; oversized, negative, non-real/boolean, or non-finite inputs now fail fast with clear `TypeError`/`ValueError` messages.
  * `thickness=0.0` now emits a warning and returns zero mass and a zero inertia tensor for the hollow part.
* **Tests**
  * Added/expanded test coverage for hollow thickness validation and sanity checks across supported primitives (sphere, box, capsule, cylinder, cone, ellipsoid).
* **Documentation**
  * Updated `CHANGELOG.md` with the fix note.


adenzler-nvidia (changes requested):
Thanks for the fix — the validation gap is real: a negative thickness silently subtracts mass from the body, and an oversized one produces a shell *heavier* than the solid, so failing fast on those is a clear improvement. Two changes are needed before this can merge.

**1. `thickness == 0` should warn and produce a zero-mass shell rather than raise.**

`ShapeConfig.margin` defaults to `0.0` and is what the builder passes as the shell thickness, so the raise fires for any hollow shape built without an explicit margin. More importantly, the USD importer creates hollow shapes for `newton:massModel = "shell"` and `mjc:shellinertia` prims and falls back to the collision margin when no `newton:shellThickness` is authored — with this PR, a single such prim aborts the entire import, including assets whose body mass is authored via `MassAPI` and overridden after shape parsing anyway. The importer's mass fallback (`_build_mass_info_from_shape_geometry`) is deliberately warn-and-skip on non-positive shell mass, and a raise inside `compute_inertia_shape()` bypasses that.

Please keep the hard errors for NaN, negative, and oversized thickness — those produce garbage under every interpretation. For `thickness == 0`, emit a warning and return zero mass/inertia directly (an early-out like the existing `density == 0.0` path), which also makes the six primitives consistent (the hollow cone currently raises on zero thickness while the others silently return zero mass). The `0.0` case in `test_hollow_primitive_thickness_must_be_positive` should assert the warning instead of the exception.

The underlying issue — the collision margin and the shell thickness sharing one field, so the contact-tuned default of 0 is nonsense as a wall thickness — is out of scope here; we'll track a dedicated `ShapeConfig.shell_thickness` field separately.

**2. The changelog entry will land in the released `[1.4.0]` section.**

This branch predates the 1.4 changelog merge-back, and the lines the new entry anchors to now sit inside `[1.4.0]` on main, so the merge will silently place the entry in a released section. Please merge main and move the entry into `[Unreleased]` → `Fixed`.

han-xudong:
@adenzler-nvidia Thank you for the detailed review. I updated zero thickness to emit a warning and return zero mass and inertia for all six analytic primitives, while negative, non-finite, and oversized values remain errors. I also moved the changelog entry to Unreleased / Fixed and merged the latest upstream/main. All five hollow primitive tests and pre-commit pass. Could you please take another look when convenient?

newton/tests/test_inertia.py
```diff
@@ -382,6 +382,91 @@ def test_compute_inertia_shape_dispatcher(self):
...
+                        geo_type,
+                        scale,
+                        None,
+                        1000.0,
+                        is_solid=False,
+                        thickness=0.0,
+                    )
+
+                self.assertEqual(mass, 0.0)
+                assert_np_equal(np.array(inertia), np.zeros(9), tol=0.0)
```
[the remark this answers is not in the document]
han-xudong: @coderabbitai This suggestion does not match the current Warp representation. With warp-lang 1.16.0.dev20260716, np.array(wp.mat33()).shape is (9,), not (3, 3). I also reproduced that np.zeros((3, 3)) fails the strict assert_array_equal check with a shape mismatch, while the current np.zeros(9) passes test_hollow_primitive_zero_thickness_returns_zero_inertia. No code change is needed here.

newton/_src/geometry/inertia.py
```diff
@@ -618,7 +664,9 @@ def compute_inertia_shape(
         if is_solid:
             return solid
         else:
-            assert isinstance(thickness, float), "thickness must be a float for a hollow cone geom"
+            thickness = _validate_hollow_thickness("cone", thickness, (("radius", scale[0]), ("half_height", scale[1])))
+            if thickness == 0.0:
+                return 0.0, solid[1], wp.mat33()
             hollow = compute_inertia_cone(density, scale[0] - thickness, scale[1] - thickness)
             m_shell = solid[0] - hollow[0]
             if m_shell <= 0.0:
```
adenzler-nvidia: Drive-by cleanup while you're here: this check is now unreachable — `_validate_hollow_thickness` already guarantees `thickness < radius` and `thickness < half_height`, so the inner cone is strictly smaller and `m_shell` is always positive. The `if m_shell <= 0.0: raise ValueError(...)` block can be dropped.

adenzler-nvidia (dismissed):
Thanks for the quick turnaround — the zero-thickness warning path and the hard errors for negative/non-finite/oversized values are exactly what we discussed, and the test coverage is thorough. LGTM.

Apologies for the changelog back-and-forth: entries in `CHANGELOG.md` conflict with almost every PR that lands, and yours got hit twice. We're working on a solution to reduce these conflicts. Could you do one more merge of upstream `main` to resolve the current one? After that this is good to go.

newton/_src/geometry/inertia.py
han-xudong: @adenzler-nvidia Removed the redundant non-positive shell-mass check in efa44370 and merged the latest upstream/main. The six focused hollow-inertia tests and the full pre-commit suite pass.

adenzler-nvidia (approved):

PR #3312, opened 2026-07-02
Fix MJCF import validation errors

### LINKED ISSUE #3315: [BUG] MJCF importer uses asserts for malformed user input
## Bug Description

The MJCF importer still uses assertions or low-context failures for malformed user-authored input. Mesh geoms can reject a non-three-value `size` even though MuJoCo ignores mesh geom `size`; nested free joints fail with `AssertionError`; and incomplete or malformed inertial data can fail with `AssertionError` or `IndexError` instead of deterministic validation errors.

Related PR: #3312

## Reproduction Script

```python
import newton

mjcf = """
<mujoco>
    <worldbody>
        <body name="parent">
            <body name="child">
                <freejoint name="bad_free"/>
            </body>
        </body>
    </worldbody>
</mujoco>
"""

builder.add_mjcf(mjcf)
```

Expected: malformed MJCF inputs produce deterministic, contextual validation errors, and mesh geom `size` length is ignored consistently.

Actual: malformed inputs can fail with `assert`/indexing errors or inconsistent mesh geom `size` handling.

## System Information

Not platform-specific.

han-xudong:
## Description

Replace user-facing MJCF importer asserts with deterministic behavior and contextual validation errors.

- Ignore mesh geom `size` length for mesh geoms, matching MuJoCo's mesh geom semantics.
- Raise contextual `ValueError`s for nested free joints instead of using `assert`.
- Raise contextual `ValueError`s for incomplete or malformed inertial `fullinertia` data.

Closes #3315.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_import_mjcf.TestImportMjcfBasic
uvx pre-commit run -a
```

## Bug fix

**Steps to reproduce:**

1. Import MJCF with a mesh geom carrying a non-three-value `size`, a nested `<freejoint>`, or an `<inertial>` element without valid `diaginertia`/`fullinertia` data.
2. Observe that mesh geom `size` can be rejected despite being ignored by MuJoCo, or malformed free-joint/inertial inputs fail with `AssertionError`/`IndexError` instead of deterministic validation errors.

**Minimal reproduction:**

```python
import newton

mjcf = """
<mujoco>
    <worldbody>
        <body name="parent">
            <body name="child">
                <freejoint name="bad_free"/>
            </body>
        </body>
    </worldbody>
</mujoco>
"""

builder.add_mjcf(mjcf)
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MJCF validation so malformed free-joints and inertial definitions fail deterministically with clearer messages.
  * Enforced correct handling of free-joints (free joint must be on the world body) and strengthened inertial checks for required value counts.
  * Mesh geom `size` length components are now consistently ignored during import.
* **Tests**
  * Added unit tests covering mesh `size` handling and the new free-joint/inertial validation error cases.
* **Documentation**
  * Updated the Unreleased “Fixed” changelog entry to reflect these corrections.


newton/_src/utils/import_mjcf.py
```diff
@@ -1976,8 +1979,17 @@ def parse_body(
                 I_m[2, 2] = diaginertia[2] * scale**2
```
adenzler-nvidia, line 1948: The `diaginertia` path has the same failure mode this PR fixes for `fullinertia`: with `diaginertia="1 2"`, `parse_vec` returns two values and this line fails with an opaque `IndexError`. A matching length check would make the validation consistent.

newton/tests/test_import_mjcf.py
```diff
@@ -230,6 +230,110 @@ def test_mjcf_maxhullvert_parsing(self):
...
+        <body name="body">
+            <inertial pos="0 0 0" mass="1.0" fullinertia="1 2 3"/>
+            <geom type="sphere" size="0.1"/>
+        </body>
+    </worldbody>
+</mujoco>
+"""
+        builder = newton.ModelBuilder()
+
+        with self.assertRaisesRegex(ValueError, "fullinertia.*6 values"):
```
adenzler-nvidia: The fewer-than-6 case already raised before this change (as `IndexError`), so this test would still pass if the new check were weakened to `< 6`. The actual behavior change is that more than 6 values now raises where the extras were previously silently ignored — could you add a 7-value case to pin that? This also matches native MuJoCo, which reads `fullinertia` with an exact count of 6.

adenzler-nvidia (commented):
Thanks for the fixes! I checked the new behavior against the native MuJoCo compiler and it matches on all three counts: `size` on mesh geoms is ignored natively (zero size params are validated for mesh and the size is overwritten from the mesh AABB), nested free joints fail with "free joint can only be used on top level", and `fullinertia` is read with an exact count of 6. Two comments inline.

han-xudong:
@adenzler-nvidia Thanks for checking this against native MuJoCo and for the detailed notes.

I pushed an update addressing both inline comments:
- added an exact length check for `diaginertia` so malformed values now raise a clear `ValueError` instead of an `IndexError`
- extended the `fullinertia` test to include a 7-value case, so the exact-count behavior is pinned on both sides

Verified locally with:
- `uv run --extra dev -m newton.tests -k test_import_mjcf.TestImportMjcfBasic`
- `uvx pre-commit run -a`

adenzler-nvidia:
Thanks for the update! The `diaginertia` check and the 7-value `fullinertia` test case address both comments — this looks good content-wise.

One more thing before this can go in: `main`'s changelog was restructured in the meantime (#3558 split the old `[Unreleased]` section into `[Unreleased]` and `[1.4.0]`). Since this branch predates that, merging as-is would silently place your entry inside the already-released `[1.4.0]` section. Could you do one more merge from `main` and move your changelog line into `[Unreleased]` → `### Fixed`?

han-xudong:
Hi @adenzler-nvidia, I merged the latest `upstream/main` and moved the MJCF validation entry to `[Unreleased]` -> `Fixed`. The targeted MJCF suite passes (27 tests), and `pre-commit` passes as well.

adenzler-nvidia (approved):
Thank you!

PR #3316, opened 2026-07-02
Add masked rigid resets to SolverVBD

### LINKED ISSUE #3256: Ensure environment resets work correctly with the VBD solver
VBD in Isaaclab is becoming an important usecase and there seems to be some gap in how to handle environment resets properly. This ticket is used to track completion.

Acceptance criteria:
* IsaacLab can reliably use environment resets with VBD

Related discussion here: https://github.com/isaac-sim/IsaacLab/pull/5730#issuecomment-4517060041

jumyungc:
## Description

Add masked rigid-body reset support to `SolverVBD`.

`SolverVBD.reset()` can now reset all rigid bodies or selected worlds while preserving caller-authored state when requested. The implementation:

- Copies selected `body_q` and/or `body_qd` from model defaults according to `StateFlags`.
- Supports `flags=0` to preserve body state while clearing solver-owned history.
- Immediately resets selected joint penalties, C0 terms, and AVBD duals.
- Rebaselines pose and enabled-cable history from the next step’s input pose, allowing state edits or forward kinematics after `reset()`.
- Invalidates selected rigid-contact warm-start history on the next fresh contact refresh.
- Supports global/unowned entities when `world_mask=None`; explicit masks select only local worlds.
- Leaves collision generation under caller control and does not modify the shared collision pipeline.
- Warns when particle reset flags are requested because particle resets remain unsupported.

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **New Features**
  * Added `SolverVBD.reset(state, world_mask=None, flags=...)` for selective per-world rigid-state restoration (including joint history and contact handling).
  * Implemented per-world pose rebaselining to keep “previous pose” history consistent after resets.
  * Added one-shot rigid contact-history reset for selected worlds, with deferred cable (Dahl) friction history reinitialization on the next step.
* **Bug Fixes**
  * Cold-starts contact warm-start history only for worlds marked as reset-pending.
  * BVH rebuild now skips particle-related work when no particle positions are present.
* **Documentation**
  * Clarified unsupported particle reset behavior and updated `body_q_prev` semantics.


mmichelis:
Thanks JC for this PR! I've tested it on the IsaacLab side, and this works for cable resets nicely, thanks!

AnkaChan:
Does this `reset()` break existing captured graph? Nevertheless, we should add a test to validate the previously-captured graph is still working.

AnkaChan:
This is more of the structural question rather than an issues specific to this PR.

Since we are predicting the next step, knowing the current step (`state_in`) is enough for friction evaluation. I wonder why why do we need to let `VBDSolver` internally implicitly maintain the `body_q_prev` variable? The only possibility I can think of is to address the kinematic body,  whose `body_q` is already modified before passing to step(). Can you confirm this is the reason for this choice? If so the particle side should probably do something similar...

jumyungc:
> Does this `reset()` break existing captured graph? Nevertheless, we should add a test to validate the previously-captured graph is still working.

Thank you, Anka for the great suggestion. The test_rigid_reset_replays_captured_step unit test now verifies that reset() works with a previously captured CUDA graph.

jumyungc:
> This is more of the structural question rather than an issues specific to this PR.
> 
> Since we are predicting the next step, knowing the current step (`state_in`) is enough for friction evaluation. I wonder why why do we need to let `VBDSolver` internally implicitly maintain the `body_q_prev` variable? The only possibility I can think of is to address the kinematic body, whose `body_q` is already modified before passing to step(). Can you confirm this is the reason for this choice? If so the particle side should probably do something similar...

Yes, they are for teleported bodies (maybe mostly kinematic bodies). As we discussed today, let's plan something separately regarding overall state, prev pose/q, etc

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -1646,6 +1657,143 @@ def step(
...
+        before stepping; contact invalidation instead waits for a fresh refresh.
+        With ``rigid_contact_history=True``, use ``contact_matching="latest"``. Reset
+        does not change ``set_rigid_history_update()``; leave rigid-history refresh
+        enabled for the next contact-bearing step. Reusing contacts
+        (``set_rigid_history_update(False)``) is unsupported only while contact
+        invalidation is still pending.
+
+        Args:
+            state: The simulation state to reset (modified in place).
+            world_mask: Boolean mask of shape ``(world_count,)`` selecting local
```
eric-heiden: 🔵 `world_mask` should have shape `(world_count + 1,)`, with the final entry selecting global entities whose world is `-1`. VBD supports resettable global bodies and joints, but the current `(world_count,)` contract makes them impossible to select with an explicit mask, and even makes an all-true mask behave differently from `None`. Since this PR introduces the VBD reset API, please include the global slot here; we can align the other solvers on the same mask format in a follow-up issue, even where that final slot is currently a no-op.

newton/_src/solvers/vbd/rigid_vbd_kernels.py
```diff
@@ -2198,6 +2359,12 @@ def init_body_body_contacts_avbd(
 
     k_floor = avg_ke if k_start < 0.0 else wp.min(k_start, avg_ke)
     slot = match_index[i]
+    # Drop the saved match for reset-selected worlds so they cold-start instead
+    # of warm-starting from pre-reset history.
+    if slot >= 0 and contact_history_reset_pending:
+        if contact_history_reset_pending[0] != 0:
+            if _contact_world_selected(s0, s1, shape_world, shape_body, body_world, contact_history_reset_mask):
+                slot = -1
```
eric-heiden, line 2375: 🟠 Setting only the local `slot = -1` does not invalidate the sticky match already replayed into this contact row. With `CollisionPipeline(contact_matching="sticky")`, `replay_matched()` runs before this kernel and replaces the point, offset, and normal with historical geometry. The reset row therefore cold-starts its lambda and penalty while still solving against pre-reset anchors and normals. Please either require `contact_matching="latest"` when resetting rigid contact history, or clear the match before sticky replay and preserve fresh narrow-phase geometry for reset-selected rows.

newton/_src/solvers/vbd/solver_vbd.py
jumyungc: Done.

newton/_src/solvers/vbd/rigid_vbd_kernels.py
jumyungc: Addressed using the first option. 
SolverVBD already documented contact_matching="latest" as the supported mode, and SolverVBD.reset() now explicitly states that it is required when rigid contact history is enabled.

eric-heiden (approved):
Thanks!

PR #3319, opened 2026-07-02
Accept labels for distance joints

adenzler-nvidia:
`ModelBuilder.add_usd` forwards a shared joint label to each joint convenience method. `add_joint_distance` was the only constructor that did not accept that argument, causing every USD distance-joint import to fail with `TypeError` before construction.

This adds the optional label parameter and forwards it to the generic joint builder, consistent with the other joint constructors. The regression test imports a minimal USD distance joint and verifies that its prim path is retained as the joint label.

## Summary by CodeRabbit

* **New Features**
  * Distance joints can now be added with an optional label, improving tracking of joint names.
* **Bug Fixes**
  * Importing USD distance joints better preserves labeled joint information, retaining the joint’s original USD path.
* **Tests**
  * Added a unit test that imports an in-memory USD stage with a labeled distance joint and verifies the stored joint label/path.

### DIFF
--- a/newton/_src/sim/builder.py
+++ b/newton/_src/sim/builder.py
@@ -4660,6 +4660,7 @@ class ModelBuilder:
         child_xform: Transform | None = None,
         min_distance: float = -1.0,
         max_distance: float = 1.0,
+        label: str | None = None,
         collision_filter_parent: bool | None = None,
         enabled: bool = True,
         custom_attributes: dict[str, Any] | None = None,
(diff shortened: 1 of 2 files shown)


newton/_src/sim/builder.py
```diff
@@ -4660,6 +4660,7 @@ def add_joint_distance(
         child_xform: Transform | None = None,
         min_distance: float = -1.0,
         max_distance: float = 1.0,
+        label: str | None = None,
```
camevor: The `deprecate_nonkeyword_arguments` decorator circumvents the keyword-only separator in line 4658, making all arguments, in particular `collision_filter_parent`, `enabled`, and `custom_attributes` reachable as positional. Inserting label before them shifts the mapping for those arguments.

The safe approach would be to move `label` after `custom_attributes`.

adenzler-nvidia: Confirmed — with label inserted there, a legacy positional call like add_joint_distance(p, c, xf_p, xf_c, -1.0, 2.0, False) silently mapped the collision_filter_parent value to label, where it then got swallowed by the auto-label fallback. Moved label after custom_attributes in d3f09eaab.

camevor (dismissed):
Thanks, ltgm! There's just a minor merge conflict

camevor (approved):

PR #3323, opened 2026-07-02
Dedupe USD single-DOF joint resolution

adenzler-nvidia:
## Description

`parse_joint`'s revolute/prismatic arm and `parse_merged_joints`' sibling loop independently implemented the same single-DOF resolution: limit gains with revolute degrees scaling, armature/friction/velocity-limit lookups, the drive block with `JointTargetMode.from_gains`, initial-state reads, and the revolute degree-to-radian conversions. A fix applied to one copy could silently miss the other.

Extract a shared `resolve_dof_params` helper returning a typed `_DofParams` dataclass in Newton units. The helper reports `has_drive` so `parse_joint` keeps omitting drive targets for undriven joints (falling back to builder defaults) while the merged path keeps forcing zeros — an intentionally preserved pre-existing divergence. `parse_joint`'s top-level armature/friction/velocity-limit reads move into the D6 arm, their only remaining consumer, so single-DOF joints resolve each attribute once. Net −60 lines.

Two minor notes for reviewers: those three attribute reads now happen after the joint's custom-attribute extraction (matching the order `parse_merged_joints` always had; the old order was inconsistent within `parse_joint` itself and only observable to a custom-attribute value transformer that mutates resolver-visible prim attributes as a side effect). The verbose prints for initial joint state now report converted units (rad / m/s) instead of raw USD values with mismatched labels.

## Checklist

## Test plan

```
```

Additionally verified behavioral equivalence by snapshotting the full `ModelBuilder` state (169 fields, exact equality) before and after the change for 10 USD test fixtures, with coverage confirmed for all touched paths: `humanoid_mjc.usda` (7 merged D6 joints with initial state), `ant.usda`/`ant_multi.usda` (drives and initial state through `parse_joint`, up to 512 joints), and `pendulum_revolute_vs_d6.usda` (native D6 arm) — all bit-identical.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved USD import for revolute and prismatic joints with more consistent per-DOF limit/drive behavior, including friction/armature, initial position/velocity, and optional velocity limits.
  * Fixed initial joint state writing to preserve correct units and reduce load discrepancies.
  * Updated merged-joint imports so each sibling DOF inherits matching limits, drives, and limit-mode settings, with proper velocity-limit defaults.


newton/_src/utils/import_usd.py
```diff
@@ -163,6 +163,30 @@ def _external_stacklevel() -> int:
         del frame
 
 
+@dataclass
+class _DofParams:
```
eric-heiden, line 167: Would it make sense to use `ModelBuilder.JointDofConfig` here instead of adding `_DofParams`? It already holds most of these DOF settings, and `add_joint_revolute()` and `add_joint_prismatic()` accept it as `axis`. The resolver could return that config together with the initial state and `limit_solref_mode`. Or is there a reason to keep a separate import-specific type here?

adenzler-nvidia: The resolver's contract is "what USD authored, in Newton units" — before each call site applies its own fallback policy. `JointDofConfig` can't express that: every field needs a concrete value (there's no "unauthored" state), and its `__init__` is active — it normalizes `axis` and clamps `target_pos` into the limits, decisions that shouldn't be made before the fallbacks are chosen.

The concrete blocker is the undriven-joint divergence flagged in the PR description: `parse_joint` omits the drive block so `add_joint_revolute()` falls back to `builder.default_joint_cfg`, while the merged path forces zero gains and an infinite effort limit. A returned `JointDofConfig` would have to bake one of those policies in, changing behavior for the other site; carrying `has_drive` alongside a config and bypassing its drive fields at one call site seemed worse than a flat record. The axis is also unknown at resolution time — the merged path remaps it into the representative joint's frame after resolving the DOF.

The plan is a follow-up that unifies the undriven case to zeros everywhere (the builder-default fallback can silently actuate passive joints under solvers that don't gate on actuator mode). Once that behavior change lands separately, the divergence disappears and the resolver can return a `JointDofConfig` plus initial state / solref mode — happy to do that migration there. This PR stays bit-identical on purpose.

newton/_src/utils/import_usd.py
```diff
@@ -1105,6 +1127,95 @@ def resolve_joint_parent_child(
...
+            if has_drive:
+                target_pos *= DegreesToRadian
+                target_vel *= DegreesToRadian
+                target_ke /= DegreesToRadian / joint_drive_gains_scaling
+                target_kd /= DegreesToRadian / joint_drive_gains_scaling
+            if velocity_limit is not None:
+                velocity_limit *= DegreesToRadian
+            if initial_position is not None:
+                initial_position *= DegreesToRadian
+
```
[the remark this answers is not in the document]
adenzler-nvidia: Intentionally preserved pre-existing behavior. The authored joint-state contract today is positions in degrees, velocities already in rad/s — `test_humanoid_d6_joint_state_initialization` in `test_schema_resolver.py` asserts velocities pass through unconverted, and the pre-refactor code carried an explicit "velocity is already in rad/s" comment. Converting here would change behavior and break that test; this PR is a behavior-preserving refactor (verified bit-identical on builder snapshots). Whether the contract itself is right — angular drive `targetVelocity` is deg/s in UsdPhysics and does get converted — is a fair consistency question, but it belongs in a separate change with its own test/asset updates.

newton/_src/utils/import_usd.py
eric-heiden: Thanks for clarifying, let's keep this separate dataclass here for now and follow up with adjusting the builder defaults to passive actuation settings.

eric-heiden (approved):
Thanks!

PR #3324, opened 2026-07-02
Support vector gravity in ModelBuilder

adenzler-nvidia:
Allow `ModelBuilder.gravity` to accept a three-component vector while preserving scalar gravity as a deprecated compatibility form along `up_axis`. Migrate Newton's tests, examples, and documentation to the vector interface.

Resolve USD `PhysicsScene.gravityDirection` and `gravityMagnitude` into a normalized builder-space gravity vector. Stage-axis conversion and whole-scene transforms rotate gravity, while root-placement overrides do not; imports inside an active world update that world's gravity directly.

Add focused coverage for scalar deprecation, vector/default behavior, builder composition, arbitrary USD directions, up-axis conversion, scene transforms, root-placement transforms, and per-world import. The changelog records both the fix and deprecation.

## Summary by CodeRabbit

* **New Features**
  * Gravity can now be provided as a 3D vector in `ModelBuilder`, making gravity setup clearer and consistent.
  * USD imports now apply authored gravity direction/magnitude with correct stage/up-axis orientation.

* **Bug Fixes**
  * Fixed USD import so joint collision behavior (for joints with two explicit bodies) is honored.
  * Fixed `ModelBuilder.add_usd()` gravity so `PhysicsScene.gravityDirection` is respected during USD import.

* **Documentation**
  * Updated collision and coordinate-system examples, and documented gravity-vector usage plus scalar-gravity deprecations.

* **Chores**
  * Updated examples and tests to use gravity vectors instead of scalars.


eric-heiden:
Thanks! This is a useful API improvement for supporting vector gravity in `ModelBuilder`.

The main issues are around default scalar warning behavior, mutable gravity object ownership, and missing public API units.

newton/_src/sim/builder.py
```diff
@@ -1236,9 +1247,10 @@ def __init__(self, up_axis: AxisType = Axis.Z, gravity: float = -9.81):
         """Internal world context backing the read-only :attr:`current_world` property."""
 
         self.up_axis: Axis = Axis.from_any(up_axis)
-        """Up axis used when expanding scalar gravity into per-world gravity vectors."""
-        self.gravity: float = gravity
-        """Gravity acceleration [m/s^2] applied along :attr:`up_axis` for newly added worlds."""
+        """Up axis used by geometry helpers and the deprecated scalar gravity API."""
+        self._gravity: float | wp.vec3 = -9.81
```
eric-heiden: 🔴 `ModelBuilder()` now seeds `_gravity` with the scalar `-9.81`, so reading `builder.gravity` takes the scalar compatibility path and emits a `DeprecationWarning` even when the caller did not pass scalar gravity. That makes normal default introspection noisy, and it can fail downstream CI that treats deprecations as errors.

Please track the omitted-default case separately from explicit scalar input. The default can still follow mutable `up_axis`, but the warning should only happen when the constructor or setter actually receives scalar gravity.

newton/_src/sim/builder.py
```diff
@@ -2026,6 +2038,32 @@ def up_vector(self, _):
...
+            self._gravity = float(value)
+            return
+        gravity = np.asarray(value, dtype=np.float32)
+        if gravity.shape != (3,):
+            raise ValueError(f"Expected gravity with shape (3,), got {gravity.shape}")
+        self._gravity = wp.vec3(*gravity)
+
+    def _gravity_as_vector(self) -> wp.vec3:
+        if np.isscalar(self._gravity):
+            return wp.vec3(*(component * self._gravity for component in self.up_vector))
```
eric-heiden: 🔴 `_gravity_as_vector()` returns the stored `wp.vec3` directly on the vector path, and callers store that object in `world_gravity`. Since `wp.vec3` is mutable, multiple worlds can share the same gravity object, and `add_builder()` can even leave a destination world aliased to the source builder's `_gravity`. Mutating one `world_gravity` entry can silently change other worlds or the source builder default.

Please make this helper always return a fresh vector, for example by copying the vector branch with `wp.vec3(*self._gravity)`, and add a regression test that checks two vector-gravity worlds do not share object identity.

newton/_src/sim/builder.py
```diff
@@ -865,15 +871,20 @@ def key(self) -> str:
...
+    ):
         """
         Initializes a new ModelBuilder instance for constructing simulation models.
 
         Args:
             up_axis: The axis to use as the "up" direction in the simulation.
                 Defaults to Axis.Z.
-            gravity: The magnitude of gravity to apply along the up axis.
-                Defaults to -9.81.
+            gravity: Default gravity vector. The deprecated scalar form applies
```
eric-heiden: 🟡 The new public `gravity` docs describe a physical acceleration but omit units. Please include `[m/s^2]` here, and apply the same unit wording to the `gravity` property and `begin_world()` gravity argument so the public API docs stay consistent.

newton/_src/sim/builder.py
adenzler-nvidia: Fixed in 9b0af6032. The untouched default is now tracked internally as `None`: reading `gravity` on a default builder returns the up-axis-resolved vector without warning, and the default still follows later `up_axis` changes. The warning only fires when a scalar is explicitly passed to the constructor or the setter.

newton/_src/sim/builder.py
adenzler-nvidia: Fixed in 9b0af6032. `_gravity_as_vector()` now returns a fresh `wp.vec3` on every path, and `begin_world()` copies its gravity argument as well. Added `test_world_gravity_entries_not_aliased` covering both world-to-world sharing and `add_builder()` aliasing the source builder.

newton/_src/sim/builder.py
adenzler-nvidia: Added `[m/s^2]` in 9b0af6032 to the constructor argument, the `gravity` property, `begin_world()`'s gravity argument, and the `world_gravity` attribute.

newton/_src/sim/builder.py
```diff
@@ -2194,6 +2206,32 @@ def up_vector(self, _):
             "The 'up_vector' property is read-only and cannot be set. Instead, use 'up_axis' to set the up axis."
         )
 
+    @property
+    def gravity(self) -> float | wp.vec3:
+        """Default gravity vector, or a deprecated scalar along :attr:`up_axis`."""
+        if np.isscalar(self._gravity):
+            warnings.warn(_SCALAR_GRAVITY_DEPRECATION_MSG, DeprecationWarning, stacklevel=2)
+        return self._gravity
```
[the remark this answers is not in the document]
adenzler-nvidia: Fixed in 9b0af6032 — the default is tracked as `None` internally and only explicit scalar assignment triggers the deprecation warning.

adenzler-nvidia:
Regarding the unauthored `gravityMagnitude` nitpick: `UsdPhysics.LoadUsdPhysicsFromRange` resolves the schema sentinels itself — with nothing authored it returns direction `-up_axis` and magnitude `9.81 / metersPerUnit`, so no `-inf` reaches the importer and there is no production issue. Added a sub-case to `test_scene_gravity_direction` in 9b0af6032 pinning this behavior (unauthored attributes, `metersPerUnit=1.0` → gravity `(0, 0, -9.81)`).

eric-heiden:
Seems like there are some CI failures to fix, otherwise LGTM!

eric-heiden (dismissed):
Thanks! I can reapprove once the merge conflicts are resolved.

eric-heiden (approved):

Issue #3038, opened 2026-06-02
[REQ] Per-Body Indexing for Volume Deformables with Material Loaded from USD


mmichelis:
## Problem

Newton's USD importer parses `UsdGeom.TetMesh` prims into simulated FEM soft bodies. However, to drive large-scale parallel (multi-environment) deformable simulation from USD, two capabilities are missing today:

1. **Per-soft-body addressability** — imported bodies cannot be located in the finalized model, so  their state cannot be read or written individually.
2. **Material-parameter parity** — only a subset of the parameters `add_soft_mesh` accepts is read   from USD, so a body's simulated behavior is not fully specified by its USD definition.

## What is currently possible

`parse_usd` (via `ModelBuilder.add_usd`) auto-discovers every `UsdGeom.TetMesh` prim — detection is purely by prim type — and imports it through `add_soft_mesh` (`utils/import_usd.py:2874-2916`):

| Capability | Detail |
|---|---|
| Discovery and import | Every `TetMesh` is imported via `add_soft_mesh`; `ignore_paths` is honored. |
| Geometry and transform | `points`, `tetVertexIndices`, left-handed winding correction, and world-transform application. |
| Material (core only) | Bound `material:binding:physics` → `youngsModulus` / `poissonsRatio` / `density`, converted to `k_mu` / `k_lambda`. |
| Multi-world replication | Replicating a prototype across worlds with `add_builder` inside `begin_world` / `end_world` yields `world_count == num_envs`. |
| Multi-copy stage read | A single `add_usd` over a stage containing N referenced `TetMesh` prims imports all N as distinct soft bodies at the correct positions, including the `instanceable=True` (instance-proxy) case. |

### 1. Locating an individual soft body

After import there is no per-body handle — no particle/tetrahedron range, no prim-path lookup, no way to separate bodies that share a world. The reasons:

- Particles are segmented only per world, via `particle_world` and `particle_world_start` (CSR,   shape `[world_count + 2]`) (`sim/model.py:170,172-188`). For three worlds with two bodies each,  `particle_world_start` is `[0, 12, 24, 36, 36]`: the two bodies in a world are merged into one  contiguous block with no recorded internal boundary.
- The `add_usd` / `parse_usd` return value (`utils/import_usd.py:3525-3544`) provides `path_body_map` and  `path_shape_map`, but no equivalent map for soft bodies.
[...]

Issue #3037, opened 2026-06-02
[REQ] Cable/Rod Parsing through USD in Newton Builder


mmichelis:
## 1. Missing feature

`newton.ModelBuilder.add_usd()` does not import cables. `add_rod_graph()` / `add_rod()` exist on the builder (they create a chain/graph of capsule rigid bodies connected by `JointType.CABLE` joints), but `parse_usd()` never calls them, and **no USD prim type is recognized as a cable**. `parse_usd()` only handles rigid bodies, the standard `UsdPhysics` joints, articulations, and the basic shape types — `UsdGeom.BasisCurves` (and curves generally) are ignored entirely.

As a result, a USD stage containing a curve marked as a simulated cable loads into Newton with **no rod bodies, joints, or shapes** — the curve is silently dropped.

A cable should be recognized by a **cable material applied to the curve prim** (a bound physics material carrying rod parameters). This differentiates a simulated cable from a plain visual/collision curve: the same `BasisCurves` is non-physical without the cable material and a simulated rod with it.

We want `add_usd()` to recognize a cable-tagged curve, import it via `add_rod_graph()` (capsule segments + cable joints), read the rod parameters from USD, and work under multi-world replication.

---

### 2.1 What loading should produce

After `builder.add_usd(stage)` + `builder.finalize()` on a stage containing a cable prim, the resulting `Model` should contain that cable's **capsule segment bodies and cable joints** (one segment body per curve edge or user-specified segment count), with parameters read from USD (radius, stretch/bend stiffness and damping, density). Each replicated instance gets its own independent segments, and the caller must be able to map a USD prim to its segment bodies/joints in the resulting `Model` (§2.2). Today no such mapping exists for rods.

Unlike cloth (particles), a cable's state is **rigid-body state**: each segment is a capsule body, so the per-cable state lives in `body_q` (transforms) and `body_qd` (spatial velocities), indexed by the cable's body range.

### 2.2 What we need: a stable prim to segment mapping

The core requirement is consumer-facing: given a cable's USD prim path, we must be able to find **which rows of `body_q` / `body_qd` (and which joints) are that cable's segments** — stably across replicated environments — so we can present its state as `(num_envs, num_segments, ...)` and reset individual instances.

[...]

Issue #3036, opened 2026-06-02
[REQ] Surface Deformable Parsing through USD in Newton Builder


mmichelis:
## 1. Missing feature

`newton.ModelBuilder.add_usd()` does not import cloth. It detects volumetric deformables via `prim.IsA(UsdGeom.TetMesh)` and calls `add_soft_mesh()`, but there is **no path that turns a triangulated `UsdGeom.Mesh` into a cloth** via `ModelBuilder.add_cloth_mesh()`. `add_cloth_mesh()` exists on the builder but `parse_usd()` never calls it, and no USD prim/schema is recognized as a cloth.

As a result, a USD stage that contains a triangle mesh marked as a simulated surface deformable loads into Newton with **no cloth particles, triangles, or bending edges** — the surface is silently ignored (or, if it also carries a collider, imported only as static collision geometry).

We want `add_usd()` to recognize a cloth `UsdGeom.Mesh` and import it as Newton cloth (particles + FEM triangles + bending edges), reading the cloth parameters from USD, and to work under multi-world replication.

A cloth should be recognized by a **cloth material applied to the `UsdGeom.Mesh`** (a bound physics material carrying cloth parameters). This is what differentiates a simulated cloth from a plain collision mesh: the same triangle mesh is a static collider without the cloth material and a simulated cloth with it.

---

### 2.1 What loading should produce

After `builder.add_usd(stage)` + `builder.finalize()` on a stage containing a cloth prim, the resulting `Model` should contain that cloth's particles, triangles, and bending edges, with parameters read from USD (stretch `tri_ke/tri_ka`, damping `tri_kd`, bending `edge_ke/edge_kd`, surface `density`, `particle_radius`). Each replicated instance gets its own independent, contiguously-indexed particles, and `add_usd()` records a per-cloth grouping on the `Model` (§2.2) — the single mechanism by which the caller maps a USD prim to its rows in `particle_q`. Today no such grouping exists for deformables.

### 2.2 Per-cloth indexing on the Model (the indexing mechanism)

"Which rows of `particle_q` belong to which cloth" lives on the `Model` as CSR start arrays, mirroring `articulation_start` / `particle_world_start`:

```python
model.cloth_count               # int
model.cloth_label               # list[str], len == cloth_count  (USD prim path per instance)
model.cloth_particle_start      # wp.array[int32], shape [cloth_count + 1]   (CSR into particle_q)
[...]

PR #3326, opened 2026-07-02
Add family-specific views for deformable selection

mzamoramora-nvidia:
## Description

Add three experimental views in `newton.selection` to find deformables by label and read or update their simulation state:

| View | Supported state in this PR |
| --- | --- |
| `DeformableCurveView` | Rod segment transforms and velocities |
| `DeformableSurfaceView` | Surface particle positions and velocities |
| `DeformableVolumeView` | Tetrahedral-volume particle positions and velocities |

The views work with native builder calls and USD imports. They share selection, range lookup, and indexed-write code internally. Each class selects its own family, so a broad surface pattern does not also select curves or volumes.

### Why separate views?

Curve, surface, and volume describe geometry, not how it is simulated. A curve could use rigid segments or particles connected by springs. The three classes leave room for those choices without tying the public names to a solver.

This PR supports Newton's existing rods, rod graphs, cloth meshes, soft meshes, and soft grids. It does not add particle-based curves or expose solver-owned MPM data. A representation filter is deferred until another implementation needs it. Available element ranges come from the selected records.

### What is a group?

A group is one deformable object, such as a cable, cloth, or soft volume. It records that object's label, world, and simulation ranges.

For example, suppose each world contains a cloth, while World 1 contains two cables:

```text
World 0: cloth, cable
World 1: cloth, cable_0, cable_1
World 2: cloth, cable
```

A surface view selecting `cloth` has three rows. A curve view selecting `cable*` has four:

```text
group index:     0        1        2        3
world:           0        1        1        2
label:         cable   cable_0  cable_1   cable
world_starts: [0, 1, 3, 4]
```

`world_starts` partitions the flat group axis by world. `world_ids` gives each row's world. Empty worlds are preserved. `ranges(kind)` returns each group's `[start, end)` simulation range; `starts(kind)` provides device-side start indices.

### Selection and state access

- Match labels with glob strings, lists of globs, or compiled regular expressions, using the same matching as `ArticulationView`.
- Read rectangular arrays when selected groups have equal counts for the requested element kind. Groups with different counts remain selectable through their raw ranges.
- Use `group_indices` to choose destination rows. Use optional `source_indices` to choose rows from the input values. Without source indices, values have one compact row per destination.
- Return zero-copy views for regular getter layouts. Irregular layouts reuse a buffer. Copy results that must remain unchanged, and use setters to write state consistently across layouts.
- Capture indexed device writes and internally staged reads after view construction and warm-up. Allocate input buffers and device selectors before capture.

Host indices must be genuine integers and within bounds. Destination duplicates are rejected; source rows may repeat. Device selectors are one-dimensional `int32` arrays on the model device. Invalid indices skip the corresponding write. Duplicate destinations use the last input row. If that row has an invalid source index, the destination stays unchanged.

A setter changes only the supplied `Model` or `State`. Applications with two alternating states must reset both. Group indices match world IDs only when exactly one group is selected in every world.

### Builder and finalization behavior

Native builder calls and USD imports record the same group information. Explicit labels support application lookup. Omitted native labels receive names such as `curve_0`, `surface_0`, and `volume_0`. USD objects use their prim paths. Groups keep their ranges through builder composition and replication.

The builder exposes experimental `curve_label` / `curve_world`, `surface_label` / `surface_world`, and `volume_label` / `volume_world` lists. Cloning tools can update object identities before finalization. Keep these lists aligned with the recorded groups. Simulation range lists stay private; the views expose the finalized ranges.

Labels do not change fixed-joint collapse. Complete curve groups follow remapped indices. A group is omitted with a warning if collapse removes one of its bodies or joints. Use `joints_to_keep` when the complete curve must remain selectable.

This PR adds no visual meshes, visual-mesh parsing, runtime attachments, or new solver behavior.

### Related issues

This addresses the finalized-model access requested in:

- [#3036](https://github.com/newton-physics/newton/issues/3036): surface particle state and topology ranges.
- [#3037](https://github.com/newton-physics/newton/issues/3037): curve segment-body state and cable ranges.
- [#3038](https://github.com/newton-physics/newton/issues/3038): soft-volume particle state and tetrahedron ranges.

For deformables, the flat group axis also supports the uneven-world use case raised for articulations in [#3462](https://github.com/newton-physics/newton/issues/3462). Visual-shape counts do not determine whether deformable state can be batched, which is related to the concern in [#3461](https://github.com/newton-physics/newton/issues/3461). This PR does not implement those requested articulation changes or close these broader issues.

## Checklist

- [x] A Towncrier fragment describes the user-facing changes

## Test plan

```bash
uv run --extra dev -m newton.tests --no-cache-clear -j 1 -k selection_deformable -k import_usd_deformable -k test_api -k test_builder_replicate
uv run --extra dev --extra docs --extra sim -m newton.tests --no-cache-clear -j 1 -k test_selection
```

Both suites passed locally, with an expected CPU VBD skip in the USD suite. Coverage includes family filtering with shared labels, native and USD replication, strict index checks, independent source rows, mixed rigid/deformable scenes, getter layouts, and CUDA capture/replay. The deformable and articulation examples ran on CPU and CUDA. The API documentation also built without warnings.

## New feature / API change

For a model with at least three worlds and one matching cable, cloth, and soft toy in every world:

```python
import warp as wp

from newton.selection import (
    DeformableCurveView,
    DeformableSurfaceView,
    DeformableVolumeView,
)

cables = DeformableCurveView(model, "hanging_cable")
cloths = DeformableSurfaceView(model, "surface_primary")
toys = DeformableVolumeView(model, "soft_cube")

transforms = cables.get_body_transforms(state)
positions = cloths.get_particle_positions(state)
toy_ranges = toys.ranges("particle")

# Save full reset buffers during setup, independently of getter storage.
positions_reset = wp.clone(cloths.get_particle_positions(model))
velocities_reset = wp.clone(cloths.get_particle_velocities(model))

# Reset worlds 0 and 2. Here, one selected cloth per world makes
# destination group IDs equal to world IDs.
destination_groups = wp.array([0, 2], dtype=wp.int32, device=model.device)
source_rows = wp.array([0, 2], dtype=wp.int32, device=model.device)
cloths.set_particle_positions(
    state,
    positions_reset,
    group_indices=destination_groups,  # Where to write.
    source_indices=source_rows,  # Which reset-buffer rows to read.
)
cloths.set_particle_velocities(
    state,
    velocities_reset,
    group_indices=destination_groups,
    source_indices=source_rows,
)

# A broader selection can contain several cloths in each world.
surfaces = DeformableSurfaceView(model, "surface_*")
world_starts = surfaces.world_starts
world_ids = surfaces.world_ids
```

The source and destination selectors are independent; they use equal values here because the saved buffer has one row for each original group.

Run the standalone example:

```bash
uv run --extra examples -m newton.examples selection_deformables --viewer gl
```

It shows cables, cloths, and soft cubes across replicated worlds. It resets selected positions and velocities in both simulation states, logs contacts, and enables the viewer's picking controls. It uses native builders and does not require USD.

## Proposed split

To make review easier, we propose splitting this work into three PRs:

1. **Record deformable groups.** Builder recording, finalized model records, composition, replication, and fixed-joint collapse handling.
2. **Add the three views.** Label matching, ranges, state access, indexed writes, and API documentation.
3. **Add the selection example.** The runnable cable/cloth/volume example, its checks, README entry, and thumbnail.

Each PR includes its own tests and relevant documentation. PR 2 builds on PR 1; PR 3 builds on PR 2. The example could stay with the views if that helps reviewers. These parts are currently combined in this PR.

## Follow-ups

These stay outside this PR to keep its scope manageable:

- Add `get_attribute()` and `set_attribute()` for broader data access, while retaining common state methods.
- Expose device-side range ends or counts for groups with different sizes.
- Consider model/state ownership checks consistently with `ArticulationView`.
- Add integration guidance for renderer or Fabric synchronization, Torch-backed selectors, and two-state resets.
- Support additional simulation representations. Add a representation selector only when another implementation needs it.
- Consider a helper for selecting all matching groups in given worlds. The current API exposes this mapping through `world_starts`.


mzamoramora-nvidia:
@coderabbitai review

mzamoramora-nvidia:
@coderabbitai review once

newton/_src/utils/import_usd_deformable_attachments.py
```diff
@@ -0,0 +1,444 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
+
+"""USD attachment / element-collision-filter import passes.
+
+Lowers ``PhysicsAttachment`` prims onto imported cables (as ball joints to xform targets) and
+``PhysicsElementCollisionFilter`` prims to shape collision-filter pairs, plus the post-collapse
+index remap. Driven by :func:`.import_usd.parse_usd` via a
+:class:`.import_usd_deformable_utils._DeformableImportContext`.
```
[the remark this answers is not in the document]
mzamoramora-nvidia: Resolving as stale for this PR. This docstring is identical to the current base branch; PR 3326 no longer introduces or changes these private Sphinx references. Any importer docstring cleanup should be handled separately against the main branch.

newton/_src/utils/import_usd_deformable_cable.py
```diff
@@ -0,0 +1,677 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
+
+"""USD cable / curve-deformable import passes.
+
+Imports linear ``UsdGeom.BasisCurves`` deformables as rods (chains of capsule bodies joined
+by cable joints, usable by any solver that supports them), welding curve-to-curve
+``PhysicsAttachment`` junctions into shared rod graphs first, then importing remaining single
+curves. Driven by :func:`.import_usd.parse_usd` via a
+:class:`.import_usd_deformable_utils._DeformableImportContext`.
```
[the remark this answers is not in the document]
mzamoramora-nvidia: Resolving as stale for this PR. The parse_usd docstring reference is already present unchanged in the current base, while PR 3326 only changes cable group bookkeeping nearby. This can be cleaned up separately against the main branch.

newton/_src/utils/import_usd_deformable_cable.py
```diff
@@ -0,0 +1,677 @@
...
+            consumed_curves.add(key)
+        if verbose:
+            print(f"Added cable graph {cid} with {len(body_ids)} segments across {len(comp_paths)} curves.")
+
+    for cid, comp_curves in components.items():
+        comp_paths = sorted(comp_curves)
+        comp_welds = welds_by_comp.get(cid, [])
+        if len(comp_paths) == 1 and not comp_welds:
+            continue  # plain single curve: leave it to the per-curve pass
+        _build_graph_component(cid, comp_paths, comp_welds)
```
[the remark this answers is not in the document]
mzamoramora-nvidia: Resolving as stale for this PR. The sorted comp_curves representative-selection behavior is identical to the current base and is not changed by PR 3326. The traversal-order question may still merit a dedicated importer follow-up, but it is outside this PR diff.

newton/_src/utils/import_usd_deformable_cloth.py
```diff
@@ -0,0 +1,204 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
+
+"""USD surface-deformable (cloth) import pass.
+
+Imports ``PhysicsSurfaceDeformableSimAPI`` polygon ``UsdGeom.Mesh`` prims as cloth, mapping the
+surface material onto the isotropic membrane. Driven by :func:`.import_usd.parse_usd` via a
+:class:`.import_usd_deformable_utils._DeformableImportContext`.
```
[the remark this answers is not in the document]
mzamoramora-nvidia: Resolving as stale for this PR. This private Sphinx reference is unchanged from the current base; PR 3326 only removes duplicate cloth-group recording in this importer. The docstring cleanup belongs in a separate change.

newton/_src/utils/import_usd_deformable_utils.py
```diff
@@ -0,0 +1,567 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
+
+"""USD deformable importer shared leaf helpers and import context.
+
+This module owns the builder-independent leaf helpers (e.g. :func:`_validate_mass_array`) and the
+shared mass / density / anchor utilities used by the cable / cloth / volume / attachment /
+collision-filter import passes, plus the :class:`_DeformableImportContext` that carries the
+:func:`parse_usd` inputs, helper closures, and result maps the passes mutate. The passes
```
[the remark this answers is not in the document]
mzamoramora-nvidia: Resolving as stale for this PR. The import_usd_deformable_utils file is no longer changed by PR 3326, and these references are identical to the current base. They can be addressed as separate importer documentation cleanup.

newton/_src/utils/import_usd_deformable_volume.py
```diff
@@ -0,0 +1,166 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
+
+"""USD volume-deformable import pass.
+
+Imports ``UsdGeom.TetMesh`` prims as soft bodies via :meth:`ModelBuilder.add_soft_mesh`. Driven by
+:func:`.import_usd.parse_usd` via a
+:class:`.import_usd_deformable_utils._DeformableImportContext`.
```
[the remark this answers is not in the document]
mzamoramora-nvidia: Resolving as stale for this PR. This private parse_usd reference is unchanged from the current base; PR 3326 only removes duplicate volume-group recording nearby. The documentation concern should be handled separately.

newton/_src/utils/import_usd_deformable_volume.py
```diff
@@ -0,0 +1,166 @@
...
+        body_root = usd._find_deformable_body_prim(prim)
+        if body_root is not None:
+            body_root_path = str(body_root.GetPath())
+            if not is_volume_deformable:
+                warnings.warn(
+                    f"{path}: non-simulation TetMesh under deformable body {body_root_path}; "
+                    f"treated as graphics/collision geometry (not simulated).",
+                    stacklevel=2,
+                )
+                continue
```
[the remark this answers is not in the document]
mzamoramora-nvidia: Resolving as stale for this PR. The PhysicsDeformableBodyAPI recognition behavior is identical to the current base and is not introduced or modified by PR 3326. Its proposal semantics can be reviewed in a focused importer follow-up.

newton/_src/utils/selection.py
```diff
@@ -2023,3 +2025,717 @@ def set_actuator_parameter(
...
+        verbose: If True, print a short selection summary.
+    """
+
+    _FAMILY_KINDS: ClassVar[dict[str, tuple[str, ...]]] = {
+        "curve": ("body", "joint"),
+        "surface": ("particle", "triangle", "edge"),
+        "volume": ("particle", "tetrahedron"),
+    }
+
+    @deprecate_nonkeyword_arguments
```
camevor: Suggest dropping the decorator here: on a new API it loosens the contract rather than protecting compatibility.

newton/_src/utils/selection.py
```diff
@@ -2023,3 +2025,717 @@ def set_actuator_parameter(
...
+
+
+class DeformableView:
+    """Select finalized deformable groups and read or update their existing state.
+
+    .. experimental::
+
+       This API may change while deformable selection is developed.
+
+    A string ``pattern`` uses label glob matching. A compiled :class:`re.Pattern`
```
camevor: Compiled-pattern support is a good suggestion for adding to `match_labels` in the context of #3272.

We should aim to use a single matching interface though. Recommend using `match_labels` and creating a separate PR for supporting compiled regular expressions in there.

newton/_src/utils/selection.py
```diff
@@ -2023,3 +2025,717 @@ def set_actuator_parameter(
...
+    uses ``fullmatch()``. Results keep model order: world order followed by builder
+    order. The public families are ``"curve"``, ``"surface"``, and ``"volume"``.
+
+    Labeled native deformables and deformables loaded by
+    :meth:`~newton.ModelBuilder.add_usd` use the same finalized group records. Native
```
camevor, line 2205: The current implementation makes deformables without a label unselectable: when label is None, the group is never recorded. If we introduce a group concept, it should not be gated by a user-provided label. Ideally, default labels should also be generated.

newton/_src/utils/selection.py
```diff
@@ -2023,3 +2025,717 @@ def set_actuator_parameter(
...
+    .. code-block:: python
+
+        import warp as wp
+
+        import newton
+
+        surface = newton.selection.DeformableView(
+            model,
+            "/World/Cloth",
+            family="surface",
```
camevor, line 2248: Does family need to be a constructor argument? The model already knows each group's family, so it could be inferred from the match (with appropriate error reporting), or it could be an optional argument for filtering.

docs/concepts/usd_parsing.rst
```diff
@@ -180,7 +181,23 @@ Every imported deformable can be looked up by its prim path in the mapping
...
 ``path_soft_map`` hold each cloth's and soft body's ``[start, end)`` particle and topology
-ranges. Without the flag the return shape carries no deformable entries.
+ranges. Without the flag the return shape carries no deformable entries. On the finalized
+:class:`~newton.Model`, select groups by label pattern with
+:class:`~newton.selection.DeformableView` (following
+:class:`~newton.selection.ArticulationView`): it batches each group's state on one flat group
+axis as ``(group_count, elements_per_group)`` arrays. ``world_starts`` partitions that axis by
+model world, including empty worlds, and ``world_ids`` identifies the world of every group.
+Setters use ``group_indices`` for flat rows or ``world_indices`` when every world has exactly
+one match. The view also exposes raw per-group ranges
```
camevor: It looks like `world_indices` is effectively a conditional alias of `group_indices`. I don't see the added benefit for this narrow use case, so I recommend removing it.
 
Alternatively, it could be generalized to work when the number of groups per world is equal across all worlds.

newton/_src/utils/selection.py
```diff
@@ -2023,3 +2025,717 @@ def set_actuator_parameter(
...
+    ranges. Global groups (world ``-1``) cannot be mixed with per-world groups.
+
+    Getters accept a :class:`~newton.Model` or :class:`~newton.State`. Writing a model
+    changes its initial arrays; writing a state changes only that state. Host
+    ``group_indices`` select flat group rows. ``world_indices`` select actual model
+    worlds when every world has exactly one matching group. Host group indices are
+    bounds-checked and duplicates are rejected. Device ``int32`` indices stay on the
+    device: out-of-range entries are ignored and the last value wins for duplicates.
+    Indexed setters using preallocated device values and indices can be captured after
+    view construction and kernel warm-up. View construction, getters, and allocations
```
camevor: Is this a constraint mandated by `DeformableView`? Enabling graph-captured particle access would be desirable.

newton/_src/utils/selection.py
```diff
@@ -2023,3 +2025,717 @@ def set_actuator_parameter(
...
+            source: Model initial state or simulation state to read.
+
+        Returns:
+            Particle positions with shape ``(count, particles_per_group)``.
+
+        Raises:
+            AttributeError: If this is not a surface or volume view.
+            ValueError: If the selected groups have different particle counts.
+        """
+        return self._gather("particle", source.particle_q, _gather_group_vec3_kernel, wp.vec3)
```
camevor: Allocating large arrays on demand and on each call (potentially within a graph) will leave performance on the table and can lead to hard-to-debug failures. Could we find a way to pre-allocate the arrays that most likely will be needed?

For homogeneous worlds, we could also use views for zero-copy access.

camevor (commented):
Thanks @mzamoramora-nvidia! I like the core mechanics here: batched particle-state access, the flat group axis for heterogeneous worlds, and private group records.

An open direction question: Is curve/surface/volume the right abstraction for the selection API? It's the USD proposal's authoring taxonomy, and family determines the accessor set only because each family currently has one discretization. However, Newton already simulates particle-backed 1D (`add_spring`) deformables and tet-free volumes (`SolverImplicitMPM`), which don't fit the fixed family -> element-kind mapping.

mzamoramora-nvidia:
@camevor thanks for the great feedback! The question about the open direction is quite valid and I hope to get some clarity on that during an architecture review.

newton/_src/utils/selection.py
mzamoramora-nvidia: Agreed. I removed `@deprecate_nonkeyword_arguments` from the new `DeformableView` constructor, so the constructor now enforces its keyword-only options directly. `family` and `verbose` remain keyword-only. This is included in `549af547` and covered by `test_constructor_rejects_positional_family`.

newton/_src/utils/selection.py
mzamoramora-nvidia: Agreed. `DeformableView` now uses the shared `find_matching_ids()` / `match_labels()` path and accepts the same string or list-of-string glob patterns as the other selection APIs. I removed the local compiled-regex implementation so regex support can be considered consistently in #3272. This is covered by `test_list_patterns_use_shared_label_matching`.

newton/_src/utils/selection.py
mzamoramora-nvidia: Agreed. Group recording is no longer conditional on a user-provided label. Unlabeled native builders now receive family-specific labels (`curve_N`, `surface_N`, or `volume_N`), while explicit labels continue to provide stable application identities. The tests cover rods, rod graphs, cloth, soft meshes, soft grids, and replication of generated labels.

newton/_src/utils/selection.py
mzamoramora-nvidia: Implemented as suggested. `family` is now optional: it is inferred when the matched groups belong to one family and remains available as a filter or to disambiguate a pattern that matches several families. A mixed-family match without the filter reports a clear error. `test_family_is_inferred_for_one_family_match` and `test_family_filters_or_disambiguates_mixed_matches` cover both paths.

docs/concepts/usd_parsing.rst
mzamoramora-nvidia: Agreed. I removed `world_indices` from all four setters and from the documentation and example. Indexed writes now use only the flat `group_indices` axis. When one selected group exists per world, those rows naturally coincide with world order; broader selections can use `world_starts` and `world_ids` to map worlds to groups. `test_group_indices_are_the_only_indexed_write_selector` protects the public signature.

newton/_src/utils/selection.py
mzamoramora-nvidia: This restriction is removed. Getter layouts are prepared and cached per view/source. Regular layouts return a live zero-copy view; irregular equal-sized layouts reuse an internal staging array whose gather can be captured after view construction and kernel warm-up. `test_irregular_getters_capture_all_state_dtypes` captures and replays particle positions, body transforms, and body spatial velocities.

newton/_src/utils/selection.py
mzamoramora-nvidia: Implemented both parts. For regular equal-sized layouts, getters return cached strided views directly into the model or state arrays, so updates to the source remain visible without a copy. For irregular equal-sized layouts, the view allocates one internal contiguous staging result and reuses it on later calls. `test_regular_getter_returns_a_live_state_view` and `test_irregular_getter_reuses_internal_staging` cover those contracts.

mmichelis:
For Isaac Lab cloning parity, could the builder-time deformable label/world metadata be exposed publicly, similar to `ModelBuilder.body_label` / `body_world` and `articulation_label` / `articulation_world`?

PR3326 currently keeps the corresponding registries private as `_cable_label` / `_cable_world`, `_cloth_label` / `_cloth_world`, and `_soft_label` / `_soft_world`, with public access available only after finalization through `DeformableView`. Isaac Lab must rebase cloned USD labels before `finalize()`, so `DeformableView` is too late and we currently have to access those private builder fields.

A minimal API could expose mutable builder properties using the public family terminology:

- `curve_label` / `curve_world`
- `surface_label` / `surface_world`
- `volume_label` / `volume_world`

Only labels and world assignments need builder-time access. Particle and topology ranges can remain private behind `DeformableView`. This would let deformables use the same generic clone-label rebasing path as rigid bodies and articulations, and remove Isaac Lab access to Newton private fields.

jcarius-nv:
IsaacLab team expressed interest again in this so their integration can be simplified. I'm putting it back on the 1.7 Milestone.

PR #3330, opened 2026-07-02
Add Kamino DR Legs benchmark

### LINKED ISSUE #3048: Kamino: DR Legs Benchmark
Benchmark of DR Legs with a pre-trained policy.

camevor:
## Description

Add benchmarks for DR Legs with a policy in the loop.

Uses 4096 worlds to simulate a typical RL locomotion workload.

Resolves #3048 

## Checklist

## Test plan

Sample results on Ubuntu 24.04 with an L40S GPU and a 32-Core AMD EPYC-Genoa CPU:
* Runtime (wall clock): 1m 18s
* KPI (ms/world-step, mean ± std over 5 runs): 0.04753 ± 0.0002

```
uv run --extra torch-cu12 asv/benchmarks/simulation/bench_kamino.py --bench KpiDRLegs

=== Benchmark Results ===
KpiDRLegs.track_simulate (4096,): 0.047539
```

## Summary by CodeRabbit

* **New Features**
  * Added Kamino ASV simulation benchmarks for the “DR Legs” setup.
  * Added an optional deployed control policy path for benchmark runs.
  * Introduced CUDA-graph-based benchmark modes to improve timing consistency.
  * Added a command-line runner to execute selected benchmarks.

* **Performance / Configuration**
  * Tuned benchmark settings for fast and Kpi modes (including updated sample sizing).
  * Updated Kpi configuration to use `4096` parameters and `25` frames per run.
  * Added stronger end-state validation checks (e.g., action sanity and motion limits).


asv/benchmarks/simulation/bench_kamino.py
```diff
@@ -0,0 +1,144 @@
...
+            self._builder = {}
+        if world_count not in self._builder:
+            self._builder[world_count] = Example.create_model_builder(self.robot, world_count)
+
+    @skip_benchmark_if(wp.get_cuda_device_count() == 0)
+    def track_simulate(self, world_count):
+        total_time = 0.0
+        for _iter in range(self.samples):
+            example = Example(
+                robot=self.robot,
```
jcarius-nv, line 82: what do you think about also tracking the startup time in the benchmark? I think next to pure FPS (throughput metrics), often people care about the time-to-first-frame, so having some insights on how we're doing on that front sounds useful to me

asv/benchmarks/benchmark_kamino.py
```diff
@@ -0,0 +1,485 @@
...
+    "j7_l_i",
+    "j2_l_o",
+    "j7_l_o",
+    "j1_r_i",
+    "j2_r_i",
+    "j6_r_i",
+    "j7_r_i",
+    "j2_r_o",
+    "j7_r_o",
+]
```
jcarius-nv, line 45: just out of interest, and independent of this PR really: isn't this something that should be stored with the policy ('output joint names'), or in the model (joints with drives)?

asv/benchmarks/benchmark_kamino.py
```diff
@@ -0,0 +1,485 @@
...
+
+# Walk-policy constants inlined from dr_legs/rl_policies/drlegs_walk.yaml at
+# the pinned asset ref, so an asset-side config change cannot silently alter
+# the benchmarked workload.
+_DRLEGS_WALK_CONFIG = {
+    # Policy
+    "action_scale": 0.4,  # joint position scale [rad]
+    "policy_file": "drlegs_walk.pt",
```
jcarius-nv, line 53: isn't Tobi's PR removing torch script policy support? I might be mistaken, but please see https://github.com/newton-physics/newton/pull/3227 . If it is, might be better to directly start with an onnx policy.

asv/benchmarks/benchmark_kamino.py
```diff
@@ -0,0 +1,485 @@
...
+            self._actions[:] = self._model(obs)
+
+        target_count = len(control.joint_target_q) // wc
+        target_q = wp.to_torch(control.joint_target_q).reshape(wc, target_count)
+        target_q.zero_()
+        target_q[:, self._driven_target_offsets] = self._action_scale * self._actions
+        wp.to_torch(control.joint_target_qd).zero_()
+
+
+class Example:
```
jcarius-nv: Nit: Maybe it's just me, but it seems we should  find a slightly more expressive name for this class? At least as a newcomer to the codebase, not so clear what this class contains or does :D

asv/benchmarks/benchmark_kamino.py
```diff
@@ -0,0 +1,485 @@
...
+            for _ in range(self.decimation):
+                wp.capture_launch(self.reset_graph)
+                wp.capture_launch(self.graph)
+        else:
+            self.simulate()
+        wp.synchronize_device()
+        self.benchmark_time += time.perf_counter() - start_time
+        self.sim_time += self.frame_dt
+
+    def test_final(self):
```
jcarius-nv, line 437: `test_final()` currently checks only that joint and body positions are finite. Could we also validate velocities and add a broad workload-health invariant for the policy-enabled case, such as finite, nontrivial actions and the pelvis remaining within a plausible height band? This should not require successful walking, but it would prevent a broken joint mapping or policy path, or a collapsed robot, from continuing to produce a deceptively valid timing result.

asv/benchmarks/simulation/bench_kamino.py
```diff
@@ -0,0 +1,144 @@
...
+
+class FastDRLegs(_FastBenchmark):
+    num_frames = 25
+    robot = "dr_legs"
+    repeat = 3
+    world_count = 32
+
+
+class KpiDRLegs(_KpiBenchmark):
+    params = [[8192]]
```
jcarius-nv: Could you please include one `KpiDRLegs` result from the target nightly hardware in the test plan? PR CI selects only `Fast*`, so the 8,192-world policy path is not exercised here. It would be useful to record the returned KPI, total runtime, variability across a few runs, and approximate peak GPU memory if readily available, along with why 8,192 is the representative point. I do not think we need many variants; I mainly want to verify that the chosen KPI completes reliably and represents an intended workload.

newton/_src/solvers/kamino/_src/linalg/factorize/rcm_batch.py
```diff
@@ -54,7 +54,8 @@
     launch()  # one zero-arg callable; CUDA-graph capturable
 """
 
-from __future__ import annotations
+# No `from __future__ import annotations`: the kernel factories' `dtype`
+# annotations must evaluate eagerly (the kernel bodies never reference it).
```
jcarius-nv: It looks like this benchmark exposed a pre-existing LLTBRCM annotation bug: postponed annotations turn the kernel-factory `dtype` annotations into strings before Warp registers the nested kernels. That fix has now landed on `main` in #3334 with dedicated solver tests. Could you rebase and drop this hunk from the PR? It may still be worth mentioning in the PR description that the benchmark depended on and helped uncover that fix.

asv/benchmarks/benchmark_kamino.py
```diff
@@ -0,0 +1,485 @@
...
+        policy_pt,
+        model,
+        world_count,
+        config,
+        command=(0.25, 0.0, 0.0),
+    ):
+        import torch  # noqa: PLC0415
+
+        # Reuse the observation helpers from the Kamino RL example to mirror the deployed workload.
+        from newton._src.solvers.kamino.examples.rl.utils import (  # noqa: PLC0415
```
jcarius-nv, line 111: This benchmark mirrors a substantial portion of `example_rl_drlegs.py` while importing private observation helpers from it. Could we either extract and reuse a shared internal fixture, or add a small equivalence test for the pinned asset, for example comparing observations, actions, or the first few states? Pinning the config is good for reproducibility, but without an equivalence check the benchmark and deployed workload can drift independently.

asv/benchmarks/benchmark_kamino.py
camevor: Yes, that PR introduces ONNX support and moves to ONNX for some policies. However, torch will continue to work. I propose continuing to use torch here and updating both the policy in `newton-assets` and this benchmark as a follow-up, once #3227 has landed.

jcarius-nv: ok, I'm fine with that.

newton/_src/solvers/kamino/_src/linalg/factorize/rcm_batch.py
camevor: Yes, rebased on main and removed the change to `rcm_batch.py`.

asv/benchmarks/simulation/bench_kamino.py
camevor: Agreed, we should track those as well. I'll add that as a follow-up: #3339

asv/benchmarks/benchmark_kamino.py
camevor: Good point, this is not really an example as it stands; it simply resembles the Example class we use for the actual examples.

asv/benchmarks/benchmark_kamino.py
camevor: Yes, I'll add some more comprehensive testing.

asv/benchmarks/benchmark_kamino.py
camevor: Agreed, this duplication is not ideal. We plan to move some examples from the Kamino-internal examples directory into Newton’s public examples. Once the DR Legs policy example has moved, this benchmark can run that workload directly, similar to how the SolverMuJoCo benchmark runs `example_robot_anymal_c_walk`. I suggest handling that consolidation as a follow-up rather than introducing another temporary abstraction here.

asv/benchmarks/benchmark_kamino.py
camevor: That's true, and the same could be said about the observation layout. I'm not sure how this would be best done either. Policy, asset and some amount of extra configuration are inseparable.

jcarius-nv (approved):
Reviewed current head `d97358e03d9f476eecdafbe98023231b1dffe2f9`. The benchmark now cleanly keeps the policy in the loop to produce RL-representative states while timing only Newton physics. The 4,096-world KPI was validated at `0.04753 ± 0.0002 ms/world-step` over five runs, and the strengthened health checks cover finite state, body-speed limits, plausible pelvis height, and nontrivial policy actions. The unrelated RCM change is gone, all review threads are resolved, and the full CI suite—including GPU benchmarks and GPU tests—is green; `FastDRLegs` remains unchanged at approximately 1.70 s. Startup benchmarking (#3339), eventual ONNX migration, and workload-harness consolidation are appropriate follow-ups and do not need to block this PR. No remaining blocker from my side.

PR #3334, opened 2026-07-03
Dev/gm/fix lltbrcm annotations

Guirec-Maloisel:
## Description

This fixes an issue in the LLTBRCM solver, where `from __future__ import annotations` broke a kernel factory taking a type as argument. The fix for this in other files is to use `Any` in type annotations depending on the factory argument, but in this case it turns out that `from __future__ import annotations` was superfluous, so I just removed it.

## Checklist

## Test plan

This PR also adds unit tests for this solver (very similar to those for LLTB, just taking into account the permutation), as this would likely have been caught earlier if this functionality was being tested.


chschuma-disney (approved):
Thanks for the quick fix and extra tests. LGTM :rocket:

PR #3335, opened 2026-07-03
Fix USD distance-joint limit handling

adenzler-nvidia:
USD's physics descriptor does not expose authored `minDistance` and `maxDistance` reliably, so `ModelBuilder.add_usd()` currently creates an unlimited distance joint. Read authored bounds from `UsdPhysics.DistanceJoint` directly while preserving the existing unlimited representation for unauthored bounds.

XPBD also used a signed, joint-local displacement as a world-space constraint gradient. Compute a normalized world-space anchor direction instead so lower and upper distance bounds converge consistently, including for rotated joint frames and near-coincident anchors.

Regression coverage exercises USD import and both XPBD limit directions on CPU and CUDA. The changelog records the corrected behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved XPBD distance-joint solving by computing constraint direction from the world-space parent/child anchor separation and enforcing minimum/maximum anchor distance.
  * Fixed USD distance-joint limit import to apply authored min/max distances directly to the joint’s distance DOF, removing conditional gating.
* **Tests**
  * Added an XPBD test to verify distance joints clamp the child-body distance within configured bounds after one solver step.
  * Updated the USD distance-joint importer test to validate min/max distance mapping for the correct DOF.


eric-heiden (dismissed):

newton/_src/solvers/xpbd/kernels.py
```diff
@@ -1620,28 +1620,28 @@ def solve_body_joints(
...
         elif upper >= 0.0 and d > upper:
             err = d - upper
 
         if wp.abs(err) > 1e-9:
             # compute gradients
-            linear_c = rel_p
+            if d > 1e-9:
+                linear_c = anchor_delta / d
+            else:
+                linear_c = wp.normalize(world_com_c - world_com_p)
```
vreutskyy: Could we also cover the case where both anchors and body COMs coincide? In that situation, `world_com_c - world_com_p` is zero, so the lower-limit correction appears to have no direction. In a quick CPU test with `min_distance=1.0`, the distance remained at `0.0` after ten iterations. Would a deterministic world-space fallback direction make sense here?

vreutskyy (commented):
One non-blocking question:

vreutskyy (approved):

PR #3350, opened 2026-07-05
Resolve bound materials with canonical ComputeBoundMaterial

ooctipus:
## Description

`_get_bound_material` returned `ComputeBoundMaterial`'s result only when the prim had `MaterialBindingAPI` formally applied, falling back to reading the prim's own `material:binding` relationship. An ancestor binding with `strongerThanDescendants` strength — the standard way to rebind an asset's appearance for domain randomization — was therefore silently ignored on meshes that author bindings without applying the schema.

Per review: the fix now calls `ComputeBoundMaterial()` **unconditionally** and deletes both manual fallbacks (the ancestor walk from the first revision and the pre-existing raw-relationship scan). Ancestor overrides, binding strength, collection-based bindings, and purpose resolution all follow the USD spec instead of a partial reimplementation.

**Warning verification** (as requested):
- The diagnostic is real: `TfWarn` from `BindingsAtPrim` (`materialBindingAPI.cpp:669`), emitted **per query** on any prim that authors `material:binding` without the applied schema — resolution still succeeds.
- There is no Python-side suppression: `TfDiagnosticMgr` has no Python binding.
- Normalizing offenders by applying the schema into the **session layer** works for editable prims but **fails on instance proxies** (`UsdStage::_ValidateEditPrim` rejects the edit), and instanced assets are the common case for these files — so suppression-by-normalization cannot be complete.
- Conclusion: the warning stands as correct diagnostics on invalid assets. The importer's per-prim resolution cache bounds it to one warning per non-conformant prim per import, and the docstring/changelog point users at `usdchecker` / `usd-validation-nvidia` to fix assets at source.

## Checklist

## Test plan

Three tests (first two fail without the fix, third is a regression guard that also passes pre-fix):
- `test_ancestor_binding_overrides_unapplied_mesh_binding` — the original bug: valid ancestor `strongerThanDescendants` bind vs a mesh's bare rel.
- `test_collection_based_ancestor_binding_resolves` — collection-based ancestor rebind, which only canonical resolution handles.
- `test_ancestor_binding_overrides_applied_mesh_binding_across_depth` — properly applied mesh binding overridden from two levels up.

```
uv run --extra dev -m newton.tests -k "ancestor_binding or collection_based"
```

Full `test_import_usd.py` shows no new failures relative to main.

## Bug fix

**Steps to reproduce:**

1. Take an asset whose meshes author `material:binding` without applying `MaterialBindingAPI`.
2. On the asset root, bind a different material with `bindMaterialAs="strongerThanDescendants"`.
3. Import with `parse_usd`: meshes keep their original material instead of the override.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed USD material import so ancestor bindings with stronger precedence correctly override a mesh’s own `material:binding`, even when the mesh doesn’t apply MaterialBinding support.
  * Updated material resolution to use canonical computed bound materials exclusively (including collection-based bindings), removing non-canonical fallbacks. Schema-less or non-conformant bindings are treated as invalid USD, surfacing warnings once per prim per import.
* **Tests**
  * Added regression coverage for ancestor override precedence (across hierarchy depth) and expected sRGB color-space conversion, including collection-based bindings.


andrewkaufman:
> Many assets author material bindings without applying the schema

This is illegal for a long time in USD. Those assets are broken & should be fixed at source. Both `usdchecker` from OpenUSD and `usd-validation-nvidia` (wheel on pypi) provide automated validation & fixing of this issue.

ooctipus:
Yeah, agreed — those assets are broken and should get fixed at source. But I think the PR title/description oversold the "unapplied bindings" angle (my bad not catching the agent wording).

I believe Newton already tolerates the bare `material:binding` rel. The bug is that a completely *valid* binding — ancestor prim, `MaterialBindingAPI` properly applied, `strongerThanDescendants` — was losing to that fallback. So if you rebind an asset's appearance at the root (the standard way to do domain randomization), the importer ignored it and kept baking the original material. That's what the regression test covers: a fully schema-correct ancestor rebind losing to a mesh's bare rel.

So the PR is really about resolution precedence — honoring the valid opinion over the tolerated-invalid one — not about adding more tolerance for broken assets.

Can you review it again ? THank you :D

newton/_src/usd/utils.py
```diff
@@ -1878,7 +1878,28 @@ def _get_bound_material(target_prim: Usd.Prim) -> UsdShade.Material | None:
     if target_prim.HasAPI(UsdShade.MaterialBindingAPI):
         binding_api = UsdShade.MaterialBindingAPI(target_prim)
         bound_material, _ = binding_api.ComputeBoundMaterial()
-        return bound_material
+        if bound_material:
+            return bound_material
+
+    # An ancestor binding with the strongerThanDescendants strength overrides any binding the
```
andrewkaufman: This is where the real question lies. `ComputeBoundMaterial()` already resolves the full binding hierarchy — ancestor overrides, `strongerThanDescendants` strength, collection bindings, everything. It's the canonical resolution function.

What happens if you just ungate it?
```python
binding_api = UsdShade.MaterialBindingAPI(target_prim)
bound_material, _ = binding_api.ComputeBoundMaterial()
if bound_material:
    return bound_material
```
(Remove the `HasAPI` check entirely.)

If the concern is USD diagnostic noise on unapplied prims, can you verify that's actually emitted? And if so, is it a TfWarn or TfStatus? If it's suppressible, that's a one-line fix vs a 20-line manual ancestor walk that partially reimplements USD's own resolution logic (and will drift as USD evolves).

newton/_src/usd/utils.py
```diff
@@ -1878,7 +1878,28 @@ def _get_bound_material(target_prim: Usd.Prim) -> UsdShade.Material | None:
...
+            ancestor_api = UsdShade.MaterialBindingAPI(ancestor)
+            rel = ancestor_api.GetDirectBindingRel()
+            if (
+                rel
+                and UsdShade.MaterialBindingAPI.GetMaterialBindingStrength(rel)
+                == UsdShade.Tokens.strongerThanDescendants
+            ):
+                material = ancestor_api.GetDirectBinding().GetMaterial()
+                if material:
+                    return material
```
andrewkaufman: This manual walk has several gaps relative to what `ComputeBoundMaterial` handles:

1. **Only checks direct bindings** — misses collection-based ancestor bindings (`material:binding:collection:*`), which are the standard way to rebind subsets of a hierarchy (e.g. different materials for different body parts in domain randomization).

2. **Only checks immediate `strongerThanDescendants`** — doesn't handle the case where multiple ancestors have bindings with different strengths. The real resolution has precedence rules for depth + strength.

3. **Stops at first matching ancestor** — what if a closer ancestor has a weaker binding and a farther ancestor has a stronger one? `ComputeBoundMaterial` handles this; this loop doesn't.

4. **Purpose-aware bindings ignored** — `material:binding:full` / `material:binding:preview` resolution is not considered here.

This is exactly why hand-rolling resolution is risky: the binding spec is more complex than direct + strength, and partial reimplementations will miss edge cases that `ComputeBoundMaterial` already handles correctly.

newton/tests/test_import_usd.py
```diff
@@ -11138,6 +11138,50 @@ def test_preview_surface_color_is_converted_to_display_space(self):
...
+            return material
+
+        original = define_material("/World/Looks/Original", (1.0, 0.0, 0.0))
+        override = define_material("/World/Looks/Override", (0.0, 1.0, 0.0))
+
+        # mesh's own binding: a bare relationship, MaterialBindingAPI deliberately NOT applied
+        mesh.GetPrim().CreateRelationship("material:binding").SetTargets([original.GetPrim().GetPath()])
+        # ancestor override with descendant-winning strength
+        ancestor_binding = UsdShade.MaterialBindingAPI.Apply(root.GetPrim())
+        ancestor_binding.Bind(override, bindingStrength=UsdShade.Tokens.strongerThanDescendants)
```
andrewkaufman, line 12251: Good test — covers the core scenario. Two suggestions:

1. Add a case where the mesh DOES have `MaterialBindingAPI` applied (not just bare rel). The current test only covers the unapplied case; need to verify the applied case still works correctly after this change (regression guard for the `if bound_material` → early-return change on line 1881).

2. The test binds ancestor at immediate parent. Consider a deeper hierarchy (grandparent binding overrides child) to exercise the `while` loop beyond one iteration.

newton/_src/usd/utils.py
```diff
@@ -1878,7 +1878,28 @@ def _get_bound_material(target_prim: Usd.Prim) -> UsdShade.Material | None:
     if target_prim.HasAPI(UsdShade.MaterialBindingAPI):
         binding_api = UsdShade.MaterialBindingAPI(target_prim)
         bound_material, _ = binding_api.ComputeBoundMaterial()
-        return bound_material
+        if bound_material:
```
andrewkaufman: Subtle behavior change here: previously, if the prim had `MaterialBindingAPI` applied but `ComputeBoundMaterial` returned nothing (no binding resolved), the function returned `None` immediately (implicit return from the early `return bound_material`). Now it falls through to the ancestor walk and then the raw-rel scan.

Is that intentional? A prim with `MaterialBindingAPI` applied + no resolved binding falling through to the raw-rel scan seems wrong — if the schema is applied, the absence of a resolved binding IS the answer (no material).

andrewkaufman (commented):
The bug is real — ancestor `strongerThanDescendants` bindings being silently ignored breaks domain randomization workflows. But the fix is over-engineered: it manually re-implements resolution logic that `ComputeBoundMaterial` already handles.

Core question: can we just remove the `HasAPI` gate and call `ComputeBoundMaterial()` unconditionally? The description claims this would "spam USD warnings" on prims with bare `material:binding` rels, but that needs verification — `ComputeBoundMaterial` may only warn if there are actual binding conflicts to report, not merely because the schema isn't applied. If the warning is confirmed real, the right fix is to suppress it (e.g. `Tf.Status` / diagnostic suppression) rather than hand-rolling ancestor resolution.

newton/_src/usd/utils.py
ooctipus: Verified it empirically, and you were right to push — reworked in 32dbb3658 to exactly your form: ungated `ComputeBoundMaterial()`, manual walk deleted (and the pre-existing raw-rel scan too, since CBM tolerates bare rels).

The warning findings: it's real — `TfWarn` from `BindingsAtPrim` (materialBindingAPI.cpp:669), fired *per query* on any prim authoring `material:binding` without the applied schema, even with zero conflicts. Not suppressible from Python (`TfDiagnosticMgr` has no binding). I also tried normalizing offenders by applying the schema into the session layer: works on editable prims, but `UsdStage::_ValidateEditPrim` rejects it on instance proxies — and instanced assets are the common case for these files, so suppression-by-normalization can't be complete either.

So the warning stands as correct diagnostics on invalid assets: bounded to one per non-conformant prim per import by the existing resolution cache, with the docstring pointing at `usdchecker` / `usd-validation-nvidia`.

newton/_src/usd/utils.py
ooctipus: All four gaps are gone by deletion — the walk no longer exists (32dbb3658). Added `test_collection_based_ancestor_binding_resolves` specifically for gap 1, since only canonical resolution passes it.

newton/tests/test_import_usd.py
ooctipus: Both added in 32dbb3658: `test_ancestor_binding_overrides_applied_mesh_binding_across_depth` covers the applied-API mesh case with the override two levels up (grandparent). Note it also passes on main — the applied case worked before — so it's a regression guard rather than a bug demo; the collection-binding test is the one only the new resolution passes.

newton/_src/usd/utils.py
ooctipus: Moot in 32dbb3658 — there are no fallbacks left. Applied API + nothing resolved now returns None from the single `ComputeBoundMaterial` call, which matches the semantics you describe: if the schema is applied and nothing resolves, that IS the answer.

newton/_src/usd/utils.py
andrewkaufman: Perfect — exactly the form I suggested. The ungated single-call approach is clean and correct. Warning investigation is thorough; agree the conclusion (let it stand, cache-bounded) is the right call given instance-proxy limitations.

newton/_src/usd/utils.py
andrewkaufman: All four gaps eliminated by deletion. The collection-binding test specifically validates gap 1 — only canonical resolution passes it. Good.

newton/tests/test_import_usd.py
andrewkaufman: Both addressed: applied-API mesh + grandparent depth in `test_ancestor_binding_overrides_applied_mesh_binding_across_depth`, collection-based binding in `test_collection_based_ancestor_binding_resolves`. The applied-API test is a proper regression guard (passes on main too, confirms the refactor didn\x27t break it). ✓

newton/_src/usd/utils.py
andrewkaufman: Moot — correct. No fallbacks remain, so the semantics are now unambiguous: CBM resolves or it doesn\x27t. Clean.

newton/_src/usd/utils.py
```diff
@@ -1872,30 +1872,21 @@ def _extract_material_input_properties(material: UsdShade.Material | None, prim:
...
+    """
     if not target_prim or not target_prim.IsValid():
         return None
-    if target_prim.HasAPI(UsdShade.MaterialBindingAPI):
-        binding_api = UsdShade.MaterialBindingAPI(target_prim)
-        bound_material, _ = binding_api.ComputeBoundMaterial()
-        return bound_material
-
-    # Some assets author material:binding relationships without applying MaterialBindingAPI.
-    rels = [rel for rel in target_prim.GetRelationships() if rel.GetName().startswith("material:binding")]
```
andrewkaufman, line 2521: I think dropping support for these anchient, broken assets is the right thing todo. But it probably merits an additional callout in the Changelog under the `# Removed` section (or wherever that kind of thing usually goes @eric-heiden?)

andrewkaufman (commented):
Clean rework. The unconditional ComputeBoundMaterial is exactly right — one call replaces three fallback paths. Warning documentation is thorough and the conclusion (let it stand, bounded by cache) is correct. Tests cover all the gaps. LGTM.

ooctipus:
Added a callout under `### Removed` in aebc746ec (happy to move it wherever @eric-heiden prefers). One precision on what's actually dropped, since `ComputeBoundMaterial` still tolerates plain bare `material:binding` rels (with USD's warning): the removal really affects assets relying on bare *purpose-specific* relationships only (e.g. `material:binding:preview` with no universal binding) or on binding targets that aren't `Material` prims — the deleted scan accepted both, canonical resolution doesn't. The entry says so and points at `usdchecker` / `usd-validation-nvidia` for repair.

andrewkaufman (approved):

eric-heiden (approved):

PR #3351, opened 2026-07-05
Split mesh subsets on binding structure, not material vocabulary

ooctipus:
## Description

Import topology currently depends on whether Newton recognizes a material's shader vocabulary: the subset splitter skips any subset whose bound material resolves no visual properties, merging its faces into the parent fallback. Rebinding one articulation variant to an unrecognized material (e.g. an MDL shader) therefore changes its imported shape count relative to identical siblings, and multi-world identical-articulation validation trips.

This PR splits on the **authored binding structure** instead: a subset that binds a material becomes its own submesh regardless of whether the material's properties resolve — unrecognized materials degrade to *unshaded*, never to *different topology*. Subsets that bind no material at all keep the existing behavior (fall through to the parent-mesh material fallback).

Per the review discussion: this replaces the original approach of teaching the property extractor OmniGlass input names. No MDL vocabulary is added — unrecognized materials import as valid-but-unshaded, and Newton stays indifferent to material dialects.

## Checklist

## Test plan

`test_subset_splitting_is_independent_of_material_vocabulary` builds two identical meshes with materialBind subsets — one bound to UsdPreviewSurface materials, one to shaders with unrecognized inputs — and asserts both import with the same shape count. Verified the test fails without the fix. The existing subset tests (`test_visual_mesh_material_subsets_create_separate_visual_shapes`, unbound-subset fallback behavior) pass unchanged; full `test_import_usd.py` shows no new failures relative to main.

```
uv run --extra dev -m newton.tests -k test_subset_splitting_is_independent_of_material_vocabulary
```

## Bug fix

**Steps to reproduce:**

1. Author a mesh with materialBind GeomSubsets, each binding a material.
2. Import once with the subsets bound to UsdPreviewSurface materials, once rebound to materials whose shader inputs Newton does not recognize.
3. Compare `builder.shape_count`: the unrecognized version imports fewer shapes (its subsets merged into the parent fallback).

**Minimal reproduction:**

```python
import newton
from pxr import Sdf, Usd, UsdGeom, UsdPhysics, UsdShade, Vt

# identical meshes; one's subsets bind PreviewSurface materials, the other's bind a shader
# with unrecognized inputs (e.g. mystery_tint). Before this PR the second imports with a
# different shape count than the first.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed USD importing so authored face subsets keep their intended split structure even when some bound materials can’t be resolved.
  * Subsets bound to unsupported/unknown materials are now imported as separate submeshes, preventing incorrect shape/mesh counts.
  * Subsets without authored material bindings still follow the existing fallback behavior.
* **Tests**
  * Added a USD importer regression test covering subset splitting independent of material vocabulary/resolution.
* **Documentation**
  * Updated the changelog with the USD import fix.


CHANGELOG.md
```diff
@@ -55,6 +55,7 @@
 
 ### Fixed
 
+- Fix USD material property extraction not recognizing OmniGlass shader inputs (`glass_color`, `frosting_roughness`); meshes bound to OmniGlass materials resolved no visual properties, which also skipped their subsets in mesh splitting and changed the imported shape count.
```
andrewkaufman: Newton does not support any MDL materials as far as I know. I don't think we should set the prescedent to start doing it. These attributes are arbitrarily named, any other MDL could use a different input name for the same meaning.

Instead, we should validate that source assets use portable materials (UsdPreviewSurface for sure, _possibly_ OpenPBR from MaterialX).

ooctipus:
Quick question on what the intended contract is, since it affects how we handle this on the Isaac Lab side: when an asset carries an MDL material Newton doesn't recognize, is the expectation that it's not valid input at all (import should error/warn and tell the user to fix the asset), or that it imports fine and just doesn't render with those properties?

Asking because right now it's neither — the import succeeds silently, but a mesh whose material doesn't resolve gets *split differently* (subsets with no resolved properties are skipped by the splitter), so the same articulation bound to different materials can import with different shape counts. That's actually the bug that bit us: multi-world validation trips because one variant has 59 shapes and its siblings 64.

If the intent is "valid but unshaded," would you be open to a PR that splits purely on authored GeomSubsets, independent of material resolution? Then unknown materials would degrade to unshaded instead of changing topology, and I'd be happy to drop the glass names from this PR — we can live with glass rendering plain in Newton.

ooctipus:
Went ahead with the decoupling since it stands regardless of the vocabulary question: the PR now splits subsets on the authored binding structure only, and the OmniGlass input names are gone. Unrecognized materials import as their own unshaded submeshes instead of changing the mesh's shape count — no MDL assumptions anywhere. Retitled accordingly.

newton/_src/utils/import_usd.py
```diff
@@ -771,13 +771,15 @@ def _get_visual_material_subset_meshes(prim: Usd.Prim) -> list[tuple[str, Mesh]]
...
-            if not any(value is not None for value in material_props.values()):
-                continue
+            # Split on authored binding structure, not on whether the bound material's properties
+            # resolve: a subset that binds a material Newton does not recognize still becomes its
+            # own (unshaded) submesh, so import topology never depends on material vocabulary.
+            # Only subsets that bind no material at all fall through to the uncovered-faces
+            # fallback below, which applies the parent mesh material.
             subset = UsdGeom.Subset(stage.GetPrimAtPath(subset_path))
+            binding_rel = subset.GetPrim().GetRelationship("material:binding")
+            if not (binding_rel and binding_rel.GetTargets()):
```
andrewkaufman: This raw relationship check is inconsistent with how `_get_bound_material` (in `usd/utils.py`) resolves materials. That function uses `UsdShade.MaterialBindingAPI.ComputeBoundMaterial()`, which handles inherited bindings, collection-based bindings, and binding-strength resolution. Here you're only checking whether `material:binding` is authored directly on the subset prim.

This creates a behavioral split: `resolve_material_properties_for_prim` (called earlier on line 767) resolves material props using the full binding API (inherited + collections), while this gate only sees direct authoring. A subset that inherits its binding from an ancestor or receives it via collection binding would pass property resolution (get real material props) but fail this gate and get skipped — the opposite of the intended behavior.

Consider:
```python
binding_api = UsdShade.MaterialBindingAPI(subset.GetPrim())
bound_material, _ = binding_api.ComputeBoundMaterial()
if not bound_material:
    continue
```
This keeps the gate consistent with the resolution path that already runs on line 767 and correctly handles inherited/collection bindings.

newton/tests/test_import_usd.py
```diff
@@ -5808,6 +5808,75 @@ def test_visual_mesh_material_subsets_create_separate_visual_shapes(self):
...
+        UsdGeom.SetStageUpAxis(stage, UsdGeom.Tokens.z)
+        UsdGeom.SetStageMetersPerUnit(stage, 1.0)
+        UsdPhysics.Scene.Define(stage, "/physicsScene")
+
+        def define_unknown_material(path: str) -> UsdShade.Material:
+            """An MDL-style material whose shader inputs Newton does not recognize."""
+            material = UsdShade.Material.Define(stage, path)
+            shader = UsdShade.Shader.Define(stage, f"{path}/Shader")
+            shader.CreateInput("mystery_tint", Sdf.ValueTypeNames.Color3f).Set((0.2, 0.6, 0.9))
+            shader.CreateInput("mystery_response", Sdf.ValueTypeNames.Float).Set(0.35)
```
andrewkaufman, line 6908: The "unknown" material helper creates a Shader prim with inputs but never connects it to the Material via a surface output. This means `_resolve_prim_material_properties` won't find it through `GetSurfaceOutput()` — it falls through to the child-scan fallback, finds the Shader, calls `_extract_shader_properties`, and gets nothing because the input names don't match.

This correctly exercises the scenario, but a real-world MDL material _would_ have an `mdl:surface` output connection (just with unrecognized input names). Consider adding a variant that wires `material.CreateOutput("mdl:surface", ...).ConnectToSource(shader, ...)` to ensure that code path also produces the same topology-independent behavior. It's a slightly different branch through `_resolve_prim_material_properties`.

newton/tests/test_import_usd.py
```diff
@@ -5808,6 +5808,75 @@ def test_visual_mesh_material_subsets_create_separate_visual_shapes(self):
...
+        result = builder.add_usd(stage)
+
+        known_shapes = [label for label in builder.shape_label if label.startswith("/Known/")]
+        unknown_shapes = [label for label in builder.shape_label if label.startswith("/Unknown/")]
+        self.assertEqual(len(known_shapes), 2)
+        self.assertEqual(len(unknown_shapes), 2, "unrecognized materials must not change import topology")
+        self.assertIn("/Unknown/VisualMesh/part_0", result["path_shape_map"])
+        self.assertIn("/Unknown/VisualMesh/part_1", result["path_shape_map"])
+
     @unittest.skipUnless(USD_AVAILABLE, "Requires usd-core")
```
andrewkaufman, line 6977: The test verifies shape *count* equivalence (both == 2), which is the key regression check. But it doesn't verify the geometry is actually correct — i.e., that both submeshes got the right face indices and the full mesh is covered. Consider adding an assertion that both Known and Unknown bodies cover all faces (no faces dropped into the fallback path). Something like verifying no fallback/parent-path entry appears in `path_shape_map` for either body.

andrewkaufman (commented):
Good fix — decoupling topology from material vocabulary is correct. The binding-structure gate is the right criterion. A few observations, one worth addressing before merge.

newton/_src/utils/import_usd.py
ooctipus: Good catch on collection bindings — fixed in dabdb45ac: the gate now accepts any `material:binding*` relationship with targets authored on the subset (direct or collection, with or without the API applied).

I deliberately didn't switch to `ComputeBoundMaterial` though, and added a comment explaining why: on a GeomSubset it always resolves *something* via inheritance — every unbound subset would inherit the parent mesh's material and pass the gate, which breaks the unbound-subset fallback (materialless subsets are supposed to keep showing the parent material). Worse, an ancestor rebind with `strongerThanDescendants` resolves identically for all subsets, so topology would depend on rebinding again — the exact bug this PR fixes. The gate needs "an opinion authored on the subset itself", which full resolution can't distinguish from inheritance. The property *resolution* on line 767 still uses the full binding API, so inherited/collection appearance is unaffected — only the split decision reads authored structure.

newton/tests/test_import_usd.py
ooctipus: Added in dabdb45ac: the unknown-material helper now has a `connect_surface` variant that wires `mdl:surface` → shader output like a real MDL material, and the test asserts a third body using it imports with identical topology — so both the surface-output branch and the child-scan fallback are covered.

newton/tests/test_import_usd.py
ooctipus: Added in dabdb45ac: the test now asserts the exact per-body shape labels (only the two subset paths — a parent-mesh fallback entry would fail the comparison) and that each submesh owns one of the mesh's two triangles, so full face coverage is verified rather than just counts.

andrewkaufman: I don't think dabdb45ac was pushed?

ooctipus:
You're right, sorry — my push raced your branch update and got rejected (misread the rejection as a success). dabdb45ac is actually up now, merged with the updated branch head (e1333f9d1).

newton/_src/utils/import_usd.py
andrewkaufman: The contributor's rationale is sound — `ComputeBoundMaterial` resolves inherited bindings for every subset (so unbound subsets would pass the gate and break the fallback path). The authored-relationship scan is the correct gate here. Note that `_get_bound_material` in `usd/utils.py` already uses the same `GetRelationships()` fallback pattern for un-applied MaterialBindingAPI cases, so this is internally consistent. The broadened `material:binding*` prefix check in dabdb45ac handles collection-based bindings. Satisfied.

newton/tests/test_import_usd.py
andrewkaufman: Addressed in dabdb45ac — the `UnknownMdl` body wires `mdl:surface` → shader output, covering the surface-output resolution branch alongside the child-scan fallback. Both paths produce identical topology. ✓

newton/tests/test_import_usd.py
andrewkaufman: Addressed in dabdb45ac — test now asserts exact shape labels (only the two subset paths, so a parent-mesh fallback entry would fail) plus per-submesh triangle count (`len(submesh.indices) == 3`). Full face coverage verified, not just counts. ✓

andrewkaufman (commented):
All three feedback items addressed in dabdb45ac. The binding-structure gate is correct and the test coverage is solid. LGTM.

ooctipus:
can you @andrewkaufman approve running the tests? Thank you!

eric-heiden (dismissed):
I think this looks good, splitting based on material binding regardless whether Newton actually supports the material makes sense.

eric-heiden:
@ooctipus can you merge the changelog once again? Then we can merge your PR.

eric-heiden (approved):

PR #3356, opened 2026-07-06
Modernize neural-actuator checkpoint loading

### LINKED ISSUE #3250: Modernize neural-actuator checkpoint loading: stop using deprecated torch.jit APIs and revisit the pickle fallback
### Context

`newton.actuators` is marked experimental (`newton/actuators.py:12`) but the neural controllers (`ControllerNeuralMLP`, `ControllerNeuralLSTM`) and their checkpoint formats shipped in v1.3.0. The checkpoint loader is `newton/_src/actuators/utils.py` (`_load`, lines 10–36). #3055 reverted the ONNX runtime "while we redesign policy loading"; with that ONNX work now back in flight, this issue records two concrete problems the strict-warnings effort (#3044) surfaced so they can be folded into that redesign.

### Problem 1 — the loader uses deprecated `torch.jit` APIs

With current PyTorch (verified on 2.12.0+cu130), both APIs the loader/fixtures rely on are deprecated:

- `` `torch.jit.script` is deprecated. Please switch to `torch.compile` or `torch.export`. ``
- `` `torch.jit.load` is deprecated. Please switch to `torch.export`. ``

The runtime hits this at `utils.py:23` (`torch.jit.load`). Note the `torch.load(..., weights_only=False)` fallback (`utils.py:29`) *also* trips it: for a TorchScript archive `torch.load` delegates to `torch.jit.load` internally, so under a `DeprecationWarning`→error policy `_load` raises even though the first `jit.load` is caught by `except Exception`. (`torch.export` is the non-deprecated torch route, but whether we keep a torch path at all depends on the format question below.)

### Problem 2 — the dict/pickle fallback (was this intentional?)

The second loader path (`utils.py:29–36`) does `torch.load(path, weights_only=False)` and returns `checkpoint["model"]` directly. Several concerns:

- **Mis-documented contract.** The docstrings say the format is `{"model": state_dict, ...}` (`utils.py` `load_checkpoint`; `controller_neural_mlp.py:34,71`; `controller_neural_lstm.py:40,77`). But the code returns `checkpoint["model"]` *as the network*, and the controllers then call `self.network.to(device)` and `self.network(net_input)` — which require a live, pickled `nn.Module`, not a `state_dict`. A real `state_dict` (an `OrderedDict`) would crash; there's no `load_state_dict`/rebuild anywhere. The tests confirm it by saving `{"model": net}` (the whole module).
- **`weights_only=False`** → arbitrary code execution when loading untrusted checkpoints, and PyTorch is moving the `weights_only` default the other way.
- **Not portable.** Unpickling a whole `nn.Module` needs the model's class importable at load time, so this path is a weaker, redundant variant rather than a distinct capability.
- **Broad `except Exception: pass`** (`utils.py:26`) masks real load failures (corrupt archive, future hard-removal of `jit.load`) behind the generic bottom-line error.

**Questions:**
1. Was the pickled-full-`nn.Module` path intentional, or was a real `state_dict` loader intended (the docstrings suggest the latter)?
2. Is there a concrete use case for it that TorchScript/ONNX don't cover? If not, can we drop it?

### Backwards compatibility

Likely a non-issue. `newton.actuators` is experimental, which reserves the right to change the format without the usual deprecation cycle (a CHANGELOG `Changed` entry suffices). Exposure is small too: v1.3.0 is recent, there's no documented checkpoint-authoring workflow (the format only appears in docstrings), and no in-tree example ships a neural-controller checkpoint. So the format can change/drop freely without keeping the deprecated paths alive for back-compat.

### Coordination with the in-flight ONNX work

The reverted Warp-backed ONNX runtime (#2662) ran inference on Warp from an ONNX file using only the `onnx`/`protobuf` parsers — no `torch`, no `onnxruntime` — and was CUDA-graph-capturable. As the ONNX path returns, the open question is **whether the runtime should be ONNX-only, or expose a torch + ONNX API.** Relevant considerations either way:

- Users train in PyTorch, so they'll bring PyTorch nets. A torch→ONNX step (`torch.onnx.export`, possibly via a thin helper) keeps that first-class while leaving torch as an author-time concern; a torch + ONNX runtime would instead carry torch as a runtime dependency and a second, non-graphable code path (and the Problem 1/2 debt with it).
- A Warp-backed ONNX runtime only runs the ONNX ops it has kernels for, so op coverage (vs. torch running arbitrary models) is part of the trade.
- Resolving this also decides the fate of Problems 1 & 2 (an ONNX-only runtime removes the torch `jit`/pickle paths entirely).

### Note

Short term, a separate PR suppresses these deprecation warnings so strict-warnings CI (#3044) is unblocked; this issue tracks the proper fix.

jvonmuralt:
## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **New Features**
  * Added support for `.pt2` (Torch export) neural-controller checkpoint archives, including reading accompanying `metadata.json` from archive contents.
* **Bug Fixes**
  * Improved checkpoint detection/loading across `.pt2`, TorchScript, and legacy dict formats with reliable metadata-only extraction and consistent evaluation behavior.
  * Neural LSTM now derives/validates `num_layers`/`hidden_size` from exported metadata (with type coercion and clearer validation errors).
* **Documentation**
  * Updated checkpoint and graphability guidance; clarified `.pt2` metadata requirements and backend behavior.  
  * Updated installation guide formatting.
* **Tests**
  * Expanded coverage for `.pt2` metadata, validation rules, and expected deprecation warnings.

### DIFF
--- a/docs/concepts/actuators.rst
+++ b/docs/concepts/actuators.rst
@@ -202,9 +202,22 @@ Whether an actuator supports differentiability and CUDA graph capture depends on
 its controller.  :class:`ControllerPD` and :class:`ControllerPID` are fully
 graphable.  Neural-network controllers (:class:`ControllerNeuralMLP`,
 :class:`ControllerNeuralLSTM`) support two checkpoint backends: ONNX checkpoints
-use Warp-NN's Warp-backed runtime and are graphable, while TorchScript
-checkpoints (``.pt`` / ``.pth``) use the Torch backend, require PyTorch, and are
-not graphable due to framework interop overhead.
+use Warp-NN's Warp-backed runtime and are graphable, while Torch checkpoints
+use the Torch backend, require PyTorch, and are not graphable due to framework
+interop overhead.
+
+Torch checkpoints are pt2 archives (``.pt2``) saved with ``torch.export.save``.
+Checkpoint metadata (scales and network configuration) is stored as a JSON
+extra file::
+
+   metadata = {"effort_scale": 2.0, "num_layers": 2, "hidden_size": 8}
+   torch.export.save(exported, "policy.pt2", extra_files={"metadata.json": json.dumps(metadata)})
+
+:class:`ControllerNeuralLSTM` requires ``num_layers`` and ``hidden_size`` in
+the metadata of pt2 checkpoints, since exported modules no longer expose the
+``torch.nn.LSTM`` submodule that legacy checkpoints are inspected for.  The
+legacy TorchScript (``.pt`` / ``.pth``, saved with ``torch.jit.save``) and
+dict (saved with ``torch.save``) formats still load but are deprecated.
 
 :meth:`Actuator.is_graphable` returns ``True`` when all components can be
 captured in a CUDA graph.
--- a/docs/concepts/actuators.rst
+++ b/docs/concepts/actuators.rst
@@ -208,8 +208,14 @@ interop overhead.
 
 Torch checkpoints are pt2 archives (``.pt2``) saved with ``torch.export.save``.
 Checkpoint metadata (scales and network configuration) is stored as a JSON
-extra file::
+extra file:
 
+.. code-block:: python
+
+   import json
+   import torch
+
+   exported = torch.export.export(net, example_inputs)
    metadata = {"effort_scale": 2.0, "num_layers": 2, "hidden_size": 8}
    torch.export.save(exported, "policy.pt2", extra_files={"metadata.json": json.dumps(metadata)})
 
--- a/docs/concepts/actuators.rst
+++ b/docs/concepts/actuators.rst
@@ -195,16 +195,13 @@ state objects — simply omit them:
 
    m2.actuators[0].step(m2.state(), m2.control())
 
-Differentiability and Graph Capture
------------------------------------
+Neural-Network Checkpoints
+--------------------------
 
-Whether an actuator supports differentiability and CUDA graph capture depends on
-its controller.  :class:`ControllerPD` and :class:`ControllerPID` are fully
-graphable.  Neural-network controllers (:class:`ControllerNeuralMLP`,
-:class:`ControllerNeuralLSTM`) support two checkpoint backends: ONNX checkpoints
-use Warp-NN's Warp-backed runtime and are graphable, while Torch checkpoints
-use the Torch backend, require PyTorch, and are not graphable due to framework
-interop overhead.
+Neural-network controllers (:class:`ControllerNeuralMLP`,
+:class:`ControllerNeuralLSTM`) support two checkpoint backends: ONNX
+checkpoints (``.onnx``) run on Warp-NN's Warp-backed runtime, while Torch
+checkpoints use the Torch backend and require PyTorch.
 
 Torch checkpoints are pt2 archives (``.pt2``) saved with ``torch.export.save``.
 Checkpoint metadata (scales and network configuration) is stored as a JSON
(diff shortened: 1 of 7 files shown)


docs/concepts/actuators.rst
```diff
@@ -202,9 +202,28 @@ Whether an actuator supports differentiability and CUDA graph capture depends on
...
 graphable.  Neural-network controllers (:class:`ControllerNeuralMLP`,
 :class:`ControllerNeuralLSTM`) support two checkpoint backends: ONNX checkpoints
-use Warp-NN's Warp-backed runtime and are graphable, while TorchScript
-checkpoints (``.pt`` / ``.pth``) use the Torch backend, require PyTorch, and are
-not graphable due to framework interop overhead.
+use Warp-NN's Warp-backed runtime and are graphable, while Torch checkpoints
+use the Torch backend, require PyTorch, and are not graphable due to framework
+interop overhead.
+
+Torch checkpoints are pt2 archives (``.pt2``) saved with ``torch.export.save``.
```
camevor, line 206: Question, not introduced by this PR: Why is this called a "checkpoint" rather than just, say, model or policy?

newton/_src/actuators/controllers/controller_neural_lstm.py
```diff
@@ -72,11 +72,21 @@ class ControllerNeuralLSTM(Controller):
...
+    ``torch.export.save``; preferred) and the deprecated TorchScript (``.pt``
+    saved with ``torch.jit.save``) and module-bundle
+    (``{"model": <network module>, "metadata": {...}}`` saved with
+    ``torch.save``) formats. ``num_layers`` and ``hidden_size`` are taken from
+    a live ``lstm`` attribute (``torch.nn.LSTM``) when the network exposes
+    one, otherwise from checkpoint metadata. Exported pt2 modules no longer
+    expose that attribute, so their checkpoints must record ``num_layers``
+    and ``hidden_size`` in metadata.
+
+    ``.onnx`` checkpoints use Warp-NN. The exported ONNX model must have three
```
camevor, line 86: This docstring could be reworded for clarity, and mention that only legacy Torch checkpoints don't need the explicit metadata, while both .pt2 and .onnx files need them. Then the ONNX paragraph stays about what's genuinely ONNX-specific (the three inputs/outputs and the name-mapping keys), and the pt2 paragraph stays about formats.

docs/concepts/actuators.rst
```diff
@@ -202,9 +202,28 @@ Whether an actuator supports differentiability and CUDA graph capture depends on
...
+   import json
+   import torch
+
+   exported = torch.export.export(net, example_inputs)
+   metadata = {"effort_scale": 2.0, "num_layers": 2, "hidden_size": 8}
+   torch.export.save(exported, "policy.pt2", extra_files={"metadata.json": json.dumps(metadata)})
+
+:class:`ControllerNeuralLSTM` requires ``num_layers`` and ``hidden_size`` in
+the metadata of pt2 checkpoints, since exported modules no longer expose the
+``torch.nn.LSTM`` submodule that legacy checkpoints are inspected for.  The
```
camevor: > since exported modules no longer expose the ``torch.nn.LSTM`` submodule that legacy checkpoints are inspected for

The wording here is confusing, as it combines the difference between legacy and new checkpoints with the removal of metadata during export.

newton/_src/actuators/utils.py
```diff
@@ -168,36 +184,92 @@ def _require_torch():
...
+    ``warn`` controls the deprecation warnings for the legacy formats; its
+    ``stacklevel`` assumes the call chain controller ``__init__`` →
+    :func:`load_checkpoint` → here, so the warning points at the user's code.
+    """
     torch = _require_torch()
+    entries = _torch_archive_entries(path)
 
-    extra_files: dict[str, str] = {"metadata.json": ""}
-    try:
+    if "archive_format" in entries:
```
camevor: Raised by a review agent: Torch 2.7 exports did not include an `archive_format` entry, causing this branch to fall through to `torch.load` (which fails). Since Torch is pinned to >= 2.7, this case seems worth handling or documenting / flagging with a specific error.

camevor (dismissed):
Looks good to me! Added a few nits / questions, no blockers.

docs/concepts/actuators.rst
jvonmuralt: To be honest, I dont have a strong preference here. I would say model kind of collides with newton model and might be confusing. Checkpoint is a bit broader then policy as it is any network, not only RL policy.

docs/concepts/actuators.rst
jvonmuralt: thanks, I reorganized this section a bit

docs/concepts/actuators.rst
camevor: Ok, thanks, I think it's fine to leave it as checkpoint, since it's pre-existing.

camevor (approved):
Thanks, LGTM!

PR #3367, opened 2026-07-06
Fix USD site import traversal

adenzler-nvidia:
## Summary

- Keep site loading independent from visual-shape loading.
- Discover sites beneath collider prims without importing visual descendants that were previously excluded.
- Use OpenUSD instance-proxy traversal instead of manually remapping prototype paths.

This preserves the existing `ModelBuilder.add_usd()` API and loading behavior outside the two site-discovery failures. Regression coverage includes disabled visual loading, collider descendants, and instance proxies.

## Summary by CodeRabbit

* **Changed**
  * Improved USD import performance for scenes with many nested transforms and instances.
  * Updated USD import behavior so site discovery no longer depends on whether visual shapes are loaded.
* **Bug Fixes**
  * Fixed USD site detection beneath non-visual containers, collider prims, and instanceable rigid-body prims.
  * Ensured visual geometry is excluded appropriately when visual-shape loading is disabled.
* **Tests**
  * Expanded USD site-import coverage for collider, instance, and visual-shapes-disabled scenarios.
* **Chores**
  * Refreshed the changelog with the latest USD import update.


newton/_src/utils/import_usd.py
```diff
@@ -883,12 +895,34 @@ def _load_visual_shapes_impl(
...
+            allow_visual_shapes: Whether non-site geometry may be loaded from this subtree.
         """
-        if _is_enabled_collider(prim) or prim.HasAPI(UsdPhysics.RigidBodyAPI):
+        if prim.HasAPI(UsdPhysics.RigidBodyAPI):
             return
         path_name = str(prim.GetPath())
         if any(re.match(path, path_name) for path in ignore_paths):
             return
+        is_collider = _is_enabled_collider(prim)
+        if is_collider or prim.IsInstance():
```
eric-heiden: 🟠 `prim.IsInstance()` returns before the instance prim's own type and site APIs are inspected. A typed instance whose referenced root is itself a `Sphere` with `NewtonSiteAPI` has no child geometry to recurse into, so `/World/link/siteInstance` is omitted from `model.shape_label`. Please classify and load supported site geometry on the instance root before traversing its proxies, or restrict this shortcut to untyped/container instances.

<details>

<summary>Minimal USDA reproduction</summary>

```usda
#usda 1.0
def Sphere "SitePrototype" (prepend apiSchemas = ["NewtonSiteAPI"]) {
    double radius = 0.1
}
def Xform "World" {
    def Xform "link" (prepend apiSchemas = ["PhysicsRigidBodyAPI"]) {
        def Sphere "siteInstance" (
            instanceable = true
            prepend references = </SitePrototype>
        ) {
        }
    }
}
```

Importing `/World` with `ModelBuilder.add_usd()` currently produces no shape labels. The expected result includes `/World/link/siteInstance` with `ShapeFlags.SITE`.

</details>

adenzler-nvidia: Fixed in e6e6a08 — the `IsInstance()` shortcut is removed entirely rather than restricted: instance prims now go through regular classification (reading composed attributes off an instance root is fine), and the child recursion already traverses instance proxies, so container instances need no special case. Your repro is added verbatim as `test_site_on_typed_instance_is_loaded`. Note this wasn't a regression from this PR — the previous prototype-children remapping also skipped the instance root's own geometry.

Since classification computes transforms/materials before knowing whether the prim type can produce a shape, removing the shortcut roughly doubled import time on instance-heavy scenes. 8520781 moves the supported-type check ahead of that work, which recovers it and makes container-heavy imports ~2x faster than before this PR (plain `Xform` hierarchies were paying the same wasted cost).

newton/_src/utils/import_usd.py
```diff
@@ -1108,7 +1128,13 @@ def add_body(
         path_body_map[label] = b
         if load_sites or load_visual_shapes:
             for child in prim.GetChildren():
-                _load_visual_shapes_impl(b, child, body_xform=xform, articulation_root_xform=articulation_root_xform)
+                _load_visual_shapes_impl(
+                    b,
+                    child,
+                    body_xform=xform,
+                    articulation_root_xform=articulation_root_xform,
+                    allow_visual_shapes=load_visual_shapes,
```
eric-heiden: 🟠 This initial loop still uses `prim.GetChildren()`, so an instanceable rigid-body root hides site children from `_load_visual_shapes_impl()`. For example, if `/World/link` has `PhysicsRigidBodyAPI`, is instanceable, and references a prototype containing a `Sphere` with `NewtonSiteAPI`, the proxy child `/World/link/site` is never reached. Please start this traversal with `GetFilteredChildren(traverse_instance_proxies)` or call `_load_visual_shape_children()` here, and add a regression test where the rigid-body prim itself is instanceable.

CHANGELOG.md
```diff
@@ -44,6 +44,8 @@
...
 - Align MJCF and USD margin/gap import with MuJoCo 3.9+. Drop the `mj_margin - mj_gap` subtraction; `Model.shape_margin` now matches authored `margin`, and `Model.shape_gap` matches authored `gap`. Pass `legacy_margin_gap=True` to `ModelBuilder.add_mjcf` or `ModelBuilder.add_usd` to restore the pre-3.9 translation for files authored against MuJoCo <= 3.8.
 - Enable particle-particle contact forces by default in standalone `SolverSemiImplicit`: the solver now rebuilds `Model.particle_grid` every substep and adds contact forces to the existing `particle_f`. Existing scenes with particles closer than `2 * particle_max_radius + particle_cohesion` may change trajectories; set `model.particle_grid = None` before stepping to preserve the previous behavior without particle-particle contacts.
+- Change `CollisionPipeline.soft_contact_max` to default to the precomputed `soft_rigid_contact_pair_count` instead of `shape_count * particle_count`.
+- Speed up USD import on scenes with many nested `Xform` or instance prims by skipping transform and material computation for prims that cannot produce shapes.
```
eric-heiden: ⚪ These additions need cleanup. The `CollisionPipeline.soft_contact_max` entry duplicates the next line while omitting its migration guidance, and the USD import entry is a fix or performance note without migration guidance for the `Changed` section. Please remove the duplicate and fold the USD speedup into the existing USD-site entry under `Fixed`.

newton/_src/utils/import_usd.py
adenzler-nvidia: Fixed in 2bf05884f — `add_body` now uses the instance-proxy-aware child traversal, and added `test_site_beneath_instanceable_rigid_body_is_loaded` as regression coverage.

CHANGELOG.md
adenzler-nvidia: Done in db056ff1c — dropped the duplicated `soft_contact_max` entry and folded the speedup note into the USD-site entry under Fixed.

eric-heiden (dismissed):
Thanks! Just need to resolve merge conflicts @adenzler-nvidia.

eric-heiden (dismissed):

eric-heiden (approved):

PR #996, opened 2025-10-28
Register and parse solver-specific custom attributes

eric-heiden:
## Description
This PR adds the function `Solver.register_custom_attributes(ModelBuilder)` which allows a solver to set up custom attributes in the builder so that the user can define them later on programmatically from the builder functions, or parse them from USD or MJCF.

As an example, this PR adds MuJoCo `condim` attribute as a per-shape custom attribute and parses it from USD and MJCF.

The USD importer now handles user-defined custom attributes for shapes, bodies, joints, and articulation.
By adding another type of attribute frequency, `ONCE`, we consider those attributes to be potential candidates for model-specific parameters which the USD importer can parse from the PhysicsScene prim (the attributes we return as `scene_attributes` are now considered for potential model-specific custom attributes).
We assume custom attributes follow the name `"newton:namespace:custom_attr"` in USD.

Addresses #976.
Closes https://github.com/newton-physics/newton/issues/538.

### ToDo's:

- [ ] Update all examples to use `Solver.register_custom_attributes(builder)`

## Before your PR is "Ready for review"

- [ ] Necessary tests have been added and new examples are tested (see `newton/tests/test_examples.py`)
- [ ] Documentation is up-to-date
- [ ] Code passes formatting and linting checks with `pre-commit run -a`

## Summary by CodeRabbit

* **New Features**
  * Unified custom-attribute system with a formal CustomAttribute descriptor and solver registration (MuJoCo condim) that propagates to geometry.

* **Improvements**
  * USD import reworked to a schema-driven flow with new USD helpers and clearer schema_attrs output.
  * URDF/MJCF import now supports custom attributes and raw-XML input.

* **Tests**
  * Added tests validating MuJoCo custom-attribute propagation from code, MJCF, URDF, and USD.

* **Breaking Changes**
  * APIs and examples now use CustomAttribute descriptors and the new USD/schema import entry points; update integrations.

PR #3368, opened 2026-07-06
Fix USD axial visual scaling

adenzler-nvidia:
## Summary

- Scale visual-only capsules, cylinders, and cones along their authored USD axis.
- Use the largest perpendicular scale for the circular radius, matching dimensions resolved by USD Physics collision descriptors.
- Apply the same conversion to sites without a site-specific code path.

Regression coverage compares visual and collision dimensions for all three axial primitives across the X, Y, and Z axes, and separately verifies site dimensions.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed USD import scaling for capsule, cylinder, and cone so visuals and site transforms follow the authored primitive axis under non-uniform scaling.
  - When `radius`/`height` aren’t authored, visuals and sites now use USD schema fallback values consistent with collision geometry.

- **Tests**
  - Added regression tests to confirm axial visual scaling matches collision shape scaling for capsule, cylinder, and cone across X/Y/Z.
  - Added site import tests validating axis-dependent site scale vectors derived from USD transforms.


eric-heiden (commented):
LGTM! Just the CodeRabbit comment is worth checking out.

newton/_src/utils/import_usd.py
```diff
@@ -836,6 +836,13 @@ def _is_uniform_scale(scale: wp.vec3) -> bool:
         scale_np = np.array(scale, dtype=np.float32)
         return bool(np.allclose(scale_np, scale_np[0], rtol=1e-6, atol=1e-6))
 
+    def _get_axial_visual_dimensions(prim: Usd.Prim, scale: wp.vec3, axis: Axis):
+        radius = usd.get_float(prim, "radius", 0.5)
+        half_height = usd.get_float(prim, "height", 2.0) / 2
+        axis_index = int(axis)
+        radius_scale = max(scale[index] for index in range(3) if index != axis_index)
+        return radius * radius_scale, half_height * scale[axis_index]
+
```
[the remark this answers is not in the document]
adenzler-nvidia: Fixed in a21d3cea3. The capsule height default had the same issue (the helper used 2.0, the Capsule schema fallback is 1.0), so `_get_axial_visual_dimensions` now takes per-type `default_radius`/`default_height` passed at each call site: capsule (0.5, 1.0), cylinder/cone (1.0, 2.0). `test_axial_visual_default_dims_match_collision` covers unauthored dimensions for all three prim types against the schema fallbacks.

adenzler-nvidia:
Addressed the CodeRabbit feedback in a21d3cea3: per-type schema-fallback defaults in `_get_axial_visual_dimensions` (including the capsule height default, which had the same issue as the cylinder/cone radius), a docstring and return annotation on the helper, and a regression test for unauthored `radius`/`height` on all three prim types.

eric-heiden (dismissed):
Thanks, I will reapprove when you fix the merge conflicts.

eric-heiden:
LGTM! Just a small suggestion while you are resolving the merge conflict.

newton/_src/utils/import_usd.py
```diff
@@ -1162,8 +1172,9 @@ def _load_visual_shapes_impl(
                 )
             elif type_name == "capsule":
                 axis = usd.get_gprim_axis(prim)
-                radius = usd.get_float(prim, "radius", 0.5) * scale[0]
-                half_height = usd.get_float(prim, "height", 2.0) / 2 * scale[1]
+                radius, half_height = _get_axial_visual_dimensions(
+                    prim, scale, axis, default_radius=0.5, default_height=1.0
+                )
                 # Apply axis rotation to transform
```
eric-heiden, line 1269: 🟠 The plane visual path in `newton/_src/utils/import_usd.py` still scales `width` with `scale[0]` and `length` with `scale[1]`, regardless of the authored `UsdGeomPlane.axis`. For X- and Y-axis planes, this selects the wrong perpendicular scale components and leaves the same axis-dependent sizing problem being fixed here for capsules, cylinders, and cones.

Please either make plane dimensions axis-aware, for example with a `_get_planar_visual_dimensions` helper, or track this explicitly as follow-up work if planes are intentionally outside this PR's scope.

adenzler-nvidia: Fixed in 72f98230b: added `_get_planar_visual_dimensions` that selects the scale components per the UsdGeomPlane spec (width aligns to Z for X-axis planes, length aligns to Z for Y-axis planes), with regression coverage across all three axes.

While fixing this I found the plane branch also dropped its axis rotation: since #996 the rotated transform was computed but the stale pre-rotation `plane_xform` was passed to `add_shape_plane`, so X- and Y-axis plane visuals were imported unrotated. The same commit restores the rotation and the test asserts the imported plane normal.

adenzler-nvidia:
Addressed the review in 72f98230b: plane visual width/length now pick the scale components per the `UsdGeomPlane` axis mapping via a `_get_planar_visual_dimensions` helper, with regression coverage across all three axes. While fixing this I found the plane branch had also silently dropped its axis rotation since #996 (X- and Y-axis plane visuals were imported unrotated); the same commit restores it. Details in the inline thread.

Also merged upstream/main (3eb93f578), so the branch is conflict-free again.

eric-heiden (approved):
Thanks!

PR #3369, opened 2026-07-06
Compute averaged contact position in SensorContact

### LINKED ISSUE #1489: Contact Sensor - Add contact position to contact sensor
To enable feature parity with PhysX the contact sensor in newton should expose the averaged contact position.

camevor:
## Description

Add `SensorContact.contact_position_matrix`, a per-counterpart matrix of world-frame contact positions. Each entry is the force-weighted average of contact midpoints (each contact's midpoint between its two surface points, weighted by linear contact force magnitude). Entries with no contributing contacts are zero. The matrix is `None` when no counterparts are specified and is only refreshed when `update()` receives a state with body transforms.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_sensor_contact
```

New `test_contact_position_matrix` verifies weighted-average positions and NaN entries for non-contacting pairs.

## New feature / API change

```python
import newton

sensor = newton.sensors.SensorContact(
    model,
    sensing_bodies=["hand"],
    counterpart_bodies=["ball", "table"],
)
sensor.update(state, contacts)

# World-frame average contact position per (sensing, counterpart) pair;
# NaN where the pair has no contacts contributing force.
positions = sensor.contact_position_matrix.numpy()
```

### Example

Example usage for contact position: PD controller balancing a sphere using a contact sensor and reading the position.

[SphereBalance.webm](https://github.com/user-attachments/assets/e7629faa-3ef0-4925-b94a-b0cb3eba568b)

## Summary by CodeRabbit

* **New Features**
  * `SensorContact` now optionally provides `position_matrix`: per-counterpart, force-weighted world-frame contact midpoints alongside existing contact force outputs.

* **Documentation**
  * Updated sensor concepts and changelog to reflect the new `position_matrix` behavior, including friction decomposition and conditional per-counterpart outputs.

* **Bug Fixes**
  * `position_matrix` is reset to **0.0** when pose/state isn’t provided, while force outputs continue to update appropriately.

* **Tests**
  * Expanded and adjusted `SensorContact` tests to validate allocation, deterministic computations, reset behavior, and zero-force cases.


newton/_src/sensors/sensor_contact.py
```diff
@@ -647,25 +712,39 @@ def update(self, state: State | None, contacts: Contacts):
...
             self.total_force_friction.zero_()
         if self.force_matrix is not None:
             self.force_matrix.zero_()
             self.force_matrix_friction.zero_()
+        update_contact_positions = (
+            self.contact_position_matrix is not None and state is not None and state.body_q is not None
+        )
+        if update_contact_positions:
+            self.contact_position_matrix.zero_()
+            self._contact_position_weight.zero_()
```
eric-heiden: 🟠 `force_matrix` is reset on every update, but `contact_position_matrix` and `_contact_position_weight` are reset only when `state.body_q` is available. After `update(state, contacts_a)` followed by `update(None, contacts_b)`, a cell can therefore contain a force from `contacts_b` paired with a position from `contacts_a`.

Please reset both position buffers whenever the force matrices are reset, even when positions cannot be recomputed. The `state=None` test should then expect zero positions instead of preserving the previous values.

newton/_src/sensors/sensor_contact.py
```diff
@@ -332,6 +381,15 @@ class SensorContact:
     dtype :class:`vec3`. Entry ``[i, j]`` is the friction force on sensing object ``i`` from counterpart
     ``counterpart_indices[i][j]``, in world frame. ``None`` when no counterparts are specified."""
 
+    contact_position_matrix: wp.array2d[wp.vec3] | None
```
eric-heiden: Consider renaming this to `position_matrix` because we use names like `force_matrix` etc. and it is obviously related to contacts since the class is a contact sensor.

I also don't really understand the motivation for this PR - what do you need the weighted contact positions for? Maybe this could also be explained in `sensors.rst` a bit more.

newton/_src/sensors/sensor_contact.py
camevor: Thanks, fixed in 9454a2fd2

newton/_src/sensors/sensor_contact.py
camevor: Agreed, renamed to `position_matrix`.

Regarding the motivation for the PR: This was an ask from IsaacLab (#1489) that I accidentally closed prematurely. I've expanded the sensor documentation in the docstring and added a simple showcase to the PR description.

eric-heiden (approved):
Thanks, LGTM!

Issue #2361, opened 2026-04-08
[BUG] Contact in nut_bolt_hydro example may lead to explosion


Kenny-Vilella:
### Bug Description

When doing QA testing for the 1.1 release, I found that the `nut_bolt_hydro` example has some stability issue when picking the nut, picking the bolt is fine though.
I could reproduce with main branch and 1.1 rc version, for two different configurations.

https://github.com/user-attachments/assets/0dec52dd-5e36-4391-877e-e765e04322e4

### Reproduction Script

```python
python3.13 -m venv .venv
source .venv/bin/activate
python -m pip install --pre newton[dev,torch-cu12]==v1.1.0rc1
python -m newton.examples nut_bolt_hydro
```

Kenny-Vilella:
Not critical to have for 1.1 since it appears only when manipulating the objects with a certain way.

mzamoramora-nvidia:
## TL;DR

The explosion is caused by the **viewer picking system**, not the hydroelastic contact pipeline. The picking spring-damper force enters a positive feedback loop on light objects (64g nut) bouncing off stiff contacts, diverging to infinity within ~20 frames.

Two fixes explored, both in `newton/_src/viewer/kernels.py`:

| | Minimal (force clamp) | Extended (velocity-tracking + inertia) |
|---|---|---|
| Prevents explosion | Yes | Yes |
| Picking feel in other examples | Unchanged | Sluggish |
| Nut vibration when held | Still present | Eliminated |
| Smooth thread following | No | Yes |
| Lines changed | ~4 | ~30 |
| API changes | None | Adds `body_inertia` to kernel |

**Recommendation:** Ship the minimal force clamp (4 lines, no API changes, no regressions). Details and code for both solutions below.

With the extended solution, we could get a very nice behavior for this demo. See video below. But that change makes picking harder for other examples.

https://github.com/user-attachments/assets/f9e83cc8-90ca-4fd9-9570-e36118407d61
 


---

## Root cause

The explosion is **not caused by the hydroelastic contact pipeline**. Contact stiffness stays bounded throughout and the MuJoCo solver handles it correctly.

The root cause is a **positive feedback loop in the viewer's picking force**. The picking kernel applies a spring-damper force:

```
F = (10 + mass) * (stiffness * displacement - damping * velocity)
```

[...]

PR #3375, opened 2026-07-07
Viewer: Limit mouse-picking torque using body inertia

jumyungc:
## Description
  Stabilize mouse picking on low-inertia bodies by automatically limiting the applied torque using each body’s mass and rotational inertia.

## Summary

  - Preserve the existing articulation-mass-based linear force clamp.
  - Limit the inertia-weighted rotational response using `body_inv_inertia` in the body frame.
  - Preserve normal point-force picking when the torque is below the limit.
  - Remove the unreleased per-body linear-only picking APIs.
  - Add coverage for clamped, unclamped, anisotropic-inertia, rotated-body, and cable-picking behavior.

  This introduces no new public API.

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **New Features**
  * Mouse picking now supports a torque scale setting, giving finer control over how picked objects rotate.
* **Bug Fixes**
  * Mouse-picking torque is now limited by each body’s rotational inertia for more stable interaction.
  * Picking now validates acceleration settings and rejects invalid values.
* **Tests**
  * Updated viewer picking coverage to verify torque limiting and inertia-aware behavior across multiple cases.


CHANGELOG.md
```diff
@@ -8,7 +8,7 @@
...
 - Add three VBD contact examples — `vbd_rigid_rigid_contact`, `vbd_soft_rigid_contact`, and `vbd_soft_rigid_mix_contact` — demonstrating rigid-rigid, soft (particle-rigid), and mixed cloth-bag contacts
 - Add viewer layer system to overlay multiple solvers/models in supported rendering viewers; call `ViewerBase.activate(layer_id)` to route subsequent `set_model` / `log_state` / `log_*` calls into a named layer, `ViewerBase.set_layer_visible()` to toggle layers independently, and `ViewerBase.set_layer_transform()` to position layers side-by-side. See `example_basic_multi_solver_overlay.py`
-- Add `viewer.set_picking_linear_only_bodies()` and `viewer.clear_picking_linear_only_bodies()` to mark bodies that should receive only the linear component of mouse-picking force, suppressing offset-induced torque.
+- Add `viewer.set_picking_torque_scale()` to scale the mouse-picking lever arm from the center of mass to the clicked point; cable examples use center-of-mass dragging to improve stability.
```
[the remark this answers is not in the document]
jumyungc: Skipped: set_picking_linear_only_bodies() and clear_picking_linear_only_bodies() were never included in a tagged release. This change replaces that unreleased API in place, and the existing Added entry now documents set_picking_torque_scale().

newton/_src/viewer/picking.py
```diff
@@ -114,54 +118,20 @@ def _apply_picking_force(self, state: newton.State) -> None:
...
                 self._pick_effective_mass,
-                self._linear_only_body_mask,
             ],
             device=self.model.device,
         )
 
-    def set_linear_only_bodies(self, body_ids: Iterable[int] | None) -> None:
-        """Configure bodies that should receive no torque from mouse picking.
+    def set_torque_scale(self, scale: float) -> None:
+        """Set the effective mouse-picking lever-arm scale."""
```
jcarius-nv: `AGENTS.md` requires Google-style docstrings. This setter accepts `scale` and raises `ValueError`, but documents neither. Please add `Args:` and `Raises:` sections, matching `ViewerBase.set_picking_torque_scale()`.

newton/tests/test_viewer_picking.py
```diff
@@ -280,22 +270,64 @@ def test_linear_only_bodies_remove_picking_torque(test: TestPickingSetup, device
...
+    half_torque_force = state.body_f.numpy()[0]
+    assert_np_equal(half_torque_force[:3], normal_force[:3], tol=1e-6)
+    assert_np_equal(half_torque_force[3:], 0.5 * normal_force[3:], tol=1e-6)
+
+    picking.set_torque_scale(0.0)
+    state.body_f.zero_()
+    picking._apply_picking_force(state)
+    force_only = state.body_f.numpy()[0]
+    assert_np_equal(force_only[:3], normal_force[:3], tol=1e-6)
+    assert_np_equal(force_only[3:], np.zeros(3), tol=1e-9)
```
jcarius-nv: Non-blocking: the PR contract also says that the default scale `1.0` preserves the previous point-force behavior and that the scaled target/damping point remains physically consistent. This test uses an unrotated, stationary body with `pick_damping=0`. Could we add regression coverage comparing scale `1.0` with the previous formulation and exercising rotation/damping at an intermediate scale? That would guard the new physics path beyond the torque endpoints.

newton/_src/viewer/picking.py
jumyungc: Done.

newton/tests/test_viewer_picking.py
jumyungc: Done

jcarius-nv (dismissed):

adenzler-nvidia:
Not blocking — this is a clear improvement over the per-body mask, and the blended lever formulation (spring, damping, and torque all from the same effective attachment, zero initial error) is nicely consistent.

Wanted to leave a note on a possible follow-up that could make the opt-out unnecessary. The root cause of the cable instability is that the picking force clamp scales with *articulation mass* while the resulting torque lands on the *segment's own inertia*. In `example_cable_pile` the clamp allows ~44 N (5g × ~0.9 kg cable), but a segment's axial inertia is ~1.6e-6 kg·m², so the unclamped angular acceleration reaches ~7e5 rad/s² — over 1000 rad/s of spin added per substep. The linear part is protected by the effective-mass clamp (#2361); the angular part has no equivalent.

MuJoCo's mouse perturb solves this without any per-scene configuration (`mjv_applyPerturbForce` in `engine_vis_interact.c`): the spring gain is scaled by the operational-space effective mass at the grab point (`pert->localmass`, from J M⁻¹ Jᵀ at click time), the torsional damping by a rotational effective inertia precomputed at model compile (`body_invweight0`), and damping is derived as critical (`sqrt(stiffness)`) rather than exposed. Because gains are *proportional* to effective mass/inertia, the perturb acceleration is the same for a cable segment and a 100 kg box — bounded by construction, with only two global constants (`map.stiffness`/`map.stiffnessrot`) that nobody tunes per model.

The Newton equivalent is small: precompute a per-body rotational effective inertia (min eigenvalue of `body_inertia`, rotation-invariant so a one-time compute) in `Picking.__init__` next to the existing `_pick_effective_mass`, and either scale the picking gains MuJoCo-style or, as a minimal delta, clamp the torque to `max_angular_acceleration * I_min` mirroring the existing force clamp. With that in place the cable examples could drop `set_picking_torque_scale(0.0)` entirely, and this knob would revert to a pure UX preference ("drag without rotating") — or could even be removed again while it's still unreleased. Happy to file an issue / prototype this as a follow-up.

jcarius-nv:
Removing from merge queue -- maybe @jumyungc could try this out for the cables and then decide it if makes sense to merge this as-is now (we can cherry-pick to 1.4 release) or instead adopt @adenzler-nvidia's proposed solution.

eric-heiden:
The global torque scale is a useful addition, I tested the cable examples and the picking works pretty well now.

Two compatibility and visualization issues need attention: the existing per-body API is removed without a migration path, and the picking line no longer shows the spring path actually used for force application.

newton/_src/viewer/viewer.py
```diff
@@ -626,25 +626,24 @@ def set_model(self, model: newton.Model | None):
...
-        if picking is not None:
-            picking.set_linear_only_bodies(body_ids)
+        scale = float(scale)
+        if not 0.0 <= scale <= 1.0:
+            raise ValueError("Picking torque scale must be in [0, 1].")
 
-    def clear_picking_linear_only_bodies(self) -> None:
-        """Restore normal mouse picking torque for all bodies."""
         picking = getattr(self, "picking", None)
         if picking is not None:
```
eric-heiden, line 646: 🟠 Removing `set_picking_linear_only_bodies()` and `clear_picking_linear_only_bodies()` breaks existing `ViewerBase` callers with `AttributeError`. The new global scale is not a compatible replacement because it cannot suppress torque for selected body IDs only. Please restore the old methods for a deprecation window, preserving their per-body behavior, or provide a scoped replacement before removing these public names.

newton/_src/viewer/kernels.py
```diff
@@ -83,53 +83,47 @@ def apply_picking_force_kernel(
...
-    vel_at_offset = vel_com + wp.cross(angular_vel, offset)
+    # pick_scale blends where the pull acts: 0 => through the COM (force only,
+    # no offset torque), 1 => at the clicked point (full point force + torque).
+    # Shift the target the same way so the spring starts at zero error for any
+    # scale (no sudden pull on click).
+    offset = pick_pos_world - com_world
+    effective_lever = pick_scale * offset
+    pick_pos_effective = com_world + effective_lever
+    pick_target_effective = pick_target_world - (1.0 - pick_scale) * pick_state[0].initial_pick_offset_world
+    pick_vel_effective = velocity_at_point(body_qd[pick_body], effective_lever)
```
eric-heiden: 🟠 The force now uses `pick_pos_effective` and `pick_target_effective`, but `picked_point_world` and `picking_target_world` still expose the raw clicked point and cursor target to the GL and RTX renderers. With a torque scale below `1`, the rendered picking line therefore shows a different spring from the one applying force, and at scale `0` it can remain stretched even when the COM spring has zero error. Please store or render the effective endpoints, or expose separate visual and force endpoints so the line accurately represents the applied spring.

eric-heiden:
> MuJoCo's mouse perturb solves this without any per-scene configuration (`mjv_applyPerturbForce` in `engine_vis_interact.c`): the spring gain is scaled by the operational-space effective mass at the grab point (`pert->localmass`, from J M⁻¹ Jᵀ at click time), the torsional damping by a rotational effective inertia precomputed at model compile (`body_invweight0`), and damping is derived as critical (`sqrt(stiffness)`) rather than exposed. Because gains are _proportional_ to effective mass/inertia, the perturb acceleration is the same for a cable segment and a 100 kg box — bounded by construction, with only two global constants (`map.stiffness`/`map.stiffnessrot`) that nobody tunes per model.

I think that would be nice to have, it's also in the spirit of @mmacklin's earlier request to have fewer knobs for the user to tune - picking should work out of the box as much as possible. Maybe this can be implemented in this PR before we have another iterative PR that again changes API? I don't think we need to rush this in for 1.4 anyway.

newton/_src/viewer/viewer.py
jumyungc: These methods were not shipped in a release, so I thought we'd better remove them, or please correct me if I am wrong.

jumyungc:
Thank you for the thoughtful reviews and suggestions. The previous PR introduced some methods that seem to be unnecessary now. I’d like to clean them up before they ship (assuming it is fine to remove unreleased methods).

I can also implement the suggested idea today. If we'd prefer not to rush it, I can limit this update to removing the unnecessary APIs and submit the stabilization separately afterward.

jumyungc:
@jcarius-nv @adenzler-nvidia @eric-heiden 

So two reasonable options I see are:
1. Update this PR(3375) following Alain’s suggestion. I already implemented it in this PR and removed the unnecessary mouse-picking APIs. The VBD cable examples seem to work fine.

2. Simply revert the unreleased mouse picking APIs introduced in #2670 and restores the previous API and behavior. I can then implement Alain and Eric’s suggestions separately in a focused follow-up PR. Here is the simple PR that revert the unreleased APIs:  https://github.com/newton-physics/newton/pull/3407

I’m happy to proceed with either option. Alain's suggestion is already implemented in this PR. If option 1 needs more review or does not look viable yet, option 2 provides a small, clean checkpoint. In either case, avoiding unnecessary public methods seems preferable. Thank you.

newton/_src/viewer/kernels.py
jumyungc: I removed the torque-scale/effective-endpoint formulation entirely now.

jumyungc:
> The global torque scale is a useful addition, I tested the cable examples and the picking works pretty well now.
> 
> Two compatibility and visualization issues need attention: the existing per-body API is removed without a migration path, and the picking line no longer shows the spring path actually used for force application.

@eric-heiden I somehow missed this comment. agents.md and compatibility.rst sounded like no migration path is needed in this case since the APIs were not released, but please correct me anyone; I can update the PR(s).

adenzler-nvidia (approved):
The inertia-based torque limit is the right approach and the implementation checks out: the bound is evaluated in the body frame consistently with how the integrator applies `body_inv_inertia`, the linear force clamp is unchanged, and the new tests pin both the clamped and unclamped paths (the anisotropic-rotation test in particular would catch scaling mistakes). Also tested `cable_pile` interactively — cable dragging is stable and feels equivalent to the removed per-body workaround.

One thing to address before merge: the PR title and description still describe the removed `set_picking_torque_scale()` API. The squash commit takes its message from these, so please rewrite them to describe the current change — picking torque limited by body inertia, no new API.

jumyungc:
> The inertia-based torque limit is the right approach and the implementation checks out: the bound is evaluated in the body frame consistently with how the integrator applies `body_inv_inertia`, the linear force clamp is unchanged, and the new tests pin both the clamped and unclamped paths (the anisotropic-rotation test in particular would catch scaling mistakes). Also tested `cable_pile` interactively — cable dragging is stable and feels equivalent to the removed per-body workaround.
> 
> One thing to address before merge: the PR title and description still describe the removed `set_picking_torque_scale()` API. The squash commit takes its message from these, so please rewrite them to describe the current change — picking torque limited by body inertia, no new API.

Thanks @adenzler-nvidia for the quick review! Applied.

PR #3376, opened 2026-07-07
Fix batched static MuJoCo geometry poses

maxkra15:
## Description

  Refresh MJWarp-derived poses for geometry attached directly to the world body after Newton applies per-world
  model transforms.

  Previously, finite static world-body colliders retained the first world’s derived pose in translated or rotated
  worlds. This could cause dynamic bodies in cloned environments to miss their local colliders and fall through.

  The fix updates `geom_xpos` and `geom_xmat` for static, non-mocap geometry after relevant model properties are
  synchronized. The regression specifically covers geometry attached directly to the world body (`body=-1`). No
  public API changes or new dependencies are introduced.

  ## Checklist

  No documentation changes are required because this fixes internal derived-state synchronization without changing
  the public API.

  ## Test plan

  The regression test was run with and without the fix:

  - Without the fix, it fails because the second world’s derived static geometry position remains at `x=0` instead
  of using its `x=4` world offset.
  - With the fix, it passes and both dynamic boxes remain supported by their respective static colliders.

  ```bash
    -k test_static_worldbody_geoms_support_offset_worlds

  uvx pre-commit run -a
```
  ## Bug fix

  Steps to reproduce:

  1. Create a Newton world containing a finite static collider attached directly to the world body.
  2. Clone it into two MJWarp worlds.
  3. Apply a translation and rotation to the second world.
  4. Initialize SolverMuJoCo with separate_worlds=True.
  5. Observe that, without the fix, the second collider’s geom_xpos remains at the first world’s position.
  6. Simulate dynamic boxes above both colliders and observe the second box falling through.

  Minimal reproduction:
```bash
  import numpy as np
  import warp as wp

  import newton


  world = newton.ModelBuilder()
  SolverMuJoCo.register_custom_attributes(world)

  world.add_shape_box(
      body=-1,
      xform=wp.transform(wp.vec3(0.0, 0.0, -0.1), wp.quat_identity()),
      hx=1.0,
      hy=1.0,
      hz=0.1,
  )

  body = world.add_link(
      xform=wp.transform(wp.vec3(0.0, 0.0, 0.1), wp.quat_identity()),
      mass=1.0,
      inertia=wp.mat33(np.eye(3)),
  )
  joint = world.add_joint_free(child=body)
  world.add_articulation([joint])
  world.add_shape_box(body=body, hx=0.1, hy=0.1, hz=0.1)

  builder.add_world(world)
  builder.add_world(
      world,
      xform=wp.transform(
          wp.vec3(4.0, 0.0, 0.0),
          wp.quat_from_axis_angle(wp.vec3(0.0, 0.0, 1.0), 0.5),
      ),
  )

  model = builder.finalize()
  solver = SolverMuJoCo(
      model,
      separate_worlds=True,
      use_mujoco_contacts=True,
  )

  print("Configured positions:")
  print(solver.mjw_model.geom_pos.numpy()[:, 0])

  print("Derived positions:")
  print(solver.mjw_data.geom_xpos.numpy()[:, 0])
```
  Before the fix, the second configured position has x=4, while its derived position incorrectly has x=0. After the
  fix, the configured and derived positions agree.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed collision handling for static world-body geometry in offset multi-world simulations.
  * Improved consistency of geometry pose data after shape-property updates so each world uses its own correct transforms.
  * Addressed a regression where supported bodies could behave incorrectly in replicated worlds with offsets/rotations.

* **Tests**
  * Added a regression test covering offset worlds with static world-body geometry to verify correct contact behavior over time.

### DIFF
--- a/newton/_src/solvers/mujoco/kernels.py
+++ b/newton/_src/solvers/mujoco/kernels.py
@@ -2496,6 +2496,31 @@ def update_geom_properties_kernel(
 
 
 @wp.kernel
+def sync_static_geom_xposes_kernel(
+    body_rootid: wp.array[int],
+    body_weldid: wp.array[int],
+    body_mocapid: wp.array[int],
+    geom_bodyid: wp.array[int],
+    geom_pos: wp.array2d[wp.vec3],
+    geom_quat: wp.array2d[wp.quat],
+    xpos: wp.array2d[wp.vec3],
+    xquat: wp.array2d[wp.quat],
+    geom_xpos: wp.array2d[wp.vec3],
+    geom_xmat: wp.array2d[wp.mat33],
+):
+    """Refresh per-world poses for geoms welded to the world body."""
+    world, geom = wp.tid()
+    body = geom_bodyid[geom]
+    if body_weldid[body] != 0 or body_mocapid[body_rootid[body]] != -1:
+        return
+
+    body_q = quat_wxyz_to_xyzw(xquat[world, body])
+    geom_q = quat_wxyz_to_xyzw(geom_quat[world, geom])
+    geom_xpos[world, geom] = xpos[world, body] + wp.quat_rotate(body_q, geom_pos[world, geom])
+    geom_xmat[world, geom] = wp.quat_to_matrix(body_q * geom_q)
+
+
+@wp.kernel
 def update_jnt_solref_from_invweight0_kernel(
     mjc_jnt_to_newton_dof: wp.array2d[wp.int32],
     joint_limit_ke: wp.array[float],
--- a/newton/_src/solvers/mujoco/kernels.py
+++ b/newton/_src/solvers/mujoco/kernels.py
@@ -2496,28 +2496,21 @@ def update_geom_properties_kernel(
 
 
 @wp.kernel
-def sync_static_geom_xposes_kernel(
-    body_rootid: wp.array[int],
-    body_weldid: wp.array[int],
-    body_mocapid: wp.array[int],
+def sync_worldbody_geom_xposes_kernel(
     geom_bodyid: wp.array[int],
     geom_pos: wp.array2d[wp.vec3],
     geom_quat: wp.array2d[wp.quat],
-    xpos: wp.array2d[wp.vec3],
-    xquat: wp.array2d[wp.quat],
     geom_xpos: wp.array2d[wp.vec3],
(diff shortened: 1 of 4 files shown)


newton/_src/solvers/mujoco/kernels.py
```diff
@@ -2495,6 +2495,31 @@ def update_geom_properties_kernel(
...
+):
+    """Refresh per-world poses for geoms welded to the world body."""
+    world, geom = wp.tid()
+    body = geom_bodyid[geom]
+    if body_weldid[body] != 0 or body_mocapid[body_rootid[body]] != -1:
+        return
+
+    body_q = quat_wxyz_to_xyzw(xquat[world, body])
+    geom_q = quat_wxyz_to_xyzw(geom_quat[world, geom])
+    geom_xpos[world, geom] = xpos[world, body] + wp.quat_rotate(body_q, geom_pos[world, geom])
```
eric-heiden: 🟠 Please scope this helper to direct worldbody geoms (`geom_bodyid[geom] == 0`) and copy/convert their per-world `geom_pos`/`geom_quat` directly, instead of composing through `xpos`/`xquat`. Those are body data and can be stale when this helper runs from `notify_model_changed(JOINT_PROPERTIES)`, so the current broader world-welded-body path can write an incorrect geom pose. For the worldbody case fixed by this PR, the body transform is identity and `JOINT_PROPERTIES` is unrelated; call this only for `SHAPE_PROPERTIES`. If support for fixed-root body geoms is intended, it needs a separate path that refreshes or derives their body pose first.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -4072,6 +4073,9 @@ def notify_model_changed(self, flags: ModelFlags | int) -> None:
                         update_connect_constraint_anchors,
                     )
 
+            if flags & (ModelFlags.JOINT_PROPERTIES | ModelFlags.SHAPE_PROPERTIES):
```
eric-heiden: This should be limited to `ModelFlags.SHAPE_PROPERTIES` to not have this overhead when only joint properties are modified. See my previous comment, we should really limit this conversion to static geometry, not geometry attached to fixed bodies or mocap bodies (could be handled separately).

eric-heiden (approved):
Thanks!

PR #3379, opened 2026-07-07
Use actuation type for implicit PD control in Kamino

chschuma-disney:
## Description

The implicit joint dynamics in Kamino have so far been all-or-nothing: If they are active, every part of the formulation (armature, damping, PD control, feed-forward force) will contribute to the dynamics. This explicitly disregards the joint's `JointActuationType`, which would specify how the joint should be actuated.

One situation where this creates an issue is when a model with non-zero PD gains and non-zero armature is imported/created, and explicit actuation is used to drive it. The explicit actuation would use the model's PD gains to generate a torque applied to the actuators. Assuming the actuation type in the model was properly set to `FORCE`, the model would currently still use its (non-zero) PD gains to add additional actuation on top of this torque. The only way to avoid this currently is to set all PD gains to zero.

I'm opening this as a draft, since the fix would be easy, but requires some thought and discussion to make sure it follows the requirements and doesn't break any existing code.

## Checklist

## Test plan

 ```
python -m unittest newton/_src/solvers/kamino/tests/test_kinematics_joints.py
```

## Bug fix

**Steps to reproduce:**

1. Load model with non-zero PD gains.
2. Set actuation type of all joints to `FORCE`.
3. Simulate model. It should still resist moving away from its rest pose, since PD control is still active.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved implicit joint dynamics to account for per-joint actuation types, producing more accurate effective inertia and total applied torque/force.
  * Ensured numerical stability by clamping the computed effective mass term before inversion.

* **Tests**
  * Expanded and updated dynamic joint tests to validate actuation-type-specific expectations, including effective mass/inverse mass and velocity terms.
  * Tightened numerical assertions using consistent tolerances for improved reliability.


Guirec-Maloisel:
This looks very reasonable to me, did you have something in mind that this could break?

chschuma-disney:
> This looks very reasonable to me, did you have something in mind that this could break?

I think it could have hidden some suboptimal simulation setups that people might have used. For example, not properly setting the actuation type to `POSITION_VELOCITY` when using implicit PD control, which would have worked until now but no longer will.

Guirec-Maloisel:
> > This looks very reasonable to me, did you have something in mind that this could break?
> 
> I think it could have hidden some suboptimal simulation setups that people might have used. For example, not properly setting the actuation type to `POSITION_VELOCITY` when using implicit PD control, which would have worked until now but no longer will.

Makes sense, I'd suggest taking a pass through the code base to fix this if needed (e.g. in examples), and adding warnings in builder.py/conversions.py if we detect non-zero gains that will not get used, due to the actuator type. That won't catch the case where these values are updated to non-zero later but should hopefully catch most downstream misuse already.

chschuma-disney:
> > > This looks very reasonable to me, did you have something in mind that this could break?
> > 
> > 
> > I think it could have hidden some suboptimal simulation setups that people might have used. For example, not properly setting the actuation type to `POSITION_VELOCITY` when using implicit PD control, which would have worked until now but no longer will.
> 
> Makes sense, I'd suggest taking a pass through the code base to fix this if needed (e.g. in examples), and adding warnings in builder.py/conversions.py if we detect non-zero gains that will not get used, due to the actuator type. That won't catch the case where these values are updated to non-zero later but should hopefully catch most downstream misuse already.

I did a pass on the examples, and everything seems to still work.
I don't think adding the warning would be warranted, since that would still be a valid use case: you might want to import a model with PD gains, but then change it to being force-actuated and supply the actuation force using external PD control.

Guirec-Maloisel:
> > > > This looks very reasonable to me, did you have something in mind that this could break?
> > > 
> > > 
> > > I think it could have hidden some suboptimal simulation setups that people might have used. For example, not properly setting the actuation type to `POSITION_VELOCITY` when using implicit PD control, which would have worked until now but no longer will.
> > 
> > 
> > Makes sense, I'd suggest taking a pass through the code base to fix this if needed (e.g. in examples), and adding warnings in builder.py/conversions.py if we detect non-zero gains that will not get used, due to the actuator type. That won't catch the case where these values are updated to non-zero later but should hopefully catch most downstream misuse already.
> 
> I did a pass on the examples, and everything seems to still work. I don't think adding the warning would be warranted, since that would still be a valid use case: you might want to import a model with PD gains, but then change it to being force-actuated and supply the actuation force using external PD control.

Hmm for sure but that's why it's only a warning and not an error. I'd think that it's still worth it to inform the user that some of the data that they imported ends up not being used (and if it's on purpose then they are free to ignore that warning).

Guirec-Maloisel (approved):
Looks good to me!

PR #3382, opened 2026-07-07
[Kamino] Add support for configurable per-shape material property mixing modes

vastsoun:
## Description

This PR introduces configurable per-shape material property mixing when using generating contacts using the external CollisionPipeline. It does so by:
1. Adding back-end support within the corresponding conversion operation/kernel
2. Exposing high-level user-facing string literals to support the min/max/multiply/average operations on contact friction and restitution.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added configurable material mixing modes for friction and restitution (average, multiply, max, min), with friction defaulting to average and restitution to min.
  * Newton-to-Kamino contact conversion now uses the selected mixing modes when deriving friction and restitution.
  * Exposed these settings through solver/model configuration inputs.
* **Bug Fixes**
  * Made mixing behavior more explicit and consistent, including the new multiply option.
* **Tests**
  * Updated test terminology to match the new mixing-mode naming.

### DIFF
--- a/newton/_src/solvers/kamino/_src/core/materials.py
+++ b/newton/_src/solvers/kamino/_src/core/materials.py
@@ -22,6 +22,7 @@ material properties that can be queried at simulation runtime. It includes:
 
 from __future__ import annotations
 
+from collections.abc import Callable
 from dataclasses import dataclass
 from enum import IntEnum
 
@@ -46,6 +47,7 @@ __all__ = [
     "MaterialPairProperties",
     "MaterialPairsModel",
     "MaterialsModel",
+    "make_get_mixed_material_pair_property",
 ]
 
 ###
@@ -75,7 +77,7 @@ Equals ``0.7``.
 ###
 
 
-class MaterialMuxMode(IntEnum):
+class MaterialMixMode(IntEnum):
     """
     An enumeration defining the heuristic modes for deriving
     pairwise material properties from individual materials.
@@ -87,11 +89,32 @@ class MaterialMuxMode(IntEnum):
     AVERAGE = 0
     """Pairwise property is the average of the two material properties."""
 
-    MAX = 1
+    MULTIPLY = 1
+    """Pairwise property is the product of the two material properties."""
+
+    MAX = 2
     """Pairwise property is the maximum of the two material properties."""
 
-    MIN = 2
+    MIN = 3
     """Pairwise property is the minimum of the two material properties."""
+
+    @classmethod
+    def from_string(cls, s: str) -> MaterialMixMode:
+        """Converts a string to a MaterialMixMode enum value."""
+        try:
+            return cls[s.upper()]
+        except KeyError as e:
+            raise ValueError(f"Invalid MaterialMixMode: {s}. Valid options are: {[e.name for e in cls]}") from e
+
+    @override
+    def __str__(self):
+        """Returns a string representation of the MaterialMixMode."""
+        return f"MaterialMixMode.{self.name} ({self.value})"
+
+    @override
+    def __repr__(self):
+        """Returns a string representation of the MaterialMixMode."""
+        return self.__str__()
 
 
 @dataclass
@@ -339,6 +362,24 @@ def material_average(
 
 
 @wp.func
+def material_multiply(
+    value1: wp.float32,
+    value2: wp.float32,
+) -> wp.float32:
+    """
+    Computes the product of two material property values.
+
+    Args:
+        value1: The first material property value.
+        value2: The second material property value.
+
+    Returns:
+        The product of the two material property values.
+    """
+    return value1 * value2
+
+
+@wp.func
 def material_max(
     value1: wp.float32,
     value2: wp.float32,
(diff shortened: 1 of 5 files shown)


newton/_src/solvers/kamino/solver_kamino.py
```diff
@@ -808,6 +821,8 @@ def step(self, state_in: State, state_out: State, control: Control | None, conta
                 contacts_in=contacts,
                 contacts_out=self._contacts_kamino,
                 convert_forces=False,
+                friction_mix_mode=self._config.materials.friction_mix_mode,
+                restitution_mix_mode=self._config.materials.restitution_mix_mode,
```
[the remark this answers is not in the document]
vastsoun: Fixed in latest commit.

newton/_src/solvers/kamino/config.py
```diff
@@ -948,3 +948,71 @@ def validate(self) -> None:
...
+    The mixing mode to use for friction.\n
+    See :class:`MaterialMixMode` for available options.\n
+    Defaults to `MaterialMixMode.AVERAGE`.
+    """
+
+    restitution_mix_mode: Literal["average", "multiply", "max", "min"] = "min"
+    """
+    The mixing mode to use for restitution.\n
+    See :class:`MaterialMixMode` for available options.\n
+    Defaults to `MaterialMixMode.AVERAGE`.
```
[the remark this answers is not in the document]
vastsoun: Fixed in latest commit.

newton/_src/solvers/kamino/_src/geometry/contacts.py
```diff
@@ -848,199 +850,222 @@ def _assert_has_data(self):
...
-    into "force on B".
+    Generates a kernel to Newton contacts to the Kamino format.
+
+    Args:
+        friction_mixing_mode: The mixing mode to use for friction.
+        restitution_mixing_mode: The mixing mode to use for restitution.
+
+    Returns:
+        A kernel function that converts Newton contacts to the Kamino format.
     """
```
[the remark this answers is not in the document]
vastsoun: Fixed in latest commit.

newton/_src/solvers/kamino/_src/core/materials.py
```diff
@@ -374,27 +415,78 @@ def material_min(
     return wp.min(value1, value2)
 
 
-def make_get_material_pair_properties(muxmode: MaterialMuxMode = MaterialMuxMode.MAX):
+def get_material_mixing_function(
+    muxmode: MaterialMixMode = MaterialMixMode.AVERAGE,
```
Guirec-Maloisel: `muxmode` should probably be `mixmode` with the renamings that happened in this file (applies to a few places in this file)

newton/_src/solvers/kamino/_src/geometry/contacts.py
```diff
@@ -848,199 +851,222 @@ def _assert_has_data(self):
...
-    # Retrieve the contact index for this thread
-    cid = wp.tid()
 
-    # Skip conversion if this contact index exceeds the number
-    # of contacts to convert.
-    num_active = newton_count[0]
-    if cid >= num_active:
-        return
+    @wp.kernel
```
Guirec-Maloisel: We should probably use `@wp.kernel(module="unique", enable_backward=False)` here, so this kernel is moved to a separate compilation unit as the rest of this module (otherwise the other kernels in this file will get recompiled as well when this factory is called with different arguments).

newton/_src/solvers/kamino/_src/geometry/contacts.py
```diff
@@ -848,199 +851,222 @@ def _assert_has_data(self):
...
-    the no-swap case, B -> A in the swap case, which is negated to restore the
-    Kamino A->B convention).
-
-    Optionally also converts Newton's :attr:`Contacts.force` (the wrench on body0
-    by body1 at the CoM of body0, in world) into Kamino's ``reaction`` (the linear
-    force on body B by body A in the local contact frame). The linear part is
-    invariant to reference-point shifts, so this is a pure rotation into the
-    contact frame, with a sign flip in the no-swap case to convert "force on A"
-    into "force on B".
+    Generates a kernel to Newton contacts to the Kamino format.
```
Guirec-Maloisel: Missing a word here (probably "to convert Newton contacts to the Kamino format")

newton/_src/solvers/kamino/config.py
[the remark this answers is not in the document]
Guirec-Maloisel: I actually don't see the fix, I still see the default as "min" while the docstring says "Defaults to `MaterialMixMode.AVERAGE`."

newton/_src/solvers/kamino/_src/geometry/contacts.py
```diff
@@ -848,199 +851,222 @@ def _assert_has_data(self):
...
-    kamino_position_B: wp.array[wp.vec3f],
-    kamino_gapfunc: wp.array[wp.vec4f],
-    kamino_frame: wp.array[wp.quatf],
-    kamino_material: wp.array[wp.vec2f],
-    kamino_margins: wp.array[wp.vec2f],
-    kamino_key: wp.array[wp.uint64],
-    kamino_reaction: wp.array[wp.vec3f],
-    kamino_remap: wp.array[wp.int32],
+@functools.cache
+def make_convert_contacts_newton_to_kamino(
```
Guirec-Maloisel, line 855: High-level comment: I see that we are bubbling up the case distinction over the mixing mode all the way up to the contact conversion function, which I guess is the best in terms of performance (since we compile kernels with the mixing type baked in). But in terms of flexibility that doesn't sound very future-proof, e.g. if we ever wanted a different mixing mode per shape. Is this something you took into account?

The version with an if directly in the low-level mixing function would allow for simpler code and more flexibility, and I wouldn't expect the impact to be noticeable.

newton/_src/solvers/kamino/_src/geometry/contacts.py
vastsoun: Good point. Will fix in upcoming commit.

newton/_src/solvers/kamino/_src/geometry/contacts.py
vastsoun: @Guirec-Maloisel That's a valid point, but currently there doesn't seem to be a need from the IsaacLab side to make this run-time configurable, so I would recommend we leave it as-is for now and modify this if needed in the future in follow-up PR.

newton/_src/solvers/kamino/config.py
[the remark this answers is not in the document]
vastsoun: Fixed in latest commit.

newton/_src/solvers/kamino/_src/geometry/contacts.py
Guirec-Maloisel: Given the need for this in a short timeline I think that makes sense, yes.

Guirec-Maloisel (approved):
Looks good to me now!

PR #3386, opened 2026-07-07
[Kamino] Fix runtime model changes of joint limits and body inertia.

rubengrandia:
## Description

Kamino was not correctly handling model changes for joint limits and body inertias.

Related issue https://github.com/vastsoun/newton/issues/382

## Checklist

## Test plan

Test added to verify the parameter update. This lives in the Kamino test folder for now because it uses details of the Kamino model directly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Joint-limit and rigid-body inertial data now stay synchronized with the model without requiring manual refreshes.
  * Updated model properties are reflected immediately and safely without mutating existing model data.
  * Removed unnecessary limit normalization during joint conversion.

* **Tests**
  * Added regression coverage for live joint-limit and inertial-property updates.
  * Verified model-change notifications remain safe and read-only.

### DIFF
--- a/newton/_src/solvers/kamino/_src/core/conversions.py
+++ b/newton/_src/solvers/kamino/_src/core/conversions.py
@@ -20,10 +20,6 @@ from .bodies import (
 from .builder import JointActuationType
 from .geometry import GeometriesModel
 from .joints import (
-    JOINT_DQMAX,
-    JOINT_QMAX,
-    JOINT_QMIN,
-    JOINT_TAUMAX,
     JointDoFType,
     JointsModel,
 )
@@ -42,7 +38,6 @@ if TYPE_CHECKING:
 __all__ = [
     "convert_geometries",
     "convert_joints",
-    "convert_model_joint_limits",
     "convert_model_joint_transforms",
     "convert_rigid_bodies",
     "convert_target_coords_to_target_dofs",
@@ -183,8 +178,6 @@ def joint_conversion_kernel(
     model_joint_target_kd: wp.array[wp.float32],
     joint_limit_lower: wp.array[wp.float32],
     joint_limit_upper: wp.array[wp.float32],
-    joint_velocity_limit: wp.array[wp.float32],
-    joint_effort_limit: wp.array[wp.float32],
     # Outputs:
     joint_jid: wp.array[wp.int32],
     joint_dof_type: wp.array[wp.int32],
@@ -248,17 +241,6 @@ def joint_conversion_kernel(
     if is_dynamic_j:
         joint_num_dynamic_cts[joint_id] = ndofs_j
     joint_num_cts[joint_id] = joint_num_dynamic_cts[joint_id] + joint_num_kinematic_cts[joint_id]
-
-    # Clip joint limits and effort/velocity limits to supported ranges
-    for i in range(qd_count_j):
-        joint_limit_lower[dofs_start_j + i] = wp.clamp(joint_limit_lower[dofs_start_j + i], JOINT_QMIN, JOINT_QMAX)
-        joint_limit_upper[dofs_start_j + i] = wp.clamp(joint_limit_upper[dofs_start_j + i], JOINT_QMIN, JOINT_QMAX)
-        joint_velocity_limit[dofs_start_j + i] = wp.clamp(
-            joint_velocity_limit[dofs_start_j + i], -JOINT_DQMAX, JOINT_DQMAX
-        )
-        joint_effort_limit[dofs_start_j + i] = wp.clamp(
-            joint_effort_limit[dofs_start_j + i], -JOINT_TAUMAX, JOINT_TAUMAX
-        )
 
 
 @wp.kernel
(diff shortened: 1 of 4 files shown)


newton/_src/solvers/kamino/tests/test_solver_kamino_notify.py
```diff
@@ -0,0 +1,164 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
+
+"""Regression tests for SolverKamino runtime model-property propagation.
```
Guirec-Maloisel: These tests are great, but I'd phrase them as functionality tests (that we should expand on as we take into account more of the flags in notify_model_changed()) rather than as regression tests, and reduce a bit the AI bloat.

More concretely:
- the comment at the top of the file could be juste one line, e.g. "Tests for SolverKamino runtime model-property progation"
- the tests for limits and mass properties could be kept mostly as-is, but without the comments about the implementation (just plainly testing the new values are taken into account, regardless of whether the implementation is trivial due to aliasing or not).

newton/_src/solvers/kamino/solver_kamino.py
```diff
@@ -1080,3 +1080,15 @@ def _update_joint_transforms(self):
         changed at runtime (e.g. animated root transforms).
         """
         self._kamino.convert_model_joint_transforms(self.model, self._model_kamino.joints)
+
+    def _update_joint_limits(self):
+        """
+        Re-copy Kamino joint DoF limits from Newton's joint limit arrays.
+
+        Called when :data:`~newton.ModelFlags.JOINT_DOF_PROPERTIES` is raised,
```
Guirec-Maloisel: I'd reduce this comment to its first line as well.

newton/_src/solvers/kamino/_src/core/conversions.py
```diff
@@ -696,6 +697,62 @@ def convert_model_joint_transforms(model: Model, joints: JointsModel) -> None:
...
+    q_j_max[i] = wp.clamp(joint_limit_upper[i], JOINT_QMIN, JOINT_QMAX)
+    dq_j_max[i] = wp.clamp(joint_velocity_limit[i], -JOINT_DQMAX, JOINT_DQMAX)
+    tau_j_max[i] = wp.clamp(joint_effort_limit[i], -JOINT_TAUMAX, JOINT_TAUMAX)
+
+
+def convert_model_joint_limits(model: Model, joints: JointsModel) -> None:
+    """
+    Copies Newton's joint DoF limits into Kamino's format.
+
+    Reads Newton's ``model.joint_limit_lower`` / ``joint_limit_upper`` /
```
Guirec-Maloisel: This comment could be compacted, e.g. "Copies Newton's joint DoF limits into Kamino's format, clamping them to Kamino's supported ranges." (+ Args description left as-is)

newton/_src/solvers/kamino/solver_kamino.py
```diff
@@ -1080,3 +1080,15 @@ def _update_joint_transforms(self):
         changed at runtime (e.g. animated root transforms).
         """
         self._kamino.convert_model_joint_transforms(self.model, self._model_kamino.joints)
+
+    def _update_joint_limits(self):
```
chschuma-disney: I think it might make sense to check whether a limit conversion is even meaningful and necessary. We're just clamping them to a specific set that is the same as Newton's for some limits, and different for others. Is there be any practical difference to just referencing the Newton arrays?

chschuma-disney (commented):
Thanks for looking into this. I think the overall idea looks good and the tests make sense, though I'm wondering if it makes sense to even keep the limit conversion, or just use the Newton data directly.

newton/_src/solvers/kamino/solver_kamino.py
vastsoun: I agree with @chschuma-disney, currently I see no need to copy/clone the source arrays any more, and we could just opt to alias the source ones in `newton.Model` and skip the additional clamping.

newton/_src/solvers/kamino/tests/test_solver_kamino_notify.py
rubengrandia: Done. Reduced the module docstring to a single line and reframed these as functionality tests (dropped the implementation/regression commentary). They now assert the Kamino arrays alias Newton's and that runtime edits are reflected; we can expand these as more notify_model_changed flags get handled.

newton/_src/solvers/kamino/solver_kamino.py
rubengrandia: Agreed, done. JointsModel now aliases Newton's joint_limit_lower/joint_limit_upper/joint_velocity_limit/joint_effort_limit directly (no clone, no clamp). The construction-time clamp loop, the convert_model_joint_limits copy kernel, and the _update_joint_limits notify handler are all removed, so runtime limit edits propagate automatically without any notify call.

rubengrandia:
@Guirec-Maloisel now that the joint-limit and body-inertia arrays are aliased directly from the Newton model, the propagation behavior is essentially trivial. Do you think we should drop the dedicated `newton/_src/solvers/kamino/tests/test_solver_kamino_notify.py` test file, or keep it as a functionality test to expand as more `notify_model_changed` flags get handled?

chschuma-disney (approved):
Looks good to me!

I would keep the tests, since we might expand those for other (non-trivial) attributes.

newton/_src/solvers/kamino/_src/core/conversions.py
```diff
@@ -1363,7 +1338,6 @@ def convert_geometries(
     sorted_excluded_pairs = model.shape_collision_filter_pairs_array()
     excluded_pairs = wp.array(sorted_excluded_pairs, dtype=wp.vec2i, device=model.device)
 
-    # Construct and return the converted geometries model
```
Guirec-Maloisel, line 1366: This unrelated comment seems to have been removed by the AI for some reason (it sometimes does that for me too)

Guirec-Maloisel (approved):
Looks good to me now, I like that we end up simplifying things and removing superfluous copies. I'd also keep the new unit tests as we'll likely implement more of the `notify_model_changed()` functionality in a near future.

newton/_src/solvers/kamino/_src/core/conversions.py
Guirec-Maloisel: It might not have been very useful in the first place though, so I'm open to leaving it removed if that avoids us a new round of CI.

PR #3387, opened 2026-07-08
Enable graph capture on CPU

c0d1f1ed:
## Description

Enable graph capture on CPU for Newton examples and tests that previously treated capture as CUDA-only.

This removes `device.is_cuda` / `wp.get_device().is_cuda` gates around `wp.ScopedCapture()` so CPU graph capture is exercised when capture is requested. An internal `is_graph_capture_allocation_enabled(device)` was added to `newton._src.utils` as the shared way to check whether allocation during graph capture is safe: CPU always returns `True`, while CUDA remains gated on mempool support.

`SolverVBD` now uses that helper for its capture-time resize guard, so CPU capture is not rejected as though it required CUDA mempool allocation. The example test matrix is also expanded so CPU graph capture coverage includes the MuJoCo VBD ADMM path.

CPU graph capture produces great benchmark results. Launch-heavy examples see the largest gains (`sensor_imu` 42x, `selection_cartpole` 16x, `diffsim_drone` 15x, `ik_custom` 12x, `basic_pendulum` 12x). Thirteen CPU-graph runs were faster than the same examples on CUDA on this hardware, which matches the expected benefit for compute-light workloads sensitive to latency.

## Checklist

## Test plan

```bash
uv run docs/generate_api.py
uv run --extra dev -m newton.tests -k _cpu
uvx pre-commit run -a

## Summary by CodeRabbit

## Summary

* **New Features**
  * Expanded graph capture support across many examples and test scenarios, including CPU where supported.
  * Graph replay is now used more consistently when a capture graph is available, reducing device-specific behavior.

* **Bug Fixes**
  * Improved graph-capture allocation behavior and validation, including clearer runtime errors when capture allocations aren’t enabled.
  * Some capture/step state handling was adjusted to match replay expectations.

* **Documentation**
  * Updated tutorials and troubleshooting guidance to consistently use “graph capture” terminology.

* **Tests**
  * Updated tests to use the new graph-capture enablement logic and flags.

### CODE REVIEW

**Comment by adenzler-nvidia** [comment]
The user-facing docs still teach the CUDA-only capture pattern this PR removes from the examples:

- `docs/concepts/collisions.rst` ("CUDA graph capture" section, ~line 1984): states "On CUDA devices, the simulation loop ... can be captured into a CUDA graph" and shows an `if wp.get_device().is_cuda:` code sample — the exact gate this PR deletes from the examples.
- `docs/tutorials/00_introduction.ipynb` ("GPU Acceleration with CUDA Graphs"): the code cell prints "Running on CPU (no CUDA graph)".
- `docs/tutorials/01_robotics.ipynb` ("3) Performance: CUDA Graph Capture"): the code cell prints "Running on CPU; graph capture skipped".

These are the snippets users copy, so they now contradict the new behavior. Please update them to the device-agnostic capture idiom used by the converted examples (the PR checklist marks documentation as up to date).

**Comment by adenzler-nvidia** [comment]
Could we avoid adding `newton.utils.is_graph_capture_allocation_enabled` to the public API in this PR?

The predicate is pure Warp device state — `device.is_cpu or device.is_mempool_enabled` — with no Newton-specific knowledge, so it is really a Warp capability fact. Warp already owns the adjacent family (`Device.is_mempool_supported`, `is_mempool_enabled`, `is_capturing`), and "can this device allocate during graph capture" should be added to Warp next to them (e.g. as a `Device` property), where the answer can be updated when Warp's internal requirements change (CPU capture semantics, future backends) without Newton having to track it.

Once this ships in a Newton release, our breaking-change policy means we'd have to carry a deprecation cycle for it when the Warp-native predicate appears, and the name is already under tension: it describes allocation safety, but the examples and tests use it as a general "should I capture" gate (on CUDA without mempool, capture without allocation is still valid). There's also precedent for keeping such wrappers internal — `check_conditional_graph_support()` wraps `wp.is_conditional_graph_supported()` and was never exported.

Concretely, for now: inline `device.is_cpu or device.is_mempool_enabled` in the two examples that use the helper (`example_robot_policy.py`, `example_robot_anymal_c_walk.py`) — arguably more instructive for readers copying the pattern anyway — and keep the helper private in `newton._src.utils` for `SolverVBD` and the tests. That drops the `newton/utils.py` re-export, the `docs/api/newton_utils.rst` entry, and the "Added" changelog entry, and leaves us free to adopt the Warp API when it exists.

File: newton/examples/contacts/example_nut_bolt_sdf.py [hunk 1]
```diff
@@ -253,12 +253,9 @@ def _build_nut_bolt_scene(self) -> newton.ModelBuilder:
...
-                self.simulate()
-            self.graph = capture.graph
-        else:
-            self.graph = None
+        with wp.ScopedCapture() as capture:
+            self.simulate()
+        self.graph = capture.graph
 
     def simulate(self):
         self.collision_pipeline.collide(self.state_0, self.contacts)
```
eric-heiden, line 265: 🟠 Enabling capture on CPU exposes an odd-substep ping-pong bug in this path and in `newton/examples/multiphysics/example_kamino_mujoco_admm_solver.py`. The Python swaps performed by `simulate()` are not replayed with the graph, so an odd number of substeps leaves the instance's `state_0` reference opposite the buffer read by the graph's first recorded launch. The Kamino example uses three substeps by default, which makes CPU replay start from stale state. Please use the odd-safe pattern from `example_robot_policy.py`, assigning the final output back into `state_0` on the last odd captured substep, or ensure the captured substep count is even.

CHANGELOG.md
```diff
@@ -49,6 +49,7 @@
...
 - Enable particle-particle contact forces by default in standalone `SolverSemiImplicit`: the solver now rebuilds `Model.particle_grid` every substep and adds contact forces to the existing `particle_f`. Existing scenes with particles closer than `2 * particle_max_radius + particle_cohesion` may change trajectories; set `model.particle_grid = None` before stepping to preserve the previous behavior without particle-particle contacts.
 - Change `CollisionPipeline.soft_contact_max` to default to the precomputed `soft_rigid_contact_pair_count` instead of `shape_count * particle_count`; pass `soft_contact_max=model.shape_count * model.particle_count` to preserve the previous capacity.
+- Enable graph capture on CPU across additional examples and tests; several multiphysics and robot examples that previously bypassed capture on CPU now capture there too, and `SolverVBD`'s contact-buffer resize guard no longer requires a CUDA mempool to allow allocation during capture.
```
jcarius-nv: The `[Unreleased]` changelog guidance in `AGENTS.md` and `docs/guide/development.rst` asks us to describe user-facing behavior and avoid internal test/implementation details. "across additional examples and tests" and "contact-buffer resize guard" expose those internals. Could we rephrase this around observable behavior, for example: "Enable graph capture on CPU in examples and allow `SolverVBD` contact buffers to grow during CPU graph capture."

jcarius-nv (dismissed):
Approving the current direction: Alain's documentation and private-API concerns are addressed, and the focused CPU suite passed (932 tests, 1 skipped).

Please still address these follow-ups before merging:

- Update the PR description: it still says `newton.utils.is_graph_capture_allocation_enabled` is public, while the current implementation intentionally keeps it in `newton._src.utils`.
- Address CodeRabbit's existing AnyMal CPU warmup finding before capture; I am not duplicating it inline here.
- The two existing duplication suggestions (shared multiphysics graph helpers and test capture boilerplate) are non-blocking and can be handled here or as follow-up cleanup.
- Eric's unresolved threads above

I added one inline changelog wording comment below.

CHANGELOG.md
c0d1f1ed: Thanks! Done.

newton/examples/contacts/example_nut_bolt_sdf.py
c0d1f1ed: Done

eric-heiden (approved):

PR #3389, opened 2026-07-08
Fix batched MuJoCo site poses

adenzler-nvidia:
## Description

Fix MuJoCo site poses for offset batched worlds.

`separate_worlds=True` builds the MuJoCo model from the first Newton world, but `site_pos` and `site_quat` remained single-world fields. Consequently, every world inherited the template site's pose in both the model and derived `site_xpos` / `site_xmat` data.

Add per-world site-to-shape mappings, expand the site pose fields, and refresh derived poses after construction and `SHAPE_PROPERTIES` notifications.

## Checklist

## Test plan

```bash
.venv/bin/python -m unittest newton.tests.test_sites_mujoco_export
.venv/bin/python -m unittest newton.tests.test_mujoco_solver.TestMuJoCoConversion.test_shape_offset_across_worlds
uvx pre-commit run -a
```

## Bug fix

**Steps to reproduce:**

1. Create two structurally identical Newton worlds containing a site attached directly to the world body.
2. Offset or rotate the second world.
3. Construct `SolverMuJoCo` with `separate_worlds=True`.
4. Inspect `mjw_model.site_pos` or `mjw_data.site_xpos`.
5. Observe that both worlds use the first world's site pose.

**Minimal reproduction:**

```python
import numpy as np
import warp as wp

import newton

world = newton.ModelBuilder()
body = world.add_link(mass=1.0, inertia=wp.mat33(np.eye(3)))
joint = world.add_joint_free(child=body)
world.add_articulation([joint])
world.add_site(
    -1,
    xform=wp.transform(wp.vec3(0.5, 0.25, 0.1), wp.quat_identity()),
)

builder.add_world(world)
builder.add_world(
    world,
    xform=wp.transform(wp.vec3(4.0, 0.0, 0.0), wp.quat_identity()),
)

model = builder.finalize()
    model,
    separate_worlds=True,
    disable_contacts=True,
)

print(solver.mjw_data.site_xpos.numpy()[:, 0])
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected MuJoCo per-site pose synchronization for offset multi-world setups.
  * Site positions/orientations now refresh correctly after runtime shape-transform updates.
* **New Features**
  * Enhanced MuJoCo export/runtime support for sites, including correct per-site world-space pose tiling across separate worlds.
  * MuJoCo site sizes now update at runtime from shape scaling (with the documented fallback when components are zero).
* **Tests**
  * Added regression coverage for site pose behavior across separate worlds.
  * Updated site transform export assertions to validate exported derived world-space site pose outputs.

### DIFF
--- a/newton/tests/test_sites_mujoco_export.py
+++ b/newton/tests/test_sites_mujoco_export.py
@@ -55,8 +55,11 @@ class TestMuJoCoSiteExport(unittest.TestCase):
         assert_site_poses_match_model()
 
         transforms = [
-            wp.transform(wp.vec3(-0.2, 0.4, 0.3), wp.quat_from_axis_angle(wp.vec3(1.0, 0.0, 0.0), 0.2)),
-            wp.transform(wp.vec3(3.0, 1.5, -0.1), wp.quat_from_axis_angle(wp.vec3(0.0, 1.0, 0.0), -0.3)),
+            wp.transform(
+                wp.vec3(-0.2 + i, 0.4 + 0.5 * i, 0.3 - 0.1 * i),
+                wp.quat_from_axis_angle(wp.vec3(0.0, 0.0, 1.0), 0.2 + 0.15 * i),
+            )
+            for i in range(model.shape_count)
         ]
         model.shape_transform.assign(wp.array(transforms, dtype=wp.transform, device=model.device))
         solver.notify_model_changed(newton.ModelFlags.SHAPE_PROPERTIES)
--- a/newton/tests/test_sites_mujoco_export.py
+++ b/newton/tests/test_sites_mujoco_export.py
@@ -66,6 +66,48 @@ class TestMuJoCoSiteExport(unittest.TestCase):
 
         assert_site_poses_match_model()
 
+    def test_site_size_updates_at_runtime(self):
+        """Propagate runtime shape_scale changes to MuJoCo site sizes."""
+        builder = newton.ModelBuilder()
+        body = builder.add_body(mass=1.0, inertia=wp.mat33(np.eye(3)))
+        builder.add_site(body, type=GeoType.BOX, scale=(0.1, 0.2, 0.3), label="box_site")
+        builder.add_site(body, type=GeoType.SPHERE, scale=(0.05, 0.05, 0.05), label="sphere_site")
+
+        model = builder.finalize()
+        solver = SolverMuJoCo(model)
+
+        np.testing.assert_allclose(solver.mjw_model.site_size.numpy(), [[0.1, 0.2, 0.3], [0.05, 0.05, 0.05]], atol=1e-6)
+
+        model.shape_scale.assign(wp.array([[0.4, 0.5, 0.6], [0.07, 0.0, 0.0]], dtype=wp.vec3, device=model.device))
+        solver.notify_model_changed(newton.ModelFlags.SHAPE_PROPERTIES)
+
+        # Zero components fall back to the first nonzero one, matching export.
+        np.testing.assert_allclose(solver.mjw_model.site_size.numpy(), [[0.4, 0.5, 0.6], [0.07, 0.07, 0.07]], atol=1e-6)
+
+    def test_site_size_updates_batched_worlds(self):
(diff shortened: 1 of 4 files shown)


newton/tests/test_sites_mujoco_export.py
```diff
@@ -16,6 +16,53 @@
...
+            )
+            np.testing.assert_allclose(solver.mjw_data.site_xmat.numpy()[:, 0], expected_xmat, atol=1.0e-6)
+
+        assert_site_poses_match_model()
+
+        transforms = [
+            wp.transform(wp.vec3(-0.2, 0.4, 0.3), wp.quat_from_axis_angle(wp.vec3(1.0, 0.0, 0.0), 0.2)),
+            wp.transform(wp.vec3(3.0, 1.5, -0.1), wp.quat_from_axis_angle(wp.vec3(0.0, 1.0, 0.0), -0.3)),
+        ]
+        model.shape_transform.assign(wp.array(transforms, dtype=wp.transform, device=model.device))
```
vreutskyy, line 64: Non-blocking: this model has four shapes (`sphere, site` in each world), but `assign()` receives only two transforms, so it updates only the first world's sphere/site and leaves the second-world site unchanged. Could we build a full-length transform array or update the site indices explicitly so the post-notify assertion exercises runtime synchronization in both worlds?

vreutskyy (dismissed):
Looks good overall. I left one non-blocking comment about strengthening the runtime-update coverage across both worlds.

newton/tests/test_sites_mujoco_export.py
adenzler-nvidia: Fixed in 09b7024ad. The runtime update now assigns a full-length transform array with distinct values for every shape, matching the other multi-world property-update tests, so both worldbody site indices are mutated before notify_model_changed().

vreutskyy (dismissed):
Re-reviewed the updated head. The follow-up now assigns distinct transforms to every shape, so runtime site synchronization is exercised in both worlds. Focused site tests and the offset-world regression pass; looks good.

eric-heiden:
🟠 Alain, would you like to handle site scaling in this PR as well? The new runtime update path syncs site poses, but it does not propagate `shape_scale` to `site_size`, so `notify_model_changed(ModelFlags.SHAPE_PROPERTIES)` leaves the MuJoCo size stale after resizing a site. I realize the current scope has been site poses so far, so this could also be split into a follow-up if you prefer.

adenzler-nvidia:
I'd prefer to split this into a follow-up — it's blocked on upstream anyway. `site_size` in mujoco_warp is `(nsite, 3)` with no world dimension (unlike `site_pos`/`site_quat`), so there's nothing to expand; batched propagation of `shape_scale` would need a mujoco_warp change first. Runtime scaling is also already a partial surface on the geom side: `geom_size` is batched and synced, but mesh collision uses the compile-time-baked `mesh_vert` (scale and re-centering baked into the vertices, also unbatched), so rescaling mesh shapes at runtime doesn't work either.

Worth noting for the follow-up: sites are always primitives, so unlike meshes there's no baked-geometry problem — once `site_size` gets a world dimension upstream, the sync is straightforward. And since `site_size` is read by the touch-sensor kernels, the stale size is a real gap for sensor users, not just cosmetic. I'll file an issue to track it.

adenzler-nvidia:
Correction to my previous comment: I misread the ask as per-world site scaling, and only that part is blocked on upstream. The plain runtime sync doesn't need any mujoco_warp change, so I implemented it here (f501c50a8): `SHAPE_PROPERTIES` now propagates `shape_scale` to `site_size`, mirroring the export-time zero-component fill. Caveat: `site_size` has no world dimension in mujoco_warp (unlike `geom_size`), so sizes are synced from the first world — per-world scale differences remain unsupported until `site_size` is batched upstream.

vreutskyy (approved):
Re-reviewed the latest head. The batched MJWarp site pose and runtime size synchronization match the intended scope, the multi-world regression coverage is in place, and the current checks are green. Looks good.

PR #3395, opened 2026-07-08
Parallelize forward kinematics on CUDA

adenzler-nvidia:
## Description

`KpiNewtonOverheadG1` showed that the MuJoCo update left a meaningful fraction of each 8k-world simulation step in Newton-side work. Profiling identified `eval_fk()` as a major contributor: the existing CUDA kernel assigns one thread to each articulation and evaluates the entire joint tree serially.

In a controlled benchmark that changes only the FK implementation, this PR reduces `KpiNewtonOverheadG1` from **9.99% to 3.50%**:

- 6.49 percentage-point reduction.
- 64.9% less Newton overhead.
- Approximately 2.85x lower remaining overhead.

This PR adds a level-parallel CUDA implementation of `eval_fk()`. Each articulation is assigned one 32-thread tile block, and joints within a tree level execute concurrently. Because a joint's parent joint always lies in exactly the previous level, the kernel stages parent transforms in two level-width shared-memory buffers that ping-pong between levels, so shared-memory usage scales with tree width rather than articulation size (G1: 336 B per block; a chain of any depth needs two slots). The level schedule is materialized once during model finalization and supports fully heterogeneous articulation sizes and joint types.

The existing serial implementation remains the fallback for CPU execution, differentiable models, non-tree articulation topologies, and tree levels wider than 512 joints. Both implementations share the joint-motion and child-state recurrence to keep their semantics aligned.

The change also:

- Uses the optimized path for MuJoCo reference-pose computation.
- Preserves masks, articulation indices, body filters, cable joints, and external parents.
- Supports articulations and tree levels wider than 32 by processing them in block-sized chunks.
- Preserves models with parallel joints or cyclic articulation topology by using the serial FK path when a unique level schedule cannot be constructed.
- Caches the specialized tile kernels so repeated eager `eval_fk()` calls avoid per-call kernel construction.

### Performance

Measured at 8,192 worlds on an NVIDIA RTX PRO 6000 Blackwell. Timings compare the final tiled kernel against the serial CUDA kernel after warmup.

| Model | Articulation topology | Tiled | Serial | Relative performance |
|---|---|---:|---:|---:|
| Cartpole | 2-joint chain | 0.0115 ms | 0.0095 ms | 0.83x |
| Franka FR3 | 10-joint chain | 0.0397 ms | 0.0471 ms | 1.19x |
| Ant | 9 joints, levels 1/4/4 | 0.0289 ms | 0.0783 ms | 2.71x |
| Allegro | heterogeneous 1- and 21-joint articulations | 0.0644 ms | 0.0891 ms | 1.38x |
| G1 | 44 joints across 14 levels | 0.0646 ms | 0.2590 ms | 4.01x |

The two-joint Cartpole is too small to amortize tile setup and level barriers, adding approximately 2 us per call. The realistic single-chain Franka model already benefits, while models with parallel tree branches show larger gains.

The table was measured with an earlier whole-articulation staging scheme; the final level-width double buffering re-measures equal or slightly faster on all models (e.g. G1 0.060 ms) while using a fraction of the shared memory.

At 8k G1 worlds, the level metadata occupies approximately 3.22 MiB and takes approximately 27 ms to construct plus 0.41 ms to upload during model finalization. The construction reuses the host joint arrays already needed for model upload and `joint_ancestor` generation.

## Checklist

## Test plan

```bash
WARP_CACHE_PATH=/tmp/newton-fk-cache .venv/bin/python -m newton.tests -k test_eval_fk --no-cache-clear
WARP_CACHE_PATH=/tmp/newton-fk-cache .venv/bin/python -m newton.tests -k test_kinematics --no-cache-clear
WARP_CACHE_PATH=/tmp/newton-fk-cache .venv/bin/python -m newton.tests -k test_mujoco_fk_consistency --no-cache-clear
uvx pre-commit run -a
```

The focused coverage compares the serial and tiled implementations across CPU and CUDA, mixed joint types, heterogeneous articulation sizes, masks and indices, body filters, cable joints, external parents, empty models, differentiable dispatch, non-tree and wide-level fallback, and deep and wide articulations exceeding 32 joints.


jcarius-nv:
@dylanturpin is this something you could take over from Alain?

camevor (commented):
Thanks @adenzler-nvidia and @dylanturpin, this looks good to me!

I found no regressions on the serial path, just one correctness issue:

A compact `ModelView` renumbers its joints, bodies, and articulations, but still inherits the parent model's `_fk_*` topology arrays, which contain parent-model indices.

`eval_fk()` sees those arrays and automatically selects the tiled fast path when running on CUDA. The kernel then uses parent-model indices against the view's compacted arrays. This can process the wrong articulation, read out of bounds, and write results to the wrong body.

The narrow fix would be, when `SolverCoupled` creates a compact `ModelView`, to clear all five inherited FK topology fields.

Once that is addressed, I think we can merge.
Nit, non-blocking: the newly added test functions should ideally have docstrings.

dylanturpin:
Thanks Chris — fixed in `c7c7741f` by clearing all five inherited FK topology fields when `SolverCoupled` compacts a `ModelView`, so compact CUDA views fall back to the correctly rebased serial FK path rather than consuming parent indices.

I added a regression to the existing non-prefix compact-articulation fixture. It fails on `e524f957` before the fix because the view still inherits the parent topology, and passes after the fix. The focused regression passes; the full coupled-solver suite passes 101/101 (CPU plus its registered CUDA cases); and pre-commit passes.

I’ve re-requested your review on the updated head.

camevor (approved):
Thanks, Dylan, lgtm!

eric-heiden (approved):

PR #3403, opened 2026-07-08
Fix handling of non-watertight meshes

### LINKED ISSUE #3242: [BUG] Collision detection uses `mesh_query_point_sign_parity` -- silently broken for non-watertight meshes

AnkaChan:
## Description

Mesh inside/outside **sign** queries are made adaptive to mesh topology. Previously every mesh sign query used parity ray-casting (`wp.mesh_query_point_sign_parity`), valid only for watertight (closed) meshes; for a non-watertight mesh (a plane, or a rigid asset with open boundaries such as the Unitree H1) the parity sign is ill-defined and flips, so SDF/soft-contact sign for open collision geometry was wrong (#3242).

Watertightness is checked once per unique collidable mesh at model finalization (visual-only meshes skipped) and stored internally (`Model._shape_mesh_properties`); the sign method is then derived automatically per path:

| mesh topology | runtime queries (soft/particle contacts, mesh-mesh SDF narrow phase) | SDF baked at finalization |
|---|---|---|
| watertight | parity | parity |
| non-watertight | angle-weighted pseudo-normal | winding numbers |

The selection machinery (`MeshProperties`, `MeshSignMethod`) is internal — no new shape-level public API. For deliberate, explicit bakes, `Mesh.build_sdf` gains a `sign_method` kwarg (`"auto" | "parity" | "winding" | "normal"`) as an advanced escape hatch, including a new pseudo-normal bake mode for open sheets.

**Continues #3249 by @nvtw** — this branch preserves his commits and addresses the review feedback on top. Supersedes #3249. Closes #3242.

**Changes on top of @nvtw's work:**
- Keep the sign selection fully automatic and internal, per @eric-heiden's review: the runtime method is a pure function of the stored per-mesh watertightness; no public `ShapeFlags` additions. (A semantic surface-vs-solid override can feed the same resolution function later if runtime ambiguity becomes a demonstrated problem.)
- Add `sign_method` to `Mesh.build_sdf` and support an angle-weighted pseudo-normal (`"normal"`) strategy in the SDF bake pipeline (@AnkaChan, @eric-heiden).
- Name the SDF culling search slack (`_SDF_QUERY_RADIUS_SLACK`) and raise `_MESH_QUERY_MAX_DIST` to an effectively unbounded `1e11` with a documented float32 scale budget (@mmacklin, @WenchaoHuang).
- Add member docstrings to `MeshSignMethod` / `MeshProperties` and document why pseudo-normal (not the winding number the bake falls back to) is the non-watertight runtime choice (@eric-heiden, @WenchaoHuang).
- Pre-allocate the empty `shape_mesh_properties` fallback in `NarrowPhase`; extract a duplicated test helper (CodeRabbit).
- Register `Model._shape_mesh_properties` in the model attribute specs — fixes `test_core_attribute_specs_cover_entity_indexed_storage`, red on all platforms after merging latest `main`.
- Add an end-to-end open-mesh soft-contact regression test exercising the actual #3242 failure at the contact level, plus `build_sdf` passthrough and numeric normal-bake tests.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -p 'test_collision_pipeline.py'   # 112 tests, incl. mesh sign-query suite
uv run --extra dev -m newton.tests -p 'test_model.py'                # 77 tests, attribute-spec coverage
uv run --extra dev -m newton.tests -p 'test_sdf_texture.py'          # 26 tests, bake sign modes
uv run --extra dev -m newton.tests -p 'test_sdf_compute.py'          # 63 tests, SDF construction
uv run --extra dev -m newton.tests -p 'test_sdf_disk_cache.py'       # 14 tests, cache keying
```

All pass on L40 (CUDA).

## Bug fix

1. Use a non-watertight mesh as collision geometry (open box, plane, or an H1 asset).
2. Run a soft/SDF-contact sim (e.g. `examples/cloth/example_cloth_h1.py`).
3. Points near an opening are signed inconsistently by parity ray-casting → inverted/dropped contact response.

**Minimal reproduction:**

```python
import numpy as np, warp as wp
wp.init()
# Open box (top face removed) — non-watertight.
v = np.array([[-.5,-.5,-.5],[.5,-.5,-.5],[.5,.5,-.5],[-.5,.5,-.5],
              [-.5,-.5,.5],[.5,-.5,.5],[.5,.5,.5],[-.5,.5,.5]], dtype=np.float32)
f = np.array([[0,2,1],[0,3,2],[0,1,5],[0,5,4],[1,2,6],[1,6,5],
              [2,3,7],[2,7,6],[3,0,4],[3,4,7]], dtype=np.int32)  # every face but +z
mesh = wp.Mesh(wp.array(v, dtype=wp.vec3), wp.array(f.reshape(-1), dtype=wp.int32))

@wp.kernel
def probe(m: wp.uint64, out: wp.array(dtype=wp.vec2)):
    p = wp.vec3(0.25, 0.0, 0.3)  # inside, just below the opening
    par = wp.mesh_query_point_sign_parity(m, p, 10.0)
    s = float(0.0); fi = int(0); u = float(0.0); vv = float(0.0)
    wp.mesh_query_point_sign_normal(m, p, 10.0, s, fi, u, vv)
    out[0] = wp.vec2(par.sign, s)  # parity: +1 (wrong, "outside"), normal: -1 (correct)

out = wp.zeros(1, dtype=wp.vec2)
wp.launch(probe, dim=1, inputs=[mesh.id, out])
print(out.numpy())  # [ 1. -1.]
```

## Summary by CodeRabbit

* **New Features**
  * Added `sign_method`/`sign_mode` support for `normal` mesh sign computation (alongside `auto`, `parity`, `winding`) for SDF baking and texture SDF generation.
  * Introduced automatic per-shape sign selection for mesh collisions and soft contacts, adapting between open and watertight meshes.
* **Bug Fixes**
  * Corrected inside/outside sign behavior across collision, soft-contact, and SDF baking stages, including clear override behavior when `sign_method` is explicitly set.
* **Documentation**
  * Updated changelog with the new options and clarified sign-query behavior.
* **Tests**
  * Expanded regression coverage for open meshes, per-shape sign routing, and `normal` mode.


AnkaChan:
@WenchaoHuang — responses to your review comments from #3249:

**[1/7]** `sdf_contact.py` — hard-coded `1000.0` → named constant, and the `wp.inf` question:
Addressed in this PR: the constant is named (`_MESH_QUERY_MAX_DIST`) and raised to `1e11` — effectively unbounded for any asset under any unit convention (the queries run in mesh-local coordinates, with per-shape scale divided out, so there's no simulation scale to tie it to).

On `wp.inf` / float32-max: Warp's mesh queries square `max_dist` internally (`native/mesh.h`: `min_dist_sq = max_dist * max_dist`), and the no-hit branches return `max_dist` as a sentinel that contact math multiplies by per-shape scale. Budgeting scale factors up to `1e6` for extreme unit conversions, `(1e11 * 1e6)^2 = 1e34` clears `FLT_MAX` (~3.4e38) by four orders of magnitude, whereas values near `sqrt(FLT_MAX) ~ 1.8e19` would overflow to `inf` and propagate into contacts.

**[2/7]** `kernels.py` — represent the constants as an enum like `GeoType`:
Done — `MeshSignMethod(IntEnum)` and `MeshProperties(IntFlag)`. In this PR the values additionally match the `ShapeFlags` field encoding (`AUTO=0, NORMAL=1, PARITY=2`; flag = `method << 5`), so the flag field and the enum are the same number.

AnkaChan:
@eric-heiden — response to your review comment from #3249:

**[3/7]** `kernels.py` — inline docstrings, maybe better names:
Member docstrings added in this PR. On names: kept `MeshSignMethod` / `MeshProperties` — after the flag-field redesign they line up 1:1 with `ShapeFlags.MESH_SIGN_*`, which reads consistently in context. If you had specific alternatives in mind, happy to rename.

AnkaChan:
@mmacklin — responses to your review comments from #3249 *(updated for the API simplification in b6d94872; see the reply to @eric-heiden below)*:

**[4/7]** `test_collision_pipeline.py` — "what is the `7`?":
Resolved by removal. The `1 << 7` literal was a phantom "invalid" bit of the mesh-sign flag mask (`0b111 << 5` claimed three bits while only two one-hot methods existed), and the flag field itself has since been dropped per @eric-heiden's review — the sign method is now selected automatically from internally stored mesh properties, and the tests exercise that resolution directly with named `MeshProperties` values.

**[5/7]** `flags.py` — document which routines the flags affect, `build_sdf()`, when to use each, perf:
The shape-level flags were removed per @eric-heiden's review, so the surface to document shrank. Up-to-date behavior:

*What selects the sign method?* Nothing user-facing — it is automatic. Watertightness is checked once per unique collidable mesh at model finalization (visual-only meshes skipped) and stored internally (`Model._shape_mesh_properties`); each consumer derives the method from it.

*Which routines use it?*
1. **Runtime mesh sign queries** — soft/particle contacts (`create_soft_contacts`) and the mesh-mesh SDF narrow phase driven by `CollisionPipeline`: parity if watertight, angle-weighted pseudo-normal otherwise.
2. **SDFs baked automatically during `ModelBuilder.finalize()`**: parity if watertight, winding numbers otherwise.

*Is `build_sdf()` affected?* Explicit `mesh.build_sdf()` calls take their own `sign_method` argument (`"auto" | "parity" | "winding" | "normal"`, default `"auto"` = the same topology rule) as a deliberate escape hatch for deliberate bakes.

*Performance:* only which Warp query runs changes — parity is the cheapest, pseudo-normal comparable, winding the most expensive and the only mode requiring winding-number acceleration on the mesh (built only when winding is actually used).

**[6/7]** `sdf_contact.py` — `_MESH_QUERY_MAX_DIST` should be larger or tied to global units:
Raised to `1e11` — effectively unbounded (~8x Earth's diameter in millimeter units). It can't be tied to global units: these queries run in mesh-local coordinates with the per-shape scale divided out. The ceiling for going larger is float32: Warp's mesh queries square `max_dist` internally, and the no-hit sentinel is multiplied by per-shape scale in contact math — budgeting scale up to `1e6`, `(1e11 * 1e6)^2 = 1e34` still clears `FLT_MAX` by four orders, while anything near `sqrt(FLT_MAX) ~ 1.8e19` would overflow. It's never binding in practice: culling passes tight `slack * threshold` radii and the remaining queries run at points the narrow phase already placed near the surface.

**[7/7]** `sdf_contact.py` — remove the `1.01` magic number:
Named: `_SDF_QUERY_RADIUS_SLACK = 1.01`, with a comment — the search radius slightly exceeds the culling threshold so points right at the threshold are classified by a real closest-point query rather than the no-hit sentinel. Applied to both kernel variants.

eric-heiden:
Thanks for addressing non-watertight mesh collisions. I recommend simplifying the public API by keeping `MeshProperties` and the concrete query method internal.

The default behavior could be:

- Check watertightness once per unique collidable mesh during finalization, while skipping visual-only meshes.
- Use parity for watertight meshes.
- Use pseudo-normal for runtime queries on non-watertight meshes.
- Use winding for automatically baked SDFs on non-watertight meshes.

I would avoid exposing the shape-level `MESH_SIGN_*` flags in this PR. Watertightness cannot distinguish an open sheet from a slightly leaky solid, but the current normal/parity override does not fully solve that ambiguity either because there is no runtime winding option. It therefore adds another public knob without covering the difficult case.

Keeping `sign_method` on explicit `Mesh.build_sdf()` calls still seems useful as an advanced escape hatch because the caller is deliberately baking an SDF there. If runtime ambiguity becomes a demonstrated problem, we could add one semantic override later, such as whether the mesh represents a surface or a solid, and derive the internal algorithm from that intent. This keeps the common path automatic while leaving room for a focused solution when needed.

AnkaChan:
@eric-heiden Makes sense — implemented in b6d94872. The PR now matches the shape you describe:

- **No public `MESH_SIGN_*` shape flags.** The runtime sign method is derived purely from the internally stored per-mesh properties: watertightness is checked once per unique collidable mesh at finalization (visual-only meshes skipped), parity for watertight, pseudo-normal for runtime queries on non-watertight, winding for automatically baked SDFs on non-watertight.
- **`MeshProperties` / `MeshSignMethod` stay internal** (`newton._src`); the properties array holds only intrinsic facts (currently `WATERTIGHT`), and the method is computed where consumed — which leaves a clean slot for the semantic surface-vs-solid override you describe (`Mesh.is_solid` already exists as a natural home) feeding the same resolution function later.
- **`Mesh.build_sdf(sign_method=...)` remains** as the deliberate escape hatch for explicit bakes, including the `"normal"` mode.

Only user-visible surface left by this PR: the automatic behavior change (open meshes now sign correctly) and the `build_sdf` kwarg.

newton/_src/geometry/kernels.py
```diff
@@ -12,6 +14,38 @@
 )
 
 
+class MeshSignMethod(enum.IntEnum):
+    """Method used to determine the inside/outside sign of a mesh point query."""
+
+    NORMAL = 0
+    """Angle-weighted closest-face pseudo-normal; robust for open surfaces."""
+    PARITY = 1
+    """Ray-crossing parity; correct and cheap for watertight (closed) meshes."""
```
eric-heiden: I think we can still unify this a bit, right? 

In sdf_texture.py you have
```
            SIGN_MODE_NORMAL,
            SIGN_MODE_PARITY,
            SIGN_MODE_WINDING,
```

Then in the `build_sdf()` method you have literals ``"auto"``, ``"parity"``, ``"winding"``, and ``"normal"``.

Can't we stick to one IntEnum that you use everywhere? Maybe make `MeshSignMethod` the only symbol for this (exposed to the user and consumed in the kernels) and add an `AUTO = -1` or something to this? Perhaps rename it to `SignMethod` if this isn't only relevant to meshes but I think we definitely don't want 3 ways to declare the same thing.

eric-heiden: I would also move these enums to `geometry/flags.py` or rename `geometry/flags.py` to `enums.py` maybe. This shouldn't be declared in kernels.py. Same for `MeshProperties` below.

AnkaChan: In `coloring()` we do it the other way around, it takes text arguments and converts it to internal `IntEnum`. I remember @mmacklin advised me to use text arguments. Should we keep the style unified?

WenchaoHuang (dismissed):
LGTM

eric-heiden (approved):
Thanks!

PR #3404, opened 2026-07-08
Hide guide-purpose prims in USD visual shape loading

ooctipus:
## Description

USD renderers hide prims whose computed purpose is `guide` by default, but the visual-shape loader in `ModelBuilder.add_usd()` only checked computed *visibility*. Guide prims were therefore imported with the `VISIBLE` flag set.

This is very noticeable with MuJoCo-converted assets (e.g. `mujoco_menagerie` via the MuJoCo USD exporter): the converter authors collision geometry as `purpose=guide` prims with `physics:collisionEnabled=false`, so Newton viewers drew all the convex collision hulls on top of the render meshes — the robot appeared wrapped in collision geometry, while usdview/Omniverse showed the expected visuals.

The fix is scoped to visual-shape loading only:

- guide prims are still imported, just as *hidden* visual shapes — the same treatment `visibility=invisible` prims already get;
- enabled guide colliders still collide and keep following the collider display policy (`force_show_colliders`, `hide_collision_shapes`);
- `proxy`-purpose prims remain visible (they are the intended preview representation).

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_guide_purpose_shapes_not_visible
uv run --extra dev -m newton.tests -k test_collision_shape_visibility_flags
```

The new regression test fails without the fix (`AssertionError: 1 is not false` on the guide prim's `VISIBLE` flag) and passes with it. The full `test_import_usd` module passes (355 tests).

## Bug fix

**Steps to reproduce:**

1. Convert a MuJoCo model that has dedicated collision geometry to USD (e.g. `mujoco_menagerie/franka_emika_panda/mjx_scene`) — collision geoms are authored with `purpose=guide` and `physics:collisionEnabled=false`.
2. Load it with `ModelBuilder.add_usd(stage)` and render with any Newton viewer.
3. The guide collision hulls are drawn on top of the render meshes.

**Minimal reproduction:**

```python
import newton
from pxr import Usd

usd_content = """#usda 1.0
def Xform "Body" (prepend apiSchemas = ["PhysicsRigidBodyAPI"])
{
    def Sphere "Visual" { double radius = 0.3 }
    def Cube "GuideHull" (prepend apiSchemas = ["PhysicsCollisionAPI"])
    {
        uniform token purpose = "guide"
        bool physics:collisionEnabled = false
        double size = 1.0
    }
}
"""
stage.GetRootLayer().ImportFromString(usd_content)

flags = builder.shape_flags[result["path_shape_map"]["/Body/GuideHull"]]
assert not (flags & newton.ShapeFlags.VISIBLE)  # fails without this PR
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved USD importing so visuals follow viewport “drawn” semantics, preventing `guide`-purpose prims from being imported as visible visual shapes (including Gaussian splats).
  * Visual-shape visibility is now limited to `default` and `proxy` purposes; `guide` visuals are suppressed unless explicitly enabled.

* **Tests**
  * Added coverage for `default`, `proxy`, `render`, and `guide`-purpose prim handling, including collider enabled/disabled behavior and how collider debugging overrides visibility.

* **Chores**
  * Updated the changelog entry for the USD importer fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -189,6 +189,7 @@
 - Fix `SensorTiledCamera.utils.convert_ray_depth_to_forward_depth()` to preserve the clear-depth sentinel for zero-direction rays and non-positive depths.
 - Fix `SolverMuJoCo` placing articulations whose root is a fully-locked D6 joint (e.g. imported from a generic USD `PhysicsJoint`) at the first world's root pose in every world; such roots now become mocap bodies like fixed-joint roots. (#3499; fixes #3430)
 - Fix `SensorTiledCamera` rendering cloth and volume deformable vertices as particle spheres while preserving standalone particle rendering in mixed scenes. (#3518)
+- Fix the USD importer loading `guide`-purpose prims as visible visual shapes; USD renderers do not draw guides by default, so viewers no longer overlay collision geometry from MuJoCo-converted assets on top of the render meshes.
 - Fix `ViewerGL.get_frame()` crashing when a CPU model is rendered while a CUDA context is active.
 - Fix `ViewerUSD` leaving stale particle geometry visible when the active particle count drops to zero. (#2992)
 - Fix `eval_inverse_dynamics_passive()` and `SolverFeatherstone` intermittently dropping descendant wrench contributions during the articulated-body backward pass on CUDA. (#3443)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -189,7 +189,7 @@
 - Fix `SensorTiledCamera.utils.convert_ray_depth_to_forward_depth()` to preserve the clear-depth sentinel for zero-direction rays and non-positive depths.
 - Fix `SolverMuJoCo` placing articulations whose root is a fully-locked D6 joint (e.g. imported from a generic USD `PhysicsJoint`) at the first world's root pose in every world; such roots now become mocap bodies like fixed-joint roots. (#3499; fixes #3430)
 - Fix `SensorTiledCamera` rendering cloth and volume deformable vertices as particle spheres while preserving standalone particle rendering in mixed scenes. (#3518)
-- Fix the USD importer loading `guide`-purpose prims as visible visual shapes; USD renderers do not draw guides by default, so viewers no longer overlay collision geometry from MuJoCo-converted assets on top of the render meshes.
+- Fix the USD importer loading `guide`- and `render`-purpose prims as visible visual shapes; visual shapes and Gaussian splats now follow USD viewport semantics and are drawn only for the `default` and `proxy` purposes.
 - Fix `ViewerGL.get_frame()` crashing when a CPU model is rendered while a CUDA context is active.
 - Fix `ViewerUSD` leaving stale particle geometry visible when the active particle count drops to zero. (#2992)
 - Fix `eval_inverse_dynamics_passive()` and `SolverFeatherstone` intermittently dropping descendant wrench contributions during the articulated-body backward pass on CUDA. (#3443)
--- a/newton/_src/utils/import_usd.py
+++ b/newton/_src/utils/import_usd.py
@@ -1038,6 +1038,17 @@ def parse_usd(
             return False
         return imageable.ComputeVisibility() != UsdGeom.Tokens.invisible
 
+    def _is_guide_purpose(prim: Usd.Prim) -> bool:
+        """Return whether ``prim``'s computed purpose is ``guide``.
+
+        Guide prims are viewport helpers that USD renderers do not draw by
+        default; e.g. MuJoCo-converted assets author collision geometry with
+        purpose ``guide``. They are excluded from visual shape loading but
+        remain eligible as collision shapes.
+        """
+        imageable = UsdGeom.Imageable(prim)
+        return bool(imageable) and imageable.ComputePurpose() == UsdGeom.Tokens.guide
+
     bodies_with_visual_shapes: set[int] = set()
 
     def _get_prim_world_mat(prim, articulation_root_xform, incoming_world_xform):
@@ -1114,7 +1125,9 @@ def parse_usd(
             return
 
         visual_shape_cfg_for_prim = copy.copy(visual_shape_cfg)
-        visual_shape_cfg_for_prim.is_visible = is_site or _is_effectively_visible(prim)
+        visual_shape_cfg_for_prim.is_visible = is_site or (
+            _is_effectively_visible(prim) and not _is_guide_purpose(prim)
+        )
         material_props = _get_material_props_cached(prim)
         shape_color = material_props.get("color")
 
--- a/newton/_src/utils/import_usd.py
+++ b/newton/_src/utils/import_usd.py
@@ -1038,6 +1038,21 @@ def parse_usd(
             return False
         return imageable.ComputeVisibility() != UsdGeom.Tokens.invisible
 
+    def _is_viewport_drawn(prim: Usd.Prim) -> bool:
+        """Return whether a prim is drawn under viewport semantics.
+
+        USD viewports draw the ``default`` and ``proxy`` purposes and hide ``guide`` and
+        ``render``; the allowlist also keeps any future purpose hidden until explicitly
+        handled. Colliders deliberately do not use this check: ``guide`` is the conventional
+        purpose for authored collision geometry (e.g. the MuJoCo USD exporter), and the
+        collider display policy (``force_show_colliders`` / ``hide_collision_shapes``) is the
+        explicit mechanism for revealing colliders — gating them on purpose would make
+        ``force_show_colliders`` a no-op on such assets.
+        """
+        if not _is_effectively_visible(prim):
+            return False
+        return UsdGeom.Imageable(prim).ComputePurpose() in (UsdGeom.Tokens.default_, UsdGeom.Tokens.proxy)
+
     bodies_with_visual_shapes: set[int] = set()
 
     def _get_prim_world_mat(prim, articulation_root_xform, incoming_world_xform):
@@ -1114,11 +1129,7 @@ def parse_usd(
             return
 
         visual_shape_cfg_for_prim = copy.copy(visual_shape_cfg)
-        # Guide-purpose prims are viewport helpers that USD renderers do not draw by
-        # default (e.g. MuJoCo-converted assets author collision geometry as guides).
-        imageable = UsdGeom.Imageable(prim)
-        is_guide = bool(imageable) and imageable.ComputePurpose() == UsdGeom.Tokens.guide
-        visual_shape_cfg_for_prim.is_visible = is_site or (_is_effectively_visible(prim) and not is_guide)
+        visual_shape_cfg_for_prim.is_visible = is_site or _is_viewport_drawn(prim)
         material_props = _get_material_props_cached(prim)
         shape_color = material_props.get("color")
 
@@ -3270,6 +3281,7 @@ def parse_usd(
                 collider_is_visible = (
                     show_collider_by_policy or collider_has_visual_material
                 ) and not hide_collider_for_body
+                # visibility only — see _is_viewport_drawn for why purpose does not gate colliders
                 collider_is_visible = collider_is_visible and _is_effectively_visible(prim)
 
                 # Contact response precedence:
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -83,6 +81,7 @@
 - Fix `ViewerFile.is_running()` to return `False` after `ViewerFile.close()` so headless recording loops can terminate like interactive viewers. (#3094)
 - Raise an error when `SolverVBD(rigid_contact_history=True)` would allocate or grow contact-history buffers during CUDA graph capture; construct `CollisionPipeline` before `SolverVBD`, or run one uncaptured solver step before capture.
 - Fix `SensorTiledCamera.utils.convert_ray_depth_to_forward_depth()` to preserve the clear-depth sentinel for zero-direction rays and non-positive depths.
+- Fix the USD importer loading `guide`-purpose prims as visible visual shapes; USD renderers do not draw guides by default, so viewers no longer overlay collision geometry from MuJoCo-converted assets on top of the render meshes.
```
eric-heiden: @ooctipus can you share an asset that has this problem? Your changes are only in `_load_visual_shapes_impl()`. That method shouldn't get called on collision geometry anyway. So I would like to understand how collision geometry can end up becoming visible here.

ooctipus:
before
image

after fix
image

[franka.zip](https://github.com/user-attachments/files/29818406/franka.zip)
load the mjx_scene.usda

newton/_src/utils/import_usd.py
```diff
@@ -1093,7 +1093,11 @@ def _load_visual_shapes_impl(
             return
 
         visual_shape_cfg_for_prim = copy.copy(visual_shape_cfg)
-        visual_shape_cfg_for_prim.is_visible = is_site or _is_effectively_visible(prim)
+        # Guide-purpose prims are viewport helpers that USD renderers do not draw by
+        # default (e.g. MuJoCo-converted assets author collision geometry as guides).
+        imageable = UsdGeom.Imageable(prim)
+        is_guide = bool(imageable) and imageable.ComputePurpose() == UsdGeom.Tokens.guide
```
andrewkaufman: This only gates on `guide`. What about `render`-purpose prims?

USD defines four purposes: `default`, `proxy`, `guide`, `render`. A conformant viewport draws `default` + `proxy` and hides `guide` + `render`. Newton's viewer is an interactive viewport — it should follow viewport semantics, not renderer semantics.

Suggestion — flip the logic to an allowlist:
```python
is_hidden_purpose = bool(imageable) and imageable.ComputePurpose() not in (
    UsdGeom.Tokens.default_,
    UsdGeom.Tokens.proxy,
)
```
This hides both `guide` AND `render` purpose prims, and is future-proof if USD adds new purposes (hidden by default until explicitly handled).

newton/_src/utils/import_usd.py
```diff
@@ -1093,7 +1093,11 @@ def _load_visual_shapes_impl(
             return
 
         visual_shape_cfg_for_prim = copy.copy(visual_shape_cfg)
-        visual_shape_cfg_for_prim.is_visible = is_site or _is_effectively_visible(prim)
+        # Guide-purpose prims are viewport helpers that USD renderers do not draw by
+        # default (e.g. MuJoCo-converted assets author collision geometry as guides).
+        imageable = UsdGeom.Imageable(prim)
```
andrewkaufman: The `UsdGeom.Imageable(prim)` construction duplicates what `_is_effectively_visible` does internally. Consider folding purpose handling into `_is_effectively_visible` so the check applies consistently to all visual paths (colliders at line 3223, Gaussian splats at line 3711):
```python
def _is_effectively_visible(prim: Usd.Prim) -> bool:
    imageable = UsdGeom.Imageable(prim)
    if not imageable:
        return False
    if imageable.ComputeVisibility() == UsdGeom.Tokens.invisible:
        return False
    purpose = imageable.ComputePurpose()
    if purpose not in (UsdGeom.Tokens.default_, UsdGeom.Tokens.proxy):
        return False
    return True
```
One function, one place to maintain, consistent behavior everywhere.

newton/tests/test_import_usd.py
```diff
@@ -9651,6 +9651,89 @@ def Sphere "CollisionSphere" (
...
+        self.assertFalse(flags_guide & ShapeFlags.COLLIDE_SHAPES)
+        self.assertFalse(flags_guide & ShapeFlags.VISIBLE)
+
+        # An enabled guide-purpose collider still collides; its display keeps following
+        # the collider policy (hidden here because the body has visual shapes).
+        flags_enabled_guide = builder.shape_flags[path_shape_map["/Body/EnabledGuideCollisionBox"]]
+        self.assertTrue(flags_enabled_guide & ShapeFlags.COLLIDE_SHAPES)
+        self.assertFalse(flags_enabled_guide & ShapeFlags.VISIBLE)
+
+        # force_show_colliders still reveals guide-purpose colliders for debugging.
```
andrewkaufman: Does `force_show_colliders` also reveal the *disabled* guide-purpose collider (`GuideCollisionBox`)? If not, how does one reveal disabled guide geometry for debugging?

Also: should there be a `render`-purpose prim in this test? If we agree Newton should hide both guide and render, the test should cover both.

newton/tests/test_import_usd.py
```diff
@@ -9651,6 +9651,89 @@ def Sphere "CollisionSphere" (
...
+        builder2 = newton.ModelBuilder()
+        result2 = builder2.add_usd(stage, force_show_colliders=True)
+        flags_forced = builder2.shape_flags[result2["path_shape_map"]["/Body/EnabledGuideCollisionBox"]]
+        self.assertTrue(flags_forced & ShapeFlags.VISIBLE)
+
     @staticmethod
     def _create_stage_with_pbr_collision_mesh(color, roughness, metallic, *, add_visual_sphere=False):
         """Create a stage with a rigid body containing a collision mesh with PBR material."""
```
andrewkaufman, line 10093: Missing a pure visual guide mesh (no `PhysicsCollisionAPI`). That's the most direct exercise of the fix — a guide mesh going through visual loading and incorrectly getting `VISIBLE=True`. The current test's `GuideCollisionBox` is disabled-collider → visual fallback, which is a more indirect path.

andrewkaufman (commented):
The bug is real but the fix is incomplete.

Newton's viewer is a viewport, not an offline renderer. USD purpose semantics for a viewport: show `default` + `proxy`, hide `guide` + `render`. This PR only handles `guide` and ignores `render`-purpose prims entirely — same class of bug (assets with high-poly render geometry authored as `purpose=render` would still import visible and clutter the viewport).

The fix should decide: is Newton's viewer a viewport or a renderer? If viewport (which it is), the check should be `purpose not in (default, proxy)` → hidden, not just `purpose == guide` → hidden. If you want to be more conservative: at minimum acknowledge `render` purpose exists and document why it's not handled here.

newton/_src/utils/import_usd.py
ooctipus: Fixed in 7c240217c — flipped to your allowlist form: `ComputePurpose() in (default_, proxy)`, so render-purpose prims are hidden too and future purposes stay hidden until explicitly handled. Agreed Newton's viewer is a viewport; the PR now says so.

newton/_src/utils/import_usd.py
ooctipus: Consolidated in 7c240217c, with one deliberate split: the shared check is a new `_is_viewport_drawn` (visibility + purpose allowlist) used by the visual-shape and Gaussian-splat paths, while colliders keep visibility-only `_is_effectively_visible`. Folding purpose into the collider path would make `force_show_colliders` a no-op on MuJoCo-converted assets — their *enabled* colliders are guide-purpose too, and the collider display policy is exactly the tool for revealing them. The reasoning lives in `_is_viewport_drawn`'s docstring and a comment at the collider site.

newton/tests/test_import_usd.py
ooctipus: Both added in 7c240217c. Semantics for the disabled guide collider under `force_show_colliders`: it stays hidden, and the test now asserts that — with collision disabled it isn't a collider in the model at all (imported as guide visual content), so the collider policy doesn't apply; it's hidden the same way usdview would hide it, and revealing it means changing purpose/visibility on the stage. A render-purpose prim is in the test now too.

newton/tests/test_import_usd.py
ooctipus: Added in 7c240217c — a plain guide-purpose sphere with no collision API, asserted hidden. You're right it's the most direct exercise of the fix; the disabled-collider case now covers the fallback path specifically rather than standing in for both.

CHANGELOG.md
ooctipus: Sure — the mechanism is the disabled-collider fallback: prims with `PhysicsCollisionAPI` but `physics:collisionEnabled = false` don't become colliders, they're imported through `_load_visual_shapes_impl()` as visual-only shapes (the regression test's `GuideCollisionBox` walks exactly that path: `COLLIDE_SHAPES` unset, loaded via visual loading). That's how the MuJoCo USD exporter authors dedicated collision geometry — `purpose = guide` + `collisionEnabled = false` on the collision meshes (visual-only capsule/hull stand-ins), with the real contact handled elsewhere. Concrete asset: `mujoco_menagerie/franka_emika_panda/mjx_scene.xml` converted with the MuJoCo USD exporter — every `collision_*` prim in the output is authored that way, and before this fix Newton drew all of them on top of the render meshes. Happy to attach the converted usda if useful.

newton/tests/test_import_usd.py
```diff
@@ -9651,6 +9651,116 @@ def Sphere "CollisionSphere" (
...
+        # An enabled guide-purpose collider still collides; its display keeps following
+        # the collider policy (hidden here because the body has visual shapes).
+        flags_enabled_guide = builder.shape_flags[path_shape_map["/Body/EnabledGuideCollisionBox"]]
+        self.assertTrue(flags_enabled_guide & ShapeFlags.COLLIDE_SHAPES)
+        self.assertFalse(flags_enabled_guide & ShapeFlags.VISIBLE)
+
+        # force_show_colliders still reveals guide-purpose colliders for debugging. A
+        # *disabled* guide collider is no collider at all (it imported as guide visual
+        # content), so the collider policy does not apply and it stays hidden — like any
+        # USD viewport would hide it; reveal it by changing purpose/visibility on the stage.
```
andrewkaufman, line 10083: This is not accurate though -- any other USD viewport would provide a purpose filter mechanism so users can change the default visualization to include guides (and potentially exclude other purposes).... editing the stage just to visualize certain purposes is not a good workflow.

I don't have a better solution within scope of this PR though, so be it I guess.

CHANGELOG.md
andrewkaufman: This explanation sounds a bit odd to me.... Are we sure that's not a bug in the mujoco usd exporter? When you say "mujoco usd exporter" does that mean `mujoco-usd-converter` (from pypi) or the Isaac Sim / Lab Asset Importer or something else entirely?

newton/_src/utils/import_usd.py
andrewkaufman: @eric-heiden I think we should discuss exporing a USD purpose-like mechanism for the Newton viewers (outside scope of this PR)... Sites are also purpose = guide in MuJoCo-converted assets. People are going to want to visualize guides sometimes.

eric-heiden: Yes, maybe we can wrap this into the shape flags we have in Newton. I've created an issue here for follow-up: https://github.com/newton-physics/newton/issues/3442.

eric-heiden (approved):

CHANGELOG.md
eric-heiden: @ooctipus can you improve the wording of this changelog entry then to remove this part `so viewers no longer overlay collision geometry from MuJoCo-converted assets on top of the render meshes`? I don't think the official MJCF-converted assets we have on newton-assets show this defect either. IIRC in the asset you shared there were no explicit visual shapes so a fallback to show collision shapes was triggered, which then requires this purpose handling to hide the true colliders. In any case, this is a strange asset. Would be good to double-check whether there is an issue in the converter as Andrew pointed out.

eric-heiden (changes requested):
A small fix to the changelog is needed, see earlier thread.

CHANGELOG.md
ooctipus: Reworded in a9c9d036b — dropped the MuJoCo overlay claim; the entry now just states the behavior change: visual shapes and Gaussian splats follow USD viewport semantics and are drawn only for the `default` and `proxy` purposes. Will follow up separately on whether the converter authoring (`purpose=guide` + `collisionEnabled=false`, no separate visual prims) is itself a converter issue.

eric-heiden (dismissed):
Thanks!

eric-heiden (approved):
Thanks!

PR #3406, opened 2026-07-08
Honor ignore_paths in custom-frequency USD traversal

### LINKED ISSUE #3405: parse_usd: custom-frequency traversal does not honor ignore_paths
## Description

`parse_usd` documents `ignore_paths` as "a list of regular expressions matching prim paths to ignore", and every prim traversal in the function honors it — bodies, articulations, joints, shapes, and the actuator pass all skip prims via `any(re.match(pattern, prim_path) for pattern in ignore_paths)` ([example](https://github.com/newton-physics/newton/blob/e5b8573a2b13a5bb964cc441e1e5d697d4452b85/newton/_src/utils/import_usd.py#L4558-L4561)).

The custom-frequency traversal is the one exception: it iterates `Usd.PrimRange(stage.GetPrimAtPath(root_path), Usd.TraverseInstanceProxies())` and consults only each frequency's `usd_prim_filter` — `ignore_paths` is never checked ([block](https://github.com/newton-physics/newton/blob/e5b8573a2b13a5bb964cc441e1e5d697d4452b85/newton/_src/utils/import_usd.py#L4627-L4689)).

## Impact

Two-pass import workflows — import the global stage with env subtrees ignored, then import the env subtrees separately (e.g. Isaac Lab's environment cloning) — get custom-frequency rows registered for prims inside the ignored subtrees. For MuJoCo-schema assets (e.g. Shadow Hand) this produces spurious MJC tendon/actuator rows that misalign the finalized model.

## Repro

1. Create a stage with prims under `/World/envs/...` that match a registered custom frequency's `usd_prim_filter`.
2. `builder.add_usd(stage, ignore_paths=["/World/envs"])`.
3. Observe rows added for prims under the ignored subtree; expected: ignored subtrees contribute no rows.

## Environment

Present on current `main` (e5b8573) and in 1.4.0.dev0. Import-time logic only; hardware-independent.

Related: #769 covers broader `ignore_paths` articulation-tree semantics; this issue is only about the custom-frequency pass skipping the check entirely.

hujc7:
## Description

Fixes #3405.

Every prim traversal in `parse_usd` skips prims matching `ignore_paths` except the custom-frequency pass, which visited ignored subtrees and registered spurious rows for prims excluded from the import. Two-pass import workflows (import the global stage with env subtrees ignored, then import env subtrees separately) ended up with misaligned custom-frequency rows — e.g. MJC tendon/actuator frequencies for MuJoCo-schema assets like the Shadow Hand under Isaac Lab's environment cloning.

The fix applies the same `ignore_paths` check used by the sibling traversals (cf. the actuator pass a few lines above) before invoking the per-frequency filters.

## Testing

- New regression test `test_custom_frequency_honors_ignore_paths`: registers a custom frequency whose filter matches prims both inside and outside an ignored subtree, imports with `ignore_paths=["/World/envs"]`, and asserts only the kept prims contribute rows. Verified it fails without the fix (4 rows instead of 2) and passes with it.
- `pytest newton/tests/test_import_usd.py -k "ignore or custom"`: 18 passed.
- pre-commit clean on both files.

Downstream context: Isaac Lab currently works around this by scoping the `usd_prim_filter` callbacks during its global-stage import (isaac-sim/IsaacLab#6411); that workaround can be dropped once this lands.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed custom-frequency imports to respect `ignore_paths` during prim traversal, so prims under ignored subtrees no longer generate custom-frequency rows or custom attribute values.
  * Improved correctness for two-pass import workflows by preventing spurious entries from ignored branches.
* **Tests**
  * Added a unit test (`test_custom_frequency_honors_ignore_paths`) validating expected custom-frequency counts and attribute lengths when `ignore_paths` is provided.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,7 +79,7 @@
 
 ### Fixed
 
-- Fix `ModelBuilder.add_usd()` to honor `ignore_paths` in the custom-frequency traversal, so prims under ignored subtrees no longer register spurious custom-frequency rows in two-pass import workflows. (#3405)
+- Fix `ModelBuilder.add_usd()` to honor `ignore_paths` in the custom-frequency traversal, so prims under ignored subtrees no longer register spurious custom-frequency rows in two-pass import workflows. (#3406)
 - Fix USD joint `physics:collisionEnabled` import so joints with two explicit bodies honor authored collision behavior; joints to world continue to allow body/world collisions, and articulation-wide self-collision filtering remains additive.
 - Fix `ViewerFile.is_running()` to return `False` after `ViewerFile.close()` so headless recording loops can terminate like interactive viewers. (#3094)
 - Raise an error when `SolverVBD(rigid_contact_history=True)` would allocate or grow contact-history buffers during CUDA graph capture; construct `CollisionPipeline` before `SolverVBD`, or run one uncaptured solver step before capture.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,7 @@
 
 ### Fixed
 
+- Fix `ModelBuilder.add_usd()` to honor `ignore_paths` in the custom-frequency traversal, so prims under ignored subtrees no longer register spurious custom-frequency rows in two-pass import workflows. (#3405)
 - Fix USD joint `physics:collisionEnabled` import so joints with two explicit bodies honor authored collision behavior; joints to world continue to allow body/world collisions, and articulation-wide self-collision filtering remains additive.
 - Fix `ViewerFile.is_running()` to return `False` after `ViewerFile.close()` so headless recording loops can terminate like interactive viewers. (#3094)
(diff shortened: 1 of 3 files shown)


eric-heiden:
Thanks @hujc7, this looks good! Could you also add an entry to the changelog?

hujc7:
> Thanks @hujc7, this looks good! Could you also add an entry to the changelog?

@eric-heiden  Updated if that's sufficient.

CHANGELOG.md
```diff
@@ -79,6 +79,7 @@
 
 ### Fixed
 
+- Fix `ModelBuilder.add_usd()` to honor `ignore_paths` in the custom-frequency traversal, so prims under ignored subtrees no longer register spurious custom-frequency rows in two-pass import workflows. (#3405)
```
eric-heiden: This is the wrong PR number, should be #3406.

hujc7: fixed. Sorry that was odd

eric-heiden (approved):
Thanks!

PR #3410, opened 2026-07-09
Clear all render layers when users change the example in the gui

nvtw:
## Description
Clear all render layers when users change the example in the GUI so stale overlay geometry from the previous example is removed.
## Checklist
## Test plan
```bash
uvx pre-commit run -a
```
## Bug fix
**Steps to reproduce:**
1. Run `python -m newton.examples basic_multi_solver_overlay`.
2. In the OpenGL GUI, switch to another example.
3. Observe that two pendulums from the overlay example remain visible.
**Minimal reproduction:**
```python
import newton

# Reproduce interactively by switching from basic_multi_solver_overlay
# to another example in the OpenGL example browser.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Switching examples and resetting the viewer now fully clears leftover overlay layers, preventing stale content from staying visible.
  * Layered viewer scenes now reset more reliably across viewer modes, returning to the default view state.

* **Tests**
  * Added coverage for `clear_all_layers()` behavior across viewer backends, including verification that only the default layer remains active.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,7 @@
 - Raise an error when `SolverVBD(rigid_contact_history=True)` would allocate or grow contact-history buffers during CUDA graph capture; construct `CollisionPipeline` before `SolverVBD`, or run one uncaptured solver step before capture.
 - Fix `SensorTiledCamera.utils.convert_ray_depth_to_forward_depth()` to preserve the clear-depth sentinel for zero-direction rays and non-positive depths.
 - Fix `ViewerGL.get_frame()` crashing when a CPU model is rendered while a CUDA context is active.
+- Fix stale overlay layers remaining visible after switching examples in the OpenGL viewer.
 - Fix XPBD particle-particle contacts to avoid non-finite particle state for exact-overlap contacts. (#1562)
 - Refer to `kf` consistently as contact friction gain in public documentation. (#2988)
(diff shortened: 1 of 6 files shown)


mzamoramora-nvidia (dismissed):
LGTM. I visually reproduced the stale-layer bug on `main`
and confirmed the PR path is set up to clear the old overlay layers when
switching examples.

Non-blocking suggestions:

- Consider extracting the repeated viewer-clear fallback in
  `newton/examples/__init__.py` into a small `_clear_viewer_scene()` helper.
  Both `switch()` and `reset()` currently repeat the same
  `clear_all_layers()` / `clear_model()` check.
- Consider adding a tiny `_ExampleBrowser.switch()` regression test with a
  stub viewer that implements both `clear_all_layers()` and `clear_model()`.
  The current tests cover the new viewer primitive well, but not the browser
  call site directly.
- Confirm that changing `reset()` to clear all layers is intentional. It seems
  reasonable for layered examples, but the PR description only mentions
  switching examples.
- Nit: the commit subject is a bit long and has no body. Local guidance prefers
  an imperative subject around 50 chars plus a short body explaining what and
  why.

Verification run:

```bash
uv run --extra dev -m newton.tests -k test_example_browser_args
uvx pre-commit run -a
```

All passed.

mzamoramora-nvidia (dismissed):

adenzler-nvidia (dismissed):
LGTM. The split between active-layer `clear_model()` and whole-scene `clear_all_layers()` matches the layered viewer lifecycle. Example reset and switch correctly discard the prior scene before the example constructor rebuilds its layers, while the RTX override safely handles its global renderer teardown. The regression coverage is appropriate.

mzamoramora-nvidia (approved):

Issue #3374, opened 2026-07-07
[REQ] Standardize Solver.reset masks with a global slot


eric-heiden:
### Description

Standardize `Solver.reset(..., world_mask=...)` across solvers on a mask shape of `(world_count + 1,)`:

- Entries `[0, world_count)` select local worlds.
- The final entry selects global entities whose world is `-1`.

Update the public `Solver.reset()` contract and each solver implementation to accept the same format. Solvers that do not support global dynamic objects should still accept the final slot and treat it consistently as a no-op. Where an existing public implementation currently requires `(world_count,)`, make the transition through the normal deprecation path.

PR #3316 should introduce this format for `SolverVBD`, which supports global bodies and joints.

### Motivation / Use Case

A `(world_count,)` mask cannot select global entities. It also makes an explicit all-true mask behave differently from `None`, because only `None` includes globals. A uniform mask contract lets callers select local and global entities independently and keeps reset code portable across solvers.

### Alternatives Considered

Keep `(world_count,)` and reserve `None` for resetting globals. This does not support global-only reset or globals plus a subset of local worlds, and it preserves surprising behavior between `None` and an all-true mask.

PR #3411, opened 2026-07-09
Isolated multi-world implicit MPM

maxkra15:
## Description

  Add isolated multi-world support to `SolverImplicitMPM`.

  This PR builds on the rebuildable sparse-grid support merged in #3412.

  Multi-world models now use an independent FEM environment for each Newton world by default. Each world
  retains its own particle, grid, collider, field, and sparse-solver topology while sharing capacity-
  bounded storage and a single solver invocation.

  This change:

  - Adds `SolverImplicitMPM.Config.separate_worlds`.
  - Packs world-local sparse grids into isolated NanoVDB environments.
  - Builds environment-aware geometry and function-space partitions.
  - Prevents particles and world-local colliders from interacting across worlds.
  - Supports global static and kinematic colliders shared by every world.
  - Validates particle ordering, world ownership, and collider ownership.
  - Adds selective reset through a per-world boolean mask.
  - Clears MPM history and warm starts only for reset worlds.
  - Preserves namespaced MPM state through non-in-place coupled-solver stepping.
  - Adds outer CUDA graph preparation and replay invariants for isolated sparse MPM.
  - Recursively forwards filtered contacts and capture lifecycle calls through nested `SolverCoupled`
  instances.
  - Reports sparse-grid overflow asynchronously across graph replays.
  - Documents isolated MPM grids, capacities, collider semantics, resets, and graph-capture
  requirements.

  ## Checklist

  ## Test plan

  Run the focused multi-world, sparse-grid, and coupled-solver coverage:

  ```bash
      -k TestImplicitMPMMultiworldSparse \
      -k TestImplicitMPMRebuildableSparse \
      -k TestSolverCoupledGraphCapture \
      -k TestImplicitMPM
 ```
  The captured tests require a CUDA device with the Warp memory pool and conditional graph support
  enabled. Coverage includes:

  - Sparse environment-grid construction.
  - Total hierarchy-capacity accounting across worlds.
  - Particle and collider world validation.
  - World-local collider isolation.
  - Global static and kinematic collider behavior.
  - Selective reset of particle history and warm-start fields.
  - Non-in-place coupled MPM reset and replay.
  - Nested coupled-solver capture preparation.
  - Persistent isolated topology during CUDA graph replay.
  - Eager-versus-captured multi-world results.
  - Sticky asynchronous overflow reporting.

  ## New feature / API change
 ```
  import newton
  import warp as wp

  world = newton.ModelBuilder()
  newton.solvers.SolverImplicitMPM.register_custom_attributes(world)
  world.add_particle(
      pos=wp.vec3(0.0, 0.0, 0.0),
      vel=wp.vec3(0.0),
      mass=1.0,
  )

  newton.solvers.SolverImplicitMPM.register_custom_attributes(builder)
  builder.add_world(world)
  builder.add_world(world)

  config = newton.solvers.SolverImplicitMPM.Config(
      separate_worlds=True,
      grid_type="sparse",
      grid_padding=0,
      max_active_cell_count=131072,
      max_leaf_node_count=131072,
      max_lower_node_count=8192,
      max_upper_node_count=512,
      velocity_basis="Q1",
      strain_basis="P0",
      collider_basis="pic27",
      solver="jacobi",
  )

  solver = newton.solvers.SolverImplicitMPM(model, config)
  solver.prepare_graph_capture()

  state_0 = model.state()
  state_1 = model.state()

  with wp.ScopedCapture(device=model.device) as capture:
      solver.step(state_0, state_1, None, None, 1.0 / 60.0)

  solver.check_status()

  # Reset only world 1 while preserving world 0 history.
  world_mask = wp.array([False, True], dtype=wp.bool, device=model.device)
  solver.reset(state_0, world_mask=world_mask)

## Summary by CodeRabbit

* **New Features**
  * Added opt-in isolated multi-world support for implicit MPM, including environment-aware colliders, rendering grains, and rheology.
  * Added rebuildable sparse grids with capacity validation, overflow status reporting, and selective world resets.
  * Added graph-capture capability and preparation/status checking for coupled solvers and implicit MPM (including sparse-grid replay management).

* **Bug Fixes**
  * Improved contact-generation counter handling and more robust empty-strain rheology processing.

* **Documentation**
  * Added comprehensive “Implicit MPM worlds” documentation for isolation, collider filtering, and sparse-grid allocation/graph requirements.

* **Tests & Chores**
  * Expanded multi-world/sparse rebuild/capture test coverage and added a CUDA ASV benchmark.


newton/_src/solvers/implicit_mpm/implicit_mpm_solver_kernels.py
```diff
@@ -896,7 +990,222 @@ def positive_modn(x: int, n: int):
...
+        min_x = int(np.min(environment_ijks[:, 0]))
+        max_x = int(np.max(environment_ijks[:, 0]))
+        environment_ijks = environment_ijks.copy()
+        environment_ijks[:, 0] += packed_start - min_x
+        packed_ijks.append(environment_ijks)
+        packed_start += max_x - min_x + 1 + guard_cells
+
+    if not packed_ijks:
+        return np.empty((0, 3), dtype=np.int64)
+    return np.unique(np.concatenate(packed_ijks), axis=0)
```
gdaviet: Even if running only at init time this does not look super scalable. Maybe using `nanogrid.from_environment_voxels()` and looking at the result would be faster

newton/_src/solvers/implicit_mpm/rasterized_collisions.py
```diff
@@ -159,72 +174,115 @@ def get_average_face_normal(
...
+        local_count = collider.query_world_offsets[environment_index + 2] - local_begin
+        query_count = global_count + local_count
+
+    for query_offset in range(query_count):
+        stable_collider_id = query_offset
+        if not shared_worlds:
+            if query_offset < global_count:
+                stable_collider_id = collider.query_collider_ids[global_begin + query_offset]
+            else:
+                stable_collider_id = collider.query_collider_ids[local_begin + query_offset - global_count]
```
gdaviet, line 266: I think this could be simplified a little bit, prepopulating just two `world_collider_offsets` (shape world_count+1) and `world_collider_ids`. Then  in each world we just have iteratore over collider ids in `world_collider_offsets[env_index]..world_collider_offsets[env_index+1]`

newton/_src/solvers/implicit_mpm/rheology_solver_kernels.py
```diff
@@ -705,15 +710,21 @@ def apply_stress_delta_impl(
...
+                for bk in range(strain_velocity_node_count):
+                    b = block_beg + bk
+                    u_i = strain_mat_columns[b]
+                    delta_u = inv_mass_matrix[u_i] * _symmetric_part_transposed_op(strain_mat_values[b], delta_stress)
+                    velocities[u_i] += delta_u
+            else:
+                for b in range(block_beg, block_end):
+                    u_i = strain_mat_columns[b]
+                    delta_u = inv_mass_matrix[u_i] * _symmetric_part_transposed_op(strain_mat_values[b], delta_stress)
+                    velocities[u_i] += delta_u
```
gdaviet: This is duplicating the  strain_velocity_node_count<=0 path, and the additional check probably nullifies perf gains from the specialized fixed-size path. I would recommend just calling the kernel with strain_node_count <= 0

newton/_src/solvers/implicit_mpm/solve_rheology.py
```diff
@@ -1574,12 +1690,41 @@ def release(self):
         super().release()
 
 
+def _strain_environment_node_counts(rheology: RheologyData) -> tuple[int, ...]:
+    if rheology.strain_environment_offsets is None:
+        return (rheology.stress.shape[0],)
+
+    offsets = rheology.strain_environment_offsets.numpy()
+    return tuple(int(offsets[i + 1] - offsets[i]) for i in range(offsets.shape[0] - 1))
```
gdaviet: this function and the other two below could be vectorized

newton/_src/solvers/implicit_mpm/solve_rheology.py
```diff
@@ -1745,6 +1891,11 @@ def solve_rheology(
             )
 
     if solvers[0] in _ITERATIVE_LINEAR_SOLVERS:
+        strain_node_counts = _strain_environment_node_counts(rheology)
+        # Warp's batched Krylov solvers accept one absolute tolerance for all
+        # batches. Scale by the smallest populated batch so no environment gets
+        # a looser stopping criterion than it would in an independent solve.
+        tolerance_scale = _linear_solver_tolerance_scale(strain_node_counts)
```
gdaviet: The issue is that it will inconditionally break graph capture of the linear solver path, we should avoid that. Let the user adjust the tolerance, for homogeneous batches per-env scaling should not matter much

gdaviet (commented):
Looks great!  (I just browsed through the MR, no agent review yet)
Please check that there is no significant time regression for the single-world path (unfortunately we don't have ASV benchmarks yet for the mpm solve)

Also I think it would make sense to mark the multi-world parameters as experimental for now, so we can iterate if required

newton/_src/solvers/implicit_mpm/implicit_mpm_solver_kernels.py
maxkra15: Done. Removed the Python per-environment packing loop. Capacity estimation now builds a temporary `Nanogrid.from_environment_voxels()` and inspects its packed cell grid.

newton/_src/solvers/implicit_mpm/rasterized_collisions.py
maxkra15: Done. Collider lookup now uses `world_collider_offsets` with shape `world_count + 1` and prepopulated `world_collider_ids`; each isolated query iterates one contiguous range.

newton/_src/solvers/implicit_mpm/rheology_solver_kernels.py
maxkra15: Done. Restored the specialized fixed-size kernel without the runtime row-length branch. Isolated multiworld solves pass a non-positive strain node count to the rheology kernels, while the single-world fast path stays specialized. The added ASV benchmark measured 60.67 ms on the sparse parent and 60.71 ms here for ten steps on an RTX 4090.

newton/_src/solvers/implicit_mpm/solve_rheology.py
maxkra15: Fixed by removing the three host-side per-environment helpers. Batched reporting now uses vectorized maximum reductions, while convergence remains device-side.

newton/_src/solvers/implicit_mpm/solve_rheology.py
maxkra15: Agreed. Removed per-environment tolerance scaling and its offset host read. All batches use the existing uniform total-size scale, so this no longer breaks graph capture in the linear solver path.

gdaviet:
A few more agent comments:  


  - [P1] separate_worlds=True is an undeclared breaking default change. solver_implicit_mpm.py:874
    (https://github.com/newton-physics/newton/blob/011818f16351624265f824fa9676e83d1a2ec1e4/newton/_src/solvers/implicit_mpm/solver_implicit_mpm.py#L874) changes
    existing multi-world simulations from shared topology to isolated topology and makes models containing global particles fail during construction. Newton’s
    compatibility policy requires a deprecation cycle before changing established behavior. This release should default to False, document True as the new opt-in,
    and flip the default only after deprecation.

  - [P2] Masked reset leaves grid-backed warm starts active in reset worlds. solver_implicit_mpm.py:1510
    (https://github.com/newton-physics/newton/blob/011818f16351624265f824fa9676e83d1a2ec1e4/newton/_src/solvers/implicit_mpm/solver_implicit_mpm.py#L1510) only
    clears masked warm starts for PointBasisSpace; fixed, dense, and allocating-sparse grid fields are silently preserved. Those stress and collider-impulse fields
    are consumed on the next step, so a reset world can inherit pre-reset solver state and produce different results with finite iteration counts. Clear selected
    environment ranges using the environment offsets, or reject masked reset when selective clearing is unavailable. The current test asserting that grid warm
    starts remain unchanged codifies the problematic behavior.

pyproject.toml
```diff
@@ -28,7 +28,7 @@ description = "A GPU-accelerated physics engine for robotics simulation"
 readme = "README.md"
 requires-python = ">=3.10"
 dependencies = [
-    "warp-lang>=1.15.0.dev20260626",
+    "warp-lang>=1.16.0.dev0",
```
eric-heiden: 🟠 `1.16.0.dev0` permits daily builds that predate the NanoVDB rebuild APIs used by this PR. The lockfile currently masks this because it resolves to a newer build, but a valid installation can select an earlier `1.16` development package and fail because those APIs are absent. The implementation landed on 2026-07-07, as shown in the [upstream commit](https://github.com/NVIDIA/warp/commit/ed6cd7e2b4dd76ed59d75258f2e38ea86c7123e4).

Please require the first known compatible build and update `uv.lock` consistently:

```suggestion
    "warp-lang>=1.16.0.dev20260708",
```

newton/_src/solvers/implicit_mpm/solve_rheology.py
```diff
@@ -151,9 +218,12 @@ def update_condition(
     condition: wp.array[int],
 ):
     cur_it = iteration[0] + solve_granularity
-    stop = (
-        residual[0, 0] < residual_threshold * l2_scale and residual[1, 0] < residual_threshold
-    ) or cur_it > max_iterations
+    converged = bool(True)
+    for batch in range(residual.shape[1]):
+        converged = converged and residual[0, batch] < residual_threshold * l2_scale
+        converged = converged and residual[1, batch] < residual_threshold
```
eric-heiden, line 253: 🟠 Each batch represents an independent world, but every residual is compared against the same global `l2_scale`, which is computed from the total Delassus operator size. In uneven or many-world simulations, this lets a small world converge under a looser tolerance than it would receive when solved alone.

Please compute an L2 scale for each range in `strain_environment_offsets` and compare each batch against its own threshold. The linear-solver tolerance and `_nonlinear_solver_result_norms()` reporting should use the same batch-local scaling so isolated and standalone solves have equivalent stopping criteria.

newton/_src/solvers/implicit_mpm/implicit_mpm_model.py
```diff
@@ -461,79 +467,250 @@ def setup_collider(
...
         if body_q is None:
             body_q = model.body_q
 
+        effective_body_mass = body_mass.numpy().copy()
+        if not body_mass_is_explicit and model.body_flags is not None:
+            body_flags = model.body_flags.numpy()
+            kinematic_bodies = (body_flags & int(newton.BodyFlags.KINEMATIC)) != 0
+            if np.any(kinematic_bodies & (effective_body_mass != 0.0)):
+                effective_body_mass[kinematic_bodies] = 0.0
+                body_mass = wp.array(effective_body_mass, dtype=body_mass.dtype, device=body_mass.device)
```
eric-heiden, line 696: When `body_mass` is omitted, this now forces every body with `BodyFlags.KINEMATIC` to have zero effective mass. Previously, `setup_collider()` used `model.body_mass` unchanged and callers explicitly passed zero masses when they wanted kinematic treatment.

I think this change makes sense to me, we should treat kinematic bodies as such and set masses to zero internally if needed. I think a changelog entry might be good since now users don't have to manually set body masses to zero anymore on the kinematic bodies in `model.body_mass`?

eric-heiden:
🟠 Could we reduce and device-neutralize the new public graph/status API before merging? `prepare_cuda_graph_capture()` is not CUDA-specific now that Warp also supports CPU graphs, so please expose `prepare_graph_capture()` instead, and likewise use `supports_graph_capture` for the capability property. `check_sparse_grid_rebuild_status()` and `clear_sparse_grid_rebuild_status()` expose NanoVDB and MPM implementation details; please make them private and keep `check_status()` as the single generic post-replay error check. This leaves callers with a small solver-level contract while allowing the internal sparse-grid status handling to evolve without another public API migration.

gdaviet (dismissed):

newton/_src/solvers/implicit_mpm/solver_implicit_mpm.py
```diff
@@ -964,6 +1201,11 @@ def __init__(
         else:
             if config.warmstart_mode not in ("particles", "grid", "smoothed"):
                 raise ValueError(f"Invalid warmstart mode: {config.warmstart_mode}")
+            if self._sparse_rebuildable and config.warmstart_mode == "grid":
```
adenzler-nvidia: Was it intentional that the default `warmstart_mode="auto"` still resolves to `"grid"` for rebuildable sparse grids? This check only rejects the explicit `"grid"` case, but `"auto"` with the default `strain_basis="P0"` resolves to `"grid"` a few lines above, and `_warmstart_fields` then skips the grid-to-grid transfers for rebuildable grids and zeroes both the impulse and stress warm starts every step. So a rebuildable sparse solver built with default settings runs every solve cold, and with capped `max_iterations` that gives different results than the same config without a capacity, without any warning.

If unintentional, resolving `"auto"` to `"particles"` when `self._sparse_rebuildable` (and treating `"smoothed"` like `"grid"` here) would preserve warm starting.

newton/_src/solvers/coupled/solver_coupled.py
```diff
@@ -3070,6 +3127,8 @@ def _prepare_filtered_contact_update_kernel(
             rigid_generation[0] = src_generation[0]
         if soft_update != 0:
             soft_generation[0] = src_generation[0]
+        if rigid_update != 0 or soft_update != 0:
+            dst_generation[0] += wp.int32(1)
```
adenzler-nvidia: Is this a leftover from a bad merge? The same condition already increments `dst_generation` a few lines above (with the int32 wraparound guard), so each contact update now bumps the generation twice, and this second increment bypasses the wraparound guard. The consumers only test the counter for change, so it looks like these two lines can simply be removed.

newton/_src/solvers/implicit_mpm/implicit_mpm_solver_kernels.py
```diff
@@ -881,6 +883,120 @@ def clamp_coordinates(
...
+    body_q: wp.array[wp.transform],
+    body_q_prev: wp.array[wp.transform],
+):
+    """Refresh previous collider poses for bodies in selected worlds."""
+    body_index = wp.tid()
+    world = body_world[body_index]
+    selected = False
+    if world >= 0 and world < world_mask.shape[0]:
+        selected = world_mask[world]
+    elif world < 0 and world_mask.shape[0] == 1:
```
adenzler-nvidia: The masked reset kernels (`reset_mpm_particle_history`, `reset_mpm_collider_history`, `reset_mpm_point_warmstart`) only select global (`world == -1`) entities when `world_mask.shape[0] == 1`, so an all-True mask over a multi-world model is not equivalent to `world_mask=None`. Example: a global kinematic collider is teleported and then all worlds are reset with an all-True mask; its `body_q_prev` entry is never refreshed, and with `collider_velocity_mode="backward"` the next step computes a phantom collider velocity from the stale pose that affects every world.

Consider selecting `world < 0` entities whenever any world is selected, or documenting the asymmetry in `reset()`.

adenzler-nvidia:
@eric-heiden could you weigh in on the graph-capture API shape here? This PR introduces `supports_graph_capture` / `prepare_graph_capture()` / `check_status()` on `SolverImplicitMPM` and `SolverCoupled` as a de-facto solver capture protocol (documented in the new worlds.rst section), but the protocol is not declared on `SolverBase`: `SolverCoupled` discovers the methods via `getattr`, and a missing `supports_graph_capture` defaults to `True`, so entry solvers that never considered capture are reported capture-safe, and implementations cannot use `@override`. Since this sets a precedent for public solver API, we should double-check the shape before it spreads — e.g. declare the trio on `SolverBase` with explicit defaults (no-op `prepare_graph_capture()`/`check_status()` and an explicit `supports_graph_capture` default).

newton/_src/solvers/implicit_mpm/solver_implicit_mpm.py
maxkra15: yes this was unintentional. We now resolve warmstart_mode="auto" to "particles" for rebuildable sparse grids, while preserving the existing particle-backed behavior for P1d/Q1d strain bases. Explicit "grid" and "smoothed" modes are both rejected for rebuildable grids

newton/_src/solvers/coupled/solver_coupled.py
maxkra15: this was a leftover duplicate increment. removed the second update, leaving the existing guarded increment as the single generation update. This preserves the int32 wraparound behavior

newton/_src/solvers/implicit_mpm/implicit_mpm_solver_kernels.py
maxkra15: the three masked reset kernels now use a shared selection helper where local entities follow their world’s mask entry, while global (world < 0) entities are selected whenever at least one world is selected. An all-false mask leaves global history unchanged. added coverage using a global kinematic collider in backward-velocity mode for full, partial, and empty masks

newton/_src/solvers/implicit_mpm/solver_implicit_mpm.py
```diff
@@ -1236,14 +1385,267 @@ def check_status(self) -> None:
...
+        cleared for selected isolated worlds; shared multi-world grids reject
+        masked reset because their warm-start nodes may combine worlds. A full
+        reset clears every warm-start field. Sparse-grid rebuild status is
+        always cleared at a valid reset boundary, and the previous-collider-pose
+        cache is refreshed from ``state``. Global particle-backed history and
+        collider poses are reset whenever the mask selects at least one world.
+
+        Args:
+            state: Simulation state whose MPM history is modified in place.
+            world_mask: Optional boolean mask of shape ``(model.world_count,)``
```
eric-heiden: Can you implement the global slot semantics from https://github.com/newton-physics/newton/issues/3374 in this reset method as well?
So the size of `world_mask` should be `model.world_count + 1` with the last element selecting the global world (particles from world -1).

eric-heiden:
@adenzler-nvidia I agree this should not become a generic solver protocol. The coupled solver can work without these methods. Please remove `supports_graph_capture`, `prepare_graph_capture()`, and `check_status()` from `SolverCoupled`, including the `getattr` forwarding.

For MPM, please keep capture setup internal instead of exposing `supports_graph_capture` and `prepare_graph_capture()`. Only `check_status()` has a concrete need because sparse-grid overflow cannot be raised during graph replay. That can remain specific to `SolverImplicitMPM` and be called directly. Please do not add the full trio to `SolverBase`.

maxkra15:
- SolverCoupled: all three methods and getattr forwarding removed
- SolverImplicitMPM: check_status() remains
- Capture resources are initialized internally
- Coupled tests call the concrete MPM solver’s check_status() directly

docs/concepts/worlds.rst
```diff
@@ -266,6 +266,166 @@ While :meth:`~newton.ModelBuilder.begin_world` and :meth:`~newton.ModelBuilder.e
    world_count: 4
 
 
+.. _implicit-mpm-worlds:
+
+Implicit MPM worlds
```
eric-heiden: I think a lot of this is very specific to SolverImplicitMPM, so perhaps a better place for these docs is in the docstring for this solver? I would also try to condense this a bit.

eric-heiden (approved):

PR #3413, opened 2026-07-09
Use velocity instead of velocity error as input to actuator nets

jvonmuralt:
## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the neural MLP and LSTM controllers to feed raw joint velocity (and velocity history where applicable) to the networks instead of velocity error, matching the expected training inputs.
  * Adjusted network feature assembly so inputs follow the `(pos_error, vel)` convention.
* **Documentation**
  * Updated controller descriptions and validation messaging to reflect `(pos_error, vel)` instead of `(pos_error, vel_error)`.
* **Tests**
  * Added a unit test verifying controller outputs depend on raw joint velocity and not on `target_qd`.


camevor (dismissed):
Thanks, lgtm!

Changes look good, although a regression test with a very simple MLP would be nice to have.

camevor (approved):

Issue #3181, opened 2026-06-15
[REQ] BatchedModelBuilder for cloning when cloned env is known


ooctipus:
### Description

Current API either only support homogeneous cloning or requires for loop for heterogeneous environment.

Say we know the 3 worlds looks like below

World 0: banana, franka
World 1: banana, franka, franka
World 2: banana, franka, franka, contact_sensor, contact_sensor

```
builders = [FrankaBuilder, BananaBuilder, ContactSensorBuilder]
builder_world_indice: wp.array = 10|110|11022
world_start_indices: wp.array  = 025
```

Requesting a BatchedModelBuilder that digest above structure and greatly speed up the cloning time for heterogeneous environment. 

### Motivation / Use Case

We have fast api for homogeneous replication but not for heterogeneous case, and in IsaacLab start up time, heterogeneous cloning can take majority of the time due to cloning is done with for loop and string, label matching on fully cloned environemts.

PR #3415, opened 2026-07-09
Batch ModelBuilder replication

adenzler-nvidia:
## Description

Batch `ModelBuilder.replicate()` through the same internal merge implementation used by `add_builder()`, instead of invoking the scalar `add_world()`/`add_builder()` composition path once per copy.

The shared merge path:

- Expands homogeneous builder copies in one batched pass.
- Keeps index remapping and field handling consistent between `add_builder()` and `replicate()`.
- Preserves the separate-world semantics of `replicate()`.
- Validates compatibility of the source builder up front, failing fast with a clear error.
- Preserves muscles and rigid-body color groups.
- Rejects incompatible custom attribute and custom frequency definitions.
- Classifies builder-owned lists so newly added fields must declare a merge strategy.

There is no public API change.

For 8,192 G1 worlds, the representative warmed benchmark improved from approximately 5.42 seconds on `main` to 2.14 seconds on this branch, or roughly 2.5× faster.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_builder_replicate
uv run --extra dev -m newton.tests -k test_custom_attributes
uvx pre-commit run -a
```

The regression coverage verifies that:

- Batched replication matches repeated `add_world()` composition.
- `replicate()` does not dispatch through scalar `add_world()` calls.
- Compatibility validation runs before the merge starts.
- Custom schemas and callbacks must be compatible.
- `body_q` entries are freshly owned, not aliased with the source or across replicas.
- Muscles, body colors, actuators, labels, and entity references are preserved.
- Every builder-owned list has an explicit merge strategy.

## Bug fix

Before this change, builder composition could silently accept incompatible custom definitions, and some builder state such as muscles and rigid-body color groups was not copied.

These cases are now validated and covered by focused regression tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved multi-world `ModelBuilder` replication by batching copies into a single merge pipeline, reducing overhead.
  * Preserved muscle structure and rigid-body color groups during copy/replication.
  * Strengthened compatibility checks for custom attributes and custom frequencies to prevent invalid merges.
  * Corrected joint/articulation start-index metadata to use the proper reference domains for compaction.
* **Tests**
  * Added extensive coverage for `replicate()` and `add_world()` composition, including transform isolation, finalization correctness, array-valued fields, zero-spacing behavior, and validation failure non-mutation.


newton/_src/sim/builder.py
```diff
@@ -2490,10 +2656,426 @@ def replicate(
...
+                    body_q[:, :3] += np.repeat(offsets, counts["body"], axis=0)
+                # Own each transform without retaining per-row views into the tiled array.
+                self.body_q.extend(wp.transform.from_buffer_copy(row) for row in body_q)
+            else:
+                for xform in xforms:
+                    if xform is None:
+                        self.body_q.extend(builder.body_q)
+                    else:
+                        self.body_q.extend(transform_mul(xform, body_q) for body_q in builder.body_q)
+
```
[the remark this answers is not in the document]
adenzler-nvidia: Fixed for `body_q`: the no-transform path now produces owned copies like the transformed branches, with a regression test. `shape_transform`/`joint_X_p` (and other mutable-element lists like `body_com`) keep the previous `add_builder` reference-sharing behavior on purpose — copying them on the batched path costs ~0.3s per 500k transforms, a large chunk of this PR's win. Happy to do a full owned-copy pass as a follow-up if we want that contract everywhere.

newton/_src/sim/builder.py
```diff
@@ -3917,125 +3963,96 @@ def transform_equality_target_value(entity_idx: int, value: Any) -> Any:
...
-                # Handle integers, preserving negative sentinels (e.g., -1 means "invalid")
-                if isinstance(v, int):
-                    return v + offset if v >= 0 else v
-                # Handle list/tuple explicitly, preserving negative sentinels in elements
-                if isinstance(v, (list, tuple)):
-                    transformed = [x + offset if isinstance(x, int) and x >= 0 else x for x in v]
-                    return type(v)(transformed)
-                # For other types (numpy, warp, etc.), try arithmetic offset
+                    return value
+                if isinstance(value, int):
```
StafaH: This only recognizes Python int. NumPy integer scalars and Warp integer scalars are not necessarily instances of int. Not sure if this causes any specific issue for the current model builder but could lead to some silent errors.

newton/_src/sim/builder.py
```diff
@@ -2517,10 +2585,385 @@ def replicate(
...
+
+        transform_mul_cfunc = wp._src.context.runtime.core.wp_builtin_mul_transformf_transformf
+
+        def transform_mul(a: wp.transform, b: wp.transform) -> wp.transform:
+            out = wp.transform.from_buffer(np.empty(7, dtype=np.float32))
+            transform_mul_cfunc(a, b, ctypes.byref(out))
+            return out
+
+        counts = self._builder_merge_counts(builder)
+        self._validate_builder_merge(builder, set(counts))
```
StafaH, line 2701: `add_world() `resolves the metadata and validates the merge before opening the world, but `add_builder() `immediately repeats the same work inside `_merge_builder_copies()`.

StafaH (dismissed):
LGTM

newton/_src/sim/builder.py
adenzler-nvidia: Fixed in 5d0c120eb — the sentinel guard now also recognizes NumPy and Warp integer scalars, so a negative sentinel like `wp.int32(-1)` no longer falls through to the arithmetic fallback and gets offset into a bogus reference. Added a regression test.

newton/_src/sim/builder.py
adenzler-nvidia: Fixed in 893af74ef — dropped the pre-validation from `add_world()` and made it roll back `begin_world()` if the merge fails instead. Validation now runs once inside `_merge_builder_copies()`, and a failed merge still can't leave a world context open (now for any error, not just validation failures).

StafaH (dismissed):

newton/_src/sim/builder.py
```diff
@@ -3472,420 +3918,30 @@ def add_world(
             for i in range(3):
                 scene.add_world(robot)  # Each robot is a separate world
         """
+        xform = None if xform is None else wp.transform(*xform)
         self.begin_world()
-        self.add_builder(builder, xform=xform, label_prefix=label_prefix)
+        try:
+            self.add_builder(builder, xform=xform, label_prefix=label_prefix)
+        except BaseException:
```
camevor, line 3970: Question: Why do we promise atomicity or even a valid destination builder state in the first place?
- Atomicity is given only for validation but broken when an error happens past validation, making this a weak promise.
- In most cases, builder construction is programmatic and unlikely to recover gracefully from an error in the first place. The actual destination builder object is also unlikely to contain much valuable work.

The upfront validation is a good choice, but I recommend dropping the promise.

newton/_src/sim/builder.py
```diff
@@ -2558,11 +2626,389 @@ def replicate(
...
+            particle_q = np.tile(np.asarray(builder.particle_q, dtype=np.float32), (world_count, 1))
+            particle_q += np.repeat(offsets, counts["particle"], axis=0)
+            self.particle_q.extend(particle_q.tolist())
+
+        shape_starts = starts("shape")
+        body_starts = starts("body")
+
+        attribute_specs.pop("shape_transform")
+        shape_transform_start = len(self.shape_transform)
+        self.shape_transform.extend(source_list("shape_transform") * world_count)
```
camevor, line 2741: Aliasing the same transforms here will lead to potentially unexpected results when the transforms of either builder are modified after replication. The supported builder life-cycle is worth documenting, although this is not newly introduced by this PR.

newton/_src/sim/builder.py
```diff
@@ -4084,14 +4106,59 @@ def transform_value(v, offset=value_offset, replace_with_world=use_current_world
...
-                entry.indices.append(idx + start_joint_dof_idx)
-            for idx in sub_entry.pos_indices:
-                entry.pos_indices.append(idx + start_joint_coord_idx)
+            entry.indices.extend(idx + joint_dof_offset for idx in sub_entry.indices)
+            entry.pos_indices.extend(idx + joint_coord_offset for idx in sub_entry.pos_indices)
             entry.controller_args.extend(sub_entry.controller_args)
             entry.delay_args.extend(sub_entry.delay_args)
             entry.clamping_args.extend(sub_entry.clamping_args)
 
+    def add_builder(
```
camevor, line 4173: Using a single, batched path for scalar `add_builder` and `replicate` is a good design. Two caveats worth mentioning:
1. Using the batch-optimized path for `add_builder` incurs a significant perf regression for repeated `add_builder` calls. This is especially important for heterogeneous workflows, where `add_builder` is typically called several times per world. There are a few things that could be optimized for the single `add_builder` path.
2. Attribute ownership across merges is now inconsistent and worth documenting: `body_q` is newly copied (previously aliased), identity merges no longer produce fresh copies for static shapes and root joints, and everything else (`shape_transform, joint_X_p, body_com`, …) remains aliased as before. Documenting the supported builder life-cycle would settle which of these is intended.

newton/_src/sim/builder.py
```diff
@@ -2558,11 +2626,389 @@ def replicate(
...
+        self.shape_transform.extend(source_list("shape_transform") * world_count)
+        if counts["shape"]:
+            static_shapes = np.flatnonzero(np.asarray(builder.shape_body, dtype=np.int64) == -1)
+            for world_index, xform in enumerate(xforms):
+                if xform is None:
+                    continue
+                for shape in static_shapes.tolist():
+                    source = builder.shape_transform[shape]
+                    target = shape_transform_start + world_index * counts["shape"] + shape
+                    self.shape_transform[target] = transform_mul(xform, source)
```
camevor, line 2750: `source_list()` tolerates array-valued fields, and `test_merge_accepts_array_valued_fields` pins that for `joint_q/particle_qd`, but the transform paths bypass it: this line (and the `joint_X_p` equivalent at L2770, and `body_q` under a rotated xform) index the source field directly and pass an ndarray row to
`transform_mul`, which raises `ctypes.ArgumentError`. Since `shape_transform` was already extended at L2732, the destination is left partially mutated.

newton/_src/sim/builder.py
```diff
@@ -2558,11 +2626,389 @@ def replicate(
...
+
+        for attr, spec in attribute_specs.items():
+            source = source_list(attr)
+            destination = getattr(self, attr)
+            if spec.compaction_policy in {"world_start", "passthrough"}:
+                continue
+            if spec.compaction_policy == "color_groups":
+                kind = self._builder_frequency_key(spec.frequency)
+                for start in starts(kind).tolist():
+                    translated = [group + start for group in source]
```
camevor: `group + start` assumes each color group is an `ndarray`. A plain Python list concatenates instead of offsetting and raises.

newton/_src/sim/builder.py
```diff
@@ -2558,11 +2626,389 @@ def replicate(
...
+                    destination = combine_independent_particle_coloring(destination, translated)
```
camevor: `combine_independent_particle_coloring` is called once per world with the ever-growing destination, and it re-sorts and `np.concatenate`s all previously accumulated groups on every call. Replicating a colored (cloth/soft/body-colored) builder is O(world_count²).

newton/_src/sim/builder.py
```diff
@@ -2558,11 +2626,389 @@ def replicate(
...
+            for suffix, reference in references.items():
+                name = f"_{group}_{suffix}"
+                specs[name] = Model.AttributeSpec(
+                    group,
+                    references=reference,
+                    compaction_policy="start" if suffix.endswith("_start") else "end",
+                )
+                declared_builder_attributes.add(name)
+
+        stale = declared_builder_attributes.difference(base_attributes)
```
camevor, line 2975: This looks like it guards against schema drift on the default builder attributes. While this is a reasonable check, I would recommend relocating it into a test (`test_all_builder_lists_have_merge_metadata`) since it's not something that changes at runtime.

camevor (commented):
Thanks, this looks very good! Added a few comments.

The main concern is that the current builder merging logic is geared toward homogeneous replication and increases the cost of scalar `add_builder`.

A quick benchmark on a heterogeneous scene (8000 worlds × (G1 + cube), 16,000 add_builder calls) indicates a > 3x regression to 15 s from 4.7s on main. Since we don't have a heterogeneous cloning path as requested in #3181 yet, there are a few things we could do to reduce this overhead:
1. Drop the runtime validation of the base class attributes (schema validation at runtime), and cache the built base attribute spec dict on a class level.
2. Bulk-copy custom-attribute values that need no remap.
3. Skip empty source lists in the generic merge loop before any branch dispatch: for a typical rigid-body source, most of the ~140 list attributes (particles, springs, tets, muscles, cloth, …) are empty yet still pay string checks, count lookups, and empty extends per merge.

newton/_src/sim/builder.py
adenzler-nvidia: Dropped the promise — the PR description no longer claims anything about destination state after a failure; the upfront validation stays as a fail-fast mechanism.

One note on the try/except here: it serves a narrower purpose than atomicity. Without the rollback, a failed merge leaves the world context open (`current_world` set, `world_count` incremented), so every subsequent `add_world`/`replicate` on that builder raises "already in world context", burying the original error. It keeps the world bookkeeping consistent, not the merged data.

newton/_src/sim/builder.py
adenzler-nvidia: Documented the life-cycle in the `add_builder`/`replicate` docstrings (1ce61ef35): the source acts as a template — merged entries may be shared with the source and across replicas; mutating entry objects in place afterwards is unsupported, slot reassignment is safe (the only pattern builder-internal code uses).

We measured copying as the alternative: even the cheap path (numpy tile + `from_buffer_copy`, as `body_q` does) costs ~0.56 s per 820k transforms vs 0.007 s for aliasing — about a quarter of the whole 8192-world G1 replicate for `shape_transform` alone, before extending the same treatment to `joint_X_p`/`body_com`/… for consistency. Since main aliases identically (with the source, and across worlds when the same source is added repeatedly), we kept aliasing and documented the contract instead.

newton/_src/sim/builder.py
adenzler-nvidia: Fixed in 761bc9b85 — the static-shape, root-joint, and rotated `body_q` paths now normalize rows with `wp.transform(*...)` before the native multiply, and a regression test covers array-valued `shape_transform`/`joint_X_p`/`body_q` under a rotated xform. (Main's scalar path fails the same way — its `transform_mul` is the same ctypes call fed raw entries — but this PR pins the tolerance contract, so all paths honor it now.) On the partial mutation: per the discussion above, we make no claim about destination state after post-validation errors.

newton/_src/sim/builder.py
adenzler-nvidia: Fixed in 4d97c2358 — each color now unions across worlds (copies are disjoint) and `combine_independent_particle_coloring` runs once, making replication O(total). 8192 worlds × 300 colored particles: 9.7 s → 1.0 s, linear growth. One behavioral note: the result is the per-color union rather than the progressively re-balanced mix the sequential loop produced — both are valid independent colorings, and the union preserves the source's group-size ratios.

newton/_src/sim/builder.py
adenzler-nvidia: Fixed in 4d97c2358 as part of the coloring rewrite (`np.asarray` per group), with a regression test covering a plain-list group. This was main-parity — the scalar path has the same expression and `set_coloring()` normalizes to ndarrays, so plain lists only arrive via direct assignment — but the merge tolerates array-valued fields elsewhere, so it does here now too.

adenzler-nvidia:
Implemented all three suggestions in 5bea32888, plus what profiling surfaced on top:

- Merge attribute specs are `functools.cache`d per (class, `use_coord_layout_targets`); schema validation now runs once per process (see the thread at the schema check for the one twist vs. moving it to a test).
- Custom-attribute values with no remap bulk-copy, and the transform closures are not even built in that case.
- Empty source lists skip before branch dispatch.
- Identity short-circuit in `_custom_attribute_defaults_match` — repeated merges compare identical default objects, and Warp value-type `__eq__` was the single largest cost in your benchmark shape.
- Identity constants hoisted to module level and free-root frames hoisted out of the per-world loop (each merge paid Warp builtin overload dispatch), a `world_count == 1` fast path in the reference remapping, and cheaper count/frequency-key lookups.

Your benchmark shape on one machine (8000 worlds × (G1 MJCF + cube), 16k `add_builder` calls): main 6.0 s, PR before 13.2 s, PR now 8.9 s. The remaining gap is the per-merge custom-attribute compatibility validation (which you rightly suggested keeping) and the generic per-attribute loop body — agreed that closing it fully is the #3181 heterogeneous cloning path. Homogeneous `replicate` also gained: 8192 G1 worlds went 3.7 s → 2.5 s.

newton/_src/sim/builder.py
adenzler-nvidia: Done in 5bea32888, with one twist: rather than relocating the check into a test, the built spec dict (validation included) is now `functools.cache`d per (class, `use_coord_layout_targets`), so the check runs once per process — same cost as dropping it, but a `ModelBuilder` subclass that misdeclares merge metadata still gets a loud `RuntimeError` instead of silently losing that attribute in merges (the returned dict filters to base list attributes, so silent drop is the alternative failure mode, and `test_all_builder_lists_have_merge_metadata` only covers the base class). That existing test keeps pinning the validation via the uncached builder.

newton/_src/sim/builder.py
adenzler-nvidia: Both caveats are addressed now: the perf side via 5bea32888 (details and numbers in the top-level reply — your benchmark shape went 13.2 s → 8.9 s vs 6.0 s main on one machine), and the ownership side via the life-cycle note added to the `add_builder`/`replicate` docstrings in 1ce61ef35: the source acts as a template, merged entries may be shared with it and across copies, in-place mutation of entries is unsupported, slot reassignment is safe. Under that contract, whether a given attribute is copied (`body_q`) or aliased (the rest) is an implementation detail rather than an inconsistency.

newton/_src/sim/builder.py
camevor: Makes sense!

newton/_src/sim/builder.py
adenzler-nvidia: Follow-up in 616f87946, superseding the union approach above: the pairing logic of `combine_independent_particle_coloring` is now extracted into `combine_independent_coloring_plan` (operating on (size, chunks) tuples), the array combine is a thin materialization over it, and the batched path runs the plan per world and materializes once. Replication now produces byte-identical composition to the sequential `add_world` loop — balancing included — while staying O(total): 8192 worlds × 300 colored particles runs in 1.2 s vs 9.7 s for the per-world combine. The union variant silently diverged from the equivalence contract for sources with uneven group sizes; `test_replicate_matches_add_world_loop` now covers that case.

newton/_src/sim/builder.py
adenzler-nvidia: One amendment in c969ccc33: the cache moved from `functools.cache` to storage in `cls.__dict__` so dynamically created `ModelBuilder` subclasses stay collectable (`functools.cache` would pin every class it ever saw for the process lifetime), and the target-layout flag is passed into the schema build explicitly instead of re-read from the global. Behavior is otherwise as described above.

camevor (approved):
All review comments have been addressed, lgtm!

PR #3416, opened 2026-07-09
Promote mujoco KPI benchmarks to the Fast suite

adenzler-nvidia:
## Description

Rename the seven `Kpi*` mujoco benchmarks to `Fast*` and delete the small `Fast*` variants they superseded. The KPI workloads are taken over unchanged (8192 worlds, 512 for KitchenG1, `track_simulate` reporting ms/world-step); `FastAllegro` additionally gets `timeout = 900` because Jetson Orin currently runs it close to the 600 s default.

The CI benchmark job selects benchmarks with `-b Fast`, so after this change CI exercises the KPI-scale workloads instead of the small variants. Since the ModelBuilder collision filter optimization (#3187), the KPI variants cost about the same wall time as the small ones on CI-class GPUs, so maintaining two variants per workload no longer pays for itself. Per-benchmark wall-clock durations recorded by asv on the newest post-#3187 runs (L40S at 5406c68e, CI-class hardware):

| Benchmark | old Fast [s] | KPI, now Fast [s] |
|---|---|---|
| Allegro | 88 | 65 |
| Cartpole | 28 | 9 |
| G1 | 61 | 33 |
| Humanoid | 38 | 35 |
| KitchenG1 | 21 | 12 |
| NewtonOverheadG1 | 8 | 31 |
| NewtonOverheadHumanoid | 6 | 33 |
| **Total** | **250** | **218** |

The KPI variants run once per benchmark while the old Fast variants paid setup twice (`rounds = 2`) plus repeats, which is why the bigger workloads come out cheaper overall. The Allegro figures predate the actuation fix (#3394) and include its roughly 2x inflation, so the renamed benchmark gets cheaper still. On the Jetson perflab machines the KPI set costs more than the old Fast set (Thor 478 s vs 292 s, Orin 772 s vs 500 s), but those machines run the full suite, where dropping the seven duplicate small variants is a net saving per run.

The `_KpiBenchmark` and `_NewtonOverheadBenchmark` base classes are unchanged, so the renamed benchmarks keep their asv version hashes and the dashboard history can be carried over to the new names in the results repository (follow-up there once this merges).

## Checklist

- [x] New or existing tests cover these changes — n/a, benchmark definitions only
- [x] The documentation is up to date with these changes — the `--bench FastAllegro` / `FastG1` examples in `docs/guide/development.rst` remain valid
- [ ] `CHANGELOG.md` has been updated (if user-facing change) — not user-facing

## Test plan

Standalone smoke run of a renamed benchmark on GPU:

```
uv run --extra dev python asv/benchmarks/simulation/bench_mujoco.py --bench FastCartpole
```

The PR's GPU benchmark CI job (`asv run -e -b Fast`) now runs all seven renamed benchmarks and is the real end-to-end verification.

## Summary by CodeRabbit

* **New Features**
  * Added a new benchmark for the Kitchen G1 simulation.
  * Expanded benchmark coverage with updated configurations for several fast simulation scenarios.

* **Bug Fixes**
  * Improved benchmark execution so timing now reflects repeated simulation steps across multiple runs.
  * Benchmarks now better adapt to available hardware and simulation settings.

* **Chores**
  * Simplified the benchmark selection list to include only the current fast benchmark set.

### DIFF
--- a/asv/benchmarks/simulation/bench_mujoco.py
+++ b/asv/benchmarks/simulation/bench_mujoco.py
@@ -9,7 +9,7 @@ import warp as wp
 wp.config.enable_backward = False
 wp.config.log_level = wp.LOG_WARNING
 
-from asv_runner.benchmarks.mark import SkipNotImplemented, skip_benchmark_if
+from asv_runner.benchmarks.mark import skip_benchmark_if
 
 parent_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
 sys.path.append(parent_dir)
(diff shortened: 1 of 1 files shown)


adenzler-nvidia:
This needs a follow-up history cleanup on the asv repo after the first results appear on the repo.

eric-heiden (approved):
LGTM!

PR #3418, opened 2026-07-09
Honor USD MassAPI value semantics

adenzler-nvidia:
## Description

Re-scoped after review feedback: the earlier revisions replaced the whole mass resolution path; the branch now reverts that rework and keeps only the minimal fix on the existing architecture (see the last two commits — the final diff against `main` is the fix itself).

Treat `UsdPhysics.MassAPI` attributes according to the schema's value semantics instead of using authoredness as the override signal. An attribute is effective only when its value is usable: positive and finite mass and density, nonzero `diagonalInertia` and `principalAxes`, finite `centerOfMass`. Explicitly authoring a schema fallback value (e.g. `physics:mass = 0`) now behaves exactly like leaving the attribute unauthored, instead of zeroing out derived mass properties or disabling inertia scaling. Non-finite mass and density values, including physics material density, are also treated as unspecified, and non-finite `ComputeMassProperties` results take the existing builder-accumulated fallback path.

The precedence cascade, collider mass-information callback, and shape-axis handling are unchanged. Deferred to follow-ups: collider-level non-finite MassAPI values (applied inside OpenUSD's aggregation), preservation of authored inertia below OpenUSD's comparison tolerance, and mass aggregation for bodies with collider-only MassAPI.

## Checklist

## Test plan

```bash
python -m newton.tests -p test_import_usd.py
```

New tests: `test_massapi_fallback_values_are_ignored`, `test_massapi_zero_inertia_with_authored_mass_scales_accumulated`, `test_massapi_non_finite_mass_and_density_are_ignored`, `test_non_finite_material_density_uses_default`. The full `test_import_usd.py` module and the other USD import test modules pass.

## Bug fix

**Steps to reproduce:**

1. Apply `PhysicsMassAPI` to a rigid body and explicitly author one or more schema fallback values (for example `physics:mass = 0.0` or `physics:diagonalInertia = (0, 0, 0)`).
3. Observe that the body imports with zero mass, or with collision-shape inertia that is not scaled to the authored mass.

**Minimal reproduction:**

```python

import newton

mass_api = UsdPhysics.MassAPI.Apply(body.GetPrim())
mass_api.CreateMassAttr().Set(0.0)  # schema fallback, means "unspecified"

collider = UsdGeom.Cube.Define(stage, "/Body/Collider")
collider.CreateSizeAttr().Set(0.2)
UsdPhysics.CollisionAPI.Apply(collider.GetPrim())

builder.add_usd(stage)
print(builder.body_mass[0])  # was 0.0, now derived from the collider
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved USD MassAPI mass/inertia/COM extraction to rely on effective (valid) values, ignoring non-finite/physically invalid data.
  * Prevented blocked/poisoned MassAPI attributes from impacting mass-property computation; clarified override vs fallback precedence (including `newton:inertia` vs diagonal inertia vs computed results).
  * Tightened material density acceptance and ensured rescaling keeps inertia and inverse-inertia consistent, with added warnings for discarded values.
* **Documentation**
  * Updated “Mass and Inertia Precedence” to describe effective-value handling and fallback equivalence for explicitly authored schema defaults.
* **Tests**
  * Expanded USDPhysics-parity and regression coverage for sentinel/blocked MassAPI scenarios, plus more defensive principal-axes quaternion handling.


newton/_src/utils/import_usd.py
```diff
@@ -3611,59 +3616,28 @@ def _collect_filtered_pairs(prim):
...
-                            )
-                        if mass_info is not None:
-                            rigid_body_mass_info_map[path] = mass_info
+                    mass_info, has_collider_mass_api_values = _prepare_collider_mass_info(
+                        prim, mass_info, body_values, shape_spec.materials
+                    )
+                    if has_collider_mass_api_values:
+                        collider_mass_api_paths_by_body[body_path].append(path)
+                    if mass_info is not None:
+                        rigid_body_mass_info_map[path] = mass_info
```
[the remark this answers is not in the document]
adenzler-nvidia: This is a pre-existing limitation, not introduced here — the importer has always gated the mass-overwrite path on body-level MassAPI, and the documented scope in `usd_parsing.rst` matches. Extending aggregation to bodies with collider-only MassAPI needs a collider pre-scan and its own tests, so I'm deferring it to a follow-up issue rather than growing this PR.

docs/concepts/usd_parsing.rst
```diff
@@ -270,53 +270,31 @@ Mass and Inertia Precedence
...
+   remain authoritative.
 
 If resolved mass is non-positive, inverse mass is set to ``0``.
 
-.. tip::
-
-   For the most predictable results, fully author ``physics:mass``, ``physics:diagonalInertia``,
-   ``physics:principalAxes``, and ``physics:centerOfMass`` on each rigid body.  This avoids any
-   fallback heuristics and is also the fastest import path since ``ComputeMassProperties(...)``
-   can be skipped entirely.
```
andrewkaufman, line 318: This tip still seems relevant, any reason to remove it entirely rather than just ~fastest import path since ``ComputeMassProperties(...)`` can be skipped entirely.~

docs/concepts/usd_parsing.rst
```diff
@@ -270,53 +270,31 @@ Mass and Inertia Precedence
...
+   overrides the inertia returned by OpenUSD.
+2. Newton supplies equivalent unit-density mass information through OpenUSD's callback. OpenUSD
+   applies collider/body ``MassAPI`` and bound ``MaterialAPI`` values while aggregating missing
+   body properties. Effective values are preserved exactly, including inertias smaller than
+   OpenUSD's internal comparison tolerance.
+3. When ``NewtonMassAPI`` is applied to a collider, ``newton:massModel`` controls whether its
+   callback inertia comes from the full volume (``"solid"``, default) or a thin shell.
+   ``newton:shellThickness`` sets the inward wall thickness [m]; its ``-inf`` fallback uses
+   ``newton:contactMargin``.
+4. A collider synthesized by a schema resolver is not discoverable by OpenUSD traversal. Bodies
```
andrewkaufman: I don't quite understand this case... What is a "collider synthesized by a schema resolver"? Can you give an example?

newton/_src/utils/import_usd.py
```diff
@@ -2958,145 +3025,86 @@ def _resolve_contact_attr(key, _prim=prim):
...
         mass_info.centerOfMass = Gf.Vec3f(*shape_com)
-        mass_info.localPos = Gf.Vec3f(*local_pos)
-        mass_info.localRot = _resolve_mass_info_local_rotation(local_rot, shape_geo_type, shape_axis)
+        mass_info.localPos = Gf.Vec3f(*shape_xform.p)
+        mass_info.localRot = Gf.Quatf(
+            float(shape_xform.q[3]),
+            float(shape_xform.q[0]),
+            float(shape_xform.q[1]),
+            float(shape_xform.q[2]),
+        )
```
andrewkaufman: I always find Quatf construction confusing... I believe there is a singature that takes float, Vec3f to make it explicit which is the imaginary element. I'd recommend always preferring that one to ensure we don't get the order wrong.

newton/_src/utils/import_usd.py
```diff
@@ -2958,145 +3025,86 @@ def _resolve_contact_attr(key, _prim=prim):
...
         mass_info.inertia = Gf.Matrix3f(*shape_inertia_np.flatten().tolist())
         return mass_info
 
+    def _prepare_collider_mass_info(prim, mass_info, body_values, material_paths):
+        if not prim.HasAPI(UsdPhysics.MassAPI):
+            return mass_info, False
+
+        mass_api = UsdPhysics.MassAPI(prim)
```
andrewkaufman: this is a double API composition check, you can do it this way instead:

```suggestion
        mass_api = UsdPhysics.MassAPI(prim)
        if not mass_api:
            return mass_info, False
```

newton/_src/utils/import_usd.py
```diff
@@ -2958,145 +3025,86 @@ def _resolve_contact_attr(key, _prim=prim):
...
+        # OpenUSD treats small nonzero inertia as its zero sentinel.
+        inertia /= inertia_scale
+        mass_info.inertia = Gf.Matrix3f(*np.diag(inertia).flatten().tolist())
+
+        if np.all(np.isfinite(com)):
+            scale = usd.get_scale(prim, local=False)
+            mass_info.centerOfMass = Gf.CompMult(
+                Gf.Vec3f(*(float(value) for value in com)),
+                Gf.Vec3f(*(float(value) for value in scale)),
+            )
```
andrewkaufman: Is there a reason to do Gf math here? You already went Gf -> np & the math is usually faster in np

newton/_src/utils/import_usd.py
```diff
@@ -2958,145 +3025,86 @@ def _resolve_contact_attr(key, _prim=prim):
...
+
+        mass_api = UsdPhysics.MassAPI(prim)
+        mass = float(mass_api.GetMassAttr().Get())
+        density = float(mass_api.GetDensityAttr().Get())
+        # Non-finite authored values are invalid; treat them as the 0.0 sentinel.
+        if not math.isfinite(mass):
+            mass = 0.0
+        if not math.isfinite(density):
+            density = 0.0
+        inertia = np.array(mass_api.GetDiagonalInertiaAttr().Get(), dtype=np.float32)
```
andrewkaufman: Why do we only look at diagonal inertia on the collider & not principal axes or the newton:inertia tensor?

newton/_src/utils/import_usd.py
```diff
@@ -2958,145 +3025,86 @@ def _resolve_contact_attr(key, _prim=prim):
...
-
     def _build_mass_info_from_shape_geometry(
-        prim: Usd.Prim,
-        local_pos,
-        local_rot,
+        shape_xform,
         shape_geo_type: int,
         shape_scale: wp.vec3,
         shape_src: Mesh | None,
-        shape_axis=None,
```
andrewkaufman, line 3075: Why doesn't this need to account for shape axis anymore?

newton/_src/utils/import_usd.py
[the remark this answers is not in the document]
andrewkaufman: Wait, I'm confused... if we are requiring MassAPI on the body, what is MassAPI on the collider providng for us? AFAIK its supposed to be ignored for mass computation when there is explicit MassAPI on the body (the body overrides whatever would've been set on the colliders)

newton/_src/utils/import_usd.py
```diff
@@ -3766,166 +3740,113 @@ def _get_collision_mass_information(collider_prim: Usd.Prim):
...
+            elif body_path in bodies_with_non_native_colliders:
+                ignored_collider_values = collider_mass_api_paths_by_body.get(body_path)
+                if ignored_collider_values:
                     warnings.warn(
-                        f"Body {body_path}: authored diagonal inertia contains negative values. "
-                        "Falling back to mass-computer result.",
+                        f"Body {body_path} contains colliders that OpenUSD cannot discover and collider-level "
+                        f"MassAPI values on {ignored_collider_values}; using descriptor geometry and material "
+                        "density without those collider overrides. Author complete body MassAPI properties or "
+                        "apply PhysicsCollisionAPI to every collider.",
```
andrewkaufman: What prim are we considering a collider that doesn't have `PhysicsCollisionAPI` applied? That should never happen... those aren't colliders...

newton/_src/utils/import_usd.py
```diff
@@ -2958,145 +3025,86 @@ def _resolve_contact_attr(key, _prim=prim):
...
+                material_prim = stage.GetPrimAtPath(material_paths[0])
+                if material_prim.HasAPI(UsdPhysics.MaterialAPI):
+                    inertia_scale = float(UsdPhysics.MaterialAPI(material_prim).GetDensityAttr().Get())
+            if not math.isfinite(inertia_scale) or inertia_scale <= 0.0:
+                meters_per_unit = float(UsdGeom.GetStageMetersPerUnit(stage))
+                kilograms_per_unit = float(UsdPhysics.GetStageKilogramsPerUnit(stage))
+                inertia_scale = 1000.0 * meters_per_unit**3 / kilograms_per_unit
+
+        # OpenUSD treats small nonzero inertia as its zero sentinel.
+        inertia /= inertia_scale
```
andrewkaufman: [bot] The division `inertia /= inertia_scale` converts authored diagonal inertia to "unit-density" form for the `MassInformation` payload — correct in principle since `ComputeMassProperties` multiplies back by density.

However, the fallback cascade (collider density → body density → material density → default 1000·m³/kg) needs to match what OpenUSD actually applies during aggregation. If OpenUSD picks a *different* density for this collider (e.g. it resolves material density from a different binding path, or uses body density instead of collider density), the round-trip breaks: we divide by one density here but OpenUSD multiplies by another, giving incorrect physical inertia.

Is there a guarantee the density source used here exactly mirrors OpenUSD's internal density resolution for the same collider during `ComputeMassProperties`? If not, this could silently produce wrong inertia for cases where the density cascade disagrees.

newton/_src/utils/import_usd.py
```diff
@@ -440,6 +440,23 @@ class PhysicsMaterial:
...
+    class BodyMassAPIValues:
+        mass: float | None
+        density: float | None
+        inertia: np.ndarray | None
+        com: np.ndarray | None
+        principal_axes: Any
+        inertia_tensor: wp.mat33 | None
+
+        @property
+        def is_complete(self) -> bool:
```
andrewkaufman: [bot] `is_complete` does not account for `density` as a mass source. A body with effective density + inertia + com (but no explicit mass) will have `is_complete = False`, triggering the `ComputeMassProperties` path. That path's result is then immediately overridden by `body_values.inertia` and `body_values.com` at the end of the loop.

This is functionally OK (the overrides win), but it's performing unnecessary work and the name is misleading — consider renaming to something like `has_direct_mass_and_inertia` or including density:
```python
self.mass is not None or self.density is not None
```
in the mass condition.

newton/_src/utils/import_usd.py
```diff
@@ -3766,166 +3740,113 @@ def _get_collision_mass_information(collider_prim: Usd.Prim):
...
-                    builder.body_inv_inertia[body_id] = wp.mat33(0.0)
-            elif i_diag_np is not None and np.linalg.norm(i_diag_np) > 0.0:
+                if body_values.density is not None and default_shape_density > 0.0:
+                    density_scale = body_values.density / default_shape_density
+                    mass *= density_scale
+                    builder.body_inertia[body_id] = wp.mat33(np.array(builder.body_inertia[body_id]) * density_scale)
+
+                if (
+                    body_values.mass is not None
+                    and body_values.inertia is None
```
andrewkaufman: [bot] When both `body_values.density` and `body_values.mass` are present, the inertia gets scaled twice in sequence:
1. `body_inertia *= density_scale` (line ~3800)
2. `inertia_scale = body_values.mass / mass` → `body_inertia *= inertia_scale` (here)

The net effect is `body_inertia *= (density_scale) * (body_values.mass / (accumulated * density_scale))` = `body_inertia *= body_values.mass / accumulated`, which is the correct final scale.

However, the intermediate state after step 1 is semantically wrong (inertia scaled to density but mass about to be overridden). Consider computing the final scale factor directly:
```python
final_scale = body_values.mass / shape_accumulated_mass
```
and applying it once. Easier to audit and less fragile if future code inserts logic between the two blocks.

newton/_src/utils/import_usd.py
```diff
@@ -2330,15 +2347,85 @@ def _resolve_contact_attr(key, _prim=prim):
...
+                )
+            elif any(value < 0.0 for value in inertia_tensor_value[:3]):
+                warnings.warn(
+                    f"Body {body_path}: newton:inertia has negative diagonal elements. Ignoring.",
+                    stacklevel=3,
+                )
+            else:
+                ixx, iyy, izz, ixy, ixz, iyz = inertia_tensor_value
+                inertia_np = np.array([[ixx, ixy, ixz], [ixy, iyy, iyz], [ixz, iyz, izz]], dtype=np.float64)
+                if np.any(np.linalg.eigvalsh(inertia_np) < 0.0):
```
andrewkaufman: [bot] The PSD check `np.linalg.eigvalsh(inertia_np) < 0.0` with exact zero threshold is numerically fragile. A physically valid inertia tensor with large off-diagonal elements (e.g. a body rotated 45° about one axis) can produce eigenvalues like `-1e-15` from floating-point error in the eigendecomposition, causing a false rejection.

Consider a small negative tolerance:
```python
if np.any(np.linalg.eigvalsh(inertia_np) < -eps * np.trace(inertia_np)):
```
where `eps` is something like `1e-6`. This rejects genuinely indefinite tensors while tolerating numerical noise proportional to the tensor magnitude.

newton/_src/utils/import_usd.py
```diff
@@ -2958,145 +3025,86 @@ def _resolve_contact_attr(key, _prim=prim):
...
+            if not math.isfinite(inertia_scale) or inertia_scale <= 0.0:
+                meters_per_unit = float(UsdGeom.GetStageMetersPerUnit(stage))
+                kilograms_per_unit = float(UsdPhysics.GetStageKilogramsPerUnit(stage))
+                inertia_scale = 1000.0 * meters_per_unit**3 / kilograms_per_unit
+
+        # OpenUSD treats small nonzero inertia as its zero sentinel.
+        inertia /= inertia_scale
+        mass_info.inertia = Gf.Matrix3f(*np.diag(inertia).flatten().tolist())
+
+        if np.all(np.isfinite(com)):
```
andrewkaufman: [bot] `usd.get_scale(prim, local=False)` returns the accumulated scale from root to prim. `MassInformation.centerOfMass` is defined in the collider's local frame, and `localPos`/`localRot` are a rigid transform (no scale component) — so scale does need to be applied separately to map COM into the body frame.

However, using root-to-prim scale means this includes the body's own scale. If the body has scale `(2, 2, 2)` and the collider has scale `(3, 3, 3)`, the non-local scale is `(6, 6, 6)`. But OpenUSD's `ComputeMassProperties` operates in body-local space — so the relevant scale should be body-to-collider (i.e. `(3, 3, 3)` in this example), not root-to-collider.

The test only covers a flat body→collider hierarchy where the body has no scale, so this distinction isn't exercised. A test with a scaled intermediate group or a scaled body would clarify the intended behavior.

newton/_src/utils/import_usd.py
```diff
@@ -3766,166 +3740,113 @@ def _get_collision_mass_information(collider_prim: Usd.Prim):
...
-            # (scaled to match authored mass below) instead of using mass computer
-            # inertia, which may already reflect the authored mass.
-            if has_inertia_tensor:
-                i_diag_np = None  # skip diagonal path; full matrix set below
-            elif has_authored_inertia:
-                i_diag_np = np.array(mass_api.GetDiagonalInertiaAttr().Get(), dtype=np.float32)
-                if np.any(i_diag_np < 0.0):
+                cmp_mass_valid = math.isfinite(cmp_mass) and cmp_mass >= 0.0
+                mass = float(cmp_mass) if cmp_mass_valid else shape_accumulated_mass
+                if not cmp_mass_valid and body_values.density is not None and default_shape_density > 0.0:
```
andrewkaufman: [bot] This rescaling assumes `shape_accumulated_mass` was computed entirely with `default_shape_density`. But earlier in the collider loop, individual shapes can get non-default density from their own `MassAPI` density, material bindings, or body density (via shape_spec density resolution).

If collider A used material density 500 and collider B used default density 1000, the accumulated mass reflects that mix. Scaling by `body_density / default_shape_density` overcorrects for collider A.

Is this intentional as a "best effort" approximation when `ComputeMassProperties` fails? If so, a comment explaining the approximation would help. If not, this path should either:
- Re-sum collider volumes and apply body density from scratch, or
- Only be reachable when all colliders used the default density.

newton/tests/test_import_usd.py
```diff
@@ -7016,6 +7020,290 @@ def test_massapi_partial_body_falls_back_to_compute(self):
...
+                builder = newton.ModelBuilder()
+                result = builder.add_usd(stage)
+                body_idx = result["path_body_map"]["/Body"]
+
+                self.assertAlmostEqual(builder.body_mass[body_idx], expected_density * 0.1**3, places=6)
+                inertia = np.array(builder.body_inertia[body_idx]).reshape(3, 3)
+                np.testing.assert_allclose(np.diag(inertia), [1.0e-6, 2.0e-6, 3.0e-6], atol=1e-10, rtol=1e-6)
+
+    @unittest.skipUnless(USD_AVAILABLE, "Requires usd-core")
+    def test_mixed_non_native_colliders_warn_for_ignored_massapi(self):
```
andrewkaufman: [bot] This test simulates a "non-native collider" by *removing* `CollisionAPI` after load. But a real schema resolver would *add* a collider that was never authored in the scene — the prim would appear in physics traversal results but never had `CollisionAPI` in the composed layer stack.

The mock approach (add API, load, remove API) creates a subtly different situation: the prim DID have `CollisionAPI` during `LoadUsdPhysicsFromRange`, so it was classified as a collision shape and appears in `ret_dict`. After removal, the importer's `prim.HasAPI(UsdPhysics.CollisionAPI)` check fails. This matches the detection logic, but doesn't prove the scenario is reachable in production without the mock.

Can you point to a real OpenUSD schema resolver that produces collision shapes without authoring `CollisionAPI`? If this is purely defensive, consider documenting it as such rather than implying it's a known production path.

andrewkaufman (commented):
This is a pretty large change... are we trying to get this for 1.4 or is it for 1.5?

I think @eric-heiden should review as well, iirc he wrote the original mass parsing code.

eric-heiden:
Thanks for addressing the fallback-value semantics. This is definitely an oversight that we used authoredness rather than the values to determine if a fallback is needed. The focused fix makes sense, but the broader mass-aggregation rewrite appears to introduce two behavior regressions and an import slowdown, while the minimal body-level example from the PR description already produces the expected mass and inertia on `main`.

- With body `physics:mass = 0.02` and collider `physics:diagonalInertia = (1e-6, 2e-6, 3e-6)`, the imported inertia becomes `(0.02, 0.04, 0.06)`. Moving the same mass to the collider preserves the authored inertia. `_prepare_collider_mass_info()` does not consider body mass when selecting its scale.
- Collider `physics:diagonalInertia = (-1, 2, 3)` is now ignored without warning. The previous path warned about invalid collider inertia, as the body-level path still does.
- Importing 200 bodies with explicitly authored MassAPI fallback values took 0.571 s versus 0.417 s on `main`, a 37.1% increase across 14 measured runs in both execution orders. The unauthored control increased only 6.2%. This is observed in the new `ComputeMassProperties()` fallback path in `import_usd.py`.

I agree with Andrew, it is a lot of code that is changed here which is probably not necessary.
Could we narrow this PR to interpreting the USD sentinel values as unspecified while retaining the existing aggregation paths, then add focused tests for the cases above?

docs/concepts/usd_parsing.rst
adenzler-nvidia: Restored — the PR is re-scoped and the docs rework is reverted, so the tip is back verbatim. Its "fastest import path" clause also still holds on this architecture: with all four attributes effective, `ComputeMassProperties` is skipped.

docs/concepts/usd_parsing.rst
adenzler-nvidia: Gone with the re-scope. This paragraph described the reverted rework's handling of colliders created by schema resolvers (shapes that aren't real USD prims, so OpenUSD traversal can't discover them). The restored code keeps main's behavior for that case: when `ComputeMassProperties` fails to find colliders, the body falls back to builder-accumulated mass properties.

newton/_src/utils/import_usd.py
adenzler-nvidia: The construction this was on is removed with the revert. The final diff only contains the identity literal and an equality compare against the zero-quat schema fallback — no component-wise construction. The one remaining `(w, x, y, z)` construction is in main's `_resolve_mass_info_local_rotation`, unchanged by this PR; switching it to the `(real, Vec3f)` signature would be a follow-up.

newton/_src/utils/import_usd.py
adenzler-nvidia: Applied, and to the other two MassAPI sites in the mass path as well — the schema object is constructed once and its bool operator replaces the separate `HasAPI` check.

newton/_src/utils/import_usd.py
adenzler-nvidia: That helper is gone with the revert. The final diff adds no Gf arithmetic: the value-semantics checks iterate components in Python (`isfinite` / nonzero), and the only remaining Gf operation is main's pre-existing `Gf.CompMult` for the authored-COM scale convention, unchanged.

newton/_src/utils/import_usd.py
adenzler-nvidia: The reader this was on ignored `principalAxes`; the restored helper honors it (an effective `principalAxes` rotates the authored diagonal before building the callback payload). Collider-level `newton:inertia` has never been read — it's body-level only on main too. Supporting it on colliders would be a follow-up.

newton/_src/utils/import_usd.py
adenzler-nvidia: Moot after the re-scope — `_resolve_mass_info_local_rotation` and its call site are preserved unchanged.

newton/_src/utils/import_usd.py
adenzler-nvidia: That path was the reverted rework's explicit modeling of resolver-synthesized colliders; it's gone. Main's behavior is restored: those bodies take the builder-accumulated fallback when `ComputeMassProperties` reports failure.

newton/_src/utils/import_usd.py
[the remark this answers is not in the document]
adenzler-nvidia: With the rework reverted this is main's behavior again. Body-level MassAPI application gates whether Newton overwrites builder-accumulated values at all. When the body's own effective mass/inertia/COM are incomplete, `ComputeMassProperties` aggregates collider contributions — effective collider MassAPI mass+diagonalInertia are converted to unit-density callback payloads, and OpenUSD applies collider/material density during aggregation. Effective body-level values then override the aggregate, which is the parent-overrides-child rule you're describing. What collider MassAPI does not do (pre-existing) is trigger the overwrite when the body has no MassAPI at all — that's the deferred follow-up from earlier in this thread.

newton/_src/utils/import_usd.py
adenzler-nvidia: The flagged code is removed, but the concern conceptually applies to main's pre-existing collider helper too (restored unchanged here): the unit-density conversion divides by `mass/volume` and assumes OpenUSD scales back consistently. The simple case is covered by `test_massapi_partial_body_falls_back_to_compute`; a collider with both authored mass and a bound material density is untested. Follow-up candidate rather than this PR.

newton/_src/utils/import_usd.py
adenzler-nvidia: Moot — this code is removed with the re-scope revert.

newton/_src/utils/import_usd.py
adenzler-nvidia: Moot — this code is removed with the re-scope revert. The analogous pre-existing interaction between the density rescale and authored-mass scaling in main's failure branch is unchanged by this PR.

newton/_src/utils/import_usd.py
adenzler-nvidia: Moot — this code is removed with the re-scope revert.

newton/_src/utils/import_usd.py
adenzler-nvidia: The flagged line is gone, but the identical check pre-exists on main in the `newton:inertia` validation (unchanged by this PR). Agreed a valid tensor with off-diagonal terms can eigen-decompose to ~-1e-15 and get falsely rejected (it warns and falls back, so non-fatal). Deferring to a follow-up with its own test.

newton/_src/utils/import_usd.py
adenzler-nvidia: The rewrite's version is gone; main's pre-existing equivalent in the `ComputeMassProperties`-failure branch has the same default-density assumption and is only reachable when collider discovery fails. Unchanged here; follow-up candidate.

newton/tests/test_import_usd.py
adenzler-nvidia: Test removed with the re-scope. The four tests in the final diff drive `ModelBuilder.add_usd()` on real stages only — no mocks or importer internals.

newton/_src/utils/import_usd.py
```diff
@@ -2330,15 +2347,85 @@ def _resolve_contact_attr(key, _prim=prim):
                 default=builder.default_shape_cfg.mu_rolling,
                 verbose=verbose,
             ),
-            # Treat non-positive/unauthored material density as "use importer default".
-            # Authored collider/body MassAPI mass+inertia is handled later.
+            # Treat non-positive material density as "use importer default".
+            # Collider/body MassAPI precedence is handled later.
             density=desc.density if desc.density > 0.0 else default_shape_density,
```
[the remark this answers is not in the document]
adenzler-nvidia: Addressed for everything the importer reads directly: body mass/density, physics material density, and `ComputeMassProperties` results are gated on `isfinite`. Collider-level authored non-finite values are applied inside OpenUSD's own aggregation and can't be gated from the callback side; deferred to a follow-up.

adenzler-nvidia:
This is re-scoped now. The earlier revisions went well beyond the bugfix — I let the change grow into a rework of the whole mass resolution path without double-checking that scope against what the fix actually required. The last commits revert the rework back to main's architecture and re-apply only the minimal value-semantics fix: the final diff against main is +240/−67, mostly tests, and the precedence cascade, collider callback, and shape-axis handling are unchanged. All inline threads are answered individually; most point at code that no longer exists.

Given the history I'd rather not rush this into 1.4 — this is for 1.5.

A request for next time: if a PR of mine looks oversized for its stated purpose, push back right away, before spending time on a detailed review. A "this is too big, re-scope it" up front is more useful than line comments on code that shouldn't exist, and cheaper for everyone.

Deferred follow-ups: collider-level non-finite MassAPI values, authored-inertia preservation below OpenUSD's comparison tolerance, collider-only MassAPI aggregation, and the exact-zero PSD tolerance in the `newton:inertia` validation.

andrewkaufman:
> A request for next time: if a PR of mine looks oversized for its stated purpose, push back right away, before spending time on a detailed review. A "this is too big, re-scope it" up front is more useful than line comments on code that shouldn't exist, and cheaper for everyone.

@adenzler-nvidia is that a comment from you to me? Or from your bot to you?

newton/_src/utils/import_usd.py
```diff
@@ -623,6 +623,30 @@ def _get_mesh_cached(prim: Usd.Prim, *, load_uvs: bool = False, load_normals: bo
...
+        mass = mass_api.GetMassAttr().Get()
+        return float(mass) if mass is not None and math.isfinite(mass) and mass > 0.0 else None
+
+    def _mass_api_effective_density(mass_api: UsdPhysics.MassAPI) -> float | None:
+        density = mass_api.GetDensityAttr().Get()
+        return float(density) if density is not None and math.isfinite(density) and density > 0.0 else None
+
+    def _mass_api_effective_diag_inertia(mass_api: UsdPhysics.MassAPI):
+        diag = mass_api.GetDiagonalInertiaAttr().Get()
+        return diag if diag is not None and any(v != 0.0 for v in diag) else None
```
eric-heiden: 🔵 This is pre-existing rather than a regression in this PR, but could we clean it up while this helper is defining MassAPI value semantics? `any(v != 0.0 for v in diag)` treats negative and non-finite components as effective. In particular, an invalid collider `diagonalInertia` can then be discarded without warning, and the body path may skip computing the fallback it later needs. Consider requiring all components to be finite and nonnegative here, and warning when a nonzero explicitly authored value fails that validation.

adenzler-nvidia:
That comment was from me to you, using my bot :-) I clearly didn't double-check what my agents flagged as a drive-by fix, so please just tell me "this is too much, I won't review this" before spending too much time on it. This is also ok for PRs coming from other.

Myself, I need to double-check everything again as well.

newton/_src/utils/import_usd.py
adenzler-nvidia: Fixed in 7378240. The helper now requires all components finite and nonnegative (at least one nonzero) and warns once per prim when a nonzero authored value fails validation. The downstream negative-value checks became unreachable and are removed — the body-path one also read `cmp_i_diag` unbound or stale from a previous loop iteration when mass and CoM were effective. Non-finite collider values applied inside OpenUSD's own aggregation remain the deferred follow-up.

adenzler-nvidia:
The CPU CI failures in `test_massapi_blocked_attributes_are_unspecified` were not flakiness — they trip undefined behavior in OpenUSD. `UsdPhysicsRigidBodyAPI::ComputeMassProperties` reads MassAPI attributes into uninitialized locals (`_ParseMassApi`/`_GetCoM` in `pxr/usd/usdPhysics/rigidBodyAPI.cpp`):

```cpp
float m;
massAttribute.Get(&m);
if (m > 0.0f)
{
    result.mass = m;
}
```

For an unauthored attribute `Get()` succeeds and returns the schema fallback `0.0`, which correctly fails the `> 0` check. For a blocked attribute `Get()` fails and leaves `m` holding stack garbage, and any positive garbage is taken as an authored mass. That's why CI saw platform-varying denormal masses (`2.67e-39` ubuntu, `1.4e-45` macos, `2.8e-45` windows; the arm job failed on the inertia/com comparison instead) while GPU CI and local runs pass by luck of benign stack contents. The same pattern affects `diagonalInertia`, `principalAxes`, and `centerOfMass`. The code is unchanged on OpenUSD `dev`, and there is no existing upstream issue; a report is drafted and will be filed separately.

3f0c2b4 works around it: when the body prim or any collider prim in the `ComputeMassProperties` traversal carries a blocked MassAPI attribute (detected via `GetResolveInfo().ValueIsBlocked()` — note `HasAuthoredValue()` returns `False` for blocks), the importer skips `ComputeMassProperties` and routes the body through the existing accumulated-property fallback. The workaround is annotated in the code with the revert condition: delete the two helpers and the call-site bypass once the minimum supported usd-core ships the upstream fix.

The test now also covers a blocked collider-level mass attribute (OpenUSD parses collider MassAPI internally even when mass information is supplied via the callback), and asserts full mass/inertia/com equality for the blocked-CoM body, which previously couldn't be asserted because it depended on `ComputeMassProperties` reading blocked data.

eric-heiden (approved):
Thanks, LGTM!

PR #3422, opened 2026-07-09
Prepare Newton 1.4.0rc1

jcarius-nv:
## Description

Prepare `release-1.4` for the first Newton 1.4 release candidate.

- Set the project version to `1.4.0rc1`.
- Replace the Warp development build with stable `warp-lang>=1.15.0`.
- Remove the NVIDIA package index and Warp source override so the release
  installs entirely from public PyPI.
- Regenerate the public API documentation and `uv.lock`.

## Checklist

- [x] `CHANGELOG.md` has been updated (not applicable; this is release metadata)

## Test plan

```bash
uv run --frozen python -c "import newton; assert newton.__version__ == '1.4.0rc1'"
```

### DIFF
--- a/asv.conf.json
+++ b/asv.conf.json
@@ -16,7 +16,7 @@
   "build_command": ["python -m build --wheel -o {build_cache_dir} {build_dir}"],
   "install_command": [
     "python -m pip install -U numpy",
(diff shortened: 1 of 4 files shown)


adenzler-nvidia:
asv.conf.json should also be updated to the right warp version, but I wouldn't block on that because it can also happen in another PR.

adenzler-nvidia (approved):
looks good!

PR #3446, opened 2026-07-10
Tune VBD contact settings in example scenes

### LINKED ISSUE #3492: [BUG] Newton-basic_shapes example (VBD variant) not showing realistic friction behavior
### Bug Description

From QA testing the samples:

08. Newton-basic_shapes — FAIL
    Variant A command:
      python -m newton.examples basic_shapes --solver xpbd
    Variant A result: PASS (user-assisted visual verification).
    All seven documented shapes appeared, fell naturally, and remained stable.
    Window close ended the process cleanly; captured stderr was empty.
    Variant B command:
      python -m newton.examples basic_shapes --solver vbd
    Variant B result: FAIL (user-assisted visual/interaction verification).
    Observed discrepancy:
    - Ground friction appeared effectively absent for non-round objects; a
      deliberate lateral drag/release check showed ice-like sliding with
      insufficient deceleration.
    - This contradicts the documented realistic-behavior expectation.
    - Inspection of the installed 1.4.0rc1 example confirmed it explicitly
      sets `builder.default_shape_cfg.mu = 0.5`, so the observation is not
      explained by an intentionally zero friction coefficient.
    - VBD does omit the XPBD-specific rolling/torsional friction settings, but
      the failure was classified using a non-round-object sliding check.
    - No console exception occurred; captured stderr was empty and window
      close ended the process cleanly.
    Overall result is FAIL because both required variants must pass.

### Reproduction Script

```python

```

### System Information

_No response_

jumyungc:
## Description
This PR makes two small example-level tuning adjustments:

  - Updates the VBD contact settings in `example_basic_shapes` for firmer resting poses and more natural floor friction.
  - Reduces ground friction in `example_cable_bundle_hysteresis` so the no-Dahl configuration relaxes after obstacle release while the Dahl configuration retains
  its curved shape.
  - Adds explicit XPBD/VBD coverage for basic shapes.
  - Adds short Dahl/no-Dahl straightness checks for the cable example.

  These changes are limited to example configuration and testing. There are no solver or API changes.

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **Bug Fixes**
  * Tuned VBD default contact parameters in the basic shapes and cable bundle hysteresis examples for more consistent friction and recovery behavior, including reducing ground-plane friction in the cable bundle example.
  * Improved cable bundle hysteresis verification by replacing the placeholder check with a post-release straightness measurement and scenario-specific pass/fail thresholds for Dahl-enabled vs no-Dahl cases.
* **Tests**
  * Updated example subprocess coverage to run basic shapes under both XPBD and VBD, and adjusted the cable hysteresis tests to use CUDA devices with separate Dahl retention and no-Dahl recovery runs and longer simulations.
* **Documentation**
  * Added the corresponding note to the Unreleased changelog.


newton/tests/test_examples.py
```diff
@@ -404,6 +413,22 @@ class TestCableExamples(unittest.TestCase):
     use_viewer=True,
     test_options={"num-frames": 20},
 )
+add_example_test(
+    TestCableExamples,
+    name="cable.example_cable_bundle_hysteresis",
+    devices=cuda_test_devices,
+    use_viewer=True,
+    test_options={"num-frames": 150},
```
adenzler-nvidia: `with_dahl` is derived from the example's parser defaults (`--eps-max 2.0`, `--tau 0.1`), and `test_final` picks which assertion to run based on it — so if the example's defaults ever change to an elastic configuration, this test silently flips to asserting elastic recovery while keeping its `dahl_retention` name. Pinning the physics here makes the test's contract self-contained:

```suggestion
    test_options={"num-frames": 150, "eps-max": 2.0, "tau": 0.1},
```

newton/examples/cable/example_cable_bundle_hysteresis.py
```diff
@@ -399,7 +402,26 @@ def render(self):
...
+            arc_length = float(np.linalg.norm(np.diff(points, axis=0), axis=1).sum())
+            end_to_end = float(np.linalg.norm(points[-1] - points[0]))
+            straightness.append(end_to_end / max(arc_length, 1.0e-8))
+
+        straightness = np.asarray(straightness)
+        if not np.all(np.isfinite(straightness)):
+            raise ValueError("Cable straightness is not finite")
+
+        mean_straightness = float(np.mean(straightness))
+        if self.with_dahl and mean_straightness > 0.9:
```
adenzler-nvidia: This check is one-sided: straightness has no lower bound, so a bundle that collapses (all bodies at one point gives `end_to_end = 0`, and the `max(arc_length, 1e-8)` clamp turns the degenerate 0/0 into a clean 0.0) or blows up into a finite tangle passes as "retained curvature" — the isfinite check only catches the NaN endpoint of a divergence. The healthy Dahl value lands around 0.73, so there is plenty of room for a lower bound. A per-cable check that `arc_length` stays within ~±20% of rest length would be even more direct: one comparison catches collapse, tangling, and joint stretch in both configurations.

newton/examples/cable/example_cable_bundle_hysteresis.py
```diff
@@ -399,7 +402,26 @@ def render(self):
...
+            points = body_positions[body_ids]
+            arc_length = float(np.linalg.norm(np.diff(points, axis=0), axis=1).sum())
+            end_to_end = float(np.linalg.norm(points[-1] - points[0]))
+            straightness.append(end_to_end / max(arc_length, 1.0e-8))
+
+        straightness = np.asarray(straightness)
+        if not np.all(np.isfinite(straightness)):
+            raise ValueError("Cable straightness is not finite")
+
+        mean_straightness = float(np.mean(straightness))
```
adenzler-nvidia: Taking the mean across the 7 cables can average away a single bad cable: six cables at 1.0 plus one stuck at 0.65 still passes the ≥ 0.95 recovery check, and one cable that slips off the obstacles and snaps straight still passes the ≤ 0.9 retention check. `np.min` (no-dahl) / `np.max` (dahl) assert the same claims per-cable, and measured values (min ≈ 0.966 no-dahl, max ≈ 0.743 dahl on GPU) pass the current thresholds unchanged. Worth also putting the per-cable values in the error message — a failure currently reports only a single number.

adenzler-nvidia (commented):
Ran the new registrations locally on GPU: both cable tests pass, `no_dahl_recovery` fails with the old `mu=2.5`, and the vbd `basic_shapes` variant fails with the old ke/kd — so the tests discriminate the tuning well. Three hardening requests on the new assertions below.

newton/tests/test_examples.py
jumyungc: Applied.

newton/examples/cable/example_cable_bundle_hysteresis.py
jumyungc: Done.

newton/examples/cable/example_cable_bundle_hysteresis.py
jumyungc: Done.

jcarius-nv (approved):

PR #3452, opened 2026-07-10
Add fast math to render kernel

StafaH:
Summary

Compile Newton render kernels with CUDA fast math for a roughly 10–13% throughput improvement across tested scenes.

Fast math can be disabled with RenderConfig.enable_fast_math.

Visual differences are negligible: relative depth error is below 1e-6, and color differences are limited to a small number of isolated pixels near object silhouettes.

| Scene | Baseline | + Fast math (`this PR`) | + Refactor (`84ce2f27`) | + Clear fill (`269b21a4`) | Total |
|---|---|---|---|---|---|
| Franka | 514k | 574k | 601k | 669k | **+30%** |
| Quadruped | 526k | 583k | 618k | 683k | **+30%** |
| Franka cabinet | 358k | 401k | 431k | 485k | **+36%** |
| Shapes 256 | 672k | 756k | 799k | 794k | **+18%** |

image

## Summary by CodeRabbit

* **New Features**
  * Added an option to enable CUDA fast math for tiled camera rendering by default.
  * Added a setting to disable fast math when bit-exact IEEE-precise output is required.

* **Bug Fixes**
  * Improved rendering behavior for scenes without Gaussian objects, preventing incorrect camera direction and hit classification.


daniela-hase (dismissed):

adenzler-nvidia:
Nice speedup! One design question before this lands: how do end-users control this if they can't reach `RenderConfig`?

With `module="unique"` + `module_options={"fast_math": ...}`, the setting is baked in at kernel creation. Unlike `mode`/`optimization_level`/`enable_mathdx_*`, which use None-means-inherit sentinels that fall back to `warp.config`, `fast_math` has no global config hook in Warp at all — its per-module default is hardcoded to `False`, and `wp.set_module_options` can't reach a dynamically created unique module. So `RenderConfig.enable_fast_math` is the only lever.

That's fine for users who construct `SensorTiledCamera` themselves, but users driving Newton through a framework that creates the sensor internally (and doesn't expose `render_config`) get fast math forced on with no opt-out. Fast math changes numerics beyond precision — denormals are flushed and IEEE special-case handling (NaN/inf) is not guaranteed — so an escape hatch may matter for reproducibility-sensitive setups.

Do we need to keep this configurable for end-users? Worth checking with the Warp team what the intended workflow is: today an explicit `module_options` value is terminal, so there's no way for a library to set its own default (fast math on) while still yielding to a user-level global override — that would need either an unset-detectable global in Warp (`wp.config.fast_math: bool | None`) or Newton resolving its own precedence (e.g. `enable_fast_math: bool | None = None` + env var fallback) at kernel creation.

adenzler-nvidia:
Follow-up to my comment above, after digging into where the fast-math win actually comes from — I think both the perf and the configurability question may have a cleaner home in Warp.

I compiled a representative intersection loop (Newton's `ray_intersect_capsule/box/sphere` plus `wp.normalize`) both ways: the entire delta is float div/sqrt. PTX swaps 40 `div.rn`/`sqrt.rn`/`rcp.rn` ops for `.approx.ftz` variants, and since ptxas expands each IEEE div/sqrt into a ~15–20 instruction Newton–Raphson sequence, SASS drops from 1800 to 1080 instructions and 64 to 53 registers (sm_120). The megakernel itself contains no transcendentals (`tan` is in camera-ray precompute, `exp` only in gaussian splatting), so of everything `--use_fast_math` implies, only `--prec-div=false --prec-sqrt=false` is doing the work here — FTZ and approximate transcendentals are just along for the ride.

So instead of baking whole-module `fast_math` into Newton, would it make more sense to ask Warp for the two narrower knobs?

1. **`prec_div` / `prec_sqrt` module options** mapping to the NVRTC flags (precedent: `fuse_fp` came out of NVIDIA/warp#379). Module-level granularity matters because much of the div/sqrt lives inside Warp's own builtins — `bvh_query_ray` computes `1.0f / dir` internally, and `mesh_query_ray` / `wp.normalize` are similar — which no Newton-side code change can reach. As a bonus, `-prec-div=false` emits full-range `div.full.f32` rather than fast_math's `div.approx.f32`, which misbehaves for divisor exponents beyond roughly ±126.
2. **Per-callsite precise opt-outs** (`wp.div_precise`, `wp.sqrt_precise`) so numerically sensitive spots inside such a module can pin back to IEEE. Warp already has the opposite direction (`wp.div_approx` / `wp.inverse_approx` from NVIDIA/warp#1199); the family is missing `wp.sqrt_approx` and has no escape hatch back to precise at all.

This also reframes my earlier question: narrower flags that don't touch denormals or transcendentals are much easier to default-on, and if Warp gives them an unset-detectable global config hook, Newton can resolve precedence itself (render kernels fast by default, user-level global wins) instead of `RenderConfig` being the only lever.

None of this blocks the PR — `fast_math` plus the `RenderConfig` opt-out is a fine state today, and we could switch to the narrower flags if/when Warp exposes them. Mainly checking whether others agree with the direction before I file the Warp feature request.

daniela-hase:
> That's fine for users who construct SensorTiledCamera themselves, but users driving Newton through a framework that creates the sensor internally (and doesn't expose render_config) get fast math forced on with no opt-out.

Is this realistic?
Whatever we may do to expose the toggle for fast math may end up in a place wrapped away by potential user-frameworks, no?

If a user drives it through a framework that doesn't expose the render config in some way, shape or form, then I'd call that a pretty bad framework 🙈

adenzler-nvidia:
fair point, yes I can't come up with a good example here either. My intention is not to block this PR at all, this is a great addition. I'm just trying to make sense of where and when we override all of the options for unique modules because it's a bit trickier with those than with per-file modules, and I stumbled over the warp-side things that are potentially missing as well while looking at this.

We can merge the PR as-is, no objections.

adenzler-nvidia (dismissed):

StafaH:
Thanks for the reviews @adenzler-nvidia and @daniela-hase. For now I will merge this PR, although I do agree that how these types of knobs are handled in newton/warp should be reconsidered, and then refactored in a future PR.

daniela-hase (approved):

PR #3459, opened 2026-07-10
Handle zero-mass particles in SolverStyle3D

### LINKED ISSUE #3451: [BUG] SolverStyle3D produces NaNs for active zero-mass particles
### Bug Description

`newton.solvers.style3d.add_cloth_mesh()` initially creates every input vertex as an active zero-mass particle, then assigns mass only to vertices referenced by valid triangles. Isolated vertices and vertices left unused after face filtering therefore remain active with zero mass.

`init_step_kernel()` checks only `ParticleFlags.ACTIVE` before computing `f_ext / mass`. A single Style3D step turns the isolated vertex position and velocity into NaN. Zero-mass particles should remain fixed and finite.

### Reproduction Script

```python
import newton
import warp as wp

builder = newton.ModelBuilder(gravity=0.0)
newton.solvers.SolverStyle3D.register_custom_attributes(builder)
newton.solvers.style3d.add_cloth_mesh(
    builder,
    pos=(0, 0, 0),
    rot=wp.quat_identity(),
    vel=(0, 0, 0),
    vertices=((0, 0, 0), (1, 0, 0), (0, 1, 0), (2, 2, 0)),
    indices=(0, 1, 2),
    density=1.0,
)
model = builder.finalize()
solver = newton.solvers.SolverStyle3D(model, iterations=1, linear_iterations=1)
state_0 = model.state()
state_1 = model.state()
solver.step(state_0, state_1, model.control(), model.contacts(), 0.01)
print(state_1.particle_q.numpy()[3])
```

### System Information

- Newton: current `main` at `9af5a9f4`
- Warp: `1.15.0.dev20260626`
- GPU: NVIDIA GeForce RTX 4080

han-xudong:
## Description

Treat Style3D particles as dynamically solvable only when they are active and have positive mass. This prevents isolated or filtered-out vertices from dividing external force by zero and keeps zero-mass rows out of optional contact and mouse-drag diagonal assembly.

Moving kinematic Style3D particles remain outside this PR's scope; zero-mass particles are kept fixed and finite.

Closes #3451.

## Checklist

## Test plan

```bash
env WARP_CACHE_ROOT=/tmp/newton-plan05-final2-cache uv run --extra dev -m newton.tests -k test_solver_style3d
env WARP_CACHE_ROOT=/tmp/newton-plan05-example-cache uv run --extra dev -m newton.tests -k test_cloth.example_cloth_hanging_style3d
uvx pre-commit run -a
git diff --check
```

## Bug fix

**Steps to reproduce:**

1. Build Style3D cloth from four vertices but only one triangle using the first three vertices.
2. Finalize the model; the isolated fourth vertex is active with zero mass.
3. Advance one Style3D step and observe non-finite position and velocity values for the isolated vertex.

**Minimal reproduction:**

```python
newton.solvers.style3d.add_cloth_mesh(
    builder,
    pos=(0.0, 0.0, 0.0),
    rot=wp.quat_identity(),
    vel=(0.0, 0.0, 0.0),
    vertices=((0, 0, 0), (1, 0, 0), (0, 1, 0), (2, 2, 0)),
    indices=(0, 1, 2),
    density=1.0,
)
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Style3D so zero-mass particles remain finite and remain fixed during solver steps.
  * Zero-mass particles are now automatically deactivated during solving (without changing the model’s stored particle flags).

* **Tests**
  * Added tests verifying isolated zero-mass particles stay stationary with zero velocity and all values remain finite after a step.
  * Added tests verifying proxy-marked zero-mass particles are deactivated internally while keeping model flags unchanged.
  * Added tests confirming solver-internal flags reflect runtime model updates.


WenchaoHuang (commented):
@han-xudong  Thanks for this contribution!

The isolated vertices could indeed cause the simulation to fail previously because `inv_A_diags` uses `mass / dt^2` in the denominator, while isolated vertices hass zero mass.
In principle, adding a check in `prepare_jacobi_preconditioner_kernel` would already be sufficient to avoid this issue.

Your fix is a robust solution. Handling the zero-mass vertices everywhere they might need to be considered is definitely the safe approach. The only overhead is one additional read from `particle_mass` per nonlinear step.

I wonder if there might be a cleaner approach. For example, similar to `ParticleFlag`, we could mark isolated vertices as `INACTIVE` upfront, instead of adding checks for them in multiple kernels. That would also avoid repeating the same condition throughout the code.

For example, during solver initialization, we could make a copy of `model.particle_flags` and mark all zero-mass particles as not ACTIVE. Then, everywhere inside the solver that currently uses `model.particle_flags` could instead use `self._particle_flags`. I think this would make the intent a bit clearer and avoid having to handle zero-mass particles repeatedly throughout the solver.

han-xudong:
Hi @WenchaoHuang, I updated the implementation in `9fcd3a78` to use a solver-local copy of `particle_flags`. The solver now clears `ACTIVE` once for non-positive-mass particles during initialization and reuses the private flags in all Style3D kernels, leaving the shared model flags unchanged and removing the repeated mass reads.

The updated CPU/CUDA tests verify zero-mass deactivation, preservation of other flag bits, and that `model.particle_flags` is not mutated. The full Style3D tests, the cloth-hanging regression, and pre-commit all pass.

WenchaoHuang (changes requested):

WenchaoHuang (dismissed):
Thanks!

han-xudong:
Hi @WenchaoHuang, I updated the implementation in `56a07083`. Zero-mass particles now receive a solver-local flag value of `0`, so no other bits remain set. I also addressed the follow-up about dynamic model changes by refreshing the local flags from `model.particle_flags` at the start of every step and reapplying zero-mass deactivation using the current masses. New CPU/CUDA coverage verifies runtime mass and flag updates, and the full Style3D and cloth-hanging regressions pass. Could you please take another look?

newton/tests/test_solver_style3d.py
```diff
@@ -33,6 +34,97 @@ def test_constructor_precomputes_fixed_pd_matrix(test, device):
...
+    )
+    model = builder.finalize(device=device)
+    solver = newton.solvers.SolverStyle3D(model, iterations=1, linear_iterations=1)
+
+    masses = model.particle_mass.numpy()
+    masses[0] = 0.0
+    model.particle_mass.assign(masses)
+    model_flags = model.particle_flags.numpy()
+    model_flags[1] = 0
+    model.particle_flags.assign(model_flags)
```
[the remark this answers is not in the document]
han-xudong: Addressed in `4fda004f`. Both runtime test snapshots now use `.numpy().copy()`, so the CPU assertions compare against independent arrays rather than aliased Warp buffers. The four CPU/CUDA solver-flag tests and `pre-commit` pass.

CHANGELOG.md
```diff
@@ -92,6 +92,7 @@
 - Tune VBD contact settings in the `basic_shapes` and `cable_bundle_hysteresis` examples for more consistent friction and recovery behavior.
 - Fix USD import ignoring ancestor material bindings with `strongerThanDescendants` strength when a mesh authors `material:binding` without applying `MaterialBindingAPI`: material resolution now uses UsdShade's canonical `ComputeBoundMaterial` unconditionally, which also adds collection-based binding support. Prims authoring bindings without the applied schema are invalid USD and now surface USD's own warning (once per prim per import) — fix such assets with `usdchecker` or `usd-validation-nvidia`.
 - Fix `ModelBuilder.add_usd()` to honor `ignore_paths` in the custom-frequency traversal, so prims under ignored subtrees no longer register spurious custom-frequency rows in two-pass import workflows. (#3406)
+- Keep zero-mass Style3D particles fixed and finite during solver steps. (#3451)
```
WenchaoHuang: Replace this with "Fix Style3D solver divergence caused by isolated vertices."

WenchaoHuang (changes requested):

CHANGELOG.md
han-xudong: Updated in `17816959` with the requested wording: “Fix Style3D solver divergence caused by isolated vertices.”

WenchaoHuang:
> Hi @WenchaoHuang, I updated the implementation in `56a07083`. Zero-mass particles now receive a solver-local flag value of `0`, so no other bits remain set. I also addressed the follow-up about dynamic model changes by refreshing the local flags from `model.particle_flags` at the start of every step and reapplying zero-mass deactivation using the current masses. New CPU/CUDA coverage verifies runtime mass and flag updates, and the full Style3D and cloth-hanging regressions pass. Could you please take another look?

I misread it as `~ParticleFlag.ACTIVE`. Your original implementation was correct. Removing only `ParticleFlag.ACTIVE` bit is the right behavior.

han-xudong:
Hi @WenchaoHuang, I restored the original bit-mask behavior in `941ded78`: zero-mass particles now clear only `ParticleFlags.ACTIVE`, while preserving `PROXY` and any other metadata bits. The regression test again verifies this explicitly on CPU and CUDA. The full Style3D suite, cloth-hanging regression, and `pre-commit` all pass.

WenchaoHuang (approved):
LGTM!

PR #2206, opened 2026-03-24
Align articulated body_qd with COM velocity contract

Idate96:
This PR is intentionally a follow-up to [#2181](https://github.com/newton-physics/newton/pull/2181).

The order is:

1. Review and merge `#2181`
2. Review this PR as the COM-convention follow-up

I'll leave as draft until 2181 is merged

## Description

This PR makes Newton's public articulated-body velocity contract internally
consistent with the existing docs and state semantics:

- `State.body_qd[:3]` is the world-frame linear velocity of the body COM
- `State.body_qd[3:6]` is world-frame angular velocity
- descendant `FREE` / `DISTANCE` `State.joint_qd` remains a parent-frame twist
  whose linear part is referenced at the child COM
- descendant `FREE` / `DISTANCE` `Control.joint_f` remains a world-frame wrench
  applied at the child COM

This also matches Kamino’s existing CoM-based body-state convention.

Before this PR, those pieces did not line up:

- articulated FK wrote many link linear velocities at the body origin rather
  than the COM
- `eval_ik()` / `eval_jacobian()` / `eval_mass_matrix()` therefore did not all
  agree on the same reference point
- Featherstone had an additional mismatch because its internal free-joint speed
  basis is not the same as the public COM-based contract

Mixed reference points break the mathematical identities Newton should satisfy:

- the public Jacobian should reproduce the published body twist, i.e.
  `J_link @ joint_qd == state.body_qd[link]`
- the public mass matrix should match kinetic energy computed from the same
  published body twists, i.e. `0.5 * qd^T H qd == T(body_qd)`
- FK and IK should round-trip without silently changing the velocity reference
  point underneath the user

If `body_qd` is origin-referenced in one path but COM-referenced in another,
then `J @ qd = v` is false for the documented `v`, the mass matrix is no
longer consistent with the public twist convention, and downstream code cannot
reason about energy, momentum, or task-space kinematics on a single basis.

This PR fixes that by:

- writing COM-referenced `body_qd` consistently in public FK and solver FK
  paths
- aligning `eval_ik()`, `eval_jacobian()`, and `eval_mass_matrix()` to the same
  COM reference point
- keeping Featherstone's internal free-joint math unchanged, but adding
  explicit public/internal boundary conversion for `FREE` / `DISTANCE`
  `joint_qd`
- routing public `FREE` / `DISTANCE` `joint_f` through Featherstone's
  `body_f_ext` path as a world-frame COM wrench
[...]

PR #3468, opened 2026-07-10
Fix misfiled and misframed 1.4 changelog entries

adenzler-nvidia:
## Description

Fixes two entries in the 1.4 release notes:

- The `SolverFeatherstone` entry for #2539 told callers they must pass FREE/DISTANCE `joint_qd` as child-COM linear velocity followed by angular velocity in the joint parent frame. That caller-facing contract has been in place since 1.1.0 (#2206); in 1.3 the solver translated to its internal basis at the step boundary, so no caller action is needed when upgrading. The entry is rewritten around the user-visible change in #2539: the solve is now centered at the root body COM, which changes floating-base results numerically while improving stability far from the world origin.
- The `deterministic` constructor argument bullet (#3300) shipped after v1.3.0 but was inserted under the `[1.3.0]` heading, so the 1.4 notes omit the feature while the shipped 1.3.0 notes claim it. Move it to `[Unreleased]`.

## Checklist

- [ ] New or existing tests cover these changes (changelog-only change)

## Test plan

```
uvx pre-commit run --files CHANGELOG.md
```

## Summary by CodeRabbit

* **Documentation**
  * Added an Unreleased changelog entry documenting the new `deterministic` constructor option for supported solvers.
  * Clarified that floating-base articulations are solved about the root body’s center of mass.
  * Reorganized the changelog entry to reflect the current release status.


eric-heiden (approved):
Thanks!

PR #3470, opened 2026-07-10
Prevent deterministic MuJoCo test leakage

### LINKED ISSUE #3420: Flaky test: test_mujoco_hydroelastic_penetration_depth_cuda_0
This GPU test failed in the merge queue, while the later combined queue head passed the same required GPU suite.

Failing job: https://github.com/newton-physics/newton/actions/runs/29010354462/job/86104250152

```text
test_mujoco_hydroelastic_penetration_depth_cuda_0
AssertionError: np.float32(0.5426573) not greater than 1.0 : Case 0: ratio 0.543 too low (measured=0.000081, expected=0.000150)
```

Investigate the instability and make the test outcome reliable.

### LINKED ISSUE #3421: Flaky test recurrence: test_stacked_small_primitive_cubes_hydroelastic_mujoco_warp_cuda_0
This previously tracked flaky test failed again in the merge queue. #1620 was closed after extended clean testing with a request to open a new issue if it recurred.

Failing job: https://github.com/newton-physics/newton/actions/runs/29010354462/job/86104250152

```text
test_stacked_small_primitive_cubes_hydroelastic_mujoco_warp_cuda_0
AssertionError: Unexpected output:
'Module _friction_dof__locals__kernel_88836a0e 88836a0 load on device 'cuda:0' took 0.81 ms  (cached)
...
nefc overflow - please increase njmax to 560
nefc overflow - please increase njmax to 552
nefc overflow - please increase njmax to 552
nefc overflow - please increase njmax to 552'
```

Investigate why this output intermittently escapes the test and make the test reliable.

mzamoramora-nvidia:
## Description

Prevent deterministic MuJoCo-Warp test configuration from affecting later
hydroelastic tests in the same Python process.

`test_mujoco_sparse_articulation_construction_cuda_0` opts into deterministic
MuJoCo-Warp execution. Warp and MuJoCo-Warp retain generated kernels and GPU
runtime state at process scope, so this configuration can affect tests that
run afterward. Depending on test ordering, the hydroelastic tests then use
the contaminated solver state and fail with incorrect penetration depth and
repeated `nefc overflow` messages.

This PR:

- Runs only the deterministic sparse-articulation test in a subprocess, so
  its generated kernels and runtime state are discarded when the test exits.
- Explicitly configures the affected hydroelastic solvers with
  `DeterministicMode.NOT_GUARANTEED`. Without this argument, `SolverMuJoCo`
  inherits `wp.config.deterministic`.
- Keeps the hydroelastic tests in the normal test process and does not isolate
  other deterministic MuJoCo tests.

This is a test-only change and does not modify production solver behavior.

Closes #3420.
Closes #3421.

## Checklist

- [x] The documentation is up to date with these changes (no documentation changes required)
- [x] `CHANGELOG.md` has been updated (not applicable; no user-facing change)

## Test plan

Run the deterministic sparse-articulation test immediately before the two
affected hydroelastic tests:

```bash
uv run --no-sync -m newton.tests \
  -k test_mujoco_sparse_articulation_construction_cuda_0 \
  -k test_mujoco_hydroelastic_penetration_depth_cuda_0 \
  -k test_stacked_small_primitive_cubes_hydroelastic_mujoco_warp_cuda_0 \
  --strict-warnings \
  --no-cache-clear \
  --serial-fallback
```

Results over 10 independent runs:

| Revision | Passed | Failed |
| --- | ---: | ---: |
| `main` | 0 | 10 |
| This PR | 10 | 0 |

On `main`, both hydroelastic tests failed in every run and produced roughly
2,986-2,998 `nefc overflow` messages per run. Running the two hydroelastic
tests without the preceding deterministic test passed, confirming that the
failure depends on process-local deterministic state.

Additional verification:

```bash
uvx pre-commit run -a
```

## Bug fix

**Steps to reproduce:**

1. Check out `main`.
2. Run the three-test command above in one Python process.
3. Observe that the deterministic test passes.
4. Observe that both subsequent hydroelastic tests fail:
   - The penetration-depth ratio is approximately `0.54`, below the required
     minimum of `1.0`.
   - The stacked-cubes test reports repeated `nefc overflow` output.

Although the failures appear flaky in CI because they depend on test ordering,
the ordered reproduction above failed in 10 out of 10 local runs.

## Summary by CodeRabbit

* **Tests**
  * Improved determinism test isolation by running checks in independent processes.
  * Reduced cross-test interference from persisted runtime state.
  * Standardized failure reporting with captured subprocess output.
  * Expanded consistent isolation across particle, articulation, and sparse construction determinism checks.

### DIFF
--- a/newton/tests/determinism/test_solver_determinism.py
+++ b/newton/tests/determinism/test_solver_determinism.py
@@ -121,7 +121,7 @@ def _build_branching_articulation(device):
     return model
 
 
-def test_mujoco_sparse_articulation_construction(test, device):
+def _check_mujoco_sparse_articulation_construction(device):
     with wp.ScopedDevice(device):
         builder = newton.ModelBuilder(gravity=0.0)
         newton.solvers.SolverMuJoCo.register_custom_attributes(builder)
@@ -146,7 +146,30 @@ def test_mujoco_sparse_articulation_construction(test, device):
         )
 
         solver.step(state_in, state_out, model.control(), None, 1.0 / 240.0)
-        test.assertTrue(np.isfinite(state_out.joint_q.numpy()).all())
+        if not np.isfinite(state_out.joint_q.numpy()).all():
+            raise AssertionError("MuJoCo sparse articulation produced non-finite joint coordinates")
+
+
+def test_mujoco_sparse_articulation_construction(test, device):
+    # MJWarp and Warp keep generated kernels and GPU runtime state globally.
+    # Keep this deterministic configuration test isolated from later simulations.
+    code = (
+        "from newton.tests.determinism.test_solver_determinism "
+        "import _check_mujoco_sparse_articulation_construction; "
+        f"_check_mujoco_sparse_articulation_construction({str(device)!r})"
+    )
+    result = subprocess.run(
+        [sys.executable, "-c", code],
+        capture_output=True,
+        text=True,
+        timeout=600,
+        check=False,
+    )
+    test.assertEqual(
+        result.returncode,
+        0,
+        f"MuJoCo sparse articulation subprocess failed\nstdout:\n{result.stdout}\nstderr:\n{result.stderr}",
+    )
 
 
 def _make_articulation_solver(model, solver_name):
--- a/newton/tests/determinism/test_solver_determinism.py
+++ b/newton/tests/determinism/test_solver_determinism.py
@@ -1,6 +1,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
 # SPDX-License-Identifier: Apache-2.0
 
+import os
 import subprocess
 import sys
 import unittest
@@ -10,6 +11,7 @@ import numpy as np
 import warp as wp
 
 import newton
+import newton.tests.unittest_utils
 from newton._src.solvers.semi_implicit import kernels_particle as semi_implicit_particle_kernels
 from newton._src.solvers.solver import _set_module_options_if_changed
 from newton._src.solvers.vbd import particle_vbd_kernels, vbd_coupling_kernels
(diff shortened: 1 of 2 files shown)


adenzler-nvidia:
I have a warp fix in flight for this one, but we should probably still merge this PR.

newton/tests/determinism/test_solver_determinism.py
```diff
@@ -146,7 +146,30 @@ def test_mujoco_sparse_articulation_construction(test, device):
...
+        "import _check_mujoco_sparse_articulation_construction; "
+        f"_check_mujoco_sparse_articulation_construction({str(device)!r})"
+    )
+    result = subprocess.run(
+        [sys.executable, "-c", code],
+        capture_output=True,
+        text=True,
+        timeout=600,
+        check=False,
+    )
```
eric-heiden: 🔵 This child process does not inherit the strict warning policy installed by `unittest_parallel.py`. Under CI, a warning raised while constructing or stepping `SolverMuJoCo` is only captured in `stderr`, so the command can still return 0 even though the previous in-process test would fail. Please mirror the existing subprocess tests by clearing ambient `PYTHONWARNINGS` and, when `newton.tests.unittest_utils.strict_warnings` is enabled, passing the strict warning flags or environment into `subprocess.run`.

newton/tests/test_hydroelastic.py
```diff
@@ -41,16 +41,22 @@ class ShapeType(Enum):
...
+    return newton.solvers.SolverMuJoCo(
         model,
         use_mujoco_cpu=False,
         use_mujoco_contacts=False,
         njmax=500,
         nconmax=200,
         solver="newton",
         ls_iterations=100,
-    ),
+        deterministic=wp.DeterministicMode.NOT_GUARANTEED,
```
eric-heiden: 🟠 Could we remove this opt-out and instead isolate every test that enables deterministic mode in a subprocess? `SolverMuJoCo` defaults to `None`, so this matches the normal fresh-process default but differs when an earlier test has changed `wp.config.deterministic`. That masks the leak in these hydroelastic regressions and removes coverage of inherited deterministic mode both here and in the penetration-depth constructor below. Keeping the hydroelastic tests on the default and moving all deterministic MuJoCo tests into subprocesses addresses the process-wide kernel/config leakage at its source; if explicit deterministic hydroelastic coverage is needed, consider adding isolated `RUN_TO_RUN` variants.

eric-heiden:
I should have moved these determinism tests always to a subprocess, can we do this in this PR? Just to be safe, because kernels will be compiled in determinism mode within the same process and reused even in tests that don't want that.

Issue #3444, opened 2026-07-09
[REQ] Make automatic USD mesh approximation configurable


eric-heiden:
## Description

USD collider approximation is selected from the authored `physics:approximation` value, but the automatic import path calls:

```python
builder.approximate_meshes(method=remeshing_method, shape_indices=shape_ids)
```

without forwarding any approximation settings. This makes method-specific settings unavailable when users rely on automatic USD approximation.

The immediate compatibility problem is CoACD. Its default `threshold` changed from `0.5` to `0.05`, and the documented migration is to pass `threshold=0.5` to `ModelBuilder.approximate_meshes()`. That migration cannot be applied through `ModelBuilder.add_usd()` while preserving the USD-authored approximation routing. The finer default can produce more convex parts, more contact pairs, longer finalization, and different contact behavior.

Users can set `skip_mesh_approximation=True` and invoke `approximate_meshes()` manually, but that requires reconstructing the USD prim-to-method routing and is especially awkward for stages containing mixed approximation methods.

## Expected outcome

Provide a typed, method-scoped configuration path that can be set before or during USD import. Avoid a single unscoped keyword dictionary whose values may be forwarded to incompatible approximation methods.

## Proposed public API

Add a persistent builder-owned default configuration. The immediate CoACD migration should be expressible as:

```python
builder.default_mesh_approximation_cfg.coacd_threshold = 0.5
builder.add_usd("robot.usda")
```

The configuration should use algorithm-prefixed fields so autocomplete groups related settings and the importer can forward only the fields supported by the selected method. For example:

```python
cfg = builder.default_mesh_approximation_cfg

# Common automatic-approximation behavior.
cfg.raise_on_failure = False
cfg.keep_visual_shapes = False

# Select the backend used for USD physics:approximation="convexDecomposition".
cfg.convex_decomposition_backend = "coacd"

# CoACD settings.
cfg.coacd_threshold = 0.5
cfg.coacd_mcts_nodes = 20
cfg.coacd_mcts_iterations = 5
cfg.coacd_mcts_max_depth = 1
cfg.coacd_merge = False

# Quadratic simplification settings.
cfg.quadratic_target_reduction = 0.5
cfg.quadratic_target_count = None

# Convex-hull settings.
cfg.convex_hull_eps = 1.0e-6

builder.add_usd("robot.usda")
```

[...]

PR #3471, opened 2026-07-10
Make USD CoACD decomposition threshold configurable

adenzler-nvidia:
## Description

PR #3196 changed the default CoACD concavity threshold from 0.5 to 0.05, but USD imports trigger convex decomposition through the authored `physics:approximation` attribute with no way to pass settings, so the documented migration was unreachable for USD assets (#3444).

This is a first slice of the configuration surface proposed in #3444: add `ModelBuilder.MeshApproximationConfig` (currently just `coacd_threshold`) and a per-builder `default_mesh_approximation_cfg`. `approximate_meshes()` reads the builder instance's config for settings not passed explicitly, so both direct calls and USD-triggered decomposition honor it, and explicit arguments win. The threshold default now lives in one place on the config class. The remaining #3444 fields can be added incrementally on this structure, so the issue stays open.

Also fixes two pre-existing bugs on the decomposition fallback chain found while testing:

- When CoACD/V-HACD fails or is not installed, `approximate_meshes()` kept forwarding decomposition-specific kwargs to the convex-hull fallback, so the documented `threshold=0.5` migration raised a `TypeError` inside the fallback on machines without coacd and silently left the original mesh in place.
- When remeshing failed for a shape, the warning claimed a bounding-box fallback but the shape silently kept its original mesh; failed shapes now receive the documented bounding-box approximation.

## Checklist

## Test plan

Tests drive a stub `coacd` backend through the real import path, so they run without the optional dependency.

```
```

## New feature / API change

```python
import newton

builder.default_mesh_approximation_cfg.coacd_threshold = 0.5
builder.add_usd("robot.usda")  # convexDecomposition prims decompose at the configured threshold
```

## Summary by CodeRabbit

* **New Features**
  * Added a configurable per-builder CoACD mesh approximation threshold (default `0.05`), automatically applied (including USD imports) with support for per-call overrides.

* **Bug Fixes**
  * Improved mesh approximation fallback behavior when CoACD/V-HACD remeshing fails, including forwarding relevant options, ignoring non-mesh entries in `shape_indices`, and preserving the original mesh when remeshing ultimately fails.
  * Fixed several stability issues around CUDA graph capture, depth conversion sentinel handling, and frame retrieval.

* **Documentation**
  * Updated guidance for mesh approximation and changelog entries.

* **Tests**
  * Expanded unit tests to validate threshold forwarding, overrides, USD behavior, and remeshing fallback scenarios.


newton/_src/sim/builder.py
```diff
@@ -7191,13 +7211,17 @@ def get_shape_custom_attributes(shape: int) -> dict[str, Any] | None:
...
                             continue
                 # note we need to copy the mesh to avoid modifying the original mesh
                 self.shape_source[shape] = self.shape_source[shape].copy(vertices=rmesh.vertices, indices=rmesh.indices)
                 # mark convex_hull result as convex mesh type for efficient collision detection
                 if method == "convex_hull":
                     self.shape_type[shape] = GeoType.CONVEX_MESH
                 remeshed_shapes.add(shape)
+            if remesh_failed:
+                # route the shapes that failed (not in remeshed_shapes) into the fallback below
+                method = "bounding_box"
```
eric-heiden, line 7239: 🔴 `approximate_meshes()` crashes when `shape_indices` includes a non-mesh shape. For a box, `shape_source[shape]` is `None`; the convex-hull attempt is caught, but this new branch then routes the same shape into the bounding-box fallback, where `mesh.vertices` raises `AttributeError`. Please filter explicit indices to `GeoType.MESH` before remeshing and fallback processing, or reject non-mesh indices consistently.

<details>

<summary>Minimal example</summary>

```py
import newton

shape = builder.add_shape_box(body=-1)
builder.approximate_meshes(method="convex_hull", shape_indices=[shape])
```

This fails at `builder.py:7232` because `mesh` is `None`.

</details>

adenzler-nvidia: Fixed in ef231619e: explicit `shape_indices` are now filtered to `GeoType.MESH` next to the default-selection branch, so remeshing, the fallbacks, and `keep_visual_shapes` never see a shape without a mesh source. This also covers the pre-existing crash when passing a primitive index directly to `bounding_box`/`bounding_sphere`; already-approximated `CONVEX_MESH` shapes pass through unchanged, matching the default selection. Covered by `test_mesh_approximation_ignores_non_mesh_shapes`.

newton/tests/test_model.py
```diff
@@ -798,6 +799,62 @@ def npsorted(x):
...
+
+        with mock.patch.dict(sys.modules, {"coacd": fake_coacd}):
+            builder.approximate_meshes(method="coacd", shape_indices=[shape])
+        self.assertEqual(captured["threshold"], 0.5)
+
+        # an explicit argument overrides the builder default
+        captured.clear()
+        with mock.patch.dict(sys.modules, {"coacd": fake_coacd}):
+            builder.approximate_meshes(method="coacd", shape_indices=[shape], threshold=0.2)
+        self.assertEqual(captured["threshold"], 0.2)
```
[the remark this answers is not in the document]
adenzler-nvidia: Done in 387661a93 — the override check now runs on a fresh shape.

newton/tests/test_model.py
```diff
@@ -798,6 +799,62 @@ def npsorted(x):
...
+        builder = ModelBuilder()
+        box = newton.Mesh.create_box(
+            1.0, 1.0, 1.0, duplicate_vertices=False, compute_normals=False, compute_uvs=False, compute_inertia=False
+        )
+        shape = builder.add_shape_mesh(body=-1, mesh=box)
+        with mock.patch.dict(sys.modules, {"coacd": None}), warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+            builder.approximate_meshes(method="coacd", shape_indices=[shape], threshold=0.5)
+        # coacd unavailable: the convex_hull fallback must not receive coacd-specific kwargs
+        self.assertEqual(builder.shape_type[shape], newton.GeoType.CONVEX_MESH)
```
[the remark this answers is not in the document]
adenzler-nvidia: The type assertion does discriminate end-to-end: `remesh_convex_hull()` accepts no `**kwargs`, so a leaked `threshold` raises `TypeError` and the shape lands at `GeoType.BOX` via the bounding-box fallback — reverting the kwargs drop makes this test fail with `BOX != CONVEX_MESH`. Renamed the test after the contract it verifies (`test_mesh_approximation_coacd_unavailable_falls_back_to_convex_hull`) in 74252006c instead of adding a mock.

newton/tests/test_model.py
```diff
@@ -798,6 +799,78 @@ def npsorted(x):
...
+            captured.update(kwargs)
+            return [cmesh]
+
+        fake_coacd.run_coacd = run_coacd
+
+        with mock.patch.dict(sys.modules, {"coacd": fake_coacd}):
+            builder.approximate_meshes(method="coacd", shape_indices=[shape])
+        self.assertEqual(captured["threshold"], 0.5)
+
+        # an explicit argument overrides the builder default
```
jcarius-nv, line 824: These two comments (here and below in `test_mesh_approximation_ignores_non_mesh_shapes`) restate behavior that the following calls and assertions already make clear. Could we remove them? Per the repository guidance, comments should explain non-obvious intent, constraints, or edge cases rather than narrate the code.

newton/_src/sim/builder.py
```diff
@@ -7099,7 +7119,7 @@ def get_shape_custom_attributes(shape: int) -> dict[str, Any] | None:
                         if method == "coacd":
                             cmesh = coacd.Mesh(mesh.vertices, mesh.indices.reshape(-1, 3))
                             coacd_settings = {
-                                "threshold": 0.05,
+                                "threshold": self.default_mesh_approximation_cfg.coacd_threshold,
```
jcarius-nv, line 7134: The empty-result branch below (`if len(decomposition) == 0: continue`) leaves the original shape as `GeoType.MESH` instead of entering the documented CoACD/V-HACD → convex-hull fallback. I reproduced this on the PR head with a stub CoACD backend returning `[]`; the shape remained `MESH`. Could we route an empty result through the fallback chain and add a regression test for it?

jcarius-nv (commented):
Two inline review comments.

newton/tests/test_model.py
jcarius-nv: let's ignore

newton/_src/sim/builder.py
jcarius-nv: Confirmed that this is pre-existing rather than introduced by this PR, but it is another gap in the fallback contract being tightened here. If CoACD/V-HACD returns no parts, the shape is not successfully remeshed: the method returns an empty set, leaves it as `MESH`, emits no warning, and does not raise even with `raise_on_failure=True`.

The intended behavior is to treat an empty result as decomposition failure—raise when requested, otherwise warn and route only the affected shape through the convex-hull fallback. I filed #3507 to track that work, including regression coverage for both modes and mixed-shape batches. No change is requested in this PR; this will be fixed in the follow-up issue.

jcarius-nv (dismissed):

camevor (dismissed):

newton/_src/sim/builder.py
```diff
@@ -7034,6 +7051,9 @@ def get_shape_custom_attributes(shape: int) -> dict[str, Any] | None:
                 for i, stype in enumerate(self.shape_type)
                 if stype == GeoType.MESH and self.shape_flags[i] & ShapeFlags.COLLIDE_SHAPES
             ]
+        else:
+            # primitives have no mesh source and convex meshes are already approximated; pass them through
+            shape_indices = [i for i in shape_indices if self.shape_type[i] == GeoType.MESH]
```
mzamoramora-nvidia: Blocking: this filter also drops explicit `GeoType.CONVEX_MESH` indices. Those shapes still have a valid `Mesh` source and, before this PR, `approximate_meshes(method="bounding_box", shape_indices=[convex_shape])` returned the shape index and converted it to `BOX`; on this head it returns an empty set and leaves it as `CONVEX_MESH`. This is an undocumented public behavior change. Could we accept both `MESH` and `CONVEX_MESH` here, filtering only incompatible/no-source shapes, and add a regression test?

newton/tests/test_import_usd.py
```diff
@@ -2699,6 +2701,59 @@ def npsorted(x):
...
+
+        fake_coacd.run_coacd = run_coacd
+
+        with mock.patch.dict(sys.modules, {"coacd": fake_coacd}):
+            builder = newton.ModelBuilder()
+            builder.add_usd(stage)
+        self.assertEqual(captured["threshold"], 0.05)
+
+        captured.clear()
+        with mock.patch.dict(sys.modules, {"coacd": fake_coacd}):
```
mzamoramora-nvidia: Blocking: this test is order-dependent because the two separate `mock.patch.dict(sys.modules, ...)` contexts restore the module mapping. The first `add_usd()` lazily imports SciPy through the convex-hull queue; exiting that context removes those modules, and the second import fails with `ImportError: cannot load module more than once per process` in a clean isolated run. It passes only when SciPy was already imported. Could both imports share one patch context so this test passes independently?

newton/_src/sim/builder.py
jcarius-nv: Fixed in e9163475. Explicit `shape_indices` now retain both `GeoType.MESH` and `GeoType.CONVEX_MESH`; incompatible non-mesh shapes are still ignored. The regression test first converts the selected mesh to `CONVEX_MESH`, then explicitly approximates that same shape as `bounding_box` and verifies that the method returns `{mesh_shape}` and changes the type to `BOX`. This restores the pre-PR behavior.

newton/tests/test_import_usd.py
jcarius-nv: Fixed in e9163475. I used a narrower module override instead of `mock.patch.dict(sys.modules, ...)`: `_patch_sys_module` saves and restores only the `coacd` entry, and both USD imports run under one override. This avoids rolling back SciPy or any other modules lazily imported by `add_usd()`. The test passes in isolation, and a same-process follow-up `import scipy.spatial` also succeeds.

newton/_src/sim/builder.py
mzamoramora-nvidia: Thanks! Confirmed that explicit `CONVEX_MESH` re-approximation is restored and covered by the updated test. This concern is resolved.

newton/tests/test_import_usd.py
mzamoramora-nvidia: Thanks! The USD test is fixed. I still reproduce the same order-dependent SciPy failure from the full `sys.modules` patch in the newly added `test_mesh_approximation_coacd_unavailable_falls_back_to_convex_hull` test. Running it immediately before `test_mesh_approximation_ignores_non_mesh_shapes` makes the latter fail with `ImportError: cannot load module more than once per process`. Could the `test_model.py` module patches use `_patch_sys_module` as well?

newton/_src/sim/builder.py
mzamoramora-nvidia: One small follow-up: the changelog still says indices of shapes other than `GeoType.MESH` are ignored, but explicit `GeoType.CONVEX_MESH` shapes are now supported again. Could we update that wording to match the restored behavior?

jcarius-nv: Follow-up fixed in 7a1d518d. The changelog now states that explicit `shape_indices` processes both `GeoType.MESH` and `GeoType.CONVEX_MESH` entries while ignoring incompatible types, matching the restored implementation and regression test. The branch is also updated to latest `main` in b6ef7371.

mzamoramora-nvidia (approved):

newton/tests/test_import_usd.py
jcarius-nv: Follow-up fixed in 7a1d518d. `patch_sys_module` now lives in `newton.tests.unittest_utils`, and every CoACD module stub added in both `test_import_usd.py` and `test_model.py` uses it instead of a full `sys.modules` snapshot. The exact two-test sequence from this comment now passes, as do all eight `test_mesh_approximation` tests. The branch is also updated to latest `main` in b6ef7371.

PR #3472, opened 2026-07-10
Viewer: Command mouse-picking acceleration instead of force

adenzler-nvidia:
## Description

Replace the mouse-picking force heuristic (`force_multiplier = 10 + mass`, followed by a post-hoc force clamp) with acceleration-space control: the spring-damper computes a commanded acceleration of the pick point, clamps its magnitude to `pick_max_acceleration`, and scales the result by the picked body's effective mass (total articulation mass, as before). The per-body inertia torque bound from #3375 is unchanged.

Picking gains become mass-independent (`pick_stiffness` [1/s²], `pick_damping` [1/s]), which is the convention used by MuJoCo (`mjv_applyPerturbForce` scales the spring by the effective mass at the grab point), Isaac Sim (both grab modes scale gains and force caps by mass/inertia), and Genesis. Heavy-body behavior is nearly unchanged since `10 + m ≈ m` for `m ≫ 10`. Light bodies previously received a large minimum force that permanently saturated the acceleration clamp (bang-bang response at the clamp boundary); they now get a smooth, bounded spring response.

This removes the hand-tuned per-example picking gains in the cable and brick-stacking examples (spanning 0.01 to 100 to compensate for the mass-dependent formula); the defaults now apply everywhere. Follow-up commits remove dead code from `Picking` and make the gains construction-only: `pick_stiffness`, `pick_damping`, and `pick_max_acceleration` are set through the constructor and live solely in the device-side pick state, with no runtime accessors. Runtime tuning was unreachable through public API anyway — the viewers construct `Picking` internally and the class is not publicly exported.

Follow-up to #3375 and realizes the "general inertia-aware stabilization can be developed and reviewed separately" that #3407 defers to, superseding that PR's remaining scope.

Draft pending interactive feel-testing of the affected examples.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_viewer_picking
uv run --extra dev -m newton.tests -k test_cable
```

New `test_picking_acceleration_mass_independent` asserts the commanded acceleration is independent of body mass below the clamp. Manual feel-testing of `cable_cross_slide_table`, `cable_pile`, `cable_y_junction`, and a heavy articulated scene is still pending (hence draft).

## New feature / API change

```python
# Picking gains are now accelerations of the pick point, independent of body
# mass, and are configured at construction only:
#   accel = pick_stiffness * (target - pick_point) - pick_damping * pick_point_velocity
#   force = clamp(accel, pick_max_acceleration) * effective_mass
picking = Picking(model, pick_stiffness=50.0, pick_damping=5.0)  # [1/s²], [1/s]
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated mouse-picking to use acceleration-based `pick_stiffness`/`pick_damping`, then convert to force using the picked body’s effective mass for mass-independent behavior.
  * Clamp is applied to the commanded acceleration before force conversion for more consistent limit handling.
  * Clarified in the changelog that custom picking gains from the old formulation must be retuned (defaults unchanged).
  * Updated cable and contact examples to no longer override viewer picking parameters.

* **Tests**
  * Expanded picking initialization assertions to validate device-backed `pick_state`.
  * Added coverage to confirm commanded acceleration is identical across body masses below the clamp.


eric-heiden:
Thanks for moving the pick controller into acceleration space. Clamping there instead of on force is a much more intuitive knob, and the mass-independence test is a nice addition.

One thing worth sorting out before this lands: the commanded acceleration is converted back into a force using the entire articulation's mass, so picking a link inside a multi-body system still applies far more acceleration than requested.

newton/_src/viewer/kernels.py
```diff
@@ -92,22 +92,19 @@ def apply_picking_force_kernel(
...
+        pick_state[0].pick_damping * pick_vel
     )
 
-    # Clamp force magnitude to prevent runaway divergence on light objects (#2361).
-    # Uses the effective mass (total articulation mass for linked bodies,
-    # own mass for free bodies) so picking a light robot link still allows
-    # enough force to move the whole chain.
+    # Clamp acceleration magnitude to prevent runaway divergence on light objects (#2361).
     max_acceleration = pick_state[0].pick_max_acceleration * 9.81
-    max_force = max_acceleration * pick_effective_mass[pick_body]
```
eric-heiden, line 107: 🟠 `pick_effective_mass` holds the total mass of the whole articulation, so using it as the acceleration-to-force multiplier over-commands links in articulated systems.

`Picking._compute_effective_mass` in `newton/_src/viewer/picking.py` assigns every body in an articulation the summed mass of that articulation. That was safe while the value only capped the force, since an overestimate just meant a looser cap. Now it scales every pick force, so the pick point no longer tracks the acceleration that was commanded. A link's operational-space effective mass depends on the articulated mass matrix, the joint configuration, the force direction, and the pick offset, and the articulation total is not a good stand-in for it.

For a light distal link the operational-space mass is close to the link's own mass, so the resulting acceleration is scaled by roughly `articulation_mass / link_mass`. Grabbing a 0.1 kg gripper finger on a 10 kg arm commands about 100x the requested acceleration, which is the opposite of the mass independence this change is going for. Fixed-base mechanisms are the worst case, since the base mass is summed in but never moves.

Two ways out:

- Derive the operational-space mass at the pick point from the articulation Jacobian and mass matrix.
- Keep the articulation total purely as a safety cap on the force, use the picked body's own mass as the multiplier here, and document the result as a heuristic rather than a true acceleration command.

The current tests will not catch either behavior: `test_picking_acceleration_mass_independent` uses a single free body, where effective mass equals body mass, and `test_picking_torque_limit_cable` only checks that state stays finite and that peak angular speed stays under 3.0. An articulated case that measures the actual pick-point acceleration for different link masses and configurations would pin this down.

Worth updating in the same pass: the `_compute_effective_mass` docstring still describes the value as being for "picking force clamping" and for allowing "enough force to move the whole chain", which no longer matches how it is used.

CHANGELOG.md
eric-heiden, line 1: Please use changelog fragments now rather than editing the changelog directly.

eric-heiden:
I'm not sure if this is resulting from my mass-scaling comment but now it is practically impossible to lift a robot by picking low-mass links, like the end-effectors. I have to pick from the torso to lift up the G1 in this example. Same for the ANYmal robot where it's now impossible to lift it from its feet. I think this is not intentional, right? The picking should always be possible and not become too "weak" just because you chose the wrong link.

https://github.com/user-attachments/assets/8e935bca-47d7-4a24-86a9-7001df47228f

jvonmuralt:
> I'm not sure if this is resulting from my mass-scaling comment but now it is practically impossible to lift a robot by picking low-mass links, like the end-effectors. I have to pick from the torso to lift up the G1 in this example. Same for the ANYmal robot where it's now impossible to lift it from its feet. I think this is not intentional, right? The picking should always be possible and not become too "weak" just because you chose the wrong link.
> picking_G1.mp4
With the operational-space mass it looks better ( I just pushed a commit), but it's still hard to lift by the feet.
There is an option to add a lower bound on the operational-space inertia at the pick point,
e.g. as a fraction of the articulation's total mass, what do you think?

eric-heiden:
> With the operational-space mass it looks better ( I just pushed a commit), but it's still hard to lift by the feet.
> There is an option to add a lower bound on the operational-space inertia at the pick point,
> e.g. as a fraction of the articulation's total mass, what do you think?

Thanks, yes this looks better! Maybe give this bounded inertia trick a try so the picking becomes more responsive at the distal links.
I see you also disabled picking on some of the cable demos and I notice picking doesn't work on the balance bird example anymore - is VBD not supported at all? I think it would be good to bring this feature also to VBD, picking for cables shouldn't be disabled.

newton/_src/viewer/picking.py
```diff
@@ -118,17 +125,79 @@ def _apply_picking_force(self, state: newton.State) -> None:
...
+        mass = float(model.body_mass.numpy()[body])
+        upper = max(float(self._pick_effective_mass.numpy()[body]), mass)
+        lower = min(max(mass, self.pick_min_inertia_fraction * upper), upper)
+        lam = np.eye(3) * upper
+
+        joints = np.nonzero(model.joint_child.numpy() == body)[0] if model.articulation_count else []
+        if len(joints) and mass > 0.0:
+            joint = int(joints[0])
+            art = int(model.joint_articulation.numpy()[joint])
+            first = int(model.articulation_start.numpy()[art])
```
eric-heiden, line 166: 🔴 `joint_articulation` is `-1` for a joint that belongs to no articulation, and nothing guards that here, so `art = -1` falls through into negative-index lookups and crashes the viewer on click.

`articulation_start[-1]` and `articulation_end[-1]` wrap around to the one-past-end sentinels rather than failing, so `first` and `ndof` come out quietly wrong. On an 8-segment closed rod that gives `first = 8`, `row = 6 * (7 - 8) = -6`, and `ndof = -4`. The negative `ndof` truncates the Jacobian columns, `jac[row + 3 : row + 6]` becomes `jac[-3:0]`, which is empty, and the cross product raises:

```
File "newton/_src/viewer/picking.py", line 180, in _update_os_inertia
    jac_point = jac[row : row + 3] + np.cross(jac[row + 3 : row + 6].T, r).T
ValueError: Both input arrays must be (arrays of) 3-dimensional vectors, but they are 0 and 3 dimensional instead.
```

`ValueError` is not caught by the `except np.linalg.LinAlgError` a few lines below, and the GUI path (`on_mouse_press` to `handle_mouse_press` to `pick`) has no handler either, so it comes straight out of the mouse callback.

Two supported models reach this today:

- A loop-closing `CABLE` joint from `add_rod(closed=True)`, which is deliberately left outside the articulation. `newton/examples/vbd/example_cable_michell_threshold.py` builds exactly that, and the ring's first body is the child of only the closing joint.
- A standalone world-root joint that is never wrapped in an articulation, which `ModelBuilder` explicitly permits.

```suggestion
        joints = np.nonzero(model.joint_child.numpy() == body)[0] if model.articulation_count else []
        art = int(model.joint_articulation.numpy()[joints[0]]) if len(joints) else -1
        if art >= 0 and mass > 0.0:
            joint = int(joints[0])
            first = int(model.articulation_start.numpy()[art])
```

Two things worth checking while you are in here:

The docstring says that where the Jacobian is unavailable, as for `CABLE` joints, the articulation total stands in. That happens for open cables, but only because `H` comes back all zeros and `np.linalg.solve` raises `LinAlgError`, which the handler swallows. It is worth making that fallback explicit rather than leaving it resting on a singular matrix.

The guard alone will not make closed rods safe. For ring bodies that do have `art >= 0`, `eval_jacobian` walks `joint_ancestor`, which is cyclic on a closed rod (joint 0's ancestor is the closing joint), so the `while j != -1` loop in `eval_articulation_jacobian` never terminates and the pick pins the GPU. The same rod built open finishes in milliseconds. The cycle is not introduced by this PR, but this PR is what first makes the viewer call that path on cable articulations, so a click can now hang the app.

eric-heiden:
Coming back to the balance bird question from earlier, since it is still open: VBD is not the reason.

`SolverVBD` does consume `state.body_f` for rigid bodies. The path is `_integrates_rigid_bodies` to `forward_step_rigid_bodies` to `integrate_rigid_body`, where `f0 * inv_mass` enters the velocity update. Dragging the bird 10 cm over 120 frames gives an identical trajectory under both solvers:

| solver | z start | z end | peak z |
|---|---|---|---|
| XPBD | 0.12975 | 0.22833 | 0.25936 |
| VBD | 0.12975 | 0.22833 | 0.25936 |

So picking works under VBD. What changed for this example is the pull strength, and it is a side effect of the new formula rather than anything solver specific.

🟠 **The pull on light free bodies is about 22x weaker than before.**

The bird is a single free body of 0.478 kg. The old multiplier was `10.0 + body_mass` = 10.478. The new one is the operational-space mass, which for a free body is exactly `body_mass` = 0.478. Same gains, same drag, so the force drops by `(10 + m) / m` = 21.9x:

```
old body_f = [26.195305,  0, 0         | 0, 0.6227884,  0]
new body_f = [ 1.1953051, 0, 4.6903777 | 0, 0.02841815, 0]
```

That is the "picking doesn't work anymore" symptom, and it will hit any light free body, not just this example. The clamp is not involved: at `5 * 9.81 * 0.478` = 23.45 N neither force is close to it. The `+10.0` used to dominate the mass term for anything under a few kg, and dropping it is a real behavior change for those objects even though the formula is more principled now.

🟠 **Weight cancellation removes the effect the demo exists to show.**

The bird balances because gravity on the two low weights restores it. `pick_force -= gravity[world] * body_mass[pick_body]` makes the picked body weightless, so that restoring behavior disappears while it is held. Setting both gains to zero, so the gravity cancellation is the only surviving term:

| | z start to end | up_z over 180 frames |
|---|---|---|
| not picked | 0.129000 to 0.128100 | wobbles, 0.9728 to 0.9986 to 0.9893 |
| picked, gains zeroed | 0.129000 to 0.129000 | 1.00000 flat, all 180 frames |

The bird hovers 0.9 mm above its resting height and never touches the pedestal. No spring is acting here, this is purely the cancellation term. Cancelling weight makes sense for a distal link whose weight the articulation carries anyway, but for a free body it turns picking into a gravity toggle. Worth deciding whether it should apply to free bodies at all.

One smaller detail in the same block: the cancellation is applied after the `max_force` clamp, so the wrench that actually reaches `body_f` can exceed the `pick_max_acceleration` bound the docstring advertises. Folding it in before the clamp would keep that limit honest.

jvonmuralt:
sorry, I requested a review again by mistake. I will try to address the comments, will let you know when I have something

PR #3478, opened 2026-07-10
Expose public viewer camera speed parameter

### LINKED ISSUE #3439: [BUG] Examples of modifying ViewerGL camera speed fail due to _cam_speed being moved to ViewerGUI
### Bug Description

The example `example_contacts_rj45_plug.py` modifies the camera speed from the fault value of 4.0 down to 0.2 due to the small scale of the simulation objects. This is useful for Newton library users who similarly simulate small-scale objects, and thus would like to modify their camera speeds.

Setting the camera speed worked in v.1.2.0, but starting in v1.3.0 this no longer works, and the example becomes difficult to view. This is caused by the following lines:

```python
#  example_contacts_rj45_plug.py: 389-390
       if hasattr(self.viewer, "_cam_speed"):
            self.viewer._cam_speed = 0.2
```

In Newton v1.3.0, the `_cam_speed` attribute has been moved from `ViewerGL` to `ViewerGui`. Thus here the `hasattr` never returns true, and even if it did, `self.viewer._cam_speed` is not the right property to set.

These lines should instead be:

```python
#  example_contacts_rj45_plug.py: 389-390
       if hasattr(self.viewer.gui, "_cam_speed"):
            self.viewer.gui._cam_speed = 0.2
```

This addresses the ViewerGui object within ViewerGL and fixes the camera speed setting.

I have implemented this fix and will create a pull-request immediately after submitting this bug.

### Reproduction Script

```python
# Just run the existing Newton example_contacts_rj45_plug.py script, and move the camera using WASD.
```

### System Information

- Python: 3.12.3
- Newton: commit #9af5a9f4
- Warp: warp-lang v1.15.0.dev20260626
- OS: Ubuntu 24.04.3
- Nvidia driver: 580.159.03
- CUDA version: 13.0

eric-heiden:
## Description

Add `ViewerBase.camera_speed` as the public API for configuring WASD/QE camera translation speed in meters per second. Interactive GL and RTX viewers now read this viewer-level setting instead of backend-specific `ViewerGui._cam_speed` state.

Update the RJ45 contact example to use the public property, document the API, and add regression coverage for its default, validation, zero-speed behavior, and GUI integration.

This fixes the regression caused when camera navigation state moved from `ViewerGL` to `ViewerGui`, leaving the example's private `_cam_speed` access ineffective.

Closes #3439.

## Checklist

## Test plan

```text
uvx pre-commit run -a
uv run --extra dev -m newton.tests -k test_viewer_controls
uv run --extra dev -m newton.tests -k test_viewer_camera
```

## Bug fix

**Steps to reproduce:**

1. Run `python -m newton.examples contacts_rj45_plug` from current main.
2. Move the camera with WASD/QE.
3. Observe that the example's intended `0.2` m/s camera speed is not applied because `_cam_speed` is no longer an attribute of `ViewerGL`.

**Minimal reproduction:**

```python

viewer = ViewerGL()
assert not hasattr(viewer, "_cam_speed")
```

## New feature / API change

```python

viewer = ViewerGL()
viewer.camera_speed = 0.2
```

## Summary by CodeRabbit

- **New Features**
  - Added a configurable viewer camera speed setting for keyboard navigation.
  - Camera speed defaults to 4.0 meters per second and supports custom non-negative, finite values.
  - Keyboard camera movement now uses the configured speed.

- **Documentation**
  - Updated the visualization guide with camera speed details and an example configuration.
  - Added the feature to the unreleased changelog.

- **Bug Fixes**
  - Invalid camera speed values are rejected with a clear validation error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Add three VBD contact examples — `vbd_rigid_rigid_contact`, `vbd_soft_rigid_contact`, and `vbd_soft_rigid_mix_contact` — demonstrating rigid-rigid, soft (particle-rigid), and mixed cloth-bag contacts
 - Add masked rigid-body reset support to `SolverVBD`; particle resets are not yet supported. (#3256)
 - Add viewer layer system to overlay multiple solvers/models in supported rendering viewers; call `ViewerBase.activate(layer_id)` to route subsequent `set_model` / `log_state` / `log_*` calls into a named layer, `ViewerBase.set_layer_visible()` to toggle layers independently, and `ViewerBase.set_layer_transform()` to position layers side-by-side. See `example_basic_multi_solver_overlay.py`
+- Add `ViewerBase.camera_speed` to configure keyboard translation speed for interactive viewers. (#3439)
 - Add SDF contact support for convex-hull shapes with mesh-attached SDFs and opt-in SDF contact generation for box shapes.
 - Add opt-in filtering of static-static, static-kinematic, and kinematic-kinematic contacts during broad-phase collision detection. Set `CollisionPipeline(include_static_kinematic_pairs=False)` to enable filtering; the default preserves existing contact generation. `Model.shape_contact_pairs` remains an unfiltered superset for direct consumers such as `SolverKamino` and hydroelastic SDF setup.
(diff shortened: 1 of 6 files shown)


adenzler-nvidia (approved):
Thanks for the fix! Promoting this to a public `ViewerBase.camera_speed` property is the right call — much nicer than continuing to reach into the private `ViewerGui` state that silently broke in the first place. Nice regression coverage too: the GUI keyboard test pins exactly the seam that regressed, and the validation tests are thorough.

PR #3480, opened 2026-07-10
Add additional scenes to tiled camera sensor benchmarks

StafaH:
Extend tiled camera rendering benchmarks beyond the single Franka scene with a set of scenes of varying visual complexity, to support hill-climbing renderer performance:

- franka: fixed-base Franka FR3 arm
- quadruped: ANYmal D in its nominal standing pose
- franka_cabinet: Isaac Lab's Franka cabinet (open-drawer) scene
- shapes_256: grid of 256 primitive shapes

Scenes are defined once as presets shared between the ASV benchmark classes and a preview image generator (--preview), which writes a single-world view and a 4x4 grid of 16 worlds per scene so the benchmarked content can be inspected visually.

New scenes look like:

image

## Summary by CodeRabbit

* **New Features**
  * Added a scene-driven tiled camera benchmark suite (quadruped, cabinet, and 256-shape scenes).
  * Added timing for color-only, depth-only, and color+depth rendering.
  * Added preview image generation and CLI support for `--preview`, `--preview-dir`, and `--preview-size` (preview skips benchmarking).
* **Changes**
  * Simplified benchmark selection to run only explicitly requested entries.
  * Removed the prior tiled vs pixel-priority benchmark modes and related naming/env gating.
* **Developer Improvements**
  * Extended `download_external_git_folder` to accept a `ref` parameter when fetching external content.


asv/benchmarks/simulation/bench_sensor_tiled_camera.py
```diff
@@ -1,225 +1,367 @@
...
+def _build_franka_cabinet() -> newton.ModelBuilder:
+    """Isaac Lab's Franka cabinet (open-drawer) scene.
 
-    return decorator
+    Layout follows FrankaCabinetDirectEnvCfg: the Sektion cabinet sits at the
+    origin raised 0.4 m, with the arm 1 m away rotated to face the drawers.
+    The cabinet URDF comes from IsaacGymEnvs (Isaac Lab itself references a
+    Nucleus-hosted USD of the same asset).
+    """
+    cabinet_folder = newton.examples.download_external_git_folder(
```
daniela-hase, line 148: Is it possible to pin this to a specific commit? Otherwise we are always pulling from `main`, right? So the assets may change unexpectedly.

asv/benchmarks/simulation/bench_sensor_tiled_camera.py
```diff
@@ -1,225 +1,367 @@
...
-                        wp.vec3f(2.4, 0.0, 0.8),
-                        wp.quatf(0.4187639653682709, 0.4224344491958618, 0.5708873867988586, 0.5659270882606506),
-                    )
-                ]
-                * world_count
-            ],
-            dtype=wp.transformf,
+        light_direction = wp.vec3f(*preset.light_direction) if preset.light_direction is not None else None
+        self.sensor = SensorTiledCamera(model=self.model)
+        self.sensor.default_render_config.render_order = RENDER_ORDER
```
daniela-hase: While `TILED` rendering makes more sense to benchmark the actual raytracing, I think keeping one of the scenes around with `PIXEL` priority can make sense too.

The `PIXEL` order greatly benefits from the environments being very similar, which of course is not very realistic in a real use-case. But, it gives us a good measurement of all the rest of the code of the renderer, not related to bvh traversal.

Like, if someone makes a small change somewhere, it may only affect the `TILED` benchmark by 1% (due to most time being spent in bvh traversal there), but may show up as a bigger performance difference in the `PIXEL` benchmark where most of the time is spent outside of bvh traversal.

Does that make sense? :)
Maybe we use the Franka Cabinet scene, make all envs/worlds identical and render it with PIXEL mode as an extra benchmark?

daniela-hase (approved):

eric-heiden (approved):
Thanks, LGTM!

PR #3484, opened 2026-07-11
Keep ViewerGL event loop alive

### LINKED ISSUE #3225: [BUG] ViewerGL once closed reopens to black screen only
### Bug Description

When the ViewerGL viewer is opened, it runs fine. But once I close it, and then re instantiate the viewer it opens a black screen. Any idea?

### Reproduction Script

```python
import newton
import warp as wp
def main():
    wp.init()
    sim_device = wp.get_device("cpu")
    model_builder = newton.ModelBuilder()
    model_builder.add_shape_plane()
    model = model_builder.finalize(sim_device)
    state_0 = model.state()
    state_1 = model.state()
    control = model.control()
    contacts = model.contacts()
    solver = newton.solvers.SolverXPBD(model, iterations=10)
    viewer.set_model(model)
    viewer.close()
    # reopen viewer
    viewer.set_model(model)
    sim_timestep = 1.0 / 240.0
    sim_time = 0.0
    while True:
        print("Sim time: ", sim_time)
        state_0.clear_forces()
        solver.step(
            state_in=state_0,
            state_out=state_1,
            control=control,
            contacts=contacts,
            dt=sim_timestep,
        )
        sim_time += sim_timestep
        viewer.begin_frame(sim_time)
        viewer.log_state(state_0)
        viewer.end_frame()
if __name__ == "__main__":
    main()
```

### System Information

- python: 3.11
- newton: 1.3.0
- warp-lang: 1.14.0
- OS: 22.04.1-Ubuntu
- GPU: NVIDIA RTX 4090

Benchan691:
## Description

Keep Pyglet’s process-wide event loop alive when an individual `ViewerGL` instance closes. This allows a viewer to be closed and recreated in the same process.

Closes #3225.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_viewer_renderer_lifecycle
uvx pre-commit run --files newton/_src/viewer/gl/opengl.py newton/tests/test_viewer_renderer_lifecycle.py CHANGELOG.md
```

Also created, rendered with, closed, and recreated `ViewerGL` twice in one process on macOS.

## Bug fix

**Steps to reproduce:**

1. Create and render a `ViewerGL` instance.
2. Call `close()`.
3. Create a second `ViewerGL` instance in the same process.

**Root cause:**

`RendererGL.close()` dispatched Pyglet’s application-wide exit event and stopped the shared platform event loop even though only one viewer window was closing. The second viewer reused that global lifecycle state and could render as a blank window.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed OpenGL viewers failing to close and be recreated in the same process.
  * Viewer shutdown now reliably keeps Pyglet’s shared event loop available for subsequent viewers.
  * Improved OpenGL renderer cleanup to reset cached textures and close the window consistently.
* **Tests**
  * Added a regression test covering the OpenGL renderer close lifecycle behavior and ensuring the shared event loop remains available.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,7 @@
 
 ### Fixed
 
+- Fix `ViewerGL` so it can be closed and recreated in the same process without shutting down Pyglet's shared event loop. (#3225)
 - Fix USD import ignoring ancestor material bindings with `strongerThanDescendants` strength when a mesh authors `material:binding` without applying `MaterialBindingAPI`: material resolution now uses UsdShade's canonical `ComputeBoundMaterial` unconditionally, which also adds collection-based binding support. Prims authoring bindings without the applied schema are invalid USD and now surface USD's own warning (once per prim per import) — fix such assets with `usdchecker` or `usd-validation-nvidia`.
 - Fix `ModelBuilder.add_usd()` to honor `ignore_paths` in the custom-frequency traversal, so prims under ignored subtrees no longer register spurious custom-frequency rows in two-pass import workflows. (#3406)
 - Fix USD joint `physics:collisionEnabled` import so joints with two explicit bodies honor authored collision behavior; joints to world continue to allow body/world collisions, and articulation-wide self-collision filtering remains additive.
--- a/newton/_src/viewer/gl/opengl.py
+++ b/newton/_src/viewer/gl/opengl.py
@@ -1369,10 +1369,6 @@ class RendererGL:
     def close(self):
         self._make_current()
 
-        if not self.headless:
-            self.app.event_loop.dispatch_event("on_exit")
-            self.app.platform_event_loop.stop()
-
         RendererGL._fallback_texture = None
         self.window.close()
 
--- a/newton/tests/test_viewer_renderer_lifecycle.py
+++ b/newton/tests/test_viewer_renderer_lifecycle.py
@@ -1,0 +1,33 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
+
+"""Regression tests for the OpenGL renderer lifecycle."""
+
+import unittest
+from types import SimpleNamespace
+from unittest import mock
+
+from newton._src.viewer.gl.opengl import RendererGL
+
+
+class TestRendererGLClose(unittest.TestCase):
+    def test_close_keeps_pyglet_event_loop_available_for_next_viewer(self):
+        """Closing one viewer must not stop Pyglet's shared event loop."""
+        renderer = RendererGL.__new__(RendererGL)
+        renderer._make_current = mock.Mock()
+        renderer.window = mock.Mock()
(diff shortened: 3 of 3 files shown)


adenzler-nvidia (changes requested):
Unfortunately I don't think this fixes #3225 — the two removed lines aren't what set the global exit state, so the black screen on recreation is still there with this change.

In the pyglet versions Newton pins (verified on 2.1.6 and 2.1.14), `EventLoop` has no runtime `on_exit` handler — it's only defined under `if _is_pyglet_doc_run:` — so `dispatch_event("on_exit")` never touched `has_exit`. And `PlatformEventLoop.stop()` is a no-op (`pass`) on Linux and macOS. The state that actually breaks recreation is set by `self.window.close()`, which this PR keeps: `BaseWindow.close()` dispatches `on_window_close` on the global event loop, whose default handler calls `exit()` when the last window closes (pyglet's hidden shadow window is excluded from `app.windows`), setting `event_loop.has_exit = True` process-wide. Nothing ever resets it, so the next viewer's `RendererGL.has_exit()` returns `True` and `ViewerGL` skips rendering — the blank window from the issue.

Since both removed calls are no-ops on macOS, this change is behavior-neutral on the platform in the test plan — the manual close/recreate cycle behaves identically with and without this patch. The test-plan note also doesn't say what the recreated viewer displayed; could you confirm whether the second instance actually renders the scene? Based on the above, I'd expect it to still come up blank.

I ran the recreate cycle with real GL contexts (newton 1.3.0, pyglet 2.1.14, EGL headless windows) against both `close()` variants:

| `close()` variant | viewer 2 `renderer.has_exit()` |
|---|---|
| pre-PR (main) | `True` |
| post-PR (this branch) | `True` |
| post-PR + `has_exit` reset in `__init__` | `False` |

Minimal repro on a desktop:

```python
import pyglet
from newton._src.viewer.gl.opengl import RendererGL

r1 = RendererGL(screen_width=320, screen_height=240, headless=False)
r1.close()
print(pyglet.app.event_loop.has_exit)  # True, with or without this PR

r2 = RendererGL(screen_width=320, screen_height=240, headless=False)
print(r2.has_exit())  # True -> ViewerGL never renders
```

Suggested fix: reset the global flag when a new viewer starts, in the non-headless branch of `RendererGL.__init__` next to `platform_event_loop.start()`:

```python
self.app.event_loop.has_exit = False
```

Two more things:

- Please keep `self.app.platform_event_loop.stop()` in `close()`. On Windows it's not a no-op: `Win32EventLoop.start()` calls `winmm.timeBeginPeriod()` and `stop()` is the matching `timeEndPeriod`. `__init__` calls `start()` per viewer, so removing `stop()` leaves the raised OS timer resolution pinned until process exit. The begin/end pairs stay balanced even with overlapping viewers. Dropping the `dispatch_event("on_exit")` line is fine — it's dead code either way.
- The new test mocks `self.window` and `self.app`, so pyglet's real close chain never runs — it asserts the two calls are absent rather than that recreation works. With the fix above it could instead assert `pyglet.app.event_loop.has_exit` is `False` after constructing a renderer.

PR #3184, opened 2026-06-15
Contact reduction improvements

nvtw:
## Description
Contact reduction improvements for mesh-SDF contacts with positive contact gaps. The reducer now prefers margin-depth contacts over gap-only directional fallbacks, keeps fresh separated speculative rows during sticky replay, and tightens SDF edge search/culling to the inner contact envelope.
## Checklist
- [x] New or existing tests cover these changes
- [x] The documentation is up to date with these changes
- [x] `CHANGELOG.md` has been updated (if user-facing change)
## Test plan
Added/updated contact reduction and SDF contact tests:
```bash
uv run --extra dev -m newton.tests -k "test_centered_two_spatial_depths_prefers_inner_then_outer or test_mesh_sdf_contact_search_precision_uses_inner_envelope"
```
## Bug fix
**Steps to reproduce:**
Run a mesh-SDF contact scenario with a positive contact gap where gap-only directional fallback contacts compete with true margin-depth contacts.
**Minimal reproduction:**
See the new `test_centered_two_spatial_depths_prefers_inner_then_outer` coverage in `newton/tests/test_contact_reduction_global.py`.

## Summary

- **Bug Fixes**
  - Mesh–SDF contacts with positive gaps now prefer margin-depth contacts over directional gap-only fallbacks.
  - Sticky contact replay now recomputes per-contact world-space geometry and uses writer-aligned shape-body inputs, avoiding overwrites when gaps are positive.

- **New Features**
  - Spatial contact reduction now supports two-tier depth prioritization (inner vs outer) with updated fast/deterministic selection behavior.
  - Mesh–SDF edge contact search accepts a caller-provided precision target and adds inner-cull consistency gating.

- **Tests**
  - Added/extended coverage for two-tier depth prioritization and the precision-target calculation.

PR #3490, opened 2026-07-13
Fix texture SDF inner contact reduction

### LINKED ISSUE #3495: [BUG] Newton-nut_bolt_sdf example QA failure
### Bug Description

Newton-nut_bolt_sdf — FAIL
    Variant A: `--world-count 4 --solver xpbd` — FAIL
      Nut motion was very jittery rather than slow and controlled. Captured
      stderr was empty and shutdown was clean.
    Variant B: `--world-count 4 --solver mujoco` — FAIL
      Nuts were jittery while only partially engaged with the bolts; motion
      became acceptable only after each nut moved completely over its bolt.
      Captured stderr was empty and shutdown was clean.
    The two solvers were materially similar, satisfying the comparison aspect,
    but both violated the expected slow, controlled screw behavior; therefore
    the overall item is FAIL.

### Reproduction Script

```python

```

### System Information

_No response_

mzamoramora-nvidia:
## Description

Closes #3495 

Account for texture-SDF voxel uncertainty when selecting the preferred inner contact-reduction tier.

The mesh-SDF reduction introduced in #3184 used the exact contact margin as its inner depth. Texture-SDF sampling uncertainty could therefore move valid near-margin contacts into the outer tier, producing an incomplete and unstable contact manifold.

For texture SDFs, this change expands the inner depth by one voxel radius, clamped to the configured contact gap. Non-texture SDF paths remain unchanged.

### Nut/bolt behavior

Mean absolute nut angular velocity was measured while the nut descended from `z = 0.060 m` to `z = 0.050 m`, using four worlds over 1,000 frames at 120 Hz.

| Version | XPBD | MuJoCo |
|---|---:|---:|
| Newton 1.3.0 (`ce11136b`) | 6.745 rad/s | 5.767 rad/s |
| Main before this fix (`8863422e`) | 3.229 rad/s | 3.454 rad/s |
| Main with this fix | 6.801 rad/s | 5.668 rad/s |

The MuJoCo 1.3.0 value uses canonical state-buffer handling to avoid including an unrelated odd-substep bug in the baseline.

## Checklist

- [x] No changelog entry is required because the affected behavior has not been released

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_mesh_sdf_voxel_tolerance_preserves_inner_contact_coverage
uv run --extra dev -m newton.tests -k test_contacts.example_nut_bolt_sdf
uv run --extra examples -m newton.examples nut_bolt_sdf --solver xpbd
uv run --extra examples -m newton.examples nut_bolt_sdf --solver mujoco
```

The regression test fails before the fix because one expected inner-tier contact is discarded. It passes with the fix.

Visual testing confirms that the nuts rotate and descend smoothly while engaged with the bolts using both XPBD and MuJoCo.

## Bug fix

**Steps to reproduce:**

1. Run `nut_bolt_sdf` with either XPBD or MuJoCo.
2. Observe slow, jittery motion while the nuts are engaged with the bolts.
3. Compare the angular speed with Newton 1.3.0.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SDF collision contact handling near margins by refining how the inner contact region depth/tolerance is computed, helping preserve valid inner contact coverage.
  * Enhanced reliability of near-margin contact placement on voxelized SDF surfaces, including positive-side (positive-X) coverage.

* **Tests**
  * Added a new CUDA test to verify that near-margin contacts on a voxelized box surface consistently produce the expected surviving contact coverage in the collision pipeline.


mzamoramora-nvidia:
Thanks for flagging this. I verified the behavior against both #3184 and current upstream: the established non-texture inner depth is `margin_sum`, not `margin_sum + gap_sum`. Using the full gap for non-texture contacts would change BVH/heightfield reduction behavior and collapse the two tiers.

Commit `892eb763` makes that distinction explicit:

```python
inner_spatial_depth = margin_sum
if use_texture_sdf_for_search:
    inner_spatial_depth += wp.min(texture_voxel_radius * min_sdf_scale, gap_sum)
```

This preserves the exact-margin non-texture path and performs the voxel calculation only for texture-SDF searches.

Validation:

- Focused texture-SDF coverage regression: passed
- All 17 `test_mesh_mesh_` tests, including SDF/BVH orderings and broad phases: passed
- SDF contact unit test: passed
- 1,000-frame `nut_bolt_sdf` final-state checks with XPBD and MuJoCo: passed

@coderabbitai review

adenzler-nvidia (approved):
Verified the fix independently on an L40S against the merge base: the nut-bolt SDF benchmark returns from 7.3 to 5.2 ms/frame (the v1.3.0 level), with worst-world solver iterations dropping from median 6 / max 13 to median 4 / max 6 — the solver stops churning against the incomplete manifold.

The approach is sound: `voxel_radius` is the texture sampling error bound (half voxel diagonal), so widening only the inner-tier classification by that uncertainty — without touching the exported contact distance — is exactly the right scope, and the `gap_sum` clamp preserves the inner ≤ outer invariant while degrading to a no-op at `gap = 0`, where tiering is vacuous anyway. This also brings tier selection in line with `mesh_sdf_contact_search_precision` and the inner-cull consistency check, which already account for texture-voxel uncertainty.

PR #3498, opened 2026-07-13
Deprecate contact match constants

jcarius-nv:
## Description

Deprecate `newton.geometry.MATCH_BROKEN` and `newton.geometry.MATCH_NOT_FOUND` because these contact-matching sentinels are implementation details. Keep both names import-compatible during the deprecation window through warning-emitting module attributes, while keeping the underlying constants local to `contact_match.py`.

Update API generation to present deprecated symbols without resolving or publishing their values. Remove the sentinel encoding from public contact documentation and retain the existing position-threshold regression coverage with explicit deprecation assertions.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_contact_matching
```

Direct and wildcard imports of both deprecated names were also verified locally.

## New feature / API change

Existing imports continue to work during the deprecation window and emit `DeprecationWarning`:

```python
from newton.geometry import MATCH_BROKEN, MATCH_NOT_FOUND
```

## Summary by CodeRabbit

* **Deprecation**
  * Deprecated `newton.geometry.MATCH_BROKEN` and `newton.geometry.MATCH_NOT_FOUND`. They remain accessible but now emit `DeprecationWarning`, are omitted from wildcard (`*`) imports, and are hidden from public API listings.

* **Documentation**
  * Updated API and concepts docs to clearly label these symbols as deprecated and to avoid relying on their numeric values.
  * Refined rigid contact matching and sticky-mode documentation to better explain index semantics and allocation/row behavior.

* **Tests**
  * Added coverage for deprecation warnings and generation of deprecated-symbol docs without exposing values, plus updated contact-matching expectations.


newton/geometry.py
```diff
@@ -24,12 +25,17 @@
...
 from ._src.geometry.kernels import sdf_box, sdf_capsule, sdf_cone, sdf_cylinder, sdf_mesh, sdf_plane, sdf_sphere
 from ._src.geometry.narrow_phase import NarrowPhase
 from ._src.geometry.sdf_hydroelastic import HydroelasticSDF
 from ._src.geometry.sdf_utils import compute_offset_mesh, create_empty_sdf_data
 
+if TYPE_CHECKING:
+    from ._src.geometry.contact_match import MATCH_BROKEN, MATCH_NOT_FOUND
+
 __all__ = [
     "MATCH_BROKEN",
```
adenzler-nvidia: Keeping `MATCH_BROKEN`/`MATCH_NOT_FOUND` in `__all__` means `from newton.geometry import *` now emits both DeprecationWarnings even for callers that never use these names: star-import getattrs every name in `__all__`, and these two now route through the module `__getattr__`. Under `-W error::DeprecationWarning` a bare star-import of this module hard-fails, and the warning is unactionable for star-importers who don't touch the constants (their only remedies are suppressing it or dropping the star-import).

There is no third option — `import *` is a plain getattr loop, so it's not possible to keep star-import silently working while warning on direct use. The alternatives are: keep them in `__all__` (star-import compat preserved at the cost of unavoidable warnings), or drop them from `__all__` and rely on `__getattr__` alone for the compat window, which is the numpy/scipy/CPython convention for PEP 562 deprecations and only breaks star-import consumers, loudly. If you drop them, `write_module_page` could union `__deprecated_symbols__` into the symbol list so the generated Deprecated table stays identical.

Fine either way — flagging so the star-import behavior is a deliberate choice.

newton/tests/test_contact_matching.py
```diff
@@ -58,6 +58,9 @@ def _collide_once(pipeline, state, contacts):
 def test_first_frame_all_not_found(test, device):
     """First frame: prev_count is 0, so every contact must get MATCH_NOT_FOUND."""
     with wp.ScopedDevice(device):
+        with test.assertWarnsRegex(DeprecationWarning, "MATCH_NOT_FOUND.*deprecated"):
+            match_not_found = newton.geometry.MATCH_NOT_FOUND
```
adenzler-nvidia: Suggest having the physics tests use the internal constants directly (`from newton._src.geometry.contact_match import MATCH_NOT_FOUND`) or the literals, like the sibling tests in this file already do (`== -1` at line 160, `== -2` at line 305), rather than fetching the values through the deprecated public alias with warning ceremony (also at line 180 below). As written, these two physics tests break with an AttributeError when the deprecation window ends and the names are removed, and rewording the warning message breaks them via the regex — failures unrelated to the contact-matching behavior they test.

The deprecated public path is better exercised once in a dedicated test (round-trip the values through `newton.geometry`), following the `EqType` precedent in `test_equality_constraints.py`. That test could also cover the shim's edge cases cheaply: unknown attribute raises `AttributeError`, and `from newton.geometry import *` still provides the names.

newton/geometry.py
jcarius-nv: Addressed in 58eb7c2a. I removed MATCH_BROKEN and MATCH_NOT_FOUND from newton.geometry.__all__, so wildcard imports no longer resolve the deprecated aliases or emit warnings. Direct attribute and named-import access still goes through __getattr__ and warns during the compatibility window. The API generator now unions __deprecated_symbols__ into its symbol list, preserving the Deprecated table without exposing the values. The dedicated API test also verifies that wildcard import remains clean when DeprecationWarning is treated as an error and omits both names.

newton/tests/test_contact_matching.py
jcarius-nv: Addressed in 58eb7c2a. The contact-matching physics tests now import the internal MATCH_BROKEN and MATCH_NOT_FOUND definitions directly, so they no longer depend on the public deprecation shim or its warning text. A dedicated API test now checks that both public aliases emit DeprecationWarning and match the internal values, that an unknown attribute raises AttributeError, and that wildcard imports stay warning-free and omit these names because they were deliberately removed from __all__.

jcarius-nv:
@coderabbitai review

camevor (commented):
Looks good, thanks!

One ask: `rigid_contact_match_index` remains public and users will observe the negative elements, but after this change nothing documents what any of the values mean. `rigid_contact_match_index` should note that non-negative elements are the matched old contact index and negative values refer to new or broken contacts.

CHANGELOG.md
```diff
@@ -71,6 +71,7 @@
 - Deprecate passing solver constructor options positionally after stable positional inputs such as `model` and explicit solver configs; migrate calls such as `SolverVBD(model, 10)` to `SolverVBD(model, iterations=10)`.
 - Deprecate `ModelBuilder.find_shape_contact_pairs()`; shape contact pairs are generated automatically by `ModelBuilder.finalize()`, so configure collision filters before finalization instead of rebuilding contact pairs manually.
 - Deprecate `newton.EqType` in favor of `newton.solvers.SolverMuJoCo.EqType`; migrate equality-constraint type references to the MuJoCo-scoped enum.
+ Deprecate `newton.geometry.MATCH_BROKEN` and `newton.geometry.MATCH_NOT_FOUND`; these were implementation-detail sentinels with no public replacement. Do not compare `rigid_contact_match_index` against them directly, and avoid importing them going forward.
```
camevor: Missing list dash

jcarius-nv: Addressed in 5ef2ff02. I restored the missing list dash and kept the migration guidance concise: the constants are deprecated without replacement, and users should not rely on or import them.

jcarius-nv:
@camevor Addressed the latest review feedback in 5ef2ff02. Contacts.rigid_contact_match_index now documents that non-negative elements index matching contacts in the previous sorted buffer, while negative elements indicate new or broken contacts, without exposing the sentinel values. I also restored the missing changelog list dash and kept its migration guidance concise and user-facing.

camevor (approved):
lgtm!

eric-heiden (approved):

PR #3500, opened 2026-07-13
Fix recording example contact gap

### LINKED ISSUE #3493: [BUG] Newton example replay_viewer failed QA testing
### Bug Description

12. Newton-replay_viewer — FAIL
    Command:
      python -m newton.examples replay_viewer
    Recording loaded:
      C:\tmp\newton-manual-runs\recording-case\humanoid_recording.bin
    Result: FAIL (user-assisted visual verification).
    Observed discrepancy:
    - Some recorded humanoid bodies visibly penetrated/fell through the ground.
    - This makes the replayed simulation state physically invalid despite the
      viewer loading the 1,000-frame artifact.
    - The recording run emitted repeated broadphase overflow/contact-capacity
      warnings. Insufficient contact-buffer capacity is a plausible correlated
      cause, but root cause was not established by this execution pass.
    - Replay viewer itself emitted no stderr output and closed cleanly.

### Reproduction Script

```python

```

### System Information

_No response_

mzamoramora-nvidia:
## Description

Closes #3493.

Fix the `recording` example humanoids falling through the ground after Newton started forwarding `Model.shape_gap` to MuJoCo.

The humanoid MJCF does not author geometry gaps, so its shapes inherited Newton's `0.1 m` default. This generated enough inactive contacts to overflow MuJoCo Warp's contact and constraint buffers, producing invalid recorded states.

This PR sets the humanoid and ground gaps to zero, matching the asset's MuJoCo behavior without increasing `nconmax` or `njmax`.

It also registers the example with the standard test harness. The regression runs eight humanoids for 120 frames on CPU and CUDA and verifies that their bodies remain above the ground. Recording serialization remains covered by `test_recorder.py`.

## Checklist

- [x] No changelog entry is needed because this fixes an unreleased regression

## Test plan

```bash
uv run --extra dev -m newton.tests --strict-warnings -k test_basic.example_recording
```

```bash
uv run -m newton.examples recording
```

For visual verification:

```bash
uv run -m newton.examples replay_viewer
```

Load `humanoid_recording.bin` and verify that the humanoids remain on or above the ground throughout playback.

Results:

- The focused regression passes on CPU and CUDA.
- Removing the zero-gap assignments causes both variants to fail with contact-capacity warnings.
- The 1,000-frame manual recording completes without broadphase or `nefc` overflow warnings.

## Bug fix

**Steps to reproduce:**

1. Run `uv run -m newton.examples recording`.
2. Observe repeated broadphase and `nefc` overflow warnings.
3. Run `uv run -m newton.examples replay_viewer`.
4. Load `humanoid_recording.bin`.
5. Observe some humanoids penetrating or falling through the ground.

**Minimal reproduction:**

```bash
uv run -m newton.examples recording
```

### DIFF
--- a/newton/tests/test_examples.py
+++ b/newton/tests/test_examples.py
@@ -27,6 +27,7 @@ from typing import Any
 import warp as wp
 
 import newton.tests.unittest_utils
+from newton.examples.basic.example_recording import Example as RecordingExample
 from newton.tests.unittest_utils import (
     USD_AVAILABLE,
     NewtonTestCase,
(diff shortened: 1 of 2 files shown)


newton/tests/test_examples.py
```diff
@@ -291,12 +292,57 @@ class TestBasicExamples(NewtonTestCase):
...
     extra_allow_output_regexes = kwargs.pop("allow_output_regexes", None) or ()
     allow_output_regexes = [*_BASIC_EXAMPLE_ALLOW_OUTPUT_REGEXES, *extra_allow_output_regexes]
     add_example_test(TestBasicExamples, allow_output_regexes=allow_output_regexes, **kwargs)
 
 
+add_function_test(
+    TestRecordingExample,
+    "test_basic.example_recording_body_stability",
+    test_recording_body_stability,
+    devices=cuda_test_devices,
```
adenzler-nvidia: not sure if this test needs to run on both CPU and GPU, as it mainly tests that the recording can be loaded again? We didn't have a test for this before?

mzamoramora-nvidia: Good point.  Recording save/load is already covered by t`est_recorder.py`.  But this example was lacking a proper test, which I added in ffc9af2

jcarius-nv (approved):
Re-reviewed the current head after the test update. The zero-gap fix is correctly applied to both the imported humanoid geometry and the ground plane, and the standard example regression covers the reported failure on CPU and CUDA while existing recorder tests cover serialization. The focused regression passed locally on both devices, and the capacity warnings return without the fix. No blocking issues found.

Non-blocking housekeeping: the PR body still mentions the removed recording round-trip test and the old recording_body_stability selector, which now runs zero tests.

mzamoramora-nvidia:
Description has been updated.

PR #3501, opened 2026-07-13
Extend CPU graph capture coverage

c0d1f1ed:
## Description

This PR finishes several CPU graph-capture follow-ups:

- Fix Kamino CG/CR capture control flow so captured solves use the capture-safe loop path on any capturing device, not only CUDA. This avoids recording a host-readback convergence decision from stale CPU capture data.
- Enable graph capture on CPU in `proxy_joint_gripper` and the remaining cable examples that were still CUDA-gated.
- Update SolverVBD graph-capture guidance and errors to describe device-specific allocation behavior instead of CUDA-only behavior.
- Move the SolverVBD graph-capture allocation helper import to module scope and rename internal Kamino CG/CR graph flags from CUDA-specific wording to device-agnostic wording.

## Checklist

## Test plan

```
uv run --extra dev -m unittest newton._src.solvers.kamino.tests.test_linalg_solve_cg -v
uv run --extra dev -m newton.tests -k proxy_joint_gripper
uv run --extra dev -m newton.tests -k cable_y_junction_cpu
uv run --extra dev -m newton.tests -k cable_twist_cpu
uv run --extra dev -m newton.tests -k cable_cross_slide_table_cpu
git diff --check origin/main...HEAD
```

The Kamino CG/CR capture-replay tests currently skip on CPU when Warp raises its known non-contiguous CPU `wp.copy()` APIC limitation; any other capture failure is still treated as a test failure.

## Bug fix

**Steps to reproduce:**

1. Capture a Kamino iterative CG/CR solve on CPU.
2. Replay the graph for a solve that requires dynamic convergence.

Without this PR, the loop can record the eager host-readback branch and bake in a stale record-time convergence decision. With this PR, captured solves use the graph-capture loop path whenever the device is capturing.

## New feature / API change

Examples that were already graph-capturable now use the same device-agnostic capture idiom on CPU and CUDA:

```python
    self.simulate()
self.graph = capture.graph
```

## Summary by CodeRabbit

- **New Features**
  - Unified conjugate-solver graph-capture control via a `use_graph` option (replacing CUDA-specific naming).

- **Bug Fixes**
  - Clarified and fixed behavior differences between graph-captured and eager execution for CG/CR (including potential under-iteration on CPU capture-and-replay).
  - Improved VBD rigid contact-history graph-capture resizing/allocation handling and corresponding guidance/error text.

- **Documentation**
  - Updated capture-time requirements and buffer pre-sizing notes for rigid contact history.

- **Tests**
  - Added CPU capture-and-replay regression tests for CG/CR matching eager results; updated to `use_graph`.

- **Chores**
  - Simplified example graph-capture to run consistently across devices; adjusted related test device coverage and help text.

### DIFF
--- a/newton/_src/solvers/kamino/_src/linalg/conjugate.py
+++ b/newton/_src/solvers/kamino/_src/linalg/conjugate.py
@@ -338,7 +338,7 @@ def _run_capturable_loop(
     maxiter: wp.array[int],
     atol_sq: wp.array[Any],
     callback: Callable | None,
-    use_cuda_graph: bool,
+    use_graph: bool,
     use_graph_conditionals: bool = True,
     maxiter_host: int | None = None,
     loop_granularity: int = 1,
@@ -384,7 +384,7 @@ def _run_capturable_loop(
         if callback_launch is not None:
             callback_launch.launch()
 
-    if use_cuda_graph and device.is_cuda and device.is_capturing:
+    if use_graph and device.is_capturing:
         if use_graph_conditionals:
             wp.capture_while(global_condition, do_cycle_with_condition)
         else:
@@ -541,7 +541,7 @@ class ConjugateSolver(Generic[ScalarType, IndexType]):
         maxiter: Maximum iterations per world. If None, defaults to 1.5 * maxdims.
         Mi: Operator applying the inverse preconditioner M^-1, such that Mi @ A has a smaller condition number than A.
         callback: Optional callback kernel invoked each iteration.
-        use_cuda_graph: Whether to use CUDA graph capture for the solve loop.
+        use_graph: Whether to use graph capture for the solve loop.
         loop_granularity: Number of iterations before termination criteria are checked.
     """
 
@@ -555,7 +555,7 @@ class ConjugateSolver(Generic[ScalarType, IndexType]):
         maxiter: wp.array[wp.int32] | None = None,
         Mi: BatchedLinearOperator[ScalarType, IndexType] | None = None,
         callback: Callable | None = None,
-        use_cuda_graph: bool = True,
+        use_graph: bool = True,
         use_graph_conditionals: bool = True,
         loop_granularity: int = 1,
     ):
@@ -593,7 +593,7 @@ class ConjugateSolver(Generic[ScalarType, IndexType]):
         self.loop_granularity = loop_granularity
 
         self.callback = callback
-        self.use_cuda_graph = use_cuda_graph
+        self.use_graph = use_graph
 
         self.dot_tile_size = min(2048, 2 ** math.ceil(math.log(self.maxdims, 2)))
         self.tiled_dot_kernel = make_dot_kernel(self.dot_tile_size, self.maxdims)
@@ -756,7 +756,7 @@ class CGSolver(ConjugateSolver[ScalarType, IndexType]):
             self.maxiter,
             self.atol_sq,
             self.callback,
-            self.use_cuda_graph,
+            self.use_graph,
             use_graph_conditionals=self.use_graph_conditionals,
             maxiter_host=self.maxiter_host,
             loop_granularity=min(self.loop_granularity, self.maxiter_host),
@@ -890,7 +890,7 @@ class CRSolver(ConjugateSolver[ScalarType, IndexType]):
             self.maxiter,
             self.atol_sq,
             self.callback,
-            self.use_cuda_graph,
+            self.use_graph,
             use_graph_conditionals=self.use_graph_conditionals,
             maxiter_host=self.maxiter_host,
             loop_granularity=min(self.loop_granularity, self.maxiter_host),
--- a/newton/_src/solvers/kamino/tests/test_linalg_solve_cg.py
+++ b/newton/_src/solvers/kamino/tests/test_linalg_solve_cg.py
@@ -75,7 +75,7 @@ class TestLinalgConjugate(unittest.TestCase):
             maxiter=maxiter,
             Mi=None,
             callback=None,
-            use_cuda_graph=False,
+            use_graph=False,
         )
         cur_iter, r_norm_sq, atol_sq = solver.solve(b_wp, x_wp)
 
@@ -223,7 +223,7 @@ class TestLinalgConjugate(unittest.TestCase):
             maxiter=None,
             Mi=None,
             callback=None,
-            use_cuda_graph=False,
+            use_graph=False,
         )
         solver_dense.solve(b_wp, x_dense)
 
@@ -237,7 +237,7 @@ class TestLinalgConjugate(unittest.TestCase):
             maxiter=None,
             Mi=None,
             callback=None,
-            use_cuda_graph=False,
+            use_graph=False,
         )
         solver_sparse.solve(b_wp, x_sparse)
 
@@ -335,7 +335,7 @@ class TestLinalgConjugate(unittest.TestCase):
             rtol=rtol,
             maxiter=None,
             Mi=None,
-            use_cuda_graph=False,
+            use_graph=False,
         )
         solver.solve(b_wp, x_wp)
 
@@ -477,7 +477,7 @@ class TestLinalgConjugate(unittest.TestCase):
             maxiter=maxiter,
             Mi=None,
             callback=None,
-            use_cuda_graph=False,
+            use_graph=False,
         )
         solver.solve(b, x_wp)
 
@@ -570,7 +570,7 @@ class TestLinalgConjugate(unittest.TestCase):
             maxiter=maxiter,
             Mi=Mi,
             callback=None,
-            use_cuda_graph=False,
+            use_graph=False,
         )
         solver.solve(b, x_wp)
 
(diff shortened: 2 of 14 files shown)


chschuma-disney (dismissed):
Thanks for looking into this! Looks good from Kamino's side.

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -2053,15 +2049,6 @@ def _initialize_rigid_bodies(
         """
         model = self.model
         internal_rigid = model.body_count > 0 and not self.integrate_with_external_rigid_solver
```
adenzler-nvidia, line 2059: Allocating the history buffers during capture doesn't behave like eager allocation: the `wp.zeros` fill is recorded into the graph and re-executes on every replay (verified empirically on both CPU APIC capture and CUDA with mempool). So when history is first allocated during capture, every replay re-zeros it before the restore kernel reads it, and warm-starting silently no-ops for the lifetime of the graph — equivalent to `rigid_contact_history=False`, with no error or warning. The check removed here (added in #3392 this cycle) raised for exactly this case regardless of mempool support, because unlike the scratch contact buffers (fully rewritten each refresh, safe to grow in-graph), the history buffers are cross-replay-persistent state.

The `rigid_contact_history` param docstring still gives the unconditional pre-allocate guidance while the class docstring, CHANGELOG, and RST now advertise growth-on-demand, so the two contradict. I'd keep the unconditional raise for the history buffers specifically and let only the ordinary buffers grow during capture. If the cold-start behavior is intentionally accepted instead, the relaxed docs need that caveat and the two docstrings need reconciling.

newton/_src/solvers/kamino/tests/test_linalg_solve_cg.py
```diff
@@ -139,6 +139,86 @@ def test_solve_cr_cuda(self):
...
+        # Warp CPU graph capture currently rejects wp.copy() on non-contiguous
+        # arrays, which the CG/CR inner loop reaches via
+        # ``rz_old.assign(rz_new)`` in ``do_iteration``. Skip only that known
+        # limitation so any other failure -- including the silent
+        # under-iteration bug the capture-safe gate protects against --
+        # surfaces as a real test failure.
+        try:
+            with wp.ScopedCapture(device) as cap:
+                cap_cur, _, _ = solver.solve(problem.b_wp, x_capture)
+        except NotImplementedError as e:
```
adenzler-nvidia, line 208: These tests currently skip both with and without the fix: the pre-fix eager path also reaches `rz_old.assign(rz_new)` inside the capture (on CPU `dot_product` is the strided view `dot_partial_sums[:, :, 0]`), so the buggy path raises the same non-contiguous `NotImplementedError` this guard swallows — reverting the gate to `use_graph and device.is_cuda and device.is_capturing` still gives `OK (skipped=2)`. So the comment's claim that the silent under-iteration bug "surfaces as a real test failure" doesn't hold, and there's no regression protection until Warp lifts the copy limitation.

Two ways to get real coverage without touching solver code: (a) a direct capture-replay test of `_run_capturable_loop` with a synthetic contiguous `do_iteration` (a trivial kernel decaying `r_norm_sq`) — that captures fine on CPU today, demonstrably fails against the old gate, and can parametrize over `use_graph_conditionals`, which currently has no capture coverage on any device; (b) also running this capture-replay pair on CUDA, where the dot buffer is contiguous and the tests exercise the real CG/CR loop without skipping. Either way, this comment should say the parity assertions only activate once Warp supports this copy.

adenzler-nvidia (dismissed):
I verified this locally on both devices. All four example tests (the three cable examples and proxy_joint_gripper) pass with the unconditional capture path on CPU and CUDA, including their `test_final`/`test_post_step` physics checks, so the example changes are sound. The CG/CR gate change is also correct: with the loop's copy limitation worked around, CPU capture-replay exactly matches the eager solve, while the old gate bakes an under-iterated result into the graph.

Three blocking items — one here because it concerns a file outside this diff, two inline.

**An existing GPU test fails at this PR's head.** `test_rigid_contact_history_capture_requires_preallocation_cuda_0` (`newton/tests/test_solver_vbd.py:1050`) asserts the exact RuntimeError message removed here, and with the default mempool enabled the growth now silently succeeds, so it fails with `AssertionError: RuntimeError not raised`. This isn't visible in the PR checks because GPU CI hasn't run yet. It needs rewriting to the new contract (mempool on: capture with lazy growth succeeds; mempool off: the `_raise_if_capturing_resize` message fires) — or it stays valid as-is if the history check is kept per the inline comment on solver_vbd.py.

newton/_src/solvers/vbd/solver_vbd.py
c0d1f1ed: Thanks! I've restore the check and updated the docstring.

newton/_src/solvers/kamino/tests/test_linalg_solve_cg.py
c0d1f1ed: Good catch! I updated the test comment and added coverage for CG/CR on CUDA.

newton/_src/solvers/kamino/_src/linalg/conjugate.py
```diff
@@ -384,7 +384,7 @@ def do_cycle_with_condition():
         if callback_launch is not None:
             callback_launch.launch()
 
-    if use_cuda_graph and device.is_cuda and device.is_capturing:
+    if use_graph and device.is_capturing:
```
eric-heiden, line 387: 🟠 This now sends CPU capture through the graph-safe loop, but the CPU dot result is still the strided view `self.dot_partial_sums[:, :, 0]`. CG and CR copy that view during captured iterations through `rz_old.assign(rz_new)` and `zAz_old.assign(zAz_new)`, which Warp CPU graph capture rejects. The new CPU test skips this failure, so the advertised CPU replay path is not actually validated. Please use a contiguous `(2, n_worlds)` result or contiguous scratch buffer and make the CPU capture test assert replay parity without skipping.

newton/examples/cable/example_cable_cross_slide_table.py
```diff
@@ -827,13 +827,10 @@ def __init__(self, viewer, args):
...
-        if self.solver.device.is_cuda:
-            with wp.ScopedCapture() as capture:
-                self.simulate()
-            self.graph = capture.graph
-        else:
-            self.graph = None
+        """Capture the simulation update into a graph for replay."""
+        with wp.ScopedCapture() as capture:
+            self.simulate()
+        self.graph = capture.graph
```
eric-heiden, line 833: 🟠 This capture relies on Warp's ambient device even though `simulate()` launches work on `self.model.device`. If the model was finalized on a different device, the graph can be recorded in the wrong device context and then replayed as though it belongs to the model. Please wrap capture, and preferably replay, in `wp.ScopedDevice(self.model.device)`. The same ambient-device pattern should also be corrected in `example_cable_twist.py` and `example_cable_y_junction.py`.

newton/examples/cable/example_cable_cross_slide_table.py
```diff
@@ -827,13 +827,10 @@ def __init__(self, viewer, args):
         self.capture()
```
eric-heiden, line 827: ⚪ Please also update the nearby comment at `newton/examples/cable/example_cable_cross_slide_table.py:771`, which still describes these arrays as supporting "CUDA graph replay." Capture is no longer CUDA-only here, so "graph replay" or "captured replay" would keep the comment accurate.

newton/_src/solvers/kamino/_src/linalg/conjugate.py
c0d1f1ed: The intention is for all operations supported by CUDA graphs to also become supported by APIC, eventually. So I'd like to avoid making changes in Newton just to work around Warp's current limitations.

I've filed https://github.com/NVIDIA/warp/issues/1681 to track it.

But let me know if you strongly prefer doing the workaround in Newton and reverting it once Warp has proper support for this.

newton/examples/cable/example_cable_cross_slide_table.py
c0d1f1ed: Is that a scenario that is actually expected to be supported at this point, finalizing things on a different device?

If so, I think it's best to make the suggested changes in a separate PR.

newton/examples/cable/example_cable_cross_slide_table.py
c0d1f1ed: Done.

newton/examples/cable/example_cable_cross_slide_table.py
eric-heiden: You're right, I thought only the `model.device` was set from the example runner, but it actually sets the device globally via `wp.set_device(...)`.

eric-heiden (approved):

eric-heiden:
@chschuma-disney could you re-approve this PR? Then we can merge it.

chschuma-disney (approved):

PR #3502, opened 2026-07-13
Update mujoco-warp to 3.10.0.2 and re-enable elliptic cone tests

### LINKED ISSUE #3391: [BUG] test_friction_ramp_mujoco_warp_cuda_0 is flaky
### Bug Description

Test failed on an unrelated change in https://github.com/newton-physics/newton/actions/runs/28928400933/job/85821719059?pr=3373 

======================================================================
test_friction_ramp_mujoco_warp_cuda_0 (test_rigid_friction_ramp.TestRigidFrictionRamp.test_friction_ramp_mujoco_warp_cuda_0)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/dlami/nvme/actions-runner/_work/newton/newton/newton/tests/unittest_utils.py", line 498, in test_func
    func(self, device, **kwargs)
  File "/opt/dlami/nvme/actions-runner/_work/newton/newton/newton/tests/test_rigid_friction_ramp.py", line 209, in test_friction_ramp
    assert_grid_behavior(test, settle_q, final_q, final_qd, mus, angles_deg, box_ids, thresholds)
  File "/opt/dlami/nvme/actions-runner/_work/newton/newton/newton/tests/test_rigid_friction_ramp.py", line 187, in assert_grid_behavior
    test.fail("\n  ".join([f"{len(failures)} friction-ramp cell(s) failed:", *failures]))
AssertionError: 1 friction-ramp cell(s) failed:
  (mu=0.30, theta=10.0deg, crit=16.7deg): expected static but disp=0.0216 >= 0.02

----------------------------------------------------------------------
Ran 4261 tests in 1048.805s

FAILED (failures=1, skipped=144)

### Reproduction Script

```python

```

### System Information

_No response_

### LINKED ISSUE #3397: [BUG] test_reset_functionality_elliptic_cuda_0 is flaky
### Bug Description

`test_reset_functionality_elliptic_cuda_0` intermittently reaches the MuJoCo Warp solver's 100-iteration cap after resetting the robot state.

The same assertion failed in two merge-queue GPU runs:

- https://github.com/newton-physics/newton/actions/runs/28785537786/job/85351535554
- https://github.com/newton-physics/newton/actions/runs/28946249274/job/85881363670

Both failures occur in the post-reset half of `TestAnymalReset._run_reset_test()`:

```text
AssertionError: 100 not less than 90.0 : Solver iterations (100) are too high (>90), max allowed is 100. Simulation is unstable!
```

The first observed occurrence coincides with the merge of #2980, which upgraded MuJoCo Warp to 3.10.0.1. That PR also documented that corrected line-search cost precision can expose poorly damped contact configurations that reach the iteration cap.

Investigation found:

- Independent isolated invocations usually pass.
- A repeated in-process diagnostic reproduced the exact failure locally on CUDA.
- Disabling CUDA graph capture did not eliminate the failure.
- The immediate post-reset comparison accepts small differences in position, velocity, acceleration, and warm-start state. These differences can be amplified by the uncontrolled floating-base contact trajectory; sampled post-reset solver iterations varied substantially while the corresponding pre-reset values remained stable.

This appears to be a brittle convergence assertion around nondeterministic contact dynamics rather than a CUDA graph issue. The reset-fidelity check and the long-horizon stability check may need to use separate, more deterministic scenarios.

### Reproduction Script

```bash
for i in {1..100}; do
  WARP_CACHE_PATH=/tmp/warp-cache-anymal-reset uv run --extra dev \
    -m newton.tests --no-cache-clear \
    -k test_reset_functionality_elliptic_cuda_0 || break
done
```

### System Information

Observed on the AWS GPU GitHub Actions runner and reproduced intermittently on an NVIDIA RTX PRO 6000 Blackwell GPU with CUDA 12.9.

adenzler-nvidia:
## Description

Update mujoco-warp to 3.10.0.2, which includes the elliptic line-search fix google-deepmind/mujoco_warp#1512, and re-enable the tests that were disabled for the resulting elliptic friction cone instabilities on CUDA:

- Revert #3465 (anymal reset elliptic cone test, closes #3397)
- Revert #3483 (MuJoCo Warp friction ramp, closes #3391)

mujoco-warp 3.10.0.2 also adds `body_simple` to its Model, which enters the menagerie model field comparison. The flag is derived from the inertia representation, so Newton's re-diagonalized inertias can produce a different but physically equivalent classification (Shadow Hand and Allegro are affected). Skip it in the comparison for all models for now.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_anymal_reset
uv run --extra dev -m newton.tests -k test_rigid_friction_ramp
uv run --extra dev -m newton.tests -k test_menagerie_mujoco
```

The re-enabled tests were run repeatedly on CUDA without failures. A sweep comparing `body_simple` between Newton-built and native models across all menagerie robots found mismatches only for Shadow Hand and Allegro; their model comparisons pass with the skip.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved MuJoCo runtime compatibility for contact-gap handling by tightening the MuJoCo-Warp version pin.
  - Applied a MuJoCo model configuration tweak to set the linesearch iterative block dimension to 32 for more consistent solver behavior.

- **Tests**
  - Expanded CUDA execution for reset functionality by removing a conditional skip for the elliptic case.
  - Enhanced friction-ramp test coverage on CUDA by removing a Mujoco-Warp-specific flakiness skip.
  - Updated model comparison checks to ignore the `body_simple` field where it may differ after inertia re-diagonalization.


jcarius-nv (commented):
The `linesearch_iterative = 32` override needs a little more justification in the patch. `mujoco_warp.put_model()` deliberately raises this to 512 for models with `nv > 500`, so this changes launch tuning—and potentially performance—for all large models, beyond the stated dependency upgrade and test re-enablement scope. Please add a brief inline comment explaining the 64K registers-per-block constraint and referencing `google-deepmind/mujoco_warp#1512`, and include or point to large-model behavior/performance coverage for forcing 32 instead of retaining the upstream size-dependent setting.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -6558,6 +6558,7 @@ def add_body_equality(i: int):
         with wp.ScopedDevice(model.device):
             # create the MuJoCo Warp model
             self.mjw_model = mujoco_warp.put_model(self.mj_model)
+            self.mjw_model.block_dim.linesearch_iterative = 32
```
vreutskyy, line 6561: The 32-thread workaround appears necessary for the high-register elliptic specialization, but this unconditional override also replaces MuJoCo Warp's 512-thread tuning for large pyramidal models (`nv > 500`). Could we restrict it to elliptic models, or provide performance coverage showing that 32 is appropriate for both cone types?

vreutskyy (commented):
One scoped concern about the line-search block-size workaround; otherwise the dependency and test changes look good.

jcarius-nv:
Follow-up on the line-search block-size performance concern: the benchmark showed no regression.

The existing MuJoCo ASV cases have per-world nv values of 2–117, so none exercises the changed nv > 500 path. I therefore ran a temporary ASV benchmark at this exact PR head (1f293a95) with nv=504 and 128 worlds on an NVIDIA L4, comparing the same large pyramidal workload at 32 versus 512 threads.

Across three clean runs:

- 32 threads: 0.0514566 ms/world-step mean
- 512 threads: 0.0517567 ms/world-step mean
- 32 threads was 0.58% faster overall; the per-run advantage ranged from 0.46% to 0.66%

There were no capacity warnings in these runs. This synthetic large-model workload therefore found no performance regression from the 32-thread setting; on this hardware it was marginally faster.

newton/_src/solvers/mujoco/solver_mujoco.py
jcarius-nv: see https://github.com/newton-physics/newton/pull/3502#issuecomment-4967051554

jcarius-nv (approved):

PR #3521, opened 2026-07-15
Refactor raycast intersection functions

StafaH:
## Description

Restructures the ray-shape intersection routines in `newton/_src/geometry/raycast.py`
(and their callers in the Warp raytrace sensor) so that the intersection math is
cleanly separated from the world↔local coordinate mapping.

- **Merge `map_ray_to_local`**: the two variants (with/without per-axis scale) collapse
  into one `map_ray_to_local(transform, ray_origin, ray_direction, scale=vec3(1))`.
- **Primitives are now pure local-space**: `ray_intersect_{sphere,box,capsule,cylinder,cone,ellipsoid,plane}`
  take a ray already in the shape's local frame and return a local-space normal — no
  `geom_to_world` argument, no per-function inverse transform. `ray_intersect_shape`
  maps the world ray into local space **once** and rotates the resulting normal back to
  world **once**.
- **`ray_intersect_mesh` is local-in / local-out**, matching the primitives; callers map
  the ray and rotate the normal. The dead `ray_intersect_mesh_no_transform` helper is
  removed.
- **`_ray_intersect_mesh_smooth`** (sensor-layer smooth shading) now reuses
  `ray_intersect_mesh` instead of duplicating the ray mapping, mesh query, and
  backface-cull logic.
- **Compile-time specialization for normal-free paths**: `ray_intersect_shape` and
  `ray_intersect_mesh` are each built by a small factory parameterized on a
  `compute_normal` bool and gated with `wp.static(...)`, yielding `*_no_normal` variants
  with **zero source duplication**. The depth-only and shadow passes use the `*_no_normal`
  variants, so the surface-normal work (and, being unused, the primitives' local-normal
  computation via dead-code elimination) is folded out of those kernels at compile time.
  The shadow triangle-mesh path also switches to the cheaper `ray_intersect_mesh_anyhit`.

No public API change.

## Benchmark results

Measured with `asv/benchmarks/simulation/bench_sensor_tiled_camera.py` scenes on an
**NVIDIA RTX PRO 6000 Blackwell**, `warp 1.15.0.dev20260626`, at the CI config
(`res=64`, `worlds=4096`, `iters=50`, TILED order, shadows + textures). Numbers are
stable to <1% across repeated runs.

### Render throughput (ms/frame, lower is better)

| Scene | Mode | main | this branch | Δ |
|---|---|---:|---:|---:|
| `franka_cabinet` (mesh) | color+depth | 10.123 | 9.557 | **−5.6%** |
| `franka_cabinet` | color-only | 10.057 | 9.494 | **−5.6%** |
| `franka_cabinet` | depth-only | 4.297 | 4.217 | **−1.9%** |
| `quadruped` (mesh) | color+depth | 6.824 | 6.472 | **−5.2%** |
| `quadruped` | color-only | 6.742 | 6.367 | **−5.6%** |
| `quadruped` | depth-only | 2.710 | 2.654 | **−2.1%** |
| `shapes_256` (primitives) | color+depth | 5.285 | 5.211 | **−1.4%** |
| `shapes_256` | color-only | 5.276 | 5.194 | **−1.6%** |
| `shapes_256` | depth-only | 1.810 | 1.760 | **−2.8%** |

The color path is ~5–6% faster on mesh-heavy scenes (map-once + simpler primitives).
The depth-only path — which regressed ~15% in an earlier revision of this branch because
the refactor computed a normal the kernel discarded — is now **faster than `main`** thanks
to the `compute_normal=False` specialization (a ~15–17% recovery from that intermediate
revision).

### Cold JIT compile time — render megakernel (ms, lower is better)

Median of 3 alternating trials with the Warp cache **and** CUDA compute cache wiped before
each run. The render megakernel is generated from the render config and is scene-independent
(shapes/quadruped/franka compile the same kernel). The `_no_normal` variants add negligible
compile cost.

| Kernel variant | main | this branch | Δ |
|---|---:|---:|---:|
| color+depth | 1583 | 1147 | **−27.5%** |
| color-only | 1417 | 982 | **−30.7%** |
| depth-only | 872 | 599 | **−31.3%** |
| **total** | **3880** | **2727** | **−29.7%** |

Collapsing the seven primitives' per-function map/transform into a single dispatcher and
removing duplicated mesh logic cuts render-kernel compile time by **~30%**.

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_raycast                  # 32 pass (CPU + CUDA)
uv run --extra dev -m newton.tests -k test_sensor_tiled_camera      # 30 pass
uv run --extra dev -m newton.tests -k forward_depth                 # 6 pass (depth path)
uv run --extra dev -m newton.tests -k particles_multiworld          # 2 pass
```

Benchmarks reproduced by checking out `main` in a worktree and running
`bench_sensor_tiled_camera.py` scenes against each branch on the same GPU / venv.

## Checklist

- [x] The documentation (docstrings) is up to date with these changes
- [ ] `CHANGELOG.md` — n/a (internal refactor, no user-facing behavior change)

## Summary by CodeRabbit

* **New Features**
  * Added optional `fast_math` control to ray intersection queries.
  * Unified raycasting/dispatch flow with consistent shape-local (and scaled) handling for meshes and primitives, including consistent backface culling.
  * Enhanced raytrace variants for depth-only and shadow “any-hit” tests to reduce unnecessary normal work.
* **Bug Fixes**
  * Fixed intersection and returned normal consistency for scaled transforms by keeping primitive math in local space and transforming normals back only when needed.
* **Tests**
  * Updated raycast test routing and normal expectations to match the revised mesh-vs-primitive intersection behavior.


daniela-hase (dismissed):

newton/_src/sensors/warp_raytrace/raytrace.py
```diff
@@ -683,59 +584,24 @@ def first_hit_shape(
...
+                        ray_origin_local, ray_direction_local = raycast.map_ray_to_local(
+                            shape_transforms[si], ray_origin_world, ray_dir_world, shape_sizes[si]
+                        )
+                        hit_dist = raycast.ray_intersect_mesh_anyhit(
+                            ray_origin_local,
+                            ray_direction_local,
                             shape_source_ptr[si],
-                            False,
                             max_dist,
                         )
```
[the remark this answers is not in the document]
daniela-hase: I think backfacing triangles casting shadows is fine and actually makes a lot of sense :)

daniela-hase (approved):

eric-heiden:
Nice cleanup of the ray intersection paths! One behavior regression remains in the triangle-mesh shadow path.

newton/_src/sensors/warp_raytrace/raytrace.py
```diff
@@ -783,9 +649,9 @@ def first_hit_triangle_mesh(
         max_dist: wp.float32,
     ) -> wp.bool:
         if triangle_mesh_id:
-            hit_dist, _normal, _bary_u, _bary_v, _face_idx = raycast.ray_intersect_mesh_no_transform(
-                triangle_mesh_id, ray_origin_world, ray_dir_world, wp.static(config.enable_backface_culling), max_dist
-            )
+            # Triangle mesh is in world space; its local frame is the world frame (see
+            # closest_hit_triangle_mesh). Shadow rays only need any hit within ``max_dist``.
+            hit_dist = raycast.ray_intersect_mesh_anyhit(ray_origin_world, ray_dir_world, triangle_mesh_id, max_dist)
```
eric-heiden, line 654: 🟠 This any-hit query no longer honors `RenderConfig.enable_backface_culling`. The previous `ray_intersect_mesh_no_transform(...)` call passed the config flag, but `ray_intersect_mesh_anyhit(...)` has no culling argument, so back-facing triangle-mesh surfaces can now block shadow rays even when culling is enabled. Please use this fast path only when culling is disabled, or add a culling-capable any-hit path that preserves the configured behavior.

StafaH: I discussed with @daniela-hase as the issue came up before and we think this is fine for shadow rays and should not be a serious issue.

eric-heiden (approved):

PR #3524, opened 2026-07-15
Require collider setup keyword arguments

jcarius-nv:
## Description

Make every parameter of `SolverImplicitMPM.setup_collider()` keyword-only. The method already has a large optional parameter list, and adding `collider_particle_ids` in the middle would otherwise silently rebind positional calls.

Update the existing coupled-solver changelog entry with the keyword-only requirement and migration guidance.

## Checklist

- [ ] New or existing tests cover these changes (Python enforces keyword-only calls from the signature)

## Test plan

```bash
uvx pre-commit run -a
```

## New feature / API change

```python
solver.setup_collider(
    collider_meshes=meshes,
    collider_body_ids=body_ids,
    collider_particle_ids=particle_ids,
)
```

## Summary by CodeRabbit

- **New Features**
  - Added support for specifying particle IDs when configuring collider coupling.

- **Breaking Changes**
  - Collider configuration options must now be provided using named arguments. Update any positional calls to use keyword arguments.


adenzler-nvidia (approved):
Makes sense, thanks!

PR #3525, opened 2026-07-15
Make ASV install pins authoritative

adenzler-nvidia:
## Description

The exact pins in the ASV `install_command` ran before the wheel install, which made them suggestions rather than pins: pip replaced any pin that violated the wheel's own dependency floors. This is how benchmark runs silently used mujoco-warp 3.10.0.2 while the config still pinned 3.10.0.1 (#3502 bumped pyproject but missed asv.conf.json, and nobody noticed because the wheel's requirement won).

Reorder the install so the wheel goes first and the exact pins apply afterwards, making the file state the actual benchmark environment:

- The wheel `[dev]` install gets the NVIDIA extra index so resolution keeps working when the warp-lang floor is a nightly that PyPI cannot satisfy.
- The exact pins now genuinely control the environment, which also supports deliberately benchmarking newer mujoco/warp builds than pyproject requires.
- `pip check` runs last so a warp-lang pin below the project floor fails the run loudly instead of distorting the time series.
- Drop the numpy line (covered by the wheel's dependencies) and update the mujoco-warp pin to 3.10.0.2 to match pyproject.

`pip check` cannot catch stale pins for extra-gated requirements because installed metadata does not record which extras were requested, and mujoco/mujoco-warp sit behind the `sim` extra. A new pre-commit hook (`scripts/check_asv_pins.py`) closes that gap by statically validating every `pkg==version` pin in asv.conf.json against the pyproject requirements whenever either file changes, e.g.:

```
asv.conf.json pins mujoco-warp==3.10.0.1, but extra 'sim' requires 'mujoco-warp>=3.10.0.2,~=3.10.0'
```

## Checklist

## Test plan

- `uvx --with virtualenv asv run <branch>^! --bench <filter>` builds the environment through asv's own machinery: all install steps execute, `pip check` passes, and the final environment holds exactly the pinned versions (warp-lang 1.15.0.dev20260626, mujoco 3.10.0, mujoco-warp 3.10.0.2, torch 2.10.0+cu130).
- The pre-commit hook passes on this branch and fails with the message above when the stale mujoco-warp==3.10.0.1 pin is restored.
- `uvx pre-commit run -a` passes.

## Summary by CodeRabbit

* **Chores**
  * Updated environment setup for benchmark runs.
  * Improved dependency validation during installation.
  * Updated simulation tooling versions and streamlined development and simulation package installation.


asv.conf.json
```diff
@@ -15,11 +15,9 @@
...
-    "python -m pip install -U numpy",
-    "python -m pip install -U --pre warp-lang==1.15.0.dev20260626 --index-url=https://pypi.nvidia.com/",
-    "python -m pip install -U mujoco==3.10.0",
-    "python -m pip install -U mujoco-warp==3.10.0.1",
-    "python -m pip install -U torch==2.10.0+cu130 --index-url https://download.pytorch.org/whl/cu130",
-    "in-dir={env_dir} python -m pip install {wheel_file}[dev]"
+    "in-dir={env_dir} python -m pip install {wheel_file}[dev] --extra-index-url=https://pypi.nvidia.com/",
+    "in-dir={env_dir} python -m pip install {wheel_file}[sim] warp-lang==1.15.0.dev20260626 mujoco==3.10.0 mujoco-warp==3.10.0.2 --extra-index-url=https://pypi.nvidia.com/",
+    "python -m pip install torch==2.10.0+cu130 --index-url https://download.pytorch.org/whl/cu130",
+    "python -m pip check"
```
adenzler-nvidia, line 21: the mujoco_warp version upgrade is a drive-by fix for a missed upgrade earlier this week.

nvtw (approved):

PR #3528, opened 2026-07-15
Fix DiffSim bear graph capture and test horizon

### LINKED ISSUE #3522: Flaky Windows CUDA timeout in example_diffsim_bear
### Bug description

During Newton 1.4.0rc2 HORDE release validation, the CUDA test for `example_diffsim_bear` intermittently exceeded its 600-second subprocess timeout on Windows 11 with an NVIDIA A40.

Full test name:

```
test_examples.TestDiffSimExamples.test_diffsim.example_diffsim_bear_cuda_0
```

Observed frequency across nine full-suite runs:

- Python 3.11.15: 1/3 failures
- Python 3.12.13: 1/3 failures
- Python 3.13.14: 0/3 failures
- Overall: 2/9 failures (7/9 passed)

Both failures ended at exactly 600 seconds:

```text
subprocess.TimeoutExpired: Command [
  'python.exe', '-m', 'newton.examples.diffsim.example_diffsim_bear',
  '--device', 'cuda:0', '--test', '--quiet', '--viewer', 'null',
  '--num-frames', '240'
] timed out after 600 seconds
```

### Reproduction

On a Windows 11 CUDA system:

```bat
py -3.11 -m venv testenv
testenv\Scripts\activate
python -m pip install --upgrade pip
python -m pip install "newton[dev,torch-cu12]==1.4.0rc2"
set WARP_CACHE_ROOT=C:\wc
python -m newton.tests
```

Repeat the full suite several times. The direct subprocess command is:

```bat
python -m newton.examples.diffsim.example_diffsim_bear --device cuda:0 --test --quiet --viewer null --num-frames 240
```

The failure has so far only been observed under the full parallel suite; isolated repetition has not yet been tested.

### System information

- Newton: 1.4.0rc2
- OS: Windows 11, x86_64
- GPU: NVIDIA A40, 48 GiB, sm_86
- Warp: 1.15.0
- Warp CUDA Toolkit: 12.9
- Warp-reported CUDA driver API: 13.2
- Driver installation was skipped; the HORDE template's preinstalled driver was used
- Torch: 2.11.0+cu128
- Full suite: 5,016 discovered tests in 533 suites across 8 worker processes
- Test harness timeout: 600 seconds per example subprocess

### Diagnostics and open question

The timeout covers subprocess startup, Warp kernel compilation, and all 240 frames. The full suite runs eight workers concurrently, so first-use compilation, cache contention, or GPU contention may contribute. However, `subprocess.run(..., capture_output=True, timeout=600)` raises `TimeoutExpired`, and the harness currently does not report the exception's partial stdout/stderr. The available logs therefore cannot determine whether the process was still compiling kernels or was stuck during simulation.

Useful follow-up diagnostics would be:

- Preserve `TimeoutExpired.stdout` and `TimeoutExpired.stderr` in the test failure
- Log timestamps around initialization/compilation and frame execution
- Compare isolated, warmed-cache, and `-j 1` runs against the default 8-worker suite

Possibly related but not the same symptom: #1386 reported a now-closed out-of-bounds read in this example.

adenzler-nvidia:
## Description

Capture the DiffSim bear forward rollout and backward pass as separate graphs. The default 300-step rollout previously created one oversized graph executable that terminated inside `cudaGraphLaunch()` on affected CUDA systems; enqueuing the two graphs on the same stream preserves their ordering while avoiding that failure.

Disable particle-particle contacts in the example by setting `model.particle_grid = None` before constructing `SolverSemiImplicit`. Since 195454cc the solver rebuilds the particle grid during `step()`, activating particle-particle contacts whose gradients are documented as unsupported; training diverged at the fourth iteration for rollouts of 120 steps or longer.

Run the CUDA example test with four training iterations of 120 simulation steps. The previous registration used the example's 300-step default and completed at most one training iteration instead of the intended four. A 120-step horizon covers the gradient-corruption regression above, which a 60-step rollout cannot: at 60 steps all iterations are numerically identical with and without the opt-out. The existing CPU override continues to use 10 simulation steps.

Closes #3522.

## Checklist

## Test plan

```console
uv run --extra dev -m newton.tests -p test_examples.py -k example_diffsim_bear_cuda_0 -j 1
uv run --extra dev -m newton.tests -p test_examples.py -k example_diffsim_bear_cpu -j 1
python -m newton.examples.diffsim.example_diffsim_bear --device cuda:0 --test --quiet --viewer null --num-frames 1 --sim-steps 300
python -m newton.examples.diffsim.example_diffsim_bear --device cuda:0 --test --verbose --viewer null --num-frames 1200 --sim-steps 300
uvx pre-commit run -a
```

## Bug fix

**Steps to reproduce:**

1. Run the minimal reproduction from `main`.
2. On affected CUDA systems, observe the combined forward/backward graph terminate with `SIGSEGV` during its first launch.
3. Run the example for four training iterations with a rollout of 120 steps or longer (e.g. `--num-frames 1200 --sim-steps 300`) and observe the loss diverge at the fourth iteration.

**Minimal reproduction:**

```console
python -m newton.examples.diffsim.example_diffsim_bear --device cuda:0 --test --quiet --viewer null --num-frames 1 --sim-steps 300
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the `diffsim_bear` example so it no longer crashes with the default CUDA configuration.
  * Improved simulation stability to prevent divergence after several training iterations.
  * Updated the example’s simulation settings for more reliable multi-frame runs.

### DIFF
--- a/newton/tests/test_examples.py
+++ b/newton/tests/test_examples.py
@@ -801,7 +801,7 @@ add_example_test(
     TestDiffSimExamples,
     name="diffsim.example_diffsim_bear",
     devices=test_devices,
-    test_options={"usd_required": True, "num-frames": 4 * 60, "sim-steps": 60},  # train_iters * sim_steps
+    test_options={"usd_required": True, "num-frames": 4 * 120, "sim-steps": 120},  # train_iters * sim_steps
     test_options_cpu={"num-frames": 2, "sim-steps": 10},
     use_viewer=True,
 )
--- a/newton/tests/test_examples.py
+++ b/newton/tests/test_examples.py
@@ -801,7 +801,7 @@ add_example_test(
     TestDiffSimExamples,
     name="diffsim.example_diffsim_bear",
     devices=test_devices,
-    test_options={"usd_required": True, "num-frames": 4 * 60},  # train_iters * sim_steps
+    test_options={"usd_required": True, "num-frames": 4 * 60, "sim-steps": 60},  # train_iters * sim_steps
     test_options_cpu={"num-frames": 2, "sim-steps": 10},
     use_viewer=True,
 )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,7 +94,7 @@
 - Fix USD import ignoring ancestor material bindings with `strongerThanDescendants` strength when a mesh authors `material:binding` without applying `MaterialBindingAPI`: material resolution now uses UsdShade's canonical `ComputeBoundMaterial` unconditionally, which also adds collection-based binding support. Prims authoring bindings without the applied schema are invalid USD and now surface USD's own warning (once per prim per import) — fix such assets with `usdchecker` or `usd-validation-nvidia`.
 - Fix `ModelBuilder.add_usd()` to honor `ignore_paths` in the custom-frequency traversal, so prims under ignored subtrees no longer register spurious custom-frequency rows in two-pass import workflows. (#3406)
 - Fix Style3D solver divergence caused by isolated vertices.
-- Fix the `diffsim_bear` example crashing with its default CUDA configuration.
+- Fix the `diffsim_bear` example crashing with its default CUDA configuration and diverging after a few training iterations.
 - Fix USD joint `physics:collisionEnabled` import so joints with two explicit bodies honor authored collision behavior; joints to world continue to allow body/world collisions, and articulation-wide self-collision filtering remains additive.
 - Fix `ModelBuilder.add_usd()` to honor `PhysicsScene.gravityDirection`, including stage-to-builder rotation and per-world imports.
 - Fix `ViewerFile.is_running()` to return `False` after `ViewerFile.close()` so headless recording loops can terminate like interactive viewers. (#3094)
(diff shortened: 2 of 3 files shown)


mzamoramora-nvidia:
Manual / visual testing (`uv run -m newton.examples.diffsim.example_diffsim_bear`) shows that the example is not working properly. This PR implements nice changes for testing but the example behavior is still not very desirable.

newton/tests/test_examples.py
```diff
@@ -801,7 +801,7 @@ class TestDiffSimExamples(unittest.TestCase):
     TestDiffSimExamples,
     name="diffsim.example_diffsim_bear",
     devices=test_devices,
-    test_options={"usd_required": True, "num-frames": 4 * 60},  # train_iters * sim_steps
+    test_options={"usd_required": True, "num-frames": 4 * 60, "sim-steps": 60},  # train_iters * sim_steps
```
mzamoramora-nvidia: Could we also disable particle-particle contacts for this differentiable example before constructing `SolverSemiImplicit`?

This test now runs four iterations, but the 60-step rollout masks a pre-existing divergence at longer rollout lengths. I reproduced it at both 120 and the default 300 steps: iterations 1–3 remain stable, then iteration 4 reaches roughly 100,000 m/s and the loss jumps from about 175 to over 300,000.

`git bisect` identified `195454cc` as the first bad commit. It began rebuilding `model.particle_grid` during `SolverSemiImplicit.step()`. The solver documentation warns that this particle-particle contact path may corrupt gradient computation.

The documented opt-out fixes the issue on both current `main` and this PR:

```python
# Particle-particle contacts do not support correct gradients in SolverSemiImplicit.
self.model.particle_grid = None
self.solver = newton.solvers.SolverSemiImplicit(self.model, enable_tri_contact=False)
```

Could we retain this 60-step test and add a CUDA-only four-iteration test at 120 steps? That regression fails before the opt-out and passes afterward. I verified the complete fix locally in `c2434584`.

adenzler-nvidia: Added the opt-out in 454efee25, before solver construction as suggested.

For the tests, I went with a single four-iteration test at 120 steps rather than keeping the 60-step one alongside it: at a 120-step rollout, iterations 1-3 are numerically identical with and without the opt-out (149.3 / 98.4 / 48.9), so a 60-step test exercises no path the 120-step test does not already cover, and dropping it keeps the CI cost down (the registered CUDA test runs in ~71 s here).

Verified the new test fails before the opt-out (iteration 4 loss 46834) and passes afterward (8.1), and the default 300-step configuration now trains monotonically over four iterations (325 / 239 / 175 / 166).

mzamoramora-nvidia (approved):
Thanks!

PR #3530, opened 2026-07-15
Refine experimental inverse dynamics API

jcarius-nv:
## Description

Refine the experimental inverse-dynamics interface before its first release:

- Replace the public container, model factory, and `EvalType` flags with caller-owned arrays.
- Add `eval_inverse_dynamics_passive()`, where each optional output is computed only when its array is provided.
- Keep `eval_inverse_dynamics_force()` separate and pass `mass_matrix`, `joint_qdd`, the bias forces, and `joint_f` explicitly as keyword-only arguments.
- Allocate temporary scratch internally so callers manage only supported inputs and outputs while retaining CUDA graph capture.
- Support articulation masks in both functions and their `ArticulationView` counterparts.
- Validate all public array and mask shapes eagerly and reject unsupported CABLE-joint models without device synchronization.
- Document exact allocation shapes and exercise the examples as Sphinx doctests.

## Checklist

## Test plan

```text
uv run --extra dev -m newton.tests -k test_inverse_dynamics
uv run --extra docs --extra sim sphinx-build -W -b doctest docs docs/_build/doctest docs/concepts/articulations.rst
```

The focused inverse-dynamics suite passes 74 tests across CPU and CUDA, with one expected CPU skip for the CUDA graph-capture test. The articulation documentation passes 17 doctests.

## New feature / API change

```python
import newton
import warp as wp

state = model.state()

mass_matrix = wp.empty(
    (
        model.articulation_count,
        model.max_dofs_per_articulation,
        model.max_dofs_per_articulation,
    ),
    dtype=wp.float32,
    device=model.device,
)
gravity_force = wp.empty_like(state.joint_qd)
coriolis_force = wp.empty_like(state.joint_qd)
joint_f = wp.empty_like(state.joint_qd)

newton.eval_inverse_dynamics_passive(
    model,
    state,
    mass_matrix=mass_matrix,
    gravity_force=gravity_force,
    coriolis_force=coriolis_force,
)

joint_qdd = wp.zeros_like(state.joint_qd)
newton.eval_inverse_dynamics_force(
    model,
    state,
    mass_matrix=mass_matrix,
    joint_qdd=joint_qdd,
    coriolis_force=coriolis_force,
    gravity_force=gravity_force,
    joint_f=joint_f,
)
```


newton/_src/sim/articulation.py
```diff
@@ -1496,35 +1506,45 @@ def eval_inverse_dynamics_force(
     parent-frame-in-world rotation) before the sum, so ``tau`` is entirely in
     that world convention.
 
+    .. experimental::
+
     Args:
         model: The model containing articulation definitions.
         state: State providing ``body_q``, used to rotate the FREE/DISTANCE
             root ``H @ qddot`` wrench into the world frame. Must be consistent
             with the ``H`` and bias buffers (i.e. the state passed to
```
adenzler-nvidia: Leftover from the signature change: `H` is no longer a parameter, so this should refer to `outputs.mass_matrix`. Same in the intro paragraph above — "writing the result into ``tau`` in place" is now `outputs.tau`.

newton/_src/sim/articulation.py
```diff
@@ -1461,19 +1474,16 @@ def eval_articulation_inverse_dynamics_force_kernel(
...
-    gap_end = joint_qd_start[articulation_start[art_idx + 1]]
     for k in range(dof_end, gap_end):
         tau[k] = float(0.0)
 
 
 def eval_inverse_dynamics_force(
     model: Model,
     state: State,
-    H: wp.array3d[wp.float32],
+    outputs: InverseDynamicsOutputs,
```
gyeomannvidia: The label "outputs" is a bit confusing because InverseDynamicsOutputs contains inputs to and outputs from eval_inverse_dynamics_force().

I don't want to hold back the change but my preference is always to spell out the terms that are const and the terms that are not.  This is one of the concerns that drove the original design. Python doesn't allow parameters to be documented as const so maybe this just isn't a concern for python programmers. If it is a concern, and I guess it is, then the original design had at least this one advantage - we separated out the buffers that were logically constant from the buffers that were to be updated.

Anyway, a simple improvement is to change the name "outputs" here. It's ok for eval_inverse_dynamic() because we only write to the buffers owned by InverseDynamicsOutputs  and don't read from them. It's more confusing here for the reasons above.

newton/_src/sim/inverse_dynamics.py
```diff
@@ -101,11 +101,18 @@ def __init__(
         self.zero_gravity = wp.zeros(world_count, dtype=wp.vec3, device=device)
 
 
-class InverseDynamics:
-    """Inverse dynamics quantities for a batch of articulated rigid-body systems."""
+class InverseDynamicsOutputs:
```
gyeomannvidia: One consequence of the name change is that eval_inverse_dynamics_force() now has a fn arg of type InverseDynamicsOutputs.  We called that arg "outputs" and I've discussed that in a separate comment.  We can fix the name of the arg but the class name in the function signature  is still InverseDynamicsOutputs, even though it contains inputs to and outputs from eval_inverse_dynamics_force().

On reflection, I think that InverseDynamicsOutputs is not the best name for this class. It would be totally fine if we did not pass it to eval_inverse_dynamics_force() and instead pass the buffers individually.

Anyway, don't want to hold up the release but my preference would be to come up with a new name for this class or to leave eval_inverse_dynamics_force() as it was. Both options avoid the confusion of something that could  easily be interpreted as an output actually being used as input and output.

gyeomannvidia (commented):
The one thing that immediately stands out is a mask test  (none, all, some) for selection.eval_inverse_dynamics_force().  We can do this as a follow-up task.

newton/_src/sim/articulation.py
jcarius-nv: Addressed across c3661424 and 24250349: removed the stale H/tau parameter wording. The docstring now refers to inverse_dynamics_buffers.mass_matrix and InverseDynamicsBuffers.tau, with public Sphinx targets.

newton/_src/sim/articulation.py
jcarius-nv: Agreed. Renamed the neutral reusable container to InverseDynamicsBuffers and the force argument to inverse_dynamics_buffers. This makes its read/write role explicit without implying const or output-only semantics.

newton/_src/sim/inverse_dynamics.py
jcarius-nv: Agreed with the naming concern. Renamed the class and factory to InverseDynamicsBuffers and Model.inverse_dynamics_buffers() while retaining the container-first force API. Buffers is intentionally direction-neutral because eval_inverse_dynamics populates M, C, and g, while eval_inverse_dynamics_force reads those buffers and writes tau. The API remains experimental.

newton/_src/sim/articulation.py
jcarius-nv: Superseded by b3135937. The final API removes the container and the stale H and implicit tau wording entirely: eval_inverse_dynamics_force() now takes mass_matrix, joint_qdd, coriolis_force, and gravity_force explicitly and writes to the explicit tau output.

newton/_src/sim/articulation.py
jcarius-nv: Addressed by the redesign in b3135937. There is no longer an outputs or buffers argument: the force function read-only quantities (mass_matrix, joint_qdd, coriolis_force, and gravity_force) and writable tau output are all explicit keyword-only parameters.

newton/_src/sim/inverse_dynamics.py
jcarius-nv: Addressed by the redesign in b3135937. InverseDynamicsOutputs, InverseDynamicsBuffers, and the model factory were removed. eval_inverse_dynamics_passive() accepts optional caller-owned output arrays, while eval_inverse_dynamics_force() accepts the passive terms as explicit inputs and tau as an explicit output.

newton/tests/test_inverse_dynamics.py
```diff
@@ -1887,44 +1949,29 @@ def test_eval_inverse_dynamics_finite_on_distance_joint(self):
...
-        model = b.finalize(device=self.device)
-
-        with self.assertRaises(ValueError) as ctx:
-            model.inverse_dynamics()
-        self.assertIn("CABLE", str(ctx.exception))
+    def test_model_has_no_inverse_dynamics_allocator(self):
+        """Inverse-dynamics arrays are caller-owned, not allocated by Model."""
+        model = newton.ModelBuilder().finalize(device=self.device)
+        self.assertFalse(hasattr(model, "inverse_dynamics_buffers"))
+        self.assertFalse(hasattr(newton, "InverseDynamicsBuffers"))
```
adenzler-nvidia: I think we can remove this test

newton/_src/sim/inverse_dynamics.py
```diff
@@ -447,22 +346,24 @@ def _compute_coriolis_force(
...
     )
 
 
-def eval_inverse_dynamics(
+def eval_inverse_dynamics_passive(
     model: Model,
     state: State,
-    eval_type: InverseDynamics.EvalType,
-    inverse_dynamics: InverseDynamics,
+    *,
```
gyeomannvidia, line 357: "*," is a new convention in the public API. We do use it elsewhere for internal code but I don't see it used anywhere else in the public API.  I just want to flag this up.

adenzler-nvidia: yes, this is intended. We want to move the entire public API to keyword-only arguments and avoid relying on positional arguments. Otherwise we can never remove parameters that sit somewhere in the middle.

newton/_src/sim/inverse_dynamics.py
```diff
@@ -486,35 +387,53 @@ def eval_inverse_dynamics(
...
     """
-    if not (eval_type & InverseDynamics.EvalType.ALL):
-        raise ValueError(
-            f"eval_type {eval_type!r} does not include any recognized flag "
-            f"(MASS_MATRIX, GRAVITY_FORCE, CORIOLIS_FORCE)."
-        )
-
-    scratch = inverse_dynamics._scratch
+    if mass_matrix is None and gravity_force is None and coriolis_force is None:
+        raise ValueError("At least one inverse-dynamics output must be provided.")
```
adenzler-nvidia, line 426: `Model.inverse_dynamics()` used to reject `JointType.CABLE` models with a `ValueError` (CABLE has no motion-subspace implementation and `eval_fk` does not reconstruct it, so results are undefined). That guard and its test were removed along with the container, and nothing mentions CABLE here anymore — a cable model now silently produces garbage. Can we restore the validation somewhere outside the capture path (e.g. a flag cached on the model), or at minimum document the limitation in both docstrings?

newton/_src/sim/inverse_dynamics.py
```diff
@@ -447,22 +346,24 @@ def _compute_coriolis_force(
...
 
 
-def eval_inverse_dynamics(
+def eval_inverse_dynamics_passive(
     model: Model,
     state: State,
-    eval_type: InverseDynamics.EvalType,
-    inverse_dynamics: InverseDynamics,
+    *,
+    mass_matrix: wp.array3d[wp.float32] | None = None,
```
adenzler-nvidia, line 358: Non-blocking: `None` now means opposite things within the same API family — `eval_mass_matrix(H=None)` allocates internally and returns it, while `mass_matrix=None` here means "skip this output", and this function calls `eval_mass_matrix` underneath. Fine while experimental, but worth reconciling before stabilizing (my preference would be migrating `eval_mass_matrix`/`eval_jacobian` to the caller-allocated style).

newton/_src/sim/inverse_dynamics.py
```diff
@@ -447,22 +346,24 @@ def _compute_coriolis_force(
...
         state.joint_qd,
         scratch.zero_gravity,
-        inverse_dynamics.coriolis_force,
+        coriolis_force,
         mask=mask,
     )
 
 
-def eval_inverse_dynamics(
+def eval_inverse_dynamics_passive(
```
adenzler-nvidia, line 354: Non-blocking question: is "passive" the right word here? The function also computes M(q), which isn't a passive force, and readers coming from MuJoCo will associate "passive" with spring/damper/fluid forces (`mj_passive`) — which this deliberately excludes (no joint damping/friction/armature). Something like `eval_inverse_dynamics_terms` would avoid both collisions. Fine to keep given the experimental marking, but renaming is cheapest now.

newton/_src/sim/articulation.py
```diff
@@ -1461,70 +1466,88 @@ def eval_articulation_inverse_dynamics_force_kernel(
...
     model: Model,
     state: State,
-    H: wp.array3d[wp.float32],
-    qddot: wp.array[wp.float32],
+    *,
+    mass_matrix: wp.array3d[wp.float32],
+    joint_qdd: wp.array[wp.float32],
     coriolis_force: wp.array[wp.float32],
     gravity_force: wp.array[wp.float32],
     tau: wp.array[wp.float32],
```
adenzler-nvidia: Should this be `joint_f` (or `joint_tau`)? `qddot` became `joint_qdd` to match the `joint_q`/`joint_qd` naming, and the docstring pins this output's layout and convention to `Control.joint_f` — `control.joint_f.assign(...)` is the canonical consumer — so `tau` is now the one odd name in the signature.

docs/concepts/articulations.rst
```diff
@@ -740,52 +742,67 @@ Newton can evaluate the **manipulator equation** for an articulated rigid-body s
...
 user-supplied :math:`\ddot{q}` to produce :math:`\tau`.
 
 Both functions require ``state.body_q`` to be consistent with
 ``state.joint_q``: callers must invoke :func:`newton.eval_fk` (or
 otherwise update ``state.body_q``) first.
 
-The inverse dynamics container :class:`~newton.InverseDynamics` is allocated using :meth:`newton.Model.inverse_dynamics`.
-It holds the public output buffers and owns the internal RNEA/Jacobian scratch privately, so callers manage only the one object.
-
 .. code-block:: python
```
adenzler-nvidia: Since these snippets are being rewritten wholesale anyway: can we make them `.. testcode::` blocks? CI runs `sphinx-build -b doctest` on PRs, and this file already has the `articulation-view` `testsetup`/`testcode` group whose model (free root + revolute) would work here nearly as-is. Executable snippets would catch exactly the kind of API drift this PR is fixing.

adenzler-nvidia (commented):
The explicit keyword-only arrays are a clear improvement over both container iterations. A few points inline: one lost validation (CABLE), two naming questions, and two non-blocking notes.

newton/tests/test_inverse_dynamics.py
jcarius-nv: Removed in de1af05c. The test only asserted that superseded symbols were absent and no longer protected supported behavior.

asv/benchmarks/simulation/bench_inverse_dynamics.py
```diff
@@ -40,31 +40,43 @@ def setup(self):
...
+                self.model.articulation_count,
+                self.model.max_dofs_per_articulation,
+                self.model.max_dofs_per_articulation,
+            ),
+            dtype=wp.float32,
+            device=self.model.device,
+        )
+        self.gravity_force = wp.empty_like(self.state.joint_qd)
+        self.coriolis_force = wp.empty_like(self.state.joint_qd)
+        self.tau = wp.empty_like(self.state.joint_qd)
```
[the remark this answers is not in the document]
jcarius-nv: Addressed in de1af05c. tau is now initialized with wp.zeros_like(), so passive-only benchmark teardown always reads a defined finite value.

newton/_src/sim/articulation.py
```diff
@@ -1461,70 +1466,88 @@ def eval_articulation_inverse_dynamics_force_kernel(
...
+            ``(model.joint_dof_count,)``, dtype float.
         gravity_force: Gravity force ``g(q) = ∂U/∂q`` [N or N·m, depending
-            on joint type], shape ``(joint_dof_count,)``, dtype float,
-            e.g. :attr:`InverseDynamics.gravity_force`.
-        tau: Output joint forces [N or N·m, depending on joint type],
-            shape ``(joint_dof_count,)``, dtype float, written in place.
-            Typically :attr:`InverseDynamics.tau`.
+            on joint type], shape ``(model.joint_dof_count,)``, dtype float.
+        tau: Output joint force [N or N·m, depending on joint type], shape
+            ``(model.joint_dof_count,)``, dtype float.
```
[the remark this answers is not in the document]
jcarius-nv: Addressed in de1af05c. The force function and both ArticulationView methods now document mass-matrix units by row and column DOF type, acceleration units, and force or torque units for every physical array.

newton/_src/sim/inverse_dynamics.py
jcarius-nv: Confirmed; the keyword-only convention is intentional here. It protects this new experimental interface from positional coupling and matches the broader direction Alan described.

newton/_src/sim/inverse_dynamics.py
jcarius-nv: Addressed in de1af05c. ModelBuilder.finalize() now caches CABLE presence as host metadata, and both inverse-dynamics entrypoints raise eagerly without inspecting device arrays or breaking graph capture. The limitation is documented, and a regression test covers both functions on CPU and CUDA.

newton/_src/sim/inverse_dynamics.py
jcarius-nv: Agreed that the existing allocation behavior of eval_mass_matrix() and eval_jacobian() should be reconciled before these interfaces stabilize. For this bounded experimental change, None remains the explicit output-selection mechanism for eval_inverse_dynamics_passive(); migrating the older functions is separate work.

newton/_src/sim/inverse_dynamics.py
jcarius-nv: We discussed this naming tradeoff explicitly and are retaining the agreed eval_inverse_dynamics_passive name for this experimental release. The docstring enumerates the exact M, gravity, and Coriolis outputs, and the experimental marker leaves room to revisit the name before stabilization if it proves confusing.

I think passive dynamics is an appropriate term here because it's the dynamics of the articulation absent of any 'active' actuation torques, see https://en.wikipedia.org/wiki/Passive_dynamics

newton/_src/sim/articulation.py
jcarius-nv: We are retaining tau as the manipulator-equation result name agreed for this interface. Control.joint_f is the canonical destination, but naming the parameter joint_f would conflate the computed mathematical quantity with that particular storage field; joint_qdd instead names an input already defined by the joint-state layout.

docs/concepts/articulations.rst
jcarius-nv: Addressed in de1af05c. All three inverse-dynamics snippets now use the existing articulation-view testcode group. The targeted Sphinx run executes 17 articulation doctests with zero failures.

jcarius-nv:
Addressed the remaining review-body items in de1af05c:

- Validate mask shape in both inverse-dynamics entrypoints before launching kernels.
- Cover malformed mass_matrix, joint_qdd, coriolis_force, gravity_force, tau, and mask arrays in the public-function tests.
- Initialize benchmark tau for passive-only teardown.
- Document SI units consistently across the free functions and ArticulationView methods.
- Restore eager CABLE rejection using graph-safe host metadata and cover both entrypoints on CPU and CUDA.
- Convert the inverse-dynamics documentation snippets to executable doctests.

Validation: 74 focused CPU/CUDA tests passed, the articulation document passed all 17 doctests, and pre-commit passed.

newton/_src/sim/inverse_dynamics.py
jcarius-nv: good point, filed https://github.com/newton-physics/newton/issues/3538 for this as a follow-up.

newton/_src/sim/articulation.py
jcarius-nv: Reconsidered after comparing this against Newton’s existing joint-state and control vocabulary. You are right: the output has exactly the shape, ordering, units, and frame convention of `Control.joint_f`, and callers can pass that buffer directly. In aca8ac53 the public output keyword is now `joint_f` across `eval_inverse_dynamics_force()`, `ArticulationView`, docs, tests, and benchmarks. We retain `tau` only as the mathematical symbol in the manipulator equation and in internal kernel terminology.

adenzler-nvidia (approved):
All feedback addressed: the CABLE guard is back as host-cached metadata with eager validation and a regression test covering both entrypoints, the output is now joint_f across API/docs/tests/benchmarks, the docs snippets run as doctests, and mask shapes are validated. Thanks for the quick turnaround!

PR #3542, opened 2026-07-16
[Kamino] Remove unused inertia statistics

rubengrandia:
## Description

Kamino collected mass and inertia statistics. The problem is that these are not updated during a `solver.notify_model_changed()`. Instead of updating them, I propose to remove them as they appear to be unused. We can add them back once they have a real use case.

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Simplified rigid-body model conversion by removing unused per-world mass and inertia summaries.
  * Reduced model metadata to focus on currently supported properties.

* **Bug Fixes**
  * Improved formatting of the dynamic-constraint topology error message.

* **Tests**
  * Updated model comparison and diagnostic output to reflect the streamlined metadata.


Guirec-Maloisel (commented):
Makes sense to me, but I'd leave @vastsoun to approve as he might have had something in mind when adding these.

chschuma-disney (approved):

PR #3549, opened 2026-07-16
[Kamino] Notify CoM changes.

rubengrandia:
## Description

Implement `notify_model_changed` for changes in the body CoM ans shape transforms. This affects 
- initial poses
- joint transforms
- geom offsets

Also fixes toDefault reset to use the initial body velocity and alias this array.

## Checklist

- [x ] New or existing tests cover these changes
- [x ] The documentation is up to date with these changes

## Summary by CodeRabbit

- **New Features**
  - Exposed functionality for converting geometry offsets to center-of-mass coordinates.
  - Runtime model updates now refresh affected body poses, joint transforms, and geometry offsets.

- **Bug Fixes**
  - Improved synchronization of updated body, inertial, and shape properties.
  - Preserved initial velocities and state reset behavior after model changes.

- **Tests**
  - Added coverage for shape, body, and inertial property updates.
  - Verified updated values propagate correctly through solver state resets.

### DIFF
--- a/newton/_src/solvers/kamino/_src/solver_kamino_impl.py
+++ b/newton/_src/solvers/kamino/_src/solver_kamino_impl.py
@@ -538,15 +538,9 @@ class SolverKaminoImpl(SolverBase):
         base_u = None
         relative_base_u = False
         if isinstance(config.base_velocity, SolverKamino.ResetConfig.ToDefault):
-            # Set base velocity to its default if body velocities are not already reset to defaults.
+            # Set base velocity to zero if body velocities are not already reset to zero
             if not isinstance(config.body_velocities, SolverKamino.ResetConfig.ToDefault):
-                get_base_u_from_joint_u_and_body_u(
-                    model=self._model,
-                    joint_u=self._model.joints.dq_j_0,
-                    body_u=self._model.bodies.u_i_0,
-                    base_u=self._base_u,
-                    world_mask=world_mask,
-                )
+                self._base_u.zero_()
                 base_u = self._base_u
         elif isinstance(config.base_velocity, SolverKamino.ResetConfig.Preserve):
             # Extract current base velocity if body poses/velocities are modified but base velocity should be preserved
--- a/newton/_src/solvers/kamino/_src/solver_kamino_impl.py
+++ b/newton/_src/solvers/kamino/_src/solver_kamino_impl.py
@@ -538,9 +538,15 @@ class SolverKaminoImpl(SolverBase):
         base_u = None
         relative_base_u = False
         if isinstance(config.base_velocity, SolverKamino.ResetConfig.ToDefault):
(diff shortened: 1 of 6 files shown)


vastsoun (dismissed):
LGTM

newton/_src/solvers/kamino/_src/solver_kamino_impl.py
```diff
@@ -538,9 +538,15 @@ def _check_length(data: wp.array[Any], name: str, expected: int):
...
             if not isinstance(config.body_velocities, SolverKamino.ResetConfig.ToDefault):
-                self._base_u.zero_()
+                get_base_u_from_joint_u_and_body_u(
+                    model=self._model,
+                    joint_u=self._model.joints.dq_j_0,
+                    body_u=self._model.bodies.u_i_0,
+                    base_u=self._base_u,
+                    world_mask=world_mask,
+                )
                 base_u = self._base_u
```
Guirec-Maloisel, line 544: Here I think we should also set `relative_base_u = False`, indicating that the base velocity should be reset to its initial value, up to the transformation applied by resetting the base pose (similarly to the block just below for the `Preserve` case)

newton/_src/solvers/kamino/solver_kamino.py
Guirec-Maloisel, line 541: The docstring here should be updated, as we don't reset this to zero anymore, but to the initial velocity as read from the initial joint/body states. Proposed update:
"To Default: reset the base velocity to its initial value, up to transformation due to the new base pose if applicable."

newton/_src/solvers/kamino/_src/solver_kamino_impl.py
Guirec-Maloisel: Actually I'm not sure that would be sufficient. I think the current code behaves like swapping the order of setting the base pose and the base velocity if `relative_base_u` is set to True. In the block below where we want to preserve the current velocity that works fine, but if we want to reset to the default we'd need to account for the current pose as well...

Guirec-Maloisel (changes requested):
The changes regarding taking into account CoM updates look good to me.

The additional change for resetting to initial instead of zero velocities in the default are a bit more problematic (see comments), I think that could have been part of the reason why I chose to reset to zero up to now (plus that's what the previous code was doing if you didn't provide actuator_u, base_u etc). I'm a bit torn now between adding the necessary changes for the corner case that wouldn't work as expected (non-trivial initial base velocity, and resetting to default velocity but non-default base pose), or keeping the current behavior (letting the user manually pass the desired non-zero velocities through FromJointU / FromActuatorU and FromBaseU)

rubengrandia:
@Guirec-Maloisel , I have reverted the use of the initial body velocity. We can try leave this as a separate issue.

Guirec-Maloisel (approved):

PR #3551, opened 2026-07-16
Fix SensorTiledCamera deformable triangle rendering to respect per-particle world indices

### LINKED ISSUE #3487: [BUG] Newton Warp renderer leaks deformable geometry across envs in multi-env tiled camera output
## Summary

In the franka_soft rendering golden images, a small horizontal bar appears in the top-middle of env_0 when using the Newton Warp renderer (newton_renderer). The artifact is not present in isaacsim_rtx or ovrtx renders, and it does not appear in env_1/env_2/env_3 tiles.

Image

Runtime investigation shows it is env_3's deformable bar bleeding into env_0's camera tile because deformable triangle-mesh geometry is raytraced from the global state.particle_q buffer without per-world scoping, while rigid shapes are correctly partitioned per world.

## Root cause

- Rigid shapes are correctly scoped per world via bvh_shapes_group_roots.
- Deformable geometry is raytraced from the flat, all-worlds state.particle_q triangle mesh and is not restricted to the tile's own world.

In the franka_soft 2×2 layout (env_spacing=3.0), env_0's camera looks diagonally across the grid. env_3's deformable at (-1.0, 1.5, 0.05) lies almost in that viewing direction at ~4.9 m, so it appears as a small bar near top-center in env_0's tile. Other env cameras do not see a diagonal deformable inside their frustum, so bleed is only visible in env_0.

### Reproduction Script

```sh
# install uv
curl -LsSf https://astral.sh/uv/install.sh | sh

# Clone huidongc's fork of IsaacLab repo and checkout the branch that increases the camera clipping range to expose the bug visually
git clone -b repro-newton-issue-3487 https://github.com/huidongc/IsaacLab.git

cd IsaacLab

# create env_isaaclab
./isaaclab.sh --uv
source env_isaaclab/bin/activate

# Install deps
./isaaclab.sh -i
./isaaclab.sh -i ov[all]

./isaaclab.sh -p -m pytest -v source/isaaclab_tasks/test/core/test_rendering_franka_soft_kitless.py::test_rendering_franka_soft_kitless[newton-newton_warp-rgb]
```

### System Information

_No response_

daniela-hase:
## Description

Fix `SensorTiledCamera` deformable triangle rendering so cloth/soft triangle meshes respect per-particle world indices in multi-world scenes.

- Build the render triangle mesh with Warp mesh groups derived from `model.particle_world`.
- Cache per-world triangle mesh group roots and pass them through color, depth-only, and shadow raytrace paths.
- Thread grouped mesh root arguments through internal mesh raycast helpers after the main merge.
- Add regression coverage proving cloth triangles render in their own world and do not leak into other worlds.
- Add a `CHANGELOG.md` entry for the user-facing fix.

## Checklist

## Test plan

```bash
```

## Bug fix

**Steps to reproduce:**

1. Create a multi-world cloth model with each world offset in space.
2. Render each world through `SensorTiledCamera`.
3. Place a camera for one world over another world's cloth.
4. Observe that deformable triangle hits can leak across worlds.

After this PR, triangle mesh ray queries are restricted to the current world plus the global world.

Closes https://github.com/newton-physics/newton/issues/3487

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed tiled camera cloth rendering so deformable triangle meshes respect per-particle world assignment.
  * Updated triangle-mesh raycasting, shadow/any-hit checks, and depth-only queries to use grouped mesh traversal with per-group root scoping to avoid cross-world hits.
* **Tests**
  * Updated raycasting tests to account for grouped traversal parameters.
  * Added a regression test ensuring cloth triangle-mesh rendering remains isolated across multiple worlds.


newton/_src/geometry/raycast.py
```diff
@@ -613,6 +614,7 @@ def ray_intersect_mesh(
         mesh_id: wp.uint64,
         enable_backface_culling: bool,
         max_t: float,
+        root: int,
```
StafaH: Can we make default be -1? That would keep it backwards compatible and avoid having to pass -1 everywhere

newton/_src/sensors/warp_raytrace/render_context.py
```diff
@@ -436,10 +444,31 @@ def triangle_indices(self, triangle_indices: wp.array[wp.int32]):
...
+            triangle_indices_np = self.triangle_indices.reshape((-1, 3)).numpy()
+            particle_world_np = self.particle_world.numpy()
+            triangle_world_np = particle_world_np[triangle_indices_np[:, 0]]
+            triangle_groups_np = np.where(triangle_world_np < 0, self.world_count, triangle_world_np).astype(np.int32)
+            triangle_groups = wp.array(triangle_groups_np, dtype=wp.int32, device=self.device)
+
+            self.triangle_mesh = wp.Mesh(
+                self.triangle_points, self.triangle_indices, groups=triangle_groups, bvh_constructor="sah"
+            )
+            self.triangle_mesh_group_roots = wp.full(self.world_count + 1, value=-1, dtype=wp.int32, device=self.device)
```
StafaH: This should already be initialized with -1 from the init methods above? Or is it possible they have no been initialized it reaches here

newton/_src/geometry/raycast.py
daniela-hase: Oh yeah, absolutely. It somehow never occurred to me that we can define default values for kernel args 😅

newton/_src/sensors/warp_raytrace/render_context.py
daniela-hase: Yep, good point. Removed the line here :)

eric-heiden (approved):
LGTM!

Issue #3383, opened 2026-07-07
Import USD deformable rest shapes


mzamoramora-nvidia:
Follow up on #3192 by adding rest-shape support for deformable bodies.

This feature requires both USD parsing and additional Newton API support:

- Cloth and volume need separate live and rest geometry, including rest topology, rest-based mass distribution, and cloth bend angles.
- Cables need model and solver support for separate rest centerlines and material frames.
- Missing or malformed rest data needs a clear validation and fallback policy.

Support should cover `restShapePoints`, `restTriVertexIndices`, `restTetVertexIndices`, `restBendAnglesDefault`, `restAdjTriPairs`, `restBendAngles`, and `restNormals`, with tests for disjoint rest topology and deformed initial states.

PR #3553, opened 2026-07-16
Import USD cloth and volume rest shapes

devshahofficial:
## Description

Import separate live and rest geometry for USD cloth and volume deformables.

- Use `restShapePoints` with optional disjoint `restTriVertexIndices` or `restTetVertexIndices` for FEM rest bases.
- Distribute density-derived mass from rest triangle areas and tetrahedron volumes.
- Honor `restBendAnglesDefault`, `restAdjTriPairs`, and `restBendAngles`.
- Validate malformed rest data and fall back to live geometry with a focused warning.

This advances #3383. Cable rest centerlines and material frames remain explicitly documented as unsupported because the current rod model cannot represent a separate live centerline and rest material frame without additional model/solver API.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_import_usd_deformable
uvx pre-commit run -a
```

128 focused tests pass on CPU; one device-specific test is skipped.

## New feature / API change

```python
import newton

builder.add_usd("deformable.usda")
model = builder.finalize()
```

Cloth and volume FEM state now uses the authored rest geometry while particle positions retain the authored live pose.

## Summary by CodeRabbit

* **New Features**
  * USD imports now support authored cloth rest geometry (rest shape + rest bend-angle data) and authored deformable volume rest geometry (tet rest shape + tet indices), using rest inputs to drive imported deformation/material state.
* **Bug Fixes**
  * Enhanced rest-geometry validation (handles non-finite, degenerate, inverted, and zero/invalid rest tet volumes) with safe fallback to simulation-derived rest; improved cloth/volume orientation and mass initialization to match authored rest.
* **Documentation**
  * Clarified cloth/volume rest-handling semantics, cable rest limitation scope, and rest-based mass distribution precedence.
* **Tests**
  * Added/expanded coverage for cloth rest-driving behavior, invalid rest/warning behavior, triangulated cloth cases, and left-handed rest index orientation; updated one mixed-import test name.


newton/_src/utils/import_usd_deformable_cloth.py
```diff
@@ -32,9 +32,80 @@
...
+    if np.any(normal_lengths <= 1.0e-12) or np.any(edge_lengths <= 1.0e-12):
+        raise ValueError("rest shape contains a degenerate triangle")
+    normals /= normal_lengths[:, None]
+    e1 = qp / edge_lengths[:, None]
+    e2 = np.cross(normals, e1)
+    rest_basis = np.stack((e1, e2), axis=1) @ np.stack((qp, rp), axis=2)
+    areas = np.linalg.det(rest_basis) * 0.5
+    if np.any(areas <= 1.0e-12):
+        raise ValueError("rest shape contains an inverted or degenerate triangle")
+    return np.linalg.inv(rest_basis), areas
```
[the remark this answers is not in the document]
devshahofficial: @coderabbitai Good catch. 4c2e668a now rejects non-finite cloth geometry, triangle areas, and tet volumes before inversion, then uses the existing warning/live-geometry fallback. I also removed the redundant test comment and added a NaN/Infinity regression. All 128 deformable-import tests pass (1 skipped), along with pre-commit.

newton/_src/utils/import_usd_deformable_volume.py
```diff
@@ -208,6 +258,21 @@ def _deformable_import_volume(ctx: _DeformableImportContext) -> None:
 
         soft_p0, soft_t0 = builder.particle_count, builder.tet_count
         builder.add_soft_mesh(**add_soft_mesh_kwargs)
+        if rest_data is not None:
+            rest_poses, rest_volumes = rest_data
+            for local_tet, pose in enumerate(rest_poses):
+                builder.tet_poses[soft_t0 + local_tet] = pose.tolist()
```
mzamoramora-nvidia, line 295: I found a case where a valid rest tet has a flat or inverted live state. `add_soft_mesh()` does not add the live tet. The assignment here then raises `IndexError`, after its particles have already been added.

Newton’s USD importers generally warn and skip unsupported geometry before changing the builder. Could you add a preflight check for this case?

Supporting inverted live tets may require a larger builder change. For this PR, warning and skipping the prim cleanly would be enough. Please add a regression test that also checks that no partial state remains.

newton/_src/utils/import_usd_deformable_cloth.py
```diff
@@ -234,6 +363,16 @@ def _deformable_import_cloth(ctx: _DeformableImportContext) -> None:
...
+        for local_tri, (pose, area) in enumerate(zip(rest_poses, rest_areas, strict=True)):
+            builder.tri_poses[t0 + local_tri] = pose.tolist()
+            builder.tri_areas[t0 + local_tri] = float(area)
+        for particle in range(p0, builder.particle_count):
+            builder.particle_mass[particle] = 0.0
+        for local_tri, area in enumerate(rest_areas):
+            for vertex in tri_faces[local_tri]:
+                builder.particle_mass[p0 + int(vertex)] += density * float(area) / 3.0
+        for local_edge, angle in enumerate(rest_angles):
+            builder.edge_rest_angle[e0 + local_edge] = float(angle)
```
mzamoramora-nvidia, line 410: `edge_rest_length` still comes from the live pose created by `add_cloth_mesh()`. VBD multiplies the bend stiffness by this value, so live stretch can change the bend response.

Could you check whether this should also be derived from the authored rest geometry? If so, please update it and add a stretched-live/rest-shape test.

Disjoint rest topology may provide two lengths for one simulation edge. We should handle that case explicitly instead of silently using the live length.

newton/_src/utils/import_usd_deformable_volume.py
```diff
@@ -186,6 +203,39 @@ def _deformable_import_volume(ctx: _DeformableImportContext) -> None:
...
+                        if len(rest_indices) != tetmesh_for_builder.tet_count:
+                            raise ValueError(
+                                "restTetVertexIndices must contain one tetrahedron per simulation tetrahedron"
+                            )
+                        if np.any(rest_indices < 0) or np.any(rest_indices >= len(rest_points)):
+                            raise ValueError("restTetVertexIndices contains an out-of-range point index")
+                    if _world_matrix_reflects(soft_mesh_mat):
+                        rest_indices = rest_indices.copy()
+                        rest_indices[:, [1, 2]] = rest_indices[:, [2, 1]]
+                    rest_data = _tet_rest_data(rest_points, rest_indices)
```
mzamoramora-nvidia, line 260: Newton already normalizes live TetMesh indices when `orientation="leftHanded"`.

Explicit `restTetVertexIndices` do not receive the same correction here. I tested this case, and a valid rest tet was treated as inverted. The importer then fell back to the live shape.

Could you apply the effective TetMesh orientation to explicit rest indices as well? Please add a test for a left-handed TetMesh with authored rest indices.

mzamoramora-nvidia (commented):
Thanks for contributing support for cloth and volume rest geometry. The element mapping and rest-based mass distribution are heading in the right direction.

I left a first round of inline comments on three areas:

- Cleanly handling unsupported live tet states before changing the builder.
- Checking whether cloth edge rest lengths also need to come from the rest shape.
- Applying the existing TetMesh orientation handling to explicit rest indices.

The focused deformable-import tests pass locally. The inline comments describe the cases I tested and suggest regression coverage.

The branch also conflicts with current `main` and brings unrelated entries back into the changelog. You can either rebase onto `main` or merge `main` into the branch. After resolving the conflicts, please check that only the new rest-geometry changelog entry remains.

This is the first review round. I will take another look after these comments are addressed and the branch is updated. Thanks again for working on this.

devshahofficial:
@mzamoramora-nvidia I worked through the full first round in 8d511bb1 and brought the branch up to current `main`. Live flat/inverted tets now preflight and skip before builder mutation, explicit rest-tet indices follow TetMesh handedness, and cloth bending-edge lengths come from the rest geometry; conflicting disjoint copies warn and fall back explicitly. I added regressions for each case and cleaned the changelog drift. Both deformable importer modules pass (71 tests, 4 CUDA-only skips), as does the full pre-commit suite. Ready for another look when you have a chance.

devshahofficial:
@mzamoramora-nvidia I’ve also refreshed this branch onto current `main` in f9e56947, so the merge blocker should be cleared. The cloth, mixed, and volume deformable importer suites pass (60 tests, 1 skip), and pre-commit is clean. The three requested fixes from your first round remain in place and are ready for another pass.

mzamoramora-nvidia:
Fresh pass over `f9e56947`. The three concerns from the previous round are addressed, and the disjoint-topology work looks right. I checked the `restShape` dihedral sign convention against the VBD kernel and the `restAdjTriPairs` → edge mapping on a larger mesh, and both are correct.

Two things I'd like to resolve before merge: explicit `restBendAngles` are authored in degrees and stored as radians, and a cloth prim that passes the new degenerate-area preflight but fails the rest-data edge-length check now aborts the whole stage import where `main` imports it. The rest are smaller: the all-or-nothing rest-shape and bend-angle fallbacks, two docs limitations that disappeared, one test to add, and imperative docstrings on the new tests.

@andrewkaufman, tagged you inline on one spec question: what the proposal intends when two disjoint rest copies of one simulation edge disagree in length. I think we should go with last-one-wins; let me know if you read it differently.

newton/_src/utils/import_usd_deformable_cloth.py
```diff
@@ -32,9 +32,110 @@
...
+        for edge_index, (tri0, tri1) in enumerate(adjacency.edge_tri_indices)
+        if tri1 != -1
+    }
+    for pair, angle in zip(pairs, angles, strict=True):
+        key = tuple(sorted((int(pair[0]), int(pair[1]))))
+        if key not in edge_by_pair:
+            raise ValueError(f"restAdjTriPairs contains non-adjacent triangle pair {key}")
+        if not math.isfinite(float(angle)):
+            raise ValueError(f"restBendAngles contains non-finite value {angle}")
+        rest_angles[edge_by_pair[key]] = float(angle)
```
mzamoramora-nvidia, line 111: Authored `restBendAngles` are stored straight into `edge_rest_angle`, which Newton documents as `[rad]` and the VBD kernel uses as radians (`particle_vbd_kernels.py:718-721`). The proposal authors them in degrees (its example is a pleat "set to something like 75 degrees", and `"flat"` is "zero degrees"), so a 75° pleat becomes 75 radians with no warning.

Could we apply `math.radians()` here? Angles computed from `restShape` (line 92) are already radians and should stay as-is. Worth a short comment citing that proposal paragraph, since the unit is stated only in prose and the schema declares a bare `float[]`.

The existing test authors `0.4` and asserts `0.4`; switching it to `75.0` / `math.radians(75.0)` would make the unit unmistakable.

newton/_src/utils/import_usd_deformable_cloth.py
```diff
@@ -139,6 +236,73 @@ def _deformable_import_cloth(ctx: _DeformableImportContext) -> None:
...
+                rest_edge_lengths = _cloth_rest_edge_lengths(
+                    np.asarray(tri_faces, dtype=np.int64), rest_points, rest_triangles
+                )
+        else:
+            if rest_indices_value is not None and len(rest_indices_value):
+                warnings.warn(
+                    f"{path}: restTriVertexIndices is authored without restShapePoints; using the simulation shape.",
+                    stacklevel=2,
+                )
+            rest_poses, rest_areas = _cloth_rest_data(rest_points, rest_triangles)
```
mzamoramora-nvidia, line 276: This `_cloth_rest_data()` call is outside any handler, as is the copy at line 266 inside the `except` block. The preflight at lines 228-237 only rejects on triangle area, while `_cloth_rest_data()` also rejects on edge length (line 50), so geometry that satisfies the first and fails the second raises out of `add_usd()` and aborts the whole stage import. A triangle with a `1e-13` edge and `5e-9` area does exactly that on this head; `main` imports it.

Rather than adding the missing thresholds to the preflight and keeping two lists in sync, could the preflight just call `_cloth_rest_data()` / `_cloth_rest_edge_lengths()` on the live geometry and warn and skip if they raise? The live result is what all three branches fall back to anyway, so keeping it lets the `else` branch and the `except` handler reuse it instead of recomputing, and the handler can no longer raise while handling an exception.

newton/_src/utils/import_usd_deformable_cloth.py
```diff
@@ -32,9 +32,110 @@
...
+                continue
+            sim_tri = sim_triangles[tri]
+            local_v0 = int(np.flatnonzero(sim_tri == v0)[0])
+            local_v1 = int(np.flatnonzero(sim_tri == v1)[0])
+            rest_edge = rest_points[rest_triangles[tri, local_v1]] - rest_points[rest_triangles[tri, local_v0]]
+            triangle_lengths.append(float(np.linalg.norm(rest_edge)))
+        if len(triangle_lengths) == 2 and not math.isclose(
+            triangle_lengths[0], triangle_lengths[1], rel_tol=1.0e-6, abs_tol=1.0e-12
+        ):
+            raise ValueError(f"disjoint rest topology assigns inconsistent lengths to simulation edge ({v0}, {v1})")
```
mzamoramora-nvidia, line 136: Disjoint rest topology can legitimately give the two rest copies of one simulation edge different lengths, since the proposal calls the rest connectivity "unconstrained". Raising here makes the handler at lines 262-269 discard the *entire* rest shape: on the mesh in the new regression test, `tri_areas` become the live `[2.0, 2.0]` instead of the rest `[0.5, 1.0]`, and mass follows.

Could we degrade only this edge, taking the last rest copy and warning with both lengths, and leave the poses, areas, masses, and bend angles rest-derived? `edge_rest_length` only scales bending stiffness (`k = stiffness * edge_rest_length`), so the pick is low-stakes. The regression test would then assert that only `edge_rest_length` degrades.

@andrewkaufman, Newton stores one `edge_rest_length` per bending edge, so it has to pick when the two copies disagree. I think we should go with last-one-wins, matching how the importer already resolves duplicate `restAdjTriPairs` entries. Does the proposal intend anything different?

newton/_src/utils/import_usd_deformable_cloth.py
```diff
@@ -139,6 +236,73 @@ def _deformable_import_cloth(ctx: _DeformableImportContext) -> None:
...
+                deformable_read(prim, "restAdjTriPairs"),
+                deformable_read(prim, "restBendAngles"),
+            )
+        except (IndexError, TypeError, ValueError) as exc:
+            warnings.warn(
+                f"{path}: invalid authored rest bend angles ({exc}); using '{rest_default}' defaults.", stacklevel=2
+            )
+            rest_angles = _cloth_rest_angles(
+                np.asarray(tri_faces, dtype=np.int64), rest_points, rest_triangles, rest_default, None, None
+            )
```
mzamoramora-nvidia, line 304: One bad entry in `restAdjTriPairs` drops every authored angle, valid ones included. A length mismatch against `restBendAngles` is global and should keep falling back wholesale, but a non-adjacent pair or a non-finite angle is per-entry, so could we skip just that entry, name it in the warning, and keep the rest? Duplicate pairs can stay last-wins (the proposal defines no precedence), but they should warn rather than resolve silently.

docs/concepts/usd_parsing.rst
```diff
@@ -122,13 +125,11 @@ Limitations
...
-  graphs drop ``restShapePoints``. For a standalone cable, a valid ``restShapePoints`` supplies
-  only the segment lengths used to convert the material moduli into joint stiffness; the rod
-  itself is still built relaxed at the current ``points`` pose, and mass distribution also uses
-  the current geometry. A body saved in a deformed pose therefore resumes relaxed at that pose
-  instead of springing back.
+* **Cable rest state** -- welded cable graphs drop ``restShapePoints``. For a standalone cable,
+  valid ``restShapePoints`` supplies only the segment lengths used to convert material moduli
+  into joint stiffness; the rod itself is still built relaxed at the current ``points`` pose,
+  ``restNormals`` is not imported, and mass distribution uses the current geometry. Cloth and
+  volume rest geometry is supported as described above.
```
mzamoramora-nvidia, line 132: The rewritten section drops the "Rest state" gap and says cloth and volume rest geometry "is supported as described above". Two gaps are worth keeping visible: malformed or ambiguous rest data falls back to the live shape for the whole prim, and time-varying rest shapes, which the proposal anticipates, are read once at import.

mzamoramora-nvidia (commented):
Inline notes on the current head, following the summary above.

newton/tests/test_import_usd_deformable_cloth.py
```diff
@@ -52,6 +52,70 @@ def test_cloth_quad_mesh_is_triangulated(self):
         self.assertEqual(group_range(builder, "cloth", "/World/Cloth", "tri"), (0, 2))
         self.assertEqual(builder.particle_count, 4)
 
+    def test_cloth_rest_shape_drives_material_state(self):
+        """Disjoint rest topology sets FEM poses, mass, edge lengths, and authored bend angles."""
```
mzamoramora-nvidia, line 56: `AGENTS.md` asks for imperative test summaries, and the seven added or renamed docstrings here are declarative: `"""Disjoint rest topology sets ..."""` rather than `"""Verify disjoint rest topology sets ..."""`.

None of the existing docstrings in these files follow the rule yet, so this is not a gap you introduced. Still, it would be good to start adhering to it with the new tests rather than widen the gap; the older ones can be brought over in a separate pass.

newton/_src/utils/import_usd_deformable_cloth.py
andrewkaufman: I defer to @sschirm on that question

PR #3555, opened 2026-07-16
Update Menagerie pin for Robotiq masses

### LINKED ISSUE #2931: Robotiq2f85V4 and WonikAllegro USD assets have body mass/inertia that disagrees with source MJCF
## Summary

The USD versions of two menagerie robots carry body mass/inertia values that don't match their source MJCF. This is upstream menagerie data, not a Newton parser bug, but it forces Newton's USD dynamics tests to use relaxed tolerances or to backfill body inertia from native MJCF before comparing.

## Robotiq2f85V4

Finger-pad bodies have a mass mismatch between USD and MJCF. **Already fixed upstream** by menagerie PR [google-deepmind/mujoco_menagerie#252](https://github.com/google-deepmind/mujoco_menagerie/pull/252) (merged 2026-04-16, thanks @mzamoramora-nvidia), but Newton's pinned `MENAGERIE_REF = "feadf76d..."` ([download_assets.py:26](https://github.com/newton-physics/newton/blob/main/newton/_src/utils/download_assets.py#L26)) is from 2026-03-18 and predates the fix. Bumping the pin to a commit at or after `affef083` will let `TestMenagerieUSD_Robotiq2f85V4.dynamics_tolerance` drop from `1e-2` to something much tighter.

## WonikAllegro

Multiple body mass/inertia values in the USD diverge from the source MJCF (not just finger pads). I did not find an existing upstream menagerie PR addressing this; needs a closer look at the USD authoring pipeline to determine whether it's a regenerable USD asset issue or a deeper menagerie source issue.

## Concrete repro

`TestMenagerieUSD_Robotiq2f85V4` and `TestMenagerieUSD_WonikAllegro` were enabled in PR #2886 with body-inertia backfill (`body_mass`, `body_inertia`, `body_iquat`, `body_ipos`, `body_subtreemass`, `body_invweight0`) plus relaxed dynamics tolerances. See [test_menagerie_usd_mujoco.py:1757-1866](https://github.com/newton-physics/newton/blob/main/newton/tests/test_menagerie_usd_mujoco.py#L1757-L1866).

## Proposed fix

1. Bump `MENAGERIE_REF` to a commit at or after `affef083` to pick up the Robotiq pad mass fix; tighten `TestMenagerieUSD_Robotiq2f85V4` tolerance afterwards.
2. Investigate WonikAllegro's USD authoring path; file an upstream menagerie issue if the divergence is genuinely in the asset.

## Follow-up to

PR #2886, Alain Denzler's review comment requesting this be tracked as a follow-up to #2420.

devshahofficial:
## Description

Advance `MENAGERIE_REF` to `affef0836947b64cc06c4ab1cbf0152835693374`, the exact upstream merge that fixes Robotiq 2F-85 finger-pad masses. Tighten its dynamics tolerance from `1e-2` to `1e-4` and add a direct source-versus-USD mass-scale regression.

I also regenerated Wonik Allegro with mujoco-usd-converter v0.3.0. The mismatch remains because source MJCF mass is inferred from non-convex visual meshes while the USD re-derives it from convex-hull collision geometry and does not carry compiled body mass properties. That converter gap is tracked in [mujoco-usd-converter#99](https://github.com/newton-physics/mujoco-usd-converter/issues/99).

Closes #2931.

## Checklist

- [x] `CHANGELOG.md` has been updated (not user-facing)

## Test plan

```
uv run --extra dev -m newton.tests -k test_menagerie_usd_mujoco.TestMenagerieUSD_Robotiq2f85V4
uv run --extra dev -m newton.tests -k test_menagerie_usd_mujoco.TestMenagerieUSD_WonikAllegro
uvx pre-commit run -a
```

Robotiq passes 5 focused tests and Wonik passes 4 focused tests on CPU.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated the Mujoco Menagerie asset revision to ensure downloads use the latest verified assets.
  - Improved simulation consistency checks for Robotiq finger-pad mass values.
  - Tightened dynamics validation for the Robotiq asset to detect smaller discrepancies.

- **Tests**
  - Added coverage verifying that USD and native model finger-pad masses match.
  - Clarified documentation around known Allegro asset conversion differences.

### DIFF
--- a/newton/tests/test_menagerie_usd_mujoco.py
+++ b/newton/tests/test_menagerie_usd_mujoco.py
@@ -1792,8 +1792,8 @@ class TestMenagerieUSD_Robotiq2f85V4(TestMenagerieUSD):
     def test_pad_mass_matches_source_scale(self):
         """The pinned MJCF and USD agree on the finger-pad mass scale."""
         self._ensure_models()
-        newton_mass = self._newton_solver.mjw_model.body_mass.numpy()[0]
-        native_mass = self._native_mjw_model.body_mass.numpy()[0]
+        newton_mass = self._newton_solver.mj_model.body_mass
+        native_mass = self._mj_model.body_mass
 
         for body_name in ("left_pad", "right_pad"):
             native_id = self._mj_model.body(body_name).id
--- a/newton/tests/test_menagerie_usd_mujoco.py
+++ b/newton/tests/test_menagerie_usd_mujoco.py
@@ -1779,18 +1779,32 @@ class TestMenagerieUSD_Robotiq2f85V4(TestMenagerieUSD):
 
     num_steps = 20
     fk_enabled = True
-    # USD asset has body_mass = 0.0033 kg for the gripper finger pads
-    # (`left_pad` / `right_pad`); the source MJCF has near-zero mass 2e-6 kg.
-    # The mass mismatch produces qvel diffs up to ~4e-3 on the gripper DOF in
-    # the first few steps before settling. Other tests (model comparison,
-    # FK) are unaffected once `_compare_inertia` is overridden to skip the
-    # body_mass check. To tighten: regenerate the USD asset from the current
-    # MJCF.
-    dynamics_tolerance = 1e-2
+    # Menagerie PR #252 corrected the source finger pads from 2e-6 kg to
+    # 0.0035 kg. The USD reconstructs them as 0.0033 kg, so exact inertia
+    # comparison remains inappropriate, but the dynamics residual is now two
+    # orders of magnitude below the tolerance required by the stale source.
+    dynamics_tolerance = 1e-4
 
     def _compare_inertia(self, newton_mjw: Any, native_mjw: Any) -> None:
         # body_mass differs for finger pads (see class docstring).
         pass
+
+    def test_pad_mass_matches_source_scale(self):
+        """The pinned MJCF and USD agree on the finger-pad mass scale."""
+        self._ensure_models()
+        newton_mass = self._newton_solver.mjw_model.body_mass.numpy()[0]
+        native_mass = self._native_mjw_model.body_mass.numpy()[0]
+
+        for body_name in ("left_pad", "right_pad"):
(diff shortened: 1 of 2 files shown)


newton/tests/test_menagerie_usd_mujoco.py
```diff
@@ -1779,19 +1779,33 @@ class TestMenagerieUSD_Robotiq2f85V4(TestMenagerieUSD):
...
 
     def _compare_inertia(self, newton_mjw: Any, native_mjw: Any) -> None:
         # body_mass differs for finger pads (see class docstring).
         pass
 
+    def test_pad_mass_matches_source_scale(self):
+        """The pinned MJCF and USD agree on the finger-pad mass scale."""
+        self._ensure_models()
+        newton_mass = self._newton_solver.mjw_model.body_mass.numpy()[0]
+        native_mass = self._native_mjw_model.body_mass.numpy()[0]
```
vreutskyy: This test class shares its models across test methods. Since `test_dynamics` runs first and calls `_backfill_and_recompute()`, `mjw_model.body_mass` may already contain the native masses here, making this assertion order-dependent.

Could we compare the original compiled models instead? `_body_map` maps between these models, and they remain unchanged by the backfill.

```suggestion
        newton_mass = self._newton_solver.mj_model.body_mass
        native_mass = self._mj_model.body_mass
```

vreutskyy (commented):
One test-order issue in the new mass-scale regression.

newton/tests/test_menagerie_usd_mujoco.py
devshahofficial: @vreutskyy Good catch — fixed in b3043054 by comparing the original compiled models, so the assertion no longer depends on the shared Warp model's backfill state. I ran the isolated regression and the full Robotiq class (5 tests, with `test_dynamics` first); both pass, along with the full pre-commit suite.

vreutskyy (approved):
Re-reviewed the latest revision. The test-order concern is addressed, and the focused Robotiq suite passes.

PR #3556, opened 2026-07-16
Respect USD schema default applicability

### LINKED ISSUE #3307: Redesign USD schema resolver precedence and provenance
## Problem

`SchemaResolverManager` does not model all value sources needed by USD import consistently. The documented fallback order is authored value → importer default → resolver mapping default, but a mapping default is only valid when its schema applies to the prim. Today the resolver has no general applicability contract, and existing SDF, hydroelastic, and MuJoCo import paths need local exceptions.

This also makes provenance difficult to preserve: callers sometimes need to know whether a value came from authored data, an applied-schema default, or an importer fallback. Custom getters add another ambiguity because a getter can find authored inputs but normalize them to `None`.

## Desired behavior

- Distinguish authored values, applied-schema defaults, and importer defaults.
- Preserve the winning value's resolver and source provenance.
- Apply mapping defaults only when the corresponding schema applies to the prim.
- Define consistent priority behavior across multiple configured resolvers.
- Preserve ordinary fallthrough when a transformed or getter-derived candidate is unusable.
- Avoid parallel generic resolution paths with subtly different semantics.

## Scope

- Define a resolver applicability contract for each prim/type/key.
- Define a single resolution result that can carry value and provenance.
- Establish and document one global precedence model.
- Audit and migrate SDF, hydroelastic, MuJoCo joint-limit, and other exception paths.
- Add cross-resolver tests for authored, applied-default, importer-default, and unusable-candidate cases.

## Context

Follow-up requested during review of #3286. That PR should keep its MuJoCo joint-limit workaround local to `import_usd.py` until this broader model is designed.

devshahofficial:
## Description

Give USD resolution a single result carrying the value, winning resolver, and source: authored, importer default, applicable schema default, or unresolved.

Built-in resolvers now declare applicability per prim type and, where necessary, per key. Mapping defaults therefore cannot leak from a configured solver schema onto unrelated prims. MuJoCo joint-limit semantics plus Newton SDF and hydroelastic defaults use the shared provenance-aware path.

Third-party resolver subclasses that do not declare applicability keep their legacy always-applicable behavior.

Closes #3307.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_schema_resolver
uv run --extra dev -m newton.tests -k test_sdf_usd
uvx pre-commit run -a
```

The resolver suite passes 34 tests. The SDF suite passes 10 CPU-capable tests and skips 14 CUDA-only tests on this machine.

## Bug fix

**Steps to reproduce:**

1. Configure Newton, PhysX, and MuJoCo schema resolvers.
2. Import a prim that does not apply the first resolver schema.
3. Observe the first configured resolver injecting its hardcoded mapping default on main.

**Minimal reproduction:**

```python
import newton
from newton.usd import SchemaResolverNewton

builder.add_usd(stage, schema_resolvers=[SchemaResolverNewton()])
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected USD schema defaults so they apply only when the corresponding API/schema is applicable to the prim.
  - Improved USD importer resolution for joint limits/gains, SDF targets, and hydroelastic parameters, with corrected precedence and mutual exclusivity handling.
  - Preserved resolution provenance so importer behavior follows the correct authored, caller default, or schema-default source.
  - Refined handling for unresolved and invalid attribute values (including hydroelastic `enabled`/`kh` defaults).

- **Documentation**
  - Updated USD “Resolution Hierarchy” guidance to match the new applicable-schema-default behavior and provenance outcomes.

- **Tests**
  - Expanded schema-resolver and provenance coverage for applicability-gated defaults and layered fallbacks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Fix MJCF, URDF, and USD imports rendering collision-only bodies as visuals when the asset authors visual geometry elsewhere. (#3291)
 - Fix `ViewerUSD` texture consumers observing partially written PNGs by publishing generated textures atomically (#3288)
 - Fix builder merging (`ModelBuilder.add_builder()`, `add_world()`, `replicate()`) offsetting negative reference sentinels in custom attribute values stored as NumPy or Warp integer scalars.
+- Fix USD schema mapping defaults applying to prims that do not carry the corresponding solver schema, and preserve resolver/source provenance for importer decisions.
 - Fix `ModelBuilder.add_usd()` requiring the optional `mujoco` package when handling `MjcActuator` prims, including during default MJC equality conversion.
 - Report malformed MJCF free-joint and inertial inputs with deterministic validation errors, and ignore MJCF mesh geom `size` lengths consistently.
 - Fix Style3D solver divergence caused by isolated vertices.
--- a/docs/concepts/usd_parsing.rst
+++ b/docs/concepts/usd_parsing.rst
@@ -470,7 +470,11 @@ The attribute resolution process follows a three-layer fallback hierarchy to det
 
 1. **Authored Values**: Resolvers are queried in priority order; the first resolver that finds an authored value on the prim returns it and remaining resolvers are not consulted.
 2. **Importer Defaults**: If no authored value is found, Newton's importer uses a property-specific fallback (e.g. ``builder.default_joint_cfg.armature`` for joint armature). This takes precedence over schema-level defaults.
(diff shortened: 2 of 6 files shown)


devshahofficial:
@coderabbitai Addressed both valid nits in d26c1c2e: applied-schema/type data is cached per manager with metadata-signature invalidation, and the SDF sentinel comment now distinguishes sentinel-backed fields from concrete defaults. I kept `_is_applicable` protected intentionally because `SchemaResolver` is public and exposing that helper would expand the API before the resolver design is approved. Validation: 34 resolver tests and 24 SDF tests pass (14 CUDA-only skips), plus the full pre-commit suite.

adenzler-nvidia:
closing this in favor of #3572. Please excuse the late notice, but I have been working on the same thing for an internal project for a while, and my PR is is part of a series trying to get the pxr- dependency out of the schemaResolvers.

PR #3564, opened 2026-07-17
Reduce Kamino's kernel compile time

nvtw:
## Description

Extend the kernel decorators of Kamino's kernels to disable grid stride and backward pass generation

## Checklist

## Test plan

     or manual steps. Example:
     ```
     ``` -->



## Summary by CodeRabbit

## Summary of Changes
* **Performance**
  * Improved GPU simulation and kernel compilation performance for SolverKamino.
* **Behavior Updates**
  * Standardized GPU kernel execution settings across core solver modules, examples, and tests to disable backward/gradient work and to turn off grid-stride scheduling by default.
  * Updated selected kernels to rely on these module-wide execution defaults rather than per-kernel settings.
* **Documentation**
  * Added an unreleased changelog entry describing the SolverKamino performance improvement.


CHANGELOG.md
```diff
@@ -10,6 +10,17 @@
...
+- Rework the API of the `reset()` operation in `SolverKamino` to use an explicit `ResetConfig` instead of many keyword arguments.
+- Non-schema per-DOF initial state attributes (`newton:{axis}:position`, `newton:{axis}:velocity`) now emit a `UserWarning` when encountered during USD import, directing users to file a feature request.
+- Change VBD Neo-Hookean membrane/tet damping to an objective metric based on the rate of `C = FᵀF`, so rigid-body rotations no longer generate damping force.
+- Change VBD spring damping to act only along the spring axis (damping edge-length rate), so transverse and rigid-rotational motion is no longer damped by springs.
+- Apply each shape's `ShapeConfig.margin` (`model.shape_margin`) to `SolverVBD` particle-rigid (soft) contacts, widening the soft-contact detection shell and reducing penetration depth per shape; previously only the global `soft_contact_margin` and particle radius were used. Re-check VBD scenes that set per-shape margins. (#2994)
```
adenzler-nvidia, line 15: some entries came along for the ride here :)

adenzler-nvidia (dismissed):
Thanks. Longer-term, we probably need better ways to configure this at least for the differentiability paths because users might want to opt-in. But I assume backwards pass doesn't really work right now anyway..

newton/_src/solvers/kamino/_src/core/bodies.py
```diff
@@ -372,7 +372,7 @@ def transform_body_inertial_properties(
 ###
 
 
-@wp.kernel
+@wp.kernel(grid_stride=False, enable_backward=False)
```
Guirec-Maloisel: Two related questions: is this not redundant with `wp.set_module_options({"enable_backward": False})` at the top of the file (in the case of `enable_backward`), and could we move the `grid_stride=False` also to the module options to avoid having it on every kernel?

nvtw: That's probably possible and indeed, I missed wp.set_module_options({"enable_backward": False}). My preference is usually to have it on the kernel but I can switch to module options

Guirec-Maloisel (dismissed):
Looks good to me! We can revisit this when we start looking into differentiability, but right now we don't anywhere in the code, so I think it's easier to disable it at the module level.

camevor (approved):
Re-approving after fixes

chschuma-disney (approved):

PR #3566, opened 2026-07-17
Surface additional simulation benchmark metrics

### LINKED ISSUE #3343: Surface additional metrics from existing benchmarks
## Description

Surface additional robot-learning-relevant physics metrics from the existing ASV simulation benchmarks. The current results emphasize aggregate elapsed time and do not directly expose simulated experience-collection rate, long-tail step behavior, or whether a performance result remained physically valid.

## Motivation / Use Case

Robot-learning workloads benefit both from executing more simulation steps per second and from advancing more simulated time per wall-clock second. Intermittent slow steps and invalid simulation states must also be visible so that aggregate runtime alone does not hide regressions or misleading speedups.

## Acceptance Criteria

- Applicable existing simulation benchmarks report simulation steps per second.
- Applicable existing simulation benchmarks report real-time factor as simulated physical time advanced per wall-clock time, such that both a larger simulation time step and a higher step rate improve the result.
- At least one high-percentile (e.g., p95) step-time metric is reported to surface tail latency in a number or representative benchmarks
- Applicable GPU simulation benchmarks report steady-state GPU memory usage after initialization and warm-up, with the measurement point and method documented.
- The additional values, units, and relevant parameters are stored in ASV and visible in the published dashboard.
- Existing aggregate benchmark results remain available and comparable.
- Adding new workloads, scenes, solvers, or startup-phase benchmarks is out of scope.

vreutskyy:
## Description

Closes #3343.

Surface robot-learning-relevant metrics from the existing ASV simulation workloads:

- simulation world-steps per second and real-time factor using synchronized complete-step timing;
- p95 complete-step latency;
- steady-state GPU memory using the device free-memory delta after warm-up;
- simulation timestep and substep configuration; and
- final-state validation for finite coordinates and velocities, normalized rotations, workload-specific health checks, and bounded body speeds.

The existing KPI mean world-step series remains physics-only and keeps its existing name and unit. Cached ASV setup runs provide all related metrics without rerunning each workload for every tracked value. The development guide documents formulas, timing scope, memory measurement, and the exclusive-GPU requirement.

## Checklist

## Test plan

- `uv --cache-dir C:\work\newton\main\.uv-cache run --frozen --extra dev -m newton.tests -k benchmark_metrics`
- `uvx --cache-dir C:\work\newton\main\.uv-cache --with virtualenv asv check`
- `uvx --cache-dir C:\work\newton\main\.uv-cache pre-commit run -a`
- Ran `FastCartpole` and `FastMetricsExampleQuadrupedXPBD` directly on an RTX 3080 Laptop GPU.
- Ran `simulation.bench_mujoco.FastCartpole.track_real_time_factor` through ASV's real setup-cache subprocess path.

The full 4,096-world `KpiDRLegs` policy setup exceeded the existing 600-second timeout on the RTX 3080 Laptop GPU; its full result remains to be verified on the nightly benchmark hardware.

## Summary by CodeRabbit

- **New Features**
  - Expanded standardized simulation benchmark metrics (adds mean/max MuJoCo solver-iteration reporting).
  - Added synchronized metric-collection utilities and new “fast metrics” benchmark variants for quicker precomputed metric tracks.
  - Introduced optional cache lifecycle support during benchmark runs.
- **Bug Fixes**
  - Fixed FastKitchenG1 KPI benchmark metrics by building the full kitchen environment for accurate results.
- **Documentation**
  - Documented the published metrics, formulas, and benchmark failure conditions.
- **Tests**
  - Added coverage for metric aggregation, validation rules, GPU memory checks, CUDA-graph behavior, and cache-aware execution.

### DIFF
--- a/.github/workflows/aws_gpu_benchmarks.yml
+++ b/.github/workflows/aws_gpu_benchmarks.yml
@@ -137,12 +137,13 @@ jobs:
 
       - name: Run Benchmarks
         run: |
+          # ASV assumes smaller values are better, so compare only metrics with that direction.
           uvx --with virtualenv asv continuous \
             --launch-method spawn \
             --interleave-rounds \
             --append-samples \
             --no-only-changed \
-            -e -b Fast \
+            -e -b '^(?!.*track_(simulation_steps_per_second|real_time_factor)).*Fast' \
             ${{ inputs.base_ref }} \
             ${{ inputs.ref }}
         continue-on-error: true
--- a/asv/benchmarks/benchmark_metrics.py
+++ b/asv/benchmarks/benchmark_metrics.py
@@ -23,6 +23,8 @@ class SimulationMetrics:
     gpu_memory_mib: float
     sim_dt: float
     sim_substeps: int
+    solver_niter_mean: float | None = None
+    solver_niter_max: float | None = None
 
 
 class _SimulationMetricTracks:
@@ -117,17 +119,6 @@ class _SimulationMetricTracksUnparameterized:
     track_sim_substeps.unit = "simulation-steps/frame"
 
 
-def _percentile(values: Sequence[float], percentile: float) -> float:
-    ordered = sorted(values)
-    rank = (len(ordered) - 1) * percentile / 100.0
-    lower = math.floor(rank)
-    upper = math.ceil(rank)
-    if lower == upper:
-        return ordered[lower]
-    fraction = rank - lower
-    return ordered[lower] + fraction * (ordered[upper] - ordered[lower])
-
-
 def compute_simulation_metrics(
     frame_times: Sequence[float],
     sim_dt: float,
@@ -159,19 +150,11 @@ def compute_simulation_metrics(
         mean_world_step_time_ms=total_time * 1000.0 / world_steps,
         world_steps_per_second=world_steps / experience_total_time,
         real_time_factor=world_steps * sim_dt / experience_total_time,
-        p95_frame_time_ms=_percentile(experience_frame_times, 95.0) * 1000.0,
+        p95_frame_time_ms=float(np.percentile(experience_frame_times, 95.0)) * 1000.0,
         gpu_memory_mib=gpu_memory_bytes / 1024**2,
         sim_dt=sim_dt,
         sim_substeps=sim_substeps,
     )
-
-
-def compute_gpu_memory_usage(device: Any, free_memory_before: int) -> int:
-    """Compute the change in total device memory usage from a free-memory baseline."""
-    used_memory = free_memory_before - device.free_memory
-    if used_memory < 0:
-        raise RuntimeError("GPU free memory increased after workload initialization")
-    return used_memory
 
 
 def validate_simulation_state(
@@ -208,31 +191,25 @@ def validate_simulation_state(
         )
 
 
-def validate_simulation_workload(workload: Any, max_linear_speed: float, max_angular_speed: float):
-    """Run generic state validation followed by a workload's specialized checks."""
-    validate_simulation_state(
-        workload.state_0,
-        max_linear_speed=max_linear_speed,
-        max_angular_speed=max_angular_speed,
-    )
-    workload.test_final()
-
-
 def collect_simulation_metrics(
     create_workload: Callable[[], Any],
     world_count: int,
     num_frames: int,
     samples: int,
-    memory_usage_bytes: Callable[[Any], int],
+    synchronize: Callable[[], None] | None = None,
     validate: Callable[[Any], None] | None = None,
     timer: Callable[[], float] = time.perf_counter,
 ) -> SimulationMetrics:
-    """Collect internal physics timing and complete synchronized step timing."""
+    """Collect simulation metrics using internal or synchronized wall timing."""
     frame_times = []
     experience_frame_times = []
     gpu_memory_bytes = None
     sim_dt = None
     sim_substeps = None
+
+    wp.synchronize_device()
+    device = wp.get_device()
+    free_memory_before = device.free_memory
 
     for sample_index in range(samples):
         workload = create_workload()
@@ -242,15 +219,27 @@ def collect_simulation_metrics(
         elif workload.sim_dt != sim_dt or workload.sim_substeps != sim_substeps:
             raise ValueError("simulation parameters changed between samples")
 
+        if synchronize is not None:
+            synchronize()
         for _ in range(num_frames):
-            start_time = workload.benchmark_time
             experience_start_time = timer()
+            benchmark_start_time = workload.benchmark_time if synchronize is None else None
             workload.step()
-            frame_times.append(workload.benchmark_time - start_time)
-            experience_frame_times.append(timer() - experience_start_time)
+            if synchronize is not None:
+                synchronize()
+            experience_frame_time = timer() - experience_start_time
+            experience_frame_times.append(experience_frame_time)
+            frame_times.append(
+                experience_frame_time
+                if benchmark_start_time is None
+                else workload.benchmark_time - benchmark_start_time
+            )
 
         if sample_index == 0:
-            gpu_memory_bytes = memory_usage_bytes(workload)
+            wp.synchronize_device()
+            gpu_memory_bytes = free_memory_before - device.free_memory
+            if gpu_memory_bytes < 0:
+                raise RuntimeError("GPU free memory increased after workload initialization")
         if validate is not None:
             validate(workload)
 
@@ -261,49 +250,4 @@ def collect_simulation_metrics(
         world_count=world_count,
         gpu_memory_bytes=gpu_memory_bytes,
         experience_frame_times=experience_frame_times,
-    )
-
-
-def collect_simulation_metrics_synchronized(
-    create_workload: Callable[[], Any],
-    world_count: int,
-    num_frames: int,
-    samples: int,
-    synchronize: Callable[[], None],
-    memory_usage_bytes: Callable[[Any], int],
-    validate: Callable[[Any], None] | None = None,
-    timer: Callable[[], float] = time.perf_counter,
-) -> SimulationMetrics:
-    """Collect metrics for workloads without an internal synchronized timer."""
-    frame_times = []
-    gpu_memory_bytes = None
-    sim_dt = None
-    sim_substeps = None
-
-    for sample_index in range(samples):
-        workload = create_workload()
-        if sim_dt is None:
-            sim_dt = workload.sim_dt
-            sim_substeps = workload.sim_substeps
-        elif workload.sim_dt != sim_dt or workload.sim_substeps != sim_substeps:
-            raise ValueError("simulation parameters changed between samples")
-
-        synchronize()
-        for _ in range(num_frames):
-            start_time = timer()
-            workload.step()
-            synchronize()
-            frame_times.append(timer() - start_time)
-
-        if sample_index == 0:
-            gpu_memory_bytes = memory_usage_bytes(workload)
-        if validate is not None:
-            validate(workload)
-
-    return compute_simulation_metrics(
-        frame_times=frame_times,
-        sim_dt=sim_dt,
-        sim_substeps=sim_substeps,
-        world_count=world_count,
-        gpu_memory_bytes=gpu_memory_bytes,
     )
--- a/asv/benchmarks/benchmark_metrics.py
+++ b/asv/benchmarks/benchmark_metrics.py
@@ -71,7 +71,7 @@ class _SimulationMetricTracks:
     track_sim_substeps.unit = "simulation-steps/frame"
 
 
-class _UnparameterizedSimulationMetricTracks:
+class _SimulationMetricTracksUnparameterized:
     """ASV track methods backed by one cached simulation configuration."""
 
     @skip_benchmark_if(wp.get_cuda_device_count() == 0)
@@ -264,7 +264,7 @@ def collect_simulation_metrics(
     )
 
 
-def collect_synchronized_simulation_metrics(
+def collect_simulation_metrics_synchronized(
     create_workload: Callable[[], Any],
     world_count: int,
     num_frames: int,
--- a/asv/benchmarks/benchmark_metrics.py
+++ b/asv/benchmarks/benchmark_metrics.py
@@ -199,8 +199,7 @@ def validate_simulation_state(
     max_measured_angular_speed = np.linalg.norm(body_qd[:, 3:], axis=-1).max()
     if max_measured_linear_speed > max_linear_speed:
         raise RuntimeError(
-            f"Maximum body linear speed is {max_measured_linear_speed:.3f} m/s, "
-            f"exceeding {max_linear_speed:.1f} m/s"
+            f"Maximum body linear speed is {max_measured_linear_speed:.3f} m/s, exceeding {max_linear_speed:.1f} m/s"
         )
     if max_measured_angular_speed > max_angular_speed:
         raise RuntimeError(
--- a/asv/benchmarks/benchmark_metrics.py
+++ b/asv/benchmarks/benchmark_metrics.py
@@ -1,0 +1,310 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
+
+import math
+import time
+from collections.abc import Callable, Sequence
+from dataclasses import dataclass
+from typing import Any
+
+import numpy as np
+import warp as wp
+from asv_runner.benchmarks.mark import skip_benchmark_if
+
+
+@dataclass(frozen=True)
+class SimulationMetrics:
+    """Metrics collected from one simulation benchmark configuration."""
+
+    mean_world_step_time_ms: float
+    world_steps_per_second: float
+    real_time_factor: float
+    p95_frame_time_ms: float
+    gpu_memory_mib: float
+    sim_dt: float
+    sim_substeps: int
+
+
+class _SimulationMetricTracks:
+    """ASV track methods backed by cached simulation metrics."""
+
+    @skip_benchmark_if(wp.get_cuda_device_count() == 0)
+    def track_simulate(self, metrics, world_count):
+        return metrics[world_count].mean_world_step_time_ms
+
+    track_simulate.unit = "ms/world-step"
+
+    @skip_benchmark_if(wp.get_cuda_device_count() == 0)
+    def track_simulation_steps_per_second(self, metrics, world_count):
+        return metrics[world_count].world_steps_per_second
+
+    track_simulation_steps_per_second.unit = "world-steps/s"
+
+    @skip_benchmark_if(wp.get_cuda_device_count() == 0)
+    def track_real_time_factor(self, metrics, world_count):
+        return metrics[world_count].real_time_factor
+
+    track_real_time_factor.unit = "x"
+
+    @skip_benchmark_if(wp.get_cuda_device_count() == 0)
+    def track_p95_step_time(self, metrics, world_count):
+        return metrics[world_count].p95_frame_time_ms
+
+    track_p95_step_time.unit = "ms/frame"
+
+    @skip_benchmark_if(wp.get_cuda_device_count() == 0)
+    def track_steady_state_gpu_memory(self, metrics, world_count):
+        return metrics[world_count].gpu_memory_mib
+
+    track_steady_state_gpu_memory.unit = "MiB"
+
+    @skip_benchmark_if(wp.get_cuda_device_count() == 0)
+    def track_sim_dt(self, metrics, world_count):
+        return metrics[world_count].sim_dt
+
+    track_sim_dt.unit = "s"
+
+    @skip_benchmark_if(wp.get_cuda_device_count() == 0)
+    def track_sim_substeps(self, metrics, world_count):
+        return metrics[world_count].sim_substeps
+
+    track_sim_substeps.unit = "simulation-steps/frame"
+
+
+class _UnparameterizedSimulationMetricTracks:
+    """ASV track methods backed by one cached simulation configuration."""
+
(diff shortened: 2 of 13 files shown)


asv/benchmarks/simulation/bench_anymal.py
```diff
@@ -31,13 +48,38 @@ def time_simulate(self):
...
+            world_count=self.world_count,
+            num_frames=self.num_frames,
+            samples=self.samples,
+            synchronize=wp.synchronize_device,
+            memory_usage_bytes=lambda workload: compute_gpu_memory_usage(device, free_memory_before),
+            validate=lambda workload: validate_simulation_state(
+                workload.state_0,
+                max_linear_speed=10.0,
+                max_angular_speed=50.0,
+            ),
```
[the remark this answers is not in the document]
vreutskyy: Thanks for flagging this. I verified the workload-specific check against this benchmark's horizon. `Example.test_final()` is calibrated for the full example and requires forward position `q[1] > 9.0`; a healthy 50-frame benchmark run reaches about `0.719 m`, so calling `validate_simulation_workload()` here would reject a valid short-horizon run. The benchmark therefore keeps `validate_simulation_state()` for the generic finite-state, quaternion, and speed checks, and `_validate_workload()` adds calibrated 50-frame base-height and forward-progress checks. Those checks are covered by `test_anymal_short_horizon_validation`, so I'm leaving this suggestion unapplied.

newton/tests/test_benchmark_simulation.py
```diff
@@ -0,0 +1,81 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
+
+import sys
+import unittest
+from pathlib import Path
+from types import SimpleNamespace
+from unittest.mock import patch, sentinel
+
+import numpy as np
```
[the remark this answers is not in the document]
vreutskyy: Thanks. I verified this against the dependency graph and the test's requirements. `warp-lang` already declares NumPy as a dependency, and Newton's existing tests import NumPy directly without making it a separate project dependency. Adding a new required dependency solely for this test would conflict with the project's guidance to avoid new dependencies. Plain lists are not equivalent here because `validate_simulation_state()` consumes the fake arrays through `.numpy()` and relies on NumPy operations including finite checks, reshape, slicing, and vector norms. I'm keeping the small NumPy-backed test fake.

adenzler-nvidia:
Question on the `FastKitchenG1` world-count drop (512 → 128): the motivation is visible in CI — in the run at 126dead06, `setup_cache` for `FastKitchenG1(512)` died with exit status -9, i.e. OOM-killed while constructing the kitchen at 512 worlds on the 64 GiB runner. Two things I'd like to understand before this lands:

1. Since the pre-v2 series never actually built the kitchen, this is effectively the first time the kitchen KPI's world count is being defined for real. Was 256 (or 384) tried? 128 halves twice from the intended KPI point, which changes GPU saturation, and "leave headroom" doesn't say how much headroom is needed. It would be good to record why 128 is the right operating point rather than the largest count that fits.

2. Needing more than 64 GiB of host RAM to construct 512 kitchen replications — even with batched replication from #3415 in the base — looks like a `ModelBuilder` memory problem worth tracking on its own. Should we file a follow-up issue so the KPI can move back up once construction is leaner, rather than baking the runner's RAM limit into the benchmark definition?

adenzler-nvidia:
Request on the metric set: please track the measured solver iteration counts for the MuJoCo workloads — record `Data.solver_niter` as tracked series (e.g. mean and max over worlds, sampled at the final frame where the state validation already runs). Iteration counts are a valuable signal for checking whether something in the simulation is broken: a solver grinding at its iteration cap indicates unhealthy physics even while the states still pass the finite/speed checks.

adenzler-nvidia (commented):
A few simplification suggestions on the metrics infrastructure, details inline. The common theme: the metric set and helpers can shrink by roughly a third without losing any information. (Related discussion in the two earlier comments: FastKitchenG1 world count and iteration-count tracking.)

asv/benchmarks/benchmark_metrics.py
```diff
@@ -0,0 +1,309 @@
...
+    """ASV track methods backed by cached simulation metrics."""
+
+    @skip_benchmark_if(wp.get_cuda_device_count() == 0)
+    def track_simulate(self, metrics, world_count):
+        return metrics[world_count].mean_world_step_time_ms
+
+    track_simulate.unit = "ms/world-step"
+
+    @skip_benchmark_if(wp.get_cuda_device_count() == 0)
+    def track_simulation_steps_per_second(self, metrics, world_count):
```
adenzler-nvidia, line 40: Suggest dropping `track_simulation_steps_per_second` and `track_real_time_factor` from both mixins. Three reasons. First, they carry no information: both are linear transforms of mean world-step time given `sim_dt`, `sim_substeps`, and `world_count`, which are all published — they can be derived at the dashboard/reporting layer. Second, and more importantly, ASV has no per-benchmark direction attribute and universally assumes smaller-is-better (`asv compare`'s `_is_result_better(a, b)` is `a < b / factor`). These two series are higher-is-better, so regression handling inverts: `asv continuous -b Fast` in PR CI will fail the job on any ≥10% *improvement* in simulation throughput, and a real slowdown renders as a green "improvement" in these rows. The dashboard's step detection has the same inversion. Third, each removed track method removes one spawned subprocess per class per commit.

asv/benchmarks/benchmark_metrics.py
```diff
@@ -0,0 +1,309 @@
...
+    """Run generic state validation followed by a workload's specialized checks."""
+    validate_simulation_state(
+        workload.state_0,
+        max_linear_speed=max_linear_speed,
+        max_angular_speed=max_angular_speed,
+    )
+    workload.test_final()
+
+
+def collect_simulation_metrics(
```
adenzler-nvidia, line 194: The GPU-memory baseline could live inside the collector. Every call site repeats the same preamble (`wp.synchronize_device(); device = wp.get_device(); free_memory_before = device.free_memory`) and then passes a `memory_usage_bytes` lambda that ignores its `workload` argument and closes over those locals. If the collector captures the baseline itself immediately before the first `create_workload()` call and computes the delta after sample 0, the `memory_usage_bytes` parameter, `compute_gpu_memory_usage`, and the preamble at all five call sites can be deleted (~25 lines). This is equivalent for every current user, since all GPU allocation happens inside `create_workload()`.

asv/benchmarks/benchmark_metrics.py
```diff
@@ -0,0 +1,309 @@
...
+        frame_times=frame_times,
+        sim_dt=sim_dt,
+        sim_substeps=sim_substeps,
+        world_count=world_count,
+        gpu_memory_bytes=gpu_memory_bytes,
+        experience_frame_times=experience_frame_times,
+    )
+
+
+def collect_simulation_metrics_synchronized(
```
adenzler-nvidia: This function and `collect_simulation_metrics` are identical except for how frame time is obtained. Consider merging them: one collector with `synchronize: Callable[[], None] | None = None` that uses the workload's internal `benchmark_time` when the workload provides it, and sync-plus-wall-timing otherwise. Halves the collector code and removes a which-one-do-I-call decision for future workloads.

asv/benchmarks/benchmark_metrics.py
```diff
@@ -0,0 +1,309 @@
...
+        raise ValueError("gpu_memory_bytes must be non-negative")
+
+    total_time = sum(frame_times)
+    experience_total_time = sum(experience_frame_times)
+    world_steps = len(frame_times) * sim_substeps * world_count
+    return SimulationMetrics(
+        mean_world_step_time_ms=total_time * 1000.0 / world_steps,
+        world_steps_per_second=world_steps / experience_total_time,
+        real_time_factor=world_steps * sim_dt / experience_total_time,
+        p95_frame_time_ms=_percentile(experience_frame_times, 95.0) * 1000.0,
```
adenzler-nvidia: `_percentile` reimplements `np.percentile`'s default linear interpolation, and numpy is a hard dependency — the helper can be deleted.
```suggestion
        p95_frame_time_ms=float(np.percentile(experience_frame_times, 95)) * 1000.0,
```

asv/benchmarks/benchmark_metrics.py
```diff
@@ -0,0 +1,309 @@
...
+            f"Maximum body linear speed is {max_measured_linear_speed:.3f} m/s, exceeding {max_linear_speed:.1f} m/s"
+        )
+    if max_measured_angular_speed > max_angular_speed:
+        raise RuntimeError(
+            f"Maximum body angular speed is {max_measured_angular_speed:.3f} rad/s, "
+            f"exceeding {max_angular_speed:.1f} rad/s"
+        )
+
+
+def validate_simulation_workload(workload: Any, max_linear_speed: float, max_angular_speed: float):
```
adenzler-nvidia: Single-use wrapper (only `bench_quadruped_xpbd`) — suggest inlining the two calls at the call site.

asv/benchmarks/benchmark_metrics.py
```diff
@@ -0,0 +1,309 @@
...
+    sim_dt = None
+    sim_substeps = None
+
+    for sample_index in range(samples):
+        workload = create_workload()
+        if sim_dt is None:
+            sim_dt = workload.sim_dt
+            sim_substeps = workload.sim_substeps
+        elif workload.sim_dt != sim_dt or workload.sim_substeps != sim_substeps:
+            raise ValueError("simulation parameters changed between samples")
```
adenzler-nvidia, line 220: Samples come from the same factory with the same builder, so `sim_dt`/`sim_substeps` can't change between samples. Reading them from the first workload and dropping the cross-sample check (in both collectors) removes a branch that can't fire.

asv/benchmarks/simulation/bench_mujoco.py
```diff
@@ -124,6 +136,9 @@ class FastCartpole(_KpiBenchmark):
     random_init = True
     environment = "None"
 
+    def setup_cache(self):
```
adenzler-nvidia, line 277: Regression vs the previous `track_simulate`-only setup: ASV's `setup_cache` has no skip mechanism — the runner creates the cache *before* the in-process `skip_benchmark_if` markers are consulted. On a machine without CUDA, each KPI class's `setup_cache` now runs the workload on CPU until the 600–900s timeout (or crashes), and all its track series are reported as *failed* instead of *skipped*. Previously the decorated `track_simulate` skipped cleanly without running anything. Suggest an early `return None` from `setup_cache` when `wp.get_cuda_device_count() == 0` — safe because every consumer track method is skipped before it touches the cache. Applies to all `setup_cache` classes in this PR (mujoco, kamino, anymal, quadruped).

asv/benchmarks/simulation/bench_mujoco.py
```diff
@@ -31,39 +36,46 @@ class _KpiBenchmark:
...
-            wp.synchronize_device()
-            for _ in range(self.num_frames):
-                example.step()
-            wp.synchronize_device()
-            total_time += example.benchmark_time
-
-        return total_time * 1000 / (self.num_frames * example.sim_substeps * world_count * self.samples)
-
-    track_simulate.unit = "ms/world-step"
+            def create_workload(builder=builder):
```
adenzler-nvidia: Regression in failure visibility: previously a missing graph with `use_cuda_graph=True` crashed in `wp.capture_launch(None)`; with the new `and self.graph is not None` fallback in `step()`, the KPI collection silently benchmarks the eager path and publishes numbers that measure something else — a graph-capture regression would show up as a mysterious step-time cliff instead of an error. The kamino path already guards this in `create_workload` ("KPI benchmark requires CUDA graph capture..."). Suggest the same assert here in `create_workload` (raise if `example.graph is None` when `use_cuda_graph` was requested), keeping the `step()` fallback itself for non-benchmark convenience.

adenzler-nvidia (commented):
Two behavioral regressions compared to the pre-PR benchmarks, details inline.

vreutskyy:
I calibrated this on the actual `g7e.2xlarge` runner.

- 128 worlds completes: ~86k world-steps/s, 14.3× real-time factor, and 1,120 MiB steady-state GPU memory.
- 256 worlds is also killed during `setup_cache` with exit status `-9`, before metrics are emitted.
- I did not attempt 384 after 256 already exceeded the 64 GiB host-memory envelope.

So 128 is the largest validated power-of-two configuration that completes on this runner, rather than a GPU-saturation optimum. The calibration run is [here](https://github.com/newton-physics/newton/actions/runs/29726644036/job/88301677747).

I also filed [#3573](https://github.com/newton-physics/newton/issues/3573) to track the `ModelBuilder` host-memory scaling problem and allow increasing the KPI world count once construction is leaner.

adenzler-nvidia:
On the new test files — split feedback:

`test_benchmark_metrics.py` is the right kind of coverage and worth keeping as-is. Benchmark bugs don't fail loudly, they publish plausible-looking wrong numbers indefinitely (the pre-v2 `FastKitchenG1` being the proof), so GPU-free contract tests on the metric formulas, the collector orchestration, and the validation branches are one of the few real guards this code can have. `test_run_benchmark_with_setup_cache` is the clearest keep of all since `newton.utils.run_benchmark` is shipped package code whose calling convention this PR changes.

`test_benchmark_simulation.py` I'd reconsider. `test_kpi_dr_legs_has_setup_cache_timeout` asserts the literal `1200` and the kitchen test asserts `version == "2"`, `128`, and `seed=123` through four layers of mocks — these fail on every legitimate retune while verifying implementation constants rather than behavior. `test_benchmark_imports_preserve_warp_config` mostly verifies the test file's own save/restore harness. The file also imports the ASV bench modules (with their `sys.path` manipulation and module-level `wp.config` mutation) into the unit suite on every CI configuration, which couples `newton.tests` to benchmark-suite import side effects.

The legitimate concern behind the kitchen test — does the workload actually build what the KPI claims? — would be better served by an assertion inside the benchmark's own validation (e.g., expected body/shape count of the built model in `_validate_workload`/`test_final`). That checks the real built model instead of a mock echo, runs where it matters, and would have caught the original bug too.

asv/benchmarks/benchmark_metrics.py
vreutskyy: Agreed on the ASV directionality problem. I kept these two tracks because #3343 explicitly requires publishing simulation steps/s and real-time factor, but changed the AWS PR comparison filter to exclude `track_simulation_steps_per_second` and `track_real_time_factor` from `asv continuous`. The smaller-is-better mean/p95 tracks remain the CI regression signals; the two higher-is-better values remain available for reporting and should not drive automatic regression direction until that layer supports it. I also verified the behavior against ASV 0.6.6's `_is_result_better` implementation.

asv/benchmarks/benchmark_metrics.py
vreutskyy: Done. The collector now captures the free-memory baseline immediately before the first `create_workload()` call and computes the delta after sample 0. The repeated call-site preambles, callback parameter, and standalone helper are gone.

asv/benchmarks/benchmark_metrics.py
vreutskyy: Done. Both timing paths now use `collect_simulation_metrics`, with optional `synchronize`: synchronized wall timing is used when supplied, otherwise the workload's internal `benchmark_time` is used.

asv/benchmarks/benchmark_metrics.py
vreutskyy: Done. The custom percentile helper was removed and p95 now uses `np.percentile`.

asv/benchmarks/benchmark_metrics.py
vreutskyy: Done. The single-use wrapper was removed and its generic-state plus workload-specific validation calls are now inlined in the quadruped callback.

asv/benchmarks/benchmark_metrics.py
vreutskyy: I kept this guard intentionally. The same factory should produce identical simulation parameters, but each workload still exposes them independently; if a future factory/configuration change makes samples disagree, combining them would publish a mislabeled aggregate. The check is one comparison per sample and keeps that aggregation contract explicit.

asv/benchmarks/simulation/bench_mujoco.py
vreutskyy: Done for all four `setup_cache` benchmark paths (MuJoCo, Kamino, ANYmal, and quadruped). They now return `None` before constructing a workload when CUDA is unavailable, with a regression test verifying that none of the builders/factories is called.

asv/benchmarks/simulation/bench_mujoco.py
vreutskyy: Done. The MuJoCo KPI workload creation now raises when CUDA graph capture was requested but no graph was produced, so KPI metrics cannot silently fall back to eager execution. `Example.step()` retains the eager fallback for non-KPI use, and both behaviors have focused coverage.

vreutskyy:
@adenzler-nvidia Thanks again for the detailed review. Update after #3574 merged:

- I rebased onto it and restored `FastKitchenG1` to the intended 512 worlds. The real 512-world workload now passes the full [`g7e.2xlarge` AWS benchmark job](https://github.com/newton-physics/newton/actions/runs/29744965135/job/88361211299) (41m29s), and all PR checks are green.
- MuJoCo KPI results now include mean and maximum measured `solver_niter` tracks.
- I replaced the brittle kitchen constant/mocking assertions with a runtime structural guard on the built workload, made the timeout test relative to ASV's configured default, and kept the Warp import no-leak regression because the import mutation had caused an actual unrelated softbody test failure. The harness now restores both `enable_backward` and `log_level`.
- I replied inline on each remaining design point, including the two places where I kept the existing behavior intentionally.

Current head is `6c526bd8bd688162974c1419e1cd5ed460d01d12`.

asv/benchmarks/simulation/bench_mujoco.py
```diff
@@ -14,12 +17,32 @@
 parent_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
 sys.path.append(parent_dir)
 
+from benchmark_metrics import (
+    _SimulationMetricTracks,
+    collect_simulation_metrics,
+)
 from benchmark_mujoco import Example
```
adenzler-nvidia: Every `track_*` entry runs in a freshly spawned process that only reads one value from the pickled `setup_cache` result, but this module-level import makes each of those processes pay the full newton/mujoco_warp import — ~2.4s of the ~3.1s module import on my machine; warp itself and the decorators' CUDA init are comparatively small. Deferring it into the paths that actually construct workloads cuts the measured per-entry process cost from ~3.5s to ~0.9s (asv A/B on `FastCartpole` at this head, identical metric values):

```python
    def _create_workload(self, builder, world_count):
        from benchmark_mujoco import Example  # noqa: PLC0415

        workload = Example(
```

Same pattern for `_collect_metrics`, `_NewtonOverheadBenchmark.setup`/`track_simulate` (with `EventTracer`), and the `newton` imports in `bench_quadruped_xpbd`, `bench_anymal`, and `bench_kamino`. Applied to all four metric bench files that's ~100 reader spawns per CI job (50 entries × 2 commits), roughly 2–3 minutes off the benchmark job. I have this as a verified commit — happy to push it onto this branch after your main merge, or as a follow-up PR if you'd rather land this one as is.

adenzler-nvidia (dismissed):
All my earlier threads are addressed in this iteration — I verified the comparison-filter regex against asv's benchmark matching semantics, the CPU `setup_cache` guards in all four paths, the KPI graph-capture error handling, and the kitchen body-count validation, and both new test files pass locally at this head (14/14). Looks good otherwise. One optional suggestion below, and the branch needs a main merge to clear the CHANGELOG/benchmark-file conflicts before the merge queue will take it — worth double-checking the two changelog entries don't get duplicated by the union merge driver.

adenzler-nvidia (approved):
Re-approving after the main merge and the deferred-imports commit. Verified at 2dcecb6fe: rebase fidelity clean (no changelog duplication, comparison-filter regex intact), deferral applied in all four bench files with the regression test, both benchmark test files pass locally, and importing all four bench modules now takes ~0.5s where bench_mujoco alone was ~3.1s before.

.github/workflows/aws_gpu_benchmarks.yml
jcarius-nv, line 1: Suggestion here, sorry I just saw this now: Can't we instead just track (or emit) the reciprocal of these values instead of straight out excluding them?

vreutskyy: That would make the ASV comparison direction correct, but I don't think we should replace the published values: #3343 explicitly asks for steps/s and real-time factor, and those are the familiar reporting units. The two reciprocals are also redundant with each other. The current lower-is-better comparison signals are physics-only mean world-step time plus complete-step p95 latency, so I excluded the human-facing higher-is-better tracks from `asv continuous`.

If we decide CI also needs mean complete-step regression sensitivity, the clean follow-up would be to add one separate lower-is-better metric—the reciprocal of throughput, e.g. complete wall time per world-step—while retaining steps/s and RTF for reporting.

adenzler-nvidia: given that solver_niter looks flaky as well, could we just report and block on actual runtime, and then have the derived values and additional data just in the dashboards?

vreutskyy: Agreed. #3604 implements this: `asv continuous` now gates only on directly measured runtime benchmarks. `solver_niter` and the other diagnostic/derived tracks remain available for dashboard reporting but no longer fail PR CI. #3575’s 5 → 6 maximum-iteration change while runtime improved is the motivating case.

jcarius-nv: ok, that's fine for me, thanks for the responses!

PR #3568, opened 2026-07-20
Add source-neutral USD schema resolution

adenzler-nvidia:
## Description

Adds one typed, source-neutral `SchemaResolution` object shared by `ModelBuilder.add_usd()` and consumers of populated or mapping-based scene data. Callers configure resolver priority and the fallback migration choice once, then pass the object to `add_usd(..., schema_resolution=resolution)` or call `resolution.resolve(...)` directly. Both paths use the same schema ownership, USD fallback catalog, conversions, and boolean compatibility choice.

The existing experimental `schema_resolvers=` argument remains as a compatibility shorthand: internally it constructs the same `SchemaResolution`. Supplying both arguments is rejected. `use_applied_schema_fallbacks=False` is the compatibility default; setting it to `True` on `SchemaResolution` selects composed applied-schema fallbacks for both PXR and non-PXR sources without migration warnings. The direct `add_usd()` boolean is shorthand when no shared resolution object is supplied.

The PXR adapter supplies registered schema fallbacks when available. Source-neutral consumers use the built-in catalog or explicit catalog additions, but do not need to construct PXR prims. Resolver definitions and evaluation internals remain private behind the small public object.

This first step deliberately leaves topology discovery, batching, Warp lowering, and direct builder-buffer population out of scope. Those can replace the scalar source adapter later without changing resolver configuration.

### PR relationship and dependencies

- Depends on #3572 and is kept synchronized with that preparatory branch.
- The independent PhysX D6 translational-limit fix from #3569 is integrated from `main`.

## Checklist

## API

```python
from newton import ModelBuilder
from newton.usd import SchemaResolution, SchemaResolverNewton

resolution = SchemaResolution(
    [SchemaResolverNewton()],
    use_applied_schema_fallbacks=True,
)

builder = ModelBuilder()
builder.add_usd("robot.usda", schema_resolution=resolution)

values = resolution.resolve(
    prim_type,
    populated_values,
    schemas=applied_schemas,
)
```

The design rationale and migration/removal sequence are documented in `docs/design/schema-resolution.md`.


docs/design/schema-resolution.md
```diff
@@ -0,0 +1,259 @@
+<!-- SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers -->
+<!-- SPDX-License-Identifier: CC-BY-4.0 -->
+
+# Source-neutral USD schema resolution
+
```
adenzler-nvidia, line 5: This file will be removed for the final version, but it motivates the design for now so I think it's useful to have it in the PR.

andrewkaufman: Can we just migrate this doc to an Issue that the PR implements?

adenzler-nvidia: I'd like to keep it up-to-date while the PR evolves, once final I will move the design to the PR description.

mzamoramora-nvidia:
Superseded by #3984

PR #3572, opened 2026-07-20
Prepare applied USD schema fallback precedence

adenzler-nvidia:
## Description

Prepares applied-schema fallback precedence while preserving current `ModelBuilder.add_usd()` results by default. Newton computes the future composed result internally: authored values follow resolver priority, while an applied, registered USD schema owns the fallback for an unauthored property. Fallbacks come from `Usd.SchemaRegistry` composed prim definitions; Newton no longer carries a copied Newton, PhysX, and MuJoCo fallback catalog.

`ModelBuilder.add_usd(..., use_applied_schema_fallbacks=True)` opts into the composed behavior without migration warnings. The boolean defaults to `False` during the compatibility period, so existing imports retain legacy precedence. Newton emits one aggregated `DeprecationWarning` when a registered schema fallback would change the effective result; explicitly authoring the affected values preserves asset behavior across the transition.

Unregistered schemas remain functional as a second-class compatibility path: authored values still participate in resolver priority, and resolver compatibility defaults remain after importer defaults when the schema applies. They do not claim first-class schema-fallback ownership. Registering the PhysX and MuJoCo codeless schema plugins is the path to authoritative fallback behavior without duplicated defaults.

The registry lookup is part of the PXR adapter, not the resolution policy. The source-neutral resolution layer consumes schema applicability and fallback values through callbacks, so a later PXR-free adapter can supply the same registered schema metadata without changing precedence or restoring a private fallback catalog.

One private compatibility path governs generic properties, joint gains and limits, and velocity-limit sentinels. Specialized authored reads are cached while auditing, and warning attribution points to the consumer call site. The composed resolver remains independent of the migration choice so the legacy branch and audit machinery can be removed after the default transition.

Existing custom resolvers, including PXR-only getters, remain compatible on the default path. Registered fallbacks take precedence in the opt-in path; unsupported PXR-only fallback getters are reported explicitly.

### PR relationship and dependencies

- This PR is a prerequisite for #3568.
- #3568 adds the source-neutral facade and uses the same boolean policy; it contains this PR in its ancestry.
- The independent PhysX D6 translational-limit fix from #3569 is now integrated from `main`, including matching multi-apply schema ownership metadata.
- Publishing and registering codeless PhysX and MuJoCo schema plugins is a follow-up needed for first-class fallback ownership for those schemas.

## Checklist

## Migration behavior

Applying a registered schema without authoring one of its properties currently allows an importer or resolver compatibility default to win. This PR detects where the schema's registered USD fallback will win after the compatibility period, preserves the current result, and asks the asset author to make the intended value explicit. Consumers can opt into the target behavior now with `use_applied_schema_fallbacks=True`.


eric-heiden:
Closing in favor of https://github.com/newton-physics/newton/pull/3888.

PR #3575, opened 2026-07-20
Use stock solver iterations in MuJoCo benchmarks

adenzler-nvidia:
## Description

The MuJoCo KPI and overhead benchmarks override `ls_iterations` per robot (3-15) and force `iterations`/`ls_iterations` fallbacks of 100/50 in `create_solver`, so the published series measure a hand-tuned solver configuration rather than what Newton does by default. mujoco-warp 3.10 terminates solver iterations per world on convergence even under CUDA graph capture (`graph_conditional` is on by default) and exits the linesearch early in-kernel, so the low fixed caps no longer buy meaningful performance: measured at 1024 worlds with random actuation, stock settings are on par with the overrides on Allegro (elliptic cone) and cartpole and slightly faster on G1, with identical convergence (`solver_niter` far below the caps in all cases).

This PR removes the per-class overrides and the forced fallbacks and passes `None` through, so `SolverMuJoCo` resolves iteration settings from model custom attributes or MuJoCo defaults. The `--solver-iteration`/`--ls-iteration` CLI flags and the `Example` parameters remain for explicit experiments.

Note for reviewers: the five KPI series keep their ASV version hashes (only class attributes and un-hashed helpers change), so their dashboards show a one-time step at the transition rather than a series reset. The two Newton-overhead series do get new hashes, since the pass-through removal is inside `track_simulate` itself; their history will be reconnected in newton-asv via the usual `replace_hash.sh` step after the first post-merge benchmark run.

## Checklist

## Test plan

- `uvx --with virtualenv asv check` (no problems found)
- Ran the affected workloads directly via `asv/benchmarks/simulation/bench_mujoco.py` comparing the removed overrides against stock settings on Allegro, cartpole, and G1, checking step time and `Data.solver_niter` convergence.

## Summary by CodeRabbit

* **Refactor**
  * Updated MuJoCo solver setup to preserve explicitly passed iteration values (including `None`) and forward them to the solver, while keeping other defaults from robot configuration.
  * Removed line-search iteration configuration across MuJoCo simulation benchmarks.
  * Updated MuJoCo teleoperation benchmark solver initialization to no longer explicitly set `iterations`/`ls_iterations`; defaults are used instead.
  * Benchmark setup and stepping behavior are otherwise unchanged.


vreutskyy (commented):
One question about ASV benchmark version continuity.

asv/benchmarks/simulation/bench_mujoco.py
```diff
@@ -249,7 +246,6 @@ def track_simulate(self, world_count):
                     world_count=world_count,
                     use_cuda_graph=True,
                     builder=self.builder[world_count],
-                    ls_iteration=self.ls_iteration,
                 )
```
vreutskyy, line 249: Was changing the ASV version hash for the Newton-overhead benchmarks intentional? ASV hashes the source of `track_simulate()`, so removing the `ls_iteration` pass-through resets the `FastNewtonOverheadG1` and `FastNewtonOverheadHumanoid` series, while the PR description says the affected series should retain their hashes and show a one-time step. If continuity is intended, could we keep the base `ls_iteration = None` and pass-through, removing only the per-robot overrides?

adenzler-nvidia: The PR body is imprecise for these two: in `_NewtonOverheadBenchmark` the pass-through sits inside `track_simulate` itself, so removing it does change the version hash — you're right that those two series reset as-is (the five KPI classes only touch class attributes and an un-hashed helper, so they keep their hashes). Rather than keeping a vestigial `ls_iteration=None` pass-through to freeze the hash, I'd land the full cleanup and reconnect the two series in newton-asv via the usual `replace_hash.sh` step once the first post-merge run lands the new hashes. I'll fix the PR description to say so.

vreutskyy (approved):
Looks good. The ASV version transition is documented, and CI is green.

PR #3580, opened 2026-07-20
Map contact kf to MuJoCo solreffriction for elliptic cones

### LINKED ISSUE #2987: Map contact `kf` to MuJoCo
Can we map `Model.shape_material_kf` to MuJoCo, e.g. tuning `impratio` based on `shape_material_ke` and `shape_material_kf`?

adenzler-nvidia:
## Description

Closes #2987.

`Model.shape_material_kf` was silently ignored by `SolverMuJoCo`. This PR maps it to MuJoCo's per-contact `solreffriction` when the solver uses elliptic friction cones (`cone="elliptic"`) with Newton contacts (`use_mujoco_contacts=False`), targeting the force-space friction slope `f = -kf * v` below the Coulomb limit so `kf` carries the same meaning as in `SolverSemiImplicit`/`SolverFeatherstone`.

Mechanism: friction rows are velocity-level and realize `f = D * (aref - A * f)`, so the row damping is chosen as `beta = kf * (1/D + A)` with MuJoCo's `A ~ body_invweight0` approximation, and written in the standard positive solref format `(2 / (kf * invw * ((1 - dmax) / impratio + dmax)), 1.0)` so the refsafe clamp bounds large `kf` at the stability limit. `kf <= 0` on either shape keeps the `(0, 0)` sentinel (inherit `solref`).

This is a behavior change for elliptic + Newton-contacts scenes (the `ShapeConfig.kf` default is 1000): the CHANGELOG entry is under Changed with migration guidance (set `kf` to `0` to restore the previous solref-inherited friction).

Limitations (documented in `docs/solvers/mujoco.rst`): elliptic cones only (pyramidal facets never read `solreffriction`); Newton contacts only (MuJoCo has no per-geom `solreffriction`, only explicit pairs); exact when `body_invweight0` matches the contact's effective mass and the contact operates at `dmax` impedance; very large `kf` saturates at the refsafe bound (time constant clamped to twice the timestep); the two shapes' `kf` combine with the usual priority/solmix weighting, except `kf = 0` on either shape disables the mapping regardless of priority; for `condim > 3` the torsional/rolling rows share the per-contact value and deviate via MuJoCo's friction-ratio scaling; the mapping is always on, independent of the shape's `solref_mode` (which only governs the normal-direction `solref`).

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k "test_mujoco_solver.TestMuJoCoSolverContactKf"
```

Contract tests assert the written `solreffriction` values against the formula, including priority/solmix mixing, the `kf = 0` sentinel, runtime `kf` updates via `notify_model_changed`, and `impratio != 1`. Behavior tests verify the force-space semantics end to end on a single-contact, rotation-locked rig (sphere pinned by a prismatic joint): equal `kf` at 8x the mass decays 8x slower, and larger `kf` decays faster.

## New feature / API change

```python
import newton

builder.default_shape_cfg.kf = 500.0  # contact friction damping [N·s/m]
# ... build scene ...
model = builder.finalize()
solver = newton.solvers.SolverMuJoCo(model, use_mujoco_contacts=False, cone="elliptic")
```

## Summary by CodeRabbit

* **Changed**
  * MuJoCo per-contact friction damping for elliptic friction cones now derives from Newton material `kf` during Newton contact processing.
  * `kf` is mixed between contacting shapes; `kf = 0` restores the previous MuJoCo friction behavior.
  * Contact response is rescaled when `impratio` changes, and runtime edits to `kf` update results accordingly.
* **Documentation**
  * Expanded MuJoCo solver docs to describe the `kf` → `solreffriction` mapping, mixing rules, and scaling behavior.
  * Updated solver index references for `kf`.
* **Tests**
  * Added `kf`→`solreffriction` validation, including runtime update and impedance scaling checks; added joint behavior coverage and native-contacts mode support in friction-ramp benchmarks.
* **Examples**
  * Updated Allegro hand default `kf` to `0.0`.


docs/solvers/mujoco.rst
```diff
@@ -281,6 +281,30 @@ the mode from user code.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For :class:`~newton.solvers.SolverMuJoCo`, ``kf`` maps to MuJoCo's per-contact
+``solreffriction`` when the solver uses elliptic friction cones (``cone="elliptic"``)
+with Newton contacts (``use_mujoco_contacts=False``); it targets the force-space
+friction slope ``f = -kf * v`` below the Coulomb limit (exact when MuJoCo's
+``body_invweight0`` matches the contact's effective mass and the contact operates at
+its maximum impedance ``dmax``). Very large ``kf`` saturates at MuJoCo's refsafe
+stability bound (reference time constant clamped to twice the timestep). Set ``kf``
+to ``0`` on a shape to fall back to MuJoCo's default friction impedance. This mapping
```
eric-heiden: When we set `kf = 0` in Featherstone and SemiImplicit solvers it means that we apply zero friction force. I don't have a good solution how to reconcile it with MuJoCo's default fallback behavior here, maybe a special case for MuJoCo is acceptable, or using `inf`? Note the default `kf` in the ModelBuilder is `1000.0` anyway. So by default you are not getting the "default" MuJoCo behavior.

newton/_src/solvers/mujoco/kernels.py
```diff
@@ -626,6 +628,21 @@ def convert_newton_contacts_to_mjwarp_kernel(
...
+        # large kf; dampratio only feeds the zero position term.
+        if shape_material_kf:
+            kf1 = shape_material_kf[shape_a]
+            kf2 = shape_material_kf[shape_b]
+            if kf1 > 0.0 and kf2 > 0.0:
+                kf = mix * kf1 + (1.0 - mix) * kf2
+                invw = body_invweight0[worldid, mj_body_a][0] + body_invweight0[worldid, mj_body_b][0]
+                ir = opt_impratio_invsqrt[worldid % opt_impratio_invsqrt.shape[0]]
+                imp = solimp[1]
+                solreffriction = wp.vec2(2.0 / (kf * invw * ((1.0 - imp) * ir * ir + imp)), 1.0)
```
eric-heiden: 🟠 Please guard this division before writing `solreffriction`. The earlier immovable-pair check does not guarantee that the sum of `body_invweight0` values is positive for every supported topology, so a zero effective inverse weight can produce `inf` here. MuJoCo then treats that nonzero value as the friction-row reference, which can corrupt the contact solve.

Please compute the denominator first and assign `solreffriction` only when it is positive and finite, leaving the existing zero fallback otherwise.

docs/solvers/index.rst
```diff
@@ -176,7 +176,9 @@ formulation.
   :class:`~newton.solvers.SolverMuJoCo`, and
   :class:`~newton.solvers.SolverVBD`.
 - ``kf`` / ``ka``: :class:`~newton.solvers.SolverFeatherstone` and
-  :class:`~newton.solvers.SolverSemiImplicit`.
+  :class:`~newton.solvers.SolverSemiImplicit`; ``kf`` also
+  :class:`~newton.solvers.SolverMuJoCo`
+  (see :ref:`mujoco-contact-friction-solreffriction`).
```
eric-heiden, line 181: ⚪ This sentence is missing a verb, so the rendered documentation reads “`kf` also `SolverMuJoCo`.” Could this say “`kf` is also used by `SolverMuJoCo`”?

docs/solvers/mujoco.rst
```diff
@@ -281,6 +281,30 @@ the mode from user code.
...
+
+Contact friction ``solreffriction`` mapping
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For :class:`~newton.solvers.SolverMuJoCo`, ``kf`` maps to MuJoCo's per-contact
+``solreffriction`` when the solver uses elliptic friction cones (``cone="elliptic"``)
+with Newton contacts (``use_mujoco_contacts=False``); it targets the force-space
+friction slope ``f = -kf * v`` below the Coulomb limit (exact when MuJoCo's
+``body_invweight0`` matches the contact's effective mass and the contact operates at
+its maximum impedance ``dmax``). Very large ``kf`` saturates at MuJoCo's refsafe
```
eric-heiden: ⚪ `body_invweight0[..., 0]` has inverse-mass units and approximates the constraint-space `J M^-1 J^T` term, so “effective mass” is dimensionally reversed here. Please call this the contact's “inverse effective mass” or explicitly describe it as the approximated `J M^-1 J^T` diagonal.

vreutskyy (approved):

PR #3584, opened 2026-07-21
Add compiled regex support to label selection

### LINKED ISSUE #3272: [REQ] Full regex support for Selection APIs

mzamoramora-nvidia:
# Add compiled regex support to label selection

## Description

Closes #3272.

This PR adds compiled regular expressions as an explicit opt-in for Newton's shared label-selection APIs. Passing a compiled `re.Pattern[str]` now uses `fullmatch()` against each complete label, while ordinary strings retain their existing `fnmatch` glob semantics.

The shared matcher supports compiled patterns for:

- `ArticulationView` articulation labels;
- `ArticulationView` joint and link inclusion/exclusion filters; and
- label selectors used by `SensorIMU`, `SensorContact`, and `SensorFrameTransform`.

Lists may combine glob strings and compiled patterns. Their union remains deduplicated and ordered by the model labels. Existing integer selectors and their validation are unchanged. Compiled byte patterns and lists that mix pattern and index selector families raise `TypeError`.

This change is intentionally separate from PR #3326 and does not modify `DeformableView`. That PR can adopt the shared compiled-pattern contract when rebased after this change.

## Checklist

## Test plan

The following verification commands cover the matcher contract, `ArticulationView` integration, public API, documentation, and repository checks:

```bash
uv run --extra dev -m newton.tests -k test_match_labels
uv run --extra dev -m newton.tests -k test_api
uv run docs/generate_api.py
uvx pre-commit run -a
```

## New feature / API change

Compile a regular expression to opt into full regex syntax:

```python
import re

import newton

robots = newton.selection.ArticulationView(
    model,
    pattern=re.compile(r"/World/envs/env_[0-9]+/Robot_(A|B|C)"),
)

feet = newton.selection.ArticulationView(
    model,
    pattern="/World/envs/*/Robot",
    include_links=[re.compile(r"(LF|RF)_FOOT")],
    exclude_joints=[re.compile(r"fixed_.*")],
)
```

Strings continue to use glob matching, so regex-looking text is not interpreted as a regular expression unless it is compiled explicitly:

```python
glob_view = newton.selection.ArticulationView(model, pattern="/World/envs/*/Robot")
regex_view = newton.selection.ArticulationView(
    model,
    pattern=re.compile(r"/World/envs/env_[0-9]+/Robot"),
)
```

## Summary by CodeRabbit

* **New Features**
  * Added compiled regular-expression support for label-based selectors across articulation selection and sensor types (including contact, frame transform, and IMU).
  * Regex uses full-string matching; existing glob-string behavior remains unchanged.
* **Documentation**
  * Expanded label-matching documentation and examples, including how to achieve substring matching with regex.
* **Tests**
  * Added coverage for full-match regex behavior, selector integration, and no-match error handling.


docs/concepts/sensors.rst
```diff
@@ -79,15 +79,25 @@ support label matching accept one of the following:
 - A **list of integer indices** -- selects directly by index.
 - A **single string pattern** -- selects all entries whose label matches the pattern via :func:`fnmatch.fnmatch`
   (supports ``*`` and ``?`` wildcards).
-- A **list of string patterns** -- selects all entries whose label matches at least one of the patterns.
+- A **compiled string regular expression** -- selects all entries whose complete label matches via
+  :meth:`re.Pattern.fullmatch`.
+- A **list of string and/or compiled patterns** -- selects all entries whose label matches at least one pattern.
```
camevor: Do we need to support a list of compiled patterns? It seems like that would trivially be handled by the regular expression itself, simplifying the signature.

newton/tests/test_match_labels.py
camevor, line 1: Many of these tests look like they are testing `re` internals. Newton's contribution here is one line of dispatch: compiled pattern → `pattern.fullmatch(label)`.

newton/_src/sensors/sensor_frame_transform.py
```diff
@@ -120,20 +122,21 @@ class SensorFrameTransform:
     def __init__(
         self,
         model: Model,
-        shapes: str | list[str] | list[int],
-        reference_sites: str | list[str] | list[int],
+        shapes: str | re.Pattern[str] | list[str | re.Pattern[str]] | list[int],
```
camevor: nit: Recommend ordering these as `str | list[str] | re.Pattern[str] | list[int]`.

camevor (commented):
Thanks, @mzamoramora-nvidia. I think supporting regular expressions via compiled patterns makes for a clean and non-obtrusive addition.

However, this change could be quite a bit more minimal by simply allowing one more option besides `str`, `list[str]`, `list[int]`, which would be `re.Pattern`.

The tests should also be improved a bit.

docs/concepts/sensors.rst
mzamoramora-nvidia: Agreed. Compiled regexes are now scalar-only; lists remain the existing `list[str]` or `list[int]` forms. Addressed in commit 35533455.

newton/tests/test_match_labels.py
mzamoramora-nvidia: Agreed. I reduced the tests to Newton-owned dispatch, full-match semantics, and glob compatibility, and simplified the matcher to the direct compiled-pattern dispatch. Addressed in commit 5c8be198.

newton/_src/sensors/sensor_frame_transform.py
mzamoramora-nvidia: Updated the selector annotations consistently to `str | list[str] | re.Pattern[str] | list[int]`. Addressed in commit 21123423.

camevor:
Thanks, @mzamoramora-nvidia ! I think this looks good. I'd like to get an opinion from @nvlukasz and / or @eric-heiden on the question of adding compiled regex's to the supported patterns.

docs/concepts/sensors.rst
```diff
@@ -79,13 +79,23 @@ support label matching accept one of the following:
 - A **list of integer indices** -- selects directly by index.
 - A **single string pattern** -- selects all entries whose label matches the pattern via :func:`fnmatch.fnmatch`
   (supports ``*`` and ``?`` wildcards).
-- A **list of string patterns** -- selects all entries whose label matches at least one of the patterns.
+- A **list of string patterns** -- selects all entries whose label matches at least one pattern.
+- A **compiled string regular expression** -- selects all entries whose complete label matches via
+  :meth:`re.Pattern.fullmatch`.
```
eric-heiden, line 84: ⚪ This description is not accurate for every selector. `ArticulationView` builds `arti_joint_names` and `arti_link_names` with `get_name_from_label`, so its joint and link filters match leaf names such as `LF_FOOT`, not complete labels such as `/World/envs/env_0/Robot_A/LF_FOOT`.

Please either document that `include_joints`, `exclude_joints`, `include_links`, and `exclude_links` operate on leaf names, or change those filters to match the full articulation labels and add a full-path regex test.

eric-heiden:
Thanks, I think this is a good feature to have! Just a nit on the wording in the docstring regarding "full" label matching since we use leaf names only in some cases.

docs/concepts/sensors.rst
mzamoramora-nvidia: Addressed in e485aab

camevor (approved):

PR #3586, opened 2026-07-21
Update PADMM test tolerances

chschuma-disney:
## Description

#3564 has lead to some test failures on some of our developer machines, which later turned out to always have failed for other developers. This seems to point to a test tolerance being set too tight, with test success being dependent on the specific hardware and compiler. The additional register availability introduced by disabling grid-stride loops in #3564 turned out to be sufficient to introduce some numerical divergence, leading to earlier PADMM convergence, but a slightly larger error in some test metrics.

This PR updates the tolerances for some of checks in the PADMM tests.

## Checklist

## Test plan

```
python -m unittest "newton/_src/solvers/kamino/tests/test_solvers_padmm.py"
```

## Summary by CodeRabbit

* **Tests**
  * Relaxed validation thresholds for PADMM dual residual checks to accommodate expected numerical differences.


nvtw (approved):

Guirec-Maloisel (approved):
Looks good to me, this is not the first instance in which PADMM unit tests have been failing seemingly randomly for me, so I think it makes sense to loosen the tolerances a bit.

PR #3590, opened 2026-07-21
Add NVTX Scope for SensorTiledCamera rendering

daniela-hase:
## Description

Add an NVTX profiling scope around `SensorTiledCamera.update()` so tiled-camera transform sync and rendering are visible under Newton’s existing `NEWTON_PROFILE` profiling switch.

- Add a module-level `PROFILE_ENABLED` gate using `NEWTON_PROFILE`.
- Wrap tiled-camera update work in `wp.ScopedTimer("Newton::SensorTiledCamera::update", use_nvtx=True)`.
- Leave default rendering behavior unchanged when profiling is disabled.

## Checklist

- [x] `CHANGELOG.md` update not required for profiling-only instrumentation

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_sensor_tiled_camera.TestSensorTiledCamera.test_tiled_render_order_accepts_partial_tiles
```

Also attempted the profiling-enabled path:

```bash
NEWTON_PROFILE=1 uv run --extra dev -m newton.tests -k test_sensor_tiled_camera.TestSensorTiledCamera.test_tiled_render_order_accepts_partial_tiles
```

## Summary by CodeRabbit

* **Performance**
  * Added optional profiling for tiled camera updates when enabled via the `NEWTON_PROFILE` environment variable.
  * When profiling is on, camera synchronization and rendering are measured together using timing markers, with synchronization performed when the profiling scope ends.

### DIFF
--- a/newton/_src/sensors/sensor_tiled_camera.py
+++ b/newton/_src/sensors/sensor_tiled_camera.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+import os
 import warnings
 from typing import Any
 
@@ -18,6 +19,8 @@ from .warp_raytrace import (
     RenderOrder,
     Utils,
 )
+
+PROFILE_ENABLED = os.environ.get("NEWTON_PROFILE", "0") != "0"
 
 _RENDER_CONFIG_DEPRECATION_MSG = (
     "SensorTiledCamera.render_config is deprecated as of Newton 1.4; "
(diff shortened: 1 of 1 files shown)


newton/_src/sensors/sensor_tiled_camera.py
```diff
@@ -224,24 +227,27 @@ def update(
...
-            forward_depth_image=forward_depth_image,
-            shape_index_image=shape_index_image,
-            normal_image=normal_image,
-            albedo_image=albedo_image,
-            clear_data=clear_data if clear_data is not None else self.default_clear_data,
-            config=render_config if render_config is not None else self.default_render_config,
-            kernel_block_dim=kernel_block_dim,
-        )
+        with wp.ScopedTimer(
+            "Newton::SensorTiledCamera::update", active=PROFILE_ENABLED, use_nvtx=True, print=False, synchronize=True
```
shi-eric: The `use_nvtx=True` and `synchronize=True` should be cleaned up before merging. The first one crashes if `nvtx` is not available, right? And the second option is an unnecessary sync you're forcing on all users even when not profiling?

shi-eric: But I see that you're just following what `viewer_rtx.py` does, right? In that case we might just leave it alone and clean it up later...

daniela-hase: > The first one crashes if nvtx is not available, right?

Only if profiling is enabled (which it isn't by default). But yes, if the user runs with profiling they need to install the nvtx pip package.

> The use_nvtx=True and synchronize=True should be cleaned up before merging. The first one crashes if nvtx is not available, right? And the second option is an unnecessary sync you're forcing on all users even when not profiling?

No, only during profiling :)

Both are covered by the `active=PROFILE_ENABLED`. The `active` parameter in the `wp.ScopedTimer` pretty much turns this into a no-op if set to `False`.

And yeah, I just followed `viewer_rtx.py` :)

shi-eric (approved):

newton/_src/sensors/sensor_tiled_camera.py
shi-eric: Thanks for the clarifications!

eric-heiden (approved):

PR #3594, opened 2026-07-21
Fix USD collider mass aggregation

### LINKED ISSUE #3591: [BUG] add_usd ignores collider MassAPI without body MassAPI
### Bug Description

`ModelBuilder.add_usd()` ignores collider-level `UsdPhysics.MassAPI` mass and density when the owning rigid-body prim does not itself have `MassAPI`.

A rigid body with an incomplete body-level `MassAPI` takes Newton's `UsdPhysics.RigidBodyAPI.ComputeMassProperties()` path and behaves correctly. The same collider subtree under a body with no body-level `MassAPI` instead uses builder-accumulated shape mass at the default density, ignoring collider mass/density and excluding disabled colliders from body mass.

Results on current `main`:

| Fixture (0.2 m cube; volume 0.008 m³) | Expected body mass | Newton |
|---|---:|---:|
| Collider `physics:mass = 0.05` | 0.05 | 8.0 |
| Collider `physics:density = 500` | 4.0 | 8.0 |
| Disabled collider mass 0.05 + enabled collider density 500 | 4.05 | 8.0 |
| Collider `physics:mass = 0` | 8.0 | 8.0 |

The last case is a control for the value semantics fixed by #3418: zero mass is ignored as an override and implicit mass is computed.

The gating is visible in:

- https://github.com/newton-physics/newton/blob/ce603b3b271c864070a4ea3b0996bdd2295d801c/newton/_src/utils/import_usd.py#L2494-L2514 — only bodies that already have `MassAPI` enter the fallback set.
- https://github.com/newton-physics/newton/blob/ce603b3b271c864070a4ea3b0996bdd2295d801c/newton/_src/utils/import_usd.py#L3749-L3801 — collider `MassInformation` is populated only for that set.
- https://github.com/newton-physics/newton/blob/ce603b3b271c864070a4ea3b0996bdd2295d801c/newton/_src/utils/import_usd.py#L3900-L3907 — bodies without body `MassAPI` are skipped during mass-property resolution.

Expected behavior: use OpenUSD aggregation for every rigid body whose effective body-level mass properties are incomplete, including bodies with no body-level `MassAPI`. Body-level mass must retain precedence; disabled collision prims must still participate in mass aggregation.

This is the collider-only aggregation follow-up explicitly deferred by #3418. It is also the `ComputeMassProperties()` follow-up mentioned in closed issue #1409. It is distinct from #3236: this report tests generic USD semantics and does not expect Newton to reconstruct MJCF intent that was not represented in USD.

### Reproduction Script

```python

import newton


def add_cube(stage, path, *, enabled=True, mass=None, density=None):
    cube = UsdGeom.Cube.Define(stage, path)
    cube.CreateSizeAttr(0.2)
    collision = UsdPhysics.CollisionAPI.Apply(cube.GetPrim())
    collision.CreateCollisionEnabledAttr(enabled)

    if mass is not None or density is not None:
        mass_api = UsdPhysics.MassAPI.Apply(cube.GetPrim())
        if mass is not None:
            mass_api.CreateMassAttr(mass)
        if density is not None:
            mass_api.CreateDensityAttr(density)


def imported_mass(cubes, partial_body_mass_api=False):
    UsdPhysics.SetStageKilogramsPerUnit(stage, 1.0)

    if partial_body_mass_api:
        UsdPhysics.MassAPI.Apply(body.GetPrim()).CreateCenterOfMassAttr(Gf.Vec3f(0.0))

    for index, kwargs in enumerate(cubes):
        add_cube(stage, f"/Body/Cube{index}", **kwargs)

    builder.add_usd(stage)
    return builder.body_mass[0]


print(imported_mass([{"mass": 0.05}]))       # expected 0.05; actual 8.0
print(imported_mass([{"density": 500.0}]))  # expected 4.0; actual 8.0
print(
    imported_mass(
        [
            {"enabled": False, "mass": 0.05},
            {"density": 500.0},
        ]
    )
)  # expected 4.05; actual 8.0

# Controls: applying only centerOfMass to the body forces the existing
# ComputeMassProperties path, which honors the same collider MassAPI.
print(imported_mass([{"mass": 0.05}], partial_body_mass_api=True))       # 0.05
print(imported_mass([{"density": 500.0}], partial_body_mass_api=True))  # 4.0
```

### System Information

- Warp: `1.16.0.dev20260716`
- Python: 3.12.3
- OS: Windows

eric-heiden:
## Description

A rigid body without `PhysicsMassAPI` ignored mass and density authored on its colliders. Newton used its default shape density instead.

This change runs OpenUSD collider aggregation when a body owns a collider with `PhysicsMassAPI`. Bodies without collider mass metadata keep Newton's existing shape-only path.

The regression test covers:

- positive collider mass
- positive collider density
- zero mass falling back to density
- mass on a disabled collider

Closes #3591.

## Checklist

## Test plan

- `uv run --extra dev -m newton.tests -k test_collider_massapi_without_body_massapi`
- `uv run --extra dev -m newton.tests -p test_import_usd.py` (264 tests)
- `uvx pre-commit run -a`

## Bug fix

Before this change, a collider mass of `0.05` kg imported as `8.0` kg when its rigid body had no `PhysicsMassAPI`. The importer now reports `0.05` kg and follows OpenUSD mass precedence.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved USD import so collider-authored mass and density are honored even when the rigid body has no mass properties.
  - Corrected mass/inertia aggregation for instanced colliders and properly accounts for enabled versus disabled colliders, using collider-derived inertials when mass-property computation fails.
  - Reduced warning noise by only warning when relevant mass properties are present and used.

- **Tests**
  - Added regression tests covering collider `MassAPI` mass/density aggregation when body `MassAPI` is missing, including instanced collider scenarios and inertia verification.

### DIFF
--- a/newton/_src/utils/import_usd.py
+++ b/newton/_src/utils/import_usd.py
@@ -2491,7 +2491,7 @@ def parse_usd(
             body_specs[body_path] = rigid_body_desc
             prim = stage.GetPrimAtPath(prim_path)
 
-    # Bodies with MassAPI that need ComputeMassProperties fallback (missing mass, inertia, or CoM).
+    # Bodies that need ComputeMassProperties fallback (no MassAPI, or missing mass, inertia, or CoM).
     bodies_requiring_mass_properties_fallback: set[str] = set()
     if UsdPhysics.ObjectType.RigidBody in ret_dict:
         prim_paths, rigid_body_descs = ret_dict[UsdPhysics.ObjectType.RigidBody]
@@ -2505,6 +2505,14 @@ def parse_usd(
             prim = stage.GetPrimAtPath(prim_path)
             mass_api = UsdPhysics.MassAPI(prim)
             if not mass_api:
+                descendants = iter(Usd.PrimRange(prim))
+                for descendant in descendants:
+                    if descendant != prim and descendant.HasAPI(UsdPhysics.RigidBodyAPI):
+                        descendants.PruneChildren()
+                        continue
+                    if descendant.HasAPI(UsdPhysics.CollisionAPI) and descendant.HasAPI(UsdPhysics.MassAPI):
+                        bodies_requiring_mass_properties_fallback.add(body_path)
+                        break
                 continue
 
             has_effective_mass = _mass_api_effective_mass(mass_api) is not None
@@ -3897,22 +3905,22 @@ def parse_usd(
             warned_missing_collider_mass_info.add(collider_path)
         return rigid_body_mass_info_map.get(collider_path, zero_mass_information)
 
-    # overwrite inertial properties of bodies that have PhysicsMassAPI schema applied
+    # Resolve body inertial properties from authored values and collider aggregation.
     if UsdPhysics.ObjectType.RigidBody in ret_dict:
         paths, rigid_body_descs = ret_dict[UsdPhysics.ObjectType.RigidBody]
         for path, rigid_body_desc in zip(paths, rigid_body_descs, strict=False):
             prim = stage.GetPrimAtPath(path)
             mass_api = UsdPhysics.MassAPI(prim)
-            if not mass_api:
-                continue
             body_path = str(path)
+            if not mass_api and body_path not in bodies_requiring_mass_properties_fallback:
+                continue
             body_id = path_body_map.get(body_path, -1)
             if body_id == -1:
                 continue
-            effective_mass = _mass_api_effective_mass(mass_api)
-            effective_density = _mass_api_effective_density(mass_api, warn_invalid=True)
-            effective_diag_inertia = _mass_api_effective_diag_inertia(mass_api)
-            effective_com = _mass_api_effective_com(mass_api)
+            effective_mass = _mass_api_effective_mass(mass_api) if mass_api else None
+            effective_density = _mass_api_effective_density(mass_api, warn_invalid=True) if mass_api else None
+            effective_diag_inertia = _mass_api_effective_diag_inertia(mass_api) if mass_api else None
+            effective_com = _mass_api_effective_com(mass_api) if mass_api else None
             has_effective_mass = effective_mass is not None
             has_effective_inertia = effective_diag_inertia is not None
             has_effective_com = effective_com is not None
@@ -4045,7 +4053,7 @@ def parse_usd(
                     builder.body_inertia[body_id] = wp.mat33(np.array(builder.body_inertia[body_id]) * scale)
                     builder.body_inv_inertia[body_id] = wp.inverse(builder.body_inertia[body_id])
             else:
-                raw_mass = mass_api.GetMassAttr().Get()
+                raw_mass = mass_api.GetMassAttr().Get() if mass_api else None
                 if raw_mass is not None and raw_mass != 0.0:
                     warnings.warn(
                         f"Body {body_path}: authored mass is not positive and finite. "
@@ -4086,7 +4094,7 @@ def parse_usd(
                         print(
                             f"Applied default inertia matrix for body {body_path}: diagonal elements = [{I_default[0, 0]}, {I_default[1, 1]}, {I_default[2, 2]}]"
                         )
-                else:
+                elif mass_api:
                     warnings.warn(
                         f"Body {body_path} has zero mass and zero inertia despite having the MassAPI USD schema applied.",
                         stacklevel=2,
--- a/newton/_src/utils/import_usd.py
+++ b/newton/_src/utils/import_usd.py
@@ -33,7 +33,7 @@ import warp as wp
 
 from ..core import quat_between_axes
 from ..core.types import Axis, Transform
-from ..geometry import GeoType, Mesh, ShapeFlags, compute_inertia_shape, compute_inertia_sphere
+from ..geometry import GeoType, Mesh, ShapeFlags, compute_inertia_shape, compute_inertia_sphere, transform_inertia
 from ..sim.builder import ModelBuilder
 from ..sim.enums import JointTargetMode, JointType
 from ..sim.model import Model
@@ -701,7 +701,7 @@ def parse_usd(
         pruning subtrees owned by nested rigid bodies."""
         if _mass_api_has_blocked_attrs(body_prim):
             return True
-        it = iter(Usd.PrimRange(body_prim))
+        it = iter(Usd.PrimRange(body_prim, Usd.TraverseInstanceProxies()))
         for prim in it:
             if prim != body_prim and prim.HasAPI(UsdPhysics.RigidBodyAPI):
                 it.PruneChildren()
@@ -2505,7 +2505,7 @@ def parse_usd(
             prim = stage.GetPrimAtPath(prim_path)
             mass_api = UsdPhysics.MassAPI(prim)
             if not mass_api:
-                descendants = iter(Usd.PrimRange(prim))
+                descendants = iter(Usd.PrimRange(prim, Usd.TraverseInstanceProxies()))
                 for descendant in descendants:
                     if descendant != prim and descendant.HasAPI(UsdPhysics.RigidBodyAPI):
                         descendants.PruneChildren()
@@ -3258,7 +3258,76 @@ def parse_usd(
     no_collision_shapes = set()
     collision_group_ids = {}
     rigid_body_mass_info_map = {}
+    rigid_body_mass_fallback_data = {}
     expected_fallback_collider_paths: set[str] = set()
+
+    def _record_fallback_collider_mass_information(
+        path: str,
+        prim: Usd.Prim,
+        shape_spec,
+        shape_type,
+        *,
+        density: float,
+        is_solid: bool,
+        thickness: float,
+    ):
+        """Record collider mass information used by the rigid-body fallback callback."""
+        if str(shape_spec.rigidBody) not in bodies_requiring_mass_properties_fallback:
+            return
+
+        shape_geo_type = None
+        shape_scale = wp.vec3(1.0, 1.0, 1.0)
+        shape_src = None
+        if shape_type == UsdPhysics.ObjectType.CubeShape:
+            shape_geo_type = GeoType.BOX
+            hx, hy, hz = shape_spec.halfExtents
+            shape_scale = wp.vec3(hx, hy, hz)
+        elif shape_type == UsdPhysics.ObjectType.SphereShape:
+            shape_geo_type = GeoType.SPHERE
+            shape_scale = wp.vec3(shape_spec.radius, 0.0, 0.0)
+        elif shape_type == UsdPhysics.ObjectType.CapsuleShape:
+            shape_geo_type = GeoType.CAPSULE
+            shape_scale = wp.vec3(shape_spec.radius, shape_spec.halfHeight, 0.0)
+        elif shape_type == UsdPhysics.ObjectType.CylinderShape:
+            shape_geo_type = GeoType.CYLINDER
+            shape_scale = wp.vec3(shape_spec.radius, shape_spec.halfHeight, 0.0)
+        elif shape_type == UsdPhysics.ObjectType.ConeShape:
+            shape_geo_type = GeoType.CONE
+            shape_scale = wp.vec3(shape_spec.radius, shape_spec.halfHeight, 0.0)
+        elif shape_type == UsdPhysics.ObjectType.MeshShape:
+            shape_geo_type = GeoType.MESH
+            shape_scale = wp.vec3(*shape_spec.meshScale)
+            shape_src = _get_mesh_cached(prim)
+        if shape_geo_type is None:
+            return
+
+        expected_fallback_collider_paths.add(path)
+        shape_axis = getattr(shape_spec, "axis", None)
+        mass_info = _build_mass_info_from_effective_properties(
+            prim,
+            shape_spec.localPos,
+            shape_spec.localRot,
+            shape_geo_type,
+            shape_scale,
+            shape_src,
+            shape_axis,
+        )
+        if mass_info is None:
+            mass_info = _build_mass_info_from_shape_geometry(
+                prim,
+                shape_spec.localPos,
+                shape_spec.localRot,
+                shape_geo_type,
+                shape_scale,
+                shape_src,
+                shape_axis,
+                is_solid=is_solid,
+                thickness=thickness,
+            )
+        if mass_info is not None:
+            rigid_body_mass_info_map[path] = mass_info
+            rigid_body_mass_fallback_data[path] = (str(shape_spec.rigidBody), density)
+
     for key, value in ret_dict.items():
         if key in {
             UsdPhysics.ObjectType.CubeShape,
@@ -3280,10 +3349,7 @@ def parse_usd(
                 # Deformable-owned meshes never reach this loop: the scout excludes them
                 # from the native parse. A sim-API mesh seen here was deliberately left
                 # rigid (e.g. its body API conflicts with RigidBodyAPI), so import it.
-                if path in path_shape_map:
-                    if verbose:
-                        print(f"Shape at {path} already added, skipping.")
-                    continue
+                shape_already_added = path in path_shape_map
                 body_path = str(shape_spec.rigidBody)
                 if verbose:
                     print(f"collision shape {prim.GetPath()} ({prim.GetTypeName()}), body = {body_path}")
@@ -3581,6 +3647,20 @@ def parse_usd(
                 else:
                     inertia_margin = margin_val
 
+                if shape_already_added:
+                    _record_fallback_collider_mass_information(
+                        path,
+                        prim,
+                        shape_spec,
+                        key,
+                        density=shape_density,
+                        is_solid=shape_is_solid,
+                        thickness=inertia_margin,
+                    )
+                    if verbose:
+                        print(f"Shape at {path} already added; skipping duplicate geometry.")
+                    continue
+
                 shape_params = {
                     "body": body_id,
                     "xform": shape_xform,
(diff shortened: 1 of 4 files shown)


newton/tests/test_import_usd.py
```diff
@@ -6988,6 +6988,58 @@ def test_collider_massapi_density_used_by_mass_properties(self):
...
+                self.assertAlmostEqual(mass, expected_mass, places=5)
+                if name == "disabled collider still contributes":
+                    continue
+                expected_diag = (1.0 / 6.0) * expected_mass * (0.2**2)
+                np.testing.assert_allclose(
+                    inertia,
+                    np.diag([expected_diag, expected_diag, expected_diag]),
+                    atol=1e-6,
+                    rtol=1e-5,
+                )
```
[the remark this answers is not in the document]
eric-heiden: Good catch. Updated the case to pair `mass=0.0` with authored `density=500.0`; the focused regression passes in commit 6efc2c237.

eric-heiden:
Addressed both findings in 8aec6f858.

- The fallback eligibility and blocked-attribute scans now traverse instance proxies.
- Disabled colliders that were already loaded as visuals now still provide mass information to the fallback callback.
- On the usd-core version used by Newton, `ComputeMassProperties` still fails to traverse instance-proxy colliders. When that happens, Newton now aggregates the recorded collider data itself, preserving collider mass/density precedence and consistent inertia.
- Added regression coverage for instance-proxy colliders with no body MassAPI (explicit mass and density), and for disabled colliders with explicit mass and density. The disabled-collider cases now assert inertia too.

Validation:
- `uv run --extra dev -m newton.tests -p test_import_usd.py` — 265 tests passed
- `uvx pre-commit run -a` — all hooks passed

newton/_src/utils/import_usd.py
```diff
@@ -3968,24 +4053,36 @@ def _get_collision_mass_information(collider_prim: Usd.Prim):
...
+                    recorded_properties = _aggregate_recorded_mass_properties(
+                        body_path, effective_density if not has_effective_mass else None
+                    )
+                    if recorded_properties is not None:
+                        cmp_mass, recorded_inertia, cmp_com = recorded_properties
+                        builder.body_inertia[body_id] = recorded_inertia
+                        if any(recorded_inertia):
+                            builder.body_inv_inertia[body_id] = wp.inverse(recorded_inertia)
                         else:
                             builder.body_inv_inertia[body_id] = wp.mat33(0.0)
```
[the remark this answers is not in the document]
eric-heiden: Verified against Newton's pinned Warp 1.16: `wp.mat33` is iterable here, so it does not raise `TypeError`. However, the underlying concern is valid because `any(wp.mat33(0.0))` returns `True` from row-object truthiness. Commit 61ae66f4d now checks `np.array(recorded_inertia).any()` instead. Both focused mass-fallback regressions and all pre-commit hooks pass.

eric-heiden:
Also addressed the per-body aggregation nit in 61ae66f4d. Recorded collider paths are now grouped by rigid-body path at record time, so fallback aggregation only visits that body's colliders.

Validation:
- `test_mass_fallback_instanced_collider_massapi_without_body_massapi`: passed
- `test_collider_massapi_without_body_massapi`: passed

eric-heiden:
Resolved the three CPU failures in 1ea7b27c3 after updating this branch from current `main`.

The failures were all the same Robotiq cross-format assertion. Current `main` pins an MJCF where the finger pads are `0.0035` kg, but the pinned USD fixture was generated from the older source and still authors collider mass `2e-6` kg. This PR now correctly honors that authored collider MassAPI value; the previous importer accidentally hid the stale fixture by reconstructing mass from geometry.

The cross-format test now skips only when both USD pad masses resolve to that exact known-stale `2e-6` value. For any other value, its normal MJCF-vs-USD comparison still runs, so regenerating the USD fixture automatically re-enables the assertion.

Validation:
- `test_pad_mass_matches_source_scale`: passes with the narrow known-fixture skip
- `test_import_usd.py`: 265 tests passed

andrewkaufman:
> On the usd-core version used by Newton, ComputeMassProperties still fails to traverse instance-proxy colliders

@eric-heiden can you remind me why that is? Did we ever report the issue upstream?

eric-heiden:
Yes — I reported it upstream as [OpenUSD #4005](https://github.com/PixarAnimationStudios/OpenUSD/issues/4005), with the same instanceable-reference repro.

The cause in Newton's current `usd-core 26.3` is that `UsdPhysicsRigidBodyAPI::ComputeMassProperties` constructs `UsdPrimRange(usdPrim)` with the default traversal predicate. That predicate skips instance proxies, so a collider below the instanceable prim is never visited, the callback is never called, and the method returns `-1`.

OpenUSD fixed it in [b2abc12](https://github.com/PixarAnimationStudios/OpenUSD/commit/b2abc12e1d827c69c8b915e61b9e5134f9f2c132) by changing that range to `UsdPrimRange(usdPrim, UsdTraverseInstanceProxies())`, with an upstream regression test. The fix shipped in OpenUSD 26.05. Newton currently constrains `usd-core>=25.5,<26.5`, and this checkout resolves 26.3, so we still need the importer fallback for the versions Newton supports today.

There is also an older report, [OpenUSD #3845](https://github.com/PixarAnimationStudios/OpenUSD/issues/3845), tracking the same root cause.

eric-heiden:
The GPU benchmark rerun failed on the same two MuJoCo solver-iteration counters, but the results show run-to-run variance rather than a #3594 regression:

- Attempt 1: Allegro `6 -> 8`; Kitchen `6 -> 7`
- Attempt 2: Allegro `7 -> 8`; Kitchen `7 -> 8`

Both the baseline and branch values changed between attempts. These workloads use random actuation, and Kitchen also enables randomized initialization. The PR diff contains only `import_usd.py`, USD importer tests, the stale USD-fixture test guard, and the changelog; it has no solver or benchmark changes.

I am rerunning the full ephemeral-runner workflow once more. No code change is warranted for these unrelated counters.

newton/_src/utils/import_usd.py
```diff
@@ -2505,6 +2505,14 @@ def _resolve_contact_attr(key, _prim=prim):
             prim = stage.GetPrimAtPath(prim_path)
             mass_api = UsdPhysics.MassAPI(prim)
             if not mass_api:
+                descendants = iter(Usd.PrimRange(prim, Usd.TraverseInstanceProxies()))
+                for descendant in descendants:
+                    if descendant != prim and descendant.HasAPI(UsdPhysics.RigidBodyAPI):
+                        descendants.PruneChildren()
+                        continue
+                    if descendant.HasAPI(UsdPhysics.CollisionAPI) and descendant.HasAPI(UsdPhysics.MassAPI):
```
andrewkaufman: why is this gated on `MassAPI`? if the collider doesnt have MassAPI then it has implicit mass from density (which can come from bound material) and still needs to be accumulated

andrewkaufman: Having discussed with AOUSD, it seems the inclusion of disabled colliders in mass calculations is considered a bug in the current mass computer impl. That won't be patched until at least usd-core 26.11 though (november release), so I'm not sure if we should include or exclude them here...

I'll change the converter to stop producing such assets in any case, but existing assets already expose the bug. Do we want to codify the current state via the tests you've added or proactively exclude such colliders from mass accumulation as we believe usd-core 26.11 will do?

andrewkaufman: Traced this. The gate here only decides whether a *no-MassAPI body* enters the ComputeMassProperties fallback path — it is not the only path that accumulates collider mass. A collider with density but no MassAPI (including density from a bound material) is still accumulated through the builder's add_shape_*() path, and that already produces the correct body mass: the existing test_material_density_without_massapi_uses_shape_material passes (250 * 8 = 2000 kg), and a mixed case — no body MassAPI, one collider with MassAPI mass plus a sibling with only material density — correctly reports the sum. So the density-only collider is not dropped.

The narrower question is whether the two accumulation paths (builder vs ComputeMassProperties) can ever disagree for the same body. The gate assumes 'if no collider carries MassAPI, the builder path is authoritative and sufficient.' That held for every case I tested, but it is an implicit assumption. Suggest either (a) a comment documenting that density-only colliders are intentionally handled by the builder path and this gate is only the fallback trigger, or (b) broaden the trigger to any collider so both paths converge. (a) is the lower-risk change for this PR.

newton/_src/utils/import_usd.py
```diff
@@ -3897,22 +3945,61 @@ def _get_collision_mass_information(collider_prim: Usd.Prim):
...
+            volume = float(mass_info.volume)
+            if volume <= 0.0:
+                continue
+            collider_prim = stage.GetPrimAtPath(collider_path)
+            collider_mass_api = UsdPhysics.MassAPI(collider_prim)
+            collider_mass = _mass_api_effective_mass(collider_mass_api) if collider_mass_api else None
+            collider_density = _mass_api_effective_density(collider_mass_api) if collider_mass_api else None
+            density = collider_mass / volume if collider_mass is not None else collider_density
+            if density is None:
+                density = body_density if body_density is not None else shape_density
```
andrewkaufman, line 3978: In _aggregate_recorded_mass_properties, when a collider has neither effective mass nor density, density falls back to body_density then shape_density (the value captured at record time — material density or default). Worth confirming this ordering matches the builder path's precedence (per-shape material density vs body-level density) so the two paths cannot assign different densities to the same collider. Not a confirmed bug — flagging for a precedence-consistency check.

newton/_src/utils/import_usd.py
```diff
@@ -3968,24 +4055,36 @@ def _get_collision_mass_information(collider_prim: Usd.Prim):
...
+                        cmp_mass, recorded_inertia, cmp_com = recorded_properties
+                        builder.body_inertia[body_id] = recorded_inertia
+                        if np.array(recorded_inertia).any():
+                            builder.body_inv_inertia[body_id] = wp.inverse(recorded_inertia)
                         else:
                             builder.body_inv_inertia[body_id] = wp.mat33(0.0)
+                    else:
+                        cmp_mass = builder.body_mass[body_id]
+                        if not has_effective_com:
+                            cmp_com = builder.body_com[body_id]
```
andrewkaufman, line 4085: this block doesnt have test coverage

newton/_src/utils/import_usd.py
```diff
@@ -3968,24 +4055,36 @@ def _get_collision_mass_information(collider_prim: Usd.Prim):
...
+                        body_density = effective_density
+                        if body_density is not None and not has_effective_mass and default_shape_density > 0.0:
+                            density_scale = body_density / default_shape_density
+                            cmp_mass *= density_scale
+                            scaled_inertia = np.array(builder.body_inertia[body_id]) * density_scale
+                            builder.body_inertia[body_id] = wp.mat33(scaled_inertia)
+                            if scaled_inertia.any():
+                                builder.body_inv_inertia[body_id] = wp.inverse(builder.body_inertia[body_id])
+                            else:
+                                builder.body_inv_inertia[body_id] = wp.mat33(0.0)
```
andrewkaufman, line 4098: this block doesnt have test coverage

newton/_src/utils/import_usd.py
```diff
@@ -3897,22 +3945,61 @@ def _get_collision_mass_information(collider_prim: Usd.Prim):
...
+        total_mass = 0.0
+        total_com = wp.vec3(0.0)
+        total_inertia = wp.mat33(0.0)
+        found = False
+        for collider_path in rigid_body_fallback_collider_paths.get(body_path, ()):
+            mass_info = rigid_body_mass_info_map[collider_path]
+            shape_density = rigid_body_mass_fallback_density[collider_path]
+            volume = float(mass_info.volume)
+            if volume <= 0.0:
+                continue
```
andrewkaufman: this block doesnt have test coverage

newton/_src/utils/import_usd.py
andrewkaufman: Re: the disabled-collider mass inclusion question — I'd suggest proactively excluding disabled colliders from mass accumulation now. Reasoning:

1. AOUSD has confirmed the inclusion is a bug in `ComputeMassProperties` (fix targeted for usd-core 26.11). Writing tests that assert disabled colliders *contribute* mass means those tests will break when Newton upgrades to 26.11+.

2. Newton's own fallback path (the new `_aggregate_recorded_mass_properties`) is not bound by usd-core behavior — we control the semantics there. We can implement the correct behavior ahead of the upstream fix without diverging: when `ComputeMassProperties` itself is called, the callback can return `zero_mass_information` for disabled colliders.

3. On the converter side, we're about to stop co-authoring density on explicit-mass geoms (mass XOR density, per MJCF docs). That eliminates the `density=1000` on `mass=0` visual geoms going forward. But existing converted assets still have it, and excluding disabled colliders from mass is the only way those assets import correctly without re-conversion.

For the test coverage gaps at L3959, L4074, L4087 — happy to help sketch fixtures if useful.

eric-heiden: Added a short explanation and a mixed MassAPI + material-density sibling test in 1db1f6e91.

newton/_src/utils/import_usd.py
eric-heiden: Covered by the same test in 1db1f6e91. Zero inertia keeps zero inverse inertia.

newton/_src/utils/import_usd.py
eric-heiden: Confirmed and covered in 1db1f6e91: collider density overrides body density, which overrides material/default density.

newton/_src/utils/import_usd.py
eric-heiden: Updated in 1db1f6e91. Disabled colliders now contribute no mass, COM, or inertia, with visual loading on or off.

newton/_src/utils/import_usd.py
eric-heiden: Removed this unreachable check in 1db1f6e91. Recorded mass information is already guaranteed to have positive volume.

newton/_src/utils/import_usd.py
eric-heiden: Covered in 1db1f6e91 with a density-only body that has no collider. The builder COM is preserved.

andrewkaufman:
Thanks for adding the disabled-collider exclusion — that resolves the AOUSD question from my last pass, and the density-precedence and mixed-source tests are good additions.

There's a bug in `1db1f6e9` though: the new zero-volume shortcut in `_get_collision_mass_information` still leaks mass when a **disabled** collider carries an authored `MassAPI` mass.

```python
def _get_collision_mass_information(collider_prim: Usd.Prim):
    if not _is_enabled_collider(collider_prim):
        return zero_mass_information
```

Returning `volume = 0.0` suppresses the *geometry* contribution, but the collider is still visited by `ComputeMassProperties`, which reads the authored mass override independently of the callback. In `_ParseCollisionShapeForMass`:

```cpp
if (shapeMassInfo.mass > 0.0f)
{
    inertia = inertia * (shapeMassInfo.mass / massInfo.volume);   // volume == 0 -> inf/NaN
}
```

So the authored mass is added to the total anyway, and the inertia scale divides by the zeroed volume.

Repro — one enabled collider with `mass = 0.05`, one disabled collider also carrying `mass = 0.05`, both `size = 0.2` cubes under a body with no `MassAPI`:

```
disabled collider has MassAPI mass=False: mass=0.050000  Ixx=0.00033333   # correct
disabled collider has MassAPI mass=True:  mass=0.100000  Ixx=0.05333334   # leaked
correct answer in both cases:             mass=0.05      Ixx=0.00033333
```

The mass is double what it should be, and `Ixx` is off by 160x. Calling `ComputeMassProperties` directly with a callback that zeroes disabled colliders shows where it comes from:

```
callback visited: ['/World/Body/EnabledMass', '/World/Body/DisabledMass']
ComputeMassProperties -> (0.10000000149011612, Gf.Vec3f(nan, nan, nan), ...)
```

The CoM/principal axes come back NaN. It doesn't surface in the imported body above because the aggregate mass is non-zero and Newton recomputes the CoM, but the mass is already wrong by then, and a NaN CoM reaching `body_com` is one asset away.

The existing test doesn't catch it because in the `"disabled collider mass"` case the disabled collider is the only one with `MassAPI`, so the enabled sibling has no override and the body falls into the shape-density path rather than the aggregation path. Both colliders need `MassAPI` for the divide to trigger.

The fix is to exclude disabled colliders from the *traversal*, not just zero their contribution — filter them before handing the set to `ComputeMassProperties` (or skip them in `_aggregate_recorded_mass_properties`, which already sees per-collider paths) so the authored override is never read. That's the same conclusion AOUSD reached: a disabled collider shouldn't be in the aggregation set at all.

Two smaller things:

**1. The removed `volume <= 0.0` guard in `_aggregate_recorded_mass_properties`.**

```python
volume = float(mass_info.volume)
-   if volume <= 0.0:
-       continue
```

`new_com = (total_com * total_mass + com * mass) / new_mass` divides by `new_mass`, so if every recorded collider for a body ends up zero-volume-and-zero-density this is a divide by zero. I couldn't reach it through `add_usd` — degenerate colliders get dropped earlier by `_build_mass_info_from_shape_geometry`'s positive-mass check, and I get the two expected warnings with a sane result — so this is defensive rather than a live bug. But the guard was cheap and removing it makes the invariant ("recorded colliders always have positive volume") implicit and unstated. If it's intentional, a comment saying why the volume is guaranteed positive here would help.

**2. `shape_density = 0.0` for disabled colliders** in the builder path reads as "this collider is weightless" rather than "this collider is not part of the mass computation." Same observable result today, but if anything later multiplies or divides by that density the intent is lost. A skip would express it better.

Happy to re-review once the traversal exclusion is in. The rest of the change looks right to me.

eric-heiden:
@andrewkaufman Good catch. Fixed in 7087e8395. Newton now bypasses OpenUSD mass aggregation when a disabled collider has MassAPI and uses only recorded enabled colliders. I added the exact two-MassAPI regression for both visual-loading modes, and documented the positive-volume invariant and zero builder density. All 268 USD importer tests and pre-commit pass.

andrewkaufman:
Verified `7087e8395` — the leak is fixed. LGTM.

I re-ran my repro against the new head; both the mass and the inertia are now correct:

```
disabled collider has MassAPI mass=False: mass=0.050000  Ixx=0.00033333
disabled collider has MassAPI mass=True:  mass=0.050000  Ixx=0.00033333   # was 0.100000 / 0.05333334
```

No NaN in `body_com` or `body_inertia` on the direct-aggregation path either.

Routing this through `_mass_computer_requires_recorded_fallback` is the right call — reusing the existing blocked-attribute bypass rather than adding a parallel mechanism keeps it to one "OpenUSD can't aggregate this safely" decision point, and the two workarounds can be retired independently as the usd-core floor moves. Linking the removal condition in the code comment is appreciated.

I probed the cases the narrower `mass`-only condition might have missed, all correct on the new head:

- disabled collider with **density** only, alongside an enabled `MassAPI` collider → `0.05`
- disabled collider with **inertia** only (no mass/density) → `0.05`
- disabled collider with `mass = 0.0` (schema fallback) + enabled with mass → `0.05`
- disabled collider with `MassAPI` and **no enabled collider at all** → `0.0`, no NaN
- the pre-existing `"disabled collider mass"` / `"disabled collider density"` cases → still `8.0`

Also checked nested-body interaction, since the bypass changes which traversal runs: a disabled collider carrying `mass = 99.0` under a *nested* rigid body does not leak into the parent (parent `0.05`, nested `0.0`), so `PruneChildren` still holds on the fallback path.

Negative control on the new regression case: reverting just the `_is_enabled_collider` condition in `_mass_computer_requires_recorded_fallback` makes `"disabled mass with MassAPI sibling"` fail with `0.10000000149011612 != 0.05` in both visual-loading modes. So the test genuinely guards the fix rather than passing incidentally.

`newton.tests -p test_import_usd.py`: 268/268 pass on my end too.

The two documentation follow-ups both address what I raised — the positive-volume invariant is now stated where the guard used to be, and the `shape_density = 0.0` intent is explicit. Good with leaving both as comments.

One note for whoever picks up the removal later, since the code comment links it without much context: <https://github.com/PixarAnimationStudios/OpenUSD/pull/4164> is an OpenUSD PR I have open to fix this on the upstream side (still unmerged). Two things to know about it:

- It excludes disabled colliders from `ComputeMassProperties`, which is the behavior this workaround is emulating.
- It *also* changes which body a collider below a disabled nested rigid body is attributed to during parsing.

So the workaround here can only be dropped once the minimum usd-core actually ships that change, not just when the PR merges. Nothing to do in this PR — just flagging it so the code comment isn't a dead end for whoever revisits it.

andrewkaufman (approved):
Approving. Verified `7087e8395` fixes the mass leak: 0.05 / Ixx 0.00033333 (was 0.100000 / 0.05333334), no NaN. Checked density-only, inertia-only, mass=0.0, and disabled-with-no-enabled-sibling variants, plus nested-body pruning. Negative control confirms the new test catches the regression. 268/268 pass.

johnnynunez:
Thank you guys! You are amazing

PR #3595, opened 2026-07-21
Fix explicit mass on MJCF mesh geoms

### LINKED ISSUE #3592: [BUG] add_mjcf drops positive explicit mass on mesh geoms
### Bug Description

`ModelBuilder.add_mjcf()` discards a positive explicit `mass` authored on an MJCF mesh geom.

When any geom has a `mass` attribute, the importer first sets the shape density to zero so density-based accumulation is skipped. It later reconstructs explicit-mass inertia for sphere, box, cylinder, capsule, and ellipsoid geoms, but mesh geoms fall into an unsupported warning branch. The result is that both mass and inertia are lost.

Minimal result on current `main`:

- MuJoCo body mass: 0.012 kg
- Newton body mass: 0.0 kg
- Newton warning: `explicit mass (0.012) on geom 'visual' with type 'mesh' is not supported — mass will be ignored`
- Replacing the mesh geom with a sphere of the same explicit mass imports correctly at 0.012 kg.

Relevant code:

- https://github.com/newton-physics/newton/blob/ce603b3b271c864070a4ea3b0996bdd2295d801c/newton/_src/utils/import_mjcf.py#L726-L735
- https://github.com/newton-physics/newton/blob/ce603b3b271c864070a4ea3b0996bdd2295d801c/newton/_src/utils/import_mjcf.py#L1134-L1189

Expected behavior: compute unit-density mass, center of mass, and inertia for the loaded mesh, scale them to the authored geom mass, and accumulate the result exactly once.

This is a focused follow-up to the body-mass findings in closed issue #2170. It is distinct from #2494, which concerns MuJoCo LEGACY versus Newton EXACT mesh-volume computation for density-derived inertia.

### Reproduction Script

```python
import tempfile
import warnings
from pathlib import Path

import mujoco
import newton


CUBE_OBJ = """\
v -0.05 -0.05 -0.05
v  0.05 -0.05 -0.05
v  0.05  0.05 -0.05
v -0.05  0.05 -0.05
v -0.05 -0.05  0.05
v  0.05 -0.05  0.05
v  0.05  0.05  0.05
v -0.05  0.05  0.05
f 1 3 2
f 1 4 3
f 5 6 7
f 5 7 8
f 1 2 6
f 1 6 5
f 2 3 7
f 2 7 6
f 3 4 8
f 3 8 7
f 4 1 5
f 4 5 8
"""

MJCF = """\
<mujoco>
  <asset>
    <mesh name="cube" file="cube.obj"/>
  </asset>
  <worldbody>
    <body name="body">
      <freejoint/>
      <geom name="visual" type="mesh" mesh="cube"
            contype="0" conaffinity="0" mass="0.012"/>
      <geom type="box" size=".015 .015 .003" mass="0"/>
      <geom type="box" size=".021 .021 .003" mass="0"/>
    </body>
  </worldbody>
</mujoco>
"""

    root = Path(directory)
    (root / "cube.obj").write_text(CUBE_OBJ)
    (root / "model.xml").write_text(MJCF)

    mj_model = mujoco.MjModel.from_xml_path(str(root / "model.xml"))

    with warnings.catch_warnings(record=True) as caught:
        warnings.simplefilter("always")
        builder.add_mjcf(str(root / "model.xml"))

    print("MuJoCo:", mj_model.body_mass[1])  # 0.012
    print("Newton:", builder.body_mass[0])   # 0.0
    print([str(w.message) for w in caught if "explicit mass" in str(w.message)])
```

### System Information

- Warp: `1.16.0.dev20260716`
- Python: 3.12.3
- OS: Windows

eric-heiden:
## Description

Positive `mass` values on MJCF mesh geoms were ignored. Newton warned that the value was unsupported and left the body without the requested mass.

This change uses the loaded meshes' unit-density mass properties. It chooses a density that makes all submeshes add up to the exact authored mass. The existing shape code then combines mass, center of mass, and inertia normally.

The regression uses the `0.012` kg value from the converter report and also checks COM, inertia, and warnings.

Closes #3592.

## Checklist

## Test plan

- `uv run --extra dev -m newton.tests -k test_explicit_mesh_geom_mass`
- `uv run --extra dev -m newton.tests -p test_import_mjcf.py` (232 tests)
- `uvx pre-commit run -a`

## Bug fix

Before this change, a mesh geom with `mass="0.012"` imported with zero mass. It now imports as `0.012` kg with matching mesh inertia.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed MJCF importing so explicitly authored **positive `mass`** on mesh geoms is no longer ignored, with correct handling of derived mesh density.
  - Prevented spurious “explicit mass” warnings and ensured imported **body mass**, **center of mass**, and **inertia** align with the authored values.
- **Tests**
  - Added a regression test covering explicit positive mass on small mesh geoms (including both `solid` and `hollow` cases), validating warnings and resulting mass properties.

### DIFF
--- a/newton/_src/utils/import_mjcf.py
+++ b/newton/_src/utils/import_mjcf.py
@@ -725,6 +725,7 @@ def parse_mjcf(
 
             geom_density = parse_float(geom_attrib, "density", density)
             geom_mass_explicit = None
+            explicit_mass_handled = False
 
             # MuJoCo: explicit mass attribute (from <geom mass="..."> or class defaults).
             # Skip density-based mass contribution and compute inertia directly from mass.
@@ -987,6 +988,12 @@ def parse_mjcf(
                     override_color=material_color,
                     override_texture=texture,
                 )
+                explicit_mesh_density = None
+                if geom_mass_explicit is not None and geom_mass_explicit > 0.0 and link >= 0:
+                    unit_density_mass = sum(float(m_mesh.mass) for m_mesh in m_meshes)
+                    if unit_density_mass > 1.0e-6:
+                        explicit_mesh_density = geom_mass_explicit / unit_density_mass
+                        explicit_mass_handled = True
                 for m_mesh in m_meshes:
                     if m_mesh.texture is not None and m_mesh.uvs is None:
                         if verbose:
@@ -998,6 +1005,8 @@ def parse_mjcf(
                     mesh_cfg.sdf_max_resolution = None
                     mesh_cfg.sdf_target_voxel_size = None
                     mesh_cfg.sdf_narrow_band_range = (-0.1, 0.1)
+                    if explicit_mesh_density is not None:
+                        mesh_cfg.density = explicit_mesh_density
                     mesh_shape_kwargs["cfg"] = mesh_cfg
                     s = builder.add_shape_mesh(
                         xform=tf,
@@ -1133,7 +1142,7 @@ def parse_mjcf(
 
             # Handle explicit mass: compute inertia using existing functions, add to body.
             # Visual geoms can still contribute authored mass when parse_visuals=True.
-            if geom_mass_explicit is not None and geom_mass_explicit > 0.0 and link >= 0:
+            if geom_mass_explicit is not None and geom_mass_explicit > 0.0 and link >= 0 and not explicit_mass_handled:
                 from ..geometry.inertia import (  # noqa: PLC0415
                     compute_inertia_box_from_mass,
                     compute_inertia_capsule,
--- a/newton/_src/utils/import_mjcf.py
+++ b/newton/_src/utils/import_mjcf.py
@@ -15,7 +15,7 @@ import warp as wp
 
 from ..core import quat_between_axes
 from ..core.types import Axis, AxisType, Sequence, Transform, vec10
-from ..geometry import Mesh, ShapeFlags
+from ..geometry import GeoType, Mesh, ShapeFlags, compute_inertia_shape
 from ..geometry.types import Heightfield
 from ..geometry.utils import compute_aabb, compute_inertia_box_mesh
 from ..sim import JointTargetMode, JointType, ModelBuilder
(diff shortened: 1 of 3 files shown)


eric-heiden:
Addressed the review finding in 7598fcd2b. The importer now accepts every strictly positive computed mesh volume, and the regression uses a scaled cube whose unit-density mass is below the old cutoff. The regression fails before the fix; the full 232-test MJCF importer suite passes after it.

newton/_src/utils/import_mjcf.py
```diff
@@ -987,6 +988,12 @@ def parse_shapes(
                     override_color=material_color,
                     override_texture=texture,
                 )
+                explicit_mesh_density = None
+                if geom_mass_explicit is not None and geom_mass_explicit > 0.0 and link >= 0:
+                    unit_density_mass = sum(float(m_mesh.mass) for m_mesh in m_meshes)
```
vreutskyy: Could we normalize using the unit-density mass that `add_shape_mesh()` will actually compute? `m_mesh.mass` assumes a solid mesh, while `mesh_cfg` inherits `default_shape_cfg.is_solid` and `margin`. With `is_solid=False` and `margin=0.01`, I get `0.00144` kg instead of the authored `0.012` kg (and zero mass with zero margin). Would using the same solid/hollow configuration for the normalization, plus a regression for this case, make sense?

vreutskyy (commented):
One question about the explicit mesh-mass normalization.

newton/_src/utils/import_mjcf.py
eric-heiden: Good catch. Confirmed with a hollow-mesh regression: before the fix, authored `0.012` kg imported as `0.0288` kg. Commit 320f22943 now derives the normalization mass with `compute_inertia_shape()` using the same mesh, unit scale, `is_solid`, and margin passed to `add_shape_mesh()`. The regression now covers both solid and hollow configurations.

Validation:
- targeted `test_explicit_small_mesh_geom_mass`: passed
- full `test_import_mjcf.py`: 232 tests passed

vreutskyy (approved):

PR #3596, opened 2026-07-21
Honor MJCF geom inertia selection

### LINKED ISSUE #3593: [BUG] add_mjcf does not honor geom-based inertia selection semantics
### Bug Description

MJCF body mass inference in `ModelBuilder.add_mjcf()` is coupled to which Newton shapes are instantiated instead of following MuJoCo's compiler semantics.

Three independently reproducible symptoms on current `main` share this root cause:

| Repro | MuJoCo body mass | Newton body mass |
|---|---:|---:|
| Mass-providing visual geom with `parse_visuals=False` | 0.012 | 0.0 |
| `inertiagrouprange="0 0"` excludes an 8 kg group-1 box | 0.5 | 8.5 |
| `inertiafromgeom="true"` should override a 1 kg `<inertial>` with a 2 kg geom | 2.0 | 1.0 |

MuJoCo selects mass-contributing geoms using `compiler/inertiafromgeom` and `compiler/inertiagrouprange`. Collision masks and whether a caller requests visual shape loading do not decide the compiled body inertial.

Newton currently partitions geoms into visual/collider lists before shape insertion, and shape insertion performs the mass accumulation:

- https://github.com/newton-physics/newton/blob/ce603b3b271c864070a4ea3b0996bdd2295d801c/newton/_src/utils/import_mjcf.py#L1304-L1389
- https://github.com/newton-physics/newton/blob/ce603b3b271c864070a4ea3b0996bdd2295d801c/newton/_src/utils/import_mjcf.py#L1926-L1989

The importer does not currently resolve `inertiafromgeom` or `inertiagrouprange`, and an authored `<inertial>` always overwrites geom-derived values.

Expected behavior:

- Determine mass-contributing geoms independently of visual/collision shape loading.
- Honor `compiler/inertiafromgeom` (`false`, `true`, and `auto`).
- Honor `compiler/inertiagrouprange`.
- Keep `parse_visuals` as a shape-loading/presentation option that does not silently change source dynamics.

This is a focused follow-up to the MJCF body-mass findings in closed issue #2170.

### Reproduction Script

```python
import mujoco
import newton


def compare(name, mjcf, **newton_options):
    mj_model = mujoco.MjModel.from_xml_string(mjcf)
    builder.add_mjcf(mjcf, **newton_options)
    print(name, "MuJoCo:", mj_model.body_mass[1], "Newton:", builder.body_mass[0])


compare(
    "parse_visuals",
    """
    <mujoco>
      <worldbody>
        <body>
          <freejoint/>
          <geom type="sphere" size=".05"
                contype="0" conaffinity="0" mass=".012"/>
          <geom type="box" size=".015 .015 .003" mass="0"/>
        </body>
      </worldbody>
    </mujoco>
    """,
    parse_visuals=False,
)  # expected 0.012; actual 0.0

compare(
    "inertiagrouprange",
    """
    <mujoco>
      <compiler inertiagrouprange="0 0"/>
      <worldbody>
        <body>
          <freejoint/>
          <geom type="sphere" size=".05" group="0" mass=".5"/>
          <geom type="box" size=".1 .1 .1" group="1" density="1000"/>
        </body>
      </worldbody>
    </mujoco>
    """,
)  # expected 0.5; actual 8.5

compare(
    "inertiafromgeom",
    """
    <mujoco>
      <compiler inertiafromgeom="true"/>
      <worldbody>
        <body>
          <freejoint/>
          <inertial mass="1" pos="0 0 0" diaginertia=".1 .1 .1"/>
          <geom type="sphere" size=".05" mass="2"/>
        </body>
      </worldbody>
    </mujoco>
    """,
)  # expected 2.0; actual 1.0
```

### System Information

- Warp: `1.16.0.dev20260716`
- Python: 3.12.3
- OS: Windows

eric-heiden:
## Description

Newton's MJCF importer did not follow MuJoCo's compiler rules for choosing body inertia.

This change adds support for:

- `inertiafromgeom="auto"`, `"true"`, and `"false"`
- `inertiagrouprange`
- inertia from visual geoms even when `parse_visuals=False`

Loaded shapes still use the normal mass path. Visual geoms that are intentionally not loaded are evaluated in a temporary builder, so they affect body mass and inertia without appearing in the model.

Closes #3593.

## Checklist

## Test plan

- `uv run --extra dev -m newton.tests -k test_visual_geom_mass_without_parsing_visuals`
- `uv run --extra dev -m newton.tests -k test_compiler_inertiagrouprange`
- `uv run --extra dev -m newton.tests -k test_compiler_inertiafromgeom_modes`
- `uv run --extra dev -m newton.tests -p test_import_mjcf.py` (234 tests)
- `uvx pre-commit run -a`

## Bug fix

Before this change, `parse_visuals=False` could change body mass, excluded geom groups still contributed inertia, and `inertiafromgeom="true"` did not override explicit inertial data. These cases now follow MuJoCo's compiler semantics.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the MJCF importer’s inertia/mass inference to properly honor `<compiler inertiafromgeom>` modes and `<compiler inertiagrouprange>` filtering.
  * Ensures inferred body mass/inertia remains consistent regardless of whether visuals are parsed as shapes.
  * Correctly applies explicit `mass` for small mesh geoms without spurious warnings.
  * Enforces expected behavior when inference is disallowed but `<inertial>` data is missing.
* **Tests**
  * Added regression tests for mesh explicit-mass handling and compiler-driven inertia/mass semantics.
* **Documentation**
  * Refreshed the unreleased changelog notes describing the fix.

### DIFF
--- a/newton/_src/utils/import_mjcf.py
+++ b/newton/_src/utils/import_mjcf.py
@@ -411,6 +411,15 @@ def parse_mjcf(
         texture_dir = "."
         fitaabb = False
 
+    inertia_from_geom = compiler_attribs.get("inertiafromgeom", "auto").lower()
+    if inertia_from_geom not in {"auto", "false", "true"}:
+        raise ValueError(
+            f"MJCF compiler inertiafromgeom must be 'auto', 'false', or 'true'; got {inertia_from_geom!r}."
+        )
+    inertia_group_range = tuple(int(value) for value in compiler_attribs.get("inertiagrouprange", "0 5").split())
+    if len(inertia_group_range) != 2:
+        raise ValueError("MJCF compiler inertiagrouprange must contain exactly 2 integers.")
+
     # Parse MJCF compiler and option tags for ONCE and WORLD frequency custom attributes
     # WORLD frequency attributes use index 0 here; they get remapped during add_world()
     # Use findall for <option> to handle multiple elements after include expansion
@@ -687,8 +696,19 @@ def parse_mjcf(
         return wp.quat_identity()
 
     def parse_shapes(
-        defaults, body_name, link, geoms, density, visible=True, just_visual=False, incoming_xform=None, label_prefix=""
+        defaults,
+        body_name,
+        link,
+        geoms,
+        density,
+        visible=True,
+        just_visual=False,
+        incoming_xform=None,
+        label_prefix="",
+        contribute_inertia=True,
+        target_builder=None,
     ):
+        shape_builder = builder if target_builder is None else target_builder
         shapes = []
         for geo_count, geom in enumerate(geoms):
             geom_class = geom.attrib.get("class")
@@ -734,6 +754,11 @@ def parse_mjcf(
                 # Set density to 0 to skip density-based mass contribution
                 # We'll add the explicit mass to the body separately
                 geom_density = 0.0
+
+            geom_group = int(geom_attrib.get("group", 0))
+            if not contribute_inertia or not (inertia_group_range[0] <= geom_group <= inertia_group_range[1]):
+                geom_density = 0.0
+                geom_mass_explicit = None
 
             shape_cfg = builder.default_shape_cfg.copy()
             shape_cfg.is_visible = visible
@@ -796,8 +821,12 @@ def parse_mjcf(
             if "gap" in geom_attrib:
                 shape_cfg.gap = mj_gap
 
-            custom_attributes = parse_custom_attributes(geom_attrib, builder_custom_attr_shape, parsing_mode="mjcf")
-            if has_solref_mode:
+            custom_attributes = (
+                parse_custom_attributes(geom_attrib, builder_custom_attr_shape, parsing_mode="mjcf")
+                if shape_builder is builder
+                else {}
+            )
+            if has_solref_mode and shape_builder is builder:
                 # Authored solref → RAW (forwarded verbatim); unauthored →
                 # MJCF_DEFAULT (force-space scaling is strictly opt-in for
                 # shapes — no auto-promote, unlike joint limits). See
@@ -946,7 +975,7 @@ def parse_mjcf(
                             tf = tf * wp.transform(center_offset, fit_rot)
 
             if geom_type == "sphere":
-                s = builder.add_shape_sphere(
+                s = shape_builder.add_shape_sphere(
                     xform=tf,
                     radius=geom_size[0],
                     **shape_kwargs,
@@ -954,7 +983,7 @@ def parse_mjcf(
                 shapes.append(s)
 
             elif geom_type == "box":
-                s = builder.add_shape_box(
+                s = shape_builder.add_shape_box(
                     xform=tf,
                     hx=geom_size[0],
                     hy=geom_size[1],
@@ -1018,7 +1047,7 @@ def parse_mjcf(
                     if explicit_mesh_density is not None:
                         mesh_cfg.density = explicit_mesh_density
                     mesh_shape_kwargs["cfg"] = mesh_cfg
-                    s = builder.add_shape_mesh(
+                    s = shape_builder.add_shape_mesh(
                         xform=tf,
                         mesh=m_mesh,
                         **mesh_shape_kwargs,
@@ -1064,7 +1093,7 @@ def parse_mjcf(
                     geom_height = geom_size[1]
 
                 if geom_type == "cylinder":
-                    s = builder.add_shape_cylinder(
+                    s = shape_builder.add_shape_cylinder(
                         xform=tf,
                         radius=geom_radius,
                         half_height=geom_height,
@@ -1072,7 +1101,7 @@ def parse_mjcf(
                     )
                     shapes.append(s)
                 else:
-                    s = builder.add_shape_capsule(
+                    s = shape_builder.add_shape_capsule(
                         xform=tf,
                         radius=geom_radius,
                         half_height=geom_height,
@@ -1117,7 +1146,7 @@ def parse_mjcf(
 
                 # Heightfields are always static — don't pass body from shape_kwargs
                 hfield_kwargs = {k: v for k, v in shape_kwargs.items() if k != "body"}
-                s = builder.add_shape_heightfield(
+                s = shape_builder.add_shape_heightfield(
                     xform=tf,
                     heightfield=heightfield,
                     **hfield_kwargs,
@@ -1128,7 +1157,7 @@ def parse_mjcf(
                 # Use xform directly - plane has local normal (0,0,1) and passes through origin
                 # The transform tf positions and orients the plane in world space
                 # MuJoCo planes are always infinite for collision; pass 0 extents.
-                s = builder.add_shape_plane(
+                s = shape_builder.add_shape_plane(
                     xform=tf,
                     width=0.0,
                     length=0.0,
@@ -1137,7 +1166,7 @@ def parse_mjcf(
                 shapes.append(s)
 
             elif geom_type == "ellipsoid":
-                s = builder.add_shape_ellipsoid(
+                s = shape_builder.add_shape_ellipsoid(
                     xform=tf,
                     rx=geom_size[0],
                     ry=geom_size[1],
@@ -1203,11 +1232,39 @@ def parse_mjcf(
                     )
 
                 # Add explicit mass and computed inertia to body (skip if inertia is locked by <inertial>)
-                if inertia_computed and not builder.body_lock_inertia[link]:
+                if inertia_computed and not shape_builder.body_lock_inertia[link]:
                     com_body = wp.transform_point(tf, com)
-                    builder._update_body_mass(link, geom_mass_explicit, inertia_tensor, com_body, tf.q)
+                    shape_builder._update_body_mass(link, geom_mass_explicit, inertia_tensor, com_body, tf.q)
 
         return shapes
+
+    def accumulate_inertia_from_unloaded_geoms(defaults, body_name, link, geoms, incoming_xform=None, label_prefix=""):
+        """Accumulate selected geom inertia without retaining visual shapes."""
+        if link < 0 or not geoms:
+            return
+        inertia_builder = ModelBuilder()
+        inertia_link = inertia_builder.add_link()
+        parse_shapes(
+            defaults,
+            body_name,
+            inertia_link,
+            geoms,
+            density=default_shape_density,
+            just_visual=True,
+            visible=False,
+            incoming_xform=incoming_xform,
+            label_prefix=label_prefix,
+            target_builder=inertia_builder,
+        )
+        mass = inertia_builder.body_mass[inertia_link]
+        if mass > 0.0:
+            builder._update_body_mass(
+                link,
+                mass,
+                inertia_builder.body_inertia[inertia_link],
+                inertia_builder.body_com[inertia_link],
+                wp.quat_identity(),
+            )
 
     def _parse_sites_impl(defaults, body_name, link, sites, incoming_xform=None, label_prefix=""):
         """Parse site elements from MJCF."""
@@ -1303,6 +1360,7 @@ def parse_mjcf(
         link: int,
         incoming_xform: wp.transform | None = None,
         label_prefix: str = "",
+        infer_inertia_from_geoms: bool = False,
     ) -> list:
         """Process geoms for a body, partitioning into visuals and colliders.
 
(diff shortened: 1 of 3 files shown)


newton/_src/utils/import_mjcf.py
```diff
@@ -1521,6 +1601,10 @@ def parse_body(
         body_attrib = merge_attrib(defaults.get("body", {}), body.attrib)
         body_name = body_attrib.get("name", f"body_{builder.body_count}")
         body_name = sanitize_name(body_name)
+        has_inertial_definition = body.find("inertial") is not None
+        infer_body_inertia_from_geoms = ignore_inertial_definitions or inertia_from_geom == "true"
+        if inertia_from_geom == "auto" and not has_inertial_definition:
+            infer_body_inertia_from_geoms = True
```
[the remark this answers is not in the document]
eric-heiden: Confirmed against MuJoCo’s XML reference and fixed in 5595d6aa3. The new regression fails before the change and passes after it; the full 234-test MJCF importer suite also passes locally.

eric-heiden:
CI failure analysis: every failing job reaches the same warning for explicit mass on the `camera_mount_geom_0_visual` mesh. This PR now evaluates visual geoms for inertia, so it exposes the existing mesh-mass bug fixed by #3595. I am keeping this PR focused instead of copying that fix here. Once #3595 lands, rebasing this branch onto `main` should clear these failures.

eric-heiden:
Addressed the fixture naming nit in d719ab264. The change only renames the local test variable to use the MJCF prefix. Validation: the targeted `test_compiler_inertiafromgeom_modes` test passes, and `uvx pre-commit run -a` passes.

eric-heiden:
To unblock CI while #3595 awaits review, this branch now includes #3595 as an explicit temporary dependency in 634f4a9a7. The mesh-mass fix was not reimplemented here. After #3595 merges, I will update this branch from `main` and remove the temporary dependency so the final diff remains focused.

Local validation on the combined branch:
- `test_menagerie_mujoco.TestMenagerie_RobotstudioSo101`: 3 tests passed
- `test_import_mjcf.py`: 235 tests passed

eric-heiden:
Addressed the inertia/COM coverage feedback in 23426d3aa. The test now gives the authored inertial and geom different COM and inertia values, then checks mass, COM, and the full inertia tensor for `auto`, `false`, and `true` modes.

Validation:
- targeted `test_compiler_inertiafromgeom_modes`: passed
- full `test_import_mjcf.py`: 235 tests passed

eric-heiden:
PR #3595 has merged, so I rebuilt this branch on current upstream `main` and removed the temporary dependency merge. The final diff now contains only the `inertiafromgeom` / `inertiagrouprange` importer changes, their regression tests, and the `#3596` changelog entry.

Validation after rebuilding:
- `uv run --extra dev -m newton.tests -p test_import_mjcf.py` — 235 tests passed
- `uvx --from pre-commit==4.5.0 pre-commit run -a` — all hooks passed

The explicit pre-commit version is only for local compatibility with Git 2.30; no project dependency changed.

adenzler-nvidia (commented):
Implementation looks good — the semantics match MuJoCo's compiler docs (`inertiafromgeom` modes, inclusive `inertiagrouprange`, compile error on `false` + missing `<inertial>`), the group-range zeroing happens before every consumer including the mesh path, and the scratch-builder transfer math checks out. Just a few nitpicks inline, mostly about hardening the tests around the new machinery.

newton/tests/test_import_mjcf.py
```diff
@@ -2849,6 +2849,97 @@ def test_visual_geom_explicit_mass_with_parse_visuals(self):
             msg="Visual geom with explicit mass should contribute non-zero inertia",
         )
 
+    def test_visual_geom_mass_without_parsing_visuals(self):
```
adenzler-nvidia: All three new tests put a single massive geom into an empty body, so `new_com == p` in `_update_body_mass` and the parallel-axis/rotation terms are identically zero — a sign or frame error in the scratch-builder transfer would pass every test here. Could you add a case with an off-origin, rotated visual plus a massive collider elsewhere (so the combined COM differs from both geom COMs), importing with `parse_visuals=True` and `False` and asserting `body_mass`, `body_com`, and the full `body_inertia` match between the two runs?

newton/_src/utils/import_mjcf.py
```diff
@@ -1371,7 +1430,10 @@ def _process_body_geoms(
 
         visual_shape_indices = []
 
+        unloaded_geoms = []
         if parse_visuals_as_colliders:
+            loaded_geom_ids = {id(geom) for geom in visuals}
+            unloaded_geoms = [geom for geom in colliders if id(geom) not in loaded_geom_ids]
```
adenzler-nvidia, line 1434: This branch is untested — nothing under `newton/tests/` uses `parse_visuals_as_colliders` at all. It's the mirror image of the tested case (here the collision geoms go through the scratch builder), so it deserves the same mass/COM/inertia equality check against a `parse_visuals=True` reference import.

newton/tests/test_import_mjcf.py
```diff
@@ -2849,6 +2849,97 @@ def test_visual_geom_explicit_mass_with_parse_visuals(self):
...
+        """Test that skipping visual shapes does not change inferred body mass."""
+        mjcf = """<?xml version="1.0" ?>
+<mujoco>
+  <worldbody>
+    <body name="test">
+      <freejoint/>
+      <geom name="visual" class="visual" type="sphere" size="0.1"
+            contype="0" conaffinity="0" group="2" mass="0.012"/>
```
adenzler-nvidia: The unloaded-geom path is only exercised with explicit `mass=`, but the mechanism behind #3593 is equally the density path (a visual geom without `mass` picking up the default density inside the scratch builder). Worth a density-only variant asserting equal mass/inertia for `parse_visuals=True` vs `False`.

newton/_src/utils/import_mjcf.py
```diff
@@ -1203,12 +1232,40 @@ def parse_shapes(
...
-                if inertia_computed and not builder.body_lock_inertia[link]:
+                if inertia_computed and not shape_builder.body_lock_inertia[link]:
                     com_body = wp.transform_point(tf, com)
-                    builder._update_body_mass(link, geom_mass_explicit, inertia_tensor, com_body, tf.q)
+                    shape_builder._update_body_mass(link, geom_mass_explicit, inertia_tensor, com_body, tf.q)
 
         return shapes
 
+    def accumulate_inertia_from_unloaded_geoms(defaults, body_name, link, geoms, incoming_xform=None, label_prefix=""):
+        """Accumulate selected geom inertia without retaining visual shapes."""
```
adenzler-nvidia: The unloaded geoms aren't always visual: in the `parse_visuals_as_colliders` branch they are the displaced collision geoms. Maybe "Accumulate inertia from geoms not loaded as shapes, via a scratch builder."

newton/tests/test_import_mjcf.py
```diff
@@ -2849,6 +2849,97 @@ def test_visual_geom_explicit_mass_with_parse_visuals(self):
...
+        self.assertAlmostEqual(builder.body_mass[0], 0.5, places=6)
+        np.testing.assert_allclose(
+            np.array(builder.body_inertia[0]).reshape(3, 3),
+            np.diag([0.002, 0.002, 0.002]),
+            atol=1e-7,
+            rtol=1e-6,
+        )
+
+    def test_compiler_inertiafromgeom_modes(self):
+        """Test compiler control over geom inference when inertial data exists."""
```
adenzler-nvidia: The second half of the test asserts the missing-`<inertial>` error, exactly the case where inertial data does not exist. Maybe "Test inertiafromgeom modes with and without an <inertial> element."

newton/_src/utils/import_mjcf.py
```diff
@@ -735,6 +755,11 @@ def parse_shapes(
                 # We'll add the explicit mass to the body separately
```
adenzler-nvidia: nit: this promise is now conditional — the group-range check below can null `geom_mass_explicit` three lines later. The block comment above already covers the intent, so these two lines could go.

newton/_src/utils/import_mjcf.py
```diff
@@ -1943,7 +2038,7 @@ def parse_body(
                 )
 
         m = builder.body_mass[link]
```
adenzler-nvidia: nit: this read is now dead — `m` is only used after being reassigned inside the branch below.

newton/_src/utils/import_mjcf.py
```diff
@@ -1203,12 +1232,40 @@ def parse_shapes(
...
+            density=default_shape_density,
+            just_visual=True,
+            visible=False,
+            incoming_xform=incoming_xform,
+            label_prefix=label_prefix,
+            target_builder=inertia_builder,
+        )
+        mass = inertia_builder.body_mass[inertia_link]
+        if mass > 0.0:
+            builder._update_body_mass(
```
adenzler-nvidia, line 1259: nit: unlike the explicit-mass path in `parse_shapes`, this skips the `body_lock_inertia` check. It's redundant today (this helper only runs when inference is on, which also skips the `<inertial>` block that sets the lock), but a guard here would keep future call sites safe.

newton/_src/utils/import_mjcf.py
```diff
@@ -1540,6 +1620,14 @@ def parse_body(
         body_attrib = merge_attrib(defaults.get("body", {}), body.attrib)
         body_name = body_attrib.get("name", f"body_{builder.body_count}")
         body_name = sanitize_name(body_name)
+        has_inertial_definition = body.find("inertial") is not None
+        if inertia_from_geom == "false" and not ignore_inertial_definitions and not has_inertial_definition:
+            raise ValueError(
+                f"MJCF body '{body_name}' requires an <inertial> element when compiler inertiafromgeom=\"false\"."
+            )
```
vreutskyy: This unconditional guard rejects every body without an `<inertial>` element when `inertiafromgeom="false"`. MuJoCo only validates mass and inertia for bodies that have joints; fixed bodies are valid with zero mass. I reproduced this for both a body fixed to the world and a fixed child of a moving body: MuJoCo 3.10 accepts them, while this PR raises `ValueError`.

vreutskyy (commented):
One concern:

newton/_src/utils/import_mjcf.py
eric-heiden: Addressed in 710516fbe. The same regression now includes `parse_visuals_as_colliders=True` and compares its mass, COM, and full inertia tensor against the `parse_visuals=True` reference.

newton/tests/test_import_mjcf.py
eric-heiden: Addressed in 710516fbe. `test_geom_inertia_independent_of_visual_loading` now combines an off-origin rotated visual box with a rotated collider at a different position. It verifies that the combined COM differs from both geom COMs and compares mass, COM, and the full inertia tensor between `parse_visuals=True` and `parse_visuals=False`.

newton/tests/test_import_mjcf.py
eric-heiden: Addressed in 710516fbe. The regression now runs both explicit-mass and density-only cases across all three visual-loading modes and compares mass, COM, and full inertia.

newton/_src/utils/import_mjcf.py
eric-heiden: Confirmed with MuJoCo 3.10 and addressed in 710516fbe. The missing-inertial guard now applies only to bodies with an authored `joint` or `freejoint`. The regression covers a fixed body attached to the world and a fixed child under a moving body, while retaining the error for a moving body without `<inertial>`. The focused tests and all 235 MJCF importer tests pass locally.

newton/tests/test_import_mjcf.py
eric-heiden: Addressed in 710516fbe. The test docstring now describes `inertiafromgeom` modes with and without an inertial element.

newton/_src/utils/import_mjcf.py
eric-heiden: Addressed in 710516fbe. I removed the dead `builder.body_mass[link]` read.

newton/_src/utils/import_mjcf.py
eric-heiden: Addressed in 710516fbe. The docstring now reads: `Accumulate inertia from geoms not loaded as shapes, via a scratch builder.`

newton/_src/utils/import_mjcf.py
eric-heiden: Addressed in 710516fbe. I removed the two stale comments; the existing block comment remains as the accurate description of the explicit-mass path.

newton/_src/utils/import_mjcf.py
eric-heiden: Addressed in 710516fbe. The scratch-builder transfer now checks `not builder.body_lock_inertia[link]` before updating body mass properties.

vreutskyy (approved):

Issue #3548, opened 2026-07-16
[REQ] implement controllers module


jeff-hough:
### Description

Implement generic `Controller` class into `newton`, including the following implementations:
* Differential Drive
* Ackermann
* Holonomic
* Operational Space
* Differential Inverse Kinematics
* Joint Impedance

Tests/Examples/Docs also needed.

### Motivation / Use Case

Centralizes controller implementations into one module, making sure they are fully CUDA-graphable and vectorized. This allows python applications (ex: Isaac Sim/Lab) to use the same controller implementations. Through warp-APIC, it lets a user run the same code on their real robot.

adenzler-nvidia:
Just as a heads-up, code freeze for 1.5 is august 4 so not a lot of time left for that milestone. Is 1.5 a hard deadline? if so, we should prioritize this and make sure we prototype/test the c++ infra side early. Maybe it makes sense to work incrementally in multiple PRs.

jeff-hough:
Having the entire migration complete for 1.5 is not a hard deadline. I would like to aim to have all of the manipulator controllers (the ones that currently exist in Isaac Lab) ported for 1.5, and then a complete migration could be pushed to 1.6, which still aligns to having a full integration of newton controllers into Isaac Sim 6.2 (which will be released in November I believe).

I completely agree with multiple PRs. I had been planning to do it like this:
- **First PR**: Controller API + one controller implementation, including an example. I think having a concrete controller to look at will help people review the API.
- From then on: **One PR per controller**: includes the implementation, tests, and an example. Prioritizing the manipulator controllers, as I mentioned above.
- PR to add a tutorial on controllers.

All of this work should be flagged as experimental to start, in case we decide on a change in API later on.

Do you agree with that plan @adenzler-nvidia ?

adenzler-nvidia:
Sounds reasonable to me - an experimental tag definitely makes sense. Looking forward to this! maybe some of the policy examples can be replaced by the controller examples, instead of adding completely new ones. Let's see :)

PR #3598, opened 2026-07-22
Adds `controllers` module and `ControllerJointImpendance` implementation.

jeff-hough:
## Description

Related issue: #3548 

This PR adds the base `Controller` API, and the `ControllerJointImpedance` (both newton model-based and model-free) as a concrete controller implementation.

Follow-up PRs will add more derived `Controller` classes, as per the referenced issue.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a new `newton.controllers` public API with a `Controller` base class and GPU-accelerated joint impedance controllers (model-based and model-free).
  * Supports stiffness/damping, optional gravity/Coriolis compensation, inertia decoupling, optional acceleration feedforward, and heterogeneous multi-DOF mappings.
* **Documentation**
  * Added Sphinx API docs for the new controllers and a heterogeneous joint impedance example.
* **Bug Fixes / Changes**
  * Updated the controller step interface to run exactly one compute step using provided inputs/outputs and a `dt` parameter.
* **Tests**
  * Added unit tests for both controllers and an example integration test.

### DIFF
--- a/docs/api/newton_controllers.rst
+++ b/docs/api/newton_controllers.rst
@@ -1,0 +1,31 @@
+.. SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+.. SPDX-License-Identifier: CC-BY-4.0
+
+newton.controllers
+==================
+
+GPU-accelerated, vectorized control laws for Newton physics simulations.
+
+This module provides standalone controllers that compute joint torques or
+other actuation signals from simulation state. Each controller is a concrete
+subclass of :class:`Controller` and operates on flat 1D arrays matching the
+layout of :class:`~newton.State` fields, making them composable with any
+Newton solver.
+
+.. experimental::
+
+    The controllers API may change without prior notice. Feedback is welcome —
+    please file issues or discussion threads.
+
+.. py:module:: newton.controllers
+.. currentmodule:: newton.controllers
+
+.. rubric:: Classes
+
+.. autosummary::
+   :toctree: _generated
+   :nosignatures:
+
+   Controller
+   ControllerJointImpedance
+   ControllerJointImpedanceModelFree
--- a/docs/api/newton_controllers.rst
+++ b/docs/api/newton_controllers.rst
@@ -4,13 +4,11 @@
 newton.controllers
 ==================
 
-GPU-accelerated, vectorized control laws for Newton physics simulations.
+GPU-accelerated, vectorized control laws.
 
-This module provides standalone controllers that compute joint torques or
-other actuation signals from simulation state. Each controller is a concrete
-subclass of :class:`ControllerBase` and operates on flat 1D arrays matching the
-layout of :class:`~newton.State` fields, making them composable with any
-Newton solver.
+This module provides standalone controllers which compute output torques, positions
+or velocities for the robot to track. Each controller is a concrete
+subclass of :class:`ControllerBase` and operates on flat 1D arrays.
 
 .. experimental::
 
--- a/docs/api/newton_controllers.rst
+++ b/docs/api/newton_controllers.rst
@@ -12,9 +12,6 @@ subclass of :class:`ControllerBase` and operates on flat 1D arrays.
 
 .. experimental::
 
-    The controllers API may change without prior notice. Feedback is welcome —
-    please file issues or discussion threads.
-
 .. py:module:: newton.controllers
 .. currentmodule:: newton.controllers
 
--- a/docs/api/newton_controllers.rst
+++ b/docs/api/newton_controllers.rst
@@ -6,7 +6,7 @@ newton.controllers
 
 GPU-accelerated, vectorized control laws.
 
-This module provides standalone controllers which compute control signals
+This module provides standalone controllers that compute signals
 for the robot to track. Each controller is a concrete
 subclass of :class:`ControllerBase`.
 
--- a/docs/api/newton_controllers.rst
+++ b/docs/api/newton_controllers.rst
@@ -6,9 +6,9 @@ newton.controllers
 
 GPU-accelerated, vectorized control laws.
 
-This module provides standalone controllers which compute output torques, positions
-or velocities for the robot to track. Each controller is a concrete
-subclass of :class:`ControllerBase` and operates on flat 1D arrays.
+This module provides standalone controllers which compute control signals
+for the robot to track. Each controller is a concrete
+subclass of :class:`ControllerBase`.
 
 .. experimental::
 
--- a/docs/api/newton_controllers.rst
+++ b/docs/api/newton_controllers.rst
@@ -8,7 +8,7 @@ GPU-accelerated, vectorized control laws for Newton physics simulations.
 
 This module provides standalone controllers that compute joint torques or
 other actuation signals from simulation state. Each controller is a concrete
-subclass of :class:`Controller` and operates on flat 1D arrays matching the
+subclass of :class:`ControllerBase` and operates on flat 1D arrays matching the
 layout of :class:`~newton.State` fields, making them composable with any
 Newton solver.
 
@@ -26,6 +26,6 @@ Newton solver.
    :toctree: _generated
    :nosignatures:
 
-   Controller
+   ControllerBase
    ControllerJointImpedance
    ControllerJointImpedanceModelFree
--- a/newton/_src/controllers/controller.py
+++ b/newton/_src/controllers/controller.py
@@ -22,8 +22,9 @@ class Controller(ABC):
 
     Subclasses are responsible for:
 
-    - :meth:`is_graphable`, :meth:`is_stateful`: predicates the user can query
-      to decide CUDA-graph capture and double-buffered state setup.
+    - :meth:`is_graphable`: predicate the user can query to decide CUDA-graph
+      capture. Check ``ctrl.state() is not None`` to determine whether
+      double-buffered state setup is needed.
     - :meth:`state`: allocate a fresh :class:`State`, or return ``None`` for
       stateless laws.
     - :meth:`input`, :meth:`output`: allocate fresh duck-typed containers
@@ -44,12 +45,14 @@ class Controller(ABC):
         """Whether :meth:`compute` is safe to capture in a CUDA graph."""
 
     @abstractmethod
-    def is_stateful(self) -> bool:
-        """Whether this controller maintains internal state between steps."""
-
-    @abstractmethod
     def state(self) -> Controller.State | None:
-        """Allocate a fresh :class:`State`, or ``None`` if stateless."""
+        """Allocate a fresh :class:`State`, or ``None`` if stateless.
+
+        The return value is the canonical indicator of statefulness:
+        ``ctrl.state() is not None`` means the controller maintains state
+        between steps and callers should allocate two buffers for
+        double-buffering.
+        """
 
     @abstractmethod
     def input(self) -> Any:
--- a/newton/_src/controllers/controller.py
+++ b/newton/_src/controllers/controller.py
@@ -1,0 +1,93 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
+
+"""Abstract base for Newton controllers."""
+
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from typing import Any
+
+import warp as wp
+
+
+class Controller(ABC):
+    """Abstract interface for a single Newton control law.
+
+    Every concrete control law (joint impedance, differential IK, …) subclasses
+    :class:`Controller` directly. There is no framework-level composition: users
+    who want to combine multiple control laws call each one's :meth:`compute`
+    in sequence themselves.
+
+    Subclasses are responsible for:
+
+    - :meth:`is_graphable`, :meth:`is_stateful`: predicates the user can query
+      to decide CUDA-graph capture and double-buffered state setup.
+    - :meth:`state`: allocate a fresh :class:`State`, or return ``None`` for
+      stateless laws.
+    - :meth:`input`, :meth:`output`: allocate fresh duck-typed containers
+      holding only the live ports declared at construction. Baked-in arrays
+      (gain ports given a :class:`wp.array` rather than an attr name) are
+      stored on the controller and do **not** appear on the input struct.
+    - :meth:`compute`: read the input struct's live arrays, run kernels, write
+      the output struct's live arrays. Writes are slot-replacing (``=``, not
+      ``+=``); composing laws is the user's job.
+    """
+
+    @dataclass
+    class State:
+        """Pure data container. Subclasses declare their fields."""
+
+    @abstractmethod
+    def is_graphable(self) -> bool:
+        """Whether :meth:`compute` is safe to capture in a CUDA graph."""
+
+    @abstractmethod
+    def is_stateful(self) -> bool:
+        """Whether this controller maintains internal state between steps."""
+
+    @abstractmethod
+    def state(self) -> Controller.State | None:
+        """Allocate a fresh :class:`State`, or ``None`` if stateless."""
+
+    @abstractmethod
+    def input(self) -> Any:
+        """Allocate a fresh input container with one field per input port.
+
+        The returned object has attributes whose names match the ``*_attr``
+        strings passed at construction. Each field is a freshly
+        :func:`wp.zeros`-allocated array of the right dtype and size. The
+        user typically reassigns these fields to point at live data buffers.
+        """
+
+    @abstractmethod
+    def output(self) -> Any:
+        """Allocate a fresh output container with one field per output port."""
+
+    @abstractmethod
+    def compute(
+        self,
+        inputs: Any,
+        outputs: Any,
+        controller_state_now: Controller.State | None,
+        controller_state_next: Controller.State | None,
+        time_step: float | wp.array[wp.float32],
+    ) -> None:
+        """Run one control step.
+
+        Args:
+            inputs: Object whose attributes hold the live read ports.
+                Resolved via ``getattr(inputs, attr_name)``. Any
+                duck-typed object works; :meth:`input` returns a
+                pre-allocated one.
+            outputs: Same contract as ``inputs`` for write ports. The
+                kernel performs slot-replacing writes — slots outside the
+                declared port indices are left untouched.
+            controller_state_now: Current state (``None`` if stateless).
+            controller_state_next: Next-step state to populate (``None``
+                if stateless). Double-buffered against
+                ``controller_state_now``.
+            time_step: Step duration [s]. Either a plain ``float`` or a
+                ``wp.array`` of shape ``(1,)`` dtype ``wp.float32``.
+        """
--- a/newton/_src/controllers/controller.py
+++ b/newton/_src/controllers/controller.py
@@ -36,23 +36,9 @@ class Controller(ABC):
       ``+=``); composing laws is the user's job.
     """
 
-    @dataclass
-    class State:
-        """Pure data container. Subclasses declare their fields."""
-
     @abstractmethod
     def is_graphable(self) -> bool:
         """Whether :meth:`compute` is safe to capture in a CUDA graph."""
-
-    @abstractmethod
-    def state(self) -> Controller.State | None:
-        """Allocate a fresh :class:`State`, or ``None`` if stateless.
-
-        The return value is the canonical indicator of statefulness:
-        ``ctrl.state() is not None`` means the controller maintains state
-        between steps and callers should allocate two buffers for
-        double-buffering.
-        """
 
     @abstractmethod
     def input(self) -> Any:
@@ -73,9 +59,7 @@ class Controller(ABC):
         self,
         inputs: Any,
         outputs: Any,
-        controller_state_now: Controller.State | None,
-        controller_state_next: Controller.State | None,
-        time_step: float | wp.array[wp.float32],
+        dt: float | wp.array[wp.float32],
     ) -> None:
         """Run one control step.
 
@@ -87,10 +71,6 @@ class Controller(ABC):
             outputs: Same contract as ``inputs`` for write ports. The
                 kernel performs slot-replacing writes — slots outside the
                 declared port indices are left untouched.
-            controller_state_now: Current state (``None`` if stateless).
-            controller_state_next: Next-step state to populate (``None``
-                if stateless). Double-buffered against
-                ``controller_state_now``.
-            time_step: Step duration [s]. Either a plain ``float`` or a
+            dt: Step duration [s]. Either a plain ``float`` or a
                 ``wp.array`` of shape ``(1,)`` dtype ``wp.float32``.
         """
--- a/newton/_src/controllers/controller.py
+++ b/newton/_src/controllers/controller.py
@@ -6,7 +6,6 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from dataclasses import dataclass
 from typing import Any
 
 import warp as wp
--- a/newton/_src/controllers/controller.py
+++ b/newton/_src/controllers/controller.py
@@ -56,6 +56,7 @@ class Controller(ABC):
     @abstractmethod
     def compute(
         self,
+        *,
         inputs: Any,
         outputs: Any,
         dt: float | wp.array[wp.float32],
--- a/newton/_src/controllers/controller.py
+++ b/newton/_src/controllers/controller.py
@@ -24,9 +24,8 @@ class ControllerBase(ABC, Generic[InputT, OutputT]):
 
     Subclasses are responsible for:
 
-    - :meth:`is_graphable`: predicate the user can query to decide graph
-      capture. Check ``ctrl.state() is not None`` to determine whether
-      double-buffered state setup is needed.
+    - :meth:`is_graphable`: predicate the user can query to decide whether
+      graph capture is possible.
     - :meth:`input`, :meth:`output`: allocate fresh typed input/output structs.
       Baked-in arrays (gains passed as a ``wp.array`` at construction) are
       stored on the controller and do **not** appear on the input struct.
--- a/newton/_src/controllers/controller.py
+++ b/newton/_src/controllers/controller.py
@@ -6,29 +6,29 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import Any
+from typing import Generic, TypeVar
 
 import warp as wp
 
+InputT = TypeVar("InputT")
+OutputT = TypeVar("OutputT")
 
-class Controller(ABC):
+
+class ControllerBase(ABC, Generic[InputT, OutputT]):
     """Abstract interface for a single Newton control law.
 
     Every concrete control law (joint impedance, differential IK, …) subclasses
-    :class:`Controller` directly. There is no framework-level composition: users
+    :class:`ControllerBase` directly. There is no framework-level composition: users
     who want to combine multiple control laws call each one's :meth:`compute`
     in sequence themselves.
 
     Subclasses are responsible for:
 
-    - :meth:`is_graphable`: predicate the user can query to decide CUDA-graph
+    - :meth:`is_graphable`: predicate the user can query to decide graph
       capture. Check ``ctrl.state() is not None`` to determine whether
       double-buffered state setup is needed.
-    - :meth:`state`: allocate a fresh :class:`State`, or return ``None`` for
-      stateless laws.
-    - :meth:`input`, :meth:`output`: allocate fresh duck-typed containers
-      holding only the live ports declared at construction. Baked-in arrays
-      (gain ports given a :class:`wp.array` rather than an attr name) are
+    - :meth:`input`, :meth:`output`: allocate fresh typed input/output structs.
+      Baked-in arrays (gains passed as a ``wp.array`` at construction) are
       stored on the controller and do **not** appear on the input struct.
     - :meth:`compute`: read the input struct's live arrays, run kernels, write
       the output struct's live arrays. Writes are slot-replacing (``=``, not
@@ -40,37 +40,29 @@ class Controller(ABC):
         """Whether :meth:`compute` is safe to capture in a graph."""
 
     @abstractmethod
-    def input(self) -> Any:
-        """Allocate a fresh input container with one field per input port.
+    def input(self) -> InputT:
+        """Allocate a fresh input struct with zero-initialised arrays.
 
-        The returned object has attributes whose names match the ``*_attr``
-        strings passed at construction. Each field is a freshly
-        :func:`wp.zeros`-allocated array of the right dtype and size. The
-        user typically reassigns these fields to point at live data buffers.
+        The user typically reassigns fields to point at live data buffers.
+        Fields for disabled features are ``None``.
         """
 
     @abstractmethod
-    def output(self) -> Any:
-        """Allocate a fresh output container with one field per output port."""
+    def output(self) -> OutputT:
+        """Allocate a fresh output struct."""
 
     @abstractmethod
     def compute(
         self,
         *,
-        inputs: Any,
-        outputs: Any,
+        inputs: InputT,
+        outputs: OutputT,
         dt: float | wp.array[wp.float32],
     ) -> None:
         """Run one control step.
 
         Args:
-            inputs: Object whose attributes hold the live read ports.
-                Resolved via ``getattr(inputs, attr_name)``. Any
-                duck-typed object works; :meth:`input` returns a
-                pre-allocated one.
-            outputs: Same contract as ``inputs`` for write ports. The
-                kernel performs slot-replacing writes — slots outside the
-                declared port indices are left untouched.
-            dt: Step duration [s]. Either a plain ``float`` or a
-                ``wp.array`` of shape ``(1,)`` dtype ``wp.float32``.
+            inputs: Populated input struct (see :meth:`input`).
+            outputs: Output struct to write into (see :meth:`output`).
+            dt: Step duration [s].
         """
--- a/newton/_src/controllers/controller.py
+++ b/newton/_src/controllers/controller.py
@@ -37,7 +37,7 @@ class Controller(ABC):
 
     @abstractmethod
     def is_graphable(self) -> bool:
-        """Whether :meth:`compute` is safe to capture in a CUDA graph."""
+        """Whether :meth:`compute` is safe to capture in a graph."""
 
     @abstractmethod
     def input(self) -> Any:
--- a/newton/_src/controllers/controller.py
+++ b/newton/_src/controllers/controller.py
@@ -19,7 +19,7 @@ class ControllerBase(ABC, Generic[InputT, OutputT]):
 
     Every concrete control law (joint impedance, differential IK, …) subclasses
     :class:`ControllerBase` directly. There is no framework-level composition: users
-    who want to combine multiple control laws call each one's :meth:`compute`
+    who want to combine multiple control laws call each one's :meth:`step`
     in sequence themselves.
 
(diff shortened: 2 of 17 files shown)


newton/_src/controllers/controller.py
```diff
@@ -0,0 +1,96 @@
...
+                pre-allocated one.
+            outputs: Same contract as ``inputs`` for write ports. The
+                kernel performs slot-replacing writes — slots outside the
+                declared port indices are left untouched.
+            controller_state_now: Current state (``None`` if stateless).
+            controller_state_next: Next-step state to populate (``None``
+                if stateless). Double-buffered against
+                ``controller_state_now``.
+            time_step: Step duration [s]. Either a plain ``float`` or a
+                ``wp.array`` of shape ``(1,)`` dtype ``wp.float32``.
```
jeff-hough: Note: time-step has to optionally be an array so that exporting with APIC can work properly.

newton/_src/controllers/impl/joint_impedance/model_based.py
```diff
@@ -0,0 +1,371 @@
...
+
+    def __init__(
+        self,
+        model_builder: ModelBuilder,
+        default_dof_indices: wp.array[wp.uint32],
+        stiffness: wp.array2d[wp.float32] | str,
+        damping: wp.array2d[wp.float32] | str,
+        use_gravity_compensation: bool = True,
+        use_coriolis_compensation: bool = True,
+        use_inertia_decoupling: bool = True,
```
jeff-hough, line 122: For simplicity, all robots have the same settings here. We could make it such that this is a per-robot setting, in which case:

* pages of the mass matrix would get filled with identity
* parts of the gravity and coriolis vectors would be filled with 0

newton/examples/controllers/example_controller_joint_impedance_heterogeneous.py
jeff-hough, line 1: Given [this communication from earlier](https://gist.github.com/jcarius-nv/3c430445d9cf5581233be065ae328210), we likely want to delete this example before merging. It shows a very simple graphed scenario, using `ControllerJointImpedance` on a heterogeneous set of robots.

jeff-hough:
Also, will the `Controller` class name need to change? Error from our pipeline:

```
/home/runner/work/newton/newton/newton/_src/sim/builder.py:docstring of newton._src.sim.builder.ModelBuilder:41: WARNING: more than one target found for cross-reference 'Controller': newton.actuators.Controller, newton._src.actuators.controllers.base.Controller, newton.controllers.Controller, newton._src.controllers.controller.Controller [ref.python]

```

I wouldn't mind changing it to something like `MotionGenerator`. I have never really loved the word `Controller` anyways, since to me it implies a feedback control loop. Many of the planned algorithms to put into this module take no feedback, they are just pure kinematic mappings (i.e. all of the mobile robot "controllers").

newton/examples/controllers/example_controller_joint_impedance_heterogeneous.py
adenzler-nvidia: In the spirit of having an example as a concept that shows how to use the newton APIs, I think we can have 1 example that shows how to set up a controller.

Otherwise, maybe fusing with some of the robot examples is a good idea?

newton/_src/controllers/impl/joint_impedance/model_based.py
```diff
@@ -0,0 +1,371 @@
...
+    def requires_grad(self) -> bool:
+        return self._requires_grad
+
+    def is_graphable(self) -> bool:
+        return True
+
+    def state(self) -> None:
+        return None
+
+    def input(self):
```
jvonmuralt: Would it be possible to reconsider the input()/output() struct pattern? In Newton context, the pre-allocated state arrays get immediately replaced when you assign inp.joint_q = state.joint_q, so the allocation doesn't add much. It also doesn't provide type safety. I would favor an approach closer to actuators, where arrays are passed directly to compute().

newton/_src/controllers/impl/joint_impedance/model_based.py
```diff
@@ -0,0 +1,371 @@
...
+        """Return a pre-allocated input struct without dynamics fields (computed internally)."""
+        ns = _allocate_namespace(self._input_specs, self._device, self._requires_grad)
+        shape_2d = (self._num_robots, self._max_dofs)
+        if self._stiffness_attr is not None:
```
jvonmuralt: If gains need to change between steps, can't the caller just hold a reference to the wp.array2d and write into it in-place? See how ControllerPD handles kp/kd in newton.actuators

newton/_src/controllers/impl/joint_impedance/_common.py
```diff
@@ -0,0 +1,97 @@
...
+    src: wp.array[wp.float32],  # flat sim array
+    indices: wp.array[wp.uint32],  # (total_dofs,) — concatenated per-robot, no padding
+    dst: wp.array[wp.float32],  # flat output (total_dofs,)
+):
+    flat = wp.tid()
+    dst[flat] = src[indices[flat]]
+
+
+@wp.kernel
+def _gather_dof_kernel(
```
jvonmuralt, line 73: Newton’s canonical joint state and control arrays are flat 1-D arrays. Could the controller keep vector quantities flat and reserve padded per-articulation layouts for inherently matrix-valued quantities such as the mass matrix? The current conversion introduces several gather/scatter kernels.

newton/_src/controllers/impl/joint_impedance/model_free.py
```diff
@@ -0,0 +1,408 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""ControllerJointImpedanceModelFree — joint-space impedance control with
+caller-supplied dynamics terms.
+
+The controller reads flat 1D sim arrays, gathers the controlled DOFs internally,
+computes a joint torque command, and scatters the result back to a flat 1D
+output array.
+
+The difference from :class:`ControllerJointImpedance` is that this controller
```
jvonmuralt, line 11: Do we need 2 separate classes if the difference is just in that ? Sounds like it adds additional API and code duplication.

newton/_src/controllers/impl/joint_impedance/model_based.py
```diff
@@ -0,0 +1,371 @@
...
+        joint_qd_des_attr: str = "joint_qd_des",
+        joint_qd_des_idx: wp.array[wp.uint32] | None = None,
+        joint_qdd_attr: str = "joint_qdd",
+        joint_qdd_idx: wp.array[wp.uint32] | None = None,
+        joint_f_attr: str = "joint_f",
+        joint_f_idx: wp.array[wp.uint32] | None = None,
+        device: Any = None,
+        requires_grad: bool = False,
+    ):
+        if not isinstance(model_builder, ModelBuilder):
```
jvonmuralt: Why does the controller require and finalize a second ModelBuilder instead of using the simulation’s finalized Model or an ArticulationView?

jvonmuralt (commented):
Thank you, Jeff! It is a large PR and of course the design decisions might be well motivated and were already discussed. From what I see high level I’m not in favor of requiring a separate ModelBuilder, exposing distinct model-based and model-free classes, and  introducing separate input/output container objects.

The tests also need more coverage, particularly for graph capture and device consistency across controller inputs, indices, and internal buffers.

newton/_src/controllers/controller.py
```diff
@@ -0,0 +1,96 @@
...
+    @dataclass
+    class State:
+        """Pure data container. Subclasses declare their fields."""
+
+    @abstractmethod
+    def is_graphable(self) -> bool:
+        """Whether :meth:`compute` is safe to capture in a CUDA graph."""
+
+    @abstractmethod
+    def state(self) -> Controller.State | None:
```
adenzler-nvidia: The state protocol is currently speculative: both concrete controllers return `None` from `state()`, ignore `controller_state_now`/`controller_state_next` in `compute()`, and `Controller.State` is an empty placeholder. Since nothing exercises it, we can't know yet whether double-buffered `State` objects are the right shape for the first stateful controller. I'd drop `state()` and the two state parameters from the ABC and add them back with the first controller that needs them — retrofitting is cheap, but shipping unused public protocol (even experimental) means a breaking change if the design turns out wrong.

newton/_src/controllers/controller.py
```diff
@@ -0,0 +1,96 @@
...
+        :func:`wp.zeros`-allocated array of the right dtype and size. The
+        user typically reassigns these fields to point at live data buffers.
+        """
+
+    @abstractmethod
+    def output(self) -> Any:
+        """Allocate a fresh output container with one field per output port."""
+
+    @abstractmethod
+    def compute(
```
adenzler-nvidia: Two naming consistency issues on this signature: (1) `time_step` should be `dt`, matching the rest of Newton (`SolverBase.step`, `newton.actuators.Controller.compute`) — the `float | wp.array` union can stay for the APIC use case. (2) The abstract method declares `inputs`/`outputs` but both concrete overrides rename them to `input_struct`/`output_struct`; for a public ABC the abstract and concrete parameter names should match so keyword calls work uniformly.

newton/_src/controllers/utils.py
```diff
@@ -0,0 +1,89 @@
...
+
+    if idx.size != default_idx.size:
+        raise TypeError(
+            f"Port '{name}': indices must be the same size as default_dof_indices: {idx.size} != {default_idx.size}."
+        )
+
+    return idx
+
+
+def _normalize_parameter_port(
```
adenzler-nvidia: `_normalize_parameter_port` is never imported or called anywhere in the PR — dead code, please remove.

newton/_src/controllers/utils.py
```diff
@@ -0,0 +1,89 @@
...
+) -> wp.array[wp.uint32]:
+    """Return ``idx`` after validation, or ``default_idx`` if ``idx`` is ``None``."""
+    if idx is None:
+        return default_idx
+
+    if (not isinstance(idx, wp.array)) or (idx.dtype != wp.uint32):
+        raise TypeError(f"Port '{name}': idx must be wp.array[uint32] or None, got {type(idx).__name__}.")
+
+    if idx.size != default_idx.size:
+        raise TypeError(
```
adenzler-nvidia: This is a size mismatch, not a type error — should raise `ValueError`, consistent with the equivalent length check in `ControllerJointImpedanceModelFree.__init__`.

newton/_src/controllers/impl/joint_impedance/model_free.py
```diff
@@ -0,0 +1,408 @@
...
+        if self._has_qdd:
+            specs.append((self._qdd_attr, wp.float32, _idx_max(self._qdd_idx)))
+        if self._use_gravity:
+            specs.append((self._gravity_attr, wp.float32, _idx_max(self._gravity_idx)))
+        if self._use_coriolis:
+            specs.append((self._coriolis_attr, wp.float32, _idx_max(self._coriolis_idx)))
+        ns = _allocate_namespace(specs, self._device, self._requires_grad)
+        shape_2d = (self._num_robots, self._max_dofs)
+        if self._stiffness_attr is not None:
+            setattr(ns, self._stiffness_attr, wp.zeros(shape_2d, dtype=wp.float32, device=self._device))
```
adenzler-nvidia: The live-gain and mass-matrix allocations here (and in `ControllerJointImpedance.input()`) don't pass `requires_grad`, while the `_allocate_namespace` fields do — inconsistent for differentiable pipelines.

newton/_src/controllers/impl/joint_impedance/model_based.py
```diff
@@ -0,0 +1,371 @@
...
+            (self._q_attr, wp.float32, _idx_max(self._q_idx)),
+            (self._qd_attr, wp.float32, _idx_max(self._qd_idx)),
+            (self._q_des_attr, wp.float32, _idx_max(self._q_des_idx)),
+            (self._qd_des_attr, wp.float32, _idx_max(self._qd_des_idx)),
+        ]
+        if self._has_qdd:
+            self._input_specs.append((self._qdd_attr, wp.float32, _idx_max(self._qdd_idx)))
+
+    @staticmethod
+    def _check_gain(
```
adenzler-nvidia: `_check_gain` duplicates the validation that `_normalize_gain` re-runs in the inner model-free instance — drift risk. Since the inner instance validates anyway, this can be dropped (or delegated) to keep a single source of truth.

adenzler-nvidia (commented):
A few additional points on the base-class contract and some smaller items, complementing the existing threads.

newton/_src/controllers/impl/joint_impedance/model_based.py
jeff-hough: The input and output arrays vary significantly between each controller - I went with this approach to keep the function signature consistent between controllers, which I think makes them easier to learn/use.

The `input` and `output` functions are for convenience only; the user is always free to create their own struct if they need to avoid the extra allocations.

newton/_src/controllers/impl/joint_impedance/model_based.py
jeff-hough: That is another option - however, I designed it this way since newton has a convention for statelessness. If variable gains are stored inside the controller, then that is introduction of state (how the controller responds depends now on prior calls I have made to the controller).

Using this approach, you either treat the gain as a _fixed_ value stored on the controller, or if it needs to vary, then you must explicitly treat it as another input.

I am open to changing this approach if it is either a) too confusing or b) too inconsistent with the rest of newton!

newton/_src/controllers/impl/joint_impedance/model_based.py
```diff
@@ -0,0 +1,371 @@
...
+            (only used when ``has_qdd_feedforward=True``).
+        joint_qdd_idx: Optional index array for feedforward read.
+        joint_f_attr: Flat sim attr name for the torque output.
+        joint_f_idx: Optional index array overriding ``default_dof_indices``
+            for the torque-output write.
+        device: Warp device.
+        requires_grad: Whether internal buffers need gradient support.
+    """
+
+    def __init__(
```
adenzler-nvidia, line 113: Since the module is still experimental, this is a good moment to lock in keyword-only parameters before anyone depends on positional order.

Newton's newer constructors keep only the model-like anchor argument positional and put everything else behind `*`, including required parameters: `SensorContact.__init__(self, model, *, ...)`, `SensorTiledCamera.__init__(self, model, *, ...)`, `SolverMuJoCo.__init__(self, model, *, ...)`, and `ViewerViser.__init__(self, *, ...)` for classes with no natural anchor arg. The two controller constructors here should follow that pattern:

```python
class ControllerJointImpedance:
    def __init__(
        self,
        model_builder: ModelBuilder,
        *,
        default_dof_indices: wp.array[wp.uint32],
        stiffness: wp.array2d[wp.float32] | str,
        damping: wp.array2d[wp.float32] | str,
        use_gravity_compensation: bool = True,
        ...
    )

class ControllerJointImpedanceModelFree:
    def __init__(
        self,
        *,
        num_robots: int,
        dofs_per_robot: wp.array[wp.int32],
        max_dofs: int,
        ...
    )
```

Rationale beyond consistency:

- Four consecutive bool flags and ten alternating `*_attr`/`*_idx` pairs make positional calls unreadable, and a positional off-by-one silently wires the wrong port.
- The docstring promises an identical interface between the two classes, but positionally they diverge: `joint_f_attr` is parameter 19 in `ControllerJointImpedance` and 24 in `ControllerJointImpedanceModelFree`. Keyword-only makes the shared parameters spell identically at every call site.
- Keyword-only leaves room to reorder or insert parameters later without a breaking change, which matters while the API is still settling.

Required keyword-only parameters (no default) are fine — `stiffness`/`damping` are required but their position carries no meaning, and `num_robots`/`max_dofs` are adjacent ints that are easy to transpose.

newton/_src/controllers/impl/joint_impedance/_common.py
jeff-hough: I could change this, yes - I agree that it introduces some unnecessary gather/scatter kernels. However, we might want to measure if it makes a significant performance difference.

Personally, I find the gather/scatter into contiguous 2D arrays makes for more intuitive compute kernels, with less inputs (no need to pass indices).

I will make the change and see how it looks!

newton/_src/controllers/impl/joint_impedance/model_free.py
jeff-hough: The classes don't overlap - the `ControllerJointImpedance` simply holds an internal `newton.Model` to compute the fk, gravity, coriolis and mass-matrix, and then passes that data to an internal `ControllerJointImpedanceModelFree`. The actual "control law" code always lives in the model-free variant.

The **model-based** variants are much more portable, since their I/O is much simpler, and the entire robot model is included in the graph-capture - the user just passes joint-angles and velocities, and then the controller itself takes care of computing gravity compensation, mass-matrix, coriolis - the user doesn't need to know how our mass-matrix/jacobians are structured, etc.

The model-free variant is useful for cases like Isaac Lab, where the jacobians, mass-matrices etc will all be pre-computed anyways, and it is okay the the robot model in physics + in the controller are identical.

I also I don't love having two classes - but this was the most straightforward way I could think of to achieve this!

newton/_src/controllers/impl/joint_impedance/model_based.py
jeff-hough: This is by design - by holding our own independent model, we can introduce modelling errors between the physical robot and the model we use in the controller - these always exist on real robots, and I think it is important to testing controller robustness. If the user doesn't want modelling errors, they can a) use the ModelFree variant, or b) pass a ModelBuilder with no modelling errors.

The reason for taking a `ModelBuilder` rather than a finalized model or `ArticulationView` is that it introduces less chance for device mismatch - i.e. the model is always finalized on the device chosen for the `Controller`.

newton/_src/controllers/controller.py
jeff-hough: Agreed - I will remove it

newton/_src/controllers/controller.py
jeff-hough: 1) Agreed, I will match
2) Yes, I missed those, thank you!

newton/_src/controllers/utils.py
jeff-hough: Sounds good - this is used in controllers which were not put into this PR, so I will remove here!

newton/_src/controllers/impl/joint_impedance/model_based.py
```diff
@@ -0,0 +1,350 @@
...
+        outputs: Any,
+        dt: float | wp.array[wp.float32],
+    ) -> None:
+        """Run one impedance-control step.
+
+        Args:
+            inputs: Namespace with flat sim arrays for joint state and
+                desired state. Dynamics terms are computed internally.
+            outputs: Namespace with a flat sim torque array.
+            dt: Unused. Accepted for API compatibility.
```
jeff-hough, line 310: Though this is unused in this specific controller, I have already implemented other articulation controllers (differential inverse kinematics) which use the `dt` parameter, so I am leaving this in the signature for now.

jeff-hough:
> Also, will the `Controller` class name need to change? Error from our pipeline:
> 
> ```
> /home/runner/work/newton/newton/newton/_src/sim/builder.py:docstring of newton._src.sim.builder.ModelBuilder:41: WARNING: more than one target found for cross-reference 'Controller': newton.actuators.Controller, newton._src.actuators.controllers.base.Controller, newton.controllers.Controller, newton._src.controllers.controller.Controller [ref.python]
> ```
> 
> I wouldn't mind changing it to something like `MotionGenerator`. I have never really loved the word `Controller` anyways, since to me it implies a feedback control loop. Many of the planned algorithms to put into this module take no feedback, they are just pure kinematic mappings (i.e. all of the mobile robot "controllers").

Motion Generator also goes well with this docs page on Isaac Lab:
https://isaac-sim.github.io/IsaacLab/main/source/overview/core-concepts/motion_generators.html

newton/_src/controllers/controller.py
eric-heiden: Maybe a more consistent name should be `step()` for this method? To match what we have in `Actuator` and `Solver`.

newton/_src/controllers/controller.py
```diff
@@ -0,0 +1,76 @@
...
+      (gain ports given a :class:`wp.array` rather than an attr name) are
+      stored on the controller and do **not** appear on the input struct.
+    - :meth:`compute`: read the input struct's live arrays, run kernels, write
+      the output struct's live arrays. Writes are slot-replacing (``=``, not
+      ``+=``); composing laws is the user's job.
+    """
+
+    @abstractmethod
+    def is_graphable(self) -> bool:
+        """Whether :meth:`compute` is safe to capture in a CUDA graph."""
```
eric-heiden: nit: the wording here should be just "graph", not only "CUDA graph" - same in the other docstrings here. We will have feature parity for CPU and CUDA graphs, so we shouldn't distinguish this.

newton/_src/controllers/controller.py
```diff
@@ -0,0 +1,76 @@
...
+
+    @abstractmethod
+    def input(self) -> Any:
```
eric-heiden: It would be nice to have type hints here instead of `Any`, like nested classes for the input/output structures.

newton/_src/controllers/impl/joint_impedance/model_based.py
```diff
@@ -0,0 +1,351 @@
...
+        joint_f_attr: Flat sim attr name for the torque output.
+        joint_f_idx: Optional index array overriding ``default_dof_indices``
+            for the torque-output write.
+        device: Warp device.
+        requires_grad: Whether internal buffers need gradient support.
+    """
+
+    def __init__(
+        self,
+        model_builder: ModelBuilder,
```
eric-heiden: We call this argument `builder` typically.

newton/_src/controllers/impl/joint_impedance/model_free.py
```diff
@@ -0,0 +1,412 @@
...
+            baked = wp.zeros(expected, dtype=wp.float32, device=self._device, requires_grad=self._requires_grad)
+            wp.copy(baked, value)
+            return None, baked
+        raise TypeError(
+            f"Port '{name}': must be wp.array2d[wp.float32] of shape (num_robots, max_dofs) "
+            f"or str; got {type(value).__name__}."
+        )
+
+    @property
+    def num_robots(self) -> int:
```
eric-heiden: Following our conventions, we prefer the word "robot_count" rather than "num_robots".

newton/_src/controllers/controller.py
jeff-hough: The `Input` and `Output` types are dynamic, since they:
* change array names depending on the `*_attr` passed in by the user
* may have extra arrays depending on whether:
  * the user wants to treat the gains as variable (i.e. another input) or fixed (baked into the controller)
  * they are using dynamic compensation (model-free variant) in which case they will have to input mass-matrix, coriolis vector, gravity etc.

I used the `*_attr` pattern to match the [dynamic naming that is done in actuators](https://github.com/newton-physics/newton/blob/de2d06f3d63ecbd1347322c36d44ab1eee1c3bf6/newton/_src/actuators/actuator.py#L89).

However, I agree that it would be nice to have more meaningful typehints. I could implement this suggestion if we are willing to give up dynamic naming of the inputs, which would also get rid of all of the many `*_attr` inputs on the constructors. Personally, I would be happy with that change. Do you agree @eric-heiden ?

docs/api/newton_controllers.rst
```diff
@@ -0,0 +1,31 @@
...
+.. py:module:: newton.controllers
+.. currentmodule:: newton.controllers
+
+.. rubric:: Classes
+
+.. autosummary::
+   :toctree: _generated
+   :nosignatures:
+
+   Controller
```
eric-heiden: As discussed, one way to avoid a naming conflict with `actuators.Controller` could be renaming this class to `ControllerBase`.

newton/_src/controllers/controller.py
eric-heiden: Yes, I think that could be a nice simplification to get rid of the `_*attr` pattern. I don't know if we envision use cases now where a user would change the name of these attributes? They could always use duck-typing to conform to our Newton attribute format anyway. Or do you know if this feature will be used outside Newton, like something in PhysX (which is I think what motivated this pattern in the actuators module - @jvonmuralt correct me if I'm wrong)?

jvonmuralt: Originally, actuators were designed to be as general as possible. They are supposed to be used not only with newton control but also accept user-specific control objects and potentially in the future control signals and their derivatives at any level (joints, tendons, or end effectors). We considered two ways to achieve this:

    1. Introduce input/output classes and rely on the user to point them to the correct arrays.

    2. Allow users to pass arbitrary objects and specify the target field names.

To avoid introducing extra objects, it was decided on Option 2 using the attribute pattern.

jcarius-nv:
## General review — Resolve actuation versus portable-controller architecture first

After further review and discussion, in particular with @jvonmuralt , I think we need to settle a more
fundamental architecture question before iterating on the detailed public API (for which I also have comments, but won't post them right now to keep the discussion focused).

**This PR currently appears to combine two different goals:**
(or maybe: Reviewers and author have opposing opinions on whether it should be (1) or (2))

1. Well-integrated simulation controllers in Newton, where a transmission layer
   may be the missing abstraction in the existing `newton.actuators` pipeline.
2. Portable Warp control algorithms designed for graph/APIC export and reuse
   across simulators and on hardware.

Those layers may share implementations, but they do not necessarily share the
same model ownership, scheduling frequency, state lifecycle, I/O adaptation, or
package boundary. MuJoCo's separation of transmission, activation dynamics, and
force generation is a relevant architecture to compare; Newton's actuator
documentation already identifies the absence of a transmission model as a
current limitation. At the same time, higher-level or coupled controllers may
still sit above an actuator transmission, so this comparison needs to classify
the intended algorithms rather than assuming one mechanism covers everything.

I suggest we pause the public `newton.controllers` contract until the relevant
stakeholders align on:

- which algorithms belong in actuator force generation, transmission, or
  higher-level command generation;
- whether the hardware/APIC-facing library belongs in Newton core or behind a
  separate package/adapter boundary; and
- how simulation and higher-frequency hardware scheduling and model lifecycles
  are expected to differ?

Once that boundary is decided, the naming, base-class protocol, typed I/O,
model-based/model-free lifecycle, APIC validation, and example shape can be
reviewed against the chosen architecture. Until then, those detailed API
decisions risk hardening the wrong seam.

jcarius-nv (commented):
Following up on the architecture concern I raised earlier: the subsequent discussion clarified the intended boundary. We agree that controllers and actuators are distinct concepts that may coexist in Newton. Actuators model hardware and map commands to simulation forces, while controllers are portable software algorithms that may run in simulation or on robots. A future transmission abstraction remains a separate concern for `newton.actuators`; `newton.controllers` should remain self-contained and not imply transmission behavior.

We also agreed to keep controllers in the main Newton repository and distribution for synchronized CI, releases, and maintenance. A package extra cannot make the module itself optional—it can only select optional dependencies—so a separate wheel is not warranted.

With placement resolved, the remaining review is about the concrete public contract. Mo's proposal implies that Newton should expose a small integration contract so users can bring their own controllers. If `ControllerBase` is that extension point, its exact protocol should be validated across representative controller types and paired with a customization path for controllers that do not fit it. Newton should own that integration contract and curated reusable primitives, rather than standardize every controller's logic.

docs/api/newton_controllers.rst
```diff
@@ -0,0 +1,31 @@
+.. SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+.. SPDX-License-Identifier: CC-BY-4.0
+
+newton.controllers
+==================
+
+GPU-accelerated, vectorized control laws for Newton physics simulations.
```
jcarius-nv: The opening still frames these as control laws “for Newton physics simulations” and says they operate on arrays matching `newton.State`. Our agreed boundary is broader: these are portable software controllers intended to work across physics engines and, through APIC, on robots; they are distinct from Newton's simulation actuator model.

Could we describe that engine-independent purpose and contract first, then identify which conveniences or implementations are specifically Newton-backed? If matching `newton.State` layout is required by the shared API, that is an important portability constraint to state explicitly rather than presenting it as the definition of this layer.

newton/controllers.py
```diff
@@ -0,0 +1,28 @@
...
+
+"""GPU-accelerated, vectorized control laws for Newton physics simulations.
+
+This module provides standalone controllers that compute joint torques or
+other actuation signals from simulation state. Each controller is a concrete
+subclass of :class:`ControllerBase` and operates on flat 1D arrays matching the
+layout of :class:`~newton.State` fields, making them composable with any
+Newton solver.
+
+.. experimental::
```
jcarius-nv, line 10: Newton's developer guidance says an entirely experimental module should use a bodyless `.. experimental::` directive, which renders the standard notice for API, behavior, defaults, and supported use cases. The custom body here only mentions API changes.

Could we switch both the public module docstring and generated API landing page to the bodyless directive? The changelog should also say that `newton.controllers` is experimental and update the stale exported base name from `Controller` to `ControllerBase`.

newton/_src/controllers/controller.py
```diff
@@ -0,0 +1,68 @@
...
+from abc import ABC, abstractmethod
+from typing import Generic, TypeVar
+
+import warp as wp
+
+InputT = TypeVar("InputT")
+OutputT = TypeVar("OutputT")
+
+
+class ControllerBase(ABC, Generic[InputT, OutputT]):
```
jcarius-nv, line 17: Mo's proposal is that Newton own a small controller integration contract while letting users bring their own controller implementations. If inheritance through `ControllerBase` is the intended customization path, then it makes sense for the base to be public. What we still need to validate is whether this exact ABC is the right extension contract for the intentionally heterogeneous controller collection.

The PR currently proves the base only with two variants of the same impedance law, where the model-based class delegates the actual control law to the model-free class. It therefore does not yet demonstrate that `input()`, `output()`, `is_graphable()`, and `step(..., dt=...)` remain a natural contract for differential IK, operational-space control, and mobile-base controllers. `is_graphable()` is also always true in the implementations, matching the feature requirement rather than demonstrating useful polymorphism.

Could we validate the protocol against representative, materially different controllers and document what Newton guarantees to user-defined subclasses? That should include the role of these methods plus the customization path for controllers that do not fit the common pattern. A draft differential-IK implementation, operational-space controller, or compact API sketch would be enough evidence; it does not need to merge in this PR. If that validation is not yet available, deferring the public base remains the safer fallback, but it is not the desired end state if `ControllerBase` is meant to enable user extensions.

newton/_src/controllers/controller.py
```diff
@@ -0,0 +1,68 @@
...
+    def output(self) -> OutputT:
+        """Allocate a fresh output struct."""
+
+    @abstractmethod
+    def step(
+        self,
+        *,
+        inputs: InputT,
+        outputs: OutputT,
+        dt: float | wp.array[wp.float32],
```
jcarius-nv, line 59: APIC requirements are shaping the public interface—most visibly the array-valued `dt`, but also preallocation, graphability, and the model-based controller lifecycle. The CUDA graph-captured example is appropriate as the canonical performance path, but graph capture alone does not validate serialization, buffer binding, artifact completeness, loading, or numerical equivalence after replay.

Please add one Python APIC round-trip integration test: capture a representative model-based controller with `wp.ScopedCapture(apic=True)`, save it with `wp.capture_save()`, reload it with `wp.capture_load()`—ideally in a fresh process—bind representative inputs including array-valued `dt`, execute it, and compare its torques with the Python path. Generate and reload the artifact in the same test run against Newton's pinned Warp version; a committed golden artifact and standalone C++ replay are out of scope.

Because Isaac is the named consumer motivating the model-free public contract, please also provide a representative adapter or equivalence test exercising its array ownership, DOF mapping, dynamics-update timing, and device assumptions. Finally, `CONTRIBUTING.md` requires benchmarking for a feature of this size; please include focused steady-state controller-step measurements for representative batch sizes, retaining graph capture as the canonical performance path.

newton/_src/controllers/controller.py
```diff
@@ -0,0 +1,68 @@
...
+
+    Every concrete control law (joint impedance, differential IK, …) subclasses
+    :class:`ControllerBase` directly. There is no framework-level composition: users
+    who want to combine multiple control laws call each one's :meth:`step`
+    in sequence themselves.
+
+    Subclasses are responsible for:
+
+    - :meth:`is_graphable`: predicate the user can query to decide graph
+      capture. Check ``ctrl.state() is not None`` to determine whether
```
jcarius-nv: `ControllerBase` still tells users to check `ctrl.state()` to determine whether double-buffered state setup is needed, but the state protocol was removed and neither the ABC nor either implementation defines `state()`.

Please remove that instruction and re-read the base documentation against the current interface before generating the API docs.

newton/_src/controllers/impl/joint_impedance/model_free.py
jcarius-nv: The Isaac model-free and portable model-based use cases explain why both implementations may be useful. What is still missing before making both public is executable evidence for their different ownership and update lifecycles.

For the model-free variant, who owns and updates the dynamics inputs, and what guarantees their topology, DOF ordering, snapshot timing, and device match the controller configuration? For the model-based variant, construction finalizes the supplied `ModelBuilder` into a private model and state: does that snapshot, consume, or constrain later reuse of the builder; how is the controller model updated or rebuilt; and how is its topology and DOF order checked against the runtime arrays?

Could we point to a runnable or draft Isaac consumer for the model-free lifecycle and an APIC consumer for the model-based lifecycle? They do not need to be merged into this PR, but seeing both paths exercised would validate that two public construction and ownership models are warranted. If only one is currently executable, keeping the other wrapper private would leave room to adjust it.

newton/examples/controllers/example_controller_joint_impedance_heterogeneous.py
jcarius-nv: I do not think we should delete controller example coverage entirely. Controllers are a major user-facing capability, so one canonical public example that teaches how to construct, bind, graph-capture, and run a representative controller is justified. I agree that one is enough.

I would simplify the current example to one self-contained articulation using the model-based joint-impedance path, retain graph capture as the canonical performance path, and preserve its meaningful `test_final()`. Heterogeneous DOF layouts and the model-free/APIC validation matrix belong in focused tests and documentation.

As the supported public entry point, please register its command in `README.md` and add the required 320x320 JPG screenshot. Future controllers should default to tests or extending this example unless they introduce a materially different user workflow.

docs/api/newton_controllers.rst
jeff-hough: Yes - reviewing this paragraph, it is in fact almost entirely wrong. I will re-write it, thank you for catching that.

jcarius-nv:
At current head bfb8bc3, all CI passes, but these remain. As discussed elsewhere, please record these as follow-up items, then we can get this merged and continue iterating on main.

  - Validate max_dofs == max(dofs_per_robot); undersizing currently silently skips DOFs. Also explicitly support or reject empty/ zero-DOF selections.

  - Clarify public API docs: Document the model-based versus model-free ownership/update lifecycles, and correct the conditional gain/mass-matrix units.

  - Complete example publication: Register the controller example in README.md and add the required 320×320 screenshot.
  - Create explicit follow-ups: APIC replay, Isaac/model-free equivalence, focused controller benchmarks, and validating ControllerBase with a
    materially different controller. These appear only in review threads/the broad #3548 umbrella—not dedicated tracked items.

  - Clean up PR state: Update the stale title/body (Impendance, Controller, compute(), state.joint_act), resolve addressed threads, and obtain
    approval. GitHub still reports REVIEW_REQUIRED with 28 unresolved threads, mostly outdated/addressed.

  Everything else from our architectural review is addressed or acceptably settled.

jcarius-nv (approved):

PR #3599, opened 2026-07-22
Create optional xforms argument for replicate function

StafaH:
## Description

Allow `builder.replicate` to accept a pre-formed xforms array for placement of environments. This is useful for downstream applications like IsaacLab that can then use it's pre-generated xforms array, and avoid a builder loop.

e.g. in IsaacLab:

```
source_xform_inv = wp.transform_inverse(world_xforms[0])
xforms = [wp.transform_multiply(world_xform, source_xform_inv) for world_xform in world_xforms]
builder.replicate(source_builders[sources[0]], num_worlds, xforms=xforms)
```

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added per-world transform support when replicating models (provide an explicit transforms sequence for each replicated world).
  * Replication can now use these transforms directly instead of auto-computing offsets (spacing is ignored when explicit transforms are provided).
* **Bug Fixes**
  * Added validation to reject explicit transform lists when their length doesn’t match the replication world count.
* **Documentation**
  * Updated the changelog to describe the new replication capability.
* **Tests**
  * Added unit tests for successful replication with explicit transforms and for proper failure on mismatched input.


newton/_src/sim/builder.py
```diff
@@ -2606,6 +2606,8 @@ def replicate(
         builder: ModelBuilder,
         world_count: int,
         spacing: tuple[float, float, float] = (0.0, 0.0, 0.0),
+        *,
+        xforms: Sequence[Transform | None] | None = None,
```
eric-heiden: It seems you are not actually handling the case when xforms contains a `None` value? So I would remove it from this type hint. Also would be good to check the earlier CodeRabbit comment.

StafaH: Fixed

newton/_src/sim/builder.py
```diff
@@ -2634,9 +2636,12 @@ def replicate(
         Args:
             builder: The builder to replicate. All entities from this builder will be copied.
             world_count: The number of worlds to create.
-            spacing: The spacing between each copy along each axis.
+            spacing: The spacing between each copy along each axis. Ignored when
+                ``xforms`` is provided.
                 For example, (5.0, 5.0, 0.0) arranges copies in a 2D grid in the XY plane.
                 Defaults to (0.0, 0.0, 0.0).
+            xforms: Optional transform for each replicated world. When provided,
+                its length must equal ``world_count``.
```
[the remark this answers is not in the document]
StafaH: Fixed

eric-heiden (approved):
Thanks, LGTM!

PR #3601, opened 2026-07-22
Load solver backends lazily to speed up `import newton`

### LINKED ISSUE #3600: [REQ] Reduce `import newton` time
## Description

`import newton` currently takes **~1.35 s** (median, warm cache, see environment below), and **~820 ms (~61%) of that is spent eagerly importing every solver backend**. Because `newton/__init__.py` imports `newton.solvers`, and `newton/_src/solvers/__init__.py` imports all solver packages, a downstream user who only needs one solver still pays for all of them.

This was reported from IsaacLab: importing `SolverMuJoCo` also imports Kamino, Style3D, implicit MPM, VBD, etc. — `SolverMuJoCo`'s own marginal import cost is only **~28 ms**, but any `import newton...` statement triggers the full ~1.35 s.

## Measurements

All numbers from Newton `main` @ `0b457032` ("Fix Featherstone ball joint damping (#3299)"), measured in fresh subprocesses (median of 7 runs, warm filesystem cache).

### Headline numbers

| what | median | notes |
|---|---:|---|
| `python -c pass` (baseline) | ~0 ms | |
| `import numpy` | 76 ms | |
| `import warp` | 221 ms | |
| **`import newton`** | **1350 ms** | 599 modules end up in `sys.modules` |
| `import newton._src.solvers.mujoco` (or *any* newton submodule) | ~1360 ms | parent `newton/__init__.py` always runs first |

### Where the 1.35 s goes (`python -X importtime -c "import newton"`, cumulative per subtree)

Direct children of `newton/__init__.py`:

| subtree | cum. ms |
|---|---:|
| **`newton.solvers`** | **819** |
| `newton._src.sim` | 162 (of which `sim.collide` 101) |
| `newton._src.core` | 160 (includes first import of most of `warp`) |
| `newton.sensors` | 81 (of which `sensor_contact` kernels 64) |
| `newton._src.geometry` | 77 |
| `newton.viewer` | 19 |
| `newton.actuators` / `newton.ik` / `newton.selection` | ~10 each |

Inside `newton._src.solvers/__init__.py` (cumulative, in import order — later solvers appear cheap because earlier ones already imported shared deps):

| import | cum. ms | notes |
|---|---:|---|
| `implicit_mpm` | 404 | first import of `warp.fem` (~180 ms) and `warp.optim.linear` (~146 ms) lands here |
| `style3d.solver_style3d` | 290 | pulls in the entire VBD solver via `style3d/collision/kernels.py` → `vbd.rigid_vbd_kernels`; `vbd.particle_vbd_kernels` alone is ~207 ms self-time |
| `featherstone` | 93 | |
| `mujoco` | 29 | `mujoco` / `mujoco_warp` are *not* imported here (already lazy 👍) |
| `kamino` | 3 | vendored kamino impl is *not* imported here (already lazy 👍) |
| `flags` | 0.2 | |

### Order-independent marginal cost per solver

To remove the import-order bias above, I also measured each solver package in isolation (stubbing `newton` / `newton._src.solvers` in `sys.modules` so a single solver can be imported without triggering the package `__init__`s; common deps `warp` + `newton._src.{core,geometry,sim,utils}` pre-imported; median of 3):

| module | marginal ms | third-party pulled in |
|---|---:|---|
| shared solver infra (`flags` + `SolverBase` + `coupled` pkg) | 117 | – |
| `implicit_mpm` | 257 | `warp.fem`, `warp.optim` |
| `vbd` | ~90–134 | – |
| `style3d.solver_style3d` | 103 (incl. its VBD dep) | – |
| `featherstone` | 39 | – |
| `mujoco` | 28 | – |
| `xpbd` | 21 | – |
| `semi_implicit` | 12 | – |
| `kamino` | 3 | – |

Almost all of this is warp kernel/struct declaration at module import time (`@wp.kernel`, `@wp.func`, `@wp.struct` in the various `*_kernels.py` files).

### Eager importing is worse than the sum of its parts

There is a superlinear interaction: once `warp.fem` / `warp.optim` are loaded (by `implicit_mpm`), subsequent warp kernel-module imports get ~2× slower. Example: importing the VBD solver in isolation takes ~90–134 ms, but with `warp.fem`+`warp.optim` pre-imported it takes **~223 ms** — matching the ~272 ms observed for the VBD subtree during the full `import newton`. So `implicit_mpm`'s true cost to `import newton` is its own ~260 ms *plus* an inflation of every solver imported after it.

### What this costs a MuJoCo-solver-only user (e.g. IsaacLab)

Needed: warp + newton core/sim/geometry (~500 ms) + solver infra (~117 ms) + `solver_mujoco` (~28 ms) ≈ **0.65 s**.
Actual: **~1.35 s** — i.e. **~0.7 s (>50%) is spent importing solvers that are never used.**

What's *not* a problem (already lazy, worth preserving): `mujoco`, `mujoco_warp`, the vendored Kamino implementation, and USD/trimesh/scipy are all **not** imported by `import newton`.

## Proposed solution

Lazy (deferred) imports for the solver backends via [PEP 562](https://peps.python.org/pep-0562/) module-level `__getattr__` in `newton/solvers.py` and `newton/_src/solvers/__init__.py`, e.g.:

```python
_SOLVER_MODULES = {
    "SolverFeatherstone": "._src.solvers.featherstone",
    "SolverImplicitMPM": "._src.solvers.implicit_mpm",
    "SolverKamino": "._src.solvers.kamino",
    "SolverMuJoCo": "._src.solvers.mujoco",
    # ...
}

def __getattr__(name):
    if name in _SOLVER_MODULES:
        module = importlib.import_module(_SOLVER_MODULES[name], __package__)
        return getattr(module, name)
    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
```

Regarding the concern that lazy imports "look messy": with PEP 562 the entire change is ~20 lines per `__init__`, the public API (`newton.solvers.SolverMuJoCo`) is unchanged, and static tooling keeps working via `if TYPE_CHECKING:` re-exports (this is the pattern numpy/scipy/jax use for exactly this reason). Alternatively `lazy_loader` (SPEC 1) provides the same with `.pyi` stub support.

Secondary follow-ups (smaller wins, independent of the above):

- `style3d` importing `vbd.rigid_vbd_kernels` means cloth users pay for the full VBD solver and vice versa — the shared `evaluate_body_particle_contact` could move to a small shared module.
- `implicit_mpm` could defer `warp.fem` / `warp.optim` to first use — both because of its own ~180+146 ms and because of the ~2× kernel-registration slowdown it induces on everything imported after it.
- `newton.sensors` (~81 ms) and `newton.viewer` (~19 ms) in `newton/__init__.py` are also candidates for PEP 562 laziness.

With lazy solvers, `import newton` should drop to roughly **0.5–0.6 s**, and `from newton.solvers import SolverMuJoCo` to ~0.65 s total.

<details>
<summary>Reproduction / methodology</summary>

Total + attribution:

```bash
python -X importtime -c "import newton" 2> importtime.log
```

Per-case timing (fresh subprocess each, median of 7):

```python
import subprocess, sys, statistics, time
code = 'import time; t0=time.perf_counter(); import newton; print((time.perf_counter()-t0)*1000)'
times = [float(subprocess.run([sys.executable, "-c", code], capture_output=True, text=True).stdout) for _ in range(7)]
print(statistics.median(times))
```

Isolated per-solver marginal cost (stub `newton` and `newton._src.solvers` in `sys.modules` with correct `__path__`/`__package__` so that importing one solver package does not execute the eager `__init__`s; pre-import `warp` and `newton._src.{core,geometry,sim,utils}`; then time `importlib.import_module("newton._src.solvers.<name>")`).

</details>

## Environment

- Newton: `main` @ `0b457032` (2026-07-22)
- Warp: 1.16.0.dev20260716
- Python: 3.12.13, Linux x86_64
- CPU: AMD EPYC 9B45, GPU: NVIDIA RTX PRO 6000 Blackwell (import timing is CPU-bound; warm FS cache)

eric-heiden:
## Description

`import newton` currently takes ~1.35 s, and about 60% of that is spent eagerly importing every solver backend: `newton/__init__.py` imports `newton.solvers`, which imports all solver packages. A user who only needs one solver (e.g. IsaacLab with `SolverMuJoCo`, whose own import cost is ~28 ms) still pays for all of them. Most of the cost is Warp kernel module creation at import time, plus `warp.fem` / `warp.optim` pulled in by `SolverImplicitMPM`.

This PR makes the solver backends load lazily. `newton/solvers.py` and `newton/_src/solvers/__init__.py` now resolve their public symbols on first attribute access via a module-level `__getattr__` ([PEP 562](https://peps.python.org/pep-0562/)), following the pattern already used in `newton/geometry.py`. Each solver is imported the first time it is actually referenced.

With this change, `import newton` drops from **~1.35 s to ~0.58 s** (warm cache, Python 3.12, Linux), and importing a single solver no longer imports the others.

Nothing changes for users:

- All symbols are still reachable exactly as before (`newton.solvers.SolverMuJoCo`, `from newton.solvers import SolverXPBD`, `newton.solvers.style3d.add_cloth_mesh`, ...), appear in `dir(newton.solvers)`, and are listed in `__all__`.
- The same imports are re-exported under `TYPE_CHECKING`, so IDEs and static type checkers resolve them as before.
- The experimental `newton.solvers.experimental.coupled` namespace keeps its eager registration, so `from newton.solvers.experimental.coupled import SolverCoupled` continues to work unchanged. (Deferring this one too is possible with an import-system hook, but it only costs ~55 ms and is left as a follow-up to keep this change simple.)

The one behavioral difference: code that relies on a solver *module* being loaded as a side effect of `import newton` (without ever referencing the solver) must import that module explicitly. A note was added to the changelog.

Closes #3600

## Checklist

## Test plan

- New `newton/tests/test_lazy_imports.py` verifies that a fresh `import newton` loads no solver backend module, that every `newton.solvers.__all__` symbol resolves and appears in `dir()`, and that the experimental coupled namespace imports in a fresh interpreter.
- Existing suites exercised locally on GPU:

```
uv run --extra dev -m newton.tests -k test_lazy_imports
uv run --extra dev -m newton.tests -k test_api
uv run --extra dev -m newton.tests -k test_coupled_solver
uv run --extra dev -m newton.tests -k test_admm_coupled_solver
uv run --extra dev -m newton.tests -k test_custom_solver
```

- Import-time measurements (median of 5 fresh interpreters, warm cache):

| | before | after |
|---|---:|---:|
| `import newton` | ~1350 ms | ~575 ms |

## Summary by CodeRabbit

* **Performance**
  * Faster `newton` imports by lazily loading solver backends only on first access.

* **Changed**
  * Solver symbols are now resolved dynamically while keeping the same public access via `newton.solvers`.
  * Solver names are improved for discovery (e.g., `dir()`), and non-exported symbols raise `AttributeError`.
  * `newton.solvers.experimental.coupled` is now lazily loaded on demand.

* **Tests**
  * Expanded coverage to verify public-surface consistency and lazy-import behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 - Compile tiled camera render kernels with CUDA fast math by default for faster rendering; set `SensorTiledCamera.render_config.enable_fast_math = False` for bit-exact, IEEE-precise output.
 - Optimize raycast/raytrace queries by restructuring ray-shape intersection into local-space primitives and compile specialized depth/shadow variants that skip unused surface-normal work (mesh shadows also use any-hit queries).
 - Improve `SolverKamino` GPU simulation and kernel compilation performance.
+- Load solver backends lazily on first access to speed up `import newton`; access solver classes through `newton.solvers` as before, and import solver modules explicitly if module-level side effects are required.
 - Speed up `ModelBuilder.replicate()` for large world counts by merging all copies in one pass; it no longer calls `add_world()` or `add_builder()` per copy, so `ModelBuilder` subclass overrides of those methods are not invoked during replication.
 
 ### Deprecated
--- a/newton/_src/solvers/__init__.py
+++ b/newton/_src/solvers/__init__.py
@@ -5,6 +5,7 @@ import importlib
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
+    from . import style3d
     from .featherstone import SolverFeatherstone
     from .flags import SolverNotifyFlags
     from .implicit_mpm import SolverImplicitMPM
@@ -27,32 +28,36 @@ __all__ = [
     "SolverStyle3D",
     "SolverVBD",
     "SolverXPBD",
+    "style3d",
 ]
 
-# Maps each public symbol to the submodule that defines it. Symbols are
+# Maps each public symbol to the module that provides it and the attribute to
+# fetch from that module (None returns the module itself). Symbols are
 # resolved on first attribute access (PEP 562) so that importing Newton does
 # not pay the import cost of every solver backend.
-_LAZY_IMPORTS = {
-    "SolverBase": ".solver",
-    "SolverFeatherstone": ".featherstone",
-    "SolverImplicitMPM": ".implicit_mpm",
-    "SolverKamino": ".kamino",
-    "SolverMuJoCo": ".mujoco",
-    "SolverNotifyFlags": ".flags",
-    "SolverSemiImplicit": ".semi_implicit",
-    "SolverStyle3D": ".style3d.solver_style3d",
-    "SolverVBD": ".vbd",
-    "SolverXPBD": ".xpbd",
+_LAZY_IMPORTS: dict[str, tuple[str, str | None]] = {
+    "SolverBase": (".solver", "SolverBase"),
+    "SolverFeatherstone": (".featherstone", "SolverFeatherstone"),
+    "SolverImplicitMPM": (".implicit_mpm", "SolverImplicitMPM"),
+    "SolverKamino": (".kamino", "SolverKamino"),
+    "SolverMuJoCo": (".mujoco", "SolverMuJoCo"),
+    "SolverNotifyFlags": (".flags", "SolverNotifyFlags"),
+    "SolverSemiImplicit": (".semi_implicit", "SolverSemiImplicit"),
+    "SolverStyle3D": (".style3d.solver_style3d", "SolverStyle3D"),
+    "SolverVBD": (".vbd", "SolverVBD"),
+    "SolverXPBD": (".xpbd", "SolverXPBD"),
+    "style3d": (".style3d", None),
 }
 
 
 def __getattr__(name: str):
     try:
-        module_name = _LAZY_IMPORTS[name]
+        module_name, attr_name = _LAZY_IMPORTS[name]
     except KeyError:
         raise AttributeError(f"module {__name__!r} has no attribute {name!r}") from None
 
-    value = getattr(importlib.import_module(module_name, __name__), name)
+    module = importlib.import_module(module_name, __name__)
+    value = module if attr_name is None else getattr(module, attr_name)
     globals()[name] = value
     return value
 
--- a/newton/_src/solvers/__init__.py
+++ b/newton/_src/solvers/__init__.py
@@ -1,16 +1,20 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 The Newton Developers
 # SPDX-License-Identifier: Apache-2.0
 
-from .featherstone import SolverFeatherstone
-from .flags import SolverNotifyFlags
-from .implicit_mpm import SolverImplicitMPM
-from .kamino import SolverKamino
-from .mujoco import SolverMuJoCo
-from .semi_implicit import SolverSemiImplicit
-from .solver import SolverBase
-from .style3d.solver_style3d import SolverStyle3D
-from .vbd import SolverVBD
-from .xpbd import SolverXPBD
+import importlib
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from .featherstone import SolverFeatherstone
+    from .flags import SolverNotifyFlags
+    from .implicit_mpm import SolverImplicitMPM
+    from .kamino import SolverKamino
+    from .mujoco import SolverMuJoCo
+    from .semi_implicit import SolverSemiImplicit
+    from .solver import SolverBase
+    from .style3d.solver_style3d import SolverStyle3D
+    from .vbd import SolverVBD
+    from .xpbd import SolverXPBD
 
 __all__ = [
     "SolverBase",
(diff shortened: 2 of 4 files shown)


adenzler-nvidia:
Thanks - yes this is also a problem in the benchmarks with the spawn option. Will review today.

adenzler-nvidia:
Nice speedup, and the mechanics (TYPE_CHECKING re-exports, `__dir__`, caching into `globals()`) look right. One structural suggestion: the symbol→module mapping now exists twice — `newton/solvers.py` and `newton/_src/solvers/__init__.py` each carry a `_LAZY_IMPORTS` map — so adding a solver means updating both maps plus both `TYPE_CHECKING` blocks and both `__all__` lists, and the public file now hardcodes internal layout (e.g. `._src.solvers.style3d`), which it didn't before this PR.

Since `newton/solvers.py` re-exports exactly the public surface of `_src.solvers`, it could delegate wholesale and let the internal package own the single mapping:

```python
# newton/solvers.py
def __getattr__(name: str):
    if name not in __all__:
        raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
    value = getattr(importlib.import_module("._src.solvers", __package__), name)
    globals()[name] = value
    return value


def __dir__() -> list[str]:
    return sorted(set(globals()) | set(__all__))
```

with your tuple-valued map moved into `newton/_src/solvers/__init__.py` so it can also serve the `style3d` submodule:

```python
# newton/_src/solvers/__init__.py
_LAZY_IMPORTS: dict[str, tuple[str, str | None]] = {
    "SolverBase": (".solver", "SolverBase"),
    "SolverFeatherstone": (".featherstone", "SolverFeatherstone"),
    "SolverImplicitMPM": (".implicit_mpm", "SolverImplicitMPM"),
    "SolverKamino": (".kamino", "SolverKamino"),
    "SolverMuJoCo": (".mujoco", "SolverMuJoCo"),
    "SolverNotifyFlags": (".flags", "SolverNotifyFlags"),
    "SolverSemiImplicit": (".semi_implicit", "SolverSemiImplicit"),
    "SolverStyle3D": (".style3d.solver_style3d", "SolverStyle3D"),
    "SolverVBD": (".vbd", "SolverVBD"),
    "SolverXPBD": (".xpbd", "SolverXPBD"),
    "style3d": (".style3d", None),
}


def __getattr__(name: str):
    try:
        module_name, attr_name = _LAZY_IMPORTS[name]
    except KeyError:
        raise AttributeError(f"module {__name__!r} has no attribute {name!r}") from None

    module = importlib.import_module(module_name, __name__)
    value = module if attr_name is None else getattr(module, attr_name)
    globals()[name] = value
    return value
```

(`experimental` stays a real module global, so it never reaches `__getattr__`.) With this, the knowledge of which submodule defines what lives in one place, `newton/solvers.py` needs no changes when a solver is added or moved, and `test_lazy_attributes_resolve` exercises the full public→internal chain for every symbol. I verified this variant against all the access paths your test covers (attribute access, from-import, `style3d` module access, `dir()`, star-import, and that touching one solver doesn't load the others).

StafaH (approved):

adenzler-nvidia (approved):
Thanks for consolidating the maps — nice catch on the eager `experimental.coupled` import too. LGTM.

PR #3602, opened 2026-07-22
[Kamino] Propagate model changes to the Kamino FK solver

rubengrandia:
## Description

Make `SolverKamino.notify_model_changed()` propagate model updates into the forward-kinematics solver internal model copies.

The FK solver now refreshes copied joint frames, synthetic tie-rod axis frames, and default floating-base poses after the corresponding model notifications. It also validates FK actuation overrides without rejecting explicit root joints that FK always replaces with an actuated free joint.

## Checklist

## Test plan

```bash
uv run --extra dev -m unittest newton._src.solvers.kamino.tests.test_solver_kamino_notify
uv run --extra dev -m unittest newton._src.solvers.kamino.tests.test_solvers_forward_kinematics.FourBarTieRodRandomPosesCheckForwardKinematics.test_axis_joint_frames_update_after_notify
uv run --extra dev -m unittest newton._src.solvers.kamino.tests.test_core_model.TestModelConversions.test_05_model_conversions_base_assignment_non_floating_root
uvx pre-commit run -a
```

## Summary by CodeRabbit

* **New Features**
  * Forward-kinematics frames now automatically refresh after model property changes (including joint/body/CoM and FK axis-joint updates).
  * Added GPU-backed validation and resolution of per-joint FK actuation overrides during model updates.
* **Bug Fixes**
  * Prevented unsupported/incompatible FK actuation partition and override changes.
  * Fixed stale FK joint frames (axis-joint frames) and default base pose after in-place notifications.
* **Tests**
  * Added coverage for axis-joint frame updates after notifications and FK override/constraint behavior.
  * Adjusted dense-vs-sparse Jacobian residual acceptance threshold for reliability.

### DIFF
--- a/newton/_src/solvers/kamino/_src/solvers/fk/kernels.py
+++ b/newton/_src/solvers/kamino/_src/solvers/fk/kernels.py
@@ -31,6 +31,9 @@ from .types import FKJointDoFType
 __all__ = [
     "_add_regularizer_to_diagonal",
     "_apply_line_search_step",
+    "_compute_base_q_default",
+    "_compute_fk_axis_joint_frames",
+    "_compute_fk_joint_frames",
     "_correct_actuator_coords",
     "_correct_universal_constraint_velocities",
     "_eval_actuator_coords",
@@ -51,6 +54,7 @@ __all__ = [
     "_newton_check",
     "_reset_state",
     "_reset_state_base_q",
+    "_resolve_fk_actuation_types",
     "_update_cg_tolerance_kernel",
     "create_1d_tile_based_kernels",
     "create_2d_tile_based_kernels",
@@ -59,6 +63,7 @@ __all__ = [
     "create_eval_joint_constraints_sparse_jacobian_kernel",
     "create_eval_min_num_iterations_kernel",
     "read_quat_from_array",
+    "validate_fk_actuation_updates",
 ]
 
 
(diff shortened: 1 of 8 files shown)


newton/_src/solvers/kamino/_src/solvers/fk/kernels.py
```diff
@@ -94,11 +99,155 @@ def read_quat_from_array(array: wp.array[wp.float32], offset: int, normalize: bo
...
+    axis_joint_1: wp.array[wp.int32],
+    model_joint_bid_B: wp.array[wp.int32],
+    model_joint_B_r_Bj: wp.array[wp.vec3f],
+    model_joint_F_r_Fj: wp.array[wp.vec3f],
+    model_body_q_0: wp.array[wp.transformf],
+    # Outputs
+    fk_X_Bj: wp.array[wp.mat33f],
+    fk_X_Fj: wp.array[wp.mat33f],
+):
+    """Compute synthetic axis-joint frames from the model data."""
```
Guirec-Maloisel, line 214: The comments in the Python code (in solver.py) that this kernel replaces should be carried over here to clarify what is happening (computing the position of the two spherical joints at the tie rod's ends on the initial pose, computing a joint frame so that the X axis is the vector that connects both spherical joints, etc.)

newton/_src/solvers/kamino/_src/solvers/fk/kernels.py
```diff
@@ -94,11 +99,155 @@ def read_quat_from_array(array: wp.array[wp.float32], offset: int, normalize: bo
...
+    model_F_r_Fj: wp.array[wp.vec3f],
+    model_X_Bj: wp.array[wp.mat33f],
+    model_X_Fj: wp.array[wp.mat33f],
+    # Outputs
+    fk_B_r_Bj: wp.array[wp.vec3f],
+    fk_F_r_Fj: wp.array[wp.vec3f],
+    fk_X_Bj: wp.array[wp.mat33f],
+    fk_X_Fj: wp.array[wp.mat33f],
+):
+    """Compute FK joint frames from the current model data."""
```
Guirec-Maloisel, line 181: Would add back a comment here as well, explaining that the FK joint frame is copied from the model when applicable, and set to the identity frame for FK-specific joints (base joint if added by FK, axis joints, the orientation being overwritten by another kernel for axis joints)

newton/_src/solvers/kamino/_src/solvers/fk/kernels.py
```diff
@@ -94,11 +99,155 @@ def read_quat_from_array(array: wp.array[wp.float32], offset: int, normalize: bo
...
+    model_base_joint: wp.array[wp.int32],
+    model_base_body: wp.array[wp.int32],
+    model_body_q_0: wp.array[wp.transformf],
+    # Outputs
+    base_q_default: wp.array[wp.transformf],
+):
+    """Compute the default FK base pose from the model data."""
+    world = wp.tid()
+    if model_base_joint[world] >= 0:
+        base_q_default[world] = wp.transform_identity()
```
Guirec-Maloisel: I'm realizing that here we should use the initial joint coordinates from the model here instead of the identity (this is only valid if the joint frame on the base and follower body coincide on the initial pose; but in newton the convention is now for "root" free joints to have identity frame and zero offset on both the base (=world) and follower, so that joint coordinates are always interpretable as the body pose).

This wasn't caught because the implementation of the reset() functionality in solver_kamino_impl.py will itself resolve the correct base_q and not call the FK solver leaving it as None, but should be fixed here as well. We might actually be able to directly use the get_base_q_from_joint_q_and_body_q() function in kinematics/resets.py here instead of a new kernel.

If we do need a new kernel, I'd add a comment that this computes the default base pose to match the initial body poses, using the initial base joint coordinates if a base joint is set, else the initial base body coordinates if the base joint was added by FK.

newton/_src/solvers/kamino/_src/solvers/fk/solver.py
```diff
@@ -376,17 +348,13 @@ def eval_joint_pos_init(jt_id_prev):
                 joints_act_type.append(JointActuationType.FORCE)
                 joints_bid_B.append(-1)
                 joints_bid_F.append(base_body_id)
-                joints_B_r_Bj.append(np.zeros(3, dtype=np.float32))
-                joints_F_r_Fj.append(np.zeros(3, dtype=np.float32))
-                joints_X_Bj.append(np.eye(3, 3, dtype=np.float32))
-                joints_X_Fj.append(np.eye(3, 3, dtype=np.float32))
+                joints_source_id.append(-1)
                 joints_num_actuated_coords.append(7)
                 # Note: we rely on the initial body orientations being identity
```
Guirec-Maloisel: I'm realizing that this note (lines 384-386 in the old code) is stale, since joint frame on base and follower are now cleanly separated. Could you remove it as part of the PR?

newton/_src/solvers/kamino/_src/solvers/fk/solver.py
Guirec-Maloisel, line 240: I would add a line here specifying that we first resolve discrete information (such as joint types and indices), and copy/compute continuous data (such as joint frames) in a second time.

newton/_src/solvers/kamino/_src/solvers/fk/solver.py
```diff
@@ -578,16 +546,27 @@ def eval_joint_pos_init(jt_id_prev):
...
+            self.fk_axis_source_joint_1 = to_warp_int32_array(fk_axis_source_joint_1)
+            self.num_axis_joints = len(fk_axis_joint)
             self.base_joint_id = to_warp_int32_array(base_joint_ids)
 
             # Default base state
-            self.base_q_default = wp.from_numpy(base_q_default, dtype=wp.transformf)
+            self.base_q_default = wp.zeros(shape=self.num_worlds, dtype=wp.transformf)
             self.base_u_default = wp.zeros(shape=(self.num_worlds,), dtype=wp.spatial_vectorf)
 
+            # Initialize mutable model-derived values through the same path used by runtime notifications.
```
Guirec-Maloisel: This comment could be a bit confusing at first read, I would more simply write "Initialize continuous joint data (e.g. joint frames)"

newton/_src/solvers/kamino/_src/solvers/fk/solver.py
```diff
@@ -931,6 +910,120 @@ def _cg_matvec_flat(x, y, world_active):
...
             )
 
+    def validate_model_changed(self, flags: ModelFlags | int) -> None:
+        """Validate FK structural invariants before model values are updated.
+
+        Args:
+            flags: Bitmask indicating which model properties changed.
+
+        Raises:
+            RuntimeError: If the effective FK actuation partition changed.
```
Guirec-Maloisel: What "actuation partition" means is not fully obvious on first read, I'd reformulate as "If the effective set of joints that are actuated for FK changed"; and for the error message below ("Changing the actuated vs passive status of joint X for FK is not supported")

newton/_src/solvers/kamino/_src/solvers/fk/solver.py
```diff
@@ -931,6 +910,120 @@ def _cg_matvec_flat(x, y, world_active):
...
+        Args:
+            flags: Bitmask indicating which model properties changed.
+        """
+        if flags & (ModelFlags.JOINT_PROPERTIES | ModelFlags.BODY_INERTIAL_PROPERTIES):
+            self._update_joint_frames()
+
+        if flags & (ModelFlags.JOINT_PROPERTIES | ModelFlags.BODY_PROPERTIES | ModelFlags.BODY_INERTIAL_PROPERTIES):
+            self._update_axis_joint_frames()
+
+        if flags & (ModelFlags.BODY_PROPERTIES | ModelFlags.BODY_INERTIAL_PROPERTIES):
```
Guirec-Maloisel: If updating _update_base_q_default() to read also from initial joint coordinates (cf other comment), we should also add ModelFlags.JOINT_PROPERTIES for this one.

newton/_src/solvers/kamino/_src/solvers/fk/solver.py
Guirec-Maloisel: I'd also take this out of the "with wp.ScopedDevice(self.device):" code block (because it's not related to it), and rather have it unindented below as a separate block

newton/_src/solvers/kamino/_src/solvers/fk/solver.py
```diff
@@ -214,31 +237,32 @@ def finalize(self, model: ModelKamino | None = None, config: ForwardKinematicsSo
...
-        base_q_default = np.zeros(7 * self.num_worlds, dtype=np.float32)  # Default base pose
-        bodies_q_0 = self.model.bodies.q_i_0.numpy()
         base_joint_ids = self.num_worlds * [-1]  # Base joint id per world
         base_joint_ids_input = self.model.info.base_joint_index.numpy().tolist()
         base_body_ids_input = self.model.info.base_body_index.numpy().tolist()
+        for base_joint_id in base_joint_ids_input:
+            if base_joint_id >= 0:
+                # FK always replaces an explicit base joint with an actuated free joint.
+                built_fk_actuated[base_joint_id] = -1
+        self._built_fk_actuated = to_warp_int32_array(built_fk_actuated, device=self.device)
```
Guirec-Maloisel: These two allocations should go below in the block that does all allocations (with wp.ScopedDevice(self.device):), with a comment above (e.g. "# Helper data for model updates validation")

newton/_src/solvers/kamino/_src/solvers/fk/solver.py
```diff
@@ -183,13 +189,30 @@ def finalize(self, model: ModelKamino | None = None, config: ForwardKinematicsSo
...
+                dim=self.model.size.sum_of_num_joints,
+                inputs=[
+                    self.model.joints.act_type,
+                    self.model.joints.fk_act_flag,
+                    resolved_act_type,
+                ],
+                device=self.device,
+            )
+        joints_act_type_prev = resolved_act_type.numpy()
+        built_fk_actuated = (joints_act_type_prev != JointActuationType.PASSIVE).astype(np.int32)
```
Guirec-Maloisel, line 217: Please add a comment here and/or in the validation kernel to explain the convention used for build_fk_actuated (0 for passive, 1 for actuated, -1 to skip the validation, indexed per joint in the model)

Guirec-Maloisel (changes requested):
Great work, I was worried about how much we'd need to change to make the FK solver updatable without having to rebuild it but the scope of changes is not that large in the end! Seems correct in terms of functionality, and the testing is quite extensive as well. I just had a few comments regarding code style / readability.

newton/_src/solvers/kamino/_src/solvers/fk/kernels.py
rubengrandia: Done

newton/_src/solvers/kamino/_src/solvers/fk/kernels.py
rubengrandia: Done

newton/_src/solvers/kamino/_src/solvers/fk/kernels.py
rubengrandia: Done

newton/_src/solvers/kamino/_src/solvers/fk/solver.py
rubengrandia: Done

newton/_src/solvers/kamino/_src/solvers/fk/solver.py
rubengrandia: Done

newton/_src/solvers/kamino/_src/solvers/fk/solver.py
rubengrandia: Done

rubengrandia: Done

newton/_src/solvers/kamino/_src/solvers/fk/solver.py
rubengrandia: Done

newton/_src/solvers/kamino/_src/solvers/fk/solver.py
rubengrandia: Done

newton/_src/solvers/kamino/_src/solvers/fk/solver.py
rubengrandia: Done

newton/_src/solvers/kamino/_src/solvers/fk/solver.py
rubengrandia: Done

newton/_src/solvers/kamino/_src/solvers/fk/kernels.py
```diff
@@ -359,20 +503,6 @@ def _eval_actuator_coords(
     _joint_transform_to_coords(dof_type, pos_rel, q_rel, coord_id, actuators_q)
 
 
-@wp.func
-def _correct_joint_angle(angle: wp.float32, angle_ref: wp.float32) -> wp.float32:
```
rubengrandia, line 363: @Guirec-Maloisel , initially I wanted to move these to avoid a circular dependency with the resets.py importing these helpers. But then I found there are duplicates in kinematics/joints.py. I import those now for use in the FK solver.

Guirec-Maloisel: Ah nice, trying to avoid this code duplication was on my todo list. I recreated the correction functions in FK because I didn't care about the limit part, but your solution of having both versions in joints.py is cleaner.

Guirec-Maloisel (approved):
Good to go from my side now!

PR #3604, opened 2026-07-22
Gate benchmark CI on runtime metrics

vreutskyy:
## Description

Gate AWS pull-request benchmark comparisons on directly measured runtime results. Diagnostic and derived tracks remain available as ASV results for dashboard reporting, but no longer determine the `asv continuous` exit status.

This prevents expected diagnostic changes from blocking unrelated work. In #3575, `FastG1.track_solver_niter_max` changed from 5 to 6 while both mean and p95 runtime improved, yet the AWS benchmark job failed.

## Checklist

- [x] `CHANGELOG.md` has been updated (not required; this only changes PR CI gating)

## Test plan

- `uv run --extra dev -m newton.tests -k test_aws_benchmark_comparison_gates_only_runtime_metrics`
- `uv run --extra dev -m newton.tests -k test_benchmark_simulation`
- `uvx --with virtualenv asv check`
- `uvx pre-commit run -a`

## Bug fix

**Steps to reproduce:**

1. Run the AWS benchmark comparison for #3575.
2. Observe that the only significant change is `FastG1.track_solver_niter_max` from 5 to 6.
3. Observe that the workflow exits with status 2 despite slightly improved mean and p95 runtime.

**Minimal reproduction:**

```bash
uvx --with virtualenv asv continuous \
  --launch-method spawn \
  --interleave-rounds \
  --append-samples \
  --no-only-changed \
  -e -b '^(?!.*track_(simulation_steps_per_second|real_time_factor)).*Fast' \
  <base-ref> <ref>
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated AWS GPU benchmark comparison gating to focus on directly measured runtime/timing metrics only, excluding overhead and dashboard-only metrics.
* **Tests**
  * Added a workflow-selection validation test to ensure the benchmark metrics filters continue matching only the intended runtime metrics.

### DIFF
--- a/.github/workflows/aws_gpu_benchmarks.yml
+++ b/.github/workflows/aws_gpu_benchmarks.yml
@@ -137,13 +137,13 @@ jobs:
 
       - name: Run Benchmarks
         run: |
-          # ASV assumes smaller values are better, so compare only metrics with that direction.
+          # Gate this PR job on directly measured runtimes only.
           uvx --with virtualenv asv continuous \
             --launch-method spawn \
             --interleave-rounds \
             --append-samples \
             --no-only-changed \
-            -e -b '^(?!.*track_(simulation_steps_per_second|real_time_factor)).*Fast' \
+            -e -b '^(?!.*FastNewtonOverhead).*Fast.*\.(?:time_.*|track_(?:simulate|(?:mean|p95)_.*(?:time|ms))(?:\(|$))' \
             ${{ inputs.base_ref }} \
             ${{ inputs.ref }}
         continue-on-error: true
--- a/.github/workflows/aws_gpu_benchmarks.yml
+++ b/.github/workflows/aws_gpu_benchmarks.yml
@@ -143,7 +143,10 @@ jobs:
             --interleave-rounds \
             --append-samples \
             --no-only-changed \
-            -e -b '^(?!.*FastNewtonOverhead).*Fast.*\.(?:time_.*|track_(?:simulate|(?:mean|p95)_.*(?:time|ms))(?:\(|$))' \
+            -e \
+            -b 'Fast\w*\.time_' \
+            -b 'Fast(?!NewtonOverhead)\w*\.track_simulate(\(|$)' \
+            -b 'Fast\w*\.track_(mean|p95)_\w*(time|ms)(\(|$)' \
             ${{ inputs.base_ref }} \
             ${{ inputs.ref }}
         continue-on-error: true
--- a/newton/tests/test_benchmark_simulation.py
+++ b/newton/tests/test_benchmark_simulation.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import json
+import re
 import sys
 import unittest
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


newton/tests/test_benchmark_simulation.py
```diff
@@ -183,6 +184,41 @@ def test_kpi_dr_legs_setup_cache_timeout_exceeds_default(self):
...
+        dashboard_benchmarks = (
+            "simulation.bench_mujoco.FastG1.track_solver_niter_mean(8192)",
+            "simulation.bench_mujoco.FastG1.track_solver_niter_max(8192)",
+            "simulation.bench_mujoco.FastG1.track_simulation_steps_per_second(8192)",
+            "simulation.bench_mujoco.FastG1.track_real_time_factor(8192)",
+            "simulation.bench_mujoco.FastG1.track_steady_state_gpu_memory(8192)",
+            "simulation.bench_mujoco.FastG1.track_sim_dt(8192)",
+            "simulation.bench_mujoco.FastG1.track_sim_substeps(8192)",
+            "simulation.bench_mujoco.FastNewtonOverheadG1.track_simulate(8192)",
+            "simulation.bench_teleop_mujoco.FastTeleopMuJoCo.track_step_rate_hz('graph')",
```
adenzler-nvidia: `FastTeleopMuJoCo` doesn't define `track_step_rate_hz` — that track only exists on `_RealtimePhysicsBenchmark` in `bench_mujoco.py`, and the teleop dashboard-only tracks all live on the nightly `TeleopMuJoCo` class. So this entry asserts against a benchmark name that never exists and passes vacuously. Suggest replacing it with a real nightly-only name such as `simulation.bench_teleop_mujoco.TeleopMuJoCo.track_frame_overrun_pct('graph')`, which would also pin that the pattern keeps non-`Fast` classes out of the PR job.

.github/workflows/aws_gpu_benchmarks.yml
```diff
@@ -137,13 +137,13 @@ jobs:
...
         run: |
-          # ASV assumes smaller values are better, so compare only metrics with that direction.
+          # Gate this PR job on directly measured runtimes only.
           uvx --with virtualenv asv continuous \
             --launch-method spawn \
             --interleave-rounds \
             --append-samples \
             --no-only-changed \
-            -e -b '^(?!.*track_(simulation_steps_per_second|real_time_factor)).*Fast' \
+            -e -b '^(?!.*FastNewtonOverhead).*Fast.*\.(?:time_.*|track_(?:simulate|(?:mean|p95)_.*(?:time|ms))(?:\(|$))' \
```
adenzler-nvidia: `-b` can be repeated (argparse `append` in asv; the patterns are OR'd), so this could be split into one pattern per family, which reads easier than the combined lookahead/alternation and keeps the `NewtonOverhead` exclusion scoped to the only family it applies to:

```yaml
            -e \
            -b 'Fast\w*\.time_' \
            -b 'Fast(?!NewtonOverhead)\w*\.track_simulate(\(|$)' \
            -b 'Fast\w*\.track_(mean|p95)_\w*(time|ms)(\(|$)' \
```

I checked that selection is identical over the current `Fast*` suite. The test would then collect all `-b` patterns and assert any-of, mirroring asv's matching.

newton/tests/test_benchmark_simulation.py
vreutskyy: Good catch. Replaced the nonexistent `FastTeleopMuJoCo.track_step_rate_hz` identifier with the real nightly-only `TeleopMuJoCo.track_frame_overrun_pct`, so the test now explicitly verifies that non-`Fast` dashboard tracks remain outside the PR job.

.github/workflows/aws_gpu_benchmarks.yml
vreutskyy: Done. Split the selector into the three repeated `-b` patterns and updated the test to collect every pattern and apply ASV’s any-pattern matching semantics. The focused test, full benchmark simulation module, ASV discovery, and pre-commit all pass at `dfd4acfd`.

adenzler-nvidia (approved):

PR #3605, opened 2026-07-22
[Kamino] Fix Kamino preserve reset for COM-offset bodies

rubengrandia:
## Description

Fix `SolverKamino.reset()` with `ResetConfig.preserve()` corrupting `joint_q` for bodies with non-zero center-of-mass offsets.

Preserve reset skips body pose updates but still re-derives joint coordinates from body state. The Newton wrapper previously skipped origin↔COM conversion on preserve paths, so Kamino inferred joints from origin-frame `body_q` as if it were COM-frame, shifting `joint_q` by the COM offset.

This PR always converts body poses between Newton body-origin and Kamino CoM frames around reset (matching `step()`), adds a regression test, and clarifies `ResetConfig` docs that joint coordinates are re-derived from the resulting body state.

## Checklist

## Test plan

```bash
uv run --extra dev python -m newton._src.solvers.kamino.tests.test_core_model \
  TestModelConversions.test_15_preserve_reset_keeps_joint_q_consistent_with_com_offset_body

uv run --extra dev python -m newton._src.solvers.kamino.tests.test_core_model \
  TestModelConversions.test_12_reset_produces_body_origin_frame \
  TestModelConversions.test_13_base_reset_produces_body_origin_frame

uv run --extra dev python -m newton._src.solvers.kamino.tests.test_solver_kamino \
  TestSolverKaminoImpl.test_06_reset_but_preserve_state
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Kamino solver reset behavior to consistently preserve body poses and joint coordinates, including models with non-zero center-of-mass offsets.
  - Ensured state is consistently converted back to the expected Newton reference frame after resets.
- **Documentation**
  - Clarified the `Preserve` reset option wording to reference preserving current body/base values.
  - Updated reset example comments for better clarity.
- **Tests**
  - Added coverage to verify preserve resets keep `body_q` and `joint_q` unchanged.

### DIFF
--- a/newton/_src/solvers/kamino/solver_kamino.py
+++ b/newton/_src/solvers/kamino/solver_kamino.py
@@ -386,11 +386,6 @@ class SolverKamino(SolverBase, CouplingInterface):
         Configuration for a call to the reset() operation, specifying the behaviour (common or separate)
         for body poses, body velocities as well as floating base pose and velocity.
 
-        After body poses and velocities are updated (or preserved), joint coordinates and velocities
-        are always re-derived from the resulting body state for consistency. Joint constraint forces
-        are reset to zero. When body and joint state are already consistent, re-derived joint values
-        match the preserved inputs.
-
         Example
         -------
 
@@ -709,6 +704,11 @@ class SolverKamino(SolverBase, CouplingInterface):
 
         All state components are reset consistently with the new body poses and velocities
         (unless prescribed otherwise by state flags), and solver-internal buffers are cleared.
+        More specifically, joint coordinates and velocities are re-derived from the
+        resulting body state for consistency, and joint constraint forces are reset to
+        zero. If flags exclude :attr:`~newton.StateFlags.JOINT_Q` or
+        :attr:`~newton.StateFlags.JOINT_QD`, the corresponding joint coordinates or
+        velocities are restored after the reset instead.
 
         Args:
             state: The simulation state to reset (modified in place).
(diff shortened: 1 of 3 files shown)


newton/_src/solvers/kamino/solver_kamino.py
```diff
@@ -386,6 +386,11 @@ class ResetConfig:
         Configuration for a call to the reset() operation, specifying the behaviour (common or separate)
         for body poses, body velocities as well as floating base pose and velocity.
 
+        After body poses and velocities are updated (or preserved), joint coordinates and velocities
```
Guirec-Maloisel: I think this comment belongs in the docstring of reset() rather than of ResetConfig(), since it describes the behavior of the operation (and that part can't be configured).

There is already a sentence there "All state components are reset consistently with the new body poses and velocities (unless prescribed otherwise by state flags), and solver-internal buffers are cleared.", so we could expand on that sentence instead (e.g. add "More specifically, joint coordinates and velocities are always re-derived from the resulting body state for consistency, and joint constraint forces are reset to zero."

Guirec-Maloisel (commented):
Great find, it seems that with all the combinatorial possibilities of the reset config I had overlooked this case. Looks good to me, I just had a comment on the placement of a docstring.

newton/_src/solvers/kamino/solver_kamino.py
```diff
@@ -386,6 +386,11 @@ class ResetConfig:
...
+        are always re-derived from the resulting body state for consistency. Joint constraint forces
+        are reset to zero. When body and joint state are already consistent, re-derived joint values
+        match the preserved inputs.
+
```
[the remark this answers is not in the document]
rubengrandia: Integrated in the updated doc

newton/_src/solvers/kamino/solver_kamino.py
rubengrandia: Moved the comment to the `reset()` doc as you suggested. I added your sentence, and also added a comment on how people can overwrite this by using the state flags.

Guirec-Maloisel (approved):

PR #3609, opened 2026-07-23
Adopt Towncrier changelog fragments

### LINKED ISSUE #3607: Use changelog deltas
When a PR has a merge conflict in most cases it is because of the CHANGELOG.md changes that GitHub is unable to merge because it doesn't follow the `merge=union` rule. This currently requires us to have the PR author fix the trivial merge conflict so the maintainer can merge the PR (if the maintainer fixed the conflict, they couldn't merge the PR even after they approved it because they made the last commit).

Isaac Lab is using a changelog deltas compilation script that works well for them, we should investigate something similar using a library such as [towncrier](https://pypi.org/project/towncrier/) or develop our own solution.

eric-heiden:
## Description

Replaces the initial custom changelog implementation with the Towncrier approach discussed in #3607 and this PR's review thread.

- Configure Towncrier 25.8.0 with issue-linked numeric fragments and leading-`+` orphan fragments for changes without an issue.
- Support multiple entries and categories by sharing one logical identifier across fragment files.
- Keep Towncrier out of Newton's runtime dependencies and `uv.lock`; pin it in pre-commit and invoke the same version from CI and release commands.
- Add a small standard-library policy adapter for `.skip` files, pull-request scope, release backports, and changelog-only `release-management` updates.
- Keep development rendering non-mutating with `towncrier build --draft`; build once on the release branch, consume shipped fragments, and synchronize the exact build commit to `main` while retaining post-cut fragments.
- Preserve the existing `[Unreleased]` content for the first Towncrier release transition.
- Update contributor guidance, CI, CodeRabbit instructions, release documentation, and release skills. Repository-tool tests live under `scripts/tests`, outside the shipped package.

## Checklist

- [x] One logical fragment set or a one-line `.skip` reason has been added under `changelog.d/`

## Test plan

- `uv run --no-project --with towncrier==25.8.0 python -m unittest scripts.tests.test_changelog_policy -v`
- `uv run --no-project python .claude/skills/release-audit/scripts/test_cleanup_report.py -v`
- `uv run --no-project python scripts/changelog_policy.py check-pr --base-ref origin/main --target-branch main`
- `uvx --from towncrier==25.8.0 towncrier build --draft --version 0.0.0 --date 1970-01-01`
- `uvx --from pre-commit==4.2.0 pre-commit run -a`

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for creating, validating, previewing, and managing changelog fragments.
  * Updated release procedures to preserve historical entries and support rolling release-note updates.

* **Chores**
  * Introduced automated changelog validation and draft previews for pull requests and release workflows.
  * Added ownership and checklist rules for changelog contributions.
  * Added secure cleanup for temporary release-audit reports.

* **Tests**
  * Expanded coverage for changelog policies, release workflows, and report cleanup.


newton/tests/test_changelog.py
shi-eric, line 1: Is it a good idea to have a meta test that tests repo utilities placed in newton/tests, which is also shipped with the wheel?

scripts/changelog.py
shi-eric, line 1: Was there a reason why towncrier didn’t work as an existing solution?

AGENTS.md
shi-eric, line 1: There might too much information placed in AGENTS.md about how to update the changelog. Why not use the fragment directory’s README.md as the source of truth and you could have one or two instructions here that instruct the agents to follow the process outlined in that file?

scripts/changelog.py
eric-heiden: Yes, towncrier wants one file per category, like `xyz.fixed`, `xyz.added`, etc. That might work when xyz is the PR number but then this means you first create a PR and then need to push a second time to add these changelog fragments. Instead, I decided to follow Isaac Lab's approach and allow each PR to have a single changelog fragment with a name that doesn't have to be the PR number - the changelog.py script creates the file for you with a unique hex code appended. I thought if we had multiple files of this sort this would look very messy - it would only look neat if the file names had the PR number but that is not possible with a single-commit PR.

Other, less important, reasons are: we can consolidate the changelog at any time just using the `changelog.py build` command. We also introduce no additional dependency, which I think also makes sense for this rather trivial functionality.

newton/tests/test_changelog.py
eric-heiden: Probably not, but I at least wanted to test this fragment compilation somehow. Maybe a better way is to just move this under `scripts/` or so? It definitely doesn't need to run like a regular unit test, I will adapt this tomorrow.

scripts/changelog.py
adenzler-nvidia: I think Towncrier can support the single-commit workflow after all. It supports orphan fragments, so contributors could hand-create files such as `changelog.d/+camera-rays-a1b2c3d4.added.md` before the PR exists. A multi-category PR would use the same random identifier with additional types, such as `.deprecated.md` or `.fixed.md`. Developers would not need to invoke Towncrier locally; CI would run `towncrier check`.

During a scheduled rollup or release, a bot could find the commit that introduced each orphan fragment, resolve that commit to its merged PR using GitHub’s associated-pull-requests API, rename the fragments in its temporary worktree to names such as `3609.added.md`, and immediately run `towncrier build`. Because the build consumes the fragments, those intermediate renames never need a separate commit. If a commit cannot be mapped unambiguously, the fragment can remain orphaned or the rollup can request manual resolution.

This would preserve single-commit contributor PRs, produce PR-linked changelog entries, support multiple categories, and leave parsing and release generation to Towncrier. Towncrier would only be a pinned CI/release tool rather than a Newton runtime dependency. Would you be open to exploring this before we maintain the custom parser and reconciliation workflow?

shi-eric: I still don't like PR-linked changelog entries. We should really move to issue-linked changelog entries.

adenzler-nvidia: we can certainly make it a requirement that everything needing a changelog entry also deserves an issue.

shi-eric: Hmm, the flexibility to have pull requests with changelog entries that don't have a linked issue still might be nice to preserve. If the fragment doesn't have an issue in the name then just don't create a link in the changelog.

adenzler-nvidia: Yes, Towncrier supports exactly that distinction, so I think we can simplify my earlier proposal and drop the bot-based PR-number rewriting.

An issue-linked change would use a fragment such as `changelog.d/3607.added.md`, with `issue_format` configured to link `#3607` to the corresponding GitHub issue. A change without an issue would use an orphan fragment such as `changelog.d/+camera-rays-a1b2c3d4.added.md`; Towncrier includes its content without adding a link.

We can configure `issue_pattern = "\\d+"` to require numeric identifiers for non-orphan fragments. Towncrier 24.8 and later explicitly exempt leading-`+` orphan fragments from that check. Multiple categories use the same identifier, for example `3607.added.md` and `3607.deprecated.md`, or their orphan equivalents.

This preserves issue-linked release notes where appropriate while retaining flexibility for changes that do not need an issue. Developers can hand-create the fragments before opening the PR, CI can run `towncrier check`, and Towncrier can handle the eventual build. A small additional CI check would only be necessary if we want to verify that a numeric identifier refers to an existing issue rather than a pull request.

shi-eric: `towncrier check` can probably be added to `pre-commit`, right? Ideally, `towncrier` won't require a `uv.lock`: Its config will be stored in `pyproject.toml` and its pin will be in `.pre-commit-config.yaml` (similar to our use of `typos`). 

My understanding is that no OSRB request should be required in principle for towncrier usage since it doesn't interact with executable Newton code in the same process.

adenzler-nvidia: We should double check but I think that's correct. Either way it's MIT licensed so should be fine?

eric-heiden: I think it would be better to follow one pattern for the fragment names *if* we decide to have a bot that renames these fragments (which may not be necessary, perhaps a bunch of fragments like `changelog.d/+camera-rays-a1b2c3d4.added.md` doesn't look as messy as I originally thought). So in that case it should be the PR number for the file name, not sometimes the issue number, sometimes the PR number.

eric-heiden:
Updated this PR to pursue the Towncrier approach from the review discussion. The branch history is now a single commit on current main; the custom parser, accumulator, provenance, release/reconcile commands, package-shipped tests, and comparison artifacts are gone. Numeric fragments link issues, leading-+ fragments remain unlinked, Towncrier stays out of uv.lock, and the release-branch synchronization behavior is covered by integration tests.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -7,7 +7,7 @@
 
 - [ ] New or existing tests cover these changes
 - [ ] The documentation is up to date with these changes
-- [ ] `CHANGELOG.md` has been updated (if user-facing change)
+- [ ] One logical fragment set or a one-line `.skip` reason has been added under `changelog.d/`
```
shi-eric: Probably would be helpful to have a hyperlink to the `changelog.d/README.md` for instructions

.claude/skills/release-audit/scripts/cleanup_report.py
```diff
@@ -0,0 +1,39 @@
...
+
+def cleanup_report(path: Path, *, temporary_directory: Path | None = None) -> None:
+    """Remove an allowed report directly beneath the temporary directory."""
+    temporary_root = (temporary_directory or Path(tempfile.gettempdir())).resolve()
+    candidate = path.resolve()
+    if candidate.parent != temporary_root:
+        raise ValueError(f"report must be directly beneath {temporary_root}")
+    if _REPORT_NAME.fullmatch(candidate.name) is None:
+        raise ValueError(f"not an allowed Newton report filename: {candidate.name}")
+    candidate.unlink(missing_ok=True)
```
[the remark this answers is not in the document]
eric-heiden: Addressed in c3310c98a: cleanup now rejects symlink paths before resolution and unlinks only the supplied path after validation. The regression test confirms the symlink target remains intact.

.claude/skills/release-audit/SKILL.md
```diff
@@ -163,13 +163,38 @@ Generates a markdown audit of a Newton release for keep/defer decisions (or, in
...
+     uvx --from towncrier==25.8.0 towncrier build --draft \
+       --version X.Y.Z --date YYYY-MM-DD
+     ```
+     Use the draft output as the canonical fragment preview; it does not modify
+     `CHANGELOG.md` or delete fragments.
+   - **Retrospective**: read `CHANGELOG.md` at HEAD (the current working tree —
+     released sections are append-only). Locate the `## [<target-version>]`
+     header and collect content up to the next dated `##` heading. If the
+     header cannot be found at HEAD, fall back to
+     `git show v<target>:CHANGELOG.md`.
```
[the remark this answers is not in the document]
eric-heiden: Addressed in c3310c98a: retrospective mode now always reads `CHANGELOG.md` from the authoritative target tag with `git show v<target>:CHANGELOG.md`; the working-tree preference and fallback were removed.

.claude/skills/release-audit/SKILL.md
```diff
@@ -179,11 +204,22 @@ Generates a markdown audit of a Newton release for keep/defer decisions (or, in
...
+       <base-ref>..<head-ref> -- changelog.d/<fragment-filename>
+     ```
+     Numeric identifiers link to issues, not necessarily to the implementing
+     pull request, so the fragment-addition commit is the authoritative join.
+   - For legacy entries and any fragment still lacking a backing commit, search
+     exact text across both storage forms:
      ```bash
-     git log --reverse -S'<distinctive substring from the entry>' --format='%H|%s|%cs' -- CHANGELOG.md
+     git log --reverse -S'<distinctive substring from the entry>' \
+       --format='%H|%s|%cs' <base-ref>..<head-ref> -- CHANGELOG.md changelog.d
```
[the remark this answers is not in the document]
eric-heiden: Addressed in c3310c98a: the guidance now keeps the distinctive text in `entry_text` and passes it as the quoted `-S"$entry_text"` argument instead of interpolating changelog text into shell source.

.claude/skills/release-changelog/SKILL.md
```diff
@@ -1,103 +1,91 @@
...
+## Checks
 
 ```bash
-git diff v<latest-release> -- CHANGELOG.md
-git log --oneline <previous-release>..<release-ref>
-rg -n "removed|removal|deprecated|will be removed|in favor of|use .* instead|renam|replac|default|breaking" CHANGELOG.md
-git diff -- CHANGELOG.md
+uv run --no-project python scripts/changelog_policy.py validate
+git diff v<latest-release> -- CHANGELOG.md changelog.d
+git diff --name-status -- CHANGELOG.md changelog.d
```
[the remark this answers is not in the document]
eric-heiden: Addressed in c3310c98a: the name-status check now uses `git diff --cached --name-status` so it reports the staged changelog update and fragment deletions.

docs/guide/release.rst
```diff
@@ -236,10 +236,13 @@ otherwise.
    * - ☐
      - Go/no-go approval obtained from maintainers.
    * - ☐
-     - Finalize ``CHANGELOG.md`` on ``release-X.Y`` using the previous release
-       tag and latest audit as references.  Date the section with the GA date
-       and merge it before preparing the final version.  The
-       ``release-changelog`` skill can assist.
+     - On ``release-X.Y``, run ``towncrier build --draft`` with the GA version
+       and date, audit and approve the preview, then rerun with ``--yes`` as
+       documented in ``changelog.d/README.md``.  Delete ``.skip`` fragments and
```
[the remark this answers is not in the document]
eric-heiden: Addressed in c3310c98a: the GA checklist now starts from the pinned `uvx --from towncrier==25.8.0 towncrier build --draft` invocation and reruns that command with `--yes` after approval.

pyproject.toml
```diff
@@ -296,6 +296,41 @@ reportUnknownParameterType = false
...
+[[tool.towncrier.type]]
+directory = "removed"
+name = "Removed"
+showcontent = true
+
+[[tool.towncrier.type]]
+directory = "fixed"
+name = "Fixed"
+showcontent = true
+
```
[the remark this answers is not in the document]
eric-heiden: Addressed in c3310c98a: removed the unsupported/no-op `header_prefix` setting from the production configuration while leaving `title_format` unchanged. Towncrier 25.8.0 draft rendering and the workflow tests still produce the intended heading structure.

scripts/tests/test_changelog_policy.py
```diff
@@ -0,0 +1,336 @@
...
+[[tool.towncrier.type]]
+directory = "removed"
+name = "Removed"
+showcontent = true
+
+[[tool.towncrier.type]]
+directory = "fixed"
+name = "Fixed"
+showcontent = true
+"""
```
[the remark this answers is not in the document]
eric-heiden: Addressed in c3310c98a: removed the duplicate `header_prefix` setting from the isolated workflow-test configuration. All eight policy and Towncrier workflow tests pass.

.github/workflows/pr.yml
```diff
@@ -8,12 +8,56 @@ on:
...
+      - name: Checkout repository
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # v6.0.1
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+      - name: Install uv
+        uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78  # v7.6.0
+        with:
+          version: "0.11.26"
+      - name: Validate pending fragments
```
jcarius-nv: The repository-tool tests added in `scripts/tests/test_changelog_policy.py` are not run by this job, so the adapter and release behavior can regress while this check remains green. Please execute them in CI with the pinned Towncrier version, for example `uv run --no-project --with towncrier==25.8.0 python -m unittest scripts.tests.test_changelog_policy -v`. Please also add a direct test that a normal PR editing `CHANGELOG.md` is rejected.

.github/workflows/pr.yml
```diff
@@ -8,12 +8,56 @@ on:
...
+        if: github.event_name == 'merge_group'
```
jcarius-nv: On `merge_group` this only validates the directory and skips the diff-aware PR policy. That leaves a transition hole where an older queued PR can still edit `CHANGELOG.md` directly or otherwise violate the normal/release-management diff rules. Please run an equivalent diff-aware policy check for the merge-group range so the policy is enforced at the final merge point as well.

pyproject.toml
```diff
@@ -296,6 +296,40 @@ reportUnknownParameterType = false
 reportOptionalOperand = false
 reportUnusedCallResult = false
 
+[tool.towncrier]
+directory = "changelog.d"
+filename = "CHANGELOG.md"
+start_string = "<!-- towncrier release notes start -->\n"
+title_format = "## [{version}] - {project_date}"
+issue_format = "[#{issue}](https://github.com/newton-physics/newton/issues/{issue})"
+issue_pattern = "\\d+"
```
jcarius-nv, line 305: Can we narrow the custom adapter by letting Towncrier own its native fragment grammar? Concretely, configure `ignore = ["*.skip"]`, run Towncrier over all of `changelog.d/`, and remove the adapter's duplicated typed-category/counter/filename parsing. The adapter would still be valuable for Newton-specific rules that Towncrier cannot express: `.skip` content, imperative style/punctuation, direct `CHANGELOG.md` edits, add-only fragments, one logical ID on `main`, backport relaxation, bootstrap, and `release-management` scope. This keeps `.skip` from rendering while making Towncrier reject every other invalid filename.

jcarius-nv (commented):
The Towncrier direction fits Newton and the implementation has a solid base. I left the concrete workflow and transition gaps I think we should close before approval, plus several smaller consistency improvements.

docs/guide/release.rst
```diff
@@ -236,10 +236,14 @@ otherwise.
    * - ☐
      - Go/no-go approval obtained from maintainers.
    * - ☐
-     - Finalize ``CHANGELOG.md`` on ``release-X.Y`` using the previous release
-       tag and latest audit as references.  Date the section with the GA date
-       and merge it before preparing the final version.  The
-       ``release-changelog`` skill can assist.
+     - On ``release-X.Y``, run
```
jcarius-nv: Please make the release changelog workflow explicit here and in the `release-changelog` skill. Once the initial release scope has been audited on `release-X.Y`, run Towncrier to assemble the dated Markdown and consume the current fragments, then invoke `release-changelog` and perform the normal completeness, grouping, deduplication, wording, category, and migration-guidance cleanup. Do this early enough during RC stabilization for maintainers to review the assembled changelog.

Treat the assembled release section as a rolling document. If further cherry-picks land before tagging, render their fragments with `--draft`, fold the entries into the existing section, delete the consumed fragments, and rerun the relevant changelog cleanup and release-audit checks. Final GA preparation should verify the completed section and confirm that no release fragments remain, rather than defer the full cleanup until the last minute.

Post-release synchronization should carry all changelog-management commits from `release-X.Y` to `main`—the initial Towncrier build, editorial cleanup, and later cherry-pick additions—while preserving fragments that exist only on `main`.

scripts/changelog_policy.py
```diff
@@ -0,0 +1,286 @@
...
+
+    allowed_paths = {CHANGELOG_PATH, *(change.path for change in fragment_changes)}
+    unexpected_paths = sorted(str(change.path) for change in changes if change.path not in allowed_paths)
+    if unexpected_paths:
+        errors.append(
+            "release-management pull requests must be changelog-only; unexpected paths: " + ", ".join(unexpected_paths)
+        )
+    if len(changelog_changes) != 1 or changelog_changes[0].status != "M":
+        errors.append("release-management pull requests must modify CHANGELOG.md exactly once")
+    if not fragment_changes:
```
jcarius-nv: This requires at least one fragment deletion, but a partial release build can still pass while other release-branch fragments remain pending. For a `release-X.Y` build PR, please require complete consumption of all release fragments (including `.skip` files); for the synchronization PR targeting `main`, continue preserving fragments added only after the branch cut. That likely requires passing the target branch into this validator and testing both build and synchronization cases.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -7,7 +7,8 @@
 
 - [ ] New or existing tests cover these changes
 - [ ] The documentation is up to date with these changes
-- [ ] `CHANGELOG.md` has been updated (if user-facing change)
+- [ ] One logical fragment set or a one-line `.skip` reason has been added by following the
```
jcarius-nv: A valid `release-management` PR cannot satisfy this checkbox because it consumes/deletes fragments rather than adding one. Please state the exception here so the template matches the policy enforced by CI.

.pre-commit-config.yaml
```diff
@@ -50,8 +50,21 @@ repos:
...
+        args: [--version, "0.0.0", --date, "1970-01-01"]
+        files: ^(CHANGELOG\.md|changelog\.d/.*\.md|pyproject\.toml)$
   - repo: local
     hooks:
+      - id: changelog-policy
+        name: check changelog fragment policy
+        entry: python scripts/changelog_policy.py validate
+        language: python
+        pass_filenames: false
+        files: ^(changelog\.d/|scripts/changelog_policy\.py|pyproject\.toml)
```
jcarius-nv: The local policy hook does not match `CHANGELOG.md`, and its directory-only `validate` command cannot reject a direct changelog edit. The Towncrier draft hook also does not enforce that Newton rule. Please include `CHANGELOG.md` and run a staged/diff-aware policy check (or an equivalent local check) so contributors receive this failure before CI.

.coderabbit.yml
```diff
@@ -17,14 +17,15 @@ reviews:
...
       instructions: |
-        Check that any user-facing change includes a corresponding entry in
-        CHANGELOG.md under the [Unreleased] section. Entries should use
-        imperative present tense ("Add X", not "Added X") and be placed under
-        the correct category (Added, Changed, Deprecated, Removed, or Fixed).
-        Entries may appear at any position within their category — do not flag
-        placement order. For Deprecated, Changed, and Removed entries, verify
-        the entry includes migration guidance telling users what replaces the
-        old behavior.
+        Follow changelog.d/README.md. Check that a normal pull request adds one
```
jcarius-nv: Please describe the `release-management` exception explicitly. As written, CodeRabbit is told that every PR must add a fragment and must not edit `CHANGELOG.md`, so it can incorrectly flag the valid build/synchronization PRs that modify the changelog and delete consumed fragments.

.claude/skills/release-audit/SKILL.md
```diff
@@ -1,9 +1,9 @@
...
+allowed-tools: Bash(git log *) Bash(git show *) Bash(git grep *) Bash(git tag *) Bash(git rev-parse *) Bash(git diff *) Bash(git ls-tree *) Bash(uv run --no-project python scripts/changelog_policy.py validate) Bash(uvx --from towncrier==25.8.0 towncrier build --draft *) Bash(python3 ${CLAUDE_SKILL_DIR}/scripts/list_commits.py *) Bash(python3 ${CLAUDE_SKILL_DIR}/scripts/license_audit.py *) Bash(python3 ${CLAUDE_SKILL_DIR}/scripts/cleanup_report.py *) Bash(gh --version) Bash(gh auth status) Bash(gh gist create *) Bash(gh gist list *) Bash(gh gist view *) Bash(gh gist edit *) Bash(gh issue view *) Bash(gh issue list *) Read Write Grep Glob
```
jcarius-nv: The new cleanup helper is invoked with bare `python3`, while Newton's documented workflow requires `uv` for commands. Please make the allowlist and the corresponding instructions use `uv run --no-project python ${CLAUDE_SKILL_DIR}/scripts/cleanup_report.py ...` consistently.

.github/workflows/pr.yml
shi-eric: Personally, I'm okay with not having these `test_changelog_policy.py` run in CI/CD as that would yet another job to the pipeline. I think we can just run these offline.

.github/CODEOWNERS
```diff
@@ -8,6 +8,7 @@
 
 # Common PR companion files do not require maintainer approval
 /CHANGELOG.md
+/changelog.d/
```
shi-eric: Could you ensure `changelog.d/README.md` remains under maintainer ownership in `CODEOWNERS`? The ownerless `/changelog.d/` rule applies recursively and currently exempts the guide along with the fragments. Please either narrow the exemption to fragment files or add a later rule assigning `changelog.d/README.md` to `@newton-physics/maintainers`.

.github/workflows/pr.yml
```diff
@@ -8,12 +8,56 @@ on:
     branches:
       - main
       - "release-*"
+    types: [opened, synchronize, reopened, labeled, unlabeled]
```
shi-eric: Could you avoid adding `labeled` and `unlabeled` to the existing PR workflow? With the current workflow-level concurrency, either event can cancel and restart the lockfile, documentation, and platform test jobs. Since changelog validation depends on the `release-management` label, please move the changelog job into a dedicated workflow with these label triggers and leave the existing PR workflow triggers unchanged.

eric-heiden:
Based on @shi-eric's feedback, I've renamed the `changelog.d` folder to `changelog`.

.github/workflows/pr.yml
eric-heiden: Done. The policy tests stay offline as suggested, and I added a direct test that rejects normal PRs which edit CHANGELOG.md.

.github/workflows/pr.yml
eric-heiden: Done. The merge-group job now runs a diff-aware policy check over the merge-group range.

pyproject.toml
eric-heiden: Done. Towncrier now ignores only *.skip and owns the native fragment filename grammar. The adapter now checks only Newton-specific rules.

docs/guide/release.rst
eric-heiden: Done. The release guide and release-changelog skill now assemble the changelog early during RC work, fold in later fragments, rerun audits, and sync all changelog-management commits to main.

scripts/changelog_policy.py
eric-heiden: Done. Release-management PRs to a release branch must consume every fragment. Sync PRs to main may keep main-only fragments, and both cases are tested.

.github/PULL_REQUEST_TEMPLATE.md
eric-heiden: Done. The PR template now calls out the release-management exception.

.pre-commit-config.yaml
eric-heiden: Done. The pre-commit hook now includes CHANGELOG.md and runs the staged, diff-aware policy check.

.coderabbit.yml
eric-heiden: Done. CodeRabbit now knows about the release-management exception and its release-branch and main-branch rules.

.claude/skills/release-audit/SKILL.md
eric-heiden: Done. The cleanup helper and the other release-audit helpers now use uv run --no-project python consistently.

.github/CODEOWNERS
eric-heiden: Done. changelog/README.md has a later CODEOWNERS rule assigning it to @newton-physics/maintainers.

.github/workflows/pr.yml
eric-heiden: Done. Changelog validation is now in its own workflow with label triggers, and the existing PR workflow triggers are unchanged.

.github/workflows/pr.yml
jcarius-nv: Thanks, the diff-aware merge-group check is the right direction. Overall this looks good from my side now, with one bootstrap case left: because this PR itself modifies `CHANGELOG.md`, `validate_merge_group_changes()` routes the entire diff through `_validate_release_management_changes()` before the bootstrap exception in `validate_pr_changes()` can apply.

I reproduced this at `b490d24` against base `de2d06f`: `check-merge-group` exits 1 because the PR contains non-changelog tooling/docs changes and adds `changelog/+towncrier-workflow-7d9e3a1c.skip`. That means this bootstrap PR would fail its own required merge-group check once queued. Please preserve the one-time bootstrap exception for the merge-group path as well, ideally with a regression test that passes this PR-shaped diff through `validate_merge_group_changes()`.

eric-heiden: Done. The merge-group check now keeps the one-time bootstrap exception. I also added a regression test for this PR-shaped diff, and the check against de2d06f now passes.

jcarius-nv: Thanks, this addresses the concrete bootstrap failure. I confirmed that the exact merge-group check now passes, and the new regression test covers this PR-shaped diff.

One minor, non-blocking follow-up: `_is_policy_bootstrap()` is evaluated over the aggregate merge-group diff. Newton’s queue can batch multiple PRs, so if a stale PR that directly edits `CHANGELOG.md` is grouped with this one, the bootstrap exception can mask that edit; I reproduced such a combined change set returning success. If you think it is worth hardening, the bootstrap path could verify that the `CHANGELOG.md` patch contains only the expected Towncrier marker insertion, or this bootstrap could be made to queue alone.

Please feel free to address this now, later, or not at all—it does not block my approval.

jcarius-nv (dismissed):
Looks good from my side. The concrete merge-group bootstrap failure is fixed; I left one optional, non-blocking batching follow-up in the existing thread.

.github/workflows/pr.yml
shi-eric: A potential issue with regular `release-management` synchronization PRs was also flagged by Codex. If one were grouped with a normal PR, the changelog update and fragment deletions from the synchronization PR, as well as the code changes and added fragment from the normal PR, would all be included in the same aggregate diff.

Since any `CHANGELOG.md` change sends the entire diff through `_validate_release_management_changes()`, the changes from the normal PR could be rejected even though both PRs passed their individual checks. Could you investigate whether this batching can occur and, if so, whether the group would be incorrectly rejected?

scripts/changelog_policy.py
shi-eric, line 1: Could you expand this module docstring with a brief high-level overview of the policy? The module handles several distinct contexts: normal PRs, release-management PRs, staged changes, release branches versus main, and aggregate merge-group diffs. A few sentences explaining these modes and what is delegated to Towncrier would make the code easier to understand without duplicating the detailed instructions in `changelog/README.md`.

.github/workflows/changelog-preview.yml
shi-eric, line 1: Consider a more descriptive workflow name since "Changelog" is not very specific. You might also add some comments near the top that explain what the workflow does to help readers understand the purpose

.pre-commit-config.yaml
```diff
@@ -50,8 +50,21 @@ repos:
...
+  - repo: https://github.com/twisted/towncrier
+    rev: 25.8.0
+    hooks:
+      - id: towncrier-check
+        entry: towncrier build --draft
+        args: [--version, "0.0.0", --date, "1970-01-01"]
+        files: ^(CHANGELOG\.md|changelog/|pyproject\.toml)
   - repo: local
     hooks:
+      - id: changelog-policy
```
shi-eric: I had Codex exercise this hook with several multi-commit workflows. A code-only follow-up commit does not require another fragment, as expected, but the staged validator rejects other valid workflows because it treats the current commit diff as if it were the complete PR diff.

For example, if a fragment is added in one commit and its wording is updated in a later commit, the second commit sees the fragment as modified and rejects it with `normal pull requests may only add fragments`. The aggregate PR diff still sees that fragment as newly added and passes CI. A release build followed by a separate editorial cleanup commit that only updates `CHANGELOG.md` is also rejected, even though that sequence is described in the release guide.

The hook also fails when pre-commit runs under Python 3.10, which Newton supports, because `changelog_policy.py` imports `tomllib`.

I tend to prefer starting with simpler changes rather than trying to automate every part of the process up front, especially when the automation can introduce edge cases of its own. It may be worth reducing this to the bare minimum for now. For example, we could omit the pre-commit hooks, have CI verify that Towncrier can render the current fragment set, and avoid encoding policies such as requiring a `.skip` file for every PR without a user-facing change. Stricter policies can always be added later if experience shows they are needed. Just a suggestion to keep the initial change focused.

eric-heiden: Thanks, agreed with your suggestion! I will simplify the automation here to remove these strict policies.

.github/workflows/pr.yml
eric-heiden: I removed the custom diff policy from both pull request and merge-group checks. The workflow now only runs a Towncrier draft render, so batched pull requests no longer go through the aggregate policy that caused these cases.

scripts/changelog_policy.py
eric-heiden: I removed `scripts/changelog_policy.py` as part of simplifying the workflow, so this module and its docstring are no longer present.

.github/workflows/changelog-preview.yml
eric-heiden: Done. I renamed the file to `changelog-preview.yml`, changed the workflow name to `Changelog fragment preview`, and added a short comment explaining its purpose.

.pre-commit-config.yaml
eric-heiden: Done. I removed both changelog pre-commit hooks and the custom policy script. CI now only runs a pinned Towncrier draft render, and `.skip` files are optional. I also verified the draft render with Python 3.10.

jcarius-nv (approved):
Looks good from my side. Let's give this a try and iterate further once we've gained some experience working with the system.

shi-eric:
Thanks for considering the simplifications @eric-heiden

shi-eric (approved):

PR #3615, opened 2026-07-23
Extended contact visualizations

Guirec-Maloisel:
## Description

This refines the (rigid) contact visualization options in the newton viewer, allowing to visualize the contact force, as well as the contact mode (open / sliding / sticking, using a color-coded disk in the tangential plane), on top of the contact normal (as per the current viewer).

The disk color (contact mode) and force visualization both require access to the contact force (e.g. through calling `builder.request_contact_attributes("force")`), but will default to black (open contact) and no force vector if not available.

## Checklist

## Test plan

This has been incorporated for a while in my local branch for validating contact modeling and comparing solvers, so it has been indirectly tested there. Not sure what the policy is for unit testing the viewer outside of this (apart from the reviewers trying it out on the newton examples). 

## Summary by CodeRabbit

* **New Features**
  * Added separate visualization controls for contact normals, contact modes, and contact forces.
  * Contact modes are shown with color-coded disks for open, sticking, and sliding behavior.
  * Contact forces appear as directional arrows when force data is available.
  * Added controls for contact scale, force scale, and arrow thickness.
  * Visualization scales adapt automatically to the model while preserving manual adjustments.

* **Documentation**
  * Updated the changelog with details about the refined contact visualizations.

### DIFF
--- a/newton/_src/viewer/kernels.py
+++ b/newton/_src/viewer/kernels.py
@@ -324,6 +324,213 @@ def compute_contact_lines(
     line_end[tid] = contact_center + line_vector
 
 
+@wp.func
+def _quat_from_normal_z(normal: wp.vec3) -> wp.quat:
+    """Build a rotation quaternion whose local +Z axis aligns with ``normal``.
+
+    The tangent axes are arbitrary. Implemented with :func:`orthonormal_basis`
+    for numerical stability near the poles.
+    """
+    n = wp.normalize(normal)
+    t1, t2 = orthonormal_basis(n)
+    R = wp.matrix_from_cols(t1, t2, n)
+    return wp.quat_from_matrix(R)
+
+
+@wp.kernel
+def compute_contact_disk_transforms(
+    body_q: wp.array[wp.transform],
+    body_qd: wp.array[wp.spatial_vector],
+    body_com: wp.array[wp.vec3],
+    shape_body: wp.array[int],
+    shape_world: wp.array[int],
+    world_offsets: wp.array[wp.vec3],
+    visible_worlds_mask: wp.array[int],
+    contact_count: wp.array[int],
+    contact_shape0: wp.array[int],
+    contact_shape1: wp.array[int],
+    contact_point0: wp.array[wp.vec3],
+    contact_point1: wp.array[wp.vec3],
+    contact_offset0: wp.array[wp.vec3],
+    contact_normal: wp.array[wp.vec3],
+    contact_force: wp.array[wp.spatial_vector],
+    disk_radius: float,
+    disk_thickness: float,
+    eps_force: float,
+    eps_velocity: float,
+    color_open: wp.vec3,
+    color_stick: wp.vec3,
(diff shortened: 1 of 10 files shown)


eric-heiden:
Thanks, looks like a useful feature! There is some z-fighting in the black discs for the open contact mode. Perhaps you want to just scale them slightly larger to prevent this flickering?


https://github.com/user-attachments/assets/0f31ff04-dc71-4746-af4a-a41a9565f00a

newton/_src/viewer/kernels.py
```diff
@@ -324,6 +324,213 @@ def compute_contact_lines(
...
+    X_wb_a = wp.transform_identity()
+    if body_a >= 0:
+        X_wb_a = body_q[body_a]
+
+    world_pos0 = wp.transform_point(X_wb_a, contact_point0[tid] + contact_offset0[tid])
+    contact_center = world_pos0
+    if world_a >= 0 or world_b >= 0:
+        contact_center += world_offsets[world_a if world_a >= 0 else world_b]
+
+    f_lin = -wp.spatial_top(contact_force[tid])  # Flip sign so positive force is along normal
```
eric-heiden, line 532: Do you think this is right? I'm not sure if it could be confusing to users when the shown contact normals are not in the true direction.

Guirec-Maloisel: According to contacts.py, the convention for the normal is "A-to-B"; and the convention for the force is "exerted on body0 by body1". Visually these two conventions didn't look consistent to me without the sign flip (forces going the ground).

One way to think about this is that the normal convention is seen "from the point of view of B" as a moving object w.r.t. a fixed/reference object A (since it then matches the outwards normal to A). But the force convention is rather from the point of view of A, since the positive force is the one applied to A. Flipping the convention means taking here again the point of view of B moving w.r.t. A, and positive forces applied to B being along the outward normal of the obstacle A.

Testing with Kamino examples, this sign seems to give intuitive results, but I'm not sure however what consistency we can expect in the contacts (e.g. would a static plane always be the A body?). As a practical fix in the viewer (if the case where e.g. B is the ground does arise) we could add a toggle in the viewer UI to flip contact normals and forces between A->B and B->A.

Guirec-Maloisel:
@eric-heiden thanks for reviewing! I've pushed a fix that addresses the z-fighting as suggested (sorry for not looking again into this earlier, I was on vacation).

https://github.com/user-attachments/assets/5299f25f-adbf-4b74-a085-713cd5a2591f

newton/_src/viewer/kernels.py
eric-heiden: Thanks for writing this up, I agree the sign flip should stay.

eric-heiden (dismissed):
Thanks for fixing the rendering! @Guirec-Maloisel can you fix the merge conflict? Then we can merge.

eric-heiden (approved):

chschuma-disney (approved):

PR #3616, opened 2026-07-23
Miscellaneous small fixes in Kamino

Guirec-Maloisel:
## Description

This groups a bunch of small fixes that would not warrant an individual PR by themselves. It is recommended to review commit by commit to see the independent changes better.

The two "bigger" changes among these are removing the density field from materials in Kamino (after an offline discussion with @vastsoun, this is not used currently nor is there a potential usage for this); and ensuring that a bunch of kernels/functions support the world_mask being passed as None (to signal no masking is needed), allowing a few tiny optimizations in some cases.

## Checklist

## Test plan

Kamino internal unit tests are all passing after these changes.

## Summary by CodeRabbit

* **New Features**
  * World masks are now optional for joint-state extraction and reset operations; when omitted, operations apply to all worlds.
  * Material definitions and imported materials now use restitution and friction properties without density.
* **Bug Fixes**
  * Improved reset success reporting when no mask or forward-kinematics result is provided.
  * Corrected floating-base and geometry transform handling.
* **Documentation**
  * Clarified joint-limit behavior and warning conditions.
* **Tests**
  * Updated coverage for material properties, model conversion, and caching behavior.

### DIFF
--- a/newton/_src/solvers/kamino/_src/core/conversions.py
+++ b/newton/_src/solvers/kamino/_src/core/conversions.py
@@ -1378,7 +1378,7 @@ def convert_joints(
                 continue
             base_body_idx_np[wid] = body_world_start_np[wid]
 
-    # Only warn for worlds where a skipped root left the world without a base
+    # Warn user if an articulation root couldn't be used as base because it is not a free joint
     if np.any(world_has_non_floating_root & (base_body_idx_np == -1)):
         msg.warning(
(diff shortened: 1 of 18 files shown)


rubengrandia (approved):
LGTM

chschuma-disney (approved):
Solid work, looks good to me :rocket:

PR #3617, opened 2026-07-23
Remove deprecated joint target aliases and deprecate the DOF target layout

adenzler-nvidia:
## Description

Remove the `joint_target_pos` / `joint_target_vel` aliases from `Model`, `Control`, and `ModelBuilder`, deprecated in 1.3.0 in favor of `joint_target_q` / `joint_target_qd`. `Actuator` now always defaults `control_target_pos_attr` / `control_target_vel_attr` to the canonical names, completing the switch its deprecation warning announced (behavior-neutral: under the legacy layout the alias resolved to the same array).

The legacy DOF-shaped `joint_target_q` layout itself had no runtime deprecation signal until now: users who migrated to the new names but rely on `use_coord_layout_targets = False` were never individually warned that the layout will change. `ModelBuilder.finalize()` now emits a `DeprecationWarning` when building a model whose joint coordinate and DOF counts differ (free/ball/distance joints — the only models where the two layouts diverge) under the legacy flag. The warning announces the terminal state, so a future release can make the coordinate layout the only layout and remove the flag in one step. Models without such joints are unaffected; the layouts coincide.

The test suite and examples now run under the coordinate layout. The suite flag is set in the test runner's entry points and worker initializers — deliberately not at `unittest_utils` import time, because `newton.examples` imports `unittest_utils` and an import-time assignment leaks the flag into production example runs (this exact leak initially broke the floating-base policy examples during development). Examples that build free/ball/distance-joint models set the flag in `Example.__init__`, following the existing opt-in pattern; `example_robot_policy` also switches its target buffer to the coordinate layout. Tests exercising the deprecated DOF layout are removed, with coord-layout ball-controller velocity coverage extended to all three axes where they were the only coverage.

Running the suite under the coordinate layout immediately surfaced a real bug: `eval_inverse_dynamics_passive()` fed `eval_rigid_tau` a DOF-sized constant-zero buffer as the position-target input, which the kernel indexes via the coord-shaped `joint_target_q_start` — an out-of-bounds read producing intermittent NaNs in gravity/Coriolis forces for models with free, ball, or distance joints. Fixed with a correctly-sized zero scratch buffer.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests --strict-warnings
uv run --extra dev -m newton.tests --strict-warnings -k test_examples
uv run --extra dev -m newton.tests --strict-warnings -k test_inverse_dynamics
uvx pre-commit run -a
```

Full suite passes with strict warnings (4862 tests; 2 machine-specific hydroelastic tolerance failures also present on clean `main` on the same box). All 186 example tests pass with zero deprecation warnings. `test_inverse_dynamics` verified stable across repeated runs (it failed intermittently under the coordinate layout before the OOB fix).

## Bug fix

**Steps to reproduce (inverse-dynamics OOB read):**

1. Set `newton.use_coord_layout_targets = True`.
2. Build a model containing a free or ball joint and call `newton.eval_inverse_dynamics_passive()`.
3. Repeat the evaluation: gravity/Coriolis outputs intermittently contain NaN, depending on the garbage past the end of the DOF-sized scratch buffer.

## Summary by CodeRabbit

* **New Features**
  * Standardized joint targets on `joint_target_q` / `joint_target_qd`.
  * Updated examples and test execution to use coordinate-layout targets.
* **Removed**
  * Removed deprecated `joint_target_pos` / `joint_target_vel` aliases.
* **Bug Fixes**
  * Fixed intermittent NaNs in passive inverse dynamics for free, ball, and distance joints.
* **Documentation**
  * Updated migration and solver guidance for coordinate-layout targets and deprecation warnings.
* **Tests**
  * Updated controller, actuator, and USD mapping coverage for canonical joint targets.


newton/tests/test_joint_controllers.py
```diff
@@ -85,103 +84,6 @@ def test_revolute_controller(
         test.assertAlmostEqual(joint_qd[0], expected_vel, delta=1e-2)
 
 
-def test_ball_controller(
```
jvonmuralt, line 88: Could we rename test_ball_controller_coord_layout  to test_ball_controller and port the missing cases from the deleted version, specifically POSITION_VELOCITY mode and the per-axis (X, Y, Z) velocity target tests updated to use the coord layout?

jvonmuralt (dismissed):

adenzler-nvidia:
moving @Guirec-Maloisel and @eric-heiden on the reviewer list because apparently this needs disney and core maintainer approval. Should be uncontroversial, just a quick approve needed.

Guirec-Maloisel (dismissed):
Looks good from the Disney side!

eric-heiden:
Thanks for this clean-up! I think we should consider raising an AttributeError now for these removed Control attributes to make sure users are aware of this API change.

newton/_src/sim/control.py
```diff
@@ -62,12 +39,12 @@ def __init__(self):
         self.joint_target_q: wp.array | None = None
         """Joint position targets [m or rad]. Shape is ``(joint_coord_count,)``
         when :data:`newton.use_coord_layout_targets` is ``True``, otherwise
-        ``(joint_dof_count,)`` for legacy compat with :attr:`joint_target_pos`.
+        ``(joint_dof_count,)`` (legacy layout).
         """
 
         self.joint_target_qd: wp.array | None = None
         """Joint velocity targets [m/s or rad/s], shape ``(joint_dof_count,)``.
```
eric-heiden, line 69: 🟠 Removing the `joint_target_pos` and `joint_target_vel` descriptors allows legacy assignments to create unused instance attributes. For example, `control.joint_target_pos = targets` succeeds but leaves `joint_target_q` unchanged, so the simulation silently ignores the requested targets.

Please retain tombstone descriptors that raise `AttributeError` with guidance to use `joint_target_q` or `joint_target_qd`. The same protection is needed for the removed aliases on `Control`, `Model`, and `ModelBuilder`.

newton/_src/actuators/actuator.py
```diff
@@ -88,8 +88,8 @@ def __init__(
         effort_indices: wp.array[wp.uint32] | None = None,
         state_pos_attr: str = "joint_q",
         state_vel_attr: str = "joint_qd",
-        control_target_pos_attr: str | None = None,
-        control_target_vel_attr: str | None = None,
+        control_target_pos_attr: str = "joint_target_q",
+        control_target_vel_attr: str = "joint_target_qd",
```
eric-heiden: 🟠 Existing callers can explicitly pass `None` for these parameters. Construction still succeeds, but `step()` later passes that value to `getattr()`, which requires a string and raises at runtime.

Please keep these parameters typed as `str | None` and normalize explicit `None` values to `"joint_target_q"` and `"joint_target_qd"`. This preserves the previous public behavior while switching the implicit defaults to the canonical names.

chschuma-disney (dismissed):

newton/_src/sim/builder.py
```diff
@@ -12078,6 +11961,17 @@ def _to_wp_array(data, dtype, requires_grad):
...
+                if self.joint_coord_count != self.joint_dof_count:
+                    warnings.warn(
+                        "The legacy DOF-shaped joint_target_q layout is deprecated for models "
+                        "whose joint coordinate and DOF counts differ (free/ball/distance "
+                        "joints). In a future release joint_target_q will always use the "
+                        "coordinate layout (matching joint_q) and newton.use_coord_layout_targets "
+                        "will be removed. Set newton.use_coord_layout_targets = True before "
+                        "building models and index targets via Model.joint_target_q_start.",
+                        DeprecationWarning,
+                        stacklevel=3,
```
jcarius-nv: The new warning uses `stacklevel=3`, but `finalize()` is not wrapped, so this skips the direct caller. I reproduced a direct `builder.finalize()` warning being attributed to `sys:1` instead of the model-construction call. Please use `stacklevel=2` (or the existing external-frame helper) and add a focused test covering affected/unaffected models plus the warning filename/line.

docs/migration.rst
```diff
@@ -146,14 +146,14 @@ corresponds to :attr:`newton.JointTargetMode.EFFORT` together with
...
 
-Historically ``Control.joint_target_pos`` was shaped ``(joint_dof_count,)`` — the same layout as
-:attr:`~newton.State.joint_qd` — even though position targets semantically match
-:attr:`~newton.State.joint_q`. The two layouts diverge whenever an articulation contains a free
-joint (7 coords vs. 6 DOFs) or ball joint (4 coords vs. 3 DOFs); every actuated DOF downstream
-then ends up indexed with the wrong stride.
+In ``warp.sim`` the target array was DOF-shaped like ``joint_qd``. Newton is moving position
+targets to the coordinate layout of :attr:`~newton.State.joint_q` instead, since that is what a
+position semantically is: the two layouts diverge whenever an articulation contains a free joint
+(7 coords vs. 6 DOFs) or ball joint (4 coords vs. 3 DOFs), and under the DOF layout every actuated
```
jcarius-nv: This reads as an exhaustive description of when the layouts diverge, but it omits `DISTANCE`, which also has 7 coordinates and 6 DOFs. Please include distance joints here so this agrees with the runtime warning and the later text on this page.

CHANGELOG.md
```diff
@@ -90,6 +90,11 @@
 - Deprecate `SensorTiledCamera(..., config=...)` in favor of `SensorTiledCamera(..., default_render_config=...)`; migrate constructor calls that pass a render config to the new keyword.
 - Deprecate `SensorTiledCamera.render_config` in favor of `SensorTiledCamera.default_render_config`; migrate `sensor.render_config.enable_shadows = True` to `sensor.default_render_config.enable_shadows = True`.
 - Deprecate `SensorTiledCamera.utils.compute_pinhole_camera_rays()` in favor of `SensorTiledCamera.utils.compute_camera_rays_pinhole()`.
+- Deprecate the legacy DOF-shaped `joint_target_q` layout (`newton.use_coord_layout_targets = False`) for models whose joint coordinate and DOF counts differ (free/ball/distance joints); `ModelBuilder.finalize()` now emits a `DeprecationWarning` for such models. Set `newton.use_coord_layout_targets = True` before building models and index targets via `Model.joint_target_q_start`. A future release will make the coordinate layout the only layout and remove the flag.
+
+### Removed
```
jcarius-nv, line 95: Please fold this entry into the existing `### Removed` section immediately below instead of creating a second consecutive category heading.

newton/tests/test_joint_controllers.py
```diff
@@ -89,170 +88,79 @@ def test_ball_controller(
...
-    device,
-    solver_fn,
-    target_axis_angle,
-    expected_quat,
-    target_ke,
-    target_kd,
-):
-    """Ball-joint position target under the coord layout: the user writes a
-    target quaternion and the MuJoCo solver must convert it to the matching
+    """Ball-joint position and velocity targets: the user writes a target
```
jcarius-nv: The changed test docstrings need concise imperative summaries per `AGENTS.md`. Please update this one and the changed summaries at lines 174 and 259 (for example, “Verify …”).

newton/tests/test_joint_controllers.py
```diff
@@ -89,170 +88,79 @@ def test_ball_controller(
...
+    for i in range(3):
+        builder.joint_target_ke[qd_start + i] = target_ke
+        builder.joint_target_kd[qd_start + i] = target_kd
+
+    model = builder.finalize(device=device)
+    solver = solver_fn(model)
+    state_0, state_1 = model.state(), model.state()
+    newton.eval_fk(model, model.joint_q, model.joint_qd, state_0)
+
+    # Convert axis-angle target to a quaternion and write to joint_target_q.
```
jcarius-nv: This axis-angle-to-quaternion conversion is duplicated again around line 215. Please extract a small shared helper so both ball-controller paths use the same conversion.

jcarius-nv (commented):
Looks good overall and fits Newton. The earlier compatibility and controller-coverage concerns appear addressed on the current head, and all required checks are passing. I found one migration-warning issue plus a few small documentation/standards cleanups; once the inline comments are addressed, this should be ready for approval.

newton/tests/test_joint_controllers.py
eric-heiden: Done. The test is now named `test_ball_controller`, uses `POSITION_VELOCITY`, and covers X, Y, and Z velocity targets with the coordinate layout.

newton/_src/sim/builder.py
eric-heiden: Done. The warning now uses `stacklevel=2`. I also added a test for affected and unaffected models that checks the warning file and line.

docs/migration.rst
eric-heiden: Done. The migration guide now includes distance joints.

CHANGELOG.md
eric-heiden: Done. I moved this entry into the existing `Removed` section.

newton/tests/test_joint_controllers.py
eric-heiden: Done. The three test summaries now use short “Verify …” wording.

newton/tests/test_joint_controllers.py
eric-heiden: Done. Both ball-controller tests now use one shared `_quat_from_axis_angle()` helper.

chschuma-disney (approved):

jcarius-nv (approved):
Looks good overall. I rechecked the latest commits and existing review history; the remaining issues have been addressed.

PR #3620, opened 2026-07-23
Clarify collision refresh tuning

mzamoramora-nvidia:
## Description

Clarify collision refresh cadence as a contact-debugging tool:

- Start by running collision detection every substep to remove stale contacts
  as a variable when investigating chatter, penetration, or weak grasps.
- Once the behavior is acceptable, collide every N substeps to trade collision
  cost for contact freshness. Keep N below the number of substeps so contacts
  are refreshed within the frame.
- Reorganize the collision guide's **Common Patterns** section so the
  soft-contact example is separate from the collision-frequency subsection and
  the three collision-loop patterns read as one group.
- Explain how Isaac Lab's `NewtonCfg.collision_decimation` maps to collision
  updates during solver substeps.

## Open question

- Should this PR include the Isaac Lab-specific
  `NewtonCfg.collision_decimation` guidance? The mapping helps Isaac Lab users
  apply the tuning advice, but it documents a downstream API that can change
  independently of Newton. If we prefer to keep this guide Newton-only, the
  `docs/lab/isaac-lab.rst` change can be removed while retaining the generic
  collision-cadence guidance.

## Checklist

- [x] `CHANGELOG.md` has been updated (not applicable: documentation-only change)

## Test plan

```bash
uv run --extra docs --extra sim sphinx-build -j auto -W -b html docs /tmp/newton-collision-refresh-html
uv run --extra docs --extra sim sphinx-build -j auto -W -b doctest docs /tmp/newton-collision-refresh-doctest
uv run --extra docs --extra sim sphinx-build -W -b linkcheck -D nbsphinx_execute=never docs /tmp/newton-collision-refresh-linkcheck docs/concepts/collisions.rst docs/concepts/simulation_tuning_solvers.rst docs/lab/isaac-lab.rst
uvx pre-commit run -a
```

The strict HTML build, all 88 Sphinx doctests, the scoped external-link check,
and the full pre-commit suite pass locally.

## Summary by CodeRabbit

* **Documentation**
  * Reorganized collision documentation, including guidance on soft contacts and interpreting contact data.
  * Clarified collision update patterns for every substep, once per frame, and every N substeps.
  * Added troubleshooting guidance for contact chatter, excessive penetration, and weak grasping.
  * Updated solver-tuning recommendations to balance contact accuracy and simulation performance.


docs/lab/isaac-lab.rst
jcarius-nv, line 1: does Isaac lab have their own tuning guides? If yes it would make sense to put this section there, I think it's a bit strange if we maintain it here. The IsaacLab tuning snippet should instead reference the more thorough explanations you added above

docs/lab/isaac-lab.rst
```diff
@@ -6,3 +6,20 @@ Isaac Lab Integration
 
 For details about Isaac Lab support for Newton, see the
 `Isaac Lab documentation <https://isaac-sim.github.io/IsaacLab/main/source/experimental-features/newton-physics-integration/index.html>`_.
+
+Collision Updates During Substeps
+---------------------------------
+
+Isaac Lab controls Newton collision updates with
```
jcarius-nv: Could we qualify this guidance by collision backend? In Isaac Lab, `collision_decimation` only performs mid-substep re-collision when the external Newton `CollisionPipeline` is active. The default `MJWarpSolverCfg` uses `use_mujoco_contacts=True`, while internal Kamino collision and MPM also bypass this path, so setting this to `1` can otherwise be a no-op. Please state that this guidance applies only when Isaac Lab is using the external Newton collision pipeline, rather than internal solver collision detection.

docs/lab/isaac-lab.rst
mzamoramora-nvidia: Done in bfb26696. I removed the Isaac Lab-specific tuning section. The Newton collision-frequency section remains the canonical explanation that downstream integration docs can reference.

docs/lab/isaac-lab.rst
mzamoramora-nvidia: Done in bfb26696. The cadence pattern is now scoped to contacts generated by `CollisionPipeline`. The guide clarifies that it does not control solver-internal collision detection, including MuJoCo contacts and VBD particle self-contact.

jcarius-nv (approved):
Thanks for addressing our review comments. Looks good to me.

PR #3623, opened 2026-07-23
Remove deprecated Model.mujoco.dof_passive_damping

adenzler-nvidia:
## Description

Remove `Model.mujoco.dof_passive_damping`, deprecated in 1.3.0 in favor of `Model.joint_damping`: the runtime alias, the `mujoco:dof_passive_damping` custom attribute, and its finalize-time merge/conflict check.

The custom attribute also acted as an importer staging channel, so this change moves both import paths onto the canonical property:

- **USD:** `SchemaResolverMjc` maps `mjc:damping` to a source-aware `damping_per_rad` key. One importer helper now applies the correct units to revolute, prismatic, spherical, and D6 joints; spherical values are broadcast to all three angular DOFs. Importing MuJoCo-authored USD now requires `schema_resolvers=[..., SchemaResolverMjc()]`.
- **MJCF:** hinge and slide damping remain natively parsed. Native ball joints now pass authored damping directly to `add_joint_ball()`, while omitted damping still uses `ModelBuilder.default_joint_cfg.damping`.

The now-unused `AttributeNamespace` deprecated-alias machinery and the coupled solver's alias-only compaction filter are removed as well.

Review follow-up adds the spherical resolver regression, removes the last legacy staging/finalizer path, centralizes USD damping-source selection, and absorbs the focused #3703 regression coverage from #3704.

The maintenance refresh merges current `main`, moves the user-facing notes to Towncrier fragments, documents explicit resolver registration, and keeps the ball/D6 regressions warning-clean under the coordinate target layout.

## Checklist

- [x] User-facing changes have Towncrier fragments under `changelog/`

## Test plan

```bash
uv run --extra dev python -W error -m unittest \
  newton.tests.test_import_usd.TestImportSampleAssetsBasic.test_joint_stiffness_damping \
  newton.tests.test_import_usd.TestImportSampleAssetsBasic.test_mjc_damping_from_spherical_joint_via_schema_resolver \
  newton.tests.test_import_usd.TestImportSampleAssetsBasic.test_mjc_damping_from_d6_joint_via_schema_resolver \
  newton.tests.test_import_mjcf.TestImportMjcfSolverParams.test_ball_joint_damping_uses_authored_value_or_default

uv run --extra dev python -m unittest \
  newton.tests.test_import_usd.TestImportSampleAssetsBasic \
  newton.tests.test_import_mjcf.TestImportMjcfSolverParams

uv run --extra dev python -m unittest newton.tests.test_schema_resolver newton.tests.test_model
uv run --extra dev python -m unittest newton.tests.test_coupled_solver newton.tests.test_mujoco_solver
uvx --from towncrier==25.8.0 towncrier build --draft --version X.Y.Z --date 2026-08-11
uvx pre-commit run -a
```

Post-merge results: 4 strict-warning focused tests, both targeted importer class runs, and all 155 schema/model tests passed. The Towncrier draft and all pre-commit hooks passed. A combined coupled/MuJoCo solver rerun reached the 5-minute local timeout without reporting a failure; the prior 350-test run on this PR passed (7 expected skips).

## Bug fix

**Steps to reproduce:**

1. Set `builder.default_joint_cfg.damping` to a nonzero value.
2. Import an MJCF ball joint with a different authored `damping`.
3. Before this change, finalization could report a conflict between the deprecated staging attribute and `joint_damping`; after this change, the authored value populates all three ball-joint DOFs directly.

**Minimal reproduction:**

```python
import newton

xml = '''
<mujoco><worldbody><body name="base"><geom type="sphere" size="0.05" mass="1"/>
  <body name="link"><joint type="ball" damping="7"/>
  <geom type="sphere" size="0.04" mass="0.5"/></body></body></worldbody></mujoco>
'''

builder.default_joint_cfg.damping = 99.0
builder.add_mjcf(xml)
model = builder.finalize()
assert (model.joint_damping.numpy()[-3:] == 7.0).all()
```

## Summary by CodeRabbit

- **New Features**
  - Added support for importing per-radian joint damping from USD.
  - MJCF ball joints now preserve authored damping values or use configured defaults when unspecified.
  - Improved damping handling for spherical and multi-DOF joints.

- **Breaking Changes**
  - Removed the deprecated `dof_passive_damping` alias; use `joint_damping` instead.
  - MuJoCo damping imports now require explicit schema resolver registration.

- **Documentation**
  - Updated MuJoCo solver documentation.

- **Tests**
  - Expanded coverage for MJCF damping and USD resolver behavior.


newton/_src/usd/schemas.py
```diff
@@ -494,6 +494,10 @@ class SchemaResolverMjc(SchemaResolver):
         },
         PrimType.JOINT: {
             "armature": SchemaAttribute("mjc:armature", 0.0),
+            # MuJoCo damping is authored in SI units (per radian for angular
+            # DOFs), unlike the USD per-degree convention behind the plain
+            # "damping" key, so it resolves through the _per_rad variant.
+            "damping_per_rad": SchemaAttribute("mjc:damping", None),
```
eric-heiden, line 500: 🟠 Please also consume this mapping for spherical joints in `newton/_src/utils/import_usd.py`. The spherical branch currently calls `add_joint_ball(**joint_params)` without resolving `damping_per_rad`, so resolver-only imports silently discard authored `mjc:damping`. MuJoCo ball joints have three angular DOFs, and their scalar damping applies uniformly to all three. Resolve the source-aware value, broadcast it across the three DOFs, pass it through `add_joint_ball(damping=...)`, and add a resolver-only spherical-joint regression test.

newton/_src/sim/model.py
```diff
@@ -691,47 +691,12 @@ def __init__(self, name: str):
             Args:
                 name: The name of the namespace
             """
-            object.__setattr__(self, "_name", name)
-            object.__setattr__(self, "_deprecated_aliases", {})
-
-        def add_deprecated_alias(self, name: str, getter: Callable[[], Any], message: str) -> None:
```
jcarius-nv, line 697: **Standards (high):** This removes the public `Model.AttributeNamespace.add_deprecated_alias()` method without deprecating the method itself in a prior release. `Model` is exported as `newton.Model`, and the PR's current API-review activity detects this exact interface removal. Newton's development guidelines require a deprecation before removing or renaming public API. Please retain and deprecate this generic method for a release, or obtain and document an explicit API-owner determination that it was never contractual. The separately deprecated `Model.mujoco.dof_passive_damping` alias can still be removed now.

newton/_src/utils/import_usd.py
```diff
@@ -1608,6 +1608,25 @@ def resolve_joint_parent_child(
...
+        regular USD damping mapping follows USD's per-degree convention.
+
+        Returns:
+            The linear and angular damping values in Newton units.
+        """
+        damping = R.get_value(jp_prim, prim_type=PrimType.JOINT, key="damping", default=None, verbose=verbose)
+        damping_per_rad = R.get_value(
+            jp_prim, prim_type=PrimType.JOINT, key="damping_per_rad", default=None, verbose=verbose
+        )
+        if damping_per_rad is not None:
```
jcarius-nv: **Spec (P2):** Resolving `damping` and `damping_per_rad` independently and then always selecting `damping_per_rad` bypasses the documented priority of `schema_resolvers`. On this head, a revolute joint authored with `newton:damping=2` and `mjc:damping=7` imports the MuJoCo value `7` for both `[SchemaResolverNewton(), SchemaResolverMjc()]` and the reverse ordering. Newton-first should select the Newton source and apply its per-degree conversion; MuJoCo-first should select the per-radian MuJoCo value. Please select one authored source according to resolver order before applying source-specific unit conversion, and cover both mixed-schema orderings with a regression test.

jcarius-nv (commented):
Reviewed exact base `1ce6499b` through head `8893d225`, keeping Standards, Spec, and Newton Fit independent.

- **Standards:** one high finding: the generic public `Model.AttributeNamespace.add_deprecated_alias()` method is removed without its own prior deprecation.
- **Spec:** one P2 correctness finding: mixed `newton:damping` / `mjc:damping` authorship ignores the documented `schema_resolvers` priority.
- **Fit:** **Fits if narrowed**. Canonicalizing damping in `Model.joint_damping` and the Newton import layer is the right ownership and a proportionate simplification; narrow the undeprecated generic API removal.

The current head addresses the existing spherical-joint, mixed-D6, legacy-finalizer, duplicate-resolution, and changelog feedback; all three inline threads are resolved. The four focused strict-warning damping regressions pass locally. Hosted unit tests, GPU unit tests, docs, pre-commit, licenses, and API detection pass. The remaining GPU benchmark failure occurred before benchmarks ran because the EC2 runner could not acquire Ubuntu's `dpkg` lock, so it is infrastructure noise even though the check remains failed.

I do not think this is ready for approval until the two inline findings are addressed.

newton/_src/sim/model.py
jcarius-nv: not sure if this function should truly be part of the public API, but my agent flagged it...

eric-heiden: OK I reverted it, but it is deprecated now since we don't need this anymore.

I would not use it for new aliases. Its callable-based design is awkward because it:

- Stores behavior in hidden runtime state.
- Special-cases assignment only for Warp arrays.
- Complicates copying, serialization, model views, and solver compaction.
- Is invisible to the model’s normal attribute metadata.

The better alternatives are:

- For a specific built-in alias, define an explicit deprecated property on the owning class or namespace that warns and forwards to the canonical attribute.
- If Newton needs generic aliases later, extend `Model.AttributeSpec` with properly supported namespaced `deprecated`/`alias_of` metadata. That lets model views, serializers, and solvers handle aliases consistently.

newton/_src/utils/import_usd.py
eric-heiden: Done, fixed USD damping selection to honor schema-resolver priority before unit conversion.

newton/_src/sim/model.py
jcarius-nv: sounds good, thank you!

jcarius-nv (approved):
Re-reviewed at exact head d0ffc5dbe57582c0439d91519226155f42cc0952 after the latest fixes.

No blocking findings remain.

- Standards: Model.AttributeNamespace.add_deprecated_alias() is retained and deprecated for 1.6, with a runtime warning, regression coverage, and a Towncrier entry. API detection no longer reports its removal.
- Spec: USD damping selection now honors schema-resolver priority before source-specific unit conversion, with both mixed-schema orderings covered. Spherical, mixed D6, and MJCF ball-joint damping paths remain covered.
- Newton Fit: Fits. The change consolidates canonical joint-damping import behavior without introducing a parallel public concept.
- Validation: all hosted checks are green, including platform unit suites, GPU unit tests and benchmarks, docs, API detection, pre-commit, and license checks. Six focused strict-warning tests and the Towncrier draft also passed locally.

The previously raised Standards and Spec threads are addressed and resolved.

Nonblocking cleanup before merge: the PR description still says the AttributeNamespace deprecated-alias machinery is removed; the current head retains and deprecates it.

PR #3628, opened 2026-07-24
Fix USD texture import: packaged .usdz, material subsets, base-color and texcoord resolution

andrewkaufman:
## Description

Fixes a series of bugs in USD material/texture import that caused textured meshes to render incorrectly (missing, tinted, or scrambled textures), especially for assets authored in DCC/Omniverse tools.

- **Packaged `.usdz` textures** — package-relative asset paths (`scene.usdz[tex.png]`) were handed to Pillow as filesystem paths and failed to load. They are now resolved through USD's asset resolver (`Ar`), which also handles nested packages.
- **Material-subset `ValueError`** — importing a mesh whose material subset binds a texture that decodes to an image array raised `The truth value of an array with more than one element is ambiguous`. The subset path now uses `is not None` (matching the non-subset path).
- **Texture tinted by palette color** — a textured mesh with no authored scalar color was assigned the default per-shape palette color, which the shader multiplied into the texture (rainbow-tinted robots). Such meshes now import with a white base color so the texture shows untinted.
- **Wrong map chosen as base color** — the non-`UsdPreviewSurface` fallback grabbed the first connected/`file` input regardless of semantics, so a roughness/metallic/normal map could become the albedo. A connected `UsdUVTexture` is now accepted only through its multi-channel color output, and an MDL direct-asset color parameter (e.g. `diffuse_texture`) is identified by name.
- **Scrambled textures on non-`st` UV sets** — `get_mesh` looked only for a primvar literally named `st`, so meshes using `st_0` (or other arbitrary names) dropped their UVs and rendered garbage; the material-subset path also collapsed faceVarying UVs per vertex. The texcoord set is now resolved from the bound material's shader network (the `UsdPreviewSurface` texture reader's `varname`, or an MDL/OmniPBR `uv_space_index`), and textured subsets slice real per-corner UVs and authored normals.

## Checklist

## Test plan

```
uv run --extra dev -m unittest newton.tests.test_texture
uv run --extra dev -m unittest \
  newton.tests.test_import_usd.TestImportSampleAssetsParsing.test_visual_mesh_material_subset_with_loaded_texture_array \
  newton.tests.test_import_usd.TestImportSampleAssetsParsing.test_textured_visual_mesh_uses_white_base_color \
  newton.tests.test_import_usd.TestImportSampleAssetsParsing.test_fallback_texture_ignores_scalar_data_maps \
  newton.tests.test_import_usd.TestImportSampleAssetsParsing.test_fallback_texture_prefers_color_output \
  newton.tests.test_import_usd.TestImportSampleAssetsParsing.test_mdl_direct_asset_selects_diffuse_texture \
  newton.tests.test_import_usd.TestImportSampleAssetsParsing.test_get_mesh_loads_alternate_texcoord_set \
  newton.tests.test_import_usd.TestImportSampleAssetsParsing.test_get_mesh_uses_material_texcoord_set
```

Each fix ships with a regression test that fails on `main` and passes here. Verified visually by importing textured `.usd`/`.usdz` assets (an Isaac lab table with `st_0` texcoords and material subsets, and a packaged Anymal C `.usdz`) in the GL viewer.

## Bug fix

**Steps to reproduce:**

1. `python -m newton.examples ...` or `ModelBuilder.add_usd(...)` on a textured USD asset that uses any of: a packaged `.usdz` texture, material `GeomSubset`s, a non-`UsdPreviewSurface` (e.g. MDL/OmniPBR) shader, or a texcoord primvar named `st_0` rather than `st`.
2. Render it in a viewer that shows textures (e.g. `--viewer gl`).
3. Observe: a raised `ValueError`, or textures that are missing, rainbow-tinted, taken from the wrong map, or scrambled across the surface.

**Minimal reproduction:**

```python
import newton

# A textured USD/USDZ whose UV set is "st_0" (common from DCC exporters) and
# whose materials use GeomSubsets. Without this PR the mesh renders scrambled;
# get_mesh() drops the UVs because it only looks for a primvar named "st".
builder.add_usd("/path/to/textured_asset.usd")
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed loading of textures embedded in `.usdz` packages.
  - Corrected USD texture-coordinate selection, including custom UV set names and material-derived texcoord primvars.
  - Fixed scrambled textures and inconsistent UVs/normals on material-subset meshes.
  - Prevented textured meshes from appearing tinted when no scalar color is authored.
  - Improved base-color texture detection to prefer actual color outputs and ignore non-color maps.
  - Fixed cases where USD imports could fail when decoded texture data is represented as image arrays.


andrewkaufman:
@ooctipus since this touches the GeomSubsets that you recently refactored

andrewkaufman:
FYI this fixes some important Lab assets. I suspect we want to get it in for 1.5 @jcarius-nv

newton/_src/utils/import_usd.py
```diff
@@ -1060,7 +1065,11 @@ def _get_visual_material_subset_meshes(prim: Usd.Prim) -> list[tuple[str, Mesh]]
             return []
 
         subset_props = [(str(subset.GetPath()), usd.resolve_material_properties_for_prim(subset)) for subset in subsets]
-        mesh = _get_mesh_cached(prim)
+        # Load UVs (and matching authored normals) so each submesh slices real
+        # per-corner texture coordinates instead of recovering per-vertex UVs,
+        # which scrambles faceVarying UV sets. UV loading unwelds vertices while
+        # preserving triangle order, so the per-face subset selection still aligns.
+        mesh = _get_mesh_cached(prim, load_uvs=True, load_normals=True)
```
eric-heiden, line 1072: Can you check this finding from my agent:

🟠 Loading normals here can still collapse face-varying UV seams. In `usd/utils.py`, normal conversion groups corners only by position and normal direction, so corners with the same smooth normal but different UVs reuse one vertex and keep only the first UV. Because that conversion marks the vertices as already split, the later UV split is skipped.

Please include the corner UV in the normal-cluster key, or split face-varying UVs before clustering normals. A subset test with shared positions and normals but different per-corner UVs would cover this case.

eric-heiden:
Thanks, this looks good to me! There seems to be a test failure, possibly because of this PR's changes to uv handling?

newton/_src/utils/import_usd.py
andrewkaufman: Good catch — the finding was valid. In the `vertex_splitting` faceVarying-normal path, clusters were keyed on normal direction only, so corners sharing a smooth normal but different faceVarying UVs (a texture seam) merged onto one vertex and one UV was dropped. And because that path sets `did_split_vertices=True`, the later UV split was skipped, so nothing recovered the seam.

Fixed in 00e938542 by adding the corner UV to the cluster key — a normal match *and* a UV match are now both required to merge. Regression test `test_vertex_splitting_preserves_uv_seams` covers it (two corners, same +Z normal, different UVs → two vertices, both UVs preserved).

This also surfaced a latent crash on the same path: faceVarying UVs whose length ≠ corner count were indexed per-corner without a bounds check (`IndexError` on the Anymal/Go2 assets — the CI failure). Those are now dropped with an info log, matching the non-splitting path; `test_vertex_splitting_drops_mismatched_facevarying_uvs` covers it. All 31 robot-policy example tests pass now.

newton/_src/usd/utils.py
```diff
@@ -2449,6 +2564,69 @@ def _extract_preview_surface_properties(shader: UsdShade.Shader | None, prim: Us
...
+        connected = surface_input.HasConnectedSource()
+    except Exception:
+        connected = False
+    if not connected and _is_color_texture_input_name(surface_input.GetBaseName()):
+        asset = surface_input.Get()
+        if asset:
+            return _resolve_color_texture_asset(asset, prim, surface_input.GetAttr())
+    return None
+
+
```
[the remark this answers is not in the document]
andrewkaufman: Applied in 00e938542. `_color_texture_from_input` now gates on a base-color input name up front — a connected `UsdUVTexture` is accepted only when the input it feeds is a base-color parameter, so a normal map wired `outputs:rgb -> inputs:normal` is no longer mistaken for the diffuse map (the >=3-channel output check alone couldn't tell them apart). The now-redundant name check in the direct-asset branch was removed.

Verified the allowlist covers the shading models this path sees in USD: UsdPreviewSurface `diffuseColor`, MDL/OmniPBR `diffuse_texture`, and OpenPBR `base_color` — while `normal`/`geometry_normal`/`normalmap_texture` and even `emission_color`/`transmission_color` are correctly rejected. Regression test `test_fallback_texture_ignores_connected_normal_map` covers the normal-map case (with UVs present so a wrongly-selected texture would actually attach).

eric-heiden (approved):
Thanks, LGTM.

PR #3631, opened 2026-07-24
Fix MJCF primitive material colors

### LINKED ISSUE #3630: [BUG] MJCF primitive geoms ignore material colors
### Bug Description

`ModelBuilder.add_mjcf()` resolves both a geom's inline `rgba` and the `rgba` of its named material, but the resolved color is only forwarded when the geom is a mesh. Primitive geoms such as spheres, boxes, capsules, cylinders, ellipsoids, and planes therefore receive Newton's default palette color instead of the color authored in the MJCF.

The expected behavior is for every imported geom type to preserve the resolved RGB color, with inline `rgba` taking precedence over the named material as in MuJoCo.

### Reproduction Script

```python
import newton

mjcf = """
<mujoco>
  <asset>
    <material name="dark" rgba="0.2 0.3 0.4 1"/>
  </asset>
  <worldbody>
    <geom type="sphere" size="0.1" material="dark" contype="0" conaffinity="0"/>
  </worldbody>
</mujoco>
"""

builder.add_mjcf(mjcf)
print(tuple(builder.shape_color[0]))
# Actual: Newton default palette color
# Expected: (0.2, 0.3, 0.4)
```

### System Information

- Newton: `main` at `77b846e0`
- Warp: `1.16.0.dev20260716`
- Python: `3.12.3`
- OS: Linux

han-xudong:
## Description

Forward resolved MJCF material and inline `rgba` colors through the common shape arguments so primitive geoms preserve their authored display colors, just like mesh geoms. Inline `rgba` continues to take precedence over a named material.

This covers spheres, boxes, capsules, cylinders, ellipsoids, and planes.

Closes #3630.

## Checklist

## Test plan

```bash
uv run --extra dev python -m unittest newton.tests.test_import_mjcf.TestMjcfPrimitiveColors -v
uv run --extra dev python -m unittest newton.tests.test_import_mjcf -v
uv run --extra dev python -m unittest newton.tests.test_shape_colors -v
uvx pre-commit run -a
```

All 234 MJCF import tests, all 7 shape-color tests, and all pre-commit hooks pass.

## Bug fix

**Steps to reproduce:**

1. Import an MJCF containing a primitive geom with a named material or inline `rgba`.
2. Inspect `builder.shape_color` or render the resulting model.
3. Observe that the primitive uses Newton's palette color instead of the authored MJCF color.

**Minimal reproduction:**

```python
import newton

mjcf = """
<mujoco>
  <asset>
    <material name="dark" rgba="0.2 0.3 0.4 1"/>
  </asset>
  <worldbody>
    <geom type="sphere" size="0.1" material="dark" contype="0" conaffinity="0"/>
  </worldbody>
</mujoco>
"""

builder.add_mjcf(mjcf)
print(tuple(builder.shape_color[0]))
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed MJCF imports so authored material colors and inline RGBA colors are applied to primitive shapes.
  * Inline geom colors now take precedence over named material colors.

* **Tests**
  * Added coverage for material and inline color handling across supported primitive shapes.

* **Documentation**
  * Updated the changelog with the MJCF color import fix.

### DIFF
--- a/newton/_src/utils/import_mjcf.py
+++ b/newton/_src/utils/import_mjcf.py
@@ -826,6 +826,7 @@ def parse_mjcf(
                         float(rgba_values[1]),
                         float(rgba_values[2]),
                     )
+                    shape_kwargs["color"] = material_color
 
             texture = None
             texture_name = material_info.get("texture")
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,7 @@
 - Fix builder merging (`ModelBuilder.add_builder()`, `add_world()`, `replicate()`) offsetting negative reference sentinels in custom attribute values stored as NumPy or Warp integer scalars.
 - Fix `ModelBuilder.add_usd()` requiring the optional `mujoco` package when handling `MjcActuator` prims, including during default MJC equality conversion.
 - Report malformed MJCF free-joint and inertial inputs with deterministic validation errors, and ignore MJCF mesh geom `size` lengths consistently.
+- Fix MJCF imports ignoring material and inline RGBA colors on primitive geoms.
 - Fix Style3D solver divergence caused by isolated vertices.
 - Fix USD site import to discover sites beneath non-visual containers, collider prims, and instanceable rigid-body prims independently of `load_visual_shapes`; the reworked traversal also speeds up import of scenes with many nested `Xform` or instance prims.
 - Fix `SolverFeatherstone` BALL joints to apply passive `joint_damping` on all three angular DOFs.
--- a/newton/tests/test_import_mjcf.py
+++ b/newton/tests/test_import_mjcf.py
@@ -9326,5 +9326,56 @@ class TestOverrideRootXform(unittest.TestCase):
             )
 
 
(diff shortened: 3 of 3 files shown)


eric-heiden:
Thanks for contributing this handling of MJCF primitive colors!

newton/_src/utils/import_mjcf.py
```diff
@@ -826,6 +826,7 @@ def parse_shapes(
                         float(rgba_values[1]),
                         float(rgba_values[2]),
                     )
+                    shape_kwargs["color"] = material_color
```
eric-heiden, line 829: 🟠 This still gives the geom RGBA precedence whenever it is present, including when it equals MuJoCo's internal geom default of `0.5 0.5 0.5 1`. MuJoCo treats that default value as non-overriding when a material is assigned, so a primitive with a red material and an explicit or class-inherited default gray RGBA should remain red. Please ignore the geom RGBA in that case and add regression coverage for both explicit and default-class values.

eric-heiden (approved):

Issue #3150, opened 2026-06-12
[BUG] Non-deterministic hydroelastic marching-cubes tessellation (degenerate triangles) — test_no_degenerate_triangles_deep_penetration flaky on warp-lang 1.14.0 / some GPUs


adenzler-nvidia:
### Summary

`test_no_degenerate_triangles_deep_penetration` (added in #2424 to fix #2348) fails non-deterministically. The hydroelastic SDF → marching-cubes contact-surface generation produces a **different tessellation run-to-run** — sometimes a clean mesh (~80k faces, ~1.3–1.7% slivers), sometimes a denser one (~93k faces, **3.56%** slivers) that trips the `degen_pct < 2.0%` assertion. The geometry is always valid (finite, in-bounds); it is the *triangle set/count* that varies.

### Proof it is non-deterministic

newton 1.3.0 + `warp-lang==1.14.0` **toggles between 79,680 and 93,299 faces in a single process**, run to run. The outcome is also sensitive to the Warp build, GPU board, and process environment (matrix below), which is why it surfaced as an intermittent failure on a pure `pip install newton==1.3.0`.

### Evidence matrix (RTX PRO 6000 Blackwell, sm_120, fresh kernel cache, in-process-verified versions)

| newton | warp-lang | 95%-overlap config | result |
|---|---|---|---|
| 1.3.0 | 1.14.0 | toggles 1.30% ↔ 3.56% | **intermittent FAIL** |
| main | 1.14.0 (minimal env) | 3.56% | FAIL |
| main | 1.14.0 (full dev env) | 1.74% | pass |
| main | 1.14.0.dev20260518 / 1.15.0.dev20260612 | 1.74% | pass |

CI stays green because it happens to land on the good side on its board (same sm_120, same `warp-lang 1.14.0`, driver 13.0).

### Reproduction

```
uv run --extra dev -m newton.tests -k test_no_degenerate_triangles_deep_penetration
```

Most reliable on `warp-lang==1.14.0` in a minimal environment; with newton 1.3.0 it is intermittent (re-run several times).

### Impact

Flaky CI; degraded hydroelastic contact-surface quality (excess sliver triangles) for users on the released `warp-lang 1.14.0`; invisible in CI.

Related: #2348 (original, fixed-by #2424), #2137, #1620 (prior flaky hydroelastic tests).

nvtw:
See https://github.com/newton-physics/newton/pull/3632

Issue #2534, opened 2026-04-22
Flaky benchmark: setup.bench_sdf.FastBuildSdf.time_build_sdf fails unrelated PRs


adenzler-nvidia:
## Summary

`setup.bench_sdf.FastBuildSdf.time_build_sdf` (added in #2483) is flagging significant regressions on the `Pull Request - AWS GPU Benchmarks` workflow for PRs that do not touch SDF code. The benchmark job fails with `PERFORMANCE DECREASED` and blocks merging until re-run.

## Evidence

Looking at recent runs on `newton-physics/newton`, the base measurement for the `128` grid is bimodal across CI runs — landing at either ~22.5 ms or ~25.5 ms — while stddev within a single run is ~0.05 ms. When base and head happen to land in different modes, asv reports a 11-14% "regression" and fails the job.

Failures in the last ~24h on unrelated PRs:

| Run | Branch | `(128)` base → head | Result |
|---|---|---|---|
| [24777344164](https://github.com/newton-physics/newton/actions/runs/24777344164) | `vreutskyy/mjcf-ball-joint-import` | 22.6 → 25.1 (+1.11) | failed |
| [24777295922](https://github.com/newton-physics/newton/actions/runs/24777295922) | `vreutskyy/enable-cassie-dynamics` | 22.6 → 25.2 (+1.11) | failed |
| [24735849179](https://github.com/newton-physics/newton/actions/runs/24735849179) | `dev/tw3/contact_matching` | 22.5 → 25.1 (+1.11) | failed |
| [24726479129](https://github.com/newton-physics/newton/actions/runs/24726479129) | `ershi/fix-removed-warp-api` | 22.7 → 25.8 (+1.14) | failed |

The `256` case shows similar bimodal behavior (ratios range 0.78 ↔ 1.16 across runs).

## Likely cause

GPU boost clock / thermal state on the AWS EC2 runners. Within a single run both measurements are very stable; across runs the whole benchmark shifts by ~15%. The benchmark's `repeat=3, rounds=1, number=1` configuration doesn't give asv enough samples to absorb this.

## Short-term action

Skipped `time_build_sdf` via `@skip_benchmark_if(True)` to unblock PRs. See the follow-up PR.

## Suggested fix

- Lock GPU clocks on the benchmark runner (`nvidia-smi -lgc <clock>`), or
- Increase `rounds`/`repeat` and add a longer warmup pass in `setup()`, or
[...]

PR #3632, opened 2026-07-24
Attempt to stabilize flaky benchmarks

nvtw:
## Description
Stabilize SDF benchmarks by limiting gating coverage to resolutions with repeatable timings, and strengthen hydroelastic contact-surface repeatability coverage.

## Checklist

## Test plan
```
uv run --extra dev -m newton.tests -k test_no_degenerate_triangles_deep_penetration
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved correctness and repeatability of hydroelastic contact surfaces for primitive shapes, especially during deep penetration.
  * Ensured generated contact surfaces avoid degenerate triangles.

* **Benchmarks**
  * Streamlined GPU benchmark runs by using a reduced set of SDF grid resolutions during pull-request checks.

* **Documentation**
  * Updated the unreleased changelog with references to the related issues.


CHANGELOG.md
```diff
@@ -79,6 +79,7 @@
 - Fix USD plane visual width and length to scale along the axes defined by the `UsdGeomPlane` schema, and orient X- and Y-axis plane visuals along the authored axis.
 - Validate `ArticulationView` mask shapes and devices before launching selection kernels. (#3448)
 - Exclude active particles with non-finite positions from rebuildable `SolverImplicitMPM` sparse-grid packing.
+- Fix hydroelastic primitive texture SDF generation to sample analytic primitive distances instead of temporary tessellated meshes. (#3150, #3239)
```
jcarius-nv: The merge from `main` left this entry duplicated two lines below. Please consolidate them into one entry referencing `#3150, #3239`. While touching it, could we also describe the user-visible fix rather than the internal implementation detail (“sample analytic primitive distances instead of temporary tessellated meshes”), per the changelog guidance?

asv/benchmarks/setup/bench_sdf.py
```diff
@@ -47,13 +47,13 @@ def _build_sdf(mesh: "newton.Mesh", *, max_resolution: int) -> "newton.SDF":
 # Number of SDFs built per timing sample, keyed by ``max_resolution``.
 # Measuring a batch rather than a single build amortizes GPU boost-clock and
 # thermal transients that otherwise make this benchmark bimodal across AWS CI
-# runs (see #2534).  Counts decrease with resolution so each sample takes
-# roughly the same wall time (~0.5 s); each SDF is released immediately after
-# construction so peak GPU memory stays bounded to one SDF at a time.
+# runs (see #2534).  Resolutions of 128 and above remain bimodal despite
```
jcarius-nv: Could you link the repeated AWS measurements showing that resolution 128 is still bimodal after the current warmup and batching changes? The latest quantitative evidence in [#2534](https://github.com/newton-physics/newton/issues/2534#issuecomment-4565444873) says 32, 64, and 128 were stable and only 256 remained bimodal; this PR's base already excludes 256. The passing benchmark check does not validate the removal because, once 128 is absent from `params`, ASV compares only 32 and 64.

asv/benchmarks/setup/bench_sdf.py
```diff
@@ -151,14 +151,13 @@ class FastBuildSdf:
...
-    CI runners that previously made single-build measurements bimodal across
-    runs (see #2534).
+    releases each immediately, reporting the total wall time.  This amortizes
+    GPU boost-clock and thermal transients on AWS CI runners.  Resolutions of
+    128 and above are excluded because they remain bimodal despite batching
+    (see #2534).
     """
 
-    params = ([32, 64, 128],)
+    params = ([32, 64],)
```
jcarius-nv: Could we preserve resolution 128 as a nightly/non-gating benchmark even if current AWS evidence justifies removing it from the PR gate? Newton's external nightly infrastructure publishes the broader ASV results into [`newton-asv`](https://github.com/newton-physics/newton-asv), which currently records this benchmark at 32, 64, and 128. As written, changing the shared `params` definition to 32/64 will also stop future nightly collection at 128.

newton/tests/test_hydroelastic.py
```diff
@@ -1750,73 +1751,83 @@ def make_cfg(kh):
...
-        )
-
-        median_area = np.median(areas)
-        num_degen = int((areas < 0.01 * median_area).sum())
-        degen_pct = num_degen / num_faces * 100
-        test.assertLess(
-            degen_pct,
-            2.0,
-            f"[{label}] {degen_pct:.1f}% degenerate triangles (< 1% median area); expected < 2%",
+            len(set(face_counts)),
```
jcarius-nv, line 2067: The docstring says the repeated builds verify “the same tessellation,” but this assertion verifies only that the face count is stable. Two different triangle sets with the same number of faces would pass. Is stable face count the intended regression criterion for the known #3150 failure? If so, please narrow the docstring to say that. If exact contact-surface repeatability is required, could the test compare an order-independent geometry signature as well?

jcarius-nv (commented):
The hydroelastic test direction is sound. I left four focused comments covering the changelog cleanup, evidence and nightly coverage for resolution 128, and the repeatability criterion.

jcarius-nv (approved):
Thanks — I confirmed separately with Tobias that resolution 128 was also not sufficiently stable for PR gating.

The revised split looks appropriate: PR CI sets `NEWTON_ASV_PR_GATE` to limit this benchmark to resolutions 32 and 64, while the external nightly automation leaves it unset and therefore retains 32, 64, and 128. The current green benchmark run still included 128 because it used the reusable workflow from `main`, so the new split will only be exercised after merge; ASV propagates the step environment into benchmark discovery, so the implementation is sound.

The changelog consolidation and the stable-face-count wording are also addressed. Approving.

PR #3634, opened 2026-07-24
Add option to include collision geometry into the shape BVH

StafaH:
## Description

Certain applications require that collision meshes also be included in the shapes BVH. One specific use case is in a raycast sensor that is using the shapes BVH for queries against collision geometry.

This fix adds the option to include collision geometry into the shapes BVH.

Note: This is probably not the most ideal solution, but is a fix until a more optimal solution can be implemented.

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in way to select which shape types are included when building the model shape BVH (now controlled by a shape-flag mask rather than only “visible” shapes).
  * Ray intersection queries can now return collision-only shapes when those shape flags are requested.
* **Documentation**
  * Updated the changelog and BVH API docs to reflect selective “shape” BVH building and the new flag-based behavior.
* **Bug Fixes**
  * If no shapes are enabled, the previously built shape BVH results are now cleared.
* **Tests**
  * Added coverage for BVH inclusion and ray hit results with collision shapes enabled.


newton/_src/sim/model.py
```diff
@@ -1860,20 +1860,29 @@ def joint_target_vel(self, value: wp.array | None) -> None:
         )
         self.joint_target_qd = value
 
-    def bvh_build_shapes(self, state: State, *, bvh_constructor: str | None = None) -> None:
+    def bvh_build_shapes(
+        self,
+        state: State,
+        *,
+        bvh_constructor: str | None = None,
+        include_collision_shapes: bool = False,
```
eric-heiden: I think it might be better to let the user pass in the shape flags here directly, so default is just `ShapeFlags.VISIBLE`. This is probably more future-proof.

newton/_src/sim/builder.py
```diff
@@ -424,6 +424,9 @@ class BvhConfig:
         shape_constructor: str | None = None
         """Warp model shape BVH constructor backend. If ``None``, Warp's default is used."""
 
+        include_collision_shapes: bool = False
```
eric-heiden: Same here and in the other places, I would prefer to have a `ShapeFlags` setting.

newton/_src/sim/model.py
```diff
@@ -1911,6 +1920,7 @@ def bvh_build_shapes(self, state: State, *, bvh_constructor: str | None = None)
             inputs=[
                 self.shape_type,
                 self.shape_flags,
+                include_collision_shapes,
                 self.bvh_shape_enabled,
                 num_enabled,
             ],
```
eric-heiden, line 1926: 🔴 Clear `self.bvh_shapes` and `self.bvh_shapes_group_roots` when `bvh_shape_count_enabled` is zero. Otherwise, building a collision-only BVH and then rebuilding with `include_collision_shapes=False` leaves the old collision BVH active while the enabled-shape arrays have been replaced. `intersect_ray()` can then query stale leaves using those new arrays, which may return excluded colliders, produce invalid shape indices, or cause out-of-bounds GPU accesses. Please also add a regression test that performs this collision-only build followed by an empty default rebuild and verifies that the stale geometry is no longer queried.

StafaH: Done

newton/_src/sim/builder.py
StafaH: Done

newton/_src/sim/model.py
StafaH: Done

eric-heiden (approved):

Issue #3633, opened 2026-07-24
Remove the dahl_defaults_enabled compatibility mode from SolverVBD


adenzler-nvidia:
## Background

The implicit positive Dahl-friction defaults in `SolverVBD.register_custom_attributes()` were deprecated in 1.3.0 and disabled by default in 1.5: `vbd:dahl_eps_max` / `vbd:dahl_tau` now default to zero, and Dahl cable friction is enabled per joint only where both are authored positive. The `dahl_defaults_enabled=True` compatibility mode remains as a warned escape hatch.

## Scope

- [ ] Remove the `dahl_defaults_enabled` parameter from `SolverVBD.register_custom_attributes()`; the attributes always default to zero.
- [ ] Remove the associated `DeprecationWarning` and the compatibility-mode docs.

Blocked until the release flipping the default (the 1.5 `Changed` entry) has shipped.

PR #3636, opened 2026-07-24
Disable implicit VBD Dahl-friction defaults

adenzler-nvidia:
## Description

Flip the default of `dahl_defaults_enabled` in `SolverVBD.register_custom_attributes()` to `False`, completing the change announced by the 1.3.0 deprecation: `vbd:dahl_eps_max` and `vbd:dahl_tau` now default to zero, and Dahl cable friction is enabled per joint only where both are authored positive. Passing `dahl_defaults_enabled=True` still restores the old positive defaults but warns that the compatibility mode will be removed (#3633 tracks that removal).

Every in-tree caller already passed `dahl_defaults_enabled=False`; the now-redundant arguments are dropped across tests and examples, and the tests exercising the compatibility mode are removed. Behavior change for callers still on the old default: cable models silently lose implicit Dahl friction — exactly what the deprecation warning has announced since 1.3.0, and restorable per model by authoring positive `vbd:dahl_eps_max` / `vbd:dahl_tau` values (or temporarily via `dahl_defaults_enabled=True`).

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests --strict-warnings -k test_solver_vbd
uv run --extra dev -m newton.tests --strict-warnings -k test_cable
uv run --extra dev -m newton.tests --strict-warnings -k test_coupled_solver
uvx pre-commit run -a
```

All pass. The Dahl registration and auto-detection tests now assert the new contract (zero defaults, no warning on the default call, friction enabled only where both values are positive).

## Summary by CodeRabbit

* **Changed**
  * VBD solver registration no longer enables implicit Dahl-friction defaults.
  * Dahl friction is enabled only when both `vbd:dahl_eps_max` and `vbd:dahl_tau` are set to positive values.
  * Existing configurations can temporarily restore legacy behavior with the compatibility option `dahl_defaults_enabled=True`.
* **Documentation**
  * Updated release notes and usage examples to reflect the new default behavior.

### DIFF
--- a/newton/tests/test_solver_vbd.py
+++ b/newton/tests/test_solver_vbd.py
@@ -2200,19 +2200,7 @@ def _vbd_custom_attribute_registration_controls_dahl_defaults(test, device):
     del device
 
     builder = newton.ModelBuilder()
-    with warnings.catch_warnings(record=True) as caught:
-        warnings.simplefilter("always", DeprecationWarning)
-        newton.solvers.SolverVBD.register_custom_attributes(builder)
-    test.assertIn("vbd:joint_is_hard", builder.custom_attributes)
-    test.assertIn("vbd:dahl_eps_max", builder.custom_attributes)
-    test.assertIn("vbd:dahl_tau", builder.custom_attributes)
-    test.assertEqual(builder.custom_attributes["vbd:joint_is_hard"].default, 1)
-    test.assertEqual(builder.custom_attributes["vbd:dahl_eps_max"].default, 0.5)
-    test.assertEqual(builder.custom_attributes["vbd:dahl_tau"].default, 1.0)
-    test.assertTrue(any(issubclass(w.category, DeprecationWarning) for w in caught))
-
-    builder = newton.ModelBuilder()
-    newton.solvers.SolverVBD.register_custom_attributes(builder, dahl_defaults_enabled=False)
+    newton.solvers.SolverVBD.register_custom_attributes(builder)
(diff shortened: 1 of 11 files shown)


newton/tests/test_solver_vbd.py
```diff
@@ -2200,19 +2200,7 @@ def _vbd_custom_attribute_registration_controls_dahl_defaults(test, device):
...
-    test.assertIn("vbd:dahl_eps_max", builder.custom_attributes)
-    test.assertIn("vbd:dahl_tau", builder.custom_attributes)
-    test.assertEqual(builder.custom_attributes["vbd:joint_is_hard"].default, 1)
-    test.assertEqual(builder.custom_attributes["vbd:dahl_eps_max"].default, 0.5)
-    test.assertEqual(builder.custom_attributes["vbd:dahl_tau"].default, 1.0)
-    test.assertTrue(any(issubclass(w.category, DeprecationWarning) for w in caught))
-
-    builder = newton.ModelBuilder()
-    newton.solvers.SolverVBD.register_custom_attributes(builder, dahl_defaults_enabled=False)
+    newton.solvers.SolverVBD.register_custom_attributes(builder)
```
adenzler-nvidia, line 2203: not sure if all of these tests can actually be removed now?

jumyungc: I see what you mean, and yeah I am okay with removing no-argument behavior test.

jumyungc (approved):

PR #3639, opened 2026-07-24
Initialize per-world MuJoCo meshes from Newton

Idate96:
## Description

I found that `SolverMuJoCo(separate_worlds=True)` compiles one template
topology even though the corresponding Newton shape in each world may reference
a different mesh or mesh scale. Before this change, those models passed
topology validation while every MuJoCo-Warp world silently used the template
world's mesh and broadphase bounds.

This PR initializes mesh selection from the Newton model during conversion. It:

- discovers the corresponding mesh shapes in every finalized Newton world;
- registers each unique `(mesh, scale)` asset once, without carrier geoms;
- resolves each Newton mesh shape to its compiled MuJoCo `dataid`;
- caches the compiled AABB center, AABB half-size, and `rbound` for each shape;
- populates the per-world geom fields through the existing shape-property update
  path, including the selected asset's compiler-generated `mesh_pos` and
  `mesh_quat` recentering transform.

I am intentionally keeping this as a construction-time feature. It does not
add a runtime mesh setter or require mutation of private MuJoCo-Warp fields.

In my downstream task, I keep one mesh assigned to each world for the lifetime
of the solver; asynchronous episode resets only reset state and pose. I still
use `use_mujoco_contacts=True` because my verified contact contract depends on
MuJoCo-Warp contact generation: MultiCCD contact manifolds and native
`contact.dist` metadata. Explicit pair parameters are covered on the
Newton-contacts path by #3645 and #3662. Newton's contact pipeline remains the
simpler option for heterogeneous meshes when those native contact semantics
are not required.

## Checklist

## Test plan

```bash
uv run --no-sync -m newton.tests -k test_mujoco_solver.TestMuJoCoConversion
uvx pre-commit run -a
```

I added regressions that cover:

- two worlds authored with different mesh objects, with contact generated only
  in the world whose mesh reaches the obstacle;
- an asymmetric, translated, rotated mesh whose compiled recentering transform
  must be applied to the selected world;
- one shared mesh authored at different per-world scales, producing distinct
  assets, AABBs, and bounding radii;
- persistence of the selected asset and bounds after
  `notify_model_changed(SHAPE_PROPERTIES)`.

## Bug fix

**Steps to reproduce:**

1. I add two Newton worlds with matching topology but different meshes in the
   corresponding mesh-shape slot.
2. I construct `SolverMuJoCo(model, separate_worlds=True)`.
3. On `main`, validation succeeds, but both MuJoCo-Warp worlds use the template
   world's mesh ID and bounds.

**Regression:**  
`TestMuJoCoConversion.test_per_world_mesh_selection_drives_contacts`


adenzler-nvidia:
Thanks for the contribution! Per-world mesh selection has been on our roadmap for a while, but we haven't prioritized it because Newton's native contact pipeline already supports heterogeneous meshes across worlds — each shape has its own mesh, so you can build worlds with different meshes today and simulate them with `use_mujoco_contacts=False`. Is that not an option for your use case? It would be useful to understand what pulls you to MuJoCo-native contacts here.

On the change itself: the two expanded fields are correct and match the field set MuJoCo Warp documents for per-world meshes ([docs](https://mujoco.readthedocs.io/en/latest/mjwarp/)). But the resulting workflow feels quite hacky — as the test shows, a user has to smuggle candidate meshes into the compiled model via zero-density carrier shapes with `collision_group=0`, then reach into `solver.mjw_model` and hand-copy six fields (`geom_dataid`, `geom_size`, `geom_aabb`, `geom_rbound`, `geom_pos`, `geom_quat`) from the carrier geom's rows. That's easy to get subtly wrong (e.g. the mesh recentering offsets baked into `geom_pos`/`geom_quat`, or mass properties when density is nonzero), it's invisible to Newton's own collision pipeline and viewer, and a later `notify_model_changed(SHAPE_PROPERTIES)` will partially overwrite the copied values.

We think there's a more natural version of this that isn't much more work: initialize per-world meshes from the Newton model itself. Since Newton natively supports a different mesh per shape per world, the solver could discover heterogeneous meshes during conversion (the mesh export path already dedupes by `(mesh, scale)` key, it just only walks the template world), register all unique meshes in the spec (unreferenced mesh assets survive compilation, so no carrier geoms needed), and populate the per-world `geom_dataid` table automatically. Most dependent fields then come for free: `update_geom_properties_kernel` already resolves `geom_pos`/`geom_quat` through the per-world `geom_dataid`, `geom_size` comes from the per-world Newton shape, and mass properties already flow from Newton's per-world `body_mass`/`body_inertia`. The only missing piece is `geom_rbound`/`geom_aabb` for the extra meshes, which can be read from a single throwaway compile with one geom per unique mesh. This would also fix a silent-wrongness issue in the current code: `separate_worlds` validation only checks shape types, so a Newton model with per-world meshes currently passes validation and silently collides every world against the template world's mesh.

Would you be open to expanding the PR in that direction? Happy to go into more implementation detail if it helps. If you'd rather keep this PR minimal, we could also land it as-is and track the model-driven setup as a follow-up — but we'd want to understand the reset-time re-selection workflow you have in mind either way, since that determines whether we need a runtime setter API on top.

Idate96:
Thanks, this makes sense. I rechecked the reset requirement: I do not need
reset-time mesh re-selection. I assign one rock mesh to each world when I build
the model and keep it fixed across asynchronous resets, so I do not need a
runtime setter.

`use_mujoco_contacts=False` supports the heterogeneous geometry, but it is not
a drop-in change for my task. I use explicit MuJoCo bucket-rock and
rock-support pair laws, including nonzero `solreffriction`, and my contact
reducer reads MuJoCo-Warp `type`, `dim`, and signed `dist`. The
Newton-to-MuJoCo-Warp contact bridge currently derives parameters from geom
fields, zeros `solreffriction`, and does not populate `contact.type`, so
switching would change the contact contract.

I pushed `7edd8205` following your model-driven suggestion: it discovers each
world's mesh and scale, registers unique assets, and initializes `dataid`,
AABB, `rbound`, and the compiled recenter transform automatically. This removes
the carrier geoms and manual field copies; runtime mutation stays out of scope. Does this
match the scope you had in mind?

adenzler-nvidia:
Yes, this matches the model-driven construction-time scope I had in mind. Deriving the bounds from the compiled mesh vertices instead of using a throwaway compile looks reasonable, and the recentering and `notify_model_changed()` coverage address my main concerns. Thanks for reworking it. I’m reviewing the implementation in depth now.

adenzler-nvidia:
Thanks, that clarifies the remaining limitation. One distinction: #3580 will map Newton’s per-shape `kf` to `solreffriction` for elliptic cones on the Newton-contact path, so Newton material parameters will no longer always produce `(0, 0)`. It does not, however, apply an authored MuJoCo `pair_solreffriction` value.

More generally, I agree that explicit MuJoCo pairs are not currently supported end-to-end with Newton-generated contacts: the bridge does not resolve the MuJoCo pair ID or use the pair-level `condim`, friction, `solref`, `solreffriction`, `solimp`, margin, and gap values. Could you open a focused issue for that support, ideally with a small representative MJCF and a note about which fields your reducer depends on? That would give us a good place to scope the pair-selection and parameter-propagation work separately from this PR.

Idate96:
I kept the pair work separate from this mesh PR: #3645 handles pair parameters, and #3662 handles collision selection. This PR remains limited to construction-time per-world mesh initialization.

Idate96:
Following up on the scoping request: the focused issue is #3644, with the small MJCF example and the fields my reducer depends on. #3645 and #3662 implement it — pair parameters and pair selection. Together they cover most of the contact-contract gaps I listed above: pair-level `condim`, friction, `solref`, `solreffriction`, `solimp`, margin, and gap now apply to Newton-generated contacts, and the bridge populates `contact.type` and `dim` for them.

That narrows my earlier answer about staying on `use_mujoco_contacts=True`. What remains native-only for us is contact generation itself:

- bucket–rock box–mesh contacts use MuJoCo Warp's MultiCCD manifolds (up to four contacts per pair); with Newton-generated contacts, native generation is skipped, so MultiCCD does not apply;
- my reducer reads native `contact.dist` as the signed distance; on the Newton path, `dist` is reconstructed from Newton witness points at conversion time.

Each world carries a different rock mesh, so this PR is what makes the native path usable for us.

On ordering: #3645 → #3662 are stacked but independent of this PR, so the pair stack and this PR can land in either order. This PR and #3645 touch the same files (`solver_mujoco.py`, `kernels.py`); whichever lands second, I will rebase.

PR #3642, opened 2026-07-24
MuJoCo joint ref fixes

camevor:
## Description

Reconcile the MuJoCo joint-reference (`ref`) convention with Newton's joint coordinates for the MuJoCo solver.

MuJoCo scalar joints (hinge/slide) may carry a reference value (MJCF `ref`, exposed as `mujoco.dof_ref`): the authored body poses correspond to `qpos = ref`, and MuJoCo interprets limits, position-actuator controls, and keyframes as **absolute** `qpos`. Newton has no reference concept — its scalar joint coordinates measure **displacement from the authored pose**. Previously these two conventions were only partially reconciled, so `ref` leaked into the wrong places.

This PR fixes the convention at every boundary:

- **Newton side** (`joint_q`, `joint_qd`, `joint_limit_lower`/`joint_limit_upper`, `Control.joint_target_q`) are displacements from the authored pose: `qpos = joint_q + ref`. The MJCF importer now shifts authored joint ranges by `-ref`.
- **MuJoCo side** (`qpos`, `jnt_range`, position-actuator `ctrl`) keeps MuJoCo's absolute convention: `SolverMuJoCo` adds `ref` back when it writes `jnt_range` (compile-time and runtime), position-actuator `ctrl`, and derives `inheritrange` ctrlrange in absolute qpos. The compiled model and its runtime observables now match the authored MJCF and native MuJoCo exactly.
- **CONNECT equality anchors** are now derived at the authored reference pose (zero scalar joint coordinates) instead of at a `dof_ref`-displaced pose. `build_ref_q_kernel` emits zero scalar coordinates rather than copying `dof_ref` (the method was renamed `_copy_dof_ref_to_qref` → `_build_ref_q`). This removes a double-application of `ref` that displaced the anchors.

Changing `mujoco.dof_ref` at runtime now relabels the MuJoCo coordinates only: Newton displacements keep their meaning, while `qpos0`, `jnt_range`, and position controls shift with the new reference, and CONNECT anchors stay put.

**Migration:** callers who authored `joint_limit_lower`/`joint_limit_upper` or `joint_target_q` in absolute MuJoCo coordinates on models with nonzero `mujoco.dof_ref` must now subtract the reference value. See the `Changed` entry in `CHANGELOG.md`.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k TestMuJoCoRefCoordinates
uv run --extra dev -m newton.tests -k test_ref_shifts_joint_limits
uv run --extra dev -m newton.tests -k TestConnectAnchorRefPose
uv run --extra dev -m newton.tests -k TestLoopJointConnectConstraint
```

New/updated coverage:

- `test_import_mjcf.py`: `test_ref_shifts_joint_limits`, `test_ref_shifts_joint_limits_degrees` — importer shifts authored ranges by `-ref` (radian and degree paths); the unlimited sentinel is left unshifted.
- `test_mujoco_solver.py::TestMuJoCoRefCoordinates` — `jnt_range` is absolute, runtime limit/ref updates keep `jnt_range = limits + ref`, a position target of `x` drives `joint_q → x` regardless of `ref`, limits clamp in displacement space, and `inheritrange` ctrlrange is authored in absolute qpos.
- `test_equality_connect_constraint_with_sim_step.py::TestConnectAnchorRefPoseBase` (new) — CONNECT anchors derived at the reference pose with nontrivial rotation+translation, identical across worlds despite differing per-world `dof_ref`; the reworked loop-joint and connect-constraint tests now assert anchors are unchanged when `dof_ref` changes at runtime. Runs under both the Warp and MuJoCo-CPU backends.

## Bug fix

CONNECT equality anchors were derived at a pose displaced by the joint reference values (`mujoco:dof_ref`), applying `ref` a second time on top of the joint-coordinate offset convention. The resulting `anchor2` was wrong for any model whose CONNECT constraint spans joints with nonzero `ref`.

1. Build an articulation with hinge/slide joints carrying nonzero `mujoco:dof_ref`, closed by a CONNECT equality constraint.
2. Create a `SolverMuJoCo` for the model.
3. Inspect `solver.mjw_model.eq_data` (or `mj_model.eq_data`): the derived `anchor2` is offset by the extra `ref` displacement, so starting the sim at the authored reference pose (zero scalar coordinates) no longer satisfies the constraint and the bodies drift.

With this PR, anchors are derived at zero scalar coordinates (the authored pose), so they are independent of `dof_ref` and the constraint is satisfied at rest.

## Summary by CodeRabbit

- **New Features**
  - Added consistent MuJoCo joint-reference support across MJCF and USD imports.
  - Joint limits, position targets, and inherited actuator control ranges now account for authored reference offsets.
  - Angular reference values respect configured degree or radian units.
  - Added support for joint-owned mimic relationships, including D6 joints.

- **Bug Fixes**
  - CONNECT equality anchors and runtime reference updates remain consistent with offset joint coordinates.
  - USD collision-group memberships now preserve complete filtering behavior.

- **Documentation**
  - Added guidance on joint-reference handling at the Newton–MuJoCo solver boundary.


newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -7445,6 +7480,19 @@ def add_body_equality(i: int):
...
         )
 
+        actuator_custom_attr_idx = np.full(len(mjc_actuator_ctrl_source_list), -1, dtype=np.int32)
+        for actuator, (ctrl_source, newton_idx) in enumerate(
+            zip(mjc_actuator_ctrl_source_list, mjc_actuator_to_newton_idx_list, strict=True)
+        ):
+            if ctrl_source == int(SolverMuJoCo.CtrlSource.JOINT_TARGET):
+                is_position = newton_idx >= 0
+                dof = newton_idx if is_position else -(newton_idx + 2)
+                info = joint_target_ranges.get((dof, is_position))
```
vreutskyy, line 7493: **[P1] Ball-joint control-range updates skip Y and Z**

For a ball joint, `joint_target_ranges` stores the authored position/velocity range under the joint's base DOF, while `mjc_actuator_to_newton_idx_list` contains the individual axis DOFs. This lookup therefore maps only X; Y and Z retain `-1` and are skipped when actuator properties are updated. The construction path applies the same source range to all three axes, so a later update leaves their limits inconsistent.

Reproduced on both CPU and Warp with a ball joint containing position and velocity actuators: after changing `model.mujoco.actuator_ctrlrange` from `[-2, 2]` to `[-0.4, 0.4]` for position and from `[-3, 3]` to `[-0.7, 0.7]` for velocity, then calling `solver.notify_model_changed(newton.ModelFlags.ACTUATOR_PROPERTIES)`, only X receives the new ranges. Y/Z retain the original wider ranges. The generated source-row mapping is `[0, 1, -1, -1, -1, -1]` for X position/velocity, Y position/velocity, and Z position/velocity.

vreutskyy (changes requested):
Fit: Fits. Both previously reported issues (per-world inherited ranges and explicit control-range overrides) are fixed in this revision, and both original reproductions now pass. All 61 focused tests passed. One P1 correctness issue remains in the new runtime actuator-range mapping, described inline and reproduced on both CPU and Warp backends.

newton/_src/solvers/mujoco/solver_mujoco.py
camevor: Thanks for spotting this! Fixed in 830dbb23

vreutskyy (approved):
Re-reviewed ece2513a. Fit: Fits. All previously reported issues are fixed, with no remaining findings from my review. The ball-joint range-update reproduction now passes on both CPU and Warp. Additional checks pass with multiple ball joints, nonzero DOF offsets, reordered actuator rows, repeated updates, and separate worlds. All 48 focused tests passed, and CI is green. Thanks for addressing the feedback!

Issue #3644, opened 2026-07-24
Honor explicit MuJoCo pair parameters for Newton-generated contacts


Idate96:
## Problem

When I use `SolverMuJoCo(..., use_mujoco_contacts=False)`, Newton generates the contacts, but the bridge ignores a matching MuJoCo `<pair>`. The contact receives geom-level values even though `mjw_model.nxn_pairid` resolves an explicit pair.

This is separate from #3580. That PR provides a per-shape `kf` fallback for `solreffriction`; an authored `pair_solreffriction` should take precedence when a pair exists.

## Small example

```xml
<mujoco model="explicit_pair_probe">
  <option cone="elliptic"/>
  <default>
    <geom condim="1" friction="0.2 0.01 0.001"
          solref="0.04 1" solimp="0.8 0.9 0.001 0.5 2"/>
  </default>
  <worldbody>
    <geom name="floor" type="plane" size="1 1 0.1"/>
    <body name="box" pos="0 0 0.095">
      <freejoint/>
      <geom name="box" type="box" size="0.1 0.1 0.1"/>
    </body>
  </worldbody>
  <contact>
    <pair geom1="floor" geom2="box" condim="3"
          friction="0.8 0.7 0.02 0.003 0.004"
          solref="0.01 2" solreffriction="0.02 3"
          solimp="0.95 0.99 0.001 0.5 2"
          margin="0.003" gap="0.001"/>
  </contact>
</mujoco>
```

The compiled model contains the pair values. The generated contact currently gets geom `dim`, friction, `solref`, and `solimp`, with `solreffriction=(0, 0)` and zero margin.

## Expected behavior

For a contact that Newton already emitted, the bridge should use the matching pair's `condim`, friction, `solref`, `solreffriction`, `solimp`, margin, and gap. Without a pair, it should keep the current geom mixing and the #3580 `kf` fallback.

The precedence should be: explicit rigid-contact overrides, then an explicit pair, then per-shape `kf`, then geom mixing. Pair lookup must also preserve per-world values and global-pair fallback.

## Reducer dependency

My downstream reducer reads `contact.type` and `contact.dim` to reject inactive rows, and signed `contact.dist` for penetration filtering. It does not depend on task-specific geometry names.

## Scope

This only applies pair values to contacts Newton already generated. It does not make an explicit MuJoCo pair create a Newton collision candidate or expand Newton's collision envelope.

Idate96:
[...]

PR #3645, opened 2026-07-24
Honor MuJoCo pairs for Newton contacts

Idate96:
## Description

I make explicit MuJoCo `<pair>` parameters apply when `SolverMuJoCo` consumes contacts generated by Newton. This covers the parameter-propagation part of #3644.

I keep collision selection out of scope. Newton must already emit the contact; an explicit MuJoCo pair does not create a Newton collision candidate. MuJoCo-filtered lookup entries now fall back to geom parameters instead of deleting a contact that Newton emitted. #3662 adds the explicit-pair selection step.

After rebasing onto current `main`, I keep #3580's `kf` mapping behind `pairid < 0`, so an explicit `pair_solreffriction` still wins.

## Checklist

- [x] Changelog fragments cover the user-facing changes

## Test plan

- `uv run --extra dev -m newton.tests -p test_mujoco_solver.py -k 'test_mujoco_solver.TestMuJoCoSolverPairProperties' --no-cache-clear`: 9 passed
- `uv run --extra dev -m newton.tests -p test_mujoco_solver.py -k 'test_mujoco_solver.TestMuJoCoSolverContactKf' --no-cache-clear`: 10 passed
- `uv run --extra dev -m newton.tests -p test_mujoco_solver.py --no-cache-clear`: 279 run, 7 skipped
- `uvx --from towncrier==25.8.0 towncrier build --draft --version 1.6.0 --date 2026-08-26`: passed

## Bug fix

Before this change, Newton-generated contacts always used geom-level parameters. Explicit pair `condim`, friction, `solref`, `solreffriction`, `solimp`, margin, and gap were ignored.

MuJoCo's compiled filter table no longer removes contacts that Newton's collision pipeline already emitted when `use_mujoco_contacts=False`; negative pair IDs use the geom-parameter path.

### DIFF
--- a/newton/_src/solvers/mujoco/kernels.py
+++ b/newton/_src/solvers/mujoco/kernels.py
@@ -37,11 +37,19 @@ def _import_contact_force_fn():
 vec10 = wp.types.vector(length=10, dtype=wp.float32)
 vec11 = wp.types.vector(length=11, dtype=wp.float32)
 
+CONTACT_TYPE_CONSTRAINT = 1
+
 
 # Utility functions
 @wp.func
 def safe_div(x: float, y: float) -> float:
     return x / wp.where(y != 0.0, y, MJ_MINVAL)
+
+
+@wp.func
+def upper_tri_index(n: int, i: int, j: int) -> int:
+    """Return the index of ``(i, j)`` in an upper triangle without its diagonal."""
+    return (i * (2 * n - i - 3)) // 2 + j - 1
 
 
 @wp.func
@@ -101,6 +109,8 @@ def write_contact(
     contact_geom_out: wp.array[wp.vec2i],
     contact_efc_address_out: wp.array2d[int],
     contact_worldid_out: wp.array[int],
+    contact_type_out: wp.array[int],
+    contact_geomcollisionid_out: wp.array[int],
 ):
     # See function write_contact in mujoco_warp, file collision_primitive.py
 
@@ -110,6 +120,9 @@ def write_contact(
     contact_frame_out[cid] = frame_in
     contact_geom_out[cid] = geoms_in
     contact_worldid_out[cid] = worldid_in
+    contact_type_out[cid] = CONTACT_TYPE_CONSTRAINT
+    # Injected Newton contacts do not carry MuJoCo collision-sensor metadata.
+    contact_geomcollisionid_out[cid] = 0
     contact_includemargin_out[cid] = margin_in
     contact_dim_out[cid] = condim_in
     contact_friction_out[cid] = friction_in
@@ -132,7 +145,15 @@ def contact_params(
     geom_friction: wp.array2d[wp.vec3],
     geom_margin: wp.array2d[float],
     geom_gap: wp.array2d[float],
+    pair_dim: wp.array[int],
+    pair_solref: wp.array2d[wp.vec2],
+    pair_solreffriction: wp.array2d[wp.vec2],
+    pair_solimp: wp.array2d[vec5],
+    pair_margin: wp.array2d[float],
+    pair_gap: wp.array2d[float],
+    pair_friction: wp.array2d[vec5],
     geoms: wp.vec2i,
+    pairid: int,
     worldid: int,
 ):
     # See function contact_params in mujoco_warp, file collision_core.py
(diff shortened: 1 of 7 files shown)


Idate96:
I extended the existing omitted-default test in a403daea. It now starts from nonzero geom `solreffriction`, steps with Newton contacts, and verifies that an omitted pair value still resolves to `(0, 0)`. The focused test and pre-commit checks pass.

Idate96:
I opened #3662 for the collision-selection half, keeping this PR limited to pair-parameter propagation. The follow-up maps compiled MuJoCo pairs once at pipeline construction and does not change automatic collision filtering.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -1350,7 +1350,7 @@ def parse_solref_mode_usd(_value: Any, context: dict[str, Any]) -> int | None:
                 name="pair_solreffriction",
                 frequency="mujoco:pair",
                 dtype=wp.vec2,
-                default=wp.vec2(0.02, 1.0),
+                default=wp.vec2(0.0, 0.0),
```
eric-heiden, line 1411: Thanks for fixing the default here. Can you please update your PR from latest on main so we have changelog fragments, and another changelog entry for this default change?

Idate96: Done, I rebased onto current `main` and added separate #3644 changelog fragments for pair propagation and the corrected `solreffriction` default. I also kept #3580's `kf` fallback behind `pairid < 0`

newton/_src/solvers/mujoco/kernels.py
```diff
@@ -548,12 +598,21 @@ def convert_newton_contacts_to_mjwarp_kernel(
...
         geoms = wp.vec2i(geom_a, geom_b)
+        geom_lo = wp.min(geom_a, geom_b)
+        geom_hi = wp.max(geom_a, geom_b)
+        if geom_lo == geom_hi:
+            tid_to_cid[tid] = -1
+            return
+        pairid = nxn_pairid[upper_tri_index(ngeom, geom_lo, geom_hi)][0]
+        if pairid == -2:
+            tid_to_cid[tid] = -1
+            return
```
eric-heiden: 🟠 These four lines let MuJoCo's contact filters delete contacts that Newton's collision pipeline already decided to emit, which is collision selection rather than parameter propagation.

`nxn_pairid` has three states: `>= 0` is an explicit `<pair>` index, `-1` means "no explicit pair, use the geom parameters", and `-2` means "MuJoCo's compiled filters excluded this geom pair". Only the first two are needed to read pair parameters. Returning early on `-2` gives MuJoCo's filter table the final say over contacts that Newton produced.

That matters because `filterparent` is always on for these models. The solver only ever sets `mjDSBL_MULTICCD`, `mjDSBL_CONTACT`, and `mjDSBL_SENSOR`, never `mjDSBL_FILTERPARENT`, and mujoco_warp then marks every parent/child geom pair `-2`:

```python
parent_child_collision = filterparent & (weldid1 != 0) & (weldid2 != 0) & (...)
nxn_pairid_contact[~(mask & ~self_collision & ~parent_child_collision & ~exclude)] = -2
```

So a USD asset whose joint sets `physics:collisionEnabled = True` ends up with `collision_filter_parent=False`, Newton intentionally emits those parent/child contacts, and with `use_mujoco_contacts=False` they now silently get no contact response at all. The same applies to geoms MuJoCo welds into a single body that Newton does not filter, such as grandparent/grandchild across two fixed joints. This also works against the paragraph this PR adds to the docs: "This only changes contacts that Newton's collision pipeline already emits."

I would keep the lookup and drop the rejection. Every downstream branch already tests `pairid >= 0` or `pairid < 0`, so `-2` then behaves exactly like `-1` and falls through to the geom parameter path with the contact still injected:

```suggestion
        pairid = nxn_pairid[upper_tri_index(ngeom, geom_lo, geom_hi)][0]
```

If honoring MuJoCo's filter table is intended after all, that is a much larger behavior change. It would need `mjDSBL_FILTERPARENT` disabled when `use_mujoco_contacts=False` so MuJoCo cannot override Newton's `collision_filter_parent`, plus a docs update and a changelog fragment covering the dropped contacts.

Idate96: Good catch. I removed the `pairid == -2` rejection, so MuJoCo-filtered entries now fall through to the geom-parameter path without deleting Newton-emitted contacts. I added a parent-child regression for `collision_filter_parent=False`. the focused suites and full MuJoCo solver test file pass.

PR #3649, opened 2026-07-24
Fix masked coupled solver resets

maxkra15:
## Description

Builds on the canonical reset-mask contract introduced by #3726, now merged into `main`: a boolean mask with shape `(world_count + 1,)`, with local-world slots followed by the global-world `-1` slot. Established public solver reset entry points still accept the deprecated `(world_count,)` form, warn once, and normalize it immediately; internal reset paths and child-solver dispatch use only the canonical form.

`SolverCoupled.reset()` previously forwarded a mask to child solvers while its surrounding distribution, synchronization, reconciliation, and history clearing could still touch complete entry states. Some child solvers also cleared private history globally, so resetting one world could perturb an unselected world.

This PR:

- applies the canonical mask to coupled gather, scatter, synchronization, reconciliation, and child-solver dispatch;
- preserves unselected proxy, ADMM, implicit-MPM, MuJoCo/MJWarp, Kamino, VBD, and contact-matching state and history;
- clears selected proxy feedback and Aitken history with one consolidated masked kernel;
- evaluates canonical mutable masks on the device so CUDA reset paths remain graph-replay compatible;
- keeps all-false masks as no-ops for per-world state and persistent coupling/contact history;
- validates mask dimensionality, dtype, device, and canonical length without reading device values.

Generic reconciliation covers registered `STATE` attributes with explicit coupled-entry ownership. Custom frequencies require coupler-specific ownership logic. Native MuJoCo CPU synchronization remains host-based, and Kamino receives a local-world slice only at its private backend adapter.

## Checklist

## Validation

- `uvx pre-commit run -a` — passed
- CI-equivalent Sphinx HTML build with warnings as errors — passed (223 sources)
- Reset-focused affected suite with unexpected `DeprecationWarning`s treated as errors — 96 passed
- Custom/coupled/ADMM/contact modules — 183 passed
- MuJoCo/VBD/implicit-MPM modules — 239 passed, 1 skipped
- Kamino DVI/MuJoCo solver modules — 295 passed, 7 skipped
- Exercised on CPU, RTX 5090 (`cuda:0`), and RTX 4090 (`cuda:1`)

## Bug reproduction

1. Advance multiple worlds through a coupled solver so solver-owned and coupling history are populated.
2. Reset one local or global world with a canonical `(world_count + 1,)` mask.
3. Without this change, coupled synchronization or solver-owned history clearing can modify an unselected world; an all-false mask can also discard history.

## Summary by CodeRabbit

* **New Features**
  * Added world-selective (“masked”) reset support across solvers, coupling state, contact history, and warmstarts.
  * Added support for selecting global entities in reset masks.
  * Added per-world contact-history reset functionality.
* **Bug Fixes**
  * Masked resets now preserve unselected worlds and their coupling/contact history.
  * Added validation for invalid reset masks while retaining support for legacy mask formats.
* **Documentation**
  * Clarified reset-mask formats and proxy-coupling contact behavior.
* **Tests**
  * Expanded coverage across solvers, contact matching, validation, and CUDA replay.


newton/_src/solvers/solver.py
```diff
@@ -346,14 +346,45 @@ def reset(
...
+            raise ValueError(
+                f"'world_mask' length {world_mask.shape[0]} must equal model.world_count "
+                f"({world_count}) or model.world_count + 1 ({world_count + 1})."
+            )
+        return world_mask
+
+    @staticmethod
+    def _reset_world_mask_selects_any(world_mask: wp.array | None) -> bool:
+        """Return whether a validated reset mask selects at least one slot."""
+        return world_mask is None or bool(world_mask.numpy().any())
```
gdaviet: Blocking for CUDA graph capture: `.numpy()` introduces a device-to-host synchronization. Branching on its result also specializes behavior to the capture-time mask—for example, capturing with an all-false mask records no reset, so changing the mask before replay has no effect. Please avoid inspecting mask values on the host and let reset kernels treat an all-false mask as a device-side no-op.

newton/_src/solvers/coupled/solver_coupled.py
```diff
@@ -1919,26 +1944,42 @@ def reset(
...
+        world_mask = self._validate_reset_world_mask(world_mask)
+        if not self._reset_world_mask_selects_any(world_mask):
+            return
 
-        self._distribute_state(state, iteration_restart=True)
+        output_state_was_valid = self._entry_output_state_valid
+        if world_mask is None:
+            self._distribute_state(state, iteration_restart=True)
+        else:
+            reset_rows = self._reset_rows(world_mask.numpy())
```
gdaviet: This makes masked coupled reset neither capturable nor replayable with a mutable mask. `_reset_rows()` performs device readbacks and constructs index arrays from the capture-time mask. Could we precompute the row-to-world mappings and launch full-range masked gather/scatter kernels using the live device mask?

gdaviet (commented):
Graph-capturability and attribute-metadata findings.

newton/_src/solvers/coupled/solver_coupled.py
```diff
@@ -2024,6 +2105,70 @@ def _step_coupled(
     # State distribution and reconciliation
     # ------------------------------------------------------------------
 
+    def _reset_rows(self, mask: np.ndarray) -> dict[str, dict[str, _ResetRows]]:
+        """Build selected view and ownership mappings for a masked reset."""
+        model = self.model
+        global_worlds = {
```
gdaviet: These string-keyed dictionaries duplicate information already represented by `Model.AttributeFrequency`, `model.attribute_specs`, and `entry.attribute_local_to_global`. Could `_reset_rows` use `AttributeFrequency` keys and have its consumers select mappings through `spec.frequency`? This would remove the parallel `"body"`/`"particle"`/`"joint_coord"`/`"joint_dof"` taxonomy and allow registered custom `STATE` attributes to follow the generic reset path.

newton/_src/solvers/coupled/solver_coupled.py
```diff
@@ -1981,6 +2022,39 @@ def _sync_entry_reset_state(self, entry: SolverEntry) -> None:
...
+        self,
+        entry: SolverEntry,
+        rows: dict[str, _ResetRows],
+        world_mask: wp.array,
+        flags: StateFlags | int | None,
+    ) -> None:
+        """Synchronize selected public and solver-owned output history."""
+        reset_flags = int(StateFlags.ALL if flags is None else flags)
+        if entry.state_1 is not None and entry.state_1 is not entry.state_0:
+            for flag, name, domain in _RESET_STATE_FIELDS:
```
gdaviet: Could this synchronization use the model’s attribute metadata instead of the hard-coded `_RESET_STATE_FIELDS` table? The current path silently misses valid top-level attributes registered with `assignment=STATE`, including Kamino’s `body_f_total`, `joint_q_prev`, and `joint_lambdas`.

Iterating the attribute specs, filtering on `AttributeAssignment.STATE`, and using `spec.frequency` for row mapping would cover both namespaced and top-level attributes.

newton/_src/geometry/contact_match.py
```diff
@@ -721,17 +759,24 @@ def prev_contact_count(self) -> wp.array[wp.int32]:
...
+        reload).
         """
-        self._prev_count.zero_()
+        if world_mask is None:
+            self._prev_count.zero_()
+            self._reset_world_mask_host.fill(False)
+            self._reset_world_mask.zero_()
+            return
+
+        mask = world_mask.numpy()
```
gdaviet: This masked reset path is not graph-capturable: it reads the mask to the host and updates a host-side mirror that will not run again during graph replay. Could a device kernel OR `world_mask` into `_reset_world_mask`, eliminating both the synchronization and `_reset_world_mask_host`?

newton/_src/solvers/solver.py
maxkra15: Yes — this was the main graph-capture issue. Validation now checks only mask metadata and never reads mask values on the host. An all-false mask still records the device work as a no-op, and the new CUDA graph test changes the mask between replays on both GPUs.

newton/_src/solvers/coupled/solver_coupled.py
maxkra15: Agreed. I removed the runtime `_reset_rows()` materialization. Row-to-world and ownership mappings are now built once during construction, and reset launches full-range masked gather, scatter, and clear kernels against the live device mask. The mutable-mask replay test covers this path.

newton/_src/solvers/coupled/solver_coupled.py
maxkra15: I switched this over to `Model.AttributeFrequency` and `spec.frequency`, so reset no longer maintains a parallel body/particle/joint string taxonomy. Custom registered frequencies still use their metadata key and follow the same projection path.

newton/_src/solvers/coupled/solver_coupled.py
maxkra15: This is metadata-driven now. The reset attributes are discovered from `model.attribute_specs` by `AttributeAssignment.STATE`, while the core transient fields retain their explicit flags. That includes namespaced and top-level custom state; I added coverage for body, joint-coordinate, and joint-constraint attributes.

newton/_src/geometry/contact_match.py
maxkra15: Good point. `ContactMatcher` no longer mirrors or reads the mask on the host. A device kernel ORs the live mask into the pending reset mask, and the regression test captures once, mutates the mask, and replays it on both GPUs.

maxkra15:
@gdaviet One follow-up from testing this that I’d appreciate your take on. I kept this PR scoped to masked reset instead of adding broad configuration rejections, but two adjacent cases remain:

- Proxy `collide_interval > 1` is driven by a Python counter. Masked reset now preserves that cadence and cached contact history, but a captured full step freezes the collide/no-collide branch recorded at capture time, and a teleported selected world may reuse cached proxy contacts until the next scheduled refresh. Would you prefer a device-side cadence/conditional graph path, or should `> 1` simply be documented as outside full-step graph capture?
- Kamino’s private pre/post reset callbacks currently receive the full converted state and can run arbitrary Python, so they cannot strictly guarantee masked or replayable behavior. Would passing the mask through a callback-aware API be the right direction, or should callback configurations remain explicitly outside the guarantee?

I left both existing behaviors intact here rather than rejecting otherwise valid configurations.

gdaviet:
> @gdaviet One follow-up from testing this that I’d appreciate your take on. I kept this PR scoped to masked reset instead of adding broad configuration rejections, but two adjacent cases remain:
> 
>     * Proxy `collide_interval > 1` is driven by a Python counter. Masked reset now preserves that cadence and cached contact history, but a captured full step freezes the collide/no-collide branch recorded at capture time, and a teleported selected world may reuse cached proxy contacts until the next scheduled refresh. Would you prefer a device-side cadence/conditional graph path, or should `> 1` simply be documented as outside full-step graph capture?
> 
>     * Kamino’s private pre/post reset callbacks currently receive the full converted state and can run arbitrary Python, so they cannot strictly guarantee masked or replayable behavior. Would passing the mask through a callback-aware API be the right direction, or should callback configurations remain explicitly outside the guarantee?
> 
> 
> I left both existing behaviors intact here rather than rejecting otherwise valid configurations.

I think 1 is fine, at least for the scope of this PR. Typically if `collide_interval > 1` we should expect that the graph granularity will be a multiple of `collide_interval` steps. 
For the Kamino callbacks I would tend to agree that passing the `world_mask` to the callbacks would be helpful, but I don;t know much about them. This can likely be done as a follow-up, @newton-physics/disney-research for viz

chschuma-disney (dismissed):
Thanks for looking into this. The changes on Kamino's side look good to me.

Good catch pointing out the issues with the callbacks. I'll make sure that we track the issue and fix it in a follow-up PR.

I'm not too familiar with the contact match code, so it might make sense to ping someone to review those changes.

newton/_src/core/reset.py
```diff
@@ -0,0 +1,41 @@
...
+@wp.func
+def reset_world_selected(world: int, world_mask: wp.array[wp.bool], world_count: int) -> bool:
+    """Return whether an entity world is selected by a validated reset mask."""
+    if world >= 0 and world < world_count:
+        return world_mask[world]
+    return world == -1 and world_mask.shape[0] == world_count + 1 and world_mask[world_count]
+
+
+def validate_reset_world_mask(
+    world_mask: wp.array | None,
```
jcarius-nv: Project standards require bracket syntax for Warp-array annotations. This helper uses bare `wp.array` for both the parameter and return type, and the same pattern appears in several new coupled-reset helpers. Please use `wp.array[wp.bool]` here and `wp.array[Any]` where the dtype is genuinely generic.

newton/_src/sim/collide.py
```diff
@@ -1222,6 +1225,16 @@ def contacts(self) -> Contacts:
         self.model._add_custom_attributes(contacts, Model.AttributeAssignment.CONTACT, requires_grad=self.requires_grad)
         return contacts
 
+    def reset_contact_matching(self, world_mask: wp.array[wp.bool] | None = None) -> None:
+        """Clear all or reset-selected previous-frame contact history."""
```
jcarius-nv: The existing open thread covers `ContactMatcher.reset()`, but this separate new public method also exposes `world_mask` without documenting it. Please add a Google-style `Args:` entry describing the `W` / `W + 1` shape contract, boolean dtype/device requirement, global-entity slot, and `None` behavior.

newton/_src/solvers/coupled/solver_coupled.py
```diff
@@ -551,6 +749,37 @@ def _joint_state_indices(self, joints: Sequence[int]) -> tuple[wp.array, wp.arra
...
+            return joint_qd_indices.numpy().astype(np.int32, copy=False).tolist()
+        if frequency == model_frequency.JOINT_CONSTRAINT:
+            rows = []
+            for raw_joint in cfg.joints:
+                joint = int(raw_joint)
+                rows.extend(
+                    range(int(self._joint_constraint_starts[joint]), int(self._joint_constraint_starts[joint + 1]))
+                )
+            return rows
+        return []
```
jcarius-nv, line 784: Custom-frequency `STATE` attributes cannot currently be reconciled into the parent state. This falls through to `[]` for every non-built-in frequency, so no `attribute_owned_rows` mapping is created and `_reconcile_reset_state()` has nothing to scatter back after a child reset. The new regression covers custom attributes on built-in BODY, JOINT_COORD, and JOINT_CONSTRAINT frequencies only. Please either support ownership/reconciliation for registered custom frequencies or explicitly narrow the metadata-driven custom-state contract.

jcarius-nv (commented):
Standards and spec findings from an exact-head review of `5dac41a7`. I did not duplicate the existing open `ContactMatcher.reset()` docstring thread.

jcarius-nv:
Fit question for @nvtw: this PR adds `CollisionPipeline.reset_contact_matching()` as stable public API, while its primary consumer here is the experimental coupled-solver path. Since you have been working on the collision pipeline, could you assess whether there is an intended durable standalone user workflow for this method? If not, I think the reset repair still fits Newton, but the hook should remain internal or be explicitly experimental to avoid an unnecessary long-term public contract.

newton/_src/solvers/coupled/solver_coupled.py
maxkra15: yes- narrowed the generic contract to `STATE` attributes on frequencies with explicit entry ownership - unsupported custom frequencies emit a construction-time warning and require coupler-specific parent reconciliation. 
Inferring ownership for arbitrary custom rows needs a separate ownership design

nvtw:
> Fit question for @nvtw: this PR adds `CollisionPipeline.reset_contact_matching()` as stable public API, while its primary consumer here is the experimental coupled-solver path. Since you have been working on the collision pipeline, could you assess whether there is an intended durable standalone user workflow for this method? If not, I think the reset repair still fits Newton, but the hook should remain internal or be explicitly experimental to avoid an unnecessary long-term public contract.

I think this new reset method makes sense. But marking it experimental sounds good - we can always make it official later once it proofed its usefulness.

nvtw:
I have not validated it by hand the the AI review agent spit out this:

  > Blocking: preserve native MuJoCo state when world 0 is not selected.
  >
  > With use_mujoco_cpu=True and update_data_interval != 1, a world-1-only or global-only reset still calls
  > _update_mjc_data(). The native path allocates temporary qpos/qvel; the conversion kernel exits because world_mask[0]
  > is false, but those unwritten buffers are then copied into MjData, corrupting the unselected template world.
  >
  > Please skip the native _update_mjc_data() call when world 0 is not selected, or initialize/preserve its temporary
  > buffers. Add a regression test using use_mujoco_cpu=True, update_data_interval=0, flags=JOINT_Q | JOINT_QD, and
  > masks (False, True) and (False, False, True).

jcarius-nv (commented):
Fit follow-up from @nvtw’s collision-pipeline assessment.

newton/_src/sim/collide.py
```diff
@@ -1222,6 +1225,28 @@ def contacts(self) -> Contacts:
         self.model._add_custom_attributes(contacts, Model.AttributeAssignment.CONTACT, requires_grad=self.requires_grad)
         return contacts
 
+    def reset_contact_matching(self, world_mask: wp.array[wp.bool] | None = None) -> None:
+        """Clear all or reset-selected previous-frame contact history.
```
jcarius-nv, line 1263: Following @nvtw’s assessment, please mark this new public method experimental before merge. The capability is useful, but its standalone public workflow is not yet proven; an `.. experimental::` directive keeps the contract reversible until usage validates promotion.

jcarius-nv:
I confirmed @nvtw’s native-MuJoCo finding on the current head with a focused reproduction: `use_mujoco_cpu=True`, `separate_worlds=True`, `update_data_interval=0`, `world_mask=(False, True)`, and `JOINT_Q | JOINT_QD` changed the unselected template world’s sentinel `qpos`/`qvel` values to uninitialized values. This should be fixed and covered by the requested world-1-only/global-only regression before merge.

Apart from this blocker and marking `CollisionPipeline.reset_contact_matching()` experimental, the original Standards/Spec findings are addressed and the change looks good to me. The two external-PR manual-approval check gates are still pending, so the full hosted test matrix has not run on this head yet.

newton/_src/sim/collide.py
maxkra15: marked experimental

maxkra15:
> I confirmed @nvtw’s native-MuJoCo finding on the current head with a focused reproduction: `use_mujoco_cpu=True`, `separate_worlds=True`, `update_data_interval=0`, `world_mask=(False, True)`, and `JOINT_Q | JOINT_QD` changed the unselected template world’s sentinel `qpos`/`qvel` values to uninitialized values. This should be fixed and covered by the requested world-1-only/global-only regression before merge.
> 
> Apart from this blocker and marking `CollisionPipeline.reset_contact_matching()` experimental, the original Standards/Spec findings are addressed and the change looks good to me. The two external-PR manual-approval check gates are still pending, so the full hosted test matrix has not run on this head yet.

Confirmed and fixed. Native MuJoCo now skips template qpos/qvel sync when world 0 is unselected, while still resetting selected Newton worlds. I added regressions for both (False, True) and (False, False, True) with use_mujoco_cpu=True, update_data_interval=0, and JOINT_Q | JOINT_QD. The full MuJoCo reset selection passes.

eric-heiden:
Thanks, @maxkra15 can you please rebase this PR on top of the changes in https://github.com/newton-physics/newton/pull/3726? There are a couple of changes here related to the coupled-reset handling we should have but it would be better to enforce one convention for the world masks across newton, i.e. to have the array with dimension `world_count + 1` rather than both `world_count` and `world_count + 1`.

maxkra15:
> Thanks, @maxkra15 can you please rebase this PR on top of the changes in #3726? There are a couple of changes here related to the coupled-reset handling we should have but it would be better to enforce one convention for the world masks across newton, i.e. to have the array with dimension `world_count + 1` rather than both `world_count` and `world_count + 1`.

had my agent rebase, can you check this matches your idea @eric-heiden ?

newton/_src/solvers/coupled/solver_coupled.py
```diff
@@ -109,6 +110,53 @@ def _compact_visible_shape_contact_pairs_kernel(
...
+    local_to_global: wp.array[int],
+    global_row_world: wp.array[int],
+    world_mask: wp.array[wp.bool],
+    world_count: int,
+    values: wp.array[Any],
+):
+    local_id = wp.tid()
+    global_id = local_to_global[local_id]
+    if global_id >= 0 and _reset_world_selected(global_row_world[global_id], world_mask, world_count):
+        values[local_id] = values.dtype(0.0)
```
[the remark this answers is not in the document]
maxkra15: I reproduced the exact kernel behavior:

  - values.dtype(0.0) works for scalars, vectors, spatial vectors, and matrices.
  - It fails Warp compilation for wp.transform and wp.quat.
  - body_q is a wp.transform, so the failure is reachable when the supplied parent state lacks body_q.
  - Normal model-created states provide body_q, so this is currently a latent edge case rather than a common reset failure.

We should not blindly replace it with values.dtype() inside the kernel: that fixes transforms but fails for scalar fields such as joint_q and joint_qd.

A fix could be to pass a typed value into the same generic kernel:

```
  def _zero_reset_view_rows_kernel(..., values:
  wp.array[Any], zero_value: Any):
      ...
      values[local_id] = zero_value
```
Launch it with values.dtype(0). This retains one kernel, requires no allocation or device synchronization, and remains graph-capturable. We should add one scalar and one transform specialization test.

eric-heiden:
🔵 Nice work making the reset mask handle global entities in world `-1` explicitly, and thanks for updating the docs and CHANGELOG along with it.

One thing to sort out before merge: the new `(world_count,)` deprecation is only written as prose in the `reset()` docstrings, so the rendered API pages do not carry the versioned deprecation marker that the project's compatibility policy asks for.

newton/_src/solvers/solver.py
```diff
@@ -346,14 +356,18 @@ def reset(
...
             state: The simulation state to reset (modified in place).
-            world_mask: Optional boolean mask of shape ``(num_worlds,)``
-                specifying which worlds to reset.  If ``None``, all worlds
-                are reset.
+            world_mask: Optional boolean mask of shape ``(world_count + 1,)``
+                specifying which worlds to reset. Entries before the last select
+                local worlds by index, and the final entry selects global entities
+                whose world is ``-1``. If ``None``, all local and global entities
+                are reset. Passing the deprecated shape ``(world_count,)`` selects
+                local worlds only and leaves global entities unselected.
```
eric-heiden: 🟠 The deprecation of the `(world_count,)` mask shape is described here only as prose, so it never renders as a deprecation admonition in the API docs.

`docs/guide/compatibility.rst` states that deprecated symbols are marked with a `.. deprecated:: X.Y` directive naming the version in which the deprecation was introduced, and the codebase already follows that elsewhere (for example the deprecated `dahl_defaults_enabled` argument of `SolverVBD.register_custom_attributes`). This PR adds the directive to `docs/concepts/worlds.rst` only. The practical effect is that someone reading the autosummary page for `reset()` or running `help(solver.reset)` gets no version number and no styled warning box, just one sentence in the middle of a parameter description.

The runtime `DeprecationWarning` and the CHANGELOG entry are already in place, so the API docs are the remaining piece.

```suggestion
            world_mask: Optional boolean mask of shape ``(world_count + 1,)``
                specifying which worlds to reset. Entries before the last select
                local worlds by index, and the final entry selects global entities
                whose world is ``-1``. If ``None``, all local and global entities
                are reset.

                .. deprecated:: 1.5
                    The shape ``(world_count,)`` is deprecated. It selects local
                    worlds only and leaves global entities unselected. Append a
                    final entry to select global entities whose world is ``-1``.
```

Once `SolverBase.reset` carries the directive, please mirror the same block in the overriding docstrings so their rendered pages match: `SolverVBD.reset`, `SolverMuJoCo.reset`, `SolverKamino.reset`, `SolverImplicitMPM.reset`, and `SolverCoupled.reset`.

maxkra15: Added `.. deprecated:: 1.5` blocks to `SolverBase.reset()` and all five overriding solver docstrings, with consistent `(world_count + 1,)` migration guidance. The CI-equivalent Sphinx HTML build now renders all 223 sources successfully with warnings treated as errors.

docs/concepts/coupling.rst
```diff
@@ -207,9 +207,11 @@ usually better served by a custom harvest hook.
 Proxy-local collision detection is optional. A proxy can provide a
 ``collision_pipeline`` factory that receives the destination ``ModelView``. If
 the factory returns a pipeline, the coupler owns a persistent contact buffer and
-refreshes it at ``collide_interval``. If the factory returns ``None`` or no
-factory is supplied, the destination solve uses contacts passed to the outer
-``step()`` call.
+refreshes it at ``collide_interval``; inner proxy iterations reuse that result.
+Masked resets preserve both this cadence and cached contacts, and mark
+selected-world matching history for invalidation on the next refresh. If the
```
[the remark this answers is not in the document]
maxkra15: Updated the coupling docs so the pending matching-history invalidation explicitly follows every entity selected by the mask, including global entities selected by the final slot.

newton/_src/solvers/coupled/solver_coupled.py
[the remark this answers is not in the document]
maxkra15: Implemented the typed-launch approach: the existing generic kernel now receives `values.dtype(0)` and assigns that typed value directly. Added regression coverage for both `wp.transform` (`body_q`) and scalar (`joint_q`) fallback rows, including selected-row clearing and unselected-row preservation. This keeps one kernel and remains allocation-free, host-read-free, and graph-capturable.

eric-heiden (dismissed):
Thanks!

newton/_src/solvers/coupled/interface.py
```diff
@@ -242,6 +245,30 @@ def coupling_notify_input_state_update(
         """
         del state, flags, iteration_restart, dt
 
+    def coupling_sync_reset_state(
```
eric-heiden: Is this new method needed? Does it have to be public API? If not, let's keep this internal. I would also recommend renaming it so the verb comes first, i.e. something like `sync_reset_state()` or so. But I don't really see why this is needed.

maxkra15: Removed coupling_sync_reset_state() from CouplingInterface, its documentationcand call site - It had no overrides and was a no-op. Its original MPM purpose is now handled generically through registered STATE attributes.

jcarius-nv (approved):

PR #3650, opened 2026-07-25
Keep guide-purpose colliders off the visual toggle

andrewkaufman:
## Description

A collider with a bound render material was given the `VISIBLE` flag regardless of its
purpose. Viewers draw a shape when `(COLLIDE and show_collision) or (VISIBLE and
show_visual)`, so such a collider was drawn by the *visual* toggle and could not be
hidden by the *collision* toggle — it occluded the very visual shapes it sat beside.

`guide` is the conventional purpose for authored collision geometry (the MuJoCo USD
exporter emits it), so honour it: purpose now gates only whether a collider inherits
visibility from its bound render material. `force_show_colliders` is checked
independently and still reveals `guide` colliders, which is what the previous
behaviour was protecting — that concern is preserved, it was just conflated with
material-inherited visibility.

Note for reviewers: this changes behaviour for existing assets. A `guide`-purpose
collider with a render material used to render and now does not, unless
`force_show_colliders=True`. Filed under `Fixed` since the old behaviour made the
viewer's collision toggle inoperative on such assets, but it is worth a second
opinion if anyone is relying on it.

## Checklist

## Test plan

```
uv run --extra dev -m unittest newton.tests.test_import_usd -k guide_purpose_collider_with_material
uv run --extra dev -m unittest newton.tests.test_import_usd   # 282 tests
uv run --extra dev -m newton.tests -k viewer                  # 221 tests
uvx pre-commit run -a
```

Verified the new test fails without the fix (`AssertionError: 1 is not false`) and
passes with it.

## Bug fix

**Steps to reproduce:**

1. Author a rigid body with a visual mesh and a separate collision mesh.
2. Give the collision mesh `uniform token purpose = "guide"` and bind a render
   material to it.
3. Import with `ModelBuilder.add_usd()` and inspect the collider's shape flags, or
   view it with `show_visual=True, show_collision=False`.
4. The collider carries `VISIBLE`, so it is drawn by the visual toggle and the
   collision toggle cannot hide it.

**Minimal reproduction:**

```python
import newton
from newton import ShapeFlags

shape = builder.add_usd(stage)["path_shape_map"]["/Body/CollisionMesh"]

# Before: True  (guide-purpose collider drawn by the visual toggle)
# After:  False (collision toggle owns it; force_show_colliders still reveals it)
print(bool(builder.shape_flags[shape] & ShapeFlags.VISIBLE))
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed visibility handling for guide-purpose colliders with bound render materials.
  * These colliders now follow collision display settings instead of being shown automatically.
  * The force-show-colliders option continues to display them when enabled.

* **Tests**
  * Added regression coverage for collider visibility and force-show behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,7 @@
 
 ### Fixed
 
+- Fix `ModelBuilder.add_usd()` marking a `guide`-purpose collider visible when it has a bound render material. Such a collider is not viewport geometry, and the extra `VISIBLE` flag left it drawn by the viewer's visual toggle instead of its collision toggle. `force_show_colliders` still reveals it.
 - Fix USD capsule, cylinder, and cone visual and site scaling to follow the authored primitive axis.
 - Fix USD plane visual width and length to scale along the axes defined by the `UsdGeomPlane` schema, and orient X- and Y-axis plane visuals along the authored axis.
 - Validate `ArticulationView` mask shapes and devices before launching selection kernels. (#3448)
--- a/newton/_src/utils/import_usd.py
+++ b/newton/_src/utils/import_usd.py
@@ -1200,11 +1200,11 @@ def parse_usd(
 
         USD viewports draw the ``default`` and ``proxy`` purposes and hide ``guide`` and
         ``render``; the allowlist also keeps any future purpose hidden until explicitly
-        handled. Colliders deliberately do not use this check: ``guide`` is the conventional
-        purpose for authored collision geometry (e.g. the MuJoCo USD exporter), and the
-        collider display policy (``force_show_colliders`` / ``hide_collision_shapes``) is the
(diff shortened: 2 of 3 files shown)


andrewkaufman:
Closing in favour of a single consolidated PR.

@eric-heiden asked for the visual/collider visibility work to land together rather than piecemeal, and while investigating this it became clear the guide-purpose fix here is one facet of a broader issue: whether a collider is drawn should follow USD purpose and visibility, not whether a render material happens to be bound to it. That same question governs approximated colliders (which currently discard their authored render mesh) and interacts with #3694.

Rolling this commit, the approximation fix, and the remaining drawability work into a `fix-shape-loading` branch. Will open that once it accounts for #3694.

Issue #1337, opened 2026-01-13
Flaky test: test_robot.example_robot_panda_hydro_cuda_0


Kenny-Vilella:
We have some occasional failures associated with this test on the main branch.
For example:
https://github.com/newton-physics/newton/actions/runs/20940565862/job/60173118807

Kenny-Vilella:
This should have been fixed in #1328.

Kenny-Vilella:
This test is being flaky again:
https://github.com/newton-physics/newton/actions/runs/24100609484/job/70310838163

Reopening.

Kenny-Vilella:
Still flaky:
https://github.com/newton-physics/newton/actions/runs/24188104026/job/70598056498

adenzler-nvidia:
Reopening — the gripper-sequence fix in #2467 reduced the failure rate but did not eliminate it.

**New failure** on PR #2593 CI attempt 1: https://github.com/newton-physics/newton/actions/runs/24982759302/job/73148853305

```
AssertionError: World 0: Object is not in the cup.
Object pos=(0.157, 0.258, 0.004), cup pos=(0.130, -0.500, 0.150)
```

The pen ends **~0.76 m off in Y** — i.e. the pen slipped out of the gripper completely and landed across the table, not just outside the 5 cm tolerance. Qualitatively worse than the small drifts seen in earlier flake reports.

**Local repro** on a Blackwell (sm_120, CUDA driver 12.9), `python -m newton.tests -k test_robot.example_robot_panda_hydro_cuda_0`: **1/3 runs fail**. Local failure mode is the small-drift case (~0.1 m off in X), not the catastrophic miss seen in CI.

Two distinct failure modes (small drift vs. complete slip) suggest the loose-grip transport phase still occasionally lets the pen go entirely on certain GPU contact orderings — #2467 narrowed the window but didn't close it.

Cloud nightly on main has been green 10/10 (Apr 17–26), but PR-target GPU runs and local runs see the failure. The PR-target runs use a different EC2 image, which may explain the rate gap.

jcarius-nv:
After discussion with @nvtw this test should become (more) stable once determinism improvementsa are merged. => depends on #2479

PR #3661, opened 2026-07-27
Add support for deterministic contact detection when hydroelastic contacts are used

nvtw:
## Description

Add support for deterministic contact detection when hydroelastic contacts are used.
`CollisionPipeline(deterministic=True)` previously left hydroelastic contacts uncovered, so a scene using them was not reproducible even in deterministic mode.
Hydroelastic contacts now reach determinism through three mechanisms:
1. **Fingerprint tiebreaking** — each face carries a stable geometric identifier (`voxel_idx * 5 + face_idx`) used as the `atomic_max` tiebreaker during reduction and as the contact sort sub-key, so the reduction winner and the final ordering no longer depend on thread scheduling.
2. **Radix sort** — the existing narrow-phase sort now receives stable sub-keys from both hydroelastic paths (reduced and unreduced).
3. **Fixed-point aggregation** — hydroelastic reduction is the only contact path that sums contributions across threads. Those per-normal-bin aggregates (contact force, center of pressure, depth volume, friction moments) are accumulated as int64 fixed point instead of with floating-point atomics. Integer addition is associative, so the sums are independent of the order threads arrive, and no ordering constraint, sorting, or kernel-lowering machinery is needed. A first pass records the largest contribution per bin as a binary exponent, which sizes that bin's fixed-point grid; the mantissa width is derived from how many terms can reach one bin, so overflow is impossible by construction rather than by assumption.
Determinism is opt-in and off by default. When it is off, the generated code and launch sequence are unchanged, so there is no behavior or performance regression for existing users.
Note: deterministic mode disables hydroelastic pre-pruning, so the generated contact set differs from the non-deterministic default.
Also adds `SolverMuJoCo(disable_sensors=...)` and a `--deterministic` flag to the Panda hydroelastic example.
Related to newton-physics/newton#1337 (the example's in-cup placement assertion stays disabled; see the Test plan).

## Checklist

## Test plan

Determinism (bit-exact repeatability across repeated `collide()` calls), covering the reduced, moment-matching, and unreduced paths:
```
uv run --extra dev -m newton.tests -k test_deterministic_hydroelastic_contacts
uv run --extra dev -m newton.tests -k test_deterministic_hydroelastic_contacts_moment_matching
uv run --extra dev -m newton.tests -k test_deterministic_hydroelastic_contacts_unreduced
```
Physics correctness of the fixed-point aggregates, compared against the unreduced reference rather than only against themselves. These guard against the "reproducibly wrong" failure mode: the first covers `agg_force` / `total_depth_reduced` which drive contact stiffness (1% tolerance), the second covers all three friction-moment accumulators (40% tolerance):
```
uv run --extra dev -m newton.tests -k test_reduced_vs_unreduced_contact_forces_deterministic
uv run --extra dev -m newton.tests -k test_reduced_vs_unreduced_contact_moments_deterministic
```
Fixed-point overflow invariant (host-side, runs on CPU-only CI):
```
uv run --extra dev -m newton.tests -k test_fixed_point_accumulator_cannot_overflow
```
Full suites:
```
uv run --extra dev -m newton.tests -k test_hydroelastic          # 60 tests, OK
uv run --extra dev -m newton.tests -k test_collision             # 259 tests, OK
uv run --extra dev --extra examples -m newton.tests -k example_robot_panda_hydro
```
Each new regression test was verified to fail without its fix. `test_deterministic_hydroelastic_contacts_unreduced` fails 7/7 repeats when the unreduced path sorts on the buffer slot instead of the fingerprint; `test_fixed_point_accumulator_cannot_overflow` fails with `9223372036854775808 not less than or equal to 9223372036854775807` if the mantissa width reserves one bit too few.
Performance, measured on an RTX 3080 Laptop with the stacked-cubes scene under CUDA graph capture (median of 300 replays):
- `deterministic=False`: 1.446 ms/collide — unchanged from `main`, confirming no regression when determinism is off.
- `deterministic=True`: 1.495 ms/collide — **+3.4%**.
The Panda example's in-cup placement assertion remains disabled. With `--deterministic` the placement does succeed (measured 19 mm from cup center against a 50 mm tolerance), but `wp.DeterministicMode.RUN_TO_RUN` only guarantees repeatability within one GPU architecture, so that margin is not established across the CI fleet. The example test does now run with determinism enabled, which removes the run-to-run variation that made it flaky.

## New feature / API change

Deterministic hydroelastic contacts, enabled through the existing `CollisionPipeline` flag:
```python
import newton

# Shapes configured with ShapeConfig(is_hydroelastic=True, kh=1e9)
    model,
    sdf_hydroelastic_config=HydroelasticSDF.Config(reduce_contacts=True),
    deterministic=True,
)

# Repeated calls on identical state now produce bit-identical contacts.
first = contacts.rigid_contact_point0.numpy().copy()
assert (contacts.rigid_contact_point0.numpy() == first).all()
```
For full end-to-end reproducibility the consuming solver must be deterministic as well:
```python
    model,
    use_mujoco_contacts=False,
    deterministic=wp.DeterministicMode.RUN_TO_RUN,
    disable_sensors=True,  # new: skips MuJoCo sensor computation
)
```
The Panda hydroelastic example takes a matching opt-in flag (default off):
```
python -m newton.examples.robot.example_robot_panda_hydro --deterministic
```

## Summary by CodeRabbit

* **New Features**
  * Added `disable_sensors` to the MuJoCo solver integration.
  * Added deterministic hydroelastic contact generation, ordering, allocation, and reduction for improved run-to-run repeatability.
  * Added separate CLI options for deterministic contact generation vs deterministic solver behavior in hydroelastic examples.

* **Bug Fixes**
  * Improved deterministic collision pipelines with unique hydroelastic sorting keys and overflow-safe fixed-point accumulation.
  * Fixed reducer/export dedup behavior during clears.

* **Documentation**
  * Updated deterministic contact ordering documentation to fully cover hydroelastic determinism.

* **Tests**
  * Added deterministic, bit-identical hydroelastic test coverage (reduced/unreduced) plus fixed-point extreme-exponent regression tests.


mzamoramora-nvidia:
@nvtw thanks for pushing this. This is a nice improvement over `main`. I left a few comments below.

newton/_src/geometry/contact_reduction_hydroelastic.py
```diff
@@ -902,6 +1361,7 @@ def export_hydroelastic_reduced_contacts_kernel(
                 contact_data.gap_sum = gap_sum
                 contact_data.contact_stiffness = c_stiffness
                 contact_data.contact_friction_scale = wp.float32(c_friction_scale)
+                contact_data.sort_sub_key = contact_fingerprints[contact_id]
```
mzamoramora-nvidia: Could we also make the exported sort keys unique? A contact can win in both a normal-bin and voxel-bin entry, but deduplication here only covers one entry. Both copies then use the same shape pair and fingerprint, so the stable sort preserves their `atomic_add` order.

Is the duplicate export intentional? If not, the generic SDF exporter’s cross-entry flags look reusable, provided they are cleared each frame. If both copies are needed, could the key include a source tag? A key-uniqueness test would also make the contract clear.

newton/tests/test_hydroelastic.py
```diff
@@ -326,6 +329,98 @@ def test_buffer_fraction_no_crash(test, device):
...
+        "rigid_contact_normal",
+        "rigid_contact_margin0",
+        "rigid_contact_margin1",
+        "rigid_contact_tids",
+        "rigid_contact_stiffness",
+        "rigid_contact_damping",
+        "rigid_contact_friction",
+    )
+
+    for _ in range(5):
```
mzamoramora-nvidia, line 388: Could this test also confirm that neither run saturated a buffer? Please check the face-contact count, `ht_insert_failures == 0`, and the final rigid-contact count against their capacities. Otherwise two identically truncated runs can still compare equal.

newton/tests/test_examples.py
```diff
@@ -638,7 +638,8 @@ class TestRobotExamples(unittest.TestCase):
     TestRobotExamples,
     name="robot.example_robot_panda_hydro",
     devices=cuda_test_devices,
-    test_options={"usd_required": True, "num-frames": 720},
+    # Deterministic contacts keep the pick-and-place check from flaking.
+    test_options={"usd_required": True, "num-frames": 720, "deterministic": True},
```
mzamoramora-nvidia, line 642: Enabling determinism for Panda is useful end-to-end coverage, but this runs once and checks only the physical outcome. Could a dedicated unittest run the scene twice in fresh processes on the same GPU and compare fixed-frame state and contact digests, with non-saturation checks?

`contacts.example_nut_bolt_hydro` could be a useful second case because it also covers dense mesh contacts, the custom-pressure API, and CUDA graph capture.

newton/_src/geometry/contact_reduction_hydroelastic.py
```diff
@@ -65,6 +67,169 @@
...
+@wp.func
+def _max_abs_component(v: wp.vec3) -> float:
+    """Largest absolute component of ``v``, used to size its fixed-point grid."""
+    return wp.max(wp.max(wp.abs(v[0]), wp.abs(v[1])), wp.abs(v[2]))
+
+
+@wp.func
+def _fixed_shift(exp_ref: int, mantissa_bits: int) -> float:
+    """Power-of-two factor mapping an entry's grid onto the fixed-point range."""
+    return wp.pow(2.0, float(mantissa_bits - wp.clamp(exp_ref, -FIXED_EXP_CLAMP, FIXED_EXP_CLAMP)))
```
mzamoramora-nvidia: `_fixed_shift()` returns float32. At the minimum hash-table size, `mantissa_bits + 80` reaches 128, so the sentinel path produces `inf` before `_to_fixed()` converts it to `int64`.

Could we form the shift in float64 and explicitly handle exponents outside the int64 overflow proof? A device test for the sentinel and high finite custom pressure values would cover both ends.

newton/_src/geometry/sdf_hydroelastic.py
```diff
@@ -404,11 +435,15 @@ def __init__(
         config: HydroelasticSDF.Config | None = None,
         device: Devicelike | None = None,
         writer_func: Any = None,
+        deterministic: bool = False,
     ) -> None:
         if config is None:
             config = HydroelasticSDF.Config()
+        if deterministic and config.pre_prune_contacts:
+            config = replace(config, pre_prune_contacts=False)
```
mzamoramora-nvidia, line 449: Disabling pre-pruning can raise this capacity past the deterministic 20-bit contact-ID limit. The resulting error suggests reducing `max_triangle_pairs`, but that does not size the hydro buffer. Could this validation report the relevant hydro configuration knobs instead?

newton/_src/geometry/sdf_hydroelastic.py
```diff
@@ -553,27 +593,44 @@ def __init__(
...
                     self.config.margin_contact_area,
                     writer_func,
                     self.pressure_func,
                 )
 
-        self.grid_size = min(self.config.grid_size, self.max_num_face_contacts)
         self._host_warning_poll_interval = 120
         self._launch_counter = 0
 
+    def validate_deterministic(self, deterministic: bool) -> None:
```
mzamoramora-nvidia: Since `HydroelasticSDF` is public and its members are documented, this helper also becomes public API. It appears to be used only by `NarrowPhase`; could it be `_validate_deterministic()` unless external callers should rely on it?

newton/_src/geometry/contact_reduction_hydroelastic.py
```diff
@@ -139,20 +305,235 @@ def export_hydroelastic_contact_to_buffer(
...
+        normal = decode_oct(reducer_data.normal[contact_id])
+        key = make_contact_key(pair[0], pair[1], get_slot(normal))
+        entry_idx = hashtable_find_or_insert(key, reducer_data.ht_keys, reducer_data.ht_active_slots)
+        reducer_data.contact_nbin_entry[contact_id] = entry_idx
+        if entry_idx >= 0:
+            reducer_data.entry_k_eff[entry_idx] = _effective_stiffness(
+                shape_material_k_hydro[pair[0]], shape_material_k_hydro[pair[1]]
+            )
+        else:
+            wp.atomic_add(reducer_data.ht_insert_failures, 0, 1)
```
mzamoramora-nvidia, line 358: This pre-registration increments `ht_insert_failures`, and the reduction kernel increments it again if the same normal-bin insertion fails. Could we count each failed registration once so the warning reports an accurate value?

mzamoramora-nvidia (dismissed):
@nvtw thanks for addressing the feedback. The latest changes resolve my earlier concerns, and this is a nice improvement over main. I am happy to approve.

I have two non-blocking follow-ups:

- The tests could later compare all unreduced contact fields and run a representative hydro example twice in fresh processes. I understand that this may add significant time to CI, so a focused or scheduled test could be a reasonable follow-up.

- The fingerprint-range warning already gives users useful guidance. As a follow-up, the docs could mention the same limit so the bit-exact guarantee and runtime behavior are fully aligned.

Neither needs to hold up this PR.

mzamoramora-nvidia (approved):
@nvtw thanks for the updates. I took another pass through the current version and the earlier functional concerns look addressed: the reduced-contact sort keys are unique, the determinism tests check buffer saturation and hashtable failures, the fixed-point conversion handles extreme exponents, and the hydroelastic validation and warning messages are clearer.

I also ran the targeted deterministic hydroelastic tests locally on Blackwell, including the reduced, moment-matching, unreduced, fixed-point, force, and moment cases. All eight passed. The current CI checks are green as well.

I do not see a functional issue that needs to hold up the PR. I have a few small, non-blocking cleanup suggestions:

- Please use imperative mood in the new test docstrings, following the repository guidelines. For example, change “Reduced hydroelastic forces must still match...” to “Verify reduced hydroelastic forces still match...”, and “`_fixed_mantissa_bits` keeps...” to “Verify `_fixed_mantissa_bits` keeps...”.
- Consider putting the deterministic qualifier first in `test_reduced_vs_unreduced_contact_forces_deterministic` and `test_reduced_vs_unreduced_contact_moments_deterministic`, so the tests group under `test_deterministic_*` for autocomplete.
- `entry_idx = int(-1)` can be simplified to `entry_idx = -1`.

None of these need to block this PR from my side. This looks like a solid improvement over `main`.

PR #3662, opened 2026-07-27
Select explicit MuJoCo pairs for Newton contacts

### LINKED ISSUE #3644: Honor explicit MuJoCo pair parameters for Newton-generated contacts

Idate96:
## Description

Stacked on #3645 — merge that first. Until it merges, its commits also show up in this diff; the selection change here is b6d5e3ac. The two PRs together complete #3644: #3645 covers pair parameters, this PR covers pair selection.

With #3645, explicit MuJoCo pairs override contact parameters after Newton generates a contact. Newton can still omit the pair earlier when collision groups or pair filters reject it. I address that missing selection step here.

`SolverMuJoCo.get_newton_collision_pairs()` maps compiled MuJoCo pairs to Newton shape pairs. `CollisionPipeline(shape_pairs_included=...)` checks those pairs in Explicit, NxN, and SAP modes, excludes them from the automatic pass to avoid duplicates, and writes them first so a tight candidate buffer cannot drop them; in Explicit mode they are merged into the precomputed pair list. The mapping is resolved once when the pipeline is constructed.

Included pairs use Newton shape margins and gaps. Pair-specific detection envelopes and hydroelastic inclusion remain out of scope.

Closes #3644

## Checklist

## Test plan

- `TestMuJoCoSolverPairProperties`: 10/10 passed on CPU
- `TestContactEstimator`: 4/4 passed on CPU
- `test_shape_pairs_included_override_filters_without_duplicates`: passed on CPU for Explicit, NxN, and SAP
- `uvx pre-commit run -a`
- M445 boulder integration: 24 explicit pairs mapped across two worlds; Newton contacts were generated in both

## New feature / API change

```python
import newton

solver = newton.solvers.SolverMuJoCo(model, use_mujoco_contacts=False)
collision_pipeline = newton.CollisionPipeline(
    model,
    shape_pairs_included=solver.get_newton_collision_pairs(),
)
```

## Summary by CodeRabbit

* **New Features**
  * Added support for explicitly including selected shape pairs in collision detection across broad-phase modes.
  * Added access to MuJoCo’s mapped collision pairs for use with Newton collision pipelines.
  * MuJoCo explicit contact pairs now apply their configured friction, stiffness, damping, margin, and gap settings.

* **Bug Fixes**
  * Fixed MuJoCo contact handling so explicit pair settings are honored when using Newton-generated contacts.
  * Improved filtering, duplicate prevention, and contact capacity handling for explicitly included pairs.

* **Documentation**
  * Clarified explicit pair inclusion, MuJoCo parameter overrides, filtering behavior, and collision pipeline configuration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,7 @@
 
 ### Fixed
 
+- Honor explicit MuJoCo contact-pair parameters when `SolverMuJoCo` consumes contacts from Newton's collision pipeline.
 - Fix USD capsule, cylinder, and cone visual and site scaling to follow the authored primitive axis.
 - Fix USD plane visual width and length to scale along the axes defined by the `UsdGeomPlane` schema, and orient X- and Y-axis plane visuals along the authored axis.
 - Validate `ArticulationView` mask shapes and devices before launching selection kernels. (#3448)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 - Add `ViewerBase.camera_speed` to configure keyboard translation speed for interactive viewers. (#3439)
 - Add opt-in DVI forward dynamics to `SolverKamino` through `SolverKamino.Config(dynamics_solver="dvi")`, with sparse and dense execution, DVI-specific diagnostics, and warm-starting. PADMM remains the default.
 - Add SDF contact support for convex-hull shapes with mesh-attached SDFs and opt-in SDF contact generation for box shapes.
+- Add `CollisionPipeline(shape_pairs_included=...)` and `SolverMuJoCo.get_newton_collision_pairs()` for selecting explicit MuJoCo pairs with Newton-generated contacts.
 - Add opt-in filtering of static-static, static-kinematic, and kinematic-kinematic contacts during broad-phase collision detection. Set `CollisionPipeline(include_static_kinematic_pairs=False)` to enable filtering; the default preserves existing contact generation. `Model.shape_contact_pairs` remains an unfiltered superset for direct consumers such as `SolverKamino` and hydroelastic SDF setup.
 - Add opt-in `body_frame_origin="com"` to `ModelBuilder.add_rod()` and `ModelBuilder.add_rod_graph()` for COM-centered cable capsule body frames.
 - Add `sign_method` argument to `Mesh.build_sdf` and `SDF.create_from_mesh` support for a `"normal"` (angle-weighted pseudo-normal) sign strategy, for selecting the inside/outside sign of the baked SDF (`"auto"`, `"parity"`, `"winding"`, or `"normal"`).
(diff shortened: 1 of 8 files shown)


Idate96:
Quick update from downstream: I ran this PR together with #3645 on the boulder scoop up task (the workload behind #3644). `get_newton_collision_pairs()` picked up all 12 explicit pairs (11 box–mesh bucket–rock, 1 plane–mesh rock–support), and `shape_pairs_included` activated them across worlds. No candidate overflow, no non-finite values through a full rest/carry/hold/press testss.

Throughput on a matched 1,024-worlds (one RTX 4090, environment only): 1,662.3 transitions/s with MuJoCo contacts vs 1,731.1 with Newton contacts (+4.1%). Single run.claim.

PR #3118, opened 2026-06-09
Add multi-world soft contact filtering

AnkaChan:
## Description

Adds multi-world filtering for soft contact paths:
- Rigid-soft contact now launches over a prebuilt `(particle, shape)` pair buffer instead of dense particle x shape work.
- Soft-soft VBD collision BVHs now use world/global group roots so queries stay same-world plus global.
- Captured examples preallocate/reuse contacts where needed.

## Checklist

- [x] New or existing tests cover these changes
- [x] The documentation is up to date with these changes
- [ ] `CHANGELOG.md` has been updated (if user-facing change)

## Test plan

```
uvx pre-commit run -a
uv run --extra dev -m newton.tests -k test_collision_pipeline
uv run --extra dev -m unittest newton.tests.test_collision_cloth
uv run --extra dev -m unittest newton.tests.test_examples.TestExampleCaptureAudit.test_captured_collision_examples_preallocate_contacts
```

Smoke-tested examples with null viewer:

```
uv run --extra dev -m newton.examples pyramid --device cuda:0 --test --quiet --viewer null --num-frames 120 --num-pyramids 3 --pyramid-size 5
uv run --extra dev -m newton.examples diffsim_soft_body --device cuda:0 --test --quiet --viewer null --num-frames 240
uv run --extra dev -m newton.examples cloth_hanging --device cuda:0 --test --quiet --viewer null --num-frames 100
uv run --extra dev -m newton.examples softbody_hanging --device cuda:0 --test --quiet --viewer null --num-frames 120
uv run --extra dev -m newton.examples cable_twist --device cuda:0 --test --quiet --viewer null --num-frames 120
```

## Bug fix

Fixes cross-world soft contact overwork/candidates by constructing/querying only compatible same-world plus global contacts.

## Summary by CodeRabbit

* **New Features**
  * Multi-world collision handling improved via per-primitive grouped BVH queries for triangle and edge detection.
  * Soft-contact generation now uses a precomputed, world-compatible particle–shape candidate list with an updated default capacity source.

* **Bug Fixes**
  * Prevented out-of-range indexing during counter increment/replay.
  * Reduced invalid cross-world collision candidates for both soft contacts and discrete collision detection.

* **Tests**
  * Expanded multi-world collision validation, rebuild consistency checks, and CUDA graph capturability coverage.
  
* **Documentation**
  * Updated notes for the new soft-contact capacity behavior and `soft_rigid_contact_pair_count`.

PR #3664, opened 2026-07-27
Fix SolverVBD over-allocating soft-contact buffers in multi-world models

### LINKED ISSUE #3660: [BUG] SolverVBD over-allocates body-particle contact buffers in multi-world models
### Bug Description

Constructing `SolverVBD` on a replicated multi-world model allocates body-particle contact buffers sized by the *cross-world* product of the global shape and particle counts. The allocation grows quadratically with world count while the contacts that can actually occur grow linearly, so the over-allocation factor grows linearly. Once the product reaches `2**31` the solver cannot be constructed at all. Single-world models are unaffected (factor is exactly 1).

Expected: the buffers are sized from the capacity the collision pipeline can actually produce, as they already are for rigid contacts.

`_init_rigid_system` ([solver_vbd.py#L826-L827](https://github.com/newton-physics/newton/blob/402d8eac02c42243a5c9ae7e4b8f223ca8bf0394/newton/_src/solvers/vbd/solver_vbd.py#L826-L827)):

```python
if model.particle_count > 0 and model.shape_count > 0:
    self._init_body_particle_contact_state(model.shape_count * model.particle_count)
```

`_init_body_particle_contact_state` ([L1167-L1172](https://github.com/newton-physics/newton/blob/402d8eac02c42243a5c9ae7e4b8f223ca8bf0394/newton/_src/solvers/vbd/solver_vbd.py#L1167-L1172)) allocates four `wp.zeros(n, dtype=float)` arrays, so 16 bytes per slot.

**This is a stale invariant, not an arbitrary constant.** When it was written it exactly matched the pipeline default. #3118 changed that default and did not update `solver_vbd.py` — CHANGELOG.md:185:

> Change `CollisionPipeline.soft_contact_max` to default to the precomputed `soft_rigid_contact_pair_count` instead of `shape_count * particle_count`; pass `soft_contact_max=model.shape_count * model.particle_count` to preserve the previous capacity. (#3118)

The pipeline now sizes from world-compatible pairs only ([collide.py#L1131-L1136](https://github.com/newton-physics/newton/blob/402d8eac02c42243a5c9ae7e4b8f223ca8bf0394/newton/_src/sim/collide.py#L1131-L1136); `_world_compatible_pairs` at [L480-L546](https://github.com/newton-physics/newton/blob/402d8eac02c42243a5c9ae7e4b8f223ca8bf0394/newton/_src/sim/collide.py#L480-L546), added later by #3262 — which did edit `solver_vbd.py`, still leaving L826-827 stale).

For `W` worlds each holding `S` world-local shapes and an equal share of particles, plus `G` global (`world == -1`) shapes and no global particles, the over-allocation factor is `(W*S + G) / (S + G)`: linear in `W`, and exactly `W` when `G = 0`, as in the repro below.

**The class docstring already describes the intended behaviour** ([L135-L138](https://github.com/newton-physics/newton/blob/402d8eac02c42243a5c9ae7e4b8f223ca8bf0394/newton/_src/solvers/vbd/solver_vbd.py#L135-L138)): *"SolverVBD pre-allocates contact state from capacities populated by `CollisionPipeline` when available; otherwise, the first `step` lazily sizes buffers from `Contacts`."* The rigid path has such a capacity, and `collide.py:908-910` writes it back specifically to keep the two in sync:

```python
# Keep model-level default in sync with the resolved pipeline capacity.
# This avoids divergence between model- and contacts-based users (e.g. VBD init).
model.rigid_contact_max = rigid_contact_max
```

There is no `model.soft_contact_max` anywhere, so the soft path always takes the worst-case branch. For the same reason, constructing the `CollisionPipeline` before the solver does not help the soft path (it does help the rigid one, via the write-back above) — L827 is computed from `model` alone.

**There is no supported way to cap it.** `SolverVBD.__init__` ([L222-L269](https://github.com/newton-physics/newton/blob/402d8eac02c42243a5c9ae7e4b8f223ca8bf0394/newton/_src/solvers/vbd/solver_vbd.py#L222-L269)) exposes no soft-contact capacity parameter; `rigid_body_particle_contact_buffer_size` is a per-body list stride, not this global capacity.

**Failure modes.** Below `2**31` slots the allocation succeeds and silently wastes VRAM. For 64 worlds each holding 64 shapes and a 10x10x10-particle soft grid — 4,096 shapes x 64,000 particles = 262,144,000 slots — `SolverVBD` reserves 4.19 GB on an RTX 5090 (measured free-memory delta 4.23 GB) where the pipeline's capacity is 4,096,000 slots = 65.5 MB. At or above `2**31` slots, construction fails outright in Warp's per-dimension int32 shape check (`check_array_shape`, `warp/_src/types.py:2976` in warp 1.17.0.dev20260725, called from `array._init_new` before any device allocation, so it reproduces on CPU):

```
ValueError: Array shapes must not exceed the maximum representable value of a signed 32-bit
integer, got 2166784000 in dimension 0.
```

Output of the reproduction script below (a few seconds on CPU, no GPU required; Warp's init banner and `Module ... load` lines elided):

```
newton 1.5.0.dev0, warp 1.17.0.dev20260725

 worlds  shapes  particles   VBD allocates  pipeline needs  ratio    VBD bytes
      1       8         27             216             216    1.0        3,456
      2      16         54             864             432    2.0       13,824
      4      32        108           3,456             864    4.0       55,296
      8      64        216          13,824           1,728    8.0      221,184
     16     128        432          55,296           3,456   16.0      884,736
     32     256        864         221,184           6,912   32.0    3,538,944
     64     512       1728         884,736          13,824   64.0   14,155,776

worlds=184  shapes=11,776  particles=184,000
SolverVBD wants 2,166,784,000 slots = 34.67 GB; pipeline capacity is 11,776,000 = 188.4 MB
ValueError: Array shapes must not exceed the maximum representable value of a signed 32-bit integer, got 2166784000 in dimension 0.
```

That model has no global shapes, so the `ratio` column equals `world_count` exactly at every row.

**Suggested fix.** Size the preallocation from the pipeline's capacity rather than the dense product. Shrinking looks safe: every index into the four buffers is a flat contact id produced under a launch dimensioned by `contacts.soft_contact_max`. The producer `build_body_particle_contact_lists` ([rigid_vbd_kernels.py:2842](https://github.com/newton-physics/newton/blob/402d8eac02c42243a5c9ae7e4b8f223ca8bf0394/newton/_src/solvers/vbd/rigid_vbd_kernels.py#L2842)) stores `tid` from a launch of that dim ([solver_vbd.py#L2490](https://github.com/newton-physics/newton/blob/402d8eac02c42243a5c9ae7e4b8f223ca8bf0394/newton/_src/solvers/vbd/solver_vbd.py#L2490)), and every consumer reads either `wp.tid()` under the same dim (`rigid_vbd_kernels.py:3340`/`:4948`, `particle_vbd_kernels.py:1820`/`:2324`, `vbd_coupling_kernels.py:154`/`:300`) or one of those stored ids (`rigid_vbd_kernels.py:3972`, reached from the per-body launch at `solver_vbd.py:2796-2798`). Nothing indexes them in the `shape_count * particle_count` space. `_initialize_rigid_bodies` also already grows them on demand ([L2473-L2481](https://github.com/newton-physics/newton/blob/402d8eac02c42243a5c9ae7e4b8f223ca8bf0394/newton/_src/solvers/vbd/solver_vbd.py#L2473-L2481), [L2512-L2520](https://github.com/newton-physics/newton/blob/402d8eac02c42243a5c9ae7e4b8f223ca8bf0394/newton/_src/solvers/vbd/solver_vbd.py#L2512-L2520)).

Two things to watch if the eager path is removed entirely rather than resized: the proxy-coupling harvest hooks at [L1016](https://github.com/newton-physics/newton/blob/402d8eac02c42243a5c9ae7e4b8f223ca8bf0394/newton/_src/solvers/vbd/solver_vbd.py#L1016) and [L1077](https://github.com/newton-physics/newton/blob/402d8eac02c42243a5c9ae7e4b8f223ca8bf0394/newton/_src/solvers/vbd/solver_vbd.py#L1077) gate, among other conditions, on `shape[0] >= contacts.soft_contact_max` and silently skip the launch when short; and `_raise_if_capturing_resize` ([L1186-L1194](https://github.com/newton-physics/newton/blob/402d8eac02c42243a5c9ae7e4b8f223ca8bf0394/newton/_src/solvers/vbd/solver_vbd.py#L1186-L1194)) turns a needed regrow during graph capture into a `RuntimeError`. Sizing from a real capacity avoids both.

One caveat on direction: with the default `enable_rigid_soft_full_surface_contact=False` the dense product is always an over-estimate, but `collide.py:1134` adds edge/face pair headroom on top, so with that flag enabled at low world counts the pipeline capacity can exceed `shape_count * particle_count` and L827 becomes an *under*-estimate. Measured on the repro model with the flag on, the crossover is between 4 and 8 worlds (1 world: 1,176 vs 216; 8 worlds: 9,408 vs 13,824). Sizing from the pipeline fixes both directions.

**Possibly related.** `newton/_src/solvers/style3d/collision/collision.py:42` has the identical expression, but feeds a `wp.launch` dim ([L203](https://github.com/newton-physics/newton/blob/402d8eac02c42243a5c9ae7e4b8f223ca8bf0394/newton/_src/solvers/style3d/collision/collision.py#L203)) rather than an allocation size, so the same world scaling applies there with a different (untested by me) failure mode.

**Downstream.** Found in Isaac Lab, which maps one env to one Newton world. For a soft-body Franka manipulation task, `SolverVBD` construction starts failing with the `ValueError` above somewhere around 700-770 envs (the exact threshold varies because the soft body's tetrahedralization is not deterministic across runs).

### Reproduction Script

```python
"""SolverVBD sizes its body-particle contact buffers with the dense cross-world product
model.shape_count * model.particle_count, ignoring model.world_count.
"""

import warp as wp

import newton
import newton.solvers


def build(world_count: int, shapes_per_world: int, soft_dim: int, device: str) -> newton.Model:
    """world_count copies of {shapes_per_world boxes + a soft_dim^3 grid}, one per world."""
    tpl = newton.ModelBuilder()
    for j in range(shapes_per_world):
        tpl.add_shape_box(
            -1,
            xform=wp.transform(wp.vec3(0.25 * j, 0.0, 0.5), wp.quat_identity()),
            hx=0.05,
            hy=0.05,
            hz=0.05,
        )
    tpl.add_soft_grid(
        pos=wp.vec3(0.0, 0.0, 1.0),
        rot=wp.quat_identity(),
        vel=wp.vec3(0.0, 0.0, 0.0),
        dim_x=soft_dim,
        dim_y=soft_dim,
        dim_z=soft_dim,
        cell_x=0.05,
        cell_y=0.05,
        cell_z=0.05,
        density=100.0,
        k_mu=1000.0,
        k_lambda=1000.0,
        k_damp=0.0,
    )
    tpl.color()  # SolverVBD requires particle coloring
    builder.replicate(tpl, world_count=world_count)
    return builder.finalize(device=device)


print(f"newton {newton.__version__}, warp {wp.__version__}\n")
print(f"{'worlds':>7}{'shapes':>8}{'particles':>11}{'VBD allocates':>16}{'pipeline needs':>16}{'ratio':>7}{'VBD bytes':>13}")
for n in (1, 2, 4, 8, 16, 32, 64):
    m = build(n, 8, 2, "cpu")
    vbd = m.shape_count * m.particle_count
    cp = newton.CollisionPipeline(m, soft_contact_max=None).soft_contact_max
    print(f"{m.world_count:>7}{m.shape_count:>8}{m.particle_count:>11}{vbd:>16,}{cp:>16,}{vbd / cp:>7.1f}{16 * vbd:>13,}")

# Same model shape, scaled until shape_count * particle_count crosses 2**31.
m = build(184, 64, 9, "cpu")
vbd = m.shape_count * m.particle_count
cp = newton.CollisionPipeline(m, soft_contact_max=None).soft_contact_max
print(f"\nworlds={m.world_count}  shapes={m.shape_count:,}  particles={m.particle_count:,}")
print(f"SolverVBD wants {vbd:,} slots = {16 * vbd / 1e9:.2f} GB; pipeline capacity is {cp:,} = {16 * cp / 1e6:.1f} MB")
try:
    newton.solvers.SolverVBD(m)
except ValueError as e:
    print(f"{type(e).__name__}: {e}")

# The 4.19 GB VRAM measurement quoted above:
#   m = build(64, 64, 9, "cuda:0")   ->  4,096 shapes x 64,000 particles = 262,144,000 slots
```

### System Information

- Newton: 1.5.0.dev0 (commit `402d8eac02c42243a5c9ae7e4b8f223ca8bf0394`, `main`)
- Warp: 1.17.0.dev20260725
- Python: 3.11.15
- OS: Ubuntu 24.04.4 LTS, Linux 6.17.0-35-generic
- GPU: NVIDIA GeForce RTX 5090 (31 GiB, sm_120), CUDA Toolkit 12.9, CUDA driver 13.2

The `ValueError` path is device-independent and reproduces on CPU; the 4.19 GB measurement needs a GPU.

mmichelis:
## Description

Closes #3660.

### The bug

`SolverVBD` sized its four body-particle contact buffers with `model.shape_count * model.particle_count`. Both are global counts, so on a replicated model the product grows quadratically with world count while the contacts that can actually occur grow linearly. At 16 bytes per slot the waste is large well before it is fatal, and past `2**31` slots Warp rejects the array shape outright and the solver cannot be constructed at all.

This was a stale invariant, not an arbitrary constant: the expression matched `CollisionPipeline`'s default until #3118 switched that default to the world-aware `soft_rigid_contact_pair_count` without updating the solver.

### The fix

`SolverVBD` now pre-sizes from `_count_soft_particle_rigid_contact_pairs`, which derives the count arithmetically from `Model.particle_world_start` and `Model.shape_world_start` in `O(world_count)`, without materializing the pair array or reading per-entity world ids.

Sizing remains a pre-size hint rather than a bound, exactly as before: every index into these buffers is a flat contact id produced under a launch dimensioned by `contacts.soft_contact_max`, and `_initialize_rigid_bodies` still grows them lazily. With `enable_rigid_soft_full_surface_contact=True` the pre-size excludes the edge/face headroom only `CollisionPipeline` knows about, and that lazy path covers the difference — as it already does today, since the dense product also under-sizes that configuration below five worlds.

Single-world models are unaffected: with one world the count equals the dense product exactly.

### Why a dedicated counter rather than reusing the builder

`CollisionPipeline` builds the pair array it needs anyway and gets the count free as `len(...)`. The solver cannot use that, because in the failing case it is constructed *before* any pipeline exists. Calling `_build_soft_particle_rigid_contact_pairs` from the solver instead works, but on a 4096-world model of 58 shapes and 66 particles each (15,949,824 pairs) it costs 4788 ms and 1026 MB of peak host memory to produce a single integer that is discarded immediately, after which the pipeline rebuilds the identical array. Skipping the packing does not rescue it: the index-array expansion alone is 2905 ms and 643 MB.

Counting from the per-world offsets avoids both the expansion and the per-entity device-to-host transfer. On 4096 worlds with 1,048,576 particles and 32,769 shapes it reads `2 * (world_count + 2)` = 8,196 entries rather than 1,081,345, taking 0.018 ms against 2.77 ms for a per-entity pass.

### Known limitation: full-surface contact under graph capture

With `enable_rigid_soft_full_surface_contact=True` the pipeline sizes `soft_contact_max` as the particle-shape pair count *plus* edge/face headroom. The pre-size deliberately omits that headroom, so it is short under that flag and `_initialize_rigid_bodies` grows the buffers on the first step. That is fine everywhere except one case: under CUDA graph capture on a device with no stream-ordered memory pool, growth raises instead.

`main` is short there too, at every world count below five — its dense product only covers the headroom above that by coincidence, since it grows as world_count squared while the requirement grows linearly. Measured on a cloth grid plus sphere per world:

| worlds | main pre-size | this PR | full-surface needs | main covers | PR covers |
|---|---|---|---|---|---|
| 1 | 16 | 16 | 67 | no | no |
| 4 | 256 | 64 | 268 | no | no |
| 8 | 1,024 | 128 | 536 | **yes** | no |
| 32 | 16,384 | 512 | 2,144 | **yes** | no |

So this removes an accidental over-allocation rather than a designed guarantee, and narrows an already-broken window from "below five worlds" to "all world counts". It fails loudly with a `RuntimeError`, never silently or with wrong physics, and needs all of: the non-default full-surface flag, graph capture, mempools disabled, and no uncaptured warm-up step.

Closing it properly would mean either publishing the pipeline's resolved capacity back to the model, or adding the headroom unconditionally — which would cost every default-path user roughly 4x, since the solver cannot see the flag. Neither seemed proportionate; the resize error message now names the remedies that actually work.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_soft_contact_presize_is_world_aware
```

`test_soft_contact_presize_is_world_aware` builds 1 and 4 worlds, with and without a global ground plane, constructs `SolverVBD` with no pipeline in existence, and asserts the buffer length equals `CollisionPipeline(model).soft_contact_max` and scales linearly. Fails `AssertionError: 128 != 32` without the fix.

Full files: `test_solver_vbd` 80 passed; `test_collision_pipeline` 163 passed bar one CUDA graph-capture error in a rigid mesh narrow-phase test, which reproduces on an unmodified tree and picks a different test name each run.

## Bug fix

**Steps to reproduce.** Build a replicated multi-world model with particles and rigid shapes, then construct `SolverVBD`. The allocation is `world_count` times the capacity the collision pipeline can produce, and past `2**31` slots construction raises.

**Minimal reproduction.**

```python
import warp as wp

import newton
import newton.solvers

tpl = newton.ModelBuilder()
for j in range(64):
    tpl.add_shape_box(-1, xform=wp.transform(wp.vec3(0.25 * j, 0.0, 0.5), wp.quat_identity()), hx=0.05, hy=0.05, hz=0.05)
tpl.add_soft_grid(
    pos=wp.vec3(0.0, 0.0, 1.0), rot=wp.quat_identity(), vel=wp.vec3(0.0, 0.0, 0.0),
    dim_x=9, dim_y=9, dim_z=9, cell_x=0.05, cell_y=0.05, cell_z=0.05,
    density=100.0, k_mu=1000.0, k_lambda=1000.0, k_damp=0.0,
)
tpl.color()

builder.replicate(tpl, world_count=184)

# Before: ValueError, array shape 2,166,784,000 exceeds the signed 32-bit limit.
# After:  constructs, sized to the 11,776,000 slots the pipeline can actually produce.
print(len(solver.body_particle_contact_penalty_k))
```

`newton/_src/solvers/style3d/collision/collision.py:42` carries the identical expression, but as a `wp.launch` dimension rather than an allocation size. Left alone here; happy to file it separately.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed `SolverVBD` initialization failures in large multi-world scenes containing particles and rigid shapes.
  - Improved body–particle soft-contact buffer sizing to be world-aware and avoid over-allocation from global particle/shape counts.

- **Documentation**
  - Clarified soft-contact buffer sizing semantics for `soft_contact_max` and `soft_rigid_contact_pair_count`, including how full-surface contact affects edge/face headroom.

- **Tests**
  - Added regression tests to verify soft particle–rigid pair counting matches built candidate pairs and that `SolverVBD` pre-sizing scales correctly across worlds.


newton/_src/sim/collide.py
```diff
@@ -562,6 +587,15 @@ def _build_soft_particle_rigid_contact_pairs(model: Model) -> wp.array[wp.vec2i]
     return _world_compatible_pairs(model.particle_world.numpy(), model.shape_world.numpy(), world_count, model.device)
 
 
+def _count_soft_particle_rigid_contact_pairs(model: Model) -> int:
+    """Count the pairs :func:`_build_soft_particle_rigid_contact_pairs` would emit for ``model``.
+
+    Lets solvers pre-size soft-contact buffers without building a :class:`CollisionPipeline`. Excludes
+    the full-surface edge/face headroom, which only the pipeline knows about.
+    """
+    return _count_world_compatible_pairs(model.particle_world.numpy(), model.shape_world.numpy(), model.world_count)
```
eric-heiden: 🔵 Please use `particle_world_start` and `shape_world_start` for this unfiltered count. The current path downloads every particle and shape world ID and allocates several entity-sized NumPy masks, while adjacent start differences plus the front/tail global counts compute the same result from only `world_count + 2` entries; in a representative host-side case with 4,096 worlds, 1,048,576 particles, and 32,768 shapes, this reduced counting time from about 6 ms to 15 µs before accounting for device-transfer cost. Since `_count_world_compatible_pairs()` has no other production caller, specializing `_count_soft_particle_rigid_contact_pairs()` around the start arrays also removes the generic invalid-ID and `shape_ok` machinery from this path.

mmichelis: Thanks for the great catch! Fixed in b83b8877. `test_soft_contact_presize_is_world_aware` now also builds a variant with a global ground plane, so the head and tail global terms the offset arithmetic introduces are exercised on a real model rather than only the same-world term.

eric-heiden (approved):

PR #3667, opened 2026-07-27
Fix MJCF site fromto parsing

### LINKED ISSUE #3665: [BUG] MJCF importer ignores site fromto
## Bug Description

The MJCF importer ignores the `fromto` attribute on `site` elements.

In MuJoCo, `fromto` is supported for capsule, cylinder, ellipsoid, and box sites. It defines the site's midpoint, longitudinal size, and local positive Z axis. Newton currently imports such a site using its default position, orientation, and length instead.

This can affect spatial tendon paths, touch sensor zones, and other features that reference the imported site.

## Reproduction Script

```python
import tempfile
from pathlib import Path

import newton

xml = \"\"\"
<mujoco>
  <worldbody>
    <body name=\"body\">
      <site name=\"site\" type=\"cylinder\" size=\"0.1\"
            fromto=\"0 0 0 0 0 2\"/>
    </body>
  </worldbody>
</mujoco>
\"\"\"

    path = Path(directory) / \"model.xml\"
    path.write_text(xml)
    builder.add_mjcf(str(path))
    print(builder.shape_transform)
    print(builder.shape_scale)
```

Expected: the site is centered at `(0, 0, 1)` with a half-length of `1`.

Actual: the site remains at the origin and uses its default longitudinal size.

## System Information

- Newton: current `main` (`c487bc87`)
- MuJoCo reference behavior: 3.10.0
- OS: Linux

han-xudong:
## Description

Import the MJCF `fromto` attribute for capsule, cylinder, ellipsoid, and box sites. The importer now derives the site midpoint, orientation, and longitudinal size instead of silently using the default site transform.

The transform calculation is shared with capsule and cylinder geoms so both paths retain the same MuJoCo orientation convention. Invalid zero-length segments and `pos` combined with `fromto` now produce clear errors.

Closes #3665

## Checklist

## Test plan

```text
uv run --extra dev python -m unittest newton.tests.test_import_mjcf.TestSiteFromto newton.tests.test_import_mjcf.TestFromtoCapsuleOrientation newton.tests.test_import_mjcf.TestImportMjcfActuatorsFrames.test_static_fromto_capsule_xform
uvx pre-commit run -a
```

## Bug fix

**Steps to reproduce:**

1. Import an MJCF model containing a cylinder, capsule, ellipsoid, or box site with `fromto`.
2. Inspect the imported site transform and scale.
3. Observe on `main` that the site is placed at its default transform and length.

**Minimal reproduction:**

See the reproduction in #3665.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected MJCF `<site>` imports so `fromto` transforms and lengths are honored consistently, including site sizing and inherited defaults.
  * Fixed `fromto` orientation for capsules and cylinders to match midpoint and endpoint semantics.
  * Added validation for conflicting `pos` and `fromto` attributes and zero-length `fromto` definitions.

* **Tests**
  * Expanded coverage for MJCF `fromto` sites and capsule/cylinder orientation cases.

* **Documentation**
  * Updated the 1.4.0 changelog with the MJCF site import fix.


newton/_src/utils/import_mjcf.py
```diff
@@ -686,6 +686,33 @@ def parse_orientation(attrib) -> wp.quat:
...
+        if length < 1.0e-6:
+            if zero_length_error is not None:
+                raise ValueError(zero_length_error)
+            return wp.transform(position, wp.quat_identity()), 0.0
+
+        direction /= length
+        if float(direction[2]) < -0.999999:
+            rotation = wp.quat(1.0, 0.0, 0.0, 0.0)
+        else:
+            rotation = wp.quat_between_vectors(wp.vec3(0.0, 0.0, 1.0), direction)
```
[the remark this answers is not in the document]
han-xudong: Addressed in 73757b19. The shared `fromto` helper now uses `end - start`, so local +Z follows MuJoCo’s first-to-second endpoint convention for both geoms and sites. I also updated the axial and diagonal regression checks to assert the rotated +Z direction rather than a specific quaternion representation. The full MJCF importer test module passes (238 tests).

eric-heiden:
Nice cleanup factoring the `fromto` parsing into one helper so sites finally get the same treatment as geoms.

One thing to settle before this lands: the axis direction change in the second commit points imported `fromto` frames the opposite way from MuJoCo's own compiler. Details in the inline comment.
I think CodeRabbit got this wrong - but please double check it!

newton/_src/utils/import_mjcf.py
```diff
@@ -686,6 +686,33 @@ def parse_orientation(attrib) -> wp.quat:
...
```
eric-heiden: 🟠 `direction = end - start` rotates the `fromto` frame 180° away from what MuJoCo's compiler produces.

MuJoCo builds this axis the other way around. In `src/user/user_objects.cc`, both `mjCGeom::Compile()` and `mjCSite::Compile()` compute `double vec[3] = {fromto[0]-fromto[3], fromto[1]-fromto[4], fromto[2]-fromto[5]}`, which is `from - to`, and pass it to `mjuu_z2quat(quat, vec)` in `src/user/user_util.cc`, which rotates local +Z *onto* `vec`. `wp.quat_between_vectors(wp.vec3(0.0, 0.0, 1.0), direction)` has the same z-to-direction semantics, and the MuJoCo side is identical in tag 3.3.7 and on `main`. So the previous `direction = start - end` matched the compiler, and this version does not.

What that means in practice: every `<geom type="capsule|cylinder" fromto=...>`, for example throughout `newton/examples/assets/nv_humanoid.xml` and `nv_ant.xml`, now imports with a `Model.shape_transform` rotated 180° relative to `mujoco.MjModel.from_xml_string()` on the same file, and the new site path inherits the same flip for sensor and tendon attachment frames. Capsules, cylinders and ellipsoids are symmetric, so contacts are unaffected, but anything that reads these frames now disagrees with the solver backend.

Worth flagging that the MuJoCo XML reference prose does say +Z is "oriented from the first towards the second point" (`doc/XMLreference.rst`), so the upstream docs and the upstream compiler contradict each other. The importer should follow the compiler, since that is what the MuJoCo backend actually runs.

```suggestion
        direction = start - end
        length = wp.length(direction)
        if length < 1.0e-6:
            if zero_length_error is not None:
                raise ValueError(zero_length_error)
            return wp.transform(position, wp.quat_identity()), 0.0

        direction /= length
        if float(direction[2]) < -0.999999:
            rotation = wp.quat(1.0, 0.0, 0.0, 0.0)
        else:
            rotation = wp.quat_between_vectors(wp.vec3(0.0, 0.0, 1.0), direction)
```

The sign assertions in `TestFromtoCapsuleOrientation` (`newton/tests/test_import_mjcf.py`) and the CHANGELOG entry for the axis change would need to be reverted along with it. The site `fromto` fix, which is what this PR set out to do, stands on its own without the axis flip.

If the change in convention is intentional, could you confirm it against a compiled `MjModel.geom_quat` / `site_quat` and describe it in the CHANGELOG as a deliberate departure from MuJoCo rather than as a fix?

han-xudong: Verified this against both the MuJoCo compiler source and native MjModel output. Commit 1c6e4297 restores the start-to-end operand order used by the compiler (start - end), updates the geom and site axis expectations, and removes the incorrect axis claim from the changelog. The focused tests pass with the fix and fail with the previous direction.

eric-heiden (approved):

PR #3673, opened 2026-07-27
Support MJCF compiler assetdir

### LINKED ISSUE #3672: [BUG] MJCF importer ignores compiler assetdir
## Bug Description

Newton's MJCF importer ignores the `assetdir` attribute on `compiler`.

MuJoCo uses `assetdir` as the common fallback for `meshdir` and `texturedir`, with explicit values for those attributes taking precedence. Newton currently looks only at `meshdir` and `texturedir`, causing valid mesh and texture paths to be resolved from the wrong directory.

The same problem applies to compiler settings declared in included MJCF files.

## Reproduction Script

```python
import tempfile
from pathlib import Path

import newton

obj = \"\"\"
v 0 0 0
v 1 0 0
v 0 1 0
v 0 0 1
f 1 3 2
f 1 2 4
f 1 4 3
f 2 3 4
\"\"\"

xml = \"\"\"
<mujoco>
  <compiler assetdir=\"assets\"/>
  <asset>
    <mesh name=\"tetrahedron\" file=\"tetrahedron.obj\"/>
  </asset>
  <worldbody>
    <geom type=\"mesh\" mesh=\"tetrahedron\"/>
  </worldbody>
</mujoco>
\"\"\"

    root = Path(directory)
    (root / \"assets\").mkdir()
    (root / \"assets\" / \"tetrahedron.obj\").write_text(obj)
    model_path = root / \"model.xml\"
    model_path.write_text(xml)

    builder.add_mjcf(str(model_path))
```

Expected: the mesh is loaded from `assets/tetrahedron.obj`.

Actual: Newton attempts to load `tetrahedron.obj` relative to the model directory and raises a file-loading error.

## System Information

- Newton: current `main` (`c487bc87`)
- MuJoCo reference behavior: 3.10.0
- OS: Linux

han-xudong:
## Description

Honor the MJCF `compiler assetdir` attribute when resolving mesh, texture, and heightfield assets. Explicit `meshdir` and `texturedir` values continue to take precedence.

The importer applies the same directory handling to included MJCF files as its existing `meshdir` and `texturedir` support. XML-string imports preserve the default working-directory fallback, while custom path resolvers receive the `assetdir`-prefixed relative path. Heightfields now consume the path resolved during MJCF expansion instead of invoking the resolver a second time.

Closes #3672

## Checklist

## Test plan

```text
uv run --extra dev python -m unittest newton.tests.test_import_mjcf.TestMjcfIncludeMeshdir
uvx pre-commit run -a
```

## Bug fix

**Steps to reproduce:**

1. Place mesh, texture, or heightfield files in a separate asset directory.
2. Set only `compiler assetdir` in the MJCF model.
3. Import the model.
4. Observe on `main` that Newton resolves the files from the model directory instead.

**Minimal reproduction:**

See the reproduction in #3672.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected MJCF import expansion so `compiler assetdir` properly drives mesh, texture, and heightfield asset resolution, including for included MJCF files.
  * Fixed how XML-string MJCF inputs and custom path resolvers resolve assets, preserving expected fallback behavior.
  * Made `compiler strippath` apply consistently when mapping authored paths to `assetdir`.
  * Improved inertia parsing for `inertiafromgeom` / `inertiagrouprange`, ensuring mass and inertia are consistent across visual vs collider geometry loading.

### DIFF
--- a/newton/_src/utils/import_mjcf.py
+++ b/newton/_src/utils/import_mjcf.py
@@ -108,6 +108,7 @@ def _load_and_expand_mjcf(
     own_compiler = root.find("compiler")
     if own_compiler is not None:
         own_assetdir = own_compiler.attrib.get("assetdir")
+        own_strippath = own_compiler.attrib.get("strippath", "false").lower() in {"true", "1"}
         if own_assetdir is None:
             own_meshdir = own_compiler.attrib.get("meshdir", ".")
             own_texturedir = own_compiler.attrib.get("texturedir", own_meshdir)
@@ -115,8 +116,12 @@ def _load_and_expand_mjcf(
             own_meshdir = own_compiler.attrib.get("meshdir", own_assetdir)
             own_texturedir = own_compiler.attrib.get("texturedir", own_assetdir)
     else:
+        own_strippath = False
         own_meshdir = "."
         own_texturedir = "."
+
+    asset_dir_by_tag = {"mesh": own_meshdir, "hfield": own_meshdir, "texture": own_texturedir}
+    own_file_assets = [elem for elem in root.iter() if elem.tag in asset_dir_by_tag and elem.get("file")]
 
     # Strip consumed asset directories so they don't leak into the parent tree
     # and affect parent-file asset resolution. Other compiler attributes (angle, etc.)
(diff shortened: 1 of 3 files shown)


eric-heiden:
Thanks for your contribution, @han-xudong! Adding `assetdir` support closes an important MJCF compatibility gap. There are just a few comments worth addressing before we can merge.

newton/_src/utils/import_mjcf.py
```diff
@@ -545,7 +558,9 @@ def resolve_element_attrib(element, tag: str, ambient_defaults: dict | None = No
             file_attr = hfield.attrib.get("file")
             file_path = None
             if file_attr:
-                file_path = path_resolver(base_dir, file_attr)
+                file_path = file_attr
+                if not os.path.isabs(file_path):
+                    file_path = os.path.abspath(os.path.join(mjcf_dirname, file_path))
```
eric-heiden, line 576: 🟠 Absolute heightfield paths now bypass `path_resolver`. `_load_and_expand_mjcf()` only calls the resolver for non-absolute paths, so a resolver that remaps an authored absolute path into an archive, cache, or sandbox is ignored and `load_heightfield_elevation()` tries to open the original path. Please resolve each heightfield exactly once while still invoking the custom resolver for absolute inputs, and add a test where the resolver remaps an absolute heightfield path.

newton/_src/utils/import_mjcf.py
```diff
@@ -150,7 +161,9 @@ def _load_and_expand_mjcf(
         if file_attr and not os.path.isabs(file_attr):
             asset_dir = _asset_dir_tags.get(elem.tag, ".")
             resolved_path = os.path.join(asset_dir, file_attr) if asset_dir != "." else file_attr
-            if base_dir is not None or os.path.isabs(resolved_path):
+            if base_dir is None and path_resolver is _default_path_resolver:
+                elem.set("file", resolved_path)
+            else:
                 elem.set("file", path_resolver(base_dir, resolved_path))
```
eric-heiden: 🟠 This joins `assetdir` with the complete `file` value without honoring `<compiler strippath="true">`. MuJoCo strips directory components before applying the asset directory, so `assetdir="assets"` with `file="../vendor/cube.stl"` should resolve to `assets/cube.stl`, not `assets/../vendor/cube.stl`. Please capture `strippath`, apply `os.path.basename(file_attr)` before this join when enabled, and add mesh, texture, and heightfield coverage for this combination.

newton/_src/utils/import_mjcf.py
han-xudong: @eric-heiden Thank you for catching this. Asset elements are now captured before include expansion, so each file resolves its own authored paths exactly once and custom resolvers are still invoked for absolute paths. I also added a regression test that remaps an absolute heightfield path.

newton/_src/utils/import_mjcf.py
han-xudong: @eric-heiden Thank you. The importer now captures `strippath` and applies it before `assetdir`, `meshdir`, or `texturedir`. The new regression covers mesh, texture, and heightfield paths with authored directory components.

eric-heiden (approved):
Thanks!

PR #3686, opened 2026-07-27
Support inline MJCF mesh data

### LINKED ISSUE #3681: Support inline vertex and face data in MJCF mesh assets
## Description

MJCF permits mesh assets to provide inline `vertex` and `face` arrays instead of loading an external file. Newton only registers `<mesh>` assets that contain a `file` attribute, so geoms referencing valid inline meshes are silently dropped.

## Minimal reproduction

```xml
<mujoco>
  <asset>
    <mesh name="tetra"
          vertex="0 0 0  1 0 0  0 1 0  0 0 1"
          face="0 2 1  0 1 3  0 3 2  1 2 3"/>
  </asset>
  <worldbody>
    <body name="body">
      <geom type="mesh" mesh="tetra"/>
    </body>
  </worldbody>
</mujoco>
```

Native MuJoCo compiles this as one mesh geom, while Newton imports no shape.

## Expected behavior

Newton should construct a mesh from valid inline triangular vertex/face data, apply mesh and importer scale consistently, and report malformed topology with a descriptive error. Inline normals and texture coordinates can remain outside the initial scope.

han-xudong:
## Description

Import MJCF mesh assets authored with inline `vertex` and triangular `face` arrays. Inline meshes now share the existing mesh-geom and primitive-fitting paths with file-backed assets, including importer/asset scaling, material overrides, and explicit mass handling. Malformed component counts or face indices raise descriptive `ValueError`s instead of silently dropping the geom.

The initial scope intentionally excludes inline normals and texture coordinates.

Closes #3681

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_inline_mesh
uvx pre-commit run -a
```

## New feature / API change

```python
import newton

mjcf = """
<mujoco>
  <asset>
    <mesh name="tetra"
          vertex="0 0 0  1 0 0  0 1 0  0 0 1"
          face="0 2 1  0 1 3  0 3 2  1 2 3"/>
  </asset>
  <worldbody><body><geom type="mesh" mesh="tetra"/></body></worldbody>
</mujoco>
"""
builder.add_mjcf(mjcf)
assert len(builder.shape_source) == 1
```

## Summary by CodeRabbit

* **New Features**
  * MJCF mesh asset import now supports inline meshes with `vertex`/`face`, plus optional `normal`/`texcoord` and reference pose transforms (`refpos`/`refquat`), with validation and correct scaling/hull limits.
  * Inline meshes are handled consistently with existing asset-based meshes for mesh geoms.
* **Bug Fixes**
  * Improved handling of `inertiafromgeom` and `inertiagrouprange`, ensuring geometry-based inertia follows inertia-group rules and matches `&lt;inertial&gt;` expectations.
* **Tests**
  * Expanded inline-mesh test coverage for whitespace-only faces and stricter error checks for invalid `refpos`/`refquat`.
* **Documentation**
  * Updated the changelog to reflect inline mesh asset import support.

### DIFF
--- a/newton/_src/utils/import_mjcf.py
+++ b/newton/_src/utils/import_mjcf.py
@@ -17,7 +17,7 @@ from ..core import quat_between_axes
 from ..core.types import Axis, AxisType, Sequence, Transform, vec10
 from ..geometry import GeoType, Mesh, ShapeFlags, compute_inertia_shape
 from ..geometry.types import Heightfield
-from ..geometry.utils import compute_aabb, compute_inertia_box_mesh
+from ..geometry.utils import compute_aabb, compute_inertia_box_mesh, remesh_convex_hull
 from ..sim import JointTargetMode, JointType, ModelBuilder
 from ..sim.model import Model
 from ..solvers.mujoco import SolverMuJoCo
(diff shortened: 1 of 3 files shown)


eric-heiden:
Thanks, this looks good! There are just a couple things for the correct MJCF parsing we should address for completeness before merging.

newton/_src/utils/import_mjcf.py
```diff
@@ -499,19 +499,58 @@ def resolve_element_attrib(element, tag: str, ambient_defaults: dict | None = No
...
+                    raise ValueError(
+                        f"Inline MJCF mesh {name!r} vertex data must contain a multiple of 3 values; "
+                        f"got {len(vertices)}."
+                    )
+                if len(vertices) < 9:
+                    raise ValueError(f"Inline MJCF mesh {name!r} must contain at least 3 vertices.")
+                vertices = vertices.reshape(-1, 3)
+
+                if "face" not in mesh_attrib:
+                    raise ValueError(f"Inline MJCF mesh {name!r} requires face data.")
```
eric-heiden: 🟠 Requiring `face` rejects valid MJCF vertex-only meshes. MuJoCo constructs faces from the point cloud's convex hull when `face` is absent or empty, so canonical assets such as its documented vertex-only tetrahedron cannot be imported here. Please build the convex hull in that case and add a vertex-only regression test.

newton/_src/utils/import_mjcf.py
```diff
@@ -499,19 +499,58 @@ def resolve_element_attrib(element, tag: str, ambient_defaults: dict | None = No
...
+                if np.any(faces < 0) or np.any(faces >= len(vertices)):
+                    raise ValueError(f"Inline MJCF mesh {name!r} face data contains an invalid vertex index.")
+                faces = faces.reshape(-1, 3)
+
+                mesh_assets[name] = {
+                    "vertices": vertices,
+                    "faces": faces,
+                    "scale": mesh_scale,
+                    "maxhullvert": maxhullvert,
+                }
```
eric-heiden, line 610: 🟠 This asset record drops the standard `texcoord` attribute, so `load_mesh_asset()` creates a mesh without UVs and textured materials lose their authored texture. Please parse exactly two texture-coordinate values per vertex, retain them here, pass them as `uvs` to `Mesh`, and add a textured inline-mesh test.

newton/_src/utils/import_mjcf.py
```diff
@@ -566,6 +605,35 @@ def resolve_element_attrib(element, tag: str, ambient_defaults: dict | None = No
...
+        vertices = mesh_asset["vertices"] * scaling
+        faces = mesh_asset["faces"]
+        return [
+            Mesh(
+                vertices,
+                faces,
+                maxhullvert=maxhullvert,
+                color=override_color,
+                texture=override_texture,
+            )
```
eric-heiden, line 720: 🟠 Authored inline vertex normals are not passed to `Mesh`, so renderers synthesize replacements and can change the intended sharp or smooth shading. Please parse `normal`, require three values per vertex, transform and normalize the normals correctly, and pass them through `Mesh(normals=...)`.

newton/_src/utils/import_mjcf.py
```diff
@@ -566,6 +605,35 @@ def resolve_element_attrib(element, tag: str, ambient_defaults: dict | None = No
...
+        if "file" in mesh_asset:
+            return load_meshes_from_file(
+                mesh_asset["file"],
+                scale=scaling,
+                maxhullvert=maxhullvert,
+                override_color=override_color,
+                override_texture=override_texture,
+            )
+
+        vertices = mesh_asset["vertices"] * scaling
```
eric-heiden: 🟠 Scaling alone ignores the mesh asset's standard `refpos` and `refquat` attributes, which imports affected geometry at the wrong position and orientation for rendering, collision, fitting, and inertia. Please retain both attributes and apply MuJoCo's reference transform order, including subtracting `refpos` and rotating by the conjugate of `refquat`, before constructing the mesh. A non-identity transform test would cover this path.

han-xudong: @eric-heiden Fixed in 8aa2e999. Inline vertices now subtract refpos, apply the conjugate of normalized refquat, and then apply asset scaling. A non-identity reference-pose regression test covers the transform order.

newton/_src/utils/import_mjcf.py
han-xudong: @eric-heiden Fixed in 8aa2e999. Inline texcoord data is now validated as two values per vertex and passed to Mesh as UVs; the new test also verifies textured-material propagation.

newton/_src/utils/import_mjcf.py
han-xudong: @eric-heiden Fixed in 8aa2e999. Inline normals are validated as three values per vertex, transformed using the reference rotation and inverse scale, normalized, and passed through to Mesh.

newton/_src/utils/import_mjcf.py
han-xudong: @eric-heiden Fixed in 8aa2e999. Inline meshes with an absent or empty face attribute now build a convex hull, with regression coverage for both forms.

newton/_src/utils/import_mjcf.py
```diff
@@ -508,19 +508,99 @@ def resolve_element_attrib(element, tag: str, ambient_defaults: dict | None = No
...
+                        faces = np.array(mesh_attrib["face"].split(), dtype=np.int32)
+                    except ValueError as exc:
+                        raise ValueError(f"Inline MJCF mesh {name!r} has invalid face data.") from exc
+                    if len(faces) % 3 != 0:
+                        raise ValueError(
+                            f"Inline MJCF mesh {name!r} face data must contain a multiple of 3 values; got {len(faces)}."
+                        )
+                    if np.any(faces < 0) or np.any(faces >= len(vertices)):
+                        raise ValueError(f"Inline MJCF mesh {name!r} face data contains an invalid vertex index.")
+                    faces = faces.reshape(-1, 3)
```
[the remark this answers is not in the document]
han-xudong: Fixed in 31d18037. The face attribute is stripped before parsing, so absent, empty, and whitespace-only values all use the convex-hull fallback. Regression coverage includes all three forms.

newton/_src/utils/import_mjcf.py
```diff
@@ -508,19 +508,99 @@ def resolve_element_attrib(element, tag: str, ambient_defaults: dict | None = No
...
+                if refpos.shape != (3,):
+                    raise ValueError(f"Inline MJCF mesh {name!r} refpos must have 3 values.")
+                if refquat.shape != (4,):
+                    raise ValueError(f"Inline MJCF mesh {name!r} refquat must have 4 values.")
+                refquat_norm = np.linalg.norm(refquat)
+                if not np.isfinite(refquat_norm) or refquat_norm == 0.0:
+                    raise ValueError(f"Inline MJCF mesh {name!r} refquat must be finite and nonzero.")
+                if not np.all(np.isfinite(refpos)):
+                    raise ValueError(f"Inline MJCF mesh {name!r} refpos must contain only finite values.")
+                refquat /= refquat_norm
```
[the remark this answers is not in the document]
han-xudong: Fixed in 31d18037. refpos and refquat conversions now raise separate mesh-scoped errors while preserving the existing shape, finiteness, and normalization checks. Both invalid-value cases are covered.

han-xudong:
I also checked the duplicate-coordinate nitpick against MuJoCo. For a face-less point cloud, MuJoCo retains duplicate vertex attributes but the generated convex hull references only one coincident vertex, so distinct normal or texcoord values at the same position have no unambiguous face provenance. The current first-occurrence mapping matches the observed MuJoCo/QHull behavior, so I left that path unchanged rather than extending the general hull API speculatively.

eric-heiden (approved):
Thanks!

PR #3694, opened 2026-07-27
Load static USD visual shapes by default

### LINKED ISSUE #3693: Enable USD importer to parse static visuals
Visual-only shapes that are not attached to any rigid body are currently skipped by the USD importer.
We should enable the parsing of static visual shapes as well.

eric-heiden:
## Description

Import visual-only USD geometry outside rigid-body hierarchies as static Newton shapes by default. Add `load_static_visual_shapes` to `ModelBuilder.add_usd()` and `parse_usd()` as an explicit opt-out.

The importer previously loaded ordinary visual geometry only while processing rigid bodies; its standalone visual pass handled Gaussian splats only. As a result, composed static visuals such as `/Table/Visuals/TableGeom` were omitted. The new path collects supported geometry candidates during the existing deformable scout traversal, avoiding another full instance-proxy walk.

Empty-case paired benchmarks found no measurable regression:

| Scene | Disabled median | Enabled median | Difference |
| --- | ---: | ---: | ---: |
| 500 rigid visual candidates, no static visuals (12 runs) | 376.53 ms | 376.12 ms | -0.11% |
| Franka robot subtree, no static visuals (6 runs) | 2444.27 ms | 2438.04 ms | -0.26% |

## Checklist

- [x] `CHANGELOG.md` has been updated for this user-facing change

## Test plan

- Verified the default-import assertion fails against unmodified `origin/main`.
- `uv run --extra dev python -m unittest newton.tests.test_import_usd.TestImportSampleAssetsBasic.test_static_visual_shapes_loading_flag -v`
- `uv run --extra dev python -m unittest newton.tests.test_import_usd` - 280 tests passed.
- Ran Ruff lint/format checks, typos 1.48.0, the Warp array-syntax checker, and `git diff --check` directly.
- `uvx pre-commit run -a` could not start because the local Git 2.30.0 does not support the `git ls-files --deduplicate` option required by the installed pre-commit version.

## Bug fix

**Steps to reproduce:**

1. Create or open a USD stage containing visual geometry outside a rigid-body hierarchy.
3. Observe that the visual prim is absent from `path_shape_map` on `main`.

**Minimal reproduction:**

```python
import newton
from pxr import Usd, UsdGeom

UsdGeom.Xform.Define(stage, "/World")
UsdGeom.Cube.Define(stage, "/World/VisualCube")

result = newton.ModelBuilder().add_usd(stage, root_path="/World")
assert "/World/VisualCube" in result["path_shape_map"]
```

## New feature / API change

```python
import newton

builder.add_usd(stage)  # Static visual-only shapes are loaded by default.
builder.add_usd(stage, load_static_visual_shapes=False)  # Explicit opt-out.
```

## Summary by CodeRabbit

- **New Features**
  - USD imports now load supported “visual-only” geometry outside rigid-body hierarchies as static shapes by default.
  - Added `load_static_visual_shapes` to control this behavior independently of other visual loading.
- **Bug Fixes**
  - Refined visual-shape loading logic, including more precise traversal/recursion handling and candidate filtering.
- **Documentation**
  - Updated USD importer documentation and the Unreleased changelog to reflect the new option.
- **Tests**
  - Added tests covering `load_static_visual_shapes` interactions with static visuals, colliders, and dynamic visuals.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
 - Add compiled regular-expression support to label-based selectors while preserving glob strings.
 - Add simulation throughput, real-time factor, p95 step-time, steady-state GPU-memory, timestep, and MuJoCo solver-iteration metrics to the ASV robot-learning benchmarks.
 - Add `joint_dof_mask` to `newton.ik.IKSolver` to keep selected joint DOFs fixed during LM optimization. (#3488)
+- Load visual-only USD geometry outside rigid-body hierarchies as static shapes by default, with a `load_static_visual_shapes` importer opt-out.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,14 +39,13 @@
 - Add compiled regular-expression support to label-based selectors while preserving glob strings.
 - Add simulation throughput, real-time factor, p95 step-time, steady-state GPU-memory, timestep, and MuJoCo solver-iteration metrics to the ASV robot-learning benchmarks.
 - Add `joint_dof_mask` to `newton.ik.IKSolver` to keep selected joint DOFs fixed during LM optimization. (#3488)
-- Load visual-only USD geometry outside rigid-body hierarchies as static shapes by default, with a `load_static_visual_shapes` importer opt-out.
-
 ### Changed
 
 - Disable the implicit positive Dahl-friction defaults in `SolverVBD.register_custom_attributes()` (deprecated in 1.3.0): `vbd:dahl_eps_max` and `vbd:dahl_tau` now default to zero, and Dahl cable friction is enabled only where both are authored positive. Pass `dahl_defaults_enabled=True` to temporarily restore the old defaults; the compatibility mode will be removed in a future release.
 - Compile tiled camera render kernels with CUDA fast math by default for faster rendering; set `SensorTiledCamera.render_config.enable_fast_math = False` for bit-exact, IEEE-precise output.
 - Optimize raycast/raytrace queries by restructuring ray-shape intersection into local-space primitives and compile specialized depth/shadow variants that skip unused surface-normal work (mesh shadows also use any-hit queries).
 - Change experimental `SolverVBD` cable constraint slots from `[STRETCH=0, BEND=1]` to `[STRETCH=0, SHEAR=1, BEND=2, TWIST=3]`, allowing each stiffness and constraint mode to be configured independently. Existing cable calls using raw `slot=1` or `JointSlot.ANGULAR` now select shear; use `JointSlot.BEND` (now slot 2) to select bending.
+- Load visual-only USD geometry outside rigid-body hierarchies as static shapes by default; pass `load_static_visual_shapes=False` to retain the previous body-associated-visuals-only behavior.
 - Improve `SolverKamino` GPU simulation and kernel compilation performance.
 - Load solver backends lazily on first access to speed up `import newton`; access solver classes through `newton.solvers` as before, and import solver modules explicitly if module-level side effects are required.
 - Speed up `ModelBuilder.replicate()` for large world counts by merging all copies in one pass; it no longer calls `add_world()` or `add_builder()` per copy, so `ModelBuilder` subclass overrides of those methods are not invoked during replication.
--- a/newton/_src/sim/builder.py
+++ b/newton/_src/sim/builder.py
@@ -3527,11 +3527,9 @@ class ModelBuilder:
             skip_mesh_approximation: If True, mesh approximation is skipped. Otherwise, meshes are approximated according to the ``physics:approximation`` attribute defined on the UsdPhysicsMeshCollisionAPI (if it is defined), using the settings from :attr:`~newton.ModelBuilder.default_mesh_approximation_cfg`. Default is False.
             load_sites: If True, sites (prims with ``NewtonSiteAPI`` or ``MjcSiteAPI``) are loaded as non-colliding reference points. If False, sites are ignored. Default is True.
             load_visual_shapes: If True, non-physics visual geometry is loaded. If False, visual-only shapes are ignored (sites are still controlled by ``load_sites``). Default is True.
-            load_static_visual_shapes: If True, visual-only geometry outside rigid-body
-                hierarchies is loaded as static shapes when ``load_visual_shapes`` is
-                also True. Static Gaussian splats retain their existing
-                ``load_visual_shapes`` behavior independently of this option. Default is
-                True.
+            load_static_visual_shapes: If True, supported visual-only geometry outside
+                rigid-body hierarchies is loaded as static shapes when
+                ``load_visual_shapes`` is also True. Default is True.
             hide_collision_shapes: If True, collision shapes on bodies that already
                 have visual-only geometry are hidden unconditionally, regardless of
                 whether the collider has authored PBR material data. Default is False.
--- a/newton/_src/sim/builder.py
+++ b/newton/_src/sim/builder.py
@@ -3420,6 +3420,7 @@ class ModelBuilder:
         skip_mesh_approximation: bool = False,
         load_sites: bool = True,
         load_visual_shapes: bool = True,
+        load_static_visual_shapes: bool = True,
         hide_collision_shapes: bool = False,
         force_show_colliders: bool = False,
         parse_mujoco_options: bool = True,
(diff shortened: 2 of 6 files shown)


newton/_src/sim/builder.py
```diff
@@ -3526,6 +3527,11 @@ def add_usd(
             skip_mesh_approximation: If True, mesh approximation is skipped. Otherwise, meshes are approximated according to the ``physics:approximation`` attribute defined on the UsdPhysicsMeshCollisionAPI (if it is defined), using the settings from :attr:`~newton.ModelBuilder.default_mesh_approximation_cfg`. Default is False.
             load_sites: If True, sites (prims with ``NewtonSiteAPI`` or ``MjcSiteAPI``) are loaded as non-colliding reference points. If False, sites are ignored. Default is True.
             load_visual_shapes: If True, non-physics visual geometry is loaded. If False, visual-only shapes are ignored (sites are still controlled by ``load_sites``). Default is True.
+            load_static_visual_shapes: If True, visual-only geometry outside rigid-body
+                hierarchies is loaded as static shapes when ``load_visual_shapes`` is
+                also True. Static Gaussian splats retain their existing
+                ``load_visual_shapes`` behavior independently of this option. Default is
```
andrewkaufman: This seems odd to me. I think splats should be driven by `load_static_visual_shapes` too

newton/_src/utils/import_usd.py
```diff
@@ -3254,6 +3275,31 @@ def _collect_filtered_pairs(prim):
...
+            path = str(prim.GetPath())
+            is_gaussian = str(prim.GetTypeName()) == "ParticleField3DGaussianSplat"
+            if (
+                (not load_static_visual_shapes and not is_gaussian)
+                or path in deformable_visual_exclude_paths
+                or path in path_shape_map
+                or _is_in_rigid_body_hierarchy(path)
+            ):
+                continue
+            _load_visual_shapes_impl(-1, prim, recurse=False)
```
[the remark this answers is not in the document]
andrewkaufman: But a disabled collider _is a visual_. It is _not a collider_. I think its correct to include here rather than in a listing of colliders

newton/_src/utils/import_usd_deformable_utils.py
```diff
@@ -862,8 +865,26 @@ def has_candidates(self) -> bool:
     }
 )
 
+_SCOUT_VISUAL_TYPE_NAMES = frozenset(
```
andrewkaufman: Why are we doing this rather than checking `UsdGeom.Imageable(prim).IsValid()`? I guess we want to only support this subset of visuals?

newton/_src/utils/import_usd.py
```diff
@@ -3834,49 +3880,6 @@ def _collect_filtered_pairs(prim):
...
-                xform=wp.transform(g_pos, g_rot),
-                scale=g_scale,
-                cfg=splat_cfg,
-                color=splat_material_props.get("color"),
-                label=gaussian_path,
-            )
-            path_shape_map[gaussian_path] = shape_id
-            path_shape_scale[gaussian_path] = g_scale
-            if verbose:
-                print(f"Added Gaussian splat shape {gaussian_path} with id {shape_id}.")
```
andrewkaufman, line 3878: I don't see this splat specific handling retained anywhere in the diff of this PR. Is it already handled the same way in `_load_visual_shapes_impl`?

newton/_src/utils/import_usd_deformable_utils.py
```diff
@@ -812,6 +812,9 @@ class _DeformablePrimBuckets:
     tetmeshes: list[Usd.Prim] = field(default_factory=list)
     attachments: list[Usd.Prim] = field(default_factory=list)
     element_filters: list[Usd.Prim] = field(default_factory=list)
+    # Optional visual candidates collected for parse_usd's static visual pass. Reusing
+    # this scout avoids a second full instance-proxy traversal of the stage.
+    static_visuals: list[Usd.Prim] = field(default_factory=list)
```
andrewkaufman: I understand the rationalle to avoid many traversals, but it is a bit odd to handle "everything non physical" inside the "special handling of deformables" codepath. We do also already have multiple traversals -- the `LoadPhysicsFromRange` call plus this extra call. I wonder if there is some way to just do one main traversal in the main `import_usd` and hand off sub-hierarchies to each system

```
for prim in root_range:
  if rigid body or joint or articulation root or actuator:
     _parse_rigid(prim)
  elif deformable:
     _parse_deformable(prim)
  elif physics scene:
    _parse_scene(prim)
  elif splat:
    _parse_splat(prim)
  else:
    _parse_static_visuals(prim)
```

Maybe its not worth it at this point :shrug:

newton/_src/sim/builder.py
eric-heiden: Agreed, I will follow `load_static_visual_shapes` for GS handling.

newton/_src/utils/import_usd_deformable_utils.py
```diff
@@ -897,6 +918,8 @@ def claim_body(prim: Usd.Prim) -> None:
 
     for prim in Usd.PrimRange(root_prim, Usd.TraverseInstanceProxies()):
         type_name = str(prim.GetTypeName())
+        if collect_static_visuals and type_name in _SCOUT_VISUAL_TYPE_NAMES:
+            buckets.static_visuals.append(prim)
```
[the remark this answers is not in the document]
eric-heiden: Fixed: ignored prims are now filtered before entering the static visual bucket.

newton/_src/utils/import_usd_deformable_utils.py
eric-heiden: Kept the explicit set because Imageable also includes containers like Xform and Scope; it now clearly names the visual types the loader supports.

newton/_src/utils/import_usd.py
eric-heiden: Yes, _load_visual_shapes_impl already handles splat data, transforms, visibility, materials, and result maps; I added flag coverage.

newton/_src/utils/import_usd_deformable_utils.py
eric-heiden: We can revisit this in a follow-up perhaps but I think we need to have at least one separate traversal for articulations/bodies to use the physics import utility function `LoadUsdPhysicsFromRange`. I would rather not replace this function.

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@
 - Add compiled regular-expression support to label-based selectors while preserving glob strings.
 - Add simulation throughput, real-time factor, p95 step-time, steady-state GPU-memory, timestep, and MuJoCo solver-iteration metrics to the ASV robot-learning benchmarks.
 - Add `joint_dof_mask` to `newton.ik.IKSolver` to keep selected joint DOFs fixed during LM optimization. (#3488)
+- Load visual-only USD geometry outside rigid-body hierarchies as static shapes by default, with a `load_static_visual_shapes` importer opt-out.
```
jcarius-nv: Could we move this entry to `Changed` and make the compatibility guidance explicit? This broadens the default behavior of every USD import. Existing callers that need the previous body-associated-visuals-only behavior should be told to pass `load_static_visual_shapes=False`. That also satisfies the repository guidance requiring migration instructions for `Changed` entries.

newton/_src/utils/import_usd_deformable_utils.py
```diff
@@ -862,8 +865,29 @@ def has_candidates(self) -> bool:
     }
 )
 
+# Keep this set aligned with _load_visual_shapes_impl's leaf-shape branches.
```
jcarius-nv: Could we keep the supported visual-shape vocabulary in one source of truth? This set duplicates `_visual_geom_types` in `import_usd.py`, and the “keep aligned” comment leaves a silent failure mode where a newly supported visual type works under rigid bodies but is omitted from the static pass. For example, define one shared canonical set and derive the lowercase lookup from it.

newton/_src/sim/builder.py
```diff
@@ -3420,6 +3420,7 @@ def add_usd(
         skip_mesh_approximation: bool = False,
         load_sites: bool = True,
         load_visual_shapes: bool = True,
+        load_static_visual_shapes: bool = True,
```
jcarius-nv, line 3423: Before making this default-on for every USD import, could we add measurements from a representative populated Isaac Lab scene—not just scenes with no static visuals? Please include the delta in imported shape count, import time, model/viewer memory, and replicated-environment cost. Also confirm the headless/no-viewer case: does merely importing these additional shapes impose runtime simulation or memory cost when no viewer or sensor renderer is created?

jcarius-nv (commented):
Three focused follow-ups from the Standards and Newton Fit review, LGTM otherwise. Thank you!

andrewkaufman (dismissed):

CHANGELOG.md
eric-heiden: Done. I moved this to `Changed` and added the opt-out: pass `load_static_visual_shapes=False` to keep the previous behavior.

newton/_src/utils/import_usd_deformable_utils.py
eric-heiden: Done. Both code paths now use one shared set of supported visual types; the lowercase lookup is derived from it.

newton/_src/sim/builder.py
eric-heiden: Measured with the official Isaac Lab Franka Cabinet scene, including its static frame marker:

| Metric | Opt-out | Default-on | Overhead |
|---|---:|---:|---:|
| Median import time | 2.639 s | 2.691 s | **+52 ms / 2.0%** |
| Shapes | 86 | 92 | +6 |
| Host geometry payload | 3.53 MB | 3.53 MB | None |

The six shapes are global, so 4,096 replicated environments still add only six shapes. Model/viewer memory changed by about 1.6/8.9 KiB, and headless collision timing stayed within noise (~0.9%).

andrewkaufman (approved):

PR #3697, opened 2026-07-28
Fix shared USD collider visibility

eric-heiden:
## Description

Preserve visibility for USD collision prims whose effective purpose is `default` or `proxy`, allowing one geometry prim to serve as both visual and collision geometry.

Previously, collider fallback visibility was scoped to the whole import: once any separate visual geometry was present, colliders elsewhere lost `ShapeFlags.VISIBLE`. This hid a default-purpose shared collider merely because another body authored a separate visual shape. The importer now uses its existing viewport-purpose predicate when deciding collider visibility. Dedicated `guide` colliders remain hidden by default, while `force_show_colliders=True` still reveals them for debugging.

The existing collision-only tests now explicitly author dedicated proxies as `purpose="guide"`, and a new in-memory USD test covers mixed shared and separate geometry.

## Checklist

## Test plan

- `uv run --extra dev python -m unittest newton.tests.test_import_usd`
- `uvx pre-commit run -a`

## Bug fix

**Steps to reproduce:**

1. Create a USD stage with a default-purpose collider that is also intended as visual geometry.
2. Add separate visual geometry on another body and a dedicated guide-purpose collider.
3. Import with `ModelBuilder.add_usd(..., force_show_colliders=False)`.
4. Observe that the shared collider is collision-enabled but lacks `ShapeFlags.VISIBLE`.

**Minimal reproduction:**

```python

import newton

shared_body = UsdGeom.Xform.Define(stage, "/World/SharedBody").GetPrim()
UsdPhysics.RigidBodyAPI.Apply(shared_body)
shared = UsdGeom.Cube.Define(stage, "/World/SharedBody/Geometry").GetPrim()
UsdPhysics.CollisionAPI.Apply(shared)

separate_body = UsdGeom.Xform.Define(stage, "/World/SeparateBody").GetPrim()
UsdPhysics.RigidBodyAPI.Apply(separate_body)
UsdGeom.Sphere.Define(stage, "/World/SeparateBody/Visual")

shared_flags = builder.shape_flags[result["path_shape_map"]["/World/SharedBody/Geometry"]]
assert shared_flags & newton.ShapeFlags.VISIBLE
assert shared_flags & newton.ShapeFlags.COLLIDE_SHAPES
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed USD collider visibility so collision geometry shared with visual purposes remains visible in the viewport.
  * Colliders marked for guide purposes remain hidden when appropriate, while their corresponding visual geometry stays visible.
  * Improved handling of collision shapes with material properties and viewport display settings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -106,6 +106,7 @@
 - Fix Style3D solver divergence caused by isolated vertices.
 - Fix compiler warnings about overflowing int32 constants when compiling SDF texture and `SensorTiledCamera` kernels.
 - Fix USD site import to discover sites beneath non-visual containers, collider prims, and instanceable rigid-body prims independently of `load_visual_shapes`; the reworked traversal also speeds up import of scenes with many nested `Xform` or instance prims.
+- Fix `ModelBuilder.add_usd()` hiding default- and proxy-purpose collision prims that also serve as visual geometry.
 - Fix `SolverFeatherstone` BALL joints to apply passive `joint_damping` on all three angular DOFs.
 - Fix excessive memory usage when importing MJCF or URDF models containing many visual-only shapes with self-collisions disabled.
 - Fix `FastKitchenG1` ASV metrics to build the kitchen scene instead of a plain G1 model.
--- a/newton/_src/utils/import_usd.py
+++ b/newton/_src/utils/import_usd.py
@@ -1200,11 +1200,9 @@ def parse_usd(
 
         USD viewports draw the ``default`` and ``proxy`` purposes and hide ``guide`` and
         ``render``; the allowlist also keeps any future purpose hidden until explicitly
-        handled. Colliders deliberately do not use this check: ``guide`` is the conventional
-        purpose for authored collision geometry (e.g. the MuJoCo USD exporter), and the
-        collider display policy (``force_show_colliders`` / ``hide_collision_shapes``) is the
-        explicit mechanism for revealing colliders — gating them on purpose would make
-        ``force_show_colliders`` a no-op on such assets.
+        handled. For colliders, this identifies geometry authored to serve as both a visual
+        and collision shape. Explicit collider display remains independent of purpose so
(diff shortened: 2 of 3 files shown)


andrewkaufman:
@eric-heiden — would you be willing to close this in favour of #3712?

We hit the same bug independently and landed on the same predicate. #3712 additionally covers two things this PR does not touch, and since all three changes are in the same few lines of `collider_is_visible`, keeping them separate means resolving a conflict where the reasoning differs rather than the code.

What #3712 adds beyond this PR:

- **`physics:approximation` no longer replaces the render mesh.** It is scoped to collision, so approximating a drawable prim now yields an approximated collider *plus* a visual carrying the authored topology. On a UR5e + Robotiq gripper this is the difference between the gripper drawing as 353 hull fragments and drawing its actual meshes.
- **The empty-viewport fallback no longer overrides `guide`.** Colliders were force-shown when an import produced no visual shapes at all, which drew `guide` geometry through the visual toggle. Removing it surfaced exactly one mislabelled bundled asset (`cartpole_single_pendulum.usda`, whose collision boxes are its only geometry yet were marked `guide`); that is fixed in the asset.

Your regression test is carried over unchanged with credit — it pins the shared-geometry contract more explicitly than anything we had. I also ran `shared_collision_visual_visual.usda` from this PR against both branches: the shared default-purpose collider is visible and the dedicated `guide` collider is not, identically. Nothing your PR fixes is lost.

One difference worth your eye rather than glossing over: this PR keeps `collider_has_visual_material` and the model-wide fallback, ORing `_is_viewport_drawn` alongside them. #3712 removes both, so a `guide` collider with a bound PBR material stays hidden (the case #3650 was opened for) and a collision-only asset opens empty with `force_show_colliders` to inspect it. If you disagree with dropping the fallback, that is the part to push back on.

Happy to go the other way and rebase #3712 onto this if you would rather land yours first.

eric-heiden:
Closing in favor of https://github.com/newton-physics/newton/pull/3712.

PR #3706, opened 2026-07-28
Fix small-angle revolute joint reconstruction

### LINKED ISSUE #3434: [BUG] Stabilize and share signed twist-angle reconstruction
### Bug Description

Newton reconstructs signed revolute-joint coordinates with the following expression in
`newton/_src/sim/articulation.py::reconstruct_angular_q_qd()`:

```python
twist = wp.quat_twist(axis, q_pc)
q = wp.acos(twist[3]) * 2.0 * wp.sign(
    wp.dot(axis, wp.vec3(twist[0], twist[1], twist[2]))
)
```

In float32, `cos(q / 2)` rounds to exactly `1.0` for sufficiently small rotations. The `acos` expression then
returns zero even though the quaternion vector still contains the signed rotation. This creates a dead zone around
zero and a discontinuous first nonzero reconstructed angle.

In an isolated test, input angles through approximately `+/-0.0005 rad` reconstruct as zero, while an input of
approximately `0.0006 rad` reconstructs as approximately `0.0006905 rad`. The behavior reproduces on both CUDA
and CPU, so it is not specific to a GPU solver backend.

This caused a slowly moving Franka joint to remain locked at zero during repeated articulation-state updates. With
16 substeps, the resulting end-effector tracking error grew to approximately `34.2 mm`. The same trajectory crossed
zero smoothly when the repeated state update was disabled or when only one substep was used. Those configuration
changes avoid repeatedly exercising the dead zone but do not fix the underlying reconstruction error.

The same unstable expression is duplicated in `newton/_src/solvers/semi_implicit/kernels_body.py`.

### Expected behavior

Signed joint angles should reconstruct continuously through zero, while retaining the current angular branch. In
particular, rotations outside `[-pi, pi]` must not be silently wrapped if the input quaternion representation retains
that branch.

### Existing related implementations

- Warp's `wp.quat_twist_angle()` is unsigned and uses `acos`, so it does not solve this issue.
- Kamino has an internal `quat_twist_angle()` using `atan2`, but it first applies `quat_positive()`. This wraps the
  result to `[-pi, pi]`; for example, `4.0 rad` becomes approximately `-2.283 rad`. It is therefore not a drop-in
  replacement for articulation reconstruction. Core simulation code also should not depend on Kamino internals.

### Reproduction Script

```python
import warp as wp


@wp.kernel
def reconstruct(
    angles: wp.array[float],
    acos_result: wp.array[float],
    atan2_result: wp.array[float],
):
    i = wp.tid()
    axis = wp.vec3(0.0, 0.0, 1.0)
    q = wp.quat_from_axis_angle(axis, angles[i])
    twist = wp.quat_twist(axis, q)
    sin_half = wp.dot(axis, wp.vec3(twist[0], twist[1], twist[2]))

    acos_result[i] = 2.0 * wp.acos(twist[3]) * wp.sign(sin_half)
    atan2_result[i] = 2.0 * wp.atan2(sin_half, twist[3])


wp.init()
angles = wp.array([-0.0006, -0.0005, -0.0001, 0.0001, 0.0005, 0.0006], dtype=float)
acos_result = wp.empty_like(angles)
atan2_result = wp.empty_like(angles)

wp.launch(reconstruct, dim=len(angles), inputs=[angles, acos_result, atan2_result])

print("input:       ", angles.numpy())
print("acos * sign: ", acos_result.numpy())
print("atan2:       ", atan2_result.numpy())
```

### Proposed Fix

Add a signed function for joint-coordinate reconstruction and use that:

```python
@wp.func
def quat_twist_angle_signed(axis: wp.vec3, q: wp.quat) -> float:
    """Return the signed twist angle without canonicalizing its angular branch."""
    sin_half = wp.dot(axis, wp.vec3(q[0], q[1], q[2]))
    return 2.0 * wp.atan2(sin_half, q[3])
```

Upstream that to warp directly and fix existing `wp.quat_twist_angle` implementation

### System Information

- Newton: `1.4.0.dev0`
- Warp: `1.15.0.dev20260626`
- Python: `3.12.12`
- OS: Windows 11 (`10.0.26200`)
- GPU: NVIDIA RTX A5500 Laptop GPU, 16 GiB, SM 8.6
- CUDA Toolkit: 12.9
- NVIDIA driver reported by Warp: 13.2

felixmey:
## Description

Fix `eval_ik()` and `SolverSemiImplicit` reconstructing small float32
revolute-joint angles as zero.

The previous calculation recovered the angle with `acos(twist.w)`. Near
the identity rotation, float32 error can make `twist.w` exactly round to `1.0`
while the quaternion vector still contains a small signed rotation. This
creates a dead zone around zero.

Use Warp's `wp.quat_twist_angle_signed()` helper instead. It uses the
remaining signed quaternion information and preserves noncanonical
branches such as `+4.0 rad`.

Added an FK/IK regression test covering:

- `±0.0001 rad`
- `±0.0005 rad`
- `±4.0 rad`
- CPU and CUDA

Closes #3434.

## Checklist

## Test plan

- `uv run python -u newton/tests/test_kinematics.py -k test_fk_ik_revolute_small_angles`
- `uv run python -u newton/tests/test_kinematics.py`
- `uv run --extra dev -m newton.tests -k semi_implicit`
- `uvx pre-commit run -a`

The new regression test was also run temporarily with the old reconstruction
expression. It failed because `-0.0005 rad` reconstructed as `-0.0`, then
passed after restoring the fix.

## Bug fix

**Steps to reproduce:**

1. Create an articulation containing one revolute joint.
2. Set its coordinate to a small float32 angle such as `-0.0005 rad`.
3. Run `eval_fk()` followed by `eval_ik()`.
4. Observe that the previous implementation returns `-0.0` instead of
   the original joint coordinate.

**Minimal reproduction:**

```python
import numpy as np
import warp as wp

import newton

builder = newton.ModelBuilder(gravity=(0.0, 0.0, 0.0))
child = builder.add_link()
joint = builder.add_joint_revolute(
    parent=-1,
    child=child,
    axis=newton.Axis.Z,
)
model = builder.finalize()

state = model.state()
state.joint_q.assign(np.array([-5.0e-4], dtype=np.float32))

q_ik = wp.zeros_like(model.joint_q)
qd_ik = wp.zeros_like(model.joint_qd)
newton.eval_ik(model, state, q_ik, qd_ik)

print(q_ik.numpy()[0])
# Before: -0.0
# After:  -0.0005
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed inverse kinematics and semi-implicit solver handling of very small revolute-joint angles, preventing valid float32 values from being rounded to zero.
  * Improved signed angle calculations for revolute and single-axis rotational joints.

* **Tests**
  * Added regression coverage for forward and inverse kinematics with small positive and negative revolute-joint angles.

* **Documentation**
  * Updated the changelog with details of the fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -107,6 +107,7 @@
 - Fix compiler warnings about overflowing int32 constants when compiling SDF texture and `SensorTiledCamera` kernels.
 - Fix USD site import to discover sites beneath non-visual containers, collider prims, and instanceable rigid-body prims independently of `load_visual_shapes`; the reworked traversal also speeds up import of scenes with many nested `Xform` or instance prims.
 - Fix `SolverFeatherstone` BALL joints to apply passive `joint_damping` on all three angular DOFs.
+- Fix `eval_ik()` and `SolverSemiImplicit` rounding small float32 revolute-joint angles to zero. (#3434)
 - Fix excessive memory usage when importing MJCF or URDF models containing many visual-only shapes with self-collisions disabled.
 - Fix `FastKitchenG1` ASV metrics to build the kitchen scene instead of a plain G1 model.
 - Fix the `diffsim_bear` example crashing with its default CUDA configuration and diverging after a few training iterations.
--- a/newton/_src/sim/articulation.py
+++ b/newton/_src/sim/articulation.py
@@ -631,8 +631,7 @@ def reconstruct_angular_q_qd(q_pc: wp.quat, w_err: wp.vec3, X_wp: wp.transform, 
         qd: The joint velocity coordinate.
     """
     axis_p = wp.transform_vector(X_wp, axis)
-    twist = wp.quat_twist(axis, q_pc)
-    q = wp.acos(twist[3]) * 2.0 * wp.sign(wp.dot(axis, wp.vec3(twist[0], twist[1], twist[2])))
+    q = wp.quat_twist_angle_signed(axis, q_pc)
     qd = wp.dot(w_err, axis_p)
     return q, qd
 
--- a/newton/_src/solvers/semi_implicit/kernels_body.py
+++ b/newton/_src/solvers/semi_implicit/kernels_body.py
@@ -203,10 +203,7 @@ def eval_body_joints(
         axis_p = wp.transform_vector(X_wp, axis)
         axis_c = wp.transform_vector(X_wc, axis)
 
-        # swing twist decomposition
-        twist = wp.quat_twist(axis, r_err)
-
-        q = wp.acos(twist[3]) * 2.0 * wp.sign(wp.dot(axis, wp.vec3(twist[0], twist[1], twist[2])))
+        q = wp.quat_twist_angle_signed(axis, r_err)
         qd = wp.dot(w_err, axis_p)
 
         t_total = axis_p * (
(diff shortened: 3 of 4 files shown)


eric-heiden (approved):
Thanks!

PR #3707, opened 2026-07-28
Update Kamino Jacobians to Newton free joint wrench convention

chschuma-disney:
## Description

Newton follows the convention that the joint wrench of a free joint (as specified in `joint_f`) is defined on the center of mass of the child body in global axes. The same holds for joint twists in `joint_qd`. To properly account for this in Kamino, we need to handle the free joint as a special case when building the DoF Jacobian.

This also adds Kamino to the CoM body force tests that would have failed before without the fix.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_body_force.TestBodyForce
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Jacobian computation for free joints to ensure more accurate kinematics and force/torque mapping.
  * Refined joint Jacobian generation consistently across dense and sparse solver paths for improved correctness.
  * Fixed Kamino solver force/torque behavior when the center of mass is offset.

* **Tests**
  * Expanded center-of-mass offset coverage to run the existing force/torque tests for the Kamino solver as well.

### DIFF
--- a/newton/_src/solvers/kamino/_src/kinematics/jacobians.py
+++ b/newton/_src/solvers/kamino/_src/kinematics/jacobians.py
@@ -59,6 +59,65 @@ wp.set_module_options({"enable_backward": False, "default_grid_stride": False})
 ###
 
 
+@wp.func
+def build_full_joint_jacobian(
+    joint_id: wp.int32,
+    dof_type: wp.int32,
+    bid_B: wp.int32,
+    bid_F: wp.int32,
+    model_joints_X_Bj: wp.array[wp.mat33f],
+    model_joints_X_Fj: wp.array[wp.mat33f],
+    state_joints_p: wp.array[wp.transformf],
+    state_bodies_q: wp.array[wp.transformf],
+):
+    """
+    Computes the full (6x6) joint Jacobian (constraint and DoFs) for a specific joint.
+    """
+    # Retrieve the pose transform of the joint
+    T_j = state_joints_p[joint_id]
+    r_j = wp.transform_get_translation(T_j)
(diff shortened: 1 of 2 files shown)


newton/_src/solvers/kamino/_src/kinematics/jacobians.py
```diff
@@ -485,23 +485,32 @@ def _build_joint_jacobians_dense(
...
-    # Compute the wrench matrices
-    # TODO: Since the lever-arm is a relative position, can we just use B_r_Bj and F_r_Fj instead?
-    W_j_B = screw_transform_matrix_from_points(r_j, r_B_j)
-    W_j_F = screw_transform_matrix_from_points(r_j, r_F_j)
-
-    # General case: Compute the effective projector to joint frame and expand to 6D
-    if dof_type != JointDoFType.UNIVERSAL:
-        R_X_bar_j = expand6d(R_X_j)
-    # Universal joint: replace R_X_j with the frame of the intermediate body for rotation constraints
+    if dof_type == JointDoFType.FREE:
```
rubengrandia: Nit: The double change in dense and sparse highlight a good opportunity to share this part of the implementation with a wp.func. As long as the function argument does not get too awkward.

rubengrandia (dismissed):
LGTM. Extracting the common code into a helper function is optional

newton/_src/solvers/kamino/_src/kinematics/jacobians.py
chschuma-disney: Good point, I pushed an update.

rubengrandia (approved):

PR #3710, opened 2026-07-28
Speedup material USD loading with clustering alg

StafaH:
# Speed up USD faceVarying vertex splitting

## Description

`newton.usd.get_mesh()` converts faceVarying normals to per-vertex normals by clustering
every face corner of a mesh: a corner joins the first cluster at its vertex whose mean
direction is within `vertex_splitting_angle_threshold_deg` and whose UV matches, otherwise
it starts a new cluster (one output vertex per cluster). That loop ran in Python, one
corner at a time, with a NumPy `norm` + `dot` per cluster comparison. On a robot's visual
mesh set it dominates USD import: importing the Unitree Go2 visuals (17 meshes,
1.2 M corners) spent **3.3 s** there, and profiling showed ~1.14 M calls to the
per-corner `_new_vertex_from` helper.

Almost all of those corners never actually needed the sequential pass. This PR adds
`_split_corners_into_vertices()`, which decides the single-cluster case for every vertex
at once with array operations and only falls back to the sequential clustering for the
vertices that fail the test:

- Group the corners by vertex with one stable `argsort`, then use `reduceat` for the
  per-vertex direction sum, the minimum corner-to-mean dot product, and the
  all-corners-share-a-UV test.
- A vertex whose corner directions all lie within **half** the threshold angle of their
  mean lie within the full threshold of each other, and a normalized convex combination of
  vectors that are pairwise within θ is itself within θ of each of them — so the sequential
  pass would necessarily place all of them in the first cluster. That makes the vectorized
  test a sound (conservative) decision procedure, not an approximation.
- The remaining vertices run the original greedy clustering, restricted to their own
  corners and rewritten on Python floats instead of per-comparison NumPy scalar ops.
- Output vertices are still numbered by the corner that created them, so the returned
  points, indices, normals, and UVs are **bit-identical** to the previous implementation.

No API or behavior change; this is purely the cost of the conversion.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k Usd                 # 162 tests, OK
uv run --extra dev -m newton.tests -k test_vertex_splitting   # 27 tests, OK
```

New test `test_vertex_splitting_clusters_against_the_running_mean` pins the
distinction the vectorized test has to respect: corners at `(0, 24, 48)` degrees all
sit within the 25-degree threshold of their overall mean, but the third one is
36 degrees off the running mean of the first two and must start a second cluster.
Relaxing the half-threshold in the fast path to the full threshold makes that case
report 7 vertices instead of 8, and the test fails.

The existing cases cover the rest of the paths this touches: hard-edge splitting
(`test_cube_facevarying_normals_vertex_splitting`), UV-seam preservation
(`test_vertex_splitting_preserves_uv_seams`), and the mismatched-UV drop
(`test_vertex_splitting_drops_mismatched_facevarying_uvs`).

Equivalence was additionally checked against a verbatim copy of the previous
implementation over every faceVarying mesh of a real asset and a synthetic worst
case, asserting exact array equality on points, indices, normals, and UVs
(`all outputs identical to the previous implementation`).

## Performance

New ASV benchmark: `asv/benchmarks/setup/bench_usd_mesh.py::GetMeshFaceVaryingSplit`,
timing `newton.usd.get_mesh(load_normals=True, load_uvs=True)` on a textured grid with
faceVarying normals. `split_fraction` is the share of corners carrying a hard crease —
`0.0` is the common case for exported visual meshes, `0.25` forces a quarter of the
corners through the sequential fallback.

| grid | split_fraction | corners | before | after | speedup |
|---|---|---|---|---|---|
| 64 | 0.00 | 24.6 k | 78.3 ms | 6.2 ms | **12.6x** |
| 64 | 0.25 | 24.6 k | 84.4 ms | 33.7 ms | **2.5x** |
| 256 | 0.00 | 393 k | 1380.6 ms | 97.4 ms | **14.2x** |
| 256 | 0.25 | 393 k | 1635.9 ms | 561.6 ms | **2.9x** |

Micro-benchmark of the splitting function alone (best of 3, RTX PRO 6000 workstation,
Python 3.12), against a verbatim copy of the previous implementation:

| case | corners | before | after | speedup |
|---|---|---|---|---|
| Unitree Go2 visual meshes (17 meshes) | 1 195 896 | 3.255 s | 0.128 s | **25.5x** |
| synthetic grid 200x200 (creases + UV seams every few corners) | 240 000 | 0.946 s | 0.219 s | **4.3x** |

The real-asset case is the one that matters and the synthetic case is deliberately
adversarial: it pushes a large share of vertices into the sequential fallback, which is
where the remaining 4.3x comes purely from dropping the NumPy scalar ops.

### Downstream effect

Isaac Lab builds its Newton model by importing each clone source with
`load_visual_shapes=True`. Warm startup to the first `env.step()` at 4096 envs,
`Isaac-Velocity-Flat-UnitreeGo2`, MJWarp backend:

| | env_creation | total |
|---|---|---|
| before | 15.73 s | 17.82 s |
| after | 10.94 s | 12.98 s |

That recovers the full USD-import cost that #3351 and #3628 added for textured assets,
while keeping their texture and material-subset correctness fixes.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved USD mesh import speed for face-varying normals by optimizing vertex splitting for the common single-cluster case.
* **Bug Fixes**
  * Refined face-varying normal conversion to better handle texture seams and ensure consistent clustering behavior near the splitting-angle threshold.
* **Tests**
  * Added new regression coverage for vertex-splitting clustering behavior (including edge/threshold cases) and to confirm imported normals remain unit-length.
* **Performance Benchmarks**
  * Added an ASV benchmark for USD mesh imports that trigger face-varying normal splitting.


newton/_src/usd/utils.py
```diff
@@ -857,6 +857,135 @@ def _expand_indexed_primvar(
...
+        end = begin + int(group_sizes[group])
+        source_vertex = int(group_vertices[group])
+        clusters: list[list] = []
+        for corner in range(begin, end):
+            dir_x, dir_y, dir_z = (float(value) for value in grouped_dirs[corner])
+            corner_uv = None if grouped_uvs is None else tuple(grouped_uvs[corner].tolist())
+            for cluster in clusters:
+                sum_x, sum_y, sum_z = cluster[0], cluster[1], cluster[2]
+                scale = max(math.sqrt(sum_x * sum_x + sum_y * sum_y + sum_z * sum_z), 1e-30)
+                if (sum_x * dir_x + sum_y * dir_y + sum_z * dir_z) / scale < cos_thresh:
```
andrewkaufman, line 950: This is the line that changes tie-breaking versus the old code. The old comparison was `float(np.dot(sum_dir / max(np.linalg.norm(sum_dir), 1e-30), n_dir))` — it normalized the sum into a vector first, then dotted. Here the dot is taken against the unnormalized sum and divided by the length afterwards. Algebraically identical, rounds differently.

For a cluster sum of the unit vectors at 0 and 20 degrees, compared against the corner at 40 degrees:

```
old:  0.8660254037844387
new:  0.8660254037844386   (1 ULP lower)
cos(30 deg) = 0.8660254037844387
```

So at `angle_threshold_deg=30.0` the old code joins the cluster and this code starts a new one.

If keeping the old rounding matters, normalizing into a mean first and then dotting should reproduce it:

```python
inv = 1.0 / max(math.sqrt(sum_x * sum_x + sum_y * sum_y + sum_z * sum_z), 1e-30)
if (sum_x * inv) * dir_x + (sum_y * inv) * dir_y + (sum_z * inv) * dir_z < cos_thresh:
```

I have not confirmed that matches NumPy's rounding in all cases, so it may not be worth chasing — but this is the line where the bit-identical claim breaks.

newton/_src/usd/utils.py
```diff
@@ -857,6 +857,135 @@ def _expand_indexed_primvar(
...
+    group_sizes = np.diff(np.append(starts, corner_count))
+    group_vertices = grouped_vertices[starts]
+    group_of_corner = np.repeat(np.arange(len(starts)), group_sizes)
+
+    # The normalized sum of a group's corner directions is the cluster mean it would end
+    # up with if every corner joined the same cluster.
+    dir_sums = np.add.reduceat(grouped_dirs, starts, axis=0)
+    dir_means = dir_sums / np.clip(np.linalg.norm(dir_sums, axis=1, keepdims=True), 1e-30, None)
+    dots = np.einsum("ij,ij->i", grouped_dirs, dir_means[group_of_corner])
+    single_cluster = np.minimum.reduceat(dots, starts) >= cos_half_thresh
```
andrewkaufman, line 920: The half-threshold test is the one genuinely new piece of reasoning in this PR, and nothing in the test suite currently exercises its boundary. The description says `test_vertex_splitting_clusters_against_the_running_mean` covers exactly this (corners at 0/24/48 degrees, 25 degree threshold, must produce 8 vertices not 7), but that test is not in the diff — `grep -rn test_vertex_splitting_clusters_against_the_running_mean newton/tests/` returns nothing on this branch.

That case is well chosen, because it separates "within the threshold of the overall mean" from "within the threshold of the running mean", which is precisely what `cos_half_thresh` exists to protect against. Please include it.

Worth also pinning the comparison itself: a corner sitting exactly at the half-threshold is treated as single-cluster by this `>=`, and that boundary deserves a test too.

andrewkaufman (commented):
Nice result, and the vectorized decision procedure is the right shape for this — deciding the single-cluster case up front and only falling back to the greedy pass is sound, and the half-threshold argument in the docstring holds up. I reproduced the equivalence and the speedup locally:

- Differential test against a verbatim copy of the pre-PR loop over 3000 randomized meshes (mixed vertex/corner counts, tight and spread direction fans, with and without UVs, thresholds 5/25/45/90): **0 mismatches** on points, indices, normals and UVs.
- Micro-benchmark at 200k corners: **42x** on smooth input, **2.6x** on 25%-creased input. Consistent with the numbers in the description.
- `python -m newton.tests -k Usd`: 161 tests, OK.

Two things I would like resolved before this merges. The first is a behavior change I can reproduce; the second is that the test and benchmark the description relies on are not in the diff.

**1. Exact-threshold ties can now split differently (float regression).**

The sequential fallback was rewritten from NumPy ops onto Python floats, and that changes the last bit of the comparison. Where a corner sits *exactly* on the threshold, the old and new code disagree:

```python
# single vertex, coplanar corner normals at 0, 20, 40, 60, 80, 120 degrees
# threshold = 30.0 exactly
old -> indices [0 0 0 1 1 2]   (3 vertices)
new -> indices [0 0 1 1 2 3]   (4 vertices)
```

The cause is a 1-ULP difference in the cluster-mean dot product for the corner at 40° against the mean of {0°, 20°}:

```
old (np.linalg.norm + np.dot):  0.8660254037844387   >= cos(30°) -> joins
new (math.sqrt + scalar math):  0.8660254037844386   <  cos(30°) -> splits
cos_thresh:                     0.8660254037844387
```

So the mesh gains a vertex and a normal discontinuity at exactly 30°. It needs contrived-but-legal input (a symmetric fan, threshold on a boundary), and `vertex_splitting_angle_threshold_deg` is user-facing with a default of 25.0, so a caller passing 30.0 on regular geometry can land on it. Given the description states the output is bit-identical, this is worth either fixing or documenting.

Sweeping thresholds at 0.1° granularity over `[0, 180]` on that fan, 30.0 is the only divergence, so it is genuinely a tie-breaking issue rather than a systematic one. Options as I see them: compare in the same order/precision the old code did, or accept it and soften the bit-identical claim to "identical except for exact-threshold ties". I would lean toward the latter plus a comment, since chasing exact NumPy bit-compatibility here is not worth the complexity — but it should be a decision rather than an accident.

**2. The new test and benchmark are missing from the PR.**

The test plan cites `test_vertex_splitting_clusters_against_the_running_mean` and describes it as the case that pins the half-threshold logic (corners at 0/24/48° with a 25° threshold), and the performance section cites `asv/benchmarks/setup/bench_usd_mesh.py::GetMeshFaceVaryingSplit`. Neither exists on the branch:

```
$ git diff --stat upstream/main...
 CHANGELOG.md              |   1 +
 newton/_src/usd/utils.py  | 142 ++++++++++++++++++-------------------
$ grep -rn test_vertex_splitting_clusters_against_the_running_mean newton/tests/   # no match
$ ls asv/benchmarks/setup/bench_usd_mesh.py                                        # no such file
```

The 'New or existing tests cover these changes' box is ticked, and the three existing tests it lists do cover hard-edge splitting, UV seams and mismatched UVs — but none of them exercise the half-threshold boundary, which is the one piece of new reasoning this PR introduces. That 0/24/48° case is exactly the right test and it should be in the diff. Same for the benchmark, otherwise the performance claims are not reproducible by anyone else and there is nothing guarding against a regression.

Everything else looks good to me. The `argsort`/`reduceat` grouping, the creation-corner renumbering to preserve output vertex order, and the UV handling (including the non-faceVarying `corner_uvs[indices]` expansion) all read correctly, and I confirmed index dtype is preserved, empty meshes work, unreferenced vertices are dropped as before, and the antiparallel sum case still hits the 1e-30 clip.

andrewkaufman (commented):
Verified both fixes on `ac7382eb`, and both look good.

**Half-threshold test** — `test_vertex_splitting_clusters_against_the_running_mean` does what it needs to. I ran a negative control: reverting `cos_half_thresh` back to `cos_thresh` in the fast-path test makes it fail with `AssertionError: 7 != 8` on the `(0, 24, 48)` subtest, so it genuinely guards the reasoning rather than just passing alongside it. The `(0, 12, 24)` / `(0, 20, 40)` subtests and the separate `..._merges_corners_exactly_at_half_the_threshold` case pin the boundary from both directions, which is more than I asked for.

**Rounding** — documenting the tie behavior rather than chasing NumPy bit-compatibility is the right call, and the comment at the comparison plus the docstring paragraph land in the two places someone would actually look.

`python -m newton.tests -k Usd` → 163 tests OK. `-k test_vertex_splitting` → 28 OK. Benchmark runs: grid=64 gives 10.3 ms at `split_fraction=0.0` and 58.6 ms at `0.25`, so it does separate the fast path from the fallback as intended.

Two small leftovers, neither blocking:

**1. The description and CHANGELOG still claim identical output.** Now that the docstring documents exact-threshold ties as undefined, these two are inconsistent with it:

- PR description: "the returned points, indices, normals, and UVs are **bit-identical** to the previous implementation"
- `CHANGELOG.md`: "the split vertices, indices, normals, and UVs are unchanged"

The CHANGELOG line is the one that matters, since it ships. Suggest something like "...are unchanged, except that a corner sitting exactly at the threshold angle may cluster differently." A reader debugging a one-vertex difference after upgrading will go to the CHANGELOG, not the docstring.

**2. `split_fraction` in the benchmark only creates hard creases, not UV seams.** CodeRabbit raised this and I confirmed it: `corner_uvs` is derived from `points[indices]`, so every corner of a given vertex gets the same UV — across the whole `grid=8` case, 0 of 81 vertices have differing corner UVs. So the `same_uv` reduction never contributes to the split decision and the comment "carrying a hard crease or a UV seam" overstates what is exercised.

Not a correctness problem, and the crease path is the one that dominates real assets, so the timings stand. But the UV-seam branch is currently unbenchmarked. Either narrow the comment to creases, or offset the UV on the creased corners to cover both — the latter is a one-line change and would make the parameter match its description.

Happy for this to land once the CHANGELOG wording is settled.

andrewkaufman (commented):
Both leftovers are handled on `634a2f40`.

**CHANGELOG** now carries the caveat and names the parameter, which is better than what I suggested — someone hitting a one-vertex difference will find it without needing the docstring.

**Benchmark UV seams** — offsetting the creased corners is the right fix, and hoisting `creased` out of the `if` keeps the two arrays consistent. Verified at `grid=8`: `split_fraction=0.25` now gives 107/384 creased corners and 62/81 vertices with genuinely differing corner UVs, so the `same_uv` reduction is actually exercised. `split_fraction=0.0` still produces zero seams, so the fast-path case stays a clean measurement of the common path.

Re-verified on the final head:

- differential test against a verbatim copy of the pre-PR loop, 3000 randomized meshes — 0 mismatches on points, indices, normals and UVs
- `python -m newton.tests -k Usd` — 163 tests, OK
- benchmark 9.8 ms at `split_fraction=0.0` vs 57.5 ms at `0.25`

LGTM from me.

---

One note on the remaining open CodeRabbit thread about non-faceVarying UVs, since it is marked Major and should not block this PR: **the bug is real, but it predates this PR and this PR does not make it worse.**

A `constant` or `uniform` UV primvar on a mesh that also has faceVarying normals with `face_varying_normal_conversion="vertex_splitting"` raises:

```
constant UV (len 1, 4 verts):     IndexError: index 1 is out of bounds for axis 0 with size 1
uniform UV (len 2 = face count):  IndexError: index 2 is out of bounds for axis 0 with size 2
```

I ran that same repro against `upstream/main` at `de2d06f3` and got byte-identical errors. The old code path had `uvs[corner_idx] if uvs_facevarying else uvs[v]`, which indexes a length-1 or length-2 array by vertex id exactly as `corner_uvs[indices]` does now. So it is a faithful translation of a pre-existing bug, not a regression — `vertex` and `faceVarying` UVs (the cases that matter in practice, and the ones the tests cover) work correctly before and after.

Worth a separate issue rather than scope creep here. Two things to decide there that are outside this PR's remit: whether `constant`/`uniform` UVs should be broadcast to corners instead of dropped (`constant` in particular has an obvious correct answer — every corner gets the one value), and whether the guard belongs in `get_mesh` or earlier at `_find_uv_primvar`. Happy to file it.

asv/benchmarks/setup/bench_usd_mesh.py
eric-heiden, line 1: I don't think we need a separate benchmark here just for the UV handling of the USD mesh importer? This should definitely not run on every PR.

StafaH: Agreed. I'm going to remove it and follow up later with modifications to benchmarks of model init, more end-end regression testing should be fine.

eric-heiden (approved):
Thanks!

PR #3711, opened 2026-07-28
Break the viewer shape count into visual and collision

andrewkaufman:
## Description

Break the viewer's shape count into visual and collision shapes, listed indented under `Shapes` in the stats overlay.

The total alone is the wrong signal when reviewing how an asset imports. An asset can go from drawing its collision hulls to drawing its authored meshes while the total barely moves, and no combination of the existing numbers reveals it. Reviewing a UR5e + Robotiq gripper import, the total went `391 → 402` while what was actually drawn went `383 → 41` — invisible in the panel as it stood.

```
Shapes: 402
    visual: 41
    collision: 361
```

The two counts deliberately need not sum to the total: a shape carrying both `VISIBLE` and `COLLIDE_SHAPES` is counted in each, and that overlap is the useful part — it shows how much of what is drawn is collision geometry. On the same asset before an importer fix the panel reads `391 / visual 383 / collision 361`, where 353 shapes are being counted twice; after, `402 / visual 41 / collision 361` with no overlap at all.

`Model.shape_flags` is a device array, so the readback is cached per model rather than repeated every frame. The cache is keyed on model identity, so replacing the model (for example reloading a file in the `load_usd` example) recomputes it.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k viewer   # 221 tests
uvx pre-commit run -a
```

Verified interactively against `newton/examples/assets` and a UR5e + Robotiq gripper USD, checking the counts against `Model.shape_flags` computed independently.

## New feature / API change

No API change; this is display only. The counts come from existing `ShapeFlags`:

```python
import numpy as np
import newton

visible = np.count_nonzero(model.shape_flags.numpy() & int(newton.ShapeFlags.VISIBLE))
colliding = np.count_nonzero(model.shape_flags.numpy() & int(newton.ShapeFlags.COLLIDE_SHAPES))
```

## Summary by CodeRabbit

* **New Features**
  * Performance statistics now display separate counts for **visual** and **collision** shapes for the loaded model.
  * These counts may overlap, since a single shape can be included in both categories.
  * The displayed counts refresh when switching or loading a model.
* **Documentation**
  * Updated the changelog “Unreleased → Added” notes to describe the new visual vs. collision shape counting behavior.

### DIFF
--- a/newton/_src/viewer/viewer_gui.py
+++ b/newton/_src/viewer/viewer_gui.py
@@ -61,9 +61,10 @@ class ViewerGui:
         self._last_fps_time: float = perf_counter()
         self._fps_frame_count: int = 0
         self._current_fps: float = 0.0
-        # (model, visible, colliding) for the stats panel. shape_flags lives on the
-        # device, so the readback is cached rather than repeated every frame.
-        self._shape_role_counts: tuple[Any, int, int] | None = None
+        # Visual / collision shape counts for the stats panel, refreshed by
+        # update_shape_counts() when the viewer is given a model.
+        self._shape_visual_count: int | None = None
+        self._shape_collision_count: int | None = None
 
         # Selection panel state (UI-local, not simulation state).
         self._selection_ui_state = {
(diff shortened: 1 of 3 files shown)


newton/_src/viewer/viewer_gui.py
```diff
@@ -950,6 +972,16 @@ def _render_stats_overlay(self):
                 imgui.text(f"Worlds: {viewer.model.world_count}")
                 imgui.text(f"Bodies: {viewer.model.body_count}")
                 imgui.text(f"Shapes: {viewer.model.shape_count}")
+                roles = self._get_shape_role_counts(viewer.model)
```
eric-heiden: Does this function run at every frame now? That seems quite intrusive, better to cache these values - I think you can assume the model passed into the viewer constructor is final.

andrewkaufman: The `.numpy()` readback runs once per model, not per frame — but you had to read the helper to see that, so I have reordered it to check the cache first and made the docstring say so. Measured on a UR5e + gripper (393 shapes):

```
first call (device readback): 0.21 ms
10000 cached calls:           1.47 ms   (0.15 us/frame)
```

On the constructor assumption: `ViewerBase.set_model()` is public and may be called again with a different model, so the counts have to follow it rather than being fixed at construction.

Keying the cache on the model object covers both: a fixed model computes once and never again, and a swapped model recomputes on the next frame. Verified the swap invalidates correctly:

```
model A -> (41, 361)
model B -> (1, 1)
model A -> (41, 361)
```

Happy to hoist it into `set_model()` instead if you would rather the cost be explicit at the assignment rather than lazy, though that puts overlay concerns into the base viewer.

*(Edited: removed a justification that referenced an unmerged example — irrelevant to this PR.)*

andrewkaufman: @eric-heiden the load-usd example its referring to is my draft unreviewed https://github.com/newton-physics/newton/pull/3629

Not sure if that'll ever merge though... I'd like it to but its not really a prio, just something I'm using for debugging assets.

eric-heiden: Can't you overload `set_model()` from ViewerGL and then compute the visual/collider count there? The overhead from doing this every frame is definitely measurable and should be avoided.

---

Benchmarked the cached stats overlay using real headless ImGui frames; across four fresh processes, the PR path remains 15.4% slower than `origin/main`.

| Benchmark | Metric | Value | vs Main | Delta |
|-----------|--------|-------|---------|-------|
| `ViewerGui._render_stats_overlay` (20,000 cached calls) | wall time | 350.85 ms | 304.15 ms | +15.4% |
| `ViewerGui._render_stats_overlay` | time per call | 17.54 µs | 15.21 µs | +15.4% |
| `ViewerGui._get_shape_role_counts` (first model) | wall time | 2.48 ms | n/a | one-time cost |

andrewkaufman: Done in 34239782a — `ViewerGL.set_model()` now calls `gui.update_shape_counts(model)`, and the overlay reads two plain ints. No per-frame compute, no cache check, no function call. The device readback happens once, where the model arrives.

Your benchmark stands and my earlier answer missed the point: I measured my helper in isolation and concluded it was free, when what you measured was the whole overlay path. The residual cost is the two `imgui.text()` calls plus `indent`/`unindent` — that is the feature itself, not overhead I can cache away:

```
1 text call (main)                   0.087 us/frame
+ indent / 2 text / unindent (PR)    0.361 us/frame
```

If ~0.3 µs/frame of text rendering is still more than this is worth, the obvious lever is to draw the two lines only when a `Model Information` style section is expanded, or behind a toggle. Say the word and I will do that rather than have you weigh it.

Guarded on `self.gui is not None` so headless viewers skip it, and `update_shape_counts(None)` clears the counts so unsetting a model does not leave stale numbers. Verified across model swaps:

```
model A -> visual=41  collision=361
model B -> visual=1   collision=1
None    -> visual=None (rows hidden)
model A -> visual=41  collision=361
```

221 viewer tests pass.

eric-heiden (approved):
Thanks!

Issue #3713, opened 2026-07-28
Preserve MuJoCo collision filtering semantics


eric-heiden:
## Summary

Newton does not yet preserve the full collision-selection semantics of MJCF models.

For automatically generated contacts, MuJoCo accepts a geom pair when:

```text
(contype_a & conaffinity_b) != 0 or (contype_b & conaffinity_a) != 0
```

Newton currently handles the special case where both values are zero by assigning `collision_group=0`, but other `contype` / `conaffinity` combinations are flattened into Newton's existing collision-group representation. This can introduce unintended contacts or remove required selective self-collisions.

Explicit MJCF `<contact><pair>` entries are a separate selection path: they intentionally bypass the automatic mask and parent-child filters, can retain otherwise zero-mask geoms, and can override contact parameters. Their import, selection, and parameter precedence must remain separate from automatic mask filtering.

The goal of this epic is to preserve MJCF collision behavior across Newton collision pipelines and `SolverMuJoCo` without regressing the existing collision-group API or replicated-model scalability.

## Goals

- Preserve exact automatic pair eligibility from MJCF `contype` / `conaffinity` masks.
- Preserve explicit `<pair>` inclusion and contact-parameter overrides, including default-class inheritance.
- Preserve `<exclude>`, same-body, parent-child, world-parent, and articulation self-collision semantics.
- Define coherent composition between `collision_group`, mask filtering, and `shape_collision_filter_pairs`.
- Avoid imposing a 32-group limit or another breaking restriction on existing Newton collision groups.
- Preserve or intentionally define `SolverMuJoCo` mask export and round-trip behavior.
- Keep model construction and collision-filter storage scalable across replicated worlds.
- Validate semantic pair equivalence instead of relying only on raw `geom_contype` / `geom_conaffinity` array equality.

### 1. Automatic mask filtering

- Decide whether exact type/affinity masks are a public API, an internal runtime representation, or importer-compiled data.
- Rebase or rework #3104 on current `main` after the newer importer, broad-phase, filter-storage, and explicit-pair changes.
- Preserve the existing unbounded collision-group behavior; do not lower every existing group into a fixed 32-bit allocation without a compatible fallback.
- Define mixed-model behavior when MJCF masks interact with Newton-authored shapes, global shapes, multiple imported assets, and existing collision groups.
- Apply the mask predicate only to automatically generated pairs; explicit included pairs must retain MuJoCo's bypass semantics.
[...]

PR #3671, opened 2026-07-27
Fix MJCF contact pair defaults

han-xudong:
## Description

Resolve MJCF default classes when importing explicit contact pairs. Pair friction, margins, dimensions, and solver parameters now match the inherited values compiled by MuJoCo.

This change is limited to explicit pair attributes and does not alter the broader `contype` and `conaffinity` collision-filter mapping.

Closes #3670

## Checklist

- [x] New or existing tests cover these changes
- [x] The documentation is up to date with these changes
- [x] `CHANGELOG.md` has been updated (if user-facing change)

## Test plan

```text
uv run --extra dev python -m unittest newton.tests.test_import_mjcf.TestContypeConaffinityZero.test_explicit_pair_inherits_default_class newton.tests.test_import_mjcf.TestContypeConaffinityZero.test_explicit_pair_generates_contacts_with_group_zero newton.tests.test_import_mjcf.TestContypeConaffinityZero.test_explicit_pair_retains_unclassified_geoms_without_visuals newton.tests.test_import_mjcf.TestContypeConaffinityZero.test_explicit_pairs_across_contact_sections_without_visuals newton.tests.test_import_mjcf.TestContypeConaffinityZero.test_explicit_pair_hierarchical_labels_without_visuals
uvx pre-commit run -a
```

## Bug fix

**Steps to reproduce:**

1. Define contact-pair properties in a global or named default class.
2. Apply the class to an explicit contact pair.
3. Import the model and inspect the pair attributes.
4. Observe on `main` that the inherited values are missing.

**Minimal reproduction:**

See the reproduction in #3670.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed MJCF contact pairs so inherited default properties merge correctly and explicit pair values take precedence.
  * Correctly scale inherited contact margins and gaps.
  * Ensured bodies and pair endpoints contribute mass and inertia when applicable, including with visuals disabled.
  * Improved handling of compiler inertia settings and validation.

* **Tests**
  * Expanded MJCF importer coverage for inheritance, scaling, endpoint retention, and inertia behavior.

PR #3714, opened 2026-07-29
Preserve MJCF collision filtering semantics

### LINKED ISSUE #3014: [BUG] contype/conaffinity collision filtering not working

eric-heiden:
## Description

Closes #3014. Tracks the automatic-mask and round-trip work in #3713.

MuJoCo accepts an automatically generated geom pair when:

```text
(contype_a & conaffinity_b) != 0 or (contype_b & conaffinity_a) != 0
```

Newton's signed collision groups cannot express every such asymmetric relation directly. This change preserves the exact pair relation without adding public `collision_type` / `collision_affinity` arrays or changing broad-phase kernels:

- MJCF import stores each geom's effective 32-bit masks in internal `model.mujoco.contype` / `model.mujoco.conaffinity` custom attributes, including inherited defaults.
- A bounded optimizer compiles the mask relation into existing signed Newton groups plus exact sparse exclusions. Import uses at most one positive group to retain the prior interaction with Newton-authored shapes, and assigns unique negative group IDs across separate imports.
- `SolverMuJoCo(use_mujoco_contacts=True)` forwards preserved source masks verbatim when every selected collision shape has them and all Newton pair filters are already enforced by those masks, same-body topology, or body-wide MuJoCo `<exclude>` elements. Otherwise it reverse-compiles the combined Newton graph so later partial pair-filter edits remain effective. Preserved masks remain authoritative over later collision-group edits.
- Native Newton models use the reverse conversion: an exact, pair-matrix-verified biclique cover of their collision groups and pair filters. A successful result cannot add or remove a pair. Any graph with at most 33 selected shapes is guaranteed to fit; larger structured graphs commonly fit as well.
- If a native graph does not fit 32 bits, the solver uses the established legacy graph-color fallback. For example, 33 disconnected collision edges provably require 33 bits.
- Reverse compilation is bounded at 256 selected shapes or 1,024 sparse exclusions. Larger workloads skip directly to the established graph-color fallback without allocating the dense reverse-cover matrix.
- Replicated worlds with group-only filtering now reuse one contact-pair template even when they have no sparse filter blocks.

Each MuJoCo bit is an oriented biclique: geoms carrying the bit in `contype` form one side and geoms carrying it in `conaffinity` form the other. Minimum biclique cover is NP-hard, so the larger-graph compiler uses safe group-derived bicliques and per-shape stars, accepts only a fully verified result, and treats an uncovered edge as capacity failure rather than silently approximating it.

MuJoCo references: [collision selection](https://mujoco.readthedocs.io/en/stable/computation/index.html#selection), [`contype` / `conaffinity` XML fields](https://mujoco.readthedocs.io/en/stable/XMLreference.html#body-geom).

## Menagerie analysis

Benchmark corpus: [`mujoco_menagerie@71f066a`](https://github.com/google-deepmind/mujoco_menagerie/commit/71f066ad0be9cd271f7ed58c030243ef157af9f4).

- Scanned all 259 XML files.
- MuJoCo compiled 233 standalone models; all 233 forward pair matrices were verified exactly.
- The other 26 files are non-standalone include fragments (missing referenced bodies, sites, defaults, or keyframe state).
- All 67 top-level asset directories have a compilable representative.
- 35 standalone XML models use masks beyond only `0/0` and `1/1`; these occur in 11 representative asset directories.

Representative complex patterns:

| Model | Geoms (active) | Effective mask counts | Group-1 exclusions | Compiled exclusions | Reverse bits |
|---|---:|---|---:|---:|---:|
| Agility Cassie | 45 (20) | `0/0×25`, `0/15×1`, `1/0×5`, `2/4×7`, `4/2×7` | 83 | 35 | 8 |
| IIT SoftFoot | 197 (51) | `0/0×146`, `1/1×1`, `1/15×1`, `1/2×49` | 1,126 | 0 | 1 |
| PAL Talos | 105 (52) | `0/0×53`, `1/1×50`, `6/1×2` | 1 | 0 | 2 |
| Tetheria Aero Hand | 94 (30) | `0/0×64`, `1/1×29`, `2/2×1` | 16 | 16 | 28 |

Across all 233 compilable XMLs:

- Naive group-1 lowering: 3,677 mask exclusions.
- Compiled lowering: 86 exclusions, a 97.66% reduction.
- Forward compile time: 33.9 μs median, 1.129 ms maximum (3 repetitions/model).
- Reverse compile: exact for 233/233 compiler-produced Newton graphs, with no capacity failures and at most 28 bits.
- Reverse compile time: 0.940 ms median, 6.171 ms maximum (3 repetitions/model).

For a 1,024-world synthetic replication of SoftFoot's 51 active geom masks (median of 5 CPU runs):

| Representation | Replicate | Finalize | Stored filters |
|---|---:|---:|---:|
| Group 1 + exclusions | 292.9 ms | 1,328.8 ms | 1,204,224 |
| 2 compiled groups | 299.4 ms | 1,287.2 ms | 0 |

This removes about 9.19 MiB of packed filter-pair storage, improves finalization by 3.1%, and improves combined replication/finalization by 2.2%. Preserved masks add two `int64` custom-attribute values, or 16 bytes per imported shape.

## Related work

- [#3104](https://github.com/newton-physics/newton/pull/3104) adds first-class public masks and applies them in every broad phase. This PR is the no-new-core-API alternative: compile masks into current groups/exclusions and preserve raw values only in the MuJoCo namespace.
- [#3283](https://github.com/newton-physics/newton/pull/3283) and [#3285](https://github.com/newton-physics/newton/pull/3285) landed geom visualization-group and explicit-pair geom preservation.
- [#3645](https://github.com/newton-physics/newton/pull/3645), [#3662](https://github.com/newton-physics/newton/pull/3662), and [#3671](https://github.com/newton-physics/newton/pull/3671) cover explicit-pair parameters, selection, and default inheritance. Explicit `<pair>` contacts remain separate from automatic mask filtering here.

## Checklist

## Test plan

```text
uv run --extra dev --with mujoco -m unittest newton.tests.test_import_mjcf
# 240 passed

uv run --extra dev -m unittest newton.tests.test_collision_mask_compiler
# 21 passed

uv run --extra dev -m unittest newton.tests.test_builder_replicate
# 19 passed

uv run --extra dev --with mujoco -m unittest newton.tests.test_mujoco_solver.TestMuJoCoSolverCollisionMasks
# 2 passed

uv run --extra dev -m unittest newton.tests.test_model.TestModelMesh
# 48 passed

uvx pre-commit run -a
# all passed
```

## Bug fix

**Steps to reproduce:**

1. Import an MJCF floor with `contype=0 conaffinity=1`.
2. Add two spheres with `contype=1 conaffinity=0`.
3. Finalize the Newton model.
4. Before this change, all three shapes shared a Newton collision group, so the two spheres incorrectly collided with each other. With this change, both floor/sphere pairs remain and the sphere/sphere pair is absent.

## Summary by CodeRabbit

* **New Features**
  * Preserved MJCF `contype`/`conaffinity` collision filtering during import and MuJoCo contact generation.
  * Added exact translation between MuJoCo masks and Newton collision groups, including sparse exclusions.
  * Improved collision-filter handling for native Newton models, with capacity-aware fallback behavior.

* **Bug Fixes**
  * Corrected collision filtering when combining preserved MuJoCo masks with Newton pair filters.
  * Improved contact-pair reuse for replicated worlds without filters.

* **Documentation**
  * Added detailed guidance on collision filtering, mask precedence, capacity limits, and fallback behavior.


newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -5352,9 +5439,36 @@ def joint_has_raw_limit_solref(dof_idx: int) -> bool:
             colliding_shapes,
         )
 
-        shape_color = self._color_collision_shapes(
-            model, colliding_shapes, visualize_graph=False, shape_labels=model.shape_label
-        )
+        use_preserved_collision_masks = (
```
vreutskyy, line 5830: Could preserved masks from separate `add_mjcf()` calls be treated as distinct mask domains here? In a test with two asymmetric floor/two-sphere imports, Newton generated 8 cross-import pairs, while the forwarded MuJoCo masks allowed only 4. Would it make sense to reverse-compile the combined graph when multiple independent imports are present?

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -4649,6 +4678,58 @@ def _find_body_collision_filter_pairs(
         excluded = unfiltered_cumulative[segment_ends] == unfiltered_cumulative[segment_ends - pair_counts]
         return [(selected_bodies[bodies_a[k]], selected_bodies[bodies_b[k]]) for k in np.flatnonzero(excluded)]
 
+    @staticmethod
+    def _compile_newton_collision_masks(model: Model, selected_shapes: np.ndarray):
+        """Compile selected Newton groups and filters into MuJoCo masks."""
+        local_a, local_b = np.triu_indices(selected_shapes.shape[0], 1)
```
vreutskyy: Could the shape-count check happen before constructing every candidate pair? `_compile_newton_collision_masks()` currently performs the quadratic pair enumeration before `compile_newton_collision_graph()` applies its 256-shape limit. Could the sparse filter-pair array be mapped directly instead, avoiding that preparatory work for graphs that will fall back anyway?

vreutskyy (commented):
A couple of questions from my review:

newton/_src/solvers/mujoco/solver_mujoco.py
eric-heiden: Yes. Each add_mjcf() call now gets a distinct mask domain. Multi-domain selections reverse-compile the combined Newton collision graph.

newton/_src/solvers/mujoco/solver_mujoco.py
eric-heiden: Yes. The size check now runs before filter-pair work, and sparse filter pairs are mapped directly without enumerating all candidate pairs.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -5690,9 +5805,42 @@ def joint_has_raw_limit_solref(dof_idx: int) -> bool:
...
+        # add_mjcf() call and the masks already enforce every Newton filter.
+        # Otherwise generate new masks from Newton's final allowed shape pairs.
+        use_preserved_collision_masks = (
+            shape_mjc_contype is not None
+            and shape_mjc_conaffinity is not None
+            and shape_mjc_collision_mask_domain is not None
+            and np.all(shape_mjc_contype[colliding_shapes] != MUJOCO_COLLISION_MASK_UNSET)
+            and np.all(shape_mjc_conaffinity[colliding_shapes] != MUJOCO_COLLISION_MASK_UNSET)
+            and np.all(shape_mjc_collision_mask_domain[colliding_shapes] != MUJOCO_COLLISION_MASK_DOMAIN_UNSET)
+            and np.unique(shape_mjc_collision_mask_domain[colliding_shapes]).shape[0] <= 1
```
vreutskyy, line 5837: Collision-mask domains collide when independently parsed builders are merged. Each source builder assigns its first `add_mjcf()` domain `0`, and `ModelBuilder.add_builder()` copies that value unchanged. The merged model is therefore treated as a single mask domain, allowing `SolverMuJoCo` to forward masks that do not represent Newton’s collision graph. In the reproduction, Newton includes pair `(0, 1)`, but the forwarded `contype=[0, 0]`, `conaffinity=[1, 1]` masks suppress it. Remapping domain IDs during builder merging would keep independent imports distinct.

vreutskyy (commented):
One finding from the follow-up review:

newton/_src/solvers/mujoco/solver_mujoco.py
eric-heiden: Fixed in 2bf6e8ef by remapping incoming mask-domain IDs when ModelBuilder.add_builder() merges a source builder. The regression test covers the reported two-builder case. This key-specific handling is a correctness fix for this PR; #3787 tracks replacing it with a solver-agnostic custom-attribute merge mechanism.

vreutskyy (dismissed):
Looks good to me. Thanks for addressing the review findings.

eric-heiden:
Just fixed a performance regression in the ModelBuilder, should be good to go now.

vreutskyy (approved):
Looks good to me. Re-reviewed the mask-domain remapping optimization and latest main merge.

PR #3721, opened 2026-07-29
[Kamino] Fix Kamino USD importer to align with Newton importer

rubengrandia:
## Description

Align Kamino USD collider visibility with the Newton USD importer by using viewport draw semantics (`default` and `proxy` purposes) instead of treating colliders as visible only when `force_show_colliders` is set.

## Checklist

## Test plan

Manual verification with USD assets that have viewport-drawn collision geometry (default/proxy purpose) and guide/invisible colliders that should only appear when `force_show_colliders` is enabled.

## Bug fix

**Steps to reproduce:**

1. Import a USD asset with collision geometry authored as viewport-drawn (`default` or `proxy` purpose).
2. Import the same asset with Kamino USD importer without `force_show_colliders`.
3. Observe colliders are hidden even though Newton importer draws them.

**Minimal reproduction:**

```python
# Import a USD file with default/proxy-purpose collision meshes using
# Kamino USDImporter and compare visible shape flags against Newton importer.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved collision geometry visibility during USD imports, now respecting viewport draw semantics (including default/proxy behavior).
  * Made “hide collision shapes” aware of whether a rigid body has viewport-drawn visual geometry, ensuring collisions are suppressed only when appropriate.
  * Enhanced consistency between “force show colliders” and “hide collision shapes” so the expected collider visibility is applied reliably.


chschuma-disney (approved):
LGTM :rocket:

PR #3730, opened 2026-07-30
Derive USD cable joint stiffness from local segment lengths

### LINKED ISSUE #3722: [BUG] Cable import derives per-joint stiffness from the mean segment length
### Bug Description

`ModelBuilder.add_usd()` converts a cable's authored elastic moduli into per-joint stiffness (`E*A/L`, `E*I/L`, ...) using the **mean** segment length of the whole curve, then passes that single scalar to `add_rod()` / `add_rod_graph()`, which broadcast it to every joint.

`E` is a material constant, so per-joint stiffness should follow each joint's own rest length. Using a curve-wide mean makes a *material* property depend on how evenly the curve happens to be sampled. A cable with deliberately graded segments, e.g. a hose whose first segment is much longer than the rest, is silently mistuned, with no warning.

Measured on a cable with segment lengths `[0.5, 0.1, 0.1, 0.1]` (mean `0.2`), every joint receives the same stiffness `3141.59`, while the values implied by the local geometry differ per joint by roughly **±50%**. A uniformly sampled cable shows no error, which is why this hasn't surfaced before.

**This is not compensated downstream.** The VBD cable path applies the stiffness raw: stretch/shear as `k * u` on the absolute anchor gap, bend/twist as `k * kappa` on the integrated curvature binormal. The only rest-length normalization in the solver is cloth-specific (`spring_rest_length` / `edge_rest_length`). So the `1/L` has to be correct at import time.

Relevant code in `newton/_src/utils/import_usd_deformable_cable.py`:

- `:689` — `seg_len = sum(seg_lengths) / max(1, num_seg)` (single curve)
- `:375` — equivalent mean over `edges` (welded rod graph)

**Expected behaviour:** per-joint stiffness derived from the local segment geometry rather than a curve-wide mean, so that an evenly sampled cable is unchanged and a graded one is no longer uniformly too stiff or too soft.

`Model.joint_target_ke` is already `AttributeFrequency.JOINT_DOF`, so per-joint stiffness is representable; the information is only lost in the authoring path, where `add_rod` / `add_rod_graph` take scalars.

### Reproduction Script

```python
import newton
from pxr import Sdf, Usd, UsdGeom, UsdShade

curves = UsdGeom.BasisCurves.Define(stage, "/World/Cable")
curves.CreateTypeAttr().Set(UsdGeom.Tokens.linear)
curves.CreateWrapAttr().Set(UsdGeom.Tokens.nonperiodic)
# Graded sampling: segment lengths 0.5, 0.1, 0.1, 0.1 (mean 0.2).
curves.CreatePointsAttr([(x, 0.0, 0.0) for x in (0.0, 0.5, 0.6, 0.7, 0.8)])
curves.CreateCurveVertexCountsAttr([5])
curves.GetPrim().AddAppliedSchema("PhysicsCurvesDeformableSimAPI")

mat = UsdShade.Material.Define(stage, "/World/Mat")
mat.GetPrim().AddAppliedSchema("PhysicsCurvesDeformableMaterialAPI")
mat.GetPrim().CreateAttribute("physics:thickness", Sdf.ValueTypeNames.Float).Set(0.02)
mat.GetPrim().CreateAttribute("physics:stretchStiffness", Sdf.ValueTypeNames.Float).Set(2.0e6)
UsdShade.MaterialBindingAPI.Apply(curves.GetPrim()).Bind(mat)

builder.add_usd(stage, root_path="/World/Cable")

for j in range(builder.joint_count):
    print(f"joint {j}: {builder.joint_target_ke[builder.joint_qd_start[j]]:.2f}")

# All joints report the same value, derived from the mean segment length,
# although the segments they connect differ in length by 5x.
```

### System Information

- Newton: `1.5.0.dev0` (reproduced on `main` @ `0a3e9e1b` and on `d7581b73`)
- Warp: `1.17.0.dev20260729`
- Python: 3.12.13
- OS: Linux (Ubuntu, glibc 2.39)

jumyungc:
## Description
The experimental USD deformable importer previously discretized cable stiffness using one curve-wide mean segment length, which was broadcast to every joint. A material therefore produced different physical behavior depending on curve sampling.

Each cable joint now uses its local dual rest length,

0.5 * (L_parent + L_child),

including welded graphs and authored restShapePoints. For a fixed structural rigidity, uniformly sampled cables remain unchanged.

This PR also aligns cable materials with the current AOUSD proposal:

Read curvesThickness and the four structural curves*Stiffness attributes.
Derive unauthored modes from youngsModulus, poissonsRatio, and thickness.
Apply AOUSD’s 1 mm diameter fallback.
Continue accepting the earlier unprefixed modulus attributes with a DeprecationWarning.

Closes #3722.

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **New Features**
  * Improved USD cable material importing with current AOUSD attributes for thickness, modulus, density, and stretch, shear, bend, and twist stiffness.
  * Cable stiffness now uses each joint’s local dual rest lengths, including welded and unevenly sampled cables.
  * Added support for authored rest-shape points and full transform handling.

* **Bug Fixes**
  * Added validation and warnings for invalid material values, malformed rest shapes, and zero-length segments.
  * Improved thickness, radius, and missing-material fallbacks.

* **Documentation**
  * Documented updated material attributes, defaults, deprecations, and migration guidance.


andrewkaufman:
I haven't really reviewed this yet, but it would be good to account for the moduli->stiffness change being added to the proposal as well, @sschirm for details.

jumyungc:
> I haven't really reviewed this yet, but it would be good to account for the moduli->stiffness change being added to the proposal as well, @sschirm for details.

Thanks for the heads-up! If it lands while this PR is open, I’ll update this PR accordingly.

mzamoramora-nvidia:
Thanks for addressing the mean-length normalization and for covering graded, periodic, welded, zero-stiffness, and standalone rest-shape cases. I ran the relevant tests locally; the PR selection passed, and the strengthened tests fail against the unfixed base as expected.

I have one related request: now that the welded graph can receive per-joint stiffness, could it also retain authored `restShapePoints` for each owning curve for per-edge normalization? The current standalone and welded paths otherwise assign different stiffness to equivalent authored rest geometry. I left an inline suggestion with a possible implementation and regression-test direction.

I am reviewing the USD material attributes according to the proposal semantics implemented by Newton today. I would not block this PR while waiting for the possible proposal revision mentioned in the discussion.

newton/_src/utils/import_usd_deformable_cable.py
```diff
@@ -319,24 +358,24 @@ def global_node(local: tuple[str, int]) -> int:
...
 
-        # add_rod_graph applies one scalar stiffness per component and auto-orients its segments, so a
-        # welded curve's authored rest shape and per-point normals cannot be honored. Warn rather than
-        # changing the curve's behavior silently (a single, unwelded curve does honor both).
+        # A welded curve's authored rest shape and per-point normals cannot be honored: the graph's
+        # nodes and edges are derived from the current points, so neither reaches its joints. Warn
+        # rather than changing the curve's behavior silently (a single, unwelded curve honors both).
         for key in comp_paths:
             kprim = curve_recs[key].prim
             if deformable_read(kprim, "restShapePoints") is not None:
```
mzamoramora-nvidia: Would it be possible to preserve authored `restShapePoints` for welded cable graphs as part of this PR?

The previous warning made sense while `add_rod_graph()` could only receive one scalar stiffness, but this PR removes that limitation by applying stiffness per joint after construction. The graph also retains the owning curve and segment index for each edge in `edge_owner`, so it seems possible to compute a rest length for each edge from the transformed `restShapePoints` of that curve, while continuing to use the current points for `node_positions` and the constructed pose.

As written, an equivalent cable gets different normalization depending on whether it is imported standalone or as part of a welded graph: the standalone path uses authored rest lengths, while the welded path drops them and uses the possibly deformed current lengths. Could we pass those per-edge rest lengths to `_apply_local_cable_stiffnesses()` and replace `test_welded_graph_drops_rest_shape_warns` with a graded welded-rest-shape regression?

newton/_src/utils/import_usd_deformable_cable.py
```diff
@@ -81,23 +81,62 @@ def _read_validated_curve_topology(curves, path: str, *, warn: bool = True):
...
     if "bendStiffness" in material:
-        bend = create_cable_stiffness_from_elastic_moduli(material["bendStiffness"], radius, segment_length)[1]
+        bend = create_cable_stiffness_from_elastic_moduli(material["bendStiffness"], radius, rest_length)[1]
     if "twistStiffness" in material:
-        twist = material["twistStiffness"] * 0.5 * math.pi * radius**4 / segment_length
+        twist = material["twistStiffness"] * 0.5 * math.pi * radius**4 / rest_length
     return stretch, shear, bend, twist
 
 
+def _apply_local_cable_stiffnesses(
```
mzamoramora-nvidia, line 272: Small simplification: this helper only reads `ctx.builder` from the import context. Would passing `builder: ModelBuilder` directly make the dependency clearer and keep the stiffness calculation independent of the larger import context?

newton/_src/utils/import_usd_deformable_cable.py
jumyungc: Thanks! Both points are addressed.

newton/_src/utils/import_usd_deformable_cable.py
jumyungc: Done.

jumyungc:
Thank you for the great feedback.

In this PR, I tried to keep the changes minimal while updating the importer for the new AOUSD proposal and adding local-segment-length stiffness normalization.

One thing I noticed is that, although AOUSD provides separate rest-state and initial-pose information, Newton’s USD cable importer cannot yet represent them as distinct simulated configurations. This should also be addressed; I guess it may require broader Newton-side changes.

mzamoramora-nvidia (dismissed):
Thanks for the update. I reviewed the latest head and reran the local cable-import selection; all tests passed. The local dual-rest-length calculation is correct for standalone, periodic, and welded cables, and the AOUSD structural-stiffness alignment matches the cited proposal change.

Both of my earlier inline comments are addressed: welded graphs now use each owning curve's authored `restShapePoints` for normalization, and the helper receives the builder directly. I have no blocking concerns.

Nit: One small, non-blocking cleanup detail: four touched test docstrings still use declarative summaries (around lines 507, 665, 942, and 1022), while the local test guideline asks each summary to start in imperative mood. Could we change them to start with "Verify ..."? The affine-transform test should also have a blank line after its one-line summary.

eric-heiden:
🔵 Thanks for this. Deriving the structural stiffness from each segment's own local rest length instead of a single global length is the right fix, and the material resolution paths (legacy, mixed, missing) look well covered.

One thing worth resolving before merge: the new assumed cable radius is small enough that `ModelBuilder`'s inertia validation silently rewrites the inertia of every imported segment. Details inline.

newton/_src/utils/import_usd_deformable_cable.py
```diff
@@ -80,22 +109,188 @@ def _read_validated_curve_topology(curves, path: str, *, warn: bool = True):
...
+    A bound current-revision material delegates an unauthored thickness to the proposal's assumed
+    value; no bound curve material, or one still on the earlier revision, keeps Newton's own.
+    """
+    if material is not None:
+        for name in _CABLE_THICKNESS_ATTRS:
+            if name in material:
+                return 0.5 * material[name]
+        if not _is_legacy_only_curve_material(material):
+            return 0.5 * _AOUSD_DEFAULT_CURVES_THICKNESS / linear_unit
+    return _CABLE_RADIUS_COMPATIBILITY_FALLBACK / linear_unit
```
eric-heiden, line 214: 🟠 When a current-revision curve material is bound but does not author a thickness, `_cable_radius_from_material()` now returns `0.5 * _AOUSD_DEFAULT_CURVES_THICKNESS`, dropping the assumed radius from 2.5 mm to 0.5 mm. At that radius the segment bodies fall below `ModelBuilder`'s inertia floor, and every segment gets silently rewritten with a much larger, nearly isotropic inertia tensor.

`validate_and_correct_inertia_kernel` (`newton/_src/geometry/inertia.py:1043`) treats any principal moment under `max(1e-6 * I3, 1e-10)` as degenerate and adds roughly `1e-6` to all three principal moments. For a 0.1 m segment at `density = 1000`, the axial moment at r = 0.5 mm is about `9.8e-12`, well under that floor:

| assumed radius | below the inertia floor? | resulting inertia |
| --- | --- | --- |
| 2.5 mm (previous fallback) | no | tracks the analytic cylinder |
| 0.5 mm (this change) | yes | about 16x the true transverse moment and about 1e5x the true axial moment |

For those segment parameters the crossover sits between 0.8 mm and 0.9 mm, so the old assumed radius was comfortably above the floor and the new one is below it. The practical effect is that twist and bend response of every imported cable is quantitatively wrong, and the only user-visible signal is a generic `Inertia validation corrected 3 bodies.` message that never points back to the assumed cable thickness.

Two ways to address it:

- Keep `_CABLE_RADIUS_COMPATIBILITY_FALLBACK` (2.5 mm) as the assumed radius when thickness is unauthored.
- Or, if the 1 mm AOUSD default diameter must be the fallback, make the consequence explicit: extend `_warn_default_cable_radius()` to state that segments this thin are subject to `ModelBuilder`'s inertia correction, and note it in `changelog/3722.changed.1.md` and `docs/concepts/usd_parsing.rst` alongside the radius change.

Worth flagging that the floor is absolute, so any explicitly authored `curvesThickness` below roughly 1.8 mm hits the same path at typical densities and segment lengths. Scaling the floor relative to the body's own inertia would be the more durable fix, but that is separate from this change.

jumyungc:
> Thanks for the update. I reviewed the latest head and reran the local cable-import selection; all tests passed. The local dual-rest-length calculation is correct for standalone, periodic, and welded cables, and the AOUSD structural-stiffness alignment matches the cited proposal change.
> 
> Both of my earlier inline comments are addressed: welded graphs now use each owning curve's authored `restShapePoints` for normalization, and the helper receives the builder directly. I have no blocking concerns.
> 
> Nit: One small, non-blocking cleanup detail: four touched test docstrings still use declarative summaries (around lines 507, 665, 942, and 1022), while the local test guideline asks each summary to start in imperative mood. Could we change them to start with "Verify ..."? The affine-transform test should also have a blank line after its one-line summary.

Thanks for the style suggestion. I’ve updated all four docstrings accordingly.

newton/_src/utils/import_usd_deformable_cable.py
jumyungc: Thank you for flagging this. I took the second option to follow the AOUSD proposal and make the inertia-correction risk explicit in the warning, documentation, and changelog.

mzamoramora-nvidia (approved):

PR #2012, opened 2026-03-09
Use newton-usd-schemas 0.1.0

andrewkaufman:
## Description

Updated to official 0.1.0 package for `newton-usd-schemas`

There are more changes to the `uv.lock` than I would expect... Is it not usualy to run `uv sync` first? Should I discard those changes & only keep the schema lines?

## Summary by CodeRabbit

* **Chores**
  * Updated Newton USD Schemas dependency to a stable release version.

PR #3733, opened 2026-07-30
Use newton-usd-schemas 0.4.1

andrewkaufman:
## Description

Bump `newton-usd-schemas` to 0.4.1, [released to PyPI today](https://pypi.org/project/newton-usd-schemas/0.4.1/).

The 0.4.1 patch contains two changes:

- **`NewtonActuator` base type corrected** from a partially implemented `UsdGeomImageable` to `UsdTyped`. Actuators are not renderable, so nothing should have been treating them as imageable. A consequence worth knowing: actuators can no longer be rendered in Hydra-compatible viewers, so a viewer wanting actuator manipulators has to special-case them, or the content author has to parent a `UsdGeomImageable` beneath them.
- **`NewtonMimicAPI` documentation clarified** to state that the leader and follower must share a joint type; multi-DOF behaviour remains undefined and unsupported. This matches what the importer already assumes — see #3728, which reads `newton:mimicCoef0` as degrees based on the follower's joint type.

Newton identifies actuator prims by type name (`prim.GetTypeName() != "NewtonActuator"` in `newton/_src/actuators/usd_parser.py`), so parsing behaves identically against either version. The floor is raised anyway so that authored stages and the importer describe actuators the same way.

No `CHANGELOG.md` entry: no Newton-side behaviour changes, matching the precedent for standalone schema bumps (#2012). Happy to add one if you'd rather record it.

## Checklist

## Test plan

```
uv run --extra dev --extra importers -m unittest newton.tests.test_import_usd          # 298 tests, OK
uv run --extra dev --extra importers -m unittest newton.tests.test_actuators newton.tests.test_schema_resolver  # 88 tests, OK (15 skipped)
uvx pre-commit run -a                                                                  # all passed, incl. uv-lock
```

Confirmed the installed schema actually changed, rather than trusting the version string:

```python
>>> import importlib.metadata as md; md.version("newton-usd-schemas")
'0.4.1'
>>> from pxr import Usd, UsdGeom
>>> prim = Usd.Stage.CreateInMemory().DefinePrim("/A", "NewtonActuator")
>>> prim.IsA(UsdGeom.Imageable)
False
```

## Summary by CodeRabbit

* **Chores**
  * Updated the optional importer dependency to require version 0.4.1 or newer.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,7 +66,7 @@ importers = [
     "usd-exchange>=2.2.0 ; platform_machine == 'aarch64' and python_version < '3.13'",
 
     # Newton USD Schemas
-    "newton-usd-schemas>=0.4.0",
+    "newton-usd-schemas>=0.4.1",
 ]
 
 # Remeshing dependencies (for SurfaceReconstructor)
--- a/uv.lock
+++ b/uv.lock
@@ -3949,7 +3949,7 @@ requires-dist = [
     { name = "newton", extras = ["importers"], marker = "extra == 'rtx'" },
     { name = "newton", extras = ["onnx"], marker = "extra == 'examples'" },
     { name = "newton", extras = ["sim"], marker = "extra == 'examples'" },
-    { name = "newton-usd-schemas", marker = "extra == 'importers'", specifier = ">=0.4.0" },
+    { name = "newton-usd-schemas", marker = "extra == 'importers'", specifier = ">=0.4.1" },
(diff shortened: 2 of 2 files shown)


andrewkaufman:
@jcarius-nv this isn't really _necessary_, as the schema patch has no runtime impact on Newton itself. Just keeping things up-to-date really, so it appears that Newton 1.5 "recommends" schemas 0.4.1

There may yet be a 0.5.0 schema release before code freeze, happy to wait for that if you'd prefer

eric-heiden (approved):

PR #3754, opened 2026-07-31
Import MJCF mesh inertia modes

### LINKED ISSUE #3753: [BUG] MJCF importer ignores mesh asset inertia modes
## Description

The MJCF importer resolves mesh asset defaults but currently keeps only the mesh file, scale, and hull limit. It ignores the asset's `inertia` attribute, so imported nonconvex meshes always use Newton's exact solid-mesh mass properties.

Native MuJoCo supports `convex`, `exact`, `legacy`, and `shell` mesh inertia modes, with `legacy` as the default. These modes can produce substantially different body mass, center of mass, and inertia for the same mesh.

## Expected behavior

- Parse explicit, inherited, and default mesh inertia modes.
- Match native MuJoCo mass properties for each supported mode.
- Keep the visual and collision triangle mesh independent from the mesh used to calculate inertia.

## Reproduction

Import a watertight nonconvex mesh with `density=\"1\"` and change only `<mesh inertia=\"...\">`. Newton currently returns identical mass properties for every mode, while native MuJoCo does not.

han-xudong:
## Description

Parse MJCF mesh asset `inertia` values and match MuJoCo's `convex`, `exact`, `legacy`, and `shell` mass-property algorithms. The computation operates on the complete asset, including multi-material submeshes, while preserving the original visual and collision triangles.

Closes #3753

## Checklist

## Test plan

```bash
uv run --extra dev python -m unittest newton.tests.test_import_mjcf.TestImportMjcfGeometry
uvx pre-commit run -a
```

## Bug fix

**Steps to reproduce:**

1. Create a watertight nonconvex mesh asset.
2. Import the same asset with `inertia=\"convex\"`, `exact`, `legacy`, and `shell`.
3. Observe that Newton currently produces exact solid-mesh mass properties for every mode, unlike native MuJoCo.

**Minimal reproduction:**

```python
import newton

builder.add_mjcf("model.xml")  # model.xml uses <mesh inertia=\"shell\" .../>
```

## Summary by CodeRabbit

* **New Features**
  * Added support for MJCF mesh inertia modes, including exact, convex, legacy, shell, and inherited-shell configurations.
  * Mesh mass, center of mass, and inertia are now computed consistently across multi-part assets and scaling configurations.
  * Mesh loading can preserve sub-mesh structure when importing file-based assets.

* **Bug Fixes**
  * Improved handling of degenerate meshes, face orientation, planar legacy meshes, and negative scaling.
  * Improved compatibility with native MuJoCo mass properties.
  * Added validation for invalid convex-hull vertex limits.

* **Documentation**
  * Added a changelog entry describing MJCF mesh inertia mode support.


newton/_src/utils/import_mjcf.py
```diff
@@ -70,6 +70,119 @@ def _default_path_resolver(base_dir: str | None, file_path: str) -> str:
...
+        raise ValueError("MJCF mesh surface area is too small to compute inertia")
+    face_centroid = np.sum(areas[:, None] * centers, axis=0) / total_area
+
+    if mode == "convex":
+        inertia_vertices, inertia_faces = remesh_convex_hull(vertices, maxhullvert=maxhullvert)
+        inertia_vertices = np.asarray(inertia_vertices, dtype=np.float64)
+        inertia_faces = np.asarray(inertia_faces, dtype=np.int32).reshape(-1, 3)
+    else:
+        inertia_vertices = vertices
+        inertia_faces = faces
```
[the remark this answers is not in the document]
han-xudong: I kept the authored `maxhullvert` for convex inertia after checking this directly against native MuJoCo. With the same OBJ and `inertia="convex"`, setting `maxhullvert="4"` changes MuJoCo body mass relative to the unlimited hull, so passing `0` here would diverge from MuJoCo. The updated regression continues comparing the imported properties with native MuJoCo, while the new multi-material fixture uses the default complete hull to avoid Qhull early-termination ordering differences.

han-xudong:
@coderabbitai, addressed the latest review in 40bdd6a2:

- The regression now exports one OBJ/MTL asset with two `usemtl` groups and verifies that both submeshes are preserved and aggregated using one MJCF geom density.
- It compares mass, COM, and inertia for every supported mode against native MuJoCo and verifies that the relevant modes produce distinct mass values.
- The repeated triangle metric calculation is shared, and precomputed mesh inertia is explicitly marked valid.
- Existing mesh parser fixtures now use closed tetrahedra so the full MJCF test file remains valid under inertia computation.

Validation: `uv run --extra dev -m newton.tests -p test_import_mjcf.py` (267 tests) and `uvx pre-commit run -a`.

newton/_src/utils/import_mjcf.py
```diff
@@ -1016,23 +1137,33 @@ def parse_shapes(
...
+                    m_meshes,
+                    inertia_mode,
+                    maxhullvert,
+                )
+                mass_mesh = m_meshes[0]
+                mass_mesh.mass = mesh_mass
+                mass_mesh.com = mesh_com
+                mass_mesh.inertia = mesh_inertia
+                for m_mesh in m_meshes:
+                    m_mesh.is_solid = shape_cfg.is_solid
```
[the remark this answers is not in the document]
han-xudong: Fixed in 40bdd6a2. The imported mass properties now explicitly set `mass_mesh.has_inertia = True` before the mesh is added to the builder.

eric-heiden:
Thanks for tackling this. Matching MuJoCo's `legacy`, `shell`, `convex`, and `exact` mesh inertia modes is a real gap to close, and the native comparison test is a good way to pin the behavior.

The main things to look at before merging: mesh assets that imported fine before can now abort `add_mjcf()`, the `exact` path accepts and rejects some meshes differently from native MuJoCo, and hollow shapes silently pick up solid mass properties. Details in the inline comments.

newton/_src/utils/import_mjcf.py
```diff
@@ -71,6 +71,110 @@ def _default_path_resolver(base_dir: str | None, file_path: str) -> str:
...
+        weights = areas
+    else:
+        weights = np.sum((centers - face_centroid) * normals, axis=1) * areas / 3.0
+        if mode == "legacy":
+            weights = np.abs(weights)
+
+    total_measure = np.sum(weights)
+    if total_measure <= 0.0:
+        qualifier = "negative" if total_measure < 0.0 else "too small"
+        raise ValueError(f"MJCF mesh volume is {qualifier}; check triangle orientation or use inertia='shell'")
```
eric-heiden: 🟠 This raise aborts imports that used to work.

Every mesh geom now runs through `_compute_mjcf_mesh_inertia` unconditionally (the call added around line 1353), so this check is reached for all of them. In the default `legacy` mode a planar mesh produces a weight of exactly `0` for every triangle, since `face_centroid` lies in the same plane as the triangles, so `total_measure` is `0` and the import raises.

Before this change a degenerate mesh simply returned zero mass and the geom was added massless, so `ModelBuilder.add_mjcf()` returned a builder. Now it throws. Two existing fixtures in `newton/tests/test_import_mjcf.py` had to be rewritten from a single triangle to a closed tetrahedron for exactly this reason, and planar mesh colliders are otherwise first class in Newton (see `newton/tests/test_solver_mujoco_planar_mesh.py` and `_make_nonplanar_mujoco_mesh` in `newton/_src/solvers/mujoco/solver_mujoco.py`).

The raise also happens before the importer decides whether the geom can contribute inertia at all, so it fires for `mass="0"` / `density="0"` visual geoms, geoms outside `inertiagrouprange`, world geoms, and bodies that carry an explicit `<inertial>` that would overwrite the result anyway.

Native MuJoCo rejects these assets too, so strictness here may be intentional, but this is still a behavior break for MJCF that Newton accepted before, landing without a deprecation period or a word in the changelog fragment.

Suggested direction: skip the integration when the geom cannot contribute body inertia, and keep the hard error for `inertia="exact"` only, degrading the other modes to a warning plus zero mass so existing models keep importing.

<details>

<summary>Sketch</summary>

```py
total_measure = np.sum(weights)
if total_measure <= 0.0:
    if mode == "exact":
        qualifier = "negative" if total_measure < 0.0 else "too small"
        raise ValueError(f"MJCF mesh volume is {qualifier}; check triangle orientation or use inertia='shell'")
    warnings.warn(
        "MJCF mesh volume is non-positive; importing with zero mass properties",
        stacklevel=2,
    )
    return 0.0, wp.vec3(), wp.mat33()
```

</details>

newton/_src/utils/import_mjcf.py
```diff
@@ -1234,23 +1348,34 @@ def parse_shapes(
...
+                    m_meshes,
+                    inertia_mode,
+                    maxhullvert,
+                )
+                mass_mesh.mass = mesh_mass
+                mass_mesh.com = mesh_com
+                mass_mesh.inertia = mesh_inertia
+                mass_mesh.has_inertia = True
+                for m_mesh in m_meshes:
+                    m_mesh.is_solid = shape_cfg.is_solid
```
eric-heiden: 🟠 Hollow mesh shapes now get solid mass properties.

Setting `mass_mesh.has_inertia = True` together with `m_mesh.is_solid = shape_cfg.is_solid` makes the cached values satisfy the reuse condition in `compute_inertia_shape`, which only reuses `src.mass/com/inertia` when `src.has_inertia and src.mass > 0.0 and src.is_solid == is_solid` (`newton/_src/geometry/inertia.py`). Previously the meshes returned by `load_meshes_from_file` always kept the `Mesh.__init__` defaults `is_solid=True` / `has_inertia=True`, so a shape configured hollow failed that comparison and fell through to `compute_inertia_mesh(density, verts, indices, is_solid=False, thickness=cfg.margin)`, the thin shell path.

With the flags aligned, the values computed here from the solid volume integral win and `cfg.margin`, the documented shell thickness, is ignored. A user who sets `builder.default_shape_cfg.is_solid = False` and then calls `add_mjcf()` silently gets the solid inertia tensor, and the full solid mass for density driven geoms, which contradicts `docs/concepts/mass_inertia.rst` ("Hollow shapes (``ShapeConfig.is_solid=False``) compute shell inertia"). The hollow subtest in `newton/tests/test_import_mjcf.py` around line 1713 exercises this exact configuration and only keeps passing because it asserts nothing stronger than `np.trace(inertia) > 0`.

Suggested fix: only take over the mass properties when the shape is solid, so the `mass_mesh.mass/com/inertia/has_inertia` assignments and this loop run under `if shape_cfg.is_solid:` and hollow shapes stay on the existing shell fallback. If MuJoCo parity is meant to apply to hollow shapes too, please call that out in the changelog fragment and tighten the hollow subtest to pin the expected tensor.

newton/_src/utils/import_mjcf.py
```diff
@@ -71,6 +71,110 @@ def _default_path_resolver(base_dir: str | None, file_path: str) -> str:
...
+        inertia_faces = faces
+
+    _, centers, areas, normals = _triangle_metrics(inertia_vertices, inertia_faces)
+
+    if mode == "shell":
+        weights = areas
+    else:
+        weights = np.sum((centers - face_centroid) * normals, axis=1) * areas / 3.0
+        if mode == "legacy":
+            weights = np.abs(weights)
```
eric-heiden, line 169: 🟠 Mirrored meshes are treated as misoriented in `exact` mode.

The signed pyramid weights computed here are rejected further down when their sum is negative. But `load_meshes_from_file` applies an asset's signed scale straight to the vertices (`newton/_src/utils/mesh.py`, around line 1367) without reversing the face winding, so a perfectly valid mesh mirrored along an odd number of axes, for example `scale="-1 1 1"`, arrives here with every normal inverted and the import fails.

Negative mesh scale is supported in MuJoCo, and it reverses triangle winding for left-handed scaling before integrating inertia (see `src/user/user_mesh.cc` in [mujoco](https://github.com/google-deepmind/mujoco/blob/main/src/user/user_mesh.cc)). The new test only uses positive unit scale, so this path is uncovered.

Suggested fix: reverse each triangle's winding when the determinant of the applied scale is negative, then integrate as usual, and add a native comparison case with `inertia="exact"` and `scale="-1 1 1"`.

newton/_src/utils/import_mjcf.py
```diff
@@ -71,6 +71,110 @@ def _default_path_resolver(base_dir: str | None, file_path: str) -> str:
...
+
```
eric-heiden, line 170: 🟠 `exact` mode skips MuJoCo's face orientation check.

The only validation on this path is that the aggregate signed volume comes out positive. A mesh with a few inconsistently wound faces can still integrate to a positive total, so the import succeeds and returns a COM and inertia tensor that are quietly wrong, with no error and no warning.

MuJoCo builds directed half-edges and rejects exact inertia when adjacent faces disagree on orientation, before it runs this integration (see `src/user/user_mesh.cc` in [mujoco](https://github.com/google-deepmind/mujoco/blob/main/src/user/user_mesh.cc)). The added test uses two consistently oriented boxes, so it cannot distinguish accepting from rejecting such a mesh.

Suggested fix: validate topology and adjacent face orientation for `exact` before integrating, taking into account vertices duplicated by material splitting, and add a test that a mesh with a single flipped face is rejected the same way native MuJoCo rejects it.

newton/_src/utils/import_mjcf.py
```diff
@@ -71,6 +71,110 @@ def _default_path_resolver(base_dir: str | None, file_path: str) -> str:
...
+        return triangles, centers, areas, normals
+
+    vertices = []
+    faces = []
+    vertex_offset = 0
+    for mesh in meshes:
+        mesh_vertices = np.asarray(mesh.vertices, dtype=np.float64)
+        mesh_faces = np.asarray(mesh.indices, dtype=np.int32).reshape(-1, 3)
+        vertices.append(mesh_vertices)
+        faces.append(mesh_faces + vertex_offset)
```
eric-heiden: 🟠 The limited convex hull is built from already scaled vertices.

The vertices gathered here have already been multiplied by the asset scale and the importer scale (applied around lines 1125 to 1136), and for `inertia="convex"` they are what `remesh_convex_hull` sees. MuJoCo builds its convex graph on the raw asset vertices and applies the mesh scale afterwards, in `ApplyTransformations` (see `src/user/user_mesh.cc` in [mujoco](https://github.com/google-deepmind/mujoco/blob/main/src/user/user_mesh.cc)).

For an unlimited hull the ordering does not matter, since the hull is affine invariant. Once `maxhullvert` is set it does: Qhull's early termination and furthest point selection can keep a different vertex subset after a nonuniform scale, so Newton can report different convex mass properties than native MuJoCo for the same model with the same authored `maxhullvert`.

Suggested fix: build the limited hull from the unscaled asset vertices, then apply the authored scale to the selected hull vertices before integrating. A native comparison using an anisotropic scale and a point cloud whose hull exceeds the budget would lock this in.

changelog/3753.fixed.md
```diff
@@ -0,0 +1 @@
+Import MJCF mesh asset `inertia` modes when computing body mass properties.
```
eric-heiden, line 1: 🔵 Mesh mass properties are recomputed for every geom reference.

In `newton/_src/utils/import_mjcf.py`, `_compute_mjcf_mesh_inertia` is called once per geom that references a mesh, and for `inertia="convex"` that includes a SciPy/Qhull hull build. The result depends only on the mesh asset, the resolved scale, the inertia mode, and the hull budget, so shared visual and collision assets and repeated instances redo the same O(faces) integration or hull construction many times over on large scenes.

It also runs for geoms that never use the result: world geoms, zero density geoms, and bodies whose explicit `<inertial>` overwrites the computed values right afterwards.

Suggested fix: cache the unit density mass properties on the entry in `mesh_assets`, keyed by resolved scale, inertia mode, and hull budget, and skip the computation entirely when the geom cannot contribute body inertia.

newton/_src/utils/import_mjcf.py
han-xudong: Addressed in 918596dd. Precomputed volume properties are now installed only for solid mesh configs. Hollow meshes remain on Newton's thickness-aware shell path, including every submesh, and the regression now pins the distinct hollow inertia.

newton/_src/utils/import_mjcf.py
han-xudong: Addressed in 918596dd. Mesh inertia integration is now skipped for world, zero-density/mass, out-of-range, and explicit-inertial geoms. Non-positive legacy/convex meshes warn and retain zero mass properties, while exact remains strict. I added regressions for planar and non-contributing meshes.

newton/_src/utils/import_mjcf.py
han-xudong: Addressed in 918596dd. Convex inertia now builds the hull from unscaled asset vertices and applies mesh scaling afterward. Limited hull selection also uses the MuJoCo Q9/TA Qhull options, with a native anisotropic-scaling comparison added.

newton/_src/utils/import_mjcf.py
han-xudong: Addressed in 918596dd. Left-handed mesh scaling now reverses the winding used for inertia integration. I added an exact-mode comparison against native MuJoCo for scale="-1 1 1".

newton/_src/utils/import_mjcf.py
han-xudong: Addressed in 918596dd. Exact mode now welds duplicate vertex positions, including material-split vertices, and rejects duplicate directed half-edges. The new regression verifies that Newton and native MuJoCo both reject inconsistent adjacent-face orientation.

changelog/3753.fixed.md
han-xudong: Addressed in 918596dd. Mass properties are now computed only for contributing solid geoms and cached per mesh asset, scale, inertia mode, and hull budget. A regression verifies that two geoms sharing one asset compute the properties once.

han-xudong:
@eric-heiden Thank you for the thorough review. I addressed all six points in 918596dd and added focused regressions for the affected behaviors. The full MJCF importer suite (274 tests), mesh-inertia tests, convex-hull tests, and pre-commit checks all pass locally. The branch is also updated with the latest upstream/main. When convenient, could you please take another look?

newton/_src/geometry/utils.py
```diff
@@ -854,9 +854,9 @@ def _warn_degenerate(rank: str) -> None:
     # General (full 3D) case.
     qhull_options = "Qt"
     if maxhullvert > 0:
-        # qhull "TA" actually means "number of vertices added after the initial simplex"
-        # from mujoco's user_mesh.cc
-        qhull_options += f" TA{maxhullvert - 4}"
+        # MuJoCo pairs furthest-point processing (Q9) with a cap on vertices
+        # added after Qhull's initial simplex (TA).
+        qhull_options += f" Q9 TA{maxhullvert - 4}"
```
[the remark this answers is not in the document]
han-xudong: Addressed in b51cfb35. remesh_convex_hull now rejects positive maxhullvert values below 4 before either the planar or Qhull path, while preserving non-positive values as unlimited. The accepted range is documented and covered by a regression for values 1 through 3.

han-xudong:
@coderabbitai Addressed the applicable review items in b51cfb35. Positive maxhullvert values below 4 now fail with a clear ValueError and have regression coverage, and the native MuJoCo orientation test no longer depends on external error wording. I left the Pillow suggestion unchanged because that heightfield test came from the merged upstream branch and is outside this PR; calling skipTest there would also mark the earlier inline and binary assertions as skipped. Validation: 274 MJCF importer tests, 19 convex-hull tests, and all pre-commit hooks pass locally.

newton/tests/test_remesh_convex_hull.py
```diff
@@ -146,6 +146,23 @@ def test_maxhullvert_limits_vertex_count(self):
...
+                [0.0, 1.0, 0.0],
+                [0.0, 0.0, 1.0],
+            ],
+            dtype=np.float64,
+        )
+
+        for maxhullvert in (1, 2, 3):
+            with self.subTest(maxhullvert=maxhullvert):
+                with self.assertRaisesRegex(ValueError, "non-positive or at least 4"):
+                    remesh_convex_hull(verts_in, maxhullvert=maxhullvert)
```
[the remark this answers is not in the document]
han-xudong: Addressed in 930124c4. The regression now exercises maxhullvert=4 with the same tetrahedron and verifies the expected four vertices and four faces, covering the lower valid boundary.

newton/_src/utils/import_mjcf.py
```diff
@@ -71,6 +71,149 @@ def _default_path_resolver(base_dir: str | None, file_path: str) -> str:
...
+
+    _, centers, areas, _ = _triangle_metrics(vertices, faces)
+    total_area = np.sum(areas)
+    if total_area <= 0.0:
+        return _zero_mass_properties("MJCF mesh surface area is too small to compute inertia")
+    face_centroid = np.sum(areas[:, None] * centers, axis=0) / total_area
+
+    if mode == "convex":
+        hull_source_vertices, _ = _combine_meshes(convex_meshes if convex_meshes is not None else meshes)
+        inertia_vertices, inertia_faces = remesh_convex_hull(hull_source_vertices, maxhullvert=maxhullvert)
```
eric-heiden, line 151: 🟠 `inertia="convex"` falls back to Newton's 64 vertex collision default where MuJoCo uses an unlimited hull.

The budget passed here comes from line 697, which falls back to `mesh_maxhullvert` (default `Mesh.MAX_HULL_VERTICES`, 64) when the asset does not author a `maxhullvert`. MuJoCo's default is `-1`, meaning unlimited. So for any mesh whose hull exceeds 64 vertices and does not author the attribute, the imported mass, COM, and inertia come from a coarser hull than the one MuJoCo integrates.

To be explicit about the earlier thread on this line: passing the authored `maxhullvert` through is correct, and this holds up. Whenever the value is authored, the import matches native MuJoCo exactly. The mismatch is only in the fallback.

On a 162 vertex icosphere at `density="1000"`:

| asset `maxhullvert` | native MuJoCo | this branch |
| --- | --- | --- |
| unset | 505.880587 | 472.310100 |
| `10` | 240.450373 | 240.450373 |
| `64` | 472.310100 | 472.310100 |
| `-1` | 505.880587 | 505.880587 |

That is a 6.6% mass error on the unset row, which is the common case for real assets.

The same coupling also lets a collision setting move the physics: importing that model with `add_mjcf(..., mesh_maxhullvert=256)` yields 505.880587 against 472.310100 at the default, so tuning collision cost silently changes body mass.

Suggested fix: keep the authored value when the asset sets `maxhullvert`, and pass an unlimited budget to the inertia hull when it does not, which keeps the collision default a collision concern. The `-1` row above shows that unlimited already reproduces MuJoCo.

A native comparison with an unset `maxhullvert` and a hull above 64 vertices would lock this in. The current convex tests either author `maxhullvert="12"` or use the 16 vertex two box fixture, so both sit on the matching side of this boundary.

<details>

<summary>Reproduction</summary>

```py
import os
import tempfile

import mujoco
import trimesh

import newton

with tempfile.TemporaryDirectory() as d:
    mesh_path = os.path.join(d, "sphere.obj")
    trimesh.creation.icosphere(subdivisions=2, radius=0.5).export(mesh_path)  # 162 vertices

    for attrib in ("", ' maxhullvert="10"', ' maxhullvert="64"', ' maxhullvert="-1"'):
        mjcf = f"""
<mujoco>
  <asset><mesh name="s" file="{mesh_path}" inertia="convex"{attrib}/></asset>
  <worldbody>
    <body name="b"><freejoint/><geom type="mesh" mesh="s" density="1000"/></body>
  </worldbody>
</mujoco>
"""
        native = mujoco.MjModel.from_xml_string(mjcf)
        builder.add_mjcf(mjcf)
        print(f"{attrib or '(unset)':<22} native={native.body_mass[1]:10.6f}  newton={builder.body_mass[0]:10.6f}")

# (unset)                native=505.880587  newton=472.310100   <-- 6.6% off
#  maxhullvert="10"      native=240.450373  newton=240.450373
#  maxhullvert="64"      native=472.310100  newton=472.310100
#  maxhullvert="-1"      native=505.880587  newton=505.880587
```

</details>

han-xudong: Implemented in `01e53d6c`. The importer now keeps the collision and inertia hull budgets separate: an authored `maxhullvert` still applies to both, while an unset value leaves Newton's `mesh_maxhullvert` collision-only and gives convex inertia MuJoCo's unlimited default (`-1`).

I added a native MuJoCo regression using a 162-vertex icosphere with collision budgets 64 and 256. Both imports now match MuJoCo's mass, COM, and inertia while retaining their respective collision budgets. The branch is also updated to the latest `upstream/main`; the related 14 tests and full pre-commit suite pass.

Issue #3515, opened 2026-07-14
Support full-surface rigid-soft contacts in SolverCoupledProxy two-way coupling


AnkaChan:
### Summary

The opt-in full-surface rigid-soft SDF contact feature (#3262) detects soft-mesh **edge and face** contacts against rigid shapes and feeds them to `SolverVBD` in the **standalone** solve. It is **not yet supported in the `SolverCoupledProxy` two-way (staggered proxy) coupling** path, so a soft body cannot yet drive a proxy body through edge/face contacts. This was explicitly deferred as a scoped follow-up during the #3262 review.

### Current behavior

`SolverVBD.coupling_prepare_proxy_contacts` (`newton/_src/solvers/vbd/solver_vbd.py`) raises `NotImplementedError` when the `Contacts` buffer carries full-surface (edge/face) records — so the cross-solver wrench harvest can't *silently* drop that edge/face feedback:

```
NotImplementedError: ... full-surface rigid-soft contacts are not yet supported with the
SolverCoupledProxy two-way coupling. Set enable_rigid_soft_full_surface_contact=False for the
coupled proxy solve, or ...
```

Note: **standalone** `SolverVBD` already consumes edge/face contacts (the per-body reaction via `accumulate_body_particle_contacts_per_body` → `_eval_soft_ef_contact`). Only the multi-solver proxy-harvest feedback path is missing.

### Proposed work

Extend the proxy wrench-harvest so it maps edge/face contact reactions onto the proxy bodies, the same way per-particle contacts are harvested today:

- Harvest the per-body reaction wrench from edge/face records at the barycentric contact point (`Σ bary_i · particle_q[corner_i]`), not just per-particle contacts.
- Carry the unified records (`soft_contact_indices` / `soft_contact_barycentric`) through the coupled contact plumbing into the proxy harvest (the per-entry contact filter was already fixed for this in #3262 — see `_filter_soft_contacts_global_shape_ids_kernel`).
- Remove the `coupling_prepare_proxy_contacts` `NotImplementedError` guard once the harvest handles edge/face.
- Add a two-way regression: a soft-body edge/face contact drives a proxy body, and the reaction matches the standalone per-body result.

### References

- Feature PR: #3262 (full-surface rigid-soft SDF contact)
- Guard: `SolverVBD.coupling_prepare_proxy_contacts`, `newton/_src/solvers/vbd/solver_vbd.py`
[...]

PR #3756, opened 2026-07-31
Support full-surface rigid-soft contacts in VBD proxy coupling

mmichelis:
## Description

Adds full-surface edge and face rigid-soft contact support to VBD proxy-body coupling under `SolverCoupledProxy`.

VBD proxy-body harvesting now consumes unified `soft_contact_indices` and `soft_contact_barycentric` records through the same contact evaluation used by standalone VBD. Existing particle-contact behavior is unchanged.

For a shared contact buffer passed to `SolverCoupled`, per-entry filtering keeps full-surface records for capable solvers when the entry owns every referenced corner. Unsupported solvers receive particle contacts only, and records spanning multiple entries are dropped. This allows mixed solvers to share one full-surface `CollisionPipeline` without duplicate collision memory or runtime.

Proxy-local collision-pipeline contacts operate on the destination view and bypass shared per-entry filtering. VBD supports full-surface proxy-body contacts through either path. VBD proxy-particle harvesting remains particle-only and rejects full-surface contact buffers.

## Checklist

## Test plan

```bash
uv run --extra dev -m unittest newton.tests.test_solver_vbd_proxy_full_surface newton.tests.test_solver_vbd newton.tests.test_coupled_solver
```

All 184 tests pass on CPU and CUDA: 6 proxy full-surface tests, 84 VBD tests, and 94 coupled-solver tests.

Coverage includes particle, edge, and face proxy-body reactions, graceful per-entry filtering for mixed solvers, proxy-particle rejection, and an end-to-end detected face contact propagated through feedback to source-body motion.

## New feature / API change

A shared full-surface pipeline can be passed directly to the coupled solver:

```python
    model,
    broad_phase="nxn",
    enable_rigid_soft_full_surface_contact=True,
)

pipeline.collide(state_in, contacts)
solver.step(state_in, state_out, control, contacts, dt)
```

Alternatively, configure a full-surface `CollisionPipeline` on the VBD proxy destination. Proxy-local contacts bypass shared per-entry filtering.


mmichelis:
Good catch — added `test_edge_contact_reacts_on_proxy_body`.

One adjustment to the suggestion: `corners=[0, 1, -1]` would not have been a meaningful test here, because those two vertices lie in the `z = 0` plane and the contact normal is `+z`, so the penetration depth is independent of the barycentric weights. The test would pass even for a harvest that ignored `bary`. The new case uses `[0, 2, -1]`, spanning the vertex deliberately lifted off that plane, so a corner-0-only evaluation would see 4x the penetration and fail.

This also surfaced a bug in the test helper, which derived the expected contact point from the first N vertices rather than the ones at the record's actual corner ids. The two coincided for the existing particle and face records and diverge for an edge that skips a corner. Fixed.

Checking each record kind against a corner-0-only evaluation:

```
particle  |F|=500.0  pen(bary)=0.0500  pen(corner0)=0.0500  discriminating=False
edge      |F|=500.0  pen(bary)=0.0500  pen(corner0)=0.2000  discriminating=True
face      |F|=500.0  pen(bary)=0.0500  pen(corner0)=0.1000  discriminating=True
```

`particle` is expected to show `False`: for a particle record the contact point is corner 0, so there is nothing to discriminate.

CHANGELOG.md
```diff
@@ -48,6 +49,7 @@
 ### Changed
 
 - Decide collider visibility from USD `purpose` and visibility rather than from a bound render material. A collider whose `purpose` resolves to `default` is viewport geometry and is drawn; mark it `guide` to state that it is collision-only. Previously an unrelated visual elsewhere in the scene could make a collider vanish. `force_show_colliders` and `hide_collision_shapes` are unchanged.
+- Reject full-surface (edge/face) soft contacts on the contact buffer passed to `SolverCoupled`, which previously accepted them and silently dropped the edge/face records. Build that pipeline with `enable_rigid_soft_full_surface_contact=False` and enable the flag on the proxy's `collision_pipeline` instead.
```
gdaviet: How hard would it be to add filtering based on the 4 corners? If this not too much code I would be tempted to add this as part of this PR, otherwise this can be left to a follow-up

gdaviet (dismissed):
Looks good to me!

CHANGELOG.md
mmichelis: I had my agent mention this before as well, but I thought full surface contact is something VBD specific. It isn't. So I folded it in now, the changes were smaller than expected: 17 lines in `solver_coupled.py` + some lines in tests.

The filter now validates each corner the record references (up to 3: `(p,-1,-1)` particle, `(v0,v1,-1)` edge, `(v0,v1,v2)` face) instead of the single `soft_contact_particle`; `dst_indices`/`dst_barycentric` were already carried through. A record is kept only when the entry owns and activates every corner, so one straddling two entries is dropped by both rather than letting an entry read a particle it doesn't own.

With that, `enable_rigid_soft_full_surface_contact` on the buffer passed to the coupled solver reaches `SolverVBD`, and a proxy-local `collision_pipeline` is no longer required. 

One consequence worth your call: the capability flag is now propagated onto the per-entry buffer. `Contacts._assert_particle_only_soft_contacts` already existed in `SolverXPBD` / `SolverSemiImplicit` / `SolverStyle3D`, but was inert inside `SolverCoupled` because the filtered buffer never carried the flag. Those solvers now raise where they previously ran with edge/face records silently removed. Note the guard keys on the flag rather than on actually receiving an edge/face record, so a mixed scene whose XPBD entry only ever sees particle contacts will raise too. I'd keep it flag-based: rejecting on capability rather than content keeps the check out of the step loop. Happy to revisit if you'd prefer otherwise.

No guard needed for ADMM: it reads its own internal pipeline, constructed without the flag, so edge/face records can't reach it.

AnkaChan (commented):
Thanks @mmichelis — this cleanly implements the proxy-body half of #3515. I read through the harvest path and ran the new tests locally; the approach is right.

**Correctness looks good:**
- Reusing `_eval_soft_ef_contact` for the harvest is exactly right — `force_on_body = -force` applied at the returned `bx` matches standalone VBD's per-body reaction. Particle records `(p, -1, -1)` + `(1, 0, 0)` reduce byte-for-byte to the single-vertex case (checked against `_eval_body_particle_contact`: same penetration, `dx`, `bv`, `relative_translation`, force, and `cp_world = bx`), so particle-only behavior is unchanged. Momentum balances by construction (body `-F`; soft particles `bary_i·F` through the unchanged §2 path).
- Making the entry filter corner-aware (keep a record only when the entry owns every active corner) and carrying the capability marker so incapable sub-solvers (`SolverXPBD` / `SolverSemiImplicit` / `SolverStyle3D`, and proxy-particle VBD) fail loud instead of silently dropping edge/face records — nice, and consistent with the rest of the feature.
- Ran the new tests locally (harvest face/edge/particle, filter survive + straddling-drop, proxy-particle guard) — all pass.

**One non-blocking suggestion:** an end-to-end wiring test would be worth adding. The current tests are all unit-level — hand-seeded records and the filter in isolation, launching the harvest kernel directly. A minimal `SolverCoupledProxy` scenario — a soft edge/face (no vertex directly opposing the body) pressing on a free proxy-coupled rigid body, stepped a few frames, asserting the proxy body picks up velocity along the contact normal — would cover the detection → filter → harvest → body-motion path the unit tests skip.

Otherwise looks great.

CHANGELOG.md
gdaviet: Thanks! I my opinion dropping the full surface contacts for solver that don't support them is the graceful degradation, instead of raising. The alternative is that the other solver needs its own CollisionPipeline with full_surface off, the result being essentially the same, but with twice the CollisionPipeline memory and runtime costs.

mmichelis: I changed this to degrade per subsolver: the shared CollisionPipeline can keep full_surface enabled, while solvers that do not support full-surface contacts receive only particle contacts. VBD explicitly opts into the full contact representation. This avoids both raising and duplicating the collision pipeline, and I added coverage for the mixed-solver case. Let me know if this fits your expectations!

mmichelis:
> Thanks @mmichelis — this cleanly implements the proxy-body half of #3515. I read through the harvest path and ran the new tests locally; the approach is right.
> 
> **Correctness looks good:**
> 
> * Reusing `_eval_soft_ef_contact` for the harvest is exactly right — `force_on_body = -force` applied at the returned `bx` matches standalone VBD's per-body reaction. Particle records `(p, -1, -1)` + `(1, 0, 0)` reduce byte-for-byte to the single-vertex case (checked against `_eval_body_particle_contact`: same penetration, `dx`, `bv`, `relative_translation`, force, and `cp_world = bx`), so particle-only behavior is unchanged. Momentum balances by construction (body `-F`; soft particles `bary_i·F` through the unchanged §2 path).
> * Making the entry filter corner-aware (keep a record only when the entry owns every active corner) and carrying the capability marker so incapable sub-solvers (`SolverXPBD` / `SolverSemiImplicit` / `SolverStyle3D`, and proxy-particle VBD) fail loud instead of silently dropping edge/face records — nice, and consistent with the rest of the feature.
> * Ran the new tests locally (harvest face/edge/particle, filter survive + straddling-drop, proxy-particle guard) — all pass.
> 
> **One non-blocking suggestion:** an end-to-end wiring test would be worth adding. The current tests are all unit-level — hand-seeded records and the filter in isolation, launching the harvest kernel directly. A minimal `SolverCoupledProxy` scenario — a soft edge/face (no vertex directly opposing the body) pressing on a free proxy-coupled rigid body, stepped a few frames, asserting the proxy body picks up velocity along the contact normal — would cover the detection → filter → harvest → body-motion path the unit tests skip.
> 
> Otherwise looks great.

Thanks for the review and the suggestions! I added test_edge_face_reacts_through_coupled_proxy, which uses a shared full-surface collision pipeline with a SemiImplicit source body and a VBD destination. The scene produces only edge/face contacts, with gravity disabled. After two lagged coupling steps, the test verifies that the harvested contact reaction gives the source sphere positive normal velocity. This covers detection - per-entry filtering - VBD proxy harvest - feedback application - body motion in one test. It also exercises the updated graceful-degradation behavior: the SemiImplicit entry drops unsupported edge/face records while VBD retains them. The test passes on CPU and CUDA. Thanks again for pointing out the integration-coverage gap!

mmichelis:
Addressed the earlier CodeRabbit barycentric-preservation suggestion in `614129f4`. The retained face-record test now verifies that `[0.6, 0.3, 0.1]` survives per-entry filtering.

newton/tests/test_coupled_solver.py
```diff
@@ -3200,6 +3317,7 @@ def _coupled_soft_contact_filter_preserves_unified_fields(test, device):
             wp.array([int(ParticleFlags.ACTIVE)], dtype=wp.int32, device=device),  # particle_flags
             int(ShapeFlags.COLLIDE_PARTICLES),
             int(ParticleFlags.ACTIVE),
+            1,  # keep_full_surface_contacts
```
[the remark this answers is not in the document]
mmichelis: Good catch. Restored `keep_full_surface_contacts=0` in `614129f4` so this continues testing unified particle fields with full-surface support disabled.

AnkaChan:
Thank you for the revision! It looks good to me now!

eric-heiden (approved):

eric-heiden (approved):

eric-heiden (approved):

PR #3758, opened 2026-07-31
Add multi-detent actuator controller

lyd405121:
## Description

  This PR 
  - Adds `ControllerDetent`, a stateless actuator controller for
  mechanical switches with two or more stable positions. 
  
  - Add an example to show magnetic door using clamp actuator and 
   `ControllerDetent` for switches with three state

https://github.com/user-attachments/assets/7a3a6de7-8beb-42e4-acb9-3677aed9c3da

- If there is a more general way to do it, please let me know
- The controller evaluates a signed piecewise-linear effort curve from the
  current joint position and applies viscous damping:

  ```text
  effort = signed_lookup(position) - damping * velocity
  ```

  Stable detents use negative-slope zero crossings, while breakover points use
  positive-slope zero crossings. This produces snap-through behavior without
  maintaining a per-frame discrete detent state.

  The controller supports:

  - High-level configuration using detent positions, holding efforts, breakaway
    efforts, transition width, and optional breakover positions.

  - Low-level configuration using an explicitly authored signed lookup table.
  - Shared lookup curves with per-DOF damping.
  - CPU and CUDA execution.
  - CUDA graph capture.

  This PR also adds the basic_actuator example, which demonstrates:

  - A refrigerator door using ControllerPD with ClampingPositionBased to model
    position-dependent magnetic closing effort.

  - A three-detent switch with stable positions at -30, 0, and 30 degrees.
  - Interactive refrigerator-door dragging.
  - Arrow-key torque control and 1/2/3 detent selection for the switch.

  The public API reference, examples table, screenshot, and changelog are
  updated accordingly.

  ## Checklist

  - [x] CHANGELOG.md has been updated (if user-facing change)

  ## Test plan

  uv run --extra dev python -m unittest -v \
      newton.tests.test_actuators.TestControllerDetent

  Verified:

  - High-level detent curve generation.
  - Low-level signed lookup interpolation.
  - Stable and unstable effort directions.
  - Velocity damping.
  - Invalid parameter rejection.
  - Builder grouping with shared curves and per-DOF damping.
  - CPU and CUDA result consistency.

## Summary by CodeRabbit

- **New Features**
  - Added a multi-detent mechanical actuator controller with configurable detent positions, holding forces, breakaway forces, transition widths, and damping.
  - Added a basic actuator example demonstrating a magnetic door actuator and a three-detent switch.
  - Made the new controller available through the public actuators API.

- **Documentation**
  - Added the actuator example to the README.
  - Added the controller to the actuator API documentation.

- **Tests**
  - Added coverage for interpolation, damping, validation, and CPU/CUDA consistency.


jvonmuralt:
Thanks for the PR and sorry for the late reply!

We've now added a section on it: CONTRIBUTING.md → "Adding Control Laws and Actuator Components" (#3887). Newton core keeps the composable building blocks and the control laws that show up across many robots — PD, PID, neural policies, clamping, delay. A control law for a specific mechanism is better implemented on the user side. A multi-detent curve is really about assets in the environment (switches, door hinges, latches) rather than the robots themselves, so it fits better outside core, and I'm going to close this PR.

Instead, you could consider contributing just the example. ModelBuilder.add_actuator() takes any Controller subclass, and Controller is public, so you can define ControllerDetent right inside the example file and only change the import.

PR #3759, opened 2026-07-31
Give coupled-solver entry views private particle hash grids

caugonnet:
## Description

`SolverCoupled` entry model views carry the parent model's `particle_grid` over by reference, so every entry's sub-solver rebuilds and queries **one shared** `wp.HashGrid`. The grid is per-step solver scratch (each particle solver rebuilds it over its own particles immediately before querying it), not shared model data:

- **Serially** this works only through ordering luck — the shared grid's contents are valid only for whichever entry stepped last.
- **Concurrently** (multi-stream or CUDA-graph task parallelism across entries) it is a data race on the cell arrays and grid descriptor, observed as run-to-run nondeterministic particle states when two XPBD entries execute in parallel graph branches.

**Fix:** allocate a private `wp.HashGrid` per particle-carrying entry view at view construction (same resolution as `ModelBuilder.finalize`); particle-less entries get `particle_grid = None`, which existing solver-side guards already support (`SolverSemiImplicit` documents `None` as the disable mechanism).

**Example:** `coupled_concurrent_entries` steps two independent XPBD particle entries concurrently on private streams via a `_step_coupled` template-method override, with the whole frame (collision + both entry solves + fork/join events) captured into a single CUDA graph. `--stepping serial` runs the default sequential entry loop; both modes produce the same trajectories.

**Cost:** +16 MB per particle-carrying entry at construction (`wp.HashGrid.__init__` allocates its cell arrays eagerly). Acceptable since solver construction is cold-path. A follow-up proposal moves grid creation from `ModelBuilder.finalize` to the consuming solvers entirely (lazy for non-consumers, removes this cost and the aliasing bug class at the root); kept separate to keep this PR a minimal correctness fix.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_coupled_concurrent_entries
uv run --extra dev -m newton.tests -k test_admm_coupled_solver
python -m newton.examples coupled_concurrent_entries --viewer null --num-frames 120 --test
python -m newton.examples coupled_concurrent_entries --viewer null --num-frames 120 --test --stepping serial
python -m newton.examples admm_contact_solver --viewer null --num-frames 60 --test
```

New tests: ownership invariant (every entry grid non-None, distinct from the parent's and from each sibling's), and concurrent per-entry stepping on private streams matching serial stepping both eagerly and under a single CUDA graph capture (bit-exact on RTX 3080 Ti; asserted at 1e-5 for cross-GPU robustness). Existing ADMM suite (17 tests) passes unchanged.

## Bug fix

**Steps to reproduce:**

1. Build a `SolverCoupled` with two particle-carrying entries (e.g. two XPBD particle sets)
2. Inspect `entry.solver.model.particle_grid` for both entries: it is the same object as the parent `model.particle_grid`
3. Step both entries concurrently (each `_step_entry` on its own stream): particle states become run-to-run nondeterministic; serially, the grid's contents after a step reflect only the last-stepped entry's particles

## Summary by CodeRabbit

* **New Features**
  * Added an example demonstrating concurrent simulation of independent particle systems on CUDA streams.
  * Added CUDA graph capture and replay support for concurrent simulations.
  * Added configurable serial or concurrent stepping modes with rendering and validation.

* **Bug Fixes**
  * Prevented coupled particle systems from sharing particle-grid state during concurrent stepping.

* **Tests**
  * Added coverage for private grid ownership and serial versus concurrent results, including CUDA graph execution.


gdaviet (commented):
I am happy with those changes, but they have implications outside of the coupling setup. XPBD/SemiImplicit solvers will now use their own grid regardless of whether they're being coupled or not. I think this weakens the point for having a grid on the model at all, @eric-heiden what would be you opinion on this?

caugonnet:
@gdaviet a less invasive PR could have been to just replace the code that takes the parent's hash grid to create a new hash grid in every case, this is a pragmatic intermediate solution too (just wasting 16MB per sub solver)

gdaviet:
> @gdaviet a less invasive PR could have been to just replace the code that takes the parent's hash grid to create a new hash grid in every case, this is a pragmatic intermediate solution too (just wasting 16MB per sub solver)

You mean on the model view ? Yes maybe that's actually a more backward compatible solution 

I am all for moving the `particle_grid` away from the model, this is mutable data after all, but technically this would be a public-API-breaking change

PR #3760, opened 2026-07-31
Support masked deformable reset in SolverVBD.reset()

### LINKED ISSUE #3400: Support deformable (cloth and soft body) resets in SolverVBD.reset()
## Description

`SolverVBD.reset()` (tracked and added under #3256) resets rigid bodies, joints, and cables per world, but does not reset **deformables** (cloth and volumetric soft bodies). Particle `StateFlags.PARTICLE_Q` / `StateFlags.PARTICLE_QD` are currently ignored, and the method warns that particle resets are unsupported.

This requests first-class masked deformable reset support so cloth and soft-body particles can be restored to their model defaults per world, matching the rigid reset semantics.

### Requested behavior

- `reset()` honors `StateFlags.PARTICLE_Q` / `StateFlags.PARTICLE_QD`, copying `model.particle_q` / `model.particle_qd` into the state for particles in the worlds selected by `world_mask`.
- Same masking semantics as the rigid path: `world_mask=None` also resets global (`world == -1`) particles; an explicit mask excludes them.
- Works when an external solver integrates the bodies, and for particle-only models.
- Covers both cloth and volumetric (tet) soft bodies through a single path.

### Notes on scope

Unlike the rigid side, the particle side needs no solver-history cold-starting: `particle_q_prev` is rebaselined from the incoming state at the start of every `step()`, self-contact and body-particle contact buffers are rebuilt per step (no match-index warm-start path for soft contacts), and tet/cloth elasticity is stateless. So the deliverable is the masked state copy plus removing the unsupported-particle warning and documenting the behavior (including regenerating contacts after moving particles, and `rebuild_bvh` for large displacements with self-contact).

## Motivation / Use Case

- Reproducible RL rollouts, sweeps, and regression tests that drop cloth or a deformable and reset to an identical start.
- Completes the VBD reset story begun in #3256, whose primary deliverable was the rigid/joint/cable reset.

## Acceptance criteria

- [ ] `SolverVBD.reset()` resets cloth and soft-body particles per world via `PARTICLE_Q` / `PARTICLE_QD`.
- [ ] Global-particle vs explicit-mask semantics match the rigid path.
- [ ] Tests cover masked restore, flag selectivity, global-particle semantics, and a finite post-reset step for both cloth and tet soft bodies.

AnkaChan:
## Description

`SolverVBD.reset()` (added in #3316) resets rigid bodies, joints, and cables per
world but ignored deformables — particle `StateFlags` were ignored and it warned
that particle resets were unsupported. This adds first-class masked deformable
reset so cloth and volumetric soft-body particles are restored to model defaults
per world, matching the rigid-body semantics.

- `reset()` now honors `StateFlags.PARTICLE_Q` / `StateFlags.PARTICLE_QD`, copying
  `model.particle_q` / `model.particle_qd` into the state for particles in the
  worlds selected by `world_mask`.
- Masking mirrors the rigid path via the shared world-selection predicate:
  `world_mask=None` also restores global (`world == -1`) particles, while an
  explicit mask restores globals only through its final entry.
- One path covers both cloth and volumetric (tet) soft bodies, and it runs even
  when an external solver integrates the bodies or the model has no bodies.
- The unsupported-particle warning is removed and the behavior is documented.

Particle and body-particle solver history is intentionally left untouched:
`particle_q_prev` rebaselines from the incoming state at the start of every
`step()`, self-contact and body-particle contacts rebuild per step, and tet/cloth
elasticity is stateless — so no particle history cold-start is required. When
particle self-contact is enabled, reset rebuilds the self-contact BVH from the
restored positions so the discontinuous jump does not degrade tree quality (a
rebuild is not graph-capturable, so issue reset outside capture in that case).

Closes #3400.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -p 'test_solver_vbd.py'
```

Full `test_solver_vbd.py` module passes (89 tests). New tests:

- `test_reset_masked_rigid_and_soft` — fixed bodies + cloth + tet in shared worlds
  and the global range; one masked `reset()` restores both rigid and particle
  state together (per-field `*_Q` / `*_QD` isolation, `world_mask=None` includes
  globals, an explicit final-entry mask selects only globals, wrong-device
  validation).
- `test_soft_reset_particle_only_and_external` — particle-only model (no bodies)
  and `integrate_with_external_rigid_solver=True`.
- `test_soft_reset_captured_graph_restores_particles` — a captured `reset()`
  restores selected-world defaults on replay (graph-capture safe).
- `test_soft_reset_rebuilds_self_contact_bvh` — reset rebuilds the self-contact
  BVH once on a position-restoring reset, and not on a velocity-only or `flags=0`
  reset.

## New feature / API change

```python
import newton
import warp as wp

# Cloth / soft-body worlds; reset only world 0's particles to the model defaults.
model = builder.finalize()
state = model.state()

world_mask = wp.array([True, False, False], dtype=wp.bool, device=model.device)
solver.reset(
    state,
    world_mask=world_mask,
    flags=newton.StateFlags.PARTICLE_Q | newton.StateFlags.PARTICLE_QD,
)
# World 0 particles restored to model defaults; world 1 and globals untouched.
# world_mask=None restores every particle, the global (world -1) range included.
```

## Summary by CodeRabbit

* **New Features**
  * Added support for resetting selected cloth and volumetric soft-body particles to their initial positions and velocities.
  * Supports masked worlds, global particles, particle-only simulations, externally integrated rigid bodies, and CUDA graph replay.
  * Particle collision data is refreshed during the next simulation step after a reset.

* **Bug Fixes**
  * Corrected deformable-particle reset behavior across supported simulation configurations.
  * Improved handling of optional particle position and velocity data during resets.


eric-heiden:
🔵 Nice to see `SolverVBD.reset()` gain masked deformable support, and the captured-graph test that comes with it is a good addition.

One thing to clean up before merge: the new reset docstring describes a graph-capture limitation that does not match how the self-contact BVH rebuild actually behaves. Details inline.

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -1861,9 +1862,23 @@ def reset(
...
+        integrates the bodies or the model has none. A requested particle field is
+        skipped if its *state* array is ``None``. Particle and body-particle solver
+        history is intentionally left untouched: ``particle_q_prev`` is rebaselined
+        from the incoming state at the start of the next :meth:`step`, self-contact
+        and body-particle contacts rebuild per step, and tet/cloth elasticity is
+        stateless, so no particle history cold-start is required. When particle
+        self-contact is enabled, reset rebuilds the self-contact BVH from the
+        post-reset positions so the discontinuous jump does not degrade tree
+        quality; a rebuild is not graph-capturable, so issue reset outside capture
+        in that case.
```
eric-heiden: ⚪ This says a self-contact BVH rebuild is not graph-capturable, but that path is already captured today. `newton/examples/cloth/example_cloth_twist.py` and `newton/examples/cloth/example_cloth_rollers.py` both call `self.solver.rebuild_bvh(self.state_0)` from inside `wp.ScopedCapture()`, and `rebuild_bvh()` is the same `self.trimesh_collision_detector.rebuild(state.particle_q)` call this paragraph is describing. Both examples run as CUDA tests in `newton/tests/test_examples.py`.

Warp backs this up: `Bvh.rebuild()` is documented as reusing the existing BVH buffers, with the native CUDA LBVH rebuild path being capture safe, and the detector builds its trees with the default LBVH constructor on CUDA.

The practical cost of leaving this in: someone running cloth with `particle_enable_self_contact=True` will move episode resets out of their captured graph for no reason, which is the opposite of what the new captured-graph reset test shows for the non-self-contact case.

```suggestion
        stateless, so no particle history cold-start is required. When particle
        self-contact is enabled, reset rebuilds the self-contact BVH from the
        post-reset positions so the discontinuous jump does not degrade tree
        quality.
```

The matching code comment around lines 1978-1980 needs the same edit. If there is a real constraint you hit here, it would be more useful to state that specific constraint instead.

Separately, worth a thought: this rebuild covers both trees across all worlds even when `world_mask` selects a single world, and the next `step()` refits from `state_in.particle_q` regardless, so callers who want an explicit rebuild already have the public `rebuild_bvh()`. Making it skippable could be a cheap win.

AnkaChan:
Addressed in f822fe3b:
- Replaced the `:func:` cross-reference to the private `_reset_world_selected` helper with an inline code literal in `reset_particle_state`'s docstring.
- Added a symmetric `PARTICLE_Q` wrong-device assertion in `_reset_masked_rigid_and_soft` covering the guard line Codecov flagged.

@coderabbitai please re-review.

newton/_src/solvers/vbd/solver_vbd.py
eric-heiden: @AnkaChan could you address this? Then we can merge.

CHANGELOG.md
```diff
@@ -22,10 +22,11 @@
...
 - Add viewer layer system to overlay multiple solvers/models in supported rendering viewers; call `ViewerBase.activate(layer_id)` to route subsequent `set_model` / `log_state` / `log_*` calls into a named layer, `ViewerBase.set_layer_visible()` to toggle layers independently, and `ViewerBase.set_layer_transform()` to position layers side-by-side. See `example_basic_multi_solver_overlay.py`
 - Add `Heightfield.create_from_mesh()` and `newton.utils.rasterize_mesh_to_heightfield()` to build a heightfield collider by ray-casting a `wp.Mesh`, replacing a large static terrain mesh with an equivalent heightfield.
+- Add masked deformable reset to `SolverVBD.reset()`: `StateFlags.PARTICLE_Q` / `StateFlags.PARTICLE_QD` restore `model.particle_q` / `model.particle_qd` for cloth and volumetric soft-body particles in the worlds selected by `world_mask`, matching the rigid-body reset masking (global `world == -1` particles are restored when `world_mask=None` and excluded by an explicit local mask). When particle self-contact is enabled, the reset rebuilds the self-contact BVH from the restored positions. (#3400)
```
jcarius-nv: The project guidelines require user-facing changelog entries to be concise, avoid internal implementation detail, and be inserted at a random position within the category. This entry is long, exposes the BVH mechanism, and is placed directly beside the related rigid-reset entry. Please reduce it to the user-visible reset capability and move it within `Added`.

newton/tests/test_solver_vbd.py
```diff
@@ -2013,6 +2013,345 @@ def step_swap():
     np.testing.assert_allclose(solver.joint_lambda_lin.numpy(), 0.0)
 
 
+def _reset_masked_rigid_and_soft(test, device):
+    """A single masked reset restores rigid bodies and deformables together per world.
```
jcarius-nv: All four new registered tests use declarative opening summaries (`A single...`, `Deformable reset...`, `A masked...`, and `Reset rebuilds...`). The test guideline requires an imperative summary. Please update each one, for example by starting with `Verify ...`.

newton/tests/test_solver_vbd.py
```diff
@@ -2013,6 +2013,345 @@ def step_swap():
...
+        cell_x=0.1,
+        cell_y=0.1,
+        mass=1.0,
+    )
+    builder.color()
+    model = builder.finalize(device=device)
+
+    solver = newton.solvers.SolverVBD(model, iterations=0, particle_enable_self_contact=True)
+
+    # Count BVH rebuilds routed through the detector.
```
jcarius-nv: This comment only restates what the following monkeypatch does. Per the project comment guideline, please remove it or replace it with the non-obvious reason the rebuild interception is needed.

newton/tests/test_solver_vbd.py
```diff
@@ -2013,6 +2013,345 @@ def step_swap():
...
+    model_bq = model.body_q.numpy()
+    model_bqd = model.body_qd.numpy()
+    model_pq = model.particle_q.numpy()
+    model_pqd = model.particle_qd.numpy()
+    body_selected = body_world == 0
+    body_global = body_world < 0
+    part_selected = particle_world == 0
+    part_global = particle_world < 0
+
+    solver = newton.solvers.SolverVBD(model, iterations=0)
```
jcarius-nv, line 2088: Issue #3400 explicitly requires a finite post-reset step for both cloth and tet soft bodies. This test constructs both but uses `iterations=0` and only verifies array copies; none of the added reset tests calls `solver.step()` after reset. Please add finite post-reset step coverage for both deformable types.

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -1861,9 +1862,23 @@ def reset(
...
-        is unchanged.
+
+        ``PARTICLE_Q`` / ``PARTICLE_QD`` likewise copy ``model.particle_q`` /
+        ``model.particle_qd`` into *state* for particles in the selected worlds,
+        using the same masking as the body fields (``world_mask=None`` also
+        restores global ``world == -1`` particles; an explicit mask restores
+        globals only through its final entry). One path covers both cloth and
+        volumetric (tet) soft bodies, and it runs even when an external solver
+        integrates the bodies or the model has none. A requested particle field is
+        skipped if its *state* array is ``None``. Particle and body-particle solver
```
jcarius-nv, line 1873: This wording conflates rebuilding solver-side contact state with regenerating contact geometry. `step()` consumes the externally supplied `Contacts`; it does not rerun `CollisionPipeline.collide()`. Since reset now moves particles, please tell callers to regenerate contacts after moving bodies or particles so stale soft contacts are not reused.

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -1918,12 +1935,50 @@ def reset(
...
+                    device=self.device,
+                )
+
+            if particle_q is not None and self.particle_enable_self_contact:
+                # A reset teleports particles discontinuously; refitting the
+                # self-contact BVH across that jump keeps a valid but low-quality
+                # tree, so rebuild it from the post-reset positions. Rebuild is not
+                # graph-capturable, so issue reset outside capture when self-contact
+                # is enabled (as with any episode reset).
+                self.trimesh_collision_detector.rebuild(state.particle_q)
```
jcarius-nv: Issue #3400 asks to document caller-directed `rebuild_bvh()` use for large displacements, while this makes every self-contact position reset perform a full rebuild across both trees and all worlds—even for a small or single-world reset. Please either keep rebuilding caller-directed or make it skippable; if automatic rebuilding is intentional, document and justify the added batched-reset cost.

jcarius-nv (commented):
Standards and specification issues are noted inline. The existing unresolved thread on `solver_vbd.py:1881` correctly covers the separate false claim that the self-contact BVH rebuild is not graph-capturable; that claim is contradicted by the pinned Warp contract, Newton’s captured cloth examples, and exact-head CUDA capture/replay, so that thread also needs resolution.

eric-heiden (approved):
Thanks Anka!

newton/_src/solvers/vbd/solver_vbd.py
AnkaChan: Thanks — right on both counts, and sorry for the incorrect claim. Confirmed `rebuild_bvh()` runs inside `wp.ScopedCapture()` in `example_cloth_twist.py` and `example_cloth_rollers.py` (both CUDA-tested), so the "not graph-capturable" wording is removed. I also took the skippable suggestion all the way: reset no longer rebuilds the BVH at all — refresh is caller-directed via the public `rebuild_bvh()`, which the docstring now points to for large post-reset displacements. Fixed in c3953696.

newton/_src/solvers/vbd/solver_vbd.py
AnkaChan: Agreed — removed the automatic rebuild. Reset no longer touches the BVH; the next `step()` refits from the incoming positions, and callers use `rebuild_bvh()` for large displacements (now documented). c3953696

newton/_src/solvers/vbd/solver_vbd.py
AnkaChan: Fixed: the docstring now states that `step()` consumes the supplied contacts and does not rerun `CollisionPipeline.collide()`, and to regenerate contacts after moving bodies *or* particles so stale soft contacts are not reused. c3953696

newton/tests/test_solver_vbd.py
AnkaChan: Added `test_soft_reset_then_step_advances_cloth_and_tet`: it resets a perturbed cloth+tet model back to the model defaults, runs a finite `step()`, and asserts reset-then-step reproduces a fresh step from the defaults (so no pre-reset state leaks through particle history), verifying both deformable types advance. c3953696

newton/tests/test_solver_vbd.py
AnkaChan: Done — the new test summaries now open with an imperative "Verify …". c3953696

newton/tests/test_solver_vbd.py
AnkaChan: That test (`_soft_reset_rebuilds_self_contact_bvh`) was removed along with the auto-rebuild behavior, so the comment is gone. c3953696

CHANGELOG.md
AnkaChan: Shortened to the user-visible reset capability and relocated it within `Added`, away from the rigid-reset entry. c3953696

PR #3761, opened 2026-07-31
Fix dropped MJC joint equality from NewtonMimicAPI

andrewkaufman:
## Description

`MjcEqualityJointAPI` builds on `NewtonMimicAPI`, and mujoco 3.10 deprecates `mjc:target`, `mjc:coef0`, and `mjc:coef1` in favor of `newton:mimicJoint`, `newton:mimicCoef0`, and `newton:mimicCoef1` (see `mjcPhysics/equalityJointAPI.h`).

`ModelBuilder.add_usd()` never read the current spellings for equality joints:

- `_parse_mjc_equality_constraints()` resolved the leader only from `mjc:target`, skipping the prim with a warning when it was absent.
- The `NewtonMimicAPI` loop skips any prim carrying `MjcEqualityJointAPI`, on the assumption the equality path owns it.

An asset authoring only the current properties matches neither path, so the constraint is dropped silently. This did not surface earlier because the pinned `newton-assets` fixtures were produced by `mujoco-usd-converter` v0.1.0a8, which still authored `mjc:target`. Converter 0.4.0 stopped authoring the deprecated aliases, so any asset converted with 0.4.x hits this.

Both paths now resolve the leader joint and coefficients through a shared `_resolve_newton_mimic()` helper, which also puts the degrees-to-radians conversion for an angular follower in one place instead of two. The deprecated `mjc:` names remain as a fallback so assets authored before the properties moved to the `newton:` namespace keep loading.

Note this is specific to the equality path: `SchemaResolverNewton` already covers the attribute-level deprecations (`newton:contactMargin`, `newton:maxHullVertices`, `newton:torsionalFriction`, and so on). `mjc:target` is a *relationship*, which `SchemaResolver.SchemaAttribute` cannot express, and the equality parser reads prim properties directly rather than going through a resolver.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_import_usd -k test_equality_constraints
```

481 tests pass. The new test fails without the fix (`AssertionError: 0 != 1`) and passes with it. `test_mjc_equality_joint_parsing` continues to cover the deprecated spelling, so both are pinned.

Also verified end to end against `newton-assets` reconverted with `mujoco-usd-converter` 0.4.1: `TestMenagerieUsdImport` goes from 1 failure (`test_import_robotiq_2f85_v4`, `equality_constraint_count` 2 != 3) to all 35 passing.

## Bug fix

**Steps to reproduce:**

1. Author a `UsdPhysics.RevoluteJoint` with `MjcEqualityJointAPI` applied.
2. Supply the leader joint via `newton:mimicJoint` only, with no `mjc:target`.
3. Import with `ModelBuilder.add_usd()` and observe `Model.mujoco.equality_constraint_count` is 0.

Equivalently, import any `usd_structured` asset converted with `mujoco-usd-converter` 0.4.x: a Robotiq 2F-85 gripper yields 2 equality constraints instead of 3, leaving the left and right driver joints uncoupled.

**Minimal reproduction:**

```python
import newton
from pxr import Sdf, Usd, UsdGeom, UsdPhysics

# ... define /World/Articulation with Root, Link1, Link2 and joints Joint1, Joint2 ...

joint2_prim = stage.GetPrimAtPath("/World/Articulation/Joint2")
joint2_prim.SetMetadata(
    "apiSchemas", Sdf.TokenListOp.Create(prependedItems=["MjcEqualityJointAPI", "NewtonMimicAPI"])
)
joint2_prim.CreateRelationship("newton:mimicJoint").SetTargets(["/World/Articulation/Joint1"])
joint2_prim.CreateAttribute("newton:mimicCoef0", Sdf.ValueTypeNames.Float).Set(90.0)
joint2_prim.CreateAttribute("newton:mimicCoef1", Sdf.ValueTypeNames.Float).Set(1.5)

builder.add_usd(stage, convert_mjc_equality_constraints=False)
model = builder.finalize()

print(model.mujoco.equality_constraint_count)  # 0 before this PR, 1 after
```

See `test_mjc_equality_joint_parsing_newton_mimic_properties` for the complete stage.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved USD import handling for MuJoCo equality-joint constraints using Newton mimic properties.
  * Added fallback support for legacy MuJoCo attributes.
  * Correctly converts revolute joint offsets from degrees to radians while preserving polynomial coefficients.
  * Improved leader-joint resolution and validation during import.
  * Respects settings that disable mimic behavior while retaining constraint records.

* **Tests**
  * Added regression coverage for Newton mimic-based equality-joint imports, conversions, and joint mapping.

### DIFF
--- a/newton/_src/utils/import_usd.py
+++ b/newton/_src/utils/import_usd.py
@@ -4511,6 +4511,7 @@ def parse_usd(
 
         coef0 = usd.get_attribute(joint_prim, "newton:mimicCoef0")
         if coef0 is None:
+            # The deprecated alias was always authored in radians, so it skips the conversion.
             coef0 = usd.get_attribute(joint_prim, "mjc:coef0", default=0.0)
         elif joint_prim.IsA(UsdPhysics.RevoluteJoint):
             coef0 *= DegreesToRadian
--- a/newton/_src/utils/import_usd.py
+++ b/newton/_src/utils/import_usd.py
@@ -4479,6 +4479,47 @@ def parse_usd(
                         existing_filter_pairs.add(pair)
                         builder.add_shape_collision_filter_pair(*pair)
 
+    def _resolve_newton_mimic(joint_prim: Usd.Prim) -> tuple[Sdf.Path | None, float, float]:
+        """Resolve the mimic leader joint and coefficients from a follower joint prim.
+
+        ``MjcEqualityJointAPI`` builds on ``NewtonMimicAPI``, so the equality and the plain
+        mimic import paths read the same properties through here. The deprecated
+        ``mjc:target``, ``mjc:coef0``, and ``mjc:coef1`` aliases are honored as a fallback
+        for assets authored before those properties moved to the ``newton:`` namespace.
+
+        ``newton:mimicCoef0`` is authored in the follower's position units, so a revolute
+        follower is converted from degrees into the joint coordinates the constraint is
+        evaluated in; the deprecated ``mjc:coef0`` is already in radians. ``coef1`` is
+        dimensionless. A multi-DOF follower has no defined unit, so its offset is passed
+        through unconverted and callers warn about it.
+
+        Returns:
+            The leader joint path, or ``None`` when no target is authored, followed by
+            ``coef0`` in joint coordinates and the dimensionless ``coef1``.
+        """
+        mimic_rel = joint_prim.GetRelationship("newton:mimicJoint")
+        targets = mimic_rel.GetTargets() if mimic_rel and mimic_rel.HasAuthoredTargets() else []
+        if not targets:
+            target_rel = joint_prim.GetRelationship("mjc:target")
+            targets = target_rel.GetTargets() if target_rel else []
+
+        leader_path = None
+        if targets:
+            leader_path = targets[0]
+            if not leader_path.IsAbsolutePath():
+                leader_path = joint_prim.GetPath().GetParentPath().AppendPath(leader_path)
+
+        coef0 = usd.get_attribute(joint_prim, "newton:mimicCoef0")
+        if coef0 is None:
+            coef0 = usd.get_attribute(joint_prim, "mjc:coef0", default=0.0)
+        elif joint_prim.IsA(UsdPhysics.RevoluteJoint):
+            coef0 *= DegreesToRadian
+        coef1 = usd.get_attribute(joint_prim, "newton:mimicCoef1")
+        if coef1 is None:
+            coef1 = usd.get_attribute(joint_prim, "mjc:coef1", default=1.0)
+
+        return leader_path, float(coef0), float(coef1)
+
     # Parse MjcEquality constraints *before* collapsing fixed joints so that the
     # builder's collapse logic can remap body/joint indices and adjust anchors/relposes
     # for any bodies that get merged.
@@ -4638,16 +4679,15 @@ def parse_usd(
                     )
                     continue
 
-                target_rel = joint_prim.GetRelationship("mjc:target")
-                targets = target_rel.GetTargets() if target_rel else []
-                if not targets:
+                leader_path, coef0, coef1 = _resolve_newton_mimic(joint_prim)
+                if leader_path is None:
                     warnings.warn(
-                        f"MjcEqualityJointAPI on '{joint_path}' has no mjc:target relationship; skipping.",
+                        f"MjcEqualityJointAPI on '{joint_path}' has no newton:mimicJoint relationship; skipping.",
                         stacklevel=2,
                     )
                     continue
 
-                target_path = str(targets[0])
+                target_path = str(leader_path)
                 joint2_idx = path_joint_map.get(target_path)
                 if joint2_idx is None:
                     warnings.warn(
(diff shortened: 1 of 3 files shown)


newton/_src/utils/import_usd.py
```diff
@@ -4479,6 +4479,47 @@ def _resolve_collision_shape_ids(path: str) -> tuple[list[int], str | None]:
...
+        if targets:
+            leader_path = targets[0]
+            if not leader_path.IsAbsolutePath():
+                leader_path = joint_prim.GetPath().GetParentPath().AppendPath(leader_path)
+
+        coef0 = usd.get_attribute(joint_prim, "newton:mimicCoef0")
+        if coef0 is None:
+            coef0 = usd.get_attribute(joint_prim, "mjc:coef0", default=0.0)
+        elif joint_prim.IsA(UsdPhysics.RevoluteJoint):
+            coef0 *= DegreesToRadian
```
andrewkaufman, line 4517: this logic is wrong, the legacy `mjc:coef0` was always in radians

andrewkaufman: Added a comment in 636e8623f to make this read clearly — the conversion is in an `elif`, so it only applies to the value read from `newton:mimicCoef0`. The deprecated `mjc:coef0` fallback lands in the `if` branch and is used as-is, in radians.

`test_mjc_equality_joint_parsing` pins that path: revolute follower, `mjc:coef0 = 0.5` authored with no `newton:mimicCoef0`, asserting `polycoef == [0.5, 1.5, 0.1, 0.05, 0.02]`. It passes unchanged.

eric-heiden:
🔵 Thanks for tightening this up. Routing `MjcEqualityJointAPI` through `_resolve_newton_mimic` means the equality path and the plain mimic path now read the same authored properties, and couplings that used to be dropped are preserved.

One gap to close: the equality path picks up the `NewtonMimicAPI` relationship and coefficients but not `newton:mimicEnabled`, so an author's opt-out is honored on one path and silently ignored on the other. Details inline.

newton/_src/utils/import_usd.py
```diff
@@ -4638,16 +4680,15 @@ def add_converted_loop_joint(
...
-                targets = target_rel.GetTargets() if target_rel else []
-                if not targets:
+                leader_path, coef0, coef1 = _resolve_newton_mimic(joint_prim)
+                if leader_path is None:
                     warnings.warn(
-                        f"MjcEqualityJointAPI on '{joint_path}' has no mjc:target relationship; skipping.",
+                        f"MjcEqualityJointAPI on '{joint_path}' has no newton:mimicJoint relationship; skipping.",
                         stacklevel=2,
                     )
                     continue
```
eric-heiden, line 4689: 🔵 `newton:mimicEnabled` is ignored on this path.

This branch now reads `NewtonMimicAPI` through `_resolve_newton_mimic`, but the constraint's active state still comes only from `enabled = bool(joint_desc.jointEnabled)` (line 4583). The loop that does read `newton:mimicEnabled` (line 4877) skips any prim carrying `MjcEqualityJointAPI` (line 4875), so nothing consumes that property for these joints.

For a joint authored with `newton:mimicEnabled = false`:

| Authoring | Result |
| --- | --- |
| `NewtonMimicAPI` alone | no constraint, opt-out honored |
| `MjcEqualityJointAPI` + `NewtonMimicAPI` | equality constraint created and active |

This also bites on the default path: `convert_mjc_equality_constraints` defaults to `True` in `ModelBuilder.add_usd()`, so the same asset additionally gets an active generic mimic constraint and the unwanted coupling is enforced for every solver, not only `SolverMuJoCo`. On `main` the opt-out appeared to work here only because the constraint was dropped entirely, which is the behavior this change fixes.

Both constraint types already expose a runtime `enabled` flag, so the opt-out can be folded into it inside the `is_eq_joint` branch instead of skipping the prim:

```python
eq_enabled = enabled and bool(usd.get_attribute(joint_prim, "newton:mimicEnabled", default=True))
```

Then pass `eq_enabled` to `mjc_add_equality_mimic` and `_add_equality_constraint` in place of `enabled`, so one property governs both spellings.

andrewkaufman: Good catch — confirmed and fixed in 191045535.

You are right that this was latent rather than new: with `mjc:target` absent the constraint was dropped outright, so the opt-out only appeared to work. For legacy assets that do author `mjc:target` the property was already being ignored on this path.

Took your suggested shape — `newton:mimicEnabled` is folded into the runtime enabled flag inside the `is_eq_joint` branch rather than skipping the prim, so the constraint is still described and can be re-enabled later:

```python
eq_enabled = enabled and bool(usd.get_attribute(joint_prim, "newton:mimicEnabled", default=True))
```

and `eq_enabled` is passed to both `mjc_add_equality_mimic` and `_add_equality_constraint`.

Added `test_mjc_equality_joint_parsing_honors_mimic_enabled`, which covers both branches since the default path is the one that also lowers a generic mimic constraint:

- `convert_mjc_equality_constraints=False` -> `equality_constraint_count == 1`, `equality_constraint_enabled[0]` false
- `convert_mjc_equality_constraints=True` -> additionally `constraint_mimic_count == 1`, `constraint_mimic_enabled[0]` false

It fails without the fix with `AssertionError: True is not false`. 741 tests across `test_import_usd`, `test_equality_constraints`, and `test_mujoco_solver` pass with it.

eric-heiden (approved):
Approving. Both review threads are resolved and the follow-up commits address what came up.

The `newton:mimicEnabled` handling landed the right way. Folding the opt-out into the runtime `enabled` flag inside the `is_eq_joint` branch keeps the constraint described in the model so it can be re-enabled later, rather than skipping the prim and dropping it. Routing both spellings through `_resolve_newton_mimic` means the equality path and the plain mimic path now read the same authored properties. `test_mjc_equality_joint_parsing_honors_mimic_enabled` covers the default conversion path too, which is the one that would otherwise lower an active generic mimic constraint and enforce the coupling for every solver.

On the red GPU check: the only failure is `test_kamino.example_kamino_robot_anymal_d_cuda_0` failing its `body velocities are small` assertion. That is a pre-existing flake on `main`, not something this change causes. It hit `main` twice earlier the same day, reporting a different set of legs each time:

| Branch | Commit | Bodies reported |
| --- | --- | --- |
| `main` | `03e597a1` | `LF_SHANK, LF_FOOT, LH_SHANK, LH_FOOT` |
| `main` | `01d43e06` | `RF_SHANK, RF_FOOT` |
| this PR | `0999077` | `LF_HIP, LF_THIGH, LF_SHANK, LF_FOOT` |

It was the sole failure in all three runs. An earlier commit on this branch passed the same GPU suite, and the ANYmal-D example loads an asset with no mimic or equality constraints, so the import change never reaches that code path. I have restarted the GPU workflow on a fresh runner.

One optional follow-up, not worth holding the merge: the `_resolve_newton_mimic` docstring says a multi-DOF follower's offset "is passed through unconverted and callers warn about it", but only the plain mimic loop warns. The equality path stays silent. Either soften the wording to say which caller warns, or add the same warning there.

PR #3765, opened 2026-08-01
Fix follow camera resetting orbit every frame

### LINKED ISSUE #3764: [BUG] Follow-camera examples reset orbit/mouse rotation every frame (set_camera overwrites orientation)
### Bug Description

When a follow-camera example (e.g. `robot_anymal_c_walk`) is running with "Follow Camera" enabled, dragging the mouse to orbit the view has no lasting effect — the camera snaps back to a fixed viewing angle on the very next frame. Orbiting a moving target doesn't work at all: the camera either stays rigidly locked to one angle, or at some drag angles ends up facing away from the tracked robot instead of toward it.

**Root cause:** `example_robot_anymal_c_walk.py`'s `render()` calls `self.viewer.set_camera(pos=<robot position> + <fixed offset>, pitch=0.0, yaw=-180.0)` unconditionally, every frame. Mouse-drag orbiting (`ViewerGui.handle_mouse_drag` → `Camera.orbit()`) mutates `camera.pitch`/`camera.yaw` through a separate path, so the very next `set_camera()` call overwrites the user's orbit input back to the hardcoded angle. The camera's *position* is also recomputed from a fixed world-space offset rather than from the pivot/orbit distance the user set, so at some angles the camera doesn't even face the target anymore.

This isn't specific to this one example — the same "recompute pos/pitch/yaw from scratch every frame" pattern is used by ~70 other `set_camera()` call sites across `newton/examples/` and `newton/_src/solvers/kamino/examples/`, including the dedicated `CameraFollowRobot` helper in `_src/solvers/kamino/examples/rl/camera_follow.py`. Any of them would exhibit the same rigidity if used interactively with orbit enabled.

Expected behavior: dragging the mouse while a follow camera is active should orbit around the tracked target at a constant distance (like a standard third-person/orbit camera), not fight or reset on the next frame.

### Reproduction Script

```bash
python -m newton.examples robot_anymal_c_walk --viewer gl
```

With "Follow Camera" checked (default), try dragging the mouse to orbit around the robot while it walks. The view angle resets to the same fixed angle on the next frame instead of staying where you left it, and at extreme angles the camera ends up looking away from the robot.

### System Information

- Newton: 1.5.0.dev0
- Warp: 1.16.0.dev20260716
- Python: 3.12.12
- OS: Ubuntu 22.04 (Linux 6.8.0-136-generic)
- GPU: NVIDIA GeForce RTX 3060 Laptop GPU

vishcomestrue:
## Description

Follow-camera code recomputed the camera's position and orientation from scratch every frame, overwriting manual mouse-drag orbit input and, at some angles, pointing the camera away from the tracked target instead of toward it.

- Add `Camera.follow(target)` (`newton/_src/viewer/camera.py`) to re-center the pivot on a moving target while preserving whatever distance/angle the user last orbited to, composing correctly with the existing `Camera.orbit()`.
- Make `pitch`/`yaw` optional (`float | None = None`) on `ViewerBase.set_camera()` and its `ViewerGL`/`ViewerRTX`/`ViewerViser` overrides, so callers can update camera position without resetting orientation.
- Update `example_robot_anymal_c_walk.py` to use `Camera.follow()` on `ViewerGL`.

Scope notes:
- Only `ViewerGL` gets the actual orbit-follow fix, since it's the only backend backed by the shared `Camera`/pivot machinery. `ViewerRTX` gets the same optional-parameter API for consistency but I could not verify it visually in my environment. `ViewerViser`'s camera is browser-native and isn't backed by `Camera`, so its follow behavior is unchanged — `pitch`/`yaw` default to `0.0` when omitted, same as before.
- This is one of ~70 `set_camera()` call sites across `newton/examples/` and `newton/_src/solvers/kamino/examples/` that share the same "recompute every frame" pattern (including the `CameraFollowRobot` helper in `_src/solvers/kamino/examples/rl/camera_follow.py`). Only `example_robot_anymal_c_walk.py` is updated here as the reference case; the rest are unaffected and could adopt `Camera.follow()` in follow-up PRs.

Closes #3764

## Checklist

## Test plan

Manual verification only, no automated test added:

```
python -m newton.examples robot_anymal_c_walk --viewer gl
```

With "Follow Camera" checked, dragging the mouse now orbits around the robot at a constant distance while it walks, instead of resetting to a fixed angle on the next frame.

`ViewerRTX` and `ViewerViser` changes are untested in my environment; flagging for reviewer attention.

## Bug fix

1. `python -m newton.examples robot_anymal_c_walk --viewer gl`
2. With "Follow Camera" checked (default), drag the mouse to orbit around the walking robot.
3. Observe the view angle reset to the same fixed angle on the next frame instead of staying where dragged; at some angles the camera ends up facing away from the robot.

**Minimal reproduction:**

```python
# newton/examples/robot/example_robot_anymal_c_walk.py, render(), before this PR:
if self.follow_cam:
    self.viewer.set_camera(
        pos=wp.vec3(*self.state_0.joint_q.numpy()[:3]) + wp.vec3(10.0, 0.0, 2.0), pitch=0.0, yaw=-180.0
    )
```

Called every frame with hardcoded `pitch`/`yaw`, this overwrites any orientation change made via mouse-drag orbit (`Camera.orbit()`) on the previous frame.

## Summary by CodeRabbit

* **New Features**
  * Added camera-follow functionality to track moving targets while preserving viewing distance and orientation.
  * Camera positioning now allows pitch or yaw to be omitted, retaining the current orientation.
  * Updated viewer examples to demonstrate smooth camera tracking.

* **Bug Fixes**
  * Fixed follow-camera behavior so manual orbit distance and angles are preserved while tracking targets.

* **Documentation**
  * Clarified camera position units and angle measurements.


CHANGELOG.md
eric-heiden, line 1: Please use Towncrier changelog fragments rather than editing the changelog directly in this PR.

newton/_src/viewer/viewer_viser.py
```diff
@@ -547,19 +547,19 @@ def _on_camera_update(_camera: Any):
...
-            pos: Requested camera position.
-            pitch: Requested camera pitch angle.
-            yaw: Requested camera yaw angle.
+            pos: Requested camera position [m].
+            pitch: Requested camera pitch angle [deg]. Defaults to 0 if None.
+            yaw: Requested camera yaw angle [deg]. Defaults to 0 if None.
         """
         position = np.asarray((float(pos[0]), float(pos[1]), float(pos[2])), dtype=np.float64)
-        front, up_direction = self._compute_camera_front_up(pitch, yaw)
+        front, up_direction = self._compute_camera_front_up(pitch or 0.0, yaw or 0.0)
```
eric-heiden: 🔵 `ViewerViser.set_camera` does not follow the base-class contract for `pitch=None` / `yaw=None`.

`ViewerBase.set_camera` documents these defaults as "the current pitch/yaw is kept", and `ViewerGL` and `ViewerRTX` behave that way. Here `pitch or 0.0` / `yaw or 0.0` does the opposite: backend-agnostic code that calls `viewer.set_camera(pos)` to move the camera without touching its orientation preserves the view on GL and RTX, but snaps to pitch 0 / yaw 0 on Viser. That is the same class of surprise this PR is removing.

The base docstring in `newton/_src/viewer/viewer.py` (lines 775-779) currently records the divergence rather than resolving it. A contract with a per-subclass carve-out is hard for callers to rely on, so it seems better to make Viser honor it and drop the exception clause.

`pitch or 0.0` is also the wrong idiom for an explicitly optional float, since it treats a caller-supplied `0.0` the same as `None`. That is harmless today because both map to 0.0, but it will silently misbehave as soon as the fallback becomes a cached value.

Caching the last requested angles handles both points:

```suggestion
    def set_camera(self, pos: wp.vec3, pitch: float | None = None, yaw: float | None = None):
        """Set camera position and orientation for connected Viser clients.

        The requested view is also cached so that newly connected clients receive
        the same camera setup as soon as they report camera state.

        Args:
            pos: Requested camera position [m].
            pitch: Requested camera pitch angle [deg]. If None, the current pitch is kept.
            yaw: Requested camera yaw angle [deg]. If None, the current yaw is kept.
        """
        pitch = self._camera_pitch if pitch is None else float(pitch)
        yaw = self._camera_yaw if yaw is None else float(yaw)
        self._camera_pitch = pitch
        self._camera_yaw = yaw
        position = np.asarray((float(pos[0]), float(pos[1]), float(pos[2])), dtype=np.float64)
        front, up_direction = self._compute_camera_front_up(pitch, yaw)
```

This needs two small companion changes: initialize `self._camera_pitch = 0.0` and `self._camera_yaw = 0.0` next to `self._camera_request` in `__init__` (around line 158), and remove the `ViewerViser` exception clause from the `ViewerBase.set_camera` docstring in `newton/_src/viewer/viewer.py` so the base class states a single contract again.

vishcomestrue:
Thanks for the review! Addressed both points:

- Moved the changelog entry to a Towncrier fragment (`changelog/3764.fixed.md`) and rebased onto `main` to pick up the Towncrier migration.
- Fixed `ViewerViser.set_camera()` to cache `_camera_pitch`/`_camera_yaw` and reuse them when `None` is passed, matching `ViewerGL`/`ViewerRTX`. Removed the now-unneeded carve-out from the `ViewerBase` docstring so there's one contract again.

newton/examples/robot/example_robot_anymal_c_walk.py
```diff
@@ -179,6 +179,11 @@ def toggle_follow_cam(imgui):
 
         newton.eval_fk(self.model, self.state_0.joint_q, self.state_0.joint_qd, self.state_0)
 
+        if isinstance(self.viewer, newton.viewer.ViewerGL):
+            base_pos = wp.vec3(*self.state_0.joint_q.numpy()[:3])
+            self.viewer.set_camera(pos=base_pos + wp.vec3(10.0, 0.0, 2.0), pitch=0.0, yaw=-180.0)
+            self.viewer.camera.look_at(base_pos)
```
eric-heiden, line 185: ⚪ The `look_at(base_pos)` call on the next line recomputes both pitch and yaw from the position-to-target direction and reassigns the pivot, so the explicit `pitch=0.0, yaw=-180.0` here never take effect. With the position offset by (10, 0, 2), the camera ends up at roughly -11.3° pitch and 180° yaw either way. As written, the call reads as if those angles establish the initial framing, which is misleading for anyone copying this block into a new example.

Dropping them keeps the exact same view and doubles as a demonstration of the optional `pitch`/`yaw` arguments this PR introduces.

```suggestion
            self.viewer.set_camera(pos=base_pos + wp.vec3(10.0, 0.0, 2.0))
            self.viewer.camera.look_at(base_pos)
```

eric-heiden:
Hi @vishcomestrue, could you please fix the merge conflict and address the last comment I've added?

vishcomestrue:
Thanks — fixed the merge conflict (rebased onto main) and dropped the dead `pitch=0.0, yaw=-180.0` args in `example_robot_anymal_c_walk.py` since `camera.look_at(base_pos)` overwrites them anyway, as you pointed out. Ready for another look.

eric-heiden (approved):
Thank you!

PR #2276, opened 2026-03-31
Collision improvements that should minimize clipping artefacts

nvtw:
## Description

Try to improve clipping of convexes through triangle ground meshes and heightfields.

Should help to address https://github.com/isaac-sim/IsaacLab/issues/5071

## Checklist

- [x] New or existing tests cover these changes
- [x] The documentation is up to date with these changes
- [x] `CHANGELOG.md` has been updated (if user-facing change)

## Test plan

<!-- How were these changes verified? Include commands, test names,
     or manual steps. Example:
     ```
     uv run --extra dev -m newton.tests -k test_relevant_test
     ``` -->

## Bug fix

<!-- DELETE this section if not a bug fix.
     Describe how to reproduce the issue WITHOUT this PR applied. -->

**Steps to reproduce:**

https://github.com/camevor/newton/tree/repro-mesh-collision


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added a triangle-prism heightfield primitive and treated it as a discrete shape; support function and closest-point projection added to improve triangle/convex interactions.
  * Triangle pose now respects heightfield local rotation and triangle center projection improved for more robust contact resolution.

* **Bug Fixes**
  * Back-face culling ignores back-side contacts to prevent inverted normals and NaNs.
  * Improved MPR degeneracy handling for extreme/skinny triangles.

* **Tests**
  * Extensive CUDA-gated tests for back-face culling, heightfield prisms, extreme triangles, and simulation stability.

* **Documentation**
  * Clarified CCW winding and back-face culling requirements for meshes.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

PR #3766, opened 2026-08-02
Fix scale-dependent MPR contacts on large mesh triangles

### LINKED ISSUE #3773: [BUG] MPR can return metre-scale penetration and nonlocal witnesses for small convex shapes on large triangles
### Bug Description

In some poses, MPR turns a local contact between a small cylinder and a large triangle into a false, tens-of-metres penetration with witness points far from the cylinder. In this reproduction, the correct penetration is `6.17 cm`, but Newton returns `58.8 m`; only the overlap boolean is correct.

The vulnerable case is near-coplanar: the cylinder center is close to the plane, while its projection lies inside the face but far from the triangle centroid. The existing `1%` centroid bias then dominates the normal separation and makes the initial MPR direction nearly tangent to the face. Depending on pose, portal refinement can follow an ill-conditioned branch and reconstruct nonlocal contact data.

The deterministic reproduction uses:

- triangle vertices: `(0, 0, 0)`, `(100, 320, 0)`, `(0, 320, 0)` m;
- cylinder radius / half-height: `0.02 / 0.07` m;
- cylinder position: `(28.87991333, 260.53430176, 0.00036136055)` m;
- cylinder orientation `(x, y, z, w)`: `(-0.16498479, 0.36682746, -0.48718625, 0.77515626)`;
- `P` is inside the face and `28.9 m` from the nearest real edge;
- `h = 0.00036136055 m` and `|G - P| = 47.4105934 m`;
- the old seed therefore adds `0.474105934 m` tangentially: a `1312.002:1` tangential/normal ratio, or `89.9563°` from the face normal.

#### Observed versus expected

| | Analytic local cylinder–plane geometry | Newton `ee84505a` |
|---|---:|---:|
| overlap boolean | `True` | `True` |
| penetration | `0.061719150 m` | **`58.826744080 m`** |
| normal (Newton's triangle-to-convex convention) | `(0, 0, 1)` | `(3.6e-11, 0.988920, 0.148452)` |
| triangle witness | near `P` | approximately on/beyond the distant `y=320 m` edge, `60.19 m` from `P` |
| convex witness | on the cylinder | `8.73 m` below the plane and metres outside the cylinder |

The boolean is therefore not the bug; the returned witness points, normal, and penetration are.

#### Why only some poses fail

Holding the triangle, cylinder XY position, and cylinder orientation fixed while changing only `h` shows an abrupt transition between two adjacent `2 μm` samples:

| `h` | returned penetration | result |
|---:|---:|---|
| `0.488 mm` | `6.268174 m` | invalid/nonlocal |
| `0.490 mm` | `0.061591 m` | correct/local |

An instrumented MPR trace shows that both runs select the same sequence of support features through phase one. At the first phase-two region test, the value used to choose the replacement changes sign:

| `h` | first `dot(cross(v4, v0), v2)` | update | final barycentric weights | penetration |
|---:|---:|---|---|---:|
| `0.488 mm` | `-0.00559139` | replace `v3` | `(0.566406, 58395.5, -58395.0)` | `6.268174 m` |
| `0.490 mm` | `+0.00639486` | replace `v1` | `(0.288423, 0.525764, 0.185814)` | `0.061591 m` |

The bad branch continues for ten refinement iterations and terminates with huge extrapolating barycentric coefficients, which place both witnesses outside the local geometry. This instrumented trace supports the portal-branch explanation. The two adjacent samples demonstrate an abrupt, pose-sensitive transition; they do not prove a mathematical discontinuity, define a universal height threshold, or measure the failure frequency in a full simulation.

#### Root cause

The `1%` centroid bias added in #2276 has a valid purpose: when a projection lies on a shared mesh edge, adjacent triangles pull their MPR initialization toward different interiors, keeping their manifold witnesses distinct. Removing the bias makes this reproduction return correct local data but regresses that shared-edge behavior.

The problem is the fixed fraction of `|G-P|`, a length that can scale with the triangle. For an interior face projection, the normal component of the initial direction is local (`h`), while the added tangential component grows with `|G-P|`. This mismatch can create a nearly tangential vulnerable seed even though the actual contact is local and the nearest edge is tens of metres away.

For `h != 0`, the scale-free quantity that describes this vulnerability is
`R = 0.01 * |G-P| / |h|`: when `R >> 1`, the artificial tangent component dominates the initial direction. This reproduction has `R = 1312`. A large `R` does not guarantee failure—the convex orientation and resulting support features still determine which portal branch is selected—but it explains why the failure class generalizes beyond this one cylinder pose.

Here the `28.9 m` nearest-edge margin is over one thousand cylinder radii, so the finite triangle is locally equivalent to an infinite plane at the contact; no physical edge feature can explain the nonlocal result.

A complete fix also needs to cover the exact `h=0` boundary. Merely replacing the old bias with `0.01 * min(|G-P|, h)` makes the explicit bias zero there, but the float32 face-point reconstruction in this case leaves a `1.5258789e-05 m` tangential residual when two large, nearby coordinates are subtracted. That residual bypasses the existing coincident-center fallback. If the residual is removed, the generic fallback then prefers a long world-space triangle axis and loses the triangle-specific shared-edge direction.

#### Proposed fix

The accompanying draft PR keeps the shared-edge intent while removing triangle-scale amplification:

1. For an orthogonal projection in the **closed face region** (including shared edges), construct the base Minkowski center offset directly as `-d * n` from signed plane distance `d` and unit face normal `n`, instead of subtracting two large reconstructed positions.
2. When `|d| > 0`, add the centroid-directed tangent bias as
   `0.01 * min(|G-P|, |d|) * normalize(G-P)`.
   For face projections, its tangential/normal ratio is therefore at most `0.01`, independent of triangle size.
3. Handle `d=0` separately: when the resulting triangle–convex centers coincide, use the face-normal direction with the same small centroid-directed bias rather than the generic world-axis probe. This is the `h -> 0` limit of the bounded rule and preserves distinct shared-edge witnesses.
4. Keep real tangential distance for projections outside the face, and keep the existing degenerate-triangle fallback.

Regression coverage includes the original `58.8 m` case, exact coplanar center, shared-edge behavior both above and exactly on the plane, a dense `0–2 mm` height sweep, and the existing large/edge/vertex triangle tests.

Draft fix: #3766. Related: #2276 and isaac-sim/IsaacLab#5071.

### Reproduction Script

```python
import numpy as np
import warp as wp

from newton import GeoType
from newton._src.geometry.mpr import create_solve_mpr
from newton._src.geometry.support_function import GenericShapeData, GeoTypeEx, SupportMapDataProvider, support_map


@wp.kernel
def run_case(
    collision_out: wp.array(dtype=int),
    point_a_out: wp.array(dtype=wp.vec3),
    point_b_out: wp.array(dtype=wp.vec3),
    normal_out: wp.array(dtype=wp.vec3),
    penetration_out: wp.array(dtype=float),
):
    triangle = GenericShapeData()
    triangle.shape_type = int(GeoTypeEx.TRIANGLE)
    triangle.scale = wp.vec3(100.0, 320.0, 0.0)
    triangle.auxiliary = wp.vec3(0.0, 320.0, 0.0)

    cylinder = GenericShapeData()
    cylinder.shape_type = int(GeoType.CYLINDER)
    cylinder.scale = wp.vec3(0.02, 0.07, 0.02)
    cylinder.auxiliary = wp.vec3(0.0)

    data_provider = SupportMapDataProvider()
    collision, point_a, point_b, normal, penetration = wp.static(create_solve_mpr(support_map).core)(
        triangle,
        cylinder,
        wp.quat(-0.16498479, 0.36682746, -0.48718625, 0.77515626),
        wp.vec3(28.87991333, 260.53430176, 0.00036136055),
        0.0,
        data_provider,
    )

    collision_out[0] = int(collision)
    point_a_out[0] = point_a
    point_b_out[0] = point_b
    normal_out[0] = normal
    penetration_out[0] = penetration


collision = wp.zeros(1, dtype=int, device="cpu")
point_a = wp.zeros(1, dtype=wp.vec3, device="cpu")
point_b = wp.zeros(1, dtype=wp.vec3, device="cpu")
normal = wp.zeros(1, dtype=wp.vec3, device="cpu")
penetration = wp.zeros(1, dtype=float, device="cpu")

wp.launch(run_case, dim=1, outputs=[collision, point_a, point_b, normal, penetration], device="cpu")

orientation = np.array([-0.16498479, 0.36682746, -0.48718625, 0.77515626])
qx, qy, _qz, _qw = orientation
axis_z = 1.0 - 2.0 * (qx * qx + qy * qy)
expected = 0.07 * abs(axis_z) + 0.02 * np.sqrt(1.0 - axis_z * axis_z) - 0.00036136055

print("collision:", collision.numpy()[0])
print("point_a:", point_a.numpy()[0])
print("point_b:", point_b.numpy()[0])
print("normal:", normal.numpy()[0])
print("penetration:", penetration.numpy()[0], "m")
print("analytic cylinder-plane overlap:", expected, "m")
```

On `ee84505a` this prints approximately:

```text
collision: 1
point_a: [ 19.584011 320.00085    0.      ]
point_b: [ 19.58408  261.8253    -8.732943]
normal: [3.62e-11 9.889196e-01 1.484519e-01]
penetration: 58.826744 m
analytic cylinder-plane overlap: 0.061719150 m
```

### System Information

- Newton: `ee84505ae44f1d2c04a38ac513bf34ab93aa81aa`
- Warp: `1.16.0.dev20260716` (the version locked by that Newton commit)
- Python: `3.12.3`
- OS: Linux `6.12.57-061257-generic`, x86_64
- GPU: NVIDIA RTX PRO 6000 Blackwell Workstation Edition
- NVIDIA driver: `570.211.01`
- The minimal reproduction above runs on CPU; the GPU is not required.

LuyiLi:
### Problem

In some poses, MPR turns a local contact between a small cylinder and a large triangle into a false, tens-of-metres penetration with witness points far from the cylinder. In the deterministic reproduction, the correct penetration is `0.061719 m`, but Newton `ee84505a` returns `58.826744 m`; only the overlap boolean is correct.

The vulnerable case is near-coplanar: the convex center is close to the triangle plane, while its projection lies inside a large face but far from the triangle centroid. The existing initialization shifts that projection toward the centroid by `1%` of their separation. Here that adds `0.474106 m` tangentially against only `0.000361 m` normally, making the initial MPR direction nearly tangent to the face. An instrumented trace in #3773 shows the resulting pose-sensitive portal branch and nonlocal witness reconstruction.

### Changes

This PR preserves the shared-edge manifold intent of #2276 without letting the initialization scale with triangle size:

- For projections in the closed face region, construct the base Minkowski center offset directly as `-d * n` from signed plane distance `d` and face normal `n`. This avoids a false tangent residual from subtracting large nearby float32 positions.
- For `|d| > 0`, bound the centroid-directed tangent bias to `0.01 * min(|G-P|, |d|)`. Its tangent/normal ratio is therefore at most `0.01`, independent of triangle size.
- For exact `d=0`, use a triangle-specific face-normal fallback with the same centroid direction instead of the generic world-axis probe.
- Preserve the real tangent distance outside the face and the existing degenerate-triangle behavior.
- Add direct regressions for the reported pose, exact coplanarity, and shared-edge witnesses both above and on the plane.

Simply removing the centroid bias is not sufficient: it fixes the large-face case but makes adjacent triangles produce seam-biased witnesses. The bounded rule and triangle-specific fallback preserve those distinct witnesses, including at `d=0`.

Closes #3773. Related: #2276 and isaac-sim/IsaacLab#5071.

## Checklist

- [x] The documentation is up to date with these changes (no public API or workflow change)
- [x] `CHANGELOG.md` has been updated (user-facing collision fix)

## Test plan

Validated with Warp `1.16.0.dev20260716`, the version locked by base commit `ee84505a`, and Python `3.12.3`:

```bash
python -m unittest -v newton.tests.test_mpr
python -m unittest -v newton.tests.test_gjk
python -m unittest -v \
  newton.tests.test_narrow_phase.TestExtremeMeshTriangles.test_huge_triangle_center \
  newton.tests.test_narrow_phase.TestExtremeMeshTriangles.test_huge_triangle_near_edge \
  newton.tests.test_narrow_phase.TestExtremeMeshTriangles.test_huge_triangle_near_vertex \
  newton.tests.test_narrow_phase.TestExtremeMeshTriangles.test_shared_edge_flat
pre-commit run --files newton/_src/geometry/mpr.py newton/tests/test_mpr.py CHANGELOG.md
```

All listed tests and hooks pass. Additional deterministic validation covered:

- `1001` heights from `0` to `2 mm` in `2 μm` steps, with no invalid depth, normal, or witness result;
- triangle scales from `0.005x` to `10x`, with penetration within `2.9e-5` relative error and `normal.z >= 0.99999994`;
- the exact reported input: `58.826744 m` on `ee84505a`, `0.061719764 m` with this PR.

## Bug fix

**Steps to reproduce:**

1. Run the standalone CPU reproduction in #3773 on Newton `ee84505a`.
2. Observe `overlap=True`, penetration `58.826744 m`, a mostly tangential normal, and nonlocal witnesses.
3. Compare with the analytic cylinder-plane penetration `0.061719150 m`.
4. Run the same case with this PR and observe local witnesses, a face-normal contact, and penetration `0.061719764 m`.

**Minimal reproduction:**

The standalone script is in #3773. The executable regression in `newton/tests/test_mpr.py` calls `create_solve_mpr(support_map).core` directly, so broad phase, contact reduction, and dynamics solvers are not involved.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved collision detection for small convex shapes contacting large mesh triangles.
  * Improved contact positioning, normals, and penetration depth accuracy across differently sized and coplanar shapes.
  * Preserved consistent collision results near shared triangle edges, including fallback contact scenarios.
  * Improved reliability when handling degenerate triangle geometry and coincident shape centers.

* **Tests**
  * Added regression coverage for triangle-versus-convex collisions, including large triangles, coplanar cylinders, shared edges, and witness data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,6 +101,7 @@
 
 ### Fixed
 
+- Fix MPR returning scale-dependent, excessively deep contacts for small convex shapes on large mesh triangles while preserving triangle-specific shared-edge manifold witnesses.
 - Make deterministic collision pipelines cover hydroelastic contact generation and reduction, including unique reduced-contact sort keys and overflow-safe fixed-point pressure accumulation.
 - Convert `newton:mimicCoef0` from degrees to radians when the mimic follower joint is angular. Assets authored against the old behavior need the value rescaled to degrees.
 - Complete Kamino RCM traversal for large and disconnected systems and reuse the resulting permutation by default; set `reuse_permutation=False` to recompute it for changing matrix topology.
--- a/newton/_src/geometry/mpr.py
+++ b/newton/_src/geometry/mpr.py
@@ -230,49 +230,53 @@ def create_support_map_function(support_func: Any):
             center_b_local = 0.5 * (min_b + max_b)
 
         center_b_world = position_b + wp.quat_rotate(orientation_b, center_b_local)
+        center_b_to_a = center_a - center_b_world
 
         if geom_a.shape_type == int(GeoTypeEx.TRIANGLE) or geom_a.shape_type == int(GeoTypeEx.TRIANGLE_PRISM):
-            # Project shape B's center onto the triangle for a starting
-            # point near the contact region — this dramatically improves
-            # MPR convergence for large triangles.
-            #
-            # Nudge toward the centroid so the point moves into the face
-            # interior when possible.  This does NOT prevent an MPR
-            # degeneracy (MPR works fine from an edge point); it improves
-            # *manifold quality*.
-            # When shape B projects onto a shared mesh edge, both adjacent
-            # triangles get the same v0, producing MPR witness points biased
-            # toward the edge.  The manifold builder (multicontact.py) uses
-            # these witness points as its center for perturbed support
-            # mapping, so edge-biased centers cause overlapping contact
-            # polygons across the two triangles instead of distinct ones —
-            # resulting in asymmetric force distribution and spurious torque.
-            # Keep that triangle-specific direction, but limit the tangential
-            # displacement to 1% of the normal center-to-plane distance.  A
-            # face projection therefore tilts the initial ray by at most
-            # atan(0.01), about 0.57 degrees, regardless of triangle size.  A
-            # pure barycentric blend scales with the triangle and can make the
-            # initial ray almost tangential when a small partner lies on a
-            # large terrain face, causing MPR to converge to a non-minimum
-            # portal.
+            # Start near the local contact region.  A small centroid bias keeps
+            # adjacent triangles' manifold witnesses distinct at a shared
+            # edge, but bound it by the normal center-to-plane distance so a
+            # large face cannot turn the initial MPR ray nearly tangential.
             tri_a = wp.vec3(0.0, 0.0, 0.0)
             tri_b = geom_a.scale
             tri_c = geom_a.auxiliary
-            proj = closest_point_on_triangle(center_b_world, tri_a, tri_b, tri_c)
-            centroid = (tri_a + tri_b + tri_c) / 3.0
-            to_centroid = centroid - proj
-            distance_to_centroid = wp.length(to_centroid)
             face_normal = wp.cross(tri_b - tri_a, tri_c - tri_a)
-            face_normal_length = wp.length(face_normal)
-            if distance_to_centroid > 1.0e-12 and face_normal_length > 1.0e-12:
-                center_to_plane = wp.abs(wp.dot(center_b_world - proj, face_normal)) / face_normal_length
-                nudge_distance = 0.01 * wp.min(distance_to_centroid, center_to_plane)
-                center_a = proj + to_centroid * (nudge_distance / distance_to_centroid)
+            face_normal_length_sq = wp.length_sq(face_normal)
+            proj = closest_point_on_triangle(center_b_world, tri_a, tri_b, tri_c)
+
+            if face_normal_length_sq >= 1.0e-20:
+                face_normal_length = wp.sqrt(face_normal_length_sq)
+                face_normal_unit = face_normal / face_normal_length
+                signed_plane_distance = wp.dot(center_b_world - tri_a, face_normal_unit)
+                plane_proj = center_b_world - signed_plane_distance * face_normal_unit
+
+                # Reconstruct face-region offsets directly.  Subtracting two
+                # large, nearby positions can otherwise leave a spurious
+                # tangential component in the initial MPR ray.
+                inside_face = (
+                    wp.dot(wp.cross(tri_b - tri_a, plane_proj - tri_a), face_normal) >= 0.0
+                    and wp.dot(wp.cross(tri_c - tri_b, plane_proj - tri_b), face_normal) >= 0.0
+                    and wp.dot(wp.cross(tri_a - tri_c, plane_proj - tri_c), face_normal) >= 0.0
+                )
+                if inside_face:
+                    proj = plane_proj
+                    center_b_to_a = -signed_plane_distance * face_normal_unit
+                else:
+                    center_b_to_a = proj - center_b_world
+
+                centroid = (tri_a + tri_b + tri_c) / 3.0
+                to_centroid = centroid - proj
+                to_centroid -= wp.dot(to_centroid, face_normal_unit) * face_normal_unit
+                distance_to_centroid = wp.length(to_centroid)
+                if distance_to_centroid > 1.0e-12:
+                    nudge_distance = 0.01 * wp.min(distance_to_centroid, wp.abs(signed_plane_distance))
+                    center_b_to_a += to_centroid * (nudge_distance / distance_to_centroid)
+
             else:
-                center_a = proj
+                center_b_to_a = proj - center_b_world
 
         center.B = center_b_world
-        center.BtoA = center_a - center_b_world
+        center.BtoA = center_b_to_a
 
         return center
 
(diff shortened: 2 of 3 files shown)


nvtw (approved):
Thanks for this stability improvement. The change gooks good.

PR #3769, opened 2026-08-03
Add compliant ALM to SolverVBD rigid constraints

### LINKED ISSUE #3333: Unified Compliant ALM for VBD rigid bodies.
Replace the legacy AVBD hard/soft constraint handling in `SolverVBD` with a
unified compliant augmented-Lagrangian (ALM) formulation for both structural
joints and body-body contacts. Each constraint keeps its authored material
stiffness `k`, adds a separate numerical penalty `rho`  and enforces the constraint through a persistent
Lagrange multiplier with cone-projected friction — separating physical authoring from numerical enforcement. The legacy non-ALM path and the per-slot hard/soft switch are deprecated with migration guidance, following
the Newton deprecation cycle.

jumyungc:
## Description

Add an opt-in unified compliant ALM formulation for experimental `SolverVBD` rigid contacts, structural joints, drives, and limits.

  With `rigid_compliant_alm=True`, authored stiffness represents physical compliance while the solver selects numerical ALM conditioning internally. This provides
  one continuous formulation from soft constraints to the large-stiffness limit.

  This PR also:

  - preserves the legacy path with `rigid_compliant_alm=False`;
  - warns when the option is omitted before its default changes to `True`;
  - deprecates the per-slot joint hard/soft mode;
  - uses projected Coulomb friction with collision-pipeline-owned contact matching;
  - updates VBD examples, documentation, and focused tests for the new stiffness semantics.

## Checklist

## New feature / API change
  ```python
  solver = newton.solvers.SolverVBD(
      model,
      rigid_compliant_alm=True,
  )



## Summary by CodeRabbit

* **New Features**
  * Added unified compliant augmented-Lagrangian handling for rigid contacts and joints in VBD simulations.
  * Added finite material stiffness, automatic numerical conditioning, improved friction behavior, and retained multiplier support.
  * Added separate stiffness and control for joint drives and limits.

* **Bug Fixes**
  * Improved contact warm-starting, stabilization, reset behavior, and structural stiffness handling.

* **Documentation**
  * Documented compliant ALM configuration, tuning guidance, migration from legacy AVBD, and deprecated rigid-contact controls.
  * Updated examples to demonstrate the new solver configuration.


newton/_src/sim/collide.py
```diff
@@ -1251,6 +1251,7 @@ def contacts(self) -> Contacts:
         # Flag the buffer so solvers that only consume particle contacts can refuse it (see
         # Contacts._enable_rigid_soft_full_surface_contact); edge/face records appear only when this is set.
         contacts._enable_rigid_soft_full_surface_contact = self.enable_rigid_soft_full_surface_contact
+        contacts._rigid_contact_matching_mode = self.contact_matching
```
jumyungc: Note: This is a temporary workaround and will be reverted once we have a proper way to query the contact-matching mode. That API work is outside the scope of this PR.

newton/_src/sim/collide.py
```diff
@@ -1308,6 +1309,7 @@ def collide(
         # particle-only solvers (XPBD, semi-implicit, Style3D) would not raise and would silently
         # ignore them. Mirrors the assignment in CollisionPipeline.contacts().
         contacts._enable_rigid_soft_full_surface_contact = self.enable_rigid_soft_full_surface_contact
+        contacts._rigid_contact_matching_mode = self.contact_matching
```
jumyungc: Note: This is a temporary workaround and will be reverted once we have a proper way to query the contact-matching mode. That API work is outside the scope of this PR.

newton/_src/sim/contacts.py
```diff
@@ -291,6 +291,8 @@ def __init__(
             # Contact matching index — filled by the collision pipeline when
             # contact_matching is enabled.
             self.contact_matching = contact_matching
+            # Internal collision-pipeline mode consumed by contact-history solvers.
+            self._rigid_contact_matching_mode = "disabled"
```
jumyungc: Note: This is a temporary workaround and will be reverted once we have a proper way to query the contact-matching mode. That API work is outside the scope of this PR.

newton/examples/basic/example_basic_conveyor_forces.py
```diff
@@ -790,13 +791,16 @@ def __init__(self, viewer, args=None):
                     builder.shape_material_mu[shape] = 0.0
                     builder.shape_material_mu_torsional[shape] = 0.0
                     builder.shape_material_mu_rolling[shape] = 0.0
+        elif self.solver_type == "vbd":
+            for shape in range(len(builder.shape_material_ke)):
+                builder.shape_material_ke[shape] = VBD_RIGID_CONTACT_KE
+                builder.shape_material_kd[shape] = VBD_RIGID_CONTACT_KD
```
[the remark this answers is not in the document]
jumyungc: Belt-box native friction is already zero.

eric-heiden:
Nice work moving the rigid constraint path onto a unified compliant ALM formulation, with an explicit `rigid_compliant_alm` opt-in and a scoped migration warning so existing callers get a clear path forward.

Two smaller points below, neither a blocker: the constructor kwargs that only exist to configure the legacy path carry no deprecation marker even though that path is now deprecated, and the new validation in `notify_model_changed` adds a blocking device to host copy to a hook that used to be free.

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -859,8 +992,18 @@ def _init_rigid_system(
...
+        refresh_structural_k = (
+            bool(flags & ModelFlags.JOINT_PROPERTIES) and self._integrates_rigid_bodies and self.model.joint_count > 0
+        )
         if flags & (ModelFlags.BODY_PROPERTIES | ModelFlags.BODY_INERTIAL_PROPERTIES):
             self._refresh_kinematic_state()
+        if self.rigid_compliant_alm:
+            if flags & ModelFlags.SHAPE_PROPERTIES:
+                self._validate_compliant_contact_materials()
+            if flags & ModelFlags.JOINT_DOF_PROPERTIES:
+                self._validate_compliant_joint_dof_materials()
```
eric-heiden: 🔵 This re-runs `_validate_compliant_contact_materials()` on every `notify_model_changed(SHAPE_PROPERTIES)` and `_validate_compliant_joint_dof_materials()` on every `notify_model_changed(JOINT_DOF_PROPERTIES)` while compliant ALM is on. Both validators read model arrays through `_to_numpy`, which is `arr.to("cpu")` followed by `.numpy()`, so each one is a blocking device to host copy.

That turns a cheap live-update hook into a pipeline stall. Randomizing shape materials per episode and then calling `solver.notify_model_changed(ModelFlags.SHAPE_PROPERTIES)` is the documented workflow (see `newton/examples/selection/example_selection_materials.py:233`), and `SolverCoupled.notify_model_changed` forwards the flags to every entry solver, so the cost is paid per solver on every reset. `SHAPE_PROPERTIES` pulls back three `shape_count`-sized arrays and `JOINT_DOF_PROPERTIES` pulls four `joint_dof_count`-sized ones. On `main` neither flag caused any host round trip here.

The construction-time validation earns its cost because it runs once. Could we keep it construction-only? If live validation is genuinely wanted, either gate it behind an opt-in debug flag, or do the finite and non-negative check with a Warp reduction kernel that writes a single flag read at an existing sync point, so no synchronous readback lands on the hot update path.

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -247,28 +290,29 @@ def __init__(
...
         rigid_body_particle_contact_buffer_size: int = 256,  # Per-body soft-contact list capacity (particle + edge/face)
         # Rigid body - joints
-        rigid_joint_linear_ke: float = 1.0e5,  # Penalty stiffness ceiling for structural linear joint constraints
-        rigid_joint_angular_ke: float = 1.0e5,  # Penalty stiffness ceiling for structural angular joint constraints
-        rigid_joint_linear_k_start: float = 1.0e2,  # Linear penalty seed (used when linear beta > 0)
-        rigid_joint_angular_k_start: float = 1.0e1,  # Angular penalty seed (used when angular beta > 0)
+        rigid_joint_linear_ke: float = 1.0e5,  # Structural linear joint stiffness
+        rigid_joint_angular_ke: float = 1.0e5,  # Structural angular joint stiffness
+        rigid_joint_linear_k_start: float = 1.0e2,  # Legacy AVBD linear joint penalty ramp seed
+        rigid_joint_angular_k_start: float = 1.0e1,  # Legacy AVBD angular joint penalty ramp seed
```
eric-heiden, line 315: 🔵 `changelog/3333.deprecated.md` deprecates the legacy non-ALM rigid path and states it will be removed. The kwargs declared here that exist only to configure that path go with it: `rigid_avbd_beta`, `rigid_avbd_linear_beta`, `rigid_avbd_angular_beta`, `rigid_contact_hard`, `rigid_contact_k_start`, `rigid_joint_linear_k_start` and `rigid_joint_angular_k_start`. None of them carries a `.. deprecated::` directive or emits a warning.

The docstring entries do describe the legacy semantics well, but prose like "Legacy AVBD penalty ramp rate per iteration" renders in Sphinx exactly like a fully supported parameter. Someone tuning `rigid_avbd_beta` today gets no signal that it is on its way out. Removing a public symbol needs a deprecation to land in an earlier release, so this PR is the natural place to start that clock.

The pattern is already in this file a few lines below, on the `rigid_contact_stick_*` entries:

```rst
            rigid_contact_stick_motion_eps: Deprecated and ignored. ...

                .. deprecated:: 1.5
```

Adding `.. deprecated:: 1.6` to those seven docstring entries would cover it.

One related point on the same surface: the `rigid_avbd_` prefix now spans two lifecycles. `rigid_avbd_gamma`, `rigid_avbd_joint_alpha` and `rigid_avbd_contact_alpha` stay live under compliant ALM, while `rigid_avbd_*beta` is legacy only, and the names give no hint which is which. A short sentence on the surviving entries noting that they outlive the legacy path would stop the shared prefix from implying a shared fate.

newton/_src/solvers/vbd/solver_vbd.py
jumyungc: Updated.

newton/_src/solvers/vbd/solver_vbd.py
jumyungc: Although body-particle contacts still consume it because they are not yet handled by compliant ALM, I think deprecating it as part of this migration is the cleanest approach. Ramping is already discouraged, with an effective beta=0 as the default behavior, and retaining the API solely for body-particle contacts is not ideal. The remaining points have been addressed.

jumyungc:
> Nice work moving the rigid constraint path onto a unified compliant ALM formulation, with an explicit `rigid_compliant_alm` opt-in and a scoped migration warning so existing callers get a clear path forward.
> 
> Two smaller points below, neither a blocker: the constructor kwargs that only exist to configure the legacy path carry no deprecation marker even though that path is now deprecated, and the new validation in `notify_model_changed` adds a blocking device to host copy to a hook that used to be free.

Thank you for the feedback - both points are now addressed.

eric-heiden (approved):
Approving. Both points I raised are addressed at `a78350f`.

`notify_model_changed` no longer runs the compliant material validators, so the blocking device to host copies are gone and the hook stays cheap for the documented per-episode material randomization workflow.

All seven legacy kwargs now carry `.. deprecated:: 1.6` directives with concrete migration guidance, so Sphinx renders them as on their way out instead of as fully supported tuning knobs. I agree with deprecating the ramping controls now even though body-particle contacts still consume them. Keeping the effective beta at `0` as the default behavior is the right story to document.

CI is green across the four platform unittest jobs, GPU tests, and GPU benchmarks, and the compliant ALM path does not add measurable per-step cost.

Nothing blocking remains. Two optional follow ups, fine to take here or separately:

- The class summary line and the "Rigid body simulation (joints, contacts) using the AVBD algorithm" bullet still present AVBD as the rigid algorithm, while the text immediately below correctly describes compliant ALM as the recommended path. Worth a short wording pass so the summary matches the body.
- `example_cable_bend_stiffness` and `example_cable_y_junction` each run about 2.2x slower from an iteration and substep increase that their own tests pass without. Worth confirming those retunes are intentional rather than leftover tuning.

PR #3776, opened 2026-08-03
Improve cube stacking stability

nvtw:
## Description

Improve cube stacking stability

## Bug Fixes

* Improved box collision stability under tiny solver rotations and near-aligned configurations.
* Preserved complete face-contact manifolds with four unique contacts and consistent normals.
* Ensured consistent collision results across supported broad-phase detection modes.
* Improved handling of nearly tied box support points.
* Improved stability of aligned box stacks during simulation.
* Added validation for collision normals, penetration depth, and contact separation.

## Documentation

* Added an unreleased changelog entry documenting the collision manifold fix.


newton/_src/geometry/mpr.py
```diff
@@ -301,9 +303,27 @@ def create_solve_mpr(support_func: Any, _support_funcs: Any = None):
...
+        _support_map_b, _minkowski_support, geometric_center = _support_funcs
     else:
-        _support_map_b, minkowski_support, geometric_center = create_support_map_function(support_func)
+        _support_map_b, _minkowski_support, geometric_center = create_support_map_function(support_func)
+
+    @wp.func
+    def centered_box_support(geom: Any, direction: wp.vec3, data_provider: Any) -> wp.vec3:
+        result = support_func(geom, direction, data_provider)
+        if geom.shape_type == GeoType.BOX:
+            # A nearly tied box face has infinitely many valid support points. Its center
```
mzamoramora-nvidia, line 314: Non-blocking / test coverage: The face has multiple equally valid support points only when this direction component is exactly zero. Here we also center components with a small nonzero contribution, so the returned point is no longer the exact support point. That may be the right robustness trade-off, but could we add one or two fixtures just below and above the tie threshold? Checking the contact/witness normal there would make the approximation's safe range explicit.

newton/tests/test_collision_pipeline.py
```diff
@@ -1553,6 +1553,105 @@ def test_box_box_quaternion_perturbation(test, device, broad_phase: str):
     )
 
 
+def test_box_box_solver_drift_manifold(test, device, broad_phase: str):
```
mzamoramora-nvidia, line 1626: Non-blocking / follow-up coverage: This is a useful narrow-phase snapshot. Since the reported symptom is an XPBD stack collapsing, could we also add a small deterministic stack test that steps through the failure window? It would protect the user-visible stability outcome if future solver changes alter how these otherwise-correct manifolds are consumed.

newton/tests/test_collision_pipeline.py
```diff
@@ -1553,6 +1553,105 @@ def test_box_box_quaternion_perturbation(test, device, broad_phase: str):
...
+        model = builder.finalize(device=device)
+        pipeline = newton.CollisionPipeline(model, broad_phase=broad_phase)
+        contacts = pipeline.contacts()
+        pipeline.collide(model.state(), contacts)
+
+        count = int(contacts.rigid_contact_count.numpy()[0])
+        points = contacts.rigid_contact_point0.numpy()[:count]
+        normals = contacts.rigid_contact_normal.numpy()[:count]
+
+        test.assertEqual(count, 4, f"Expected 4 face-corner contacts, got {count}")
```
mzamoramora-nvidia: Non-blocking / assertion strength: Could this assert the full contact contract as well? In particular, validate the expected four tangent-plane corner positions (with tolerance), `point1`/separation, and a normalized near-vertical normal. Four distinct rounded `point0` projections and positive `normal.y` are a good smoke test, but a malformed manifold could still satisfy them. The same helper/assertions could cover the unequal-box fixture below.

mzamoramora-nvidia (dismissed):
Looks good to me. The focused regressions cover the reported equal- and unequal-box manifold failures across all three supported broad phases, and the MPR-local handling keeps the robustness policy out of the shared support map.

I left a few non-blocking follow-ups on boundary and end-to-end coverage, but they should not hold up this fix. Approving.

mzamoramora-nvidia (approved):

PR #3779, opened 2026-08-03
Add culling of speculative contacts in Kamino

chschuma-disney:
## Description

Speculative contacts (contacts with positive distance) can currently lead to restitution issues in Kamino. As a temporary workaround while investigating possible strategies to allow speculative contacts, these contacts are now culled during contact conversion in Kamino.

## Checklist

## Test plan

```
python -m unittest newton/_src/solvers/kamino/tests/test_geometry_contacts.py
```

## Summary by CodeRabbit

* **New Features**
  * Added default culling for speculative contacts during Newton-to-Kamino conversion.
  * Contacts separated by a positive margin-adjusted distance are excluded by default, reducing unnecessary contact processing.
  * Genuine contacts continue to be converted correctly across margin and gap configurations.

* **Bug Fixes**
  * Cleared force values for contacts removed during conversion, preventing stale force data from being reported.

* **Tests**
  * Added coverage for speculative contact culling, genuine contacts, and force clearing.

### DIFF
--- a/newton/_src/solvers/kamino/_src/geometry/contacts.py
+++ b/newton/_src/solvers/kamino/_src/geometry/contacts.py
@@ -112,6 +112,12 @@ threshold.
 Set to `1e-5`.
 """
 
+DEFAULT_CULL_SPECULATIVE_CONTACTS: bool = True
+"""
+The global default for culling speculative contacts (with positive
+margin-shifted distance) during Newton->Kamino conversion.
+"""
+
 
 ###
 # Types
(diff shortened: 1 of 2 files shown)


rubengrandia (changes requested):
A local review agent flagged that we also need to write contact forces for culled contacts in `_convert_existing_contacts_kamino_to_newton`. If the contacts are culled, no one is updating the contact force results anymore, and we might end up with stale data for contacts that were in contact in a previous step and culled in the current step.

chschuma-disney:
> A local review agent flagged that we also need to write contact forces for culled contacts in `_convert_existing_contacts_kamino_to_newton`. If the contacts are culled, no one is updating the contact force results anymore, and we might end up with stale data for contacts that were in contact in a previous step and culled in the current step.

Great point! I pushed a fix and regression test, I hope that should resolve it.

rubengrandia (approved):
LGTM 🚀

PR #3780, opened 2026-08-03
Stabilize Warp 1.16 dependency

jcarius-nv:
## Description

Stabilize Newton on the released Warp 1.16.0 package while preserving the established split between `main` and release branches.

- Replace the Warp development requirement and ASV pin with stable Warp 1.16.0.
- Keep the NVIDIA `warp-lang` package-index override on `main` and resolve the stable package through that source, leaving `main` ready for the next dev-nightly update.
- Keep ASV's NVIDIA extra-index option on `main` so its wheel installation remains ready for the next Warp dev-nightly requirement.
- Document that pre-branch stabilization retains the NVIDIA source on `main`, while release-candidate preparation removes it from `release-X.Y` so release wheels install purely from public PyPI.
- Leave previously released Newton branches unchanged.
- Add the Warp 1.16 requirement to the unreleased changelog.

This follows the same flow used for prior Warp and Newton releases: stabilize the dependency on `main`, then remove the NVIDIA source later on the new release branch.

## Checklist

## Test plan

```bash
uv lock --check
uv run --frozen python -c "import warp as wp; print(wp.__version__); assert wp.__version__ == '1.16.0'"
uv run --extra docs --extra sim sphinx-build -j auto -W -b html docs /tmp/newton-warp-1.16-docs
```

The full test run completed 5,195 tests successfully with 162 skips. After correcting the package-source configuration, the lock check, frozen Warp version check, pre-commit suite, and warning-as-error Sphinx build all passed again.

## Summary by CodeRabbit

* **Chores**
  * Updated the project to use the stable Warp 1.16.0 release.
  * Updated dependency requirements to support Warp 1.16.0 and newer.
* **Documentation**
  * Added release checklist guidance for stabilizing dependency and benchmark references before branch creation.
  * Updated the unreleased changelog with the new minimum Warp version.


asv.conf.json
```diff
@@ -15,8 +15,8 @@
   "default_benchmark_timeout": 600,
   "build_command": ["python -m build --wheel -o {build_cache_dir} {build_dir}"],
   "install_command": [
-    "in-dir={env_dir} python -m pip install {wheel_file}[dev] --extra-index-url=https://pypi.nvidia.com/",
-    "in-dir={env_dir} python -m pip install warp-lang==1.16.0.dev20260716 mujoco==3.11.0 mujoco-warp==3.11.0 --extra-index-url=https://pypi.nvidia.com/",
```
eric-heiden, line 19: ⚪ Dropping `--extra-index-url=https://pypi.nvidia.com/` from both `pip install` lines is correct for the current pins. `warp-lang==1.16.0`, `mujoco==3.11.0`, and `mujoco-warp==3.11.0` all resolve from public PyPI, and `uv.lock` shows `pypi.nvidia.com` was the source for `warp-lang` only, so nothing in `{wheel_file}[dev]` needs that index anymore.

The residual risk is in the process rather than in this change. Warp dev builds (`1.17.0.devYYYYMMDD`) are published only to `pypi.nvidia.com`, so the next time someone re-pins `asv.conf.json` to a nightly, which is exactly the flow the new item in `docs/guide/release.rst` anticipates, the benchmark environment build fails with `No matching distribution found for warp-lang==1.17.0.devXXXX`. That new checklist item only mentions the `pyproject.toml`/`uv` index override, so the pip-level flag removed here has no home in the docs.

Can we leave `--extra-index-url=https://pypi.nvidia.com/` in place here? It is a no-op for stable releases, and `aws_gpu_benchmarks.yml` runs with `egress-policy: audit`, so it costs nothing to keep.

jcarius-nv: Agreed. We originally removed these flags by following Alain’s Warp 1.14 stabilization precedent (#3046), where ASV switched from an RC on the NVIDIA index to the stable package from public PyPI while `main` retained the `tool.uv.sources` entry. As you point out, the current ASV setup added the flags specifically so the wheel install survives a nightly Warp floor (#3525), and retaining them avoids a second, easy-to-miss edit on the next nightly bump. Restored on both install commands in 62e62ce4 and documented retaining them on `main`; the release branch will still remove them.

eric-heiden (approved):
Thanks!

asv.conf.json
```diff
@@ -16,7 +16,7 @@
   "build_command": ["python -m build --wheel -o {build_cache_dir} {build_dir}"],
   "install_command": [
     "in-dir={env_dir} python -m pip install {wheel_file}[dev] --extra-index-url=https://pypi.nvidia.com/",
-    "in-dir={env_dir} python -m pip install warp-lang==1.16.0.dev20260716 mujoco==3.11.0 mujoco-warp==3.11.0 --extra-index-url=https://pypi.nvidia.com/",
+    "in-dir={env_dir} python -m pip install warp-lang==1.16.0 mujoco==3.11.0 mujoco-warp==3.11.0 --extra-index-url=https://pypi.nvidia.com/",
```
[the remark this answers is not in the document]
jcarius-nv: intentionally left, see Eric's comment

PR #3781, opened 2026-08-03
Stabilize friction stopping-distance test

### LINKED ISSUE #3777: [BUG] test_friction_stopping_distance_mujoco_warp_cuda_0 is flaky again
### Bug Description

`test_friction_stopping_distance_mujoco_warp_cuda_0` is failing intermittently in GPU CI again after the previous flakiness report, #3266, was closed by #3396.

A recent failure on an unrelated change occurred in [PR #3671's GPU test job](https://github.com/newton-physics/newton/actions/runs/30504704054/job/91261894835?pr=3671):

```text
AssertionError: 1 stopping-distance failure(s):
  (mu=0.70): |v_final|=0.0533 m/s >= 0.05 after measurement window
```

The current symptom differs from #3266: the earlier failure exceeded the 1% stopping-distance tolerance, while this failure exceeds the final-velocity threshold. The test should pass consistently when the code under test does not affect rigid friction behavior.

### Reproduction Script

```bash
for i in {1..20}; do
    -k test_friction_stopping_distance_mujoco_warp_cuda_0 \
    --no-cache-clear || break
done
```

### System Information

Observed on the AWS GPU GitHub Actions runner using `cuda:0`. See the linked CI job for the complete environment and logs.

vreutskyy:
## Description

Closes #3777.

Replace the stopping-distance test's single-frame 3-D velocity assertion with
average planar drift over a 0.25-second trailing rest window. The analytical
stopping-distance measurement remains at its original simulation time.

The old assertion could fail on normal-direction contact jitter even though the
test's horizontal stopping-distance oracle passed. A failing rest-window check
now reports the start and end XYZ velocities to distinguish residual sliding
from contact jitter.

## Checklist

- [x] The documentation is up to date with these changes (test-only change)
- [x] `CHANGELOG.md` has been updated (not applicable; no user-facing change)

## Test plan

```text
uv run --extra dev -m unittest newton.tests.test_rigid_friction_ramp.TestRigidFrictionRamp.test_friction_stopping_distance_xpbd_cpu -v
uv run --extra dev -m unittest newton.tests.test_rigid_friction_ramp.TestRigidFrictionRamp.test_friction_stopping_distance_mujoco_cpu_cpu -v
uv run --extra dev -m unittest newton.tests.test_rigid_friction_ramp.TestRigidFrictionRamp.test_friction_stopping_distance_vbd_cpu -v
uvx pre-commit run -a
```

All listed checks pass. The exact MuJoCo-Warp CUDA case was also attempted
locally, but kernel compilation did not complete within the bounded timeout
while the GPU was shared with other workloads. GPU CI provides the authoritative
Blackwell result.

## Bug fix

**Steps to reproduce:**

1. Run the MuJoCo-Warp CUDA stopping-distance test repeatedly.
2. Observe an intermittent failure of the final full 3-D velocity assertion,
   such as `0.0533 m/s >= 0.05 m/s`, while the stopping-distance check passes.

**Minimal reproduction:**

```bash
for i in {1..20}; do
    -k test_friction_stopping_distance_mujoco_warp_cuda_0 \
    --no-cache-clear || break
done
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved friction stopping validation by measuring residual motion over a short rest period instead of relying on a single final velocity reading.
  * Added checks for invalid numeric results during both motion and rest measurements.
  * Improved reporting of residual drift and endpoint velocities.
  * Updated stopping criteria to use solver-specific rest-speed thresholds, improving consistency across supported configurations.

### DIFF
--- a/newton/tests/test_rigid_friction_ramp.py
+++ b/newton/tests/test_rigid_friction_ramp.py
@@ -316,8 +316,8 @@ def test_friction_stopping_distance(test, device, solver_fn, rel_tol, rest_speed
 
     state_0, state_1 = simulate(solver, model, state_0, state_1, control, collision_pipeline, contacts, num_frames)
 
-    final_q = state_0.body_q.numpy()
-    final_qd = state_0.body_qd.numpy()
+    final_q = state_0.body_q.numpy().copy()
+    final_qd = state_0.body_qd.numpy().copy()
 
     # A trailing displacement is insensitive to one-frame normal contact jitter
     # while still detecting residual sliding after the stopping-distance check.
--- a/newton/tests/test_rigid_friction_ramp.py
+++ b/newton/tests/test_rigid_friction_ramp.py
@@ -99,7 +99,8 @@ STOPPING_PATCH_HY = 0.6
 STOPPING_PATCH_HZ = 0.05
 STOPPING_BOX_PITCH_Y = 5.0
 STOPPING_SETTLE_FRAMES = 30
-STOPPING_V_FINAL_MAX = 0.05  # m/s - sanity bound: box must have come to rest
+STOPPING_REST_FRAMES = 15  # 0.25 s window for averaging residual planar motion
+STOPPING_REST_SPEED_MAX = 0.05  # m/s - sanity bound on average planar motion after stopping
 
 _ROW_COLORS = (
     (0.90, 0.30, 0.30),
@@ -269,13 +270,14 @@ def build_stopping_distance_scene(device):
     return builder.finalize(device=device), box_ids
 
 
-def test_friction_stopping_distance(test, device, solver_fn, rel_tol, v_final_max, native_contacts=False):
+def test_friction_stopping_distance(test, device, solver_fn, rel_tol, rest_speed_max, native_contacts=False):
     """Verify a sliding box stops at d = v0^2 / (2 mu g).
 
     Three boxes at mu in STOPPING_MUS settle on matching ground patches, then
     start with v0 along world-X. Run for 1.5 * t_stop(mu_min) so every box has
     come to rest for Coulomb-cone solvers, then compare measured stopping
-    distance against the analytical value with per-solver bounds.
+    distance against the analytical value and check average planar motion over
+    a short trailing rest window.
     """
     model, box_ids = build_stopping_distance_scene(device)
     solver = solver_fn(model)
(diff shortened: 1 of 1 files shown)


eric-heiden:
Thanks for tightening this test, the trailing rest window is a nice way to catch residual sliding without tripping over a single frame of normal contact jitter.

One thing to sort out before this lands: the new drift check does not actually exercise anything on CPU devices, so those variants would pass for the wrong reason. Details inline.

newton/tests/test_rigid_friction_ramp.py
```diff
@@ -317,25 +319,42 @@ def test_friction_stopping_distance(test, device, solver_fn, rel_tol, v_final_ma
...
+    rest_q = state_0.body_q.numpy()
+    rest_qd = state_0.body_qd.numpy()
+
+    if any(np.any(np.isnan(values)) for values in (final_q, final_qd, rest_q, rest_qd)):
         test.fail("Simulation produced NaN values (numerical instability)")
 
+    rest_duration = STOPPING_REST_FRAMES * SIM_DT
     failures = []
     for bid, mu in zip(box_ids, STOPPING_MUS, strict=True):
         d_expected = STOPPING_V0 * STOPPING_V0 / (2.0 * mu * g)
```
eric-heiden: 🟠 `final_q` and `final_qd` are captured without `.copy()` on lines 319-320, so on a CPU device they are views into the live `state_0` buffers instead of snapshots. That was harmless before, because nothing stepped the sim after the capture. With the new `simulate(..., STOPPING_REST_FRAMES)` call inserted here, the solver keeps writing through those same buffers.

`STOPPING_REST_FRAMES` runs an even number of `state_0`/`state_1` swaps, so the returned `state_0` is the same object `final_q`/`final_qd` point into. The effect on CPU:

| Expression | Intended | Actual on CPU |
| --- | --- | --- |
| `rest_q[bid, :2] - final_q[bid, :2]` | drift during the rest window | always `0.0` |
| `rest_speed` | residual sliding speed | always `0.0`, never reaches `rest_speed_max` |

So the residual-drift assertion this PR adds is dead on `test_friction_stopping_distance_xpbd`, `_mujoco_cpu`, and `_vbd`. On CUDA `.numpy()` copies, so the check behaves as intended there, which makes this easy to miss locally.

There is a second effect worth noting: `d_measured` on line 339 is now computed from body positions sampled after the extra rest window rather than at the end of the measurement window the docstring describes, which quietly shifts the primary stopping-distance oracle on CPU.

The fix is the same guard the file already uses for `settle_q` (line 211) and `initial_q` (line 296):

```python
final_q = state_0.body_q.numpy().copy()
final_qd = state_0.body_qd.numpy().copy()
```

vreutskyy: Fixed in b3123532 by copying both measurement arrays before advancing the rest window, preserving the intended CPU snapshots for the stopping-distance and drift checks. Verified the XPBD CPU, MuJoCo CPU, and VBD CPU stopping-distance tests, and `uvx pre-commit run -a` passes.

eric-heiden (approved):

PR #3122, opened 2026-06-10
Split VBD cable stretch/shear and bend/twist constraints

jumyungc:
## Description

- Splits the VBD cable response into independent stretch, shear, bend, and twist modes.
- Adds geometric rest bend/twist tracking for cable joints, plus cable stiffness utilities that map physical EA, EI, and GJ values into solver parameters.

- Full validation report: https://jumyungc.github.io/newton/reports/cable_bend_twist_verification/
- Strain-measure comparison, Körner et al. vs. DER: https://jumyungc.github.io/newton/reports/cable_twist_wiggle_buckling/, especially case 3.

## Checklist

- [x] New or existing tests cover these changes
- [x] The documentation is up to date with these changes
- [x] `CHANGELOG.md` has been updated (if user-facing change)

## Test plan

```python
import newton

# Code that demonstrates the new capability
```

## Summary by CodeRabbit

- **New Features**
  - Added optional shear and twist stiffness/damping controls to cable/rod builder APIs and documentation (with sensible defaults).
  - Introduced `CableStiffness` and enhanced material-based stiffness utilities to support torsional stiffness.
  - Updated rigid VBD cable behavior to a split `[stretch, shear, bend, twist]` layout for improved bend/twist fidelity.
- **Examples**
  - Added multiple VBD cable validation examples (analytic, stiffness, twist transfer, buckling, torsion mapping, Michell threshold, Dahl hysteresis) and a new cable plectoneme example.
- **Documentation**
  - Updated README and API docs, plus clarified example run instructions; added a viewer camera helper for VBD examples.
- **Tests**
  - Expanded split-cable solver and example regression coverage, including additional finite-difference and Dahl active-subspace checks.

PR #3783, opened 2026-08-04
Refine the VBD cable stretch/shear solve Hessian

jumyungc:
## Description

  Follow-up to #3122, which split VBD cable constraints into independent stretch/shear and bend/twist rows.

  When stretch and shear elastic stiffness are equal, the material-frame and original world-space formulations define the same elastic energy and forces. Their Gauss–Newton solve Hessians need not be identical, however, because they omit different second-order terms.

  This PR decomposes the stretch/shear solve metric into:

  - a common isotropic elastic positive-semidefinite block, evaluated with the world-form parent-anchor Jacobian; and
  - a remaining material-frame positive-semidefinite block containing anisotropic elasticity and damping.

  The common elastic block uses the minimum directional stiffness, which keeps the remainder positive semidefinite. When stretch and shear stiffness are equal, the anisotropic elastic remainder vanishes. Damping remains in the material frame because its previous-step material residual is frozen.

  This changes only the local VBD solve Hessian. It does not change the cable material energy, forces, damping law, authored parameters, or converged solution.
  Since VBD uses a finite iteration budget, simulated trajectories may still change; this PR does not claim a universal convergence improvement.

  There are no API changes.

  No new changelog entry is needed because this refines the unreleased changes from #3122 rather than changing behavior from a previously released version.

## Checklist

## Test plan

```python
import newton

```

## Bug Fixes
* Improved cable stretch and shear force calculations for translated and rotated parent bodies.
* Improved stiffness handling during parent-body response assembly, including distinct stretch and shear behavior.
* Preserved accurate parent-body responses for isotropic cable stiffness.

## Tests
* Added coverage for cable force and Hessian calculations with separate stretch and shear stiffness values.
* Expanded device-level verification across translated and rotated reference frames.


jumyungc:
This is a small follow-up PR that could clean up the previous stretch/shear split. I set the milestone to 1.5 in case they can ship together, but if more review time is needed or the change is not appropriate, I’ll remove the milestone. Thank you! 
@gdaviet @eric-heiden @jcarius-nv  for viz.

gdaviet (commented):
The derivation looks good to me. Would be good to add a CHANGELOG entry since this will affect solver convergence behavior

jcarius-nv (commented):
This looks good to me. The implementation matches the intended Hessian decomposition, all checks pass, and the existing review thread has been addressed.

Optional: consider adding direct coverage for the equal-stiffness, zero-damping limit by comparing `H_al` and `H_aa` against the world-form parent-anchor Hessian. The current unequal-stiffness test covers the general decomposition, so I don’t consider this blocking.

jumyungc:
> The derivation looks good to me. Would be good to add a CHANGELOG entry since this will affect solver convergence behavior

Thanks! Assuming this ships with the stretch/shear and bend/twist split PR, I amended the existing changelog entry to clarify the final behavior.

jumyungc:
> This looks good to me. The implementation matches the intended Hessian decomposition, all checks pass, and the existing review thread has been addressed.
> 
> Optional: consider adding direct coverage for the equal-stiffness, zero-damping limit by comparing `H_al` and `H_aa` against the world-form parent-anchor Hessian. The current unequal-stiffness test covers the general decomposition, so I don’t consider this blocking.

Thanks! Since you marked it as optional, I left out the additional test to keep this follow-up’s coverage minimal. It ships with the earlier stretch/shear split, which already added broader cable tests.

jcarius-nv (approved):

PR #3613, opened 2026-07-23
DVI solver - contact improvements

nvtw:
## Description

DVI solver - contact improvements

Requires MR https://github.com/newton-physics/newton/pull/3610 to be merged first.

Unify dense and sparse DVI inequality handling around one conflict-free, graph-colored projected Gauss-Seidel schedule for joint limits and contacts. Solve bilateral equalities with batched Cholesky between inequality blocks, use automatic normal effective-mass scaling to improve stacked-contact and high-mass-ratio convergence, and apply the same contact formulation on CPU and CUDA.

Improve packed multi-world contact mapping, warm-start integration, friction behavior, and DR Legs contact stability. Deprecate `DVISolverConfig.contact_jacobi_omega`, `contact_jacobi_relaxation`, and `contact_block_preconditioner` without replacement because DVI now selects its contact updates automatically.

## Checklist

- [x] New or existing tests cover these changes
- [ ] The documentation is up to date with these changes
- [ ] `CHANGELOG.md` has been updated (if user-facing change)

## Test plan

```bash
uv run --extra dev -m unittest newton._src.solvers.kamino.tests.test_solvers_dvi
uv run --extra dev -m unittest newton.tests.test_solver_kamino_dvi
uv run --extra dev -m unittest newton.tests.test_solver_kamino_dvi_cuda
uvx pre-commit run -a
```

Coverage includes dense and sparse solves, joint-limit/contact coloring, packed multi-world contacts, stacked and high-mass-ratio systems, zero and kinetic friction, warm-starting, contact release, and DR Legs stability.

## New feature / API change

Remove manual DVI contact tuning from application configuration; contact scaling and preconditioning are now automatic:

```python
import newton

config = newton.solvers.SolverKamino.Config(dynamics_solver="dvi")
solver = newton.solvers.SolverKamino(model, config=config)
```

## Summary by CodeRabbit

* **New Features**
  * Added unified, graph-colored inequality-only DVI solving for dense and sparse workflows, including improved bilateral direct-block iteration.
  * Updated DVI contact preconditioning to a diagonal-based approach and enhanced warm-start reuse of cached topology.
  * Improved collision detection contact allocation via a per-world heuristic (auto-sized when unset).
* **Bug Fixes**
  * Improved constraint stability, including regression coverage for zero-friction tangent momentum preservation.
* **Breaking Changes**
  * Switched from contact-coloring to inequality-coloring.
[...]

PR #3791, opened 2026-08-04
Report Kamino contact buffer overflow

rubengrandia:
## Description

Add graph-capture-friendly, one-time `wp.printf` warnings when Kamino truncates contacts because either per-world or model contact capacity is exhausted. Apply the diagnostic to primitive narrow-phase contact generation and Newton-to-Kamino contact conversion, with regression coverage for both paths.

Implements [disneyresearch/newton#115](https://github.com/disneyresearch/newton/issues/115).

A user-facing overflow API is out of scope: Newton does not yet have an established API for this diagnostic. This PR therefore limits behavior to a single warning, following the precedent introduced for DVI in [#3613](https://github.com/newton-physics/newton/pull/3613).

## Checklist

## Test plan

```bash
uv run --extra dev -m unittest newton._src.solvers.kamino.tests.test_geometry_contacts newton._src.solvers.kamino.tests.test_geometry_primitive
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved contact-capacity management during collision detection, including safer handling when storage limits are reached.
  * Overflow warnings are now emitted once per collision pass and reset appropriately for subsequent passes.
  * Standardized capacity handling across supported collision types, with consistent contact truncation behavior.

* **Tests**
  * Added coverage for warning state transitions after overflow, successful processing, and repeated overflow.
  * Expanded validation for undersized contact storage across collision-processing scenarios.


rubengrandia:
Open question is how often the warning should emit. @Guirec-Maloisel , @chschuma-disney . We have it currently configured to warn only once. Alternative would be to reset and warn every step.

newton/_src/solvers/kamino/_src/geometry/primitive/narrowphase.py
```diff
@@ -369,30 +430,20 @@ def add_multiple_contacts(
             gid_AB = wp.vec2i(gid_1, gid_2)
             bid_AB = wp.vec2i(bid_1, bid_2)
 
-        # Safely increment the per-world active contact counter (see notes in _write_contact_unified_kamino in unified.py)
```
Guirec-Maloisel, line 372: The same pattern is also in unified.py for the unified collision pipeline, so we could probably call the new function there as well. Also, I think that's the place where there were comments about in which order to do the varios atomic_add() and atomic_sub(), as this was not obvious. It would be great to keep these in the new function.

newton/_src/solvers/kamino/_src/geometry/contacts.py
```diff
@@ -244,6 +238,9 @@ def _default_num_world_max_contacts() -> list[int]:
     Shape of ``(num_worlds,)``.
     """
 
+    contact_overflow_warning_emitted: wp.array[wp.int32] | None = None
```
Guirec-Maloisel: I'd maybe start this one with an underscore (_contact_overflow_warning_emitted) to indicate better that it's an internal flag managed automatically.

newton/_src/solvers/kamino/_src/geometry/contacts.py
```diff
@@ -1043,11 +1031,21 @@ def _convert_contacts_newton_to_kamino(
         wcid = wp.atomic_add(kamino_world_active, wid, 1)
```
Guirec-Maloisel: Is there a reason why the new function (reserve_contact_capacity) can't be used here as well?

Guirec-Maloisel (dismissed):
Regarding the open question (warning frequency): not really sure, but I think I'm leaning towards having this warning more frequently (once per time step with culled contacts), because by construction it can't really be a false positive. So I'd say it's worse if it gets potentially overlooked (because hidden in the middle of many lines about modules being loaded etc; with wp.printf() we can't send this through the proper warning channel with colored text); than to annoy some users (for which the fix should be relatively straightforward).

newton/_src/solvers/kamino/_src/geometry/primitive/narrowphase.py
rubengrandia: I didn't initially do this because the allocation for 1 contact looked different than the helper that was for multiple contacts. But I see now the multiple contact helper also generalizes to that case. Migrated note.

newton/_src/solvers/kamino/_src/geometry/contacts.py
rubengrandia: Same reason as above, the single contact case seemed different. Also re-using the helper here required generalizing the warning to be more vague. The original warning pointed to configs that do not apply in this workflow. But all acceptable consequences I think

rubengrandia:
> Regarding the open question (warning frequency): not really sure, but I think I'm leaning towards having this warning more frequently (once per time step with culled contacts), because by construction it can't really be a false positive. So I'd say it's worse if it gets potentially overlooked (because hidden in the middle of many lines about modules being loaded etc; with wp.printf() we can't send this through the proper warning channel with colored text); than to annoy some users (for which the fix should be relatively straightforward).

Sounds good. I updated it to reset the flag every simulation step. If it becomes too annoying we can always make it less verbose again.

Guirec-Maloisel (commented):
Thanks for the updates, looks good to me now! I think I can live with the warnings being slightly less specific.

Leaving approval to @chschuma-disney in case he has a different opinion on the warning frequency.

newton/_src/solvers/kamino/tests/test_geometry_primitive.py
```diff
@@ -1089,6 +1096,30 @@ def test_02_make_and_collide(self):
...
+        """Reset the primitive overflow warning flag between collision runs."""
+        builder = testing.make_shape_pairs_builder(shape_pairs=[("box", "box")])
+        model = builder.finalize(device=self.default_device)
+        data = model.data()
+        state = model.state()
+        contacts_small = ContactsKamino(capacity=1, device=self.default_device)
+        contacts_large = ContactsKamino(capacity=8, device=self.default_device)
+        pipeline = CollisionPipelinePrimitive(model=model)
+
+        pipeline.collide(data, state, contacts_small)
```
chschuma-disney, line 1109: Same here with wrapping these in `with self.assertLogs(level="WARNING"): ...`.

newton/_src/solvers/kamino/tests/test_geometry_contacts.py
```diff
@@ -1409,6 +1391,13 @@ def test_08_edge_cases(self):
             small_capacity,
             f"Active count {nc_small} must not exceed Kamino capacity {small_capacity}",
         )
+        self.assertEqual(int(kamino_small._data._contact_overflow_warning_emitted.numpy()[0]), 1)
+
+        convert_contacts_newton_to_kamino(model_nc, state_nc, contacts_nc, kamino_small)
+        self.assertEqual(int(kamino_small._data._contact_overflow_warning_emitted.numpy()[0]), 0)
+
+        convert_contacts_newton_to_kamino(model, state, contacts, kamino_small)
```
chschuma-disney: Could we wrap these calls where we expect a warning in `with self.assertLogs(level="WARNING"): ...` (not sure about the level for warp's printf) to avoid leaking to stdout?

chschuma-disney (changes requested):
Great work, looks mostly good to me. Just a small comment on capturing the output.

Guirec-Maloisel (dismissed):
Marking as approved from my side (in case my previous "Request changes" would block the merge)

newton/_src/solvers/kamino/tests/test_geometry_contacts.py
rubengrandia: Added the `assertLogs` for the existing python warning that this line emits. It does not capture the `wp.printf` though. I did not find a similar way to capture that warning (but we do assert our internal flag to check it was emitted).

newton/_src/solvers/kamino/tests/test_geometry_primitive.py
rubengrandia: This line only has a `wp.printf`, which we cannot assert like that.

chschuma-disney (approved):

Issue #3762, opened 2026-07-31
[BUG] newton:limitStiffness units disagree between Newton and newton-usd-schemas


andrewkaufman:
## Bug Description

Newton and `newton-usd-schemas` disagree about the units of `NewtonJointAPI`'s `newton:limitStiffness` / `newton:limitDamping`, and I would like guidance on which side is authoritative before a fix is written.

**The schema defines them in effort space.** From `NewtonJointAPI` in `newton-usd-schemas`:

> `newton:limitStiffness`: "When a DOF's position exceeds the range defined by the joint limits, a restoring effort is applied: `effort = limitStiffness * penetration` … **Units: effort / degrees** (angular DOFs) or effort / distance (linear DOFs)."

> `newton:limitDamping`: "**Units: effort * seconds / degrees** (angular DOFs) or effort * seconds / distance (linear DOFs)."

**Newton's constants describe the same quantity in acceleration space.** From `newton/_src/solvers/mujoco/constants.py`:

```python
DEFAULT_LIMIT_KD = 100.0
"""Newton damping gain equivalent to MuJoCo's implicit default joint-limit solref."""

DEFAULT_LIMIT_KE = 2500.0
"""Newton stiffness gain equivalent to MuJoCo's implicit default joint-limit solref."""
```

MuJoCo's implicit default `solreflimit` is `(timeconst, dampratio) = (0.02, 1.0)`, and `2500 = 1 / (0.02² · 1²)`, `100 = 2 / 0.02`. Those have units of s⁻² and s⁻¹ — a normalized/acceleration-space gain, not a torque per radian. Calling `2500` the "equivalent" Newton stiffness gain asserts a dimensionally different quantity than the schema's `effort / degrees`.

The two readings differ by the joint's effective inertia, so this is not a cosmetic wording issue — it changes simulation results.

### Why this matters now

`mujoco-usd-converter` 0.4.0 began authoring these attributes, deriving them from `mjc:solreflimit` with only a per-degree scale (`_impl/joint.py`):

```python
stiffness = 1.0 / (timeconst * timeconst * dampratio * dampratio)   # 2500 s⁻²
damping   = 2.0 / timeconst                                          # 100 s⁻¹
return to_newton_angular_gain(joint, stiffness), to_newton_angular_gain(joint, damping)
```

That matches Newton's `DEFAULT_LIMIT_KE` / `DEFAULT_LIMIT_KD` exactly, but not the schema's stated units.

[...]

PR #3804, opened 2026-08-05
Preserve Newton limit gains during MJCF import

vreutskyy:
## Description

Implement items 4 and 5 from #3762:

- preserve authored and implicit MJCF and MuJoCo-schema USD `solreflimit` through MuJoCo-specific attributes and provenance modes;
- leave Newton's generic force-space joint-limit gains at the configured builder defaults or generic schema values instead of deriving `2500/100`-style normalized values;
- keep native MuJoCo solref and generic Newton gains independently when both are authored;
- replace magic-value/tolerance edit detection with per-DOF provenance and import-time gain baselines, preserving gain edits made before or after solver construction;
- document the semantic split between native MuJoCo parameters and Newton-generic gains.

## Checklist

- [x] A Towncrier fragment has been added for the user-facing change

## Test plan

- `uv run --extra dev -m unittest newton.tests.test_import_mjcf newton.tests.test_import_usd newton.tests.test_mujoco_solver.TestMuJoCoSolverInvweightScaledSolref` (599 tests)
- `uvx pre-commit run -a`
- `uvx --from towncrier==25.8.0 towncrier build --draft --version X.Y.Z --date 2026-08-14`

## Bug fix

**Steps to reproduce:**

1. Configure non-default `ModelBuilder.default_joint_cfg.limit_ke` and `limit_kd` values.
2. Import an MJCF joint with an authored or implicit `solreflimit`.
3. Observe that the importer replaces the generic Newton gains with an inertia-independent conversion from the native MuJoCo constraint response.

**Minimal reproduction:**

```python
import newton

builder.default_joint_cfg.limit_ke = 4321.0
builder.default_joint_cfg.limit_kd = 76.0
builder.add_mjcf("""
<mujoco>
  <worldbody>
    <body>
      <joint type="hinge" solreflimit="0.03 0.9"/>
    </body>
  </worldbody>
</mujoco>
""")
model = builder.finalize()
assert model.joint_limit_ke.numpy()[0] == 4321.0
assert model.joint_limit_kd.numpy()[0] == 76.0
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved native MJCF/USD joint-limit `solreflimit` values separately from Newton stiffness and damping defaults.
  * Corrected force-space scaling when generic joint-limit gains are edited, with consistent CPU and GPU behavior.
  * Added validation and warnings for invalid solver-reference values while preserving valid defaults.
  * Improved handling of authored, default, and zero-valued solver-reference settings.

* **Documentation**
  * Clarified interactions between native solver values and Newton generic gains.
  * Documented updated MJCF and USD import behavior.


eric-heiden:
Thanks for tightening this up. Treating `solreflimit` as a native MuJoCo solver parameter instead of folding it into Newton's generic limit gains is the right call, and the explicit provenance mode makes the intent easy to follow.

Two things look worth another pass before merge: the USD import path still derives the generic gains from `mjc:solreflimit`, so the two importers now disagree for the same asset, and the solver-side gain snapshot is taken before the MJCF conversion runs, so gains configured before `SolverMuJoCo(...)` never take effect.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -3755,6 +3753,21 @@ def __init__(
...
+            np.array(model.joint_limit_ke.numpy(), copy=True) if model.joint_limit_ke is not None else None
+        )
+        self._joint_limit_kd_snapshot = (
+            np.array(model.joint_limit_kd.numpy(), copy=True) if model.joint_limit_kd is not None else None
+        )
+        mujoco_attrs = getattr(model, "mujoco", None)
+        solreflimit_mode = getattr(mujoco_attrs, "solreflimit_mode", None) if mujoco_attrs is not None else None
+        self._solreflimit_mode_snapshot = (
+            np.array(solreflimit_mode.numpy(), copy=True) if solreflimit_mode is not None else None
+        )
```
eric-heiden, line 4028: 🟠 This snapshot is taken in `__init__` before `_convert_to_mjc` runs, and `_update_solref_from_invweight0` only promotes `SOLREF_MODE_MJCF_DEFAULT` to `SOLREF_MODE_FORCE_SPACE` when the current gains differ from it. The first promotion pass happens during construction itself, inside `_convert_to_mjc` via `notify_model_changed(ModelFlags.ALL)`, so the gains always match the snapshot and the mode can never be promoted on that pass. Both the kernel and the CPU fallback then return early on `SOLREF_MODE_MJCF_DEFAULT` (`kernels.py` around line 2708, `solver_mujoco.py` around line 8669) before `ke`/`kd` are read at all.

The effect is that any limit gain configured before the solver exists is silently dropped.

<details>

<summary>Example</summary>

```py
builder.add_mjcf(mjcf_with_a_hinge_and_no_solreflimit)
model = builder.finalize()

model.joint_limit_ke.assign([5000.0])

# jnt_solref stays at (0.02, 1.0) instead of the invweight-scaled
# force-space pair, so the limit responds with MuJoCo's default
# stiffness while model.joint_limit_ke reports 5000.
```

</details>

Setting `builder.default_joint_cfg.limit_ke` before `add_mjcf` hits the same path, and the new test in `test_mujoco_solver.py` (around lines 11311-11321) captures the behavior: it asserts `initial_solref == [0.02, 1.0]` while `model.joint_limit_ke == 4321`.

On `main` the check was value-based against `DEFAULT_LIMIT_KE`, so 5000 differed from 2500 and the joint was promoted to force space on that first pass. `docs/solvers/mujoco.rst` (lines 182-184) still promises the MuJoCo default holds "until either generic gain is changed", without saying the change only counts if it happens after the solver is built.

Anchoring the provenance at import time rather than at solver construction time would close this. Either have `parse_mjcf` stash the gains it assigned alongside `mujoco:solreflimit_mode` so the comparison uses import-time values, or keep the solver-side snapshot but seed it from the importer-written defaults and state in the docs that gains must be changed after `SolverMuJoCo(...)`, followed by `notify_model_changed(ModelFlags.JOINT_DOF_PROPERTIES)`.

newton/_src/utils/import_mjcf.py
```diff
@@ -1948,46 +1968,13 @@ def parse_body(
...
-                    limit_ke = DEFAULT_LIMIT_KE  # From MuJoCo's default solref.
-                if limit_kd is None:
-                    limit_kd = DEFAULT_LIMIT_KD  # From MuJoCo's default solref.
+                # ``solreflimit`` is a native MuJoCo solver parameter, not a
+                # force-space gain. Preserve it through the custom attribute and
+                # keep Newton's generic joint-limit gains at their configured
+                # defaults instead of assigning an inertia-independent conversion.
+                limit_ke = default_joint_limit_ke
+                limit_kd = default_joint_limit_kd
+                warn_invalid_joint_solreflimit(joint_attrib)
```
eric-heiden, line 2025: 🔵 MJCF import no longer derives `limit_ke`/`limit_kd` from `solreflimit`, but the USD path still does. `_resolve_joint_limit_gain` in `newton/_src/utils/import_usd.py` (around lines 819-832) keeps returning the `mjc:solreflimit`-derived value as `limit_ke`/`limit_kd`, via `_solref_to_stiffness_per_rad` and `solref_to_stiffness_damping` in `newton/_src/utils/usd/schemas.py`.

So the same MuJoCo asset now lands on different generic gains depending on which importer reads it:

| Import path | Authored value | Resulting `model.joint_limit_ke` |
| --- | --- | --- |
| MJCF | `solreflimit="0.03 0.9"` | `builder.default_joint_cfg.limit_ke` (1e4 by default) |
| USD | `mjc:solreflimit = [0.03, 0.9]` | about 1371.7 |

That matters outside MuJoCo: `SolverXPBD`, `SolverFeatherstone` and the other solvers consume `joint_limit_ke`/`joint_limit_kd` directly, so joint-limit response for one asset now depends on the file format it was loaded from. The tests encode both behaviors at once, with `test_import_usd.py` still asserting the solref-derived gains (2500/100) and the updated `test_import_mjcf.py` asserting the builder defaults.

The new documentation text is carefully scoped to "MJCF import", so it is not wrong, just silent about the remaining divergence. Two ways to close it:

1. Apply the same rule in `_resolve_joint_limit_gain` / `SchemaResolverMjc` so USD also preserves `mjc:solreflimit` as a native parameter and leaves the generic gains alone. This could be a follow-up referenced from the changelog fragment.
2. Or add one sentence to `docs/solvers/mujoco.rst` (near line 790) noting that USD `mjc:solreflimit` still populates `joint_limit_ke`/`joint_limit_kd` while MJCF no longer does.

docs/solvers/mujoco.rst
```diff
@@ -176,11 +176,15 @@ solver restores MuJoCo's default ``solreflimit`` value ``(0.02, 1.0)``.
...
-used.
+``model.mujoco.solreflimit`` custom attribute and are not rescaled. Neither
+MJCF import nor USD import through the MuJoCo schema derives generic
+``joint_limit_ke`` or ``joint_limit_kd`` values from ``solreflimit``; those
+gains retain Newton builder defaults or values supplied by a generic schema.
+An authored native value wins in :class:`SolverMuJoCo`, while the generic gains
+remain available to other solvers. Imported joints that did not author
+``solreflimit`` keep MuJoCo's implicit default ``(0.02, 1.0)`` until either
+generic gain is configured or edited, including edits made before constructing
+the solver, at which point the Newton force-space scaling above is used.
```
eric-heiden: ⚪ Two small cosmetic things in this paragraph.

`:class:SolverMuJoCo` on line 183 is the only unqualified reference to the solver in this file. With `.. currentmodule:: newton` at the top, it resolves to `newton.SolverMuJoCo`, which is not re-exported from `newton/__init__.py`, so Sphinx renders it as plain literal text instead of a link. Line 156 in this same section already uses `:class:`~newton.solvers.SolverMuJoCo``.

The section also enumerates `model.mujoco.solreflimit` and `model.mujoco.solreflimit_mode`, but this change adds a third user-visible model attribute, `model.mujoco.solreflimit_gain_baseline`, which is exactly what makes the newly documented "including edits made before constructing the solver" behavior work. One sentence here would keep the three attributes documented together.

```suggestion
An authored native value wins in :class:`~newton.solvers.SolverMuJoCo`, while
the generic gains remain available to other solvers. Imported joints that did
not author ``solreflimit`` keep MuJoCo's implicit default ``(0.02, 1.0)`` until
either generic gain is configured or edited, including edits made before
constructing the solver, at which point the Newton force-space scaling above is
used. ``model.mujoco.solreflimit_gain_baseline`` records the generic gains seen
at import time so that edits made before construction can be detected.
```

jcarius-nv (approved):
Reviewed at exact head `768f86f568dd868c8cd7100e59e122b9818fcf26`, including the requirements and discussion in #3762 and the existing review history. I found no blocking correctness, specification, project-fit, or coding-standards issues.

The current head:

- keeps native MuJoCo `solreflimit` values separate from Newton's generic effort-space joint-limit gains;
- preserves authored native values and implicit MuJoCo defaults across MJCF and USD imports;
- uses import-time provenance to detect gain edits made before solver construction, including merged D6 joints; and
- covers the unauthored USD default case and MuJoCo CPU/MJWarp parity with focused regressions.

Local exact-head validation passed 47 focused tests covering solref handling and pre-construction provenance. All current CI checks are passing. Prior review concerns are addressed.

The remaining asset-level workaround cleanup discussed outside this PR is downstream follow-up and does not block this Newton runtime fix.

eric-heiden (approved):

PR #3810, opened 2026-08-05
Added option to render all worlds in one view for SensorTiledCamera

daniela-hase:
## Description

Add an opt-in `SensorTiledCamera` mode that renders all worlds into one shared camera view by traversing every BVH group with per-world display offsets. This matches the OpenGL visualizer layout
while preserving the existing per-world tiled rendering path as the default.

This also updates the tiled camera example with a runtime UI toggle, reuses the same image buffers by reshaping them between tiled and combined modes, and de-duplicates grouped BVH traversal setup
in the Warp raytrace kernels.

## Checklist

## Test plan

```bash
python -m compileall -q newton/_src/sensors newton/examples/sensors/example_sensor_tiled_camera.py newton/tests/test_sensor_tiled_camera.py
uv run -m newton.examples sensor_tiled_camera --viewer null --headless --test --num-frames 1
uvx pre-commit run -a

## New feature / API change

sensor = SensorTiledCamera(model, width=1024, height=1536)
render_config = SensorTiledCamera.RenderConfig(render_worlds_together=True)

sensor.update(
    state,
    camera_transforms,
    color_image=color_image.reshape((1, camera_count, height, width)),
    depth_image=depth_image.reshape((1, camera_count, height, width)),
    world_offsets=viewer.world_offsets,
    render_config=render_config,
)
```

## Summary by CodeRabbit

* **New Features**
  * Tiled-camera rendering can combine multiple worlds into a single view using configurable world offsets.
  * Combined rendering supports color, depth, lighting, and ray-traced outputs.
  * Output buffers automatically adapt to the selected number of rendered worlds.

* **Enhancements**
  * Added a configuration option for combined-world rendering.
  * Updated the tiled-camera example with a live rendering toggle and resolution details.

* **Tests**
  * Added coverage for world offsets, output dimensions, and detected scene objects.


newton/examples/sensors/example_sensor_tiled_camera.py
```diff
@@ -317,44 +369,86 @@ def get_camera_transforms(self) -> wp.array[wp.transformf]:
...
+            ],
+            dtype=np.float32,
+        )
+        correction *= 0.5
+        correction[newton.Axis.from_any(self.model.up_axis)] = 0.0
+        offsets -= correction
+        return wp.array(offsets, dtype=wp.vec3f, device=self.model.device)
+
     def test_final(self):
         self.render_sensors()
```
jcarius-nv, line 403: Please add the required test docstring. Newton's test guidelines require every test function or method, including `Example.test_final()`, to use a triple-double-quoted imperative summary. For example:

```python
"""Verify tiled and combined-world sensor outputs."""
```

jcarius-nv (commented):
## Fit discussion

This is a coherent extension of `SensorTiledCamera` and the existing `world_offsets` model. I'd like to challenge the scope before we commit to the public API:

- What concrete robotics or simulation workflow needs one perspective-correct composite sensor image rather than the existing per-world outputs or viewer offsets?
- Do all six output helpers need a new public `world_count` parameter, or could this be supported with a narrower public surface?
- Can you confirm, using a representative measurement or existing benchmark evidence, that this refactor does not regress the existing per-world rendering path? A new benchmark is not necessarily required. For the intended combined-mode world counts, it would also be useful to establish that visiting every world's BVH group per ray has acceptable cost.

The capability itself seems to fit; these answers would help validate that the public surface and performance cost are proportionate.

eric-heiden:
Nice feature! Gating the new code behind `wp.static(config.render_worlds_together)` keeps the existing render path clean.

The comments below are all small: an output allocation default that ignores the new flag, a docs gap around when `world_offsets` actually applies, and a per-frame allocation added to the default render path.

newton/_src/sensors/warp_raytrace/utils.py
```diff
@@ -295,119 +295,158 @@ def __warn_implicit_render_config_update(self, method_name: str, config_field: s
...
+        return int(world_count)
+
+    def create_color_image_output(
+        self, width: int, height: int, camera_count: int = 1, *, world_count: int | None = None
+    ) -> wp.array4d[wp.uint32]:
         """Create a color output array for :meth:`~newton.sensors.SensorTiledCamera.update`.
 
         Args:
             width: Image width [px].
             height: Image height [px].
```
eric-heiden, line 303: 🔵 I wonder if you actually need this `world_count` argument in these `create_*_image_output()` functions now? It wasn't needed before, and you have `render_config.render_worlds_together` available to query if that flag is true. So I would prefer we remove this argument if possible.

`Utils` already holds the config, so it could supply the default itself:

```python
def __resolve_output_world_count(self, world_count: int | None) -> int:
    if world_count is None:
        if self.__render_config is not None and self.__render_config.render_worlds_together:
            return 1
        return self.__render_context.world_count
    if world_count < 1:
        raise ValueError(f"world_count must be >= 1, got {world_count}")
    return int(world_count)
```

newton/_src/sensors/warp_raytrace/render_context.py
```diff
@@ -254,10 +259,18 @@ def render(
             width = camera_rays.shape[2]
             height = camera_rays.shape[1]
             camera_count = camera_rays.shape[0]
+            output_world_count = 1 if config.render_worlds_together else self.world_count
 
             if clear_data is None:
                 clear_data = RenderContext.DEFAULT_CLEAR_DATA
 
+            if world_offsets is None:
+                world_offsets = wp.array([], dtype=wp.vec3f, device=self.device)
```
eric-heiden: 🔵 This sentinel is rebuilt on every `render()` call that does not pass `world_offsets`, which today is every caller of `SensorTiledCamera.update()`. Its contents never depend on the call, so each frame pays for a numpy conversion, runtime and device resolution, and a fresh Python object on a path designed for per-step image rendering.

Can you just pass `world_offsets` as `None` to the kernels to prevent this allocation altogether? Or at least allocate this empty sentinel just once and reuse it.

newton/_src/sensors/warp_raytrace/utils.py
daniela-hase: Oh yeah, that's true, good point! Updated it in https://github.com/newton-physics/newton/pull/3810/changes/794a1fac6d8b467367947d49e1ae6e3f0f221b85

newton/_src/sensors/warp_raytrace/render_context.py
daniela-hase: Good catch, thank you! Update in https://github.com/newton-physics/newton/pull/3810/changes/794a1fac6d8b467367947d49e1ae6e3f0f221b85

newton/examples/sensors/example_sensor_tiled_camera.py
daniela-hase: Added the docstring in https://github.com/newton-physics/newton/pull/3810/changes/794a1fac6d8b467367947d49e1ae6e3f0f221b85 :)

jcarius-nv:
Re-reviewed at `ae68740e`.

The follow-up addresses the review feedback:

- adds the required `Example.test_final()` docstring;
- removes the extra public `world_count` parameters from the output helpers and derives the output count from the render config;
- avoids the per-frame empty-offset allocation on the existing rendering path; and
- corrects `__reshape_buffer_for_flatten()` to declare its actual `(buffer, worlds_per_row)` return value with bracket-style Warp array syntax.

Local validation passed:

- 41 selected tests, including `test_render_worlds_together_applies_world_offsets`;
- repository-wide pre-commit; and
- `git diff --check`.

The opt-in combined-world feature looks sound and the public surface is now proportionate. It would still be good to confirm, with a representative measurement or equivalent evidence, that the existing per-world path does not regress and that combined-mode traversal cost is acceptable for the intended world counts, but I do not consider that performance confirmation blocking for this PR.

Fresh CI is running on the pushed head. Thanks for the contribution!

jcarius-nv (approved):
Approved at `ae68740e`. The review feedback has been addressed and local validation passes. Performance confirmation would still be useful follow-up, but is not blocking this PR.

eric-heiden:
In our discussion with @mmacklin, @daniela-hase and @nvlukasz we decided to prefer using the Viewer classes if you want to see all worlds at the same time. Otherwise, a separate BVH needs to be constructed and a lot of overhead is introduced just for this feature. Let's close this PR for now.

PR #3811, opened 2026-08-05
Add fullscreen image rendering to ViewerGL

daniela-hase:
## Description

Add a `fullscreen` option to `ViewerBase.log_image()` / `ViewerGL.log_image()` so ViewerGL can draw a logged image as the main viewer surface for the current frame. This lets sensor outputs, such as
tiled camera color images, replace the 3D scene while keeping the ViewerGL UI available.

This PR also:
- Keeps fullscreen image logs separate from dockable image logs
- Preserves aspect ratio when rendering fullscreen textures
- Makes fullscreen image selection frame-scoped so stale images do not replace later 3D frames
- Updates the tiled camera sensor example with a GL-only main-view toggle
- Documents `fullscreen=True` usage and adds a Towncrier fragment

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_viewer_image_logger_class
uv run --extra dev -m newton.tests -k test_headless_capture_main_image_frame

Both commands passed locally.

## New feature / API change

import numpy as np
import newton

viewer = newton.viewer.ViewerGL(width=640, height=480, headless=True)
image = np.zeros((480, 640, 3), dtype=np.uint8)

viewer.begin_frame(0.0)
viewer.log_image("sensor_color", image, fullscreen=True)
viewer.end_frame()
```

## Summary by CodeRabbit

* **New Features**
  * Added support for displaying a logged image as the viewer’s main surface during a frame.
  * Supports tiled camera outputs while preserving UI overlays.
  * Added a tiled-camera option to show sensor color output as the main view.
  * Automatically returns to normal 3D rendering when no main image is selected.
  * Improved handling of missing or unavailable logged images.

* **Documentation**
  * Documented main-image display behavior and usage.


mmacklin:
Instead of a new function could we make this an option on the `viewer.log_image()` method?

In general we should name things with common prefixes so they're grouped together, e.g.: `viewer.log_image_main()`, but in this case I think an option is better, e.g.: "maximize=True" or "docked=True".

mmacklin:
Also a little surprised at the size of this change, if it's just adding a docked image log function then do we really need all the internal refactoring as well? Given the late stage of the PR it would be good to keep it as minimal as possible.

newton/tests/test_example_sensor_tiled_camera.py
eric-heiden, line 1: I don't think we need another test file just for this example? We have `test_examples.py` for this.

daniela-hase:
I reduced the size of the PR to the bare minimum, removing some of the refactoring I did and changed `log_main_image` to `log_image(..., fullscreen=True)`.

I kept the part of rendering the image as a texture directly via OpenGL instead of using the ImGui Window and forcing it to be a borderless fullscreen window that passes through mouse and keyboard events. Using ImGui for this would currently mean the fullscreen image gets hidden when the user hides the UI.

We could of course make an exception for the fullscreen window, like always show that one regardless of if the user shows or hides the rest of the UI, but that doesn't feel too great to me.

But, let me know what you think and I can make more changes to this :)

jcarius-nv (commented):
Thanks, Daniela, for addressing Miles's API and scope concerns. Moving this to `log_image(..., fullscreen=True)` and reducing the patch materially improves the direction.

I still found the following issues to address. I'm not super familiar with the code, so take it with a grain of salt and address those that make sense.

### Spec

- In [`example_sensor_tiled_camera.py`](https://github.com/newton-physics/newton/blob/b51c76886d1248955613a3ca79bf95085e0f673d/newton/examples/sensors/example_sensor_tiled_camera.py#L308-L316), the example logs the regular `color` image before logging the separate fullscreen image. The regular image is therefore auto-selected as a floating window, and `ImageLogger.draw()` no longer suppresses an image that is also used fullscreen, so the default UI overlays that window on the fullscreen surface. Please suppress the matching overlay or avoid logging both simultaneously, and cover both normal UI rendering and `get_frame(render_ui=True)`.

- [`test_viewer_get_frame.py`](https://github.com/newton-physics/newton/blob/b51c76886d1248955613a3ca79bf95085e0f673d/newton/tests/test_viewer_get_frame.py#L115-L146) catches every `Exception` raised while constructing `ViewerGL` and turns it into a skip. That can hide an actual initialization regression as a green test. Please catch only the known display/backend-unavailable cases and retain non-GL lifecycle/control-flow coverage.

- The PR description still documents removed pieces such as `log_main_image`, `uv_rect`, `LoggedImageTexture`, `hidden_name`, and deleted test modules. It also promises per-tile `compute_grid_layout` behavior, while the current [`render_texture()`](https://github.com/newton-physics/newton/blob/b51c76886d1248955613a3ca79bf95085e0f673d/newton/_src/viewer/gl/opengl.py#L1337-L1384) blits the packed atlas as a single image. Please update the description and test plan, and either preserve the advertised batched-image behavior or document the narrower behavior.

### Standards

- [`docs/guide/visualization.rst`](https://github.com/newton-physics/newton/blob/b51c76886d1248955613a3ca79bf95085e0f673d/docs/guide/visualization.rst) still describes `log_image()` output only as dockable windows. Please add a focused `fullscreen=True` example and document the per-frame requirement and backend limitations.

- The focused coverage for reset/3D fallback, missing textures, resource cleanup, and the example toggle was removed. Together with the broad skip above, the remaining GL happy-path test does not provide comprehensive regression coverage for the feature.

- The comment at [`viewer_gl.py:1740`](https://github.com/newton-physics/newton/blob/b51c76886d1248955613a3ca79bf95085e0f673d/newton/_src/viewer/viewer_gl.py#L1740) narrates the branch rather than explaining a non-obvious constraint. Please remove it or rewrite it to explain why the distinction matters.

- Please rename the PR to a concise imperative squash subject, for example: `Add fullscreen image rendering to ViewerGL`.

jcarius-nv (dismissed):
Thanks for addressing all review comments!

jcarius-nv (approved):

PR #3814, opened 2026-08-06
Add USD physics scene helper

eric-heiden:
## Description

Expose `newton.usd.get_physics_scenes()` so integrations can retrieve every `UsdPhysics.Scene` reported by OpenUSD in parser order and call typed schema methods directly.

`ModelBuilder.add_usd()` now returns the selected `physics_scene_path`. During import, Newton calls `UsdPhysics.LoadUsdPhysicsFromRange()` once and reuses that result for both scene selection and descriptor processing, so this change does not add another native physics parse to the import path. Calling the standalone helper performs one native parse of its requested range.

Scene construction uses `UsdPhysics.Scene.Get(stage, path)`. In cached-result microbenchmarks, this was about 3.1× faster for one scene and 4.2× faster for 100–1,000 scenes than constructing each schema from an intermediate prim.

## Checklist

## Test plan

```console
uv run python -m unittest newton.tests.test_import_usd.TestPhysicsSceneAccessor newton.tests.test_import_usd.TestImportSampleAssetsParsing.test_scene_path newton.tests.test_import_usd.TestImportSampleAssetsParsing.test_scene_gravity_enabled_parsing newton.tests.test_import_usd.TestImportSampleAssetsParsing.test_scene_gravity_non_unit_linear_unit newton.tests.test_import_usd.TestImportSampleAssetsParsing.test_scene_gravity_direction newton.tests.test_api.TestApi.test_builder_usd_signature_parity
uvx pre-commit run -a
uvx --from towncrier==25.8.0 towncrier build --draft --version 0.0.0 --date 2026-08-07
```

## New feature / API change

```python
from pxr import Usd

import newton
import newton.usd

stage = Usd.Stage.Open("scene.usda")
scenes = newton.usd.get_physics_scenes(stage)
gravity_magnitude = scenes[0].GetGravityMagnitudeAttr().Get() if scenes else None

selected_scene_path = result["physics_scene_path"]
```


jcarius-nv (dismissed):
Technically ready. This implements the right public seam for Isaac to query the same physics scene Newton imports, without duplicating Newton's discovery policy.

Standards:

- Minor, non-blocking: the `UsdPhysics.Scene` return-docstring reference could use a Sphinx `:class:` cross-reference.
- Minor, non-blocking: `ret_dict` would be clearer as `physics_results`, matching the helper.

Spec: no findings. The helper uses `UsdPhysics.LoadUsdPhysicsFromRange`, returns the same first scene Newton imports, preserves range/exclusion and instance-proxy behavior, and lets the importer reuse the parser result so it performs one native physics traversal.

Newton Fit: **Fits.** Isaac has a concrete need to select its solver before `ModelBuilder.add_usd()`, while Newton owns scene-selection policy. `newton.usd` is the appropriate existing seam, and the singular API can remain the primary-scene query if plural scene support is added later.

Validation:

- Focused USD-enabled tests: 9/9 passed locally.
- API generation and `git diff --check`: clean.
- Required checks: passing.
- Review threads: none.

One non-blocking downstream question: does Isaac MR !9278 test that the helper-selected scene produces the same solver choice as the subsequent Newton import? I could not access that internal MR from this environment.

newton/_src/usd/utils.py
```diff
@@ -2294,6 +2295,56 @@ def _get_deformable_point_masses(prim: Usd.Prim, read_attr: Callable[[Usd.Prim,
...
+    return stage.GetPrimAtPath(scene_paths[0])
+
+
+def _load_physics_scene_prim(
+    stage: Usd.Stage,
+    root_path: str = "/",
+    exclude_paths: Sequence[str] | None = None,
+) -> tuple[Usd.Prim | None, dict[Any, Any]]:
+    """Parse a USD range and return its selected scene prim and physics results."""
+    physics_results = UsdPhysics.LoadUsdPhysicsFromRange(
```
Milad-Rakhsha-NV: I was mainly hoping we can avoid calling loadUsdPhysicsFromRange. I was thinking to do something like 
```
result = {
    # ...
    "physics_scene_path": (
        str(physics_scene_prim.GetPath())
        if physics_scene_prim is not None
        else None
    ),
}
```
in add_usd() so we can do this on the sim side 
```
import_result = builder.add_usd(self.usd_stage, ...)
physics_scene_path = import_result["physics_scene_path"]

if physics_scene_path is not None:
    physics_scene_prim = self.usd_stage.GetPrimAtPath(physics_scene_path)

```

eric-heiden: OK, I can add this result argument to give you the prim path of the PhysicsScene.
But, to be clear, I reuse the results from `UsdPhysics.LoadUsdPhysicsFromRange`  here for the full USD traversal (this is in the `_get_physics_scene_prim_from_results` function) so there is no overhead here. Do you think I should remove the `newton.usd.get_physics_scene_prim()` function then?

newton/_src/usd/utils.py
```diff
@@ -2294,6 +2295,42 @@ def _get_deformable_point_masses(prim: Usd.Prim, read_attr: Callable[[Usd.Prim,
...
+
+    Args:
+        stage: The USD stage to search.
+        root_path: The root of the subtree to search.
+        exclude_paths: Prim paths whose subtrees should be excluded from the search.
+
+    Returns:
+        Prims with the ``UsdPhysics.Scene`` schema in parser order.
+    """
+    physics_results = UsdPhysics.LoadUsdPhysicsFromRange(
```
andrewkaufman, line 2326: We just return the actual `Usd.Prim` for the scene in the end, and do nothing with the bodies/colliders/joints, right? Why bother with the full `LoadUsdPhysicsFromRange` logic when all we really need is a standard traversal that includes instance proxies?

Also, would it be more useful to the end user if we return `list[UsdPhysics.Scene]` instead so they can directly call `result[0].GetGravityAttr().Get()` etc?

newton/_src/usd/utils.py
```diff
@@ -2294,6 +2295,42 @@ def _get_deformable_point_masses(prim: Usd.Prim, read_attr: Callable[[Usd.Prim,
...
+    """Get physics scene prims from parsed OpenUSD physics results."""
+    scene_results = physics_results.get(UsdPhysics.ObjectType.Scene)
+    if scene_results is None:
+        return []
+
+    scene_paths, _ = scene_results
+    return [stage.GetPrimAtPath(path) for path in scene_paths]
+
+
+def get_physics_scene_prims(
```
andrewkaufman: I'm curious about the client workflow for this. It seems a reasonable thing to want, but wouldn't they also want some dict associating scene->things they own? Any multi-scene stage is likely to have `simulationOwner` set on the bodies so each scene controls some subset of the physics prims.

newton/_src/utils/import_usd.py
```diff
@@ -583,6 +585,8 @@ class PhysicsMaterial:
         dict.fromkeys([*non_regex_ignore_paths, *_deformable_prims.native_physics_exclude_paths])
     )
     ret_dict = UsdPhysics.LoadUsdPhysicsFromRange(stage, [root_path], excludePaths=native_exclude_paths)
+    physics_scene_prims = usd._get_physics_scene_prims(stage, ret_dict)
+    physics_scene_prim = physics_scene_prims[0] if physics_scene_prims else None
```
andrewkaufman: We should spawn a new issue to handle `simulationOwner` properly... if there are multiple scenes in the stage its pretty likely that was intentionally done to control certain bodies with different solver settings.

newton/_src/usd/utils.py
Milad-Rakhsha-NV: yeah the reason I modified the original method returning a single scene to this was in anticipation of when we start handling multiple scenes for different simulation owners, since this is also part of the public API to also avoid future deprecation needs

newton/_src/utils/import_usd.py
Milad-Rakhsha-NV: yep we should work on that as part of the co-simulation developments;

newton/_src/usd/utils.py
eric-heiden: We use `LoadUsdPhysicsFromRange` here because we want to make sure we return exactly the PhysicsScene prims in the same order as they are parsed by `ModelBuilder.add_usd()`. 

I would be OK changing the return type to `list[UsdPhysics.Scene]` if @Milad-Rakhsha-NV has no objections? You can easily extract the original prims from these descriptors if you need them.

Milad-Rakhsha-NV: would this be fine?
https://github.com/Milad-Rakhsha-NV/newton/commit/cfc385d2d4345ad04bc5737e631a4813f360b4c1
if so we can cherry pick it in the current branch

Milad-Rakhsha-NV: > We use LoadUsdPhysicsFromRange here because we want to make sure we return exactly the PhysicsScene prims in the same order as they are parsed by ModelBuilder.add_usd().

I think this is a valid point, I used instanceProxies approach before seeing Eric's comment but just reverted that part in the above commit; and just return the `list[UsdPhysics.Scene]`

@eric-heiden could you please cherry-pick this when you can ? https://github.com/Milad-Rakhsha-NV/newton/commit/63db0a1182260ccf6ca18d02fc4f910371777fd1

eric-heiden: Thanks @Milad-Rakhsha-NV — I cherry-picked your change as 343afdd3, preserving your NV authorship. I added 5fd20249 on top to use `UsdPhysics.Scene.Get(stage, path)` instead of constructing the schema from an intermediate prim. In cached-result microbenchmarks this conversion was about 3.1× faster for one scene and 4.2× faster for 100–1,000 scenes. We still use `LoadUsdPhysicsFromRange`, so parser ordering and importer result reuse remain unchanged.

andrewkaufman: > `UsdPhysics.Scene.Get`

I did not even know that existed

>  this conversion was about 3.1× faster for one scene 

That is very unexpected to me... we might want to ask openusd folks why that would be... I bet a lot of code can get faster if this really holds

> 4.2× faster for 100–1,000 scenes

If we ever parse a stage with that many scenes I'm outta here :grimacing:

andrewkaufman (approved):

jcarius-nv (approved):

PR #3816, opened 2026-08-06
Validate mesh and particle topology

jcarius-nv:
## Description

A reported validation gap allowed malformed triangle-mesh and particle-topology indices to reach inertia computation, device upload, and solver-specific precomputation.

This change validates each topology at the Newton boundary that owns it:

- `Mesh` now rejects triangle index arrays whose length is not divisible by three or whose values are outside the vertex array. Validation uses the normalized arrays before inertia computation, vertex/index setters validate candidates before replacing the current data, and `finalize()` revalidates mutable arrays before native mesh creation.
- Convex-hull replacement stages and validates its new vertices and indices as one geometry update, so a valid hull can reduce the vertex count without being checked against stale connectivity.
- `ModelBuilder._validate_structure()` now checks the shape and bounds of spring, triangle, bending-edge, and tetrahedron connectivity before model arrays are created. Bending-edge opposite vertices retain the supported `-1` boundary sentinel, while edge endpoints must reference particles.
- Style3D coverage verifies malformed cloth connectivity is rejected before solver precomputation instead of duplicating checks in the hot device helper.

The root cause was that mesh and model construction assumed connectivity had already been validated by importers or builder helpers. The public containers remain mutable, however, and some data paths can bypass those earlier checks. Central validation makes the invariant explicit and produces an actionable `ValueError` before native code consumes malformed topology.

Valid models are unchanged. The existing `skip_validation_structure` option remains available for callers that intentionally trust their input and need to skip structural checks.

## Checklist

## Test plan

```console
uv run --extra dev -m newton.tests --serial-fallback --level test -p test_model.py
uv run --extra dev -m newton.tests --serial-fallback --level test -p test_solver_style3d.py
uvx --from towncrier==25.8.0 towncrier build --draft --version 0.0.0 --date 1970-01-01
```

Results:

- `test_model.py`: 130 tests passed.
- `test_solver_style3d.py`: 12 tests passed across CPU and CUDA.
- Pre-commit: all hooks passed.
- Towncrier: the `Fixed` fragment rendered successfully.

## Bug fix

**Steps to reproduce:**

1. Construct a `Mesh` whose triangle indices are malformed or outside the vertex array, or mutate a builder topology array to reference a nonexistent particle.
2. Finalize the mesh or model.
3. Observe that malformed connectivity reaches later inertia, adjacency, device-upload, or solver code instead of being rejected at the owning boundary.

**Minimal reproduction:**

```python
import newton

vertices = [
    (0.0, 0.0, 0.0),
    (1.0, 0.0, 0.0),
    (0.0, 1.0, 0.0),
]

# Now raises ValueError before inertia or native mesh creation.
newton.Mesh(vertices, [0, 1, 3])
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for triangle-mesh connectivity and particle topology, including malformed, negative, and out-of-range indices.
  * Invalid topology is now rejected before simulation data creation and solver precomputation.
  * Empty mesh connectivity and valid edge boundary markers remain supported.
  * Geometry updates now preserve valid data when invalid replacements are rejected, while refreshing derived collision data.

* **Tests**
  * Added coverage for topology validation, boundary markers, cache invalidation, and early rejection of invalid solver input.


nvtw (dismissed):
AI found one smaller issue, feel free to address or not:

P2 – Lossy casts bypass validation (3816.diff.txt:52-54, 118-127): indices are converted to integers before validation. NumPy silently turns 2.9 into 2, and an int64 value 4294967298 into int32(2). Malformed/out-of-range topology can therefore be accepted with altered connectivity.
Validate the original values for integer-ness and bounds before narrowing to int32/int64, and add regression tests for fractional and oversized indices.

nvtw (approved):

PR #3817, opened 2026-08-06
Remove PR close cancellation workflow

jcarius-nv:
## Description

Remove the dedicated pull-request-close workflow that searches for and
cancels queued or in-progress runs.

Newton's PR workflows already use concurrency groups with
`cancel-in-progress: true`, which handles the higher-value case of cancelling
superseded runs. Maintaining a separate close-time cancellation path adds API
filtering complexity and platform-specific edge cases for comparatively limited
savings.

## Checklist

No changelog fragment is needed because this changes repository CI behavior
rather than user-facing Newton behavior.

## Test plan

- `uvx --python 3.12 pre-commit run -a`
- `git diff --check`
- Confirm no remaining repository references to `pr_closed.yml` or its
  dedicated regression test.


shi-eric:
I recommend changing this PR instead to delete the workflow `pr_closed.yml`. I believe I was the one who added it as a cost optimization, but I think it's not worth the complexity and edge  case detection for the money saved in the rare event a pull request is closed while the AWS workflows are still running. There are more robust mechanisms in place already to deal with cost optimizations (e.g. concurrency cancellation) and those are providing a lot more value.

shi-eric (approved):

PR #3818, opened 2026-08-06
Harden untrusted file loading

jcarius-nv:
## Description

Harden three existing file-loading boundaries without adding public API, changing cache locations, or adding dependencies:

- Validate cooked SDF cache entries before any array is uploaded to a GPU texture. The loader now requires the cache marker/version, exact stored dtypes and ranks, producer-consistent array shapes, valid scalar ranges, and consistent subgrid occupancy/slot bounds. Malformed entries, including corrupt or truncated archives, remain ordinary cache misses, so Newton recooks and replaces them through the existing path. Safe overprovisioned packed textures remain valid.
- Restrict viewer-recording `numpy.*` tags to concrete NumPy numeric scalar classes that the serializer can emit. `numpy.ndarray` keeps its dedicated path, while callable and unknown names are rejected rather than invoked.
- Normalize USD reference paths under both POSIX and Windows path rules, then resolve every cache destination canonically and require path-aware containment beneath the selected cache root before writing. Raw nested references are checked before their parent path is prefixed, so drive-qualified and UNC paths cannot hide beneath a local subdirectory. For reference assignments handled by the recursive resolver, any path that cannot be localized safely causes resolution to fail with `ValueError`; no asset path is returned. Downloaded layer bytes may remain in the selected cache directory after failure and must still be treated as untrusted input. This rejects backslash traversal, mixed separators, drive-qualified paths, UNC paths, prefix collisions, and symlink escapes while preserving valid nested references such as `../common.usd` when they remain inside the cache.

### Root causes and rationale

The SDF cache loader previously checked only the format version and cast the remaining fields into the expected in-memory types. A cache entry could therefore supply mismatched arrays or out-of-range packed subgrid slots that disagreed with the coarse-grid metadata, while a corrupt ZIP member could raise instead of becoming a miss. Validating the private on-disk schema and treating the specific corrupt-archive error as a miss keeps malformed data away from native texture consumers while preserving the existing caller-selected cache contract.

The recording deserializer previously resolved any `numpy.*` suffix with `getattr()` and immediately called the result. Recording files only need the concrete `np.number` subclasses emitted by the serializer, so open-ended callable dispatch was unnecessary.

The USD downloader screened reference strings with POSIX lexical checks but wrote them with native filesystem operations. That left Windows separators, drives, UNC roots, and filesystem indirection outside the model used by the check. Skipping a rejected download also left its asset path in cached layer text, allowing a later `Usd.Stage.LoadAll` open to resolve it independently. The new helper first rejects rooted/drive-qualified Windows forms, normalizes both separator styles, combines [`realpath`](https://docs.python.org/3/library/os.path.html#os.path.realpath) with path-aware [`commonpath`](https://docs.python.org/3/library/os.path.html#os.path.commonpath) containment, and fails the resolution rather than attempting to reconstruct or partially neutralize authored USD composition.

### Impact

Malformed SDF caches are ignored and regenerated instead of reaching GPU texture creation. Viewer recordings continue to round-trip legitimate NumPy arrays and numeric scalars but reject callable/unknown NumPy tags. Remote USD assets continue to resolve valid nested and absolute HTTPS references, but resolution fails if a recognized reference would select a destination outside the download cache. The function returns no asset path in that case, although downloaded bytes may remain in the selected cache directory.

No example cache location or directory permissions change. This PR does not introduce a per-user cache, mutate caller-supplied directory modes, or add dependencies.

## Checklist

## Test plan

```console
uv run --extra dev -m unittest newton.tests.test_sdf_disk_cache newton.tests.test_recorder newton.tests.test_import_usd.TestResolveUsdFromUrl
env UV_CACHE_DIR=/tmp/newton-file-cache-uv-cache UV_TOOL_DIR=/tmp/newton-file-cache-uv-tools PRE_COMMIT_HOME=/tmp/newton-file-cache-pre-commit-cache uvx --python 3.12 pre-commit run -a
uvx --from towncrier==25.8.0 towncrier build --draft --version 1.5.0 --date 2026-08-06
```

The focused resolver, pure SDF-cache, and recorder suites pass, as does the full pre-commit suite. GitHub CI passes across Ubuntu x86_64 and ARM, Windows, macOS, and AWS GPU tests. Focused regressions cover CRC-corrupt and truncated archives, nested drive/UNC references, a valid nested `../common.usd` reference, fail-closed traversal and symlink cases, and mixed reference-list and prim-path syntax. The behavioral regressions were first run against the previous implementation and reproduced the corresponding failures.

## Bug fix

**Steps to reproduce:**

1. Place a corrupt or truncated SDF cache archive, or an entry with the expected version but inconsistent array metadata or an encoded subgrid slot outside the packed texture.
2. Deserialize a viewer-recording value tagged as a NumPy callable such as `numpy.ones` or `numpy.fromfile`.
3. Resolve a remote USD layer, including a nested layer, containing a backslash-relative, drive-qualified, UNC, mixed-separator, or symlink-routed reference.
4. Observe that the old paths raised on corrupt archives, accepted malformed cache fields, invoked the named NumPy callable, selected a filesystem destination using semantics that the lexical check did not model, or retained a skipped unsafe reference in the cached layer text.

**Minimal reproduction:**

```python
from newton._src.viewer.viewer_file import deserialize

deserialize(
    {"__type__": "numpy.ones", "value": [1024]},
    lambda value, _path: value,
)
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of malformed or inconsistent SDF cache data.
  - Blocked unsafe USD cache paths, traversal attempts, and symlink escapes.
  - Prevented unsupported NumPy scalar types from being deserialized.
  - Added safer handling for Windows-style USD references and downloads.

- **Tests**
  - Expanded coverage for invalid caches, references, paths, and serialized types.

- **Documentation**
  - Added changelog details covering these validation improvements.


nvtw (changes requested):
AI found 2 issues:

[P1] Handle reference-list syntax — newton/_src/utils/import_usd.py:5329

The regex only recognizes singular references = @path@. Valid USDA such as references = [@../secret.usd@] is untouched, allowing Usd.Stage.LoadAll to resolve a path outside the cache. This bypasses the intended security boundary.

[P2] Remove optional referenced prim paths — newton/_src/utils/import_usd.py:5326

For references = @../secret.usd@</Root>, replacement produces invalid USDA:

references = []</Root>
The full assignment, including the optional prim path, must be neutralized.

nvtw (dismissed):
AI found some corner cases, feel free to ignore:

  - import_usd.py silently deletes valid internal references from mixed reference lists. For example, valid USDA:
    references = [</Local>, @safe.usd@</Remote>]
    becomes:
    references = [@safe.usd@</Remote>]

    The new regex reconstruction preserves only @...@ items. This can change scene composition. The rewriter should
    preserve unmatched internal references and needs a regression test.

  - test_reference_symlink_escape_is_rejected fails on Windows with WinError 1314. Checking hasattr(os, "symlink") does
    not mean the process has symlink privileges. The test should skip when symlink creation is denied.

jcarius-nv:
Thanks for the maintainability feedback. We reconsidered the approach and narrowed the PR accordingly. The resolver now validates referenced assets and fails closed when one cannot be safely localized, while preserving authored USD composition instead of attempting to comprehensively rewrite reference syntax. This keeps the security boundary explicit without taking on a partial USDA rewriter that would be difficult to maintain. We also documented the packed SDF slot sentinel layout. Focused resolver, SDF cache, and recorder tests, plus the full pre-commit suite, pass locally.

nvtw (approved):
I think it's good now, thanks

Issue #48, opened 2025-05-07
Add shape flags


eric-heiden:
Similar to the particle flags (see `Model.particle_flags`), we need flags for the shapes, i.e. `Model.shape_flags`.
They will control the visibility, whether they collide with other shapes and with the ground:

```py
SHAPE_FLAG_VISIBLE = wp.constant(wp.uint32(1 << 0))
SHAPE_FLAG_COLLIDE_SHAPES = wp.constant(wp.uint32(1 << 1))
SHAPE_FLAG_COLLIDE_GROUND = wp.constant(wp.uint32(1 << 2))
```
The `Model.shape_flags` array will therefore subsume `Model.shape_visible`, `Model.shape_ground_collision`, and `Model.shape_shape_collision`.

Users can apply their own flags to be passed to their custom solvers, collision handlers, renderers, etc.
We can do the same for joints, bodies, and articulations.

vreutskyy:
@eric-heiden @adenzler-nvidia 
There's a problem with this though. `Model.shape_visible` is a `wp.array`, while `Model.shape_ground_collision` and `Model.shape_shape_collision` are simple Python lists and are indexed on CPU (in `find_shape_contact_pairs()`) which `wp.array` doesn't support. We can't easily combine them. Or do I miss some point?

eric-heiden:
I think we can still query the flags from `ModelBuilder.find_shape_contact_pairs` (where the `shape_flags` will be a Python list), just need to use binary checks to figure out the flags:
```py
if self.shape_flags[shape_a] & SHAPE_FLAG_COLLIDE_SHAPES:
   ...
```
Having the flags as Warp array on the `Model` would also allow us to use them in the new collision library.

vreutskyy:
Hmm. But it's `Model.find_shape_contact_pairs`. and it's a Warp array there, not a Python list. What do I miss?

eric-heiden:
Oh indeed, I missed that! But I think it can be moved to `ModelBuilder`, or at least get refactored that way until the Warp arrays `shape_contact_pairs` and `shape_ground_contact_pairs` are constructed.

vreutskyy:
There're 2 more places where it tries to read that array. In MuJoCoSolver.convert_to_mjc.add_geoms and SimRenderer.populate

eric-heiden:
Fixed in https://github.com/newton-physics/newton/pull/69

Issue #47, opened 2025-05-07
Include floating-base dofs in control inputs


eric-heiden:
Modify the `Control` class handling so that it includes:

```py
joint_f = wp.zeros(model.joint_dof_count, dtype=wp.float32)
joint_ctrl = wp.zeros(model.joint_dof_count, dtype=wp.float32)
```

This will replace the previous `Control.joint_act` array.

Note that this new joint torque and control array will match `joint_qd` instead of having `joint_axis_count` dimensions.
This means we will include torques and PD control inputs for floating-base (free) joints in these arrays.

With these changes the solvers need to be adapted.

Issue #44, opened 2025-05-06
use_mujoco -> mujoco_cpu


erwincoumans:
The use_mujoco boolean is unclear. Can we modify it to mujoco_cpu to make it more explicit?

Issue #43, opened 2025-05-06
Create free joints for non-static bodies in ModelBuilder.finalize()


eric-heiden:
If a body has nonzero mass (dynamic) it should be simulated as a rigid body but the MuJoCo and Featherstone solver will only do that if they have a free joint assigned to them. We should have `ModelBuilder.finalize()` to automatically add free joints for such bodies if they have no connected joints.

j-pos:
Decision made not to include this as per sync with @eric-heiden 
Closing.

PR #3821, opened 2026-08-06
Use newton-assets with converter 0.5.0 structured USD

andrewkaufman:
## Description

Points `NEWTON_ASSETS_REF` at [newton-physics/newton-assets#48](https://github.com/newton-physics/newton-assets/pull/48), which regenerated the eight structured USD menagerie robots with `mujoco-usd-converter` 0.5.0, replacing output from 0.1.0a8.

Two changes reach Newton:

- Bodies whose mass came from visual geoms inside `inertiagrouprange` now carry MuJoCo's compiled mass rather than accumulating it from disabled colliders. The Allegro palm goes from 0.35039040 kg to 0.35158747 kg, matching the source MJCF's compiled 0.35158749 kg to float32 precision.
- `newton:limitStiffness` and `newton:limitDamping` are no longer authored, per the resolution of #3762. `SolverMuJoCo` reads joint limits from the authored `mjc:solreflimit`, and other solvers fall back to their own limit defaults.

The Menagerie pin inside newton-assets is unchanged, so the robot sources are the same; only the converter version moved. Bumping that pin to match `MENAGERIE_REF` is a follow-up.

## Checklist

## Test plan

```
uv run --extra dev -m unittest newton.tests.test_menagerie_usd_mujoco
uv run --extra dev -m unittest newton.tests.test_robot_composer
uv run --extra dev -m unittest newton.tests.test_anymal_reset
uv run --extra dev -m unittest newton.tests.test_import_usd
```

`test_menagerie_usd_mujoco` was run on this branch both with and without the ref bump, on the same revision and machine:

| `NEWTON_ASSETS_REF` | result |
| --- | --- |
| `261cd1f` (before) | 44 tests, 2 failures + 1 error |
| `26c7f5c` (this PR) | 44 tests, 2 failures + 1 error |

The failure sets are identical, so the bump introduces no regressions. Those three — `ApptronikApollo` and `BoosterT1` `test_dynamics`, plus a `Robotiq2f85V4` `IndexError` — are pre-existing on `main` with the currently pinned assets.

The other asset-dependent suites pass: `test_robot_composer` (1 test), `test_anymal_reset` (4 tests), `test_import_usd` (302 tests).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved physical accuracy for structured USD robot assets by using compiled body masses for visual geometry.
  * Updated solver-specific joint-limit settings and defaults.
  * Improved compatibility when importing and simulating Kamino DR legs robot assets.

* **Chores**
  * Refreshed the bundled Newton asset collection and pinned the DR legs asset revision.
  * Updated support for the MuJoCo USD converter to version 0.5.0.


changelog/+structured-usd-assets-4c7e91b2.changed.md
```diff
@@ -0,0 +1 @@
+Update the pinned `newton-assets` revision so the eight structured USD menagerie robots come from `mujoco-usd-converter` 0.5.0. Bodies whose mass came from visual geoms inside `inertiagrouprange` now carry MuJoCo's compiled mass rather than accumulating it from disabled colliders, and `newton:limitStiffness` and `newton:limitDamping` are no longer authored, so `SolverMuJoCo` reads joint limits from the authored `mjc:solreflimit` and other solvers fall back to their own limit defaults.
```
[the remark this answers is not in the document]
andrewkaufman: Added in df5de9f1 — the fragment now points authors of custom structured USD at `mjc:solreflimit` for `SolverMuJoCo` and explains why the Newton-generic gains should stay unset.

One correction to the framing though: this PR changes only `NEWTON_ASSETS_REF` and touches no parsing code, so existing custom assets are unaffected and there is no migration for them to perform. I verified the resolution order in `_resolve_joint_limit_gain` — an authored `mjc:solreflimit` yields `SOLREF_MODE_RAW` and is used verbatim, authored Newton-generic gains yield `SOLREF_MODE_FORCE_SPACE`, and with neither the MuJoCo defaults apply. That behavior is already on `main` and is unchanged here. The guidance is worth stating for people authoring their own assets, so I have added it as such rather than as a migration step.

eric-heiden:
🔵 Nice cleanup, moving the menagerie robots onto the converter 0.5.0 structured USD layout makes the asset story a lot more consistent.

One blocker before this can land: the new asset pin also pulls in rewritten DisneyResearch assets that break several in-repo Kamino tests and the default Kamino benchmark. Details in the inline comment.

newton/_src/utils/download_assets.py
```diff
@@ -20,7 +20,7 @@
 # commit.  Update these SHAs when assets change upstream and the new versions
 # have been validated against Newton's test suite.
 NEWTON_ASSETS_URL = "https://github.com/newton-physics/newton-assets.git"
-NEWTON_ASSETS_REF = "261cd1f429619d8ef4f546bd788ab9dea906b5e1"
+NEWTON_ASSETS_REF = "26c7f5ccaadd9a163f4d6c8f867c789b3ac57d83"
```
eric-heiden, line 23: 🔴 This pin bump brings in more than the eight structured USD menagerie robots described in the commit subject and the changelog. It also spans 5 upstream commits that rewrite 81 files under `disneyresearch/`, and those rewrites break the Kamino USD import path.

Running the same test code against both asset revisions:

| Asset ref | Result |
| --- | --- |
| `261cd1f` (current pin) | ran 6, passed 6 |
| `26c7f5c` (this PR) | ran 6, passed 1, failed 2, errored 3 |

The five regressions:

| Test | Failure |
| --- | --- |
| `test_import_model_dr_legs` | `ValueError: cannot reshape array of size 1 into shape (3)` |
| `test_import_model_dr_legs_with_meshes_and_boxes` | same `ValueError` |
| `test_03_model_conversions_dr_legs_from_usd` | same `ValueError` |
| `test_import_model_dr_legs_with_boxes` | `AssertionError: 37 != 36` (joint count) |
| `test_02_model_conversions_dr_testmech_from_usd` | `RigidBodiesModel.label are not equal` (body ordering) |

**Root cause:** `dr_legs.usda` goes from 31 to 81 `UsdGeom.Mesh` prims, and four new `merged_collision` prims (`lowerleg_link_{l,r}_{i,o}`) declare `normals` without an authored value, so `IsDefined()` is `True`, `HasAuthoredValue()` is `False`, and `Get()` returns `None`. The Kamino USD importer wraps that in `np.array(...)`, which yields size 1, and the reshape to `(-1, 3)` throws. The same crash hits the default Kamino benchmark problem `dr_legs`, whose `builder_fn()` goes from OK to `ValueError`.

Worth noting this is not a uniform regression. `ModelBuilder.add_usd` on `dr_legs.usda` actually improves (orphan-joint `ValueError` before, OK now), and `newton.examples.kamino.example_kamino_robot_dr_legs` constructs fine at both refs. The breakage is specific to the Kamino `USDImporter` path.

The comment directly above this line states the contract here: "Update these SHAs when assets change upstream and the new versions have been validated against Newton's test suite."

**Suggested next step:** either hold the DisneyResearch asset changes out of this bump, or fix it in the same PR by making the Kamino USD importer treat declared-but-unauthored `normals` as absent (fall back to computed normals) and updating the two topology-sensitive expectations (joint count 36 to 37, and the `dr_testmech` body ordering).

andrewkaufman: Confirmed, thanks — you're right on all of it, and my validation was too narrow. I reproduced your numbers with a proper control: the current pin gives 78/78 OK on `test_utils_io_usd` + `test_core_model`, the new pin gives 2 failures + 3 errors. I'd only run the menagerie and USD-import suites and never touched Kamino.

Also worth correcting my own PR description: the bump spans **six** newton-assets commits, not just #48 — #42, #43, #44, #46, #47 and #48, 81 files under `disneyresearch/`. I'll fix that wording.

I split the causes into two, because they're different problems:

**The three `ValueError` crashes — [#3845](https://github.com/newton-physics/newton/pull/3845).** Worse than it looked: `normals` is a built-in attribute of the `UsdGeom.Mesh` schema, so `IsDefined()` is true for *every* mesh prim regardless of authoring. Kamino's importer therefore could not load **any** mesh without authored normals; `dr_legs` is just the first asset we've pinned that contains some. I fixed it by delegating both mesh-read sites to `newton._src.usd.utils.get_mesh` rather than patching the guard, since the duplicate parser is the actual problem — that also picks up `primvars:normals`, indexed primvars, face-varying interpolation and n-gon support, none of which that path had.

**The two topology failures — [#3844](https://github.com/newton-physics/newton/issues/3844).** These are not really about the assets. #42 added `PhysicsArticulationRootAPI` to the DR models; `add_usd` is articulation-aware and Kamino's importer isn't, so the two paths diverge:

| `dr_testmech.usda` | `ArticulationRootAPI` | `add_usd` order | Kamino order |
| --- | --- | --- | --- |
| `261cd1f` | 0 | `base, driver, link1, ...` | identical |
| `26c7f5c` | 1 | `base, cart, clevis, ...` | unchanged |

Kamino's ordering never moved — `add_usd` did. And on the joint count, `dr_legs_with_boxes.usda` has 36 joint prims and 31 bodies at *both* revisions; the 37th is `world_to_pelvis`, synthesized for the new articulation root. So that expectation is legitimately 37 now, but the body-ordering one isn't an expectation at all — `test_02_model_conversions_dr_testmech_from_usd` asserts two import paths agree with each other, so there's no number to update. That needs a call from your side, which is why I filed it rather than picking one.

On holding the DisneyResearch changes out: I don't think that's available. `NEWTON_ASSETS_REF` is a single SHA into newton-assets history, so there's no way to take #48 without #42–#47 short of an orphan commit off `main`, which defeats pinning to main.

Plan for this PR: rebase on #3845, update the three `num_joints` expectations to 37, correct the description and changelog, and re-validate across Kamino plus the suites I originally ran. It stays blocked on #3844 for the two ordering assertions — happy to follow whichever direction you and the DisneyResearch folks prefer there.

andrewkaufman:
Status on the Kamino regressions blocking this, measured today against `main` + #3871 + this pin.

**Two of the five are resolved.** #3845 merged as `ae5ef001a`, which clears the three `ValueError` crashes. #3871 makes the model comparison order-independent and fixes `test_02_model_conversions_dr_testmech_from_usd`.

**One remains, and it is not what #3844 describes.** `test_03_model_conversions_dr_legs_from_usd` still fails, in two stages:

1. Only test_02 gained `allow_reordering=True` in #3871; test_03 did not, so it fails first on body label ordering.
2. Adding the flag to test_03 locally moves the failure to `Label sets differ in world 0`:

```
First differing element 36:
'joint_1'
'world_to_/dr_legs/pelvis'
```

The two import paths give the pelvis base joint **different names** — Kamino's importer calls it `joint_1`, `add_usd` calls it `world_to_/dr_legs/pelvis`. Elements 0-35 are identical, so this is one joint, named two ways, not an ordering difference. Order-independent matching cannot fix it: `_label_mapping` matches on labels, so differing label *sets* fail before any permutation is computed.

This was invisible until today because test_03 previously died earlier, on the `normals` `ValueError` that #3845 fixed.

For the record, the rest is clean at this pin: the remaining 20 tests in `test_kamino_core_model` pass, and the wider Kamino suite shows only the pre-existing `test_10_dr_legs_dvi_tipped_contact_does_not_creep`, which also fails on unmodified `main`.

So the chain here is: #3871 lands (currently red on all four platforms, fix offered there) → test_03 gets `allow_reordering=True` → someone decides what the base joint should be called. @chschuma-disney given your point on #3844 that the Kamino USD importer is legacy and shouldn't dictate changes to the Newton importer, is renaming on the Kamino side the direction you'd want here too?

chschuma-disney:
> So the chain here is: #3871 lands (currently red on all four platforms, fix offered there) → test_03 gets `allow_reordering=True` → someone decides what the base joint should be called. @chschuma-disney given your point on #3844 that the Kamino USD importer is legacy and shouldn't dictate changes to the Newton importer, is renaming on the Kamino side the direction you'd want here too?

Yes, the Kamino importer should be the one to change. For this case, there's actually already a flag for the `importer.import_from(...)` call: `use_articulation_root_name=False`
That should resolve the naming issues, though it looks like there might be another floating precision issue that might require an update of the `atol`/`rtol` tolerances in the test.

@andrewkaufman Do you want me to address these issues in a separate PR, or will you add the changes to this one?

andrewkaufman:
Thanks — `use_articulation_root_name=False` was the right pointer, and I've taken the changes onto this PR rather than splitting them out. Rebased onto current `main` at the same time, since this branch had drifted 49 commits.

**test_03 now passes.** Your flag fixed the naming, and behind it were two more divergences of the same kind — both of which turned out to have an existing option too, so the comparison stays strict rather than losing attributes:

| divergence | resolved by |
| --- | --- |
| `add_usd` honours `physics:approximation`, Kamino's importer does not: 8 collision meshes became `CONVEX_MESH` and 4 became `BOX` on one side only | `skip_mesh_approximation=True` on `add_usd` |
| the assets now use mesh colliders, which Kamino's importer leaves non-collidable by default — 12 shapes had `VISIBLE` where `add_usd` had `VISIBLE\|COLLIDE_SHAPES\|COLLIDE_PARTICLES` | `meshes_are_collidable=True` |
| base joint named after the articulation root vs numbered | `use_articulation_root_name=False` |
| bodies and joints stored in a different order by each path | `allow_reordering=True` |

That collidability difference alone accounted for `num_collidable` (15 vs 3 — exactly the 12), its pair count (95 vs 3) and both minimum-contact totals (3448 vs 24), so all four fall out once the flag is set rather than needing exclusions.

**On the tolerances:** it isn't a precision issue. What remains is one joint out of 37 — the base joint itself:

```
[0] joint_1
    converted (newton add_usd): [0. 0. 0.]
    kamino importer:            [-0.018934786  0.0002243507 -0.0073442967]
    |delta| = 0.020310 m
```

20 mm, with a relative difference of 1.0 because one side is exactly zero. An `atol` able to absorb that would have to be 2 cm on joint anchors, which would blind the test to real errors, so I've excluded `B_r_Bj` and `q_j_0` with the same `TODO` wording `test_04` already uses for this family. Happy to swap that for a tolerance if you would rather, but it looked like the wrong lever.

The DR legs example also needed `newton.use_coord_layout_targets = True`, matching `example_kamino_robot_anymal_d.py`: the new base joint makes the coordinate and DOF counts differ, and the example suite runs with `-W error::DeprecationWarning`.

---

**One failure left, and I don't think it's mine to silence.** `test_10_dr_legs_dvi_tipped_contact_does_not_creep` passes at the current pin and fails at this one — the tipped base drifts 0.0269 m against a 0.008 m threshold.

It is not the mesh approximation; forcing raw meshes changes it by 5e-5 (0.026897 → 0.026850). The robot itself is materially different between the two asset revisions:

| | current pin | this pin |
| --- | --- | --- |
| total mass | 2.4708 | 2.2921 |
| base mass | 0.6032 | 0.4379 |
| base inertia diag | 0.00161, 0.00108, 0.00142 | 0.00114, 0.00028, 0.00110 |
| shapes | 34 | 84 |
| collidable shapes | 3 | 15 |

A 27% lighter base with roughly a quarter of its previous `Iyy`, and five times the collidable geometry, is going to land differently when you tip it on its side. This is the rewritten DisneyResearch asset that @eric-heiden flagged, arriving through the same SHA — it is not something the converter reconversion touched.

So it needs a judgement I can't make: either the rewritten asset is what you want and the DVI thresholds want re-tuning against it, or the mass and inertia changes are themselves unintended and belong upstream in newton-assets. @chschuma-disney @eric-heiden — could one of you take a view? Everything else in the Kamino suite passes at this pin (679 tests, that one failure).

chschuma-disney:
There are some issues with the newer model on the assets repo that we're in the process of analyzing, but it has been a bit of a rabbit hole and we've not yet found a solution. For some reason, the newer model is better for RL, which is why it was pushed, but worse for sim, which is we hadn't yet updated the asset commit reference in `newton`.

In any case, I think my preferred solution would be to pin the DR Legs example to the previous reference while we investigate this further. So in `example_kamino_robot_dr_legs.py`, we should update the asset path to:

```
asset_path = newton.utils.download_asset("disneyresearch", ref="261cd1f429619d8ef4f546bd788ab9dea906b5e1")
```

I believe that should resolve the issue.

andrewkaufman:
Done — pinned exactly as you suggested, and it resolves the failure cleanly.

```
test_10_dr_legs_dvi_tipped_contact_does_not_creep ... ok
test_11_dr_legs_dvi_contact_force_balances_weight ... ok
```

The full Kamino suite is now green at this pin: **679 tests, OK (4 skipped)**, and the USD suites are unaffected (606 tests, OK).

Two notes on the shape of it:

The pin is narrow by design — only the DR legs example takes the older ref. Everything else stays on the new assets and passes, including `test_02`/`test_03` in `test_kamino_core_model`, the four forward-kinematics tests, and `example_kamino_basic_dr_testmech`. So the pin bump still validates the reconverted assets everywhere except the one place whose thresholds are tuned to the old model's mass and inertia.

I also dropped the `newton.use_coord_layout_targets = True` I had added to that example. It was only needed because the newer model gives it a free base joint, so with the older revision pinned it is redundant — and since `test_kamino_solvers_dvi` imports `Example`, setting a module-level global there would have leaked into whatever else shares the process. Verified the example still passes under `-W error::DeprecationWarning` without it.

I left a comment on the pin pointing at the RL-versus-sim difference and saying to drop the `ref` once you have got to the bottom of it, so it does not quietly become permanent. Happy to open a tracking issue for the mass and inertia deltas if that would help — the base is 27% lighter with roughly a quarter of its previous `Iyy`, which seemed worth writing down somewhere even while the analysis is ongoing.

andrewkaufman:
The unit tests are green on all four platforms and the GPU runner. The one red check is **Run GPU Benchmarks**, and I believe it is runner variance rather than anything in this PR.

It flags the MuJoCo teleop loop as 10-15% slower:

| Change | Before [8ce54fac] | After [1cb639fc] | Ratio | Benchmark |
| + | 246±1ms | 272±1ms | 1.10 | `bench_teleop_mujoco.FastTeleopMuJoCo.time_teleop_loop('mjwarp_cpu_graph')` |
| + | 2.0400 | 2.3375 | 1.15 | `track_mean_loop_ms('mjwarp_cpu_graph')` |
| + | 2.2582 | 2.5622 | 1.13 | `track_p95_loop_ms('mjwarp_cpu_graph')` |

Three things point away from this PR being the cause:

**The compared delta contains nothing that runs.** The baseline `8ce54fac` differs from this branch's base by exactly one commit — "Remove PR close cancellation workflow (#3817)" — which is a GitHub Actions file. On top of that this PR adds the asset ref, a Kamino test change, a changelog fragment, and an example pinned to the *older* asset.

**The benchmark's model is unchanged by the pin.** `bench_teleop_mujoco` loads `unitree_g1/usd_structured/g1_29dof_with_hand_rev_1_0.usda`. Built at both refs it is identical — 44 joints, 49 DOFs, 104 shapes, 44 bodies, and the same `joint_limit_ke`/`joint_limit_kd` (10000 / 10). That is consistent with newton-assets#48 having been a surgical regen of the structured USD, so there is no model difference for the solver to be slower on.

**Only the CPU variants moved.** Every flagged entry is `mjwarp_cpu_graph` or `mjwarp_cpu_eager`; the GPU benchmarks and `bench_viewer` are flat at 1.00, and `mujoco_cpu_eager` came in at 0.99. A change that genuinely added solver work would show up on the GPU side too.

The reason it reproduced across both of my runs with near-identical numbers is that ASV compares fresh measurements against *stored* baseline results rather than re-running the baseline, so a busier runner shifts everything CPU-side by roughly a constant factor.

Happy to be wrong about this — if someone wants it re-run once the queue is quieter, that would settle it. Otherwise all five of the Kamino regressions from the original review are now resolved: 679 Kamino tests and 606 USD tests pass locally at this pin, and the full matrix is green in CI.

@eric-heiden this should be ready for another look when you have a moment, since you raised the original block.

eric-heiden (approved):
Thanks for investigating the potential benchmark regression, @andrewkaufman! It indeed seems unrelated to this PR so let's merge it.

Issue #3826, opened 2026-08-06
Follow-up PR fixes for `controllers` module


jeff-hough:
- [ ] Validate max_dofs == max(dofs_per_robot); undersizing currently silently skips DOFs. Also explicitly support or reject empty/ zero-DOF selections.
- [ ] Clarify public API docs: Document the model-based versus model-free ownership/update lifecycles, and correct the conditional gain/mass-matrix units.
- [ ] Complete example publication: Register the controller example in README.md and add the required 320×320 screenshot.
- [ ] Create explicit follow-ups: APIC replay, Isaac/model-free equivalence, focused controller benchmarks, and validating ControllerBase with a
    materially different controller. These appear only in review threads/the broad #3548 umbrella—not dedicated tracked items.

_Originally posted by @jcarius-nv in https://github.com/newton-physics/newton/issues/3598#issuecomment-5183796367_

PR #3827, opened 2026-08-06
Fix max dof checks, and clarify public API docs

jeff-hough:
## Description

This PR addresses the first two points of the issue #3826 

## Checklist

## New feature / API change

Upon adding max-dof checking, I discovered that this variable was not needed, given it can be derived from the `dofs_per_robot` input. For the same reasoning, `robot_count` was not needed in the constructor.

Other than this small change to the constructor API, this PR updates the class docstrings to clarify units, object life-cycles, and the relationship between the `ControllerJointImpedance` and `ControllerJointImpedanceModelFree` variants.

## Summary by CodeRabbit

- **Improvements**
  - Simplified joint impedance controller setup by deriving robot count and maximum degrees of freedom from the provided configuration.
  - Added comprehensive validation for robot configurations, input arrays, devices, data types, shapes, gains, mass matrices, and indexed data.
  - Improved error detection for invalid or incomplete inputs before execution, including graph-capture scenarios.
- **Documentation**
  - Clarified dynamics requirements, gain shapes and units, and mass-matrix units for model-based and model-free controllers.

### DIFF
--- a/newton/_src/controllers/impl/joint_impedance/model_based.py
+++ b/newton/_src/controllers/impl/joint_impedance/model_based.py
@@ -24,7 +24,7 @@ from newton._src.sim.builder import ModelBuilder
 from newton._src.sim.inverse_dynamics import eval_inverse_dynamics_passive
 
 from ...controller import ControllerBase
-from ...utils import _normalize_indices
+from ...utils import _normalize_indices, _validate_array, _validate_flat_port
 from ._common import _gather_dof_flat_kernel, _idx_max
 from .model_free import ControllerJointImpedanceModelFree
 
@@ -143,9 +143,6 @@ class ControllerJointImpedance(ControllerBase):
     ):
         if not isinstance(builder, ModelBuilder):
             raise TypeError(f"builder must be a newton.ModelBuilder, got {type(builder).__name__}.")
-        if not isinstance(default_dof_indices, wp.array) or default_dof_indices.dtype != wp.uint32:
-            raise TypeError("default_dof_indices must be wp.array[uint32].")
-
         robot_count = builder.articulation_count
         if robot_count < 1:
             raise ValueError("builder has no articulations.")
@@ -163,7 +160,7 @@ class ControllerJointImpedance(ControllerBase):
                 f"zero-DOF fixed joints; found unsupported joint types: {unsupported_joints}"
             )
 
-        self._device = device if device is not None else wp.get_device()
+        self._device = wp.get_device(device)
         self._requires_grad = requires_grad
         self._use_gravity = bool(use_gravity_compensation)
         self._use_coriolis = bool(use_coriolis_compensation)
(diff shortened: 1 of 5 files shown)


newton/_src/controllers/impl/joint_impedance/model_free.py
```diff
@@ -315,6 +347,30 @@ def step(
...
+            raise ValueError(f"stiffness has shape {tuple(stiffness.shape)}, expected {expected_2d}.")
+        if tuple(damping.shape) != expected_2d:
+            raise ValueError(f"damping has shape {tuple(damping.shape)}, expected {expected_2d}.")
+        if self._use_inertia:
+            expected_3d = (self._robot_count, self._max_dofs, self._max_dofs)
+            if tuple(inputs.mass_matrix.shape) != expected_3d:
+                raise ValueError(
+                    f"inputs.mass_matrix has shape {tuple(inputs.mass_matrix.shape)}, expected {expected_3d}."
+                )
+
```
jeff-hough: All of these functions add more runtime validation. Obviously, they do not apply if the controller has been captured in a graph.

newton/_src/controllers/impl/joint_impedance/model_free.py
```diff
@@ -123,9 +134,7 @@ class Outputs:
     def __init__(
         self,
         *,
-        robot_count: int,
         dofs_per_robot: wp.array[wp.int32],
-        max_dofs: int,
```
jeff-hough, line 128: This is the API change - both `robot_count` and `max_dofs` were left-over from prior refactoring. They can both be derived from `dofs_per_robot`, and so are not needed.

More validation of the `dofs_per_robot` input was added instead.

newton/_src/controllers/impl/joint_impedance/model_free.py
```diff
@@ -144,18 +153,23 @@ def __init__(
...
+            raise ValueError(f"default_dof_indices must be 1-D, got shape {tuple(default_dof_indices.shape)}.")
 
         dofs_per_robot_np = dofs_per_robot.numpy()
+        robot_count = int(dofs_per_robot_np.size)
+        if robot_count < 1:
+            raise ValueError("dofs_per_robot must not be empty.")
+        if dofs_per_robot_np.min() < 1:
+            raise ValueError(f"every robot must have >= 1 DOF, got dofs_per_robot={dofs_per_robot_np.tolist()}.")
+
+        max_dofs = int(dofs_per_robot_np.max())
```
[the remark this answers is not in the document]
jcarius-nv: Confirmed at `24129dfe`. `dofs_per_robot`, `default_dof_indices`, and every normalized index override are retained on their original devices, while the kernels launch on `self._device`. Please resolve the controller device first, reject or copy mismatched topology arrays, and add cross-device construction coverage.

newton/_src/controllers/impl/joint_impedance/model_free.py
```diff
@@ -235,6 +260,13 @@ def _normalize_gain(
             f"or None; got {type(value).__name__}."
         )
 
+    def _validate_flat_port(self, array: wp.array[wp.float32], name: str, min_length: int) -> None:
+        """Check a caller-bound flat array's device and length before any kernel reads it."""
+        if array.device != self._device:
+            raise ValueError(f"{name} is on device {array.device}, expected {self._device}.")
+        if array.size < min_length:
+            raise ValueError(f"{name} has length {array.size}, expected at least {min_length}.")
+
```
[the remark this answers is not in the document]
jcarius-nv: Confirmed. Please also make the rebound-port validation complete while fixing this: `_validate_flat_port()` currently checks only device and total size, while live gains and the mass matrix check only shape. Wrong types, dtypes, or dimensions can therefore still reach `wp.launch`, contrary to the new claim that shapes and devices are checked. Validate type, `wp.float32`, dimensionality/shape, and device before launch.

newton/_src/controllers/impl/joint_impedance/model_free.py
```diff
@@ -315,6 +347,30 @@ def step(
...
+        self._validate_flat_port(inputs.joint_qd, "inputs.joint_qd", self._min_len_qd)
+        self._validate_flat_port(inputs.joint_q_des, "inputs.joint_q_des", self._min_len_q_des)
+        self._validate_flat_port(inputs.joint_qd_des, "inputs.joint_qd_des", self._min_len_qd_des)
+        if self._has_qdd:
+            self._validate_flat_port(inputs.joint_qdd, "inputs.joint_qdd", self._min_len_qdd)
+        if self._use_gravity:
+            self._validate_flat_port(inputs.gravity_force, "inputs.gravity_force", self._min_len_gravity)
+        if self._use_coriolis:
+            self._validate_flat_port(inputs.coriolis_force, "inputs.coriolis_force", self._min_len_coriolis)
+        self._validate_flat_port(outputs.joint_f, "outputs.joint_f", self._min_len_f)
```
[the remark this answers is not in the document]
jcarius-nv: Confirmed at `24129dfe`. The model-based path launches both gathers—and may run dynamics evaluation—before delegating to the model-free validation. Please preflight `joint_q` and `joint_qd` before the first gather and add a model-based short/cross-device regression test.

newton/_src/controllers/impl/joint_impedance/model_free.py
jcarius-nv: Since graph capture is the canonical performance path, please state this lifecycle explicitly in the public docs: these Python validations run when `step()` executes during capture, but they are not re-executed on graph replay or APIC rebinding. “Shapes and devices are checked on each call” otherwise overstates the captured behavior.

newton/_src/controllers/impl/joint_impedance/model_based.py
```diff
@@ -30,13 +30,17 @@
...
-    :class:`ControllerJointImpedanceModelFree` — flat 1D sim arrays in,
-    flat 1D torque array out — except that the dynamics terms (mass matrix,
-    gravity force, Coriolis force) are computed internally from the Newton
-    model rather than supplied by the caller.
+    Implements the joint-space impedance control law. This model-based variant
+    computes the mass matrix, gravity, and Coriolis terms itself: it holds a
+    private :class:`~newton.Model` built from ``builder`` and evaluates forward
+    kinematics and the enabled dynamics terms on every :meth:`step`, so the
+    caller supplies only joint positions and velocities.
+
```
jcarius-nv, line 40: This lifecycle claim is not correct. `ModelBuilder.finalize()` mutates builder bookkeeping: it assigns `world_count` and rebuilds start-index arrays. On this exact head, constructing the controller changed a fresh test builder's `world_count` from 0 to 1. The new test compares only `articulation_count`, `joint_dof_count`, and `joint_type`, then confirms that a second finalization succeeds; it does not establish that the builder was untouched. Please either document that construction finalizes and may mutate the supplied builder, or create an actual independent snapshot before finalization if non-mutation is intended to be the contract.

jcarius-nv (commented):
Standards review at `24129dfe`. The runtime-validation findings are recorded in the existing threads. Two remaining review-wide items:

- Please add a Towncrier fragment for the removal of `robot_count` and `max_dofs`, with migration guidance telling callers that both values are now derived from `dofs_per_robot`. The breaking signature change is acceptable here because this module is experimental, but it still needs to be visible to users.
- Please clean up the final commit history: several subjects are non-imperative or vague/overlong (`adds...`, `improves...`, `confirms...`).

newton/_src/controllers/impl/joint_impedance/model_based.py
jeff-hough: Understood.

I don't love either of these options:
* **construction finalizes and may mutate the supplied builder**: this doesn't inform the user whether the builder is still usable afterwards (just that it _may_ mutate)
* **create an actual independent snapshot**: this is not explicitly supported by newton as far as I can tell? I.e. I don't think there are guarantees that builders are `deepcopy`-able?

I prefer to just change the controller to take a `Model` object rather than a `ModelBuilder`, which should avoid the ownership question altogether.

jeff-hough:
This has been superceded by the merged changes here:
#3979

PR #3828, opened 2026-08-06
Unify soft self-contact into CollisionPipeline; solver-owned detection scheduling

### LINKED ISSUE #3846: Unify cloth self-contact into CollisionPipeline and let solvers own detection scheduling
## Problem

Newton currently runs collision detection in two disconnected places:

1. `CollisionPipeline` produces rigid (shape-shape) and soft (particle-shape) contacts into `Contacts`, driven by user code once per step.
2. `SolverVBD` owns a private `TriMeshCollisionDetector` for cloth self-contact, driven inside `step()` with its own scheduling knob (`particle_collision_detection_interval`) and detector-owned result buffers.

Consequences:

- Self-contact results live outside `Contacts`, unlike every other contact family, so downstream consumers need solver-specific access paths.
- Detection frequency has no single control point: skipping rigid detection requires the manual `collide()` + `set_rigid_history_update()` pattern, while self-contact frequency is a solver-internal setting with different semantics.
- There is no way to run rigid detection *more* than once per step (e.g. re-detecting at the current iterate inside the solver loop), which displacement-bounded methods need: contact distances frozen at step start cap how far bodies may move.
- The layering prevents fixing this where it belongs: `newton/_src/sim/` must not import from `newton/_src/solvers/`, so the pipeline cannot reference the detector at its current location.

## Requested feature

- Fold cloth self-contact detection into `CollisionPipeline` / `Contacts`: results stored as `contacts.soft_self_contact_data`, detection available to standalone users via the pipeline, detector relocated to the geometry layer.
- Let a solver own the collision pipeline and control when detection runs, per contact family (rigid pass vs. self-contact), including: never (externally driven), once per step, before/after solver initialization, and every k-th solver iteration — with settings changeable at runtime and no hidden cross-step scheduling state.
- Mid-solve rigid re-detection should preserve in-flight per-contact solver state (e.g. AVBD multipliers) by riding the existing contact-matching machinery rather than cold-starting.
- Keep the existing external-driving workflow (`pipeline.collide(state, contacts)` + `solver.step(..., contacts, ...)`) working unchanged, with deprecations for renamed parameters.

AnkaChan:
## Description

Unify collision detection: every kind of contact is produced through one `CollisionPipeline` and stored in one `Contacts` object, and a solver can own the pipeline and control when detection runs.

- **Soft self-contact joins `Contacts`.** `CollisionPipeline.init_soft_self_contact()` configures cloth self-contact (margin/gap scheme: `margin` = interaction distance, detection queries use `margin + gap`, mirroring `ShapeConfig.margin`/`gap`); `collide(..., soft_self_contact=True)` runs vertex-triangle + edge-edge detection into `Contacts.soft_self_contact_data` (a `TriMeshCollisionInfo`, one packed kernel argument). BVH upkeep stays with the caller via `refit_soft_self_contact_bvh(new_pos, rebuild=False)` — `collide()` never touches the BVHs.
- **Solver-owned pipelines.** `SolverBase` accepts `pipeline=...` (opt-in via `supports_collision_pipeline`), allocates its own contacts buffer (read-only `SolverBase.contacts`), and schedules detection per slot with `collision_frequency` / `collision_frequency_type` (`NONE` / `PRE_INIT` / `PRE_POST_INIT` / `ITERATIONS` / `AUTO` — the in-step detection point). Settings are runtime-changeable via `set_collision_frequency()`, and the solver keeps **no hidden cross-step scheduling state**: skipping detection across steps is expressed by toggling a slot to `NONE` from the calling loop.
- **`SolverVBD` adopts both.** With a pipeline, VBD seeds its self-contact configuration into the pipeline, binds the shared `TriMeshCollisionDetector` to the owned buffer (results become part of `solver.contacts`), and runs the rigid pass per the resolved schedule; without one, everything behaves exactly as today. `particle_self_contact_radius` and `particle_collision_detection_interval` are deprecated (exact legacy behavior preserved when set); rigid `ITERATIONS` raises `NotImplementedError` pending the follow-up work that needs it.
- Mechanically, `TriMeshCollisionInfo`/`TriMeshCollisionDetector` moved from `newton/_src/solvers/vbd/` to `newton/_src/geometry/` (the `sim` layer must not import from `solvers`; the old module re-exports), and result allocation was extracted into `build_tri_mesh_collision_info()` so a contact container can own the arrays while one detector serves any number of buffers.

## Checklist

## Test plan

All on GPU (L40) + CPU:

```
uv run --extra dev -m newton.tests -k test_collision_cloth              # 79 OK (incl. new injected-struct equivalence + pipeline-vs-standalone parity)
uv run --extra dev -m newton.tests -k TestSolverCollisionFrequency      # 50 OK (schedule/toggle/validation + VBD pipeline-vs-legacy parity + deprecations)
uv run --extra dev -m newton.tests -k vbd                               # 224 OK
uv run --extra dev -m newton.tests -k "test_collision_pipeline or test_sensor_contact or test_contact_matching"
```

The `SolverVBD` pipeline path is verified to match the legacy externally-driven path on identical cloth scenes, and all pre-existing self-contact examples/tests were migrated to the new parameters.

## New feature / API change

```python
import newton

pipeline = newton.CollisionPipeline(model, broad_phase="nxn")

solver = newton.solvers.SolverVBD(
    model,
    pipeline=pipeline,                       # solver owns detection
    particle_enable_self_contact=True,
    particle_self_contact_margin=0.2,        # interaction distance
    particle_self_contact_gap=0.15,          # detection reach = margin + gap
)

for i in range(num_steps):
    solver.step(state_in, state_out, control, None, dt)   # detection runs inside step()
    viewer.log_contacts(solver.contacts, state_out)       # incl. soft_self_contact_data
```

## Summary by CodeRabbit

- **New Features**
  - Added triangle-mesh self-collision detection for cloth and soft-body simulations.
  - Added configurable collision scheduling, including automatic and iteration-based detection.
  - Added separate contact margins and detection gaps for more precise collision tuning.
  - Exposed collision information and collision-query utilities through the public geometry API.

- **Documentation**
  - Updated collision and solver-tuning guides with the new configuration names and scheduling options.

- **Deprecations**
  - Retained legacy collision parameters as deprecated aliases with warnings.

Closes #3846


nvtw (changes requested):
Result from AI review:

  Findings, highest severity first:

  1. High: self-contact NONE can consume uninitialized contacts.

     /C:/AnkaChen/newton/newton/_src/geometry/tri_mesh_collision.py:277 is allocated with uninitialized wp.array, not wp.zeros. When scheduling is NONE, VBD skips detection at /C:/AnkaChen/newton/newton/_src/solvers/vbd/solver_vbd.py:2765 but still evaluates self-contact at /C:/AnkaChen/newton/newton/_src/solvers/
     vbd/solver_vbd.py:2837. On the first step this can interpret garbage counts as contacts.

     At minimum, initialize all publicly readable counts to zero and add a first-step NONE regression test.

  2. High: externally populated owned contacts become stale in rigid NONE mode.

     The step initially honors _update_rigid_history, but /C:/AnkaChen/newton/newton/_src/solvers/vbd/solver_vbd.py:1950 forcibly changes it to False. This contradicts the documented workflow where users populate solver.contacts externally while the slot is NONE. VBD then retains old per-body contact indices while
     the buffer contains new rows.

     Remove that override and test external pipeline.collide(..., solver.contacts) across multiple steps with changing contacts.

  3. Medium: pipeline/model ownership is not validated.

     /C:/AnkaChen/newton/newton/_src/solvers/solver.py:238 accepts a pipeline created for a different model. It then allocates contacts from that pipeline at line 249. Detection can consequently use one model while VBD consumes the results using another.

     Require pipeline.model is model during construction.

  4. Medium: arbitrary soft-contact buffers are rebound without validation.

     /C:/AnkaChen/newton/newton/_src/sim/collide.py:1460 checks only that soft_self_contact_data exists; /C:/AnkaChen/newton/newton/_src/geometry/tri_mesh_collision.py:471 performs plain assignment. A buffer from a differently sized model, device, or detector configuration may be passed into kernels without a clear
     error. Validate device, array sizes, and optional triangle-side storage.

  5. Medium: SolverVBD loses its class docstring.

     /C:/AnkaChen/newton/newton/_src/solvers/vbd/solver_vbd.py:95 appears before the string literal, so SolverVBD.__doc__ becomes None. Move the attribute below the class docstring. This also contributes to the PR’s documented docstring-coverage warning. The PR is currently marked draft on GitHub
     (https://github.com/newton-physics/newton/pull/3828).

  Architecture assessment:

  - Geometry placement: yes. The triangle-mesh module contains topology filtering, BVHs, collision queries, and result storage—not VBD force evaluation. Model is only imported under TYPE_CHECKING. It is general enough for Newton’s internal geometry layer, although it remains Newton-model-oriented rather than a
    standalone mesh primitive.

  - Pipeline integration: functional, but not seamless. VBD calls the detector directly through private methods, while standalone pipeline users must explicitly refit BVHs and request soft_self_contact=True. This creates two lifecycle paths. Ideally, VBD should ask the pipeline to run a selective self-contact pass,
    with the pipeline owning refit, buffer validation, and detector binding while avoiding unnecessary rigid detection.

  - Efficiency: I found no obvious regression in the normal covered path. Sharing one detector/BVH set and allocating results in Contacts is sensible. However, the PR provides correctness tests rather than benchmark evidence, so I would not certify performance broadly.

nvtw (dismissed):
The changes look fine from my side. We can still submit adjustments in later MRs if we notice that further improvements are needed.

nvtw:
@eric-heiden should we mark the cloth/soft body part of the collision pipeline experimental to allow for easier API changes?

AnkaChan:
@coderabbitai The remaining review notes are addressed in 6220a1f6: `soft_contact_gap` is validated in the `CollisionPipeline` constructor (covering the deprecated `soft_contact_margin` alias path), the collisions.rst comment now says gap, and the changed-fragment carries concrete legacy-to-new migration snippets. The mid-solve `C0` recompute concern was resolved in acbc7297 (`step_body_body_contact_C0_lambda` now runs inside `_refresh_rigid_contact_state`).

newton/_src/solvers/solver.py
```diff
@@ -196,13 +197,187 @@ class SolverBase:
...
 
-    def __init__(self, model: Model):
+    def __init__(
+        self,
+        model: Model,
+        *,
+        pipeline: CollisionPipeline | None = None,
+        collision_frequency: list[int] | None = None,
+        collision_frequency_type: list[CollisionFrequencyType] | None = None,
+    ):
```
eric-heiden, line 246: I realize we didn't have a docstring for this constructor before, can you add it here to also document the arguments?

newton/_src/solvers/solver.py
```diff
@@ -196,13 +197,187 @@ class SolverBase:
...
+    _COLLISION_SLOT_SOFT_SELF = 1
+
     _module_options_revision = 0
 
-    def __init__(self, model: Model):
+    def __init__(
+        self,
+        model: Model,
+        *,
+        pipeline: CollisionPipeline | None = None,
```
eric-heiden: Maybe `collision_pipeline` is a clearer name? In that case I would rename this everywhere, like `Solver.collision_pipeline` also in the subclasses.

newton/_src/geometry/tri_mesh_collision.py
```diff
@@ -0,0 +1,985 @@
...
+    triangle_colliding_vertices_count: wp.array[wp.int32]
+    triangle_colliding_vertices_min_dist: wp.array[float]
+
+    # size: 2 x sum(edge_colliding_edges_buffer_sizes)
+    # every two elements records the edge index and an edge index it collides to
+    edge_colliding_edges: wp.array[wp.int32]
+    edge_colliding_edges_offsets: wp.array[wp.int32]
+    edge_colliding_edges_buffer_sizes: wp.array[wp.int32]
+    edge_colliding_edges_count: wp.array[wp.int32]
+    edge_colliding_edges_min_dist: wp.array[float]
```
eric-heiden, line 74: Now that this is part of the `newton.geometry` public API we should have docstrings for the members of this struct, perhaps also a better docstring than `Store bounded triangle-mesh self-collision query results.` for this struct in general.

eric-heiden:
Nice work pulling soft self-contact into `CollisionPipeline` and giving solvers a single explicit place to schedule detection.

Most of the comments below are about the new public surface (naming, defaults, and arguments that are accepted but not used), plus one correctness issue in the mid-solve rigid re-detection path that lets resting bodies sink through the ground in the default `ITERATIONS` configuration.

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -2401,6 +2568,224 @@ def _initialize_particles(self, state_in: State, state_out: State, dt: float):
...
+                            self.rigid_contact_k_start_value,
+                        ],
+                        outputs=[
+                            self.body_body_contact_penalty_k,
+                            self.body_body_contact_material_kd,
+                            self.body_body_contact_material_mu,
+                            self.body_body_contact_material_ke,
+                        ],
+                        dim=contact_launch_dim,
+                        device=self.device,
```
eric-heiden, line 2777: 🔴 The mid-solve rigid re-detection zeroes the accumulated contact duals, so `CollisionFrequencyType.ITERATIONS` sinks resting bodies in its default configuration.

`step()` calls `_refresh_rigid_contact_state(contacts, refresh=True)` from the in-loop re-detection block. With `SolverVBD`'s default `rigid_contact_history=False` that lands in the `elif contact_launch_dim > 0:` branch, which ends in `self.body_body_contact_lambda.zero_()` on this line, discarding the augmented Lagrangian contact impulse accumulated so far on every k-th iteration:

```
history=False (default)  redetect[0] max|lambda| 36.4007 -> 0.0000
history=False (default)  redetect[3] max|lambda| 36.4007 -> 0.0000
history=True, matching   redetect[0] max|lambda| 77.9438 -> 77.9438
```

For a single box resting on a ground plane the correct resting height is z = 0.200. `ITERATIONS` moves away from that as the solve is refined, and keeps sinking over time:

<details>

<summary>Measured resting height</summary>

```
convergence under refinement (density=500):
    iters/hz |   PRE_INIT   ITER k=1   ITER k=2   ITER k=5
      10/120 |    0.14857   -0.03875    0.01857    0.10242
      30/240 |    0.18065   -0.05431   -0.00137    0.08604
      60/480 |    0.18994   -0.06357   -0.01341    0.07730

long run:
  t= 0.5s  PRE_INIT z= 0.18065   ITER k=1 z= -0.05431   ITER k=1 + history z= 0.19900
  t= 4.0s  PRE_INIT z= 0.18060   ITER k=1 z= -0.42314   ITER k=1 + history z= 0.19901
```

</details>

This is the default configuration for a documented public enum member, and `test_vbd_rigid_iterations_mode` cannot catch it: the `k=10` case never enters the block because the run does fewer iterations than that, and the `k=1` case only asserts `np.isfinite`.

Suggested fix: do not zero `body_body_contact_lambda` on a mid-solve refresh. Either carry the duals across re-detection (the `rigid_contact_history=True` plus matching path already shows the correct behavior, z stays at 0.199), or make `ITERATIONS` require contact matching at construction the way `rigid_contact_history=True` already does, so the broken default combination is rejected instead of silently sinking bodies.

newton/_src/solvers/solver.py
```diff
@@ -196,13 +197,187 @@ class SolverBase:
...
+                raise ValueError(
+                    "step(contacts=...) must be None when the solver owns a collision "
+                    "pipeline; the solver detects into its own buffer (solver.contacts)."
+                )
+            return self._pipeline_contacts
+        return contacts
+
+    def _run_rigid_collision(self, state: State) -> None:
+        """Run the owned pipeline into the owned contacts buffer."""
+        self.pipeline.collide(state, self._pipeline_contacts)
```
eric-heiden: 🔵 `_run_rigid_collision` drops `dt`, which makes solver-owned pipelines and speculative contacts mutually exclusive.

This is the only path by which an owning solver drives detection (called from `solver_vbd.py:2216` and `:2237`), and it calls `collide()` without `dt`. `CollisionPipeline.collide` raises `ValueError("dt must be provided when speculative contacts are enabled")` when `self._speculative_enabled and dt is None`, so `SolverVBD(model, pipeline=CollisionPipeline(model, speculative_config=...))` fails on the very first `step()` even though the caller did pass `dt` to `step()`, and the message names an argument the user has no way to supply.

Both call sites are inside `SolverVBD.step(..., dt)`, so `dt` is already in scope at each one:

```suggestion
    def _run_rigid_collision(self, state: State, dt: float | None = None) -> None:
        """Run the owned pipeline into the owned contacts buffer."""
        self.pipeline.collide(state, self._pipeline_contacts, dt=dt)
```

and then pass `dt` from both `SolverVBD.step` call sites. If supporting that combination is out of scope here, raising a clear `ValueError` at construction when `pipeline._speculative_enabled` would at least make the limitation visible. Speculative contacts currently appear only in `newton/tests/test_speculative_contacts.py`, which is why CI does not surface this.

newton/_src/sim/collide.py
```diff
@@ -2196,12 +2445,32 @@ def collide(
...
+            detector.vertex_triangle_collision_detection(
+                query_radius,
+                min_query_radius=self.soft_self_contact_rest_shape_exclusion_radius,
+                min_distance_filtering_ref_pos=self.model.particle_q,
+            )
+            detector.edge_edge_collision_detection(
+                query_radius,
+                min_query_radius=self.soft_self_contact_rest_shape_exclusion_radius,
+                min_distance_filtering_ref_pos=self.model.particle_q,
+            )
```
eric-heiden: 🔵 `collide()` takes `state`, but the self-contact half of the call never reads it, so detection runs against stale BVHs.

With `soft_self_contact=True`, detection uses whatever the self-contact BVHs were last fitted to by `refit_soft_self_contact_bvh()`. At construction those are built from `model.particle_q` (the rest pose), and `collide()` never updates them, as the comment on these lines says outright. The rigid and particle-shape parts of the same call do use `state`, so one call ends up with two different notions of "current positions".

Measured with two 8x8 cloth layers whose rest pose is 2 m apart but whose state places them 1 cm apart, with a 2 cm margin:

```
collide(state, contacts, soft_self_contact=True) without refit -> vertex-tri pairs: 0
collide(state, contacts, soft_self_contact=True) with refit    -> vertex-tri pairs: 768

detector.vertex_positions is model.particle_q  ->  True
detector.vertex_positions is state.particle_q  ->  False
```

So the natural first-use loop, calling `pipeline.collide(state, contacts, soft_self_contact=True)` each step, misses every real self-contact with no error and no warning. The docstring does note that the self-contact BVHs are not updated by this call, but a parameter that is accepted and then ignored for one of the outputs is easy to miss. The owning-solver path is unaffected because `SolverVBD._collision_detection_penetration_free` refits internally, so this only bites standalone pipeline users, which is exactly the audience `init_soft_self_contact` and `refit_soft_self_contact_bvh` were added for.

Suggested direction: refit inside `collide()` from `state.particle_q` when `soft_self_contact=True` (the positions are already in hand), and keep `refit_soft_self_contact_bvh(rebuild=True)` as the explicit escape hatch for periodic full rebuilds. If caller-driven refitting is deliberate for graph capture reasons, at minimum remember the array the BVHs were last fitted to and warn or raise when `state.particle_q` is not that array, so the failure is loud instead of a silent zero.

newton/_src/solvers/solver.py
```diff
@@ -196,13 +197,187 @@ class SolverBase:
...
+
+    supports_collision_pipeline: bool = False
+    """Whether this solver can own a :class:`~newton.CollisionPipeline` and drive detection itself.
+
+    Currently only :class:`~newton.solvers.SolverVBD` opts in; passing ``pipeline``
+    to any other solver raises ``ValueError`` (drive detection externally instead).
+    """
+
+    _COLLISION_SLOT_RIGID = 0
+    _COLLISION_SLOT_SOFT_SELF = 1
```
eric-heiden: 🔵 The new schedule API is two fixed-length-2 positional lists whose slot meaning lives only in private constants.

`collision_frequency` and `collision_frequency_type` are ordered `[rigid, soft_self_contact]`, but the only names for those positions are `_COLLISION_SLOT_RIGID` and `_COLLISION_SLOT_SOFT_SELF`, which are private and not exported from `newton.solvers`. Call sites therefore read as positional magic: `newton/examples/cloth/example_cloth_franka.py:256`, `example_cloth_rollers.py:372`, and `example_softbody_franka.py:158` all pass literals like `[AUTO, PRE_INIT]`, where the reader has to know that slot 1 is self-contact and that slot 0 was set to `AUTO` only to leave it alone.

| issue | effect |
|---|---|
| both entries are the same enum type | `[PRE_INIT, AUTO]` and `[AUTO, PRE_INIT]` both construct successfully with opposite meanings |
| `set_collision_frequency` raises on `len(...) != 2` (lines 313 and 322) | adding a third detection slot later is a hard break for every existing caller |

The surrounding API already models this better: `SolverVBD.JointSlot` is a public named-slot class, and `CollisionPipeline.set_collision_detection_range()` uses one keyword per quantity with `None` meaning "leave alone".

Suggested direction: accept keywords or a mapping, for example `set_collision_frequency(rigid=..., soft_self_contact=...)`, or `collision_frequency_type={CollisionSlot.RIGID: PRE_INIT}` backed by a public `SolverBase.CollisionSlot` IntEnum mirroring the existing `JointSlot` pattern. That makes each call site self-describing, lets a caller set one slot without restating the other, and leaves room for a third slot without breaking existing code. If the positional lists have to stay, please at least promote `_COLLISION_SLOT_*` to public constants and export them from `newton.solvers`.

newton/_src/sim/contacts.py
```diff
@@ -176,6 +177,13 @@ def __init__(
         requested_attributes: set[str] | None = None,
         contact_matching: bool = False,
         contact_report: bool = False,
+        soft_self_contact: bool = False,
+        particle_count: int = 0,
+        tri_count: int = 0,
+        edge_count: int = 0,
+        soft_self_contact_vertex_buffer_pre_alloc: int = 16,
+        soft_self_contact_edge_buffer_pre_alloc: int = 32,
+        soft_self_contact_record_triangle_vertices: bool = False,
```
eric-heiden, line 184: 🔵 These six new constructor parameters are public API (`Contacts` is exported as `newton.Contacts`), and they are inconsistent both internally and with the rest of the self-contact surface.

Three are namespaced (`soft_self_contact`, `soft_self_contact_vertex_buffer_pre_alloc`, `soft_self_contact_edge_buffer_pre_alloc`) and three are bare (`particle_count`, `tri_count`, `edge_count`), even though the docstring says all three are used only when `soft_self_contact=True`. Under the prefix-first naming convention the bare names lose autocomplete grouping and read like general purpose sizing arguments that a caller might reasonably set with self-contact off, where they are silently ignored.

The same two buffer capacities are also spelled three different ways across three public entry points, with two different default pairs:

| entry point | vertex / edge parameter | defaults |
|---|---|---|
| `newton.Contacts.__init__` | `soft_self_contact_vertex_buffer_pre_alloc` / `soft_self_contact_edge_buffer_pre_alloc` | 16 / 32 |
| `build_tri_mesh_collision_info()` (internal) | `vertex_collision_buffer_pre_alloc` / `edge_collision_buffer_pre_alloc` | 16 / 32 |
| `CollisionPipeline.init_soft_self_contact()` | `vertex_buffer_pre_alloc` / `edge_buffer_pre_alloc` | 32 / 64 |
| `SolverVBD.__init__` | `particle_vertex_contact_buffer_size` / `particle_edge_contact_buffer_size` | 32 / 64 |

The two default pairs are not merely cosmetic, they are mutually incompatible, because `_get_soft_self_contact_detector` validates the `Contacts` array sizes against the detector's pre-alloc. Building both objects at their own defaults is rejected:

<details>

<summary>Rejected at default settings</summary>

```python
p.init_soft_self_contact(...)                       # defaults 32 / 64
c = newton.Contacts(0, 0, soft_self_contact=True,   # defaults 16 / 32
                    particle_count=..., tri_count=..., edge_count=...)
p._get_soft_self_contact_detector(c)
# ValueError: collision_info.vertex_colliding_triangles has size 800, expected 1600
```

</details>

`CollisionPipeline.contacts()` avoids this by forwarding the detector's values explicitly, so only the hand-built `Contacts` path trips, but that path is public and the error names an internal CSR array rather than the two defaults that actually disagree.

Suggested direction: pick one name pair for the two capacities and use it at every entry point (for example `self_contact_vertex_buffer_size` / `self_contact_edge_buffer_size`), prefix the three bare mesh counts (`soft_self_contact_particle_count` and friends), and align the defaults on the 32 / 64 that the solver and pipeline already agree on. If `Contacts` has to keep 16 / 32, make the validation error name both capacities so the mismatch explains itself.

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -548,7 +578,114 @@ def __init__(
...
+                    "particle_self_contact_margin without particle_self_contact_gap retains its deprecated "
+                    "meaning as the detection query radius; pass particle_self_contact_gap explicitly "
+                    "to use particle_self_contact_margin as the interaction distance.",
+                    DeprecationWarning,
+                    stacklevel=3,
+                )
+                _sc_margin = 0.2
+            else:
+                _sc_margin = particle_self_contact_margin if particle_self_contact_margin is not None else 0.2
+                _sc_gap = particle_self_contact_gap if particle_self_contact_gap is not None else 0.0
```
eric-heiden, line 635: 🔵 Passing only `particle_self_contact_margin` now raises even when self-contact is disabled.

When `particle_self_contact_radius` is omitted, this block applies the legacy reading (`_sc_gap = particle_self_contact_margin - 0.2`, `_sc_margin = 0.2`) and raises `ValueError` for any margin below 0.2. The check now sits in the unconditional `__init__` prologue, whereas on `main` the equivalent check lived inside `if particle_enable_self_contact:` in `_init_particle_system`.

So `SolverVBD(model, particle_enable_self_contact=False, particle_self_contact_margin=0.005)` was accepted and ignored before and is now a hard failure, with no deprecation cycle. Configuring self-contact geometry while the feature is off is a pattern this repo uses itself in `example_softbody_franka.py`, `example_cloth_stiff_material_hanging.py`, and `example_vbd_soft_rigid_contact.py`. Those three also pass `radius`, which routes into the other branch, so nothing in-repo breaks, but an external caller who set only a metre-scale `margin` with self-contact disabled now gets an exception.

The other two deprecation branches (`radius` set, and `margin` plus `gap`) preserve the previous semantics exactly, so this is the one behavior change in the block. Separately, and much lower impact: `self.particle_self_contact_radius` is no longer assigned (`_init_particle_system` now sets only margin and gap). There are no in-repo readers of that attribute, so it affects external users only.

Suggested fix: gate the margin-only `ValueError` on `particle_enable_self_contact` to match the previous behavior, and otherwise clamp `_sc_gap` to 0 and skip the check.

newton/_src/sim/collide.py
```diff
@@ -1729,6 +1808,161 @@ def contacts(self) -> Contacts:
...
+        gap: float = 0.0,
+        rest_shape_exclusion_radius: float = 0.0,
+        vertex_buffer_pre_alloc: int = 32,
+        edge_buffer_pre_alloc: int = 64,
+        edge_edge_parallel_epsilon: float = 1e-5,
+        record_triangle_contacting_vertices: bool = False,
+        topological_filter_threshold: int = 2,
+        external_vertex_filter_map: dict | None = None,
+        external_edge_filter_map: dict | None = None,
+    ) -> None:
```
eric-heiden, line 1911: ⚪ `CollisionPipeline` now offers two spellings for the same two values:

```
init_soft_self_contact(margin=, gap=, ...)
set_collision_detection_range(soft_contact_gap=, soft_self_contact_margin=, soft_self_contact_gap=)
```

Both write `pipeline.soft_self_contact_margin` and `pipeline.soft_self_contact_gap`. The short form is defensible in isolation, since the method name already says `soft_self_contact`, and the multi-domain setter has to qualify in order to keep self-contact distinct from the particle-shape `soft_contact_gap`. The cost is that neither name round-trips: a user sets `init_soft_self_contact(margin=0.2)`, reads it back as `pipeline.soft_self_contact_margin`, and updates it with `set_collision_detection_range(soft_self_contact_margin=...)`, so one quantity has three spellings across set, get, and update on the same object.

Only a nit, since the docstrings do connect them and the failure mode is a `TypeError` on an unexpected keyword rather than wrong physics. Using the qualified names in `init_soft_self_contact` as well would make all three agree, or an explicit cross-reference in both docstrings would cover it.

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -686,10 +823,11 @@ def _init_particle_system(
             raise ValueError("model.soft_mesh_adjacency_device is missing; finalize the model with ModelBuilder.")
```
eric-heiden, line 838: ⚪ A few leftovers in `newton/_src/solvers/vbd/solver_vbd.py` that look ready to clean up:

1. Line 810: `self.particle_collision_detection_interval` is assigned but read nowhere, since scheduling now flows through `_deprecated_particle_interval` (line 648) into `_resolve_self_contact_schedule` (line 3780). It also stores `None` by default instead of `0`, so anyone still reading it for back-compat gets a non-numeric value. The now unused `particle_collision_detection_interval` parameter of `_init_particle_system` (line 797) and its call-site argument (line 751) can go with it.
2. Line 2192: the `contacts:` doc for `step()` still says "If None, rigid contact handling is skipped", but with an owned pipeline `contacts=None` is required and `_resolve_step_contacts` substitutes `solver.contacts`.
3. Line 2583: the "Contact C0 + history restore BEFORE integration" comment now sits inside `_refresh_rigid_contact_state`, which no longer touches C0. That launch stayed in `_initialize_rigid_bodies` (line 2848).
4. Line 2810: stray blank line before the closing `"""` of the `_initialize_rigid_bodies` docstring.

newton/_src/solvers/solver.py
eric-heiden: To be concrete about the shape, since this is worth locking in before the API ships.

The annotations are a length lie. `list[int]` and `list[CollisionFrequencyType]` promise "any length, homogeneous", but the constructor (which forwards straight to `set_collision_frequency`) and the setter both reject anything other than exactly 2 at runtime:

| line | check |
|---|---|
| `solver.py:318` | `if len(collision_frequency) != 2: raise ValueError("collision_frequency must have length 2, ...")` |
| `solver.py:327` | `if len(collision_frequency_type) != 2: raise ValueError("collision_frequency_type must have length 2, ...")` |

`tuple[int, int]` / `tuple[CollisionFrequencyType, CollisionFrequencyType]` would at least make the length honest, but it keeps the slot meaning positional, so `[AUTO, PRE_INIT]` and `[PRE_INIT, AUTO]` still both type-check while meaning opposite things.

The preferred fix is to drop the sequence and give each slot its own keyword, prefixed by the slot name:

```python
def __init__(
    self,
    model: Model,
    *,
    pipeline: CollisionPipeline | None = None,
    rigid_collision_frequency: int | None = None,
    rigid_collision_frequency_type: CollisionFrequencyType | None = None,
    soft_self_contact_collision_frequency: int | None = None,
    soft_self_contact_collision_frequency_type: CollisionFrequencyType | None = None,
): ...
```

with `set_collision_frequency` taking the same four keywords and `None` meaning "leave this slot alone", which is already how `CollisionPipeline.set_collision_detection_range()` behaves. The read-side properties would split the same way.

That buys four things:

- Call sites read for themselves. `example_cloth_franka.py:256`'s `[AUTO, PRE_INIT]` becomes `soft_self_contact_collision_frequency_type=PRE_INIT`, and the rigid slot is simply not mentioned instead of being pinned to `AUTO` to mean "don't touch this one".
- The two slot-specific rejections become ordinary per-argument checks instead of index lookups into a list: `PRE_POST_INIT` is invalid for rigid (line 330) and an active rigid type needs an owned pipeline (line 335).
- `_COLLISION_SLOT_RIGID` and `_COLLISION_SLOT_SOFT_SELF` can stay private, because no caller needs the index any more.
- Adding a third detection slot later is a new keyword instead of a hard break for every existing caller.

The cost is four keywords instead of two, which seems like the right trade for an API where the current spelling lets a caller silently swap the two slots.

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -1814,13 +1952,42 @@ def step(
...
+            ):
+                # Re-detect all pipeline contacts at the current iterate. This
+                # must also run without internally integrated bodies because
+                # the pipeline owns particle-shape contacts. In-flight rigid
+                # lambdas are snapshotted first so matching can carry them onto
+                # the refreshed contact set.
+                self._snapshot_rigid_contact_history(contacts)
+                self._run_rigid_collision(self._rigid_iterate_view(state_in, state_out))
+                self._refresh_rigid_contact_state(contacts, refresh=True)
+                self._refresh_body_particle_contact_state(contacts, refresh=True)
```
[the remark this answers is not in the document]
eric-heiden: Confirming this, and it reaches a bit wider than `C0` alone.

`step_body_body_contact_C0_lambda` is the only kernel that writes `body_body_contact_C0`, `body_body_contact_normal_rho` and `body_body_contact_tangent_rho`, and it is launched from exactly one place: `_initialize_rigid_bodies` line 2848, immediately after the once-per-step `_refresh_rigid_contact_state(contacts, refresh)` on line 2839. The mid-solve `ITERATIONS` path on line 2238 calls `_refresh_rigid_contact_state(contacts, refresh=True)` on its own with no matching launch, so after re-detection all three arrays still describe the previous contact set while `rigid_contact_point0`, `rigid_contact_point1` and `rigid_contact_normal` have been replaced underneath them.

Each refreshed row is then solved with a rest offset and an ALM metric that were computed for a different contact. `normal_rho` and `tangent_rho` are not warm-start hints that wash out over iterations, they set the effective stiffness `k_eff = k*rho/(k+rho)`, so this changes the material the contact is solved with.

Moving that launch into `_refresh_rigid_contact_state` behind the same `refresh` flag would keep the prologue and the mid-solve path in sync. It needs `dt` and the body frame that detection actually ran against, which is `state_in.body_q` in the prologue and the `_rigid_iterate_view(state_in, state_out)` frame on line 2237 here, so the frame has to be passed in rather than assumed.

newton/_src/geometry/tri_mesh_collision.py
eric-heiden: One more item for the same public-surface pass: the struct is exported but the functions needed to read it are not.

The only supported way to access this data from a kernel is the `@wp.func` accessor set right below the struct: `get_vertex_colliding_triangles_count`, `get_vertex_colliding_triangles`, `get_vertex_collision_buffer_vertex_index`, `get_triangle_colliding_vertices_count`, `get_triangle_colliding_vertices`, `get_edge_colliding_edges_count`, `get_edge_colliding_edges` and `get_edge_collision_buffer_edge_index`, plus `build_tri_mesh_collision_info` to allocate one. None of those appear in `newton/_src/geometry/__init__.py`'s `__all__`, so `TriMeshCollisionInfo` reaches `newton.geometry` without them.

They are not optional convenience wrappers. Reading the raw fields directly is unsafe in two specific ways that the accessors handle:

- `vertex_colliding_triangles_count` can exceed `vertex_colliding_triangles_buffer_sizes` when a buffer overflows, which is exactly what `get_vertex_colliding_triangles_count` clamps with `wp.min`.
- `vertex_colliding_triangles` and `edge_colliding_edges` store interleaved pairs indexed as `2 * (offset + i) + 1`, while `triangle_colliding_vertices` is a plain `offset + i`. Nothing in the field names says so.

Exporting the accessors alongside the documented field layout would let external kernels use the struct without reimplementing those rules.

newton/_src/solvers/solver.py
```diff
@@ -196,13 +197,187 @@ class SolverBase:
...
+            SolverBase.CollisionFrequencyType.AUTO,
+            SolverBase.CollisionFrequencyType.AUTO,
+        ]
+        self.set_collision_frequency(
+            collision_frequency=collision_frequency,
+            collision_frequency_type=collision_frequency_type,
+        )
+
+    @property
+    def contacts(self) -> Contacts | None:
```
eric-heiden, line 289: 🔵 `SolverBase.contacts` is a property, but the two sibling contact accessors in the same public surface are methods.

| API | form | call |
|---|---|---|
| `Model.contacts()` (`model.py:1972`) | method | `model.contacts()` |
| `CollisionPipeline.contacts()` (`collide.py:1753`) | method | `pipeline.contacts()` |
| `SolverBase.contacts` (here) | property | `solver.contacts` |

These get used side by side, including in this docstring's own example (`pipeline.collide(state, solver.contacts)` right after `pipeline.contacts()`), so `solver.contacts()` is a natural thing to type. It fails with `TypeError: 'Contacts' object is not callable`, which does not point at the mistake.

The property itself is defensible since it does not allocate, unlike the other two. But the getter and setter here only proxy `self._pipeline_contacts` with no validation, so a plain attribute would behave identically and be one less shape in the API. If the property stays, could the "does not allocate" contrast move to the first line of the docstring instead of the third? That is the whole reason this one is spelled differently.

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -699,24 +837,24 @@ def _init_particle_system(
...
-                edge_collision_buffer_pre_alloc=particle_edge_contact_buffer_size,
-                edge_edge_parallel_epsilon=particle_edge_parallel_epsilon,
-                topological_contact_filter_threshold=particle_topological_contact_filter_threshold,
-                external_vertex_triangle_filtering_map=particle_external_vertex_contact_filtering_map,
-                external_edge_edge_filtering_map=particle_external_edge_contact_filtering_map,
-            )
+            if self.pipeline is not None:
+                # Solver-owned pipeline: use its shared detector bound to the owned
+                # Contacts buffer so self-contact results land in solver.contacts.
+                self.trimesh_collision_detector = self.pipeline._get_soft_self_contact_detector(self._pipeline_contacts)
```
eric-heiden: 🔵 The solver captures a device-side snapshot of a detector that the pipeline can later re-point at a different buffer.

`CollisionPipeline._get_soft_self_contact_detector()` (`collide.py:1931`) keeps one shared detector per pipeline and rebinds it to whichever `Contacts` it is handed:

```python
detector = self._ensure_soft_self_contact_detector()
if detector.collision_info is not data:
    detector._bind_external_buffers(data)
```

`CollisionPipeline.collide(..., soft_self_contact=True)` goes through that same call (`collide.py:2461`). So any later `pipeline.collide(state, other_contacts, soft_self_contact=True)` moves the shared detector onto `other_contacts`, while `self.trimesh_collision_info` on line 859 still holds the struct captured here at construction. From then on the detector writes into `other_contacts` and the solver keeps evaluating the frozen contents of its own buffer. No error, and the output still looks plausible.

This is easy to hit by accident, because `pipeline.contacts()` allocates a fresh buffer on every call, so a user who writes `contacts = pipeline.contacts()` after constructing the solver already holds a second buffer.

Two ways out: rebuild `self.trimesh_collision_info` from `self.trimesh_collision_detector.collision_info` at the start of each self-contact detection instead of caching it once, or have `_get_soft_self_contact_detector` refuse to rebind a detector that a solver has already claimed and point the caller at `solver.contacts` instead.

newton/tests/test_solver_collision_frequency.py
```diff
@@ -0,0 +1,456 @@
...
+    def run(model, solver, contacts):
+        s0, s1 = model.state(), model.state()
+        for _ in range(3):
+            solver.step(s0, s1, None, contacts, 1e-3)
+            s0, s1 = s1, s0
+        return s0.particle_q.numpy()
+
+    q_a = run(model_a, solver_a, None)
+    q_b = run(model_b, solver_b, None)
+    assert_np_equal(q_b, q_a, tol=1e-6)
```
eric-heiden, line 540: 🔵 This parity check passes for the wrong reason: the scene never produces a self-contact, so the path it is meant to cover contributes nothing.

`_build_cloth_model` (line 300) creates a flat 8x8 grid at z = 1.0 with no ground plane and no collider, and `run` steps it three times at `dt = 1e-3`. A planar grid in 3 ms of free fall stays planar, so both solvers detect zero self-contacts and `assert_np_equal(q_b, q_a)` compares two runs that never entered self-contact evaluation. `test.assertIsNotNone(solver_b.contacts.soft_self_contact_data)` on line 340 only confirms the buffer was allocated, not that anything was written into it.

Asserting that the owned run actually found self-contacts would make the parity claim real:

<details>

<summary>Suggested check</summary>

```python
data = solver_b.contacts.soft_self_contact_data
test.assertGreater(
    int(data.vertex_colliding_triangles_count.numpy().sum())
    + int(data.edge_colliding_edges_count.numpy().sum()),
    0,
)
```

</details>

That assertion fails against the current scene, which is the point. The cloth needs to actually fold before the two paths are meaningfully compared, for example by dropping it onto a ground plane or starting it creased, and the step count needs to be long enough to reach contact.

newton/_src/solvers/solver.py
eric-heiden: The constructor has a docstring now, covering `model`, `collision_pipeline`, `collision_frequency` and `collision_frequency_type`. Resolving.

newton/_src/solvers/solver.py
eric-heiden: Renamed. The parameter and the attribute are both `collision_pipeline` now, and `SolverVBD` plus the examples use the same name throughout. Resolving.

newton/_src/geometry/tri_mesh_collision.py
eric-heiden: Both parts are covered now. Every field of `TriMeshCollisionInfo` has a docstring, the struct docstring calls out the interleaved pair layout and that counts can exceed row capacity, and all eight `get_*` accessors plus `build_tri_mesh_collision_info` are exported from `newton.geometry` and listed in `docs/api/newton_geometry.rst`. Resolving.

newton/_src/solvers/vbd/solver_vbd.py
eric-heiden: Fixed. The mid-solve block now snapshots the in-flight lambdas first and calls `_refresh_rigid_contact_state(..., restore_history=True)`, which takes the `init_body_body_contacts_alm` branch instead of the one that ended in `body_body_contact_lambda.zero_()`, and the following `_step_body_body_contact_frame` rebuilds `C0` and the rho terms on the refreshed contact set. The new `RuntimeError` when the pipeline has no contact matching also makes the requirement visible instead of silently cold-starting. Resolving.

newton/_src/solvers/solver.py
eric-heiden: `_run_rigid_collision` takes `dt` now and forwards it to `collide()`, and both `SolverVBD.step` call sites pass it, so an owned pipeline with speculative contacts works. Resolving.

newton/_src/sim/collide.py
eric-heiden: `collide()` now calls `detector.refit(state.particle_q)` before running self-contact detection, so both halves of the call use the same positions, and the docstring states it. Resolving.

newton/_src/solvers/solver.py
eric-heiden: The schedule is a `Mapping[CollisionSlot, ...]` now, with `CollisionSlot` public on `SolverBase`, unspecified slots keeping their defaults, and both the length-2 checks and the private `_COLLISION_SLOT_*` constants gone. Adding a slot later is no longer a break for existing callers, and the example call sites read as named slots instead of positional literals. Resolving.

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -706,24 +854,26 @@ def _init_particle_system(
         self.use_particle_tile_solve = particle_enable_tile_solve and model.device.is_cuda
 
         if particle_enable_self_contact:
```
eric-heiden, line 854: 🔵 The margin-only validation moved from `_init_particle_system` into `__init__` and left both of its guards behind.

On `main` this check sat right here, inside `if particle_enable_self_contact:`, and below the `if model.particle_count == 0: return` early exit at the top of the method. Both guards still exist in `_init_particle_system`, but the check they protected now runs in the unconditional `__init__` prologue, in the margin-only branch that computes `_sc_gap = particle_self_contact_margin - 0.2` and raises when that goes negative.

So a sub-0.2 `particle_self_contact_margin` is now a hard failure in two configurations that were previously accepted and ignored:

| Call | `main` | This PR |
|---|---|---|
| `SolverVBD(model, particle_self_contact_margin=0.1)` on a model with no particles | accepted | `ValueError` |
| `SolverVBD(model, particle_enable_self_contact=False, particle_self_contact_margin=0.1)` | accepted | `ValueError` |

`particle_enable_self_contact` defaults to `False`, so the first row is what an external caller hits on any rigid-only model. Nothing in this repo breaks, because the examples that configure self-contact geometry also pass `particle_self_contact_radius` and route into the other branch. That also means there is no test covering it, and outside callers get the exception with no deprecation cycle.

The validated value is only ever consumed when self-contact is on and particles exist, so the check can simply be skipped when it cannot matter. Keep the legacy reinterpretation and its `DeprecationWarning`, but only raise when `particle_enable_self_contact` is true and `model.particle_count > 0`, and clamp `_sc_gap` to `0.0` otherwise.

<details>

<summary>Both calls raise on the current branch</summary>

```py
import newton

b = newton.ModelBuilder()
b.add_body()
b.add_shape_sphere(body=0, radius=0.1)
m = b.finalize()  # m.particle_count == 0

# Both raise ValueError: "particle_self_contact_margin is smaller than the
# legacy interaction radius 0.2; this would miss contacts."
newton.solvers.SolverVBD(m, particle_self_contact_margin=0.1)
newton.solvers.SolverVBD(m, particle_enable_self_contact=False, particle_self_contact_margin=0.1)
```

</details>

This is the same underlying regression as https://github.com/newton-physics/newton/pull/3828#discussion_r3859736421. That thread covers the self-contact-disabled half; this adds the zero-particle case and the reason both slipped through.

nvtw: Should be addressed with the latest commits

newton/_src/sim/collide.py
```diff
@@ -1774,6 +1853,158 @@ def contacts(self) -> Contacts:
...
+    def _get_soft_self_contact_detector(self, contacts: Contacts) -> TriMeshCollisionDetector:
+        """Return the shared detector re-pointed at ``contacts.soft_self_contact_data``."""
+        data = contacts.soft_self_contact_data
+        if data is None:
+            raise ValueError(
+                "This Contacts buffer has no soft_self_contact_data; allocate it with "
+                "CollisionPipeline.contacts() after init_soft_self_contact()."
+            )
+        detector = self._ensure_soft_self_contact_detector()
+        if detector.collision_info is not data:
```
jcarius-nv, line 2027: The shared detector can still be rebound away from the solver-owned buffer. `SolverVBD` captures this detector and a device-side `TriMeshCollisionInfo` snapshot at construction, but any later `pipeline.collide(..., other_contacts, soft_self_contact=True)` comes through here and repoints the detector at `other_contacts`. The solver's `_collision_detection_penetration_free()` path does not bind it back, so detection writes to `other_contacts` while the force kernels read the stale owned snapshot.

I reproduced this at the current head: the detector was initially bound to `solver.contacts`, became bound to `other_contacts`, and remained there after invoking the solver detection path. Could we either rebind to the owned buffer before every solver detection, or prevent a solver-claimed detector from being rebound? A regression covering the solver-plus-second-Contacts sequence would protect this.

newton/_src/solvers/solver.py
```diff
@@ -196,13 +198,209 @@ class SolverBase:
...
+    def collision_frequency(self) -> dict[CollisionSlot, int]:
+        """Per-slot detection frequency numbers as a read-only copy."""
+        return dict(self._collision_frequency)
+
+    @property
+    def collision_frequency_type(self) -> dict[CollisionSlot, CollisionFrequencyType]:
+        """Per-slot :class:`CollisionFrequencyType` values as a read-only copy."""
+        return dict(self._collision_frequency_type)
+
+    def set_collision_frequency(
```
jcarius-nv, line 314: `CODING_GUIDELINES.rst` says new optional public parameters should normally be keyword-only. Since both parameters here are optional configuration, could we add `*` after `self`? We should transition these new APIs consistently to keywords before they ship.

newton/_src/sim/collide.py
```diff
@@ -1774,6 +1853,158 @@ def contacts(self) -> Contacts:
...
+            detector._bind_external_buffers(data)
+        return detector
+
+    def refit_soft_self_contact_bvh(self, new_pos: wp.array[wp.vec3], rebuild: bool = False) -> None:
```
jcarius-nv: Same keyword-only point for this new public method: `new_pos` is the primary operand, while `rebuild` is an optional mode. Could this be `refit_soft_self_contact_bvh(self, new_pos, *, rebuild=False)`?

newton/_src/geometry/tri_mesh_collision.py
```diff
@@ -0,0 +1,1025 @@
...
+    edge_colliding_edges_buffer_sizes: wp.array[wp.int32]
+    """Maximum stored collision count for each edge row."""
+    edge_colliding_edges_count: wp.array[wp.int32]
+    """Detected collision count for each edge; values may exceed row capacity."""
+    edge_colliding_edges_min_dist: wp.array[float]
+    """Minimum detected edge-edge distance for each edge [m]."""
+
+
+@wp.func
+def get_vertex_colliding_triangles_count(col_info: TriMeshCollisionInfo, v: int):
```
jcarius-nv: Could we reconsider the parameter names across this newly public accessor family before release? In generated API docs and external kernels, `col_info`, `v`, `e`, `tri`, and `i_collision` are quite terse out of context. `collision_info`, `vertex`, `edge`, `triangle`, and `collision_index` would better match Newton's public-name guidance and make the signatures self-explanatory.

newton/_src/solvers/solver.py
```diff
@@ -196,13 +198,209 @@ class SolverBase:
...
+                frequency = int(frequency_value)
+                if frequency < 1:
+                    raise ValueError(f"collision_frequency[{slot.name}] must be >= 1, got {frequency}")
+                freq[slot] = frequency
+
+        ftype = dict(self._collision_frequency_type)
+        if collision_frequency_type is not None:
+            for slot, value in collision_frequency_type.items():
+                ftype[Slot(slot)] = Frequency(value)
+            if ftype[Slot.RIGID] == Frequency.PRE_POST_INIT:
```
jcarius-nv: Issue #3846 asks for solver-owned scheduling per contact family, including detection before/after solver initialization. This rejects `PRE_POST_INIT` for the rigid slot, so that requested mode is not available for rigid contacts. Is there a technical reason a rigid post-initialization pass cannot exist? Otherwise, could we implement it or explicitly narrow the issue/API contract so the limitation is deliberate rather than implicit?

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -2114,13 +2270,45 @@ def step(
...
         if control is None:
             control = self.model.control(clone_variables=False)
 
         self._initialize_rigid_bodies(state_in, control, contacts, dt, update_rigid)
         self._initialize_particles(state_in, state_out, dt)
 
         for iter_num in range(self.iterations):
+            if (
+                self._rigid_mode_this_step == _Frequency.ITERATIONS
+                and iter_num > 0
```
jcarius-nv: `ITERATIONS` currently has different timing by contact family: rigid skips `iter_num == 0`, while soft self-contact runs when `iter_num % k == 0`, including iteration zero. With `k=1`, the first rigid solve uses the pre-initialization contacts, whereas soft self-contact refreshes after initialization before that solve. Could we define one precise meaning for "every k-th solver iteration" and align both families, or document why the difference is intentional?

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -3628,12 +3919,12 @@ def _collision_detection_penetration_free(self, current_state: State):
 
         self.trimesh_collision_detector.refit(current_state.particle_q)
         self.trimesh_collision_detector.vertex_triangle_collision_detection(
-            self.particle_self_contact_margin,
+            self._self_contact_query_radius,
```
jcarius-nv: Issue #3846 asks to fold cloth self-contact detection into `CollisionPipeline` / `Contacts`, but the owned-solver path still obtains the pipeline's private detector and performs refit plus vertex/edge queries directly in `SolverVBD`. That preserves a solver-specific execution path and also enables the shared-detector rebinding bug noted above. Could the pipeline own an operation that detects into the solver-owned `Contacts`, leaving VBD responsible only for scheduling it?

newton/geometry.py
```diff
@@ -36,6 +46,8 @@
     "HydroelasticSDF",
     "NarrowPhase",
     "ParticleSurface",
+    "TriMeshCollisionInfo",
```
jcarius-nv, line 40: Fit discussion only, not a general flag or blocker: exporting `TriMeshCollisionInfo`, its builder, and the eight storage-level accessors makes the packed buffer layout a durable public commitment. Do we have a standalone public consumer that needs this full surface? It may be worth keeping the builder/accessors internal or explicitly experimental until that consumer is demonstrated, while exposing a narrower result/query interface. Happy to discuss the intended use case.

jcarius-nv (commented):
Thank you for the substantial work here. Unifying the collision pipeline while preserving the existing external-driving workflow is a large and valuable change. I did another exact-head pass and left a focused set of comments covering one reproduced correctness issue, two public-API standards points, the three remaining issue #3846 requirements, and one non-blocking Fit question about the low-level geometry surface.

newton/_src/sim/collide.py
jcarius-nv: (or maybe we ignore this and just state the assumption somewhere, that's also fine)

AnkaChan:
Huge thanks @nvtw for carrying the review response on this branch! The PR came out of this review substantially stronger. Thanks also @eric-heiden and @jcarius-nv for the exceptionally thorough reviews.

newton/geometry.py
AnkaChan: I think it makes sense to make it experimental module. In fact we are also experimenting with different design choices, like global buffer vs per-element buffer, each has its benefits. I confirmed that we have no public consumer except the VBDSolver (Style3D has its own data structure, but it results our collision detection kernel). @nvtw has already marked this experimental.

newton/_src/solvers/vbd/solver_vbd.py
jcarius-nv: Addressed in `6db1f674`: margin-only validation now raises only when self-contact is enabled and the model has particles; otherwise the ignored legacy gap is clamped to zero. The disabled/no-particle and active-error cases are covered by the focused regression tests on CPU and CUDA.

newton/_src/sim/collide.py
jcarius-nv: Addressed in `6db1f674`: VBD now delegates to `CollisionPipeline._detect_soft_self_contact(...)`, which rebinds the shared detector to the supplied solver-owned `Contacts` on every pass. `test_vbd_self_contact_rebinds_owned_buffer` covers the second-buffer sequence on CPU and CUDA.

newton/_src/solvers/solver.py
jcarius-nv: Addressed in `6db1f674`: `set_collision_frequency()` now makes both optional configuration arguments keyword-only with `*`.

newton/_src/sim/collide.py
jcarius-nv: Addressed in `6db1f674`: `refit_soft_self_contact_bvh()` now keeps `new_pos` as the primary operand and makes `rebuild` keyword-only.

newton/_src/geometry/tri_mesh_collision.py
jcarius-nv: Addressed in `6db1f674`: the builder and eight packed-buffer accessors were removed from the public `newton.geometry` surface, and the internal accessor parameters were expanded to descriptive names.

newton/_src/solvers/solver.py
jcarius-nv: Addressed in `6db1f674`: the rigid-slot rejection was removed and VBD now implements `PRE_POST_INIT` with detection before initialization and a second pass plus contact-state refresh afterward. The new mode has CPU/CUDA regression coverage.

newton/_src/solvers/vbd/solver_vbd.py
jcarius-nv: Addressed in `6db1f674`: rigid and soft-self-contact `ITERATIONS` now share the documented meaning—one pre-initialization baseline, then detection immediately before iterations k, 2k, and so on—and the aligned schedules are regression-tested on CPU and CUDA.

newton/_src/solvers/vbd/solver_vbd.py
jcarius-nv: Addressed in `6db1f674`: self-contact execution now lives in `CollisionPipeline._detect_soft_self_contact(...)`; VBD is responsible only for scheduling it and selecting its owned `Contacts` buffer.

newton/geometry.py
jcarius-nv: Addressed for this PR: the builder and storage-level accessors remain internal, while the remaining public `TriMeshCollisionInfo` type and `Contacts.soft_self_contact_data` are explicitly experimental. Thanks for clarifying the current consumer and design-exploration status; this sufficiently narrows the concern.

jcarius-nv (approved):
Thanks for the work and the thoughtful follow-up. I re-reviewed head 6223e816: the detector ownership bug, keyword-only and naming points, and all three scheduling/spec findings are addressed. The low-level geometry surface is now appropriately narrow and experimental. All 30 focused CPU/CUDA tests and all required checks pass; the API is meaningful and consistent with the Newton coding guidelines, and the remaining Eric thread is addressed. Approving.

PR #3829, opened 2026-08-07
Collision performance improvements

nvtw:
## Description

Improve collision pipeline performance across broad phase, GJK/MPR, mesh and SDF contacts, hydroelastic processing, contact reduction, sorting, and matching. The changes reduce redundant work and temporary storage while preserving collision results.

This also adds optional scalar SDF texture storage for users who prefer lower memory usage over faster sampling.

## Checklist

## Test plan

- Full unit-test suite passes in CI.
- `uvx pre-commit run -a`
- Added regression coverage for optimized SAP scheduling, collision routing, contact matching and reduction, hydroelastic processing, mesh edge filtering, and scalar SDF textures.

## New feature / API change

Disable paired SDF samples to approximately halve SDF texture storage at the cost of slower sampling:

```python
import newton

builder = newton.ModelBuilder(sdf_texture_paired_samples=False)
mesh.build_sdf(paired_samples=False)
```

## Summary by CodeRabbit

* **New Features**
  * Added faster collision handling for convex, mesh-SDF, and hydroelastic shapes.
  * Mesh SDF builds now filter fully inward-facing edges by default, with an option to retain them.
  * Added cached mesh-edge data and specialized SDF processing for improved performance.
  * Added convex collision-mesh deduplication and improved sparse collision routing.

* **Bug Fixes**
  * Improved contact reduction stability, duplicate suppression, and dense-scene contact preservation.
  * Fixed edge cases in plane, capsule, cylinder, mesh, and hydroelastic collision responses.
  * Improved contact normal consistency and deterministic behavior.

* **Tests**
  * Expanded coverage for collision routing, SDF sampling, mesh filtering, and contact reduction.


newton/_src/sim/builder.py
```diff
@@ -11090,6 +11135,18 @@ def _shape_requests_planar_sdf(shape_idx: int) -> bool:
...
+                source = generated_shape_sources[shape_idx]
+                if shape_type != GeoType.CONVEX_MESH or not isinstance(source, Mesh):
+                    continue
+                source_identity = id(source)
+                if source_identity in deduplicated_convex_sources:
+                    generated_shape_sources[shape_idx] = deduplicated_convex_sources[source_identity]
+                    continue
+                deduplicated_source = _deduplicate_convex_collision_mesh(source)
+                deduplicated_convex_sources[source_identity] = deduplicated_source
+                generated_shape_sources[shape_idx] = deduplicated_source
```
mzamoramora-nvidia, line 11151: Could we remap `deferred_edges` to the deduplicated convex topology before indexing `mesh.vertices` here? The deferred SDF is built on a copy of the original mesh, while `mesh` here comes from `generated_shape_sources` and may have fewer vertices after `_deduplicate_convex_collision_mesh()`.

I can reproduce an `IndexError` with a duplicated-vertex box by adding it as a convex hull, setting `builder.shape_sdf_max_resolution[shape] = 16`, and calling `finalize()`:

`IndexError: index 10 is out of bounds for axis 0 with size 8`

This uses the deferred per-shape SDF builder state rather than the usual prebuilt-SDF convenience path, but `finalize()` explicitly handles that state for these shape types. Please either carry the vertex remap into the deferred-edge cache or keep the edges and vertices from the same topology. It would be good to add this case as a CUDA regression test.

newton/_src/geometry/contact_reduction_global.py
```diff
@@ -1794,33 +1870,18 @@ def export_reduced_contacts_kernel(
...
+                        contact_fingerprints,
+                        exported_flags,
+                        shape_types,
+                        shape_data,
+                        shape_gap,
+                        writer_data,
+                    )
 
             # Keep lanes together before the shared tile is reused.
             _sync = wp.tile_extract(duplicate_bits, lane)
```
mzamoramora-nvidia: `wp.tile_extract()` on a shared tile is only a read in the pinned Warp 1.17 native implementation; it does not issue a block barrier. That means other warps can begin the next `active_idx` iteration and overwrite `duplicate_bits` while lane 0 is still serially exporting the current entry.

Could we replace this with a supported cooperative block synchronization, or restructure the loop so the shared tile cannot be reused until all lanes have completed? A deterministic multi-entry stress test would help cover the race.

newton/_src/geometry/broad_phase_sap.py
```diff
@@ -318,93 +403,77 @@ def _sap_broadphase_kernel(
...
                 shape_body,
                 body_flags,
                 include_static_kinematic_pairs,
+                candidate_pair,
+                candidate_pair_count,
+                max_candidate_pair,
             )
-
-        workid += nsweep_in
+        chunk_base += chunk_stride
```
mzamoramora-nvidia: The earlier guard keeps `nsweep_in * 4` in int32 range, but repeated additions here can still overflow `chunk_base`. With roughly 65k mutually overlapping shapes, `total_work_packages` can remain a valid positive int32 while this sum wraps negative; the loop then continues with an invalid work ID and can hang.

Could we use wider arithmetic or check the remaining distance before adding the stride? A synthetic near-`INT32_MAX` mapping test could cover the boundary without allocating the corresponding candidate-pair array.

newton/_src/sim/builder.py
```diff
@@ -11497,6 +11555,16 @@ def compute_voxel_resolution_from_aabb(aabb_lower, aabb_upper, voxel_budget):
                         if deferred_key in deferred_collision_edges_cache:
                             deferred_collision_edges[i] = deferred_collision_edges_cache[deferred_key]
                     if mesh_sdf is not None:
+                        coarse_texture = getattr(mesh_sdf, "_coarse_texture", None)
+                        if coarse_texture is not None and (
+                            (coarse_texture.num_channels == 2) != self.sdf_texture_paired_samples
+                        ):
+                            mode = "paired" if self.sdf_texture_paired_samples else "scalar"
+                            raise ValueError(
```
mzamoramora-nvidia, line 11568: This validation makes `paired_samples` a model-wide layout choice: every prebuilt mesh SDF must agree with `ModelBuilder.sdf_texture_paired_samples`. The public `Mesh.build_sdf()` and `ModelBuilder` docstrings currently read like independent per-object controls. Could we document the matching requirement in both public parameter descriptions? Otherwise a valid prebuilt SDF can fail only when the model is finalized.

There are also several added specialization parameters, including `use_precomputed_center` and `paired_samples`, that are missing from their Google-style `Args:` sections. Could those be included in the documentation cleanup?

newton/tests/test_collision_pipeline.py
```diff
@@ -2390,6 +2441,28 @@ def test_mesh_convex_with_sdf_routes_to_sdf_contact(test, device):
...
+    builder.add_shape_mesh(body=body_mesh, mesh=mesh)
+    builder.add_shape_convex_hull(body=body_convex, mesh=convex)
+
+    model = builder.finalize(device=device)
+    pipeline = newton.CollisionPipeline(model, broad_phase="sap", rigid_contact_max=256)
+    contacts = pipeline.contacts()
+    pipeline.collide(model.state(), contacts)
+
+    test.assertEqual(mesh.sdf._coarse_texture.num_channels, 1)
+    test.assertGreater(int(contacts.rigid_contact_count.numpy()[0]), 0)
```
mzamoramora-nvidia: Could we strengthen the scalar collision and hydroelastic regressions into paired-vs-scalar differential tests? Checking only `contact_count > 0` proves that each route runs, but not the PR's "preserving collision results" claim. Comparing canonicalized exported contacts, including shape pairs, points, normals, penetration depth, and hydroelastic stiffness where applicable, would preserve that equivalence with an appropriate numerical tolerance. I did not observe a mismatch in the current test scenes; this is optional regression strengthening.

mzamoramora-nvidia (commented):
Thanks for putting this together. The performance direction looks worthwhile, and all reported CI checks pass, including the AWS GPU benchmark job.

An AI-assisted review also identified several points for improvement, noted in the inline comments below.

Given the size and breadth of the changes, it would be good to consider splitting this PR into smaller, independently reviewable chunks.

I found three correctness issues that seem worth addressing before merge:

- Deferred convex-SDF finalization can index deduplicated vertices with edge indices from the original mesh and raises `IndexError` for a duplicated-vertex box.
- `tile_extract()` is being used as if it were a block barrier before a shared tile is reused, but shared-tile extraction does not synchronize in the pinned Warp implementation.
- Dense SAP work scheduling can wrap `chunk_base` through signed int32 and continue with a negative work ID.

I left inline comments for those, plus two smaller documentation and test-coverage suggestions.

nvtw:
> Given the size and breadth of the changes, it would be good to consider splitting this PR into smaller, independently reviewable chunks.

I agree, splitting would be a good thing. But let's test the changes as monolith and split before merging. Otherwise we just test multiple times (for every split fragment).

newton/_src/sim/builder.py
```diff
@@ -11477,6 +11536,7 @@ def compute_voxel_resolution_from_aabb(aabb_lower, aabb_upper, voxel_budget):
                         sdf_kwargs["margin"] = sdf_gen_margin
                         sdf_kwargs["scale"] = tuple(shape_scale)
                         sdf_kwargs["texture_format"] = sdf_tex_fmt
+                        sdf_kwargs["paired_samples"] = self.sdf_texture_paired_samples
                         deferred_key = (
```
mzamoramora-nvidia, line 11543: Could this cache key distinguish the effective mesh topology? A regular mesh
and a convex hull created from the same `Mesh` currently use the same key, even
though the convex source is deduplicated before its deferred SDF is built.

If the regular mesh is cached first, the convex shape can reuse collision-edge
indices for the original topology. Those indices are then invalid for the
deduplicated mesh when `NarrowPhase` uses the raw-edge fallback. In a
duplicated-vertex box case, the precomputed route produced 12 contacts while
the fallback produced 14.

Could the key include the effective source topology or shape type? A regression
using one source as both a regular mesh and a convex hull, in both insertion
orders, would cover this.

newton/_src/geometry/narrow_phase.py
```diff
@@ -1838,6 +1973,7 @@ def launch_custom_write(
         mesh_edge_centers: wp.array[wp.vec4] | None = None,
```
mzamoramora-nvidia, line 2624: Could `mesh_edge_centers` and `mesh_edge_halves` be added to the `Args:`
section? It would help to state that they are paired precomputed edge arrays
and that omitting them selects the raw mesh-index fallback. That distinction is
important for understanding the two supported paths here.

newton/_src/geometry/sdf_hydroelastic.py
```diff
@@ -1225,8 +1395,8 @@ def decode_coords_8(bit_pos: wp.uint8) -> wp.vec3ub:
     )
 
 
-def create_count_iso_voxels_block_kernel(pressure_func: Any):
-    """Factory specializing :func:`count_iso_voxels_block` to a pressure callback.
+def create_count_iso_voxels_block_kernel(pressure_func: Any, integer_center: bool, paired_samples: bool):
```
mzamoramora-nvidia, line 1399: Could `paired_samples` be included in this factory's `Args:` section? The same
parameter is also missing from `get_generate_contacts_kernel()` below. A short
description of whether the generated kernel reads paired or scalar SDF texture
storage would complete the specialization documentation.

newton/_src/geometry/types.py
```diff
@@ -778,6 +778,7 @@ def build_sdf(
         texture_format: str = "uint16",
         sign_method: "SignMethod" = "auto",
         cache_dir: str | os.PathLike[str] | None = None,
+        paired_samples: bool = True,
         edge_lower_angle_threshold_rad: float = math.radians(0.1),
         edge_upper_angle_threshold_rad: float = math.radians(10.0),
         edge_inward_filter: bool = True,
```
mzamoramora-nvidia, line 784: `edge_inward_filter` is a new default-on behavior in the public
`Mesh.build_sdf()` API, but the PR's changelog fragment currently covers only
paired/scalar texture storage. Could this default change get a Towncrier
fragment as well?

mzamoramora-nvidia (commented):
Thanks for the updates. The current CI is green, and the aggregate benchmark
results move in the intended direction.

An AI-assisted review found one remaining correctness concern and a few smaller
points to improve, noted in the inline comments.

newton/_src/sim/builder.py
nvtw: Should be addressed by now

newton/tests/test_collision_pipeline.py
nvtw: Should be addressed by now

newton/_src/geometry/types.py
nvtw: Should be addressed by now

PR #3833, opened 2026-08-07
Fix 2D Warp mesh indices in normal computation

ruziniuuuuu:
## Summary

- flatten contiguous `(N, 3)` Warp triangle-index arrays before launching the 1D normal kernels
- reject unsupported Warp and NumPy index layouts before flattening
- cover flat and `(N, 3)` Warp inputs on CPU and CUDA when available
- cover flat and `(N, 3)` NumPy inputs plus invalid Warp and NumPy dimensions
- add a Towncrier fix fragment

Closes #3832.

## Validation

- `uv run --extra dev python -m unittest newton.tests.test_mesh_utils` — 3 passed on CPU and CUDA
- `uvx pre-commit run -a` — passed
- Towncrier draft build — passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation and handling of triangle index arrays when computing mesh vertex normals.
  * Added support for flattened and two-dimensional `(N, 3)` index formats across supported input types.
  * Invalid index dimensions and column counts now produce clear errors.

* **Tests**
  * Added coverage for CPU and Warp meshes, including expected triangle normals and invalid index layouts.

* **Documentation**
  * Documented support for two-dimensional triangle index arrays.

PR #3835, opened 2026-08-07
Keep API detection advisory on metadata 403

ruziniuuuuu:
## Summary

- keep API-change detection successful when repository policy denies label/comment writes
- emit a workflow warning for metadata-sync HTTP 403 responses
- continue failing on all other unexpected metadata errors
- preserve the existing missing-label HTTP 404 behavior

This fixes the `detect-api-changes` failures currently blocking #3831 and #3833. Both runs finish detection with `NEEDS_REVIEW=false`, then fail while removing the `api-changes` label because the workflow token receives `Resource not accessible by integration`. Since `pull_request_target` loads this workflow from the base branch, the affected PR branches cannot fix the failure themselves.

## Validation

- regression test failed before the workflow change
- `uv run --no-project -m unittest scripts/ci/tests/test_pr_api_changes_workflow.py` (6 passed)
- `uv run --no-project -m unittest discover -s scripts/ci/tests` (20 passed)
- `uvx pre-commit run -a` (passed)

## Risk

When repository policy denies metadata writes, the `api-changes` label and bot comment may remain stale or absent. The detector result remains available in the workflow log, and detector or non-permission failures are not suppressed.

## Summary by CodeRabbit

* **Bug Fixes**
  * API review metadata synchronization now continues safely when write permissions are insufficient.
  * Permission-related failures are reported as warnings without blocking the workflow.
  * Non-permission-related errors continue to be reported normally.
  * Synchronization warnings and skipped actions are surfaced clearly in workflow results.

* **Tests**
  * Added regression coverage for handling forbidden metadata permissions without blocking the workflow.

* **Documentation**
  * Documented that API-change detection remains advisory when metadata synchronization cannot update review information.

### DIFF
--- a/.github/workflows/pr_api_changes.yml
+++ b/.github/workflows/pr_api_changes.yml
@@ -97,56 +97,64 @@ jobs:
               issue_number: context.issue.number,
             };
 
-            if (needsReview) {
-              await github.rest.issues.addLabels({
-                ...issue,
-                labels: ['api-changes'],
-              });
-            } else {
-              try {
-                await github.rest.issues.removeLabel({
+            try {
+              if (needsReview) {
+                await github.rest.issues.addLabels({
                   ...issue,
-                  name: 'api-changes',
+                  labels: ['api-changes'],
                 });
-              } catch (error) {
-                // The current state is already correct when the label is absent.
-                if (error.status !== 404) throw error;
+              } else {
+                try {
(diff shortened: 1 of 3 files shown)


jcarius-nv:
Thank you for the contribution. We addressed this just now by changing the permission for the job. Please file follow-up issues in case there are still problems, otherwise we consider it solved.

PR #3838, opened 2026-08-07
Skip CUDA-only browser examples without CUDA

jcarius-nv:
## Description

Skip examples with known CUDA requirements in the manual example-browser smoke test when CUDA is unavailable.

The regular example test suite already limits these examples to `cuda_test_devices`, but `test_example_browser.py` attempted to instantiate them on CPU-only systems. On macOS this produced six expected switch failures from the CUDA-only texture SDF pipeline and MuJoCo sleeping example.

The browser smoke test now reports capability-based skips explicitly, includes the skip count in its summary, and exits successfully when every selected example is unsupported. Unmatched patterns and unexpected runtime failures remain errors.

## Checklist

- [x] `CHANGELOG.md` has been updated (not applicable; test-only behavior)

## Test plan

```bash
CUDA_VISIBLE_DEVICES=-1 uv run python newton/tests/test_example_browser.py brick_stacking contacts_rj45_plug mujoco_sleeping nut_bolt_hydro nut_bolt_sdf robot_panda_hydro
uvx --python 3.12 pre-commit run --files newton/tests/test_example_browser.py
```

The no-CUDA run reports all six examples as skipped and exits with status 0.

## Bug fix

**Steps to reproduce:**

1. Run on macOS or another system without a CUDA-capable device.
2. Run `uv run python newton/tests/test_example_browser.py`.
3. Observe six switch failures for examples that require CUDA.

**Minimal reproduction:**

```bash
CUDA_VISIBLE_DEVICES=-1 uv run python newton/tests/test_example_browser.py nut_bolt_sdf
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved example selection and skip handling in the smoke test.
  * Skip reasons are now reported clearly, including for CUDA-dependent examples.
  * Requests with no matching examples are handled before skip filtering.
  * When all selected examples are skipped, the test completes successfully without opening a viewer.
  * Final results now include skipped-example counts and distinguish between all examples passing and all runnable examples passing.

### DIFF
--- a/newton/tests/test_example_browser.py
+++ b/newton/tests/test_example_browser.py
@@ -32,10 +32,6 @@ import newton.viewer
 
 wp.init()
 
-SKIP_EXAMPLES = {
-    "robot_policy": "uses a non-standard constructor",
-}
-
 CUDA_REQUIRED_EXAMPLES = {
     "brick_stacking",
     "contacts_rj45_plug",
@@ -55,8 +51,6 @@ def _step_and_render(example, num_frames):
 
 
 def _get_skip_reason(name, cuda_available):
-    if name in SKIP_EXAMPLES:
-        return SKIP_EXAMPLES[name]
     if not cuda_available and name in CUDA_REQUIRED_EXAMPLES:
         return "requires CUDA"
     return None
--- a/newton/tests/test_example_browser.py
+++ b/newton/tests/test_example_browser.py
@@ -172,7 +172,7 @@ def main():
                 print(f"    {line}")
 
     if not switch_fail and not reset_fail:
-        print("\nAll examples passed!")
+        print("\nAll runnable examples passed!" if skipped else "\nAll examples passed!")
 
     return 1 if (switch_fail or reset_fail) else 0
 
--- a/newton/tests/test_example_browser.py
+++ b/newton/tests/test_example_browser.py
@@ -33,7 +33,16 @@ import newton.viewer
 wp.init()
 
 SKIP_EXAMPLES = {
-    "robot_policy",  # non-standard constructor: (viewer, config, asset_directory, mjc_to_physx, physx_to_mjc)
+    "robot_policy": "uses a non-standard constructor",
+}
+
+CUDA_REQUIRED_EXAMPLES = {
+    "brick_stacking",
+    "contacts_rj45_plug",
+    "mujoco_sleeping",
+    "nut_bolt_hydro",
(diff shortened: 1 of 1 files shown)


newton/tests/test_example_browser.py
```diff
@@ -33,7 +33,16 @@
...
+}
+
+CUDA_REQUIRED_EXAMPLES = {
+    "brick_stacking",
+    "contacts_rj45_plug",
+    "mujoco_sleeping",
+    "nut_bolt_hydro",
+    "nut_bolt_sdf",
+    "robot_panda_hydro",
 }
```
eric-heiden, line 37: ⚪ The skip reason for `robot_policy` is out of date. In `newton/examples/robot/example_robot_policy.py`, the class is now declared as `def __init__(self, viewer, args)`, and its `create_parser()` gives `--robot` a default of `g1_29dof`. That is exactly the contract this script drives: `getattr(example_class, "create_parser", create_parser)()` then `default_args(...)` then `example_class(viewer, args)`. The signature named in the removed comment, `(viewer, config, asset_directory, mjc_to_physx, physx_to_mjc)`, does not exist in that file anymore, so the claim was already stale before this PR and the rewrite restates it in prose.

Since the line is being rewritten anyway, it is cheap to make it accurate. If the example runs and resets cleanly now, the nicest fix is to drop the entry and let the sweep cover it. If it still needs to be skipped, the real blockers look like the module level `from warp_nn.runtime import OnnxRuntime` and the asset download in the constructor:

```suggestion
    "robot_policy": "requires warp-nn/ONNX runtime and downloads robot assets",
```

Either way, the reason string is what future readers will trust when deciding whether the skip is still needed, so it is worth having it describe something that is actually true.

jcarius-nv: Good catch. I removed the `robot_policy` skip entirely in 43107a2d rather than replacing it with another reason. The example now follows the standard `(viewer, args)` contract, and the required warp-nn/ONNX and asset-download dependencies are included in `newton[examples]`.

I verified the current behavior with a browser-style smoke check that constructs the default G1 policy example, steps and renders one frame, destroys it, and constructs it again in the same process. That passed both on CUDA and with `CUDA_VISIBLE_DEVICES=-1`; pre-commit also passes. The browser sweep will now cover `robot_policy` on both CPU-only and CUDA systems, while retaining conditional skips only for the six examples that actually require CUDA.

eric-heiden (approved):

PR #3839, opened 2026-08-07
Speed up MuJoCo collision mask compilation

eric-heiden:
## Description

Speed up `SolverMuJoCo` initialization for replicated models by querying collision exclusions only for the shapes selected for the MuJoCo template.

The reverse collision-mask compiler previously materialized every Newton shape-filter pair and then discarded pairs outside the selected template. For the G1 benchmark with 8,192 worlds, that expanded 11,173,888 filter pairs even though the generated MuJoCo model contains only five collision geoms.

This change instead enumerates the selected graph's upper triangle and queries those pairs through `Model.shape_collision_filter_mask()`. The work is bounded at 32,640 candidate pairs by the reverse compiler's existing 256-shape limit.

Focused measurements on an RTX 4090 Laptop GPU system:

| G1, 8,192 worlds | Before | After | Change |
|---|---:|---:|---:|
| Collision-mask compilation | 528 ms | 2.31 ms | 228x faster |
| `SolverMuJoCo` construction | 4.609 s | 4.268 s | -342 ms (-7.4%) |

The before/after mask arrays were verified to be exactly identical.

## Checklist

## Test plan

```text
uv run --extra dev -m unittest newton.tests.test_mujoco_solver.TestMuJoCoSolverCollisionMasks
# 4 passed

uv run --extra dev -m unittest newton.tests.test_collision_mask_compiler
# 21 passed

uvx --from towncrier==25.8.0 towncrier build --draft --version X.Y.Z --date 2026-08-07
# fragment rendered successfully

uvx pre-commit run -a
# all passed
```

The focused regression test was also run with the original global-scan implementation and failed as expected because it attempted to materialize the full sparse-filter array.

## Bug fix

**Steps to reproduce:**

1. Build the ASV G1 workload with 8,192 replicated worlds.
2. Construct `SolverMuJoCo` with MuJoCo contacts enabled.
3. Observe that reverse mask compilation scans every replicated-world filter pair to generate masks for the five selected template geoms.

## Summary by CodeRabbit

* **Performance**
  * Optimized MuJoCo solver initialization for replicated models by querying collision filters only for selected shapes.
  * Reduced unnecessary processing of model-wide collision-pair data.

* **Bug Fixes**
  * Improved collision-mask handling for selected shape pairs, ensuring filtering remains accurate during solver setup.

* **Tests**
  * Added coverage verifying that only relevant collision-graph pairs are queried and processed.

### DIFF
--- a/changelog/+mujoco-mask-init-7c2f1a4e.fixed.md
+++ b/changelog/+mujoco-mask-init-7c2f1a4e.fixed.md
@@ -1,0 +1,1 @@
+Speed up :class:`~newton.solvers.SolverMuJoCo` initialization for replicated models by querying collision filters only for the selected MuJoCo template shapes.
--- a/newton/_src/solvers/mujoco/solver_mujoco.py
+++ b/newton/_src/solvers/mujoco/solver_mujoco.py
@@ -5026,16 +5026,12 @@ class SolverMuJoCo(SolverBase, CouplingInterface):
         if selected_shapes.shape[0] > NEWTON_COLLISION_MASK_MAX_SHAPE_COUNT:
             return compile_newton_collision_graph(groups)
 
-        # The compiler derives group-based edges itself. Map only explicit
-        # exclusions whose two endpoints are in this solver's shape selection.
-        filter_pairs = model.shape_collision_filter_pairs_array()
-        if filter_pairs.shape[0]:
-            to_local = np.full(model.shape_count, -1, dtype=np.int64)
-            to_local[selected_shapes] = np.arange(selected_shapes.shape[0], dtype=np.int64)
-            local_filter_pairs = to_local[filter_pairs]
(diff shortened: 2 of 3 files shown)


jcarius-nv (approved):
Reviewed at `83217a15`.

- **Standards:** No blocking violations. The small duplicated upper-triangle traversal is not worth expanding this focused change to refactor.
- **Spec:** The implementation restricts collision-filter queries to the selected MuJoCo template shapes, preserves the 256-shape bound, and maps filtered global pairs back to local compiler indices. No gaps or scope creep found.
- **Fit:** **Fits.** This is a measured improvement to Newton's existing solver/collision path with no new public API or dependencies.
- **Validation:** The 4 focused MuJoCo collision-mask tests and 21 compiler tests pass. An additional 64-world replicated-model equivalence check produced the same local exclusions through the old and new paths. All GitHub checks, including the AWS GPU benchmark, pass.

This is a focused, well-tested improvement. Thanks, Eric!

PR #3842, opened 2026-08-07
Simplify omni wheel simulation

nvtw:
## Description

Simplify the omni-wheel simulation and add barrel-cylinder geometry for curved wheel rollers.

- Add optional `barrel_radius` support to cylinder construction, mesh generation, collision detection, bounds, mass properties, implicit MPM geometry, and viewer geometry while preserving straight cylinders by default.
- Add a robot example that drives a mecanum platform and a self-balancing ballbot using omni-wheels with barrel-shaped rollers.
- Document the cylinder convention and mass-property behavior, and add focused regression coverage for geometry, collision, inertia, bounds, validation, and viewer generation.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_cylinder_barrel_support
uv run --extra dev -m newton.tests -k test_barrel_cylinder
uv run --extra dev -m newton.tests -k test_cylinder_barrel_radius
uv run --extra dev -m newton.tests -k test_robot.example_robot_omniwheel
uvx pre-commit run -a
```

## New feature / API change

```python
import newton

body = builder.add_body()
builder.add_shape_cylinder(body, radius=0.5, half_height=1.0, barrel_radius=2.0)
```

## Summary by CodeRabbit

- **New Features**
  - Added optional circular-arc barrel profiles for cylinders.
  - Barrel cylinders support mesh generation, collision detection, raycasting, SDFs, mass properties, bounds, and viewer rendering.
  - Added a warehouse omniwheel robot simulation example.
- **Bug Fixes**
  - Improved cylinder contact handling and preserved original shape dimensions during solver setup.
  - Added barrel-radius validation while retaining standard-cylinder behavior by default.
- **Documentation**
  - Documented barrel-cylinder parameters, constraints, and mass properties.
  - Added the omniwheel example to the README.
- **Tests**
  - Added comprehensive coverage for barrel geometry and related simulation features.


eric-heiden:
Nice work simplifying the cylinder path, and the barrel profile fits naturally into the existing shape API.

A few things look worth resolving before merge: the new omniwheel example raises on construction, the viewer can throw on models imported from MJCF files that contain cylinder sites, and the new example is not registered for CI or the docs gallery.

newton/examples/robot/example_robot_omniwheel.py
```diff
@@ -0,0 +1,693 @@
...
+        self.ballbot_yaw_rate = math.inf
+
+        self.viewer.set_model(self.model)
+        self.viewer.set_camera(pos=wp.vec3(2.4, -3.2, 2.0), pitch=-24.0, yaw=128.0)
+        self.capture()
+
+    def capture(self):
+        """Capture one simulation frame."""
+        self.graph = None
+        if self.device.is_cuda:
```
eric-heiden, line 509: 🔴 `self.device` is never assigned anywhere in this file, but `capture()` reads `self.device.is_cuda`, and `__init__` calls `self.capture()` as its last statement. Constructing the example always fails:

```
File "newton/examples/robot/example_robot_omniwheel.py", line 503, in __init__
    self.capture()
File "newton/examples/robot/example_robot_omniwheel.py", line 508, in capture
    if self.device.is_cuda:
AttributeError: 'Example' object has no attribute 'device'
```

Since `__init__` raises before the run loop starts, the `test_final()` and `test_post_step()` checks in this file never execute either, so all of that validation is dead code today.

Assigning the device in `__init__` just before `self.capture()` is enough to let the example run to completion, and it matches what the other robot examples do (`example_robot_anymal_d.py`, `example_robot_h1.py`):

```python
self.device = wp.get_device()
self.capture()
```

newton/_src/viewer/viewer.py
```diff
@@ -1651,7 +1652,14 @@ def log_geo(
...
+            mesh = newton.Mesh.create_cylinder(
+                radius,
+                half_height,
+                up_axis=newton.Axis.Z,
+                barrel_radius=barrel_radius,
+                compute_inertia=False,
+            )
 
         elif geo_type == newton.GeoType.CONE:
             radius, half_height = geo_scale[:2]
```
eric-heiden, line 1657: 🟠 This branch passes `geo_scale[2]` as `barrel_radius` for every cylinder shape, but MuJoCo cylinder sites carry an unrelated third size component that is only for display. `_populate_shapes` builds geometry for every shape including sites, with no `ShapeFlags.SITE` filter, so `viewer.set_model(model)` on any model imported from an MJCF containing a cylinder site raises:

```
ValueError: barrel_radius must be zero or at least half_height
```

Concretely, `<site type="cylinder" size="0.02 0.1"/>` is imported with MuJoCo's default third component of `0.005`, giving `shape_scale = (0.02, 0.1, 0.005)`. That value is nonzero and smaller than the half height, so the new guard in `create_mesh_cylinder` fires. `parse_sites` defaults to `True`, so this is the normal import path, not an exotic one.

The builder side of this change already recognizes the hazard twice: `add_shape_cylinder` skips the barrel validation for sites, and the local AABB path zeroes the third component with a comment about the unused site display size. The viewer never got the same guard, and `implicit_mpm_model.py` has the same gap.

Suggested fix: in `_populate_shapes`, zero `geo_scale[2]` for `GeoType.CYLINDER` shapes whose `shape_flags` include `ShapeFlags.SITE` before the geometry is hashed and built. That also keeps site and non-site straight cylinders sharing one cached mesh.

newton/examples/robot/example_robot_omniwheel.py
```diff
@@ -0,0 +1,693 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
```
eric-heiden, line 1: 🔵 This new example is not registered anywhere, so it never runs in CI and it does not appear in the docs gallery.

- `newton/tests/test_examples.py` lists every example explicitly and has no auto discovery path, and there is no `robot.example_robot_omniwheel` entry. That means the assertions in this file (balance, velocity tracking, yaw rate, wheel engagement, chassis travel) never execute, so the barrel cylinder feature the example is written to exercise has no end to end regression coverage.
- `README.md` has no entry for it, and `docs/images/examples/` has no `example_robot_omniwheel.jpg`. All nine sibling robot examples have both.

Next step: add a `robot.example_robot_omniwheel` entry in `newton/tests/test_examples.py` next to the other robot examples, and add the `README.md` gallery entry with a 320x320 screenshot at `docs/images/examples/example_robot_omniwheel.jpg`, matching the sibling examples. If this one is too slow or too hardware dependent for the default CI tier, register it with the same gating the other heavy examples use rather than leaving it unregistered.

newton/_src/geometry/kernels.py
```diff
@@ -360,11 +432,14 @@ def sdf_cylinder(
         half_height [m]: Half-height along the cylinder axis.
         up_axis: Cylinder long axis as ``int(newton.Axis.*)``.
         top_radius [m]: Top radius. Negative values use ``radius``.
+        barrel_radius [m]: Radius of the circular side profile. Zero creates straight sides.
 
     Returns:
         Signed distance [m], negative inside, zero on surface, positive outside.
     """
     point_z_up = _sdf_point_to_z_up(point, up_axis)
+    if barrel_radius > 0.0:
```
eric-heiden, line 442: 🔵 `barrel_radius` is now part of three public surfaces, and each one enforces a different contract:

| API | `barrel_radius < half_height` | `barrel_radius` plus a different `top_radius` |
| --- | --- | --- |
| `Mesh.create_cylinder` | raises `ValueError` | raises `ValueError` |
| `ModelBuilder.add_shape_cylinder` | raises `ValueError` (sites exempt) | no `top_radius` parameter exists |
| `newton.geometry.sdf_cylinder` | accepted, clamped into a degenerate profile | accepted, `barrel_radius` silently wins |

The docstring here says only that zero creates straight sides, so someone calling the documented API has no way to learn either rule, and the combination the mesh path rejects is silently accepted by the SDF path. It would help to state the `barrel_radius >= half_height` requirement and the precedence over `top_radius` in the `sdf_cylinder` and `sdf_cylinder_grad` docstrings, so all three entry points describe the same contract.

Small unrelated point in the same change: the new `Mesh.create_cylinder` docstring lines in `types.py` (around line 400) use single backticks around `0.0`, while the surrounding lines use RST double backticks such as ``None`` and ``radius``. Single backticks render as italics in Sphinx rather than literal code.

newton/_src/utils/mesh.py
```diff
@@ -1757,46 +1762,60 @@ def add_vertex(position: np.ndarray, normal: np.ndarray | None, uv: tuple[float,
...
+                        ],
+                        dtype=np.float32,
+                    )[[x_dir, y_dir, z_dir]]
+
+                side_uv = (i / max(segments - 1, 1), v) if compute_uvs else None
+                ring_indices.append(add_vertex(position, side_normal, side_uv))
+            side_ring_indices.append(ring_indices)
+
+        side_bottom_indices = side_ring_indices[0]
+        side_top_indices = side_ring_indices[-1]
```
eric-heiden: ⚪ Two small cleanups.

1. `side_bottom_indices` and `side_top_indices` are assigned here but never read. The barrel side face loop iterates `side_ring_indices` directly, and the cap rings are rebuilt from scratch further down, so both assignments can be dropped.
2. `newton/examples/basic/example_basic_shapes.py` turns the gallery's canonical `CYLINDER` entry into a barrel via `barrel_radius=0.8`. The module header still advertises a variety of collision shapes, and the README screenshot for `basic_shapes` becomes stale. Adding a separate barrel entry, or leaving this one straight, keeps the reference gallery easy to read. If the change stays, the screenshot should be refreshed.

newton/_src/geometry/raycast.py
```diff
@@ -389,10 +392,39 @@ def ray_intersect_cylinder(ray_origin: wp.vec3, ray_direction: wp.vec3, r: float
...
+        tolerance = 1.0e-5 * wp.max(1.0, r + h)
+        for _ in range(64):
+            point = ray_origin + distance_along_ray * direction
+            distance = _sdf_barrel_cylinder_z(point, r, h, barrel_radius)
+            if wp.abs(distance) <= tolerance:
+                return distance_along_ray / direction_length, _sdf_barrel_cylinder_grad_z(point, r, h, barrel_radius)
+            distance_along_ray += wp.max(wp.abs(distance), tolerance)
+            if distance_along_ray > max_distance:
+                break
+        return -1.0, wp.vec3(0.0)
```
eric-heiden: ⚪ Not blocking, and possibly not even worth a follow-up. Just flagging it in case it is useful.

The new barrel branch lives inside `ray_intersect_cylinder`, which gets inlined into the tiled-camera render megakernel through `ray_intersect_shape`. Even though the branch is guarded by `barrel_radius > 0.0` and is never taken in a cylinder-free scene, the megakernel still gets measurably slower. That points at the added code affecting the whole kernel, for example register pressure or occupancy, rather than at the branch actually executing.

Measured on an L40: 1024 worlds at 64x64, 50 renders per sample, 9 samples per launch, min across 6+ independent process launches per config. Baseline is the merge base `8e938a02`, so only this branch's changes differ. The `nocyl` scene contains spheres, boxes and capsules only, so `ray_intersect_cylinder` is never called there at all.

| Scene | Metric | This branch | `8e938a02` | Delta |
| `nocyl` (no cylinders at all) | render megakernel | 3.807 ms | 3.466 ms | **+9.8%** |
| `nocyl` (no cylinders at all) | wall time / 50 renders | 0.2050 s | 0.1878 s | **+9.2%** |
| `mixed` (straight cylinders) | render megakernel | 3.786 ms | 3.474 ms | **+9.0%** |
| this branch, only `raycast.py` reverted | render megakernel | 3.469 ms | 3.466 ms | +0.1% |

The last row is the one that convinced me. Checking out just the merge-base `raycast.py` on top of this branch, leaving `support_function.py`, `kernels.py` and `narrow_phase.py` at branch state, puts the number back on baseline exactly. `ray_intersect_cylinder` has no callers outside `raycast.py`, so that ablation is fairly clean.

For contrast, the cost when barrel cylinders *are* used looks entirely reasonable and is not what I am asking about: an all-barrel scene renders at 6.749 ms versus 3.992 ms for straight cylinders with a matched footprint, which is about what an exact sphere-traced profile should cost.

So the only question is whether a roughly 10% tax on every ray-shape query is acceptable for users who never touch barrel cylinders. If the render megakernel turns out to be register limited, that would be worth knowing independently of this change.

<details>

<summary>Benchmark script (run from the repo root)</summary>

```python
"""Tiled-camera raycast throughput. Run from the newton repo root:

    BENCH_SCENE=nocyl BENCH_WORLDS=1024 BENCH_ITERS=50 BENCH_REPEAT=9 \
        python bench_tiled_camera_shapes.py

BENCH_SCENE picks the shape mix:
  mixed  - spheres/boxes/capsules/straight cylinders
  nocyl  - spheres/boxes/capsules only, so no cylinder is ever hit
  barrel - barrel cylinders only, exercising the new sphere tracer
  allcyl - straight cylinders matched to the barrel scene's footprint
"""

import importlib.util
import math
import os
import pathlib
import random
import sys
import time

import warp as wp

wp.config.enable_backward = False

import newton  # noqa: E402
from newton.sensors import SensorTiledCamera  # noqa: E402

print(f"newton package: {pathlib.Path(newton.__file__).resolve().parent}")

# Reuse the ASV tiled-camera rig so the harness is identical on both checkouts.
_BENCH = pathlib.Path("asv/benchmarks/simulation/bench_sensor_tiled_camera.py").resolve()
_spec = importlib.util.spec_from_file_location("_bench_tc", _BENCH)
_mod = importlib.util.module_from_spec(_spec)
sys.modules["_bench_tc"] = _mod
_spec.loader.exec_module(_mod)

SCENE = os.environ.get("BENCH_SCENE", "mixed")
WORLDS = int(os.environ.get("BENCH_WORLDS", "1024"))
ITERS = int(os.environ.get("BENCH_ITERS", "50"))
REPEAT = int(os.environ.get("BENCH_REPEAT", "9"))


def build() -> newton.ModelBuilder:
    """Build a 16x16 grid of primitives with the configured shape mix."""
    rng = random.Random(1234)
    for index in range(256):
        row, col = divmod(index, 16)
        position = wp.vec3((col - 7.5) * 0.8, (row - 7.5) * 0.8, 0.3 + 0.4 * rng.random())
        orientation = wp.quat_rpy(*(rng.uniform(-math.pi, math.pi) for _ in range(3)))
        body = builder.add_body(xform=wp.transform(position, orientation))
        kind = index % (3 if SCENE == "nocyl" else 4)
        if SCENE in ("barrel", "allcyl"):
            kind = 3
        if kind == 0:
            builder.add_shape_sphere(body, radius=0.18)
        elif kind == 1:
            builder.add_shape_box(body, hx=0.16, hy=0.16, hz=0.16)
        elif kind == 2:
            builder.add_shape_capsule(body, radius=0.12, half_height=0.14)
        elif SCENE == "barrel":
            builder.add_shape_cylinder(body, radius=0.12, half_height=0.18, barrel_radius=0.36)
        elif SCENE == "allcyl":
            # Same footprint as the barrel scene: equatorial radius of r=0.12, br=0.36.
            builder.add_shape_cylinder(body, radius=0.16823, half_height=0.18)
        else:
            builder.add_shape_cylinder(body, radius=0.16, half_height=0.18)
    return builder


preset = _mod.ScenePreset(build, camera_eye=(9.0, 9.0, 6.0), camera_target=(0.0, 0.0, 0.0))
rig = _mod._TiledCameraSceneRig(preset, WORLDS, 64, SensorTiledCamera.RenderOrder.TILED)
rig.render(color=True, depth=True)
wp.synchronize()

# Capture the render loop so timings measure GPU work, not per-launch CPU overhead.
    for _ in range(ITERS):
        rig.render(color=True, depth=True)
graph = capture.graph
wp.capture_launch(graph)
wp.synchronize()

samples = []
for _ in range(REPEAT):
    start = time.perf_counter()
    wp.capture_launch(graph)
    wp.synchronize()
    samples.append(time.perf_counter() - start)

samples.sort()
print(f"\n=== tiled_camera scene={SCENE} worlds={WORLDS} iters={ITERS} repeat={REPEAT} ===")
print(f"min={min(samples):.4f}s median={samples[len(samples) // 2]:.4f}s mean={sum(samples) / len(samples):.4f}s")

# Per-kernel device time for the raytrace megakernel via CUDA events.
mega = []
for _ in range(REPEAT):
    with wp.ScopedTimer("render", cuda_filter=wp.TIMING_KERNEL, print=False) as timer:
        rig.render(color=True, depth=True)
        wp.synchronize()
    mega.append(sum(r.elapsed for r in timer.timing_results if "megakernel" in r.name))
mega.sort()
print(f"megakernel_ms min={mega[0]:.4f} median={mega[len(mega) // 2]:.4f}")
```

</details>

nvtw: Should be improved now

newton/_src/solvers/implicit_mpm/implicit_mpm_model.py
```diff
@@ -130,10 +130,13 @@ def _get_shape_mesh(model: newton.Model, shape_id: int, geo_type: newton.GeoType
 
     elif geo_type == newton.GeoType.CYLINDER:
         radius, half_height = geo_scale[:2]
+        is_site = model.shape_flags.numpy()[shape_id] & int(newton.ShapeFlags.SITE)
+        barrel_radius = 0.0 if is_site else geo_scale[2]
```
eric-heiden: Why do we ignore the barrel radius if the cylinder is a site? That seems inconsistent, or is it an artifact from MuJoCo?
I would argue the user explicitly declared barrel radius so it should be considered here.

newton/_src/sim/builder.py
```diff
@@ -6425,6 +6425,8 @@ def add_shape(
             GeoType.GAUSSIAN,
         ):
             scale = (abs(float(scale[0])), abs(float(scale[1])), abs(float(scale[2])))
+            if type == GeoType.CYLINDER and not cfg.is_site and scale[2] != 0.0 and scale[2] < scale[1]:
+                raise ValueError(f"Cylinder barrel radius must be zero or at least the half-height; got scale={scale}.")
```
eric-heiden: Why can't cylinder sites not have a barrel radius?

newton/_src/solvers/implicit_mpm/implicit_mpm_model.py
nvtw: Should now be addressed

newton/_src/sim/builder.py
nvtw: Should now be addressed

eric-heiden:
Thanks, LGTM! @nvtw I think you need to update the tests that still encode some prior cylinder shape scale assumptions.
Should be good to merge then.

eric-heiden (approved):

Issue #3844, opened 2026-08-07
Kamino USD importer and ModelBuilder.add_usd diverge on assets with ArticulationRootAPI


andrewkaufman:
## Bug Description

`ModelBuilder.add_usd` and Kamino's `USDImporter` disagree about a USD asset once it declares `PhysicsArticulationRootAPI`. `add_usd` is articulation-aware; Kamino's importer has its own USD parser that ignores articulations entirely. On an asset without an articulation root the two agree, so the divergence is invisible until one is added.

This surfaced when bumping `NEWTON_ASSETS_REF` in #3821. newton-assets [#42](https://github.com/newton-physics/newton-assets/pull/42) added `PhysicsArticulationRootAPI` to the DisneyResearch models, and two Kamino tests started failing. Neither path looks obviously wrong to me, so this needs a decision from the Kamino/DisneyResearch side rather than a fix from mine.

### 1. Body ordering diverges

`newton/_src/solvers/kamino/tests/test_core_model.py::test_02_model_conversions_dr_testmech_from_usd` asserts that `ModelKamino.from_newton(model_newton)` equals a model built directly by Kamino's `USDImporter`. It now fails on `RigidBodiesModel.label`.

It is a **pure reorder** — same ten bodies, same set, no additions or removals. Loading `dr_testmech.usda` both ways at each asset revision:

| `dr_testmech.usda` | `ArticulationRootAPI` | `add_usd` order | Kamino `USDImporter` order |
| --- | --- | --- | --- |
| newton-assets `261cd1f` | 0 | `base, driver, link1, link2, link3, slider, cart, cartlink, endeffector, clevis` | identical |
| newton-assets `26c7f5c` | 1 | `base, cart, clevis, slider, cartlink, driver, endeffector, link1, link2, link3` | unchanged from above |

Kamino's ordering is the same at both revisions. `add_usd` is what changed, because it now groups bodies by articulation.

### 2. A root joint appears

`newton/_src/solvers/kamino/tests/test_utils_io_usd.py::test_import_model_dr_legs_with_boxes` asserts `num_joints == 36` and now sees 37.

`dr_legs_with_boxes.usda` contains 36 `PhysicsRevoluteJoint` prims and 31 rigid bodies at **both** revisions — only `ArticulationRootAPI` changed, 0 to 1. The extra joint is `world_to_pelvis`, synthesized for the articulation root; the other 36 are the expected `j1..j9` across four legs. `num_bodies` is still 31.

[...]

PR #3845, opened 2026-08-07
Read Kamino USD meshes via Newton's shared reader

andrewkaufman:
## Description

Kamino's USD importer cannot load a mesh that does not author normals. `normals` is a **built-in attribute of the `UsdGeom.Mesh` schema**, so `GetNormalsAttr().IsDefined()` is true for every mesh prim whether or not a value was ever authored:

```python
mesh = UsdGeom.Mesh.Define(stage, "/M")   # nothing authored
mesh.GetNormalsAttr().IsDefined()         # True
mesh.GetNormalsAttr().HasAuthoredValue()  # False
mesh.GetNormalsAttr().Get()               # None
```

Guarding on `IsDefined()` therefore passed `None` into `Mesh`, where `np.array(None).size == 1` and the reshape raised:

```
ValueError: cannot reshape array of size 1 into shape (3)
```

Rather than correct the guard in the two places that read mesh data, this delegates to `newton._src.usd.utils.get_mesh`, which already handles it and which the rest of Newton's USD import path uses. Kamino was maintaining a second, weaker USD mesh parser; that is the underlying problem, and the normals crash was one symptom of it.

The delegation also brings capabilities the Kamino path did not have:

| | before | after |
| --- | --- | --- |
| `primvars:normals` precedence | not read at all | honored |
| indexed primvars | unsupported | expanded |
| face-varying interpolation | ignored | converted |
| polygon support | triangles and quads only | any n-gon |

Behaviour is otherwise held constant: stage units stay unapplied and inertia uncomputed (`apply_stage_units=False`, `compute_inertia=False`), matching what this importer did before, and fan triangulation reproduces the previous triangle and quad output exactly — `(0,1,2)` and `(0,2,3)`.

Found while bumping `NEWTON_ASSETS_REF` in #3821: the DisneyResearch `dr_legs` asset hits this on four `merged_collision` prims.

## Checklist

## Test plan

Two new tests in `newton/_src/solvers/kamino/tests/test_utils_io_usd.py`, both verified to fail without the fix with the exact `cannot reshape array of size 1 into shape (3)` error:

- `test_import_mesh_with_declared_but_unauthored_normals` — a declared-but-unauthored `normals` attribute is treated as absent, and the quad is fan-triangulated to six indices.
- `test_import_mesh_prefers_primvars_normals` — `primvars:normals` is read in preference to the mesh attribute.

```
uv run --extra dev -m unittest newton._src.solvers.kamino.tests.test_utils_io_usd
uv run --extra dev -m unittest newton._src.solvers.kamino.tests.test_core_model
uv run --extra dev -m unittest newton.tests.test_import_usd
```

| suite | result |
| --- | --- |
| Kamino `test_utils_io_usd`, `test_core_model`, `test_utils_control_animation`, `test_solvers_forward_kinematics` | 98 tests, OK |
| `newton.tests.test_import_usd` | 307 tests, OK |
| `newton.tests.test_robot_composer` | 1 test, OK |
| `newton.tests.test_anymal_reset` | 4 tests, OK |

Against the assets proposed in #3821, this clears all three `ValueError` crashes. The remaining failures there are unrelated to normals and are tracked in #3844.

## Bug fix

**Steps to reproduce:** import any USD mesh that does not author normals through Kamino's `USDImporter`.

**Minimal reproduction:**

```python
from newton._src.solvers.kamino._src.utils.io.usd import USDImporter

xform = UsdGeom.Xform.Define(stage, "/Body")
UsdPhysics.RigidBodyAPI.Apply(xform.GetPrim())
mass = UsdPhysics.MassAPI.Apply(xform.GetPrim())
mass.CreateMassAttr(1.0)
mass.CreateDiagonalInertiaAttr(Gf.Vec3f(1.0, 1.0, 1.0))

mesh = UsdGeom.Mesh.Define(stage, "/Body/Mesh")
UsdPhysics.CollisionAPI.Apply(mesh.GetPrim())
mesh.CreatePointsAttr([Gf.Vec3f(0, 0, 0), Gf.Vec3f(1, 0, 0), Gf.Vec3f(1, 1, 0), Gf.Vec3f(0, 1, 0)])
mesh.CreateFaceVertexIndicesAttr([0, 1, 2, 3])
mesh.CreateFaceVertexCountsAttr([4])

USDImporter().import_from(stage, load_materials=False)
# ValueError: cannot reshape array of size 1 into shape (3)
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved USD mesh importing for meshes without authored normals.
  * Added support for normals precedence, indexed primvars, face-varying interpolation, and n-gon triangulation.
  * Preserved mesh units authored in the USD stage.
  * Applied consistent mesh handling to visual and physics shapes.

* **Tests**
  * Added coverage for missing normals and primvar normal precedence.

* **Documentation**
  * Documented the USD importer improvements.


changelog/+kamino-shared-mesh-reader-9d3f47ac.fixed.md
```diff
@@ -0,0 +1 @@
+Fix Kamino's USD importer failing to load any mesh that does not author normals. `normals` is a built-in attribute of the `UsdGeom.Mesh` schema, so the previous `IsDefined()` guard was always true and passed `None` through, raising `cannot reshape array of size 1 into shape (3)`. Mesh reading now goes through Newton's shared USD reader, which also adds `primvars:normals` precedence, indexed primvars, face-varying interpolation, and n-gon triangulation to that importer.
```
chschuma-disney, line 1: As an experimental solver, Kamino does not require a changelog entry, so it's up to you whether you want to keep this or drop it.

chschuma-disney (dismissed):
This looks good to me, thanks for fixing this!

The tests might have to be moved based on a recently merged PR that migrated all tests, sorry for the inconvenience.

changelog/+kamino-shared-mesh-reader-9d3f47ac.fixed.md
andrewkaufman: Thanks — I've kept it. The crash is user-visible (any mesh without authored normals fails to import outright, not just degrades), so it seemed worth a line for anyone who hits the `ValueError` and searches for it. Happy to drop it if you'd rather keep Kamino out of the changelog entirely as a matter of policy; it's a one-file delete.

andrewkaufman:
Thanks for the review! Rebased onto current `main` and addressed the CodeRabbit nitpicks. No behaviour change since you looked — the diff against my previous head is two renames plus one docstring cross-ref:

- `_read_usd_mesh` → `_read_mesh`. CodeRabbit suggested `_read_mesh_usd`, but every method on `USDImporter` reads USD, so the qualifier is redundant; this matches the file's existing `_get_prim_*` / `_parse_joint_*` verb-then-object pattern.
- `_define_quad_mesh` → `_define_mesh_quad` in the tests.
- The `get_mesh` reference now points at the public `newton.usd.get_mesh` rather than the `newton._src` path.

Heads up that the force-push dismissed your approval, so this needs a fresh one when you have a moment.

On the test-move note: #3807 left `test_utils_io_usd.py` in `newton/_src/solvers/kamino/tests/`, so nothing needed to move and the rebase came through clean — no inconvenience at all.

One thing that did surface while checking, worth separating from this PR: that file isn't under CI discovery. `-m newton.tests` starts at `newton/tests`, so all 59 tests in it — the two new ones included — only run when invoked explicitly. Not a regression from #3807; it's been that way since the file landed in #1915. I trialled the migration locally (move to `newton/tests/kamino/test_kamino_utils_io_usd.py`, retarget the `setup_tests` import) and all 59 pass under the runner, but since #3807 deliberately left non-user-facing test files behind, that call is yours — happy to open a follow-up if you want it in CI, or leave it alone.

For what it's worth, the fix itself is already covered in CI through a different path: `newton/tests/kamino/test_kamino_core_model.py::test_03_model_conversions_dr_legs_from_usd` runs `USDImporter` over `dr_legs_with_meshes_and_boxes.usda`. That asset has no normals-less meshes at the currently pinned `NEWTON_ASSETS_REF`, but at the ref proposed in #3821 it gains four (the `merged_collision` prims on the lower legs), and that test errors with the `ValueError` on `main` and passes with this change.

chschuma-disney (approved):
LGTM!

And leaving the test where it is was a deliberate choice, since it is planned to be removed in the near future.

PR #3851, opened 2026-08-08
Improve AWS runner attribution and watchdog coverage

shi-eric:
## Description

EC2 instances and attached EBS volumes created by Newton's AWS runner workflows now receive launch-time tags for the repository, normalized trigger category, workload, GitHub run ID, and run attempt. The repository tag uses `${{ github.repository }}` so authorized fork runs remain attributable.

Caller workflows explicitly classify pull requests, merge queue runs, pushes, scheduled nightly runs, and manual dispatches. This makes AWS Console spend easier to map back to the CI activity that created it.

This PR also checks the existing overdue-runner watchdog CloudFormation template into `scripts/ci/aws` with deployment instructions. The watchdog scans all running instances with Newton's ownership tag instead of filtering on a single repository, closing the monitoring gap for temporary development runs from authorized forks. Instances created before the new attribution tags are still reported with unknown metadata rather than ignored.

The watchdog continues to calculate runner age from the EC2 launch time, so this change does not add an expiration tag. It implements the attribution portion of the [Newton AWS EC2 runner mitigation plan](https://gist.github.com/shi-eric/92509b1b263c7bf398c4752729e06a48) without changing the runner provisioning or cleanup mechanism.

This is a CI infrastructure change and does not affect Newton's public API.

## Checklist

- [x] No changelog fragment is needed because this change affects CI infrastructure only

## Test plan

- [x] `uv run --no-project -m unittest discover -s scripts/ci/tests -v`
- [x] `uvx pre-commit run -a`
- [x] Ran the AWS GPU workflow from `shi-eric/newton` and confirmed that the EC2 instance and attached EBS volume received the expected dynamic repository, trigger, workload, run ID, and run-attempt tags before cleanup: https://github.com/shi-eric/newton/actions/runs/31274733298
- [x] Validated and deployed the committed CloudFormation template to the existing watchdog stack, invoked the Lambda function, and confirmed the stack was `UPDATE_COMPLETE`, drift status was `IN_SYNC`, and the alarm was `OK`

The local tests cover workflow configuration contracts and watchdog logic. They do not call AWS; the fork workflow run and watchdog deployment provide the live AWS verification.

## Summary by CodeRabbit

* **New Features**
  * Added automated monitoring for overdue CI runners, including scheduled checks, metrics, alarms, and notifications.
  * Added normalized workflow trigger categories and detailed attribution metadata to CI runner resources across benchmark, test, pull-request, push, and scheduled runs.
* **Documentation**
  * Added AWS CI infrastructure setup, validation, deployment, and rollback guidance.
* **Tests**
  * Added validation for workflow metadata, trigger configuration, runner tagging, and watchdog behavior.

### DIFF
--- a/.github/workflows/aws_gpu_tests.yml
+++ b/.github/workflows/aws_gpu_tests.yml
@@ -66,7 +66,10 @@ on:
       trigger-category:
         description: 'Normalized workflow trigger category'
         required: false
-        type: string
+        type: choice
+        options:
+          - manual
+          - scheduled-nightly
         default: 'manual'
 
 jobs:
--- a/.github/workflows/aws_gpu_tests.yml
+++ b/.github/workflows/aws_gpu_tests.yml
@@ -42,6 +42,10 @@ on:
         required: false
         type: string
         default: ''
+      trigger-category:
+        description: 'Normalized workflow trigger category'
+        required: true
+        type: string
     secrets:
       GH_PERSONAL_ACCESS_TOKEN:
         required: true
@@ -59,6 +63,11 @@ on:
         required: false
         type: string
         default: ''
+      trigger-category:
+        description: 'Normalized workflow trigger category'
+        required: false
+        type: string
+        default: 'manual'
 
 jobs:
   start-runner:
@@ -117,7 +126,11 @@ jobs:
             [
               {"Key": "Name", "Value": "ec2-github-runner"},
               {"Key": "created-by", "Value": "github-actions-newton-role"},
-              {"Key": "GitHub-Repository", "Value": "${{ github.repository }}"}
+              {"Key": "GitHub-Repository", "Value": "${{ github.repository }}"},
+              {"Key": "Newton-Trigger", "Value": ${{ toJSON(inputs['trigger-category']) }}},
+              {"Key": "Newton-Workload", "Value": "gpu-unit-tests"},
+              {"Key": "GitHub-Run-ID", "Value": "${{ github.run_id }}"},
+              {"Key": "GitHub-Run-Attempt", "Value": "${{ github.run_attempt }}"}
             ]
 
   gpu-unit-tests:
--- a/.github/workflows/mujoco_warp_tests.yml
+++ b/.github/workflows/mujoco_warp_tests.yml
@@ -31,7 +31,10 @@ on:
       trigger-category:
         description: 'Normalized workflow trigger category'
         required: false
-        type: string
+        type: choice
+        options:
+          - manual
+          - scheduled-nightly
         default: 'manual'
 
 jobs:
--- a/.github/workflows/mujoco_warp_tests.yml
+++ b/.github/workflows/mujoco_warp_tests.yml
@@ -16,12 +16,23 @@ env:
 
 on:
   workflow_call:
+    inputs:
+      trigger-category:
+        description: 'Normalized workflow trigger category'
+        required: true
+        type: string
     secrets:
       GH_PERSONAL_ACCESS_TOKEN:
         required: true
       CODECOV_TOKEN:
         required: true
   workflow_dispatch:
+    inputs:
+      trigger-category:
+        description: 'Normalized workflow trigger category'
+        required: false
+        type: string
+        default: 'manual'
 
 jobs:
   start-runner:
@@ -80,7 +91,11 @@ jobs:
             [
               {"Key": "Name", "Value": "ec2-github-runner"},
               {"Key": "created-by", "Value": "github-actions-newton-role"},
-              {"Key": "GitHub-Repository", "Value": "${{ github.repository }}"}
+              {"Key": "GitHub-Repository", "Value": "${{ github.repository }}"},
+              {"Key": "Newton-Trigger", "Value": ${{ toJSON(inputs['trigger-category']) }}},
+              {"Key": "Newton-Workload", "Value": "mujoco-warp-tests"},
+              {"Key": "GitHub-Run-ID", "Value": "${{ github.run_id }}"},
+              {"Key": "GitHub-Run-Attempt", "Value": "${{ github.run_attempt }}"}
             ]
 
   mujoco-warp-tests:
--- a/scripts/ci/aws/README.md
+++ b/scripts/ci/aws/README.md
@@ -1,0 +1,104 @@
+# AWS CI infrastructure
+
+This directory contains the source templates for AWS resources used by Newton
+CI. If the templates exist only in AWS, maintainers have to reconstruct the
+deployed configuration from console history during an incident. Checking them
+into Git gives us a reviewable change history and a known version to restore.
+
+That history is why these files belong in the repository. They are source files,
+not examples or generated output. The templates contain no credentials,
+secrets, private endpoints, or notification subscribers. They do not grant
+access to AWS, and commits to this directory do not deploy automatically.
+
+Put future AWS CI templates here too. This gives the small group that maintains
+the infrastructure one place to look and keeps AWS details out of the rest of
+the project.
+
+## Inventory
+
+| Template | Stack | Region | Required parameters |
+| --- | --- | --- | --- |
+| `overdue-newton-github-runner-watchdog.yaml` | `overdue-newton-github-runner-watchdog` | `us-east-1` | `AlertTopicArn` |
+
+The deployer must supply `AlertTopicArn` because the template has no default.
+This keeps notification routing and subscriber details outside the repository.
+
+## Validate and review a change
+
+Authenticate the AWS CLI, check that the selected profile points to the intended
+account, then validate the template:
+
+```bash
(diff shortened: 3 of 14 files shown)


.github/workflows/mujoco_warp_tests.yml
```diff
@@ -80,7 +91,11 @@ jobs:
             [
               {"Key": "Name", "Value": "ec2-github-runner"},
               {"Key": "created-by", "Value": "github-actions-newton-role"},
-              {"Key": "GitHub-Repository", "Value": "${{ github.repository }}"}
+              {"Key": "GitHub-Repository", "Value": "${{ github.repository }}"},
+              {"Key": "Newton-Trigger", "Value": "${{ inputs['trigger-category'] }}"},
+              {"Key": "Newton-Workload", "Value": "mujoco-warp-tests"},
+              {"Key": "GitHub-Run-ID", "Value": "${{ github.run_id }}"},
+              {"Key": "GitHub-Run-Attempt", "Value": "${{ github.run_attempt }}"}
```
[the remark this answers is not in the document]
shi-eric: @coderabbitai Fixed in 01d8b7479. All five runner workflows now serialize trigger-category with toJSON before constructing aws-resource-tags, covering the two noted sites and the three matching sites. Added a contract regression test that verifies quote- and newline-containing values remain valid JSON. Verified with uv run --no-project -m unittest discover -s scripts/ci/tests -v (25 tests) and uvx pre-commit run -a.

scripts/ci/tests/test_watchdog_logic.py
```diff
@@ -0,0 +1,147 @@
...
+        marker = "        ZipFile: |\n"
+        self.assertIn(marker, template)
+        source_lines = []
+        for line in template[template.index(marker) + len(marker) :].splitlines(keepends=True):
+            if line.strip() and not line.startswith("          "):
+                break
+            source_lines.append(line)
+        source = textwrap.dedent("".join(source_lines))
+        namespace = {}
+        exec(compile(source, str(TEMPLATE), "exec"), namespace)
```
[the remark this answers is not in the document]
shi-eric: @coderabbitai We are not adding this suppression. The repository Ruff configuration does not select the S rule family, and the actual pre-commit suite passes this file without S102. The exec is intentional test-only evaluation of Lambda source extracted from the checked-in CloudFormation template, so a noqa for a disabled rule would be misleading noise.

scripts/ci/aws/overdue-newton-github-runner-watchdog.yaml
```diff
@@ -0,0 +1,242 @@
...
+      Period: 21600
+      EvaluationPeriods: 1
+      DatapointsToAlarm: 1
+      Threshold: 1
+      ComparisonOperator: GreaterThanOrEqualToThreshold
+      TreatMissingData: breaching
+      AlarmActions:
+        - Ref: AlertTopicArn
+      OKActions:
+        - Ref: AlertTopicArn
```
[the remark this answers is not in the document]
shi-eric: @coderabbitai We are leaving this behavior unchanged. The one-time alarm notifications during initial stack creation are expected and acceptable. TreatMissingData: breaching is intentional so the alarm reports if the watchdog stops publishing after deployment, and the deployment procedure invokes the Lambda for operational verification. Adding metric seeding or staged alarm activation would add complexity solely to suppress an acceptable creation-time notification.

.github/workflows/aws_gpu_tests.yml
```diff
@@ -59,6 +63,11 @@ on:
         required: false
         type: string
         default: ''
+      trigger-category:
+        description: 'Normalized workflow trigger category'
+        required: false
+        type: string
```
jcarius-nv: ### Suggestion

Could we constrain or validate `trigger-category` against the supported vocabulary (`pull-request`, `merge-queue`, `push`, `scheduled-nightly`, `manual`) before constructing `Newton-Trigger`? This direct-dispatch input is a free-form string, so manual runs can create arbitrary or misleading categories and fragment the Cost Explorer grouping this change is intended to stabilize. It can also exceed EC2's 256-character tag-value limit. The same direct-dispatch definition appears in `minimum_deps_tests.yml`, `mujoco_warp_tests.yml`, and `warp_nightly_tests.yml`; a choice input plus allowlist validation at the leaf boundary would keep the value genuinely normalized.

GitHub permits large `workflow_dispatch` input payloads: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_dispatchinputs

EC2 tag values are limited to 256 Unicode characters: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#tag-restrictions

scripts/ci/aws/README.md
```diff
@@ -0,0 +1,104 @@
...
+the project.
+
+## Inventory
+
+| Template | Stack | Region | Required parameters |
+| --- | --- | --- | --- |
+| `overdue-newton-github-runner-watchdog.yaml` | `overdue-newton-github-runner-watchdog` | `us-east-1` | `AlertTopicArn` |
+
+The deployer must supply `AlertTopicArn` because the template has no default.
+This keeps notification routing and subscriber details outside the repository.
```
jcarius-nv, line 32: ### Missing acceptance evidence

Could we add the account-level activation and verification step for the billing goal? AWS requires user-defined keys such as `Newton-Trigger` and `Newton-Workload` to be activated as cost-allocation tags before they appear in billing reports. The current smoke test establishes that EC2 and EBS receive the tags, but not that Cost Explorer/CUR can filter or group by them. If activation has already been completed outside the repository, recording that evidence in the PR test plan would close the loop; otherwise this README is a good place to document activation, propagation delay, and one verification query.

AWS activation requirements: https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/activating-tags.html

scripts/ci/tests/test_watchdog_logic.py
```diff
@@ -0,0 +1,147 @@
...
+            "LaunchTime": now - timedelta(minutes=90),
+            "State": {"Name": "running"},
+            "Tags": tags,
+        }
+
+    def test_watchdog_finds_owned_runners_from_any_repository(self):
+        """Report an overdue owned runner regardless of repository tag value."""
+        namespace = self._lambda_namespace()
+        now = datetime(2026, 8, 8, 12, 0, tzinfo=timezone.utc)
+        instance = self._instance(now, repository="example/newton")
```
jcarius-nv, line 94: ### Optional / non-blocking

Consider adding a second instance whose `created-by` tag is not `github-actions-newton-role` and keeping the expected result unchanged. This test currently proves that an owned fork runner is included, but it would still pass if the production ownership filter were accidentally removed. The implementation is correct today; this would only strengthen regression coverage for the shared-account boundary.

jcarius-nv (commented):
Request one focused change before approval.

This fits Newton and materially improves attribution for AWS runner spend without adding public API or dependency burden. The implementation consistently adds repository, trigger, workload, run ID, and run-attempt tags across all five runner workflows, and the checked-in watchdog is appropriately owned alongside Newton's CI infrastructure. Standards review found no hard violations; the repeated tag schema across workflow files is a maintainability consideration, not a blocker.

Before approval, I think two things should be addressed:

- Enforce the promised normalized trigger vocabulary rather than accepting arbitrary direct-dispatch strings.
~~- Confirm and document that the new billing dimensions are activated and visible in Cost Explorer/CUR; applying EC2/EBS resource tags alone does not establish the shared-account billing outcome.~~

_Edit: I realized the second point does not work because this PR does not yet use the new CI configuration_

The existing aggregate filter remains `GitHub-Repository=newton-physics/newton`; this PR's main value is the additional trigger/workload/run breakdown. The watchdog ownership-filter test gap noted inline is optional and non-blocking because the production filter is currently correct.

Validation at `21093ea0`: all hosted checks pass, all 30 local CI-tool tests pass, and the three earlier inline threads are resolved. The reported fork smoke run provides live resource-tagging evidence, although the exact-head `pull_request_target` checks execute workflow definitions from the base branch.

.github/workflows/aws_gpu_tests.yml
shi-eric: Thanks for the reference. Updated all four direct-dispatch workflows to present manual and scheduled-nightly as the only choices. The reusable `workflow_call` input remains a string, and the contract test now checks the dispatch options and default.

scripts/ci/aws/README.md
shi-eric: We don't have the ability to create user-defined cost-allocation tags in the AWS account. The README has been updated to note this.

scripts/ci/tests/test_watchdog_logic.py
shi-eric: Added a focused regression test with an unowned instance to confirm the watchdog still returns only Newton-owned runners.

jcarius-nv (approved):
Thank you!

PR #3801, opened 2026-08-05
Improve sdf collision performance

nvtw:
## Description

Improve mesh-SDF contact performance by caching and pre-scaling collision-edge geometry, specializing texture-SDF and identity-scale paths, reducing redundant SDF samples and transforms, deduplicating corner contacts, filtering fully inward mesh edges, and reducing contact-reduction overhead.

### Performance

Nut-and-bolt captured collision pass:

| Worlds | `main` | This branch | Time reduction | Throughput |
|---:|---:|---:|---:|---:|
| 100 | 29.13 ms | 15.45 ms | **47.0%** | **1.89x** |
| 500 | 125.14 ms | 72.52 ms | **42.1%** | **1.73x** |
| 2,000 | 475.04 ms | 290.86 ms | **38.8%** | **1.63x** |

Hot mesh-SDF collision kernel:

| Worlds | `main` | This branch | Time reduction | Throughput |
|---:|---:|---:|---:|---:|
| 100 | 28.79 ms | 13.58 ms | **52.8%** | **2.12x** |
| 2,000 | 495.53 ms | 278.86 ms | **43.7%** | **1.78x** |

Measured on an RTX 3080 Laptop GPU with one nut-and-bolt assembly per world, SAP broad phase, contact reduction, and the example's default collision configuration.

## Checklist

- [x] New or existing tests cover these changes
- [x] The documentation is up to date with these changes
- [x] Towncrier changelog fragments have been added for user-facing changes

## Test plan

```text
uv run --extra dev -m newton.tests -k test_finalize_packs_collision_edge_geometry
uv run --extra dev -m newton.tests -k test_texture_sdf_paired_hw_gradient_matches_scalar_cuda_0
uv run --extra dev -m newton.tests -k test_mesh_mesh_sdf_vs_bvh_nxn_cuda_0
uv run --extra dev -m newton.tests -k test_11_dr_legs_dvi_contact_force_balances_weight
uv run --extra dev python -m unittest newton.tests.test_mesh_edge_angle_filter newton.tests.test_sdf_contact
uvx pre-commit run -a
```

## Summary by CodeRabbit

* **New Features**
  * Mesh SDF construction now filters fully inward collision edges by default, with an option to preserve them.
  * Added support for scaled meshes and precomputed edge geometry during collision detection.
  * Improved hardware SDF sampling and gradient accuracy.
  * Added face-normal utilities for contact reduction.

* **Bug Fixes**
  * Improved contact deduplication and retention in dense scenes.
[...]

PR #2366, opened 2026-04-08
Add support for speculative contacts

nvtw:
## Description

Add support for speculative contacts. They are the easiest way to get a decent amount of continuous collision detection for fast moving objects while keeping the complexity low.

## Checklist

- [x] New or existing tests cover these changes
- [x] The documentation is up to date with these changes
- [x] `CHANGELOG.md` has been updated (if user-facing change)

## Summary by CodeRabbit

* **New Features**
  * Speculative contact detection to catch fast-moving objects (configurable extension and update interval); new top-level SpeculativeContactConfig export
  * Contact matching for frame-to-frame rigid contact consistency with three matching modes; per-call dt override for collision update

* **Bug Fixes**
  * Per-frame contact report buffers are reliably cleared on reset

* **Documentation**
  * Guides and API docs for speculative contacts and contact matching

* **Tests**
  * New tests covering speculative contacts and tunneling scenarios (primitive and mesh)

PR #3854, opened 2026-08-10
Add predictive rigid contact manifolds

ooctipus:
## Description

This draft provides a standalone reference implementation of predictive rigid-contact manifolds for collision pipelines whose collision discovery runs less frequently than the solver.

The branch is a single commit directly on Newton `main`. It neither contains nor depends on #3801 or #2366.

The implementation adds:

- An opt-in `CollisionPipeline.SpeculativeContactConfig`, compiled out when disabled.
- An exact per-call collision horizon through `CollisionPipeline.collide(..., dt=...)`.
- Conservative swept broad-phase bounds with exact contact-point closing-speed eligibility, including angular velocity.
- Separate predictive search distance and physical contact distance so separated objects do not hover.
- A bounded mesh-SDF predictive manifold with six fingerprint-sharded nearest-clearance candidates and one earliest-impact guard.
- Atomic provisional reducer claims that prevent concurrent predictive candidates from exhausting a bounded contact buffer.
- Explicit rejection of unsupported hydroelastic combinations.

Predictive candidates preserve current physical geometry. They become available to the solver before the next collision refresh, but do not manufacture penetration or apply an impulse while the bodies remain physically separated.

### Motivation

A thin, tilted rod can move far enough between collision refreshes that its physical contact feature disappears before the solver can activate it. Raising collision frequency reduces the failure, but adds substantial collision cost. This change instead retains only candidates that exact point velocity predicts will close within the collision-update horizon.

### Known limitations

The exported predictive manifold is bounded to seven contacts per shape pair. Atomic provisional claims remove the observed concurrent allocation stampede, but transient allocations are not formally constant-bounded under adversarial monotonic winner ordering; strict allocation bounds would require two-pass selection or fixed per-pair staging.

Hydroelastic contact generation is not supported when predictive contacts are enabled.

## Test plan

Clean `main`-based branch:

- Speculative-contact suite: 48/48 passed across CPU, CUDA 0, and CUDA 1.
- Speculative mesh/SDF suite: 4/4 passed across CUDA 0 and CUDA 1.
- Global contact-reducer suite: 69/69 passed across CPU, CUDA 0, and CUDA 1.
- SDF-contact suite: 1/1 passed on CUDA.
- Full pre-commit suite passed.

Regression-gate verification:

- Before the atomic reservation fix, 256 concurrent candidates with a 16-contact buffer failed 20/20 trials in both fast and deterministic modes, allocated 256 entries, and lost nearest/earliest-impact winners.
- With the fix, the combined contention/capacity regression retains the exact seven expected winners and stays within the 16-entry capacity in both modes.

The same predictive-manifold behavior was also validated in the downstream IsaacLab Factory rod-on-board reproduction. That integration branch carried unrelated SDF performance work which is deliberately absent here:

- Existing collision path: 37/128 rods crossed the board medial surface and remained stuck.
- Predictive sharded manifold: 0/128 crossed and 0/128 remained stuck.

External reproduction:

    uv run isaaclab play --task=IsaacContrib-Factory-Franka --rl_library=rsl_rl physics=newton_mjwarp presets=rod_insert_16mm --num_envs=128 --viz=newton

## New feature / API change

    config = newton.CollisionPipeline.SpeculativeContactConfig(
        collision_update_dt=0.01,
        max_speculative_extension=0.01,
    )
    pipeline = newton.CollisionPipeline(model, speculative_config=config)
    pipeline.collide(state, contacts, dt=time_until_next_collision_refresh)

## Checklist

- [x] New and existing tests cover these changes
- [ ] A changelog fragment has been added

Documentation and a changelog fragment are intentionally deferred while maintainers decide whether this reference implementation should become the final API.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in speculative contacts to predict collisions over a configurable time horizon, with support for linear and angular motion, and mesh/SDF contacts.
  * Added swept-AABB detection across broad-phase collision modes.
  * Improved predictive contact selection, retention, and buffer reuse.

* **Bug Fixes**
  * Ensured existing collision behavior is unchanged when speculative contacts are disabled.
  * Added validation for unsupported speculative configurations including hydroelastic combinations.
  * Prevented false collision candidates from swept volume overlaps.


newton/_src/geometry/broad_phase_nxn.py
```diff
@@ -500,6 +499,10 @@ def launch(
             shape_body = wp.empty(0, dtype=wp.int32, device=device)
         if body_flags is None:
             body_flags = wp.empty(0, dtype=wp.int32, device=device)
+        if shape_sweep is None:
```
camevor: This allocation and the checks could be avoided by passing `shape_sweep` directly to the kernel, even if it's None (also for other broad phases). One check that would be nice is on the length of the shape_sweep array.

changelog/+speculative-contacts-d4e9c2a1.added.md
```diff
@@ -0,0 +1 @@
+Add opt-in speculative rigid-contact manifolds through `CollisionPipeline.SpeculativeContactConfig`, the `speculative_config` constructor argument, and per-call `CollisionPipeline.collide(..., dt=...)` horizon overrides, with continuous swept-AABB filtering and predictive contact-reduction lanes, and reject unsupported hydroelastic combinations.
```
camevor: I think the term contact manifold is neither necessary nor uniformly applicable. Could this be simplified to just speculative contacts?

newton/_src/geometry/broad_phase_sap.py
camevor, line 558: Pre-existing: shape_count appears to be unused

newton/_src/geometry/broad_phase_sap.py
```diff
@@ -572,6 +597,9 @@ def launch(
                 an all-static model when ``shape_body`` is provided.
             include_static_kinematic_pairs: Whether to include pairs where both shapes are immovable. Set to
                 ``False`` to filter static-static, static-kinematic, and kinematic-kinematic pairs.
+            shape_sweep: Optional world-space AABB translation over normalized time ``[0, 1]`` [m].
+            shape_sweep_projection_limit: Optional non-negative per-shape projection limit used only by the coarse
+                SAP sort [m]. The exact pair test continues to use the full relative sweep.
```
camevor: The wording here risks implying that this is a coarse, conservative pre-filter. However, the limit can cause the SAP stage to drop pairs, creating false negatives.

newton/_src/geometry/contact_data.py
```diff
@@ -87,6 +87,59 @@ def make_contact_sort_key(shape_a: int, shape_b: int, sort_sub_key: int) -> wp.i
...
+    point_a: wp.vec3,
+    point_b: wp.vec3,
+    normal_a_to_b: wp.vec3,
+    shape_transform: wp.array[wp.transform],
+    shape_linear_velocity: wp.array[wp.vec3],
+    shape_angular_velocity: wp.array[wp.vec3],
+) -> float:
+    """Return the closing speed of two shape points along a contact normal.
+
+    ``shape_linear_velocity`` is the world-space velocity of each shape
```
camevor: This explanation is confusing and seems redundant

newton/_src/geometry/narrow_phase.py
camevor, line 885: The speculative broad phase accounts for translational motion through shape_sweep, but the external-AABB infinite-plane precheck uses the current AABB without that sweep and runs before the expanded shape_gap is applied. This can reject a valid swept GJK candidate. For example, a cone moving toward an infinite plane crosses it within the configured horizon and reaches broad phase, but produces no contact until its current bounding sphere overlaps the plane. Could this precheck include the translational sweep/search extension, or be skipped for speculative candidates?

newton/_src/sim/collide.py
```diff
@@ -1044,11 +1182,20 @@ def __init__(
             else:
                 raise ValueError(f"Unsupported broad phase mode: {self.broad_phase_mode}")
 
+            if self._speculative_enabled:
+                shape_flags_np = model.shape_flags.numpy()
+                is_hydroelastic = (shape_flags_np & int(ShapeFlags.HYDROELASTIC)) != 0
+                shape_pairs_np = model.shape_contact_pairs.numpy().reshape(-1, 2)
```
camevor: This should have a None guard

newton/_src/geometry/contact_reduction_global.py
```diff
@@ -1087,12 +1199,22 @@ def export_contact_to_buffer(
...
-    contact_id = buffer_idx + 1  # ID zero is reserved for provisional winners.
+    if buffer_idx < reducer_data.capacity:
+        contact_id = buffer_idx + 1  # ID zero is reserved for provisional winners.
+    else:
+        contact_id = int(0)
+        if reducer_data.reclaimed_contact_head.shape[0] > 0:
+            contact_id = _pop_reclaimed_contact_id(reducer_data)
+        if contact_id == 0:
+            # Failed allocations remain above capacity so the difference is
+            # the drop count reported by buffer diagnostics.
```
camevor: No such diagnostic is being reported

newton/_src/geometry/narrow_phase.py
camevor, line 1592: `NarrowPhase(speculative=True, contact_writer_warp_func=...)` silently ignores the speculative flag.

camevor (commented):
Thanks for implementing this! The implementation looks good to me.
Since this is a nontrivial feature, the PR description should be updated to the current state for future reference.

A fully dynamic regression test for tunneling would make sense.
Finally, a test comparing non-moving contacts across speculative-enabled and -disabled pipelines would make sense, since speculative contact generation is enabled globally, affecting also processing of contacts that are between non-moving objects.

newton/_src/sim/collide.py
```diff
@@ -887,6 +1013,11 @@ def __init__(
                 ``True``.  Overhead is one extra kernel launch per collision
                 pass; disable in hot loops or CUDA graph capture once buffer
                 sizes are known to be adequate.
+            speculative_config: Optional predictive-contact configuration.
```
eric-heiden: I think the user-facing documentation here is still a bit lacking. We should explain speculative contacts in `collisions.rst` as well, and add more explanations here to the user-facing arguments, like this `speculative_config` (which should also link to the section in `collisions.rst` that will be added) and explain further what the settings in `SpeculativeContactConfig` mean exactly. These are the only places the user sees so we should make sure everything is easy to understand.

newton/_src/geometry/contact_reduction_global.py
```diff
@@ -1063,6 +1131,50 @@ def get_data_struct(self) -> GlobalContactReducerData:
...
 
+@wp.func
+def reclaim_contact_id(contact_id: int, reducer_data: GlobalContactReducerData):
+    """Return an unreferenced materialized contact ID to the reservation stack."""
+    if reducer_data.reclaimed_contact_head.shape[0] == 0:
+        return
+    head_value = reducer_data.reclaimed_contact_head[0]
+    while True:
+        head_id = int(head_value & wp.uint64(0xFFFFFFFF))
+        reducer_data.reclaimed_contact_next[contact_id] = head_id
```
camevor: `wp.atomic_cas` does not order the preceding write to `reclaimed_contact_next`. Another thread removing an ID from the free stack can therefore observe the new head with a stale next pointer, corrupting the stack and potentially allocating the same contact ID twice. The version tag prevents ABA on the head but does not provide this ordering.

camevor (dismissed):

newton/_src/geometry/contact_reduction_global.py
```diff
@@ -1602,6 +1945,67 @@ def reduce_buffered_contacts_kernel(
...
+                shape_transform,
+                shape_collision_aabb_lower,
+                shape_collision_aabb_upper,
+                shape_voxel_resolution,
+            )
+        else:
+            export_and_reduce_predictive_contact(
+                shape_a,
+                shape_b,
+                wp.vec3(pd[0], pd[1], pd[2]),
```
eric-heiden, line 1992: 🟠 `surface_offset_sum` adds the effective radii on top of the margins, but `pd[3]` has already had those radii removed. So the clearance used for routing here, and the one re-derived inside `export_and_reduce_predictive_contact`, is short by `radius_eff_a + radius_eff_b`.

`create_compute_gjk_mpr_contacts` shrinks sphere and capsule cores to `small_radius`, and `post_process_axial_on_discrete_contact` then does `contact_distance -= radius_eff_a` / `-= radius_eff_b` under exactly the `SPHERE or CAPSULE` test that makes `compute_effective_radius` non-zero. Everywhere else the canonical separation is `depth - margin_a - margin_b`: in `prepare_speculative_contact` the radii are added back into `point_a`/`point_b` and then cancel out, and the sibling call in `sdf_contact.py` passes `margin_sum` with both effective radii set to `0.0`.

Impact: for sphere or capsule against mesh or heightfield, which route to `shape_pairs_mesh` and are buffered by `mesh_triangle_contacts_to_reducer_kernel`, the clearance is understated by `r_a + r_b`. With a radius 0.1 sphere over a collision mesh, every candidate up to 0.1 m past the configured gap is pushed into the regular normal-bin competition instead of the bounded speculative manifold, while the candidates that do reach the manifold are admitted with an understated clearance (`clearance > closing_speed * dt` under-rejects) and are then dropped by the exact writer check at export, after consuming the limited predictive slots. Both directions lose speculative contacts on the shape combination this feature is aimed at.

Passing margins only, matching the SDF caller, fixes both sides. `radius_eff_a` and `radius_eff_b` are forwarded separately, so the contact-point reconstruction inside `export_and_reduce_predictive_contact` stays correct:

```python
surface_offset_sum = shape_data[shape_a][3] + shape_data[shape_b][3]
```

Two follow-ups while you are in there: the `surface_offset_sum` parameter docstring on `export_and_reduce_predictive_contact` still says "Combined effective radii and collision margins", and `test_speculative_contacts.py` has no sphere or capsule against mesh case, which is why this slips through today.

newton/_src/geometry/narrow_phase.py
```diff
@@ -752,6 +863,15 @@ def narrow_phase_kernel_gjk_mpr(
...
                 bsphere_center_b, bsphere_radius_b = compute_bounding_sphere_from_aabb(aabb_b_lower, aabb_b_upper)
 
+                # External AABBs describe the current geometry. Include both shapes' search
+                # extensions so relative translational motion cannot be culled before GJK.
+                if wp.static(external_aabb):
+                    pair_search_extension = shape_gap[shape_a] + shape_gap[shape_b]
+                    if is_infinite_plane_a:
+                        bsphere_radius_b += pair_search_extension
+                    else:
+                        bsphere_radius_a += pair_search_extension
```
eric-heiden, line 883: 🔵 This block is gated on `wp.static(external_aabb)` only, not on speculative mode, so it also runs for the default non-speculative `CollisionPipeline`, which supplies external AABBs as well.

`bsphere_radius_a` / `bsphere_radius_b` are not just a cull radius here. They are forwarded into `create_find_contacts` a few lines below, which sizes the infinite-plane proxy via `convert_infinite_plane_to_cube(..., bsphere_radius_b + rigid_gap)` with `lateral_size = depth = other_radius * 10.0`. With the default `rigid_gap = 0.1` on both shapes, every ground-plane pair in the default pipeline now hands GJK/MPR a proxy box whose half extents grew by 2.0 m, plus a correspondingly looser bounding-sphere overlap test.

On that path the expansion is also redundant: `compute_shape_aabbs` already enlarges each external AABB by `margin + shape_gap[shape_id]`, which is why `collide()` passes `None` for the broad-phase gap array with the note that AABBs are already expanded by effective gaps. The comment above describes the speculative requirement, where `shape_gap` is the velocity-expanded `_shape_search_gap`; on the default path `shape_gap` is `model.shape_gap` and the motion term is zero.

Two ways to scope it:

- give `create_narrow_phase_kernel_gjk_mpr` a `speculative` specialization flag, the way `create_narrow_phase_primitive_kernel` already takes one, and apply the extension only when it is set, or
- pass `shape_base_gap` alongside and add only the velocity part, `(shape_gap[shape_a] - shape_base_gap[shape_a]) + (shape_gap[shape_b] - shape_base_gap[shape_b])`, which is zero on the default path.

newton/_src/sim/collide.py
```diff
@@ -887,6 +1018,13 @@ def __init__(
...
```
eric-heiden: ⚪ The public API is named `speculative`, but a few docstrings still call it "predictive". This line reads "Optional predictive-contact configuration" while the very next sentence says "disables speculative contacts".

```suggestion
            speculative_config: Optional speculative-contact configuration.
```

The same wording survives in a few other spots:

| Location | Current text |
| --- | --- |
| `newton/_src/geometry/narrow_phase.py:1661` | "exact predictive-contact data" |
| `newton/_src/geometry/narrow_phase.py:2020` | "Maximum predictive clearance [m]" for `max_speculative_extension` |
| `newton/_src/geometry/narrow_phase.py:2565-2566` | "Predictive collision horizon [s]" and "Maximum predictive clearance [m]" |

One more for the sweep: `docs/concepts/simulation_tuning_mujoco.rst:271` still says that `gap` adds "speculative detection distance", which is the older meaning this PR moved away from in `docs/concepts/collisions.rst:1306`. Right now the two concept pages define "speculative" differently. Rewording it the same way, for example "an early detection distance", would keep the term reserved for the new velocity-adapted mechanism.

newton/_src/geometry/contact_reduction_global.py
nvtw: Done

newton/_src/geometry/narrow_phase.py
nvtw: Done

newton/_src/sim/collide.py

newton/_src/geometry/broad_phase_sap.py
```diff
@@ -572,6 +602,10 @@ def launch(
                 an all-static model when ``shape_body`` is provided.
             include_static_kinematic_pairs: Whether to include pairs where both shapes are immovable. Set to
                 ``False`` to filter static-static, static-kinematic, and kinematic-kinematic pairs.
+            shape_sweep: Optional world-space AABB translation over normalized time ``[0, 1]`` [m].
+            shape_sweep_projection_limit: Optional non-negative per-shape projection limit used by the SAP sort [m].
+                Limiting the projection is an approximation that may exclude pairs whose projected sweep exceeds
+                the limit. Candidate pairs retained by the sort are tested using the full relative sweep.
```
eric-heiden: Can you explain more what this means? "normalized time" - is that the dt interval? Would be good to link to the speculative contact section in the docs, and/or explain that these arguments are related to speculative contacts.

nvtw: The formulation is indeed bad - fixed in the latest commit.

nvtw: I tried to name it better. But maybe we should keep an eye on it during the API review. I'll come up with a follow up MR in case the API review before the 1.6 release comes up with better names.

eric-heiden (approved):
Thanks!

PR #3855, opened 2026-08-10
Implicit actuation

### LINKED ISSUE #3175: Linearly-implicit external actuators for IsaacLab / Newton
## Problem and Motivation

IsaacLab wants users to move toward explicit actuator models because they are the better sim-to-real path: external/user-provided actuator models, system-identified parameters, and a clear "state/targets in, torques out" contract shared by Newton, IsaacLab, Isaac Sim, and potentially other simulators.

The gap is stability. Existing implicit-actuator users are sensitive to the numerical robustness they had before, and purely explicit actuator integration can require smaller timesteps, extra substeps, or fragile tuning for high-stiffness/high-damping models.

We should not re-add the old implicit actuator path as the product direction. The target is to keep explicit/external actuator modeling while adding an implicit-style stability mechanism that is cleanly supported by Newton. Usecase might often be explicit neural actuator where Jacobians are almost trivial to obtain from.

## Proposal

Extend the Newton actuator interface so external actuator forces can be integrated linearly implicitly, without embedding actuator implementations into solver internals.

The goal is to provide implicit-PD-like stability for explicit/external actuator models while preserving the explicit-actuator sim-to-real story.

See https://www.jie-tan.net/project/spd.pdf for inspiration and https://reports.mmacklin.com/implicit-actuation-newton/ for Miles' idea how to do this.

## Proposed Shape

- Actuators remain external/decoupled models: state and targets in, torques/forces out.
- Add an optional Jacobian path to the actuator interface for derivatives of actuator force with respect to relevant state variables.
- Support analytic Jacobians where actuator models can provide them.
- Consider finite-difference Jacobians as a fallback where performance, graph capture, and determinism allow it.
- Keep the implementation separable from solver internals: the solver consumes a well-defined force/Jacobian contribution rather than owning arbitrary actuator logic.
- Provide an IsaacLab migration/adapter path for Lab-native or user-provided actuator models.

## Acceptance Criteria

- Sign off on the actuator/Jacobian API direction.
- The interface can represent at least PD-like actuator Jacobians and one externally provided/custom actuator use case.
- Unsupported solver/backend combinations fail early with meaningful errors or warnings.
- At least one IsaacLab-relevant benchmark demonstrates improved stability versus pure explicit integration at the same timestep/substep setup.
- Documentation explains when to use explicit-only vs. linearly-implicit actuator integration, and what old implicit-actuator users should do

## Scope

In scope:

- Newton actuator-interface design for optional Jacobian output.
- Linearly-implicit actuator-force integration path.
- IsaacLab adapter/migration requirements.
- One or more concrete IsaacLab validation environments.
- Initial performance/determinism expectations, especially for graphable RL workloads.

Out of scope:

- Reintroducing old IsaacLab implicit actuators as the main path.
- Broad system-identification or real-to-sim tuning tooling.
- Making every Newton solver support the feature immediately, unless stakeholders decide this is required.
- Coupling arbitrary actuator code directly into solver internals.

jvonmuralt:
## Description

Adds an experimental **implicit effort mode** to `newton.actuators.Actuator`.

By default an actuator evaluates its control law at the current state and holds
the effort constant over the step. At stiff gains and large timesteps that
overshoots or goes unstable. The implicit mode instead solves the control law
against the predicted end-of-step state (a Stable-PD style solve), which stays
stable at higher gains.

```python
oracle = ResponseOracle(model)
actuator.set_effort_mode_implicit(response=oracle)
```

### Design

The effort is computed **before** the solver step and accumulated into
`control.joint_f`, exactly as the explicit mode already does. Nothing about the
solver interface changes, so the mode works with every backend as-is.

For each articulation the actuator solves for the joint impulse `p` that is
consistent with the state it produces:

```
r(p)  = p - h*g(q(p), qd(p)) = 0
qd(p) = qd + A*p
q(p)  = q + h*qd(p)
```

`A` is the joint-space inverse mass matrix, supplied by the new `ResponseOracle`
and refreshed once per step. An impulse on one DOF changes the velocity of every
DOF in its articulation, so an articulation's actuated DOFs are solved together
as one coupled system, using the submatrix of `A` for the DOFs the actuator
drives. Newton's method, a few iterations per articulation, one GPU thread per
articulation, float64 internally.

Clamping is folded into the equation being solved rather than applied
afterwards, so a saturated effort is still consistent with the state it produces.

### Direction chosen, relative to #3175

Issue #3175 describes solver-consumed optional Jacobians: each solver would take
actuator derivatives and fold them into its own implicit step. This PR follows
the *Separable Implicit Actuation for Newton* report instead — the actuator
solves for its own impulse before the step and writes a plain force.

The pre-solver approach was selected because it requires no solver changes and
therefore works with every existing backend, including MuJoCo, without touching
solver internals or coordinating a cross-solver API change. The cost is that the
prediction accounts only for the actuator's own impulse: gravity, other applied
forces, other actuators on the same articulation, and joint drive applied
outside the actuator do not enter it. A solver-integrated formulation would
capture those, at the price of per-solver support. The two are not mutually
exclusive — this mode can be superseded later without an API break, since
`set_effort_mode_implicit()` is the only entry point.

### Public API

Added:

- `Actuator.set_effort_mode_implicit(response, options=None)` and
  `Actuator.set_effort_mode_explicit()`
- `Actuator.ImplicitOptions` — iteration count, convergence tolerances, warm start
- `newton.actuators.ResponseOracle` — supplies the inverse mass matrix, with
  `refresh(state)` (assembles it from the model) and
  `refresh_from_solve(solve_inverse, dof_map=...)` (reuses a solver's own
  inertia; MuJoCo is currently the only Newton solver that can provide it)
- Opt-in hooks on `Controller` (`evaluate_force`, `bind_params`,
  `prepare_implicit`) and on `Clamping` (`evaluate_clamp`, `param_width`,
  `bind_params`), documented under Customization

Deprecated:

- `ClampingDCMotor.corner_velocity` — now derived from the live parameters on
  access instead of being stored at construction, so it no longer goes stale
  when a parameter is retuned

Behaviour:

- Supported for all four controllers. The neural controllers enter the solve as
  a per-step linearization and need an ONNX checkpoint; a Torch checkpoint
  raises `NotImplementedError`
- Requires one position coordinate per DOF, so revolute, prismatic and D6 work;
  ball, free and distance joints raise `ValueError`
- Not differentiable: `set_effort_mode_implicit()` on an actuator built with
  `requires_grad=True` raises `NotImplementedError`
- CUDA-graph capturable, including the response refresh

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **New Features**
  * Added experimental implicit-effort actuator solving across coupled articulations.
  * Added response-data refresh from simulation state, inertia solvers, or mass matrices.
  * Added configurable solver tolerances, iteration limits, warm starts, and derivative handling.
  * Added implicit support for PD, PID, compatible neural controllers, and clamping behaviors.
  * Explicit effort mode remains the default and can be restored.
  * Added gradient-enabled ONNX checkpoint loading.

* **Documentation**
  * Added API and conceptual documentation for effort modes and response data.

* **Tests**
  * Added coverage for actuator solving, clamping, controller integrations, graph replay, and cloth regression scenarios.


docs/concepts/actuators.rst
```diff
@@ -222,6 +222,64 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
 into a computation graph that no longer exposes it.
 
+Effort Modes
+------------
+
+By default an actuator computes effort **explicitly**: the control law is
+evaluated at the current state and held constant over the step (zero-order
+hold). At stiff gains and large timesteps this can overshoot or go unstable.
+
+The **implicit** effort mode instead solves the control law against the
```
gyeomannvidia, line 237: I'm a big believer in short sentences!  I'd also drop terms like "much higher" for "higher" because we can't really quantify in general how much more stability we get. 

How about this?

The implicit effort mode instead solves the control law against the predicted end-of-step state (a Stable-PD style solve). A key advantage of this formulation over the explicit effort mode is that it stays stable at higher gains. The implicit effort mode necessarily requires the joint-space effective inverse mass. This is supplied by a ResponseOracle, which is refreshed once per step at the current pose:

docs/concepts/actuators.rst
```diff
@@ -222,6 +222,64 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
+   from newton.actuators import ResponseOracle
+
+   oracle = ResponseOracle(model)
+   actuator.set_effort_mode_implicit(response=oracle)
+
+   # Simulation loop
+   oracle.refresh(state)
+   actuator.step(sim_state, sim_control, state_a, state_b, dt=0.01)
+
+The solve couples the DOFs of each articulation, so the response is the full
```
gyeomannvidia: I didn't quite follow this sentence because I was expecting there to be a function called solve() in the python snippet. 

Can you provide more detail in the text about what is being solved?  Is this referring to actuator.step()?  

I had the same issue with "the response".  Is this the return of the ResponseOracle? or is it the outcome of actuator.step()? 

I think the point being made is that ResponseOracle is responsible for computing the per-articulation inverse mass rather than just a per-doc scalar. Maybe a connected point is that actuator.step() considers the mass matrix of the system and not just per-dof state. I may have completely misunderstood, though.

docs/concepts/actuators.rst
```diff
@@ -222,6 +222,64 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
+per-articulation inverse mass rather than a per-DOF scalar.
+
+:meth:`~newton.actuators.ResponseOracle.refresh` assembles its own mass matrix.
+It omits joint damping, contacts and constraint regularization, so the response
```
gyeomannvidia: I guess it also omits limits and friction.  Does it also omit armature?

docs/concepts/actuators.rst
```diff
@@ -222,6 +222,64 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
+is too large and the actuator under-drives the joint. Reuse the solver's own
```
gyeomannvidia: ...the actuator under-drives the joint...

Would it not be over-drive?  I'm thinking of an actuator that will take the joint pos from X to X+Delta but adding contact to the actuation model would have limited the outcome at X+Delta/2.  That would mean that the actuator has over-driven the joint due to the omission of contact. The same would be true for damping.   If we ignore damping, the actuator might take the joint pos from X to X+Delta but if the actuator model accounted for damping the outcome would be X + Delta - dt * damping * otherstuff.

I think my confusion is that I've not understood the source of "the response".  I'm imagining that the actuator computes a new state and that the "response" is that difference between new and old state.  With this definition, a response that is "too large" would over-drive the joints.

I think a really important thing here is to define "the response".

docs/concepts/actuators.rst
```diff
@@ -222,6 +222,64 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
+inertia instead. MuJoCo keeps it factorized, so
```
gyeomannvidia: I found this a little hard to follow.

I think the point being made is that ResponseOracle omits contact, damping, limits etc and, as a consequence, is only an approximation of the full system.  This can be improved by replacing the mass matrix computed by ResponseOracle (my guess is that this uses eval_mass_matrix() in Newton) with the mass matrix of the full system. This is possible with some solvers.  The MuJoCo solver is a good example so here's how to compute the mass matrix with MuJoCo. Other solvers may or may not support similar functionality.

Have I got this right?

docs/concepts/actuators.rst
```diff
@@ -222,6 +222,64 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
+.. code-block:: python
+
+   from newton.actuators import ResponseOracle
+
+   oracle = ResponseOracle(model)
+   actuator.set_effort_mode_implicit(response=oracle)
+
+   # Simulation loop
+   oracle.refresh(state)
+   actuator.step(sim_state, sim_control, state_a, state_b, dt=0.01)
```
gyeomannvidia, line 265: do we still need a newton.step() to complete the picture?

eg 

solver = SolverMuJoCo(sim_model)


...


solver.step(dt)

It looks as though actuator.step() replaces solver.step() but I doubt this is the case ( or I have completely misunderstood without yet looking at code).  If actuator.step() takes care of solver.step() internally then it would be great to mention that.

docs/concepts/actuators.rst
```diff
@@ -222,6 +222,64 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
+.. code-block:: python
+
+   def solve_inverse(x, y):
+       # x = M^-1 y, using the factorization the solver already built
+       mujoco_warp.solve_m(solver.mjw_model, solver.mjw_data, x, y)
+
+   # Simulation loop, in place of oracle.refresh(state)
+   oracle.refresh_from_solve(solve_inverse, dof_map=solver.mjc_dof_to_newton_dof)
+
+Both refresh paths only launch kernels, so the actuator, the solver step and the
```
gyeomannvidia: launch only kernels

docs/concepts/actuators.rst
```diff
@@ -222,6 +222,64 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
+response update can be captured in one CUDA graph.
+
+Call :meth:`~newton.actuators.Actuator.set_effort_mode_explicit` to switch back.
+
+Solver tuning lives in :class:`~newton.actuators.Actuator.ImplicitOptions`.
+
+All controllers support the implicit mode. The neural controllers enter the
```
gyeomannvidia: If there a fixed list of controller types  supported by Newton.  I think it is worth spelling out the list, unless it is really long.

All controllers support the implicit mode. These are X, Y, Z, W. The neural...

docs/concepts/actuators.rst
```diff
@@ -222,6 +222,64 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
+solve as a per-step linearization of the network, so they need an ONNX
```
gyeomannvidia: I am not familiar with ONNX. Is it possible to add a link?

docs/concepts/actuators.rst
jvonmuralt: much better, updated!

docs/concepts/actuators.rst
jvonmuralt: Yes, one definitely needs to step the solver too. There were full snippets already in this file, but I agree it is more readable if I extend this snippet too.

docs/concepts/actuators.rst
jvonmuralt: I linked the relevant section now

docs/concepts/actuators.rst
jvonmuralt: Added, thanks!

docs/concepts/actuators.rst
jvonmuralt: Yes, thats correct, I tried to reword it now a bit

docs/concepts/actuators.rst
jvonmuralt: Covers the armature and the rest is omitted, improved the doc to mention this

docs/concepts/actuators.rst
gyeomannvidia: One thing I think worth mentioning is that in some parameter regimes stability is improved by applying less force. We had this with the implicit spring in physx.   I remember just a few % of the expected stiffness*bias force being applied by the implicit spring.  For me the key point is the trade-off between stability at large timesteps/implicit drive and simulation integrity with small timesteps/explicit drive.

docs/concepts/actuators.rst
```diff
@@ -222,6 +225,89 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
+   oracle = ResponseOracle(model)
+   actuator.set_effort_mode_implicit(response=oracle)
+
+   # Simulation loop
+   oracle.refresh(sim_state)
+   sim_control.joint_f.zero_()
+   actuator.step(sim_state, sim_control, state_a, state_b, dt=0.01)
+   solver.step(sim_state, next_sim_state, sim_control, contacts, dt=0.01)
+
+In this mode :meth:`~newton.actuators.Actuator.step` solves for the joint
```
gyeomannvidia: The link is great. The remaining issue is that the doc builder just writes out "step" so it's hard to immediately know if step refers to the actuator or the solver.  Can anything be done?

docs/concepts/actuators.rst
```diff
@@ -222,6 +225,89 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
+impulse of each actuated DOF. An impulse on one DOF changes the velocity of
```
gyeomannvidia: I just want to check that we really do write out joint impulses.  From the code snippet, I expect that actuator.step() writes to sim_control.joint_f. I believe that joint_f is per-dof forces.

solves for the joint impulses (I think that should be forces)... this might be understood as evaluating joint impulses or applying them and updating joint velocities.  Can we replace "solves" with "evaluates"?  We have many functions in newton named evaluate_thing() so evaluate feels quite natural here.

docs/concepts/actuators.rst
```diff
@@ -222,6 +225,89 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
+
+The effective inverse mass, called *the response* below, is computed for a whole
+articulation. The actuator then reads only the entries for the DOFs it drives.
+:class:`~newton.actuators.ResponseOracle` is responsible for providing that
+matrix, and there are two ways to obtain it: compute it from scratch, or reuse
+what the solver already has.
+
+:meth:`~newton.actuators.ResponseOracle.refresh` builds the mass matrix itself,
+from :func:`~newton.eval_mass_matrix` and joint armature. Joint damping, joint
+limits, friction, contacts and constraint regularization are absent. All of
```
gyeomannvidia: I have a hunch that loop closure joints are also ignored. My guess is that oracle.refresh() ignores loop closure joints because eval_mass_matrix() also ignores them.  I'm far less sure of MuJoCo.solve_m.

docs/concepts/actuators.rst
```diff
@@ -222,6 +225,89 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
+
+In this mode :meth:`~newton.actuators.Actuator.step` solves for the joint
+impulse of each actuated DOF. An impulse on one DOF changes the velocity of
+every DOF in its articulation, so an articulation's actuated DOFs are solved
+together as one coupled system.
+
+The effective inverse mass, called *the response* below, is computed for a whole
+articulation. The actuator then reads only the entries for the DOFs it drives.
+:class:`~newton.actuators.ResponseOracle` is responsible for providing that
+matrix, and there are two ways to obtain it: compute it from scratch, or reuse
```
gyeomannvidia: It would make it even clearer if you added the corresponding functions eg

from scratch (actuator.refresh()), or reuse what the solver already has (actuator.refresh_from_solve()).  

if you can add links to the functions that would be even better.

docs/concepts/actuators.rst
```diff
@@ -222,6 +225,89 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
+what the solver already has.
+
+:meth:`~newton.actuators.ResponseOracle.refresh` builds the mass matrix itself,
+from :func:`~newton.eval_mass_matrix` and joint armature. Joint damping, joint
+limits, friction, contacts and constraint regularization are absent. All of
+those resist motion, so the response comes out too large. A larger response
+divides the control law harder and so yields a smaller effort, which errs
```
gyeomannvidia: small improvement

...so the response comes out larger than anticipated.

small improvement:

A larger response further divides the control law and...

another:

and so yields a smaller effort, which errs...

I read this and at first thought that stability was a positive outcome of the model approximations.  I don't think this is the point you're trying to make. How about this?

and so yields a smaller effort than would have been evaluated without the simplifications listed above.

docs/concepts/actuators.rst
```diff
@@ -222,6 +225,89 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
+toward stability.
+
+A solver that can apply its own inertia removes that approximation.
```
gyeomannvidia: How about this to begin the paragraph?

The approximations inherent in actuator.refresh() may be avoided when working with a solver that is able to evaluate the inverse mass matrix more directly. To this end, :meth:`~newton.actuators.ResponseOracle.refresh_from_solve`  takes a callable function that computes ``x = M^-1 y``.

To end the paragraph, I think we just need to quickly explain that M^-1 is computed with back substitution and that MuJoCo has this functionality alone.  Just from the API, the only thing the solver has to do is compute x=M^-1 * y so I don't feel that adding the detail about factorized/dense is necessary.  You know more about this so please disagree if you think this is an important detail eg is it necessary for refresh_from_solve() to work.

I added "callable function" above but I noted that you just had "callable". Is "callable" ok on its own? 

I just want to check again that the goal is the inverse mass matrix and not the inverse effective mass matrix (J * M^-1 * J^T).

docs/concepts/actuators.rst
```diff
@@ -222,6 +225,89 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
+       # x = M^-1 y, using the factorization the solver already built
+       mujoco_warp.solve_m(solver.mjw_model, solver.mjw_data, x, y)
+
+   # Simulation loop, in place of oracle.refresh(sim_state)
+   oracle.refresh_from_solve(solve_inverse, dof_map=solver.mjc_dof_to_newton_dof)
+
+Both refresh paths launch only kernels, so the actuator, the solver step and the
+response update can be captured in one CUDA graph.
+
+:meth:`~newton.actuators.Actuator.set_effort_mode_explicit` switches back, and
```
gyeomannvidia: it's kind of obvious but we can still spell it out:

switches back to explicit mode.

docs/concepts/actuators.rst
```diff
@@ -222,6 +225,89 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
+every DOF in its articulation, so an articulation's actuated DOFs are solved
+together as one coupled system.
+
+The effective inverse mass, called *the response* below, is computed for a whole
+articulation. The actuator then reads only the entries for the DOFs it drives.
+:class:`~newton.actuators.ResponseOracle` is responsible for providing that
+matrix, and there are two ways to obtain it: compute it from scratch, or reuse
+what the solver already has.
+
+:meth:`~newton.actuators.ResponseOracle.refresh` builds the mass matrix itself,
```
gyeomannvidia, line 281: I just want to check whether it builds the mass matrix or the inverse mass matrix or the inverse effective mass matrix ( J * M^-1 * J^T).

jvonmuralt: So ResponseOracle.refresh builds mass matrix by calling eval_mass_matrix and then inverts the result so the outcome of this method is inverse mass matrix.

docs/concepts/actuators.rst
jvonmuralt: You are right, I think this detail was unnecessarily added during an iterative discussion with Claude ...

newton/_src/actuators/effort_mode_implicit.py
```diff
@@ -0,0 +1,636 @@
...
+    max_iters: int = 4
+    """Maximum Newton iterations per articulation group."""
+
+    residual_tol: float = 1.0e-5
+    """Stop when the residual vector norm falls below this [N·s or N·m·s]."""
+
+    update_tol: float = 1.0e-5
+    """Stop when the impulse-update vector norm falls below this [N·s or N·m·s]."""
+
+    fd_epsilon: float = 1.0e-4
```
gyeomannvidia, line 67: Is it necessary to guard against fd_epsilon=0?  It looks as thought this would generate 1/0. It would be a quick job to add a warning and a test of the warning in test_validation_rejects_bad_options

newton/tests/test_actuators.py
```diff
@@ -190,6 +206,103 @@ def _write_dof_values(model, array, dof_indices, values):
...
+    builder.body_com[body] = wp.vec3(0.5, 0.0, 0.0)
+    joint = builder.add_joint_revolute(parent=-1, child=body, axis=newton.Axis.Z)
+    builder.add_articulation([joint])
+    return builder.finalize(device=device)
+
+
+def _two_link_builder(armature: float = 0.0):
+    """Builder for a two-link revolute chain — one articulation, two coupled DOFs."""
+    builder = newton.ModelBuilder(gravity=(0.0, 0.0, 0.0))
+    base = builder.add_link(mass=1.5)
```
gyeomannvidia: I noticed that you gave the two masses carefully chosen values. The actual masses will be greater than that because Newton adds on volume*density.  I don't think the chosen masses play an explicit role in any test so I don't think this really matters but the intention would be clearer if the density was set to 0 for add_shape_box.  I can imagine someone later looking at a test and wondering why the mass is not0.8. This is a really small thing.

newton/tests/test_actuators.py
```diff
@@ -190,6 +206,103 @@ def _write_dof_values(model, array, dof_indices, values):
...
+    builder.body_com[tip] = wp.vec3(0.25, 0.0, 0.0)
+    j0 = builder.add_joint_revolute(parent=-1, child=base, axis=newton.Axis.Z, armature=armature)
+    j1 = builder.add_joint_revolute(
+        parent=base,
+        child=tip,
+        axis=newton.Axis.Z,
+        parent_xform=wp.transform(wp.vec3(0.6, 0.0, 0.0), wp.quat_identity()),
+        armature=armature,
+    )
+    builder.add_articulation([j0, j1])
```
gyeomannvidia, line 260: I ran the tests and verified that this creates a dof_map that is 1:1 ie joint N in newton is joint N in mujoco.  The testing would be improved with a non trivial dof_map. Can you think about how that might be achieved with a quick change to _two_link_builder and/or _build_pendulum?

I wonder if we could try the following:

1.  add a dummy body before adding the articulation. . 
2. disable collision on every body and link added or have no shapes on the dummy body

I think that might lead to a non trivial mapping without upsetting the physics of the scene. You will have better ideas, I'm sure.

newton/tests/test_actuators.py
```diff
@@ -190,6 +206,103 @@ def _write_dof_values(model, array, dof_indices, values):
...
+    return builder
+
+
+def _build_two_link(device):
+    """Two-link revolute chain — one articulation, two inertially coupled DOFs."""
```
gyeomannvidia: How hard would it be to refactor here to allow multiple worlds to be created?  I guess it is easy to change _build_two_link but the hard part is to refactor the tests so that they test the results across multiple worlds.

It is more than enough to just verify that world N has the same outcome as world 1.   So all testing for world1 remains as is and a helper function can then test that all worlds match world 1. This does not have to be for all tests. Can you identity a few that would make most sense with N worlds?  2 worlds is already a lot.

newton/tests/test_actuators.py
```diff
@@ -190,6 +206,103 @@ def _write_dof_values(model, array, dof_indices, values):
...
+    """Build an implicit PD Actuator over all DOFs, with an optional max-effort clamp.
+
+    Returns the actuator together with the response oracle driving its solve.
+    """
+    n = model.joint_dof_count
+    clamping = None
+    if max_effort is not None:
+        clamping = [ClampingMaxEffort(max_effort=wp.array(max_effort, dtype=float, device=device))]
+    oracle = kwargs.setdefault("response", ResponseOracle(model))
+    actuator = Actuator(
```
gyeomannvidia, line 390: The documentation states that the actuator selects the joints that will be actuated.  In test, all joints are driven by the actuator.  How hard would it be to create a separate test where only 1 joint was driven?

This test limitation likely predates the work on implicit actuators.  If it's a big job please add it as a follow-up task .

newton/tests/test_actuators.py
```diff
@@ -190,6 +206,103 @@ def _write_dof_values(model, array, dof_indices, values):
...
+
+def _expected_implicit_pd(model, state, kp, kd, target, dt):
+    """Closed-form single-DOF implicit PD effort from the pose held by *state*."""
+    alpha = _response_at_state(model, state)[0, 0]
+    q = float(state.joint_q.numpy()[0])
+    qd = float(state.joint_qd.numpy()[0])
+    return (kp * (target - q - dt * qd) - kd * qd) / (1.0 + alpha * dt * kd + alpha * dt * dt * kp)
+
+
+def _make_implicit_actuator(model, device, kp, kd, max_effort=None, **kwargs):
```
gyeomannvidia: Could you annotate all fn args?  eg kp is a np.array[float32] or np.list or whatever.  

If  possible, could you do that for every function in the test file.  I've found Claude to be really good at this. Purely for my benefit because my peanut brain is having a hard time understanding if I am looking at a float or an array of floats etc.

newton/tests/test_actuators.py
```diff
@@ -190,6 +206,103 @@ def _write_dof_values(model, array, dof_indices, values):
...
+    """Build an implicit PD Actuator over all DOFs, with an optional max-effort clamp.
+
+    Returns the actuator together with the response oracle driving its solve.
+    """
+    n = model.joint_dof_count
+    clamping = None
+    if max_effort is not None:
+        clamping = [ClampingMaxEffort(max_effort=wp.array(max_effort, dtype=float, device=device))]
```
gyeomannvidia, line 388: Does the feature support an array with different types of Clamping applied to each dof? Would it be easy to write a test for this?  The simplest test would be a 2nd Clamping that does effectively nothing. So we would test with 

     Clamping = [ClampingMaxEffort, ClampingSomethingElseButConfiguredToDoNothing]

and then 

     Clamping = [ClampingSomethingElseButConfiguredToDoNothing, ClampingMaxEffort]

On a similar theme, what would happen if I created two Actuator instances working on the same dof list. The 1st Actuator applies a max effort A to dof 0, and inf for dof 1.  The 2nd Actuator applies a max effort of inf to dof 0 and a max effort A to dof 1.  Would this be identical to a single Actuator with max_effort=[A,A]? Can this be tested?

docs/concepts/actuators.rst
```diff
@@ -222,6 +225,102 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
+matrix, and there are two ways to obtain it: compute it from scratch
+(:meth:`ResponseOracle.refresh <newton.actuators.ResponseOracle.refresh>`), or
+reuse what the solver already has (:meth:`ResponseOracle.refresh_from_solve
+<newton.actuators.ResponseOracle.refresh_from_solve>`).
+
+:meth:`~newton.actuators.ResponseOracle.refresh` builds the mass matrix itself,
+from :func:`~newton.eval_mass_matrix` and joint armature. This comes with
+approximations. First, joint damping, joint limits, friction, contacts and
+constraint regularization are absent. All of those resist motion, so the
+response comes out larger than anticipated. A larger response further divides
```
gyeomannvidia, line 285: I've been digging into the code a bit further and I think the list of caveats is incomplete but also more complicated than I first thought. 

The predicted state qd += A*p ignores gravity, applied forces, other actuators on the same articulation.

Right at top where you mention "predicted end-of-step state" could you add the caveats about gravity, other forces, other actuators?  I'm guessing we also ignore joint drive that is added without using the Actuator class.

I think it is also worth adding these caveats to the relevant docstrings in the public API.  Two places spring to mind 1) ResponseOracle.refresh and 2) the docstring  `qd(p) = qd + A p``

newton/_src/actuators/response_oracle.py
```diff
@@ -0,0 +1,346 @@
...
+    then backward substitution) and store x as column c of the inverse.
+    """
+    a = wp.tid()
+    n = art_dof_count[a]
+
+    for j in range(n):
+        s = H[a, j, j]
+        for k in range(j):
+            s -= L[a, j, k] * L[a, j, k]
+        s = wp.max(s, 1.0e-9 * wp.max(H[a, j, j], 1.0e-9))
```
gyeomannvidia: I hope I've got this right.  We're working in float32 here but using 1e-9 as a threshold. This is a very small number far below float32 epsilon.  

The problem is that s can fall to 1e-18, then we take d = sqrt(s) then we compute 1/d.  

I'm actually not sure what to do here. I'm guessing that H[j, j] < 1e-9 signals the degenerate case so this is tricky.

newton/_src/actuators/effort_mode_implicit.py
```diff
@@ -0,0 +1,636 @@
...
+
+        The controller law at the final predicted state is written to
+        *computed_forces*. Clamps are enforced inside the solve against that
+        state, and the solved effort is written to *applied_forces*.
+        """
+        if dt is None:
+            raise ValueError("Implicit actuation requires dt")
+        if dt <= 0.0:
+            raise ValueError(f"Implicit actuation requires dt > 0, got {dt}")
+        if applied_forces is None:
```
gyeomannvidia: Claude caught this one.
applied_forces  is never none now so we don't need the guard.

newton/_src/actuators/controllers/controller_pid.py
```diff
@@ -161,6 +213,65 @@ def is_stateful(self) -> bool:
...
+            requires_grad=self.kp.requires_grad,
+        )
+        pack[:, 0].assign(self.kp)
+        pack[:, 1].assign(self.kd)
+        self.kp = pack[:, 0]
+        self.kd = pack[:, 1]
+        self._param_pack = pack
+        return pack
+
+    def prepare_implicit(
```
gyeomannvidia, line 235: Missing type annotations. 
The same is true for the other prepare_implicit functions added to Newton.

newton/_src/actuators/actuator.py
```diff
@@ -182,13 +185,57 @@ def __init__(
         for clamp in self.clamping:
             clamp.finalize(self.device, self.num_actuators)
 
+        self._effort_mode = _EffortModeExplicit(controller, self.clamping, self.device)
+
+    ImplicitOptions = ImplicitOptions
```
gyeomannvidia, line 192: I'm afraid I don't understand this line of code. Claude is recommending a comment. Here's what it says:

"In a language with explicit class keywords for nesting, Actuator.State and Actuator.ImplicitOptions would look structurally different from a field. In Python, ImplicitOptions = ImplicitOptions is syntactically indistinguishable from declaring a field with a default — which is what makes the comment worth adding."

newton/_src/actuators/effort_mode_implicit.py
```diff
@@ -0,0 +1,636 @@
...
+        self._num_actuators = num_actuators
+        self._device = device
+        if not isinstance(response, ResponseOracle):
+            raise ValueError(
+                "Implicit actuation requires response to be a ResponseOracle; "
+                "build one with newton.actuators.ResponseOracle(model)."
+            )
+        self._response = response
+        self._controller = controller
+        # Only controllers with their own prepare_implicit (the neural linearizers)
```
gyeomannvidia: Minor.

I would never have found this without asking Claude to check all code comments.

Only controllers with their own prepare_implicit (the neural linearizers)...

The set of controllers that have their own prepare_implicit and set of controllers that are neural linearizers are not the same. The comment isn't quite right, as a consequence.

Claude also found this:

...so PD/PID skip both launches.

but only PD  skips both.

docs/concepts/actuators.rst
gyeomannvidia: Room for one small improvement.  In the docstring for set_effort_mode_implicit(), would it be possible to refer readers to the docstring at the top of effort_mode_implicit.py?

jvonmuralt: I think I cannot link the effort_mode_implicit docstring. 
I've linked the Effort Modes section of the concepts page instead. Let me know if you'd like the content of this docstring from effort_mode_implicit.py be moved/replicated in the concepts page too.

gyeomannvidia: That also works for me.  Anything that takes readers to the full list of caveats is great.

gyeomannvidia (dismissed):
Good job!

jcarius-nv (commented):
Thank you, Julia, for this substantial contribution, and thank you, Gordon, for the thorough review.

I have four small follow-ups before approval:

1. API compatibility: ClampingDCMotor.corner_velocity is removed. Because it is a public attribute and Newton requires deprecation before removal, please preserve and deprecate it first, or otherwise retain compatibility.

2. PR description: Please replace the empty template with a concise description of the selected design, the public API changes, and the test plan.

3. Specification direction: issue #3175 describes solver-consumed optional Jacobians, while the linked Separable Implicit Actuation for Newton report describes the pre-solver joint_f approach implemented here. This can probably be addressed with a good PR description that records which direction was selected and why.

4. Custom actuators: Please add a little documentation showing how an externally provided or custom Controller opts into implicit actuation through evaluate_force, bind_params, and prepare_implicit, including how unsupported configurations behave.

After these small follow-ups, I think the PR can be approved.

jcarius-nv (commented):
Thanks for restoring and deprecating ClampingDCMotor.corner_velocity. One compatibility detail remains: it was previously a persistent, writable wp.array, whereas the new property returns a freshly derived array on each access. As a result, assigning clamp.corner_velocity raises AttributeError, and clamp.corner_velocity.assign(...) only mutates a temporary and no longer affects clamping behavior. Was this change in write semantics intentional? If so, I am comfortable treating it as an explicit compatibility decision and approving the PR.

jvonmuralt:
Yes, intentional. It's a derived value,  velocity_limit * (1 + max_motor_effort / saturation_effort). Caching it at construction meant that retuning any of those three left the clamp using the old motor's envelope.

Actually it shouldn't have been public at all. Writing to it never really made sense, since it just contradicts the parameters it comes from. The property is kept so existing reads keep working, and they now return the right value after a retune

jcarius-nv (approved):

PR #3858, opened 2026-08-10
[Kamino] check massless runtime update

rubengrandia:
## Description

Follow-up on https://github.com/newton-physics/newton/pull/3840. Adds runtime model change validation for massless bodies: 
- converting a normal body into a massless body is rejected. 
- converting a massless body to something with mass (and afterwards converting it back) is allowed.

With this restriction we do not have to do the full topology check again at runtime. I also excluded checking eigenvalues and only check for full zero Inertia matrix. It feels like setting singular inertias is a user error and not common enough to warrent an eigenvalue check.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Expanded model-change validation to cover structural and inertial updates, including bodies becoming massless.
  - Added clear detection and reporting for unsupported structural changes.

- **Bug Fixes**
  - The solver now requires recreation when a body becomes massless or its inverse inertia is fully zeroed.

- **Tests**
  - Added regression coverage for massless-body and zero-inertia updates.

### DIFF
--- a/newton/_src/solvers/kamino/_src/core/conversions.py
+++ b/newton/_src/solvers/kamino/_src/core/conversions.py
@@ -40,7 +40,7 @@ if TYPE_CHECKING:
 ###
 
 __all__ = [
-    "JointUpdateViolation",
+    "StructuralUpdateViolation",
     "convert_geometries",
     "convert_joints",
     "convert_model_joint_actuation",
@@ -49,7 +49,7 @@ __all__ = [
     "convert_rigid_bodies",
     "convert_target_coords_to_target_dofs",
     "convert_target_dofs_to_target_coords",
-    "validate_model_joint_updates",
+    "validate_model_structural_updates",
 ]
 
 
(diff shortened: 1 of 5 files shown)


newton/_src/solvers/kamino/_src/core/conversions.py
```diff
@@ -318,10 +319,42 @@ def validate_joint_axes_kernel(
...
     if not valid:
-        wp.atomic_min(violations, JointUpdateViolation.NONORTHONORMAL_AXES, joint)
+        wp.atomic_min(violations, StructuralUpdateViolation.NONORTHONORMAL_AXES, joint)
+
+
+@wp.func
+def has_zero_inverse_inertia(inv_inertia: wp.mat33f) -> bool:
+    """Return whether every element of an inverse inertia matrix is zero."""
+    return (
+        inv_inertia[0, 0] == 0.0
```
Guirec-Maloisel, line 331: It seems that here we're back to checking whether the inverse inertia is full zero, as opposed to singular. Should we also check for singular inertia in this PR, or is there a reason not to?

rubengrandia: yes, I mentioned this in the description. It feels like that is too expensive to do at runtime. It think for updating the model we can make the user responsible for ensuring this themselves.

Guirec-Maloisel: Knowing that it's a 3x3 matrix, I'd have expected that this can be done relatively cheaply. But I think I'm fine merging as it is, this starts to really be a corner case of a corner case.

Guirec-Maloisel (dismissed):

Guirec-Maloisel (approved):
Re-approving after the rebase

PR #3859, opened 2026-08-10
Fix NotifyDRLegs ASV sampling

rubengrandia:
## Description

Update the NotifyDRLegs ASV configuration for asv_runner 0.3.0 setup semantics. One timed call is now collected per sample, removing repeated model and solver construction from explicitly batched calls. The suite uses five repeats, no timed warmups, and a 600-second timeout.

## Checklist

- [x] For user-facing changes, a fragment has been added by following the [changelog fragment instructions](https://github.com/newton-physics/newton/blob/main/changelog/README.md) (not user-facing)

## Test plan

- `uvx pre-commit run -a`
- `uv run --extra dev asv check`
- `uv run --extra dev asv run --launch-method spawn -b 'NotifyDRLegs'`

On an RTX 5090, the seven-sample validation run completed in 88 seconds. Per-method spreads were 1–5.5%; no repeat increase is proposed initially. The PR ASV gate does not select NotifyDRLegs, so this needs a new nightly baseline rather than a threshold change.

## Summary by CodeRabbit

* **Performance**
  * Updated benchmark settings to complete faster by reducing repetitions, disabling warmup, and shortening the timeout.
  * Maintained a single benchmark round for consistent results.

### DIFF
--- a/asv/benchmarks/simulation/bench_kamino.py
+++ b/asv/benchmarks/simulation/bench_kamino.py
@@ -158,10 +158,11 @@ class KpiDRLegs(_KpiBenchmark):
 class NotifyDRLegs:
     """Benchmark Kamino model notifications for 2048 DR Legs worlds."""
 
-    number = 10
-    repeat = 7
+    number = 1
+    repeat = 5
     rounds = 1
-    timeout = 3600
+    warmup_time = 0
+    timeout = 600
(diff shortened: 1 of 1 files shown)


jcarius-nv (approved):
Thank you!

PR #3864, opened 2026-08-10
Decouple ASV tests from installed package

jcarius-nv:
## Description

Move benchmark-specific tests out of the installed `newton.tests` package and
run them directly from `asv/tests` in source CI. This prevents wheel test
discovery from importing repository-only ASV modules that are intentionally
not packaged.

Separate the ASV `track_*` adapters from metric computation, collection, and
simulation-state validation. Keep the `newton.utils.run_benchmark` cache
lifecycle contract test in `newton.tests` because it covers shipped package
functionality.

## Checklist

      [changelog fragment instructions](https://github.com/newton-physics/newton/blob/main/changelog/README.md) (not user-facing)

## Test plan

```text
uv run --extra dev -m unittest discover -s asv/tests -v
python -m unittest newton.tests.test_run_benchmark -v
asv check
uv build
```

The rebuilt wheel was installed into an isolated environment and exercised
through `python -m newton.tests`; full discovery completed without the former
benchmark-module import errors, and all 45 selected tests passed.

## Bug fix

**Steps to reproduce:**

1. Build the wheel with `uv build`.
2. Install it into an environment with Newton's test dependencies.
3. Run `python -m newton.tests`.
4. Observe `ModuleNotFoundError: No module named 'benchmark_metrics'` while
   discovering `test_benchmark_metrics` and `test_benchmark_simulation`.

**Minimal reproduction:**

```text
python -m newton.tests -k benchmark_metrics
```

## Summary by CodeRabbit

* **Testing**
  * Added automated benchmark execution to continuous integration, including test reports, coverage, and downloadable artifacts.
  * Added coverage for benchmark setup-cache behavior and lifecycle handling.
  * Updated benchmark tests for the current repository layout.
* **Benchmarking**
  * Added simulation metrics for timing, throughput, frame times, GPU memory, timestep, and substeps.
  * Metrics are skipped automatically when CUDA is unavailable.
* **Maintenance**
  * Improved installed test discovery and benchmark component organization.
  * Added licensing headers to benchmark test support files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,6 +149,8 @@ jobs:
           --junit-report-xml rspec.xml
           --coverage
           --coverage-xml coverage.xml
+      - name: Run ASV benchmark tests
+        run: uv run --no-sync -m unittest discover -s asv/tests
       - name: Test Summary
         uses: test-summary/action@37b508cfee6d4d080eedd00b5bb240a6a784a6a5  # v2.6
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,7 +149,7 @@ jobs:
           --junit-report-xml rspec.xml
           --coverage
           --coverage-xml coverage.xml
-      - name: Run ASV benchmark tests
+      - name: Run ASV harness unit tests
         run: >
           uv run --no-sync -m newton.tests
           --strict-warnings
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -150,11 +150,18 @@ jobs:
           --coverage
           --coverage-xml coverage.xml
       - name: Run ASV benchmark tests
-        run: uv run --no-sync -m unittest discover -s asv/tests
+        run: >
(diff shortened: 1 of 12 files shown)


.github/workflows/ci.yml
```diff
@@ -149,18 +149,29 @@ jobs:
           --junit-report-xml rspec.xml
           --coverage
           --coverage-xml coverage.xml
+      - name: Run ASV benchmark tests
```
eric-heiden: Could we rename this step to `Run ASV harness unit tests`? These are unit/contract tests for the ASV harness rather than the performance benchmarks themselves, and the current name could make it sound like CI is executing the benchmarks.

eric-heiden (commented):
One small naming suggestion for the new CI step.

.github/workflows/ci.yml
jcarius-nv: Addressed in 7cfbc48b: renamed the step to Run ASV harness unit tests.

eric-heiden (approved):
Thanks!

PR #3868, opened 2026-08-10
Fix distance-dependent drift in the free-joint analytic IK Jacobian

### LINKED ISSUE #3866: Floating-base free-joint analytic IK Jacobian drifts with distance from world origin (regression from #2539)
## Summary

In `newton.ik.IKJacobianType.ANALYTIC` mode, the inverse-kinematics Jacobian for a floating base (a root `JointType.FREE`, and equivalently `JointType.DISTANCE`) is wrong once the body is translated away from the world origin. The error in the analytic motion subspace grows linearly with the body's distance from the origin, so floating-base ANALYTIC IK drifts the farther the character/robot travels from the origin. In-place solves at the origin are unaffected, and `AUTODIFF` mode is correct throughout.

On a traveling motion clip we observed roughly **150 mm** of end-effector drift in ANALYTIC mode that does not appear in AUTODIFF mode.

## Affected versions

- Introduced in **1.4.0** by #2539 (commit `250fae60`).
- Still present in **1.5.0rc2** and on `main` (`1.6.0.dev0`).
- `AUTODIFF` mode is correct in all versions.

## Root cause

#2539 rewrote the FREE/DISTANCE analytic motion subspace (`write_free_distance_motion_subspace` in `newton/_src/sim/articulation.py`) so that the angular DOFs' rotation-to-translation coupling is taken about the **child body COM expressed in world coordinates**:

```python
x_child_com_world = wp.transform_point(X_wc, body_com_child)
...
joint_S_s[qd_start + 3] = wp.spatial_vector(-wp.cross(axis_world_x, x_child_com_world), axis_world_x)
```

For a floating base whose parent anchor is fixed, the origin-invariant coupling must be taken about the **fixed parent-joint anchor** (the translation of `X_pa_world`), not the traveling child COM. Because the child COM moves with the body, the `-cross(axis_world, x_child_com_world)` terms pick up a spurious contribution proportional to the body's world position, so the Jacobian's rotation-to-translation block is wrong by an amount that scales with distance from the origin. The public world-frame twist convention introduced by #2539 is otherwise correct and should be preserved.

## AUTODIFF-confirmed diagnosis

Comparing the ANALYTIC and AUTODIFF IK Jacobians for a single root FREE (or DISTANCE) body at increasing world translations shows exact agreement at the origin and a divergence that grows with translation. AUTODIFF matches finite differences at every translation; only ANALYTIC drifts. Equivalently, the analytic motion subspace `joint_S_s` for the root joint is not invariant under pure translation of the base, when it should be.

## Steps to reproduce

1. Build a model with a single root `FREE` joint (identity parent/child anchors).
2. Set the joint's translation to increasing world positions (0, 1, 5, 20 m) with an identity orientation.
3. Build two `IKSolver`s over an `IKObjectivePosition` on that body: one `AUTODIFF`, one `ANALYTIC`.
4. Compute each solver's Jacobian and compare. The max abs difference is ~0 at translation 0 and grows with translation under current `main`/1.5.0rc2.

## Minimal reproduction

```python
import numpy as np
import warp as wp
import newton
import newton.ik as ik

device = "cuda:0"

with wp.ScopedDevice(device):
    joint = builder.add_joint_free(
        parent=-1,
        child=body,
        parent_xform=wp.transform_identity(),
        child_xform=wp.transform_identity(),
    )
    model = builder.finalize(device=device, requires_grad=True)

    translations = (0.0, 1.0, 5.0, 20.0)
    joint_q = np.zeros((len(translations), model.joint_coord_count), dtype=np.float32)
    joint_q[:, 0] = translations   # translate the base along +x
    joint_q[:, 6] = 1.0            # identity quaternion (w = 1)

    targets = wp.zeros(len(translations), dtype=wp.vec3, device=device)
    objectives = [ik.IKObjectivePosition(body, wp.vec3(), targets)]

    solver_auto = ik.IKSolver(model, len(translations), objectives,
                              jacobian_mode=ik.IKJacobianType.AUTODIFF)
    solver_ana = ik.IKSolver(model, len(translations), objectives,
                             jacobian_mode=ik.IKJacobianType.ANALYTIC)

    q_auto = wp.array(joint_q, device=device, requires_grad=True)
    q_ana = wp.array(joint_q, device=device)
    solver_auto._impl._compute_residuals(q_auto)
    solver_ana._impl._compute_residuals(q_ana)
    J_auto = solver_auto._impl._jacobian_at(solver_auto._impl._ctx_solver(q_auto)).numpy()
    J_ana = solver_ana._impl._jacobian_at(solver_ana._impl._ctx_solver(q_ana)).numpy()

    for i, t in enumerate(translations):
        print(f"translation {t:>5} m: max|J_ana - J_auto| = {np.abs(J_ana[i] - J_auto[i]).max():.6e}")
```

Under current `main`/1.5.0rc2 the analytic-vs-autodiff error is ~0 at translation 0 and grows with translation (nonzero at 1/5/20 m). It should be 0 at every translation.

## Expected behavior

The ANALYTIC free/distance-joint IK Jacobian should equal the AUTODIFF Jacobian (and finite differences) regardless of the base's world position. The analytic motion subspace for a floating base must be origin-invariant: the angular DOFs' translation coupling should reference the fixed parent-joint anchor rather than the moving child COM.

dylanturpin:
## Description

Fixes distance-dependent drift in the floating-base free-joint analytic IK Jacobian, a regression from #2539.

#2539 (commit `250fae60`) rewrote the FREE/DISTANCE analytic motion subspace (`write_free_distance_motion_subspace` in `newton/_src/sim/articulation.py`) so the angular DOFs' rotation-to-translation coupling is taken about the **child body COM in world coordinates**. Because the child COM travels with the body, the `-cross(axis_world, x_child_com_world)` terms gain a spurious contribution proportional to the base's world position, so the analytic Jacobian's rotation-to-translation block is wrong by an amount that scales with distance from the origin. Floating-base `ANALYTIC`-mode IK therefore drifts the farther the body is from the origin (~150 mm observed on a traveling clip); in-place solves at the origin are fine, and `AUTODIFF` mode is correct throughout.

This PR references the **fixed parent-joint anchor** (`wp.transform_get_translation(X_pa_world)`) instead of the traveling child COM, restoring origin-invariant coupling. It preserves the public world-frame twist convention introduced by #2539, so the analytic motion subspace now matches AUTODIFF at any world position. The change is ~14 lines in `articulation.py` plus direct analytic-vs-AUTODIFF regression coverage in `test_ik.py`.

Closes #3866

> **Regression / 1.5 release candidate:** this bug is present in **1.4.0** and still in **1.5.0rc2**, so this is a candidate for the **1.5 release**. Could a maintainer consider cherry-picking it onto `release-1.5` as part of the backport?

## Checklist

## Test plan

New tests add direct analytic-vs-AUTODIFF IK Jacobian coverage for a translated floating body (both `FREE` and `DISTANCE`) at 0 / 1 / 5 / 20 m, and assert the root motion subspace is translation-invariant.

```
uv run --extra dev -m newton.tests -k ik
uv run --extra dev -m newton.tests -k featherstone
```

Verified on an NVIDIA RTX PRO 6000 (Blackwell) GPU:

- `-k ik`: **160 passed** (3 skipped), including the new `test_translated_free_jacobian_compare` and `test_translated_distance_jacobian_compare` on both CPU and CUDA.
- `-k featherstone`: **274 passed**.
- New analytic-vs-AUTODIFF tests: max abs Jacobian error is **exactly 0.0** at translations 0 / 1 / 5 / 20 m (was growing with distance before the fix).
- A warm-started traveling sweep that drifted ~**150 mm** in ANALYTIC mode now tracks AUTODIFF to **~2e-6 m**.

## Bug fix

**Steps to reproduce:**

1. Build a model with a single root `FREE` joint (identity parent/child anchors).
2. Set the joint translation to increasing world positions (0, 1, 5, 20 m) with an identity orientation.
3. Build two `IKSolver`s over an `IKObjectivePosition` on that body: one `AUTODIFF`, one `ANALYTIC`.
4. Compute and compare each solver's Jacobian. Without this PR, the max abs difference is ~0 at translation 0 and grows with translation.

**Minimal reproduction:**

```python
import numpy as np
import warp as wp
import newton
import newton.ik as ik

device = "cuda:0"

with wp.ScopedDevice(device):
    joint = builder.add_joint_free(
        parent=-1,
        child=body,
        parent_xform=wp.transform_identity(),
        child_xform=wp.transform_identity(),
    )
    model = builder.finalize(device=device, requires_grad=True)

    translations = (0.0, 1.0, 5.0, 20.0)
    joint_q = np.zeros((len(translations), model.joint_coord_count), dtype=np.float32)
    joint_q[:, 0] = translations   # translate the base along +x
    joint_q[:, 6] = 1.0            # identity quaternion (w = 1)

    targets = wp.zeros(len(translations), dtype=wp.vec3, device=device)
    objectives = [ik.IKObjectivePosition(body, wp.vec3(), targets)]

    solver_auto = ik.IKSolver(model, len(translations), objectives,
                              jacobian_mode=ik.IKJacobianType.AUTODIFF)
    solver_ana = ik.IKSolver(model, len(translations), objectives,
                             jacobian_mode=ik.IKJacobianType.ANALYTIC)

    q_auto = wp.array(joint_q, device=device, requires_grad=True)
    q_ana = wp.array(joint_q, device=device)
    solver_auto._impl._compute_residuals(q_auto)
    solver_ana._impl._compute_residuals(q_ana)
    J_auto = solver_auto._impl._jacobian_at(solver_auto._impl._ctx_solver(q_auto)).numpy()
    J_ana = solver_ana._impl._jacobian_at(solver_ana._impl._ctx_solver(q_ana)).numpy()

    for i, t in enumerate(translations):
        print(f"translation {t:>5} m: max|J_ana - J_auto| = {np.abs(J_ana[i] - J_auto[i]).max():.6e}")
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected analytic inverse-kinematics Jacobians for floating-base FREE and DISTANCE joints.
  * Ensured Jacobian results remain consistent regardless of world-space translation.
  * Aligned motion-subspace calculations with the parent-anchor coordinate convention.

* **Tests**
  * Added coverage comparing analytic and autodiff Jacobians across multiple translations and supported devices.

* **Documentation**
  * Added a changelog entry describing the Jacobian drift fix.

### DIFF
--- a/newton/_src/sim/articulation.py
+++ b/newton/_src/sim/articulation.py
@@ -934,7 +934,6 @@ def eval_ik(
 @wp.func
 def write_free_distance_motion_subspace(
     X_pa_world: wp.transform,
-    x_child_com_world: wp.vec3,
     qd_start: int,
     # outputs
     joint_S_s: wp.array[wp.spatial_vector],
@@ -944,7 +943,7 @@ def write_free_distance_motion_subspace(
     Used by both the Featherstone inverse-dynamics path (``jcalc_motion``) and
     the IK/Jacobian path (``jcalc_motion_subspace``) so they agree on the exact
     column layout. Linear DOFs act at the child body's COM; angular DOFs are
-    world-aligned axes expressed through ``X_pa_world``.
+    world-aligned axes referenced to the fixed parent anchor.
 
     Args:
         X_pa_world: Parent-anchor world transform (``X_wp * joint_X_p``) used
@@ -952,7 +951,6 @@ def write_free_distance_motion_subspace(
             This is *not* the classical Featherstone ``X_sc`` (spatial-to-
             child); Newton's FREE/DISTANCE joint coordinates live in the
             parent-anchor basis.
-        x_child_com_world: World-space position of the child body's COM.
         qd_start: Starting velocity-DOF index for this joint.
         joint_S_s: Output spatial-vector subspace array; six slots starting at
             ``qd_start`` are overwritten.
@@ -960,13 +958,14 @@ def write_free_distance_motion_subspace(
     axis_world_x = wp.transform_vector(X_pa_world, wp.vec3(1.0, 0.0, 0.0))
     axis_world_y = wp.transform_vector(X_pa_world, wp.vec3(0.0, 1.0, 0.0))
     axis_world_z = wp.transform_vector(X_pa_world, wp.vec3(0.0, 0.0, 1.0))
+    x_anchor_world = wp.transform_get_translation(X_pa_world)
 
     joint_S_s[qd_start + 0] = wp.spatial_vector(axis_world_x, wp.vec3())
     joint_S_s[qd_start + 1] = wp.spatial_vector(axis_world_y, wp.vec3())
     joint_S_s[qd_start + 2] = wp.spatial_vector(axis_world_z, wp.vec3())
-    joint_S_s[qd_start + 3] = wp.spatial_vector(-wp.cross(axis_world_x, x_child_com_world), axis_world_x)
-    joint_S_s[qd_start + 4] = wp.spatial_vector(-wp.cross(axis_world_y, x_child_com_world), axis_world_y)
-    joint_S_s[qd_start + 5] = wp.spatial_vector(-wp.cross(axis_world_z, x_child_com_world), axis_world_z)
+    joint_S_s[qd_start + 3] = wp.spatial_vector(-wp.cross(axis_world_x, x_anchor_world), axis_world_x)
+    joint_S_s[qd_start + 4] = wp.spatial_vector(-wp.cross(axis_world_y, x_anchor_world), axis_world_y)
+    joint_S_s[qd_start + 5] = wp.spatial_vector(-wp.cross(axis_world_z, x_anchor_world), axis_world_z)
 
 
 @wp.func
@@ -1066,8 +1065,7 @@ def jcalc_motion_subspace(
         joint_S_s[qd_start + 2] = S_2
 
     elif joint_type_value == JointType.FREE or joint_type_value == JointType.DISTANCE:
-        x_child_com_world = wp.transform_point(X_wc, body_com_child)
-        write_free_distance_motion_subspace(X_pa_world, x_child_com_world, qd_start, joint_S_s)
+        write_free_distance_motion_subspace(X_pa_world, qd_start, joint_S_s)
 
 
 @wp.kernel
--- a/newton/tests/test_ik.py
+++ b/newton/tests/test_ik.py
@@ -163,6 +163,21 @@ def _build_descendant_free_distance(device, joint_type) -> tuple[newton.Model, i
     return builder.finalize(device=device, requires_grad=True), child
 
 
+def _build_root_free_distance(device, joint_type) -> tuple[newton.Model, int]:
+    builder = newton.ModelBuilder()
+    body = builder.add_link(mass=1.0)
+    joint = _add_free_distance_joint(
+        builder=builder,
+        joint_type=joint_type,
+        parent=-1,
+        child=body,
+        parent_xform=wp.transform_identity(),
+        child_xform=wp.transform_identity(),
+    )
+    builder.add_articulation([joint])
+    return builder.finalize(device=device, requires_grad=True), body
+
+
 # ----------------------------------------------------------------------------
 # helpers - D6
 # ----------------------------------------------------------------------------
@@ -711,6 +726,46 @@ def test_d6_jacobian_compare(test, device):
     _jacobian_compare(test, device, _d6_objective_builder)
 
 
+def test_free_distance_translated_jacobian_compare(test, device, joint_type):
+    """Match analytic and autodiff Jacobians for a translated floating body."""
+    with wp.ScopedDevice(device):
+        translations = (0.0, 1.0, 5.0, 20.0)
+        model, body = _build_root_free_distance(device, joint_type)
+        joint_q = np.zeros((len(translations), model.joint_coord_count), dtype=np.float32)
+        joint_q[:, 0] = translations
+        joint_q[:, 6] = 1.0
+        target_positions = wp.zeros(len(translations), dtype=wp.vec3, device=device)
+
+        solver_auto = ik.IKSolver(
+            model,
+            len(translations),
+            [ik.IKObjectivePosition(body, wp.vec3(), target_positions)],
+            jacobian_mode=ik.IKJacobianType.AUTODIFF,
+        )
+        solver_ana = ik.IKSolver(
+            model,
+            len(translations),
+            [ik.IKObjectivePosition(body, wp.vec3(), target_positions)],
+            jacobian_mode=ik.IKJacobianType.ANALYTIC,
+        )
+        q_auto = wp.array(joint_q, device=device, requires_grad=True)
+        q_ana = wp.array(joint_q, device=device)
+
+        solver_auto._impl._compute_residuals(q_auto)
+        solver_ana._impl._compute_residuals(q_ana)
+        jacobian_auto = solver_auto._impl._jacobian_at(solver_auto._impl._ctx_solver(q_auto)).numpy()
+        jacobian_ana = solver_ana._impl._jacobian_at(solver_ana._impl._ctx_solver(q_ana)).numpy()
+
+        for problem_idx, translation in enumerate(translations):
+            with test.subTest(translation=translation):
+                assert_np_equal(jacobian_auto[problem_idx], jacobian_ana[problem_idx], tol=1e-4)
+
+        motion_subspace = solver_ana._impl.joint_S_s.numpy()
+        for problem_idx, translation in enumerate(translations[1:], start=1):
+            with test.subTest(motion_subspace_translation=translation):
+                assert_np_equal(motion_subspace[0], motion_subspace[problem_idx], tol=1e-6)
+
+
 # ----------------------------------------------------------------------------
 # 3.  Test-class registration per device
 # ----------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


eric-heiden:
Thanks for tracking this down, Dylan! The write-up of the free joint integration convention is clear and shipping a regression test with the fix is the right call. The main concern is that the shared motion subspace helper also feeds the public Jacobian and mass matrix path, so this change moves those results off their documented COM convention, and the new test does not yet cover a nonzero anchor offset.

newton/_src/sim/articulation.py
```diff
@@ -944,29 +943,29 @@ def write_free_distance_motion_subspace(
...
 
     joint_S_s[qd_start + 0] = wp.spatial_vector(axis_world_x, wp.vec3())
     joint_S_s[qd_start + 1] = wp.spatial_vector(axis_world_y, wp.vec3())
     joint_S_s[qd_start + 2] = wp.spatial_vector(axis_world_z, wp.vec3())
-    joint_S_s[qd_start + 3] = wp.spatial_vector(-wp.cross(axis_world_x, x_child_com_world), axis_world_x)
-    joint_S_s[qd_start + 4] = wp.spatial_vector(-wp.cross(axis_world_y, x_child_com_world), axis_world_y)
-    joint_S_s[qd_start + 5] = wp.spatial_vector(-wp.cross(axis_world_z, x_child_com_world), axis_world_z)
+    joint_S_s[qd_start + 3] = wp.spatial_vector(-wp.cross(axis_world_x, x_anchor_world), axis_world_x)
+    joint_S_s[qd_start + 4] = wp.spatial_vector(-wp.cross(axis_world_y, x_anchor_world), axis_world_y)
+    joint_S_s[qd_start + 5] = wp.spatial_vector(-wp.cross(axis_world_z, x_anchor_world), axis_world_z)
```
eric-heiden, line 968: 🔴 `write_free_distance_motion_subspace` now pivots the three angular columns at the parent anchor (`x_anchor_world`) instead of the child COM, but this helper is shared by two consumers that need different reference points.

`jcalc_motion_subspace` calls it for the IK optimizers *and* for `eval_articulation_jacobian`, which is the kernel behind public `newton.eval_jacobian`, `ArticulationView.eval_jacobian` (`newton/_src/utils/selection.py`), `newton.eval_mass_matrix`, and the `eval_inverse_dynamics_*` entry points (`newton/_src/sim/inverse_dynamics.py`).

| Consumer | Reference point it needs |
| --- | --- |
| IK optimizers (`_compute_motion_subspace_2d`) | parent anchor, to match how `jcalc_integrate` rotates the free joint |
| `eval_articulation_jacobian` | child COM, to match `state.body_qd` |

`eval_jacobian` documents that it satisfies `J_link @ joint_qd == state.body_qd[link]` under Newton's public COM/world body twist convention. With the anchor pivot, the free joint angular columns inject a spurious `ω × (x_com - x_anchor)` into the link's COM linear velocity, while FK gives those angular DOFs zero COM linear contribution (`eval_single_articulation_fk`). So for any floating base whose COM is offset from its joint anchor, which is essentially every free body not sitting at its anchor, `J @ joint_qd != body_qd`. The mass matrix assembled from these columns is then no longer the COM convention inertia, so gravity, Coriolis, and inverse dynamics results for floating bases inherit the same error.

Four existing tests in `newton/tests/test_jacobian_mass_matrix.py` assert exactly this contract and should fail with this change: `test_floating_free_jacobian_matches_body_qd_com`, `test_descendant_free_jacobian_matches_body_qd_com`, `test_floating_free_mass_matrix_matches_com_kinetic_energy`, `test_descendant_free_mass_matrix_matches_com_kinetic_energy`. Their fixture puts the anchor at the origin and the COM at roughly `(1.3, -0.4, 2.2)`, then compares `J @ qd` against `body_qd` at `atol=1e-5`. The COM pivot in this helper was introduced deliberately along with those tests, so it is not leftover from the IK unification.

The IK side of the diagnosis looks right, so the cleanest fix is probably to stop making one pivot serve both callers: keep `jcalc_motion_subspace` COM referenced and pass the reference point in explicitly, so the IK kernels can ask for the anchor.

<details>

<summary>Sketch of an explicit pivot parameter</summary>

```py
@wp.func
def write_free_distance_motion_subspace(
    ...,
    pivot_world: wp.vec3,
    ...
):
    ...
    joint_S_s[qd_start + 3] = wp.spatial_vector(-wp.cross(axis_world_x, pivot_world), axis_world_x)
    joint_S_s[qd_start + 4] = wp.spatial_vector(-wp.cross(axis_world_y, pivot_world), axis_world_y)
    joint_S_s[qd_start + 5] = wp.spatial_vector(-wp.cross(axis_world_z, pivot_world), axis_world_z)


# jcalc_motion_subspace: keep the COM convention
write_free_distance_motion_subspace(..., wp.transform_point(X_wc, body_com_child), ...)

# _compute_motion_subspace_2d in ik_lm_optimizer.py and ik_lbfgs_optimizer.py: use the joint anchor
write_free_distance_motion_subspace(..., wp.transform_get_translation(X_wpj), ...)
```

</details>

Worth running `newton/tests/test_jacobian_mass_matrix.py` alongside `newton/tests/test_ik.py`, since both contracts need to stay green.

newton/tests/test_ik.py
```diff
@@ -163,6 +163,21 @@ def _build_descendant_free_distance(device, joint_type) -> tuple[newton.Model, i
...
+    joint = _add_free_distance_joint(
+        builder=builder,
+        joint_type=joint_type,
+        parent=-1,
+        child=body,
+        parent_xform=wp.transform_identity(),
+        child_xform=wp.transform_identity(),
+    )
+    builder.add_articulation([joint])
+    return builder.finalize(device=device, requires_grad=True), body
```
eric-heiden, line 178: 🔵 `_build_root_free_distance` creates the joint with `parent=-1` and `parent_xform=wp.transform_identity()`, so inside `_compute_motion_subspace_2d` (`ik_lm_optimizer.py`) `X_wpj` is the identity and `x_anchor_world` is `(0, 0, 0)` in all four translation subtests.

That makes the new `-wp.cross(axis_world_*, x_anchor_world)` term identically zero everywhere this test looks: replacing `x_anchor_world` with `wp.vec3()` would still pass. The test pins down that the pivot is no longer the child COM, but it never checks that the pivot is the parent anchor, which is the actual claim of the change, and the anchor offset case is exactly where the candidate conventions differ.

This file already has `_build_descendant_free_distance`, which uses a non-identity free joint `parent_xform` of `(0.7, -0.2, 0.4)` plus a rotation, a rotated moving parent, and a nonzero child COM, so it is already the configuration that would exercise the term. Parameterizing the new test over both fixtures would cover it. A smaller alternative is to give the root fixture a nonzero anchor:

```py
parent_xform=wp.transform(wp.vec3(0.3, -0.4, 0.6), wp.quat_identity()),
```

newton/_src/sim/articulation.py
eric-heiden: Actually, I wonder if it wouldn't be easier if the IK optimizers followed Newton's conventions throughout and also considered the child COM as reference point rather than the parent anchor? Seems inconsistent to how everything else is implemented.

eric-heiden:
Closing in favor of https://github.com/newton-physics/newton/pull/4088.

PR #3869, opened 2026-08-10
Import authored MPM particles from USD

### LINKED ISSUE #3861: Import authored MPM USD schemas into Newton
## Summary

Add Newton-side parsing for the authored MPM USD schemas defined in [newton-usd-schemas#75](https://github.com/newton-physics/newton-usd-schemas/issues/75) and implemented in [newton-usd-schemas#76](https://github.com/newton-physics/newton-usd-schemas/pull/76).

The importer should translate authored MPM scene, particle, and material data into the corresponding Newton model and `SolverImplicitMPM` configuration.

## Acceptance criteria

- Newton loads a USD stage using the finalized MPM schemas and constructs the expected MPM particles, materials, and solver configuration.
- Schema fallbacks, stage units, and material bindings are handled consistently with the schema definitions.
- Automated coverage exercises a representative authored MPM USD fixture.

maxkra15:
## Description

Add experimental authored-MPM import to `ModelBuilder.add_usd()`.

This is the parser layer (PR 2 of 3) and depends on
[newton-usd-schemas#76](https://github.com/newton-physics/newton-usd-schemas/pull/76).
It does not add the Isaac Sim / Kit runtime path.

- Discover `UsdGeom.Points` carrying `NewtonPointsDeformableSimAPI` and governed
  by `PhysicsDeformableBodyAPI` on the Points prim or its direct parent.
- Resolve `physics:simulationOwner` to one `NewtonMPMSceneAPI` scene and expose
  `SolverImplicitMPM.Config.create_from_usd()` for validated solver setup.
- Register MPM custom attributes before particle creation; import positions,
  velocities, world transforms, stage units, widths as radii, and standard
  per-point/body/material mass precedence.
- Read physics-purpose whole-prim and point-`GeomSubset` material bindings,
  including Newton MPM elasticity and constitutive fields.
- Return stable prim-path to half-open particle index ranges for downstream
  synchronization and reset workflows.
- Preserve authored initial plastic volume strain across solver resets.
- Add checked-in multi-prim and material-subset USDA fixtures and a
  `mpm_granular --from-usd` example path.

The minimum schema-wheel version is intentionally not bumped yet because the
MPM APIs are not released. Before this leaves draft, schema PR #76 should merge,
a wheel should be released, and Newton's minimum `newton-usd-schemas` version
should be bumped to that release.

## Checklist

- [x] A changelog fragment has been added for the user-facing change

## Test plan

Validated against the local schema checkout from newton-usd-schemas#76:

- `.venv/bin/python -m unittest newton.tests.test_import_usd_mpm -v`
  (45 passed)
- `.venv/bin/python -m unittest newton.tests.test_implicit_mpm -v`
  (68 passed, 1 skipped)
- `.venv/bin/python -m newton.examples mpm_granular --from-usd --viewer null --num-frames 2 --device cpu --test`
- `uvx --from towncrier==25.8.0 towncrier build --draft --version X.Y.Z --date 2026-08-10`
- `uvx pre-commit run -a`

## New feature / API change

```python
result = builder.add_usd("authored_mpm.usda", load_visual_shapes=False)

config = result["mpm_config"]
particle_ranges = result["path_mpm_particle_map"]
solver = SolverImplicitMPM(builder.finalize(), config=config)
```

## Summary by CodeRabbit

* **New Features**
  * Added experimental USD import support for MPM particles, materials, scene settings, transforms, masses, and particle ranges.
  * Added a granular MPM example option to load simulations directly from USD.
  * Added a complete sand simulation USD asset.
  * Preserved configured initial plastic strain when resetting particle history.

* **Bug Fixes**
  * Improved validation and handling of units, gravity, material subsets, ownership, and invalid particle data.

* **Documentation**
  * Added guidance and examples for USD-based MPM workflows.


newton/_src/sim/builder.py
```diff
@@ -3361,7 +3361,7 @@ def add_usd(
         legacy_margin_gap: bool = False,
         return_deformable_results: bool = False,
     ) -> dict[str, Any]:
-        """Parses a Universal Scene Description (USD) stage and adds rigid bodies, soft bodies, shapes, and joints to the given ModelBuilder.
+        """Parses a Universal Scene Description (USD) stage and adds rigid bodies, MPM particles, soft bodies, shapes, and joints to the given ModelBuilder.
```
eric-heiden: I would leave out "MPM" here, just particles, not tied to a specific solver.

newton/_src/sim/builder.py
```diff
@@ -3537,6 +3567,8 @@ def add_usd(
                   - Mapping from prim path (str) of the UsdGeom to the respective shape index in :class:`~newton.ModelBuilder`
                 * - ``"path_shape_scale"``
                   - Mapping from prim path (str) of the UsdGeom to its respective 3D world scale
+                * - ``"path_mpm_particle_map"``
+                  - Mapping from an MPM ``UsdGeom.Points`` prim path to its half-open ``(particle_start, particle_end)`` builder range
```
eric-heiden: Same here, why call this "MPM" points? Better to keep things solver-agnostic, so `path_particle_map`. Doc strings also need to be updated.

newton/_src/usd/particles.py
eric-heiden, line 1: This file should probably go under `newton/_src/usd/particles.py`.

newton/_src/sim/builder.py
maxkra15: yes, done

newton/_src/sim/builder.py
maxkra15: left out

newton/_src/usd/particles.py
maxkra15: moved

newton/_src/utils/import_usd.py
```diff
@@ -550,26 +585,13 @@ class PhysicsMaterial:
...
     except Exception as e:
         if verbose:
             print(f"Failed to get linear unit: {e}")
-    if not math.isclose(linear_unit, 1.0):
-        warnings.warn(
-            "USD stages with non-unit linear units are not supported. "
-            f"Set metersPerUnit to 1.0 before import. Found metersPerUnit={linear_unit}.",
-            stacklevel=_external_stacklevel(),
-        )
-
```
eric-heiden, line 583: Please keep these unit checks at the beginning of the parse function so that the user will see these warnings before any error etc. is raised later in this function.

newton/tests/test_import_usd_mpm.py
```diff
@@ -0,0 +1,1408 @@
...
+
+        stage = Usd.Stage.CreateInMemory()
+        scene = UsdPhysics.Scene.Define(stage, "/World/PhysicsScene").GetPrim()
+
+        with self.assertRaisesRegex(ValueError, "NewtonMPMSceneAPI is not applied"):
+            SolverImplicitMPM.Config.create_from_usd(scene)
+
+
+@unittest.skipUnless(_has_mpm_schemas(), "Requires Newton USD schemas with MPM support")
+class TestImportUsdMPM(unittest.TestCase):
```
eric-heiden, line 116: 🔵 This `skipUnless` sits on the whole `TestImportUsdMPM` class, which is roughly 40 tests and about 1,300 of this file's 1,408 lines.

`_has_mpm_schemas()` only returns `True` when the installed `newton-usd-schemas` registers `NewtonPointsDeformableSimAPI`, and the pinned version does not ship it. `pyproject.toml` still requires `newton-usd-schemas>=0.4.1`, which resolves to 0.4.1, and that package's `generatedSchema.usda` declares `NewtonSceneAPI`, `NewtonXpbdSceneAPI`, `NewtonKaminoSceneAPI`, `NewtonJointAPI` and similar, with no MPM or Points APIs at all. So `prim.ApplyAPI("NewtonPointsDeformableSimAPI")` fails and the gate is always `False`.

@andrewkaufman, do these schemas (NewtonPointsDeformableSimAPI, NewtonMPMSceneAPI, and NewtonMPMMaterialAPI) ship with a newer newton-usd-schemas version that we should update in this PR?

The practical effect on a stock `uv sync`: the only tests in this file that execute are the two error-path cases in `TestMpmConfigUsdEntryGuards`. That leaves `newton/_src/usd/particles.py` (711 new lines) and `Config.create_from_usd` (~250 new lines) with no unit coverage, and the only remaining signal is the CUDA-only `mpm_granular --from-usd` example test.

The gate is also stricter than part of the suite needs. The programmatic tests genuinely require schema registration, since `_stage()` calls `scene.ApplyAPI("NewtonMPMSceneAPI")`. But the two checked-in fixture tests do not: `usd.has_applied_api_schema()` in `newton/_src/usd/utils.py` falls back to the raw `apiSchemas` metadata, and the `.usda` fixtures author it with `prepend apiSchemas`.

Moving `test_imports_checked_in_two_prim_fixture` and `test_imports_checked_in_material_subset_fixture` into a class gated only on `USD_AVAILABLE`, or applying `_has_mpm_schemas()` per method instead of per class, would give unit conversion, world transforms, point-subset materials, density-derived mass and `particle_Jp` seeding real coverage on the dependency that is actually installed today.

newton/_src/utils/import_usd.py
```diff
@@ -5015,6 +5166,7 @@ def add_converted_loop_joint(
         # "articulation_bodies": articulation_bodies,
         "path_body_relative_transform": path_body_relative_transform,
         "max_solver_iterations": max_solver_iters,
+        "mpm_config": mpm_config,
```
eric-heiden: Instead of coupling the MPM solver config into the USD importer, can we just return the scene prim, or its prim path here? This gets quite messy if we bring solver-specific things into the USD importer. We also have custom attributes for this purpose that the MPM solver can  use, but that would be a larger refactor probably.

eric-heiden (changes requested):

newton/tests/test_import_usd_mpm.py
maxkra15: Done

andrewkaufman:
If you havent yet, you should rebase on main to pickup the v0.5.0 schemas

maxkra15:
rebased and resolved threads

newton/_src/utils/import_usd.py
maxkra15: importer now returns particle_scene_prim instead of an MPM config

newton/_src/utils/import_usd.py
```diff
@@ -5035,6 +5176,7 @@ def add_converted_loop_joint(
         # "articulation_bodies": articulation_bodies,
         "path_body_relative_transform": path_body_relative_transform,
         "max_solver_iterations": max_solver_iters,
+        "particle_scene_prim": particle_scene_prim,
```
eric-heiden: ⚪ This is the only entry in the `add_usd()` result that hands back a live `pxr.Usd.Prim`. Every other value is plain data (indices, maps, floats, path strings), and the same concept is already exposed as a string a few lines above in this same dict:

```python
"physics_scene_path": str(physics_scene_prim.GetPath()) if physics_scene_prim is not None else None,
```

Returning the object rather than its path has two effects that are easy to miss. The result dict now holds a reference into the USD stage, so an import result that used to be a self-contained snapshot keeps the stage alive for as long as anything holds the dict. And the prim can go invalid later if the stage is closed or the scene prim is removed, which is not a failure mode callers expect from a plain-data result.

Could you return the path instead, for symmetry with `physics_scene_path`?

```python
"particle_scene_path": str(particle_scene_prim.GetPath()) if particle_scene_prim is not None else None,
```

`SolverImplicitMPM.Config.create_from_usd()` still wants a prim and `add_usd()` never returns the stage, so a caller who needs the solver config would open the stage itself and pass it in, which `add_usd()` already accepts:

```python
stage = Usd.Stage.Open(newton.examples.get_asset("mpm_sand.usda"))
import_result = builder.add_usd(stage, load_visual_shapes=False)
scene_path = import_result["particle_scene_path"]
mpm_options = SolverImplicitMPM.Config.create_from_usd(stage.GetPrimAtPath(scene_path))
```

The rename would also touch the two result tables in this file and in `builder.py`, the snippet in `docs/concepts/usd_parsing.rst`, `example_mpm_granular.py`, and `test_import_usd_mpm.py`.

maxkra15: done

eric-heiden (approved):
Thanks!

PR #3871, opened 2026-08-10
Make Kamino model comparison order-independent

chschuma-disney:
## Description

The comparison between two Kamino models currently assumes that the order of bodies and joints is independent of the import/conversion path. For USD models with an articulation, importing through the Kamino USD importer will yield a different order than importing through the Newton importer and then converting to a Kamino model. See #3844 

With Kamino's USD importer set to be removed at some point in the future, the sensible short-term option would be to make the comparison order-independent.

## Checklist

## Summary by CodeRabbit

* **Tests**
  * Added coverage for comparing equivalent models with reordered bodies and entities.
  * Added validation for strict ordering, corrupted local IDs, and invalid geometry references.
  * Updated conversion tests for reordered model elements and renamed joint transform fields.

* **Improvements**
  * Model comparisons can optionally support label-based reordering while validating indexed references, geometry relationships, and tensor-backed data.
  * Enhanced comparison checks provide more reliable results when equivalent model elements appear in different orders.


andrewkaufman (commented):
Thanks for picking this up so quickly — and agreed on the direction: making the comparison order-independent is the right short-term move, and sunsetting the Kamino USD importer is the right medium-term one.

I confirmed this does fix the motivating failure. With #3821's asset pin applied locally, `test_02_model_conversions_dr_testmech_from_usd` goes from failing (`RigidBodiesModel.label are not equal`) to passing. `test_03_model_conversions_dr_legs_from_usd` still errors with `ValueError: cannot reshape array of size 1 into shape (3)`, but that's the normals fix in #3845, not this PR's concern.

Two things I think need addressing before this lands, and two that will bite later.

## 1. The `bid` / `jid` / `gid` comparisons become unconditionally vacuous

The `entity_index_remaps` branch in `assert_array_attributes_equal` overwrites `desired` without ever reading `attr1`:

```python
for i, other_idx in enumerate(mapping):
    desired[i] = remap[other_idx]
```

`_entity_index_remap(mapping, ref_local)` builds `remap[mapping[i]] = ref_local[i]`, so this evaluates to `desired[i] = ref_local[i]` — and `ref_local` is `model0`'s own array, i.e. `actual`. The assertion reduces to `actual == actual`.

Setting every `bid`, `jid` and `gid` in `model1` to `12345` and running `test_core_model`:

| | corrupted ids detected |
| --- | --- |
| `allow_reordering=False` | **7 test failures** |
| `allow_reordering=True` (this PR's default) | **0** |

I don't think a remap can rescue this one: `bid`, `jid` and `gid` are all documented as world-local *positional* ids, so once the rows are reordered there is no cross-model comparison left to make. The honest options are to assert the per-model invariant (`bid[row] == row - world_offset`) on each model independently, or to drop these three from the compared list with a comment explaining why. A skip that looks like a comparison is worse than an explicit skip, since it reads as coverage in review.

## 2. `allow_reordering=True` relaxes all 9 call sites, not just the one that needs it

`mapping` ends up non-`None` whenever the flag is on — **including when the computed mapping is the identity** — so the strict path is now unreachable in the test suite. That also drops the nine per-joint offset arrays (`coords_offset` through `kinematic_cts_offset`) from comparison everywhere, since they are only appended under `if mapping is None`.

Defaulting to `False` and opting in at the one affected call site would keep the blast radius at the problem you're solving. If you'd rather keep the default on, falling back to the strict path when the computed mapping turns out to be the identity gets most of the way there — it would preserve existing coverage in the 6 of 7 comparisons that aren't reordered.

## 3. `collidable_pairs` / `excluded_pairs` hold geom indices that are never remapped

Both are in the geoms `array_attributes`, but their leading dimension is the pair count rather than `num_geoms`, so `len(mapping) == desired.shape[0]` is false and they skip permutation entirely — while their *contents* are still `model1` geom indices, and no `index_remaps` entry covers them.

This doesn't fire today only because `dr_testmech` reorders bodies and joints but leaves geometries alone (`bodies=PERM(10) joints=PERM(14) geoms=ident(10)`). The first asset that reorders geometry will produce a spurious failure with 32 collidable pairs to eyeball.

## 4. `ModelInfo.base_body_index` / `base_joint_index` have the same gap

Both are documented as indices "w.r.t the model" and are compared in `assert_model_info_equal` with no mapping applied. `dr_testmech` permutes bodies and still matches, so the base body happens to keep its row — same failure mode as #3, just not triggered yet.

## On test coverage

Worth knowing before you rely on CI here: at the asset pin currently on `main`, every one of the 8 `assert_model_equal` calls in `test_core_model` produces an identity mapping. None of the new permutation, remapping or packed-slice logic executes, so a green run on this PR isn't evidence about any of it. The path only comes alive once #3821 bumps the pin, and even then it's 1 permuted comparison out of 7.

A unit test that hands the checker a synthetic reordered pair directly would cover all of the above cheaply, and would have caught #1 immediately.

Happy to take any of these on myself if that's easier — this is unblocking my PR, so I'm glad to share the load.

chschuma-disney:
I unfortunately ran out of time today to have another look at this. @andrewkaufman If you want to take on the revision, feel free to do so. Otherwise, I'll look into it tomorrow.

chschuma-disney:
I hope all issues should be addressed now. I did not add `fk_act_flag` to the checked attributes, since it was not present before and is not currently relevant in this context.

andrewkaufman (commented):
CI is red on the current head, and I think it's this PR rather than infra flake.

`newton-unittests (ubuntu-24.04-arm)` came back `Ran 4844 tests, FAILED (errors=6)`, and all six are `test_kamino_core_model.TestModelConversions` — tests 00, 01, 02, 21, 30 and 40. It isn't arm-specific; it just needs `--strict-warnings`, which every CI workflow passes:

| branch | `uv run --extra dev -m newton.tests -k test_kamino_core_model --strict-warnings` |
| --- | --- |
| `main` @ `ae5ef001a` | **OK** (21 tests) |
| this PR @ `0925c379` | **FAILED (errors=6)** |

The cause is the debug diff at `checks.py:188` moving from warp to numpy:

```python
diff = attr0 - attr1        # before: warp array subtraction
diff = actual - desired     # after:  numpy, on the .numpy() results
```

`JointsModel.tau_j_max` is `inf` in both models, so `inf - inf` now yields a NaN plus `RuntimeWarning: invalid value encountered in subtract`. Warp's subtraction never warned, which is why this only appears now. Traceback from the arm run:

```
  File ".../newton/tests/kamino/utils/checks.py", line 188, in assert_array_attributes_equal
    diff = actual - desired
           ~~~~~~~^~~~~~~~~
RuntimeWarning: invalid value encountered in subtract
```

Since `diff` only feeds the `msg.debug(...)` call, wrapping it in `np.errstate(invalid="ignore")` — or computing it only when debug logging is enabled — should be enough.

I have a few smaller comments on coverage and on test discovery, but they're all secondary to this one; I'll follow up separately so it doesn't hold up the fix.

andrewkaufman:
Since this is blocking my pin bump I went ahead and fixed it, in case it saves you a round trip: [`akaufman/3871-strict-warnings-fix`](https://github.com/andrewkaufman/newton/tree/akaufman/3871-strict-warnings-fix), one commit on top of your `0925c379`. Cherry-pick it, copy the diff, or ignore it entirely — no attachment to my version.

```python
# Unbounded limits are stored as inf (e.g. JointsModel.tau_j_max), so this purely
# informational diff hits inf - inf. Left unguarded it raises a RuntimeWarning, which
# CI turns into a test error via --strict-warnings.
with np.errstate(invalid="ignore"):
    diff = actual - desired
```

`uv run --extra dev -m newton.tests -k test_kamino_core_model --strict-warnings` goes from 6 errors to 21 passing.

One update since my last comment: it is not just the arm runner. The whole matrix has now reported and **macos-latest, windows-latest, ubuntu-latest and ubuntu-24.04-arm all failed**, which fits — `--strict-warnings` is passed by every workflow, and nothing about `inf - inf` is platform-specific.

andrewkaufman (approved):
CI is green across all four platforms and the GPU runner, including `ubuntu-24.04-arm` which was the one that surfaced the `--strict-warnings` failure. The cherry-pick is byte-identical to what I offered, so nothing to re-check there.

All four of my original findings are addressed, and addressed properly rather than papered over:

1. **`bid`/`jid`/`gid`** — dropped from the cross-model comparison under reordering and replaced with the per-model invariant on *both* models via `_assert_world_local_ids`. That is the honest version of the two options I suggested; the new `test_reordered_model_detects_corrupted_local_ids` covers it.
2. **Blanket relaxation** — `allow_reordering` now defaults to `False` *and* `_none_if_identity` collapses an identity permutation back to the strict path, so the six unreordered comparisons keep their original coverage. Only `test_02` opts in.
3. **`collidable_pairs`/`excluded_pairs`** — `_assert_geom_pairs_equal` remaps the geom indices and canonicalises pair order, so the case I flagged as latent is handled before it can bite.
4. **`base_body_index`/`base_joint_index`** — index remaps wired through `assert_model_info_equal`.

I also re-verified the thing that mattered most to me: at the asset pin currently on `main` this is a clean no-op, 21/21 in `test_kamino_core_model` on both CUDA and CPU, so it carries no risk for anyone not on the new pin.

Thanks for turning this around quickly, and for taking the `fk_act_flag` call the conservative way.

One thing for whoever picks up #3844 next, not for this PR: with #3845 merged, `test_03_model_conversions_dr_legs_from_usd` gets past the `normals` `ValueError` and now needs `allow_reordering=True` as well — after which it surfaces a genuine naming divergence (`joint_1` vs `world_to_/dr_legs/pelvis`) that order-independent matching cannot resolve. Details are on #3821.

rubengrandia (approved):

PR #3772, opened 2026-08-03
Fix XPBD joint separation after large errors

eric-heiden:
## Description

Fix XPBD joint constraints becoming permanently separated after a large transient anchor error.

The joint solver used the child anchor to compute the parent body's moment arm. As the anchors separated, that moment arm grew with the error, inflated the angular effective-mass term, and progressively reduced positional correction until the joint could no longer catch up.

Compute the parent moment arm from the relative offset projected onto the joint's admissible linear coordinates. Locked coordinates project to zero, limit violations project to the nearest boundary, position-driven coordinates project to their target, and unconstrained coordinates retain their valid extension. This removes separation error from the lever arm without discarding legitimate prismatic or D6 motion.

## Checklist

- [x] New or existing tests cover these changes
- [x] The documentation is up to date with these changes
- [x] `CHANGELOG.md` has been updated (if user-facing change)

## Test plan

```text
uv run --extra dev -m newton.tests -k test_prismatic_joint
uv run --extra dev -m newton.tests -k test_ball_joint_recovers_from_large_anchor_separation
uvx pre-commit run -a
```

The new ball and prismatic regressions pass on CPU and CUDA. The selected test runs completed 46 and 42 tests respectively, and all pre-commit hooks pass.

The regression tests were also verified to fail before their corresponding fix: the original ball-joint case retained a 1.318 m gap after one step, and the fully-locked-only fix left 1.389 m of transverse error in the prismatic case.

## Bug fix

**Steps to reproduce:**

1. Connect thin rigid bodies with a ball or prismatic joint.
2. Introduce a large anchor separation transverse to every permitted linear coordinate.
3. Advance `SolverXPBD` with a low iteration count.
4. Observe the old solver convert much of the correction into rotation and leave a large residual gap. Under sustained load, the gap grows and appears to be a broken joint.

**Minimal reproduction:**

```python
body_q = state_0.body_q.numpy()
body_q[child, :3] += np.array((0.5, 1.0, 1.0), dtype=np.float32)
state_0.body_q.assign(body_q)

solver = newton.solvers.SolverXPBD(model, iterations=2)
solver.step(state_0, state_1, None, None, 1.0 / 240.0)
```

The complete recovery and valid-extension torque assertions are in `test_solver_xpbd.py`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed fully locked joints becoming permanently separated after large temporary anchor displacements.
[...]

PR #3877, opened 2026-08-11
Fix API review analysis output

eric-heiden:
## Description

Fix the public API review detector so it reports stable APIs accurately and keeps analyzer limitations from obscuring concrete findings.

- Track public instance attributes initialized in `__init__`, so attribute-to-property transitions are reported as modifications rather than additions.
- Resolve assignments nested in module-level control flow, including `newton.__version__`.
- Fingerprint normalized inherited public surfaces instead of entire provider modules.
- Hide uncertainty diagnostics when concrete API changes already request review. For uncertainty-only cases, show a concise explanation with technical details collapsed.

This reduces false positives observed on #3803 while preserving conservative review requests when static analysis is genuinely inconclusive.

## Checklist

- [x] For user-facing changes, a fragment has been added by following the [changelog fragment instructions](https://github.com/newton-physics/newton/blob/main/changelog/README.md) — not applicable; this is a CI-only change

## Test plan

- Verified the four focused regression tests fail against the unmodified detector and pass with this change.
- `uv run --no-project -m unittest scripts.ci.tests.test_detect_api_changes scripts.ci.tests.test_pr_api_changes_workflow -v`
- `uvx pre-commit run -a`
- Replayed the detector against PRs #3053, #3772, #3775, #3752, and #3623.

## Summary by CodeRabbit

* **New Features**
  * Improved API change detection for instance attributes, inherited APIs, and exports defined through control flow.
  * More accurately distinguishes implementation changes from public API changes.

* **Bug Fixes**
  * Correctly identifies changes when instance attributes become properties.
  * Flags uncertain inherited API changes for manual review while ignoring unrelated inherited implementation updates.

* **Documentation**
  * Clarified generated review comments and hides analyzer uncertainty details when concrete API changes are detected.

### DIFF
--- a/scripts/ci/detect_api_changes.py
+++ b/scripts/ci/detect_api_changes.py
@@ -47,6 +47,7 @@ class Definition:
         self.module = module
         self.node = node
         self.members = members or {}
+        self.surface_fingerprint: str | None = None
 
 
 class Module:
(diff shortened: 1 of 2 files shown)


shi-eric:
Non-blocking: the collector resolves the `try`-body assignment for `newton.__version__`, but does not traverse its `except PackageNotFoundError` handler because `ExceptHandler` is not an `ast.stmt`. Consequently, changing only the fallback value still reports `unchanged`. This is unlikely to matter for API review because `__version__` is release metadata, but it is a limitation of the control-flow support described in the PR.

shi-eric (approved):

PR #3885, opened 2026-08-11
[Kamino] Add config validation rejecting dense solver + balanced penalty

rubengrandia:
## Description

It was possible to select the balanced penalty update together with the dense solver. This creates a problem because then only the RHS sees the updated rho, but the LHS was factorized with the old rho. 

This PR also adds validation of the config to the solver init. This catches the common pattern of modifying the config after constructing it, which never validated again.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Added validation to reject adaptive PADMM penalty updates when sparse dynamics are not enabled.
  - Explicitly supplied solver configurations are now validated before use, helping prevent invalid setups and providing clearer configuration errors.
  - Invalid solver settings are now identified earlier, reducing the risk of failed or unexpected solver runs.

### DIFF
--- a/newton/_src/solvers/kamino/solver_kamino.py
+++ b/newton/_src/solvers/kamino/solver_kamino.py
@@ -735,7 +735,7 @@ class SolverKamino(SolverBase, CouplingInterface):
         if config is None:
             config = self.Config.from_model(model)
         else:
-            # Validate the user-provided config. Protects against motifying the config after initialization.
+            # Validate the user-provided config. Protects against modifying the config after initialization.
             config.validate()
         self._config = config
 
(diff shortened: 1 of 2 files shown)


newton/_src/solvers/kamino/solver_kamino.py
```diff
@@ -728,6 +734,9 @@ def __init__(
         # found on the model, so `self._config` will always be fully initialized after this step.
         if config is None:
             config = self.Config.from_model(model)
+        else:
+            # Validate the user-provided config. Protects against motifying the config after initialization.
```
chschuma-disney: `motifying` -> `modifying`

chschuma-disney (commented):
LGTM, just a small typo

Guirec-Maloisel (commented):
LGTM too (up to the typo)

newton/_src/solvers/kamino/solver_kamino.py
rubengrandia: That's what I get for manually writing docs. Should just stick to prompting..

Guirec-Maloisel (approved):

chschuma-disney (approved):

PR #3888, opened 2026-08-11
Add opt-in USD schema fallback precedence

mzamoramora-nvidia:
## Description

Continues and supersedes #3572.

Addresses the fallback-precedence part of #3307 and provides the internal foundation for the public, source-neutral resolution API in #3984, which continues and supersedes #3568.

This PR lets the USD importer resolve unauthored properties from composed definitions in `Usd.SchemaRegistry`. Newton therefore does not need copied PhysX or MuJoCo fallback tables.

A temporary [design note](https://github.com/mzamoramora-nvidia/newton/blob/978a9888f02fd7c68c87f7b21bc0f80ac2be8d51/docs/guide/usd_schema_fallback_resolution.md) explains the intended behavior and module boundary for review. I plan to remove it before merge once we agree on the design.

### Resolution policy

| `use_registered_schema_fallbacks` | `audit_registered_schema_fallbacks` | Behavior |
| --- | --- | --- |
| `False` | `False` | Keep legacy precedence and evaluate only the legacy policy |
| `False` | `True` | Keep the legacy result and audit it against registered-schema precedence |
| `True` | `False` | Use registered-schema precedence and evaluate only that policy |

The two flags cannot both be `True`.

Legacy precedence remains:

1. Authored USD value
2. Importer argument or default
3. Compatibility resolver default

Registered-schema precedence is:

1. Explicit supported importer override
2. For each resolver in priority order, its authored value and then its registered schema fallback
3. Importer default
4. Eligible compatibility default from an unregistered schema or a resolver without declared schema ownership

A blocked value is treated as unauthored, so a registered fallback can still be used. Unset or unusable candidates fall through to the next candidate. A registered property with no fallback does not receive a compatibility default.

Some `add_usd()` arguments must distinguish omission from an explicit value. This PR supports that distinction for `enable_self_collisions`, `mesh_maxhullvert`, `joint_drive_gains_scaling`, and `collapse_fixed_joints`. Their public Python defaults and reflected signatures remain unchanged.

Set `audit_registered_schema_fallbacks=True` to emit one `DeprecationWarning` when the migration audit proves that registered-schema precedence changes an interpreted property value or a source that changes its meaning. The audit compares values after property-specific sentinel handling, unit conversion, compound assembly, and validation. For omitted SDF padding, it uses `gap` for ordinary collision and `margin + gap` for colliding hydroelastic shapes. The audit is disabled by default because it evaluates both policies.

The importer-specific policy and audit code lives in the private `_usd_resolution_policy.py` module. `schema_resolver.py` remains responsible for schema applicability, authored reads, registered fallbacks, and compatibility defaults.

Schema ownership remains an internal detail of the built-in resolvers in this PR. Existing custom `SchemaResolver` subclasses keep their compatibility behavior without adopting a new public ownership API.

### Scope

The `newton[importers]` extra requires `newton-usd-schemas>=0.5.0`, and the repository lock currently tests version 0.5.0. PhysX and MuJoCo schema plugins remain optional: their resolvers can read authored vendor attributes without those plugins, while registered vendor fallbacks require the corresponding plugin. Newton does not copy vendor fallback catalogs.

Most deformable properties still read authored `physics:` values first, followed by resolver-enabled vendor namespaces. Cloth `mass_model` and `shell_thickness` already use the common resolver. These properties are described by the experimental [AOUSD UsdPhysics Deformables proposal](https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/111). Registered-fallback support for the remaining deformable properties will be expanded in the future.

Public source-neutral results and provenance are implemented by follow-up #3984. A reviewed public ownership contract and broader dynamic applicability remain future work. This PR addresses #3307 but does not close it.

## Checklist

## Test plan

```bash
uvx pre-commit run -a
uv run --extra dev -m newton.tests -s newton/tests -p test_schema_resolver.py --strict-warnings -j 1
uv run --extra dev -m newton.tests -s newton/tests -p test_import_usd.py --strict-warnings -j 1
uv run --extra dev -m newton.tests -s newton/tests -p 'test_import_usd_deformable*.py' --strict-warnings -j 1
uv run --extra dev -m newton.tests -s newton/tests -p test_api.py --strict-warnings -j 1
uv run --extra docs --extra sim sphinx-build -j auto -W -b html docs /tmp/newton-pr3888-html
uv run --extra docs --extra sim sphinx-build -j auto -W -b doctest docs /tmp/newton-pr3888-doctest
```

Results:

- All pre-commit hooks passed.
- All 87 schema-resolver tests passed with strict warnings.
- All 372 USD-importer tests passed with strict warnings.
- 173 deformable USD-importer tests passed with strict warnings; one expected
  test was skipped.
- All 13 API and signature tests passed with strict warnings.
- The strict Sphinx HTML build and all 88 doctests passed.

## New feature / API change

Adopt registered schema fallbacks:

```python
builder.add_usd(stage, use_registered_schema_fallbacks=True)
```

Audit a legacy import before migrating:

```python
builder.add_usd(stage, audit_registered_schema_fallbacks=True)
```

Explicitly override a supported importer value:

```python
builder.add_usd(
    stage,
    enable_self_collisions=False,
    use_registered_schema_fallbacks=True,
)
```


newton/_src/utils/import_usd.py
```diff
@@ -386,7 +423,14 @@ def parse_usd(
...
 
-                The ``schema_resolvers`` argument may change without prior notice.
+                The ``schema_resolvers`` and ``use_applied_schema_fallbacks``
+                arguments may change without prior notice.
+        use_applied_schema_fallbacks: When omitted, retain legacy resolution and
+            warn when registered schema fallbacks would change imported values.
+            Pass True to use registered schema fallbacks without migration
+            warnings. Pass False explicitly to retain legacy resolution without
+            migration warnings. Unregistered resolver defaults remain compatibility
+            defaults after importer defaults.
```
eric-heiden: I think it's better to explicitly deprecate the case when `use_applied_schema_fallbacks=False`. It would be confusing to change the way parameters are resolved through a flag long-term, we should converge on one approach that is always predictable. So we should add deprecated notes here in the docstring and probably also show deprecation warnings when a value is resolved in a way that would resolve differently going forward with these applied schema fallbacks.

newton/_src/utils/import_usd.py
```diff
@@ -3853,8 +4085,9 @@ def _record_fallback_collider_mass_information(
                         prim,
                         prim_type=PrimType.SHAPE,
                         key="max_hull_vertices",
-                        default=mesh_maxhullvert,
+                        default=max_hull_vertices_default,
                         verbose=verbose,
+                        override=max_hull_vertices_override,
                     )
```
eric-heiden, line 3802: 🟠 With `use_applied_schema_fallbacks=True`, `newton:maxHullVertices` resolves to the registered applied-schema fallback, which is the sentinel `-1` (`Usd.SchemaRegistry().FindAppliedAPIPrimDefinition('NewtonMeshCollisionAPI').GetAttributeFallbackValue('newton:maxHullVertices')` returns `-1`, while `newton.Mesh.MAX_HULL_VERTICES` is `64`). That sentinel is then assigned verbatim to `Mesh.maxhullvert`, and nothing downstream normalizes it:

- `Mesh.__init__` only maps `None` to `MAX_HULL_VERTICES` (`newton/_src/geometry/types.py:227-228`)
- `remesh_convex_hull` only decimates when `maxhullvert > 0` (`newton/_src/geometry/utils.py:834` and `:856`), and its docstring states that a non-positive budget means no limit

So a resolved `-1` means **unbounded**, not the documented default of 64.

Measured on the stock `unitree_g1` `usd_structured` asset (the one loaded by `newton.examples.robot.example_robot_g1`) with `load_visual_shapes=False`, followed by `approximate_meshes("convex_hull")`:

| policy | meshes | resolved maxhullvert | total hull verts | max verts / hull |
|---|---|---|---|---|
| `use_applied_schema_fallbacks=False` | 37 | 64 | 2,368 | 64 |
| `use_applied_schema_fallbacks=True` | 37 | **-1** | **42,534** | **5,583** |

That is an 18x increase in collision hull vertices on a shipped asset for any caller who follows the documented opt-in without also passing `mesh_maxhullvert`. It is not only a cost regression either: the resolved collision geometry changes, so contact behavior diverges as the scene settles. Since the changelog and `docs/concepts/usd_parsing.rst` both point users at `use_applied_schema_fallbacks=True` as the future behavior, this sits on the main migration path rather than an exotic configuration.

Could we normalize the sentinel where the resolved value is consumed, so a non-positive `max_hull_vertices` falls back to `Mesh.MAX_HULL_VERTICES` before it reaches `mesh.maxhullvert`? Note that `test_fallback_policy_distinguishes_omitted_and_explicit_false` in `newton/tests/test_import_usd.py` currently asserts `composed_builder.shape_source[0].maxhullvert == -1`, so it codifies the sentinel leak as expected behavior and would need to assert the normalized value instead.

newton/_src/utils/import_usd.py
```diff
@@ -811,36 +884,59 @@ def _resolve_joint_limit_gain(
...
+    def _resolve_joint_velocity_limit(prim: Usd.Prim) -> float | None:
+        return R.get_value(
+            prim,
+            prim_type=PrimType.JOINT,
+            key="velocity_limit",
+            default=None,
+            verbose=verbose,
+            legacy_value_transformer=lambda value, _resolver: None if value == float("inf") else value,
+            comparison_key=lambda value, _resolver: default_joint_velocity_limit if value is None else value,
+        )
```
eric-heiden: 🔵 Moving the `inf -> None` normalization into `R.get_value` via `legacy_value_transformer` means the transform is applied first and the transformed value is what gets handed to `_report_missing` (`newton/_src/usd/schema_resolver.py:544`), which prints an error whenever that value is `None` (`schema_resolver.py:826-829`).

With the default resolver set (`schema_resolvers = [SchemaResolverNewton()]`) and `newton:velocityLimit` unauthored, the legacy lookup falls through to the Newton mapping default of `float("inf")`, the transformer converts it to `None`, and `add_usd(..., verbose=True)` prints:

```
Error: Cannot resolve value for 'joint:velocity_limit' on prim '<path>'; no authored value, explicit default, or applicable resolver fallback.
```

This happens once per revolute/prismatic joint and once per merged D6 joint, for a value that resolved perfectly well and simply means "unlimited". A 30-DOF robot produces 30 spurious `Error:` lines on a healthy import.

On `main` both call sites converted `inf` to `None` after the call returned, so the resolver saw the non-`None` mapping default and never reported anything. `velocity_limit` is the only caller passing `legacy_value_transformer`, so this affects just this key.

Could we report against the pre-transform value, for example by having `_get_value_with_policy` remember it and pass that to `_report_missing`? Alternatively, move the `inf -> None` conversion back to the call site here and keep `comparison_key` for the audit path only.

eric-heiden:
🟠 **The new precedence adopts registered schema fallbacks without translating Newton's "unset" sentinels back into "no opinion"**

The precedence change itself is reasonable. What is missing is a step that recognizes sentinel fallbacks. Resolution decides "this resolver has an opinion" by testing `value is not None`, but Newton's schemas do not encode "no opinion" as `None`. They encode it as `-1`, `inf`, and `-inf`. `schemas.py` states this convention directly:

```python
# `-inf` is the "unset" sentinel (same convention as gap / shell_thickness above).
```

Because `GetAttributeFallbackValue()` returns one of those sentinels rather than `None`, the fallback passes every `None` check and is returned as a resolved value. I checked all 29 Newton mapping keys that have a registered schema, and **every one of them has a registry fallback**. So the schema-fallback stage always fires, and the importer-default and compatibility-default stages are unreachable for every registered Newton property whenever its schema is applied and the attribute is unauthored.

Three gates produce this, none of them sentinel aware:

| Location | Gate | Consequence |
|---|---|---|
| `schema_resolver.py:386` | `if value is not None` when harvesting registry fallbacks | a sentinel is recorded as a real fallback |
| `schema_resolver.py:422` | `elif fallback is not _MISSING_FALLBACK` | a sentinel short-circuits the chain before the importer default at line 425 |
| `schema_resolver.py:618` | `if resolved.authored and legacy_value_transformer is not None` | sentinel normalization runs for authored values but not for schema fallbacks |

### Where this actually fails

Measured with `SchemaResolverNewton` on a prim that has the API schema applied and the property unauthored:

| Property | Registry fallback | What that value means to Newton | Legacy | Opt-in | Result |
|---|---|---|---|---|---|
| `newton:maxHullVertices` | `-1` | unset, use `mesh_maxhullvert` | `64` | `-1` | `remesh_convex_hull` caps only when `> 0`, so `-1` is unbounded |
| `newton:limitStiffness` | `-inf` | force the builder default **and** outrank lower-priority schemas | `None` | `-inf` | authored PhysX limit gains are silently discarded |
| `newton:limitDamping` | `-inf` | same | `None` | `-inf` | same |
| `newton:velocityLimit` | `inf` | unlimited, normalized to `None` | `None` | `inf` | normalization is skipped, so `inf` reaches the builder |

These survive only because the call site normalizes the sentinel itself:

| Property | Registry fallback | Legacy | Opt-in | Why it is fine |
|---|---|---|---|---|
| `newton:contactStiffness`, `contactDamping`, `contactFrictionGain`, `contactAdhesion` | `-inf` | `None` | `-inf` | `_interpret_usd_contact_parameter` maps non-finite to `None` |
| `newton:contactGap`, `sdfPadding`, `sdfTargetVoxelSize`, `shellThickness` | `-inf` | `-inf` | `-inf` | consumers test `!= float("-inf")` / `isfinite` |
| `newton:maxSolverIterations` | `-1` | `-1` | `-1` | unchanged |

And these are genuine value changes rather than sentinel bugs, but they make the importer default unreachable, so they are worth confirming as intended:

| Property | Legacy | Opt-in |
|---|---|---|
| `newton:torsionalFriction` | `0.25` | `0.005` |
| `newton:rollingFriction` | `0.0005` | `0.0001` |
| `newton:sdfMaxResolution` | `-inf` | `64` |
| `newton:sdfNarrowBandInner` / `Outer` | `-inf` | `-0.1` / `0.1` |
| `newton:hydroelasticStiffness` | `-inf` | `1e10` |
| `newton:hydroelasticEnabled` | `None` | `False` |
| `newton:sdfTextureFormat` | `None` | `uint16` |
| `newton:damping` | `None` | `0.0` |

For the two friction rows this also means `builder.default_shape_cfg.mu_torsional` and `mu_rolling` stop having any effect on materials that carry `NewtonMaterialAPI`.

### The sharpest case: applying `NewtonJointAPI` discards authored PhysX limit gains

`_resolve_newton_limit_ke` documents `-inf` as "an **authored** `-inf` selects the builder default, taking precedence over per-DOF fallbacks from lower-priority schemas". The opt-in path now hands it an *unauthored* `-inf` that came from the schema registry, so it takes that override branch. On a revolute joint with `physxLimit:angular:stiffness = 777` and `damping = 33` authored, and `NewtonJointAPI` applied but `newton:limitStiffness` left unauthored:

| `NewtonJointAPI` applied | `use_applied_schema_fallbacks` | `limit_ke` | `limit_kd` |
|---|---|---|---|
| no | `False` | 44518.8 | 1890.8 |
| no | `True` | 44518.8 | 1890.8 |
| yes | `False` | 44518.8 | 1890.8 |
| yes | `True` | **10000.0** | **10.0** |

The last row is the builder default. Merely applying `NewtonJointAPI`, without authoring anything on it, throws away the authored PhysX gains. The migration warning does flag this prim, but it describes it as "schema fallbacks will take precedence", which reads like an intentional schema opinion rather than an artifact of an unset sentinel.

### Suggested direction

The cleanest fix is to translate sentinels back into "no opinion" at the point where a registered fallback enters the chain, rather than at each consumer. Something like a per-attribute "unset" predicate on `SchemaAttribute`, defaulting to the existing convention (`-1` for the count-like properties, non-finite for the float ones), consulted in `_registered_attribute_fallbacks` or right after `read_fallback` in `_resolve_value`. A fallback that matches the sentinel would then be skipped so resolution continues to the next resolver, the importer default, and the compatibility default.

Two smaller points that fall out of the same root cause:

- `legacy_value_transformer` should apply to schema fallbacks too, not only to authored values, otherwise an authored `inf` and the schema's own `inf` resolve differently for the same property.
- The material contact path already gets this right by normalizing at the call site with `_interpret_usd_contact_parameter`. The velocity limit path moved its normalization into `get_value`, which is what breaks it, so moving it back would fix both that and the spurious verbose error.

newton/_src/utils/import_usd.py
eric-heiden: 🟠 Thanks for adding `fallback_is_unset=_fallback_is_negative_one` (`newton/_src/usd/schemas.py:299`). That is the right fix for the unauthored case and it fully covers what I flagged here.

The issue now is that `_interpret_usd_max_hull_vertices` (`newton/_src/utils/import_usd.py:96-100`) was added for the same purpose and is still applied to the resolved value at line 4103:

```py
if resolver is not None and value == -1:
    return Mesh.MAX_HULL_VERTICES
```

Because the sentinel is already dropped during resolution, this helper no longer sees the unauthored case at all. The only `-1` values that still reach it are ones an author explicitly wrote, and it rewrites those too. The `resolver is not None` guard cannot separate the two, since an authored `-1` also arrives with a resolver attached.

That matters because `-1` is a documented authoring value, not only an internal sentinel. `NewtonMeshCollisionAPI` states that "If `newton:maxHullVertices` is set to -1, the hull computation should use as many vertices as necessary to produce a perfect convex hull", and `remesh_convex_hull` implements it, with both gates written as `maxhullvert > 0` (`newton/_src/geometry/utils.py:834`, `:856`).

Importing a mesh with `NewtonMeshCollisionAPI` applied and `newton:maxHullVertices` explicitly authored:

| authored value | `main` | this branch, flag off | this branch, flag on |
| --- | --- | --- | --- |
| `-1` | `-1` (exact hull) | **`64`** | **`64`** |
| `0` | `0` | `0` | `0` |
| `32` | `32` | `32` | `32` |
| unauthored | `64` | `64` | `64` |

So unlike the other precedence changes, this one is not gated behind `use_applied_schema_fallbacks`; it changes the default path too. On a test mesh the collider drops from 400 hull vertices to 64, an 84% reduction, and nothing is warned.

Since `fallback_is_unset` already covers the unauthored case, the simplest fix is to drop `_interpret_usd_max_hull_vertices` and pass the resolved value straight through. If you would rather keep it as a safety net, it needs the authored flag instead of the resolver identity, because only an unauthored value should ever be reinterpreted.

Worth adding a regression test for authored `-1` next to the existing authored `0` case at `newton/tests/test_import_usd.py:11470`.

newton/_src/utils/import_usd.py
```diff
@@ -3482,33 +3643,53 @@ def _record_fallback_collider_mass_information(
...
+                    comparison_key=lambda value, _resolver: (
+                        builder.rigid_gap
+                        if value is None or (value == float("-inf") and builder.default_shape_cfg.gap is None)
+                        else builder.default_shape_cfg.gap
+                        if value == float("-inf")
+                        else value
+                    ),
                 )
                 if gap_val == float("-inf"):
                     gap_val = builder.default_shape_cfg.gap
```
eric-heiden: 🟠 `gap` is the only shape key resolved without a `default=`, and with `use_applied_schema_fallbacks=True` that makes `builder.default_shape_cfg.gap` unreachable.

With the flag on, the registered `newton:contactGap = -inf` fallback is dropped by `fallback_is_unset=_fallback_is_negative_infinity` (`newton/_src/usd/schemas.py:306`), there is no importer default to fall back to because none is passed at line 3668, and Newton's compatibility default is skipped since `_use_legacy_unowned_defaults = False`. `_resolve_value` therefore returns `None`, the `== float("-inf")` test below never fires, and `gap=None` reaches `ShapeConfig`, where `ModelBuilder` substitutes `rigid_gap`.

Setting `builder.default_shape_cfg.gap = 0.015` and then importing a shape with `NewtonCollisionAPI` applied:

| `use_applied_schema_fallbacks` | resulting `shape.gap` | expected |
| --- | --- | --- |
| `False` | `0.015` | `0.015` |
| `True` | **`0.1`** | `0.015` |

The adjacent `margin` lookup is unaffected precisely because it passes `default=builder.default_shape_cfg.margin` on line 3664.

This is easy to hit, since the changelog tells users to "Pass `use_applied_schema_fallbacks=True` to adopt the new precedence", and around 16 files in the repo set a custom gap before importing USD, including `newton/examples/robot/example_robot_allegro_hand.py:85`.

There is also no migration warning for it. The new `comparison_key` lambda on line 3673 does model the right effective value, but `_record_legacy_fallback` returns at `if resolved.resolver is None or resolved.authored` (`newton/_src/usd/schema_resolver.py:757`) before that comparison is ever reached, and `resolved.resolver` is exactly `None` in this case. So the audit stays quiet and the contact gap changes silently.

Normalizing `None` alongside the sentinel keeps the fix local:

```suggestion
                if gap_val is None or gap_val == float("-inf"):
                    gap_val = builder.default_shape_cfg.gap
```

Passing `default=` on the lookup at line 3668, the way `margin` does, would work as well and would be more consistent with the surrounding keys.

newton/_src/utils/import_usd.py
eric-heiden: Confirmed on `6c928e73`, this is addressed.

`use_applied_schema_fallbacks=False` is now explicitly deprecated rather than presented as a neutral flag:

- the argument carries its own `.. deprecated:: 1.6` note in the docstring (`newton/_src/sim/builder.py:3514`, mirrored in `newton/_src/utils/import_usd.py:446`)
- there is a `changelog/3307.deprecated.md` fragment saying the same thing
- the legacy path now emits a `DeprecationWarning` when registered-schema precedence would change the resolved value or its source, and it names the affected schema properties and prim paths

Verified on an import with `NewtonCollisionAPI` applied and the property left unauthored:

```
This import used deprecated legacy precedence for applied but unauthored USD schema
properties; schema fallbacks will take precedence for NewtonCollisionAPI
(newton:contactMargin) on /Body/Collider. In a future release, applied-schema fallbacks
will take precedence; pass use_applied_schema_fallbacks=True to adopt that behavior now,
or author the intended values explicitly to preserve them.
```

That covers both halves of what I asked for, so resolving this thread.

newton/_src/utils/import_usd.py
eric-heiden: Fixed on `6c928e73`, resolving.

`_resolve_joint_velocity_limit` no longer passes `legacy_value_transformer`. It now passes `default=_ImporterDefault(None)` and applies `_interpret_usd_joint_velocity_limit` after the call returns (`newton/_src/utils/import_usd.py:940-953`), leaving `comparison_key` for the audit path only. That is the second option I suggested, and it means the resolver sees the untransformed value when it decides whether to report anything.

Verified with `verbose=True` on a stage with three revolute joints and `newton:velocityLimit` unauthored:

| policy | spurious `joint:velocity_limit` error lines | resolved limit |
| --- | --- | --- |
| `use_applied_schema_fallbacks=False` | 0 | `1e6` (builder default) |
| `use_applied_schema_fallbacks=True` | 0 | `1e6` (builder default) |

No error lines on either policy, and the value still resolves correctly.

newton/_src/sim/builder.py
```diff
@@ -3482,7 +3503,17 @@ def add_usd(
...
+                    arguments may change without prior notice.
+            use_applied_schema_fallbacks: If True, use registered schema fallbacks
+                before importer defaults. False retains deprecated legacy precedence
+                and warns when the effective imported value or source would change.
+                Unregistered resolver defaults remain compatibility defaults after importer
+                defaults.
+
+                .. deprecated:: 1.6
+                    Passing False selects deprecated legacy fallback precedence. Pass
+                    True to adopt registered schema fallback precedence.
```
eric-heiden: ⚪ The `.. experimental::` note that names `use_applied_schema_fallbacks` is nested inside the `schema_resolvers` entry, so Sphinx renders it under `schema_resolvers` only.

The admonition sits at the `schema_resolvers` continuation indent (16 spaces), and `use_applied_schema_fallbacks` only opens its own `Args:` entry on the following line, at the 12-space argument indent. Google-style parsing therefore attaches the note to `schema_resolvers`. The `.. deprecated:: 1.6` block a few lines down is scoped correctly to its own entry, which confirms the intended nesting depth, so this reads as a placement slip rather than a deliberate choice.

It matters because `use_applied_schema_fallbacks` is the argument this change asks users to adopt, including in two shipped examples (`newton/examples/robot/example_robot_g1.py:52` and `example_robot_h1.py:52`). Someone reading that entry in the rendered docs never sees that the argument may change without notice.

Giving each argument its own admonition also lets each note stand on its own, without the shared cross-reference:

```suggestion
                    The ``schema_resolvers`` argument may change without prior notice.
            use_applied_schema_fallbacks: If True, use registered schema fallbacks
                before importer defaults. False retains deprecated legacy precedence
                and warns when the effective imported value or source would change.
                Unregistered resolver defaults remain compatibility defaults after importer
                defaults.

                .. experimental::

                    The ``use_applied_schema_fallbacks`` argument may change without
                    prior notice.

                .. deprecated:: 1.6
                    Passing False selects deprecated legacy fallback precedence. Pass
                    True to adopt registered schema fallback precedence.
```

The same docstring is duplicated at `newton/_src/utils/import_usd.py:436-439`, so it needs the same split.

eric-heiden:
Thanks for the recent updates here. The sentinel handling reads much better now, and the two threads I had open about the deprecation wording and the velocity limit transform are addressed, so I resolved those. Here is a short list of what I think is still left.

- 🟠 **`builder.default_shape_cfg.gap` gets lost when `use_applied_schema_fallbacks=True`.** `gap` is the only shape property resolved without a `default=`, so under the new precedence it comes back as `None` and the builder substitutes `rigid_gap` (0.1) instead. Anyone who sets `default_shape_cfg.gap = 0.015` and then opts in, as the changelog suggests, quietly gets a 0.1 m contact gap. Passing the importer default explicitly, the way the neighbouring `margin` lookup already does, should be enough.

- 🟠 **An authored `newton:maxHullVertices = -1` becomes 64.** The schema documents `-1` as "use as many vertices as necessary to produce a perfect convex hull", but `_interpret_usd_max_hull_vertices` cannot tell an authored `-1` from an unauthored schema fallback and rewrites both. This one is not gated behind the flag, it also happens on the default path. Since `fallback_is_unset=_fallback_is_negative_one` already handles the unauthored case, applying the rewrite only to non authored values would do it.

- 🔵 **`test_applied_sdf_effective_defaults_do_not_warn` does not check for warnings.** It asserts the effective values but nothing about `DeprecationWarning`, so the "do not warn" part of the name is not actually covered. A `warnings.catch_warnings()` block with `simplefilter("error", DeprecationWarning)` around the import would close that gap.

- ⚪ **Two small polish items.** The `.. experimental::` note that mentions `use_applied_schema_fallbacks` is indented under the `schema_resolvers` entry, so Sphinx renders it under `schema_resolvers` only (same in both copies of the docstring). And `legacy_value_transformer` no longer has any caller now that the velocity limit path stopped using it, so it can be dropped. It would also be nice to keep `_ImporterDefault` out of the rendered signature, since `add_usd` currently shows defaults like `<omitted; importer default=1.0>`.

Nothing here looks structural to me. Only the first two feel like merge blockers, the rest are easy follow ups. Happy to re-check once they are in.

mzamoramora-nvidia:
@eric-heiden , thanks for the feedback. I have pushed a reworked version, which should address all of the comments. 

I added a temporary [design note](https://github.com/mzamoramora-nvidia/newton/blob/9745d662588e5d8b78bd2d0213e1d5a13db643f3/docs/guide/usd_schema_fallback_resolution.md), which I plan to remove before merge. 

The PR is still large. Suggestions on how to simplify it are more than welcome.

docs/concepts/usd_parsing.rst
```diff
@@ -148,6 +148,10 @@ material that authors its values only under the vendor namespaces is still read
 resolver, with a ``DeprecationWarning``, so existing assets keep their stiffness and density
 during the deprecation window.
 
+.. note::
+
+   ``use_applied_schema_fallbacks`` does not yet change proposal deformable attribute reads. Those attributes continue to read canonical ``physics:`` values first, followed by resolver-enabled vendor namespaces. They do not yet read composed fallbacks from ``Usd.SchemaRegistry``. Cloth shell mass properties are the limited exception: their shared ``mass_model`` and ``shell_thickness`` properties already use the ordinary schema resolver. Applying the common registered-fallback policy to the remaining deformable properties is follow-up work.
```
eric-heiden: I don't know if this is a typo but things like "proposal deformable attribute reads" is hard to understand - do you mean proposed deformable schema attribute? This paragraph should be rewritten in simpler language.

newton/_src/usd/schema_resolver.py
```diff
@@ -72,6 +207,10 @@ class SchemaAttribute:
     # Name of the schema resolver
     name: ClassVar[str]
 
+    # Applied or typed schema that owns each mapping entry.
+    _schema_names: ClassVar[dict[PrimType, str | dict[str, str]]] = {}
+    _use_legacy_unowned_defaults: ClassVar[bool] = True
```
eric-heiden: 🔵 `SchemaResolver` is public API (re-exported from `newton/usd.py`, documented in `docs/api/newton_usd.rst`), and `add_usd(schema_resolvers=...)` is typed `list[SchemaResolver] | None`, so user subclasses are a supported extension path. But the entire new precedence is gated on the two private ClassVars added here, `_schema_names` and `_use_legacy_unowned_defaults`.

Two resolvers that are identical across the whole public surface (`name`, `mapping`) and differ only by the private ClassVar give different results on a prim with `NewtonSDFCollisionAPI` applied and nothing authored, both run with `use_applied_schema_fallbacks=True`:

| Resolver | Resolved `kh` |
|---|---|
| public API only | `500.0` (the resolver's own compatibility default) |
| same resolver plus private `_schema_names` | `1e10` (the registered schema fallback) |

The mechanism: with the inherited `_schema_names = {}`, `_schema_name()` returns `None`, so `_schema_is_applied()` is `False`, so the registered-fallback branch is never entered and resolution falls through to the compatibility default loop, where a custom resolver passes every skip clause and gets `spec.default`. So a custom resolver silently keeps legacy precedence even when the caller explicitly opted in, and nothing warns, because the migration audit only fires when the legacy and composed results diverge, which for such a resolver they never do. The two new public `SchemaAttribute` fields, `fallback_is_unset` and `angular_unit`, are inert for custom resolvers for the same reason: `fallback_is_unset` is only consulted inside the branch that private ownership gates.

The test suite reads like confirmation that ownership really is part of the extension contract: the resolvers defined in `newton/tests/test_schema_resolver.py` all set `_schema_names` to get the behavior under test.

Suggested direction: promote schema ownership to public API before this ships as the future default, for example a public `schema_names` ClassVar, or a `SchemaAttribute(schema=...)` field that keeps ownership right next to `fallback_is_unset` where authors already look, with `_schema_names` kept as a deprecated private alias. If keeping the gate private for one release is deliberate, please state that in the `schema_resolvers` docstring and in the experimental note, so custom resolver authors know they are not yet participating in registered-fallback precedence.

newton/examples/robot/example_robot_h1.py
```diff
@@ -49,6 +49,7 @@ def __init__(self, viewer, args):
             asset_file,
             ignore_paths=["/GroundPlane"],
             enable_self_collisions=False,
+            use_applied_schema_fallbacks=True,
```
eric-heiden: 🔵 Opting in here silently discards this example's own `default_joint_cfg.friction = 1e-5` on 19 of its 25 DOFs.

19 joint prims in `h1.usda` carry `NewtonJointAPI`, and `Usd.SchemaRegistry().FindAppliedAPIPrimDefinition("NewtonJointAPI").GetAttributeFallbackValue("newton:friction")` returns `0.0`. Under the new precedence the registered fallback outranks the importer default, so the configured friction is dropped on exactly those joints. The remaining 6 DOFs carry no `NewtonJointAPI` and keep `1e-5`.

Running this example's exact `add_usd()` configuration against the merge base:

| Metric | main | this PR, `=True` (as shipped here) | this PR, `=False` |
|---|---|---|---|
| `joint_friction` sum over 25 DOFs | 1.9e-4 | 0.0 | 1.9e-4 |
| DOFs changed from 1e-5 to 0 | n/a | 19 | 0 |
| armature, limit_ke, limit_kd, velocity_limit, target_ke, target_kd | baseline | identical | identical |
| shape count / joint DOF count | 54 / 25 | 54 / 25 | 54 / 25 |

The g1 example is unaffected, because its joints author friction explicitly and authored values still win.

At 1e-5 the physical effect is negligible, so this is not a merge blocker. It is still worth addressing: it is an unintended behavior change in shipped example code, and it illustrates the general hazard the new precedence introduces, namely that any caller who tunes `default_joint_cfg` and opts in loses those settings on every prim carrying the matching applied schema. Either drop `friction` from this example's joint config so the code matches what actually happens, or call the interaction out in the migration docs.

newton/_src/usd/schema_resolver.py
```diff
@@ -218,65 +709,755 @@ def _collect_on_first_use(self, resolver: SchemaResolver, prim: Usd.Prim) -> Non
...
+        active_default = self._active_default(default, legacy_default)
+        self._report_missing(
+            prim,
+            prim_type,
+            key,
+            resolved.value,
+            active_default,
+            verbose and override is _NO_OVERRIDE,
+        )
+        return self._InterpretedPolicyValue(resolved, resolved.comparison)
```
eric-heiden, line 932: 🔵 `_get_interpreted_value()` passes the raw resolved value and the raw `default` to `_report_missing()`, and `_report_missing` prints an `Error: Cannot resolve value for ...` line whenever the resolved value is `None`, no `_ImporterDefault` was supplied, and `verbose` is set.

With `add_usd(..., use_applied_schema_fallbacks=True, verbose=True)`, the combination this PR asks users to adopt, that produces three bogus error lines for every collision shape: `shape:sdf_narrow_band_inner`, `shape:sdf_narrow_band_outer`, and `shape:kh`. The resolved values are correct, since `interpret_narrow_band` and `interpret_stiffness` substitute the builder default. Only the diagnostic is wrong.

Why it happens: those three keys are requested via `_get_interpreted_value(..., verbose=self._verbose)` with no `default=` in `_usd_resolution_policy.py` (lines 686-704 and 880-886), so `default` is `None` and `_importer_default(None)` returns `(False, None)`. On the opt-in path `SchemaResolverNewton._use_legacy_unowned_defaults` is `False`, so `_resolve_value` skips the compatibility default and returns `UNRESOLVED` with `value=None` when `NewtonSDFCollisionAPI` is not applied. On the legacy path the same keys resolve to the compatibility default `float("-inf")`, so nothing prints. This is specific to the new path.

Two ways to fix it: pass `default=_ImporterDefault(<builder default>)` at those three call sites so the resolver knows an importer default exists, or skip `_report_missing` in `_get_interpreted_value` when the interpreter produced a usable value, which is what the sibling `_resolve_interpreted_policies` already does.

newton/_src/sim/builder.py
```diff
@@ -3365,6 +3366,7 @@ def add_usd(
         parse_mujoco_options: bool = True,
         mesh_maxhullvert: int | None = None,
         schema_resolvers: list[SchemaResolver] | None = None,
+        use_applied_schema_fallbacks: bool = False,
```
eric-heiden: ⚪ The new argument is named `use_applied_schema_fallbacks`, but every piece of prose describing it, including its own docstring, calls the mechanism *registered* schema fallbacks:

- `changelog/3307.changed.md`: "Add opt-in registered USD schema fallback precedence"
- `changelog/3307.deprecated.md`: "pass `True` to adopt registered USD schema fallback precedence"
- its own docstring here and the copy in `import_usd.py`: "authored value and registered schema fallback"
- `docs/concepts/usd_parsing.rst`: "Registered schema fallback", "registered-schema precedence"
- the runtime `DeprecationWarning`: "registered schema fallbacks will be considered before importer defaults"
- the internals: `_ValueSource.REGISTERED_FALLBACK`, `_registered_schema_fallbacks`, `_registered_attribute_fallbacks`

The name is also narrower than the behavior. `_schema_is_applied` matches typed schemas as well as applied API schemas (`str(prim.GetTypeName()) == schema_name or usd.has_applied_api_schema(...)`), and `_schema_fallback` calls `FindConcretePrimDefinition` for the typed case, so "applied" covers only half of what the flag does.

The flag is still experimental and headed for a deprecation-driven default flip, so the rename is at its cheapest right now:

```suggestion
        use_registered_schema_fallbacks: bool = False,
```

The same rename would need to follow in `import_usd.py`, the docstrings, `docs/concepts/usd_parsing.rst`, and the changelog entries. If the current spelling has already circulated, accepting it for one release with a `DeprecationWarning` would cover existing callers.

newton/_src/usd/schema_resolver.py
```diff
@@ -189,20 +481,297 @@ def _collect_attrs_by_namespace(prim: Usd.Prim, namespaces: Sequence[str]) -> di
...
+        schema_is_applied: Callable[[SchemaResolver, str], bool],
+        read_fallback: Callable[[SchemaResolver, str], Any],
+        prim_type: PrimType,
+        key: str,
+        *,
+        default: Any = None,
+        authored_aliases: Sequence[str] = (),
+    ) -> _ResolvedValue:
+        compatibility_fallbacks: set[int] = set()
+        for resolver in self._resolvers:
```
jcarius-nv, line 505: Was this precedence order already decided elsewhere? The current resolver-first loop allows an earlier resolver's registered fallback to mask a later resolver's explicitly authored value. For example, an unauthored `newton:armature` fallback can beat authored `mjc:armature`. I had expected: explicit override -> authored candidates in resolver order -> registered fallbacks in resolver order -> importer default -> compatibility default. Issue #3307 requires consistency but does not appear to prescribe resolver-first source authority. If this was already decided, could we link the rationale and the mixed-schema use case that requires it? Otherwise, could we settle that ordering before making it the future default?

changelog/3307.changed.md
```diff
@@ -0,0 +1 @@
+Add opt-in registered USD schema fallback precedence to `ModelBuilder.add_usd()`, with one migration warning when future precedence changes an interpreted property or its meaning. Pass `use_registered_schema_fallbacks=True` to adopt the new precedence; see the USD parsing guide for ways to preserve current results.
```
jcarius-nv: Could the permanent guide include one concrete before/after example showing the same mixed-schema asset and resolver order under legacy versus registered precedence? The changelog should also surface the key behavioral change in one sentence rather than only directing users to enable the option. This should follow the final resolution of the authored-versus-fallback ordering.

docs/guide/usd_schema_fallback_resolution.md
```diff
@@ -0,0 +1,131 @@
...
+
+Legacy imports resolve the registered-schema result using cached reads and emit at most one aggregated `DeprecationWarning` per import. The warning is emitted only for a proven interpreted value or relevant source change; audit failures alone do not warn. It identifies the old and new suppliers, bounds the list of prim paths, and explains how to adopt or preserve the behavior.
+
+An explicit importer override suppresses migration auditing for that property because the caller has already selected its future value.
+
+## Performance
+
+The policy extraction adds no stage traversal and does not build a second model. Registered-schema mode evaluates only the selected policy. Legacy mode also evaluates the future result for migration diagnostics, but authored reads and composed schema definitions are cached for the import.
+
+Local before-and-after timings should cover a large authored-value stage and a fallback-heavy stage. This is an engineering check rather than a committed benchmark or public performance guarantee.
```
jcarius-nv, line 110: Could we include import timings for a representative large authored-heavy asset and a fallback-heavy asset under both policies? In particular, the legacy path performs the registered-policy shadow audit, so the existing green benchmark check does not establish the overhead introduced on normal imports.

newton/_src/usd/schema_resolver.py
```diff
@@ -72,6 +224,12 @@ class SchemaAttribute:
     # Name of the schema resolver
     name: ClassVar[str]
 
+    #: Schema ownership by prim type. A schema name owns every mapped key for
+    #: that prim type. A key-to-schema mapping declares ownership per key.
+    schema_names: ClassVar[Mapping[PrimType, str | Mapping[str, str]]] = {}
```
jcarius-nv: `schema_names` does more than identify schemas: it grants fallback ownership for logical properties. Could this contract remain private until the source-neutral provenance work in #3984, or otherwise use a name that communicates authority, such as `schema_ownership`? This seems worth settling before custom resolvers depend on the experimental public spelling.

jcarius-nv (commented):
Two broader design questions from this review:

**Landing structure**

Would you consider factoring this into stacked PRs: first a semantic-neutral `_UsdResolutionPolicy` extraction, then the private registered-schema selection engine and tests, and finally public integration/migration? This would let us review the durable algorithm independently from the temporary audit machinery and public API. If exact per-property migration warnings are not a firm requirement, a general deprecation warning or on-demand audit could also avoid running two fully interpreted policies on every legacy import.

**Schema dependency contract**

Registered fallbacks make installed schema versions part of Newton's import semantics, but the dependency contract is not yet clear. `newton-usd-schemas>=0.5.0` is unbounded, while PhysX and MuJoCo schemas are not registered by a normal Newton installation and are currently exposed through much broader packages. Before registered precedence becomes the default, could we define required versus optional schemas, supported version ranges, early-registration ownership, missing-plugin behavior, and who publishes lightweight PhysX/MuJoCo schema-only wheels?

newton/_src/usd/schema_resolver.py
mzamoramora-nvidia: Thanks, this is a good point. #3888 inherited the resolver-first order proposed in #3572. 

The motivation was summarized in [this comment on #3307](https://github.com/newton-physics/newton/issues/3307#issuecomment-5050177784):

> Precedence is: per resolver in priority order, authored wins, otherwise an applied schema's fallback wins and stops fallthrough — the same result native `Get()` produces on registered schemas.

The suggested: `explicit override -> authored candidates in resolver order -> registered fallbacks in resolver order -> importer default -> compatibility default` also makes sense and shouldn't be difficult to implement but we need to decide which approach to follow.

mzamoramora-nvidia:
About the landing structure: yes, we could split this into stacked PRs. Let's first settle the schema resolver precedence and address the remaining concerns on this branch. Once that is clear, we can split the work into smaller PRs with stable boundaries.

newton/_src/usd/schema_resolver.py
jcarius-nv: Related to that decision: Andrew's proposal in #3307 was to rely on USD's existing mechanisms for registered schemas—`GetResolveInfo()` for authored versus fallback provenance, `HasAPI()` for applicability, and prim definitions for schema fallbacks—with custom applicability only as the second-class unregistered path. Could we clarify how much of `_SchemaResolutionPolicy` remains necessary for registered schemas, and whether native USD resolution can be the first-class path while custom selection is confined to unregistered or legacy schemas? The "same result native `Get()` produces" statement describes resolution of one USD attribute; here Newton is also establishing precedence between different attributes such as `newton:armature` and `mjc:armature`, which USD itself does not order.

mzamoramora-nvidia: Yes, I think we can simplify `_SchemaResolutionPolicy`. But we also want to keep it independent of PXR: Newton currently reads a PXR `Usd.Stage`, while OVNewton reads an OVStage. Both paths should provide the policy with the same basic information: the value, where it came from, and whether its schema applies.

When resolving a Newton value, we mostly have three kinds of input:

- **Registered schema:** The schema plugin and definition are available to USD. Newton's PXR path can use `HasAPI()`, `Get()`, and `GetResolveInfo()` to obtain the value, source, and applicability directly from USD.
- **Unregistered schema:** USD can still return authored namespaced attributes, but without a registered plugin it has no schema definition or authoritative fallback. Newton therefore needs resolver-specific rules to decide when the mapping applies. A resolver may carry an old hard-coded default, which Newton uses only when neither the USD file nor the importer provides a value.
- **Importer argument:** This value comes from `add_usd()`, not USD. A value passed explicitly is an override, while the value used when the argument is omitted is an importer default.

Even if native USD does more of the work, we still need a small Newton policy for decisions that neither USD nor OVStage should make:

- Choose between different attributes that represent the same Newton setting, such as `newton:armature` and `mjc:armature`.
- Order explicit importer overrides, USD values, importer defaults, and old resolver defaults.
- Skip registered fallback values that mean "keep looking," such as `-inf` for several contact and SDF settings, `+inf` for an unspecified velocity limit, and the registered `-1` fallback for maximum hull vertices. An explicitly authored `-1` must still be preserved.

At the same time, property conversions, such as converting MuJoCo `solreflimit` into stiffness and damping or converting between degrees and radians, can stay in shared property-specific Newton code rather than making `_SchemaResolutionPolicy` larger. With this split, Newton's PXR reader and OVNewton's OVStage reader can share the same smaller resolution policy without OVNewton opening and traversing the USD stage again.

jcarius-nv: that's a good point, especially keeping an ovnewton (or any other such data source) in mind 👍

docs/concepts/usd_parsing.rst
```diff
@@ -550,13 +562,84 @@ Priority-Based Resolution
...
 
 1. **Authored Values**: Resolvers are queried in priority order; the first resolver that finds an authored value on the prim returns it and remaining resolvers are not consulted.
 2. **Importer Defaults**: If no authored value is found, Newton's importer uses a property-specific fallback (e.g. ``builder.default_joint_cfg.armature`` for joint armature). This takes precedence over schema-level defaults.
-3. **Approximated Schema Defaults**: If neither an authored value nor an importer default is available, Newton falls back to a hardcoded approximation of each solver's schema default, defined in Newton's resolver mapping. These approximations will be replaced by actual USD schema defaults in a future release.
+3. **Resolver Compatibility Defaults**: If neither an authored value nor an importer default is available, Newton falls back to the resolver mapping's compatibility default.
+
+Pass ``use_registered_schema_fallbacks=True`` to enable schema ownership. Resolution then follows this order:
+
+1. **Resolver candidate**: For each resolver in priority order, use its authored value or the fallback from its applicable registered schema.
```
gyeomannvidia: I found this really quite hard to understand.  Part of that is that the text has a lot of technical terms like "property-specific importer and "compatibility default".  Part of it is just the complexity of the logic.

I think it makes sense to attempt a flow diagram setting out the sequence of decisions and outcomes. 

I mapped out pseudo-code to help my own understanding. Not sure if it is useful at all to help create a flow diagram.

```
def resolver_default_is_eligible(resolver, prim, key):
    # Q1: does the resolver specify a responsible schema for this key?
    schema_name = get_responsible_schema(resolver, key)
    if schema_name is None:
        return True   # unowned - no schema to check, resolver default is always OK

    # Q2: is the responsible schema applied to the prim?
    if not schema_is_applied(prim, schema_name):
        return False  # a real/claimed schema was expected but never invoked - don't guess

    # Q3: is the responsible schema registered in Usd.SchemaRegistry?
    if schema_is_registered(schema_name):
        return False  # a real, live schema was in play - it should have answered, don't guess
    else:
        return True   # applied, but not a real schema - safe to use the resolver default
		
def get_armature(prim, resolvers, registry, use_registered_schema_fallbacks):

	if !use_registered_schema_fallbacks

		# Look for an authored value. 
		for i in range (0, nb_resolvers):	
			resolver = resolvers[i]
			key = resolver.get_key("armature")		
			if prim_has_authored_value(prim, key):
				val = prim_get_authored_value(prim, key)
				return val				
				
		# Now look for an importer default.
		if builder.default_joint_cfg.armature != None:
			return builder.default_joint_cfg.armature
			
		# Now look for a default value in any resolver
		for i in range (0, nb_resolvers):	
			resolver = resolvers[i]
			key = resolver.get_key("armature")		
			if resolver_has_compatibility_default(resolver, key):
				val = resolver_get_compatibility_default(resolver, key)
				return val
				
	else:
	
		# Look for an authored value or a schema default value.
		# The schema is taken to be the schema registered as the 
		# responsible schema in the the resolver.
		for i in range (0, nb_resolvers):	
			resolver = resolvers[i]
			
			# Look for an authored value.
			key = resolver.get_key("armature")		
			if prim_has_authored_value(prim, key):
				val = prim_get_authored_value(prim, key)
				return val
				
			# Look for a schema default.
			responsible_schema_name = get_responsible_schema(resolver, key)	
			if responsible_schema_name is not None:
				if schema_is_applied(prim, responsible_schema_name):
					if responsible_schema_has_default_value_in_registry(responsible_schema_name, key, registry):
						val = get_default_value_from_registry(responsible_schema_name, key, registry)
						return val		
						
		# Now look for an importer default.
		if builder.default_joint_cfg.armature != None:
			return builder.default_joint_cfg.armature
					
		# Now look for a default value in any resolver.
		# Care must be taken here to make sure that the failure to find a schema 
		# default above was intentional.
		# We need to answer 3 questions.
		# Q1) does the resolver specify an API that is responsible for "armature"?
		# Q2) is the responsible API applied to the prim?
		# Q3) is the responsible API registered in the USD registry ie does Newton?
		# know about the API.

		# Q1: 	No. Use the resolver armature default. see note 1.
				Yes. Go to Q2

		# Q2:	No. Do not use the resolver armature default. see note 2.
				Yes. Go to Q3.
				
		# Q3:	No.  Use the resolver armature default. see note 3.
				Yes. Do not use the resolver armature default. see note 4.

		note 1: the prim did not signal an intent to use a schema default.
		note 2: the prim did not signal an intent to use the schema default of the 
		responsible API.
		note 3: the responsible API was applied to the prim, but Newton's USD registry
		does not recognize it as a real schema. There is no genuine schema authority
		to defer to, so the resolver's own guess is safe to use.
		note 4: the responsible API was applied to the prim and is a real, registered
		schema. It had the opportunity to answer (either the property was found already
		in an earlier step, or the registered schema simply has no fallback declared
		for it) — either way, a real authority was in play, so the resolver must not
		substitute its own guess in its place.
		
		# Q1, Q2, Q3 are coded in resolver_default_is_eligible().
				
		for i in range(0, nb_resolvers):
			resolver = resolvers[i]
			key = resolver.get_key("armature")
			if resolver.use_compatibility_defaults:
				if resolver_default_is_eligible(resolver, prim, key):
					if resolver_has_compatibility_default(resolver, key): 
						val = resolver_get_compatibility_default(resolver, key)
						return val
```

newton/_src/usd/schema_resolver.py
```diff
@@ -72,6 +224,12 @@ class SchemaAttribute:
...
+    #: Whether unregistered or unowned mappings may use compatibility defaults.
+    use_compatibility_defaults: ClassVar[bool] = True
```
gyeomannvidia: I had to ask claude to explain the meaning of compatibility default. I am not a USD or parsing expert so maybe use_compatibility_defaults is a well understood term. 

My understanding is that the flag determines whether to consider the default set in the resolver in the event that it could not find a value from USD or from the builder cfg.

Can we find a less ambiguous name here and say a bit more in the docstring?

docs/concepts/usd_parsing.rst
gyeomannvidia: I asked claude to make a flow diagram based on my pseudo-code.  I've attached the diagram and the python file used to create the diagram.


[draw_flow.py](https://github.com/user-attachments/files/31795478/draw_flow.py)

flow

docs/guide/usd_schema_fallback_resolution.md
mzamoramora-nvidia: Good point. Addressed partially in 201285b8. I made the audit opt-in so normal imports do not pay the cost of evaluating both policies. Local tests showed 3.7% overhead for registered precedence on Unitree G1 and 16% on a generated fallback-heavy stage with 500 shapes. More performance work may be needed, although I am not sure the synthetic case is representative of a real workload.

newton/_src/usd/schema_resolver.py
mzamoramora-nvidia: Agreed. Addressed in 1e0df85c. Schema ownership is private again as `_schema_ownership`, and I removed the public custom-resolver example. Existing custom resolvers keep their current behavior. We can review a public ownership contract together with the source-neutral API in #3984.

docs/concepts/usd_parsing.rst
mzamoramora-nvidia: Yeah, the pseudocode and diagram were really helpful. Addressed in 611a12de. The documentation now starts with a short overview and splits the detailed flow into two diagrams so the resolver-candidate and importer/mapping-default checks remain readable.

newton/_src/usd/schema_resolver.py
mzamoramora-nvidia: Agreed, the old name was not clear enough. Addressed in d31c10e1. I renamed `_use_compatibility_defaults` to `_use_mapping_defaults` and expanded the comment to explain that it controls whether defaults stored in the resolver mapping may be considered after the importer default.

docs/concepts/usd_parsing.rst
mzamoramora-nvidia: Addressed in 611a12de. I added a short overview and split the detailed flow into two diagrams so the registered-schema and mapping-default checks remain readable.

gyeomannvidia:
My main comment is that the logic governing USD parsing is really quite complicated.  This makes it hard to maintain but even harder to explain to users.
Do we have a document that sets out the full set of requirements?  If we do, can we revisit that to verify that everything in the doc remains a requirement?  For example, Newton is open source so could we limit the supported schema flavours and leave users to add their own separate parser.

PR #3892, opened 2026-08-12
Fix torsional/rolling friction defaults and units

andrewkaufman:
## Description

`SchemaResolverNewton` declared 0.25 and 0.0005 as the schema fallbacks for `newton:torsionalFriction` and `newton:rollingFriction`, but newton-usd-schemas defines 0.005 and 0.0001. That field is documented as "Default USD-authored value from schema" and is used when nothing is authored, so importing a material that applies `NewtonMaterialAPI` without setting either attribute produced coefficients **50x and 5x too large** — and inconsistent with the `ModelBuilder.ShapeConfig` defaults for the same quantities.

Both coefficients were also documented as `[dimensionless]`. They are length-dimensioned, and Newton's own consumers are unambiguous about it:

- `SolverXPBD` (`xpbd/kernels.py:2368,2381`) clamps an **angular** impulse to `lambda_n * mu`, where `lambda_n` is the normal **linear** impulse. Torque = force x length, so `mu` must be a length.
- `SolverMuJoCo` (`mujoco/kernels.py:2488`) writes them straight into `geom_friction` as `vec3f(mu, torsional, rolling)`, and MuJoCo documents both slots as having [units of length](https://mujoco.readthedocs.io/en/stable/computation/index.html) — "the diameter of the surface contact patch" and "the depth of the local deformation".
- `SolverKamino` does not implement either coefficient (`kamino/_src/core/materials.py:256,263` carry only a `TODO`), so it neither confirms nor contradicts.

The `ShapeConfig` defaults of 0.005 / 0.0001 are also byte-identical to MuJoCo's `friction="1 0.005 0.0001"`, which is only coherent if they are the same quantity.

The distinction is not cosmetic: a length-dimensioned coefficient has to scale with a stage's `metersPerUnit` while a dimensionless one must not, so the label decides what a consumer is obliged to do. Newton itself sidesteps that today by declining non-unit stages outright (`import_usd.py:577`), so nothing here changes scaling behaviour — but the label is what other consumers of the schema read.

Related: [google-deepmind/mujoco#3472](https://github.com/google-deepmind/mujoco/issues/3472), which covers unit and semantic errors on the MuJoCo side of the same schema. The wording of the units in the schema itself is corrected in [newton-physics/newton-usd-schemas#83](https://github.com/newton-physics/newton-usd-schemas/pull/83); this PR is correct independently of it, since it aligns Newton with the schema's *values* rather than its prose.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_schema_resolver --strict-warnings   # 34 tests, OK
uv run --extra dev -m newton.tests -k test_import_usd --strict-warnings        # 442 tests, OK
uv run --extra dev -m newton.tests -k test_import_mjcf --strict-warnings       # 264 tests, OK
```

`test_material_friction_attributes` already existed but asserted the incorrect values, so it codified the bug rather than catching it. It now compares the resolved fallback against the schema's own fallback instead of a literal, so the two cannot drift apart again. Reverting `schemas.py` alone fails it:

```
AssertionError: 0.0005 != 9.999999747378752e-05 within 7 places
```

## Bug fix

**Steps to reproduce:**

1. Author a `UsdShade.Material` with `NewtonMaterialAPI` applied.
2. Do not author `newton:torsionalFriction` or `newton:rollingFriction`.
3. Import the stage and read back the resolved coefficients.
4. Observe 0.25 and 0.0005 rather than the schema's 0.005 and 0.0001.

**Minimal reproduction:**

```python
from pxr import Usd, UsdShade
from newton._src.usd.schema_resolver import PrimType, SchemaResolverManager
from newton._src.usd.schemas import SchemaResolverNewton

material = UsdShade.Material.Define(stage, "/material").GetPrim()
material.ApplyAPI("NewtonMaterialAPI")

resolver = SchemaResolverManager([SchemaResolverNewton()])
print(resolver.get_value(material, PrimType.MATERIAL, "mu_torsional"))  # 0.25, expected 0.005
print(resolver.get_value(material, PrimType.MATERIAL, "mu_rolling"))    # 0.0005, expected 0.0001

# the schema's own fallbacks, for comparison
print(material.GetAttribute("newton:torsionalFriction").Get())  # 0.005
print(material.GetAttribute("newton:rollingFriction").Get())    # 0.0001
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected default torsional and rolling friction values when importing USD materials.
  * Updated friction units to meters, aligning documentation and material behavior with solver expectations.

* **Documentation**
  * Clarified the physical meaning and units of torsional and rolling friction settings.

* **Tests**
  * Improved validation to ensure imported values match the declared USD schema defaults.


newton/_src/sim/builder.py
```diff
@@ -481,9 +481,17 @@ class ShapeConfig:
         on the solver constructor for this field to take effect.
         """
         mu_torsional: float = 0.005
-        """The coefficient of torsional friction (resistance to spinning at contact point)."""
+        """The coefficient of torsional friction [m] (resistance to spinning at contact point).
+
+        Length-dimensioned rather than dimensionless: it scales a normal impulse into a
+        torsional one, and can be read as the diameter of the contact patch.
```
andrewkaufman: This is oddly worded, its like my robot wrote its thoughts & arguments rather than just documenting the unit

andrewkaufman: Agreed, that was editorialising rather than documenting. Trimmed to match the neighbouring `ShapeConfig` fields, which state the unit and stop:

```python
mu_torsional: float = 0.005
"""The coefficient of torsional friction [m] (resistance to spinning at contact point)."""
mu_rolling: float = 0.0001
"""The coefficient of rolling friction [m] (resistance to rolling motion)."""
```

So it is now the original wording plus `[m]`, in the same shape as `ka` ("The contact adhesion distance [m]."). The reasoning lives in the PR description and commit message, which is where it belongs.

`Model.shape_material_mu_torsional` and `mu_rolling` were already one-liners and are unchanged apart from the unit.

andrewkaufman:
@coderabbitai on the `metersPerUnit` suggestion — that scaling does not exist to test. `import_usd` refuses non-unit stages rather than converting them (`newton/_src/utils/import_usd.py:577-582`):

```python
if not math.isclose(linear_unit, 1.0):
    warnings.warn(
        "USD stages with non-unit linear units are not supported. "
        f"Set metersPerUnit to 1.0 before import. Found metersPerUnit={linear_unit}.",
    )
```

Nothing is scaled by `linear_unit` on the material path, so a test asserting scaled friction coefficients would be asserting behaviour that neither exists nor is intended. Adding one would lock in an expectation the importer explicitly declines to meet.

The `metersPerUnit` remark in the PR description is about why the *dimension* matters for the label, not a claim that Newton scales by it — I have reworded it to avoid the implication. If non-unit stage support is ever added, the two friction coefficients would need to scale along with every other length, and that is worth handling as one change rather than piecemeal here.

eric-heiden (approved):
Thanks, LGTM!

PR #3893, opened 2026-08-13
Expanded MPM solver names for consistency

Milad-Rakhsha-NV:
## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

- **New Features**
  - Added full-name options for conjugate-gradient, conjugate-residual, and generalized-minimal-residual solvers.
  - Existing abbreviated solver names remain supported.

- **Documentation**
  - Updated solver configuration and usage documentation to include the new names and aliases.
  - Added a changelog entry describing the expanded solver naming options.


eric-heiden (commented):
Thanks @Milad-Rakhsha-NV! What is the reason for having these expanded names? Generally I think it's better we only keep one value name, not support different ways of expressing the same setting. If this is not critical, can we close this?

Milad-Rakhsha-NV:
> Thanks @Milad-Rakhsha-NV! What is the reason for having these expanded names? Generally I think it's better we only keep one value name, not support different ways of expressing the same setting. If this is not critical, can we close this?

Hi Eric please see this https://github.com/newton-physics/newton-usd-schemas/pull/84
In the original schema PR the MPM solvers were inconsistent, and for USD we shouldn't have used all acronyms or just expanded the names. So because we need a way to map the expanded names to the correct runtime solver I create this

eric-heiden (approved):

PR #3898, opened 2026-08-13
Fix MJCF heightfield row order and uniform-data normalization

### LINKED ISSUE #3897: Heightfield elevation ignored by XPBD contact generation (collides a flat plane at min_z); MJCF path also ejects bodies spawned near the base when elevation is non-uniform
**Environment:** newton 1.5.0, warp 1.16.0, CPU device (macOS arm64,
Python 3.12). SolverXPBD.

## Symptom 1 — elevation surface never collides (native API and MJCF)

A sphere dropped onto a heightfield rests at `base + radius` regardless
of the elevation data. Repro with the native API, no MJCF involved:

```python
import warp as wp, numpy as np, newton
from newton import Heightfield

b = newton.ModelBuilder()
hf = Heightfield(data=np.full((9, 9), 0.5, dtype=np.float32),
                 nrow=9, ncol=9, hx=4.0, hy=4.0, min_z=0.0, max_z=0.8)
b.add_shape_heightfield(xform=wp.transform(), heightfield=hf)
bod = b.add_body(xform=wp.transform(wp.vec3(0.0, 0.0, 1.5), wp.quat_identity()), mass=1.0)
b.add_shape_sphere(body=bod, radius=0.1)
b.add_joint_free(bod)
b.gravity = wp.vec3(0.0, 0.0, -9.81)
m = b.finalize()
solver = newton.solvers.SolverXPBD(m, iterations=10)
s0, s1 = m.state(), m.state()
ctrl = m.control()
for _ in range(1500):
    s0.clear_forces()
    contacts = m.collide(s0)
    solver.step(s0, s1, ctrl, contacts, 0.002)
    s0, s1 = s1, s0
print(s0.body_q.numpy()[0][2])   # -> 0.100
```

Expected: the surface sits well above min_z (uniform 0.5 data in
[0, 0.8]), so the ball should rest near z ≈ 0.5–0.6.
Observed: **z = 0.100** — it rests on a flat plane at min_z.

Same result via the (non-deprecated) `CollisionPipeline`, via the MJCF
import path (`<hfield ... elevation="...">` inline data, which
`import_mjcf` parses correctly into `Heightfield`), with a box instead
of a sphere, and for ramp-shaped data. Rest-height probes at four
locations of an asymmetric ramp all landed at base + radius, where
MuJoCo (3.10) rests the same model on the profile (0.0–0.7 m). The
1.5.0 release notes wire the heightfield BVH into **raycast**; contact
generation appears not to consult the surface at all.

## Symptom 2 — violent ejection (MJCF path, non-uniform data,
near-base spawn)

With the MJCF importer, NON-uniform inline elevation plus a body
spawned near the base plane is ejected violently instead of resting:

```xml
<mujoco model="m">
  <asset>
    <hfield name="terrain" nrow="9" ncol="9" size="4 4 0.8 0.05"
            elevation="0 0 0 0 0 0 0 0 0  0 0 0 0 0 0 0 0 0
                       0 0 0 0 0 0 0 0 0  0 0 0 0 0 0 0 0 0
                       0 0 0 0 0 0 0 0 0
                       0.2 0.2 0.2 0.2 0.2 0.2 0.2 0.2 0.2
                       0.4 0.4 0.4 0.4 0.4 0.4 0.4 0.4 0.4
                       0.6 0.6 0.6 0.6 0.6 0.6 0.6 0.6 0.6
                       0.8 0.8 0.8 0.8 0.8 0.8 0.8 0.8 0.8"/>
  </asset>
  <worldbody>
    <geom type="hfield" hfield="terrain" pos="0 0 0" friction="0.05"/>
    <body name="ball" pos="-2 2 0.15"><freejoint/>
      <geom type="sphere" size="0.1" density="600" friction="0.8"/></body>
  </worldbody>
</mujoco>
```

Stepped exactly as above (XPBD, dt=0.002): |drift| reaches **37.6 m by
t = 0.1 s** and **2,522 km by t = 6 s**. Insensitive to dt (0.002 /
0.001 / 0.0005) and to the body type (a box at the same spot: 2,340 km).

Control experiments that do NOT eject:
- identical model with ALL-ZERO elevation: settles calmly (base rest);
- identical model with uniform non-zero elevation: settles (on the base
  plane per symptom 1);
- the same ramp data + near-base spawn through the NATIVE API: settles
  (base rest) — the ejection needs the MJCF path specifically;
- the same MJCF model with the spawn raised (drop from z = 1.5):
  settles at base + radius.

## Context

We run Newton as a cross-validation oracle beside MuJoCo for our own
engine's conformance suite; heightfield rows are currently excluded
because of this. Happy to test a fix branch — the ramp model above is
part of a standing corpus we re-run continuously.

CharlesKZW:
## Description

Fixes the two remaining defects behind #3897 (heightfield collision):

1. **MJCF heightfields import with mirrored row order (inline and PNG
   sources).** MuJoCo's row conventions are per source (verified against
   mujoco 3.10/3.11 with `mj_ray` probes on asymmetric fields): it
   reverses rows when parsing the inline `elevation` string and when
   loading PNG images (row 0 lands at *maximum* y, the image-top
   convention), but loads its custom binary format as stored (row 0 =
   minimum y, which already matches Newton's grid). The importer passed
   inline and PNG rows straight through, so that terrain was mirrored
   across the x axis; as a follow-on, bodies authored near the true low
   side spawned deep inside the mirrored terrain and were ejected at
   extreme velocity (the 2,500 km drifts in #3897). The importer now
   flips rows per source — inline and PNG flip, binary stays as stored.

2. **The importer now applies MuJoCo's elevation normalization
   exactly** — `(v - min) / (max - min)`, with constant data compiling
   to zeros (surface at the geom origin) — before constructing the
   `Heightfield`, preserving exact MuJoCo parity on import for all
   three sources. Per review, the `Heightfield` constructor's handling
   of uniform data is left byte-for-byte unchanged (zeros -> surface at
   `min_z`): reading the constant as an absolute height would have made
   `data` unit-bearing in one entry point only, diverged from the `data`
   setter, and split `SolverXPBD` from `SolverMuJoCo` on the same model.
   The documented contract (and the escape hatch: omit `min_z`/`max_z`
   to place a flat field at its value) is now stated explicitly in the
   docstring and pinned by a test.

Note for cross-referencing #3897: that issue was measured against the
1.5.0 wheel, where heightfield contact generation was not wired at all;
current `main` already resolves the general case. This PR addresses what
remains on `main`, and the new tests assert rest *heights* on non-uniform
elevation — a case the existing suite did not cover (its contact tests
use all-zero elevation, where a surface collapsed to `min_z` is
indistinguishable from a correct one).

Closes #3897.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_heightfield
uv run --extra dev -m newton.tests -k TestImportMjcfHeightfieldOrientation
```

The orientation regressions fail without the fix, with the exact
pre-fix symptoms; all focused tests pass under `-W error::UserWarning`:

- `test_hfield_row_order_per_source`: asserts the imported grids
  directly for all three sources (inline, custom binary, PNG — the PNG
  leg skips without Pillow), no simulation involved; a blanket
  (non-per-source) flip fails its binary leg, a missing flip fails the
  inline and PNG legs;
- `test_hfield_row_orientation_matches_mujoco`: end-to-end contact
  check on an asymmetric quadrant field (`0.100 not greater than 0.9:
  rows mirrored` without the fix);
- `test_hfield_constant_inline_data_matches_mujoco` and
  `test_heightfield_uniform_data_normalization_contract`: pin the
  constant-data normalization (surface at the geom origin / `min_z`,
  matching MuJoCo) and the auto-derived-range escape hatch.

Also cross-checked externally against MuJoCo 3.10 on an asymmetric-ramp
MJCF (a sphere on the flat region, a high-friction box holding on the
slope, a low-friction box sliding down it): with this PR, Newton's rest
positions and slide direction match MuJoCo's, and the ejection from
#3897 no longer reproduces.

## Bug fix

**Steps to reproduce:**

1. Build a `Heightfield` with uniform data and an explicit range, e.g.
   `data=0.5` everywhere, `min_z=0.0`, `max_z=0.8`.
2. Drop a sphere on it with `SolverXPBD` + `CollisionPipeline`.
3. The sphere rests at `min_z + radius` (0.1) instead of 0.6.

**Minimal reproduction:**

```python
import warp as wp
import numpy as np
import newton
from newton import Heightfield

b = newton.ModelBuilder()
hf = Heightfield(data=np.full((9, 9), 0.5, dtype=np.float32),
                 nrow=9, ncol=9, hx=4.0, hy=4.0, min_z=0.0, max_z=0.8)
b.add_shape_heightfield(heightfield=hf)
bod = b.add_body(xform=wp.transform(wp.vec3(0.0, 0.0, 1.5), wp.quat_identity()), mass=1.0)
b.add_shape_sphere(body=bod, radius=0.1)
b.gravity = wp.vec3(0.0, 0.0, -9.81)
m = b.finalize()

solver = newton.solvers.SolverXPBD(m, iterations=10)
pipe = newton.CollisionPipeline(m)
contacts = pipe.contacts()
s0, s1 = m.state(), m.state()
ctrl = m.control()
for _ in range(1500):
    s0.clear_forces()
    pipe.collide(s0, contacts)
    solver.step(s0, s1, ctrl, contacts, 0.002)
    s0, s1 = s1, s0
print(s0.body_q.numpy()[0][2])  # 0.100 before the fix; 0.600 after
```

For the row-order defect, load any MJCF with asymmetric inline
`elevation` (e.g. the ramp model in #3897) and compare rest positions
against MuJoCo: pre-fix terrain is mirrored across x, and the #3897
ball is ejected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected imported MJCF heightfields so terrain orientation matches MuJoCo.
  * Fixed low-side body ejection during heightfield collisions.
  * Corrected uniform heightfields with explicit bounds so surfaces retain their actual height instead of collapsing to the minimum.
  * Improved body settling behavior on heightfield terrain.

* **Tests**
  * Added regression coverage for heightfield surface placement, collision settling, uniform terrain, and imported terrain orientation.


CharlesKZW:
Thanks — fixed: the fragment now uses imperative present tense (`Fix ...`), matching the other entries. Head is 8c78848.

newton/_src/utils/import_mjcf.py
```diff
@@ -1318,6 +1318,14 @@ def parse_shapes(
...
                     elevation = np.zeros((nrow, ncol), dtype=np.float32)
 
+                # MuJoCo stores heightfield rows starting at MAXIMUM y (the
+                # image-top convention, shared by its PNG and custom-binary
+                # assets); Newton's grid puts row 0 at minimum y. Flip the
+                # rows so the imported terrain matches MuJoCo's orientation
+                # (verified against mujoco 3.10 with mj_ray probes on an
+                # asymmetric ramp).
+                elevation = np.ascontiguousarray(np.asarray(elevation)[::-1])
+
```
Miko997, line 1345: This changes constant inline MJCF heightfields away from MuJoCo. With `size="2 2 0.8 0.1"` and all elevations set to `0.5`, this branch imports normalized data as `0.625` and world height `0.5`. MuJoCo 3.11 compiles the same data to zeros and world height `0.0`. The native `Heightfield` fallback can remain separate, but the MJCF importer should preserve MuJoCo normalization. Could you add a regression with constant inline data and non-unit `size_z`?

newton/tests/test_heightfield.py
```diff
@@ -643,6 +643,48 @@ def test_rasterize_mesh_missed_rays_use_floor(self):
...
+            hy=4.0,
+            min_z=0.0,
+            max_z=0.8,
+        )
+        builder.add_shape_heightfield(heightfield=hfield)
+
+        sphere_body = builder.add_body(xform=wp.transform((0.0, 0.0, 1.0), wp.quat_identity()), mass=1.0)
+        builder.add_shape_sphere(body=sphere_body, radius=0.1)
+        builder.add_joint_free(sphere_body)
+        builder.gravity = wp.vec3(0.0, 0.0, -9.81)
```
Miko997: `add_body()` already creates a free joint and articulation, so this adds a second inconsistent joint. The focused test fails here under `-W error::UserWarning`. Please remove this call.

newton/tests/test_import_mjcf.py
```diff
@@ -10429,5 +10429,73 @@ def test_inline_rgba_overrides_primitive_material(self):
...
+            builder = newton.ModelBuilder()
+            builder.add_mjcf(model_path)
+            # (x, y, expected-high?) probes: the plateau covers (+y, -x).
+            probe_spots = [(-1.5, 1.5, True), (1.5, 1.5, False), (-1.5, -1.5, False)]
+            probes = []
+            for x_pos, y_pos, _high in probe_spots:
+                body = builder.add_body(xform=wp.transform(wp.vec3(x_pos, y_pos, 1.5), wp.quat_identity()), mass=1.0)
+                builder.add_shape_sphere(body=body, radius=0.1)
+                builder.add_joint_free(body)
+                probes.append(body)
```
Miko997, line 10499: Same issue for each probe: `add_body()` already creates a free joint. Removing this call avoids three warnings and keeps the test compatible with strict-warning runs.

Miko997 (commented):
I ran both focused tests on CPU. They pass normally, but the comments below show one MJCF behavior mismatch and a strict-warning failure.

newton/_src/utils/import_mjcf.py
CharlesKZW: Confirmed against mujoco 3.10 as well (constant data compiles to zeros, surface at the geom origin) — you're right that the importer must preserve that. Fixed in b102efd: the importer now applies MuJoCo's normalization exactly (including the constant→zeros case) before constructing the `Heightfield`, so the constructor's explicit-range handling of uniform data stays a native-API semantic and never reaches the MJCF path. Added the regression you asked for: constant inline data with `size_z=0.8`, asserting the sphere rests at the geom origin + radius.

newton/tests/test_heightfield.py
CharlesKZW: Removed in b102efd — thanks, `add_body()` indeed already creates the free joint. Both focused tests now pass under `-W error::UserWarning`.

newton/tests/test_import_mjcf.py
CharlesKZW: Removed for all three probes in b102efd; the focused tests run clean under `-W error::UserWarning`.

newton/_src/utils/import_mjcf.py
```diff
@@ -1318,6 +1318,27 @@ def parse_shapes(
...
+                else:
+                    elevation = np.zeros_like(elevation)
+
+                # MuJoCo stores heightfield rows starting at MAXIMUM y (the
+                # image-top convention, shared by its PNG and custom-binary
+                # assets); Newton's grid puts row 0 at minimum y. Flip the
+                # rows so the imported terrain matches MuJoCo's orientation
+                # (verified against mujoco 3.10 with mj_ray probes on an
+                # asymmetric ramp).
+                elevation = np.ascontiguousarray(elevation[::-1])
```
eric-heiden: 🟠 The row flip is right for inline XML and PNG, but it also lands on the custom binary branch, and MuJoCo loads that format without reversing rows. So this fixes the XML and PNG orientation while mirroring `.bin` terrain, which was already correct before this change.

Checked against MuJoCo 3.11 with a 4x4 field whose first two stored rows are high, `size="2 2 1 0.1"`, probing the surface at `y = +1.5` and `y = -1.5`:

| elevation source | MuJoCo high side | after this change |
| --- | --- | --- |
| inline `elevation` | `+y` | `+y`, matches |
| `file="terrain.bin"` | `-y` | `+y`, mirrored |

They differ because MuJoCo reverses rows when it parses the XML `elevation` string and when it loads a PNG, but its custom binary loader copies the payload straight through. For `.bin` the stored row 0 stays file row 0, which is minimum y. Newton's grid also puts row 0 at minimum y, which is why `load_heightfield_elevation` already lined up for binary files on `main`. `SolverMuJoCo` exports `hfield_src.data` unflipped as well, so the mirrored terrain carries into that path too.

Applying the flip per source, rather than after the three branches have merged, keeps all of them correct:

<details>

<summary>Suggested approach</summary>

```py
if hfield_asset["elevation"] is not None:
    elevation = hfield_asset["elevation"]
    flip_rows = True  # MuJoCo reverses the XML elevation string
elif hfield_asset["file"] is not None:
    elevation = load_heightfield_elevation(hfield_asset["file"], nrow, ncol)
    # PNG is top-to-bottom like the XML string; the custom binary
    # format is stored bottom-up and needs no flip.
    flip_rows = os.path.splitext(hfield_asset["file"])[1].lower() == ".png"
else:
    elevation = np.zeros((nrow, ncol), dtype=np.float32)
    flip_rows = False

# ... normalization stays as it is ...

if flip_rows:
    elevation = np.ascontiguousarray(elevation[::-1])
```

</details>

It would also help to reword the comment above, since the image-top convention applies to the XML and PNG assets but not to the custom binary one.

newton/_src/geometry/types.py
```diff
@@ -2229,6 +2231,13 @@ def __init__(
         # Normalize data to [0, 1]
         if d_max > d_min:
             self._data = (raw - d_min) / (d_max - d_min)
+        elif min_z is not None and float(max_z) > float(min_z):
+            # Uniform data with an explicit height range: keep the flat
+            # surface at its true height inside [min_z, max_z] instead of
+            # collapsing it to min_z (uniform data has no range of its own
+            # to normalize by).
+            t = (d_min - float(min_z)) / (float(max_z) - float(min_z))
```
eric-heiden: 🟠 For uniform data with an explicit range this reads the constant as an absolute world Z, but everywhere else in this class `data` is unitless. The class docstring and the `data` argument both still say values are normalized to `[0, 1]` with `z = min_z + data[r, c] * (max_z - min_z)`, and the `data` setter just below still normalizes the old way, so the same array now produces different internal state depending on which entry point you use.

The larger problem is that the two solvers stop agreeing. With `Heightfield(data=np.full((9, 9), 0.5), nrow=9, ncol=9, hx=4.0, hy=4.0, min_z=0.0, max_z=0.8)` the data normalizes to `0.625` and the surface lands at `0.5`. Dropping a 0.1 m sphere onto it:

| solver | rest z | implied surface |
| --- | --- | --- |
| `SolverXPBD` with `CollisionPipeline` | `0.6001` | `0.5` |
| `SolverMuJoCo` | `0.0996` | `0.0` |

`SolverMuJoCo` exports `hfield_src.data` as `userdata` with `size_z = max_z - min_z`, and MuJoCo re-normalizes elevation when it compiles the model, so constant data becomes zeros and the surface falls back to `min_z`. On `main` both solvers agreed. The new `test_heightfield_uniform_data_explicit_range_rest_height` case pins the XPBD number only, so this disagreement is not caught.

The mapping is also discontinuous. A constant `0.5` field sits at `0.5`, the same field plus a tiny bit of noise gets stretched across the full range, and a constant outside `[min_z, max_z]` is silently clipped.

Since the importer now reproduces MuJoCo's normalization itself, the MJCF half of this PR no longer depends on this branch. Dropping it keeps `data` unitless, and someone who wants a flat surface at height `h` can already pass `min_z=max_z=h`. If the absolute reading is intentional, it would need to be stated in the class docstring and the `data` argument, mirrored in the setter, and folded into the `SolverMuJoCo` export so both solvers place the surface at the same height.

newton/tests/test_import_mjcf.py
```diff
@@ -10493,5 +10493,121 @@ def test_inline_rgba_overrides_primitive_material(self):
         np.testing.assert_allclose(builder.shape_color[0], [0.0, 1.0, 0.0], atol=1.0e-6)
 
 
+class TestImportMjcfHeightfieldOrientation(unittest.TestCase):
```
eric-heiden, line 10459: 🔵 This pins the row order for the inline `elevation` attribute only, but the importer has three elevation sources and they do not share one row convention. MuJoCo reverses rows for the XML string and for PNG, and stores the custom binary format as-is, so covering one source leaves the other two unverified.

The existing binary heightfield tests assert resolved paths and `shape_count` only, and there is no PNG heightfield test at all, so a mirrored `.bin` import would pass the whole suite today.

A data level assertion per source would be cheap and would pin the contract exactly, with no simulation involved:

<details>

<summary>Suggested test shape</summary>

```py
# Write a small .bin and a small PNG with a known asymmetric pattern,
# then check the imported grid directly.
builder.add_mjcf(model_path)
data = np.asarray(builder.shape_source[0].data)
# Newton grid: row 0 is minimum y, column 0 is minimum x.
np.testing.assert_allclose(data, expected)
```

</details>

Both new tests also read orientation back through 750 to 1000 XPBD steps, so they pin it only as tightly as the contact solver's rest height. Asserting the grid directly would make them faster and more precise, and the same style would tighten the inline XML case.

newton/_src/utils/import_mjcf.py
CharlesKZW: Verified against mujoco 3.10 as well — same result as your 3.11 table: the identical stored pattern lands high at −y for `.bin` and high at +y for inline, so the blanket flip was indeed mirroring binary terrain that was correct before. Fixed in 3293271 with your suggested shape: the flip is now decided per source (inline and PNG flip, binary stays as stored, matching `load_heightfield_elevation`'s own dispatch). Good catch on the `SolverMuJoCo` export side carrying the native-`Heightfield` convention unflipped — that looks like a separate seam from the importer; happy to file it as its own issue if useful.

newton/_src/geometry/types.py
CharlesKZW: You're right — reading the constant as an absolute height made `data` unit-bearing in one entry point only, diverged from the `data` setter, and split `SolverXPBD` from `SolverMuJoCo` on the same model. Reverted in 3293271: the constructor's normalization is byte-for-byte back to `main` (uniform → zeros → surface at `min_z`), and the behavior is now stated explicitly in the docstring, with the auto-derived-range escape hatch for placing a flat field at its value. `test_heightfield_uniform_data_normalization_contract` pins both halves of the contract, and the MJCF parity case is covered by the importer-side normalization plus its own regression.

newton/tests/test_import_mjcf.py
CharlesKZW: Added in 3293271: `test_hfield_row_order_per_source` asserts the imported grids directly for all three sources (inline, `.bin`, PNG — the PNG leg skips without Pillow), no simulation involved. It fails both ways: a blanket flip trips the binary leg, a missing flip trips inline and PNG. I kept the quadrant simulation test alongside as an end-to-end contact check, since the corpus that surfaced #3897 was contact-level.

eric-heiden (approved):
Thanks!

PR #3899, opened 2026-08-13
Retry transient asset download errors

### LINKED ISSUE #3774: [BUG] Test asset downloads have no retry, so one transient HTTP 5xx from GitHub fails a whole CI job
## Bug Description

`download_git_folder()` in `newton/_src/utils/download_assets.py` performs its network fetch exactly once. When GitHub returns a transient error, the exception is re-raised as a `RuntimeError` and whichever test triggered the download errors out, failing the entire CI job even though every other test in the run passed.

This hit two jobs within four minutes on the same PR, on different runners and against two different upstream repositories, which points at a short GitHub-side degradation rather than anything in the code under test:

| Job | Tests run | Repo being fetched | Test that errored | git stderr |
|---|---|---|---|---|
| GPU Unit Tests (AWS EC2) | 5205, `errors=1` | `mujoco_menagerie` @ `affef083` | `setUpClass (test_menagerie_mujoco.TestMenagerie_AnyboticsAnymalC)` | `HTTP 503` then `fatal: error reading section header 'shallow-info'` |
| newton-unittests (windows-latest) | 4183, `errors=1` | `newton-assets` @ `261cd1f4` | `test_import_usd.TestImportSampleAssetsBasic.test_g1` | `HTTP 503` then `fatal: expected 'packfile'` |

Both tracebacks bottom out at the same unprotected call:

```
File "newton/_src/utils/download_assets.py", line 379, in download_git_folder
    repo.git.fetch("origin", ref, "--depth=1", "--filter=blob:none")
...
git.exc.GitCommandError: Cmd('git') failed due to: exit code(128)
  cmdline: git fetch origin affef0836947b64cc06c4ab1cbf0152835693374 --depth=1 --filter=blob:none
  stderr: 'error: RPC failed; HTTP 503 curl 22 The requested URL returned error: 503
fatal: error reading section header 'shallow-info''

The above exception was the direct cause of the following exception:
...
File "newton/_src/utils/download_assets.py", line 420, in download_git_folder
    raise RuntimeError(f"Git operation failed: {e}") from e
RuntimeError: Git operation failed: ...
```

## Why This Is Worth Fixing

- One dropped connection throws away a 22 minute GPU run and a 28 minute Windows run.
- Every occurrence costs a maintainer a manual re-run plus the runner time to redo the whole suite.
- It teaches reviewers to wave off red CI as "probably the asset server again", which is how a real regression eventually slips through.
- The module already establishes the pattern. `_safe_rename()` (line 52) retries up to 5 times with a delay to absorb transient filesystem races. The network calls, which fail far more often than a rename does, get no such treatment.

## Suggested Fix

1. **Add a bounded retry with exponential backoff** around the two network operations in `download_git_folder()`: `repo.git.fetch(...)` (line 379) and `gitpython.Repo.clone_from(...)` (line 384). Three or four attempts with a growing delay would have absorbed both failures above.
2. **Only retry transient signatures.** Inspect the `GitCommandError` stderr and retry on `HTTP 5xx`, `RPC failed`, `early EOF`, `the remote end hung up`, `error reading section header`, and `expected 'packfile'`. Do not retry an unknown ref or an auth failure, because those never succeed and retrying only delays a legitimate error.
3. **Reset the temp directory between attempts.** A failed shallow fetch can leave a partial object store behind, so each retry should start from a clean `temp_dir` rather than resuming into the remains of the previous attempt.
4. **Optionally cache assets in CI.** The workflows already restore the Warp kernel cache (`ci.yml`, `aws_gpu_tests.yml`) but not the asset cache. Assets are pinned to fixed SHAs (`NEWTON_ASSETS_REF`, `MENAGERIE_REF`), so an `actions/cache` entry keyed on those SHAs covering `NEWTON_CACHE_PATH` would make most runs a cache hit and take the network off the common path entirely.

One clarification on the current fallback behavior: `_get_latest_commit_via_git()` (line 141) already swallows network errors and falls back to a cached copy, but when the ref is a pinned 40-character SHA it returns immediately without contacting the network at all. The fetch is therefore the only call that touches the network in the pinned-SHA path, and it is exactly the one with no protection.

## Reproduction Script

```python
# Requires a cold cache, e.g. NEWTON_CACHE_PATH pointed at an empty directory.
# A single simulated transient error is enough to lose the download,
# because there is no second attempt.
from unittest import mock

import git

import newton

transient = git.exc.GitCommandError(
    "git fetch",
    128,
    b"error: RPC failed; HTTP 503 curl 22 The requested URL returned error: 503\n"
    b"fatal: expected 'packfile'",
)

with mock.patch("git.Repo.init", side_effect=transient):
    newton.utils.download_asset("unitree_g1/usd")
    # RuntimeError: Git operation failed: Cmd('git') failed due to: exit code(128)
```

## System Information

- Reproduced on CI rather than locally, on two platforms at once:
  - `Run GPU Tests / Run GPU Unit Tests on AWS EC2`, Linux, Python 3.12
  - `run-newton-tests / newton-unittests (windows-latest)`
- Warp: 1.16.0.dev20260716
- The affected file is `newton/_src/utils/download_assets.py`, which was not modified by the PR whose CI surfaced this.

Miko997:
## Description

Retry transient Git fetch and clone failures up to three times with 1 and 2 second backoff. Each retry starts with a clean temporary directory. Permanent and unknown errors still fail immediately.

Closes #3774

## Checklist

## Test plan

```
uv run --extra dev -m unittest newton.tests.test_download_assets -v
uvx pre-commit run -a
uvx --from towncrier==25.8.0 towncrier build --draft --version 1.6.0 --date 2026-08-13
```

37 tests passed. All pre-commit hooks passed. The changelog draft rendered successfully.

## Bug fix

**Steps to reproduce:**

1. Start with an empty asset cache.
2. Return a transient HTTP 5xx from Git during a sparse fetch or clone.
3. Observe that the first `GitCommandError` aborts the download.

**Minimal reproduction:**

The issue contains a mock-based reproduction for the failing fetch path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Asset downloads now automatically retry transient Git failures up to three times with increasing delays.
  * Temporary download data is reset between attempts to improve recovery from incomplete downloads.
  * Permanent errors, such as authentication failures, are reported immediately without retries.
* **Documentation**
  * Added changelog information describing the improved retry behavior.

### DIFF
--- a/newton/_src/utils/download_assets.py
+++ b/newton/_src/utils/download_assets.py
@@ -26,6 +26,38 @@ MENAGERIE_URL = "https://github.com/google-deepmind/mujoco_menagerie.git"
 MENAGERIE_REF = "affef0836947b64cc06c4ab1cbf0152835693374"
 
 _SHA_RE = re.compile(r"[0-9a-f]{40}")
+_HTTP_SERVER_ERROR_RE = re.compile(r"\bhttp 5\d\d\b", re.IGNORECASE)
+_GIT_DOWNLOAD_ATTEMPTS = 3
+_GIT_DOWNLOAD_RETRY_DELAY = 1.0
+_TRANSIENT_GIT_ERROR_PATTERNS = (
+    "rpc failed",
+    "early eof",
+    "the remote end hung up",
+    "error reading section header",
+    "expected 'packfile'",
+)
+_PERMANENT_GIT_ERROR_PATTERNS = (
+    "authentication failed",
+    "permission denied",
+    "repository not found",
+    "couldn't find remote ref",
+    "http 401",
+    "http 403",
+    "http 404",
+)
+
+
+def _is_transient_git_error(error: Exception) -> bool:
+    """Return whether a Git error indicates a transient network failure."""
+    stderr = getattr(error, "stderr", "") or ""
+    if isinstance(stderr, bytes):
+        stderr = stderr.decode(errors="replace")
+    details = stderr.lower()
+    if any(pattern in details for pattern in _PERMANENT_GIT_ERROR_PATTERNS):
+        return False
+    return bool(_HTTP_SERVER_ERROR_RE.search(details)) or any(
+        pattern in details for pattern in _TRANSIENT_GIT_ERROR_PATTERNS
+    )
 
 
 def _get_newton_cache_dir() -> str:
@@ -356,9 +388,6 @@ def download_git_folder(
     _cleanup_stale_temp_dirs(cache_path, base_prefix)
 
     try:
-        if temp_dir.exists():
-            _safe_rmtree(temp_dir)
-
         if cached is not None:
             print(
                 f"New version of {folder_path} found "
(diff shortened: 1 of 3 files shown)


newton/_src/utils/download_assets.py
```diff
@@ -26,6 +26,38 @@
 MENAGERIE_REF = "affef0836947b64cc06c4ab1cbf0152835693374"
 
 _SHA_RE = re.compile(r"[0-9a-f]{40}")
+_HTTP_SERVER_ERROR_RE = re.compile(r"\bhttp 5\d\d\b", re.IGNORECASE)
+_GIT_DOWNLOAD_ATTEMPTS = 3
+_GIT_DOWNLOAD_RETRY_DELAY = 1.0
+_TRANSIENT_GIT_ERROR_PATTERNS = (
```
camevor, line 32: The transient allowlist misses Git’s common `The requested URL returned error: 503` error, so that failure still receives no retry. Could we invert this policy and retry fetch()/clone_from() failures unless they match a known non-retryable pattern?

Since that would make unknown `GitCommandErrors` retryable, please also scope the catch to the actual fetch/clone call so failures from init, remote setup, sparse checkout, or checkout still fail immediately.

newton/_src/utils/download_assets.py
```diff
@@ -26,6 +26,38 @@
...
+)
+_PERMANENT_GIT_ERROR_PATTERNS = (
+    "authentication failed",
+    "permission denied",
+    "repository not found",
+    "couldn't find remote ref",
+    "http 401",
+    "http 403",
+    "http 404",
+)
```
camevor, line 47: If we switch to retry-by-default, some additional high-confidence non-retryable conditions are:
- host key verification failed
- no space left on device
- read-only file system

newton/_src/utils/download_assets.py
```diff
@@ -368,32 +397,44 @@ def download_git_folder(
...
-                depth=1,
-                no_checkout=True,
-                multi_options=["--filter=blob:none", "--sparse"],
-            )
-            try:
-                repo.git.sparse_checkout("set", folder_path)
-                repo.git.checkout(ref)
-            finally:
-                repo.close()
+                if temp_dir.exists():
```
camevor: Nit: please scope the caught `GitCommandError` to the actual fetch/clone call so failures from init, remote setup, sparse checkout, or checkout still fail immediately.

newton/_src/utils/download_assets.py
```diff
@@ -368,32 +397,44 @@ def download_git_folder(
...
+                        repo.git.sparse_checkout("set", folder_path)
+                        repo.git.checkout(ref)
+                    finally:
+                        repo.close()
+                break
+            except GitCommandError as e:
+                if attempt == _GIT_DOWNLOAD_ATTEMPTS - 1 or not _is_transient_git_error(e):
+                    raise
+                delay = _GIT_DOWNLOAD_RETRY_DELAY * 2**attempt
+                print(f"Transient Git error. Retrying in {delay:g} seconds...")
```
camevor, line 436: Please include the error caught here or print it separately. Otherwise, a successful retry hides the diagnostic.

camevor (commented):
Thanks for your contribution @Miko997!

Looks good to me overall. One simplification I’d like before merging is to replace the transient/permanent classification with a single list of non-retryable patterns.

camevor:
On rereading the issue’s requested behavior, the transient allowlist is intentional. Please disregard my comments about inverting the retry policy and expanding the permanent list.

camevor (approved):

PR #3901, opened 2026-08-13
Add OpenCV pinhole camera rays

liorbenhorin:
Support calibrated OpenCV pinhole cameras in the tiled camera renderer, including rational radial, tangential, and thin-prism distortion.

## Description

Add `SensorTiledCamera.utils.compute_camera_rays_pinhole_opencv()` for generating camera-space rays from OpenCV pinhole intrinsics and distortion coefficients.

The helper supports:

- Rational radial coefficients `k1`–`k6`
- Tangential coefficients `p1` and `p2`
- Thin-prism coefficients `s1`–`s4`
- Calibration and output images with different resolutions
- Preallocated multi-camera ray buffers

During review of [IsaacLab PR #6851](https://github.com/isaac-sim/IsaacLab/pull/6851), we agreed that the OpenCV pinhole ray-generation helper belongs in Newton alongside its existing fisheye helpers. This PR upstreams that functionality, allowing the pending IsaacLab PR to remove its temporary local kernel while making the API available to other Newton users.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_opencv_pinhole
uv run --extra dev -m newton.tests -k TestSensorCameraRays
uvx pre-commit run -a
```

The complete test suite passed 6,336 tests.

## New feature / API change

```python
import newton

sensor = SensorTiledCamera(model)

camera_rays = sensor.utils.compute_camera_rays_pinhole_opencv(
    width=640,
    height=480,
    fx=339.27,
    fy=338.82,
    cx=323.56,
    cy=250.27,
    k1=0.1,
    k2=-0.05,
    p1=0.001,
    p2=-0.002,
)
```

## Summary by CodeRabbit

- **New Features**
  - Added support for generating camera rays from OpenCV-calibrated pinhole cameras.
  - Supports radial, rational, tangential, and thin-prism distortion coefficients.
  - Supports calibrated image dimensions, reusable output buffers, and multi-camera output selection.
  - Added validation for calibration dimensions, focal lengths, principal points, and distortion coefficients.
  - Provides a safe zero-direction fallback when distorted rays cannot be reliably inverted.

- **Documentation**
  - Expanded camera documentation with OpenCV calibration and ray-generation guidance.
  - Added a changelog entry for the new capability.


eric-heiden:
Nice addition to our camera sensor models!

The main concern is in the distortion inversion: on strong but still invertible calibrations the fixed-point loop settles on the wrong root, and the kernel returns confident, finite rays that are several degrees off. Details inline.

newton/_src/sensors/warp_raytrace/camera_utils.py
```diff
@@ -440,6 +440,63 @@ def compute_camera_rays_pinhole_from_aperture_kernel(
...
+        )
+        tangential_prism_y = (
+            p1 * (radius_squared + 2.0 * y * y) + 2.0 * p2 * x * y + s3 * radius_squared + s4 * radius_fourth
+        )
+        x = (x_distorted - tangential_prism_x) / radial
+        y = (y_distorted - tangential_prism_y) / radial
+
+    ray_direction_camera_space = wp.normalize(wp.vec3f(x, -y, -1.0))
+    out_rays[camera_index, py, px, 0] = wp.vec3f(0.0)
+    out_rays[camera_index, py, px, 1] = ray_direction_camera_space
```
eric-heiden: 🟠 This fixed-point inversion can converge to a solution that is not the true inverse, and the kernel still emits a confident ray for it.

For `k1=-0.45, k2=0.25` at 640x480 with `fx=fy=320`, `cx=320, cy=240` (roughly a 103 degree diagonal field of view), the true inverse radius at the worst pixel is `1.249837`, but the loop settles somewhere else and stays there:

| iterations | converged `r` |
| --- | --- |
| 10 | 1.051137 |
| 50 | 1.042110 |
| 1000 | 1.042110 |
| true value | 1.249837 |

This is not just an iteration budget problem. The loop genuinely converges, only to a spurious root, so raising the count or adding an iterate-to-iterate delta check would not catch it.

That calibration satisfies the precondition documented in `utils.py` ("The forward distortion mapping must be invertible over the calibrated image domain"): `rho(r) = r * radial(r)` is strictly increasing across the whole calibrated disc, so the mapping really is a bijection there. The resulting rays are off by up to 4.9 degrees, about 64 px of reprojection error, with 188 pixels in the image off by more than a degree.

Nothing downstream can flag this. Across 144 realistic rational-model coefficient sets, every bad ray came back finite, unit length, and marked valid, so a non-finite check does not help. `compute_camera_rays_fisheye_opencv` in this same module already has the right convention: it writes `wp.vec3f(0.0)` for rays it cannot resolve.

Two changes would address it:

1. After the loop, re-apply the forward distortion model to the converged `(x, y)` and compare against `(x_distorted, y_distorted)`. The two cases separate by about five orders of magnitude (residual `0.293621` at the spurious root versus `0.000000` at the true inverse), so a small tolerance is enough. If the residual is too large, write the `wp.vec3f(0.0)` sentinel instead of a normalized direction.
2. Either raise `_PINHOLE_OPENCV_INVERSION_ITERATIONS` or replace the fixed-point step with a Newton step on the forward model. Ten iterations is too few even when the iteration is well behaved: a mild `k1=-0.20` barrel is still 0.655 degrees (about 9.1 px) off after 10 iterations and needs roughly 200 to reach the true `r = 1.234563`. Newton also converges to the correct root rather than the spurious one.

It would also be worth extending `test_opencv_pinhole_full_distortion_round_trips`. It uses `k1=0.1, k2=-0.05`, which sits well inside the contractive regime and round-trips to 1e-5 regardless. A strong but still invertible set such as `k1=-0.45, k2=0.25` fails today.

liorbenhorin: Thanks @eric-heiden ! Fixed in ffacbba3 - replaced the fixed-point iteration with a
damped Newton solve and added a forward-residual check. Extended the tests to
cover the strong-but-invertible distortion you flagged, plus a non-invertible case.

eric-heiden (approved):
Thanks!

PR #3905, opened 2026-08-13
Fix frictionless Coulomb dual cone projection in PADMM

rubengrandia:
## Description

Fix branch ordering in `project_to_coulomb_dual_cone()` so the membership test runs before the polar-cone test. At `mu = 0` the dual cone is the half-space `xn >= 0`; evaluating the polar-cone condition first incorrectly maps feasible axis points `(0, 0, +xn)` to zero.

Add regression coverage that compares the Warp Coulomb cone and dual-cone projections against a float64 reference across interior, boundary, and polar-cone samples, including degenerate friction coefficients (`mu = 0`, `mu → ∞`).

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_12_coulomb_cone_projections_match_analytic_reference
```

## Bug fix

**Steps to reproduce:**

1. Project `(0, 0, xn)` with `xn > 0` onto the Coulomb dual cone with `mu = 0`.
2. Observe the projection returns `(0, 0, 0)` instead of the input.

**Minimal reproduction:**

```python
import numpy as np
import warp as wp

from newton._src.solvers.kamino._src.solvers.padmm.math import project_to_coulomb_dual_cone

@wp.kernel
def project_kernel(x: wp.array[wp.vec3f], out: wp.array[wp.vec3f]):
    out[0] = project_to_coulomb_dual_cone(x[0], wp.float32(0.0))

x = wp.array([wp.vec3f(0.0, 0.0, 1.0)], dtype=wp.vec3f, device="cpu")
out = wp.zeros(1, dtype=wp.vec3f, device="cpu")
wp.launch(project_kernel, dim=1, inputs=[x], outputs=[out], device="cpu")

assert np.allclose(out.numpy()[0], [0.0, 0.0, 1.0])  # fails before this PR
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Coulomb cone projections for zero-friction cases, preserving valid vectors while correctly projecting invalid inputs.
  * Improved handling across interior, boundary, polar, and general projection scenarios.
  * Simplified projection behavior for more consistent results across friction levels.

* **Tests**
  * Added comprehensive validation for primal and dual Coulomb cone projections.
  * Expanded coverage across zero, small, unit, and large friction values, including representative projection cases.

### DIFF
--- a/newton/_src/solvers/kamino/_src/solvers/padmm/math.py
+++ b/newton/_src/solvers/kamino/_src/solvers/padmm/math.py
@@ -45,14 +45,13 @@ wp.set_module_options({"enable_backward": False})
 
 
 @wp.func
-def project_to_coulomb_cone(x: wp.vec3f, mu: wp.float32, epsilon: wp.float32 = 0.0) -> wp.vec3f:
+def project_to_coulomb_cone(x: wp.vec3f, mu: wp.float32) -> wp.vec3f:
     """
     Projects a 3D vector `x` onto an isotropic Coulomb friction cone defined by the friction coefficient `mu`.
 
     Args:
         x: The input vector to be projected.
         mu: The friction coefficient defining the aperture of the cone.
-        epsilon: A numerical tolerance applied to the cone boundary. Defaults to 0.0.
 
     Returns:
         The vector projected onto the Coulomb cone.
@@ -60,8 +59,11 @@ def project_to_coulomb_cone(x: wp.vec3f, mu: wp.float32, epsilon: wp.float32 = 0
     xn = x[2]
     xt_norm = wp.sqrt(x[0] * x[0] + x[1] * x[1])
     y = wp.vec3f(0.0)
-    if mu * xt_norm > -xn + epsilon:
-        if xt_norm <= mu * xn + epsilon:
+    # The polar-cone test must precede the membership test: at mu = 0 the primal
+    # cone degenerates to the ray { xt = 0, xn >= 0 }, testing xt_norm <= mu * xn
+    # first would leave the infeasible point (0, 0, -xn) unchanged.
+    if mu * xt_norm > -xn:
+        if xt_norm <= mu * xn:
             y = x
         else:
             ys = (mu * xt_norm + xn) / (mu * mu + 1.0)
@@ -73,7 +75,7 @@ def project_to_coulomb_cone(x: wp.vec3f, mu: wp.float32, epsilon: wp.float32 = 0
 
 
 @wp.func
-def project_to_coulomb_dual_cone(x: wp.vec3f, mu: wp.float32, epsilon: wp.float32 = 0.0) -> wp.vec3f:
+def project_to_coulomb_dual_cone(x: wp.vec3f, mu: wp.float32) -> wp.vec3f:
     """
     Projects a 3D vector `x` onto the dual of an isotropic Coulomb
     friction cone defined by the friction coefficient `mu`.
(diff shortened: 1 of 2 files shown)


rubengrandia:
Tracking issue: disneyresearch/newton#430

newton/_src/solvers/kamino/_src/solvers/padmm/math.py
Guirec-Maloisel: Could we update this one as well to also check first if we just return the input? I don't think we'd have the same corner case here (it would be for mu -> infinity) but it's probably cleaner to ensure first if the point is already in the cone

Guirec-Maloisel: Slightly outside the diff, but this epsilon argument is not used anywhere in the code (left at zero). I'm wondering if we could get rid of it? I don't see a reason why you'd want to introduce a tolerance here.

newton/tests/kamino/test_kamino_solvers_padmm.py
```diff
@@ -985,6 +1078,42 @@ def test_11_apadmm_honors_per_world_iteration_budgets(self):
...
+        wp.launch(
+            kernel=project_onto_coulomb_cones_kernel,
+            dim=len(x),
+            inputs=[x_arg, mu_arg],
+            outputs=[primal, dual],
+            device=self.default_device,
+        )
+        return primal.numpy(), dual.numpy()
+
+    def test_12_coulomb_cone_projections_match_analytic_reference(self):
```
Guirec-Maloisel: This PR introduces a whole architecture to test a single math function. I'm wondering if we could directly list a few hardcoded examples of inputs and outputs (including corner cases), and call directly the wp.func from Python in the text case.

newton/_src/solvers/kamino/_src/solvers/padmm/math.py
rubengrandia: This was not possible and introduced the same bug we had before. Added a comment to the code the explain why the order is important

rubengrandia: I agree, I removed it.

newton/tests/kamino/test_kamino_solvers_padmm.py
rubengrandia: I simplified to tests to hardcoded cases. I checked that they still catch the original bug. Calling the wp.func directly is not possible, but the whole test logic should be much more readable now.

chschuma-disney (dismissed):
Looks good from my side!

newton/_src/solvers/kamino/_src/solvers/padmm/math.py
Guirec-Maloisel: Yes, it makes sense that you'd have the opposite problem there, you basically swap the 0 and infinity cases. Thanks for clarifying!

Guirec-Maloisel (dismissed):
Thanks for the simplifications to the code, looks good to me for merging now!

newton/tests/kamino/test_kamino_solvers_padmm.py
Guirec-Maloisel: Hmm interesting, it works for me in a notebook if I manually create a wp.vec3f and call the wp.func directly (returns a wp.vec3f). But it might be different in a unit test script then.

chschuma-disney (approved):

Guirec-Maloisel (approved):

PR #3906, opened 2026-08-13
Apply MJCF compiler settotalmass on import

### LINKED ISSUE #3689: [BUG] MJCF compiler settotalmass is ignored
## Description

Newton currently ignores MJCF `<compiler settotalmass="...">`. Native MuJoCo rescales the compiled model's dynamic body masses and inertias so their total reaches the requested value, whereas Newton retains the authored values.

This changes accelerations, gravity response, contacts, actuator behavior, and controller tuning.

## Minimal reproduction

```xml
<mujoco>
  <compiler settotalmass="10"/>
  <worldbody>
    <body name="body">
      <freejoint/>
      <geom type="sphere" size="0.1" mass="2"/>
    </body>
  </worldbody>
</mujoco>
```

Native MuJoCo compiles this body with mass 10, while Newton retains mass 2.

## Expected behavior

Each MJCF import should rescale the bodies created by that import to the requested total mass, without modifying bodies that already existed in the destination `ModelBuilder`. Both mass and inertia must use the same scale factor.

## Edge cases and validation

- Multiple bodies with mixed explicit geom masses and `<inertial>` definitions.
- Zero-total-mass and static-only models.
- Import into a nonempty builder.
- Fixed-joint collapse and builder composition.
- Comparison of body masses and inertias with native MuJoCo.

## Coordination

PR #3596 changes MJCF inertia selection and mass aggregation. An implementation should coordinate with or follow that work to avoid conflicting mass-computation paths.

deepspace28:
## Description

MJCF `<compiler settotalmass>` rescales a compiled model so its body masses sum to the requested value. Newton ignored the attribute, so imported models kept their authored masses — changing accelerations, gravity response, contacts, and actuator tuning relative to native MuJoCo.

The rescale is applied at the end of `parse_mjcf`, over only the bodies the import created.

I confirmed the semantics against native `mujoco` 3.11 by compiling each case rather than inferring them from the docs, and **two of my initial assumptions were wrong** (both corrected [in the issue thread](https://github.com/newton-physics/newton/issues/3689#issuecomment-5281731550)):

- **Every non-world body counts, jointless ones included.** A static body of mass 3 beside a dynamic body of mass 1 rescales to `7.5` and `2.5` for a requested total of 10 — not to 10 for the dynamic body alone.
- **A zero-mass import needs no diagnostic.** It is simply left alone. A *moving* body with zero mass never reaches the rescale, being rejected earlier against `mjMINVAL` for unrelated reasons.

Confirmed as expected: mass and inertia take the **same linear factor**, not its square (`mass 2 → 10` alongside `inertia 0.008 → 0.04`), and a non-positive `settotalmass` is a no-op (`"0"` and `"-1"` both leave masses untouched).

Two decisions worth review attention:

1. **Scope.** `settotalmass` describes the imported model, so importing into a populated `ModelBuilder` leaves its existing bodies untouched, and they do not dilute the scale factor. This is what the issue asks for, but it is a deliberate divergence from a literal reading of the attribute name.
2. **The inverses are scaled, not recomputed.** `1/(m·k)` is exactly `inv_mass/k`, so scaling keeps `body_inv_mass` and `body_inv_inertia` in step without inverting a possibly-singular inertia, and leaves the zero placeholders marking static bodies undisturbed.

The rescale runs before any fixed-joint collapse, while the imported bodies still occupy a contiguous index range. A uniform scale commutes with the collapse's additive merge, so the ordering is a matter of index stability rather than of result.

Closes #3689

## Checklist

## Test plan

```
uv run --extra dev --extra sim -m newton.tests -k TestImportMjcfSetTotalMass
```

10 new tests, all passing. The full `test_import_mjcf` module also passes — 276 tests, no regressions.

**Verified the tests fail without the fix:** stashing the `import_mjcf.py` change leaves **6 of the 10 red**. The other 4 assert no-change behaviour, so they are guards against a careless fix rather than against the missing feature. Rather than assume they earn their place, I mutation-tested them against the finished fix:

| Mutation | Caught by |
| --- | --- |
| Drop the two inverse updates | `test_inverse_mass_and_inertia_stay_consistent` only |
| Scale inertia by `mass_scale ** 2` | both inertia tests |

Expected values throughout are those native MuJoCo 3.11 produces for the same XML, read from `MjModel.body_mass` / `body_inertia`.

## Bug fix

**Steps to reproduce:**

1. Author an MJCF model with `<compiler settotalmass="10"/>` and a single body whose geom mass is 2.
2. Import it with `ModelBuilder.add_mjcf()`.
3. Observe the body retains mass 2, where native MuJoCo compiles it to 10.

**Minimal reproduction:**

```python
import newton

MJCF = """
<mujoco>
  <compiler settotalmass="10"/>
  <worldbody>
    <body name="body">
      <freejoint/>
      <geom type="sphere" size="0.1" mass="2"/>
    </body>
  </worldbody>
</mujoco>
"""

builder.add_mjcf(MJCF)
print(builder.body_mass[0])  # 2.0 before this PR, 10.0 after — MuJoCo compiles 10.0
```

---

### Unrelated observation, not addressed here

While working on the inertia path I noticed the guard at `import_mjcf.py:2304`:

```python
if any(x for x in I_m):
    builder.body_inv_inertia[link] = wp.inverse(I_m)
else:
    builder.body_inv_inertia[link] = I_m
```

Iterating a `wp.mat33` yields its rows, and a `wp.vec3` is always truthy, so `any(...)` returns `True` even for an all-zero matrix and the `else` branch is unreachable — `wp.inverse` is called on a singular inertia regardless. It predates this change and I have left it alone to keep the diff focused, but I am happy to open a separate issue or fold a fix in here if you would prefer.

## Summary by CodeRabbit

* **New Features**
  * MJCF imports now support the `<compiler settotalmass>` setting, proportionally adjusting imported body masses and inertias.
  * Heightfield elevation data is normalized consistently across supported sources.
  * MJCF actuator imports now support additional joint, site, and slider-crank transmission configurations.

* **Bug Fixes**
  * Ball-joint damping now preserves authored values or applies builder defaults correctly.
  * Heightfield row orientation and constant-data handling are corrected.
  * Non-positive mass targets remain unchanged, and existing builder bodies are excluded from rescaling.


deepspace28:
Thanks — the slop flag was fair on the code comments, so I've tightened them in d96e6a5.

The block above the rescale opened by restating what the code does, and ran to thirteen comment lines against twelve lines of code. `AGENTS.md` asks for the opposite ("brief, and only for non-obvious code... explain *why*, not *what* the code already shows"), so that was a real miss rather than a false positive.

Now eight lines, keeping only the four things the code cannot show:

- why the scope is this import rather than the whole builder,
- why the rescale precedes the fixed-joint collapse,
- why a zero total is skipped,
- why the inverses are scaled rather than recomputed.

The fuller derivation still exists — it lives in the commit message and the PR description, which is where it belongs.

I left the test docstrings as they are: `AGENTS.md` requires a docstring on every test and permits a body elaborating on complex ones, and they carry the provenance of the expected numbers (native MuJoCo 3.11 output) rather than restating the assertions. Happy to trim those too if a reviewer disagrees.

No behaviour change: the ten tests still pass, and six still fail against `main`'s importer.

deepspace28:
Thanks — both actionable points are in `fb6754b6`.

**Helper name.** Renamed `_settotalmass_mjcf` to `_mjcf_settotalmass` across all eleven call sites, matching the prefix-first rule in `AGENTS.md`.

**Sentinel comment.** This was a genuine inaccuracy, not just verbosity: an explicit `settotalmass="0"` stays `0.0`, and only an *absent* attribute becomes `-1.0`, so "share a sentinel" was wrong. What they actually share is the behaviour — both are non-positive, so both disable rescaling. Reworded to say that.

**The seven-line comment is already gone.** This review covered `afcf119..cada685`, but `d96e6a5` had trimmed that block from thirteen comment lines to three about an hour earlier, for the same reason. It currently reads:

```python
# settotalmass describes the imported model, so bodies the destination builder
# already held are excluded from both the sum and the scale. Runs before any
# fixed-joint collapse, while the imported bodies are still contiguous.
```

Happy to shorten it further if a reviewer still finds it long, though I'd want to keep the scope and ordering rationale somewhere — both are decisions the code cannot show.

No behaviour change in either commit: ten tests still pass, and six still fail against `main`'s importer.

deepspace28:
Gentle ping on this one.

The blocker looks procedural rather than technical: `Check Author Membership` and `detect-api-changes` are both green, but `Require Manual Approval for External PRs` has been sitting at *waiting* since the 13th, so the full suite has never actually run. Could a maintainer approve the workflow run?

The branch is 15 commits behind `main` now, though still conflict-free. Happy to rebase before the run if that's preferred — just say the word and I'll push.

Issue #1524, opened 2026-02-04
Handle USD parsing of multi-dof joints, MjcActuators


eric-heiden:
[Converting](https://github.com/newton-physics/mujoco-usd-converter) the gym humanoid (`nv_humanoid.xml` from `newton/tests/assets`) from MJCF to USD results in multi-dof joints getting mapped not to D6 joints but to individual revolute joints for every dof, so we have potentially more than one joint in USD between 2 bodies. This currently breaks our USD parser (it thinks there are cycles in the joint graph).

See the converted [humanoid_mjc.txt](https://github.com/user-attachments/files/25063858/humanoid_mjc.txt) (replace .txt with .usda) for such a multi-dof articulation with MjcActuator prims.

Besides adding parsing support for such assets, we need to update the handling of MJC actuators: `mjc:target` (the transmission target of an MjcActuator) needs to be resolved to the MuJoCo dof index from the USD prim path of the respective joint dof. So we need to keep a mapping of these prim paths to dof indices somewhere to be able to resolve the `mjc:target` index. Maybe something like `ModelBuilder.joint_dof_key` could be used added so every dof (including the multi-dof D6 joint) can register its own USD path.

andrewkaufman:
@eric-heiden re the parsing issue
> This currently breaks our USD parser (it thinks there are cycles in the joint graph).

Is this a UsdPhysics parser issue or a Newton parser issue?

andrewkaufman:
> mjc:target 

Does the mjc target ever target a single dof of a multi-dof joint? I think it only targets all dofs equally?

eric-heiden:
This is a Newton parser issue, I assumed so far there can only be one joint between bodies.

@andrewkaufman The `mjc:target` appears to target the entire joint indeed, but since we combine multiple dofs into a single D6 joint in Newton (we do this in the MJCF parser, and have to do that also in the USD parser as part of this issue) we need to keep track of these individual joint prim paths of the dofs that got subsumed in those D6 joints. I think for now this "joint dof path" can be handled through CustomAttributes on the MJC schema parser at least.

andrewkaufman:
> since we combine multiple dofs into a single D6 joint in Newton (we do this in the MJCF parser

I guess I'm not understanding why we do that. Is that beneficial for some other solvers? It seems like it could only be lossy (or at best equivalent but slow/confusing) for SolverMujoco

eric-heiden:
[...]

PR #1512, opened 2026-02-02
Introduce Extended Contact Attributes & Refactor `SensorContact`

camevor:
## Description
This PR introduce Extended Contact Attributes – analogous to Extended State Attributes – for interfacing with solver outputs, such as contact forces. This is used to access contact force information in `SensorContact` via a specified interface.

Further changes:
* fill & use Newton fields on `Contacts` struct to avoid ad-hoc assignment of structs
* remove `populate_contacts` function

Resolves #1283, resolves #624

## Newton Migration Guide

Please ensure the migration guide for **warp.sim** users is up-to-date with the changes made in this PR.

- [ ] The migration guide in ``docs/migration.rst`` is up-to date

## Before your PR is "Ready for review"

- [ ] Necessary tests have been added and new examples are tested (see `newton/tests/test_examples.py`)
- [ ] Documentation is up-to-date
- [ ] Code passes formatting and linting checks with `pre-commit run -a`

## Summary by CodeRabbit

* **New Features**
  * Models and builders can request optional per-contact attributes; sensors auto-request contact force by default.

* **Documentation**
  * New consolidated "Extended Attributes" guide covering contact and state attributes; sensor docs updated to reference it.

* **Breaking Changes**
  * Legacy contact-population helper removed—use the solver-based contact update flow.
  * Contact data format changed to shape-indexed contacts and spatial-force vectors (fields updated).

PR #3908, opened 2026-08-13
Drop the line-search block-dimension workaround

### LINKED ISSUE #3544: Remove SolverMuJoCo line-search block-dimension workaround
Newton currently forces `self.mjw_model.block_dim.linesearch_iterative = 32` after `mujoco_warp.put_model()`. This workaround was added in #3502 because MuJoCo Warp selected a 512-thread block for models with `nv > 500`, causing the elliptic line-search kernel to exceed the CUDA registers-per-block limit on some architectures and fail with CUDA error 701.

The issue was resolved upstream by [google-deepmind/mujoco_warp#1524](https://github.com/google-deepmind/mujoco_warp/pull/1524), which caps the large-model line-search block dimension at 256.

Once Newton integrates the first MuJoCo Warp release containing that fix:

- Remove the `linesearch_iterative = 32` override from `SolverMuJoCo`.
- Let MuJoCo Warp select the upstream block dimension.
- Verify large-`nv` models using elliptic cones run successfully on CUDA, including Ada GPUs.
- Run the existing `SolverMuJoCo` regression tests.

deepspace28:
## Description

`SolverMuJoCo` forced `block_dim.linesearch_iterative = 32` immediately after `mujoco_warp.put_model()`. #3502 added it because MuJoCo Warp selected a 512-thread block for models with `nv > 500`, overrunning the CUDA registers-per-block limit and failing with error 701.

google-deepmind/mujoco_warp#1524 capped that upstream, and the cap ships in `mujoco-warp` 3.11.0 — the version `pyproject.toml` already pins. `mujoco_warp/_src/io.py`:

```python
if mjm.nv > 500:
    m.block_dim.linesearch_iterative = 256
```

So the workaround now overrides a value upstream has already chosen safely, holding large models to a needlessly small line-search block.

**The change is narrower than "remove an unconditional assignment" suggests.** I raised on the issue that deleting the line would also hand models below the threshold back to the upstream default, and that this was a wider behaviour change than the issue implied. Having checked, that concern was unfounded — upstream's `BlockDim()` default for this field is **also 32**:

```
upstream BlockDim() default = 32
bodies=   2  nv=  12  linesearch_iterative=32
bodies= 100  nv= 600  linesearch_iterative=256
```

Models at or below `nv > 500` are therefore bit-identical. The only behavioural change is `32 → 256` above the threshold, which is exactly what the issue asks for.

Closes #3544

## Checklist

## Test plan

```
uv run --extra dev --extra sim -m newton.tests -k TestMuJoCoLinesearchBlockDim
```

The new test builds a 100-body free-joint model (`nv = 600`, over the threshold) and asserts the solver leaves the field at whatever `put_model()` chose. It compares against `put_model()`'s own value rather than a literal `256`, so it keeps its meaning if MuJoCo Warp retunes the number.

**Verified it fails without the change** — restoring the override gives `AssertionError: 32 != 256`. A model over the threshold is the only one that can detect the regression, since it is the only one whose value differs from the default.

Regression runs, all passing:

| Suite | Result |
| --- | --- |
| `test_mujoco_solver` | 272 passed, 7 skipped |
| `test_mujoco_reset`, `test_mujoco_fk_consistency`, `test_mujoco_general_actuators` | 39 passed |
| `TestMuJoCoLinesearchBlockDim` on CPU | passed |

**One limitation worth stating plainly:** the original error 701 was architecture-dependent. I verified on an RTX 3050 (sm_86) that a `nv = 600` model builds and steps without it, so this demonstrates the override is unnecessary on that architecture — not on every architecture that originally hit the failure. If there is a specific model or `nv` range from #3502 that reproduced it more reliably, I am happy to run against that instead.

## Bug fix

**Steps to reproduce:**

1. Build a model with `nv > 500` (100 bodies with free joints suffices).
2. Construct a `SolverMuJoCo` for it.
3. Read `solver.mjw_model.block_dim.linesearch_iterative`.
4. Observe `32`, where MuJoCo Warp had selected `256`.

**Minimal reproduction:**

```python
import warp as wp
import newton

for i in range(100):
    body = builder.add_body(xform=wp.transform(wp.vec3(i * 0.5, 0.0, 1.0), wp.quat_identity()), mass=1.0)
    builder.add_shape_sphere(body, radius=0.1)
    builder.add_joint_free(body)

solver = SolverMuJoCo(builder.finalize())
print(solver.mj_model.nv)                                   # 600
print(solver.mjw_model.block_dim.linesearch_iterative)      # 32 before, 256 after
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Removed a fixed line-search block-size restriction, allowing MuJoCo Warp to select an appropriate size automatically.
  - Improved reliability when solving large models, including models with more than 500 degrees of freedom.

- **Tests**
  - Added regression coverage for large models and repeated CUDA simulation steps.


Miko997 (commented):
Checked the pinned MuJoCo Warp version and the upstream cap. Models at or below `nv = 500` keep the default 32, while larger models retain MuJoCo Warp's 256 selection. The regression would catch the override returning. Looks good.

deepspace28:
Both points were fair — addressed in `e777b480`.

**Optional-dependency handling.** Replaced the bare `import mujoco_warp` with `SolverMuJoCo.import_mujoco()` inside a `try`/`except ImportError` that skips, matching the existing pattern at `test_menagerie_mujoco.py:61`. You're right that the direct import turned a missing backend into a failure instead of a skip.

**Runtime coverage.** Also correct, and the better catch — the sizing assertion cannot detect a launch fault, because the block dimension only matters once the kernel actually runs. Added `test_large_elliptic_model_steps_on_cuda`, which targets the original failure mode rather than a generic solve: error 701 occurred in the **elliptic** line-search kernel, so the model is imported with `cone="elliptic"` and `nv > 500`, stepped five times, then `wp.synchronize()`d — launch failures surface asynchronously, so without the sync the test would pass through a real fault. It skips when the default device is not CUDA, verified by running with the default device set to `cpu`:

```
test_large_elliptic_model_steps_on_cuda ... skipped 'block dimensions only constrain CUDA kernel launches'
test_linesearch_block_dim_is_left_to_mujoco_warp ... ok
```

Both models now come from one MJCF helper so the cone option is set where MuJoCo parses it, rather than patched onto an already-built model.

274 tests pass in `test_mujoco_solver`; the sizing test still fails against `main` with `32 != 256`.

One correction on the framing, since it would otherwise end up in the record: the review describes the PR objective as including elliptic-cone execution "including Ada GPUs". The PR doesn't claim that — the hardware here is an **RTX 3050, sm_86 (Ampere)**, and I flagged in the description that error 701 was architecture-dependent, so this demonstrates the override is unnecessary on that one architecture rather than on every architecture that originally hit it. The new test narrows that gap but does not close it; only CI on the affected hardware can.

jcarius-nv (approved):
Reviewed at `60aa507d9e43dbc22ab177c9ef0ee255271108a4`.

## Review summary

- **Standards:** No substantive findings. The regression uses `unittest`, public Newton APIs, compliant test docstrings, and a justified synchronization point for asynchronous CUDA launch failures.
- **Spec:** The change satisfies #3544: Newton removes its obsolete `linesearch_iterative = 32` override, leaves block-size selection to the pinned MuJoCo Warp 3.11.0, and covers both delegation and the large-`nv` elliptic CUDA runtime path.
- **Fit:** **Fits.** `SolverMuJoCo` is the right owner for removing the downstream workaround it introduced, while MuJoCo Warp remains the sustainable owner of kernel launch tuning. The change adds no public API, configuration, dependency, or maintenance surface.

Validation at this head:

- `TestMuJoCoLinesearchBlockDim`: 2/2 passed under `--strict-warnings` on an NVIDIA L4 (Ada, `sm_89`).
- The large elliptic test compiled and launched `_linesearch_iterative` with `nv > 500`.
- `git diff --check` passed.

## Context

This closes a useful loop in Alain Denzler's earlier MuJoCo Warp work:

1. [MuJoCo Warp #1512](https://github.com/google-deepmind/mujoco_warp/pull/1512) stabilized elliptic line search by computing small float32 cost changes directly instead of subtracting nearly equal absolute costs.
2. The resulting register-heavy elliptic kernel exposed a launch problem on Ada: for `nv > 500`, the prior 512-thread selection required about 78K registers against CUDA's 64K-per-block limit and failed with error 701.
3. [Newton #3502](https://github.com/newton-physics/newton/pull/3502) integrated that numerical fix and added the emergency downstream override to 32 threads. Review-time coverage found no performance regression from that temporary setting on the measured large pyramidal workload.
4. Alain then implemented the permanent upstream solution in [MuJoCo Warp #1524](https://github.com/google-deepmind/mujoco_warp/pull/1524), capping the large-model block size at 256. This is safe by construction because `256 * 255 <= 65,536`, and the accompanying benchmarks retained essentially all of the wider block's benefit.
5. Alain opened #3544 to remove Newton's override once the upstream cap shipped. Newton now pins MuJoCo Warp 3.11.0, which contains that fix.

The resulting behavior is narrow: models with `nv <= 500` remain at the upstream default of 32, while larger models use the upstream-selected 256. No line-search algorithm or convergence behavior is intentionally changed.

Approved. Thanks for carrying the workaround through to a focused regression and clean upstream ownership.

Issue #3691, opened 2026-07-27
[BUG] MJCF import scale misses linear joint and constraint quantities


han-xudong:
## Description

`ModelBuilder.add_mjcf(scale=...)` scales body positions, geom dimensions, sites, and inertial offsets, but does not consistently scale every length-valued MJCF quantity. This can produce a mechanism whose geometry and motion or constraint parameters use different length scales.

## Confirmed reproduction

```xml
<mujoco>
  <worldbody>
    <body name="body">
      <joint name="slide" type="slide" range="-0.3 0.3"/>
      <geom type="box" size="0.1 0.1 0.1" mass="1"/>
    </body>
  </worldbody>
</mujoco>
```

Importing with `scale=2` doubles the body and geom dimensions, but the slide-joint range remains `[-0.3, 0.3]` rather than `[-0.6, 0.6]`.

## Audit scope

Other likely length-valued fields requiring review include:

- slide-joint `range`, `ref`, and `springref`;
- equality weld `relpose` translation;
- tendon range, margin, width, and spring length;
- actuator length ranges;
- remaining linear margins and anchors.

Angular, mass, force, time, and dimensionless quantities must not be length-scaled.

## Expected behavior

Every linear quantity belonging to the imported mechanism should be scaled consistently, while non-linear quantities retain their existing semantics. A shared quantity-aware helper or attribute table would be preferable to isolated multiplications.

## Validation

- Compare scale 1 and scale 2 imports for every audited linear attribute.
- Verify hinge ranges and references are not length-scaled.
- Verify equality, tendon, actuator, and `SolverMuJoCo` representations remain consistent.

## Coordination

PR #3642 modifies joint range and reference conversion. Slide range/ref scaling should be implemented against its final coordinate convention.

deepspace28:
Opened #3909 covering the equality-weld item.

Worth recording what the rest of the audit list looks like against the current importer, since most of it turns out not to need work:

| Field | Finding |
| --- | --- |
| geom `margin`, `gap` | already scaled |
| equality weld `anchor` | already scaled |
| **equality weld `relpose`** | **not scaled — fixed in #3909** |
| joint `ref`, `springref` | not imported at all, so nothing to scale yet |
| tendon `range`, `margin`, `width`, spring length | not parsed as length quantities; only `fixed` tendons with joint coefficients are imported |
| actuator `lengthrange` | not imported |
| slide-joint `range` | not scaled — real, but see below |

[...]

PR #3909, opened 2026-08-13
Scale MJCF equality weld relpose translations

deepspace28:
## Description

`ModelBuilder.add_mjcf(scale=...)` scaled the equality weld `anchor` but not the `relpose` sitting beside it, so a scaled import produced a constraint whose two length-valued fields disagreed — bodies welded at their authored offset inside geometry that had moved. Measured before the fix, importing `anchor="0.3 0 0"` and `relpose="0.4 0 0 1 0 0 0"`:

| | scale 1 | scale 2 |
| --- | --- | --- |
| `anchor` | 0.3 | **0.6** (scaled) |
| `relpose` translation | 0.4 | **0.4** (not scaled) |

Both weld forms build the transform through separate branches, so the body-based and site-based paths were each wrong independently. Rather than multiply by `scale` in two places, the parsing moves into a `parse_relpose()` helper that scales the translation and leaves the quaternion alone — #3691 asks for a shared quantity-aware helper over isolated multiplications, and this duplication is exactly what let the two branches drift.

### Scope — this is one item from the audit list, and I want to be explicit about why

I worked through the rest of #3691's list against the current importer:

| Field | Finding |
| --- | --- |
| geom `margin`, `gap` | **Already scaled** (`import_mjcf.py`, the `mj_gap`/`mj_margin` parse) |
| equality weld `anchor` | **Already scaled** |
| **equality weld `relpose`** | **Not scaled — fixed here** |
| joint `ref`, `springref` | **Not imported at all**, so there is nothing to scale yet |
| tendon `range`, `margin`, `width`, spring length | Not parsed as length quantities; only `fixed` tendons with joint coefficients are imported |
| actuator `lengthrange` | Not imported |
| slide-joint `range` | **Not scaled — deliberately left out**, see below |

Slide-joint `range` is the issue's own headline reproduction, and it is genuinely broken (`import_mjcf.py`, where `limit_lower`/`limit_upper` take `joint_range` unscaled for non-angular joints). I have not touched it because #3691's Coordination section asks that slide range/ref scaling be implemented against #3642's final coordinate convention, and #3642 is still open. Happy to follow up once it lands, or to fold it in now if you would rather not wait.

So the audit list is smaller than it appears: most entries are either already correct or describe fields the importer does not yet read.

Addresses #3691

## Checklist

## Test plan

```
uv run --extra dev --extra sim -m newton.tests -k TestImportMjcfWeldRelposeScale
```

Four tests. **Verified they fail without the fix** — the two translation cases go red, while the rotation and anchor cases stay green, which is the point of including them: they are guards against a fix that over-reaches by scaling the quaternion or regressing the anchor.

Full `test_import_mjcf` passes (270 tests), as does `test_equality_connect_constraint_with_sim_step` (6 tests).

## Bug fix

**Steps to reproduce:**

1. Author an MJCF with an equality weld carrying both `anchor` and `relpose`.
2. Import it twice, with `scale=1` and `scale=2`.
3. Compare `mujoco:equality_constraint_anchor` against `mujoco:equality_constraint_relpose`.
4. Observe the anchor doubles while the relpose translation does not.

**Minimal reproduction:**

```python
import newton

MJCF = """
<mujoco>
  <worldbody>
    <body name="a"><freejoint/><geom type="sphere" size="0.1" mass="1"/></body>
    <body name="b" pos="1 0 0"><freejoint/><geom type="sphere" size="0.1" mass="1"/></body>
  </worldbody>
  <equality>
    <weld body1="a" body2="b" anchor="0.3 0 0" relpose="0.4 0 0 1 0 0 0"/>
  </equality>
</mujoco>
"""

builder.add_mjcf(MJCF, scale=2.0, convert_mjc_equality_constraints=False)
print(list(builder.custom_attributes["mujoco:equality_constraint_anchor"].values[0])[0])   # 0.6
print(list(builder.custom_attributes["mujoco:equality_constraint_relpose"].values[0])[0])  # 0.4 before, 0.8 after
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected scaling of MJCF weld relative-position translations during import.
  * Preserved weld rotation values while scaling translations and anchors consistently for body- and site-based welds.
  * Improved heightfield imports with correct row orientation, inline data normalization, and consistent elevation ordering across supported formats.

* **Tests**
  * Added coverage for weld scaling and rotation preservation.
  * Added regression coverage for heightfield orientation, normalization, and data ordering.


deepspace28:
Gentle ping.

Same procedural gate as #3906: `Check Author Membership` and `detect-api-changes` are green, but `Require Manual Approval for External PRs` is still *waiting*, so the full suite hasn't run. Could a maintainer approve the workflow run?

This branch is current with `main` and conflict-free, so it should be ready to test as-is. CodeRabbit raised no actionable comments.

deepspace28:
On the heightfield comments: I'm keeping them. They record which row convention was verified against which MuJoCo version (3.10 vs 3.11 differ on image vs binary sources), and dropping that history turns a verified behavior back into folklore the next person has to re-derive. The PR itself is unchanged and still waiting on the external-PR CI approval gate - same procedural hold as #3906.

PR #3911, opened 2026-08-14
Use authored colors for imported shapes with no bound material

### LINKED ISSUE #3910: Imported visual shapes with no material get debug palette colours
### Summary

When `add_usd` imports a visual shape that has no bound material, the shape ends up with a colour from `ModelBuilder`'s per-shape debug palette. The result is that an imported scene renders in saturated rainbow colours, which reads as "Newton got the colours wrong" rather than "these prims have no material".

Every other renderer draws an unmaterialed prim in a neutral default. Comparing Newton's GL viewer against an RTX render of the same stage, the differences are entirely accounted for by this: a ground plane that is light grey in the reference is saturated green in Newton, an obelisk that is grey renders yellow with a pink tip, and a grid of grey/chartreuse cubes renders as pink, cyan, blue and orange.

### Cause

`ModelBuilder._add_shape` falls back to the palette whenever no colour is supplied (`newton/_src/sim/builder.py:6494-6498`):

```python
resolved_color = ModelBuilder._coerce_shape_color(color)
if resolved_color is None and src is not None:
    resolved_color = ModelBuilder._coerce_shape_color(getattr(src, "color", None))
if resolved_color is None:
    resolved_color = ModelBuilder._shape_palette_color(shape)
```

That fallback is right for procedurally built scenes — `example_basic_shapes` and friends rely on it to tell shapes apart, and the builder cannot distinguish "the caller did not bother to set a colour" from "this asset genuinely has no material here". The USD importer *can* make that distinction, so the fix belongs there rather than in the builder.

The palette is also right for collision shapes, where distinguishability is the point. This is only about shapes that carry `VISIBLE`.

### Suggested behaviour

When `add_usd` creates a shape carrying `ShapeFlags.VISIBLE` and resolves no material and no authored `primvars:displayColor`, pass an explicit neutral colour instead of leaving it `None`. Procedurally built scenes keep the palette; imported scenes stop inventing colours.

Worth noting `primvars:displayColor` is not a substitute on its own. On the asset below it is *defined but unauthored* on the affected prims — `HasAuthoredValue()` is `False` and `Get()` returns `None` with no schema fallback — so there is nothing to read. Newton already honours it when it is authored (`newton/_src/usd/utils.py:2944`).

### Reproduction

Any stage with unmaterialed visual prims. Measured on the OpenUSD Exchange sample asset, 105 shapes:

| | count | outcome |
| --- | --- | --- |
| constant `diffuseColor` | 35 | correct |
| interface-connected constant, or textured mesh | 18 | correct |
| **no bound material** | **52** | **debug palette** |

```python
import numpy as np, newton
from pxr import Usd, UsdShade

b = newton.ModelBuilder(up_axis=newton.Axis.Z)
r = b.add_usd("sample.usdc", load_visual_shapes=True)
stage = Usd.Stage.Open("sample.usdc")
col = np.asarray(b.shape_color)

for path, sid in r["path_shape_map"].items():
    prim = stage.GetPrimAtPath(path)
    if prim and prim.IsValid() and not UsdShade.MaterialBindingAPI(prim).ComputeBoundMaterial()[0]:
        print(f"{tuple(round(float(x), 2) for x in col[sid])}  {path}")
```

```
(0.13, 0.53, 0.2)   /World/groundXform/groundCube     <- light grey in the reference render
(0.27, 0.47, 0.67)  /World/cube
(0.8, 0.73, 0.27)   /World/quatCube
(0.93, 0.4, 0.47)   /World/cubeMesh
(0.67, 0.2, 0.47)   /World/referencePrim/Cube_0_0_0
...
```

### Not part of this

Two things I checked while narrowing this down, neither of which is a defect:

- The linear→sRGB conversion in `newton/_src/usd/utils.py:2500` is deliberate and self-consistent — `ViewerGL` has no gamma stage, so writing an already-encoded value straight out is correct. It is why an authored `(0.6, 0.4, 0.2)` is stored as `(0.798, 0.665, 0.485)`, and why the only colours that survive a round trip unchanged are those built from 0.0 and 1.0, the fixed points of the transfer function.
- Textures being mesh-only is by design; procedural primitives carry no UVs.

andrewkaufman:
## Description

Closes #3910.

**The scope is wider than the issue describes, and the issue's proposed fix would have been wrong on its own.** #3910 reported that imported visual shapes with no bound material get a colour from `ModelBuilder`'s per-shape debug palette, and proposed substituting a neutral default. Building that revealed the larger half of the problem: those prims mostly *do* carry a colour, in `primvars:displayColor`, and the importer was never reading it.

`primvars:displayColor` was only consulted after a material had been resolved but supplied no colour. A prim that binds no material at all returns early from `resolve_material_properties_for_prim` with empty properties, so the one case where `displayColor` is the only colour the prim carries was precisely the case that never read it.

So this lands as two commits, and the order matters for review:

1. **`Use a neutral color for unmaterialed visual shapes`** — what #3910 asked for. A visual shape that resolves no colour takes `UsdPreviewSurface`'s `diffuseColor` default of `0.18` linear, which is what USD renderers draw an unmaterialed prim as, rather than a shape-index palette entry. Defined in linear and passed through the same `color_linear_to_srgb` as authored colours instead of hardcoding the encoded value.
2. **`Read displayColor when a prim binds no material`** — the actual bug. Moves the `displayColor` lookup into a helper and calls it from the no-material path too. This supersedes most of the first commit: on the sample asset, 77 prims author `displayColor` and would otherwise have gone grey.

Had only the neutral default landed, the result would have been *differently* wrong — grey where the asset authored chartreuse, green and blue.

The palette itself is left alone. It is correct for procedurally built scenes, where it is what makes shapes distinguishable and `ModelBuilder` genuinely cannot tell "the caller did not set a colour" from "this prim binds no material", and it is correct for collision shapes, where distinguishability is the point. Only the USD importer has the context to decide, so only the USD importer changes.

### Measured

On the [OpenUSD Exchange](https://github.com/NVIDIA-Omniverse/usd-exchange) sample stage — 105 shapes, mixed `UsdPreviewSurface`, MaterialX, textured, unmaterialed, plus `GeomSubset`s and instancing:

| | before | after |
| --- | --- | --- |
| shapes taking a palette colour | 52 | **0** |
| distinct colours in the scene | 18 | 15 |

Spot checks against the authored values:

| prim | source | result |
| --- | --- | --- |
| `/World/cubeMesh` | `displayColor (0.463, 0.725, 0)` | `(0.71, 0.868, 0.0)` |
| `/World/matrixXform/matrixCube` | `displayColor (0.5, 0.7, 1)` | `(0.735, 0.854, 1.0)` |
| `/World/groundXform/groundCube` | neither material nor `displayColor` | `(0.461, 0.461, 0.461)` |

Rendering the stage in the viewer alongside an RTX render of the same stage at matching framing, the colours now agree. The differences that remain are unrelated missing capability — no PBR, no opacity, and one light versus an HDRI with GI.

## Checklist

## Test plan

```
uv run --extra dev --extra examples -m newton.tests -k usd --strict-warnings     # 608 tests, OK
uv run --extra dev --extra examples -m newton.tests -k viewer --strict-warnings  # 210 tests, OK
```

Two regression tests, each failing without its commit:

- `test_unmaterialed_visual_shape_uses_neutral_color` — asserts the neutral, and separately that the three channels are equal, since the palette varies with shape index and a neutral cannot. Without the fix: `0.2667 != 0.4614`.
- `test_display_color_used_without_bound_material` — a cube with authored `displayColor` and no material. Without the fix: `0.4614 != 0.7104`, grey where chartreuse belongs.

## Bug fix

**Steps to reproduce:**

1. Author a stage with a visual prim that binds no material — with `primvars:displayColor` for the second case, without it for the first.
2. `add_usd` the stage and read back `builder.shape_color`.
3. Observe a palette colour that varies with shape index, rather than the authored display colour or a neutral.

**Minimal reproduction:**

```python
import newton
from pxr import Usd, UsdGeom

UsdGeom.Xform.Define(stage, "/World")
UsdGeom.Cube.Define(stage, "/World/Bare")
UsdGeom.Cube.Define(stage, "/World/Colored").GetDisplayColorAttr().Set([(0.463, 0.725, 0.0)])

result = builder.add_usd(stage, load_visual_shapes=True)
for path in ("/World/Bare", "/World/Colored"):
    print(path, builder.shape_color[result["path_shape_map"][path]])

# before: two unrelated palette colours
# after:  (0.4614, 0.4614, 0.4614) and (0.7104, 0.8677, 0.0)
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Imported visual and collision shapes without materials or textures now display a consistent dark-gray fallback color.
  - Authored and inherited display colors are now respected when importing shapes.
  - Color values are validated and converted correctly for display.
  - Textured meshes retain their expected white builder color while displaying their textures properly.

### DIFF
--- a/newton/_src/usd/utils.py
+++ b/newton/_src/usd/utils.py
@@ -2600,6 +2600,10 @@ def _coerce_color(value: Any) -> tuple[float, float, float] | None:
     """Coerce a value to an RGB color tuple, or None if not possible."""
     if value is None:
         return None
+    # A per-vertex or per-face primvar holds one entry per element and only the leading one
+    # is used, so take it before the numpy conversion rather than flattening the whole array.
+    if hasattr(value, "__len__") and len(value) > 0 and hasattr(value[0], "__len__"):
+        value = value[0]
     color_np = np.array(value, dtype=np.float32).reshape(-1)
     if color_np.size >= 3:
         return (float(color_np[0]), float(color_np[1]), float(color_np[2]))
--- a/newton/_src/usd/utils.py
+++ b/newton/_src/usd/utils.py
@@ -2941,13 +2941,22 @@ def _resolve_prim_material_properties(target_prim: Usd.Prim) -> dict[str, Any] |
         if properties.get(key) is None and material_props.get(key) is not None:
             properties[key] = material_props[key]
     if properties["color"] is None and properties["texture"] is None:
-        display_color = UsdGeom.PrimvarsAPI(target_prim).GetPrimvar("displayColor")
-        if display_color:
-            color = _coerce_color(display_color.Get())
-            if color is not None:
-                properties["color"] = _color_to_display_space(color, display_color.GetAttr())
+        properties["color"] = _display_color_for_prim(target_prim)
 
     return properties
+
+
+def _display_color_for_prim(prim: Usd.Prim) -> tuple[float, float, float] | None:
+    """Read ``primvars:displayColor`` off a prim, in Newton's display color space."""
+    if UsdGeom is None or not prim or not prim.IsValid():
+        return None
+    display_color = UsdGeom.PrimvarsAPI(prim).GetPrimvar("displayColor")
+    if not display_color:
+        return None
+    color = _coerce_color(display_color.Get())
+    if color is None:
+        return None
+    return _color_to_display_space(color, display_color.GetAttr())
 
 
 def resolve_material_properties_for_prim(prim: Usd.Prim) -> dict[str, Any]:
@@ -3003,7 +3012,12 @@ def resolve_material_properties_for_prim(prim: Usd.Prim) -> dict[str, Any]:
             if fallback_props is not None:
                 return fallback_props
 
-    return _empty_material_properties()
+    # No material is bound anywhere, which is exactly when ``primvars:displayColor`` is the
+    # only color the prim carries. The fallback above only runs once a material has been
+    # resolved, so it never reaches these prims.
+    properties = _empty_material_properties()
+    properties["color"] = _display_color_for_prim(prim)
+    return properties
 
 
 def get_gaussian(prim: Usd.Prim, min_response: float = 0.1) -> Gaussian:
--- a/newton/_src/usd/utils.py
+++ b/newton/_src/usd/utils.py
@@ -2947,10 +2947,14 @@ def _resolve_prim_material_properties(target_prim: Usd.Prim) -> dict[str, Any] |
 
 
 def _display_color_for_prim(prim: Usd.Prim) -> tuple[float, float, float] | None:
-    """Read ``primvars:displayColor`` off a prim, in Newton's display color space."""
+    """Read ``primvars:displayColor`` off a prim, in Newton's display color space.
+
+    Resolved with inheritance: a ``constant`` primvar applies to every descendant, so an
+    ancestor is a legitimate place to author the color for a whole subtree.
+    """
     if UsdGeom is None or not prim or not prim.IsValid():
         return None
-    display_color = UsdGeom.PrimvarsAPI(prim).GetPrimvar("displayColor")
+    display_color = UsdGeom.PrimvarsAPI(prim).FindPrimvarWithInheritance("displayColor")
     if not display_color:
         return None
     color = _coerce_color(display_color.Get())
--- a/newton/_src/utils/import_usd.py
+++ b/newton/_src/utils/import_usd.py
@@ -1368,9 +1368,11 @@ def parse_usd(
         visual_shape_cfg_for_prim.is_visible = is_site or _is_viewport_drawn(prim)
         material_props = _get_material_props_cached(prim)
         shape_color = material_props.get("color")
-        # A textured prim also resolves no scalar color, deliberately, so that the texture
-        # is not tinted. Substituting the neutral there would multiply into every texel.
-        if shape_color is None and material_props.get("texture") is None and visual_shape_cfg_for_prim.is_visible:
+        # A textured mesh resolves no scalar color on purpose, so the texture is not tinted;
+        # the mesh path gives it white. Geometry that never receives the texture still wants
+        # the neutral, otherwise it falls through to a palette color.
+        carries_texture = material_props.get("texture") is not None and type_name == "mesh"
+        if shape_color is None and not carries_texture and visual_shape_cfg_for_prim.is_visible:
             shape_color = _UNMATERIALED_VISUAL_COLOR
 
         if path_name not in path_shape_map:
@@ -3572,7 +3574,8 @@ def parse_usd(
                 shape_ka = shape_contact["ka"]
 
                 shape_color = material_props.get("color")
-                if shape_color is None and material_props.get("texture") is None and collider_is_visible:
+                carries_texture = material_props.get("texture") is not None and key == UsdPhysics.ObjectType.MeshShape
+                if shape_color is None and not carries_texture and collider_is_visible:
                     shape_color = _UNMATERIALED_VISUAL_COLOR
 
                 # SDF parameters. Applying NewtonSDFCollisionAPI is the canonical
--- a/newton/_src/utils/import_usd.py
+++ b/newton/_src/utils/import_usd.py
@@ -1368,7 +1368,9 @@ def parse_usd(
         visual_shape_cfg_for_prim.is_visible = is_site or _is_viewport_drawn(prim)
         material_props = _get_material_props_cached(prim)
         shape_color = material_props.get("color")
-        if shape_color is None and visual_shape_cfg_for_prim.is_visible:
+        # A textured prim also resolves no scalar color, deliberately, so that the texture
+        # is not tinted. Substituting the neutral there would multiply into every texel.
+        if shape_color is None and material_props.get("texture") is None and visual_shape_cfg_for_prim.is_visible:
             shape_color = _UNMATERIALED_VISUAL_COLOR
 
         if path_name not in path_shape_map:
@@ -3570,7 +3572,7 @@ def parse_usd(
                 shape_ka = shape_contact["ka"]
 
                 shape_color = material_props.get("color")
-                if shape_color is None and collider_is_visible:
+                if shape_color is None and material_props.get("texture") is None and collider_is_visible:
                     shape_color = _UNMATERIALED_VISUAL_COLOR
 
                 # SDF parameters. Applying NewtonSDFCollisionAPI is the canonical
--- a/newton/_src/utils/import_usd.py
+++ b/newton/_src/utils/import_usd.py
@@ -53,6 +53,7 @@ from ..usd import require_newton_usd_schemas
 from ..usd import utils as usd
 from ..usd.schema_resolver import PrimType, SchemaResolver, SchemaResolverManager
 from ..usd.schemas import SchemaResolverNewton
+from .color import color_linear_to_srgb
 from .import_usd_deformable_attachments import (
     _deformable_import_attachments,
     _deformable_import_element_collision_filters,
@@ -75,6 +76,12 @@ _NEWTON_SRC_DIR = os.path.normpath(os.path.join(os.path.dirname(__file__), os.pa
 
 # Stiffness used for a hard joint limit (NewtonJointAPI newton:limitStiffness == +inf).
 _HARD_LIMIT_KE = 1.0e8
+
+# `UsdPreviewSurface`'s schema default for `diffuseColor`. A visual shape whose prim binds no
+# material is given this rather than left for ModelBuilder's per-shape debug palette, which
+# would render an unmaterialed scene in colours the asset never authored. Display-encoded to
+# match the colours that are resolved from a material.
+_UNMATERIALED_VISUAL_COLOR = color_linear_to_srgb((0.18, 0.18, 0.18))
 
 
 def _resolve_newton_limit_ke(
@@ -1361,6 +1368,8 @@ def parse_usd(
         visual_shape_cfg_for_prim.is_visible = is_site or _is_viewport_drawn(prim)
         material_props = _get_material_props_cached(prim)
         shape_color = material_props.get("color")
+        if shape_color is None and visual_shape_cfg_for_prim.is_visible:
+            shape_color = _UNMATERIALED_VISUAL_COLOR
 
         if path_name not in path_shape_map:
             if type_name == "cube":
@@ -3561,6 +3570,8 @@ def parse_usd(
                 shape_ka = shape_contact["ka"]
 
                 shape_color = material_props.get("color")
+                if shape_color is None and collider_is_visible:
+                    shape_color = _UNMATERIALED_VISUAL_COLOR
 
                 # SDF parameters. Applying NewtonSDFCollisionAPI is the canonical
                 # signal that SDF generation is configured for this shape.
--- a/changelog/+usd-unmaterialed-color-7c3e9a15.fixed.md
+++ b/changelog/+usd-unmaterialed-color-7c3e9a15.fixed.md
@@ -1,1 +1,0 @@
-Honor `primvars:displayColor` on imported prims that bind no material, and give those that carry neither a material nor a display color the `UsdPreviewSurface` default instead of a shape-index color from `ModelBuilder`'s debug palette, so a stage renders in the colors it authored rather than colors it never did.
--- a/changelog/+usd-unmaterialed-color-7c3e9a15.fixed.md
+++ b/changelog/+usd-unmaterialed-color-7c3e9a15.fixed.md
@@ -1,0 +1,1 @@
+Give imported visual shapes that bind no material the `UsdPreviewSurface` default color rather than a shape-index color from `ModelBuilder`'s debug palette, so a stage renders in the colors it authored instead of colors it never did.
--- a/changelog/+usd-unmaterialed-color-7c3e9a15.fixed.md
+++ b/changelog/+usd-unmaterialed-color-7c3e9a15.fixed.md
@@ -1,1 +1,1 @@
-Give imported visual shapes that bind no material the `UsdPreviewSurface` default color rather than a shape-index color from `ModelBuilder`'s debug palette, so a stage renders in the colors it authored instead of colors it never did.
+Honor `primvars:displayColor` on imported prims that bind no material, and give those that carry neither a material nor a display color the `UsdPreviewSurface` default instead of a shape-index color from `ModelBuilder`'s debug palette, so a stage renders in the colors it authored rather than colors it never did.
--- a/changelog/3910.fixed.md
+++ b/changelog/3910.fixed.md
@@ -1,0 +1,1 @@
+Use authored colors for imported shapes that bind no material, rather than a debug palette color.
--- a/newton/tests/test_import_usd.py
+++ b/newton/tests/test_import_usd.py
@@ -7398,9 +7398,13 @@ def Xform "Articulation" (
         builder = newton.ModelBuilder()
         result = builder.add_usd(stage)
 
-        src = builder.shape_source[result["path_shape_map"]["/Body/VisualMesh"]]
+        shape = result["path_shape_map"]["/Body/VisualMesh"]
+        src = builder.shape_source[shape]
         self.assertIsNotNone(src.texture)
         np.testing.assert_allclose(np.array(src.color), np.array([1.0, 1.0, 1.0]))
+        # The viewer reads shape_color, and ModelBuilder prefers it over src.color, so the
+        # white has to survive there too or the shader multiplies it into every texel.
+        np.testing.assert_allclose(np.array(builder.shape_color[shape]), np.array([1.0, 1.0, 1.0]))
 
     @staticmethod
     def _build_uvless_textured_visual_mesh_stage(*, material_subset: bool):
--- a/newton/tests/test_import_usd.py
+++ b/newton/tests/test_import_usd.py
@@ -32,6 +32,7 @@ from newton._src.solvers.mujoco.constants import (
     SOLREF_MODE_RAW,
 )
 from newton._src.solvers.mujoco.utils import MjcEqualityTargetKind
+from newton._src.utils.color import color_linear_to_srgb
 from newton._src.utils.import_usd import _is_uniform_scale
 from newton.math import quat_between_axes
 from newton.solvers import SolverMuJoCo
@@ -12440,6 +12441,36 @@ def Xform "Body" (
         self.assertEqual(builder.shape_count, 2)
         drawn = [s for s in range(builder.shape_count) if builder.shape_flags[s] & ShapeFlags.VISIBLE]
         self.assertEqual(drawn, [path_shape_map["/Body/VisualSphere"]])
+
+    @unittest.skipUnless(USD_AVAILABLE, "Requires usd-core")
+    def test_unmaterialed_visual_shape_uses_neutral_color(self):
+        """Verify a visual prim with no bound material gets the neutral default, not the palette.
+
+        ModelBuilder colors an uncolored shape from a per-shape debug palette, which suits
+        procedurally built scenes but makes an imported stage render in colors the asset
+        never authored. USD renderers draw an unmaterialed prim in UsdPreviewSurface's
+        ``diffuseColor`` default instead, so the importer supplies that.
+        """
+        from pxr import Usd, UsdGeom
+
+        stage = Usd.Stage.CreateInMemory()
+        UsdGeom.SetStageUpAxis(stage, UsdGeom.Tokens.z)
+        UsdGeom.Xform.Define(stage, "/World")
+        UsdGeom.Cube.Define(stage, "/World/Bare")
+
+        builder = newton.ModelBuilder()
+        result = builder.add_usd(stage, load_visual_shapes=True)
+        shape = result["path_shape_map"]["/World/Bare"]
+
+        self.assertTrue(builder.shape_flags[shape] & ShapeFlags.VISIBLE)
+        color = builder.shape_color[shape]
+        # 0.18 linear, display-encoded the same way an authored color would be.
+        expected = color_linear_to_srgb((0.18, 0.18, 0.18))
+        for channel, want in zip(color, expected, strict=True):
+            self.assertAlmostEqual(channel, want, places=5)
+        # Guard the actual defect: the palette varies with shape index, a neutral does not.
+        self.assertAlmostEqual(color[0], color[1], places=6)
+        self.assertAlmostEqual(color[1], color[2], places=6)
 
     @unittest.skipUnless(USD_AVAILABLE, "Requires usd-core")
     def test_hide_collision_shapes_fallback_with_material(self):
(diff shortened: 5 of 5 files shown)


newton/_src/usd/utils.py
```diff
@@ -2941,15 +2941,24 @@ def _resolve_prim_material_properties(target_prim: Usd.Prim) -> dict[str, Any] |
...
+        properties["color"] = _display_color_for_prim(target_prim)
 
     return properties
 
 
+def _display_color_for_prim(prim: Usd.Prim) -> tuple[float, float, float] | None:
+    """Read ``primvars:displayColor`` off a prim, in Newton's display color space."""
+    if UsdGeom is None or not prim or not prim.IsValid():
+        return None
+    display_color = UsdGeom.PrimvarsAPI(prim).GetPrimvar("displayColor")
```
andrewkaufman: This is not the correct way to get display color. constant primvars inherit down hierarchy

andrewkaufman: You're right, fixed in `2db9676`. Switched to `FindPrimvarWithInheritance`, which is also what MuJoCo's USD decoder uses for this.

The failure mode was quieter than a plain miss, which is why it got past me. On a Gprim, `GetPrimvar("displayColor")` returns a primvar that is **truthy** — the schema defines the attribute — but whose `Get()` is `None`. So the guard passed and the value read as "no color authored" rather than "look further up":

```
constant displayColor on /World, nothing on /World/Child:
  GetPrimvar on child        : True   None
  FindPrimvarWithInheritance : True   [(0.1, 0.2, 0.3)]
```

Since both call sites go through the helper this PR introduced, this also fixes the pre-existing lookup on the has-a-material-but-no-color path, which had the same bug.

Added `test_display_color_inherited_from_ancestor`: a constant primvar on an Xform with a bare cube beneath it. Fails without the change with `0.4614 != 0.3492` — the neutral default where the inherited color belongs.

609 USD tests and 210 viewer tests pass.

newton/_src/usd/utils.py
```diff
@@ -2941,15 +2941,28 @@ def _resolve_prim_material_properties(target_prim: Usd.Prim) -> dict[str, Any] |
...
+
+    Resolved with inheritance: a ``constant`` primvar applies to every descendant, so an
+    ancestor is a legitimate place to author the color for a whole subtree.
+    """
+    if UsdGeom is None or not prim or not prim.IsValid():
+        return None
+    display_color = UsdGeom.PrimvarsAPI(prim).FindPrimvarWithInheritance("displayColor")
+    if not display_color:
+        return None
+    color = _coerce_color(display_color.Get())
```
eric-heiden, line 2964: ⚪ The inheritance fix here is the right call, and none of this is a merge blocker, but it does move some import-time cost onto every prim that binds no material. One piece of that is easy to make constant time.

A synthetic stage of 1500 unmaterialed cubes that author `displayColor` imports about 27% slower. Interleaved in-process A/B (ABBA order, min of 8), with a null control (identical code on both sides) measuring this machine's noise floor, so a delta only counts as signal if it clearly beats the control:

| Stage (prims) | This PR | Main | Delta | Control | Verdict |
|---|---|---|---|---|---|
| cubes_displaycolor (1500) | 985.2 ms | 776.9 ms | **+26.8%** | -8.8% | signal |
| cubes_bare (1500) | 741.9 ms | 749.6 ms | -1.0% | -4.8% | noise |
| cubes_deep (1500, 6 levels) | 901.9 ms | 938.9 ms | -3.9% | +4.5% | noise |
| meshes_plain (150) | 505.8 ms | 486.9 ms | +3.9% | -7.0% | noise |
| meshes_vertexcolor (150) | 510.1 ms | 497.0 ms | +2.6% | +3.2% | noise |

Timing `resolve_material_properties_for_prim` on its own (A/B by checking the two changed files out from `main`, min of 11) separates two mechanisms:

| Stage (prims) | This PR | Main | Delta |
|---|---|---|---|
| cubes_displaycolor (1500) | 85.65 ms | 21.12 ms | +306% |
| meshes_vertexcolor (150) | 7.64 ms | 2.25 ms | +239% |
| cubes_bare (1500) | 37.09 ms | 19.63 ms | +89% |
| meshes_plain (150) | 4.22 ms | 2.25 ms | +88% |
| cubes_deep (1500) | 48.17 ms | 26.61 ms | +81% |

**1. The lookup is paid even when it misses.** `FindPrimvarWithInheritance` walks the ancestor chain to the pseudo-root before giving up, where `GetPrimvar` only inspected the prim itself. So prims that carry no `displayColor` anywhere now pay roughly +89% inside `resolve_material_properties_for_prim` for a lookup that returns nothing. That is the cost of resolving inheritance correctly, and it stays inside the noise floor end to end, so I would leave it as is. Just worth knowing it is there.

**2. The color extraction is O(n) in the array length.** This is the actionable one. `_coerce_color` does `np.array(value).reshape(-1)` over the whole `displayColor` array just to read three floats, so a per-vertex or per-face color costs time proportional to the vertex count:

| displayColor values | `_coerce_color` today | leading element only | speedup |
|---|---|---|---|
| 3,600 | 3.1 µs | 3.3 µs | 0.9x |
| 60,000 | 36.9 µs | 3.4 µs | 11x |
| 300,000 | 241.2 µs | 3.3 µs | 72x |
| 900,000 | 723.5 µs | 3.3 µs | 218x |

`_coerce_color` is not touched by this PR, but on `main` an unmaterialed prim never reached it, so this cost is newly exposed here. Taking the leading element before the numpy round trip makes it constant time:

```python
def _coerce_color(value: Any) -> tuple[float, float, float] | None:
    """Coerce a value to an RGB color tuple, or None if not possible."""
    if value is None:
        return None
    # displayColor is commonly per-vertex or per-face, and only the leading element is
    # needed, so avoid converting the whole array to read three floats.
    if hasattr(value, "__len__") and len(value) > 0 and hasattr(value[0], "__len__"):
        value = value[0]
    color_np = np.array(value, dtype=np.float32).reshape(-1)
    if color_np.size >= 3:
        return (float(color_np[0]), float(color_np[1]), float(color_np[2]))
    return None
```

Output is identical on every size in the table above, and on a bare `Gf.Vec3f`, a flat tuple, and a single element color array. The added checks cost about 1.5 µs on the smallest arrays, which is nothing against the ~700 µs saved on a large one.

Caching is not the thing to reach for here: instrumenting an import showed exactly one `resolve_material_properties_for_prim` and one `_display_color_for_prim` call per prim, so there is no repeated lookup to eliminate.

One small extra: the changelog fragment could note that prims with no bound material now perform a primvar lookup, since that is the part of this change a user would feel on a large stage.

<details>

<summary>Benchmark script: import, isolated resolve, and array length scaling</summary>

Run as `python bench_usd_import_displaycolor.py ab` for the end to end A/B with the null control, `micro` for the isolated `resolve_material_properties_for_prim` numbers, and `scale` for per-call cost against array length. For `micro`, A/B by running it, then `git checkout origin/main -- newton/_src/usd/utils.py newton/_src/utils/import_usd.py`, running again, then restoring with `git checkout HEAD -- <same files>`.

```python
"""Benchmark the USD import path this PR touches: displayColor on unmaterialed prims.

The PR adds a ``primvars:displayColor`` read to ``resolve_material_properties_for_prim``'s
no-material-bound tail, so every prim without a bound material now pays a primvar lookup
(and, when authored, a full ``Get()`` + numpy coercion of the color array).

Modes (argv[1]):
  micro  resolve_material_properties_for_prim only -- the tight, reproducible number.
  import end-to-end ModelBuilder.add_usd (host-noise limited).
  ab     end-to-end, interleaved in-process A/B (ABBA) with the new lookup toggled off.
  scale  per-call cost vs. displayColor array length.
"""

import statistics
import sys
import time

import warp as wp

wp.config.quiet = True

import newton  # noqa: E402
from newton._src.usd import utils as usd_utils  # noqa: E402
from pxr import Gf, Usd, UsdGeom, Vt  # noqa: E402

N_CUBES = 1500
N_MESHES = 150
N_TRIS = 1200  # 3600 points per mesh


def build_stage(kind: str) -> Usd.Stage:
    UsdGeom.Xform.Define(stage, "/World")
    if kind == "cubes_deep":
        # Bare prims nested 6 deep: FindPrimvarWithInheritance walks every ancestor on a miss.
        for i in range(N_CUBES):
            path = f"/World/A{i}/B/C/D/E"
            for d in range(1, 5):
                UsdGeom.Xform.Define(stage, "/".join(path.split("/")[: d + 2]))
            UsdGeom.Cube.Define(stage, path + "/Cube")
    elif kind.startswith("cubes"):
        for i in range(N_CUBES):
            cube = UsdGeom.Cube.Define(stage, f"/World/C{i}")
            if kind == "cubes_displaycolor":
                cube.GetDisplayColorAttr().Set([(0.463, 0.725, 0.0)])
    else:
        pts = Vt.Vec3fArray([Gf.Vec3f(i % 7, (i * 3) % 5, (i * 11) % 3) for i in range(N_TRIS * 3)])
        counts = Vt.IntArray([3] * N_TRIS)
        idx = Vt.IntArray(list(range(N_TRIS * 3)))
        # Per-vertex displayColor: the array main never fetched for an unmaterialed prim.
        cols = Vt.Vec3fArray([Gf.Vec3f(0.5, 0.5, 0.5)] * (N_TRIS * 3))
        for i in range(N_MESHES):
            m = UsdGeom.Mesh.Define(stage, f"/World/M{i}")
            m.GetPointsAttr().Set(pts)
            m.GetFaceVertexCountsAttr().Set(counts)
            m.GetFaceVertexIndicesAttr().Set(idx)
            if kind == "meshes_vertexcolor":
                m.GetDisplayColorAttr().Set(cols)
                m.GetDisplayColorPrimvar().SetInterpolation(UsdGeom.Tokens.vertex)
    return stage


def timeit(fn, repeat=7):
    samples = []
    for _ in range(repeat):
        t0 = time.perf_counter()
        fn()
        samples.append(time.perf_counter() - t0)
    return min(samples), statistics.median(samples)


def scale_test():
    """Per-call cost of the new read vs. displayColor array length."""
    stages = []
    for nverts in (3, 3600, 60000, 300000, 900000):
        stages.append(stage)  # kept alive: prims need the stage
        ntri = max(nverts // 3, 1)
        m = UsdGeom.Mesh.Define(stage, "/M")
        m.GetPointsAttr().Set(Vt.Vec3fArray([Gf.Vec3f(i % 7, i % 5, i % 3) for i in range(ntri * 3)]))
        m.GetFaceVertexCountsAttr().Set(Vt.IntArray([3] * ntri))
        m.GetFaceVertexIndicesAttr().Set(Vt.IntArray(list(range(ntri * 3))))
        m.GetDisplayColorAttr().Set(Vt.Vec3fArray([Gf.Vec3f(0.5, 0.5, 0.5)] * (ntri * 3)))
        m.GetDisplayColorPrimvar().SetInterpolation(UsdGeom.Tokens.vertex)
        prim = stage.GetPrimAtPath("/M")
        assert usd_utils._display_color_for_prim(prim) is not None, "color not read"
        lo, med = timeit(lambda p=prim: usd_utils._display_color_for_prim(p), repeat=20)
        print(f"SCALE displayColor_verts={nverts:7d} per_call_min={lo * 1e6:8.1f}us med={med * 1e6:8.1f}us")


def main():
    # "micro" isolates resolve_material_properties_for_prim (no ModelBuilder allocation,
    # so GC pressure from retained builders cannot confound the per-prim cost).
    mode = sys.argv[1] if len(sys.argv) > 1 else "all"
    if mode == "scale":
        scale_test()
        return
    for kind in ("cubes_bare", "cubes_deep", "cubes_displaycolor", "meshes_plain", "meshes_vertexcolor"):
        stage = build_stage(kind)
        prims = [p for p in stage.Traverse() if p.IsA(UsdGeom.Gprim)]
        lo = med = mlo = mmed = float("nan")

        if mode in ("all", "import"):
            newton.ModelBuilder().add_usd(stage, load_visual_shapes=True)  # warmup
            lo, med = timeit(lambda s=stage: newton.ModelBuilder().add_usd(s, load_visual_shapes=True))

        if mode == "ab":
            # Interleaved in-process A/B: "off" reproduces main (no displayColor read on an
            # unmaterialed prim); downstream work is identical since both yield a color tuple.
            # A null control (real vs. real, same interleaving) measures this host's noise
            # floor -- cross-process runs on this box drift by tens of percent, so a delta
            # only counts as signal if it clearly exceeds |control|.
            real = usd_utils._display_color_for_prim
            build = lambda s=stage: newton.ModelBuilder().add_usd(s, load_visual_shapes=True)  # noqa: E731
            on, off, ctl_a, ctl_b = [], [], [], []
            build()

            def sample(fn, bucket):
                usd_utils._display_color_for_prim = fn
                t = time.perf_counter()
                build()
                bucket.append(time.perf_counter() - t)

            disabled = lambda _p: None  # noqa: E731
            for i in range(8):  # ABBA order so allocation/GC drift cannot bias one side
                order = (real, on, disabled, off) if i % 2 == 0 else (disabled, off, real, on)
                sample(order[0], order[1])
                sample(order[2], order[3])
                ctl = (ctl_a, ctl_b) if i % 2 == 0 else (ctl_b, ctl_a)
                sample(real, ctl[0])  # null control: identical code both sides
                sample(real, ctl[1])
            usd_utils._display_color_for_prim = real
            d = (min(on) - min(off)) / min(off) * 100
            noise = (min(ctl_a) - min(ctl_b)) / min(ctl_b) * 100
            print(
                f"AB {kind:22s} prims={len(prims):5d} main_min={min(off) * 1e3:8.2f}ms "
                f"pr_min={min(on) * 1e3:8.2f}ms delta={d:+6.1f}% control={noise:+5.1f}% "
                f"main_med={statistics.median(off) * 1e3:8.2f}ms pr_med={statistics.median(on) * 1e3:8.2f}ms"
            )
            continue

        if mode in ("all", "micro"):

            def resolve(ps=prims):
                for p in ps:
                    usd_utils.resolve_material_properties_for_prim(p)

            resolve()
            mlo, mmed = timeit(resolve, repeat=11)
        print(
            f"RESULT {kind:22s} prims={len(prims):5d} "
            f"add_usd_min={lo * 1e3:8.2f}ms add_usd_med={med * 1e3:8.2f}ms "
            f"resolve_min={mlo * 1e3:8.3f}ms resolve_med={mmed * 1e3:8.3f}ms"
        )


if __name__ == "__main__":
    main()
```

</details>

<details>

<summary>Equality and speedup check for the suggested `_coerce_color` change</summary>

Asserts identical output before timing, so the speedup is a like for like comparison and not a behavior change.

```python
"""Compare _coerce_color's full-array numpy round-trip against slicing the leading element."""

import time
from typing import Any

import numpy as np
from pxr import Gf, Usd, UsdGeom, Vt


def coerce_current(value: Any) -> tuple[float, float, float] | None:
    """Reproduce _coerce_color as it stands today."""
    if value is None:
        return None
    color_np = np.array(value, dtype=np.float32).reshape(-1)
    if color_np.size >= 3:
        return (float(color_np[0]), float(color_np[1]), float(color_np[2]))
    return None


def coerce_slice_first(value: Any) -> tuple[float, float, float] | None:
    """Take the leading element of an array-valued primvar before converting."""
    if value is None:
        return None
    if hasattr(value, "__len__") and len(value) > 0 and hasattr(value[0], "__len__"):
        value = value[0]
    color_np = np.array(value, dtype=np.float32).reshape(-1)
    if color_np.size >= 3:
        return (float(color_np[0]), float(color_np[1]), float(color_np[2]))
    return None


def _timed(fn):
    t0 = time.perf_counter()
    fn()
    return time.perf_counter() - t0


def timeit(fn, repeat=20):
    return min(_timed(fn) for _ in range(repeat))


def main():
    # Shapes _coerce_color is actually handed: a bare Gf.Vec3f, a flat tuple, a
    # single-element color array (constant interpolation), and per-vertex arrays.
    fixed = [
        ("Gf.Vec3f", Gf.Vec3f(0.463, 0.725, 0.0)),
        ("tuple", (0.463, 0.725, 0.0)),
        ("Vec3fArray[1]", Vt.Vec3fArray([Gf.Vec3f(0.463, 0.725, 0.0)])),
    ]
    for label, value in fixed:
        a, b = coerce_current(value), coerce_slice_first(value)
        assert a == b, f"{label}: {a} != {b}"
        print(f"EQUAL {label:16s} -> {a}")

    for nverts in (3, 3600, 60000, 300000, 900000):
        value = Vt.Vec3fArray([Gf.Vec3f(0.463, 0.725, 0.0)] * nverts)
        a, b = coerce_current(value), coerce_slice_first(value)
        assert a == b, f"n={nverts}: {a} != {b}"
        cur = timeit(lambda v=value: coerce_current(v))
        new = timeit(lambda v=value: coerce_slice_first(v))
        print(
            f"SCALE verts={nverts:7d} current={cur * 1e6:8.1f}us slice_first={new * 1e6:6.1f}us "
            f"speedup={cur / new:6.1f}x equal={a == b}"
        )

    # End of the path that matters: the same comparison through a real primvar Get().
    mesh = UsdGeom.Mesh.Define(stage, "/M")
    mesh.GetDisplayColorAttr().Set(Vt.Vec3fArray([Gf.Vec3f(0.463, 0.725, 0.0)] * 300000))
    mesh.GetDisplayColorPrimvar().SetInterpolation(UsdGeom.Tokens.vertex)
    primvar = UsdGeom.PrimvarsAPI(stage.GetPrimAtPath("/M")).FindPrimvarWithInheritance("displayColor")
    got = primvar.Get()
    assert coerce_current(got) == coerce_slice_first(got)
    print(f"EQUAL via_primvar_get   -> {coerce_slice_first(got)}")


if __name__ == "__main__":
    main()
```

Output:

```
EQUAL Gf.Vec3f         -> (0.46299999952316284, 0.7250000238418579, 0.0)
EQUAL tuple            -> (0.46299999952316284, 0.7250000238418579, 0.0)
EQUAL Vec3fArray[1]    -> (0.46299999952316284, 0.7250000238418579, 0.0)
SCALE verts=      3 current=     1.8us slice_first=   3.3us speedup=   0.5x equal=True
SCALE verts=   3600 current=     3.1us slice_first=   3.3us speedup=   0.9x equal=True
SCALE verts=  60000 current=    36.9us slice_first=   3.4us speedup=  10.9x equal=True
SCALE verts= 300000 current=   241.2us slice_first=   3.3us speedup=  72.4x equal=True
SCALE verts= 900000 current=   723.5us slice_first=   3.3us speedup= 218.1x equal=True
EQUAL via_primvar_get   -> (0.46299999952316284, 0.7250000238418579, 0.0)
```

</details>

newton/_src/utils/import_usd.py
```diff
@@ -1361,6 +1368,8 @@ def _load_visual_shapes_impl(
         visual_shape_cfg_for_prim.is_visible = is_site or _is_viewport_drawn(prim)
         material_props = _get_material_props_cached(prim)
         shape_color = material_props.get("color")
+        if shape_color is None and visual_shape_cfg_for_prim.is_visible:
+            shape_color = _UNMATERIALED_VISUAL_COLOR
```
eric-heiden: 🟠 This fallback also fires for textured prims, so textured imports lose their white base color.

When `diffuseColor` is connected to a `UsdUVTexture`, `_extract_preview_surface_properties` in `newton/_src/usd/utils.py` records the texture and deliberately leaves the scalar color unset, so `material_props.get("color")` is `None` for every textured prim. That is exactly the case this branch treats as "no material", so textured prims now pick up the neutral gray.

The gray then wins over the white that the mesh path sets on purpose. `_get_mesh_with_visual_material` sets `mesh.color = (1.0, 1.0, 1.0)` when a texture is present ("The texture provides albedo, so avoid tinting it with a scalar color"), but `ModelBuilder._add_shape` prefers the explicit `color=` argument over `src.color`, so `builder.shape_color` ends up at roughly `(0.461, 0.461, 0.461)`. The viewer reads `model.shape_color` before falling back to `geo_src.color`, and the GL shader multiplies the object color into the sampled texel (`albedo = pow(ObjectColor, 2.2)`, then `albedo *= pow(tex_color, 2.2)`), so any textured USD asset now renders at 0.18x linear intensity.

Skipping the fallback when a texture was resolved keeps both behaviors:

```suggestion
        if shape_color is None and material_props.get("texture") is None and visual_shape_cfg_for_prim.is_visible:
            shape_color = _UNMATERIALED_VISUAL_COLOR
```

The collider block further down (`if shape_color is None and collider_is_visible:`) needs the same guard, since drawn textured colliders also go through `_get_mesh_with_visual_material`.

It would also help to extend `test_textured_visual_mesh_uses_white_base_color` in `newton/tests/test_import_usd.py`. It asserts on `builder.shape_source[...].color`, which is still white, so it keeps passing. Adding an assertion that `builder.shape_color[shape]` is `(1, 1, 1)` checks the invariant where the viewer actually reads it.

andrewkaufman: Good catch, and it was worse than a wrong shade — fixed in `9a50162`.

Reproduced before changing anything: every textured mesh in the OpenUSD Exchange sample came out at `(0.461, 0.461, 0.461)` while `src.color` was still `(1, 1, 1)`, exactly the path you traced. Applied the guard at both sites; all 6 textured shapes in that stage are back to white.

Also took your point about the test. It asserted on `shape_source[...].color`, which never moved, so it passed throughout. It now also asserts `builder.shape_color[shape]`, and with that added it does fail on the unfixed code — `Mismatched elements: 3 / 3` — so it would have caught this.

newton/_src/usd/utils.py
andrewkaufman: Took the `_coerce_color` half, in `9a50162`. Confirms your numbers:

| displayColor values | before | after |
| --- | --- | --- |
| 3,600 | 6.4 us | 6.4 us |
| 300,000 | 241 us | 3.5 us |
| 900,000 | 723 us | 2.7 us |

Identical results for `Gf.Vec3f`, plain tuple, `Vt.Vec3fArray` and a `(3,)` ndarray, so the leading-element slice does not change behaviour for any of the shapes this receives.

Left the miss cost alone, as you suggested. It is the price of resolving inheritance correctly, and your end-to-end table has it inside the noise floor. Worth noting it is also gated: the lookup only runs when a prim resolves no color and no texture, and it is memoized per prim path by `_get_material_props_cached`, so it is one walk per unmaterialed prim rather than per shape.

If it does show up in a real asset, the next step is caching per ancestor path rather than per prim, since siblings share the answer. I would rather wait for a profile than build that on spec.

Thanks for the ABBA-with-control methodology — that is what made the 27% legible as signal rather than noise.

andrewkaufman:
Addressed the fragment naming from the review summary: renamed to `3910.fixed.md` so Towncrier links it to the issue this closes, per `changelog/README.md`. Also cut the entry to one line, since it was several times the length of the surrounding fix entries.

Both of @eric-heiden's findings are in as well:

- Textured prims were picking up the neutral, because a texture deliberately leaves the scalar color unset. Reproduced it on the OpenUSD Exchange sample — every textured mesh at `(0.461, 0.461, 0.461)` with `src.color` still white — then narrowed the guard. One follow-up beyond the suggestion: guarding on "the material has a texture" was too broad, since only a mesh receives one. A sphere bound to an MDL projection shader resolved no color, was skipped, and fell through to a palette entry, which is the exact thing this PR removes. The guard now checks that the shape is a mesh.
- `_coerce_color` is constant time now: 900k values go from ~723us to ~2.7us, identical results for `Gf.Vec3f`, tuple, `Vt.Vec3fArray` and ndarray inputs. Left the miss cost alone as suggested.

Palette use on that stage is now 0 of 105 shapes, and all 6 textured meshes keep their white base. 609 USD tests and 210 viewer tests pass.

eric-heiden (approved):
Thanks for addressing my comments, LGTM!

PR #3913, opened 2026-08-14
Add SensorCamera and deprecate SensorTiledCamera

daniela-hase:
## Description

Adds `newton.sensors.SensorCamera`, a redesigned raytraced camera sensor, and the standalone renderer that backs it under `newton/_src/sensors/sensor_camera_render/`. The previous tiled renderer (`newton/_src/sensors/warp_raytrace/`) and `SensorTiledCamera` are **frozen** and **deprecated** rather than modified in place.

The entire public surface of this feature is a single class: **`newton.sensors.SensorCamera`**. Nothing else — no render context, no config module, no config types — is exposed at the top level.

### Why fork the renderer instead of evolving it?

`SensorTiledCamera` and its `warp_raytrace` backend model rendering as a 4-D `(world, camera, height, width)` problem with cameras attached to model sites. The new API is deliberately simpler — one camera per view, an explicit view→world mapping, caller-owned poses, and a fully encapsulated renderer — which touches nearly every kernel signature. To de-risk that, the renderer was duplicated and the two now evolve independently:

- **`newton/_src/sensors/sensor_camera_render/`** (new, active) — backs `SensorCamera`. 3-D `(view, height, width)` outputs, per-view world selection.
- **`newton/_src/sensors/warp_raytrace/`** (frozen) — backs the deprecated `SensorTiledCamera`. A SHA-256 manifest guard (`test_frozen_tiled_renderer.py`) fails CI on any edit, so bug fixes are consciously ported, not accidentally diverged.

### `SensorCamera` design

- **Model in, renderer hidden.** `SensorCamera(model)` builds and owns its render context internally. There is no public `RenderContext`. Scene setup is surfaced as methods on the camera: `create_default_light(...)`, `assign_checkerboard_material(...)`.
- **Caller owns the camera poses.** World-space transforms are passed to `update(state, camera_transforms, ...)` each frame; the sensor doesn't allocate or hold them. `update()`.
- **Views are decoupled from worlds.** A per-view `world_indices` array maps each view to a world; negative `SensorCamera.WorldRenderFlag` sentinels skip a view (`DISABLE_PRESERVE` leaves pixels unchanged, `DISABLE_CLEAR` writes clear values). `view_count` may differ from `world_count` (e.g. three cameras of one shared world).
- **Config types are nested.** `RenderConfig`, `ClearData`, `WorldRenderFlag`, `RenderOrder`, `TextureProjectionMode`, `LightType`, and `GaussianRenderMode` are exposed as `SensorCamera.*` nested attributes (the `SensorTiledCamera` pattern), not on the `newton` namespace.

### Breaking changes

- `SensorTiledCamera` is deprecated (emits `DeprecationWarning`); use `SensorCamera`.
- The tiled example and benchmark are replaced by `example_sensor_camera` / `bench_sensor_camera` (reusing the existing screenshot).

## Checklist

## Test plan

```bash
# New SensorCamera unit suite (44 tests) under strict warnings
uv run --extra dev -m newton.tests --strict-warnings -k TestSensorCamera

# Frozen-renderer guard + deprecation warning
uv run --extra dev -m newton.tests -k test_frozen_tiled_renderer

# Example (CUDA); smoke-tested headless on CPU
uv run --extra examples -m newton.examples sensor_camera --num-frames 4 --viewer null --device cpu
```

- `uvx pre-commit run -a` clean.
- `TestSensorCamera`: 44 passed, 1 skipped (CUDA-only device test).
- `TestFrozenTiledRenderer` passes; `TestSensorTiledCamera`: 28 passed, 5 skipped.
- Example runs to completion headless on CPU.

## New feature / API change

```python
import math
import numpy as np
import warp as wp
import newton
from newton.sensors import SensorCamera

# 1. Build a model (one or more worlds).
# ... add worlds / shapes ...
model = builder.finalize()

# 2. Construct the camera for the model. It builds its own internal renderer;
#    view_count defaults to model.world_count.
rays = SensorCamera.compute_camera_rays_pinhole(256, 256, camera_fov=math.radians(45.0))
camera = SensorCamera(model)

# 3. Configure scene lighting/materials and render settings on the camera.
#    Config types are nested attributes of SensorCamera.
camera.create_default_light(enable_shadows=True)
camera.render_config = SensorCamera.RenderConfig(enable_textures=True)

# 4. The caller owns the per-view world-space camera poses.
identity = np.tile([0, 0, 0, 0, 0, 0, 1], (camera.view_count, 1))
camera_transforms = wp.array(identity, dtype=wp.transformf, device=model.device)

color = camera.create_color_image_output()
depth = camera.create_depth_image_output()

# 5. Render. Optionally remap/disable views via world_indices, e.g.
#    [0, int(SensorCamera.WorldRenderFlag.DISABLE_CLEAR)].
state = model.state()
camera.update(state, camera_transforms, rays, color_image=color, depth_image=depth)
```

## New Features
- Added `SensorCamera` for raytraced rendering across multiple views and worlds.
- Added pinhole, USD pinhole, and fisheye camera models with configurable lighting, shadows, textures, Gaussian rendering, and checkerboard materials.
- Added color, depth, normals, albedo, shape-index, and HDR outputs with visualization utilities.
- Added flexible camera transforms, render ordering, world selection, and render configurations.

## Documentation
- Updated examples and API documentation for `SensorCamera`.

## Deprecation
- Deprecated `SensorTiledCamera`; use `SensorCamera` instead.


newton/_src/sensors/sensor_camera_render/raytrace.py
```diff
@@ -0,0 +1,769 @@
...
+        camera_forward: wp.vec3f,
+    ) -> ClosestHit:
+        if bvh_shapes_size:
+            for i in range(wp.static(2 if config.enable_global_world else 1)):
+                group_root = get_group_roots(bvh_shapes_group_roots, world_index, i)
+                if group_root < 0:
+                    continue
+
+                gaussians_hit = wp.vector(length=wp.static(state.num_gaussians), dtype=wp.uint32)
+                num_gaussians_hit = wp.int32(0)
```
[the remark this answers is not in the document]
daniela-hase: exercised by the no-gaussian sphere tests and works (Warp tolerates it); identical to the frozen renderer.

newton/_src/sensors/sensor_camera_render/render.py
```diff
@@ -0,0 +1,353 @@
...
+        is_gaussian = wp.bool(False)
+        if wp.static(state.num_gaussians > 0):
+            if closest_hit.shape_index < raytrace.MAX_SHAPE_ID:
+                if shape_types[closest_hit.shape_index] == GeoType.GAUSSIAN:
+                    is_gaussian = wp.bool(True)
+
+        albedo_color = wp.vec3f(0.0)
+
+        if not is_gaussian:
+            hit_point = ray_origin_world + ray_dir_world * closest_hit.distance
```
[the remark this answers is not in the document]
daniela-hase: `is_gaussian`/`albedo_color` are initialized before the conditionals; no bug.

newton/_src/sensors/sensor_camera.py
```diff
@@ -0,0 +1,597 @@
...
+    def utils(self, view_count: int) -> Utils:
+        """Return post-processing helpers (``to_rgba``/``flatten``/depth conversion) for ``view_count`` views.
+
+        Args:
+            view_count: Number of views the processed images carry (their leading dimension).
+
+        Returns:
+            A :class:`~newton.sensors.SensorCamera.Utils` bound to ``view_count`` and the model device.
+        """
+        return Utils(view_count=int(view_count), device=self.device)
```
mmacklin, line 153: This is a slightly unusual pattern - returning a Utils object each time the user does `camera.utils`? Why not just add these functions onto the main class? We already have helpers for creating outputs, creating rays, etc.

daniela-hase: Primarily to keep the auto-complete list a bit more organized.

All those methods to flatten the image, convert them etc. are only for quite specific use cases and most often only for debugging/visualizing the outputs. So they are more helper'y than the ones on the main class.

But I have no strong opinion on this, so I can move them out of utils onto the main class if that is preferred.

newton/_src/sensors/sensor_camera_render/types.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+class WorldRenderFlag(enum.IntEnum):
+    """Negative disable sentinels for the per-view ``world_indices`` array.
+
+    Each entry of ``world_indices`` is either a non-negative world index to
+    render for that view, or one of these negative sentinels to skip the view.
+    """
+
+    DISABLE_PRESERVE = -1
```
eric-heiden: Please keep world IDs consistent with [`newton.intersect_ray()`](https://github.com/newton-physics/newton/blob/42e49ba35f02be7702721f37dc7842fd71e98834/newton/_src/geometry/raycast.py#L936-L950): `-1` selects global-world geometry only, while a local world index also includes global geometry by default.

Assigning `-1` to `DISABLE_PRESERVE` instead makes a valid Newton world ID silently skip rendering and retain stale pixels. I reproduced this on CPU and CUDA: for the same scene and world IDs `[-1, 0, 1]`, `newton.intersect_ray()` returns depths `[3.5, 1.5, 3.5]`, whereas `SensorCamera` leaves the first output at its prefilled value of `77.0`.

Please reserve `-1` for the global world and keep view enable/disable and clear/preserve behavior separate from world IDs. The corresponding `update()` documentation and renderer handling should follow the same convention.

daniela-hase: Rendering **just** the global world is currently not supported by the renderer. But I can change those consts here to not conflict with a `-1` world if we decide to add support for rendering just the global world in the future.

eric-heiden: Thanks, your fix looks better. At least -1 is reserved for the global world once we have it supported.

PR #3916, opened 2026-08-14
Add multi-RHS Kamino FK velocity solves

### LINKED ISSUE #3915: [REQ] Add multi-RHS constraint-consistent velocity evaluation to SolverKamino
## Description

Kamino currently evaluates velocity-level forward kinematics for one right-hand side at a time. I would like to discuss a public way to evaluate several independent velocity right-hand sides at the same configuration while reusing the constraint work that is common to all of them.

Desired behavior:

- Support floating-base and fixed-base constrained mechanisms.
- Produce the same constraint-consistent body twists as repeated single-RHS evaluations.
- Support heterogeneous batched worlds and optional world masking.
- Preserve caller-owned Newton state.
- Avoid repeated Jacobian assembly and factorization where possible.
- Remain compatible with CUDA graph capture after workspace initialization.

## Motivation / Use Case

Closed-loop and otherwise constrained mechanisms cannot generally use tree-based kinematic Jacobians because those Jacobians do not account for loop-closure constraints.

Task-space control, mechanism analysis, sensitivity evaluation, and reduced-order dynamics often require responses to multiple independent velocity basis vectors at one configuration. Repeating the current single-RHS operation rebuilds and refactorizes the same constraint system.

## Questions for Maintainers

- Should this be a public `SolverKamino` operation or remain an internal batched forward-kinematics primitive?
- Would an initial dense-FK-only scope fit the Kamino roadmap, or should sparse FK be part of the first design?
- What input/output layout and workspace ownership would best fit existing Kamino APIs?
- Is there planned work that already covers this use case?

## Alternatives Considered

Calling the existing single-RHS velocity FK operation repeatedly produces the required result but repeats common assembly and factorization work.

Applications can also implement their own constraint elimination or Jacobian projection, but that duplicates Kamino conventions and relies on private implementation details.

han-xudong:
## Description

Closes #3915.

Add homogeneous multi-RHS velocity-level forward kinematics to Kamino's internal FK solver.

The implementation:

- Extends the existing batched semi-sparse blocked Cholesky solve and FK kernels with an RHS dimension instead of adding parallel implementations.
- Shares one constraint Jacobian, normal matrix, and factorization across all velocity right-hand sides.
- Routes single- and multi-RHS velocity FK through the same kernels and `solve_for_body_velocities()` entry point using array views for the single-RHS case.
- Adds `request_rhs_size()` and `request_velocity_solve_batch_size()` so temporary buffers are allocated before CUDA graph capture; batch size 1 is available by default.
- Uses one homogeneous RHS size across all worlds.
- Keeps multi-RHS velocity FK limited to the dense FK solver.

## Checklist

- [x] No changelog fragment is needed because Kamino remains in beta

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_kamino_solvers_forward_kinematics
uv run --extra dev -m newton.tests -k test_kamino_linalg_factorize_llt_blocked_semi_sparse
uvx pre-commit run -a
```

The tests compare each batched result against an independent single-RHS solve, verify linearity, cover gimbal-coordinate refresh and shared base-velocity broadcasting, and compare the LLT solutions against NumPy.

## New feature / API change

The functionality remains on the internal `ForwardKinematicsSolver`. Call `request_velocity_solve_batch_size(batch_size)` before graph capture, then use `solve_for_body_velocities()` with arrays whose leading dimension is the requested batch size.


newton/_src/solvers/kamino/_src/solvers/fk/solver.py
```diff
@@ -2075,6 +2081,133 @@ def solve_for_body_velocities(
...
+                self.actuated_coord_offsets,
+                self.actuated_dof_offsets,
+                self.constraint_full_to_red_map,
+                self.actuators_q_next,
+                fk_actuators_u,
+                world_mask,
+                target_cts_u,
+            ],
+            device=self.device,
+        )
```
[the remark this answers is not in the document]
han-xudong: Fixed in 334ef80c. The multi-RHS path now refreshes actuator coordinates from the current body poses regardless of whether target transforms are supplied. I also added a gimbal regression test that poisons the cached coordinates and verifies the multi-RHS result against the single-RHS solve.

han-xudong:
Addressed the review feedback in 334ef80c:

- Refresh actuator coordinates for explicit target transforms, with a gimbal regression test.
- Reject multi-RHS linear systems whose row dimension is smaller than the configured matrix size.
- Document the RHS axis layout on the new FK kernels.

I left the Jacobian-transpose tile optimization out of this correctness update. Supporting it cleanly requires extending the tile-kernel factory with a three-dimensional RHS layout rather than reusing the current 2D kernel directly. The current path is dense-FK-only and already shares the Jacobian assembly and factorization; I would prefer to benchmark the matrix-product contribution before expanding this PR further.

Guirec-Maloisel:
Hi @han-xudong, thanks for your interest in Kamino, and great that you're willing to contribute to the solver!
However, for your contributions to get in, please follow the contribution guidelines: https://github.com/newton-physics/newton/blob/main/CONTRIBUTING.md
In particular, to introduce new features, you should first open an issue to propose and motivate a new feature, and wait for feedback - we might already have something planned ourselves, that covers or conflicts with that feature. For this PR and the other ones that you opened recently, it looks like you created an issue and a PR simultaneously and have the issue mostly describe what your PR brings in.
I'm converting your PRs to drafts now, but will leave the corresponding issues open. Please clean up the issues if needed and add the kamino label for visibility; we'll then have a more detailed look this week and give feedback on how to move on with the implementation, at which point we could re-open some of the PRs.

han-xudong:
@Guirec-Maloisel Thanks for your feedback! I revised #3915 to focus on the problem, motivation, desired behavior, alternatives, and open design questions for the Kamino team. I also tried to add the kamino label, but GitHub reports that I do not have AddLabelsToLabelable permission in the upstream repository. Could a maintainer please add that label? I will keep this PR in draft and pause implementation work until the issue receives project-member feedback.

Guirec-Maloisel (changes requested):
I've had a look at the code changes and made up my mind on the API question. Here are some high-level comments:

- There is a lot of code duplication in llt_block_semi_sparse.py and fk/kernels.py. Please correct me if I missed something, but I think it would be possible to simply update existing kernels to work with one more thread id and one more input/output dimension (the kernels in the LLT solver even already are taking 3d inputs, just reading/writing at the 0 index). Then update the call sites with the right thread dimensions and view on arrays (e.g. with reshape or by passing ptr= to wp.array()).
- I think it would be cleaner to not expose this functionality in solver_kamino.py for now, as this deviates too much from the main purpose of the solver. I'd be open to adding (sparingly) convenience members to make it easier to call the functionality on the FK solver through SolverKamino (e.g. property for the FK solver, convenience for the body pose conversion at the newton<->kamino interface etc).
- The new functionality seems limited to the homogeneous case when it comes to RHS dimension. Since this is a new functionality I'm fine with this (as there is nothing that fundamentally prevents heterogenous rhs dimensions across worlds, it's just the API the becomes a bit more complicated), but this should be documented clearly at all levels (so LLT solver and FK solver)

I have more detailed comments that I will write at the appropriate location

newton/_src/solvers/kamino/_src/linalg/factorize/llt_blocked_semi_sparse.py
```diff
@@ -109,6 +109,21 @@ def reorder_rows_kernel_col_vector(
         dst[batch_id, i, 0] = src[batch_id, src_row, 0]
 
 
+@wp.kernel
+def reorder_rows_kernel_col_vector_multi_rhs(
```
Guirec-Maloisel: Assuming this new kernel disappears (and the old kernel is adapted instead), it would make sense to change the naming of these kernels and add docstrings for clarity. More specifically:
reorder_rows_kernel -> reorder_rows_cols_kernel (+docstring: applies a permutation to a matrix along rows and columns)
reorder_rows_kernel_col_vector -> reorder_rows_kernel (+docstring: applies a permutation to matrix/vector along rows)

newton/_src/solvers/kamino/_src/linalg/factorize/llt_blocked_semi_sparse.py
Guirec-Maloisel: This assignment in the old kernel is not useful, and was pruned in the new kernel. When instead adapting the old kernel to support multi-rhs, it would make sense to indeed remove it (but not all the comments that got pruned in the duplicate kernel)

newton/_src/solvers/kamino/_src/linalg/factorize/llt_blocked_semi_sparse.py
```diff
@@ -589,7 +675,6 @@ def solve(
 
             rhs = self.rhs_swizzled
 
-        # Then solve the system using blocked_cholesky_solve kernel
```
Guirec-Maloisel, line 592: This comment and at least another below are not related to the new functionality, and should not be removed

newton/_src/solvers/kamino/_src/linalg/factorize/llt_blocked_semi_sparse.py
```diff
@@ -599,10 +684,83 @@ def solve(
...
         if self.enable_reordering:
-            # Undo reordering
             wp.launch(
                 reorder_rows_kernel_col_vector,
                 dim=[self.num_batches, self.max_num_equations],
                 inputs=[R, result, self.inverse_ordering, self.num_active_equations, batch_mask],
                 device=self.device,
             )
+
+    def solve_multi_rhs(
```
Guirec-Maloisel: Similar here, the argument types are the same as solve(), so the existing function should be adapted instead.

newton/_src/solvers/kamino/_src/linalg/factorize/llt_blocked_semi_sparse.py
```diff
@@ -433,6 +518,7 @@ def __init__(self, num_batches: int, max_num_equations: int, block_size=16, devi
         self.rhs_swizzled = wp.zeros(
             shape=(num_batches, self.padded_num_equations, 1), dtype=float, device=self.device
         )  # temp memory
+        self._multi_rhs_workspaces: dict[int, tuple[wp.array, wp.array, wp.array]] = {}
```
Guirec-Maloisel: I'd not separate the single-rhs case anymore, and remove the existing y, result_swizzled, rhs_swizzled.
I'd name the new dictionary `temp_buffers` and have a docstring listing the 3 arrays that it stores per rhs size.
The type annotation for wp.array should also have square brackets with the type

newton/_src/solvers/kamino/_src/linalg/factorize/llt_blocked_semi_sparse.py
```diff
@@ -599,10 +684,83 @@ def solve(
...
+        """
+        if rhs.shape != result.shape:
+            raise ValueError("rhs and result must have identical shapes")
+        if rhs.ndim != 3 or rhs.shape[0] != self.num_batches:
+            raise ValueError("rhs must have shape (batch_count, n, rhs_count)")
+        if rhs.shape[1] < self.max_num_equations:
+            raise ValueError(f"rhs must have at least {self.max_num_equations} rows")
+
+        rhs_count = rhs.shape[2]
+        workspace = self._multi_rhs_workspaces.get(rhs_count)
```
Guirec-Maloisel: To avoid overhead and complications that come with allocations within a CUDA graph (especially if the allocated data survives after the end of the capture like here) we should allocate the temporary arrays up-front before calling solve(), and raise an exception if solve() is called for a rhs size that is not pre-allocated.

This could be done with constructor arguments, and/or helpers like request_rhs_size() to be called before graph capture.

(unless there is a real need for rhs sizes varying over time and not known upfront)

newton/_src/solvers/kamino/_src/linalg/factorize/llt_blocked_semi_sparse.py
```diff
@@ -599,10 +684,83 @@ def solve(
...
+    def solve_multi_rhs(
+        self,
+        rhs: wp.array3d[float],
+        result: wp.array3d[float],
+        batch_mask: wp.array[wp.bool],
+    ):
+        """Solve multiple right-hand sides using the most recent factorization.
+
+        Args:
+            rhs: Right-hand sides with shape ``(batch_count, n, rhs_count)``.
```
Guirec-Maloisel: Minor point, but in other docstrings in this file we've used batch_size, so please update to batch_size and rhs_size here as well for consistency

newton/_src/solvers/kamino/_src/linalg/factorize/llt_blocked_semi_sparse.py
Guirec-Maloisel: The same applies then to the FK solver then, with e.g. a request_velocity_rhs_size() or similar that would call the one from the LLT solver. The size 1 case would be allocated by default by the FK solver, and with the LLT solver with default arguments.

newton/tests/kamino/test_kamino_solver_kamino_linearization.py
```diff
@@ -0,0 +1,68 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
+
+"""Test the public Kamino body-velocity linearization API."""
```
Guirec-Maloisel: If we don't expose this functionality in SolverKamino, this file should go, but the unit test itself could be kept as part of the FK unit tests below (which are also on a four-bar model already).

newton/tests/kamino/test_kamino_solvers_forward_kinematics.py
```diff
@@ -959,6 +959,119 @@ def test_solve_fk_warns_without_base_body_when_base_provided(self):
...
+        actuator_u = wp.array(actuator_u_np, dtype=wp.float32, device=self.default_device)
+        base_u = wp.array(base_u_np[:, None, :], dtype=wp.spatial_vectorf, device=self.default_device)
+        bodies_u = wp.zeros(
+            (rhs_count, model.size.sum_of_num_bodies), dtype=wp.spatial_vectorf, device=self.default_device
+        )
+        solver.solve_for_body_velocities_multi_rhs(actuator_u, bodies_q, bodies_u, base_u=base_u)
+
+        np.testing.assert_allclose(bodies_u.numpy(), np.asarray(expected), rtol=2.0e-4, atol=2.0e-4)
+
+    def test_multi_rhs_refreshes_gimbal_coords_with_explicit_transforms(self):
```
Guirec-Maloisel, line 1064: I don't fully understand what this unit test is about (or if it is even really about the multi-rhs solve, since there seems to be a single rhs). Could you clarify/add more comments?

changelog/3915.added.md
```diff
@@ -0,0 +1 @@
+Add `SolverKamino.eval_body_velocity_linearization()` for evaluating multiple constraint-consistent velocity responses with a shared factorization.
```
Guirec-Maloisel: We don't need a changelog entry for this (especially if we remove the SolverKamino part), as Kamino is still in beta (we only add entries for major changes).

newton/_src/solvers/kamino/_src/linalg/factorize/llt_blocked_semi_sparse.py
han-xudong: Thank you for pointing this out. I restored the unrelated solve and reordering comments while adapting the original implementation for multiple right-hand sides.

newton/_src/solvers/kamino/_src/linalg/factorize/llt_blocked_semi_sparse.py
han-xudong: Updated as suggested. The single-RHS special buffers were removed; `temp_buffers` now stores the reordered RHS, reordered result, and forward-substitution workspace for each requested RHS size, using bracket-style Warp array annotations.

newton/_src/solvers/kamino/_src/linalg/factorize/llt_blocked_semi_sparse.py
han-xudong: Updated the new documentation and tests to use `batch_size` and `rhs_size` consistently.

newton/_src/solvers/kamino/_src/linalg/factorize/llt_blocked_semi_sparse.py
han-xudong: Thank you. I removed the duplicate kernel, renamed the existing kernels to `reorder_rows_cols_kernel` and `reorder_rows_kernel`, and added docstrings describing the permutations.

newton/_src/solvers/kamino/_src/linalg/factorize/llt_blocked_semi_sparse.py
han-xudong: Agreed. I removed `solve_multi_rhs()` and adapted the existing `solve()` method to accept a homogeneous `rhs_size` in its existing 3D inputs.

newton/_src/solvers/kamino/_src/linalg/factorize/llt_blocked_semi_sparse.py
han-xudong: Thank you. The existing solve kernel now handles the RHS dimension directly, the unused `i_end` assignment was removed, and the surrounding forward/backward substitution comments were retained.

newton/tests/kamino/test_kamino_solver_kamino_linearization.py
han-xudong: Agreed. I removed the `SolverKamino` API and its dedicated test file, then retained the linearity coverage in the existing FK test module.

newton/_src/solvers/kamino/_src/linalg/factorize/llt_blocked_semi_sparse.py
han-xudong: Implemented for both layers. `request_rhs_size()` preallocates LLT buffers, and `request_velocity_rhs_size()` preallocates FK buffers and forwards the request to LLT. RHS size 1 is available by default, and solves now raise if another size was not requested before graph capture.

changelog/3915.added.md
han-xudong: Removed as suggested. The final diff no longer contains a changelog fragment.

newton/tests/kamino/test_kamino_solvers_forward_kinematics.py
han-xudong: I clarified and strengthened this test: it now uses two genuinely different gimbal velocity right-hand sides, deliberately poisons the cached actuator coordinates, and compares both batched results against independent single-RHS solves. This verifies that the shared path refreshes the current gimbal coordinates for every RHS.

newton/_src/solvers/kamino/_src/linalg/factorize/llt_blocked_semi_sparse.py
```diff
@@ -566,43 +575,73 @@ def solve(
...
+
+        Raises:
+            ValueError: If shapes are incompatible or ``rhs_size`` was not requested.
         """
+        if rhs.shape != result.shape:
+            raise ValueError("rhs and result must have identical shapes")
+        if rhs.ndim != 3 or rhs.shape[0] != self.num_batches:
+            raise ValueError("rhs must have shape (batch_size, n, rhs_size)")
+        if rhs.shape[1] < self.max_num_equations:
+            raise ValueError(f"rhs must have at least {self.max_num_equations} rows")
```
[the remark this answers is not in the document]
Guirec-Maloisel: Is this really problematic? I thought that tile_load, tile_store operations were able to deal with this (considering out-of-bounds values as zero for read, and no-op for write). Could you double-check?

Guirec-Maloisel (changes requested):
@han-xudong I took a cleaning pass through the PR as it felt easier than commenting. I restored some of the removed comments, added a bit more at other places, and merged back the single/multiple-rhs functions in the FK solver.

The only thing missing from my side would be more comments on the gimbal unit test, explaining more what is done on a high level and what you are trying to achieve by "poisoning" actuators_q_next. Once you address this I'll trigger the CI so we can merge.

newton/_src/solvers/kamino/_src/solvers/fk/kernels.py
```diff
@@ -2271,47 +2289,51 @@ def _eval_body_velocities(
...
         num_bodies: Number of bodies per world
         first_body_id: First body id per world
         bodies_q: Body poses
-        bodies_q_dot: Time derivative of body poses
+        bodies_q_dot: Time derivative of body poses, with shape ``(num_worlds, num_states_max, batch_size)``
         world_mask: Per-world boolean flag to perform the computation (False = skip)
     Outputs:
-        bodies_u: Body velocities (twists)
+        bodies_u: Body velocities (twists), with shape ``(batch_size, num_bodies_max)``
     """
```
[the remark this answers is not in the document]
Guirec-Maloisel: Good point, I fixed the docstring

newton/_src/solvers/kamino/_src/solvers/fk/solver.py
```diff
@@ -1826,40 +1836,53 @@ def _run_newton_iteration(self, bodies_q: wp.array[wp.transformf]):
...
+                    shape=(batch_size, 6 * self.num_worlds),
+                    device=self.device,
+                    copy=False,
                 ),
                 actuators_u,
                 self.actuated_dofs_map,
-                self.actuators_u,
+                fk_actuators_u,
             ],
             device=self.device,
```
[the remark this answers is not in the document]
Guirec-Maloisel: Right, good catch. I replaced batch_size with base_u.shape[0] to adapt to either case.

han-xudong:
@Guirec-Maloisel, thank you for taking the cleaning pass through the PR. I added high-level comments to the gimbal test explaining the independent single-RHS reference and why `actuators_q_next` is deliberately poisoned before the batched solve. The test now makes explicit that the solver must refresh the gimbal coordinates from the current body pose rather than reuse stale scratch state.

While verifying the updated path, I also aligned the validation and annotation with the single-row `base_u` broadcast already supported by the kernel, and covered that case in the same test. The branch is now up to date with `upstream/main`; all 17 FK tests, both semi-sparse LLT tests, and the full pre-commit suite pass. It should be ready for CI.

Guirec-Maloisel (approved):
Looks good to me, good catch for the transformf vs spatial_vectorf!

I had added the support for size 1 along the batch for base_u to be able to handle the case when base_u is not provided and falls back to the default (which is not sized based on the number of rhs). But I guess it doesn't hurt to support this more generally since it's implemented.

PR #3918, opened 2026-08-14
Improve convex collision kernel performance

nvtw:
## Description

Improve collision kernel performance by reducing broad- and narrow-phase launch and synchronization overhead and optimizing the SAP, MPR, GJK, multicontact, primitive-contact, and SDF-contact paths. Large CUDA workloads use sparse candidate routing, while CPU execution retains a safe atomic fallback.

The sparse CUDA path changes only the layout of the internal `gjk_candidate_pairs` work buffer; contact outputs remain unchanged. `CollisionPipeline` selects the optimized routing, kernel, and buffer configuration automatically.

For direct `NarrowPhase` construction, this adds optional expert controls for mesh work-buffer bounds, generic-convex stage availability, sparse GJK routing, candidate-work estimates, and SDF texture layout. These keyword arguments are additive and retain conservative defaults, so no migration is required.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_compacted_append_matches_cpu_and_cuda
uv run --extra dev -m newton.tests -k test_sap_adaptive_chunk_enumerates_dense_pairs
uv run --extra dev -m newton.tests -k test_sparse_gjk_routing
uv run --extra dev -m newton.tests -k test_cylinder_scale_update_keeps_generic_stage
uvx pre-commit run -a
```

## Summary by CodeRabbit

* **New Features**
  * Added CUDA-accelerated convex collision processing for large and replicated scenes.
  * Expanded support for generic convex shapes, mesh scaling, SDFs, and replicated worlds.
* **Performance**
  * Improved collision workload scheduling, adaptive processing, and memory sizing.
  * Reduced unnecessary processing for separated or non-colliding shape pairs.
* **Bug Fixes**
  * Improved contact accuracy, distance calculations, back-face handling, and mesh-scale support.
  * Improved contact matching, sorting, and reporting consistency.
* **Compatibility**
  * Existing collision paths and results remain supported without migration requirements.


mzamoramora-nvidia (commented):
Thanks for splitting this work out. The benchmark results look strong, especially the contact-pyramid improvement. I found one correctness issue around changing a cylinder's scale after the collision pipeline has been constructed, plus a few smaller API, documentation, and test-coverage points in the inline comments.

One small PR-description fix: the test plan names `test_warp_aggregated_append_matches_cpu_and_cuda`, but the test is now called `test_compacted_append_matches_cpu_and_cuda`.

newton/_src/sim/collide.py
```diff
@@ -52,6 +52,117 @@ def _shape_collide_mask(model: Model, shape_count: int | None = None) -> np.ndar
...
+        shape_scale_array = getattr(model, "shape_scale", None)
+        shape_scales = shape_scale_array.numpy() if shape_scale_array is not None else None
+        requirements = []
+        for shape_a, shape_b in pairs.reshape(-1, 2):
+            type_a = int(shape_types[shape_a])
+            type_b = int(shape_types[shape_b])
+            cylinder_barrel_radius = None
+            if shape_scales is not None:
+                if type_a == int(GeoType.CYLINDER):
+                    cylinder_barrel_radius = float(shape_scales[shape_a, 2])
```
mzamoramora-nvidia: Could we make this check conservative when routing depends on `shape_scale`? `ModelFlags.SHAPE_PROPERTIES` explicitly includes runtime `shape_scale` updates, but the current barrel radius is used to decide permanently whether this pipeline needs a GJK stage.

I tested a horizontal plane-cylinder pair on CPU and CUDA. I constructed the pipeline for a straight cylinder, then changed its barrel radius in `model.shape_scale` from `0.0` to `0.6`. Reusing the pipeline produced zero contacts, while rebuilding it produced four. The primitive stage routes the changed cylinder to GJK, but the cached `has_generic_convex_pairs=False` means the GJK stage is not launched.

Could we keep GJK available for pairs whose routing depends on mutable shape properties, or add a supported way to invalidate and re-specialize the pipeline? A regression should change the scale and reuse the same pipeline.

newton/_src/geometry/narrow_phase.py
```diff
@@ -1604,8 +2219,16 @@ def __init__(
         has_meshes: bool = True,
         has_heightfields: bool = False,
         use_lean_gjk_mpr: bool = False,
+        has_generic_convex_pairs: bool = True,
```
mzamoramora-nvidia, line 2144: These scheduling options are being added to the public `newton.geometry.NarrowPhase` constructor, while the PR description says the documented `NarrowPhase` interfaces remain unchanged. Since the arguments have defaults, this is not a breaking change, but could we either keep these details internal or update the PR description to mention the new expert controls?

It would also be helpful to reject inconsistent combinations such as `all_pairs_generic_convex=True` with `has_generic_convex_pairs=False`.

newton/_src/geometry/collision_core.py
```diff
@@ -974,40 +970,128 @@ def mesh_vs_convex_midphase(
...
+    v1 = mesh.points[idx1]
+    v2 = mesh.points[idx2]
+    face_normal = wp.cross(v1 - v0, v2 - v0)
+    center_dist = wp.dot(face_normal, center_in_bvh - v0)
+    return not (center_dist < 0.0)
+
+
+@wp.func
+def mesh_vs_convex_midphase(
+    idx_in_thread_block: int,
```
mzamoramora-nvidia, line 1018: Could we document `idx_in_thread_block` in the `Args:` section? It is part of the updated signature but is the only argument missing there. Something like "Lane index within the mesh-query thread block" would make the tiled sharing behavior clear.

newton/tests/test_narrow_phase.py
```diff
@@ -2026,6 +2079,80 @@ def test_gjk_buffer_overflow(self):
         count = contact_count.numpy()[0]
         self.assertGreater(count, 0, "Should still produce contacts for pairs that fit in the buffer")
 
+    @unittest.skipUnless(_cuda_available, "Sparse GJK routing is enabled only on CUDA")
+    def test_sparse_gjk_routing_preserves_filtered_slots(self):
```
mzamoramora-nvidia, line 2084: This covers the sparse buffer layout well. Could we also run the same scene through compact and sparse routing and compare the resulting contact count, pairs, positions, normals, and penetration depths? That would directly protect the PR's promise that sparse routing changes the work-buffer layout without changing contact results.

A small separate assertion around the one-million-pair threshold would also cover automatic sparse-mode selection.

newton/_src/sim/collide.py
```diff
@@ -52,6 +52,84 @@ def _shape_collide_mask(model: Model, shape_count: int | None = None) -> np.ndar
...
+        (int(GeoType.PLANE), int(GeoType.BOX)),
+        (int(GeoType.SPHERE), int(GeoType.SPHERE)),
+        (int(GeoType.SPHERE), int(GeoType.CAPSULE)),
+        (int(GeoType.SPHERE), int(GeoType.BOX)),
+        (int(GeoType.CAPSULE), int(GeoType.CAPSULE)),
+    }
+)
+
+
+def _pair_requires_generic_convex_narrow_phase(
```
camevor, line 69: Likely bug: this function newly rejects pairs of finite planes. The documentation states (`collisions.rst`):

> The plane--plane upper bound is five when at least one plane is finite. Two infinite planes produce no contacts.

newton/_src/sim/collide.py
```diff
@@ -1278,6 +1438,14 @@ def __init__(
                             bool(scale_baked[shape_sdf_index[shape_idx]]) or identity_shape_scale[shape_idx]
                             for shape_idx in np.flatnonzero(mesh_sdf_shapes)
                         )
+                max_mesh_mesh_pairs = min(
```
camevor: Likely bug: `_compute_per_world_mask_pair_max()` is a valid bound for NXN/SAP, but not for `BroadPhaseExplicit`. This can lead to overflow, contact loss, or a crash.

Repro: two overlapping mesh shapes in different worlds with explicit pair [[0,1]] provisions `max_mesh_mesh_pairs=0`. On this branch, returns no contacts are returned with reduction disabled and a segfault is encountered on CPU with default reduction, while the implementation on `main` provisions capacity 1 and returns contacts.

newton/_src/sim/collide.py
```diff
@@ -52,6 +52,84 @@ def _shape_collide_mask(model: Model, shape_count: int | None = None) -> np.ndar
...
+    """Collect generic-convex requirements for possible shape-type pairs."""
+    shape_types_array = getattr(model, "shape_type", None)
+    if shape_types_array is None:
+        return None
+
+    shape_types = shape_types_array.numpy()
+    if broad_phase_mode == "explicit":
+        if shape_pairs_filtered is None:
+            return None
+        pairs = shape_pairs_filtered.numpy()
```
camevor, line 97: The documentation does not specify whether the shape pairs may be modified at runtime. Main consumes the live contents of a fixed-size explicit-pair buffer on every collision pass, and same-size content updates work, although resizing does not.

This implementation does not allow updating the explicit shape pairs, so this should be documented.

camevor (commented):
The comments by @mzamoramora-nvidia were all addressed. Overall, the changes look solid. Added three comments for correctness and documentation issues found.

mzamoramora-nvidia (approved):

PR #3919, opened 2026-08-14
Fix Torch neural controller target indexing and reset

### LINKED ISSUE #3923: [BUG] Torch neural controllers ignore target_pos_indices and crash on masked reset
### Bug Description

 Two defects in the PyTorch checkpoint path of ControllerNeuralMLP and ControllerNeuralLSTM
 1. target_pos_indices is ignored
 2. Masked reset raises on inference tensors. ControllerNeuralLSTM stores network outputs produced under torch.inference_mode() as persistent hidden/cell state. State.reset(mask) then writes to them in place, giving RuntimeError: Inplace update to inference tensor outside InferenceMode is not allowed. The unmasked path is unaffected, because it uses new_zeros.

### Reproduction Script

```python

```

### System Information

_No response_

jvonmuralt:
ControllerNeuralMLP and ControllerNeuralLSTM chose between position and sequential indices using Python object identity, so the supplied target_pos_indices was never read. Whenever the target and position layouts differ -- DOF-layout targets on a floating-base robot, or an explicit target_pos_indices argument -- the controllers read the wrong target slots and computed wrong position errors. Index with the supplied array directly, as the Warp kernels and ControllerPD/PID already do.

ControllerNeuralLSTM stored network outputs produced under torch.inference_mode() as persistent hidden and cell state. A masked reset then wrote to them in place outside that mode, raising RuntimeError. Build new tensors instead, matching the unmasked branch.

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed neural controller target-position selection when target and position layouts differ.
  - Improved compatibility with Torch checkpoint outputs by safely resetting masked recurrent states.
  - Corrected Torch-based indexing for more reliable behavior across supported neural controllers.
  - Improved handling of position and velocity data, reducing the risk of incorrect actuator responses during inference.


jcarius-nv (commented):
Reviewed at `7a4f9fdc`. The implementation is small, coherent, and looks good in general: it uses the supplied target-position indices in both Torch neural controllers and avoids in-place masked writes to LSTM inference tensors. The current CI, GPU tests, and benchmarks are green.

Before merging—especially if this is a candidate for the 1.5 patch release—please add focused regression tests that exercise the previously failing paths:

- `ControllerNeuralMLP` and `ControllerNeuralLSTM` with `target_pos_indices` different from `pos_indices`, verifying that the target is read from the supplied target slot. Existing Torch tests pass identical index arrays and therefore do not catch this regression.
- `ControllerNeuralLSTM.State.reset(mask)` after hidden/cell outputs have been produced under `torch.inference_mode()`, verifying that the selected entries are zeroed, unselected entries are preserved, and no `RuntimeError` is raised.

These tests should fail without the fix and pass at this head. Please also add `Fixes #3923` to the PR description so the bug and its closure are linked explicitly.

Two non-blocking judgment calls:

- Mysterious name: `newton/_src/actuators/controllers/controller_neural_lstm.py:111` uses `t` for a broadcast Torch mask; a name such as `broadcast_mask` or `torch_mask` would make its role clearer.
- Duplicated code: the LSTM path around lines 395–405 and MLP path around lines 408–418 duplicate index conversion and input gathering, leaving two paths vulnerable to future drift. Consider sharing that preparation where it can remain simple; this need not expand the patch-release fix into a broader refactor.

jvonmuralt:
@jcarius-nv I added the regression tests and linked the issue, the other problems I will address in the next release.

jcarius-nv:
> @jcarius-nv I added the regression tests and linked the issue, the other problems I will address in the next release.

Sounds good, thanks! Ready from my side, but let's hear from @AntoineRichard before merging.

AntoineRichard (approved):
I tried the changes, and I they seem to work better!

jcarius-nv (approved):

PR #3924, opened 2026-08-14
Import uniform mesh normals instead of dropping them

andrewkaufman:
## Description

Authored `uniform` normals were discarded on import, with a warning, and the mesh rendered with none.

Normals have a single interpolation branch, `faceVarying`, and the indexed expansion lives inside it. A flat-shaded mesh authoring one normal per face — indexed so each distinct direction is stored once — therefore never had its indices applied. The raw value count was compared against the vertex count and the corner count, matched neither, and the normals were dropped:

```
UserWarning: Normals length (6) does not match vertices after UV splitting
for mesh /World/materialSampleGrid/meshWithGeomsubsets; dropping normals.
```

For that mesh: 12 points, 10 faces, 40 corners, and `primvars:normals` holding 6 values with `interpolation = uniform` and 10 indices. Entirely valid, and other renderers read it without complaint.

The fix resolves the indices, then hands each face's normal to its own corners, which is the `faceVarying` form the rest of the function already handles.

Worth knowing for review scope: authored normals are only loaded for meshes carrying material subsets (`load_normals=True` is passed in one place, the subset path), so that is the only path this reaches.

## Checklist

## Test plan

```
uv run --extra dev --extra examples -m newton.tests -k usd --strict-warnings   # 607 tests, OK
```

`test_uniform_indexed_normals_are_expanded` builds two quads meeting at a right angle with an indexed `uniform` primvar holding both face directions, and subsets so the normals are loaded at all. It asserts no normals warning, one normal per vertex, unit length, and that both authored directions survive — dropping the indices would collapse them to one. Without the fix it fails at `2 != 6`.

Checked against the real asset as well: the mesh splits into `front` and `screen`, flat with a single direction each, and `body` with five distinct face directions.

## Bug fix

**Steps to reproduce:**

1. Author a mesh with `primvars:normals` at `uniform` interpolation, indexed, plus a material subset so the normals are loaded.
2. `add_usd` the stage.
3. Observe the "does not match vertices" warning and `shape_source[...].normals` of `None`.

**Minimal reproduction:**

```python
import newton
from pxr import Sdf, Usd, UsdGeom, Vt

mesh = UsdGeom.Mesh.Define(stage, "/Mesh")
mesh.CreatePointsAttr().Set([(0, 0, 0), (1, 0, 0), (1, 1, 0), (0, 1, 0), (1, 0, 1), (0, 1, 1)])
mesh.CreateFaceVertexCountsAttr().Set([4, 4])
mesh.CreateFaceVertexIndicesAttr().Set([0, 1, 2, 3, 1, 4, 5, 2])
normals = UsdGeom.PrimvarsAPI(mesh).CreatePrimvar(
    "normals", Sdf.ValueTypeNames.Normal3fArray, UsdGeom.Tokens.uniform
)
normals.Set(Vt.Vec3fArray([(0, 0, 1), (1, 0, 0)]))
normals.SetIndices(Vt.IntArray([0, 1]))
for name, face in (("a", 0), ("b", 1)):
    subset = UsdGeom.Subset.Define(stage, f"/Mesh/{name}")
    subset.CreateElementTypeAttr().Set(UsdGeom.Tokens.face)
    subset.CreateIndicesAttr().Set(Vt.IntArray([face]))
    subset.CreateFamilyNameAttr().Set("materialBind")

result = builder.add_usd(stage, load_visual_shapes=True)
print(builder.shape_source[result["path_shape_map"]["/Mesh/a"]].normals)
# before: None, with a warning
# after:  four unit normals of (0, 0, 1)
```

## Summary by CodeRabbit

* **Bug Fixes**
  * USD imports now preserve authored uniform mesh normals instead of discarding them.
  * Indexed face normals are correctly expanded across mesh corners, including meshes with material subsets.
  * Invalid normal counts now produce a clear validation error.

* **Tests**
  * Added regression coverage for importing uniform mesh normals without warnings.

### DIFF
--- a/newton/tests/test_import_usd.py
+++ b/newton/tests/test_import_usd.py
@@ -12646,6 +12646,57 @@ def Xform "Body" (
             self.assertAlmostEqual(channel, want, places=5)
 
     @unittest.skipUnless(USD_AVAILABLE, "Requires usd-core")
+    def test_uniform_indexed_normals_are_expanded(self):
+        """Verify indexed uniform normals survive import rather than being dropped.
+
+        A flat-shaded mesh commonly authors one normal per face, indexed so each distinct
+        direction is stored once. Only faceVarying was handled, so the raw value count was
+        compared against the vertex and corner counts, matched neither, and the normals
+        were discarded with a warning.
+        """
+        from pxr import Sdf, Usd, UsdGeom, Vt
+
+        stage = Usd.Stage.CreateInMemory()
+        UsdGeom.SetStageUpAxis(stage, UsdGeom.Tokens.z)
+        mesh = UsdGeom.Mesh.Define(stage, "/Mesh")
+        # Two quads meeting at a right angle, so the two face normals differ.
+        mesh.CreatePointsAttr().Set([(0, 0, 0), (1, 0, 0), (1, 1, 0), (0, 1, 0), (1, 0, 1), (0, 1, 1)])
+        mesh.CreateFaceVertexCountsAttr().Set([4, 4])
+        mesh.CreateFaceVertexIndicesAttr().Set([0, 1, 2, 3, 1, 4, 5, 2])
+        normals = UsdGeom.PrimvarsAPI(mesh).CreatePrimvar(
+            "normals", Sdf.ValueTypeNames.Normal3fArray, UsdGeom.Tokens.uniform
+        )
+        normals.Set(Vt.Vec3fArray([(0, 0, 1), (1, 0, 0)]))
+        normals.SetIndices(Vt.IntArray([0, 1]))
+        # Authored normals are only loaded for meshes carrying material subsets, so the
+        # subsets are what makes this reachable at all.
+        for name, face in (("a", 0), ("b", 1)):
+            subset = UsdGeom.Subset.Define(stage, f"/Mesh/{name}")
+            subset.CreateElementTypeAttr().Set(UsdGeom.Tokens.face)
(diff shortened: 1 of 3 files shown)


newton/tests/test_import_usd.py
```diff
@@ -12442,6 +12442,57 @@ def test_hide_collision_shapes_suppresses_approximated_visual_copy(self):
         self.assertEqual(drawn, [path_shape_map["/Body/VisualSphere"]])
 
     @unittest.skipUnless(USD_AVAILABLE, "Requires usd-core")
+    @unittest.skipUnless(USD_AVAILABLE, "Requires usd-core")
```
eric-heiden, line 12648: There is a problem here with this decorator that got cut from the test function below and now appears twice in front of the new test method.

eric-heiden (commented):
LGTM! Just a small issue with the test method and then need to resolve the merge conflict.

newton/tests/test_import_usd.py
andrewkaufman: Fixed — the stray decorator is removed and `test_hide_collision_shapes_fallback_with_material` has its own decorator back, in [f3007ae](https://github.com/newton-physics/newton/commit/f3007ae25bde5b925bff767c2766b2a322c18190).

andrewkaufman (commented):
Both points addressed in the force-push: the duplicated decorator is gone, and the branch is rebased onto `main` so the conflict is resolved.

eric-heiden (approved):

PR #1966, opened 2026-03-06
Add scheduled CI for nightly mujoco-warp from source

adenzler-nvidia:
## Description

Add a daily GitHub Actions workflow that tests Newton against the latest
mujoco-warp built from source (`google-deepmind/mujoco_warp` HEAD).

This complements the existing Warp nightly workflow. The workflow:
- Runs daily at 12 PM UTC (staggered 1 hour after the Warp nightly)
- Spins up a GPU EC2 instance using the same infrastructure as existing workflows
- Installs Newton via `uv sync`, then overrides mujoco-warp with a git install
- Uses `uv run --no-sync` to preserve the source-built mujoco-warp during tests
- Supports `workflow_dispatch` for manual triggers

## Checklist

- [x] New or existing tests cover these changes
- [x] The documentation is up to date with these changes
- [ ] `CHANGELOG.md` has been updated (if user-facing change)

## Test plan

Trigger via `workflow_dispatch` after merge to verify the workflow runs correctly.

PR #3925, opened 2026-08-14
Remove dormant MuJoCo Warp CI workflow

shi-eric:
## Description

Nothing triggers `.github/workflows/mujoco_warp_tests.yml`.

It was added in #1966 with its own nightly cron. #2236 moved the scheduled workflows behind `scheduled_nightly.yml` and converted each one into a `workflow_call` target. `aws_gpu_tests`, `minimum_deps_tests` and `warp_nightly_tests` got dispatch entries. This one didn't, and its header comment has said so ever since. The API reports 0 runs against a retained history going back to 2026-05-06, and it has still collected edits from nine repo-wide CI sweeps in that time: SHA pinning, harden-runner, uv bumps, HOME scoping.

This is the only place we install mujoco-warp from git HEAD, but nobody has asked for it to run in all that time, which suggests it wasn't providing much value in the first place. It's in the git history, so hooking it up properly is easy if we decide the extra spend is worth it.

## Checklist

- [ ] For user-facing changes, a fragment has been added by following the [changelog fragment instructions](https://github.com/newton-physics/newton/blob/main/changelog/README.md)

None apply. There's no code to test, nothing in the docs refers to this workflow, and a CI-only change has no user-facing impact, so no changelog fragment.

## Test plan

`git grep mujoco_warp_tests` has no hits after the removal. `uv run --no-project -m unittest discover -s scripts/ci/tests` passes 25 tests and `uvx pre-commit run -a` is clean. The Codecov flag `mujoco-warp-nightly` and the `test-results-mujoco-warp-nightly` artifact were only referenced from inside the deleted file.

## Summary by CodeRabbit

* **Chores**
  * Removed the MuJoCo Warp test workflow from automated CI.
  * MuJoCo Warp tests are no longer included in workflow contract checks or direct-dispatch handling.


jcarius-nv:
I think we need @adenzler-nvidia's input, we had a discussion about it some time ago that running it regularly is not very helpful because it breaks frequently and doesn't provide useful signal.
The comment says "Not currently dispatched by scheduled_nightly.yml; kept available for manual dispatch/reuse.", maybe @eric-heiden knows how this was used.

jcarius-nv (dismissed):
Discussed with @eric-heiden yesterday and we agreed it's ok to remove this.

jcarius-nv (approved):

PR #3929, opened 2026-08-15
Add per-particle display colors for deformable surfaces

### LINKED ISSUE #3852: [REQ] Deformable surfaces have no colour: Model carries no appearance data for soft bodies, cloth or particles
### Description

Deformable surfaces have no appearance data at the `Model` level, so both the camera sensors and the viewer render them in a fixed colour that cannot be set.

In `SensorTiledCamera`, a soft-body surface-triangle hit is tagged with a sentinel rather than a shape index (`newton/_src/sensors/warp_raytrace/raytrace.py`):

```python
TRIANGLE_MESH_SHAPE_ID = wp.uint32(0xFFFFFFFD)
PARTICLES_SHAPE_ID     = wp.uint32(0xFFFFFFFE)
MAX_SHAPE_ID           = wp.uint32(0xFFFFFFF0)
```

and the albedo lookup in `render.py` is guarded on that index being a real shape:

```python
albedo_color = wp.vec3f(1.0)
if closest_hit.shape_index < raytrace.MAX_SHAPE_ID:
    albedo_color = srgb_to_linear_wp(shape_colors[closest_hit.shape_index])

if wp.static(config.enable_textures) and closest_hit.shape_index < raytrace.MAX_SHAPE_ID:
    ...
```

Both sentinels are above `MAX_SHAPE_ID`, so neither branch is reachable for deformables: the albedo stays at the literal `wp.vec3f(1.0)` and the texture path is unreachable as well. Every soft body, cloth and particle renders pure white regardless of the scene.

The same gap shows up in the viewer, from the other direction. `Viewer.log_mesh()` accepts `color`, `roughness` and `metallic`, but `ViewerBase._log_triangles()` calls it without any of them:

```python
def _log_triangles(self, state: newton.State):
    if self.model.tri_count:
        points = self._apply_layer_transform_to_points(state.particle_q)
        self.log_mesh(
            self._qualify("/model/triangles"),
            points,
            self.model.tri_indices.flatten(),
            hidden=not self.show_triangles or self._layer_force_hidden(),
            backface_culling=False,
        )
```

There is nothing to pass: `Model` carries `shape_color` for rigid shapes and no counterpart for deformables (`Model.particle_colors` is VBD graph colouring, not appearance), and `ModelBuilder.add_soft_mesh` / `add_soft_grid` / `add_cloth_mesh` take no colour argument. So the backends are not the limitation — the data is missing one layer up, which is why both are affected.

### Motivation / Use Case

We generate a synthetic manipulation dataset with a UR7e + 2F-140 grasping deformable objects, and ship the sensor's RGB and depth alongside per-element stress. Three different objects — a silicone cube, an apple and a banana — currently render as the same white, so the images carry shape and nothing else. Colour is the cheapest signal separating one object from another, and its absence is visible in every frame we would train or evaluate on.

More generally, anything using the camera sensors for perception on deformables is affected: cloth, soft grippers, food handling. And since a corpus is expensive to regenerate, the appearance has to be right before the run rather than after.

Two notes on timing, offered in case they are useful:

* #3327 (draft) adds an `elif ... == TRIANGLE_MESH_SHAPE_ID` branch in `render.py`, but keyed on a new per-face texture id and assigning the sampled texture directly. An untextured deformable still resolves to `wp.vec3f(1.0)`, and particles are not covered. A base colour looks orthogonal to it and much smaller.
* #3276 and #3627 introduce further camera implementations, and both carry the same guard forward unchanged, so the behaviour would be inherited by the replacements.

This may also be the "cheap, self-contained first step" #3223 describes — carrying material onto the sim mesh — without the embedding/skinning runtime that issue is really about.

### Alternatives Considered

* **Per-shape colour only.** Does not apply: deformables never receive a shape index, which is the root of the guard.
* **A visual-only rigid proxy mesh tracking the particles.** Works for colour and textures, but `bvh_shape_bounds` are computed once in local space and only re-transformed by `bvh_refit_shapes`, so a deforming proxy needs oversized bounds or a per-frame rebuild. Fragile, and it duplicates geometry that already exists.
* **A single hard-coded non-white default.** Distinguishes deformables from the background but not from each other.

### Possible shape (open to discussion)

Mirroring `shape_color`, an optional per-surface-triangle array, with the builder methods that create those triangles taking a colour:

* `Model.tri_color: wp.array[wp.vec3] | None` (and `Model.particle_color` for the particle path)
* `color=` on `add_soft_mesh` / `add_soft_grid` / `add_cloth_mesh`, defaulting to `None` so current behaviour is unchanged
* in `render.py`, a lookup on the existing sentinel branch — `ClosestHit` already carries `face_idx`, `bary_u` and `bary_v`, and they are already written on the triangle-mesh hit path, so the renderer side is a lookup rather than new geometry work
* in the viewer, passing it through the existing `log_mesh(color=...)` parameter

Per-vertex colour on the particles would give smooth interpolation via the barycentrics that are already there, if that is preferred to flat per-face.

We would be glad to prepare a PR if that shape sounds reasonable, or a different one if you would rather it landed elsewhere — in particular whether it should go into `SensorTiledCamera` given #3276, or wait for the newer sensor.

Checked against `main` @ 71fe5e63 as well as our pinned `d58e7026`; the guard is identical in both.

kms8720:
## Description

Closes #3852.

Add optional per-particle display colors as shared appearance data for standalone particles and deformable surfaces.

- Store authored colors in `Model.particle_display_color`, separate from the existing VBD graph-colouring `Model.particle_colors`.
- Add display-color authoring across particle, cloth, and soft-body builder entry points.
- Use direct lookup for particle hits and barycentric interpolation for simulation-triangle hits in `SensorTiledCamera`.
- Add optional per-vertex `colors=` support to `Viewer.log_mesh()` and forward deformable colors through the viewer backends.
- Preserve existing fallbacks when no color is authored; use white for missing entries in partially authored models and validate the finalized array length.
- Preserve display colors across builder merging and replication.
- Append the new color arguments after the existing keyword-only parameters so
  deprecated positional calls keep their bindings.

Per-particle colors interpolate across shared vertices; hard color discontinuities require splitting the corresponding particles.

## Checklist

## Test plan

```bash
uv run --extra dev python -m unittest -q \
  newton.tests.test_particle_display_color \
  newton.tests.test_sensor_tiled_camera \
  newton.tests.test_viewer_layers \
  newton.tests.test_viewer_mesh_colors \
  newton.tests.test_viewer_particle_flags \
  newton.tests.test_viewer_rerun_hidden \
  newton.tests.test_viewer_usd

uv run --extra dev python -m unittest -q \
  newton.tests.test_builder_replicate

UV_PYTHON=3.12 uvx --python 3.12 pre-commit run -a
```

- Focused builder, camera, and viewer coverage: 87 tests run, 4 skipped on a CPU-only macOS ARM host.
- Builder replication: 20 tests passed.
- Model coverage: 170 passed, 110 skipped.
- Exact-head validation on an NVIDIA H200: 17/17 targeted checks passed, including full and depth-only particle/triangle rendering, color conversion, multi-world indexing, fallbacks, ViewerGL packing, and structural validation.
- All pre-commit hooks passed.

## New feature / API change

```python
import newton
import warp as wp

builder.add_particles(
    pos=[wp.vec3(0.0, 0.0, 0.0), wp.vec3(1.0, 0.0, 0.0)],
    vel=[wp.vec3(), wp.vec3()],
    mass=[1.0, 1.0],
    colors=[(1.0, 0.2, 0.1), None],
)

model = builder.finalize()
# The second particle uses the white display fallback.
```

## Summary by CodeRabbit

* **New Features**
  * Added per-particle display colors for standalone particles, deformable surfaces, and tiled-camera rendering.
  * Added per-vertex mesh color support across viewers, with texture-aware precedence.
  * Added color options to particle, cloth, soft-body, and mesh creation workflows.
  * Display colors interpolate across deformable triangles and persist during particle filtering.

* **Bug Fixes**
  * Uncolored particles and meshes retain the white fallback.
  * Added validation for mismatched color data and improved color preservation across frames and viewer backends.


kms8720:
CLA is sorted now, and I have merged current `main` — the viewer conflicts came from the dynamic-mesh rework. `log_mesh` gained `dynamic` upstream while this branch was adding `colors`; I kept `colors` last so an upstream positional call that passes `dynamic` retains its binding.

Two things worth pointing a reviewer at:

- `MeshInstancerGL` now sets the vertex-colour attribute up inside `_bind_mesh_buffers()` rather than inline, so `set_mesh()` re-establishes it. The divisor is left unset, keeping location 9 per-vertex.
- `ViewerGL` carries `base_color` onto a replacement mesh. This feature makes `color` a derived field and `base_color` the authored one, so the existing `replacement.color = existing.color` had become dead — a mesh logged once with `color=` and later re-logged at a different size reverted to the constructor default. Both fixes have regression coverage; the replace branch in particular was the one path the colour tests were not exercising.

Local run on a CPU-only macOS ARM host: 115 tests in the affected modules pass, 4 skipped, and all pre-commit hooks are clean.

kms8720:
Hi @mzamoramora-nvidia @daniela-hase, could you review this PR and advise on the scope following #3053?

I left a [follow-up in #3852 last week](https://github.com/newton-physics/newton/issues/3852#issuecomment-5487653703) about avoiding overlap with the newly merged `tri_color` support. Once the direction is clear, I can update the branch and resolve the current conflicts.

GPU CI is also awaiting maintainer approval. If another maintainer would be better placed to review this, could you point me to them? Thanks!

Issue #3931, opened 2026-08-15
[BUG] ViewerGL(headless=True) raises NoSuchDisplayException unless pyglet is put in headless mode first


rakhimovv:
### Bug Description

The headless snippet in the visualization guide does not run on a machine without a display.

`docs/guide/visualization.rst` presents `ViewerGL(headless=True)` followed by `get_frame()` as a
self-contained example, but executing it on a headless Linux machine raises
`pyglet.display.xlib.NoSuchDisplayException: Cannot connect to "None"` while the viewer is being
constructed.

The cause is ordering. pyglet picks its windowing backend when `pyglet.window` is first imported,
and defaults to Xlib on Linux. Newton only ever *reads* the corresponding option — in
`newton/_src/viewer/gl/opengl.py`, `self.headless = pyglet.options.get("headless", False)` — and
never sets it. Nothing under `newton/` or `docs/` mentions `PYGLET_HEADLESS`, so the constructor
argument `headless=True` on its own is not sufficient to get a headless context.

Measured on a headless Linux box, all three against the same commit:

| attempt | result |
|---|---|
| documented snippet exactly as written | `NoSuchDisplayException: Cannot connect to "None"` |
| `pyglet.options["headless"] = True` set **before** `import newton.viewer` | viewer constructs |
| `pyglet.options["headless"] = True` set **after** `import newton.viewer`, before constructing | viewer constructs |

The third row is the useful one: importing `newton.viewer` does not yet import `pyglet.window`
(`"pyglet.window" in sys.modules` is `False` at that point), so the option can still be set from
inside `ViewerGL.__init__`.

### Suggested fix

Have `ViewerGL.__init__` set `pyglet.options["headless"] = True` when `headless=True` is
requested. The documented snippet would then work as written, with no environment variable and no
import-order rule for users to know about.

One trade-off is worth a maintainer's view before anyone implements it: `pyglet.options` is
process-global, so a constructor writing to it would also affect a later
`ViewerGL(headless=False)` in the same process. Reasonable variants are to set it only when it is
currently unset, to set it and document the precedence, or to leave the code as is and document
the requirement instead. I am happy to send a patch for whichever option you prefer.

Independently of that decision, the current requirement is undocumented, so I have opened a
documentation-only PR that records it. If the code fix lands, that note shrinks to a sentence.

### Reproduction Script

```python
# Headless Linux machine, DISPLAY unset, PYGLET_HEADLESS not set in the environment.
import newton

[...]

PR #3932, opened 2026-08-15
Document pyglet headless requirement for ViewerGL

rakhimovv:
## Description

`docs/guide/visualization.rst` presents `ViewerGL(headless=True)` followed by `get_frame()` as a
self-contained snippet, but on a machine without a display it raises
`pyglet.display.xlib.NoSuchDisplayException: Cannot connect to "None"` at construction. pyglet
selects its windowing backend when `pyglet.window` is first imported and defaults to Xlib on
Linux; Newton reads `pyglet.options["headless"]` in `newton/_src/viewer/gl/opengl.py` but never
sets it, and neither `newton/` nor `docs/` mentions `PYGLET_HEADLESS`.

This adds a note beside that snippet recording the requirement and both ways to satisfy it —
`PYGLET_HEADLESS=1` in the environment, or `pyglet.options["headless"] = True` set before
`newton` is imported — along with `PYGLET_HEADLESS_DEVICE` for choosing the GPU on a machine with
more than one.

Documentation only; no behavior change.

Refs #3931, which proposes having `ViewerGL.__init__` set the option itself. If that lands, this
note shrinks to a sentence.

## Checklist

- [ ] New or existing tests cover these changes — not applicable, documentation only
      — added `changelog/+docs-viewergl-headless-pyglet-3f7a9c21.skip`, as there is no
      user-facing code change

## Test plan

```
```

Builds clean with warnings as errors, which is the same invocation `.github/workflows/pr.yml`
uses. `uvx pre-commit run -a` passes.

The statements in the note were checked against the pyglet in the lockfile (2.1.14) rather than
written from memory:

- `pyglet/__init__.py` resolves every option from `PYGLET_<OPTION_NAME>`, so `PYGLET_HEADLESS`
  and `PYGLET_HEADLESS_DEVICE` correspond to the `headless` and `headless_device` options.
- `headless_device` defaults to `0`.
- The exception raised when the option is not set is `pyglet.display.xlib.NoSuchDisplayException`
  (`pyglet.canvas` was renamed to `pyglet.display` in pyglet 2).

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for enabling pyglet headless mode before using ViewerGL on Linux.
  * Documented environment-variable and Python configuration options.
  * Added instructions for selecting the GPU used in headless mode.
  * Recorded the existing headless-mode requirement in the changelog.

### DIFF
--- a/docs/guide/visualization.rst
+++ b/docs/guide/visualization.rst
@@ -188,6 +188,30 @@ Warp array on the viewer device:
     # Returns a wp.array with shape (height, width, 3), dtype wp.uint8
     frame = viewer.get_frame()
 
+.. note::
+
+    On a machine without a display, pyglet must also be put in headless mode. pyglet binds its
+    display backend the first time that backend is imported, and Newton imports pyglet's window
+    and display modules when the first :class:`~newton.viewer.ViewerGL` is constructed, so the
+    option has to be set before that point. Otherwise the snippet above fails with
+    ``pyglet.display.xlib.NoSuchDisplayException: Cannot connect to "None"`` on Linux, since
+    pyglet defaults to Xlib. Either set the environment variable::
+
+        PYGLET_HEADLESS=1 python your_script.py
+
+    or set the option in Python before creating the viewer::
+
+        import newton
+        import pyglet
+
+        pyglet.options["headless"] = True
+
+        viewer = newton.viewer.ViewerGL(headless=True)
+
+    On a machine with several GPUs, ``PYGLET_HEADLESS_DEVICE`` (or
+    ``pyglet.options["headless_device"]``) selects which one renders; it defaults to ``0``,
+    which is not necessarily the device the rest of the simulation runs on.
+
 **Custom UI panels:**
 
 :meth:`~newton.viewer.ViewerGL.register_ui_callback` adds custom imgui UI elements to the viewer.
--- a/changelog/+docs-viewergl-headless-pyglet-3f7a9c21.skip
+++ b/changelog/+docs-viewergl-headless-pyglet-3f7a9c21.skip
@@ -1,0 +1,1 @@
+Documentation only: records the existing pyglet headless requirement for ViewerGL.


eric-heiden:
Thanks for improving our documentation. Getting `ViewerGL` running on a machine without a display is a common stumbling block, and having the exact `NoSuchDisplayException` message plus both ways to set the option in the visualization guide will save people real time.

One small correction below: the note ties the deadline for enabling pyglet's headless mode to `import newton`, which is stricter than what Newton actually requires.

docs/guide/visualization.rst
```diff
@@ -188,6 +188,28 @@ Warp array on the viewer device:
...
+
+        PYGLET_HEADLESS=1 python your_script.py
+
+    or set the option in Python before the import::
+
+        import pyglet
+
+        pyglet.options["headless"] = True
+
+        import newton  # noqa: E402
```
eric-heiden: ⚪ The pyglet half of this is correct, but the deadline is not `import newton`.

Newton has no module-level pyglet import. Every one of them is function-local, and `newton/viewer.py` calls this out at the top: "Import all viewer classes (they handle missing dependencies at instantiation time)". The first code that touches the display backend lives in `RendererGL.__init__` in `newton/_src/viewer/gl/opengl.py`, which imports pyglet and then opens `pyglet.window.Window(...)`. So the real deadline is the first `ViewerGL()` construction, not the `import newton` statement.

The advice as written does work, it is just stricter than it needs to be. It pushes readers into reordering their imports and adding a `# noqa: E402` pragma they do not need, and anyone who already has `import newton` at the top of a module or a notebook cell will read this and conclude the Python option is closed to them and that they have to restart the interpreter.

Tying the deadline to viewer construction keeps the guidance accurate and lets the snippet drop the pragma:

```suggestion
    On a machine without a display, pyglet must also be put in headless mode. pyglet binds its
    display backend the first time that backend is imported, and Newton imports pyglet's window
    and display modules when the first :class:`~newton.viewer.ViewerGL` is constructed, so the
    option has to be set before that point. Otherwise the snippet above fails with
    ``pyglet.display.xlib.NoSuchDisplayException: Cannot connect to "None"`` on Linux, since
    pyglet defaults to Xlib. Either set the environment variable::

        PYGLET_HEADLESS=1 python your_script.py

    or set the option in Python before creating the viewer::

        import newton
        import pyglet

        pyglet.options["headless"] = True

        viewer = newton.viewer.ViewerGL(headless=True)
```

rakhimovv: You're right, the deadline is viewer construction, not `import newton`. Applied your wording in a0c0e96.

I checked it on a display-less Linux box with a single H100: `import newton` at the top of the file, then `pyglet.options["headless"] = True`, then `ViewerGL(headless=True)`, and the viewer builds without the exception. The `# noqa: E402` was not earning its keep.

I reworded the commit body too, since it made the same claim about the import.

eric-heiden (approved):
Thanks!

PR #3942, opened 2026-08-16
fix(solvers): defer annotation evaluation so ModelBuilder() works on Python 3.10

### LINKED ISSUE #3941: ModelBuilder() raises TypeError on Python 3.10: solver.py uses PEP 604 over wp.array[...] without postponed evaluation
### Summary

`newton.ModelBuilder()` raises `TypeError` on **Python 3.10**, which `pyproject.toml` declares as supported (`requires-python = ">=3.10"`, plus a 3.10 classifier). Python 3.11+ is unaffected.

`newton/_src/solvers/solver.py` annotates with PEP 604 over a subscripted Warp type (`wp.array[wp.bool] | None`) without postponed evaluation, so the annotation is evaluated at class-definition time — and `wp.array[wp.bool]` is not a `type`.

### Reproduction

```python
import newton
newton.ModelBuilder()
```

```
TypeError: Union[arg, ...]: each arg must be a type. Got wp.array[wp.bool].
```

Import chain: `ModelBuilder.__init__` → `builder.py:1614` → `solvers/mujoco/__init__.py` → `solver_mujoco.py:40` → `solvers/coupled/__init__.py` → `solver_coupled.py:19` → `solvers/solver.py:188 class SolverBase` → **`solver.py:221`**:

```python
def _normalize_reset_world_mask(self, world_mask: wp.array[wp.bool] | None) -> wp.array[wp.bool] | None:
```

(also `solver.py:345`)

### Environment

| | |
|---|---|
| newton | 1.5.0 |
| warp-lang | 1.16.0 |
| Python | 3.10.20 |
| OS | Linux x86_64 |

### Why this is Python 3.10 only

`typing._type_check` was relaxed in 3.11. Isolated from Newton and Warp entirely:

```python
import typing
class Fake: pass          # not callable, like warp's _ArrayAnnotation
typing.Union[Fake(), None]
```

| Python | result |
|---|---|
| 3.10.20 | `TypeError: Union[arg, ...]: each arg must be a type` |
| 3.11.15 | `typing.Optional[<__main__.Fake object ...>]` — accepted |

Warp's side, for completeness: `wp.array[wp.bool]` returns `_ArrayAnnotation`, not a type. In 1.14 it has no `__or__` (`TypeError: unsupported operand type(s) for |`); in 1.16 `__or__` is implemented as `typing.Union[self, other]`, which is what reaches `_type_check`. Since newton 1.5.0 requires `warp-lang>=1.16`, the 1.16 path is the relevant one, and it fails only on 3.10.

newton 1.3.0 does not define `_normalize_reset_world_mask` and is unaffected.

### Why CI does not catch it

This looks like a CI blind spot rather than an untested area:

* the full unit-test suite runs under the pinned `.python-version` — **3.12** — where the annotation evaluates fine;
* the multi-version jobs in `ci.yml` run `python -c "import newton"` on 3.10–3.14, and **plain `import newton` succeeds on 3.10**. The failure is at `ModelBuilder()` construction, which that job never reaches.

So every declared-supported Python is exercised, but only for import, and the one version that breaks breaks just past that point.

A test that constructs `ModelBuilder()` under 3.10 would close the gap.

### Fix

Add to the top of `newton/_src/solvers/solver.py`:

```python
```

`newton/_src/solvers/coupled/solver_coupled.py` in the same subtree already has it.

Verified: with that single line, `newton.ModelBuilder()` constructs on 3.10, and `add_mjcf` then imports a 271-geom MuJoCo model (`body=37 shape=302 joint=37 dof=43`, matching the source model's `nv=43`).

Behaviour-preserving for this module: it contains no `typing.get_type_hints`, `__annotations__`, `pydantic` or `@dataclass` usage, so nothing reads these annotations at runtime. `newton/tests/test_api.py` does call `get_type_hints`, but only on importer functions and `ModelBuilder` methods — not on `SolverBase` — and its `SolverSemiImplicit` assertions use `inspect.signature(...).kind`, which is unaffected by PEP 563.

If you would rather keep eager evaluation, quoting the two annotations or using `Optional[...]` works equally well; the future import just fixes the file as a whole.

### Notes

Happy to open a PR with the fix plus a regression test that constructs `ModelBuilder()`, if that is welcome.

Separately, and probably its own issue: the MJCF importer routes meshes through `trimesh`, which cannot read MuJoCo's own `.msh` format (`NotImplementedError: file_type 'msh' not supported`, `utils/mesh.py:1443`). Models shipping `.msh` assets — as LIBERO/robosuite do — need their meshes converted before `add_mjcf` accepts them. Let me know if you'd like that filed.

sparkle2820:
## Description

Fixes `newton.ModelBuilder()` raising `TypeError` on Python 3.10, and closes the CI gap that let it through.

`SolverBase` annotates `wp.array[wp.bool] | None`. That is evaluated at class-definition time, and `wp.array[wp.bool]` is not a `type` — so `typing._type_check` rejects it. Python 3.11 relaxed that check, which is why this reproduces only on 3.10, a version `pyproject.toml` declares as supported.

Three commits:

1. **`from __future__ import annotations` in `newton/_src/solvers/solver.py`.** One line, placed to match `solvers/coupled/solver_coupled.py` in the same subtree, which already defers.
2. **A regression test, `newton/tests/test_python_compatibility.py`**, run by CI on 3.10–3.14. See below.
3. Changelog fragment.

Closes #3941

### Where the regression guard lives

*(Revised after review — the first version of this PR appended `newton.ModelBuilder()`
to the five "minimal import" steps instead. @jcarius-nv pointed out that this catches
the defect only because constructing a builder happens to traverse the solver import
chain, which is incidental rather than a stated contract, and that CONTRIBUTING.md asks
bug fixes to carry a regression test. That is right, and this is his suggested test.)*

The guard has two halves, because either alone would miss this:

- **A named test.** `newton/tests/test_python_compatibility.py` resolves every stable
  export of `newton.solvers`. Resolving one imports the module that defines it, so the
  test fails on any module whose import-time annotations do not evaluate on a supported
  interpreter — the class of defect this is, stated as a contract rather than reached
  by accident.
- **Run on every supported version.** The full suite runs under the pinned
  `.python-version` (3.12), where the annotation evaluates fine, so a test alone would
  stay green. CI now invokes this one test on 3.10–3.14, replacing the smoke steps.

Verified on Python 3.10.20 / warp-lang 1.16.0 / newton 1.5.0, which carries the defect
verbatim (no postponed annotations; `wp.array[wp.bool] | None` at `solver.py:221`
and `:345`), by running the test with and without the one-line fix in fresh
interpreters:

```
defect present   FAIL — 9 errors, every one
                 TypeError: Union[arg, ...]: each arg must be a type. Got wp.array[wp.bool].
patched          PASS
```

Nine of the ten stable exports fail, not one: most solvers subclass `SolverBase`.

## Checklist

- [x] For user-facing changes, a fragment has been added

## Test plan

Verified on newton 1.5.0 + warp-lang 1.16.0 + Python 3.10.20:

```
# without the fix
$ python -c "import newton; newton.ModelBuilder()"
TypeError: Union[arg, ...]: each arg must be a type. Got wp.array[wp.bool].

# with the fix
$ python -c "import newton; newton.ModelBuilder(); print('ok')"
ok
```

Beyond construction, `add_mjcf` then imports a 271-geom MuJoCo model successfully
(`body=37 shape=302 joint=37 dof=43`, matching the source model's `nv=43`).

Checked that deferring evaluation is behaviour-preserving for this module:

- `solver.py` contains no `typing.get_type_hints`, `__annotations__`, `pydantic` or `@dataclass` usage, so nothing reads these annotations at runtime;
- `newton/tests/test_api.py` does call `get_type_hints`, but only on importer functions and `ModelBuilder` methods, not on `SolverBase`; its `SolverSemiImplicit` assertions use `inspect.signature(...).kind`, which PEP 563 does not affect.

## Bug fix

**Steps to reproduce:**

1. Install `newton==1.5.0` under Python 3.10 (warp-lang 1.16.0 comes with it).
2. Construct a `ModelBuilder`.
3. `TypeError` is raised from `solver.py:221` during the solver import chain, before any user code runs.

**Minimal reproduction:**

```python
import newton

newton.ModelBuilder()
# TypeError: Union[arg, ...]: each arg must be a type. Got wp.array[wp.bool].
```

The `typing` behaviour difference, isolated from Newton and Warp:

```python
import typing
class Fake: pass          # not callable, like warp's _ArrayAnnotation
typing.Union[Fake(), None]
# 3.10 -> TypeError: Union[arg, ...]: each arg must be a type
# 3.11 -> typing.Optional[<__main__.Fake object ...>]
```

### Alternatives

If you would rather keep eager evaluation in this file, **quoting** the two annotations
works. I went with the future import because it fixes the file as a whole and matches a
sibling module.

~~or using `Optional[...]`~~ — **correction, thanks @jcarius-nv:** `Optional[...]` does
not work. It fails the same `_type_check`, just with a different message. Checked on
both warp-lang 1.16.0 and 1.14.0 under Python 3.10:

```python
wp.array[wp.bool] | None      # TypeError: Union[arg, ...]: each arg must be a type.
Optional[wp.array[wp.bool]]   # TypeError: typing.Optional requires a single type.
'wp.array[wp.bool] | None'    # ok
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue that caused `ModelBuilder()` to fail during initialization on Python 3.10.
  * Improved compatibility with supported Python versions by safely handling type annotations.
* **Tests**
  * Added compatibility checks for importing stable solver functionality across Python 3.10–3.14.
* **Documentation**
  * Added a changelog entry describing the Python 3.10 compatibility fix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,15 +42,15 @@ jobs:
       - name: Test CI tooling
         run: uv run --no-project -m unittest discover -s scripts/ci/tests
       - name: Test minimal import (Python 3.10)
-        run: uv run --python 3.10 python -c "import newton; print(f'Newton {newton.__version__} imported successfully')"
+        run: uv run --python 3.10 python -c "import newton; newton.ModelBuilder(); print(f'Newton {newton.__version__} imported successfully')"
       - name: Test minimal import (Python 3.11)
-        run: uv run --python 3.11 python -c "import newton; print(f'Newton {newton.__version__} imported successfully')"
+        run: uv run --python 3.11 python -c "import newton; newton.ModelBuilder(); print(f'Newton {newton.__version__} imported successfully')"
       - name: Test minimal import (Python 3.12)
-        run: uv run --python 3.12 python -c "import newton; print(f'Newton {newton.__version__} imported successfully')"
+        run: uv run --python 3.12 python -c "import newton; newton.ModelBuilder(); print(f'Newton {newton.__version__} imported successfully')"
       - name: Test minimal import (Python 3.13)
-        run: uv run --python 3.13 python -c "import newton; print(f'Newton {newton.__version__} imported successfully')"
+        run: uv run --python 3.13 python -c "import newton; newton.ModelBuilder(); print(f'Newton {newton.__version__} imported successfully')"
       - name: Test minimal import (Python 3.14)
-        run: uv run --python 3.14 python -c "import newton; print(f'Newton {newton.__version__} imported successfully')"
+        run: uv run --python 3.14 python -c "import newton; newton.ModelBuilder(); print(f'Newton {newton.__version__} imported successfully')"
 
   dependency-install-test:
     needs: minimal-import-test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ on:
       - release-*
 
 jobs:
-  minimal-import-test:
+  python-compatibility-test:
     if: github.event_name != 'push' || github.repository == 'newton-physics/newton'
     runs-on: ubuntu-latest
     steps:
@@ -41,19 +41,19 @@ jobs:
           version: "0.11.26"
       - name: Test CI tooling
         run: uv run --no-project -m unittest discover -s scripts/ci/tests
-      - name: Test minimal import (Python 3.10)
-        run: uv run --python 3.10 python -c "import newton; newton.ModelBuilder(); print(f'Newton {newton.__version__} imported successfully')"
-      - name: Test minimal import (Python 3.11)
-        run: uv run --python 3.11 python -c "import newton; newton.ModelBuilder(); print(f'Newton {newton.__version__} imported successfully')"
-      - name: Test minimal import (Python 3.12)
-        run: uv run --python 3.12 python -c "import newton; newton.ModelBuilder(); print(f'Newton {newton.__version__} imported successfully')"
-      - name: Test minimal import (Python 3.13)
-        run: uv run --python 3.13 python -c "import newton; newton.ModelBuilder(); print(f'Newton {newton.__version__} imported successfully')"
-      - name: Test minimal import (Python 3.14)
-        run: uv run --python 3.14 python -c "import newton; newton.ModelBuilder(); print(f'Newton {newton.__version__} imported successfully')"
+      - name: Python compatibility (Python 3.10)
+        run: uv run --python 3.10 python -m unittest newton.tests.test_python_compatibility
+      - name: Python compatibility (Python 3.11)
+        run: uv run --python 3.11 python -m unittest newton.tests.test_python_compatibility
+      - name: Python compatibility (Python 3.12)
+        run: uv run --python 3.12 python -m unittest newton.tests.test_python_compatibility
+      - name: Python compatibility (Python 3.13)
+        run: uv run --python 3.13 python -m unittest newton.tests.test_python_compatibility
+      - name: Python compatibility (Python 3.14)
+        run: uv run --python 3.14 python -m unittest newton.tests.test_python_compatibility
 
   dependency-install-test:
-    needs: minimal-import-test
+    needs: python-compatibility-test
     strategy:
       fail-fast: false
       matrix:
--- a/changelog/3941.fixed.md
+++ b/changelog/3941.fixed.md
@@ -1,0 +1,1 @@
+Fix `ModelBuilder()` raising `TypeError` on Python 3.10, caused by eager evaluation of a PEP 604 annotation over a subscripted Warp type in `SolverBase`.
--- a/newton/_src/solvers/solver.py
+++ b/newton/_src/solvers/solver.py
@@ -1,6 +1,8 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 The Newton Developers
 # SPDX-License-Identifier: Apache-2.0
 
+from __future__ import annotations
+
 from typing import Any
 
 import warp as wp
--- a/newton/tests/test_python_compatibility.py
+++ b/newton/tests/test_python_compatibility.py
@@ -12,12 +12,12 @@ postponed -- which is how a Python 3.10 ``TypeError`` reached users through
 
 import unittest
 
+import newton
+
 
 class TestPythonCompatibility(unittest.TestCase):
     def test_stable_solver_exports_resolve(self):
         """Resolve every stable public solver export."""
-        import newton
-
         for name in newton.solvers.__all__:
             if name == "experimental":
                 continue
--- a/newton/tests/test_python_compatibility.py
+++ b/newton/tests/test_python_compatibility.py
@@ -1,0 +1,29 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
+
+"""Compatibility checks that must hold on every supported Python version.
+
+``pyproject.toml`` declares support down to Python 3.10, so the stable public
+API has to resolve there. Resolving a solver export imports the module that
+defines it, and annotations in that module are evaluated at import time unless
+postponed -- which is how a Python 3.10 ``TypeError`` reached users through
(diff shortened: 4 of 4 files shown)


jcarius-nv (commented):
Thanks for the thorough diagnosis. I confirmed that Python 3.10 is still an intentional Newton support target—not just stale packaging metadata—and reproduced the reported failure at the base commit and success with the `from __future__ import annotations` fix. The solver change and changelog fragment look good.

I would, however, reshape the regression coverage. Constructing `ModelBuilder` inside a step named “minimal import” selects one constructor because it currently happens to traverse the solver import chain. It does not provide a systematic compatibility boundary, and `CONTRIBUTING.md` asks bug fixes to include a regression unit test.

Could we add a small compatibility test such as:

```python
import unittest


class TestPythonCompatibility(unittest.TestCase):
    def test_stable_solver_exports_resolve(self):
        """Resolve every stable public solver export."""
        import newton

        for name in newton.solvers.__all__:
            if name == "experimental":
                continue
            with self.subTest(name=name):
                _ = getattr(newton.solvers, name)
```

Resolving `SolverBase` imports `solver.py` and catches the exact Python 3.10 annotation failure, while resolving the complete stable solver surface gives the test a broader, principled purpose. I verified that all of these exports resolve successfully on Python 3.10 with the proposed fix.

The existing multi-version steps can then invoke this test directly:

```yaml
run: uv run --python 3.10 python -m unittest newton.tests.test_python_compatibility
```

and likewise for Python 3.11–3.14. Renaming the job/steps from “minimal import” to “Python compatibility” would make the new contract explicit. A separate `ModelBuilder` construction test is not necessary: the compatibility test catches this defect at its source, while the normal suite already exercises `ModelBuilder` extensively on the pinned Python version.

One small PR-description correction: `Optional[wp.array[wp.bool]]` also fails `_type_check` on Python 3.10, so quoting or postponing the annotations are the viable alternatives.

With that test change, I have no other concerns.

sparkle2820:
Thanks — applied, and the reasoning holds up. Constructing a `ModelBuilder` inside a
"minimal import" step did reach the defect, but only via the solver import chain, which
is a route rather than a contract. `newton/tests/test_python_compatibility.py` is your
test as written; the CI steps now invoke it on 3.10–3.14 and the job is renamed. No
separate `ModelBuilder` construction test, per your note.

I checked that the test actually fails without the fix rather than assuming it, on the
environment #3941 was reported from (Python 3.10.20 / warp-lang 1.16.0 / newton 1.5.0 —
no postponed annotations, `wp.array[wp.bool] | None` at `solver.py:221` and `:345`),
running it with and without the one-line fix in fresh interpreters:

```
defect present   FAIL — 9 errors, every one
                 TypeError: Union[arg, ...]: each arg must be a type. Got wp.array[wp.bool].
patched          PASS
```

Two details from doing it that may be worth having:

- **Nine of the ten stable exports fail, not one.** Most solvers subclass `SolverBase`,
  so the whole surface goes down together. The test's breadth is doing real work here,
  not just giving the check a tidier name.
- **`dependency-install-test` declares `needs: minimal-import-test`**, so the rename has
  a second edit site. Updated; `yaml.safe_load` plus a check that no `needs:` points at a
  job id that no longer exists.

Your `Optional[...]` correction is right and the PR description is fixed. Confirmed on
warp-lang 1.16.0 and 1.14.0 under 3.10:

```python
wp.array[wp.bool] | None      # TypeError: Union[arg, ...]: each arg must be a type.
Optional[wp.array[wp.bool]]   # TypeError: typing.Optional requires a single type.
'wp.array[wp.bool] | None'    # ok
```

One caveat on my verification, so it is not read as stronger than it is: the check above
ran against the installed 1.5.0 wheel, patched in place, because a full source checkout
of this repo is slow on my link. That wheel matches the PR base in the two lines that
matter, but it is not the base commit itself.

eric-heiden:
Thanks for addressing Jan's comments @sparkle2820! Could you please fix the pre-commit CI failure so this PR becomes mergeable?

sparkle2820:
@eric-heiden Fixed — pre-commit.ci is green now. Thanks!

jcarius-nv (approved):

PR #3949, opened 2026-08-17
Optimize linear texture conversion

camevor:
## Description

Speeds up USD texture import in two independent ways.

**Linear→sRGB conversion for `uint8` textures.** `linear_texture_to_srgb()` evaluated
`np.power()` over every pixel. Since `uint8` has only 256 distinct inputs, the transfer
function is now precomputed into a 256-entry lookup table and applied per channel. The
LUT is built with the same float32 arithmetic the general integer path uses, so results
are bit-identical to before — verified exhaustively over all 256 values.

`uint8` RGBA is the only form that reaches this function in practice: it is called from
`_resolve_color_texture_asset()`, whose input always comes from `load_texture()` →
PIL `.convert("RGBA")`. The general integer and float paths are unchanged.

**Redundant visual-material resolution during import.** `get_mesh()` unconditionally
called `resolve_material_properties_for_prim()`, which for a linear-space texture
eagerly decodes the image and converts it. `parse_usd()` already resolves and memoizes
the same material itself via `_get_material_props_cached()`, so this work was performed
twice per mesh prim — once uncached. A new `load_visual_materials` keyword lets callers
skip it, and `parse_usd()` now passes `False`.

Colliders that are not viewport-drawn still reach the viewer through its
`show_collision` / `show_static` overlays, so `parse_usd()` re-attaches `texture`,
`roughness`, and `metallic` on that path. Display color was already passed separately
via `shape_params["color"]` and is unaffected.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - USD mesh loading can skip visual-material processing when only geometry data is needed, improving import efficiency.
  - Displayed collision meshes retain available texture and physically based material properties.
  - Faster, consistent linear-to-sRGB conversion for 8-bit textures.

- **Bug Fixes**
  - Preserved texture channels, alpha data, and writable output during conversion.

- **Documentation**
  - Added guidance to use keyword arguments for USD mesh-loading options.

- **Tests**
  - Added coverage for exact 8-bit texture conversion and USD material-loading behavior.

### DIFF
--- a/newton/_src/usd/utils.py
+++ b/newton/_src/usd/utils.py
@@ -1110,6 +1110,7 @@ def _get_mesh_from_source(
     preserve_facevarying_uvs: bool,
     compute_inertia: bool,
     apply_stage_units: bool,
+    load_visual_materials: bool,
 ) -> Mesh:
     """Load and merge mesh prims from a USD stage, path, URL, or prim subtree."""
     if Usd is None or UsdGeom is None:
@@ -1162,6 +1163,7 @@ def _get_mesh_from_source(
             vertex_splitting_angle_threshold_deg=vertex_splitting_angle_threshold_deg,
             preserve_facevarying_uvs=preserve_facevarying_uvs,
             compute_inertia=False,
+            load_visual_materials=load_visual_materials,
         )
         source_meshes.append(source_mesh)
         matrix = _relative_transform_matrix(prim, root, xform_cache)
@@ -1333,6 +1335,8 @@ def get_mesh(
     root_path: str | None = None,
     compute_inertia: bool = True,
     apply_stage_units: bool = True,
+    *,
+    load_visual_materials: bool = True,
 ) -> Mesh: ...
 
 
@@ -1351,6 +1355,8 @@ def get_mesh(
     root_path: None = None,
     compute_inertia: bool = True,
     apply_stage_units: bool = True,
+    *,
+    load_visual_materials: bool = True,
 ) -> tuple[Mesh, np.ndarray | None]: ...
 
 
@@ -1371,6 +1377,7 @@ def get_mesh(
     apply_stage_units: bool = True,
     *,
     prim: Usd.Prim,
+    load_visual_materials: bool = True,
 ) -> Mesh: ...
 
 
@@ -1391,6 +1398,7 @@ def get_mesh(
     apply_stage_units: bool = True,
     *,
     prim: Usd.Prim,
+    load_visual_materials: bool = True,
 ) -> tuple[Mesh, np.ndarray | None]: ...
 
 
@@ -1410,6 +1418,7 @@ def get_mesh(
     apply_stage_units: bool = True,
     *,
     prim: Usd.Prim | None = None,
+    load_visual_materials: bool = True,
 ) -> Mesh | tuple[Mesh, np.ndarray | None]:
     """
     Load a triangle mesh from a USD mesh prim, stage, file path, or URL.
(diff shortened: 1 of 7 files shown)


newton/_src/usd/utils.py
```diff
@@ -1333,6 +1335,8 @@ def get_mesh(
     root_path: str | None = None,
     compute_inertia: bool = True,
     apply_stage_units: bool = True,
+    *,
+    load_visual_materials: bool = True,
```
eric-heiden, line 1340: While you're at it, can you make all the arguments after `source` keyword-only?
Then you need to apply the `@deprecate_nonkeyword_arguments` decorator and add a changelog fragment for this change.

eric-heiden: Same for the other `get_mesh` overloads here.

newton/_src/usd/utils.py
```diff
@@ -1484,6 +1493,11 @@ def get_mesh(
             non-mesh prim sources from authored USD distance units to meters.
             Single mesh prim sources keep their authored coordinates for
             backward compatibility unless ``root_path`` is provided.
+        load_visual_materials: If True, resolve and attach the bound visual
+            material's color, texture, metallic, and roughness. Set to False
+            when only mesh geometry is needed. If ``load_uvs`` is True, the
+            material's shader network may still be inspected to select the UV
+            primvar used by the texture.
```
eric-heiden: ⚪ The docstring describes a narrower behavior than `load_visual_materials=False` actually has. It mentions only the bound visual material's color, texture, metallic, and roughness, but the flag gates the entire material resolution step, which also covers:

| Also skipped when `False` | Where it comes from |
| --- | --- |
| `displayColor` primvar fallback, used when a material is bound but authors neither color nor texture | `resolve_material_properties_for_prim` |
| Materials resolved through an instance prototype | `resolve_material_properties_for_prim` |
| Materials bound on `UsdGeom.Subset` children of the mesh | `resolve_material_properties_for_prim` |

As written, someone could reasonably expect a `displayColor`-only mesh to still come back tinted when they pass `False`. It is also worth stating the observable result: with `False`, `Mesh.color`, `Mesh.texture`, `Mesh.metallic`, and `Mesh.roughness` keep their `newton.Mesh` defaults instead of being resolved.

```suggestion
        load_visual_materials: If True, resolve the mesh's visual material and
            populate :attr:`newton.Mesh.color`, :attr:`newton.Mesh.texture`,
            :attr:`newton.Mesh.metallic`, and :attr:`newton.Mesh.roughness`.
            Resolution also covers materials bound through an instance
            prototype or a ``UsdGeom.Subset`` child, and the ``displayColor``
            primvar fallback. If False, those attributes keep their
            :class:`newton.Mesh` defaults; set it to False when only mesh
            geometry is needed. If ``load_uvs`` is True, the material's shader
            network may still be inspected to select the UV primvar used by
            the texture.
```

newton/_src/utils/import_usd.py
```diff
@@ -3857,7 +3862,16 @@ def _record_fallback_collider_mass_information(
...
+                        # Not viewport-drawn, but the viewer still draws these under show_collision /
+                        # show_static. Mutating the shared cache entry is safe: both caches key on the
+                        # prim path, so every consumer resolves the same values.
                         mesh = _get_mesh_cached(prim)
+                        if material_props.get("texture") is not None:
+                            mesh.texture = material_props["texture"]
+                        if material_props.get("roughness") is not None:
+                            mesh.roughness = material_props["roughness"]
+                        if material_props.get("metallic") is not None:
+                            mesh.metallic = material_props["metallic"]
```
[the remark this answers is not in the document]
camevor: Fixed in 7ee1b7a60

newton/_src/usd/utils.py
camevor: Yes, migrated them to keyword-only.

nvtw (approved):
LGTM

PR #3956, opened 2026-08-18
Add asRoBallet control example

### LINKED ISSUE #3955: [REQ] Add an asRoBallet robot control example
## Description

[asRoBallet](https://github.com/asRoBallet/asRoBallet_mujoco) is an underactuated humanoid ballbot driven by three omni-wheel actuators. The system was introduced in an [RSS 2026 paper](https://arxiv.org/abs/2604.24916), and its original MuJoCo repository provides the canonical MJCF model, mesh geometry, contact configuration, and control tasks.

I propose adding an asRoBallet example to Newton with:

- policy control as the default mode;
- an LQR controller as an alternative;
- keyboard-controlled planar velocity and yaw-rate commands;
- an independently floating ball as the default topology;
- an optional virtual ball joint matching the original MJCF topology;
- a shared simulation environment for both controllers.

The robot-specific control implementations would remain local to the example. No public Newton APIs would be added or changed. The required MJCF models, meshes, and ONNX policies would be proposed separately to `newton-assets`.

## Motivation / Use Case

asRoBallet provides a compact example of a contact-rich, underactuated robot whose body is supported and driven through wheel-sphere-ground interactions.

A Newton example would demonstrate how existing Newton capabilities can be combined for this type of system, including MJCF import, MuJoCo solver integration, independently floating rigid bodies, collision filtering, ONNX policy execution, model-based control, and interactive viewer input.

Providing policy and LQR controllers in the same environment also shows how learned and model-based controllers can share one Newton simulation setup without adding robot-specific control APIs to the library.

## Alternatives Considered

The Newton implementation could remain exclusively in the external asRoBallet repositories. That remains useful for project-specific development, but it would not provide an upstream-tested and documented Newton example.

A procedurally generated approximation was also considered. It would not preserve the robot's detailed geometry, omni-wheel contact structure, or policy observation and action interfaces.

han-xudong:
## Description

Closes #3955.

Add an asRoBallet robot example with policy and LQR control modes.

The example:

- uses policy control by default and supports `--controller lqr`;
- accepts interactive planar velocity and yaw-rate commands;
- switches from velocity tracking to station keeping after the robot slows;
- models the ball as an independent floating body by default;
- provides `--virtual-ball-joint` for the original spherical-joint topology;
- applies the model's intended collision filtering;
- shares the same model, simulation setup, and command interface between both controllers.

Companion assets pull request: https://github.com/newton-physics/newton-assets/pull/50 (merged as `f8fb7abcbeba2318814a74f3eeb02780ad7925d6`).

No public Newton APIs are added or changed.

## Checklist

## Test plan

The registered example tests run both the default policy controller and the LQR controller for 500 simulation frames. They pass locally against the merged official assets revision.

```bash
uv run --extra dev -m newton.tests -k test_robot.example_robot_asroballet

uv run --extra examples -m newton.examples robot_asroballet \
  --viewer null --num-frames 100 --virtual-ball-joint

uv run --extra examples -m newton.examples robot_asroballet \
  --viewer null --num-frames 100 --virtual-ball-joint --controller lqr

uvx pre-commit run -a

uvx --from towncrier==25.8.0 towncrier build --draft \
  --version 0.0.0 --date 2026-08-18
```

## Summary by CodeRabbit

* **New Features**
  * Added the asRoBallet robot simulation example.
  * Supports keyboard or fixed-velocity commands, LQR control, and ONNX-based learned policies.
  * Added controller selection, ball configuration, planar velocity, and yaw-rate options.
  * Includes interactive controls, motion tracking, station keeping, and automatic setup and validation.

* **Documentation**
  * Added setup, preview, source, and execution details to the examples documentation.
  * Expanded contribution guidance for local development and workflows.
  * Added a changelog entry for the new example.

* **Tests**
  * Added automated coverage for controller configurations and interactive controls.


newton/examples/robot/example_robot_asroballet.py
```diff
@@ -0,0 +1,1355 @@
...
+    yaw_torque = -yaw_gain[0] * (angular_velocity[2] - YAW_COMMAND_SCALE * desired_velocity[2])
+
+    virtual_torque = wp.vec3(roll_torque, pitch_torque, yaw_torque)
+    for wheel in range(3):
+        wheel_control = (
+            ball_to_control[wheel, 0] * virtual_torque[0]
+            + ball_to_control[wheel, 1] * virtual_torque[1]
+            + ball_to_control[wheel, 2] * virtual_torque[2]
+        )
+        control[wheel] = wp.clamp(wheel_control, -control_limit, control_limit)
```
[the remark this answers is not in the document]
han-xudong: Resolved in 0870cea5. Both controllers now resolve wheel control indices from `mujoco:actuator_trnid` and pass those indices into their Warp kernels, so actuator ordering is no longer assumed.

newton/examples/robot/example_robot_asroballet.py
```diff
@@ -0,0 +1,1372 @@
...
+    lateral = float(viewer.is_key_down("j")) - float(viewer.is_key_down("l"))
+    yaw = float(viewer.is_key_down("u")) - float(viewer.is_key_down("o"))
+    return np.array(
+        [
+            KEYBOARD_LINEAR_SPEED * forward,
+            KEYBOARD_LINEAR_SPEED * lateral,
+            KEYBOARD_YAW_RATE * yaw,
+        ],
+        dtype=np.float64,
+    )
```
eric-heiden, line 76: The keyboard bindings already work, but they are only discoverable from the source comments. Please combine keyboard and GUI input in one helper so both paths share the same limits.

```suggestion
    )


def interactive_velocity_command(viewer, gui_command: np.ndarray | None = None) -> np.ndarray:
    """Combine held keyboard and GUI controls into one body-relative command."""
    if gui_command is None:
        gui_command = np.zeros(3, dtype=np.float64)
    command = keyboard_velocity_command(viewer) + gui_command
    limits = np.array([KEYBOARD_LINEAR_SPEED, KEYBOARD_LINEAR_SPEED, KEYBOARD_YAW_RATE])
    return np.clip(command, -limits, limits)
```

newton/examples/robot/example_robot_asroballet.py
```diff
@@ -0,0 +1,1372 @@
...
+                self.state_0.clear_forces()
+                self.viewer.apply_forces(self.state_0)
+                self.solver.step(self.state_0, self.state_1, self.control, None, self.sim_dt)
+                self.state_0, self.state_1 = self.state_1, self.state_0
+
+    def update_control(self):
+        raise NotImplementedError
+
+    def after_simulate(self):
+        pass
```
eric-heiden, line 626: Could we surface the existing bindings in the viewer and add mouse-accessible controls? Using `is_item_active()` makes these press-and-hold buttons match the keyboard semantics; the panel also shows the applied command and explains when fixed CLI mode disables interaction.

```suggestion
        pass

    def gui(self, ui):
        """Draw press-and-hold motion controls and keyboard help."""
        ui.text("Motion Controls")
        ui.text("Hold a button or its keyboard shortcut.")
        ui.text("Forward/back: I / K")
        ui.text("Left/right: J / L")
        ui.text("Turn left/right: U / O")
        ui.separator()

        interactive = getattr(self, "keyboard_control", False)
        ui.begin_disabled(not interactive)

        ui.button("Forward [I]##asroballet_forward")
        forward = ui.is_item_active()
        ui.same_line()
        ui.button("Backward [K]##asroballet_backward")
        backward = ui.is_item_active()

        ui.button("Left [J]##asroballet_left")
        left = ui.is_item_active()
        ui.same_line()
        ui.button("Right [L]##asroballet_right")
        right = ui.is_item_active()

        ui.button("Turn Left [U]##asroballet_turn_left")
        turn_left = ui.is_item_active()
        ui.same_line()
        ui.button("Turn Right [O]##asroballet_turn_right")
        turn_right = ui.is_item_active()

        ui.end_disabled()

        self.gui_velocity_command = np.zeros(3, dtype=np.float64)
        if interactive:
            self.gui_velocity_command[:] = [
                KEYBOARD_LINEAR_SPEED * (forward - backward),
                KEYBOARD_LINEAR_SPEED * (left - right),
                KEYBOARD_YAW_RATE * (turn_left - turn_right),
            ]
        else:
            ui.text("Fixed CLI command active; interactive controls are disabled.")

        command = getattr(self, "velocity_command", np.zeros(3))
        ui.separator()
        ui.text(f"Command: vx={command[0]:+.2f}, vy={command[1]:+.2f} m/s")
        ui.text(f"Yaw rate: {command[2]:+.2f} rad/s")
```

newton/examples/robot/example_robot_asroballet.py
```diff
@@ -0,0 +1,1372 @@
...
+            sphere_radius=self.sphere_radius,
+            control_limit=self.control_limit,
+            device=self.device,
+        )
+
+    def update_command(self):
+        if not self.keyboard_control:
+            return
+
+        target_command = keyboard_velocity_command(self.viewer)
```
eric-heiden: Feed the held GUI state through the same acceleration-limited LQR command path.

```suggestion
        target_command = interactive_velocity_command(
            self.viewer,
            getattr(self, "gui_velocity_command", None),
        )
```

newton/examples/robot/example_robot_asroballet.py
```diff
@@ -0,0 +1,1372 @@
...
+            [wp.vec3(*self.velocity_command)],
+            dtype=wp.vec3,
+            device=self.device,
+        )
+
+    def update_command(self):
+        if not self.keyboard_control:
+            return
+
+        self.velocity_command = keyboard_velocity_command(self.viewer)
```
eric-heiden: Feed the held GUI state through the policy command path as well.

```suggestion
        self.velocity_command = interactive_velocity_command(
            self.viewer,
            getattr(self, "gui_velocity_command", None),
        )
```

newton/examples/robot/example_robot_asroballet.py
```diff
@@ -0,0 +1,1372 @@
...
+        self._controller = controller_type(viewer, args)
+
+    def __getattr__(self, name):
+        return getattr(self._controller, name)
+
+    def step(self):
+        self._controller.step()
+
+    def render(self):
+        self._controller.render()
```
eric-heiden, line 1401: Delegate the example GUI hook so Newton registers the controller panel under “Example Options.”

```suggestion
        self._controller.render()

    def gui(self, ui):
        self._controller.gui(ui)
```

eric-heiden (commented):
The keyboard controls already work, but the bindings are not discoverable in the viewer. I prototyped and locally validated an on-screen press-and-hold panel with visible shortcuts and live command feedback. The inline suggestions below are intended to be applied together.

newton/examples/robot/example_robot_asroballet.py
han-xudong: Implemented in 151068de. Keyboard and held GUI input now share one command helper and the same linear/yaw limits; I also added a focused regression test for combining and clipping both inputs.

newton/examples/robot/example_robot_asroballet.py
han-xudong: Implemented in 151068de. The Example Options panel now exposes press-and-hold motion buttons, visible keyboard shortcuts, fixed-command guidance, and live applied-command feedback.

newton/examples/robot/example_robot_asroballet.py
han-xudong: Implemented in 151068de. The LQR path now consumes the combined keyboard/GUI target while preserving its existing acceleration-limited command ramp.

newton/examples/robot/example_robot_asroballet.py
han-xudong: Implemented in 151068de. The policy path now reads the same combined and clipped interactive command before updating policy mode and observations.

newton/examples/robot/example_robot_asroballet.py
han-xudong: Implemented in 151068de. The top-level Example now delegates gui() to the selected controller, so Newton registers the panel under Example Options.

eric-heiden:
Nice addition, the asRoBallet example is well structured and the LQR and policy controllers are cleanly separated behind a single flag.

Thanks for addressing my comments! Two things are worth a look before merge: the action history fed back into the policy observation is one control step older than it should be, and the new LQR test variant can fail at import time in environments without warp-nn.

newton/examples/robot/example_robot_asroballet.py
```diff
@@ -0,0 +1,1431 @@
...
+    wheel_ctrl_indices: wp.array[int],
+    control: wp.array[float],
+    current_action: wp.array2d[float],
+    last_action: wp.array2d[float],
+):
+    wheel = wp.tid()
+    value = wp.clamp(action[0, wheel], -1.0, 1.0)
+    last_action[0, wheel] = current_action[0, wheel]
+    current_action[0, wheel] = value
+    control[wheel_ctrl_indices[wheel]] = value
```
eric-heiden, line 1072: 🟠 The policy receives an action that is two control steps old.

`last_action` is overwritten with the previous value of `current_action` before `current_action` is updated, so once control step `k` finishes the buffers hold the action from step `k` in `current_action` and the action from step `k-1` in `last_action`. `update_control()` builds the observation from `self.last_action` before it queries the policy, so on the next step the action history slots get the action from step `k-1` instead of the one that was just applied.

That leaves `observation[0, 13..15]` in `_build_policy_observation_kernel` and `observation[0, 14..16]` in `_build_station_policy_observation_kernel` one control step (20 ms at the 100 Hz control rate set by `sim_substeps=10` and `control_decimation=5`) staler than the training convention, so both ONNX policies get an action history feature that does not match what they were trained on on every default run of the example.

`current_action` is not read anywhere else in the file, so the second buffer only adds the delay. Writing the applied value straight into the history buffer matches the sibling example, where `example_robot_policy.py` copies the action into its previous action buffer right after the policy call:

```suggestion
    wheel = wp.tid()
    value = wp.clamp(action[0, wheel], -1.0, 1.0)
    last_action[0, wheel] = value
    control[wheel_ctrl_indices[wheel]] = value
```

With that change, `current_action` can also be dropped from the kernel signature and the launch call. If you would rather keep both buffers, pass `self.current_action` to the observation launches instead. Either way it would match the helper docstring, which says the applied actions are retained for the next observation.

newton/tests/test_examples.py
```diff
@@ -667,6 +683,21 @@ class TestRobotExamples(unittest.TestCase):
...
+    use_viewer=True,
+)
+add_example_test(
+    TestRobotExamples,
+    name="robot.example_robot_asroballet",
+    devices=cuda_test_devices,
+    test_options={"controller": "lqr", "num-frames": 500},
+    use_viewer=True,
+    test_suffix="LQR",
+)
```
eric-heiden, line 684: 🔵 This variant should be registered with `onnx_required` too, or the import should be made lazy.

`newton/examples/robot/example_robot_asroballet.py` does `from warp_nn.runtime import OnnxRuntime` at module scope, so importing the example requires warp-nn even when the LQR controller is selected, which otherwise only needs SciPy and Warp. The policy variant just above sets `"onnx_required": True` and skips cleanly when the dependency is missing, while this one would fail with `ModuleNotFoundError: No module named 'warp_nn'`. The same applies to anyone who installs a narrower set of extras than `newton[examples]` and runs the example with `--controller lqr`.

The three other examples with a module level `warp_nn` import (`example_robot_anymal_c_walk.py`, `example_robot_policy.py`, `example_mpm_anymal.py`) are all registered with `onnx_required: True`, so the smallest fix is to match them:

```suggestion
add_example_test(
    TestRobotExamples,
    name="robot.example_robot_asroballet",
    devices=cuda_test_devices,
    test_options={"controller": "lqr", "num-frames": 500, "onnx_required": True},
    use_viewer=True,
    test_suffix="LQR",
)
```

If you would rather have the LQR path genuinely run without warp-nn, the lazy import idiom already used by `_require_onnx()` in `newton/examples/robot/onnx_policy_utils.py` and `_require_warp_nn_runtime()` in `newton/_src/actuators/utils.py` fits well here.

Every documented CI invocation installs the extras, so this is latent today rather than a failing test.

newton/examples/robot/example_robot_asroballet.py
```diff
@@ -0,0 +1,1431 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
+
+###########################################################################
+# Example Robot asRoBallet
```
eric-heiden, line 5: Feel free to add a citation to your RSS paper here so users can find it.

newton/examples/robot/example_robot_asroballet.py
han-xudong: Fixed in 32c5a59f. The clipped action is now written directly to `last_action` for the next observation, and the redundant `current_action` buffer and data flow have been removed. I also added a regression test covering the retained action and resolved actuator mapping.

newton/tests/test_examples.py
han-xudong: Fixed in 32c5a59f using the lazy-import option. `warp_nn.runtime` is now imported only when a policy runtime is created, so the LQR-capable module imports without Warp-NN. A subprocess regression test explicitly blocks `warp_nn` to verify this path.

newton/examples/robot/example_robot_asroballet.py
han-xudong: Added the RSS 2026 paper link to the example header in 32c5a59f so users can find the project reference directly from the source.

newton/tests/test_examples.py
```diff
@@ -623,6 +624,63 @@ class TestRobotExamples(unittest.TestCase):
     pass
 
 
+class TestAsRoBalletExample(unittest.TestCase):
```
eric-heiden: We should not have an extra class just for this example here.
Please fold `test_policy_action_history` into the test function from `example_robot_asroballet.py` like we do for the other examples. Generally, that test function should probably already cover the behavior from the test functions you declare here, so you might be able to just remove them.

han-xudong: Updated in c3d7315e. I removed the dedicated `TestAsRoBalletExample` class and folded the applied-action history check into `_PolicyController.test_final()`, so it now runs through the standard registered example test. The other helper-specific tests were removed; I also revalidated the LQR import manually with `warp_nn` blocked.

eric-heiden (approved):
Thanks for your contribution @han-xudong and for addressing my comments, the example looks great!

PR #3961, opened 2026-08-18
Improve collision performance through better preprocessing

nvtw:
## Description

Improve collision performance by deduplicating convex collision vertices, streamlining mesh work preprocessing, and selecting split GJK/MPR based on the active convex workload. Add targeted regression coverage and a convex collision benchmark.

## Checklist

- [ ] For user-facing changes, a fragment has been added by following the [changelog fragment instructions](https://github.com/newton-physics/newton/blob/main/changelog/README.md)

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_narrow_phase
uv run --extra dev -m newton.tests -k test_sdf_contact
```

## Summary by CodeRabbit

- **New Features**
  - Added collision benchmarks for irregular convex meshes across varied sizes and simulation scales.
  - Added automatic duplicate-vertex cleanup for convex collision meshes while preserving connectivity.
  - Added optional split GJK/MPR processing for large convex-collision workloads.

- **Bug Fixes**
  - Improved handling of varying mesh and world workloads.
  - Prevented stale inactive entries from affecting contact calculations.
  - Improved deferred SDF generation and cache separation for different mesh topologies.

- **Tests**
  - Added regression coverage for mesh cleanup, SDF behavior, workload scaling, and collision graph capture.


asv/benchmarks/simulation/bench_contacts.py
```diff
@@ -145,6 +248,54 @@ def time_simulate(self):
         wp.synchronize_device()
 
 
+class FastConvexCollision:
+    """Benchmark lean-hull and mixed-type convex collision workloads."""
+
+    params = (CONVEX_COLLISION_CASES,)
+    param_names = ["case"]
```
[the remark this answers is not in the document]
nvtw: Done

asv/benchmarks/simulation/bench_contacts.py
```diff
@@ -145,6 +248,54 @@ def time_simulate(self):
         wp.synchronize_device()
 
 
+class FastConvexCollision:
+    """Benchmark lean-hull and mixed-type convex collision workloads."""
```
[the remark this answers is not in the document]
nvtw: Done

newton/_src/sim/collide.py
```diff
@@ -1471,6 +1472,11 @@ def __init__(
             candidate_pair_work_estimate = min(self.shape_pairs_max, _compute_per_world_shape_pairs_max(model))
             if self.broad_phase_mode == "explicit":
                 candidate_pair_work_estimate = self.shape_pairs_max
+            split_gjk_mpr = (
```
mzamoramora-nvidia, line 1526: Could we tune this using the routed convex work and account for the lean path? The new selector helps smaller scenes, but some larger cases below the cutoff are slower than the base. Across three interleaved runs on an RTX PRO 6000, the 191-world lean case took 0.184 ms on this branch and 0.140 ms on the base. The mixed path was also 9% slower below the cutoff. At 192 worlds both versions split and match. Adding cases below the cutoff would help us tune this range.

asv/benchmarks/simulation/bench_contacts.py
```diff
@@ -43,6 +66,87 @@ def _import_example_class(module_names: list[str]):
     raise SkipNotImplemented
 
 
+def _make_irregular_rock(vertex_count: int, seed: int) -> newton.Mesh:
```
mzamoramora-nvidia: Could one case use triangle-local duplicate vertices? The generated rocks are already unique, so this benchmark does not exercise deduplication. With duplicate-heavy rocks at 192 worlds, three local runs improved from 0.218 ms to 0.140 ms. Candidates and contacts were unchanged. This would make the benchmark show a real benefit from the PR.

asv/benchmarks/simulation/bench_contacts.py
```diff
@@ -145,6 +249,54 @@ def time_simulate(self):
         wp.synchronize_device()
 
 
+class FastConvexCollision:
```
mzamoramora-nvidia, line 265: Could we add a focused mesh preprocessing benchmark? This convex workload does not run either changed mesh path. With 65,536 pairs, the mesh-plane helper was about 18% faster and the mesh-mesh helper was about 18% faster on the same GPU.

mzamoramora-nvidia (commented):
Thanks for the careful correctness work here. The GPU measurements show real wins from vertex deduplication and the mesh preprocessing changes. The new selector also improves several smaller convex workloads.

The main improvement I suggest is tuning the selector threshold. Lean and full-support workloads have different measured crossovers. Some workloads below the cutoff are slower than the base. Across three interleaved runs, the lean hull workload was 31% slower and the mixed workload was 9% slower just below 6,144 shapes.

The added benchmark begins at that cutoff and uses unique vertices, so it misses the slower sub-threshold cases and the main improvements. I left focused suggestions for workloads that expose each change. It would also be good to add the Towncrier fragment described in the contribution guidelines.

mzamoramora-nvidia (commented):
Thanks for the updates. I left three focused follow-ups from a fresh local pass.

asv/benchmarks/simulation/bench_contacts.py
```diff
@@ -145,6 +266,146 @@ def time_simulate(self):
...
+            block_counts,
+            total_weight,
+            shape_heightfield_index,
+            heightfield_data,
+        )
+
+        def preprocess():
+            total_weight.zero_()
+            if pair_type == "mesh_plane":
+                compute_mesh_plane_block_offsets_scan(
```
mzamoramora-nvidia: Could we make this benchmark work with both helper signatures? ASV uses this new benchmark code while testing both the base commit and the PR commit. I reproduced the base-side setup locally, where it fails because these helpers expect `weight_prefix_sums`. The new arguments `total_vert_count`, `total_edge_count`, and `total_num_threads` do not exist there. The current CI run lost its runner before it reached this result, so it did not confirm the problem. Could the benchmark detect which signature is installed and call it with the matching buffers? That would let ASV measure both revisions on a healthy run.

newton/_src/sim/collide.py
```diff
@@ -747,6 +747,47 @@ def _resolve_shape_pairs_max(model: Model, override: int | None) -> int:
 
 
 BROAD_PHASE_MODES = ("nxn", "sap", "explicit")
+_SPLIT_GJK_MPR_LEAN_PAIR_COUNT_THRESHOLD = 32_768
```
mzamoramora-nvidia: Could we add a lean crossover case and revisit this cutoff? On an RTX PRO 6000, the head was 3.5–5.8% slower than base at 56–66 hull worlds because it stayed fused. At 67 worlds it switched to split and matched base. The current ASV hull cases are all above this cutoff, so they miss this range.

newton/_src/geometry/sdf_contact.py
```diff
@@ -1019,50 +1013,53 @@ def compute_mesh_mesh_block_offsets_scan(
     target_blocks: int,
     block_offsets: wp.array,
     block_counts: wp.array,
-    weight_prefix_sums: wp.array,
+    total_edge_count: wp.array,
```
mzamoramora-nvidia: Small typing fix: could this be `wp.array[wp.int32]`? That is the Warp array style required by this repo.

mzamoramora-nvidia: We can also do this on a follow up.

mzamoramora-nvidia (approved):

PR #3964, opened 2026-08-18
[Kamino] Implement joint friction (Coulomb)

### LINKED ISSUE #3935: [REQ] Support joint friction in Kamino
## Description

Newton models expose per-DOF `Model.joint_friction`, including values imported from common robot formats, but Kamino currently does not consume this property. I would like to discuss the friction semantics and integration point that should be supported by Kamino.

Desired behavior:

- Apply friction consistently across PADMM and DVI and supported storage modes.
- Oppose motion without injecting energy.
- Support CPU, CUDA, batched worlds, runtime property updates, and CUDA graph capture.
- Define predictable behavior around zero velocity and for invalid authored or runtime values.

## Motivation / Use Case

Ignoring imported joint friction can leave constrained and geared mechanisms with unrealistically persistent motion or force applications to inject solver-specific efforts. Supporting the existing model property would improve consistency across Newton solvers.

## Questions for Maintainers

- Which initial model is preferred: regularized sliding friction, a stiction-capable formulation, or another existing Kamino mechanism?
- Should the near-zero regularization threshold be fixed internally or exposed in `SolverKamino.Config`?
- At what stage should friction modify the effective joint effort so PADMM and DVI share the same semantics?
- How should invalid runtime device updates be handled without introducing host synchronization?
- Is joint friction already covered by planned actuator or passive-force work?

## Alternatives Considered

- Use joint damping, which is velocity-proportional and does not model bounded Coulomb-like resistance.
- Apply external joint efforts in application code, which duplicates solver logic across worlds and graph replays.
- Implement full static friction immediately, which requires a larger complementarity and stiction design.

rubengrandia:
## Description

Adds Coulomb (dry) joint friction to `SolverKamino` by enforcing `Model.joint_friction`.

For each frictional joint DoF, the solver adds a bounded multiplier row with the DoF-direction Jacobian:

```
b_j = dt · f_j                        (f_j = Model.joint_friction[dof], N or N·m)
Λ_f ∈ −b_j · Sign(dq⁺)   ⟺   Λ_f = Π_[−b_j, b_j](Λ_f − ρ·dq⁺)   ∀ρ > 0
```

This maximum-dissipation formulation provides exact stiction: a joint remains stationary under loads below the friction limit, without the creep associated with damping approximations. Once the limit is exceeded, the friction torque saturates and a freely spinning joint slows linearly rather than exponentially. Unlike contact friction, the bound is independent of normal force, so the box projection fully defines the friction law (no De Saxce correction needed).

The friction row has no `inv_m_j` regularization. When a joint also has a dynamic row, their coupling includes both body inertia and armature inertia; adding another armature term to the friction row would double-count it. The same row formulation also applies to joints without a dynamic row. The test file verifies this by testing all variants agains their analytical solution.

## Implementation

- Adds a generic bounded-multiplier constraint group with per-row lower and upper bounds. The constraint layout is now `[ joint dynamic | joint kinematic | bounded | limits | contacts ]`. The more general bounded formulation will be re-used in the joint effort limit constraint.
- Adds box projection, warm-starting, and natural-map convergence residuals for bounded rows.
- Renames the former `unilaterals` offsets and counts to `inequalities`, reflecting that projected rows now include bounded, limit, and contact constraints.
- Exposes the recovered friction torque through `data.joints.lambda_f_j`.
- Supports joint friction with both dense and spars PADMM solver paths. DVI support will follow in a separate PR.

## Checklist

- [ ] A Towncrier changelog fragment will be added before merging to `main`.

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_solver_kamino_joint_friction
uv run --extra dev python newton/_src/solvers/kamino/tests/test_solver_kamino_joint_friction.py
```

## Summary by CodeRabbit

* **New Features**
  * Added Coulomb joint-friction support to SolverKamino, including dense and sparse simulation workflows.
  * Added friction limits, constraint handling, warm starts, wrenches, and solver metrics.
  * Added bounded-constraint support for joint friction and related inequality processing.
  * Added validation and warnings for unsupported configurations, missing capacity, and free-joint friction.

* **Documentation**
  * Updated SolverKamino documentation and support tables to describe joint friction and bounded constraints.

* **Tests**
  * Added coverage for friction behavior, limits, multiworld setups, solver modes, state updates, and validation errors.


newton/_src/solvers/kamino/_src/core/builder.py
```diff
@@ -193,6 +195,16 @@ def num_kinematic_joint_cts(self) -> int:
...
         return self._num_joint_kinematic_cts
 
+    @property
+    def num_bounded_cts(self) -> int:
+        """Returns the number of bounded-multiplier constraint rows contained in the model."""
+        return self._num_bounded_cts
+
+    @property
+    def num_friction_cts(self) -> int:
+        """Returns the number of Coulomb-friction constraint rows contained in the model."""
```
Guirec-Maloisel, line 205: To avoid any confusion, we should make sure that all docstrings referring to the new joint Coulomb friction constraints always mention _joint_ friction, and ideally also _Coulomb_ friction to be precise (because viscous friction is applied separately through the damping term in joint dynamics). Picking one occurence for this comment, but applies at multiple places.

In this specific place, I'd also rename the property (and corresponding internal member) to `num_joint_friction_cts` (especially as there's already a `num_joint_kinematic_cts` and `num_joint_dynamic_cts`)

newton/_src/solvers/kamino/_src/core/bodies.py
```diff
@@ -262,7 +262,13 @@ class RigidBodiesData:
 
     w_j_i: wp.array[wp.spatial_vectorf] | None = None
     """
-    Joint constraint wrench applied to each body (in world coordinates).
+    Bilateral joint constraint wrench applied to each body (in world coordinates) [N, N·m].
+    Shape of ``(num_bodies,)``.
+    """
+
+    w_f_i: wp.array[wp.spatial_vectorf] | None = None
```
Guirec-Maloisel, line 269: Maybe to keep for a separate pass for other similar variables, but `w_f_i` doesn't really scream what it is to me. We could be more explicit in naming.

newton/_src/solvers/kamino/_src/core/builder.py
```diff
@@ -193,6 +195,16 @@ def num_kinematic_joint_cts(self) -> int:
         """Returns the number of kinematic joint constraints contained in the model."""
         return self._num_joint_kinematic_cts
 
+    @property
+    def num_bounded_cts(self) -> int:
+        """Returns the number of bounded-multiplier constraint rows contained in the model."""
```
Guirec-Maloisel, line 200: We're missing a place where the whole set of constraints and their categories is described. Maybe model.py?
Here it's not so clear what bounded cts are, or the fact that they include (and are currently limited to) the friction cts below.

newton/_src/solvers/kamino/_src/core/builder.py
```diff
@@ -485,6 +500,8 @@ def add_joint_descriptor(self, joint: JointDescriptor, world_index: int = 0) ->
         # Check if the descriptor is valid
         if not isinstance(joint, JointDescriptor):
             raise TypeError(f"Invalid joint descriptor type: {type(joint)}. Must be `JointDescriptor`.")
+        if joint.dof_type == JointDoFType.FREE and any(value > 0.0 for value in joint.f_j):
```
Guirec-Maloisel, line 503: Makes sense to ignore friction for free joints, but I was wondering how the new code behaves for spherical/gimbal joints?

newton/_src/solvers/kamino/_src/dynamics/dual.py
```diff
@@ -873,10 +933,11 @@ def _build_dual_preconditioner_all_constraints(
     # Retrieve the vector index offset of the world
     vio = problem_vio[wid]
 
-    # Retrieve the number of active joint and limit constraints of the world
```
Guirec-Maloisel, line 876: This comment should probably be updated instead of removed

newton/_src/solvers/kamino/_src/kinematics/constraints.py
```diff
@@ -200,21 +208,25 @@ def _make_empty_model_contacts_info():
...
-        njc + maxnl + maxnc for njc, maxnl, maxnc in zip(world_njc, world_maxnlc, world_maxncc, strict=False)
+        njc + nbc + maxnl + maxnc
+        for njc, nbc, maxnl, maxnc in zip(world_njc, world_nbc, world_maxnlc, world_maxncc, strict=False)
     ]
     model.size.sum_of_max_total_cts = sum(world_maxncts)
     model.size.max_of_max_total_cts = max(world_maxncts)
 
-    # Compute the entity index offsets for limits, contacts and unilaterals
-    # NOTE: unilaterals is simply the concatenation of limits and contacts
+    # Compute the entity index offsets for limits, contacts, and inequalities
```
Guirec-Maloisel, line 217: Should this just be "for inequalities"? since limits and contacts are inequality constraints

newton/_src/solvers/kamino/_src/solvers/common.py
```diff
@@ -90,14 +96,17 @@ def warmstart_joint_constraints(
...
     dt = model_time_dt[wid_j]
     joint_dyn_cts_start = joint_dynamic_cts_offset_joint_cts[jid]
     joint_kin_cts_start = joint_kinematic_cts_offset_joint_cts[jid]
+    joint_friction_cts_start = joint_friction_cts_offset[jid]
+    joint_dofs_start = joint_dofs_offset[jid]
     dyn_cts_row_start_j = joint_dynamic_cts_offset_total_cts[jid]
     kin_cts_row_start_j = joint_kinematic_cts_offset_total_cts[jid]
+    friction_cts_row_start_j = joint_friction_cts_offset_total_cts[jid]
 
-    # Convert cached forces to preconditioned impulses. Joint constraints do
```
Guirec-Maloisel, line 99: This comment also seems unduly shortened

newton/_src/solvers/kamino/tests/test_solver_kamino_joint_friction.py
```diff
@@ -0,0 +1,428 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
```
Guirec-Maloisel, line 2: Is there a reason this test file is not added to newton tests but kept internal?

newton/_src/solvers/kamino/_src/core/model.py
```diff
@@ -527,19 +568,18 @@ def data(
...
+        njfccts = self.size.sum_of_num_friction_cts
 
         # Construct the model data on the specified device
         with wp.ScopedDevice(device=device):
             # Create a new model data info with the total constraint
-            # counts initialized to the joint constraints count
+            # counts initialized to the joint + bounded constraints count
             info = DataKaminoInfo(
-                num_total_cts=wp.clone(self.info.num_joint_cts),
-                num_limits=wp.zeros(shape=nw, dtype=wp.int32) if unilateral_cts else None,
```
Guirec-Maloisel, line 537: Is it safe to remove all these other arguments to the constructor of DataKaminoInfo() ?

newton/_src/solvers/kamino/_src/solvers/padmm/types.py
```diff
@@ -293,9 +295,12 @@ class PADMMStatus:
             Computed using the L-inf norm as `r_primal := || x - y ||_inf`.
         r_d: The total dual residual.
             Computed using the L-inf norm as `r_dual := || eta * (x - x_p) + rho * (y - y_p) ||_inf`.
-        r_c: The total complementarity residual.
-            Computed using the L-inf norm as `r_compl := || [x_k.T @ z_k] ||_inf`,
-            with `k` indexing each unilateral constraint set, i.e. 1D limits and 3D contacts.
+        r_c: The generalized inequality optimality residual used by the convergence test.
```
Guirec-Maloisel, line 298: Note sure we need all this fuss about "generalized inequality optimality residual". To be fully correct we could call it a "generalized" complementarity residual, but I'd be fine to keep calling this just the complementarity residual (but have the note about what it means for bounded multiplier constraints). Wouldn't this be a complementarity between the joint velocity, and the friction force minus the bound, or is it really a different quantity?

Guirec-Maloisel (commented):
More general comments about the constraint categories and naming.

As a first comment, it doesn't feel so intuitive anymore that it is now bilateral vs inequality, maybe we should have it equality vs inequality?

I also said it in another comment, but we should detail the categories in a longer docstring somewhere in the model or solver.
We'd have equality (=bilateral) comprising joint kinematics and dynamics; and inequality comprising unilateral (= joint position limits and contacts) and the new bounded constraints.

Regarding the bounded constraints, maybe "bounded effort constraints" could be more intuitive than bounded multiplier? Then say that it corresponds mathematically to bounded normal cone constraints, and applies for joint Coulomb friction, and (in the future) to joint effort limits.

I'd be explicit also about limits (joint position vs joint effort), I wouldn't rename the Limits container etc but talk about joint position limits in the docstrings whenever possible.

Let me know what you think!

newton/_src/solvers/kamino/_src/core/builder.py
rubengrandia: Agree. I did the local rename you suggested and made all docs reflect that it is Coulomb joint friction.

newton/_src/solvers/kamino/_src/core/builder.py
rubengrandia: Added documentation to ModelKaminoInfo

newton/_src/solvers/kamino/_src/core/builder.py
rubengrandia: For spherical joints it will apply the friction per-coordinate in local joint angular velocity space. This will results in a box shaped friction though. Extending to be a isotropic 3D friction would be an additional effort.

For Gimbal joints it applies it in the consecutive rotation (euler) coordinates. This is a bit odd, but consistent with how the D6 is defined.

newton/_src/solvers/kamino/_src/solvers/common.py
rubengrandia: Added back an updated comment

newton/_src/solvers/kamino/tests/test_solver_kamino_joint_friction.py
rubengrandia: Moved it.

newton/_src/solvers/kamino/_src/core/model.py
rubengrandia: yes, the unilateral_cts flag was never set to true on any call site. So this was only assigning None, which is already the default.

newton/_src/solvers/kamino/_src/kinematics/constraints.py
rubengrandia: fixed

newton/_src/solvers/kamino/_src/solvers/padmm/types.py
rubengrandia: Reverted to just say complementarity residual with a note on how it works for bounded multipliers. With the updated way this is computed, this is again a true complementarity residual.

rubengrandia:
Due to broader renaming I am dropping the PR split and continue on just the 1 PR. https://github.com/newton-physics/newton/pull/3966

PR #3969, opened 2026-08-18
Reject cross-articulation joints

### LINKED ISSUE #3968: [BUG] non-rooted articulations cause incorrect `eval_fk`, `eval_jacobian`, `eval_mass_matrix`
### Bug Description

I have a case where I have placed one articulation at the end of a second articulation. While this case is contrived, I think it represents a case where a user mounts an arm onto a mobile base, leaving them as two separate articulations, which seems sensible.

The robot setup looks like this. In the "standalone" case, it is a 1DOF robot. In the "mounted" case, it is a 1DOF robot mounted on another 1DOF robot:

```
            standalone      mounted
            ----------      -------
            j0    mass      j0         j1        mass
            o----------*    o----------o----------*
            |<- 1 m -->|    |<- 1 m -->|<- 1 m -->|
                            \_"base"_/\_"pendulum"_/
```

## Jacobian (and by extension, the mass matrix) are inconsistent:
To compute the Jacobian, `eval_jacobian`
1. walks "up" the kinematic tree, all the way up to the root link (newton/_src/sim/articulation.py:1155).
2. calculates an index of the jacobian to write: (newton/_src/sim/articulation.py:1162)
3. writes it (newton/_src/sim/articulation.py:1167)

However, if the articulation is mounted on _another_ articulation, then the index computed in step 2 is negative. The result is that joints that are **not in the articulation** start to write to columns of the `jacobian[...-1]`, `jacobian[...-2]`, overwriting the columns that were just computed.

In this particular case, looping `while j >= joint_start` should be the fix?

Probably as a consequence, the Mass Matrix of "mounted" articulations are incorrect.

### Jacobian Test:
    def test_jacobian_predicts_the_velocity_fk_computes_for_a_mounted_pendulum(self):
        r"""Verify ``J @ joint_qd`` reproduces the body velocity ``eval_fk`` computes.

        :func:`newton.eval_jacobian` documents that ``J_link @ joint_qd ==
        state.body_qd[link]``, so the two ways of asking "how fast is this body
        moving?" must agree. Only the pendulum joint spins, at 1 rad/s::

            j0         j1        mass
            o----------o--------->*   the COM is 1 m from j1,
            |<- 1 m -->|<- 1 m -->|   so it moves at 1 m/s
            \_"base"_/\_"pendulum"_/

        ``eval_fk`` gets 1 m/s. The Jacobian says 2 m/s — the speed the COM would
        have if it were spinning about j0, 2 m away — because the pendulum's
        ancestor is the base joint, which lies outside the pendulum articulation.
        Its column index comes out negative and lands on the pendulum's own
        column, overwriting it.
        """
        inertia = wp.mat33(np.diag([0.1, 0.1, 0.1]).astype(np.float32))  # about each body's own COM [kg m^2]
        axis = wp.vec3(0.0, 0.0, 1.0)

        for device in get_test_devices():
            with self.subTest(device=str(device)):
                base = builder.add_link(mass=1.0, inertia=inertia, lock_inertia=True)
                builder.add_articulation([builder.add_joint_revolute(parent=-1, child=base, axis=axis)], label="base")
                mass = builder.add_link(
                    mass=2.0,
                    com=wp.vec3(1.0, 0.0, 0.0),  # 1 m out from its own joint
                    inertia=inertia,
                    lock_inertia=True,
                )
                builder.add_articulation(
                    [
                        builder.add_joint_revolute(
                            parent=base,
                            child=mass,
                            axis=axis,
                            parent_xform=wp.transform(p=wp.vec3(1.0, 0.0, 0.0)),  # 1 m out from the base body
                        )
                    ],
                    label="pendulum",
                )
                model = builder.finalize(device=device)

                state = model.state()
                state.joint_qd.assign(np.array([0.0, 1.0], dtype=np.float32))  # spin only the pendulum joint

                pendulum_jacobian = newton.eval_jacobian(model, state).numpy()[1]
                predicted = pendulum_jacobian @ np.array([1.0])
                np.testing.assert_allclose(
                    predicted,
                    state.body_qd.numpy()[1],
                    atol=1e-5,
                    err_msg=(
                        "the Jacobian and forward kinematics disagree about the body's velocity\n"
                        f"  J @ joint_qd:   {np.array2string(predicted, precision=4)}\n"
                        f"  state.body_qd:  {np.array2string(state.body_qd.numpy()[1], precision=4)}"
                    ),
                )

### Jacobian test result:
The Jacobian that we actually get is a single column describing how "J0" affects the speed of the mass at the end of the pendulum (which I assume is caused by writing to index -1). The result is consistent across devices.
```
======================================================================
test_jacobian_predicts_the_velocity_fk_computes_for_a_mounted_pendulum (test_articulation_rootedness.TestArticulationRootedness.test_jacobian_predicts_the_velocity_fk_c
Verify ``J @ joint_qd`` reproduces the body velocity ``eval_fk`` computes.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/jhough/Code/newton/newton/tests/test_articulation_rootedness.py", line 96, in test_jacobian_predicts_the_velocity_fk_computes_for_a_mounted_pendulum
    np.testing.assert_allclose(
  File "/home/jhough/Code/newton/.venv/lib/python3.12/site-packages/numpy/testing/_private/utils.py", line 1780, in assert_allclose
    assert_array_compare(compare, actual, desired, err_msg=str(err_msg),
  File "/home/jhough/Code/newton/.venv/lib/python3.12/site-packages/numpy/testing/_private/utils.py", line 988, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Not equal to tolerance rtol=1e-07, atol=1e-05
the Jacobian and forward kinematics disagree about the body's velocity
  J @ joint_qd:   [0. 2. 0. 0. 0. 1.]
  state.body_qd:  [0. 1. 0. 0. 0. 1.]
Mismatched elements: 1 / 6 (16.7%)
Mismatch at index:
 [1]: 2.0 (ACTUAL), 1.0 (DESIRED)
Max absolute difference among violations: 1.
Max relative difference among violations: 1.
 ACTUAL: array([0., 2., 0., 0., 0., 1.])
 DESIRED: array([0., 1., 0., 0., 0., 1.], dtype=float32)
```

### Mass Matrix Test:
```
    def test_mass_matrix_matches_when_pendulum_is_mounted_on_another_articulation(self):
        r"""Verify a pendulum's mass matrix is the same mounted as it is standalone.

        Both models drawn at q=0, with (o) a revolute joint about z and (*) the
        2 kg mass. The pendulum is identical in both: same mass, same Izz, same
        1 m radius about its own joint::

            standalone      mounted
            ----------      -------
            j0    mass      j0         j1        mass
            o----------*    o----------o----------*
            |<- 1 m -->|    |<- 1 m -->|<- 1 m -->|
                            \_"base"_/\_"pendulum"_/

        A 2 kg mass on a 1 m arm has moment of inertia ``Izz + m*r^2 = 2.1``
        about its own joint axis. That is a property of the pendulum alone, so
        mounting it on another body cannot change it.
        """
        inertia = wp.mat33(np.diag([0.1, 0.1, 0.1]).astype(np.float32))  # about each body's own COM [kg m^2]
        axis = wp.vec3(0.0, 0.0, 1.0)
        expected = 0.1 + 2.0 * 1.0**2  # Izz + m*r^2 about the pendulum's own joint

        for device in get_test_devices():
            with self.subTest(device=str(device)):
                alone = newton.ModelBuilder()
                mass = alone.add_link(mass=2.0, com=wp.vec3(1.0, 0.0, 0.0), inertia=inertia, lock_inertia=True)
                alone.add_articulation([alone.add_joint_revolute(parent=-1, child=mass, axis=axis)], label="pendulum")

                mounted = newton.ModelBuilder()
                base = mounted.add_link(mass=1.0, inertia=inertia, lock_inertia=True)
                mass = mounted.add_link(mass=2.0, com=wp.vec3(1.0, 0.0, 0.0), inertia=inertia, lock_inertia=True)
                mounted.add_articulation([mounted.add_joint_revolute(parent=-1, child=base, axis=axis)], label="base")
                hanging_joint = mounted.add_joint_revolute(
                    parent=base,
                    child=mass,
                    axis=axis,
                    parent_xform=wp.transform(p=wp.vec3(1.0, 0.0, 0.0)),  # 1 m out from the base body
                )
                mounted.add_articulation([hanging_joint], label="pendulum")

                alone = alone.finalize(device=device)
                mounted = mounted.finalize(device=device)
                h_alone = float(newton.eval_mass_matrix(alone, _fk(alone)).numpy()[0, 0, 0])
                h_mounted = float(newton.eval_mass_matrix(mounted, _fk(mounted)).numpy()[1, 0, 0])

                self.assertAlmostEqual(h_alone, expected, places=5, msg="reference is wrong")
                self.assertAlmostEqual(
                    h_mounted,
                    expected,
                    places=5,
                    msg=f"mounted pendulum H={h_mounted}, standalone H={expected}",
                )

```

### Result
On both the `cpu` and `cuda:0` device, the wrong answer for the mass-matrix is obtained for the pendulum if it is mounted on another body. It gets the answer of 8.1, which would be correct if the pendulum joint was J0.
```
======================================================================
test_mass_matrix_matches_when_pendulum_is_mounted_on_another_articulation (test_articulation_rootedness.TestArticulationRootedness.test_mass_matrix_matches_when_pendulum_is_mounted_on_another_articulation) (device='cpu')
Verify a pendulum's mass matrix is the same mounted as it is standalone.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/jhough/Code/newton/newton/tests/test_articulation_rootedness.py", line 222, in test_mass_matrix_matches_when_pendulum_is_mounted_on_another_articulation
    self.assertAlmostEqual(
AssertionError: 8.100000381469727 != 2.1 within 5 places (6.000000381469727 difference) : mounted pendulum H=8.100000381469727, standalone H=2.1
```

## `eval_fk` datarace
In the "mounted" case described above, `eval_fk` seems to have a datarace. The pose of the pendulum depends on the pose of the base articulation. However, these `eval_fk` are performed in parallel **by articulation**. Therefore, there is no guarantee that the `base` articulation fk runs before the `pendulum` articulation.

### Test:
```
    def test_body_poses_match_when_mechanism_split_across_articulations(self):
        r"""Verify body poses do not depend on how joints are grouped into articulations.

        The base joint is turned to 0.7 rad and the hanging joint left at 0, so
        the whole arm should swing rigidly::

            one articulation (and CPU)   split across two, on CUDA
            --------------------------   -------------------------
                     * mass
                    /                          + base [0.765, 0.644]
               j1  o                          /  (correct in both)
                  /                          /
            base +                          /
                /                          /
            j0 o                    j0    o---------------* mass [1, 0, 0]

            base at [0.765, 0.644]       base is right, the mass is not:
            mass at [1.530, 1.288]       the link was laid out from the
                                         world origin

        ``body_q`` is a function of ``joint_q`` and the parent/child topology.
        Splitting the same two joints across two articulations changes neither,
        so the poses must be identical.

        On CUDA the pendulum articulation reads ``body_q[base]`` before the base
        articulation has written it. What it reads is the *unwritten* value — the
        identity transform ``state.body_q`` was allocated with — so the link
        lands 1 m from the world origin at ``[1, 0, 0]``. Note that is not the
        base's q=0 pose either, which would put it at ``[2, 0, 0]``. CPU happens
        to schedule the base first, so this test passes there.
        """
        inertia = wp.mat33(np.diag([0.1, 0.1, 0.1]).astype(np.float32))  # about each body's own COM [kg m^2]
        axis = wp.vec3(0.0, 0.0, 1.0)

        def build(device, *, split):
            base = builder.add_link(mass=1.0, inertia=inertia, lock_inertia=True)
            mass = builder.add_link(mass=2.0, com=wp.vec3(1.0, 0.0, 0.0), inertia=inertia, lock_inertia=True)
            base_joint = builder.add_joint_revolute(
                parent=-1,
                child=base,
                axis=axis,
                child_xform=wp.transform(p=wp.vec3(-1.0, 0.0, 0.0)),  # base origin sits 1 m out from its axis
            )
            hanging_joint = builder.add_joint_revolute(
                parent=base,
                child=mass,
                axis=axis,
                parent_xform=wp.transform(p=wp.vec3(1.0, 0.0, 0.0)),  # 1 m out from the base body
            )
            if split:
                builder.add_articulation([base_joint], label="base")
                builder.add_articulation([hanging_joint], label="pendulum")
            else:
                builder.add_articulation([base_joint, hanging_joint], label="whole")
            return builder.finalize(device=device)

        for device in get_test_devices():
            with self.subTest(device=str(device)):
                whole = _fk(build(device, split=False), 0.7).body_q.numpy()[:, :3]
                apart = _fk(build(device, split=True), 0.7).body_q.numpy()[:, :3]
                np.testing.assert_allclose(
                    apart,
                    whole,
                    atol=1e-5,
                    err_msg=(
                        "the same mechanism produced different body poses when split across articulations\n"
                        f"  base body: one articulation={np.array2string(whole[0], precision=4)}"
                        f"  split={np.array2string(apart[0], precision=4)}\n"
                        f"  mass body: one articulation={np.array2string(whole[1], precision=4)}"
                        f"  split={np.array2string(apart[1], precision=4)}"
                    ),
                )
```

### Result:
On this test, I get the correct answer when it runs on the `cpu`, and an incorrect answer when it runs on `cuda:0`:
```
======================================================================
test_body_poses_match_when_mechanism_split_across_articulations (test_articulation_rootedness.TestArticulationRootedness.test_body_poses_match_when_mechanism_split_across_articulations) (device='cuda:0')
Verify body poses do not depend on how joints are grouped into articulations.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/jhough/Code/newton/newton/tests/test_articulation_rootedness.py", line 290, in test_body_poses_match_when_mechanism_split_across_articulations
    np.testing.assert_allclose(
  File "/home/jhough/Code/newton/.venv/lib/python3.12/site-packages/numpy/testing/_private/utils.py", line 1780, in assert_allclose
    assert_array_compare(compare, actual, desired, err_msg=str(err_msg),
  File "/home/jhough/Code/newton/.venv/lib/python3.12/site-packages/numpy/testing/_private/utils.py", line 988, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Not equal to tolerance rtol=1e-07, atol=1e-05
the same mechanism produced different body poses when split across articulations
  base body: one articulation=[0.7648 0.6442 0.    ]  split=[0.7648 0.6442 0.    ]
  mass body: one articulation=[1.5297 1.2884 0.    ]  split=[1. 0. 0.]
Mismatched elements: 2 / 6 (33.3%)
Mismatch at indices:
 [1, 0]: 1.0 (ACTUAL), 1.5296844244003296 (DESIRED)
 [1, 1]: 0.0 (ACTUAL), 1.2884353399276733 (DESIRED)
Max absolute difference among violations: 1.2884353
Max relative difference among violations: 1.
 ACTUAL: array([[0.764842, 0.644218, 0.      ],
       [1.      , 0.      , 0.      ]], dtype=float32)
 DESIRED: array([[0.764842, 0.644218, 0.      ],
       [1.529684, 1.288435, 0.      ]], dtype=float32)
```

# Way forward
The simplest way to avoid all of the issues above would be to require that all articulations use the root (`-1`) as their parent.

Otherwise, there are probably a few different spots where we need to make sure these kinematic dependencies are all handled in the correct order. Presumably this would also affect `masking` in each of `eval_fk`, `eval_jacobian` - if one articulation is mounted on another, then when `eval_fk` gets a mask, it must also compute the `fk` of all the articulations it depends on (i.e. more than just the mask would imply).

### Reproduction Script

```python

```

### System Information

- Warp 1.17.0.dev20260807, CUDA 12.9 / driver 13.3, Python 3.12, Ubuntu 24.04
- newton (main branch), commit 5a631742cdd26e14ce803ef7f18909f9325a1383

eric-heiden:
## Description

Reject articulated joints whose parent body belongs to a different articulation during `ModelBuilder.finalize()`.

Newton evaluates and assembles an articulation under the assumption that connected ancestor joints share its articulation metadata. A joint between two separately registered articulations violates that invariant and can produce incorrect kinematics and dynamics results. Finalization now identifies both articulations, the offending joint, and its parent body before the invalid model can be used.

The check is intentionally limited to cross-articulation connections. Existing rootless mechanisms whose parent body is not owned by another articulation remain accepted for solver-specific use. Existing test fixtures that crossed articulation boundaries were corrected: connected chains now share one articulation, while cable anchors outside the cable articulation are created as unarticulated links instead of separate single-body articulations.

The articulation documentation also links directly to `ModelBuilder.add_urdf`, `ModelBuilder.add_mjcf`, and `ModelBuilder.add_usd`: callers should use `parent_body` and `base_joint` when composing imported assets so the new joints are appended to the parent's articulation.

Closes #3968.

## Checklist

## Test plan

```console
uv run --extra dev -m newton.tests -k test_custom_attributes
uv run --extra dev -m newton.tests -k test_05_model_conversions_base_assignment_non_floating_root
uv run --extra dev -m newton.tests -k test_cable_chain_connectivity_cpu
uv run --extra dev -m newton.tests -k test_xpbd_parent_force_
uv run --extra docs --extra sim sphinx-build -j auto -W -b doctest docs docs/_build/doctest docs/concepts/custom_attributes.rst
uv run --extra docs --extra sim sphinx-build -j auto -W -b doctest docs docs/_build/doctest docs/concepts/articulations.rst
uvx pre-commit run -a
```

All 17 cable attachment, drive, limit, and enabled-toggle tests reported by the Ubuntu ARM job were also rerun together and passed.

The new regression was run before the validation change and failed because `ModelBuilder.finalize()` did not raise an error.

## Bug fix

**Steps to reproduce:**

1. Create an articulation rooted at the world.
2. Create a second articulation whose root joint uses a body from the first articulation as its parent.
3. Finalize the model.
4. Observe that finalization accepts the unsupported cross-articulation topology.

**Minimal reproduction:**

```python
import newton

base = builder.add_link(label="base")
base_joint = builder.add_joint_revolute(parent=-1, child=base)
builder.add_articulation([base_joint], label="base_articulation")

pendulum = builder.add_link(label="pendulum")
mount_joint = builder.add_joint_revolute(parent=base, child=pendulum)
builder.add_articulation([mount_joint], label="pendulum_articulation")

builder.finalize()
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Model finalization now rejects joints connecting separate articulations.
  * Articulations must form complete kinematic trees rooted in the world.
  * Improved validation messages identify conflicting articulation relationships.

* **Documentation**
  * Clarified unsupported cross-articulation joints and articulation requirements.
  * Added guidance for appending imported assets using importer configuration options.

* **Tests**
  * Expanded coverage for articulation validation, free and fixed root joints, joint attributes, kinematic bodies, and cable anchors.

### DIFF
--- a/newton/_src/sim/builder.py
+++ b/newton/_src/sim/builder.py
@@ -10419,11 +10419,10 @@ class ModelBuilder:
                 )
 
     def _validate_joints(self):
-        """Validate articulation rootedness and joint membership.
+        """Validate articulation topology and joint membership.
 
-        Every non-world parent of an articulated joint must be the child of another
-        joint in the same articulation. This keeps each articulation rooted at the
-        world and prevents joints from connecting separate articulations.
+        A joint in one articulation cannot have a parent body that belongs to a
+        different articulation. Connected joints must use the same articulation.
 
         Loop-closing joints are allowed when their child is already reachable through
         an articulation. Standalone world-root joints (``parent == -1``) are also
@@ -10435,9 +10434,12 @@ class ModelBuilder:
         """
         if self.joint_count > 0:
             articulation_children = {}
+            body_articulations = {}
             for joint_idx, articulation_idx in enumerate(self.joint_articulation):
                 if articulation_idx >= 0:
-                    articulation_children.setdefault(articulation_idx, set()).add(self.joint_child[joint_idx])
+                    child = self.joint_child[joint_idx]
+                    articulation_children.setdefault(articulation_idx, set()).add(child)
+                    body_articulations.setdefault(child, set()).add(articulation_idx)
 
             for joint_idx, articulation_idx in enumerate(self.joint_articulation):
                 if articulation_idx < 0:
@@ -10451,14 +10453,20 @@ class ModelBuilder:
                 if parent < -1 or parent >= self.body_count:
                     continue
 
+                parent_articulations = body_articulations.get(parent)
+                if not parent_articulations:
+                    continue
+
+                parent_articulation_idx = min(parent_articulations)
                 articulation_label = self.articulation_label[articulation_idx]
+                parent_articulation_label = self.articulation_label[parent_articulation_idx]
                 joint_label = self.joint_label[joint_idx]
                 parent_label = self.body_label[parent]
                 raise ValueError(
-                    f"Articulation {articulation_idx} ('{articulation_label}') is not rooted at the world: "
-                    f"joint {joint_idx} ('{joint_label}') has parent body {parent} ('{parent_label}') outside the "
-                    "same articulation. Articulations cannot be connected through joints. Add all connected joints "
-                    "to the same articulation."
+                    f"Joint {joint_idx} ('{joint_label}') in articulation {articulation_idx} "
+                    f"('{articulation_label}') has parent body {parent} ('{parent_label}') in articulation "
+                    f"{parent_articulation_idx} ('{parent_articulation_label}'). Articulations cannot be connected "
+                    "through joints. Add all connected joints to the same articulation."
                 )
 
             # Find all bodies reachable via articulated joints.
@@ -11151,7 +11159,7 @@ class ModelBuilder:
                 you are confident the model is valid. Default is False.
             skip_validation_worlds: If True, skips validation of world ordering and contiguity. Default is False.
             skip_validation_joints: If True, skips articulation-topology and membership validation. By default,
-                articulations must be rooted at the world, and non-root joints must belong to an articulation or
+                joints cannot connect separate articulations, and non-root joints must belong to an articulation or
                 close a loop; standalone world-root joints are allowed.
             skip_validation_shapes: If True, skips validation of shapes having valid contact margins. Default is False.
             skip_validation_structure: If True, skips validation of structural invariants (body/joint references,
--- a/newton/_src/sim/builder.py
+++ b/newton/_src/sim/builder.py
@@ -10419,7 +10419,11 @@ class ModelBuilder:
                 )
 
     def _validate_joints(self):
-        """Validate that joints belong to an articulation, with two exceptions.
+        """Validate articulation rootedness and joint membership.
+
+        Every non-world parent of an articulated joint must be the child of another
+        joint in the same articulation. This keeps each articulation rooted at the
+        world and prevents joints from connecting separate articulations.
 
         Loop-closing joints are allowed when their child is already reachable through
         an articulation. Standalone world-root joints (``parent == -1``) are also
@@ -10430,7 +10434,34 @@ class ModelBuilder:
             ValueError: If any validation check fails.
         """
         if self.joint_count > 0:
-            # First, find all bodies reachable via articulated joints
+            articulation_children = {}
+            for joint_idx, articulation_idx in enumerate(self.joint_articulation):
+                if articulation_idx >= 0:
+                    articulation_children.setdefault(articulation_idx, set()).add(self.joint_child[joint_idx])
+
+            for joint_idx, articulation_idx in enumerate(self.joint_articulation):
+                if articulation_idx < 0:
+                    continue
+
+                parent = self.joint_parent[joint_idx]
+                if parent == -1 or parent in articulation_children[articulation_idx]:
+                    continue
+
+                # Let structural validation report malformed body references.
+                if parent < -1 or parent >= self.body_count:
+                    continue
+
+                articulation_label = self.articulation_label[articulation_idx]
+                joint_label = self.joint_label[joint_idx]
+                parent_label = self.body_label[parent]
+                raise ValueError(
+                    f"Articulation {articulation_idx} ('{articulation_label}') is not rooted at the world: "
+                    f"joint {joint_idx} ('{joint_label}') has parent body {parent} ('{parent_label}') outside the "
+                    "same articulation. Articulations cannot be connected through joints. Add all connected joints "
+                    "to the same articulation."
+                )
+
+            # Find all bodies reachable via articulated joints.
             articulated_bodies = set()
             articulated_bodies.add(-1)  # World is always reachable
             for i, art in enumerate(self.joint_articulation):
@@ -11119,8 +11150,9 @@ class ModelBuilder:
             skip_all_validations: If True, skips all validation checks. Use for maximum performance when
                 you are confident the model is valid. Default is False.
             skip_validation_worlds: If True, skips validation of world ordering and contiguity. Default is False.
-            skip_validation_joints: If True, skips articulation-membership validation. By default, non-root joints
-                must belong to an articulation or close a loop; standalone world-root joints are allowed.
+            skip_validation_joints: If True, skips articulation-topology and membership validation. By default,
+                articulations must be rooted at the world, and non-root joints must belong to an articulation or
+                close a loop; standalone world-root joints are allowed.
             skip_validation_shapes: If True, skips validation of shapes having valid contact margins. Default is False.
             skip_validation_structure: If True, skips validation of structural invariants (body/joint references,
                 particle topology, array lengths, monotonicity). Default is False.
--- a/newton/_src/sim/builder.py
+++ b/newton/_src/sim/builder.py
@@ -10433,27 +10433,33 @@ class ModelBuilder:
             ValueError: If any validation check fails.
         """
         if self.joint_count > 0:
-            articulation_children = {}
-            body_articulations = {}
+            articulation_to_child_bodies = {}
+            child_body_to_articulations = {}
             for joint_idx, articulation_idx in enumerate(self.joint_articulation):
-                if articulation_idx >= 0:
-                    child = self.joint_child[joint_idx]
-                    articulation_children.setdefault(articulation_idx, set()).add(child)
-                    body_articulations.setdefault(child, set()).add(articulation_idx)
+                if articulation_idx < 0:
+                    continue
+
+                child = self.joint_child[joint_idx]
+                if articulation_idx not in articulation_to_child_bodies:
+                    articulation_to_child_bodies[articulation_idx] = set()
+                articulation_to_child_bodies[articulation_idx].add(child)
+                if child not in child_body_to_articulations:
+                    child_body_to_articulations[child] = set()
+                child_body_to_articulations[child].add(articulation_idx)
 
             for joint_idx, articulation_idx in enumerate(self.joint_articulation):
                 if articulation_idx < 0:
                     continue
 
                 parent = self.joint_parent[joint_idx]
-                if parent == -1 or parent in articulation_children[articulation_idx]:
+                if parent == -1 or parent in articulation_to_child_bodies[articulation_idx]:
                     continue
 
                 # Let structural validation report malformed body references.
                 if parent < -1 or parent >= self.body_count:
                     continue
 
-                parent_articulations = body_articulations.get(parent)
+                parent_articulations = child_body_to_articulations.get(parent)
                 if not parent_articulations:
                     continue
 
--- a/changelog/3968.fixed.md
+++ b/changelog/3968.fixed.md
@@ -1,0 +1,1 @@
+Reject joints that connect separate articulations during model finalization, preventing incorrect kinematics and dynamics results.
--- a/docs/concepts/articulations.rst
+++ b/docs/concepts/articulations.rst
@@ -864,12 +864,12 @@ with a desired acceleration:
 Orphan joints
 -------------
 
-Each articulation must form a kinematic tree rooted at the world. In particular,
-the parent body of every articulated joint must either be the world or be the child
-of another joint in the same articulation. Connecting two articulations with a
-joint is unsupported and :meth:`~newton.ModelBuilder.finalize` rejects that
-topology. To compose imported robots, pass ``parent_body`` and ``base_joint`` to
-the importer so the new joints are appended to the parent's articulation.
+A joint in one articulation cannot use a body from another articulation as its
+parent. This cross-articulation topology is unsupported, and
+:meth:`~newton.ModelBuilder.finalize` rejects it. To compose imported assets with
+:meth:`~newton.ModelBuilder.add_urdf`, :meth:`~newton.ModelBuilder.add_mjcf`, or
+:meth:`~newton.ModelBuilder.add_usd`, pass ``parent_body`` and ``base_joint`` so
+the new joints are appended to the parent's articulation.
 
 An **orphan joint** is a joint that is not part of any articulation **and** whose child body is not reachable through any articulated joint (i.e. the child has no articulated path back to the rest of the model). This situation can arise when:
 
--- a/docs/concepts/articulations.rst
+++ b/docs/concepts/articulations.rst
@@ -882,7 +882,7 @@ USD import preserves joints outside authored articulations without emitting an a
 
 **Validation and finalization**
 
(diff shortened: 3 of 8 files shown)


newton/_src/sim/builder.py
```diff
@@ -10430,7 +10433,43 @@ def _validate_joints(self):
...
         """
         if self.joint_count > 0:
-            # First, find all bodies reachable via articulated joints
+            articulation_children = {}
+            body_articulations = {}
+            for joint_idx, articulation_idx in enumerate(self.joint_articulation):
+                if articulation_idx >= 0:
+                    child = self.joint_child[joint_idx]
+                    articulation_children.setdefault(articulation_idx, set()).add(child)
+                    body_articulations.setdefault(child, set()).add(articulation_idx)
```
jeff-hough: nit: this part took me a second to mentally parse, I think writing it like this might be simpler to read:

```
  articulation_children = {}
  body_articulations = {}
  for joint_idx, articulation_idx in enumerate(self.joint_articulation):
      if articulation_idx < 0:
          continue
      child = self.joint_child[joint_idx]
      if articulation_idx not in articulation_children:
          articulation_children[articulation_idx] = set()
      articulation_children[articulation_idx].add(child)
      if child not in body_articulations:
          body_articulations[child] = set()
      body_articulations[child].add(articulation_idx)

```

jeff-hough: another nit, but I also think naming the variables something like `articulation_to_child_bodies` and `child_body_to_articulations` makes the rest of the checks easier to read as well

jeff-hough (commented):
One nit found by an agent:

>  docs/concepts/articulations.rst:866 — The "Validation and finalization" section (which explains the skip_validation_joints escape hatch) still describes finalize() as raising ValueError only "for non-root orphan joints." It doesn't mention that the new check also rejects joints whose parent belongs to a different articulation. A reader who hits the new cross-articulation error and jumps to that section for guidance on suppressing it won't realize skip_validation_joints=True also covers this new case.

Otherwise just very minor suggestions which (for me personally) improve the readability slightly.

jcarius-nv:
Tagging @camevor here in case you want to take a look, this seems closely related to some of your earlier work on warning about illegal joint configurations.

newton/_src/sim/builder.py
eric-heiden: Addressed in 56f90708a. I expanded the lookup construction and renamed the mappings to articulation_to_child_bodies and child_body_to_articulations.

eric-heiden:
Addressed the remaining review notes in 56f90708a: documented the add_articulation fixture option, updated the validation/finalization section to cover cross-articulation joints and skip_validation_joints=True, and simplified the articulation/body lookup construction and names.

newton/_src/sim/builder.py
```diff
@@ -10419,7 +10419,10 @@ def _validate_world_ordering(self):
                 )
 
     def _validate_joints(self):
-        """Validate that joints belong to an articulation, with two exceptions.
+        """Validate articulation topology and joint membership.
+
+        A joint in one articulation cannot have a parent body that belongs to a
```
camevor, line 10424: This makes sense. This raises two tangential questions:
- Should we check that any non-articulation parent has 0 dofs? `eval_jacobian` will produce invalid results for such configurations.
- Should we allow parents that are in a different articulation but have 0 dofs? That is not obviously wrong, although stating that it's a limitation is reasonable.

camevor (approved):
Looks good overall. One comment about granularity added.

Another concern is the startup performance impact. Would it be possible to speed this up without causing issues? A quick A/B test on this PR (6d3fdf6f and 56f90708) with 4096 G1 worlds shows a 67 ms increase in `finalize()` time. End to end startup time also seems to increase by ~300 ms across these refs, which is puzzling.

Issue #42, opened 2025-05-05
Bring bugfix for fixed particles over to Newton


shi-eric:
Warp: https://github.com/NVIDIA/warp/commit/e95adb60a2aa1cde0efe91d59fdd2821c16b8a85

PR #3970, opened 2026-08-19
Bump newton-assets and Menagerie pins

andrewkaufman:
## Description

The structured USD robots in `newton-assets` have been regenerated so they all derive from a single Menagerie revision ([newton-physics/newton-assets#49](https://github.com/newton-physics/newton-assets/pull/49)). This bumps `NEWTON_ASSETS_REF` to pick that up, and moves `MENAGERIE_REF` to the same revision the assets were converted from.

Both pins have to move together. `test_menagerie_usd_mujoco.py` compares a USD import against the native MJCF parse of the same robot, so a USD asset built from `da76818` has to be compared against `da76818` MJCF — leaving `MENAGERIE_REF` on `affef08` compares two different revisions of the robot.

Opening as a draft to get CI results on the asset update.

## Checklist

- [x] I am familiar with the [Contributing Guidelines](https://github.com/newton-physics/newton/blob/main/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

## Summary by CodeRabbit

* **Chores**
  * Updated the bundled Newton and MuJoCo Menagerie assets to newer revisions.

* **Bug Fixes**
  * Corrected Robotiq 2F-85 finger-pad mass data and improved consistency checks between USD and MJCF formats.
  * Updated DR Legs model expectations to recognize the corrected joint count across supported USD import variants.

### DIFF
--- a/newton/_src/utils/download_assets.py
+++ b/newton/_src/utils/download_assets.py
@@ -20,10 +20,10 @@ from warp._src.thirdparty.appdirs import user_cache_dir
 # commit.  Update these SHAs when assets change upstream and the new versions
 # have been validated against Newton's test suite.
 NEWTON_ASSETS_URL = "https://github.com/newton-physics/newton-assets.git"
-NEWTON_ASSETS_REF = "26c7f5ccaadd9a163f4d6c8f867c789b3ac57d83"
+NEWTON_ASSETS_REF = "a96f0973b6ae69c90609f9fafef9d2c1db2d6431"
 
 MENAGERIE_URL = "https://github.com/google-deepmind/mujoco_menagerie.git"
-MENAGERIE_REF = "affef0836947b64cc06c4ab1cbf0152835693374"
+MENAGERIE_REF = "da76818e269b82289eba39808e2fb91d679d6994"
 
 _SHA_RE = re.compile(r"[0-9a-f]{40}")
 
--- a/newton/_src/solvers/kamino/tests/test_utils_io_usd.py
+++ b/newton/_src/solvers/kamino/tests/test_utils_io_usd.py
@@ -1185,7 +1185,7 @@ class TestUSDImporter(unittest.TestCase):
 
         # Check the loaded contents
         self.assertEqual(builder_usd.num_bodies, 31)
-        self.assertEqual(builder_usd.num_joints, 36)
+        self.assertEqual(builder_usd.num_joints, 37)
         self.assertEqual(builder_usd.num_geoms, 31)
 
     def test_import_model_dr_legs_with_boxes(self):
@@ -1200,7 +1200,7 @@ class TestUSDImporter(unittest.TestCase):
 
         # Check the loaded contents
         self.assertEqual(builder_usd.num_bodies, 31)
-        self.assertEqual(builder_usd.num_joints, 36)
+        self.assertEqual(builder_usd.num_joints, 37)
         self.assertEqual(builder_usd.num_geoms, 3)
 
     def test_import_model_dr_legs_with_meshes_and_boxes(self):
@@ -1215,7 +1215,7 @@ class TestUSDImporter(unittest.TestCase):
 
         # Check the loaded contents
         self.assertEqual(builder_usd.num_bodies, 31)
-        self.assertEqual(builder_usd.num_joints, 36)
+        self.assertEqual(builder_usd.num_joints, 37)
         self.assertEqual(builder_usd.num_geoms, 34)
 
     @unittest.skipUnless(USD_AVAILABLE, "Requires usd-core")
--- a/newton/tests/test_menagerie_usd_mujoco.py
+++ b/newton/tests/test_menagerie_usd_mujoco.py
@@ -1805,13 +1805,10 @@ class TestMenagerieUSD_Robotiq2f85V4(TestMenagerieUSD):
     num_steps = 20
     fk_enabled = True
     # Menagerie PR #252 corrected the source finger pads from 2e-6 kg to
-    # 0.0035 kg. The pinned USD was generated from the old source and still
-    # authors the old mass on its colliders; the test below tracks that fixture.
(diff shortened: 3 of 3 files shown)


newton/_src/utils/download_assets.py
```diff
@@ -20,10 +20,10 @@
 # commit.  Update these SHAs when assets change upstream and the new versions
 # have been validated against Newton's test suite.
 NEWTON_ASSETS_URL = "https://github.com/newton-physics/newton-assets.git"
-NEWTON_ASSETS_REF = "26c7f5ccaadd9a163f4d6c8f867c789b3ac57d83"
+NEWTON_ASSETS_REF = "a96f0973b6ae69c90609f9fafef9d2c1db2d6431"
```
eric-heiden, line 23: 🔵 This bump changes the pinned Robotiq 2F-85 v4 structured USD, which leaves the workaround in `newton/tests/test_menagerie_usd_mujoco.py` out of date.

At `a96f097`, `robotiq_2f85_v4/usd_structured/Payload/Physics.usda` authors `physics:mass = 0.00175` on `left_pad1`, `left_pad2`, `right_pad1`, and `right_pad2`. The old pin `26c7f5c` authored `0.000001` on those same colliders, so the USD now agrees with the `mass="0.00175"` geoms in `robotiq_2f85_v4/2f85.xml`.

Three spots in that test file still assume the old values:

| Location | Problem |
| --- | --- |
| lines 1828-1829 | The `np.isclose(usd_mass, 2.0e-6)` skip guard can no longer fire, so it is dead code. |
| lines 1807-1809 | The class comment says the pinned USD still authors the old mass on its colliders, which is no longer true. |
| lines 1812-1814 | The `_compare_inertia` no-op keeps the whole body mass and inertia comparison disabled for this robot, even though the finger pad mismatch it was excusing is now fixed. |

The last one is the part that costs real coverage: a future regression in Robotiq inertia would pass silently.

Could we fold that test update into this PR? Concretely: remove the dead `2.0e-6` guard, rewrite the class comment to describe the current fixture, and either drop the `_compare_inertia` override so the mapped inertia comparison runs again, or keep it with a justification based on the COM and `principalAxes` differences that this asset commit also introduces. The previous Menagerie pin bump updated this same test class in lockstep, so doing it here keeps the two in sync.

andrewkaufman: Good catch on all three, and the inertia one was worth more than it looked.

Folded the test update in as [`9d7b7f1`](https://github.com/newton-physics/newton/commit/9d7b7f13cf132d63f4d115402d215f147ec47096). Confirmed the fixture change first: at `a96f097`, `robotiq_2f85_v4/usd_structured/Payload/Physics.usda` authors `physics:mass = 0.00175` on `left_pad1`, `left_pad2`, `right_pad1` and `right_pad2`, where `26c7f5c` authored `0.000001` at those same four prims, so the USD now agrees with the `mass="0.00175"` pad geoms in `2f85.xml`.

On the third point, I dropped the `_compare_inertia` override outright rather than keeping it with a justification. Reindexing Newton's arrays to native ordering through `_body_map`, body mass is now exact and inertia is within float noise:

```
body_mass     max abs diff: 0.0
body_inertia  max abs diff: 5.8e-11
body_ipos     max abs diff: 1.5e-08
```

`body_iquat` does differ on seven bodies (up to 1.4), but those are sign and axis-order permutations of the same principal frame, and `compare_inertia_tensors` reconstructs `I = R @ diag(d) @ R.T` before comparing, so it is insensitive to them. Running the real `compare_inertia_tensors_mapped` that removing the override re-enables passes. So the COM and `principalAxes` differences you flagged as a possible reason to keep it are present but not load-bearing.

Also removed the now-unreachable `2.0e-6` guard and rewrote the class comment to describe the current fixture.

Verified with a negative control, since this class has three failures in my environment either way: `test_dynamics`, `test_forward_kinematics` and `test_model_comparison` fail identically on the unmodified branch and on the merge base, all of them `opt.sleep_tolerance` mismatches from a local mujoco_warp version skew rather than anything in this PR. The sorted failure sets are identical with and without my change. The signal is `test_pad_mass_matches_source_scale`, which skipped on the merge base and passes on the branch — exactly the dead guard you identified.

Separately, and beyond what you asked for: I added [`3efa87e`](https://github.com/newton-physics/newton/commit/3efa87e8b6b7461d295bd2edae34653c6c7efa39) updating three DR Legs joint expectations in `newton/_src/solvers/kamino/tests/test_utils_io_usd.py` from 36 to 37. That is the remaining half of #3844: newton-assets #42 added `PhysicsArticulationRootAPI` to the DisneyResearch models, so the importer now yields a synthesized `world_to_pelvis` root joint alongside the 36 authored revolute joints. The plan recorded in that issue was to land this with the assets pin bump once newton-assets#49 merged, which is this PR, so it seemed better here than in a standalone follow-up. Happy to split it back out if you would rather keep this PR to the two pins.

One thing that commit deliberately does not fix. With the joint counts corrected, two geom expectations in the same file fail: `test_import_model_dr_legs` asserts `num_geoms == 31` and sees 81, and `test_import_model_dr_legs_with_meshes_and_boxes` asserts 34 and sees 84. Those are not from this bump — the `disneyresearch` asset tree is byte-identical between `26c7f5c` and `a96f097` (`diff -rq` is empty), and applying only the 36 to 37 change on top of the merge base reproduces both failures with the old pins. They were previously masked because the joint assertion failed first. That directory is also manual-only: it has its own `__main__.py` and README describing `python -m unittest discover`, and no workflow references it, which is why CI stayed green through all of this. Left alone as out of scope here.

andrewkaufman (commented):
Addressed the review; details in the thread reply.

eric-heiden (approved):
Thanks!

eric-heiden:
@chschuma-disney or @rubengrandia can you also approve this PR if it looks good to you? We need your approval to merge.

Guirec-Maloisel (approved):

PR #3975, opened 2026-08-19
[Kamino] Fix metrics for NaN in the solution

rubengrandia:
## Description

Warp max does not propagate NaN value: `wp.max(v, NaN) -> v`. The result was that all metrics remain 0.0 in case the solution contains NaNs. This PR propagates the NaN values over the max finite such that it is correctly reflected in the metrics.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved solver metric handling when calculations contain invalid or missing numerical values.
  - Residual and objective metrics now consistently report undefined results instead of misleading values.
  - Preserved relevant maximum-error locations for easier diagnosis.
  - Resetting metrics now reliably clears previous invalid results before new calculations.
  - Free joints no longer contribute incorrectly to kinematics metrics.

- **Tests**
  - Expanded coverage across contact, joint, limit, primal, and dual solver metrics in dense and sparse modes.
  - Added regression coverage confirming correct handling of free-joint kinematics.

### DIFF
--- a/newton/_src/solvers/kamino/_src/solvers/metrics.py
+++ b/newton/_src/solvers/kamino/_src/solvers/metrics.py
@@ -802,6 +802,8 @@ def _compute_joint_kinematics_residual_dense(
 
     # Retrieve the size and index offset of the joint constraint
     num_cts_j = model_joint_num_kinematic_cts[jid]
+    if num_cts_j == 0:
+        return
     cts_offset_j = model_joint_kinematic_cts_offset_total_cts[jid] - model_info_total_cts_offset[wid]
 
     # Retrieve the world-specific info
@@ -865,6 +867,8 @@ def _compute_joint_kinematics_residual_sparse(
     # Retrieve the size and index offset of the joint constraint
     num_dyn_cts_j = model_joint_num_dynamic_cts[jid]
     num_kin_cts_j = model_joint_num_kinematic_cts[jid]
+    if num_kin_cts_j == 0:
+        return
     kin_cts_offset_j = model_joint_kinematic_cts_offset[jid] - model_info_joint_kinematic_cts_offset[wid]
 
     # Retrieve the starting index for the non-zero blocks for the current joint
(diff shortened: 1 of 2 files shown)


newton/_src/solvers/kamino/_src/solvers/metrics.py
```diff
@@ -965,6 +1082,24 @@ def _compute_dual_problem_metrics(
...
+        r_ncp_p = wp.nan
+        r_ncp_p_argmax = wp.int32(-1)
+        r_ncp_d = wp.nan
+        r_ncp_d_argmax = wp.int32(-1)
+        r_ncp_c = wp.nan
+        r_ncp_c_argmax = wp.int32(-1)
+        r_ncp_natmap = wp.nan
+        r_ncp_natmap_argmax = wp.int32(-1)
+
+    if _vector_has_nan(ncts, vio, solution_lambdas) or _vector_has_nan(ncts, vio, problem_v_f):
```
Guirec-Maloisel: Small optimization, _vector_has_nan(ncts, vio, solution_lambdas) is computed twice (and loops sequentially over all lambdas), we could store that result. Also applies below (line 1200+). Maybe the true solution would be some tile-based kernels (for this, and possibly for other operations in this module), but that sounds like a bigger refactor, and performance isn't critical here.

Guirec-Maloisel (commented):
Nice find, looks good to me up to some nitpick (+ CodeRabbit's comment which I agree with). I'm realizing now that I had to write very similar code for the FK solver a while ago because of the same issue (see `_eval_max_residual` in solvers/fk/kernels.py) but since then had forgotten about this...

newton/_src/solvers/kamino/_src/solvers/metrics.py
rubengrandia: Thanks — commit 832fbf09 caches the `solution_lambdas` NaN result once per kernel and reuses it for both invalidation checks. I kept the broader tile-based optimization out of scope for this PR.

newton/_src/solvers/kamino/_src/solvers/metrics.py
```diff
@@ -550,6 +551,131 @@ def compute_v_plus_sparse(
...
+        if wp.isnan(raw):
+            has_nan = True
+        else:
+            v = wp.abs(raw)
+            if v > max_val:
+                max_val = v
+                argmax = wp.int32(i)
+    if has_nan:
+        return wp.nan, wp.int32(-1)
+    return max_val, argmax
```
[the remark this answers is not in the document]
rubengrandia: Good catch — commit 832fbf09 skips kinematics metric updates for zero-constraint FREE joints before the norm/helper and argmax-key construction. It also adds dense and sparse coverage to preserve the unset `-1` argmax.

Guirec-Maloisel (approved):

PR #3976, opened 2026-08-19
Fix headless pyglet backend selection for offscreen video recording

rubengrandia:
## Description

Kamino records simulation videos by rendering frames offscreen through `RendererGL` with `headless=True` and `visible=False`. On machines without a display (CI runners, remote servers, SSH sessions), this currently fails because pyglet opens a display connection during `pyglet.graphics` import before the headless backend is selected.

This PR sets `pyglet.options["headless"] = True` immediately after importing pyglet and **before** importing `pyglet.graphics`, matching pyglet's requirement that the headless backend be chosen at import time.

This is a prerequisite for Kamino headless video recording. The Kamino examples still need follow-up wiring (create a headless `ViewerKamino` when `--record` is set, call `render()` in the headless loop) — they currently skip viewer creation and disable recording when `--headless` is passed.

## Checklist

## Test plan

Verified that `RendererGL` initializes headless without a display connection:

```python
from newton._src.viewer.gl.opengl import RendererGL

renderer = RendererGL(headless=True, screen_width=640, screen_height=480, vsync=False)
assert renderer.headless is True
renderer.close()
```

Also verified end-to-end offscreen frame capture by creating a headless `ViewerKamino` with `record_video=True` and calling `render()` after each simulation step — PNG frames are written successfully.

## Bug fix

**Steps to reproduce:**

1. Construct `RendererGL(headless=True)` on a machine without a display (e.g. CI, SSH without X forwarding).
2. Observe pyglet failing to initialize OpenGL because `pyglet.graphics` creates a display connection before the headless backend is selected.

**Minimal reproduction:**

```python
from newton._src.viewer.gl.opengl import RendererGL

# Fails without this PR when $DISPLAY is unset
renderer = RendererGL(headless=True, screen_width=640, screen_height=480)
```

## Summary by CodeRabbit

* **New Features**
  * Added reliable headless OpenGL rendering and frame capture on Linux systems without an X display.
  * Automatically selects EGL for displayless Linux environments while preserving native rendering elsewhere.
* **Bug Fixes**
  * Improved headless renderer initialization, visibility, event handling, and callbacks.
  * Added clear errors for incompatible backend configurations and missing EGL runtimes.
* **Documentation**
  * Clarified EGL requirements, hidden-window behavior, GPU selection, and restrictions when mixing native and headless viewers.
* **Tests**
  * Expanded coverage for backend selection, headless settings, compatibility checks, and subprocess rendering.


newton/_src/viewer/gl/opengl.py
```diff
@@ -1018,6 +1018,11 @@ def __init__(self, title="Newton", screen_width=1920, screen_height=1080, vsync=
         try:
             import pyglet
 
+            if headless:
+                # Pyglet must select its headless backend before importing
+                # pyglet.graphics, which otherwise creates a display connection.
+                pyglet.options["headless"] = True
```
eric-heiden: 🟠 `pyglet.options["headless"] = True` does more than skip the window: it switches pyglet's entire backend to EGL, on every platform. When `pyglet.gl` is imported a few lines below, it takes the headless path (`pyglet/gl/headless.py`), which does `from pyglet.libs.egl import egl` and ends up calling `pyglet.lib.load_library('EGL')`. With no libEGL present that raises `ImportError('Library "EGL" not found.')` inside this `try` block, and it is re-raised as the misleading `OpenGLRenderer requires pyglet (version >= 2.0) to be installed.`

macOS ships no libEGL and most Windows installs have none, so `ViewerGL(headless=True)` now hard-fails there. That covers `python -m newton.examples <name> --headless` and the headless usage documented in `docs/guide/visualization.rst`. Before this change the same call created a hidden native window via `visible=not headless`, which works on Linux, macOS and Windows.

CI is affected too. `newton-unittests` runs on `windows-latest` and `macos-latest` with the `dev` extra that pulls in pyglet, and `_is_viewer_gl_unavailable_error` in `newton/tests/test_viewer_get_frame.py` only converts pyglet-namespaced exceptions into skips. A builtins `Exception` propagates, so `test_headless_capture_main_image_frame` and its siblings will error instead of skipping.

One way to keep the EGL backend limited to the case where the native path genuinely cannot work (`os` and `sys` are already imported at the top of this file):

```suggestion
            has_display = bool(os.environ.get("DISPLAY") or os.environ.get("WAYLAND_DISPLAY"))
            if headless and sys.platform.startswith("linux") and not has_display:
                # Pyglet must select its headless backend before importing
                # pyglet.graphics, which otherwise creates a display connection.
                pyglet.options["headless"] = True
```

If you would rather keep it unconditional, wrap the `pyglet.gl` import in its own `try/except ImportError`, reset `pyglet.options["headless"] = False`, drop the partially imported `pyglet.gl` and `pyglet.window` entries from `sys.modules`, and fall back to the hidden-window path so non-EGL platforms keep working.

Worth handling at the same time: `pyglet.options` is process global and this flag is never restored, so a windowed `ViewerGL` created later in the same process stays pinned to the EGL backend and loses its GL context. Saving the previous value and restoring it once the window exists would keep mixed headless and windowed usage working.

rubengrandia: Thanks, this was correct. I narrowed pyglet's EGL backend to headless Linux without `DISPLAY` and preserved the native hidden-window path elsewhere, including when only `WAYLAND_DISPLAY` is set because pyglet's Linux backend is Xlib. Since pyglet binds its GL/window/display backend classes at import time, the implementation now rejects incompatible backend reuse instead of attempting to restore the option afterward.

I added selection and conflict regression tests plus an isolated no-display Linux initialization smoke test. `test_viewer_gl_pyglet_backend`, `test_viewer_get_frame`, and pre-commit now pass.

rubengrandia:
I tried to incorporate all the feedback from @eric-heiden and CodeRabbit and landed on the current machinery. To be honest, it became quite a bit more complex than I hoped. Leaving it up to your Eric if this is the way to go or if the environment variable as noted in https://github.com/newton-physics/newton/pull/3932 is the solution for now.

eric-heiden:
> I tried to incorporate all the feedback from @eric-heiden and CodeRabbit and landed on the current machinery. To be honest, it became quite a bit more complex than I hoped. Leaving it up to your Eric if this is the way to go or if the environment variable as noted in #3932 is the solution for now.

Thanks @rubengrandia! I think it's better we leave this as it is and avoid such a complex solution. Passing `PYGLET_HEADLESS=1` as environment variable should be the easiest and preferred solution here, which is now properly documented as well.

PR #3979, opened 2026-08-19
Reworking `ControllerJointImpedance` (`Model` in place of `ModelBuilder`, `wp.indexedarray` in place of locally stored indices)

jeff-hough:
## Description

This PR updates the interfaces of `ControllerJointImpedance` and `ControllerJointImpedanceModelFree` to simplify working with these controllers in `newton`. The main changes:

### The `ControllerJointImpedance` constructor takes a `Model` instead of a `ModelBuilder`.
* The `ModelBuilder` has awkward ownership semantics (it was not clearly `deepcopy`-able, nor is it _guaranteed_ to be usable after `finalize()` was called), so the only safe thing for a user to do is pass a completely independent `ModelBuilder` from the sim-model.
* By taking a `Model`, the common use case of "adding a controller to the simulated model" is much easier to do.
* An independent model is still totally possible - you create a second `ModelBuilder`, which is the path expected currently in `main` anyways.

### All `*_idx` arguments have been deleted from the controller constructors.
* Instead, the controllers work almost exclusively with `wp.indexedarray | wp.array` on inputs/outputs; we assume that only the needed data is passed in, nothing more.
* This is not any more complex for the user (they pass the indices to the `view`, instead of passing them to the controller constructor), but it significantly reduces the number of local variables which have to be tracked and verified in the controllers.
* This cuts the number of constructor arguments, and all of the default `*_idx` values which had to be picked (which were probably going to trip people up at some point)
* This simplifies array-validation, since we can check the _exact_ size of the view we are passed, rather than a "minimum" size which can still let many bugs slip through.
* It is verified that this does not break APIC-export (see draft PR on my fork here: https://github.com/jeff-hough/newton/pull/1)
* The constructor for `ControllerJointImpedance` now selects joints via `articulations` and `joints` arguments, and uses a private heterogeneous selection helper internally to find the correct coord/dof indices to control. By default if `joints` is set to `None`, then only valid 1DOF joints in the model will be selected for control. Selecting an invalid joint (i.e. Ball joint) will raise an error at construction.

### `ControllerJointImpedance` takes the "full" model `joint_q` and `joint_qd` as inputs
* This is an active bug on `main` - `eval_fk`, `eval_mass_matrix`, `eval_inverse_dynamics_passive` need the full estimated robot-state, but only the controlled DOFs are written into the internal state today. Every uncontrolled joint is silently evaluated at whatever value was last written, which corrupts all computations.

### The controller is only constrained by the joints it actually controls
* A model with a free, ball, distance, or D6 joint in it is now accepted; only a joint you _control_ has to span a single coordinate and a single DOF. Previously any one of these anywhere in the model was rejected outright.
* An articulation can also be left uncontrolled entirely, in which case it takes up no slot in any buffer and is masked out of the FK and dynamics evaluations.

### `damping` and `stiffness` are now 1d array inputs
* This was flagged by @jvonmuralt in the last PR, and was indeed a bit awkward - these were 2d arrays with padding when almost everything else was 1d with no padding.

## Checklist

## Test plan

```bash
``` 

## New feature / API change

```python
import newton
import warp as wp
from newton.controllers import ControllerJointImpedance

device = wp.get_device()

scene = newton.ModelBuilder()
for i in range(512):

control = model.control()

# -- pick what to control --
# Here every panda articulation (not the conveyor),

controller = ControllerJointImpedance(
    model,
    articulations="panda_*",
    joints=[f"panda_joint{k}" for k in range(1, 8)], # don't select the finger joints.
    stiffness=100.0,  # None to drive the gains live via inputs.stiffness
    damping=10.0, # None to drive the gains live via inputs.damping
    use_gravity_compensation=True,
    use_coriolis_compensation=True,
    use_inertia_decoupling=True,
)

inputs = controller.input()
outputs = controller.output()

# joint_q/joint_qd are whole-model — an unselected joint still moves the bodies
# downstream of it, and so the mass matrix and gravity terms of the joints that
# are controlled. They are bound in the loop below, since the physics loop
# double-buffers. Every other port is compact.
inputs.joint_q_des = wp.zeros(controller.total_controlled_dofs, dtype=wp.float32, device=device)
inputs.joint_qd_des = wp.zeros(controller.total_controlled_dofs, dtype=wp.float32, device=device)

# An indexed view scatters the compact torque command straight into the sim,
# leaving the conveyor's and the fingers' entries in control.joint_f untouched.
outputs.joint_f = control.joint_f[controller.qd_start]

# -- sim loop --
for _ in range(1000):
    inputs.joint_q_des.assign(desired_q)  # compact, shape [total_controlled_dofs]

    for _ in range(sim_substeps):
        state_0.clear_forces()

```

## New Features
- Added joint selection by index, name, glob pattern, or regular expression.
- Added compact impedance-controller ports sized to controlled degrees of freedom.
- Added support for heterogeneous robots and indexed array views.

## Changes
- Controllers now use finalized models and explicit joint indices.
- Simplified gain and torque interfaces with compact one-dimensional arrays.
- Uncontrolled articulations remain unchanged during control.
- Added controlled-DOF count properties and stronger input validation.
- Removed legacy joint-index override and sizing arguments.

## Documentation
- Documented joint selection and compact controller behavior.


docs/api/newton_controllers.rst
```diff
@@ -24,3 +27,12 @@ subclass of :class:`ControllerBase`.
...
    ControllerJointImpedanceModelFree
+   JointSelection
+
+.. rubric:: Functions
+
+.. autosummary::
+   :toctree: _generated
+   :signatures: long
+
+   select_joints
```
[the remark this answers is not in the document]
jeff-hough: not going to do this in a controller PR

jcarius-nv (changes requested):
# Review of PR #3979

Reviewed PR: https://github.com/newton-physics/newton/pull/3979

- Head: `c727679731ed3e40843c53de1d168467951ce564`
- Base: `904aef2bbfd2a2603cc85cfcefc19d8fa9b3934d`
- Recommendation at this head: **Request changes; do not approve yet.**

## Spec

### P1 — The fleet-selection example does not work

`select_joints()` matches complete model labels. After
`add_builder(..., label_prefix="panda_0")`, labels are slash-prefixed, such as
`panda_0/panda` and `panda_0/panda_joint1`.

- The PR-body argument `articulations="panda_.*"` is treated as a glob, not a
  regular expression, and matches nothing.
- The reusable bare joint names such as `"panda_joint1"` do not match the
  prefixed full labels.

Both failures were reproduced at the pinned head. This contradicts the PR's
central example and its statement that one joint list can serve the whole
prefixed fleet. Either support articulation-relative joint matching or revise
the promised API and all corresponding examples.

Evidence:

- `newton/_src/controllers/joint_selection.py:137`
- `newton/_src/controllers/joint_selection.py:164`

### P1 — Caller mutation silently changes what the controller controls

`ControllerJointImpedance` retains the caller-owned `joint_q_idx` and
`joint_qd_idx` arrays and constructs live indexed views from them. The robot
packing, local-DOF tables, and masks remain derived from the original host
snapshot.

An exact-head probe changed the computed torque from `-1` to `-10` after
mutating those arrays. Depending on the new indices, this can combine state
from one selection with topology and dynamics derived from another.

Copy both arrays into private Warp arrays before constructing the views, and
add a regression test that mutates the caller's arrays after construction.

Evidence:

- `newton/_src/controllers/impl/joint_impedance/model_based.py:296-298`
- `newton/_src/controllers/impl/joint_impedance/model_based.py:360-367`

### P2 — APIC verification remains incomplete

The linked fork PR, https://github.com/jeff-hough/newton/pull/1, is still a
draft with an empty test plan/checklist. Its test and GPU jobs were skipped.

The focused Newton suite confirms CUDA graph capture, but it does not
independently verify the claimed C++ APIC capture/save/load/replay path. Put
durable executable evidence for that path in-tree or narrow the claim.

No other Spec omissions or unasked scope creep were found.

## Standards

### Medium — Public port annotations exclude supported indexed arrays

The public controller port annotations say `wp.array`/`wp.array3d`, while the
documentation and runtime validation explicitly accept `wp.indexedarray`.
The annotations therefore describe a narrower API than the implementation.

Evidence:

- `newton/_src/controllers/impl/joint_impedance/model_based.py:118-136`
- `newton/_src/controllers/impl/joint_impedance/model_free.py:112-136`
- `AGENTS.md`: types belong in annotations, not only in docstrings.

### Medium — `select_joints()` annotations exclude supported inputs

The implementation and tests accept mixed index/pattern lists and list entries
containing compiled regular expressions, but the annotations permit only
homogeneous `list[int]`, homogeneous `list[str]`, or a single compiled pattern.
Broaden the element union so the public type contract matches the supported
behavior.

Evidence:

- `newton/_src/controllers/joint_selection.py:53-57`
- `newton/tests/test_controllers_joint_impedance.py:1296-1298`

No other Standards violations were found. The module-level experimental marker
permits the breaking controller changes; the Towncrier fragments, test
docstrings, public exports, and API inclusion otherwise conform.

## Architecture discussion

The following are interface ideas for discussion, not additional merge
blockers beyond the reproduced correctness findings above.

### Should the canonical controller depend on a Newton `Model`?

The earlier controller discussion established a portable controller layer that
can run across physics engines and, through APIC, on hardware. Against that
goal, requiring a Newton `Model` at the portable controller seam would be the
wrong ownership choice.

The current implementation does retain an engine-independent route through
`ControllerJointImpedanceModelFree`: callers provide compact state arrays and
the enabled mass-matrix, gravity, and Coriolis terms. The unsuffixed
`ControllerJointImpedance` is instead a Newton-specific convenience adapter that
computes those terms from `Model`.

Using a finalized `Model` rather than `ModelBuilder` is the right choice inside
that adapter. The point for discussion is whether the naming and layering
should make the portable law canonical and the Newton-backed adapter explicit,
for example:

- `ControllerJointImpedance`: portable law with caller-supplied dynamics.
- `ControllerJointImpedanceNewton` or
  `ControllerJointImpedanceNewtonDynamics`: Newton `Model` adapter.

This would let another simulator use the controller without constructing a
parallel Newton model, while retaining the convenient Newton-integrated path.
`ModelFree` is also slightly misleading because the caller may still supply a
model-derived mass matrix and compensation terms; the distinction is really
caller-supplied versus Newton-computed dynamics.

### Should callers supply two parallel index arrays?

The implementation genuinely needs separate coordinate and DOF mappings:
`joint_q` and `joint_qd` occupy different index spaces when joints such as
quaternions make `nq != nv`. That does not mean callers should identify every
controlled joint twice.

`JointSelection` already bundles `q_idx` and `qd_idx`, but the constructor
immediately splits them back into two independently mutable arguments and then
needs extensive length, pairing, grouping, and ownership validation.

Consider accepting one coherent `selection: JointSelection` argument, with the
selection derived from joint identity once. The controller should expose the
resolved selection for indexed input/output binding and copy its device arrays
so later caller mutation cannot alter configuration. Raw index pairs could
remain only as an explicitly advanced/experimental construction path if there
is a demonstrated need.

The helper is therefore moving in the right direction, but the interface would
be deeper if it hid the dual coordinate/DOF bookkeeping instead of asking users
to unpack the result again.

## Fit

**Verdict: Fits if narrowed.**

Portable, broadly reusable joint impedance belongs in Newton. The ownership
boundary was explicitly settled in PR #3598, and using a finalized `Model`,
compact ports, and controlled-articulation packing is a coherent improvement.

Conditions:

- Copy the selection configuration so it remains stable after construction.
- Put APIC deployment evidence in-tree.
- Make the catalog example demonstrate the advertised same-simulation-model
  path, or explicitly classify its duplicate-model setup as an advanced
  scenario and provide a canonical normal-path example.

USD-schema disposition is not applicable: this is runtime control logic and
array selection, not user-authorable/importer configuration.

## Consistency with the proposed PR #3922 guidelines

PR #3922 is still an open draft at
`0851ff17c3e15c368848620cbc1a2e39cf3a45e6`; these are proposed-policy
findings, not violations of merged policy.

### P2 — The new autosummary option is unsupported in a locked environment

`docs/api/newton_controllers.rst:34-36` adds `:signatures: long`, which requires
Sphinx 8.2. The project still supports Python 3.10, for which `uv.lock` resolves
Sphinx 8.1.3. Current Python 3.12 docs CI passes, but the supported-environment
contract is inconsistent.

### P2 — New public count properties lack return-contract docstrings

Seven new public count properties have no property docstrings of their own:

- `newton/_src/controllers/impl/joint_impedance/model_based.py:393-407`
- `newton/_src/controllers/impl/joint_impedance/model_free.py:280-290`

The proposed review guidelines require checking the behavior, return contract,
documentation, and generated reference for changed public APIs.

### Design question — Public naming and concept-family prefix

`JointSelection` and `select_joints` are introduced specifically for
`ControllerJointImpedance`, but their public names do not use the `Controller*`
concept-family prefix emphasized by the proposed naming guidance. This is not a
definite violation, but the PR should either justify the broader names or
tighten them.

Exports, `__all__`, generated API inclusion, experimental status, Warp notation,
SI units, SPDX years, and migration fragments otherwise align with the proposed
guidelines.

## Validation at the reviewed head

- Focused command: `uv run --extra dev -m newton.tests -k test_controllers`
- Local result: 73/73 tests passed, including the heterogeneous CUDA
  graph-capture example.
- `git diff --check` passed.
- Hosted checks: 30 successful, 2 skipped; every required check passed.
- Review threads: 7 total, 0 unresolved; 2 resolved/current and 5
  resolved/outdated.

## Current axis summary

- Standards: 2 findings; worst issue is inaccurate public type contracts.
- Spec: 3 findings; worst issues are the broken fleet-selection example and
  mutable controller selection.
- Fit: **Fits if narrowed**.
- PR #3922 comparison: 2 concrete P2 findings and 1 naming question.
- Architecture: 2 non-blocking interface questions for discussion.

jeff-hough:
Thanks @jcarius-nv for the feedback! I will take a look into the `joint_selection` bugs - but to respond to some of the higher-level comments:

>The earlier controller discussion established a portable controller layer that
can run across physics engines and, through APIC, on hardware. Against that
goal, requiring a Newton Model at the portable controller seam would be the
wrong ownership choice.

I would mostly disagree with this comment:
### in the case of hardware deployment
* the `newton.Model` functions I am using are fully graph-capturable. Therefore, when exporting the controller to hardware, you retain the exact `newton.Model` in a fully portable way. You don't have to instantiate this model in C++, it is just "in" the `.wrp` file.

### in the case of non-newton sim applications
* If the application is using this controller, then by definition they have already installed `newton` into their environment. 
* We require a model of the robot in this version of the controller - that is fundamental to calculating fk, mass matrices etc; the `newton.Model` is the only logical choice. The other option would be to accept any class which meets some interface, and I think that could quickly become unusable (i.e. we might end up with requirements on quaternion ordering, reference frames for Jacobians, ordering of twist/wrench vectors, units, etc).
* The application will likely have robot models stored in either `usd` or `urdf`, in which case they build the model directly from that file with very little knowledge of newton required.

To address your point though - i.e. that the user may have solely installed `newton` to use this controller and they do not otherwise have any knowledge of `newton` or the `newton.Model`, we could add a `ControllerJointImpedance.from_urdf()`/`ControllerJointImpedance.from_usd` type helper, such that the user themselves never has to build the independent `newton.Model`

> Should callers supply two parallel index arrays?

This is a good idea - I prefer the idea of passing the `JointSelection` directly. I will update to that. If a user wants to pass their own custom indexes for whatever reason, then they can easily create their own custom `JointSelection` instead of using the helper.

## On the `ModelFree` naming:
> ModelFree is also slightly misleading because the caller may still supply a
> model-derived mass matrix and compensation terms; the distinction is really
> caller-supplied versus Newton-computed dynamics.

This is fair enough; personally I am still quite comfortable calling this controller `ModelFree`, as the controller _is_ model-free. But as you say, we do need mass-matrix, gravity and Coriolis vectors - and of course, these have to be computed from _some model_ which is external to the controller.

Is there a strong preference for a different naming scheme? For example `*ModelInteral` and `*ModelExternal`?

## On the conditions
> 1. Copy the selection configuration so it remains stable after construction.
> 2. Put APIC deployment evidence in-tree.
> 3. Make the catalog example demonstrate the advertised same-simulation-model path, or explicitly classify its duplicate-model setup as an advanced scenario and provide a canonical normal-path example.

For 1 and 3 I completely agree. For 2, I didn't do this only because I felt it is too large a PR. I would still prefer to do this in a separate PR. But let me know if this is your strong preference.

## On the naming of `select_joints` and `JointSelection`
> JointSelection and select_joints are introduced specifically for
ControllerJointImpedance, but their public names do not use the Controller*
concept-family prefix emphasized by the proposed naming guidance

To me, the name `ControllerJointSelection` makes the class sound like it is a controller? Is it not sufficient that it comes packaged in the `newton.controllers` module?

jcarius-nv:
Thanks for the elaborate response!

>we could add a ControllerJointImpedance.from_urdf()/ControllerJointImpedance.from_usd type helper, such that the user themselves never has to build the independent newton.Model

That's an interesting idea. Definitely to be looked into in a follow-up PR, but it would align with our goal to allow (not mandate!) more of Newton to be data-defined via a USD-file. Do you see cases, for example in IsaacSim where it would be convenient to have the robot model and its corresponding controller be directly defined in USD so you don't need to do this programatically? I'm not sure yet how useful it is if the user then anyways needs to instantiate the class to set inputs and such, but a good thing to explore nonetheless.

I agree otherwise with your argument that the newton model is a good interface for configuring the controller, so no changes needed there.

ModelFree naming: seems fine, I don't think the alternatives are better. So let's leave it :)

APIC deployment: Sure, follow-up PR is good. Smaller PRs are always better for reviews.

> the name ControllerJointSelection makes the class sound like it is a controller? 

that's true. My point was mainly that you don't always see the module name when looking at documentation or for example doing a `rg` search through the codebase, so having an expressive name, independent of where it's nested in the hierarchy is generally helpful. We can leave it and see if we find a unifying prefix as we add more controllers.

jcarius-nv (changes requested):
Re-review at `a01d0e643bd585143015749d9697d2ee153cc261`.

Thanks—the recent updates address the mutable caller-owned indices, give the controller one `JointSelection` boundary, fix leaf-name matching across prefixed fleets, and move the scalar-joint restriction out of `select_joints` as discussed.

The earlier review already records the `articulations="panda_.*"` glob problem and the catalog example's duplicate-model condition, so I am not duplicating those as new findings here. Three Spec issues remain:

### P1 — Update the PR body's runnable example to the current API

The example still passes the removed `joint_q_idx=` and `joint_qd_idx=` constructor arguments. It must pass `joint_selection=selection`. It also still contains the previously reported `articulations="panda_.*"`; ordinary strings use glob syntax against full articulation labels such as `panda_0/panda`, so this needs an actual glob such as `panda_*` or a compiled regular expression.

### P2 — Do not classify loose joints as members of the next articulation

`select_joints()` builds `joint_to_art` with:

```python
np.searchsorted(art_end, np.arange(model.joint_count), side="right")
```

This ignores `articulation_start`. I reproduced a model with a loose joint before articulation 0 where `model.joint_articulation == [-1, 0]`; `select_joints(model, articulations=[0], joints=[loose_joint])` nevertheless succeeds and returns the loose joint. This contradicts the documented requirement that explicit joints belong to a selected articulation.

Please derive ownership from `model.joint_articulation` (or validate both articulation boundaries) and add a regression test for a loose joint before/between articulations.

### P2 — Reconcile “all joints” with the `JointSelection` representation

Following the Slack discussion, `select_joints(model)` now selects every joint and leaves controllability validation to `ControllerJointImpedance`. However, `JointSelection` documents `q_idx` and `qd_idx` as one valid entry per controlled DOF, while the helper appends only one coordinate/DOF *start* per joint.

For a fixed-only articulation I reproduced `q_idx == [0]` and `qd_idx == [0]` even though both model spaces have length zero. Those values cannot be used as the documented indexed-view subscripts, and the controller raises an out-of-range error before reaching its intended one-coordinate/one-DOF validation. Multi-DOF joints have the related problem that one start-index pair does not represent all coordinates or DOFs.

Please make joint identity/span information sufficient to validate fixed and multi-DOF selections before deriving the scalar controller mappings, or narrow and correct the public contract so every returned index remains valid. Filtering unsupported joint types inside the helper would undo the intentional separation agreed in Slack.

Focused evidence at this head remains otherwise clean: 20/20 selection tests and 59/59 controller tests passed locally on CPU/CUDA, and every required hosted check is green.

newton/_src/controllers/impl/joint_impedance/model_based.py
```diff
@@ -89,70 +113,53 @@ class Inputs:
...
-                f"ControllerJointImpedance only supports 1-DOF joints (Revolute/Prismatic) and "
-                f"zero-DOF fixed joints; found unsupported joint types: {unsupported_joints}"
-            )
+        if not isinstance(model, Model):
+            raise TypeError(f"model must be a newton.Model, got {type(model).__name__}.")
+        model_robot_count = model.articulation_count
+        if model_robot_count < 1:
+            raise ValueError("model has no articulations.")
+
+        self._device = wp.get_device(device)
```
eric-heiden: **[P2] Derive this configuration from `model`**

This model-backed controller immediately requires `device == model.device` and `requires_grad == model.requires_grad`, so accepting both values from the caller creates redundant sources of truth. In particular, omitting `device` resolves against the current Warp device and can reject a model created on another device, while the default `requires_grad=False` rejects a differentiable model unless the caller repeats its setting. Please remove both constructor arguments, set `self._device = model.device` and `self._requires_grad = model.requires_grad`, and use those derived values for the internal state and allocations.

eric-heiden (commented):
Follow-up API review after the latest update.

newton/_src/controllers/joint_selection.py
```diff
@@ -0,0 +1,204 @@
...
+                    )
+                matched_arts.append(entry)
+        # An index and a label can name the same articulation; selecting it
+        # twice would duplicate every one of its joints in the output.
+        selected_arts = sorted(dict.fromkeys(matched_arts))
+
+    robot_joints_by_art: dict[int, list[int]] = {art: [] for art in selected_arts}
+    if joints is None:
+        for art in selected_arts:
+            robot_joints_by_art[art] = np.arange(art_start[art], art_end[art]).tolist()
```
eric-heiden, line 169: **[P1] Make the default select only controllable joints.** With `joints=None`, this includes every joint in the selected articulations; the later filter removes only zero-coordinate/zero-DOF joints, so Free and Ball joints survive and are then rejected by `ControllerJointImpedance` for not being 1-coordinate/1-DOF. As a result, the default constructor fails on common floating-base robots even though the controller docs say unsupported joints may remain uncontrolled. The implicit/default path should skip non-1x1 joints, while explicitly naming one can continue to raise.

newton/_src/controllers/impl/joint_impedance/model_based.py
```diff
@@ -89,71 +122,48 @@ class Inputs:
...
         self,
-        builder: ModelBuilder,
+        model: Model,
         *,
-        default_dof_indices: wp.array[wp.uint32],
-        stiffness: wp.array2d[wp.float32] | None,
-        damping: wp.array2d[wp.float32] | None,
+        articulations: list[int | str | re.Pattern[str]] | str | re.Pattern[str] | None = None,
+        joints: list[int | str | re.Pattern[str]] | str | re.Pattern[str] | None = None,
+        stiffness: wp.array[wp.float32] | None,
```
eric-heiden: **[P2] Allow gains without precomputing the private selection size.** The controller resolves `articulations`/`joints` internally, but a caller supplying baked gains must allocate these compact arrays before construction and therefore already know `total_controlled_dofs`. The revised PR example has to hardcode `512 * 7`, and heterogeneous pattern selections cannot generally compute this without duplicating the private resolver. Please support scalar gains with internal broadcasting (and retain compact arrays for per-DOF gains), or otherwise provide a post-resolution way to bind static gains.

jcarius-nv (approved):

PR #3982, opened 2026-08-20
Fix heightfield contact using prism skirt instead of top face

Double7sBurger:
A heightfield triangle is extruded 1 m along -Z into a solid prism (`support_map` in
  `newton/_src/geometry/support_function.py`) so GJK/MPR can resolve shapes reaching its
  back side. The skirt and bottom cap are interior to the terrain and cannot carry a
  contact, but once the other shape's center passes below the top face the portal can
  settle on them. The reported penetration then jumps to roughly the extrusion depth with
  a normal of (0, 0, -1), so the solver drives the shape further into the terrain instead
  of separating it, and the next substep reads an even deeper overlap. 

  Penetration depth is the minimum support width over all directions, so the width along
  the triangle's face normal bounds the true depth from above. This PR takes it when it is
  the smaller of the two, which discards the skirt solution and leaves the set of colliding
  pairs unchanged. The change is scoped to `GeoTypeEx.TRIANGLE_PRISM`; the triangle-mesh
  path is untouched.

  Targeting `release-1.5` because that is what Isaac Lab pins. `mpr.py` is identical on
  `main`, so the commit cherry-picks cleanly if you would rather land it there first.

  ## Checklist
 
  - [ ] Changelog fragment — `changelog/` does not exist on `release-1.5`; happy to add one
        if this is retargeted at `main

 ## Test plan

`  uv run --extra dev -m newton.tests -k test_heightfield_box_penetration_uses_top_face`
 
  Fails on `release-1.5` — all four subtests, e.g. `-1.0190 m` reported for a `0.0210 m`
  overlap — and passes with this change.

  ## Bug fix

  **Steps to reproduce:**

  1. Build a flat heightfield and rest a thin box on it.
  2. Lower the box past its own half-thickness.
  3. Observe the reported penetration jump to about the 1 m extrusion depth, with the
     contact normal pointing into the terrain.

  **Minimal reproduction:**

  ```python
  import numpy as np
  import warp as wp

  import newton

  HALF_Z = 0.02
  DEPTH = 0.021  # just past the box's half-thickness

  elevation = np.zeros((21, 21), dtype=np.float32)
  builder.add_shape_heightfield(
      heightfield=newton.Heightfield(
          data=elevation, nrow=21, ncol=21, hx=1.0, hy=1.0, min_z=0.0, max_z=1.0
      )
  )   
  body = builder.add_body(xform=wp.transform((0.0, 0.0, HALF_Z - DEPTH), wp.quat_identity()))
  builder.add_shape_box(body=body, hx=0.2, hy=0.065, hz=HALF_Z)

  model = builder.finalize()
  pipeline = newton.CollisionPipeline(model, requires_grad=True)
  pipeline.collide(model.state(), contacts)

  n = int(contacts.rigid_contact_count.numpy()[0])
  d = contacts.rigid_contact_diff_distance.numpy()[:n]
  k = int(np.argmin(d))
  print(d[k], contacts.rigid_contact_normal.numpy()[:n][k])
  # release-1.5: -1.019  [ 0.  0. -1.]
  # this PR:     -0.021  [ 0.  0.  1.]
 ```
  
  
  Downstream effect. On Isaac Lab Isaac-Velocity-Rough-G1 / UnitreeA1 / UnitreeGo2 all had NaN issues by default. After changing they all complete the full training loop.

### DIFF
--- a/newton/_src/geometry/mpr.py
+++ b/newton/_src/geometry/mpr.py
@@ -546,6 +546,34 @@ def create_solve_mpr(support_func: Any, _support_funcs: Any = None):
                     point_a = alpha * vert_a(v1) + beta * vert_a(v2) + gamma * vert_a(v3)
                     point_b = alpha * v1.B + beta * v2.B + gamma * v3.B
 
+                    # A heightfield triangle is extruded 1 m along -Z into a solid prism
+                    # (see support_map) so GJK/MPR can resolve shapes reaching its back side.
+                    # Only the top face is a real surface; the skirt and bottom cap are inside
+                    # the terrain. Once B's center passes below the face the portal can settle
+                    # on the skirt, reporting about a metre of penetration along -Z, which
+                    # pushes B further in instead of separating it. Penetration depth is the
+                    # minimum support width over all directions, so the width along the face
+                    # normal bounds the true depth from above; taking it when it is smaller
+                    # discards the skirt solution without changing which pairs collide.
+                    if geom_a.shape_type == int(GeoTypeEx.TRIANGLE_PRISM):
+                        prism_n = wp.cross(geom_a.scale, geom_a.auxiliary)
+                        prism_n_sq = wp.length_sq(prism_n)
+                        if prism_n_sq >= 1.0e-24:
+                            prism_n = prism_n / wp.sqrt(prism_n_sq)
+                            # The extrusion is along -Z in this frame, so the outward face
+                            # normal is +Z whatever the triangle's winding.
+                            if prism_n[2] < 0.0:
+                                prism_n = -prism_n
+                            prism_sv = mpr_support(
+                                geom_a, geom_b, prism_n, orientation_b, position_b, extend, data_provider
+                            )
+                            prism_pen = wp.dot(prism_sv.BtoA, prism_n)
+                            if prism_pen < penetration:
+                                normal = prism_n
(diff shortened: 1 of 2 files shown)


nvtw:
Attempt to fix the same problem but including cleanup in mpr.py where shape specific code has started to accumulate (which is a non-goal): https://github.com/newton-physics/newton/pull/3983

eric-heiden:
Thanks for finding this and for the clear write up. The bug was real: once a body's center dropped below the heightfield surface, the contact could settle on the extruded prism skirt instead of the top face, so the normal pointed down and pushed the body deeper into the terrain instead of out of it.

Closing this in favor of #3983, which is already merged and fixes the same bug. That change takes a broader approach: it moves the heightfield specific logic out of `mpr.py` and into `support_function.py`, which drops about 226 lines from `mpr.py` and keeps the general MPR code free of per shape special cases. It also landed with wider tests for the same behavior (box, sphere, sloped, and rotated cases, plus heightfield boundary handling), so the test added here is covered too.

One thing to be aware of: this PR targets `release-1.5`, and #3983 was merged into `main`. The fix is not on `release-1.5` today. If 1.5.x needs it, the next step is a backport of #3983 to that branch rather than merging this one.

Thanks again for the reproduction test. It made the problem easy to confirm.

PR #3987, opened 2026-08-20
Negotiate the GL window config at the scene sample count

rakhimovv:
## Description

`RendererGL` asks pyglet for an 8x multi-sampled window config, then renders the scene
into offscreen renderbuffers allocated at a hardcoded 4x:

```python
config = pyglet.gl.Config(sample_buffers=1, samples=8, double_buffer=True)
...
self.msaa_samples = 4
```

The window's own buffer never carries the scene. `render()` draws into `_frame_msaa_fbo`
and resolves that into `_frame_fbo`'s texture, and it is the texture that gets drawn to
the window, so `msaa_samples` — not the window config — decides how the scene is
anti-aliased.

Two effects follow:

- An 8x default framebuffer is negotiated and allocated that nothing reads.
- A driver that cannot match `samples=8` raises `NoSuchConfigException`, and the handler
  for it set `msaa_samples = 0`. Hardware that offers 4x but not 8x therefore lost scene
  anti-aliasing entirely, even though the offscreen framebuffer does not depend on the
  window config at all.

This requests the window config at the count the scene actually uses — `MSAA_SAMPLES`,
still 4, so rendered output is unchanged wherever the 8x request already succeeded — and
keeps that count when no multi-sampled config is available. `_setup_frame_buffer` already
disables MSAA when the offscreen framebuffer comes back incomplete, which is the case
that genuinely warrants it.

Happy to split the fallback half out into its own PR if you would rather take the two
changes separately.

## Checklist

## Test plan

```
PYGLET_HEADLESS=1 uv run --extra dev -m newton.tests -k TestViewerGLMultisampling  # Ran 2 tests, OK
PYGLET_HEADLESS=1 uv run --extra dev -m newton.tests -k viewer                     # Ran 229 tests, OK
uvx pre-commit run -a                                                      # all hooks pass
```

Reverting `newton/_src/viewer/gl/opengl.py` alone and keeping the new tests fails both of
them, each on the assertion that names the defect:

```
test_window_config_requests_the_scene_sample_count
    AssertionError: 8 not found in (0, 4)
test_scene_keeps_msaa_without_a_multisampled_window_config
    AssertionError: 0 not greater than 0
```

Measured on a display-less Linux box, NVIDIA H100, pyglet 2.1.14 on the EGL headless
backend.

## Bug fix

**Steps to reproduce:**

1. Construct a `ViewerGL` on any machine where the driver grants the requested 8x config.
2. Compare the sample count the window was granted with the count the scene is rendered at.
3. They disagree: the window carries 8x, the scene is drawn at 4x.

**Minimal reproduction:**

```python
import pyglet

pyglet.options["headless"] = True  # only needed on a display-less machine

import newton

viewer = newton.viewer.ViewerGL(width=64, height=48, headless=True)
print(viewer.renderer.window.config.samples)  # 8  <- negotiated, and unused by the scene
print(viewer.renderer.msaa_samples)           # 4  <- what the scene is actually rendered at
viewer.close()
```

The second half — losing MSAA outright when the request cannot be matched — is what a
driver without an 8x config hits. Standing in for such a driver by asking for a config no
implementation can provide shows both the failure and that the offscreen path never needed
the window config:

```python
from pyglet import gl
import pyglet

try:
    cfg = gl.Config(sample_buffers=1, samples=128, double_buffer=True)
    pyglet.window.Window(width=32, height=32, visible=False, config=cfg)
except pyglet.window.NoSuchConfigException:
    print("no matching config")  # <- this is the branch that used to set msaa_samples = 0

w = pyglet.window.Window(width=32, height=32, visible=False)
print(w.config.samples)  # 0 -- no multi-sampling on the default framebuffer

rb = gl.GLuint()
gl.glGenRenderbuffers(1, rb)
gl.glBindRenderbuffer(gl.GL_RENDERBUFFER, rb)
gl.glRenderbufferStorageMultisample(gl.GL_RENDERBUFFER, 4, gl.GL_RGB8, 32, 32)

fbo = gl.GLuint()
gl.glGenFramebuffers(1, fbo)
gl.glBindFramebuffer(gl.GL_FRAMEBUFFER, fbo)
gl.glFramebufferRenderbuffer(gl.GL_FRAMEBUFFER, gl.GL_COLOR_ATTACHMENT0, gl.GL_RENDERBUFFER, rb)
print(gl.glCheckFramebufferStatus(gl.GL_FRAMEBUFFER) == gl.GL_FRAMEBUFFER_COMPLETE)  # True
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved scene anti-aliasing when multisampled window configurations are unavailable.
  * Improved multisampling setup by requesting the scene’s rendering sample count when supported.
  * Improved shadow rendering by aligning lighting with the scene rather than the camera position.
  * Improved rendering updates when the viewport or camera position changes.

* **Tests**
  * Added coverage for multisampling configuration and fallback behavior.


newton/tests/test_viewer_gl_msaa.py
eric-heiden, line 1: We don't need a separate test file for this, please remove it.

newton/_src/viewer/gl/opengl.py
```diff
@@ -1044,11 +1050,11 @@ def __init__(self, title="Newton", screen_width=1920, screen_height=1080, vsync=
                 visible=not headless,
                 config=config,
             )
-            gl.glEnable(gl.GL_MULTISAMPLE)
-            # remember sample count for later (e.g., resolving FBO)
-            self.msaa_samples = 4
         except pyglet.window.NoSuchConfigException:
-            print("Warning: Could not get MSAA config, falling back to non-AA.")
+            # Only the window's own buffer loses multi-sampling here, so the
+            # scene keeps its MSAA framebuffer. _setup_framebuffer still turns
```
eric-heiden: typo: The function is called `_setup_frame_buffer()`.

newton/tests/test_viewer_gl_msaa.py
rakhimovv: Moved both tests into `newton/tests/test_viewer_get_frame.py` and dropped the file. They reuse that module's `_make_headless_viewer_gl_or_skip()`, which already skips on the backend errors a machine without a usable GL display raises.

newton/_src/viewer/gl/opengl.py
rakhimovv: Fixed, in both comments that named it.

rakhimovv:
Refreshed this branch so it is easy to review against recent `main`.

Rebased onto `b74df534`. No conflicts — #4006 touched `newton/_src/viewer/gl/opengl.py` as well, but in the shadow and shader path, which does not overlap the window-config change here.

Also picked up the pre-merge review points:

- `window.config.samples` is `None` on a backend that leaves the attribute unset, which the membership check would have read as a mismatch, so it is normalized before comparing. On the headless backend here the config-less window reports `0`, so this is hardening for backends I cannot exercise rather than a live failure.
- Added a docstring to the window stand-in in the fallback test.
- Dropped a `return` after `skipTest`, which never runs because `skipTest` raises.

Re-verified on a display-less Linux box with H100s:

```
uv run --extra dev -m newton.tests -k test_viewer_get_frame
Ran 11 tests ... OK

uvx pre-commit run -a
all hooks pass
```

No rush from my side — just flagging that CI here is still waiting on the external-PR approval gate, in case that is easy to release.

rakhimovv:
Picked up the remaining review points on the two MSAA tests in `45d50d9`. Kept it to the tests — no new helpers or scaffolding, and they still live in `test_viewer_get_frame.py`.

- **The scene-MSAA assertion was host-dependent.** It asserted only `msaa_samples > 0`, but `_setup_frame_buffer` zeroes that itself when the driver cannot complete the multi-sampled framebuffer, so a valid environment could fail the test. The two cases are separable: the framebuffer object is created *before* its completeness is checked, so a missing one means the window-config fallback disabled the scene's MSAA — the regression under test, and a failure — while a zeroed count beside a live framebuffer is the driver's own limit, and skips.
- **Neither test noticed if the multi-sampled window request were deleted outright.** The stand-in now records what it was asked for, and the test compares that with the scene's sample count.
- **The `import pyglet.window` guard swallowed everything** into a skip. It now classifies through the module's existing `_is_viewer_gl_unavailable_error()` and re-raises anything else.

One suggestion I did not take: normalizing `config.samples` through `.value` for ctypes. I checked pyglet 2.1.14 rather than assume — every backend stores a plain `int` (`xlib`, `headless` and `cocoa` assign `value.value`/`vals.value`; `win32` iterates a `c_int` array, which yields ints), and measured here `repr(window.config.samples)` → `4 <class 'int'>`. The existing `or 0` normalization covers the one real case, a backend that leaves the attribute unset. Adding `.value` handling would be dead code.

Re-verified on a display-less Linux box, with the renderer reverted to its pre-PR state to confirm the tests still fail without the change:

```
before: FAILED (failures=2)
  test_scene_keeps_msaa_...  AssertionError: unexpectedly None
  test_window_config_...     AssertionError: 8 not found in (0, 4)
after:  both ok
uvx pre-commit run -a  →  all hooks pass
```

(The unrelated `test_headless_capture_main_image_frame` stalls intermittently on this particular machine inside the CUDA↔GL buffer map — a known bad card here, not something this branch touches.)

Still nothing red on my side, and the branch is mergeable. `Require Manual Approval for External PRs` has not run since the PR was opened, so CI has never actually executed against it — if a Project Member could release that, it would be good to see it green rather than pending.

PR #3988, opened 2026-08-20
Update AOUSD deformable material API

### LINKED ISSUE #3809: [REQ] Update USD Deformables based on latest proposal
### Description

The [AOUSD proposal ](https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/111) has been updated to include Young's Modulus and Poisson Ratio on all deformable materials, as well as a units changes for the various stiffnesses and some attribute renames in [61d83b54b7efbe97ad2f480de885255cd1e593be](https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/111/commits/61d83b54b7efbe97ad2f480de885255cd1e593be). See commit message & diff for complete details.

There will also be another change coming which re-names one of the APIs (commit not yet posted).

We should re-align our parsing with the proposal & cite the commit we aligned to.

### Motivation / Use Case

Parsing UsdPhysics Deformables needs to align with AOUSD standards

### Alternatives Considered

We could maintain backwards compatibility for the current implementation, but I fear that gets quite complicated, especially with the units change on the stiffnesses. Since its marked experimental I'd suggest we just adapt to the changes & explain in the changelog.

mzamoramora-nvidia:
## Description

Closes #3809.

Align the experimental USD deformable importer with AOUSD deformables proposal
commit
[`63c74d79`](https://github.com/PixarAnimationStudios/OpenUSD-proposals/commit/63c74d79aa67dc98d8707af8ad68f2e1a10622b5).
That revision moves surface and curve thickness from materials to typed arrays on
the simulation geometry, makes non-empty mass arrays require an element type,
and defines how element and point masses are converted before lumping to the
simulation points.

This PR:

- reads geometry-owned `physics:thicknesses` with the required
  `physics:thicknesses:elementType` for surfaces and curves;
- applies constant and varying thickness to mass, collision radius, and any
  stiffness derived from Young's modulus and Poisson's ratio, while keeping
  cable collision-radius caches and attachment-local weld decisions consistent
  with the resolved local radii;
- reads typed `physics:masses` for volume, surface, and curve deformables and
  converts constant, simplex, curve, segment, and point values through the
  proposal's element-mass model;
- follows the proposal's body mass, body density, material density, and final
  1000 kg/m³ density precedence;
- reads simulation geometry fallbacks at the USD default time and uses
  `physics:restBendAnglesDefault = "flat"` when cloth rest bend data is absent;
- keeps the family-prefixed structural stiffness API introduced by the earlier
  proposal revision, including independent isotropic fallbacks for missing
  surface and curve modes; and
- adds actionable diagnostics for malformed array/type pairs, invalid physical
  values, and rest topology that Newton cannot currently represent.

Proposal fallbacks apply only when the corresponding value is unauthored or
uses its defined sentinel. Valid values explicitly authored through the
simulation, body, or bound material APIs retain their documented precedence.

For unauthored cable thickness, AOUSD's 1 mm diameter (0.5 mm radius) fallback
replaces Newton's previous 2.5 mm radius. To preserve the previous physical
dimensions, author a 5 mm diameter expressed in stage units as a one-element
`physics:thicknesses` array with
`physics:thicknesses:elementType = "constant"`.

The removed material `physics:surfaceThickness` and
`physics:curvesThickness` attributes remain accepted during a deprecation
window as thickness fallbacks. Earlier unprefixed material attributes and
untyped mass arrays also remain compatible with their previously released
interpretation and emit `DeprecationWarning`. Current geometry-owned arrays take
precedence when both revisions are authored.

The proposed core APIs are not registered by OpenUSD or by the packaged
`newton-usd-schemas` 0.5.x dependency, so Newton resolves their attributes by
token name. Numeric scalar and array values are validated against the
proposal's USD `float` storage and semantic ranges. This PR intentionally does
not introduce a general policy for derived-value overflow in Newton's float32
simulation storage. Such a policy should cover masses, inertia, stiffness,
geometry scaling, and other USD import paths consistently in separate work.
The importer remains experimental, and no schema-package dependency change is
required for this proposal revision.

Newton's current isotropic cloth representation cannot apply
`physics:surfaceShearStiffness` independently. The importer warns and preserves
the authored value in `path_cloth_attrs`. Other supported-subset limitations are
documented in `docs/concepts/usd_parsing.rst`.

### Open question: keep the deprecation window?

This branch preserves the previously released experimental spellings behind
`DeprecationWarning`, following Newton's policy that breaking changes receive a
deprecation first. Issue #3809 notes that the deformable importer is
experimental, so maintainers may prefer a direct cutover.

Maintainer guidance is requested before merge:

1. **Keep the compatibility path implemented here.** Existing assets retain
   their former interpretation while receiving actionable migration warnings.
2. **Treat the experimental importer as an exception and cut over directly.**
   Remove the legacy readers, warnings, compatibility tests, and deprecation
   fragment from this PR.

Unless the experimental API is explicitly exempted from the repository's
deprecation policy, this PR defaults to option 1.

## Checklist

## Test plan

The focused USD-deformable test slice covers the three family importers and
their shared attachment, filtering, grouping, and mixed-scene behavior:

```console
uv run --extra dev -m newton.tests -k usd_deformable --serial-fallback --no-cache-clear
# Ran 160 tests: OK (skipped=1)
```

Documentation, changelog rendering, and repository checks:

```console
uv run --extra dev sphinx-build -W --keep-going -b html docs docs/_build/html
uv run --extra dev sphinx-build -W --keep-going -b doctest docs docs/_build/doctest
uvx --from towncrier==25.8.0 towncrier build --draft \
  --version X.Y.Z --date 2026-08-26
uvx pre-commit run -a
```

CI is responsible for the full Newton test matrix.

## New feature / API change

Thickness and mass now belong to the simulation geometry and require an element
type when non-empty:

```usda
def Mesh "Cloth" (
    prepend apiSchemas = ["PhysicsSurfaceDeformableSimAPI"]
)
{
    float[] physics:thicknesses = [0.001]
    token physics:thicknesses:elementType = "constant"
    float[] physics:masses = [2.0]
    token physics:masses:elementType = "constant"
}

def Material "ClothMaterial" (
    prepend apiSchemas = ["PhysicsSurfaceDeformableMaterialAPI"]
)
{
    float physics:surfaceStretchStiffness = 100
    float physics:surfaceShearStiffness = 40
    float physics:surfaceBendStiffness = 1e-8
    float physics:density = 1000
}
```

Alternatively, omit any structural mode to derive it independently from the
shared isotropic material properties:

```usda
def Material "DerivedClothMaterial" (
    prepend apiSchemas = ["PhysicsSurfaceDeformableMaterialAPI"]
)
{
    float physics:youngsModulus = 1000000
    float physics:poissonsRatio = 0.3
}
```

To migrate from the earlier proposal revision:

- move `physics:surfaceThickness` from the material to geometry-owned
  `physics:thicknesses` and author the corresponding element type;
- migrate scalar `physics:curvesThickness = d` to geometry-owned
  `physics:thicknesses = [d]` with
  `physics:thicknesses:elementType = "constant"`;
- add `physics:masses:elementType` to every non-empty mass array and ensure all
  typed values are finite and strictly positive; one invalid value rejects the
  entire array and continues with lower-precedence mass sources;
- expect `physics:masses:elementType = "point"` conversion to preserve total
  mass while potentially changing its distribution relative to the deprecated
  direct-point interpretation; and
- convert earlier unprefixed material moduli to the corresponding
  family-prefixed structural stiffness attributes.


jumyungc:
Thanks @mzamoramora-nvidia!
Some cable side review:

1. `shape_collision_radius` is stale after changing segment radius 
`add_rod()` and `add_rod_graph()` initially create every shape with one representative radius (`import_usd_deformable_cable.py:714,1058`). `_set_cable_body_radius()` later corrects each segment’s `shape_scale` and inertia (`import_usd_deformable_utils.py:729-746`) but does not update `builder.shape_collision_radius[shape]`. That cached value is calculated during shape creation (`sim/builder.py:6688`) and copied unchanged into the finalized model (`sim/builder.py:11515`). The default collision path currently derives capsule AABBs from the corrected shape_scale, which may mask the issue there, but the finalized model still contains an inconsistent collision radius consumed by other paths.
Could _set_cable_body_radius() recompute shape_collision_radius and the local-thickness test assert it?

2. Welding tolerance uses a non-local radius
_CurveDeformableRecord.radius is initialized from segment_radii[0] at import_usd_deformable_cable.py:468. Line 524 then uses that value to test hard curve-to-curve point attachments, although attachment indices a and b may refer to points elsewhere on the curves.
With varying thickness, welding therefore depends on an unrelated first-segment radius and can change when the curve order is reversed. Could the tolerance be calculated per attachment pair from point_radii[a] and point_radii[b], with a varying-thickness welding test?

3. Document the changed fallback radius
At import_usd_deformable_cable.py:255, unauthored thickness now resolves to the AOUSD radius of 0.5 mm. Previously, cables with no material, or only a legacy material without thickness, used Newton’s 2.5 mm compatibility radius. Could changelog/3722.changed.1.md retain the removed “instead of Newton’s previous 2.5 mm radius” comparison and recommend explicitly authoring physics:thicknesses to preserve existing dimensions?

4. Clarify migration behavior
  Could the changelog or user documentation also state that:
  - Scalar physics:curvesThickness maps to a one-element physics:thicknesses array with elementType = "constant".
  - Typed masses must be strictly positive. One zero causes the entire array to be ignored and mass resolution to continue with lower-precedence sources.
  - For cables, elementType = "point" uses volume-weighted point-to-segment conversion. Total mass is preserved, but its distribution across segment bodies may change.
  - For proposal-marked cables without authored density, the AOUSD 1000 kg/m³ fallback is used instead of builder.default_shape_cfg.density.

jumyungc:
@AnkaChan for viz for surface / volume.

mzamoramora-nvidia:
Thanks @jumyungc for the detailed cable-side review. I went through all four points and
addressed them:

1. `_set_cable_body_radius()` now recomputes `shape_collision_radius` with Newton's
   canonical `compute_shape_radius()` helper immediately after updating `shape_scale`.
   `test_cable_segment_thickness_controls_local_physics` checks the cached value on both
   the builder and the finalized model for varying segment radii.

2. Curve-to-curve weld tolerance is now evaluated independently for each attachment pair
   from `point_radii[a]` and `point_radii[b]`, rather than either curve's first segment.
   `test_weld_tolerance_uses_attachment_point_radii` covers both inverse cases: a thin first
   segment with a thick attached point welds, while a thick first segment with a thin
   attached point is rejected.

3. `changelog/3722.changed.1.md` now calls out the change from Newton's previous 2.5 mm
   radius to AOUSD's 1 mm diameter / 0.5 mm radius fallback. It also explains how to retain
   the previous physical size by authoring a 5 mm diameter, expressed in stage units, as a
   one-element constant `physics:thicknesses` array.

4. The migration and fallback behavior is now explicit in the Towncrier fragments, the USD
   parsing concept page, and the existing warnings:

   - `physics:curvesThickness = d` maps to `physics:thicknesses = [d]` with
     `physics:thicknesses:elementType = "constant"`;
   - typed mass values must be finite and strictly positive, and one invalid value rejects
     the entire array before resolution continues with lower-precedence sources;
   - point mass conversion preserves total mass but can redistribute it across cable segment
     bodies relative to the deprecated direct-point interpretation; and
   - proposal-marked cables without authored density use the final 1000 kg/m³ fallback rather
     than `builder.default_shape_cfg.density`.

newton/_src/usd/utils.py
```diff
@@ -2350,29 +2454,29 @@ def _get_deformable_body_overrides(
...
-    Per-point masses take precedence over body and material mass/density (proposal
-    "Simulation Geometry and Rest Shape"). Returns ``None`` when unauthored/empty.
-    """
-    val = read_attr(prim, "masses")
-    if val is None:
+    def read_override(name: str) -> float | None:
+        raw_value = read_attr(body_prim, name)
+        if raw_value is None:
+            return None
+        value = float(raw_value)
```
jcarius-nv: **P1 — Validate wrong-declared scalar types before conversion.**

The raw-token compatibility path allows these attributes to arrive with a non-numeric USD storage type. Both this conversion and `_read_deformable_material()`'s `float(val)` currently propagate `TypeError`/`ValueError` instead of diagnosing malformed authoring and continuing with the next precedence/default source.

I reproduced this at the current head through `ModelBuilder.add_usd()` with string-typed `physics:youngsModulus = "bad"`; string-typed body `physics:mass` and `physics:density` fail the same way here. This contradicts the PR's stated validation/diagnostic contract and lets one malformed scalar abort an otherwise valid stage. Could conversion failures warn and fall through, with regression coverage for both material and body scalars?

newton/_src/utils/import_usd_deformable_volume.py
```diff
@@ -187,29 +202,41 @@ def _deformable_import_volume(ctx: _DeformableImportContext) -> None:
...
 
         soft_p0, soft_t0 = builder.particle_count, builder.tet_count
         builder.add_soft_mesh(**add_soft_mesh_kwargs)
+        authored_masses = None
         if is_volume_deformable:
-            _apply_particle_masses(builder, prim, soft_p0, builder.particle_count, deformable_read)
+            mass_indices = np.asarray(
+                add_soft_mesh_kwargs.get("indices", tetmesh_for_builder.tet_indices), dtype=np.int64
+            ).reshape(-1, 4)
+            mass_points = np.asarray(world_vertices, dtype=np.float64)
```
jcarius-nv, line 228: **P2 — Keep the rest-shape limitation explicit in the mass-compliance claim.**

The proposal defines element volume from the rest shape, while volume mass conversion here uses the imported/current `world_vertices`; cloth similarly uses current triangle areas. Because authored rest geometry is currently warned and ignored, constant/point typed masses and density-derived masses can be distributed differently for a saved deformed pose.

The concept page documents this limitation, which is good. Could the PR/changelog's claim that masses use the proposal element-mass model be narrowed accordingly, or could a concrete follow-up be linked for completing rest-shape-based mass distribution?

changelog/3809.changed.md
```diff
@@ -0,0 +1 @@
...ical fallback, and apply varying thickness to mass, collision radius, and derived stiffness. Read strictly positive typed `physics:masses` as constant, per-simplex, per-curve, or per-point values and convert them through element masses before lowering to Newton particles or cable segments. Ignore the entire typed array when any value is invalid and continue with lower-precedence mass sources; point conversion preserves total mass but can change its distribution relative to the deprecated direct-point interpretation. Missing cloth bend arrays now follow `physics:restBendAnglesDefault` (`flat` by default), and rest fallbacks read simulation attributes at the USD default time. Unsupported authored rest topology, an element-type token without its array, and invalid body mass/density overrides now warn and use safe fallbacks. Use the proposal's final 1000 kg/m³ density instead of Newton builder density defaults when a proposal-marked deformable has no authored density. Remove unsupported rest topology or bake it into default-time simulation topology, author array/type pairs together, and replace invalid physical values; valid assets within Newton's supported subset require no migration.
```
jcarius-nv: **P2 — Carry the complete migration and public-API behavior into the changelog.**

Thanks for the recent additions covering typed-mass rejection, redistribution, density fallback, and cable thickness migration. Two release-visible pieces still appear missing:

- `newton.usd.get_tetmesh()` and `TetMesh.create_from_usd()` now give a current volume material the proposal's 1 MPa/0.3 elasticity fallbacks where missing values previously remained `None`; and
- the complete cross-family migration guidance currently present in the PR description should live in the relevant Towncrier fragments, even if that makes them long.

Correction to my accompanying review: I initially misunderstood where the proposed deformable schemas are published. `newton-usd-schemas` provides Newton-owned extensions, not the AOUSD core `Physics*Deformable*` APIs; those currently exist only as draft definitions in [OpenUSD-proposals #111](https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/111), without a registered schema plugin. Therefore the current migration guidance should explain the raw-token authoring path rather than direct users to a nonexistent compatible core-schema package. Once an upstream implementation is published, the preferred migration can name the exact first OpenUSD/schema version that registers these APIs. Any `newton-usd-schemas` release should be referenced separately for updates to its Newton-owned extensions.

jcarius-nv (commented):
## Fit / upstream schema follow-up question

Correction: I initially misunderstood where the proposed deformable schemas are currently published. `newton-usd-schemas` 0.5.0 does not provide the AOUSD core `Physics*Deformable*` APIs; it only provides Newton-owned extensions such as `NewtonPointsDeformableSimAPI` and `NewtonCurvesDeformableMaterialAPI`. The core schema definitions currently exist as draft snippets in open [OpenUSD-proposals #111](https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/111), not as a registered OpenUSD schema plugin or package.

Given that, are there plans or a tracked follow-up for the upstream schema implementation and a schema-aware authoring/validation path, and what is currently blocking it? Is the direct token reader intended to bridge the period until the proposal is accepted and implemented upstream? Naming an owner and tracking point here would make that transition explicit. A registered author -> export -> Newton-import round trip can follow once such a schema implementation exists; the raw-token tests are appropriate coverage in the meantime.

Separately, could `newton-usd-schemas` be updated for the Newton-owned extensions that depend on the proposal? Its `NewtonCurvesDeformableMaterialAPI` documentation still describes the removed material-owned `physics:curvesThickness`, so its extension contract and documentation should be aligned with proposal revision `63c74d79`. That update should not be presented as providing registration for the AOUSD core APIs.

On compatibility, retaining easy-to-maintain aliases/readers for one release cycle seems proportionate given how new the schema is. I would not expect behavioral compatibility across the stiffness units change where the old and new meanings cannot be represented faithfully at once; that should instead produce an explicit migration warning and conversion guidance. It would be good to notify Lightwheel, ask them to validate the migration against their assets, and collect their feedback before removing the retained aliases.

mzamoramora-nvidia:
Thanks @jcarius-nv. I addressed the three inline comments:

- Non-numeric deformable material values and body `mass`/`density` now warn and use the documented fallback instead of aborting import. Public regressions cover `ModelBuilder.add_usd()` and `newton.usd.get_tetmesh()`.
- The changelog now says mass measures use default-time simulation geometry rather than rest geometry and links [newton#3383](https://github.com/newton-physics/newton/issues/3383).
- The Towncrier fragments now include the public TetMesh 1 MPa/0.3 fallback, the complete cross-family migration guidance, and the current raw-token authoring path.

The direct token reader is the bridge until the proposed core schemas are published. That work and the stale Newton-extension documentation will be tracked in [newton-usd-schemas#88](https://github.com/newton-physics/newton-usd-schemas/issues/88). The compatibility readers remain; Lightwheel should be consulted before they are removed.

newton/_src/utils/import_usd_deformable_cable.py
```diff
@@ -481,11 +564,12 @@ def _deformable_import_cable_graphs(ctx: _DeformableImportContext) -> tuple[set[
                 stacklevel=2,
             )
             continue
-        # A tenth of the thinner cable's radius: welding then moves geometry by well under the
-        # junction bodies' own overlap, so the weld is equivalent to the authored constraint.
-        coincidence_tol = 0.1 * min(curve_recs[src0].radius, curve_recs[src1].radius)
+        # A tenth of the thinner attached point's radius: welding then moves geometry by well
+        # under the junction bodies' own overlap, so the weld is equivalent to the authored
```
jcarius-nv, line 594: **P1 — Contain malformed attachment scalar storage types.**

The new coercion path covers material and body scalars, but this welded-graph path still calls `float(stiffness_val)` directly. The attachment post-pass does the same for both stiffness and damping in [`import_usd_deformable_attachments.py`](https://github.com/newton-physics/newton/blob/9a087fb963c1df18ddb9063b7f983b432a976e79/newton/_src/utils/import_usd_deformable_attachments.py#L102-L105). Because the proposed core schemas are unregistered and these values are read by token, a wrong-declared token/string value can reach either conversion.

I reproduced both cases at this head through `ModelBuilder.add_usd()` with `physics:stiffness = "bad"`: the normal attachment post-pass and a curve-to-curve welded-graph candidate each raise `ValueError`, aborting the entire stage import rather than warning and preserving/skipping the unsupported attachment. This also leaves an exception to the PR description's claim that proposal numeric scalars are validated against USD float storage and semantic ranges.

Could attachment stiffness/damping reuse the scalar coercion and diagnostic path before hardness/range validation, with regression coverage for both the welded-graph pre-pass and attachment post-pass?

jcarius-nv (commented):
One additional correctness finding from the current-head re-review.

newton/_src/usd/utils.py
mzamoramora-nvidia: Addressed in commit `066fb6b8`.

newton/_src/utils/import_usd_deformable_volume.py
mzamoramora-nvidia: Addressed in commit `36640c23`.

changelog/3809.changed.md
mzamoramora-nvidia: Addressed in commit `36640c23`.

newton/_src/utils/import_usd_deformable_cable.py
mzamoramora-nvidia: Addressed in commit `6b211ee3`.

newton/_src/utils/import_usd_deformable_cloth.py
```diff
@@ -234,7 +367,63 @@ def _deformable_import_cloth(ctx: _DeformableImportContext) -> None:
...
+        for element_mass, face in zip(density_element_masses, tri_faces, strict=True):
+            for point in face:
+                density_point_masses[int(point)] += float(element_mass) / 3.0
+        for point, mass in enumerate(density_point_masses):
+            builder.particle_mass[p0 + point] = mass
+
+        for tri_offset, face_thickness in enumerate(face_thicknesses):
+            resolved = _resolve_surface_structural_stiffnesses(surface_material, face_thickness, ctx.linear_unit)
+            stretch = builder.default_tri_ke if resolved is None or resolved[0] is None else resolved[0]
+            material = builder.tri_materials[t0 + tri_offset]
```
AnkaChan: **P1 — a single NaN vertex aborts the whole import.**

This loop assumes `add_cloth_mesh` kept every triangle, but `add_triangles` drops `area > 0` failures (builder.py:9147-9151), while the pre-filter at line 272 tests `area < 1e-12` — NaN fails both comparisons, the counts diverge, and this indexes out of range. Repro: a quad with one NaN point → `IndexError` here, killing the whole `parse_usd`; main imported the same asset (silent triangle drop). Contradicts the module's own "cannot crash the import" contract (lines 214-216).

Fix: NaN-safe pre-filter (`not (area >= 1e-12)`, or `np.isfinite` on the baked points), and/or guard `builder.tri_count - t0 == len(tri_faces)` after `add_cloth_mesh`.

newton/_src/utils/import_usd_deformable_utils.py
```diff
@@ -592,18 +732,148 @@ def _set_body_mass(builder: ModelBuilder, b: int, m: float) -> None:
...
+def _element_masses_from_points(
+    point_masses: Sequence[float], element_indices: Sequence[Sequence[int]], element_volumes: Sequence[float]
+) -> list[float] | None:
+    """Convert point masses to element masses using the proposal's volume-weighted relation."""
+    point_volumes = [0.0] * len(point_masses)
+    for indices, volume in zip(element_indices, element_volumes, strict=True):
+        share = float(volume) / len(indices)
+        for point in indices:
+            point_volumes[int(point)] += share
+    if any(volume <= 0.0 or not math.isfinite(volume) for volume in point_volumes):
```
AnkaChan: **P1 — one unreferenced vertex discards the whole authored point-mass array.**

An orphan vertex (legal USD, routine in DCC exports) keeps its `point_volumes` entry at 0.0, so this guard rejects the entire array. Repro: quad + 1 orphan, `elementType="point"`, masses `[2,2,2,2,2]` → array ignored, total mass 10.0 → 1.0 (density fallback), orphan particle mass 0.0 (kinematic). The legacy untyped path imports the same data exactly — and the deprecation warning (utils.py:134-142) steers those users into this break. Since the conversion never reads the orphan's mass (line 773), whole-array rejection is stricter than the division-safety it guards.

Suggest: reject only when a *referenced* point has non-positive/non-finite adjacent volume, skip orphans with a warning, and document topology-based rejection (usd_parsing.rst:208-215 covers only value-based).

newton/_src/utils/import_usd_deformable_cloth.py
```diff
@@ -234,7 +367,63 @@ def _deformable_import_cloth(ctx: _DeformableImportContext) -> None:
...
+            adjacent_faces = edge_face_indices[key]
+            if authored_thicknesses is not None and authored_thicknesses.element_type == "point":
+                edge_thickness = 0.5 * (point_thicknesses[key[0]] + point_thicknesses[key[1]])
+            else:
+                edge_thickness = sum(face_thicknesses[index] for index in adjacent_faces) / len(adjacent_faces)
+            resolved = _resolve_surface_structural_stiffnesses(surface_material, edge_thickness, ctx.linear_unit)
+            bend = builder.default_edge_ke if resolved is None or resolved[2] is None else resolved[2]
+            properties = builder.edge_bending_properties[edge_offset]
+            builder.edge_bending_properties[edge_offset] = (bend, properties[1])
+        if rest_bend_angles_default == "flat":
```
AnkaChan, line 467: **P2 — unauthored `restBendAnglesDefault` silently flattens curved cloth.**

On main, rest angles came from the mesh dihedrals, so a draped asset rested in its authored shape; this loop zeroes them with no signal. `flat` is the proposal's fallback, but nothing in the ecosystem authors this token yet, so its only effect today is rewriting the rest state of every existing curved asset.

My preference: keep Newton's previous initial-shape rest (`restShape` semantics) for unauthored data while the schemas are unregistered. If we track the proposal exactly instead, then at minimum fix changelog/3809.changed.md ("no migration required" isn't true here; authored `restShape` restores the old behavior) and warn once when the unauthored token flattens a non-planar cloth. Worth raising on OpenUSD-proposals #111 either way.

newton/_src/utils/import_usd_deformable_cloth.py
```diff
@@ -152,65 +297,53 @@ def _deformable_import_cloth(ctx: _DeformableImportContext) -> None:
...
-            )
+        vol_density = _resolve_deformable_density(
+            prim,
+            cloth_mat.get("density"),
+            deformable_read,
+            ctx.linear_unit,
+            read_base_material=surface_material is None,
+        )
+        if thickness is None:
+            thickness = _AOUSD_DEFAULT_THICKNESS / ctx.linear_unit
```
AnkaChan, line 330: **P2 — cloth thickness fallback: 2mm → 1mm, now unconditional and silent.**

This replaces the gated `_DEFAULT_CLOTH_THICKNESS = 0.002` + "assuming the default thickness" warning (tests asserting it were deleted). Impact: legacy materials import tri_ke 2x / bend 8x softer with half the mass and radius — undocumented, while cables got the explicit "instead of Newton's previous 2.5 mm radius" + recipe in 3722.changed.1.md. Bare assets drop from `builder.default_particle_radius` (0.1) to 0.0005 — ~200x smaller contact radius, and that knob no longer affects imported cloth at all.

The 1 mm-derived radius is the saner physical default on meter stages, so no revert — just: changelog the delta like the cable fragment (recipe: `physics:thicknesses = [0.002]`, `elementType = "constant"`), and restore an import-time notice, at minimum for the bare-asset case.

newton/_src/usd/utils.py
```diff
@@ -1969,58 +2023,82 @@ def get_tetmesh(
...
-                # division by zero in the Lame parameter conversion.
-                nu = max(-0.999, min(nu, 0.499))
-                k_mu = E / (2.0 * (1.0 + nu))
-                k_lambda = E * nu / ((1.0 + nu) * (1.0 - 2.0 * nu))
+                youngs = None
+        if youngs is None and is_current_volume_material:
+            youngs = _AOUSD_DEFAULT_YOUNGS_MODULUS * linear_unit
+        if poissons is not None:
+            nu = poissons
+            if not math.isfinite(nu) or not (-1.0 < nu <= 0.5):
```
AnkaChan: **P2 — Poisson out-of-range: silent clamp → discard-to-0.3, inconsistent at the boundary.**

Main clamped nu into (-0.999, 0.499); now authored nu=0.6 warns and imports with 0.3 → k_lambda goes from ≈166·E (clamped) to ≈0.577·E, ~288x softer. Meanwhile exactly 0.5 is *approximated* to 0.499 (`_deformable_lame_parameters`, lines 1870-1883) while 0.5+ε is *discarded* — 0.5 and 0.5+ε shouldn't be 288x apart.

0.3 is too soft for authored near-incompressible intent: please extend the 0.499 approximation (with its warning) to the whole [0.5, 1) band; everything else can keep treat-as-unauthored.

newton/_src/utils/import_usd_deformable_cable.py
```diff
@@ -470,22 +553,44 @@ def _deformable_import_cable_graphs(ctx: _DeformableImportContext) -> tuple[set[
         # the attachment post-pass, which preserves it in path_attachment_attrs as
         # unsupported instead of silently snapping the geometry together.
         stiffness_val = deformable_read(prim, "stiffness")
+        damping_val = deformable_read(prim, "damping")
+        stiffness = (
+            math.inf
+            if stiffness_val is None
+            else usd._coerce_deformable_float(stiffness_val, prim, "stiffness", warn_on_failure=False)
```
AnkaChan: **P3 (optional) — shared attachment-gains helper.**

This coerce+validate block is duplicated at attachments.py:106-115 (validity predicate: lines 572-574 ≡ attachments:153), and the weld pass relies on both copies classifying identical values identically. One shared helper makes that invariant structural; the divergent consequences (skip-weld vs mark-unsupported) can stay at the call sites.

newton/_src/usd/utils.py
```diff
@@ -1969,58 +2023,82 @@ def get_tetmesh(
...
+        youngs = _coerce_deformable_float(
+            _read_physics_attr(material_prim, "youngsModulus", compat_namespaces), material_prim, "youngsModulus"
+        )
+        poissons = _coerce_deformable_float(
+            _read_physics_attr(material_prim, "poissonsRatio", compat_namespaces), material_prim, "poissonsRatio"
+        )
+        raw_density = _read_physics_attr(material_prim, "density", compat_namespaces)
+        density_val = _coerce_deformable_float(raw_density, material_prim, "density")
+
+        # The unregistered proposal schema cannot inject its sentinel fallback, so apply
```
AnkaChan, line 2064: **P3 (optional) — two validation stacks for the same volume-material attributes.**

This hand-rolled block re-implements what `_read_deformable_material` (lines 2249-2347) centralizes for the importer pass, and they already double-warn in one corner (invalid unscoped `physics:density`: line 2090 + lines 2481-2486). Delegating get_tetmesh's material read to the shared reader would single-source the policy for the next proposal revision.

newton/_src/utils/import_usd_deformable_cloth.py
```diff
@@ -234,7 +367,63 @@ def _deformable_import_cloth(ctx: _DeformableImportContext) -> None:
...
+        element_volumes = tri_areas * np.asarray(face_thicknesses, dtype=np.float64)
+        density_element_masses = resolved_cloth_density * element_volumes
+        density_point_masses = [0.0] * len(mesh_points)
+        for element_mass, face in zip(density_element_masses, tri_faces, strict=True):
+            for point in face:
+                density_point_masses[int(point)] += float(element_mass) / 3.0
+        for point, mass in enumerate(density_point_masses):
+            builder.particle_mass[p0 + point] = mass
+
+        for tri_offset, face_thickness in enumerate(face_thicknesses):
```
AnkaChan: **P3 (optional) — the new per-element loops are pure Python.**

`_resolve_surface_structural_stiffnesses` re-runs per triangle here and per edge below (~250k calls for a 100k-tri cloth; both loops are no-ops when thickness is uniform). The mass lumping (utils.py:759-787, plus lines 375-380 above) loops over data that is already ndarrays — `np.add.at`/`np.bincount` are drop-ins. Cable gains resolve 3x per joint (cable.py:389-402) — cache by radius. Import-time only, but it scales with exactly the assets per-element thickness targets.

AnkaChan (commented):
Reviewed the surface/volume side per jumyung's ping. Solid alignment with the pinned proposal revision — and I verified Jan's 08-31 attachment-gain case is fixed at head, with tests. Findings inline: 2 correctness (P1), 3 silent behavior changes for existing assets (P2), 3 optional cleanups (P3).

Small nits (optional): unreachable `len(point_masses) != n` branch (import_usd_deformable_utils.py:878-883 — both producers guarantee length n); stale "simulator default" comment wording (usd/utils.py:2273, :2297 — the proposal defines -inf as "assumed 10^6 Pa", which the code correctly implements); a docs sentence contrasting empty-bound-material vs no-material defaults would help authors; `path_*_attrs["material"]` exposes both legacy and canonical spellings — consider canonicalizing with a provenance flag before the window closes.

On the open question: keep **option 1** (deprecation window) as implemented, matching Jan — with the addendum that the P2 behavior changes inline should get the same migration-note treatment as the renames.

jcarius-nv (approved):
Re-reviewed the current head, including the latest top-level review and all thread responses.

The previously raised correctness concerns are addressed: malformed material, body, and attachment scalars are contained; non-finite cloth geometry no longer aborts the stage; referenced point masses survive orphan vertices; and the Poisson-ratio compatibility behavior is consistent and covered. The cloth thickness and rest-bend fallback changes now emit actionable warnings and are documented in the changelog with preservation recipes.

The compatibility approach looks appropriate for this experimental schema: retain the easy-to-maintain legacy readers for one release cycle, while documenting the stiffness-unit conversion that cannot be made behaviorally compatible by aliases alone. Please notify Lightwheel and collect asset-level feedback before removing those readers.

Direct token parsing is a reasonable bridge while the AOUSD core deformable APIs remain unregistered. The upstream schema path and the separate update to Newton-owned extension documentation should continue through the tracked schema follow-up; neither needs to block this importer update.

All current required checks pass. I also ran the focused deformable test slice at this head: 174 tests passed, with one expected skip.

Approved.

newton/_src/utils/import_usd_deformable_cloth.py
mzamoramora-nvidia: Addressed in `382c0d3a`. Cloth geometry is now checked for non-finite points and triangle areas before the builder is modified. A malformed cloth is skipped with a warning, while valid sibling cloth still imports. The regression also verifies that the skipped cloth leaves no partial builder state or result metadata.

newton/_src/utils/import_usd_deformable_utils.py
mzamoramora-nvidia: Addressed in `382c0d3a`. Adjacent volume is now required only for referenced points. Authored masses on orphan points are ignored with a warning, while masses on referenced points are preserved. Cloth and TetMesh regressions cover this behavior.

newton/_src/utils/import_usd_deformable_cloth.py
mzamoramora-nvidia: Addressed in `9bcbab8f`. The importer keeps the proposal's `flat` fallback, but now warns once when that unauthored fallback changes a non-planar rest shape. The documentation and changelog explain that authoring `restShape` preserves the previous Newton behavior.

newton/_src/utils/import_usd_deformable_cloth.py
mzamoramora-nvidia: Addressed in `9bcbab8f`, with strict-warning coverage completed in `1e9c95e4`. The proposal's 1 mm fallback remains and now warns only when selected, reports the resolved stage-unit value, and provides the explicit 2 mm migration recipe.

newton/_src/usd/utils.py
mzamoramora-nvidia: Addressed in `c993a991`. Values in `[0.5, 1)` now resolve to `0.499` with a compatibility warning. Values outside the recoverable range and non-finite values use the `0.3` fallback. The same resolver is used by `get_tetmesh()` and all deformable importer families.

newton/_src/utils/import_usd_deformable_cable.py
mzamoramora-nvidia: Addressed in `38f95357`. The weld prepass and attachment post-pass now share one private gain resolver while retaining their different consequences and warning ownership.

newton/_src/usd/utils.py
mzamoramora-nvidia: Agreed that this duplication is worth removing, but I left it for a follow-up. `get_tetmesh()` supports public material and namespace compatibility paths that the internal reader does not, so consolidating them here could change fallback precedence and warning behavior. I would first add coverage for those compatibility paths, then unify the readers.

newton/_src/utils/import_usd_deformable_cloth.py
mzamoramora-nvidia: Addressed in `13657de3`. Mass accumulation and cloth stiffness resolution are vectorized, and cable gains are reused. Local benchmarks improved constant-thickness cloth by 50%, varying-thickness cloth by 14 to 17%, and point-mass conversion by 16.1x.

PR #3990, opened 2026-08-20
[Kamino] Implement joint effort limits

rubengrandia:
## Description

Add per-DoF joint effort limits to SolverKamino for both explicit effort commands and implicit PD drives. Explicit commands on the body-wrench path are clamped directly. Finite effort limits on implicit PD drives are represented as bounded solver rows, so saturation is enforced inside the coupled dynamics solve rather than by clamping an unconstrained PD result afterward.

The change supports dense and sparse PADMM and DVI paths, preserves separate dynamic, friction, and effort multipliers, reports effort reactions through the Newton state, and extends structural-change validation and multi-world bookkeeping for the new rows.

All dynamic, friction, and effort constraints are now **configured per-DoF instead of per joint**. This allows the exact number constraints required for cases where multiple DoFs have different properties. The per-joint setup no longer scaled.

### Joint effort constraint

For one joint DoF, let:

- $h$ be the time step;
- $q^-$ and $\dot q^+$ be the current position and unknown post-step velocity;
- $J_a u^+ = \dot q^+$ be the joint-axis velocity extracted from the unknown body velocity;
- $k_p$, $k_d$, and $\tau_{\max}$ be the PD gains and symmetric effort limit;
- $c_a$ contain the explicit command and the reference-dependent part of the selected controller mode.

For example, with position and velocity targets,

$$
c_a = \tau_{\mathrm{cmd}} + k_p(q_{\mathrm{ref}}-q^-) + k_d\dot q_{\mathrm{ref}}.
$$

The velocity-dependent part is treated implicitly. Define

$$
\rho_a = h k_d + h^2 k_p,
$$

with the $h^2 k_p$ term omitted for velocity-only control. The implementation floors $\rho_a$ at a small positive value. Using the actuator impulse $\lambda_\tau=h\tau^+$, the new row is

$$
J_a u^+ + \rho_a^{-1}\lambda_\tau = \rho_a^{-1}h c_a,
\qquad
-h\tau_{\max} \le \lambda_\tau \le h\tau_{\max}.
$$

Equivalently, at the solution,

$$
\lambda_\tau = \text{clamp}\left(hc_a-\rho_a\dot q^+,-h\tau_{\max},h\tau_{\max}\right).
$$

This adds $\rho_a^{-1}$ to the effort row's Delassus diagonal and uses $h c_a/\rho_a$ as its velocity bias. The solved generalized force is $\tau^+=\lambda_\tau/h$.

Controller-mode terms in $c_a$ are:

- `POSITION`: $\tau_{\mathrm{cmd}} + k_p(q_{\mathrm{ref}}-q^-)$; $k_d$ implicitly damps toward zero velocity.
- `VELOCITY`: $\tau_{\mathrm{cmd}} + k_d\dot q_{\mathrm{ref}}$.
- `POSITION_VELOCITY`: $\tau_{\mathrm{cmd}} + k_p(q_{\mathrm{ref}}-q^-) + k_d\dot q_{\mathrm{ref}}$.
- `POSITION_VELOCITY_FORCE`: the previous expression plus feed-forward effort $\tau_{\mathrm{ff}}$.

### Interaction with the dynamic constraint

The dynamic row represents passive joint-space armature $a$ and viscous damping $b$. When a bounded effort row shares the axis, actuator terms are removed from the dynamic row:

$$
m_p = a + h b,
\qquad
J_a u^+ + m_p^{-1}\lambda_d = m_p^{-1}a\dot q^-.
$$

The dynamic multiplier $\lambda_d$ remains unbounded and represents the passive armature/damping reaction. The effort multiplier $\lambda_\tau$ represents the bounded actuator impulse and operates on the body equations directly. If $a>0$ or $b>0$, both rows exist on the same axis and are solved simultaneously through the same body velocities and the coupled Delassus matrix. If both are zero, bounded implicit PD needs only the effort row.

Without a finite effort bound, implicit PD stays fused into the dynamic row:

$$
m = a + h b + \rho_a,
\qquad
J_a u^+ + m^{-1}\lambda_d = m^{-1}\left(a\dot q^- + h c_a\right).
$$

Thus adding a finite limit splits the actuator from passive dynamics without double-counting either contribution.

### Per-DoF assignment

Rows and actuation routing are selected independently for each DoF:

- Implicit PD is active for `VELOCITY` when $k_d>0$, and for `POSITION`, `POSITION_VELOCITY`, or `POSITION_VELOCITY_FORCE` when $k_p>0$ or $k_d>0$.
- A bounded effort row is created when implicit PD is active and `joint_effort_limit` is finite and below Newton's default unbounded sentinel.
- A dynamic row is created when armature is positive, damping is positive, or implicit PD is active without a bounded effort limit.
- Routing priority is effort row, then dynamic row, then body wrench. Therefore bounded implicit PD always uses the effort row; unbounded implicit PD uses the dynamic row; and a DoF with neither row applies explicit effort as a clamped body wrench.

## Checklist

## Test plan

```bash
uv run --extra dev -m newton.tests -k test_kamino_solver_joint_effort_limit
uv run --extra dev -m newton.tests -k test_kamino_core_model
uv run --extra dev -m newton.tests -k test_kamino_solver_kamino
uv run --extra dev -m newton.tests -k test_kamino_solver_kamino_joint_friction
uvx pre-commit run -a
```

## Summary by CodeRabbit

* **New Features**
  * Added joint effort-limit support for explicit commands and implicit PD drives.
  * Added per-DoF actuation modes and routing, including mixed joint configurations.
  * Added joint friction and effort constraints across dense and sparse solving paths.
  * Added effort clamping, compliance, saturation, warm-starting, and state reporting.
  * Added spherical-joint position-error handling and improved actuator torque filtering.

* **Documentation**
  * Updated solver feature tables, terminology, README content, and constraint documentation.

* **Bug Fixes**
  * Improved structural-update validation for dynamic, friction, and effort constraint changes.
  * Corrected constraint indexing, offsets, and bounded-row handling across solver workflows.


newton/_src/solvers/kamino/_src/core/builder.py
```diff
@@ -485,6 +509,8 @@ def add_joint_descriptor(self, joint: JointDescriptor, world_index: int = 0) ->
         # Check if the descriptor is valid
         if not isinstance(joint, JointDescriptor):
             raise TypeError(f"Invalid joint descriptor type: {type(joint)}. Must be `JointDescriptor`.")
+        if joint.dof_type == JointDoFType.FREE and any(value > 0.0 for value in joint.f_j):
+            msg.warning("Ignoring joint friction on FREE joint %r.", joint.name)
```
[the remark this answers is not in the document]
rubengrandia: This is stale on the current head: `builder.py` imports the existing `msg` utility, so the FREE-joint warning path no longer raises `NameError`.

newton/_src/solvers/kamino/_src/core/builder.py
```diff
@@ -432,13 +453,15 @@ def add_joint(
...
-
         # Check if the DoF type is valid
         if not isinstance(dof_type, JointDoFType):
             raise TypeError(f"Invalid DoF type: {dof_type}. Must be `JointDoFType`.")
+        if isinstance(act_type, JointActuationType):
+            act_type_dof = [act_type] * dof_type.num_dofs
+        elif isinstance(act_type, list):
+            act_type_dof = act_type
+        else:
+            raise TypeError(f"Invalid actuation type: {act_type}. Must be `JointActuationType` or a list of them.")
```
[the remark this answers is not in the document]
rubengrandia: The invariant is already enforced by `JointDescriptor.__post_init__`: a mismatched per-DoF list raises `ValueError` before the descriptor can be finalized or registered. I’m keeping the validation in one place rather than duplicating it in `add_joint`.

newton/_src/solvers/kamino/_src/core/model.py
```diff
@@ -527,19 +585,19 @@ def data(
...
         njdyncts = self.size.sum_of_num_dynamic_joint_cts
         njkincts = self.size.sum_of_num_kinematic_joint_cts
+        njfccts = self.size.sum_of_num_friction_cts
+        njeccts = self.size.sum_of_num_effort_cts
 
         # Construct the model data on the specified device
         with wp.ScopedDevice(device=device):
             # Create a new model data info with the total constraint
-            # counts initialized to the joint constraints count
+            # counts initialized to the joint + bounded constraints count
```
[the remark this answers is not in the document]
rubengrandia: The referenced ranges are outdated on the current head. I don’t plan to expand this feature fix into a broader comment-cleanup pass.

newton/tests/kamino/test_kamino_solver_joint_effort_limit.py
```diff
@@ -0,0 +1,448 @@
...
+import unittest
+
+import numpy as np
+import warp as wp
+
+import newton
+from newton._src.solvers.kamino._src.core.joints import JOINT_TAUMAX
+from newton._src.solvers.kamino.solver_kamino import SolverKamino
+
+newton.use_coord_layout_targets = True
```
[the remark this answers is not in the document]
rubengrandia: Will move tests to public space and adapt accordingly.

newton/_src/solvers/kamino/_src/core/joints.py
```diff
@@ -1465,6 +1687,13 @@ def is_unary(self) -> bool:
         """
         return self.bid_B == -1 or self.bid_F == -1
 
+    @property
+    def act_type(self) -> JointActuationType:
```
Guirec-Maloisel, line 1723: Is there still a use for the overall joint actuation type, aggregated from per-dof types? If yes, does the aggregation with max() make sense in all cases? If no, should we just move officially to per-dof actuation type?

newton/_src/solvers/kamino/_src/core/conversions.py
```diff
@@ -1779,6 +1914,30 @@ def convert_joints(
             joint_kinematic_cts_start,
             joint_bounded_cts_start,
             joint_friction_cts_start,
+            joint_effort_cts_start,
+        ],
+        device=model.device,
+    )
+
+    wp.launch(
```
Guirec-Maloisel, line 1942: Could we add a small comment here to explain what is done, and why it needs to be done in two steps instead of a single pass (if I got it right one need to know first the total number of joint dynamic, effort limit and friction constraints before allocating the arrays that keep track of what dof axis they correspond to)

newton/_src/solvers/kamino/_src/core/joints.py
rubengrandia: The main use for this is the FK actually. It operates on active / passive joints, not DoFs.

newton/_src/solvers/kamino/_src/core/joints.py
```diff
@@ -193,6 +199,169 @@ def from_newton_wp(target_mode: int) -> int:
...
+        aggregate = int(JointActuationType.PASSIVE)
+        for dof in range(dof_start, dof_end):
+            act_type = JointActuationType.from_newton_wp(target_mode[dof])
+            if act_type < 0:
+                return -1
+            aggregate = max(aggregate, act_type)
+        return aggregate
+
+
+class JointActuationPath(IntEnum):
```
Guirec-Maloisel: Multiple comments here:
- This, and JointActuationType, should probably be renamed with a "Dof" instead of "Joint" prefix, since they are now per-joint
- After looking through the rest I now get what this field does, but was not so obvious from the naming. A possible suggestion would be to make this an "ActuationMode" instead of path, with options EXPLICIT, IMPLICIT_BOUNDED and IMPLICIT_UNBOUNDED, so that it makes a bit more sense from a high-level without going into implementation details.
- We should make clear (here and/or in the docstring of the corresponding fields in joint descriptor / model) that this field is inferred automatically from actuation types and numerical values (like gains, torque limits etc), rather than a separate parameter that can be set independently.

newton/_src/solvers/kamino/_src/core/joints.py
```diff
@@ -1479,19 +1708,29 @@ def is_actuated(self) -> bool:
         """
         return self.act_type > JointActuationType.PASSIVE
 
-    @property
-    def is_dynamic(self) -> bool:
+    def dynamic_cts_axes(self) -> list[int]:
         """
-        Returns whether the joint's dynamics is simulated implicitly.
+        Returns sorted local DoF axes with dynamic rows.
```
Guirec-Maloisel: If possible I'd avoid mentioning "dynamic rows" here (one needs to know a bit about Jacobians etc to understand what row means in this context). Instead rather stay a bit higher level (like for `effort_cts_axes`), e.g. DoF axes with joint dynamics or joint Coulomb friction (for "friction rows" below)

newton/_src/solvers/kamino/_src/core/joints.py
```diff
@@ -1566,6 +1798,20 @@ def __post_init__(self):
         # Validate that the specified parameters are valid
         self._check_parameter_values()
 
+        # Check if DoF type + dynamic/implicit PD settings are compatible.
+        if self.dof_type == JointDoFType.FREE and (self.num_dynamic_cts > 0 or self.num_effort_cts > 0):
```
Guirec-Maloisel, line 1840: Minor point, but the case for FIXED and FREE could be merged and have the same message (now the message for FIXED is a bit more detailed than for FREE). Although it's not so clear to me how you could end up in the FIXED case, because there are 0 DoFs in the first place.

newton/_src/solvers/kamino/_src/core/joints.py
```diff
@@ -1566,6 +1798,20 @@ def __post_init__(self):
...
+        if self.dof_type == JointDoFType.FIXED and (self.num_dynamic_cts > 0 or self.num_effort_cts > 0):
+            if self.num_dynamic_cts > 0 and self.num_effort_cts > 0:
+                violation = "dynamic or implicit PD"
+            else:
+                violation = "dynamic" if self.num_dynamic_cts > 0 else "implicit PD"
+            raise ValueError(
+                f"Invalid joint: FIXED joints cannot have {violation} DoFs (name={self.name}, uid={self.uid})."
+            )
+
         # TODO: Add support for missing multi-DOF joint types in the future.
```
Guirec-Maloisel, line 1844: Is there something that is still blocking for supporting multi-DOF types with dynamic/friction/effort constraints? Now that everything is per-DoF...

newton/_src/solvers/kamino/_src/core/joints.py
```diff
@@ -1574,46 +1820,14 @@ def __post_init__(self):
...
                 "Invalid joint: Kamino currently supports dynamic/implicit joints "
                 f"for REVOLUTE, PRISMATIC, or GIMBAL types (name={self.name}, uid={self.uid})."
             )
 
-        # TODO: Add more checks based on JointDoFType because how do we
-        # handle iterating in DoF-like CTS space when num_coords != num_dofs?
-        # Ensure that PD gains are only specified for actuated joints
-        if self.is_passive and (np.any(self.k_p_j) or np.any(self.k_d_j)):
-            raise ValueError(
-                f"Joint `{self.name}` has non-zero PD gains but the joint is defined as passive:"
```
Guirec-Maloisel, line 1588: Should these exceptions really be removed? Would it make sense to keep these as warnings? I think that now an actuator set e.g. to position or velocity actuation type, but with zero gains there, would behave silently like a passive or lower actuation type (based on the new functions decided what dofs need each type of constraint).

newton/_src/solvers/kamino/_src/core/joints.py
```diff
@@ -1374,6 +1585,12 @@ class JointDescriptor(Descriptor):
     Coulomb joint friction constraints in its world.
     """
 
+    effort_cts_offset: int = -1
+    """
+    Index offset of this joint's effort-limited actuator rows among all
+    effort-limited actuator constraints in its world.
```
Guirec-Maloisel: "effort-limited actuator constraints" (several occurences, also "effort-limited actuator Lagrange multipliers" sounds a bit wrong, it's not the constraints that are limited. I think it's short for effort-limited implicit PD, so we should use it in full, or maybe use effort-limit actuator constraints in more compact form.

newton/_src/solvers/kamino/_src/core/joints.py
```diff
@@ -1093,8 +1298,13 @@ class JointDescriptor(Descriptor):
     # Attributes
     ###
 
-    act_type: JointActuationType = JointActuationType.PASSIVE
-    """Actuation type of the joint."""
+    act_type_dof: list[JointActuationType] = field(default_factory=list)
```
Guirec-Maloisel: I'd suggest renaming this to act_types (if one removes the aggregate act_type), or dof_act_types (if one keeps it) to make it clearer that it's a list (or expand to act_type_per_dof)

newton/_src/solvers/kamino/_src/core/joints.py
```diff
@@ -193,6 +199,169 @@ def from_newton_wp(target_mode: int) -> int:
...
+            aggregate = max(aggregate, act_type)
+        return aggregate
+
+
+class JointActuationPath(IntEnum):
+    """
+    An enumeration of per-DoF actuation routing paths.
+    """
+
+    BODY_WRENCHES = 0
```
Guirec-Maloisel, line 274: To be sure I get all details right: the "BODY_WRENCHES" path corresponds to explicit actuation (torque applied with other external forces), and matches the "FORCE" actuation type? Is there any situation where this could end up being applied on top of the torque from implicit PD? What about the feed-forward torque, does that one still get applied if gains are set to zero?
If that's correct, we could add a sentence here that one gets this mode as a result of using "FORCE" as actuation type.

newton/_src/solvers/kamino/_src/core/joints.py
```diff
@@ -2282,13 +2605,17 @@ class JointsData:
     used for implicit integration of joint dynamics.
 
     Let ``m_j_0 := a_j + dt * b_j``, where ``dt`` is the simulation time step.
-    The actuation mode determines the remaining terms:
+    When a joint's dynamic rows retain fused actuation (no passive/actuator
```
Guirec-Maloisel: I get what this is about, but what "retain fused actuation" means is not clear at first read (here + other occurence below).
It feels like we're lacking an explanation somewhere about bounded/unbounded implicit PD and passive joint dynamics (armature/damping), saying that dynamic joint constraints are accounting for passive dynamics, and for implicit PD in the unbounded case; while effort limit rows account for bounded implicit PD.

newton/_src/solvers/kamino/_src/solvers/fk/solver.py
```diff
@@ -213,6 +213,19 @@ def finalize(self, model: ModelKamino | None = None, config: ForwardKinematicsSo
...
+        base_joint_ids_input = self.model.info.base_joint_index.numpy().tolist()
+        base_joint_id_set = {joint_id for joint_id in base_joint_ids_input if joint_id >= 0}
+        for joint in range(self.model.size.sum_of_num_joints):
+            if joint in base_joint_id_set:
+                continue
+            dof_actuation = joint_act_type_dof[joint_dof_start[joint] : joint_dof_start[joint + 1]]
+            passive = dof_actuation == JointActuationType.PASSIVE
+            if np.any(passive) and not np.all(passive):
+                raise ValueError(
+                    f"Invalid FK actuation for joint {joint}: all DoFs must be passive or all must be actuated."
```
Guirec-Maloisel, line 231: Makes sense that passive/actuated mixes aren't supported yet by FK, but maybe there should be a note about this at a higher level (in the list of supported features, or in the FK/reset documentation, or in the actuation type documentation). Now the only way to learn about this is by triggering the exception.

We should probably create an issue to add support for this too, as I don't see any fundamental blockers. Could be tackled as part of the D6 implementation, possibly.

newton/_src/solvers/kamino/_src/core/joints.py
```diff
@@ -2349,12 +2677,49 @@ class JointsData:
...
+    := 0`` on each passive dynamic row and the bounded effort rows supply the
+    actuator contribution.
+
     For ``POSITION``, the ``dt * k_d_j`` term in :attr:`m_j` supplies derivative
     damping toward zero velocity without consuming ``dq_j_ref``.
 
     Shape of ``(sum_of_num_dynamic_joint_cts,)``.
     """
 
+    inv_rho_a: wp.array[wp.float32] | None = None
```
Guirec-Maloisel: rho is a bit unusual for an inertia, why not inv_m_a? To be consistent/parallel with dq_b_j, m_j vs dq_b_a, m_a.

newton/_src/solvers/kamino/_src/core/joints.py
```diff
@@ -2349,12 +2677,49 @@ class JointsData:
...
+    division by zero.
+
+    Shape of ``(sum_of_num_effort_cts,)``.
+    """
+
+    dq_b_a: wp.array[wp.float32] | None = None
+    """
+    Velocity bias of each effort-limited implicit-PD actuator row [m/s, rad/s].
+
+    ``dq_b_a := c_a / rho_a`` with ``c_a`` including the explicit command,
```
Guirec-Maloisel: I don't think c_a is being detailed in these docstrings. If it only intervenes in dq_b_a, we could also just expand the formula here directly.

newton/_src/solvers/kamino/_src/dynamics/delassus.py
```diff
@@ -1609,7 +1737,7 @@ def diagonal(self, diag: wp.array[wp.float32]):
         """Stores the diagonal of the Delassus matrix in the given array.
 
         Note:
-            This uses the diagonal of the pure Delassus matrix, without any regularization or
+            This includes armature and effort compliance regularization, but excludes ``eta`` and
```
Guirec-Maloisel: We could make it clear that this regularization is still part of the pure Delassus matrix. E.g.:
"This uses the diagonal of the pure Delassus matrix, without proximal regularization (`` eta``) or preconditioning. This does include armature and effort compliance diagonal terms which are part of the Delassus matrix."

newton/_src/solvers/kamino/_src/dynamics/dual.py
```diff
@@ -1666,7 +1719,7 @@ def _build_free_velocity_bias(
 
         Primarily builds the free-velocity bias vector ``v_b`` (joint dynamics and
         kinematics, limits, contacts). Also fills auxiliary inequality data that is
-        zeroed in :meth:`zero` and consumed later by the solver: joint friction
+        zeroed in :meth:`zero` and consumed later by the solver: joint-friction
```
Guirec-Maloisel, line 1722: This should mention the new auxiliary data (effort bounds and bias)

newton/_src/solvers/kamino/_src/dynamics/wrenches.py
```diff
@@ -740,6 +832,31 @@ def compute_constraint_body_wrenches_dense(
             device=model.device,
         )
 
+    if model.size.sum_of_num_effort_joint_cts > 0:
+        # Joint-DoF wrenches own zeroing `w_a_i` and must run before this accumulation.
```
Guirec-Maloisel: I don't get this comment. I assume this is about what class or function is responsible for zeroing the array before accumulating wrenches, but "Joint-DoF wrenches" is not a class, so can't really "own zeroing". (one more occurence below)

newton/_src/solvers/kamino/_src/kinematics/joints.py
```diff
@@ -892,6 +928,90 @@ def compute_and_write_joint_implicit_dynamics(
...
+        if act_type == JointActuationType.POSITION:
+            tau_j_tot += k_p_j * q_j_err
+        elif act_type == JointActuationType.VELOCITY:
+            tau_j_tot += k_d_j * pd_dq_j_ref
+        elif act_type == JointActuationType.POSITION_VELOCITY:
+            tau_j_tot += k_p_j * q_j_err + k_d_j * pd_dq_j_ref
+        elif act_type == JointActuationType.POSITION_VELOCITY_FORCE:
+            tau_j_tot += pd_tau_j_ff + k_p_j * q_j_err + k_d_j * pd_dq_j_ref
+
+        # Enforce minimum compliance to avoid division by zero
```
Guirec-Maloisel: I think this is minimum inertia / maximum compliance (compliance = inverse inertia would be infinite for zero gains)

newton/_src/solvers/kamino/_src/utils/control/pid.py
```diff
@@ -113,6 +104,10 @@ def _reset_jointspace_pid_references(
         # Compute the DoF index in the global DoF vector
         dof_index = dofs_offset + dof
 
+        # Only proceed for force actuated DoFs
```
Guirec-Maloisel: Nit: should be "force-actuated" (also below)

newton/_src/solvers/kamino/solver_kamino.py
```diff
@@ -1415,14 +1415,15 @@ def _validate_structural_invariants(self, flags: ModelFlags | int) -> None:
...
+        effort_joint = violations[self._kamino.StructuralUpdateViolation.EFFORT_CTS]
 
         if dynamic_joint != sentinel:
             joint = int(dynamic_joint)
             raise RuntimeError(
                 f"Changing dynamic constraint topology for joint {joint} "
                 f"({self.model.joint_label[joint]!r}) is not supported; recreate SolverKamino to apply the change. "
-                "The dynamic constraint topology changes if armature, damping, target stiffness, or target damping are updated to non-zero values, while they were zero when creating the solver. "
-                "The opposite is also true: if the values are updated to zero, while they were non-zero when creating the solver, the dynamic constraint topology also changes."
+                "Each dynamic row is selected per DoF by passive armature or damping, or by active implicit PD "
```
Guirec-Maloisel: It makes sense to update this message, but this should remain high-level and not reference solver internals (such as "dynamic row"). Ultimately the user needs to know what they did wrong, so the part about variables being updated from zero to non-zero should stay. Also applies for "effort-limit row topology" message below.

Guirec-Maloisel (changes requested):
Great job adding this feature! The formulation checks out to me, and the refactor from per-joint to per-dof actuation makes a lot of sense as well. I had mostly comments about naming / documentation, to avoid creating confusion (especially about the various ways things can be actuated).

newton/examples/kamino/example_kamino_robot_dr_legs.py
```diff
@@ -79,6 +79,10 @@ def __init__(self, viewer: newton.viewer.ViewerBase, args=None):
                 2.0 if mode != newton.JointTargetMode.NONE else 0.0 for mode in robot_builder.joint_target_mode
             ]
 
+        effort_limit_override = getattr(args, "joint_effort_limit", None) if args else None
```
chschuma-disney, line 82: Is this command line argument not registered with `argparse` on purpose, or by mistake?

newton/_src/solvers/kamino/_src/core/joints.py
rubengrandia: Yes — FK and the world/notify bookkeeping still need a joint-level passive-versus-actuated classification, while dynamics uses the per-DoF values. `max()` is only used to preserve that coarse classification; no control path interprets the selected non-passive enum. I made that contract explicit and renamed the per-DoF storage so the distinction is clear.

newton/_src/solvers/kamino/_src/core/joints.py
rubengrandia: Agreed on making the per-DoF nature and inference explicit. I used `DofActuationPath`, `dof_act_types`, and `dof_act_paths`. I kept “path” rather than “mode”: armature/damping-only models route explicit effort through the dynamic constraint without implicit PD, and “mode” would clash with target modes such as POSITION and POSITION_VELOCITY.

newton/_src/solvers/kamino/solver_kamino.py
rubengrandia: I rewrote the dynamic, friction, and effort-limit errors around the authored properties that changed (zero/non-zero armature, damping, gains, friction, or effort-limit status), removed row/topology jargon, and kept the actionable instruction to recreate `SolverKamino`.

newton/_src/solvers/kamino/_src/core/joints.py
rubengrandia: I replaced the unexplained “fused” wording with a routing overview: passive armature/damping and unbounded implicit PD share the dynamic constraint, while bounded implicit PD uses an effort-limit constraint and any passive dynamics remain separate.

newton/_src/solvers/kamino/_src/core/joints.py
rubengrandia: Since the joint-level aggregate remains for FK and bookkeeping, I renamed the list and array to `dof_act_types` so its scope and plurality are explicit.

newton/_src/solvers/kamino/_src/core/joints.py
rubengrandia: Agreed — this quantity plays the same effective-inertia role as `m_j`. I renamed it to `inv_m_a` and documented `m_a = dt * k_d + dt² * k_p` as the effective actuator inertia, with `inv_m_a` as the corresponding compliance.

newton/_src/solvers/kamino/_src/core/joints.py
rubengrandia: Friction is already per-DoF for supported non-free joint types, but dynamic and effort constraints still have an explicit REVOLUTE/PRISMATIC/GIMBAL whitelist. I documented that remaining limitation and prepared a focused follow-up for the other multi-DoF types rather than expanding the solver scope in this PR.

newton/_src/solvers/kamino/_src/core/joints.py
rubengrandia: Correct: `BODY_WRENCHES` applies explicit `tau_j` with external body wrenches and is normally the `FORCE` path. However, this explicit forces will be routed through the dynamics constraint if there is armature or damping for that DoF. Routing is exclusive per DoF, so it is not added on top of implicit PD. Feed-forward effort is included for `POSITION_VELOCITY_FORCE` on the implicit dynamic/effort path; zero-gain PD configurations are now rejected rather than silently falling through.

newton/_src/solvers/kamino/_src/solvers/fk/solver.py
rubengrandia: Good call — I documented the joint-homogeneous FK restriction in the Kamino guide and FK API documentation, and prepared a follow-up for mixed passive/actuated DoFs on multi-DoF joints. The current universal-joint regression test remains as the explicit limitation until FK partitioning and kernels are extended.

newton/_src/solvers/kamino/_src/core/joints.py
rubengrandia: Agreed on rejecting silent no-op PD modes. Validation now rejects selected PD modes whose relevant gains are all zero (`VELOCITY` requires `k_d`; position-based modes require `k_p` or `k_d`). Gains on `PASSIVE` and `FORCE` remain allowed, since models may retain authored gains while switching modes.

rubengrandia:
Thanks for the detailed review @Guirec-Maloisel . I had an agent take a pass, because most comments we're straightforward. Here is an overview of what changed:

Changes made:

- Renamed per-DoF routing to `DofActuationPath` and `dof_act_paths`.
- Renamed `act_type_dof` to `dof_act_type`
- Documented that `act_type` is only a coarse passive/actuated aggregate for FK and bookkeeping.
- Added validation that selected implicit-PD modes have a non-zero applicable gain; unused gains on PASSIVE/FORCE remain allowed.
- Clarified inferred routing, explicit body-wrench behavior, exclusive paths, and feed-forward effort.
- Documented the conversion count-then-index pass for per-DoF constraints.
- Reworded constraint-axis APIs and standardized effort-limit implicit-PD terminology.
- Renamed actuator compliance storage from `inv_rho_a` to `inv_m_a`; documented effective actuator inertia and bias formulas.
- Clarified Delassus, dual-data, wrench-zeroing, compliance-floor, and PID comments.
- Rewrote structural-update errors in user-facing terms.
- Documented current multi-DoF dynamic/effort and mixed-actuation FK limitations.

Guirec-Maloisel (dismissed):
Thanks for the quick refinement pass, looks all good from my side now!

chschuma-disney (dismissed):
Fantastic work. Just a tiny comment/question from my side, otherwise everything looks good. Feel free to ignore it, it's ultimately not important.
Approving for now since I'm out tomorrow.

newton/examples/kamino/example_kamino_robot_dr_legs.py
rubengrandia: I only needed it for turning it of for the DVI unit test. But you are right:dding it to the CLI could be useful. Added it.

Guirec-Maloisel (approved):

PR #3993, opened 2026-08-20
[asv] NotifyDRLegs: bump benchmark version to "2" (sampling change in #3859)

### LINKED ISSUE #3992: [asv] NotifyDRLegs.time_notify_* ~2-7x step is a sampling-change artifact of #3859, not the Warp 1.16 pin (#3780) — bump the benchmark's ASV version to split the series
## Summary

The fleet-wide ~2–7× step in `bench_kamino.NotifyDRLegs.time_notify_*` that the daily ASV triage
has been carrying as a P1 regression is a **measurement artifact of the sampling change in #3859**
("Fix NotifyDRLegs ASV sampling", merged 2026-08-10), not a product or dependency regression.
The earlier suspect, the Warp 1.16 pin bump **#3780, is exonerated**: an A/B by @RubenGrandia
showed the elevated timing follows the sampling config (`number=1` vs `number=10`) independent of
Warp version.

It shows up as an *in-series* regression only because ASV's benchmark version hash covers
benchmark/setup **source code**, not timing attributes — so the methodology change never split the
series. Proposed fix: add an explicit `version = "2"` to `NotifyDRLegs`.

## What happened

#3859 changed only class-level timing/config attributes on `NotifyDRLegs`
(`asv/benchmarks/simulation/bench_kamino.py`):

```diff
-    number = 10
-    repeat = 7
+    number = 1
+    repeat = 5
     rounds = 1
-    timeout = 3600
+    warmup_time = 0
+    timeout = 600
```

The benchmark's first `wp.synchronize_device()` carries a one-time setup cost. With `number = 10`
that cost was amortized across 10 in-sample iterations; with `number = 1` every sample pays it.
Measured magnitude jumps ~2–7× across the `time_notify_*` family (worst: `time_notify_model_properties`,
~7× on the Jetsons) at the first nightly after the merge — with no change to the code being measured.

## Why ASV shows it as an in-series regression (the mechanics)

ASV computes each benchmark's version as (asv_runner `benchmarks/_base.py`, 0.3.x):

```python
self.code = get_source_code([self.func] + self._setups + [self._setup_cache])
code_hash = sha256(self.code.encode("utf-8")).hexdigest()
self.version = str(_get_first_attr(attr_sources, "version", code_hash))
```

i.e. `sha256(timed-function source + setup + setup_cache)`, unless an explicit `version` attribute
overrides it. `number`, `repeat`, `rounds`, `warmup_time`, `timeout` are **not hashed**, and
`NotifyDRLegs` has no explicit `version` attribute (none exists anywhere in `asv/benchmarks/`).
#3859 changed nothing that ASV hashes, so the version stayed put and old/new results landed in one
continuous series — the step then reads as a regression at the boundary.

Result-file evidence (`time_notify_model_properties`, machine SC-PV-10 / RTX 5090):

| Date | Commit | `stats_number` | `stats_repeat` | version hash |
|---|---|---|---|---|
| 2026-08-07 | `71fe5e63` | `[10]` | `[7]` | `a00bd83135d3…` |
| 2026-08-10 | `cc6f60c8` | `[10]` | `[7]` | `a00bd83135d3…` |
| 2026-08-11 | `98ba6dc6` | `[1]` | `[5]` | `a00bd83135d3…` |
| 2026-08-19 | `8f1dbad7` | `[1]` | `[5]` | `a00bd83135d3…` |

The hash is identical in **all 57 result files** for this machine from the benchmark's first
appearance (2026-07-21) through 2026-08-19; the sampling flip at 08-11 is plainly visible in
`stats_number`/`stats_repeat` while the version never moves.

Why #3780 looked guilty: the initial cross-machine bracket of the step contained the Warp pin bump
and no `asv/` change, so it classified as a dependency bump. The A/B (same benchmark config across
Warp pins → no step; `number` 10 vs 1 at fixed Warp → full step) plus the result-file metadata above
overturn that. No action needed on #3780 or Warp 1.16.

## Proposed fix

Add an explicit version pin to the class in `asv/benchmarks/simulation/bench_kamino.py`:

```python
class NotifyDRLegs:
    version = "2"  # sampling/methodology changed in #3859 -> new ASV series
    number = 1
    repeat = 5
    ...
```

Per `_get_first_attr(attr_sources, "version", code_hash)`, the explicit `version` overrides the
code hash for **every** `time_notify_*` method in the class. Pre-#3859 results stay under the old
hash, post-bump runs start a new series, and the 08-11 discontinuity becomes a series boundary
instead of a flagged regression. This should arguably have shipped inside #3859; it is a one-line
follow-up. (The bump is not retroactive — the 08-11→bump rows already stored under `a00bd83135d3…`
remain in the old series; optionally re-run/re-baseline post-08-10 commits if a perfectly clean cut
at the boundary is wanted. Any new string works; only "different from before" matters.)

## General foot-gun worth noting

Any change to ASV timing attributes (`number`, `repeat`, `rounds`, `warmup_time`, `timeout`, or
`setup`-independent config like world counts held in un-hashed attrs) **silently continues the
existing series**, because none of it enters the version hash. Whenever a change re-tunes sampling
in a way that shifts measured magnitude, the PR should bump an explicit `version` on the affected
class in the same commit. Suggest adopting that as a convention for `asv/benchmarks/`.

## How to verify

1. **Hash constancy across the sampling flip** (the artifact): in the ASV results repo, for any
   machine dir, dump `version` + `stats_number` for
   `simulation.bench_kamino.NotifyDRLegs.time_notify_model_properties` per result JSON
   (`result_columns` gives the column order). Expect `stats_number` `[10]→[1]` at the 08-11 nightly
   with the version constant `a00bd83135d3…` on both sides.
2. **A/B the artifact** (repro of Ruben's check): at a fixed newton commit, run the benchmark with
   `number=10` vs `number=1` under the same Warp — the ~2–7× step reproduces; run `number` fixed
   across `warp-lang==1.16.0.dev20260716` vs `1.16.0` — no step.
3. **The fix works**: apply the `version = "2"` patch, run the ASV suite once, and check
   `results/benchmarks.json` — `NotifyDRLegs.time_notify_*` entries now report `"version": "2"`;
   the dashboard shows the old series ending and a new one starting at the bump, and the boundary
   is no longer flagged as an in-series regression.

Refs: #3859 (cause, sampling change), #3780 (exonerated), #3858 (separate, real ~1.3–1.9×
`time_notify_body_inertial_properties` cost from the validation kernel — not this issue).

dylanturpin:
Adds an explicit `version = "2"` to `NotifyDRLegs` so ASV splits the series at the #3859 sampling change. ASV's per-benchmark version hashes only benchmark/setup source, never timing attrs, so #3859 (`number` 10→1) continued the old series and the ~2–7x step read as an in-series regression.

Fixes #3992. Cause: #3859; #3780 is exonerated (A/B: the step follows the sampling config, not the Warp version) — no action needed there.

Not retroactive: old rows keep the old hash; the new series starts at the bump.

## Summary by CodeRabbit

* **Chores**
  * Started a new benchmark series for the NotifyDRLegs simulation to reflect updated sampling and methodology.

### DIFF
--- a/asv/benchmarks/simulation/bench_kamino.py
+++ b/asv/benchmarks/simulation/bench_kamino.py
@@ -157,6 +157,7 @@ class KpiDRLegs(_KpiBenchmark):
 class NotifyDRLegs:
     """Benchmark Kamino model notifications for 2048 DR Legs worlds."""
 
+    version = "2"  # sampling/methodology changed in #3859 -> new ASV series
     number = 1
     repeat = 5
     rounds = 1


jcarius-nv:
Nice find, so this explains why the hashes stayed the same even though the underlying benchmark configuration did actually change.

jcarius-nv (approved):

rubengrandia (approved):
Thanks!

PR #3994, opened 2026-08-20
Derive graph-conditional usage from platform support

tomasthoresen:
## Summary

Several solvers and examples assume CUDA graph conditional nodes are available and fail at capture time on platforms where they are not (Warp reports availability via `wp.is_conditional_graph_supported()`; it is false on CPU-only setups, CUDA toolkits below 12.4, and HIP/ROCm devices). This PR derives the behavior from that query instead of assuming it:

- **kamino/PADMM**: `use_graph_conditionals` is clamped to platform support, taking the solver's own documented unrolled-loop alternative instead of raising mid-capture. Five of the six kamino examples go from failing to running on such platforms.
- **mujoco_warp integration**: conditional-graph use is disabled when unsupported, matching what mujoco_warp itself does at runtime.
- **implicit-MPM examples**: graph capture is skipped when the convergence loop would need conditional nodes; the host-driven residual loop cannot run under an active capture.
- **rj45 example**: guards viewer picking/gizmo access so the example runs under headless/non-interactive viewers.
- **tests**: model-comparison expectations updated to accept the forced `graph_conditional` flag either way.

No behavior changes on platforms where conditional nodes are available: the query returns true there and every path is unchanged.

## Testing

Validated over the full Newton unit suite and examples on a platform without conditional graph nodes (gfx1151/ROCm via a Warp HIP port: 4851 passing units; kamino and coupled-MPM examples run where they previously raised), and the test suite on CUDA (RTX A4000), where the query returns true and behavior is unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility on devices that do not support conditional execution graphs.
  * Automatically uses reliable non-graph simulation fallbacks when conditional graph support is unavailable.
  * Updated coupled simulation examples to avoid unsupported graph capture paths.
  * Improved model comparison handling across devices with different graph capabilities.
  * Clarified warnings when conditional graph execution is unavailable.

* **Documentation**
  * Documented platform-dependent graph behavior and fallback handling.


newton/_src/solvers/kamino/_src/solvers/padmm/solver.py
```diff
@@ -242,7 +242,7 @@ def finalize(
         # Cache high-level solver options shared across all worlds
         self._warmstart = warmstart
         self._use_acceleration = use_acceleration
-        self._use_graph_conditionals = use_graph_conditionals
+        self._use_graph_conditionals = use_graph_conditionals and wp.is_conditional_graph_supported()
```
Guirec-Maloisel, line 246: Could we print a warning (with Kamino's logger, you can search for `msg.warning` in Kamino for example usage) if graph conditionals get disabled because of the platform?
E.g. "Graph conditionals are unavailable on this platform, using unrolled for-loops over max iterations in PADMM solver."

tomasthoresen:
Both review items addressed: added the PADMM warning via Kamino's logger (fires only when the caller requested conditionals and the platform lacks them), and a Towncrier fragment (3994.changed.md) describing the compatibility behavior. Thanks for the review!

newton/_src/solvers/kamino/_src/solvers/padmm/solver.py
```diff
@@ -242,7 +242,12 @@ def finalize(
...
+        if use_graph_conditionals and not self._use_graph_conditionals:
+            msg.warning(
```
Guirec-Maloisel, line 248: I think you're missing an import for msg. Also, the message is missing the the of "in the PADMM solver" (forgot it in my comment above).

tomasthoresen:
Good catch on the missing import — fixed (verified the module imports and resolves the logger at runtime), plus the message wording and the fragment capitalization. The NameError could only fire on platforms without conditional-node support, which is exactly the path CI here can't execute — apologies for missing it.

Guirec-Maloisel (commented):
Looks good from the Kamino side! Will approve when the rest of the changes gets reviewed by a code owner.

eric-heiden (dismissed):
Thanks, LGTM! @Guirec-Maloisel can you also approve?

tomasthoresen:
The ubuntu-24.04-arm unittest failure looks unrelated to this change: the failing assertion is 'Unexpected stderr' triggered by a USD 'Could not open asset' warning on the runner, and this PR touches only solver conditional-graph gating (no USD/asset paths). The same job passed on the previous revision of this branch. Could someone re-run that job when convenient?

tomasthoresen:
Correction to my earlier comment: the ARM/macOS failures were caused by this PR after all — the new PADMM warning fired on every solver construction on CPU-only runners (where the conditional-node query is false by definition), tripping strict-stderr tests. The USD notice in the ARM log was a red herring. Fixed by gating the warning on the model's device being CUDA: on CPU there is no graph capture to fall back from, so the warning is noise there. GPU runners were unaffected (query true, warning silent).

Guirec-Maloisel (approved):

tomasthoresen:
Gentle ping — this is fully green and approved, and just needs the workflow-approval click for the final revision (its diff vs the already-passed GPU run is a single guarded line from the review). Happy to make any further changes if needed. Thanks!

tomasthoresen:
Quick note for whoever looks next: after the re-run, every individual test leg is green on the current revision (including ubuntu-24.04-arm, which had caught the earlier warning issue — confirms the fix). The remaining red entries are the aggregate jobs from the partially re-run older workflow, and the GPU jobs are still waiting on the external-PR approval click. Thanks!

tomasthoresen:
Looked at the ubuntu/windows leg failures: both are test_import_usd hanging (log shows a single test running 3300s before the job gave up, around the intentional does_not_exist.usda fixture) — nothing in this PR touches USD import. Meanwhile arm64 and macOS, which caught the earlier warning issue, are green on this revision. Looks like the USD test flake is hitting different legs on different runs.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -6619,6 +6619,11 @@ def add_body_equality(i: int):
         with wp.ScopedDevice(model.device):
             # create the MuJoCo Warp model
             self.mjw_model = mujoco_warp.put_model(self.mj_model)
+            if not wp.is_conditional_graph_supported():
+                # No conditional graph nodes on this platform (e.g. HIP/ROCm);
+                # use mujoco_warp's non-conditional solver loop (the same
+                # fallback path JAX uses).
+                self.mjw_model.opt.graph_conditional = False
```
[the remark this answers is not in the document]
tomasthoresen: Added in changelog/3994.changed.md

PR #3995, opened 2026-08-20
Optimize VBD soft-contact processing and elasticity

AnkaChan:
## Description

Accelerate `SolverVBD`'s rigid-soft (body-particle) contact processing and deformable elasticity on CUDA, without changing solver settings, contact capacity, or simulation results. Self-contact paths are untouched. All changes are internal; no public API is added or removed.

The baseline spent most of its kernel time on capacity-proportional work: the soft-contact buffer is sized for the worst case (millions of candidate slots at 1024 envs), while only a small prefix holds active contacts, and the solver re-scanned the full capacity once per particle color per VBD iteration for contact forces and once per iteration for AVBD dual updates. Contact generation enumerated every world-compatible (feature, shape) candidate in feature-major order in register-bound edge/face SDF kernels, and the tiled elasticity solve occupied only half of each hardware warp.

Changes, in decreasing measured impact:

- **Per-particle contact gather over an active-prefix adjacency.** After each contact refresh, a graph-safe fixed-size grid walks the clamped active contact prefix once and builds linked per-particle incidence lists (`build_particle_body_contact_adjacency_active`). Each color/iteration then launches one thread per colored particle, which walks only its own list and writes its force/Hessian once — no capacity scan, no output atomics (`gather_particle_body_contact_force_and_hessian`). Enabled only on CUDA with `DeterministicMode.NOT_GUARANTEED`; CPU and deterministic modes keep the legacy scatter.
- **Shape-grouped candidate pairs.** Candidate (feature, shape) pairs are stably sorted by shape index on every device, so on CUDA warps read the same shape data and take the same type-dispatch branch in the generation kernels. The order is deterministic; contact records store their candidate tid, so downstream mapping is order-independent. CPU and deterministic paths emit records in the new order (identical contact set; results shift only within floating-point summation tolerance).
- **Conservative feature-AABB rejection.** Edge/face candidates are tested against the shape's narrow-phase world AABB (expanded by margin + particle radius + negative gap) before any shape-frame math or SDF search. Planes bypass the test; touching bounds are kept.
- **Two-particle cloth elasticity.** For tiled CUDA models with no tetrahedra, one 32-thread block solves two independent particles; a native half-warp XOR reduction preserves the legacy 16-lane reduction order, so results are bit-identical.
- **Tetrahedron-only elasticity.** When tetrahedra are present and all triangle/edge stiffnesses are inactive (static model data, decided at solver construction), a specialized kernel skips the compiled-in triangle/edge paths.
- **Active-prefix dual update** with a fixed graph-safe worker grid.

Measured on 1024-env IsaacLab lift tasks (Franka + cloth / Franka + soft volume, 10 VBD iterations, 1 substep, ABBA protocol with paired bootstrap, reproduced independently on RTX 4090/Windows and L40/Linux): cloth **1.66x** and soft-volume **1.32x** end-to-end environment throughput; summed CUDA kernel time per env step drops ~2x (cloth 1.94x, volume 2.24x). One-env 20-step trajectories on the CUDA nondeterministic path match the baseline bitwise (`max_abs = 0.0` on every saved field) on a contact-rich cloth scene (526 contacts/step) and a frozen-topology volume scene.

CUDA graph capture stays safe under changing contact counts: launch dimensions are host-static, and the device-side count is read inside the kernels; replay tests cover zero -> active -> overflow count transitions.

## Checklist

## Test plan

New coverage in `test_solver_vbd.py`: gather-vs-legacy-scatter equivalence on mixed particle/edge/face records with repeated corners; adjacency counts at zero, worker boundaries, capacity, and raw overflow; changing device-side counts under a captured graph; full `SolverVBD.step()` dispatch and adjacency refresh; deterministic and CPU fallbacks; active-prefix dual exactness with stale-tail preservation; two-particle and tet-only elasticity bitwise equivalence on odd sizes, high-valence fixtures, and shuffled indices. New coverage in `test_collision_pipeline.py`: AABB touching/separation/negative-gap/plane-bypass boundaries; exact contact multisets with rejection on and off; CUDA shape grouping; graph capture/replay.

```
```

On an L40 (CUDA + CPU device parameterizations), both files pass except two tests that fail identically on clean upstream main without this branch: `TestVBDFullSurfaceContact.test_edge_face_pushes_vertices_out_cuda_0` and `TestFullSurfaceSoftContact.test_optimize_against_mesh_texture_sdf_cuda_0` (same assertion values with and without this change). They appear related to an out-of-band mesh texture-SDF extrapolation change on main: at identity scale, a query 0.5 m outside a unit box's surface returns phi = 0.4, while the scaled-path evaluation returns the exact 0.5 (release-1.5 returned 0.5 on both paths). Happy to file that separately.

The port adapts one thing relative to the original release-1.5 branch: the mirrored-mesh-scale test asserts sign and gradient direction with a loose magnitude bound, since exact mirror symmetry of |phi| no longer holds for the full evaluator on main (see above).

**Update (f1165618):** the original branch also split the SDF evaluation into value-only and gradient-only variants inside the contact search. An independent per-change ablation measured no benefit from that split on either workload (the contact-generation gains come from the shape-sorted candidate ordering and the AABB rejection), so it was removed from this PR, cutting about 280 lines.

## Summary by CodeRabbit

- **Performance**
  - Improved CUDA performance for rigid-soft contact processing and deformable elasticity simulations.
  - Added specialized processing for tetrahedral, cloth, and mixed elastic materials.
  - Reduced unnecessary collision work through improved spatial filtering and contact grouping.

- **Collision Handling**
  - Improved soft-contact detection for particles, edges, and faces.
  - Contact ordering is now consistent across devices while preserving the contact set.

- **Compatibility**
  - Maintained existing solver settings, contact capacity, CPU behavior, and deterministic execution.
  - Preserved simulation results within expected floating-point tolerances.


newton/tests/test_solver_vbd.py
```diff
@@ -3735,6 +4210,293 @@ def _soft_contact_presize_is_world_aware(test, device):
...
+        ],
+        outputs=[packed_displacements],
+        device=device,
+    )
+    legacy_bits = legacy_displacements.numpy().view(np.uint32)
+    packed_bits = packed_displacements.numpy().view(np.uint32)
+    np.testing.assert_array_equal(packed_bits, legacy_bits)
+
+
+def _tet_only_elasticity_eligibility_matches_active_materials(test, device):
```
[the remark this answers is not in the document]
AnkaChan: Added in 5dc3c38b.

newton/_src/geometry/soft_contacts_sdf.py
```diff
@@ -134,41 +139,128 @@ def eval_shape_sdf(
...
+            fc = eval_shape_sdf_lower(geo, scale, (1.0 - c) * p + c * q, shape_sdf_index, texture_sdf_table)
         else:
             lo = c
             c = d
             fc = fd
             d = lo + (hi - lo) * inv_phi
-            fd, _fd_a, _gd = eval_shape_sdf(geo, scale, (1.0 - d) * p + d * q, shape_sdf_index, texture_sdf_table)
-    u = 0.5 * (lo + hi)
+            fd = eval_shape_sdf_lower(geo, scale, (1.0 - d) * p + d * q, shape_sdf_index, texture_sdf_table)
+    return 0.5 * (lo + hi)
```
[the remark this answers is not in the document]
AnkaChan: The value-only/gradient-only evaluator split this thread refers to was removed in f1165618: an independent per-change ablation measured no benefit from it on either benchmark workload (the contact-generation gains come from the shape-sorted candidate ordering and the conservative AABB rejection). The edge/face search now uses the full evaluator exactly as on main, so the nonuniform/mirrored-scale concern no longer applies to new code paths.

eric-heiden:
Nice work here, the soft-contact path is noticeably tighter and pulling the AABB cull ahead of the SDF evaluation makes the edge/face flow much easier to follow.

Two of the performance claims did not hold up when I measured them (the shape-major candidate sort and the new per-particle contact gather), this is from my agent running on a L40, so it might not be useful - feel free to ignore those.

newton/_src/sim/collide.py
```diff
@@ -829,12 +834,15 @@ def _world_compatible_pairs(
...
 
-    def _pairs(f_idx: np.ndarray, s_idx: np.ndarray) -> wp.array[wp.vec2i]:
+    def _pairs(f_idx: np.ndarray, s_idx: np.ndarray):
         # ``shape_ok`` (optional, indexed by shape) drops pairs whose shape cannot participate -- e.g.
         # full-surface edge/face excludes shapes without a usable SDF, which fall back to per-particle.
         if shape_ok is not None and len(s_idx):
             keep = shape_ok[s_idx.astype(np.intp)]
             f_idx, s_idx = f_idx[keep], s_idx[keep]
+        if len(s_idx):
+            order = np.argsort(s_idx, kind="stable")
```
eric-heiden, line 967: 🔵 Sorting candidate pairs by shape is a clear win when soft features are far from the rigid shapes, but it costs about 20% in the contact-heavy case, which is exactly the regime full-surface contact exists to handle. The changelog fragment documents the ordering change and its float-tolerance caveat, but not this trade-off.

Measured on a 32x32 cloth in contact with 8 mesh-SDF boxes (3864 emitted contacts), with all arms interleaved in one process and emitted contact counts asserted equal across arms, so only the ordering differs:

| Arm | Median | min | max |
|---|---|---|---|
| feature-major, no cull (current behavior) | 0.5550 ms | 0.5547 | 0.5555 |
| shape-major sort only, no cull | 0.6638 ms | 0.6634 | 0.6642 |
| this PR (shape-major + cull) | 0.6633 ms | 0.6630 | 0.6641 |

The bands do not overlap, and the attribution is unambiguous: the sort accounts for the full +19.6%, while the AABB cull is roughly free here (0.6638 to 0.6633).

How much of that reaches a user is much smaller. The delta over a full `collide()` call is about +14.1%, and once a VBD step is included it lands at 1 to 2%:

| `collide()` + VBD step | before | after | Delta |
|---|---|---|---|
| run 1 | 2.1933 ms | 2.2170 ms | +1.1% |
| run 2 | 2.1840 ms | 2.2319 ms | +2.2% |

One note on the latest commit: the sort used to be gated on `group_by_shape=device.is_cuda` and now applies on every device. CPU gains no warp-coherence benefit, so I measured that arm on its own (24x24 cloth, 8 boxes, 466 contacts) and it is performance-neutral, 3.910 ms median before versus 3.828 ms after, well inside the run-to-run spread. So extending the sort to CPU costs nothing measurable, and its only real consequence is the float-tolerance caveat that is already written down.

The gap is just documentation. One sentence in the changelog fragment noting that this pass gets roughly 20% slower when most candidates are actually in contact would set the right expectation for anyone reading the entry as a pure speedup.

newton/_src/solvers/vbd/particle_vbd_kernels.py
```diff
@@ -2432,6 +2449,146 @@ def accumulate_particle_body_contact_force_and_hessian(
...
+            if particle_index >= 0:
+                node = 3 * contact_index + corner
+                previous = wp.atomic_exch(particle_contact_head, particle_index, node)
+                particle_contact_next[node] = previous
+
+        contact_index += contact_stride
+
+
+@wp.kernel
+def gather_particle_body_contact_force_and_hessian(
```
eric-heiden, line 2376: 🔵 The changelog lists this gather as one of the accelerations, on the grounds that it gathers contact forces per particle over the active contact prefix instead of scanning the full contact capacity every color and iteration. In `example_vbd_soft_rigid_contact`, the soft-rigid workload this change targets, I measure it consistently slower than the scatter kernel it replaces.

That scene has 3126 particles, 6 colors, and 511 active contacts against a capacity of 18756 (2.7% occupancy), which is precisely the low-occupancy regime the gather is designed to exploit.

| Kernel benchmark | scatter (before) | gather (after) | Delta |
|---|---|---|---|
| reps=80, run 1 | 0.0446 ms | 0.0476 ms | +6.8% |
| reps=80, run 2 | 0.0510 ms | 0.0534 ms | +4.6% |
| reps=80, run 3 | 0.0503 ms | 0.0536 ms | +6.5% |

At reps=120 the three runs came in at +7.9%, +7.9%, and +5.3%, so the direction is consistent across six consecutive runs. Even though the sweep launches roughly 3126 threads in total versus 6 x 18756 = 112k for the old path, the per-particle linked-list traversal (`particle_contact_head` into `particle_contact_next`) appears to cost more than the old kernel's very cheap `idx >= count` early-out.

This does not reach the user, which is why I would not block on it. The kernel is around 0.05 ms against a 19 to 23 ms frame, and full-frame numbers land on both sides of zero with fully overlapping bands:

| Full VBD frame | before | after | Delta |
|---|---|---|---|
| run 1 | 19.7070 ms | 20.3068 ms | +3.04% |
| run 2 | 19.6392 ms | 20.0899 ms | +2.29% |
| run 3 | 23.7072 ms | 23.3751 ms | -1.40% |
| run 4 | 18.1261 ms | 18.4314 ms | +1.68% |

So the ask is a documentation correction rather than a revert: either identify the regime where the gather actually pays off (much higher contact counts, or capacity ratios well beyond 6x) and state that in the changelog, or drop it from the list of accelerations. It is worth settling before this lands, because this path carries real complexity the other three changes do not: an extra adjacency-build kernel, two new persistent buffers, and a graph-capture re-init flag.

newton/_src/geometry/soft_contacts_sdf.py
```diff
@@ -238,6 +238,81 @@ def _shape_frames(
...
+    shape_aabb_lower: wp.array[wp.vec3],
+    shape_aabb_upper: wp.array[wp.vec3],
+    feature_lower: wp.vec3,
+    feature_upper: wp.vec3,
+    margin: float,
+    radius: float,
+) -> bool:
+    """Analytic non-plane variant of :func:`_soft_feature_aabb_misses_shape`."""
+    if shape_aabb_lower.shape[0] == 0:
+        return False
```
eric-heiden: ⚪ A few cosmetic leftovers in the new code, all optional and none of them behavioral.

**1. This helper reads like a leftover from an earlier split.** The docstring calls `_soft_feature_aabb_misses_analytic_shape` the "Analytic non-plane variant", but the only caller routes every non-plane shape through it, `MESH` and `CONVEX_MESH` included, so the name and the doc no longer describe what it does. The `shape_aabb_lower.shape[0] == 0` early-out just below is also unreachable: `_soft_feature_aabb_misses_shape` already returns at line 269 for that case, and nothing else calls in here. Folding the two functions into one and passing `geo` instead of `shape_type` would remove both.

**2. `_shape_frames` (line 224) no longer has a production caller.** Both kernels moved to `_shape_world_frame` plus `wp.transform_inverse`, and the only remaining uses are in `test_collision_pipeline.py` at lines 4633 and 4676. Either delete it and inline the two-line frame computation in those tests, or keep it with a short note that it is test-only.

**3. A few redundant global loads in the emit path.** Lines 468-470 re-read `tri_indices[t, 0..2]` into `out_a_idx`/`out_b_idx`/`out_c_idx`, but `a_idx`/`b_idx`/`c_idx` at lines 410-412 already hold exactly those values. Same at lines 592-593 for `out_v0`/`out_v1` versus `v0`/`v1` at 543-544. `_soft_feature_aabb_misses_shape` also re-reads `shape_type[shape_index]` at line 269 when `geo` is already in hand at both call sites (403 and 537). Minor, but these are easy loads to drop in a change that is about performance.

**4. `collide.py:837`:** `_pairs` lost its `-> wp.array[wp.vec2i]` return annotation. Worth restoring to stay consistent with the bracket-annotation convention described in `AGENTS.md`.

newton/_src/geometry/soft_contacts_sdf.py
```diff
@@ -471,7 +614,19 @@ def create_soft_edge_contacts(
...
+    state,
+    contacts,
+    margin: float,
+    device,
+    edge_pairs,
+    face_pairs,
+    n_particle_pairs,
+    shape_aabb_lower=None,
+    shape_aabb_upper=None,
+):
```
eric-heiden, line 582: Can you add type hints here?

changelog/+vbd-soft-contact-perf-9394aa08.changed.md
```diff
@@ -0,0 +1 @@
...very color and iteration, sort contact-detection candidate pairs by shape, reject far-apart edge/face candidates with a conservative AABB test before the SDF search, pack two cloth particles per warp in the tiled elasticity solve, and select a tetrahedra-only elasticity kernel when triangle and edge stiffnesses are inactive. Solver settings and contact capacity are unchanged, and CPU and deterministic modes keep the previous kernels. The shape-major candidate ordering applies on every device, so soft-contact records are emitted in a different order: the contact set is identical, but summation-order differences can shift results within floating-point tolerance on CPU and deterministic paths. Both contact-processing changes target scenes whose candidate sets are much larger than their active contacts (for example many replicated worlds): when most candidates are in active contact, the shape-major ordering can make the edge/face generation pass itself roughly 20% slower (full-step impact within 1-2%), and at small contact capacities (tens of thousands of slots) the per-particle gather kernel can be a few percent slower than the scatter it replaces, with no measurable full-step effect.
```
eric-heiden: Can you make this entry a bit shorter, or split it up into multiple pieces?

newton/_src/geometry/soft_contacts_sdf.py
```diff
@@ -221,21 +221,55 @@ def optimize_face_sdf(
...
+    """Return the shape-local-to-world transform for a supplied body-local frame."""
     rigid_body = shape_body[shape_index]
     X_wb = wp.transform_identity()
     if rigid_body >= 0:
         X_wb = body_q[rigid_body]
-    X_bs = shape_transform[shape_index]
-    X_ws = wp.transform_multiply(X_wb, X_bs)
-    X_sw = wp.transform_inverse(X_ws)
-    return X_bs, X_ws, X_sw
+    return wp.transform_multiply(X_wb, X_bs)
```
eric-heiden: Nit: you can also write `X_wb * X_bs`.

newton/_src/geometry/soft_contacts_sdf.py
```diff
@@ -471,7 +568,19 @@ def create_soft_edge_contacts(
...
+    *,
+    model,
+    state,
+    contacts,
+    margin: float,
+    device,
+    edge_pairs: wp.array[wp.vec2i],
+    face_pairs: wp.array[wp.vec2i],
+    n_particle_pairs: int,
+    shape_aabb_lower: wp.array[wp.vec3] | None = None,
```
eric-heiden, line 580: 🟠 `wp.array[wp.vec3] | None` gets evaluated at import time here, and on Python 3.10 that raises before the module finishes loading:

```
TypeError: Union[arg, ...]: each arg must be a type. Got wp.array[wp.vec3f].
```

`wp.array[wp.vec3]` returns a `_ArrayAnnotation` instance rather than a class, and Warp implements `__or__` as `typing.Union[self, other]`. Python 3.11 and newer accept the non-type and produce `typing.Optional[wp.array[wp.vec3f]]`, but 3.10's `typing._type_check` rejects it. Since `pyproject.toml` sets `requires-python = ">=3.10"` and lists 3.10 as a supported classifier, this makes `newton` unimportable on a version we advertise. CI stays green only because `.python-version` pins 3.12.

There is no need to drop the `|` here. Adding the future import at the top of the module keeps the union syntax and fixes 3.10:

```py
Nothing imports this module except the collision launcher, so it is cycle-free.
"""

import warp as wp
```

That defers every annotation in the file to a string, which is safe for the `wp.func` and `wp.kernel` definitions in this module: Warp resolves the strings at kernel build time. Several Newton modules that mix kernels with this import already depend on that behavior, for example `newton/_src/solvers/mujoco/kernels.py` and `newton/_src/sensors/sensor_contact.py`.

I checked both directions on 3.10: the file as written fails to import, and it imports cleanly with the future import added. On 3.12 the full-surface soft contact tests behave identically with and without it.

newton/_src/sim/collide.py
AnkaChan: Reproduced your setup's direction, and measured two more points on an L40: with eight large mesh-SDF shapes (30 MB sparse textures each) under dense contact the sorted order flips back to a 1.39x win on the edge/face passes, and on the 1024-world workload it is the main generation win (2.6x). So the slow case is specifically a few small same-type shapes in dense contact, where the whole pass is sub-millisecond anyway (+0.1 ms in your scene). The changelog now states this. Net: the cost lands in scenes with headroom, the win is milliseconds per step in the scenes that are bound, so we kept the sort. Thanks also for checking the CPU arm; matches what we measured.

newton/_src/solvers/vbd/particle_vbd_kernels.py
AnkaChan: Confirmed your numbers on the same example on an L40: scatter is about 7 percent faster at kernel level across 4 runs per arm, matching your 5 to 8, and frame times fully overlap. At the other end (12.7M-slot capacity, 3 percent occupancy, 1024 worlds) the gather is the largest single win in this PR, about 9 ms per env step. The changelog now states the regime next to the claim. If you would rather have this settled in code, a capacity threshold on the gather dispatch is a small follow-up; we would measure the crossover before picking a number.

newton/_src/geometry/soft_contacts_sdf.py
AnkaChan: All four addressed in 2ebcad95. We also went one step further: an isolated ablation showed the deferred shape-frame restructuring had no measurable benefit on either workload, so 25dbb015 restores the eager _shape_frames computation, which puts the helper back in production use. The generation kernels now differ from main only by the AABB rejection.

newton/_src/geometry/soft_contacts_sdf.py
AnkaChan: Added in 2ebcad95 (and see the 3.10 fix below, 819b5ee7).

newton/_src/geometry/soft_contacts_sdf.py
AnkaChan: Good catch, and thanks for verifying both directions. Fixed in 819b5ee7 with the future import as suggested.

changelog/+vbd-soft-contact-perf-9394aa08.changed.md
AnkaChan: Split into two shorter entries in 819b5ee7: the acceleration summary and a separate candidate-ordering entry carrying the ordering and regime notes.

AnkaChan:
Thanks for the thorough review and the measurements. We reproduced both of your perf results on our side and measured the opposite regimes as well (details in the threads). The short version: where the new ordering and the per-particle gather lose, the affected pass is a fraction of a millisecond in scenes with plenty of headroom; where they win, they save milliseconds per step in the many-world and large-mesh scenes this PR targets. We think that trade is worthwhile, and the changelog now states both sides. All code comments are addressed (2ebcad95, 25dbb015, 819b5ee7), the changelog is split and shortened, and the branch is merged with main.

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -697,7 +726,9 @@ def _init_particle_system(
             print("Info: Tiled solve requires model.device='cuda'. Tiled solve is disabled.")
 
         self.use_particle_tile_solve = particle_enable_tile_solve and model.device.is_cuda
-
+        # Model element materials are static solver inputs. Rebuild the solver after changing
+        # triangle or edge stiffness so this graph-stable specialization is recomputed.
+        self._use_tet_only_tile_solve = self.use_particle_tile_solve and _is_tet_only_elasticity_model(model)
```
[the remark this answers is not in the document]
AnkaChan: Resolving with the static contract, which now has a precise justification: SolverVBD.notify_model_changed's ModelFlags set (JOINT_PROPERTIES, JOINT_DOF_PROPERTIES, BODY_PROPERTIES, BODY_INERTIAL_PROPERTIES, SHAPE_PROPERTIES, MODEL_PROPERTIES, CONSTRAINT_PROPERTIES) contains no flag for triangle or edge material arrays, so live tri/edge stiffness mutation is not a supported update channel on main. The tet-only eligibility check is therefore consistent with the solver's update model: like the other construction-time choices it is recomputed on solver rebuild, and a comment at the specialization site documents that contract. If a tri/edge-material ModelFlags entry is added later, wiring the gate into notify_model_changed is a two-line follow-up.

newton/_src/sim/collide.py
AnkaChan: Reproduced your result. Two more datapoints on an L40: with eight large mesh-SDF shapes in dense contact the sort is 1.39x faster on the edge/face passes, and on the 1024-world scene it's the main generation win (2.6x). So it only loses on small scenes with a few same-type shapes -- cheap enough that the slowdown doesn't matter -- and wins on the big ones. Kept it; changelog now says so. Thanks for checking the CPU arm too.

newton/_src/solvers/vbd/particle_vbd_kernels.py
AnkaChan: Confirmed your numbers -- scatter ~7% faster here, frames overlap. That's a cheap scene where it doesn't matter; on the 1024-world scene the gather is the biggest single win in the PR (~9 ms/step). Changelog states the regime. If you'd rather gate it in code, a capacity threshold is an easy follow-up -- I'd measure the crossover first.

newton/_src/geometry/soft_contacts_sdf.py
AnkaChan: All four fixed in 2ebcad95. Also removed the deferred shape-frame refactor -- an ablation showed no benefit -- so _shape_frames is back in use and the generation kernels now differ from main only by the AABB reject.

newton/_src/geometry/soft_contacts_sdf.py
AnkaChan: Added in 2ebcad95 (and the 3.10 fix below, 819b5ee7).

newton/_src/geometry/soft_contacts_sdf.py
AnkaChan: Good catch, and thanks for checking both directions. Fixed in 819b5ee7 with the future import.

newton/_src/geometry/soft_contacts_sdf.py
AnkaChan: Done in 819b5ee7.

changelog/+vbd-soft-contact-perf-9394aa08.changed.md
AnkaChan: Split into two shorter entries in 819b5ee7.

AnkaChan:
Thanks for the careful review and measurements. We reproduced both your results and also measured the cases they don't cover. Short version: the sort and the gather can be a bit slower on small, cheap scenes -- where nothing is performance-bound and it doesn't matter -- and win big on the large many-world and big-mesh scenes this PR targets. The changelog states both sides. All code comments addressed (2ebcad95, 25dbb015, 819b5ee7); changelog split and shortened; merged with main.

eric-heiden (dismissed):
Thanks!

newton/tests/test_solver_vbd.py
```diff
@@ -397,6 +405,152 @@ def _eval_directional_joint_projection_kernel(
...
+        contact_count[0] = raw_count[0]
+    if i < particle_contact_head.shape[0]:
+        particle_contact_head[i] = -1
+        particle_forces[i] = wp.vec3(0.0)
+        particle_hessians[i] = wp.mat33(0.0)
+
+
+@wp.kernel
+def accumulate_particle_body_contact_force_and_hessian(
+    # inputs
```
eric-heiden, line 440: 🔵 `accumulate_particle_body_contact_force_and_hessian` was deleted from `newton/_src/solvers/vbd/particle_vbd_kernels.py` and re-declared verbatim here, so the tests that launch it now check a test-local copy instead of code that ships.

Four tests that ran against the production kernel on `main` are affected. They are all registered with `devices=devices` and reach this copy through `_run_face_section2` and the damping helper:

| Test | No longer guards |
|---|---|
| `test_barycentric_force_distribution` | the barycentric split (`weight = bary[corner]`, `weight * weight` on the Hessian) |
| `test_edge_face_uses_shape_margin` | per-shape margin selection |
| `test_edge_face_mixes_shape_material` | per-shape material mixing |
| `test_body_particle_contact_damping_ignores_penalty_ramp` | damping staying independent of the penalty ramp |

A regression in `gather_particle_body_contact_force_and_hessian` would leave all four of them green. The equivalence test that replaces them, `test_particle_contact_gather_matches_legacy`, is registered with `devices=cuda_devices`, so the shipping gather has no value-comparing coverage on CPU even though `_solve_particle_iteration` launches it on every device. `test_particle_contact_gather_order_pinned` is the only gather test that runs on CPU, and it compares gather output against gather output rather than against a reference.

Either of these would close the gap:

1. Point `_run_face_section2` and the damping wrapper at `gather_particle_body_contact_force_and_hessian`, building the adjacency with `build_particle_body_contact_adjacency_active` the way `_launch_particle_contact_gather` already does. That puts the four tests back on shipping code.
2. At minimum, register `test_particle_contact_gather_matches_legacy` with `devices=devices` so the shipping kernel's values are checked on CPU too.

AnkaChan: Good catch — confirmed. Moving `accumulate_particle_body_contact_force_and_hessian` into the test module left `test_barycentric_force_distribution`, `test_edge_face_uses_shape_margin`, `test_edge_face_mixes_shape_material`, and `test_body_particle_contact_damping_ignores_penalty_ramp` running the test-local copy, and `test_particle_contact_gather_matches_legacy` was CUDA-only, so `gather_particle_body_contact_force_and_hessian` had no value-comparing coverage on CPU.

Both suggestions done in 584d2bad:
1. `_run_face_section2` and the damping test now launch `build_particle_body_contact_adjacency_active` + `gather_particle_body_contact_force_and_hessian` — the same two-kernel sequence `SolverVBD` uses — so those four tests guard the production kernel again. The copy remains only as the reference oracle inside `test_particle_contact_gather_matches_legacy`.
2. `test_particle_contact_gather_matches_legacy` is registered with `devices=devices`, so the gather kernel's values are checked against the reference on CPU as well as CUDA.

Verified: with a deliberately perturbed Hessian weight in `gather_particle_body_contact_force_and_hessian` the repointed tests fail; unperturbed, all six affected tests pass on CPU and cuda:0.

eric-heiden (approved):
Thanks!

PR #4004, opened 2026-08-22
Harvest VBD proxy wrenches from the capped per-body contact list

### LINKED ISSUE #3795: [BUG] VBD proxy wrench harvest ignores the per-body soft-contact buffer cap
### Bug Description

`SolverVBD`'s proxy-body wrench harvest reports contact reactions that its own solve never applied, once a proxy-coupled rigid body exceeds the per-body soft-contact buffer.

**The destination solve is capped per body.** `build_body_particle_contact_lists` bins soft contacts into a per-body list whose capacity is `rigid_body_particle_contact_buffer_size` (default `256`, `newton/_src/solvers/vbd/solver_vbd.py:266`). The rigid soft-contact solve then truncates explicitly:

```python
# newton/_src/solvers/vbd/rigid_vbd_kernels.py:3911-3913
num_contacts = body_particle_contact_counts[body_id]
if num_contacts > body_particle_contact_buffer_pre_alloc:
    num_contacts = body_particle_contact_buffer_pre_alloc
```

Overflow is reported only by a device-side `wp.printf` warning (`check_contact_overflow`, `rigid_vbd_kernels.py:2835-2854`). It is not an error, and the step continues.

**The harvest is not capped.** `coupling_harvest_proxy_wrenches()` launches over the flat global contact stream (`solver_vbd.py:1007-1037`, `dim=contacts.soft_contact_max`), and the kernel's only bound is the *global* soft-contact count:

```python
# newton/_src/solvers/vbd/vbd_coupling_kernels.py:155-157
contact_idx = wp.tid()
if contact_idx >= body_particle_contact_count[0]:
    return
```

Every soft contact whose shape maps to a proxy-coupled body therefore contributes a wrench through `wp.atomic_add(out_body_f, ...)` -- including the contacts that were dropped from that body's list and never acted on the soft body.

**Consequence.** For a body carrying more than `rigid_body_particle_contact_buffer_size` soft contacts in a step, the soft body is pushed by at most `N` contacts while the source solver's rigid body is pushed back by all of them. The two sides of the coupling disagree about the same contact set, so momentum is injected into the system -- silently, apart from a `printf` on the device. The error grows with the size of the overflow, so it shows up precisely in the dense-contact scenarios that motivate full-surface contacts (cloth wrapping a gripper, a bag closing around a box).

**Pre-existing, but #3756 makes it much easier to reach.** The unbounded launch predates PR #3756 -- the previous kernel read `soft_contact_particle` under the same guard and the same `dim`. What changes is record density: enabling full-surface (edge/face) records produces several records where the particle-only path produced one for the same overlap, so the per-body cap is hit far sooner. Filing as a follow-up rather than a blocker on #3756.

### Suggested fix

**Preferred -- drive the harvest from the same list the solve uses.** `self.body_particle_contact_indices` and `self.body_particle_contact_counts` are already built on the solver each refresh, and they are guaranteed to exist whenever this harvest runs: the VBD-specific harvest is gated on `_coupling_has_rigid_avbd_state` (`solver_vbd.py:961`), which is exactly `not integrate_with_external_rigid_solver and body_count > 0` (`solver_vbd.py:528`) -- the same condition that builds the lists (`solver_vbd.py:2508`). The harvest can iterate `min(counts[body], pre_alloc)` entries per body (launch `dim = body_count * pre_alloc`, or keep the flat launch and skip a record whose slot in its owning body's list is beyond `pre_alloc`). The feedback wrench then equals, by construction, the sum of the forces the destination solve applied, and it stays consistent with the list's refresh cadence.

**Fallback, if the harvest must stay flat** -- surface the overflow on the host for proxy-coupled bodies (raise, or log a warning once per step) so silent momentum injection cannot pass unnoticed.

### Reproduction Script

```python
# Sketch, not yet executed -- described rather than run, to keep the repro small.
#
# 1. Build a scene with one rigid body and a soft mesh dense enough to generate
#    more than N soft contacts against that body in a single step.
# 2. Drive it with SolverCoupledProxy: a rigid source entry proxied into a
#    SolverVBD destination entry.
# 3. Construct the destination solver with a small buffer so the repro stays cheap:
#       newton.solvers.SolverVBD(view, rigid_body_particle_contact_buffer_size=4)
# 4. Step once and compare:
#       - the wrench harvested into out_body_f (all soft contacts), against
#       - the reaction implied by the <= 4 contacts the body's list actually kept.
#    They diverge; the gap is the injected momentum. Below the cap they agree.
```

### Acceptance criteria

- Regression test along the lines above, using a small `rigid_body_particle_contact_buffer_size`, asserting the harvested `out_body_f` matches the reaction from the truncated contact set. Verify it fails before the fix.
- No behavior change below the cap.

### System Information

- Branch `mym/vbd-proxy-full-surface-contact` @ `87b6513` (PR #3756); the unbounded launch is also present on `main`.
- Applies to `SolverCoupledProxy` with a `SolverVBD` destination on all devices; no GPU-specific behavior.

deepspace28:
## Description

`SolverVBD`'s proxy-body wrench harvest launched over the flat soft-contact stream, bounded
only by the *global* soft-contact count. The destination rigid solve reads a **per-body**
adjacency list truncated at `rigid_body_particle_contact_buffer_size`
(`accumulate_body_particle_contacts_per_body`, `rigid_vbd_kernels.py:5426-5428`). A
proxy-coupled body carrying more soft contacts than that list holds was pushed by at most
`N` of them, while the source solver was handed the reaction from all of them — momentum
the two sides disagree about, reported on the device by a `wp.printf` and nothing else.

This takes the issue's preferred fix: drive the harvest from the same list the solve uses.
The kernel now mirrors `accumulate_body_particle_contacts_per_body` — `dim = body_count *
_NUM_CONTACT_THREADS_PER_BODY`, each thread striding its body's list and applying the same
`min(counts[body], pre_alloc)` truncation. The feedback wrench is then, by construction, the
sum of the reactions the destination solve applied, and it tracks the list's refresh cadence
for free.

The per-contact evaluation is byte-for-byte the same call to `_eval_soft_ef_contact`, so
nothing changes below the cap. Above it, the harvest now reports what the solve applied
rather than what the collider detected.

Two incidental consequences worth naming:

- **Accumulation is now per body, not per contact.** Each thread sums into registers and
  issues one `wp.atomic_add`, instead of one per contact. Fewer atomics; the sum is still
  order-nondeterministic across threads, as before.
- **Bodies the list-builder skips are no longer harvested.**
  `build_body_particle_contact_lists` bins only bodies with `body_inv_mass_effective > 0`,
  so a proxy whose destination copy is static or kinematic now contributes nothing. It
  contributed nothing to the destination solve either, and a proxy is only immobile in the
  destination when its source counterpart is, where the fed-back wrench was inert. The
  coupling machinery never puts a proxy body in the dynamics-disabled set
  (`solver_coupled.py:590-591` excludes `proxy_body_keep`), so this is not reachable through
  `SolverCoupledProxy` configuration.

Closes #3795.

## Checklist

## Test plan

`test_proxy_wrench_harvest_respects_body_particle_contact_buffer_cap` in
`newton/tests/test_coupled_solver.py`, built to the issue's acceptance criteria.

Twelve particles are pressed into one flat, uniform box face at the same depth, so every
soft contact carries the same force. Which `cap` records survive the list-builder's
`wp.atomic_add` is not deterministic, but with identical contacts *any* subset of a given
size sums to the same linear reaction — so the assertion holds regardless of which ones the
build kept. The body is given mass and inertia `1e6` so one step at `dt = 1/60` leaves the
contact geometry intact; gravity is zero.

```
buffer = 12 (reference): soft_contact_count 12, listed 12, overflow 0
buffer =  4 (overflows): soft_contact_count 12, overflow_max 12 > 4
assert  F_capped == F_full * 4/12
```

**Verified to fail without the fix** (`git stash` of the two source files, test kept):

```
AssertionError:
Not equal to tolerance rtol=1e-05, atol=1e-06
 [2]: -141.30818 (ACTUAL), -47.102547 (DESIRED)
Max relative difference among violations: 2.0000114
```

`-141.308 / -47.103 = 3.000` exactly — twelve contacts harvested where four were applied.
With the fix the test passes, and the run still prints
`Warning: Per-body particle contact buffer overflowed 12 > 4`, so the test cannot silently
stop exercising the overflow path.

Suites, on `cpu` and `cuda:0`:

```console
python -m unittest newton.tests.test_coupled_solver   # 101 tests
python -m unittest newton.tests.test_solver_vbd       # 123 tests
uvx pre-commit run --files newton/_src/solvers/vbd/vbd_coupling_kernels.py \
    newton/_src/solvers/vbd/solver_vbd.py newton/tests/test_coupled_solver.py \
    changelog/3795.fixed.md
uvx --from towncrier==25.8.0 towncrier build --draft --version 1.6.0 --date 2026-08-21
```

Three failures, all reproduced on clean `main` at `f12f7c81` and unrelated to this change:

| Test | Baseline on `main` |
| --- | --- |
| `TestSolverCoupledBasic.test_compacted_joint_targets_use_local_layout` | fails identically (`copy source buffer with size (36) to offset (0) is larger than destination size (32)`) |
| `TestSolverVBDCouplingHooks.test_reset_preserves_pose_history_cuda_0` | fails identically; NVRTC on this machine (`CUDA kernel build failed with error code 6`) |
| `TestVBDFullSurfaceContact.test_edge_face_pushes_vertices_out_cuda_0` | fails identically |

## Bug fix

**Steps to reproduce:**

1. Proxy-couple a rigid source entry into a `SolverVBD` destination entry.
2. Give the proxied body more soft contacts in one step than
   `rigid_body_particle_contact_buffer_size` (default `256`). Enabling full-surface
   edge/face records reaches this quickly — a cloth of `V` vertices needs `V + F + E`
   entries, and @AhmedFetouh-26 [measured 56,362 against the default 256](https://github.com/newton-physics/newton/issues/3795#issuecomment-5307883938)
   in production.
3. Step. The soft body is pushed by the contacts the collider found; the destination solve
   pushes the proxy back with at most `256` of them; the source solver is handed all of
   them.

**Before:** the harvested wrench scales with the *detected* contact count, so the source
body accelerates with no physical counterpart. Nothing on the host reports it — only a
device-side `wp.printf` from `check_contact_overflow`.

**After:** the harvested wrench scales with the *applied* contact count.

## Note on the issue's fallback suggestion

The issue offers surfacing the overflow host-side as a fallback "if the harvest must stay
flat", and the confirming reporter asked for it on the grounds that sizing the buffer
requires knowing `V + F + E` in advance. That is not included here: reading
`body_particle_contact_overflow_max` back on the host costs a synchronisation every step and
would not survive CUDA graph capture, which is presumably why the existing warning is a
device-side `printf`. It is also a separate concern from this fix — with the harvest capped,
overflow degrades contact resolution but no longer breaks the coupling's momentum balance.
Happy to add an opt-in accessor (a property that reads the counter only when asked) in a
follow-up if that is wanted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rigid-body reaction handling when body-particle contact buffers reach capacity.
  * Prevented overflow contacts from contributing unintended momentum during coupled solves.
  * Ensured proxy-body wrench feedback reflects only contacts retained by the configured buffer limit.

* **Tests**
  * Added regression coverage for capped and uncapped contact buffers, including overflow accounting and reaction scaling.


gdaviet (dismissed):
Great find, thanks for the MR!

ix looks good to me

deepspace28:
@gdaviet thanks for looking at it.

Heads-up that the approval was auto-dismissed about an hour later by my push of `42b498e`. That commit is tests-only (+9/-1 across two test files): the direct kernel launch in `test_solver_vbd_proxy_full_surface` now has to pass the buffer capacity and the per-body index list to match the new harvest signature, and the new coupling test passes `rigid_compliant_alm` explicitly so the constructor stops raising its deprecation warning. Nothing in `solver_vbd.py` or `vbd_coupling_kernels.py` has changed since you reviewed.

The full matrix is green now — unittests on all four platforms, docs, licenses, lockfile, and Codecov reports every modified line covered. Could you re-approve when you get a chance? Happy to make any other change it needs.

gdaviet (approved):

eric-heiden (approved):

Issue #3977, opened 2026-08-19
[BUG] Newton Visualizer used within Isaac Lab has rendering issues


aneangel:
### Bug Description

When viewing a policy trained in Isaac Lab with the newton visualize flag the newton visualizer seems to have trouble rendering the simulation when moving the camera from any point other than origin whether that is panning, zooming, or moving along XYZ axis.

### Reproduction Script

```
python scripts/environments/zero_agent.py \
    --task Isaac-Velocity-Flat-G1 \
    --num_envs 9 \
    --visualizer newton
```

aneangel:
Has been verified as a stand alone issue from Isaac Lab's side, with the resolving PR already filed on Isaac Lab

jcarius-nv:
Closing since not a Newton issue.

PR #4006, opened 2026-08-23
Fix ViewerGL large-mesh depth instability

ooctipus:
## Description

Fix ViewerGL corruption when the camera moves over a mesh whose triangles are much larger than the view frustum.

The renderer now:

- transforms shapes relative to the camera before view rotation;
- reconstructs each fragment's camera-relative position from `gl_FragCoord.w` and writes corrected depth;
- uses the reconstructed position consistently for shading, fog, shadows, and spotlights;
- renders the shadow pass relative to the same camera origin; and
- gives the edge-overlay pass that same camera-relative origin.

There are no terrain names, mesh-size thresholds, texture changes, or Isaac Lab special cases. Collision geometry and terrain dimensions are unchanged. This fixes the rendering failure reported in #3977 instead of requiring Isaac Lab to reduce its visual ground from 2,000 km to 2 km.

The separate USD projected-texture mapping fix is #4007.

## Checklist

## Test plan

- `uv run --extra dev -m newton.tests -k test_large_quad_is_stable_under_parallel_camera_translation`
- `uv run --extra dev -m newton.tests -k test_viewer_get_frame` (9 passed)
- `uv run --extra dev -m newton.tests -k viewer` (228 passed)
- `uvx pre-commit run -a`
- The strengthened 200,000 km stress case fails with the exact `origin/main` shape shaders: camera translation changes 1.16% of pixels on an RTX 5090, above the 1% limit. The fixed shaders change 0.61%.
- Forcing the edge shader back to a zero camera origin fails the new visible-edge assertion.
- Representative headless Isaac Lab G1 render benchmark: approximately 0.19 ms/frame added at more than 400 rendered frames/s. The inverse projection is cached until the camera projection changes.

## Bug fix

**Steps to reproduce:**

1. Render a two-triangle quad scaled to 2,000,000 m in ViewerGL.
2. Place ordinary geometry on the quad and enable shadows.
3. Translate the camera and target together by 1 m.
4. Observe large regions of the quad incorrectly occlude or erase geometry, with unstable depth and shadows.

**Minimal reproduction:**

```console
uv run --extra dev -m newton.tests -k test_large_quad_is_stable_under_parallel_camera_translation
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rendering stability for very large or offset meshes while moving the camera.
  * Stabilized depth, shading, fog, lighting, and shadows across camera movement.
  * Improved shadow positioning and localization in large scenes.

* **Tests**
  * Added coverage validating stable rendering with shadows enabled and disabled.

### DIFF
--- a/newton/_src/viewer/gl/shaders.py
+++ b/newton/_src/viewer/gl/shaders.py
@@ -16,11 +16,13 @@ layout (location = 5) in vec4 aInstanceTransform2;
 layout (location = 6) in vec4 aInstanceTransform3;
 
 uniform mat4 light_space_matrix;
+uniform vec3 render_origin;
 
 void main()
 {
     mat4 transform = mat4(aInstanceTransform0, aInstanceTransform1, aInstanceTransform2, aInstanceTransform3);
-    gl_Position = light_space_matrix * transform * vec4(aPos, 1.0);
+    vec3 render_position = mat3(transform) * aPos + transform[3].xyz - render_origin;
+    gl_Position = light_space_matrix * vec4(render_position, 1.0);
 }
 """
 
@@ -51,23 +53,22 @@ layout (location = 8) in vec4 aMaterial;
 
 uniform mat4 view;
 uniform mat4 projection;
-uniform mat4 light_space_matrix;
+uniform vec3 view_pos;
 
 out vec3 Normal;
-out vec3 FragPos;
 out vec3 LocalPos;
 out vec2 TexCoord;
 out vec3 ObjectColor;
-out vec4 FragPosLightSpace;
 out vec4 Material;
 
 void main()
 {
     mat4 transform = mat4(aInstanceTransform0, aInstanceTransform1, aInstanceTransform2, aInstanceTransform3);
 
-    vec4 worldPos = transform * vec4(aPos, 1.0);
-    gl_Position = projection * view * worldPos;
-    FragPos = vec3(worldPos);
+    // Subtract before rotating to avoid cancellation in a translated view matrix for large coordinates.
+    vec3 camera_relative_position = mat3(transform) * aPos + transform[3].xyz - view_pos;
+    vec3 view_position = mat3(view) * camera_relative_position;
+    gl_Position = projection * vec4(view_position, 1.0);
     LocalPos = aPos;
 
     mat3 rotation = mat3(transform);
@@ -81,7 +82,6 @@ void main()
     Normal = normalMatrix * aNormal;
     TexCoord = aTexCoord;
     ObjectColor = aObjectColor;
-    FragPosLightSpace = light_space_matrix * worldPos;
     Material = aMaterial;
 }
 """
@@ -91,13 +91,15 @@ shape_fragment_shader = """
 out vec4 FragColor;
 
 in vec3 Normal;
-in vec3 FragPos;
 in vec3 LocalPos;
 in vec2 TexCoord;
-in vec3 ObjectColor; // used as albedo
-in vec4 FragPosLightSpace;
+in vec3 ObjectColor;
 in vec4 Material;
 
+uniform mat4 view;
+uniform mat4 projection;
+uniform mat4 inverse_projection;
+uniform vec2 viewport_size;
 uniform vec3 view_pos;
 uniform vec3 light_color;
 uniform vec3 sky_color;
@@ -167,7 +169,33 @@ vec2 poissonDisk[16] = vec2[](
    vec2( 0.14383161, -0.14100790 )
 );
 
-float ShadowCalculation()
+vec3 ReconstructCameraRelativePosition()
+{
+    vec2 ndc_xy = gl_FragCoord.xy / viewport_size * 2.0 - 1.0;
+    vec4 near_h = inverse_projection * vec4(ndc_xy, -1.0, 1.0);
+    vec3 near_view = near_h.xyz / near_h.w;
+
+    // gl_FragCoord.w retains reciprocal clip-space W even when interpolated
+    // clip-space depth loses precision across a triangle much larger than the
+    // view frustum. Recover the view-space point along this pixel's camera ray.
+    vec4 clip_w_row = vec4(projection[0][3], projection[1][3], projection[2][3], projection[3][3]);
+    float denominator = dot(clip_w_row.xyz, near_view);
+    if (abs(gl_FragCoord.w) < 1.0e-8 || abs(denominator) < 1.0e-8)
+    {
+        float ndc_z = gl_FragCoord.z * 2.0 - 1.0;
+        vec4 view_h = inverse_projection * vec4(ndc_xy, ndc_z, 1.0);
+        gl_FragDepth = gl_FragCoord.z;
+        return transpose(mat3(view)) * (view_h.xyz / view_h.w);
+    }
+
+    float amount = (1.0 / gl_FragCoord.w - clip_w_row.w) / denominator;
+    vec3 view_position = near_view * amount;
+    vec4 clip_position = projection * vec4(view_position, 1.0);
+    gl_FragDepth = clip_position.z / clip_position.w * 0.5 + 0.5;
+    return transpose(mat3(view)) * view_position;
+}
+
+float ShadowCalculation(vec3 camera_to_fragment)
 {
     vec3 normal = normalize(Normal);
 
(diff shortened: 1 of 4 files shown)


eric-heiden:
Nice work tracking this down, and moving the camera offset into a shader uniform is a clean way to keep large scenes numerically stable.

Two things look worth resolving before merge: the edge overlay pass links the updated vertex shader but never uploads the new `view_pos` uniform, and the new regression test still passes with the pre-fix shaders on the software GL stack that headless CI runs on. Details are inline.

newton/_src/viewer/gl/shaders.py
```diff
@@ -51,23 +53,22 @@
 
 uniform mat4 view;
 uniform mat4 projection;
-uniform mat4 light_space_matrix;
+uniform vec3 view_pos;
```
eric-heiden, line 56: 🟠 `ShaderEdge` links this same vertex shader but never queries or uploads `view_pos`, so the edge overlay runs with this uniform at its GL default of `vec3(0)`.

`ShaderEdge.__init__` looks up only `view`, `projection`, `edge_color` and `light_space_matrix`, and `update()` uploads those same four. Because `view * vec4(p, 1.0)` equals `mat3(view) * p + t` with `t = -R * eye`, every edge vertex is now rasterized at view space `p_view + R * eye`. With the stock camera roughly 12 units from its pivot, geometry near the pivot lands at view z close to 0 and gets near-clipped, while everything else is grossly displaced.

This is reachable from the public `renderer.draw_edges` toggle, which the GUI exposes as the "Show Edges" checkbox and which drives the edge pass in `opengl.py`.

To give the edge program the same camera relative origin as the shape pass:

- add `self.loc_view_pos = self._get_uniform_location("view_pos")` in `ShaderEdge.__init__`
- thread a `view_pos` argument through `ShaderEdge.update` and upload it with `glUniform3f`
- pass `view_pos=self.camera.pos` from the `self._edge_shader.update(...)` call in `_render_scene`

A small test would help here as well, since nothing under `newton/tests/` currently exercises `draw_edges`.

newton/tests/test_viewer_get_frame.py
```diff
@@ -183,6 +183,103 @@ def test_headless_frame_capture_across_devices(self):
...
+            normals = wp.array([(0.0, 0.0, 1.0)] * 4, dtype=wp.vec3, device=viewer.device)
+            # Deliberately offset the giant triangle pair from the camera. This mirrors
+            # imported USD ground assets and exposes clip-depth interpolation error that
+            # a perfectly centered quad can accidentally hide.
+            xforms = wp.array(
+                [wp.transform((-8500.0, 8500.0, 0.0), wp.quat_identity())],
+                dtype=wp.transform,
+                device=viewer.device,
+            )
+            scales = wp.array([(2.0e6, 2.0e6, 1.0)], dtype=wp.vec3, device=viewer.device)
```
eric-heiden: 🔵 This test passes on the branch, but it also passes with the shaders from `origin/main` swapped back in, so it does not actually guard the fix. `AGENTS.md` asks that regression tests fail without the fix.

Replaying the test's exact scene and camera moves against both shader versions, with only the ground quad scale and matching instance offset varied (thresholds are changed fraction under 1% and mean absolute delta under 1.0):

| quad scale / offset | this branch | pre-fix shaders | fails without the fix? |
| --- | --- | --- | --- |
| 2e6 / 8500 (current) | 0.24%, mean 0.551 | 0.78%, mean 0.703 | no |
| 2e7 / 85000 | 3.68%, mean 7.388 | 6.60%, mean 12.253 | yes |
| 2e8 / 850000 | 0.59%, mean 0.253 | 100.00%, mean 222.517 | yes |

At the scale the test picks today, the pre-fix shaders score 0.78% against the 1% limit and 0.703 against the 1.0 limit, so a revert of the shader change would leave this test green. Bumping the scale gives real separation, roughly 380x instead of the current 1.3x, while the branch stays comfortably inside both thresholds:

```suggestion
            scales = wp.array([(2.0e8, 2.0e8, 1.0)], dtype=wp.vec3, device=viewer.device)
```

Pair that with the instance transform above set to `(-850000.0, 850000.0, 0.0)`. Keeping the offset in step with the scale matters: at 2e8 with the current 8500 offset the frame stops changing at all, and the check would pass vacuously.

One caveat: these numbers come from Mesa llvmpipe. Discrete GPU interpolation precision may differ, so the current test could discriminate on some hardware, but this test only ever runs through the headless path, and headless CI is the software path where it does not.

The underlying fix itself looks right, this is only about the strength of the guard.

newton/_src/viewer/gl/shaders.py
```diff
@@ -264,6 +289,9 @@
 
 void main()
 {
+    vec3 camera_to_fragment = ReconstructCameraRelativePosition();
+    vec3 view_position = mat3(view) * camera_to_fragment;
```
eric-heiden: ⚪ Three small optional cleanups, grouped here so they are easy to skip or take together.

1. **Dead `light_space_matrix` plumbing.** `ShaderEdge` still queries and uploads `light_space_matrix`, and `_render_scene` still threads it in, but this PR removed that uniform from `shape_vertex_shader` and `edge_fragment_shader` never declared it. `loc_light_space_matrix` is now `-1`, so the upload is a silent no-op and the whole argument can be dropped from `ShaderEdge.update` and its call site.
2. **Redundant view round trip.** `ReconstructCameraRelativePosition()` returns `transpose(mat3(view)) * view_position`, and this line converts it straight back. The only consumer is the fog distance further down, and `mat3(view)` is orthonormal, so `length(view_position)` equals `length(camera_to_fragment)`. Using `length(camera_to_fragment)` for the fog lets this per-fragment 3x3 multiply go away entirely.
3. **Hidden depth side effect.** `ReconstructCameraRelativePosition()` also writes `gl_FragDepth` on both exits, which the name does not hint at. A one-line comment right at this call site would save the next reader a trip through the function.

newton/_src/viewer/gl/shaders.py
ooctipus: Addressed in 61af2067. ShaderEdge now uploads the same camera position used by the shape pass, RendererGL threads it through, and the dead edge-only light-space-matrix plumbing is removed. The framebuffer regression now exercises draw_edges and asserts that a visible overlay is produced; forcing the edge origin back to zero makes it fail.

newton/tests/test_viewer_get_frame.py
ooctipus: Addressed in 61af2067 using the suggested 2e8 scale and 850000 offset. I also verified the guard by swapping in the exact origin/main shape shaders: they fail at 1.16% changed pixels on an RTX 5090, while the fixed shaders pass at 0.61%. The mean-delta ceiling is 2.0 because the fixed NVIDIA path measures 1.075 even though the llvmpipe result is lower.

newton/_src/viewer/gl/shaders.py
ooctipus: Applied all three cleanups in 61af2067: removed dead light_space_matrix state from ShaderEdge, use length(camera_to_fragment) directly for fog, and document at the reconstruction call that it also corrects fragment depth.

newton/_src/viewer/viewer.py
```diff
@@ -1632,6 +1632,10 @@ def log_geo(
                 uvs,
                 hidden=hidden,
                 texture=texture,
+                texture_scale=geo_src.texture_scale,
+                texture_translate=geo_src.texture_translate,
+                texture_rotate=geo_src.texture_rotate,
+                texture_projection=geo_src.texture_projection,
```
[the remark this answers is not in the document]
ooctipus: This thread is obsolete after the texture-projection work was split into PR #4007. PR #4006 no longer changes this file.

newton/tests/test_mesh_cache.py
```diff
@@ -42,6 +42,24 @@ def test_invalidate_cache_refreshes_hash(self):
         mesh.invalidate_cache()
         self.assertNotEqual(hash(mesh), old_hash)
 
+    def test_texture_mapping_is_copied_and_hashed(self):
+        """Texture-coordinate semantics survive copies and distinguish cached prototypes."""
+        mesh = _make_tet_mesh()
```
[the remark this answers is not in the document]
ooctipus: This thread is obsolete after the texture-projection work was split into PR #4007. PR #4006 no longer changes this file.

eric-heiden (approved):
Thanks!

PR #4008, opened 2026-08-23
Refit the MuJoCo midphase BVH after syncing inertial frames on CPU

### LINKED ISSUE #3805: [BUG] `SolverMuJoCo(use_mujoco_cpu=True)`: post-compile `body_iquat` sync leaves the MuJoCo body BVH in the stale compile-time inertial frame, silently dropping valid contacts
## Summary

`_sync_mjw_inertias_to_mjc_cpu()` overwrites `mj_model.body_iquat` with Newton's inertia principal axes after `spec.compile()`. MuJoCo stores each body's midphase BVH boxes in the **compile-time inertial frame** (`mj_collideTree` transforms them by runtime `ximat = xmat · iquat`), so after the sync the BVH is rotated by the wrong frame and the CPU backend's `mj_collision` culls contact pairs that genuinely overlap — while `mj_geomDistance` (no midphase) still reports penetration for the same pair.

The sync runs during solver construction via `notify_model_changed(ModelFlags.ALL)`, so every `use_mujoco_cpu=True` model whose Newton-computed inertial frame differs from MuJoCo's compiled one is affected from step 0. The error grows with how far the two principal-axis frames diverge: near-symmetric single-hull bodies lose little, multi-shape bodies (e.g. CoACD convex decompositions) can lose **all** contacts.

Introduced in #3018 (fix for #3017, first released in v1.3.0). #3018 correctly identified that `body_inertia`/`body_iquat` are coupled to the compiled simple-path metadata and invalidates `body_sameframe`/`body_simple` — but the body BVH is expressed in that same frame and is not touched. **Reproduced empirically on v1.3.0 (mujoco 3.8.1), v1.4.0 (mujoco 3.11.0), and `main` @ b05d7f27b1 / 1.5.0.dev0 (mujoco 3.11.0)** — both at init and after a runtime `notify_model_changed(BODY_INERTIAL_PROPERTIES)` inertia update.

Note for reproducing on ≥ v1.4.0: the builder's default `rigid_gap = 0.1` is now exported as a 10 cm geom `margin`, which inflates the midphase boxes and masks the culling for shallow overlaps — the repro below sets `builder.rigid_gap = 0.0`. Real scenes with realistic margins (mm-scale) remain affected, since the BVH frame error is much larger than the margin.

The MJWarp backend is unaffected (its broadphase works from runtime-transformed geom AABBs and never reads `mjModel.bvh_*`), which makes this an easy-to-miss CPU/GPU behavioral divergence.

## Suggested fix

Any of:

1. When the sync changes `body_iquat[b]`, re-express body `b`'s BVH nodes in the new inertial frame (rotate the stored `bvh_aabb` boxes by `iquat_old⁻¹ · iquat_new` with conservative re-axis-alignment), analogous to the existing `body_sameframe`/`body_simple` invalidation.
2. Cheaper and always safe: set `mj_model.opt.disableflags |= mjDSBL_MIDPHASE` whenever the sync modifies any `body_iquat` (with a warning), trading midphase culling for correctness.
3. Avoid the frame change where possible: keep the compiled `body_iquat` when it still diagonalizes Newton's tensor and only permute the diagonal moments (the direction #3008 originally explored), falling back to 1/2 otherwise.

## Workaround for users

```python
solver.mj_model.opt.disableflags |= mujoco.mjtDisableBit.mjDSBL_MIDPHASE
```

after constructing `SolverMuJoCo(use_mujoco_cpu=True, ...)`. Verified to restore contact sets identical to the MJWarp backend in the scene below.

Report written by Fable.

## Repro 1 — through Newton's public API (~45 lines)

A static 2 cm cube and a free body (8×2×1 cm slab + off-center blob, so the composite inertia frame is rotated), overlapping by 1 mm:

```python
import mujoco
import numpy as np
import warp as wp

import newton

CUBE_FACES = np.array(
    [
        [0, 1, 3], [0, 3, 2],
        [4, 7, 5], [4, 6, 7],
        [0, 4, 1], [1, 4, 5],
        [2, 3, 7], [2, 7, 6],
        [0, 2, 6], [0, 6, 4],
        [1, 5, 7], [1, 7, 3],
    ],
    dtype=np.int32,
)


def box(hx, hy, hz):
    v = np.array(
        [[sx * hx, sy * hy, sz * hz] for sx in (-1, 1) for sy in (-1, 1) for sz in (-1, 1)],
        dtype=np.float32,
    )
    return newton.Mesh(v.flatten().tolist(), CUBE_FACES.flatten().tolist())


builder.rigid_gap = 0.0  # >= v1.4.0 exports the 0.1 default as geom margin, masking the bug
builder.add_shape_mesh(body=-1, mesh=box(0.01, 0.01, 0.01))          # static cube
obj = builder.add_body(xform=wp.transform(wp.vec3(0.049, 0.0, 0.0), wp.quat_identity()))
builder.add_joint_free(obj)
builder.add_shape_mesh(body=obj, mesh=box(0.04, 0.01, 0.005))        # slab
builder.add_shape_mesh(                                              # off-center blob
    body=obj, mesh=box(0.005, 0.005, 0.005),
    xform=wp.transform(wp.vec3(0.02, 0.02, 0.0), wp.quat_identity()),
)

solver = newton.solvers.SolverMuJoCo(model, use_mujoco_cpu=True, use_mujoco_contacts=True)

m, d = solver.mj_model, solver.mj_data
print("body_iquat after solver init:", np.round(m.body_iquat[1], 4))

fromto = np.zeros(6)
for label, disable in [("midphase ON ", False), ("midphase OFF", True)]:
    m.opt.disableflags &= ~int(mujoco.mjtDisableBit.mjDSBL_MIDPHASE)
    if disable:
        m.opt.disableflags |= int(mujoco.mjtDisableBit.mjDSBL_MIDPHASE)
    mujoco.mj_forward(m, d)
    dist = mujoco.mj_geomDistance(m, d, 0, 1, 10.0, fromto)
    print(f"{label}: ncon={d.ncon}  mj_geomDistance(cube, slab)={dist * 1000:7.3f} mm")
```

Output (identical on newton 1.3.0 + mujoco 3.8.1, newton 1.4.0 + mujoco 3.11.0, and newton main/1.5.0.dev0 + mujoco 3.11.0):

```
body_iquat after solver init: [ 0.7069  0.0166 -0.7069  0.0166]
midphase ON : ncon=0  mj_geomDistance(cube, slab)= -1.000 mm
midphase OFF: ncon=1  mj_geomDistance(cube, slab)= -1.000 mm
```

The pair penetrates 1 mm, the narrowphase can prove it, and `mj_collision` drops it unless the midphase is disabled. The same failure occurs after a runtime inertia update: rotating the body's inertia tensor and calling `solver.notify_model_changed(newton.ModelFlags.BODY_INERTIAL_PROPERTIES)` re-syncs `body_iquat` and the midphase culls again (so domain-randomization-style inertia updates re-trigger the bug even if init-time frames happened to coincide).

## Repro 2 — pure MuJoCo, isolating the invariant

No Newton involved: compile the same scene from MJCF, then perform exactly the mutation the sync does. Shows MuJoCo behaves correctly on a self-consistent model and that overwriting `body_iquat` post-compile is what breaks it:

```python
import mujoco
import numpy as np


def box_mesh(hx, hy, hz):
    v = []
    for sx in (-1, 1):
        for sy in (-1, 1):
            for sz in (-1, 1):
                v += [sx * hx, sy * hy, sz * hz]
    return " ".join(f"{x:.6f}" for x in v)


XML = f"""
<mujoco>
  <asset>
    <mesh name="finger" vertex="{box_mesh(0.01, 0.01, 0.01)}"/>
    <mesh name="slab"   vertex="{box_mesh(0.04, 0.01, 0.005)}"/>
    <mesh name="blob"   vertex="{box_mesh(0.005, 0.005, 0.005)}"/>
  </asset>
  <worldbody>
    <geom name="finger" type="mesh" mesh="finger"/>
    <body name="obj" pos="0.049 0 0">
      <freejoint/>
      <geom name="slab" type="mesh" mesh="slab"/>
      <geom name="blob" type="mesh" mesh="blob" pos="0.02 0.02 0"/>
    </body>
  </worldbody>
</mujoco>
"""

m = mujoco.MjModel.from_xml_string(XML)
d = mujoco.MjData(m)


def report():
    for label, disable in [("midphase ON ", False), ("midphase OFF", True)]:
        m.opt.disableflags &= ~int(mujoco.mjtDisableBit.mjDSBL_MIDPHASE)
        if disable:
            m.opt.disableflags |= int(mujoco.mjtDisableBit.mjDSBL_MIDPHASE)
        mujoco.mj_forward(m, d)
        fromto = np.zeros(6)
        dist = mujoco.mj_geomDistance(m, d, 0, 1, 10.0, fromto)
        print(f"  {label}: ncon={d.ncon}  dist={dist * 1000:7.3f} mm")


print("1) freshly compiled model:")
report()

# what _sync_mjw_inertias_to_mjc_cpu does, minus the BVH it forgets:
b = mujoco.mj_name2id(m, mujoco.mjtObj.mjOBJ_BODY, "obj")
m.body_iquat[b] = [np.cos(np.pi / 4), 0.0, 0.0, np.sin(np.pi / 4)]
m.body_sameframe[b] = int(mujoco.mjtSameFrame.mjSAMEFRAME_NONE)
m.body_simple[b] = 0
m.dof_simplenum[:] = 0
mujoco.mj_setConst(m, d)

print("2) after overwriting body_iquat post-compile:")
report()
```

```
1) freshly compiled model:
  midphase ON : ncon=4  dist= -1.000 mm
  midphase OFF: ncon=4  dist= -1.000 mm
2) after overwriting body_iquat post-compile:
  midphase ON : ncon=0  dist= -1.000 mm
  midphase OFF: ncon=4  dist= -1.000 mm
```

### System Information

- Reproduced on: newton v1.3.0 + mujoco 3.8.1; newton v1.4.0 + mujoco 3.11.0; newton `main` @ b05d7f27b1 (1.5.0.dev0) + mujoco 3.11.0
- macOS arm64 / CPU; mechanism is platform-independent

catplotlib:
## Description

MuJoCo stores each body's midphase BVH boxes in the **compile-time** inertial frame, and `mj_collideTree` transforms them at runtime by `ximat = xmat @ iquat`. `_sync_mjw_inertias_to_mjc_cpu()` overwrites `body_iquat` after `spec.compile()`, so those boxes were left describing a frame that no longer exists and the midphase culled geom pairs that genuinely overlap — while `mj_geomDistance` (no midphase) still reported penetration for the same pair.

This re-expresses each rotated body's BVH boxes in its new inertial frame using `R = R_new.T @ R_old`. A rotated box is not axis-aligned, so the half-extents are grown into the enclosing box via `|R|`; that is conservative and can never cull a real overlap.

The sync runs during solver construction via `notify_model_changed(ModelFlags.ALL)`, so every `use_mujoco_cpu=True` model whose Newton-computed inertial frame differs from the compiled one was affected from step 0. Near-symmetric single-hull bodies lose little; multi-shape bodies (for example CoACD convex decompositions) can lose every contact.

Closes #3805

## Checklist

## Test plan

New regression test, which fails on `main` and passes with this change:

```
uv run --extra dev -m newton.tests -k test_mujoco_cpu_midphase_keeps_contacts_after_inertia_frame_sync
```

It builds a body carrying two shapes far apart in the body frame, so its principal axes diverge from the compiled ones, and asserts the midphase reports the same contact count as the narrowphase alone. Without the fix it fails with `AssertionError: 0 != 1`.

Full solver suite on this branch:

```
python -m unittest newton.tests.test_mujoco_solver
```

274 tests, 19 skipped, 1 error — the error is `test_mjc_damping_from_usd_via_schema_resolver` failing with `ModuleNotFoundError: No module named 'pxr'`, which fails identically on pristine `main` here (no USD installed). `ruff check` and `ruff format` are clean.

Environment: Linux x86_64, Python 3.12, CPU device, mujoco 3.11.0 / mujoco-warp 3.11.0 / warp-lang 1.16.0.

## Bug fix

To reproduce **without** this PR applied:

```python
import mujoco
import numpy as np
import warp as wp

import newton

CUBE_FACES = [
    0, 1, 3, 0, 3, 2, 4, 7, 5, 4, 6, 7, 0, 4, 1, 1, 4, 5,
    2, 3, 7, 2, 7, 6, 0, 2, 6, 0, 6, 4, 1, 5, 7, 1, 7, 3,
]

def box(hx, hy, hz):
    v = np.array([[sx * hx, sy * hy, sz * hz] for sx in (-1, 1) for sy in (-1, 1) for sz in (-1, 1)], dtype=np.float32)
    return newton.Mesh(v.flatten().tolist(), CUBE_FACES)

builder.rigid_gap = 0.0  # a non-zero gap is exported as a geom margin and masks the miss
builder.add_shape_mesh(body=-1, mesh=box(0.01, 0.01, 0.01))
obj = builder.add_body(xform=wp.transform(wp.vec3(0.049, 0.0, 0.0), wp.quat_identity()))
builder.add_joint_free(obj)
builder.add_shape_mesh(body=obj, mesh=box(0.04, 0.01, 0.005))
builder.add_shape_mesh(body=obj, mesh=box(0.005, 0.005, 0.005),
                       xform=wp.transform(wp.vec3(0.02, 0.02, 0.0), wp.quat_identity()))

solver = newton.solvers.SolverMuJoCo(model, use_mujoco_cpu=True, use_mujoco_contacts=True)
m, d = solver.mj_model, solver.mj_data

for label, disable in (("midphase ON ", False), ("midphase OFF", True)):
    m.opt.disableflags &= ~int(mujoco.mjtDisableBit.mjDSBL_MIDPHASE)
    if disable:
        m.opt.disableflags |= int(mujoco.mjtDisableBit.mjDSBL_MIDPHASE)
    mujoco.mj_forward(m, d)
    dist = mujoco.mj_geomDistance(m, d, 0, 1, 10.0, np.zeros(6))
    print(f"{label}: ncon={d.ncon}  mj_geomDistance={dist * 1000:7.3f} mm")
```

On `main`:

```
midphase ON : ncon=0  mj_geomDistance= -1.000 mm
midphase OFF: ncon=1  mj_geomDistance= -1.000 mm
```

With this change both report `ncon=1`. Credit to @lenroe, whose report isolated the invariant precisely.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed CPU collision detection after inertial frame updates, preventing valid overlapping geometry from being incorrectly culled.
  * Improved collision-bound updates when body positions, center-of-mass locations, or orientations change.
  * Prevented collision bounds from drifting during repeated inertial-frame updates or switches between configurations.

* **Tests**
  * Added regression coverage for penetrating geometry and collision bounds following runtime center-of-mass and orientation updates.

### DIFF
--- a/changelog/3805.fixed.md
+++ b/changelog/3805.fixed.md
@@ -1,0 +1,1 @@
+Re-express a body's MuJoCo midphase BVH in its new inertial frame when `SolverMuJoCo(use_mujoco_cpu=True)` syncs inertias after compilation, so the midphase no longer culls geom pairs that genuinely overlap.
--- a/newton/_src/solvers/mujoco/solver_mujoco.py
+++ b/newton/_src/solvers/mujoco/solver_mujoco.py
@@ -4379,6 +4379,8 @@ class SolverMuJoCo(SolverBase, CouplingInterface):
         if not np.any(changed_bodies):
             return
 
+        old_body_iquat = self.mj_model.body_iquat.copy()
+
         self.mj_model.body_inertia[:] = mjw_body_inertia
         self.mj_model.body_iquat[:] = mjw_body_iquat
 
@@ -4388,6 +4390,40 @@ class SolverMuJoCo(SolverBase, CouplingInterface):
         self.mj_model.body_sameframe[changed_bodies] = int(self._mujoco.mjtSameFrame.mjSAMEFRAME_NONE)
         self.mj_model.body_simple[changed_bodies] = 0
         self.mj_model.dof_simplenum[:] = 0
+
+        self._refit_body_bvh_to_inertial_frame(old_body_iquat, mjw_body_iquat, iquat_changed)
+
+    def _refit_body_bvh_to_inertial_frame(
+        self, old_body_iquat: np.ndarray, new_body_iquat: np.ndarray, iquat_changed: np.ndarray
+    ) -> None:
+        """Re-express each rotated body's midphase BVH boxes in its new inertial frame.
+
+        MuJoCo stores the body midphase BVH in the compile-time inertial frame and
+        ``mj_collideTree`` transforms it by ``ximat = xmat @ iquat`` at runtime. Overwriting
+        ``body_iquat`` after ``spec.compile()`` therefore leaves those boxes expressed in a frame
+        that no longer exists, and the midphase culls geom pairs that genuinely overlap.
+
+        Args:
+            old_body_iquat: Inertial frame orientations before the sync, shape ``(nbody, 4)``.
+            new_body_iquat: Inertial frame orientations after the sync, shape ``(nbody, 4)``.
+            iquat_changed: Boolean mask selecting the bodies whose orientation changed.
+        """
+        rot_old = np.empty(9)
+        rot_new = np.empty(9)
+        for body in np.flatnonzero(iquat_changed):
+            bvh_num = int(self.mj_model.body_bvhnum[body])
+            if bvh_num == 0:
+                continue
+            self._mujoco.mju_quat2Mat(rot_old, old_body_iquat[body])
+            self._mujoco.mju_quat2Mat(rot_new, new_body_iquat[body])
+            # Maps a point from the old inertial frame to the new one.
+            rot = rot_new.reshape(3, 3).T @ rot_old.reshape(3, 3)
+            bvh_adr = int(self.mj_model.body_bvhadr[body])
+            aabb = self.mj_model.bvh_aabb[bvh_adr : bvh_adr + bvh_num]
+            # A rotated box is not axis-aligned, so its half-extents grow into the enclosing box.
+            self.mj_model.bvh_aabb[bvh_adr : bvh_adr + bvh_num] = np.hstack(
+                (aabb[:, :3] @ rot.T, aabb[:, 3:] @ np.abs(rot).T)
+            )
 
     @override
     def notify_model_changed(self, flags: ModelFlags | int) -> None:
--- a/newton/_src/solvers/mujoco/solver_mujoco.py
+++ b/newton/_src/solvers/mujoco/solver_mujoco.py
@@ -3664,6 +3664,11 @@ class SolverMuJoCo(SolverBase, CouplingInterface):
         if disable_sensors:
             disableflags |= mujoco.mjtDisableBit.mjDSBL_SENSOR
         self.use_mujoco_cpu = use_mujoco_cpu
+        # Compiled midphase boxes and the inertial frame they are expressed in, captured
+        # before the first inertial sync so every refit starts from the same bounds.
+        self._canonical_bvh_aabb: np.ndarray | None = None
+        self._canonical_body_ipos: np.ndarray | None = None
+        self._canonical_body_iquat: np.ndarray | None = None
         if use_mujoco_contacts or use_mujoco_cpu:
             mujoco_attrs_for_warn = getattr(model, "mujoco", None)
             solref_mode_attr = (
@@ -4361,6 +4366,7 @@ class SolverMuJoCo(SolverBase, CouplingInterface):
         """Synchronize the complete MJWarp inertial representation to MuJoCo CPU."""
         mjw_body_inertia = self.mjw_model.body_inertia.numpy()[0]
         mjw_body_iquat = self.mjw_model.body_iquat.numpy()[0]
+        mjw_body_ipos = self.mjw_model.body_ipos.numpy()[0]
 
         inertia_changed = ~np.isclose(
             self.mj_model.body_inertia,
@@ -4374,15 +4380,25 @@ class SolverMuJoCo(SolverBase, CouplingInterface):
             rtol=1.0e-6,
             atol=1.0e-8,
         ).all(axis=1)
-        changed_bodies = inertia_changed | iquat_changed
+        ipos_changed = ~np.isclose(
+            self.mj_model.body_ipos,
+            mjw_body_ipos,
+            rtol=1.0e-6,
+            atol=1.0e-8,
+        ).all(axis=1)
+        # The inertial frame is the pair (ipos, iquat); either one moving invalidates the
+        # compiled metadata and the midphase boxes expressed in that frame.
+        frame_changed = iquat_changed | ipos_changed
+        changed_bodies = inertia_changed | frame_changed
 
         if not np.any(changed_bodies):
             return
 
-        old_body_iquat = self.mj_model.body_iquat.copy()
+        self._cache_canonical_body_bvh()
 
         self.mj_model.body_inertia[:] = mjw_body_inertia
         self.mj_model.body_iquat[:] = mjw_body_iquat
+        self.mj_model.body_ipos[:] = mjw_body_ipos
 
         # ``body_inertia`` and ``body_iquat`` are coupled. Once the inertial
         # frame changes, MuJoCo CPU's compiled simple-path metadata may still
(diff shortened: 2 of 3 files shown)


vreutskyy (changes requested):
I found two issues that should be addressed before merging:

1. The BVH refit transforms only the inertial-frame orientation, but body_ipos is synchronized as part of the same update. body_ipos is overwritten before _sync_mjw_inertias_to_mjc_cpu(), while the helper receives only quaternions and visits only iquat_changed bodies. The center transform is therefore missing R_new.T @ (p_old - p_new). This can still cull real contacts when the COM changes, including supported runtime body_com updates. In an isolated MuJoCo reproduction applying this PR's orientation-only refit, midphase found 0 contacts while narrowphase found 4; adding the translation restored 4/4. Please snapshot the old body_ipos, include position changes in the affected-body mask and metadata invalidation, apply the full affine inertial-frame transform, and cover this case with a regression test.

2. Each update refits the already-expanded bvh_aabb. The abs(R) enclosure is not reversible, so repeated runtime inertia-frame updates can grow the bounds without limit. For example, extents [1, 0.1, 0.1] subjected to repeated 45-degree round trips become [1.1, 1.1, 0.1], then [2.2, 2.2, 0.1], then [4.4, 4.4, 0.1]. On multi-shape bodies this can make the midphase ineffective. Please recompute from cached canonical bounds or otherwise rebuild tight bounds rather than transforming the previous conservative enclosure.

catplotlib:
Thanks @vreutskyy — both were real, and I could reproduce both. Pushed 6d60c55.

**1. `body_ipos`.** You're right that it is written before the sync runs, and that the helper never saw it. The caller wrote `body_ipos` three lines above the call, so `ipos_changed` could not have been true even if I had computed it. The sync now owns that write, the mask keys on either component of the frame moving, and the transform is the full affine one: a point `p` in the compiled frame sits at `R_new.T @ (R_c @ p + ipos_c - ipos_new)`.

I confirmed the failure through a runtime `body_com` update, which moves `ipos` while leaving `iquat` untouched — the case you described. On the previous commit the inertial origin moved 0.02000 and the box centres moved 0.00000; now they move 0.02092, the difference being the rotation into the new frame.

**2. Unbounded growth.** Also correct, and reproducible in isolation — half-extents `[1, 0.1, 0.1]` under repeated 45-degree round trips go `[1.1, 1.1, 0.1]`, `[2.2, 2.2, 0.1]`, `[4.4, 4.4, 0.1]`, exactly as you said. The compiled boxes and the frame they were compiled in are now snapshotted before the first sync, and every refit is computed from that snapshot, so the operation is idempotent rather than accumulating.

**Regression test.** `test_mujoco_cpu_midphase_survives_runtime_com_update` covers the centre-of-mass case: it asserts the boxes track the origin, and that repeating the update leaves the extents unchanged. It fails on the previous commit with `AssertionError: 0.0 not greater than 0.0001`.

Full `test_mujoco_solver` suite is 275 tests, 19 skipped, with one pre-existing error in `test_mjc_damping_from_usd_via_schema_resolver` (`No module named 'pxr'` in my environment, fails the same way on a clean checkout). `ruff check` and `ruff format` are clean.

One note on scope: moving the `body_ipos` write into `_sync_mjw_inertias_to_mjc_cpu` means that method now owns the whole inertial-frame sync rather than just inertia and orientation. That seemed truer to its name and is what lets the mask see the origin, but say the word if you would rather it stayed in the caller and the old values were passed in instead.

catplotlib:
Good catch, fixed in 90a647e.

You were right that the assertion was vacuous: repeating an identical update returns early from the sync before any refit happens, and a centre-of-mass translation cannot change half-extents in either direction. It passed regardless of the implementation.

The check now rotates the inertia tensor to move the body to a second inertial frame, returns to the first, and asserts the bounds match the ones recorded for that frame.

Both halves are verified against implementations that should fail them:

- against the first commit, the centre assertion fails with `0.0 not greater than 0.0001`
- against a variant that keeps the `body_ipos` fix but refits from the current bounds instead of the compiled ones, the round-trip assertion fails with `Not equal to tolerance rtol=1e-07, atol=1e-06`

vreutskyy (approved):
Re-reviewed at ce951717. The inertial-origin transform and canonical-BVH refit address the requested changes, and I found no remaining blocking issues.

catplotlib:
Hey, this has been approved and green since Aug 31, but the external-PR approval check is still pending so it can't merge. It's 82 commits behind main now, happy to rebase if that helps.

PR #3857, opened 2026-08-10
Standardize cable and rod API terminology

jumyungc:
## Description
Standardize cable and rod terminology so “cable” describes the modeled object and “rod” describes its joint/element representation.

- Add JointType.ROD and ModelBuilder.add_joint_rod() as the preferred joint APIs.
- Add prefix-first cable geometry/frame helpers:
  - cable_straight_points()
  - cable_parallel_transport_quaternions()
  - cable_straight_points_and_quaternions()
  - Add RodStiffness and rod_stiffness_from_elastic_moduli() for per-joint rod stiffness.
- Deprecate JointType.CABLE, add_joint_cable(), CableStiffness, and the existing create_* cable helpers while preserving compatibility through Newton 1.6.
- Update rod-only solver helpers/state, examples, tests, and documentation to use rod terminology.
- Clarify that rod joints allocate four VBD material slots—stretch, shear, bend, and twist—rather than conventional generalized-coordinate DOFs.
The enum value remains `7`, existing positional signatures remain compatible, and generated cable labels remain unchanged.

### Terminology boundary
- Cable refers to the physical object and representation-independent geometry/frame concepts.
- Rod refers to the discrete joint/element representation, including per-joint stiffness, material slots, and VBD mechanics.

## Checklist

- [x] New or existing tests cover these changes
- [x] The documentation is up to date with these changes
- [x] For user-facing changes, a fragment has been added by following the
      [changelog fragment instructions](https://github.com/newton-physics/newton/blob/main/changelog/README.md)

## New feature / API change

```python
import newton

# Code that demonstrates the new capability
```

## Summary by CodeRabbit

* **New Features**
  * Added preferred rod-joint APIs and `ROD` naming for cable-like simulations.
  * Added streamlined cable geometry and stiffness utility names.
  * Rod joints support stretch, shear, bend, and twist behavior.

* **Deprecations**
  * Previous cable-joint and utility names remain available through version 1.6 with warnings.
  * Existing names may be removed in version 1.7; migration guidance is included.

* **Documentation**
  * Updated solver support, articulation, USD import, and API documentation to use rod terminology.

PR #4009, opened 2026-08-23
Add newton.Rod and deprecate cable helper APIs

jumyungc:
## Description

This builds on the cable-to-rod joint terminology introduced in #3857 by adding a canonical public representation for prepared discrete rods.

- Add `newton.Rod` as the public input type for centerline geometry, segment topology, material frames, and optional constitutive data.
- Support ordered chains and explicit graph topology without introducing a `newton.rod` submodule.
- Make `ModelBuilder.add_rod(rod=...)` the preferred construction path.
- Allow a `Rod` to specify either isotropic elastic material properties or a complete set of stretch, shear, bend, and twist section rigidities.
- Convert section rigidities into per-joint stiffness using each joint’s local dual rest length. Branched graphs require explicit builder stiffness because their junctions do not define a unique material law.
- Keep direct per-joint stiffness and damping arguments as assembly-level overrides.
- Update the documentation, examples, USD cable importer, and focused tests to use the new API.

This preserves the terminology distinction discussed in #3857: a cable is the modeled object, while `Rod` represents Newton’s discrete body-and-joint formulation of it.

## Deprecations

Deprecate the following APIs in favor of `newton.Rod` and `ModelBuilder.add_rod(rod=...)`:

- `ModelBuilder.add_rod(positions=...)`
- `ModelBuilder.add_rod_graph()`
- `newton.utils.CableStiffness`
- `newton.utils.create_cable_stiffness_from_elastic_moduli()`
- `newton.utils.create_parallel_transport_cable_quaternions()`
- `newton.utils.create_straight_cable_points()`
- `newton.utils.create_straight_cable_points_and_quaternions()`

These APIs remain functional through 1.6 and are eligible for removal in 1.7. The deprecated stiffness helper preserves its released conditional return shape.

Temporary rod-prefixed utility spellings introduced only during 1.6 development were not part of a stable release and do not require an additional deprecation cycle.

## Checklist


newton/utils.py
```diff
@@ -89,29 +89,33 @@
 __all__ += [
     "CableStiffness",
     "RodStiffness",
-    "cable_straight_points",
+    "cable_generate_straight_points",
```
jcarius-nv: **P1 — Standards (proposed policy):** PR #3922's source-code guide says related API families belong in a purposeful domain submodule and explicitly says not to use a general-purpose module such as `utils` as a substitute for a clear domain boundary. These preferred exports make a cohesive cable/rod simulation family permanent `newton.utils` API. The resulting public contract weakens ownership and discoverability and makes future rod/cable expansion further enlarge the catch-all namespace. Because #3922 remains open, this is a proposed-policy finding rather than a violation of merged Newton policy.

newton/_src/utils/cable.py
```diff
@@ -157,64 +119,64 @@ def rod_stiffness_from_elastic_moduli(
...
-        stretch=stretch_stiffness,
-        bend=bend_stiffness,
-        twist=G * polar_inertia / L,
-    )
+    return RodStiffness(stretch=stretch, bend=bend, twist=shear * polar_inertia / L)
 
 
-# Cable-level geometry and frames
-def cable_straight_points(
+def cable_generate_straight_points(
```
jcarius-nv: **P1 — Requirements and Standards:** The PR description and these docstrings say the generated values may be passed to `ModelBuilder.add_rod()`, but the accepted bounds do not satisfy that consumer contract. `num_segments=1` returns two points and one quaternion while `add_rod()` rejects fewer than two segments; `length=0` returns duplicate points that `add_rod_graph()` rejects; and `twist_total=NaN` returns NaN quaternions despite the combined helper promising unit orientations. The focused test exercises the one-segment helper result without passing it to the documented consumer. Callers can therefore successfully generate advertised construction outputs that fail only downstream or silently contain invalid orientations. This does not repeat CodeRabbit's existing non-finite direction/segment-vector concern.

newton/_src/utils/cable.py
```diff
@@ -363,23 +338,62 @@ def create_cable_stiffness_from_elastic_moduli(
...
+    if poissons_ratio is None:
+        if shear_modulus is not None:
+            return rod_compute_stiffness_from_elastic_moduli(
+                youngs_modulus,
+                radius,
+                segment_length,
+                shear_modulus=shear_modulus,
+            )
+
+        E = float(youngs_modulus)
```
jcarius-nv: **P2 — Standards judgment call:** This deprecated no-modulus branch duplicates the canonical factory's finite/range validation and stretch/bend calculations. During the required 1.6 compatibility window, numerical-validation and error-contract changes must remain synchronized across both paths. Current compatibility coverage exercises the valid legacy return but not this duplicated invalid-input branch; the maintenance cost is bounded by the deprecation period.

jcarius-nv (commented):
Reviewed exact head `04382ec265fc887e2eb0d977e693243a9f7e0a8f` using the proposed Fit, Requirements, and Standards rubric from PR #3922 at `d8df5fa89d444647710c494118e0febb59a273e4`.

## Fit

**Verdict: Fits.** Newton owns these construction helpers, the surface is bounded, existing workloads exercise it, and no new serialized or USD-authored concept is introduced.

## Requirements

One P1 finding: the construction helpers accept inputs that do not satisfy their documented `ModelBuilder.add_rod()` consumer contract. Details are inline.

## Standards

Two P1 findings and one P2 judgment call: the preferred API family is assigned to the general-purpose `newton.utils` namespace despite the proposed source-code guideline, the helper-family contracts are inconsistent at documented boundaries, and the deprecated stiffness path duplicates canonical validation/formulas. Details are inline.

The already-raised CodeRabbit concern about non-finite direction/segment vectors is acknowledged and not repeated. Its reported stale `newton.rod` test call no longer reproduces at this head. Focused helper tests passed 5/5; live CI has 30 passing and 2 skipped checks, with no pending or failing checks. GitHub still reports `reviewDecision=REVIEW_REQUIRED`, `mergeable=MERGEABLE`, and `mergeStateStatus=BLOCKED`.

jcarius-nv:
@jumyungc I got a bit confused, I think you changed the PR content quite a bit again with the latest commit. Can you explain so I can re-review it properly?

jumyungc:
> @jumyungc I got a bit confused, I think you changed the PR content quite a bit again with the latest commit. Can you explain so I can re-review it properly?

Sorry for the confusion - I did not realize that you had already started reviewing the PR. Yes in the last commit, it changed from using rod submodule back to using utils. Now there is a separate proposal to use a public newton.Rod class, and I tagged you in that discussion. I will mark this PR as a draft until we settle the design, so you do not need to re-review the current revision yet. Thanks!

jumyungc:
The new Rod class-based approach is now implemented, and the PR is ready for review.

newton/_src/utils/cable.py
```diff
@@ -10,323 +12,46 @@
...
-    Returned by :func:`rod_stiffness_from_elastic_moduli` when the caller
-    supplies either ``poissons_ratio`` or ``shear_modulus``.
+from ..sim.rod import _compute_parallel_transport_quaternions
 
-    Fields:
 
-    * ``stretch`` -- axial stiffness ``E * A / L`` [N/m]
-    * ``bend``    -- bending stiffness ``E * I / L`` [N·m/rad]
-    * ``twist``   -- torsional stiffness ``G * J / L`` [N·m/rad]
+class CableStiffness(NamedTuple):
```
eric-heiden, line 18: Isn't this a breaking change if you rename `RodStiffness` to `CableStiffness` here?

newton/_src/sim/rod.py
```diff
@@ -0,0 +1,582 @@
...
+            exclusive with elastic material inputs.
+        twist_rigidity: Optional torsional rigidity ``GJ`` [N·m²]. Mutually
+            exclusive with elastic material inputs.
+
+    Raises:
+        ValueError: If the geometry, frames, or constitutive inputs are invalid
+            or mutually inconsistent.
+
+    Note:
+        Material-derived transverse shear follows the AOUSD deformable-body
```
eric-heiden: Nit: I think this should be called "OpenUSD" rather than "AOUSD" because OpenUSD is the standard, AOUSD is just the alliance behind it.
Would be good to update in other places here in this PR where AOUSD is mentioned.

newton/_src/utils/cable.py
jumyungc: I think this is not a released breaking change because Newton 1.5 exposes CableStiffness, which remains available in 1.6 as a deprecated compatibility type. RodStiffness was introduced only during 1.6 development as a better-named replacement. With the new Rod-based API in the PR, I don’t think we need a separate preferred stiffness class, so the temporary RodStiffness type has been removed.

newton/_src/sim/rod.py
jumyungc: Agreed, thanks!

newton/_src/sim/builder.py
```diff
@@ -8618,6 +8754,254 @@ def add_rod(
...
+        wrap_in_articulation: bool = True,
+        color: Vec3 | None = None,
+        body_frame_origin: Literal["start", "com"] | None = None,
+        rod: None = None,
+        junction_collision_filter: bool = True,
+    ) -> tuple[list[int], list[int]]: ...
+
+    def add_rod(
+        self,
+        positions: list[Vec3] | None = None,
```
eric-heiden, line 8768: Nothing here in the docstring says this version of the `add_rod()` method is deprecated and you are also not raising a DeprecationWarning if a user passes positions as input here. I think we should do that, right? Going forward, users should use `newton.Rod` and pass that to `add_rod()`.

I also don't understand why we need all these `@overload` declarations of this method now - is this just for type hinting? Again, since we are deprecating the old version that takes positions as input, I don't think we need those.

newton/_src/sim/rod.py
```diff
@@ -0,0 +1,582 @@
...
+        self.stretch_rigidity = (
+            None if stretch_rigidity is None else _validate_nonnegative_float("stretch_rigidity", stretch_rigidity)
+        )
+        self.shear_rigidity = (
+            None if shear_rigidity is None else _validate_nonnegative_float("shear_rigidity", shear_rigidity)
+        )
+        self.bend_rigidity = (
+            None if bend_rigidity is None else _validate_nonnegative_float("bend_rigidity", bend_rigidity)
+        )
+        self.twist_rigidity = (
```
eric-heiden, line 227: Can you please add inline docstrings for these user-writable/accessible attributes via `"""..."""` here like we have in `newton.Model` so that we see these attributes in the generated docs for `newton.Rod`?
Right now, only the points, edges, and quaternions properties have docstrings.

newton/_src/sim/builder.py
jumyungc: I was initially unsure whether we wanted to deprecate the existing add_rod(positions=...) form and add_rod_graph() as part of this change. They are now explicitly documented as deprecated and emit a DeprecationWarning, while add_rod(rod=...) remains the preferred warning-free path.

newton/_src/sim/rod.py
jumyungc: Done.

newton/utils.py
jumyungc: Addressed: the preferred API now lives on newton.Rod instead of in newton.utils.

newton/_src/utils/cable.py
jumyungc: Addressed.

newton/_src/utils/cable.py
jumyungc: No longer applicable in the current revision.

newton/_src/sim/builder.py
```diff
@@ -5984,6 +5993,94 @@ def _set_joint_rod_material_gains(
...
+    ) -> None:
+        """Validate direct per-joint rod stiffness inputs."""
+        stiffnesses = (
+            ("stretch_stiffness", stretch_stiffness),
+            ("shear_stiffness", shear_stiffness),
+            ("bend_stiffness", bend_stiffness),
+            ("twist_stiffness", twist_stiffness),
+        )
+        for name, stiffness in stiffnesses:
+            if stiffness is not None and stiffness < 0.0:
```
jcarius-nv: **P1 — Requirements: Non-finite stiffness overrides create invalid models.**

This validation only rejects values `< 0.0`, so `NaN` and positive infinity pass for every direct stiffness override. I reproduced `stretch_stiffness=np.nan` yielding joint gains `[nan, nan, 0.0, 0.0]` and `twist_stiffness=np.inf` yielding `[100000.0, 100000.0, 0.0, inf]`; these values survive model finalization and can reach VBD. The preferred `add_rod(rod=...)` path therefore accepts and assembles an invalid model.

newton/_src/sim/builder.py
```diff
@@ -8618,6 +8763,215 @@ def add_rod(
...
+        label: str | None = None,
+        wrap_in_articulation: bool = True,
+        color: Vec3 | None = None,
+        body_frame_origin: Literal["start", "com"] | None = None,
+        rod: Rod | None = None,
+        junction_collision_filter: bool = True,
+    ) -> tuple[list[int], list[int]]:
+        """Adds a rod composed of capsule bodies connected by rod joints.
+
+        .. deprecated:: 1.6
```
jcarius-nv: **P2 — Standards: The documentation marks the canonical method as deprecated.**

I verified this in a generated API HTML build: because this directive is at method scope, the rendered `ModelBuilder.add_rod` entry displays a method-level **Deprecated since version 1.6** notice. Only the `positions=` form is deprecated; `add_rod(rod=...)` is the canonical replacement, so the generated API reference currently misstates the method's status.

jcarius-nv (commented):
Reviewed exact head `7d789c79`.

This is a good rework of the earlier helper-based direction. The `newton.Rod` type gives the discrete rod representation a coherent Newton-owned home, while `ModelBuilder.add_rod()` remains the assembly boundary. **Fit verdict: Fits.**

I found two implementation/documentation issues at this head; details are inline.

jumyungc:
> Reviewed exact head `7d789c79`.
> 
> This is a good rework of the earlier helper-based direction. The `newton.Rod` type gives the discrete rod representation a coherent Newton-owned home, while `ModelBuilder.add_rod()` remains the assembly boundary. **Fit verdict: Fits.**
> 
> I found two implementation/documentation issues at this head; details are inline.

Thank you for the thorough review. Both findings have now been addressed and pushed.

newton/_src/sim/builder.py
jumyungc: Done.

newton/_src/sim/builder.py
jumyungc: Done.

newton/_src/sim/rod.py
```diff
@@ -0,0 +1,594 @@
...
+        stretch_rigidity: float | None = None,
+        shear_rigidity: float | None = None,
+        bend_rigidity: float | None = None,
+        twist_rigidity: float | None = None,
+    ) -> Rod:
+        """Create a uniformly discretized straight rod.
+
+        Args:
+            start: First centerline point in world space [m].
+            direction: World-space direction, which need not be normalized.
```
jcarius-nv: Could we make the normalization and discretization behavior explicit here? The implementation normalizes `direction`, so its magnitude is ignored, then creates `segment_count + 1` points at spacing `length / segment_count`, yielding `segment_count` equal-length segments. As written, "need not be normalized" does not make it clear whether the magnitude affects the result or how it interacts with `length` and `segment_count`.

jumyungc: Done.

jcarius-nv:
Thanks @jumyungc for addressing all comments. Looks good to me now!

jcarius-nv (approved):

PR #4011, opened 2026-08-24
Preserve collision extension compatibility

nvtw:
## Description

Preserve the Newton 1.5 collision extension contract for custom broad- and narrow-phase implementations.

Speculative-contact support added optional inputs to the built-in `BroadPhase*.launch()` and `NarrowPhase.launch_custom_write()` methods. `CollisionPipeline` previously passed these new keywords even when speculative contacts were disabled. A custom implementation overriding the Newton 1.5 signature would therefore fail with an unexpected-keyword `TypeError`.

This change:

- Passes broad-phase motion inputs only when speculative contacts are active.
- Passes narrow-phase speculative inputs only when speculative contacts are configured.
- Continues passing preprocessing caches to the built-in `NarrowPhase`.
- Avoids passing post-1.5 preprocessing inputs to custom narrow-phase overrides.
- Adds regression coverage representing 1.5-compatible expert components.

### Questions for reviewers

1. **Should internal broad- and narrow-phase fields use `_` prefixes?**

   The release audit also identified runtime-owned preprocessing fields exposed without an underscore, including fields removed since 1.5. Should implementation buffers and counters on public expert classes consistently use `_` prefixes so they are not mistaken for supported API?

   This PR does not rename any fields.

2. **Should inward-edge filtering be opt-in or opt-out?**

   `Mesh.build_sdf()` now enables inward-edge filtering by default through `edge_inward_filter=True`. This is an opt-out design: users who need the Newton 1.5 contact set can pass `edge_inward_filter=False`.

   I prefer retaining this opt-out behavior: fully inward manifold edges should not contribute meaningful contacts, so filtering them by default avoids redundant work and contacts for most users. The compatibility tradeoff is that changing the default can alter the generated contact set for existing scenes. Do reviewers agree that opt-out is the right default for 1.6, given that `edge_inward_filter=False` preserves the earlier behavior, or should the filter initially require explicit opt-in?

   This PR does not change the inward-edge filtering default.

## Checklist

- [x] For user-facing changes, a changelog fragment has been added

## Test plan

```text
uv run --extra dev -m newton.tests -k test_legacy_expert_components_ignore_unconfigured_extension_inputs
uv run --extra dev -m newton.tests -k test_speculative_candidates_are_opt_in
uvx pre-commit run -a
```

The compatibility regression was verified to fail without the fix and pass with it. Speculative-contact tests pass on CPU and CUDA.

## Bug fix

**Steps to reproduce:**

1. Subclass a built-in broad or narrow phase.
2. Override `launch()` or `launch_custom_write()` using its Newton 1.5 signature.
3. Supply the expert component to a non-speculative `CollisionPipeline`.
4. Call `CollisionPipeline.collide()`.
5. Observe an unexpected-keyword `TypeError` for a post-1.5 input such as `shape_displacement`.

**Minimal reproduction:**

```python
class BroadPhase15(newton.geometry.BroadPhaseAllPairs):
    def launch(
        self,
        *args,
        device=None,
        filter_pairs=None,
        num_filter_pairs=None,
        skip_count_zero=False,
        shape_body=None,
        body_flags=None,
        include_static_kinematic_pairs=True,
    ):
        return super().launch(
            *args,
            device=device,
            filter_pairs=filter_pairs,
            num_filter_pairs=num_filter_pairs,
            skip_count_zero=skip_count_zero,
            shape_body=shape_body,
            body_flags=body_flags,
            include_static_kinematic_pairs=include_static_kinematic_pairs,
        )

# Before this PR, a non-speculative CollisionPipeline passes
# shape_displacement=None and the override raises TypeError.
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved compatibility with Newton 1.5 broad-phase and narrow-phase extensions when speculative contacts are disabled.
  - Fixed collision processing for custom extension implementations that do not support newer optional arguments.
  - Ensured rigid contacts continue to be generated correctly with legacy-compatible collision components.

- **Documentation**
  - Added a changelog entry describing the extension compatibility fix.

### DIFF
--- a/changelog/+collision-extension-compatibility-5f8d2c1a.fixed.md
+++ b/changelog/+collision-extension-compatibility-5f8d2c1a.fixed.md
@@ -1,0 +1,1 @@
+Preserve the Newton 1.5 broad-phase and narrow-phase extension call contracts unless speculative contacts are enabled.
--- a/newton/_src/sim/collide.py
+++ b/newton/_src/sim/collide.py
@@ -1904,6 +1904,7 @@ class CollisionPipeline:
             )
 
         # Run broad phase (AABBs are already expanded by effective gaps, so pass None)
+        broad_phase_speculative_kwargs = {"shape_displacement": self._shape_displacement} if speculative_active else {}
         if isinstance(self.broad_phase, BroadPhaseAllPairs):
             self.broad_phase.launch(
                 self.narrow_phase.shape_aabb_lower,
@@ -1921,9 +1922,11 @@ class CollisionPipeline:
                 filter_pairs=self.shape_pairs_excluded,
                 num_filter_pairs=self.shape_pairs_excluded_count,
                 skip_count_zero=True,  # Already zeroed by compute_shape_aabbs
-                shape_displacement=self._shape_displacement if speculative_active else None,
+                **broad_phase_speculative_kwargs,
             )
         elif isinstance(self.broad_phase, BroadPhaseSAP):
+            if speculative_active:
+                broad_phase_speculative_kwargs["sort_axis_displacement_limit"] = max_speculative_extension
             self.broad_phase.launch(
                 self.narrow_phase.shape_aabb_lower,
                 self.narrow_phase.shape_aabb_upper,
@@ -1940,8 +1943,7 @@ class CollisionPipeline:
                 filter_pairs=self.shape_pairs_excluded,
                 num_filter_pairs=self.shape_pairs_excluded_count,
                 skip_count_zero=True,  # Already zeroed by compute_shape_aabbs
-                shape_displacement=self._shape_displacement if speculative_active else None,
-                sort_axis_displacement_limit=max_speculative_extension if speculative_active else None,
+                **broad_phase_speculative_kwargs,
             )
         else:  # BroadPhaseExplicit
             self.broad_phase.launch(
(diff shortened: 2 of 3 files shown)


jcarius-nv (commented):
Thank you for the contribution and for addressing the collision-extension compatibility concerns raised by the Newton 1.6 release audit.

I reviewed exact head `0dac2e85` using the proposed Fit, Requirements, and Standards guidance from #3922:

- **Fit: Fits.** `CollisionPipeline` owns these extension calls, and the conditional forwarding is a small, proportionate compatibility shim with no new public API, dependency, or default.
- **Requirements:** The release-audit compatibility concern is addressed for non-speculative pipelines. The focused compatibility test passes, as do the existing opt-in speculative-contact tests on CPU and CUDA.
- **Standards:** No findings. The implementation, regression test, and `fixed` Towncrier fragment follow the repository guidance.

One non-blocking P2 evidence note: the added regression directly exercises a legacy `BroadPhaseAllPairs` override, while the implementation also changes the SAP and explicit branches. Those branches use the same conditional forwarding and look correct, but legacy `BroadPhaseSAP` and `BroadPhaseExplicit` signatures are not directly covered by this test.

On the reviewer questions:

1. **Internal fields:** Newly introduced runtime-owned buffers, kernels, work queues, counters, and specialization state should use `_` prefixes. Intentionally supported configuration or diagnostics can remain public; this should be an inventory, not a blanket rename. Already released public-looking fields need separate compatibility/deprecation handling or an explicit decision about the supported expert-component contract.

2. **Inward-edge filtering:** I agree with retaining the opt-out `edge_inward_filter=True` default. This is a reasonable performance improvement, and `edge_inward_filter=False` preserves the earlier contact set, so existing workflows have a direct compatibility path. The current changelog wording provides the essential migration guidance; it should remain explicit in the final 1.6 changelog and release notes.

Overall, this is a focused fix that addresses the audit concern without expanding the public surface.

jcarius-nv (approved):
Approved based on the Fit, Requirements, and Standards review above. Thank you again for addressing the Newton 1.6 release-audit compatibility concern.

PR #4015, opened 2026-08-24
Reduce CI time

nvtw:
## Description
Reduce CI time by removing redundant synchronization, device-to-host copies, and allocation cleanup, and by bounding the work performed by the PR benchmark gate. Full ASV collection remains unchanged where PR-only reductions apply.

### Benchmark changes
- Add a separate PR ASV environment that omits PyTorch because the selected DR Legs benchmark disables policy inference; all other example dependencies and the pinned Warp/MuJoCo stack remain aligned with full ASV.
- Interleave base/head rounds, cap PR comparisons to 30 seconds, and reduce expensive fast-benchmark repeats to at most three for cloth, convex contacts, IK, inverse dynamics, XPBD quadruped, and MuJoCo selection. Benchmark cases and parameters are unchanged, and full ASV retains the original repeat counts.
- Reduce the PR-only Allegro KPI workload from 300 frames x 2 samples to 200 frames x 1 sample. All configured world counts and the blocking simulation and p95 runtime metrics remain covered; full ASV retains the larger workload.
- Measure representative tiled-camera modes in the PR gate: color+depth and depth-only for image-priority rendering, plus color+depth for pixel-priority rendering. Use two interleaved rounds with one repeat and warm only selected modes; full ASV still measures all three output modes for both render orders.
- Keep directly measured `time_*`, simulation, mean-latency, and p95-latency regressions in the PR gate while dropping redundant camera modes, Teleop mean latency, and Newton-overhead simulation metrics. Teleop p95 latency and representative camera modes remain blocking, and full ASV collection retains every metric.
- Add ASV configuration tests covering PR/full dependency parity, repeat caps, Allegro workload reduction, benchmark-selection regexes, and camera mode selection/warmup so future edits cannot silently reduce the intended gate.

### Test changes
- Reuse the cross-slide-table body-state copy already collected once per frame instead of copying it again in `test_post_step()` and `test_final()`. The same current-frame state and all existing bounds/tracking assertions are retained.
- Record only the final D6 state for the Kamino gimbal limit sweep instead of copying every intermediate state. The final limit oracle is unchanged; tests that require trajectories keep full recording.
- Remove per-step host synchronization from Kamino rollout and multi-world tests, retaining synchronization for progress output and relying on ordered Warp operations plus the existing final host reads. Simulation steps, contacts, and final comparisons are unchanged.
- Move the Kamino joint-friction spin-down test and the long Kamino/VBD friction-ramp and stopping-distance cases into separate test classes so class-level parallel scheduling can balance them independently. No test cases or solver/configuration variants are dropped by the split.
- Reduce only the Kamino friction-ramp grid from 25 to 16 cells, retaining the minimum/maximum coefficient and angle, static and sliding regimes for representative rows, and the full stopping-distance oracle. XPBD and MuJoCo keep the original grid, while VBD keeps its denser transition-focused grid.
- Record pendulum period and energy trajectories on-device and copy them once after simulation instead of synchronizing every step. The same timesteps, solver matrix, trajectory samples, and analytical assertions are retained.
- Skip garbage collection for statically skipped tests, batch CPU cleanup every eight executed tests, and keep per-test CUDA cleanup to preserve peak GPU-memory behavior. New tests cover static/runtime skips, executed tests, CPU batching, final cleanup, and both text/JUnit result classes.

## Checklist
- [x] The documentation is up to date with these changes (no user-facing documentation is affected)
- [x] No changelog fragment is required because this only changes CI, benchmarks, and tests

## Test plan
- MR CI exercises the existing Newton test matrix and AWS GPU benchmark comparison.
- `asv/tests/test_benchmark_simulation.py` verifies PR benchmark selection, dependency parity, repeat caps, Allegro sampling, and camera coverage.
- `newton/tests/test_unittest_utils.py` verifies the revised allocation-cleanup behavior for both parallel result implementations.
- Existing cable, Kamino gimbal/solver/joint-friction, physics-verification, and rigid-friction tests retain their original behavioral assertions while avoiding redundant work.

## Summary by CodeRabbit

- **New Features**
  - Added streamlined pull-request performance checks for camera rendering, simulation tracking, representative metrics, and teleoperation latency.
  - Added configurable benchmark execution with targeted coverage, shorter runs, and reduced sampling for faster feedback.

- **Bug Fixes**
  - Improved diagnostic validation for cable simulations and physics verification.
  - Reduced unnecessary solver synchronization and improved test resource cleanup, including CUDA allocation handling.

- **Tests**
  - Expanded coverage for benchmark configuration, camera warmup, friction ramps, trajectory recording, energy conservation, pendulum timing, and cleanup behavior.
  - Added dedicated validation suites for supported friction solvers.


nvtw:
According to AI, part of the benchmark speedup will only become effective after this MR is merged

asv/benchmarks/simulation/bench_sensor_tiled_camera.py
```diff
@@ -340,14 +349,32 @@ class TiledCameraQuadruped(_SceneBenchmark):
 
 class FastSensorTiledCamera(_SceneBenchmark):
     scene = "franka_cabinet"
+    repeat = 1
```
jcarius-nv: Was it intentional for `repeat = 1` to apply to full ASV as well as the PR gate? Previously, with no explicit repeat, full ASV used its adaptive sampling. I do not see this as a merge blocker, but it seems worth confirming because the description says the reductions are PR-only and full ASV remains unchanged. If the global reduction is intentional, could the description say so; otherwise should this depend on `NEWTON_ASV_PR_GATE`?

asv/tests/test_benchmark_simulation.py
```diff
@@ -209,16 +243,25 @@ def test_aws_benchmark_comparison_gates_only_runtime_metrics(self):
         """Gate PR comparisons on runtime while retaining dashboard metrics."""
         workflow_path = BENCHMARK_DIR.parents[1] / ".github" / "workflows" / "aws_gpu_benchmarks.yml"
         workflow = workflow_path.read_text(encoding="utf-8")
+        self.assertIn("--config asv-pr.conf.json", workflow)
+        self.assertIn("--interleave-rounds", workflow)
         patterns = tuple(re.compile(selection) for selection in re.findall(r"-b '([^']+)'", workflow))
         self.assertTrue(patterns)
 
         blocking_benchmarks = (
```
jcarius-nv, line 299: Could this assertion use ASV's discovered benchmark inventory rather than hand-written benchmark-name strings? As written, renaming or deleting an actual blocking benchmark could leave these literals and the test green. Since the stated goal is to prevent future edits from silently reducing the gate, a discovery-derived check would cover that gap.

newton/tests/test_rigid_friction_ramp.py
```diff
@@ -66,6 +66,10 @@
 # is noisy near 50 deg. mu=1.00 therefore exercises only the static side.
 _DEFAULT_MUS = (0.10, 0.30, 0.50, 0.70, 1.00)
 _DEFAULT_ANGLES_DEG = (3.0, 10.0, 20.0, 30.0, 40.0)
+# Kamino retains the full coefficient and angle range while omitting middle
+# cells already covered by the default grid on XPBD and MuJoCo.
+_KAMINO_MUS = (0.10, 0.30, 0.70, 1.00)
+_KAMINO_ANGLES_DEG = (3.0, 10.0, 30.0, 40.0)
```
jcarius-nv: Could you confirm that permanently reducing the Kamino grid from 25 to 16 cells is acceptable? Unlike the ASV changes, this reduction applies to every run, not only the PR gate. Do we have historical-failure, mutation, or sensitivity evidence that the omitted middle cells are redundant, or is there another full/nightly path that retains them? I do not consider this a merge blocker; I mainly want to confirm the intended coverage tradeoff.

asv-pr.conf.json
```diff
@@ -0,0 +1,22 @@
...
+  "branches": ["main"],
+  "dvcs": "git",
+  "environment_type": "virtualenv",
+  "show_commit_url": "https://github.com/newton-physics/newton/commit/",
+  "benchmark_dir": "asv/benchmarks",
+  "env_dir": "asv/pr-env",
+  "results_dir": "asv/results",
+  "html_dir": "asv/html",
+  "build_cache_size": 20,
+  "default_benchmark_timeout": 600,
```
jcarius-nv: How do you see the separate PR ASV configuration being maintained over time? Since it largely mirrors `asv.conf.json`, could an overlay, generated configuration, or supported command-line override express only the PR-specific differences and avoid configuration drift? Or is keeping a complete explicit file the least fragile option with ASV today?

asv-pr.conf.json
```diff
@@ -0,0 +1,22 @@
...
+  "build_command": ["python -m build --wheel -o {build_cache_dir} {build_dir}"],
+  "install_command": [
+    "in-dir={env_dir} python -m pip install warp-lang==1.17.0.dev20260807 mujoco==3.11.0 mujoco-warp==3.11.0 --extra-index-url=https://pypi.nvidia.com/",
```
jcarius-nv: Is there a way to keep the Warp, MuJoCo, and MuJoCo Warp pins in one source of truth so dependency bumps cannot update one ASV configuration without the other? Even if both configuration files remain, generating or validating this command from shared pin data would make updates atomic.

.github/workflows/aws_gpu_benchmarks.yml
```diff
@@ -149,16 +149,25 @@ jobs:
...
             --no-only-changed \
             -e \
-            -b 'Fast\w*\.time_' \
+            -b 'Fast(?!SensorTiledCamera)\w*\.time_' \
+            -b 'FastSensorTiledCamera\.time_render_(color_depth|depth_only)(\(|$)' \
+            -b 'FastSensorTiledCameraPixel\.time_render_color_depth(\(|$)' \
             -b 'Fast(?!NewtonOverhead)\w*\.track_simulate(\(|$)' \
-            -b 'Fast\w*\.track_(mean|p95)_\w*(time|ms)(\(|$)' \
+            -b 'Fast(?!TeleopMuJoCo)\w*\.track_mean_\w*(time|ms)(\(|$)' \
+            -b 'Fast\w*\.track_p95_\w*(time|ms)(\(|$)' \
```
jcarius-nv: Is there a less fragile way to define the PR benchmark set than keeping this filter policy in workflow regexes? For example, could a checked-in selection manifest or benchmark support module own the group while the workflow invokes one named selection? That would make benchmark renames easier to validate and review.

asv.conf.json
```diff
@@ -15,8 +15,8 @@
   "default_benchmark_timeout": 600,
   "build_command": ["python -m build --wheel -o {build_cache_dir} {build_dir}"],
   "install_command": [
-    "in-dir={env_dir} python -m pip install {wheel_file}[dev] --extra-index-url=https://pypi.nvidia.com/",
     "in-dir={env_dir} python -m pip install warp-lang==1.17.0.dev20260807 mujoco==3.11.0 mujoco-warp==3.11.0 --extra-index-url=https://pypi.nvidia.com/",
+    "in-dir={env_dir} python -m pip install {wheel_file}[examples] --extra-index-url=https://pypi.nvidia.com/",
```
jcarius-nv, line 19: Was the full ASV suite, including discovery/import of every benchmark, verified with only the `[examples]` extra? `dev` currently appears to add coverage on top of examples, so this may be sufficient, but the PR gate exercises only its selected subset. It would be useful to confirm that no full-collection benchmark relies on a development-only dependency.

jcarius-nv (commented):
Thanks for putting this together. This is a solid change and should materially shorten PR benchmark feedback while preserving representative coverage. The removal of redundant synchronization and copies, together with the explicit PR gate, is thoughtful, and the 35m52s benchmark run supports the direction.

One clarification for the description: the earlier `--durations 30` argument did not cap comparisons to 30 seconds. In ASV, that value means to report the 30 slowest benchmark durations, and `asv continuous` does not accept the option. Since it has now been removed, could you reword or remove the claim that PR comparisons are capped at 30 seconds so the description matches the implementation?

I left a few questions inline, mostly about confirmation of intended coverage and long-term maintainability. I do not view the camera-repeat or Kamino-grid questions as merge blockers. More generally, is there now a risk that the lower repeat counts and narrower PR selection make benchmark comparisons flaky again? If so, do we have enough noise monitoring, thresholds, or retry policy to prevent false regression signals?

Thanks again for the careful work here.

nvtw:
> Thanks for putting this together. This is a solid change and should materially shorten PR benchmark feedback while preserving representative coverage. The removal of redundant synchronization and copies, together with the explicit PR gate, is thoughtful, and the 35m52s benchmark run supports the direction.
> 
> One clarification for the description: the earlier `--durations 30` argument did not cap comparisons to 30 seconds. In ASV, that value means to report the 30 slowest benchmark durations, and `asv continuous` does not accept the option. Since it has now been removed, could you reword or remove the claim that PR comparisons are capped at 30 seconds so the description matches the implementation?
> 
> I left a few questions inline, mostly about confirmation of intended coverage and long-term maintainability. I do not view the camera-repeat or Kamino-grid questions as merge blockers. More generally, is there now a risk that the lower repeat counts and narrower PR selection make benchmark comparisons flaky again? If so, do we have enough noise monitoring, thresholds, or retry policy to prevent false regression signals?
> 
> Thanks again for the careful work here.

I don't know about the flakiness. And we won't know unless we try. So I'm in favor of trying and if flakiness appears, we can still revert pieces of this MR.

asv/benchmarks/simulation/bench_sensor_tiled_camera.py
nvtw: Not intentional for full ASV. Fixed: repeat=1 is now PR-gate-only; full ASV retains adaptive sampling.

asv/tests/test_benchmark_simulation.py
nvtw: Done. The test discovers ASV’s actual inventory and verifies every protected/excluded benchmark exists before applying the selection policy.

newton/tests/test_rigid_friction_ramp.py
nvtw: We lacked evidence that the omitted cells were redundant, so I restored the full 5×5 Kamino grid and added a regression assertion.

asv-pr.conf.json
nvtw: Replaced the duplicated config with a temporary PR config generated from asv.conf.json; only the PR environment directory and Torch omission differ.

asv-pr.conf.json
nvtw: Resolved by generating the PR config from asv.conf.json. Warp, MuJoCo, and MuJoCo Warp now have one source of truth.

.github/workflows/aws_gpu_benchmarks.yml
nvtw: Moved selection into asv/pr_benchmarks.txt. The workflow invokes one runner, and tests validate the manifest against discovered benchmarks.

asv.conf.json
nvtw: Confirmed in an isolated environment containing only the pinned stack, Newton [examples], and asv_runner: all 152 benchmarks imported and were discovered successfully.

jcarius-nv (approved):
The review feedback has been addressed and the focused tests and CI checks pass. This is a solid change that should make the benchmark gate faster. Thank you!

PR #4018, opened 2026-08-24
Render Gaussian splat shapes in ViewerViser

### LINKED ISSUE #2099: Add Gaussian splats rendering in ViewerViser
Viser has direct support for Gaussian splats: https://viser.studio/main/examples/scene/gaussian_splats/

aneangel:
## Description

`ViewerViser.log_gaussian()` was a no-op, so `newton.Gaussian` splat shapes silently disappeared when viewing a scene with `--viewer viser`. This adds a real implementation that maps each Gaussian to viser's native `add_gaussian_splats` renderer:

- Local-space covariance is computed as `Σ = R · diag(scale²) · Rᵀ` from each Gaussian's quaternion rotation and per-axis scale (verified against viser's own `01_scene/09_gaussian_splats.py` reference example, which computes covariance the same way).
- Color is decoded from the DC spherical-harmonics term using the same `0.5 + SH_C0 * sh` formula `ViewerGL` already uses, so colors match across backends.
- The point cloud (centers/covariances/colors/opacities) is uploaded once per asset and cached by name; subsequent frames just update the scene node's `position`/`wxyz`, avoiding a full re-upload every step.
- `clear_model()` now drops the Gaussian cache entry alongside the existing per-name caches (meshes, instances) so a model reset can't leave a stale cache key pointing at an already-removed scene handle.

Note: viser's `add_gaussian_splats` is marked experimental upstream; documented as such in the docstring.

Closes #2099

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_viewer_viser_gaussian
```

All 4 tests confirmed to fail against a no-op `log_gaussian()` baseline and pass with the fix. Also manually verified end-to-end against both a synthetic Gaussian asset and viser's own reference `nike.splat` asset (parsed independently, fed through `newton.Gaussian`) rendered live in a browser via `--viewer viser`.

## Bug fix

**Steps to reproduce:**

1. Build a model with a body carrying a `newton.Gaussian` shape (`builder.add_shape_gaussian(...)`).
2. Run with `--viewer viser` and open the browser.
3. Observe: no Gaussian splat renders, even though `viewer.show_gaussians = True`.

**Minimal reproduction:**

```python
import numpy as np
import newton
import warp as wp

gaussian = newton.Gaussian(
    positions=np.random.normal(size=(1000, 3)).astype(np.float32),
    scales=np.full((1000, 3), 0.05, dtype=np.float32),
)
body = builder.add_link()
builder.add_shape_gaussian(body, gaussian=gaussian)
model = builder.finalize()

viewer = newton.viewer.ViewerViser()
viewer.set_model(model)
viewer.show_gaussians = True
state = model.state()
viewer.begin_frame(0.0)
viewer.log_state(state)
viewer.end_frame()
# Before this fix: nothing appears in the browser.
```

## Summary by CodeRabbit

* **New Features**
  * Added support for rendering Gaussian splat shapes in ViewerViser.
  * Gaussian data is cached and reused efficiently, with per-frame position and rotation updates.
  * Added support for visibility, opacity, colors, covariance, optional transforms, and asset replacement.

* **Bug Fixes**
  * Empty or removed Gaussian assets are now cleared correctly.
  * Gaussian rendering state is cleared when the displayed model is reset.

* **Tests**
  * Added coverage for rendering, caching, transforms, visibility, replacement, and cleanup behavior.


newton/tests/test_viewer_viser_gaussian.py
eric-heiden, line 1: I think we don't need this file because it only does mock testing.

aneangel: I went ahead and removed the file from the PR, was following the struct from test_view_layers.py but if you don't think it is needed since it can't check for Viser API drift we won't include it we can always add back in if you think otherwise.

eric-heiden (approved):
Thanks!

PR #4019, opened 2026-08-24
Stop ground-plane grids from flickering in ViewerViser

### LINKED ISSUE #2099: Add Gaussian splats rendering in ViewerViser

aneangel:
## Description

`_log_plane_instances()` called `self._remove_plane_handles(name)` unconditionally at the top of every call, then rebuilt the grid from scratch — even when re-logging the exact same static plane on every simulation frame. Since a ground plane gets re-logged every frame just like everything else, this meant: destroy the plane's viser handles, recreate them, destroy, recreate — once per frame, forever. The visible symptom is the ground grid flickering in the browser.

This caches each plane batch's instance count and extents (`geometry_key`) alongside its handles. When a later call's geometry is unchanged, the existing handles are just repositioned (`handle.position` / `handle.wxyz`) instead of torn down and rebuilt — mirroring the pattern `log_instances()` already uses for batched meshes elsewhere in this file. Only a genuine change in instance count or extents triggers a rebuild.

Found while validating #4018 (Gaussian splat rendering) in the same viewer backend.

Closes #2099

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_viewer_viser_plane
```

Both tests confirmed to fail against the pre-fix "always remove and recreate" baseline and pass with the fix. Also manually verified via `--viewer viser` that a ground plane stays visually stable across frames instead of flickering.

## Bug fix

**Steps to reproduce:**

1. Build any model with `builder.add_ground_plane()`.
2. Run with `--viewer viser` and open the browser.
3. Observe the ground grid flickering every frame.

**Minimal reproduction:**

```python
import newton
import warp as wp

builder.add_ground_plane()
model = builder.finalize()

viewer = newton.viewer.ViewerViser()
viewer.set_model(model)
state = model.state()

for i in range(100):
    viewer.begin_frame(float(i))
    viewer.log_state(state)
    viewer.end_frame()
# Before this fix: every log_state() call tears down and rebuilds the
# ground plane's grid handles, producing a visible flicker in the browser.
```

## Summary by CodeRabbit

- **New Features**
  - Added support for displaying Gaussian splat assets in the viewer, including transforms, visibility controls, replacement, and removal.
  - Improved rendering of Gaussian splats with correct colors, rotations, scales, and positioning.
  - Gaussian splats without an explicit transform now use the identity transform by default.

- **Bug Fixes**
  - Prevented plane-grid flicker during updates when geometry remains unchanged.
  - Improved cleanup of removed mesh and point visualizations.
  - Ensured stale Gaussian splat visuals are removed during asset updates.


changelog/2099.fixed.1.md
```diff
@@ -0,0 +1 @@
+Stop `ViewerViser` from flickering ground-plane and plane-shape grids. Plane grid handles were torn down and rebuilt on every frame regardless of whether the pose changed, so the grid briefly disappeared each frame while the removal and re-add messages round-tripped over the websocket. Grid handles are now reused and just repositioned when their instance count and extents are unchanged.
```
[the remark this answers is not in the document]
aneangel: Keeping the `.1.md` suffix on purpose: this fix and #4018 (Gaussian splat rendering) both close issue #2099 as separate PRs, and #4018 already added `changelog/2099.fixed.md`. Renaming this one to the same filename would collide once both land. Multiple entries per issue is exactly what Towncrier's numeric-suffix convention is for (see the repo's changelog/README.md, e.g. `3607.fixed.md` / `3607.fixed.1.md`), so leaving this as `2099.fixed.1.md`.

newton/tests/test_viewer_viser_plane.py
eric-heiden, line 1: This file is also just a bunch of mock tests that can't really test actual functionality of the viser viewer, so I think they should be removed.

aneangel: went ahead and removed this one as well.

eric-heiden (approved):
Thanks!

PR #4022, opened 2026-08-24
Allow usd-core 26.08 and adopt usd-exchange 3.x on aarch64

### LINKED ISSUE #2216: Remove USD multithreading workaround after OpenUSD fix lands
## Context

Newton documents a known limitation where importing USD files with many (>30) mesh colliders under the same rigid body can crash in `UsdPhysics.LoadUsdPhysicsFromRange` due to a thread-safety bug in OpenUSD (see #1743).

The upstream fix is tracked in [PixarAnimationStudios/OpenUSD#4002](https://github.com/PixarAnimationStudios/OpenUSD/pull/4002). Once that PR is merged and ships in a new `usd-core` release, we should clean up our workaround.

## What to do when the fix lands

1. **Remove the "Limitations" section** in `docs/concepts/usd_parsing.rst` (lines ~540–598) that documents the `PXR_WORK_THREAD_LIMIT` workaround.
2. **Remove the workaround** in `newton/tests/thirdparty/unittest_parallel.py` that sets `PXR_WORK_THREAD_LIMIT=1` before tests.
3. **Update minimum `usd-core` version** if appropriate.
4. Add a `CHANGELOG.md` entry under `Removed`.

## Upstream references

- Newton issue: #1743
- OpenUSD PR: https://github.com/PixarAnimationStudios/OpenUSD/pull/4002
- Pixar internal tracking: USD-11936
- Blocked by release of `usd-exchange`. See comments in: 
  -  https://github.com/newton-physics/newton/pull/2594#issuecomment-4337425949
  -  https://github.com/newton-physics/newton/pull/2594#issuecomment-4337506554

andrewkaufman:
## Description

Allows `usd-core` 26.08 and moves aarch64 to `usd-exchange` 3.x.

`usd-core` was capped at `<26.5` and `usd-exchange` at `<3`, gated to Python `< 3.13`. That gate left aarch64 on Python 3.13 resolving to no USD provider at all, since neither requirement applied.

OpenUSD 26.08 renamed `UsdPhysics.LoadUsdPhysicsFromRange` to [`UsdPhysicsLoadStageFromPrimRange`](https://github.com/PixarAnimationStudios/OpenUSD/blob/v26.08/pxr/usd/usdPhysics/parseUtils.h) and deprecated the old name, so the three call sites now go through a wrapper that prefers the new name where it exists and falls back otherwise. Both spellings take the same arguments and return the same descriptor dict. The two tests that patched the entry point by name patch the wrapper instead, so they assert the parser ran without depending on the installed runtime's spelling.

The second commit addresses #2216. The test suite set `PXR_WORK_THREAD_LIMIT=1` unconditionally to dodge the collider-parsing race, which serialized all OpenUSD work in CI. That race is fixed in 26.08, so the limit is now applied only to older runtimes and CI regains multi-threaded USD parsing on 26.08.

The workaround cannot be removed outright, since the `usd-core` floor still allows releases that predate the fix. The docs are rescoped accordingly rather than describing the fix as pending.

Reading the version is more awkward than expected: OpenUSD latches `PXR_WORK_THREAD_LIMIT` when `pxr` initializes, so the check must run before any test module imports `pxr`. That rules out `Usd.GetVersion()`, and also rules out importing a newton USD module, because `newton_usd_schemas` imports `pxr` at module scope. The check therefore reads distribution metadata: `usd-core` is versioned directly, and `usd-exchange` advertises its bundled OpenUSD as a `usd<major><minor>` extra. An unidentifiable runtime is treated as affected, and a caller-provided override is still preserved.

## Checklist

## Test plan

[OpenUSD PR 4002](https://github.com/PixarAnimationStudios/OpenUSD/pull/4002) is closed unmerged, so the 26.08 claim in #2216 was measured rather than assumed. With 200 colliders under one rigid body and no thread limit, 26.3 segfaults and 26.08 completes 40 iterations with stable descriptor counts, also at 500 colliders over 60 iterations. As negative controls, 26.3 with the limit set passes, and the same script still segfaults on 26.3.

Ran 17 USD, MJCF and URDF test modules, 1111 tests, against `usd-core` 26.3 and 26.08 in environments differing only in the USD provider: no failures on either, identical counts. The first commit was also verified alone, since it must stand on its own when bisecting. Separately, `usd-exchange` 3.0.0 on its own resolves and passes the same modules, and all 25 `newton-usd-schemas` 0.5.0 schema types register against its bundled 26.08.

USD requirements resolve across Python 3.10 to 3.14 on linux x86_64 and aarch64, macOS and Windows.

aarch64 and macOS are verified by resolution only; the tests above ran on linux x86_64.

Test modules, run in each environment:

```
uv run --extra dev --extra importers -m newton.tests -k test_import_usd
```

with `test_import_usd_deformable_{attachments,cable,cloth,filtered_pairs,groups,mixed,volume}`,
`test_import_usd_mpm`, `test_import_usd_multi_dof`, `test_sdf_usd`, `test_sites_usd_import`,
`test_usd_mesh_loading`, `test_viewer_usd`, `test_import_mjcf`, `test_import_urdf` and
`test_lazy_imports`.

The thread-safety measurement, run with `PXR_WORK_THREAD_LIMIT` unset against each runtime:

```python

body = UsdGeom.Xform.Define(stage, "/World/Body")
for i in range(200):
    cube = UsdGeom.Cube.Define(stage, f"/World/Body/col_{i:04d}")
    cube.CreateSizeAttr(0.05)
    UsdPhysics.CollisionAPI.Apply(cube.GetPrim())

load = getattr(UsdPhysics, "UsdPhysicsLoadStageFromPrimRange", UsdPhysics.LoadUsdPhysicsFromRange)
for _ in range(40):
    load(stage, ["/"], excludePaths=[])
```

On `usd-core` 26.3 this segfaults; on 26.08 it completes with stable descriptor counts.

## Summary by CodeRabbit

- **New Features**
  - Expanded USD support for aarch64 systems, including Python 3.13 compatibility.
  - Added compatibility with OpenUSD 26.08 and newer physics parsing APIs.

- **Bug Fixes**
  - Resolved USD mesh-collider parsing crashes on OpenUSD 26.08 and newer.
  - Retained the concurrency workaround only for older OpenUSD versions.

- **Documentation**
  - Updated USD parsing guidance to recommend upgrading OpenUSD rather than applying global thread limits.

Fixes #2216


eric-heiden (dismissed):
Thanks!

chschuma-disney (dismissed):
LGTM

jcarius-nv (changes requested):
Since this is a rather fundamental change, we're holding this until after the 1.6 release branch is cut. At this point Andrew can merge it.

Currently there's also this open issue that Warp is experiencing with the new packages: https://github.com/PixarAnimationStudios/OpenUSD/issues/4198

jcarius-nv (approved):
https://github.com/PixarAnimationStudios/OpenUSD/issues/4198 is still open, but I don't think we need to block on it.

Coderabbit points out a test coverage gap, but it seems a bit excessive to run all tests with two usd versions. I'm going to approve and leave it to @andrewkaufman 's judgement whether more testing makes sense or not.

PR #4023, opened 2026-08-25
Add per-layer viewer render styles

adityaajay-nv:
## Description

Add render-only appearance controls for named viewer layers. `LayerRenderStyle`
lets a caller replace source colors and set opacity without changing model
materials or simulation state. `set_layer_shape_visibility()` accepts an
explicit Boolean mask, leaving scene semantics with the application that owns
them.

`ViewerGL` stores the resolved style in a per-instance buffer. It draws opaque
geometry first, then draws translucent geometry with alpha blending enabled and
depth writes disabled. Hidden and translucent instances don't enter the shadow
or opaque-edge passes.

The public API and visualization guide include the new layer controls.

## Checklist

- [x] `CHANGELOG.md` has been updated (user-facing change)

## Test plan

```bash
uv run docs/generate_api.py
uv run --extra dev -m newton.tests -k test_viewer_visible_worlds
uv run --extra dev -m newton.tests -k test_viewer_world_offsets
uvx pre-commit run -a
```

The three viewer suites ran 46 tests on an RTX 6000 Ada and passed. Applying
only the new layer tests to unmodified `main` fails at import because
`LayerRenderStyle` doesn't exist there.

I also rendered two overlapping boxes through headless `ViewerGL`: an opaque
orange reference and a blue candidate at 0.35 opacity. The captured overlap
contained the expected mixed color while the non-overlapping faces retained
their assigned colors.

## New feature / API change

```python
import numpy as np
from newton.viewer import LayerRenderStyle

viewer.activate("candidate")
viewer.set_model(model)
viewer.set_layer_render_style(
    "candidate",
    LayerRenderStyle(color=(0.3, 0.7, 1.0), opacity=0.3),
)

visibility = np.ones(model.shape_count, dtype=bool)
visibility[fixture_shape_indices] = False
viewer.set_layer_shape_visibility("candidate", visibility)
```

## Summary by CodeRabbit

* **New Features**
  * Added experimental layer appearance controls in `ViewerGL`, including solid colors, opacity, and per-shape visibility.
  * Added `LayerRenderStyle` to the public viewer API.
  * Added styling support for meshes, capsules, Gaussian shapes, and SDF isosurfaces.
  * Improved translucent rendering with depth sorting and opacity-aware rendering.

* **Bug Fixes**
  * Added validation for shape-visibility masks and reset masks when switching models.

* **Documentation**
  * Updated viewer API and visualization guides with layer appearance details.


nvtw:
AI found an issue that needs a closer look and might need adjustments:

  - Performance blocker: mixed visible/hidden or opaque/translucent batches degrade from one instanced draw to one draw call per opaque instance. Large multi-world
    scenes could produce thousands of draw calls in the main, shadow, and edge passes. This is already an unresolved inline review comment
    (https://github.com/newton-physics/newton/pull/4023#discussion_r3857700407).

nvtw:
@adityaajay-nv  keep in mind that the CLA must be signed - I think there are some legal reasons for that

nvtw (changes requested):
CI must pass

eric-heiden:
Thanks @adityaajay-nv, please rebase on current `main` so this can reuse the transparency work from #3053.

This branch was cut before #3053 ("Add opacity support to shapes and viewers") landed, so it never picked that code up. As it stands, the viewer would end up with two separate transparency implementations.

What `main` already has, and what this PR builds a second time:

| Concern | Already on `main` (#3053) | Added again here |
| --- | --- | --- |
| Per-instance alpha | `aOpacity` at attribute location `9` | `aRenderStyle` at the same location `9` |
| Opaque vs transparent split | `_split_transparent_objects` in `_render_scene` | a second split in `_render_scene` |
| Draw ordering | weighted OIT, with alpha blending as the fallback | CPU back-to-front sort, one draw call per instance |

Rebasing should also clear up the performance concern raised earlier in this PR. The existing path keeps the single instanced draw call and lets OIT sort out ordering, so there is no per-instance Python sort and no per-instance draw. After the rebase, the per-layer feature should mostly reduce to feeding opacity into the existing pipeline, plus the new solid-color override.

PR #4034, opened 2026-08-25
Fix hydroelastic SDF tests on multiple GPUs

### LINKED ISSUE #4032: [BUG] Multi-GPU nightly crashes on cross-device hydroelastic SDF test
## Bug Description

The scheduled multi-GPU test suite has failed on four consecutive nightly runs (August 22-25, 2026):

- https://github.com/newton-physics/newton/actions/runs/32564178555
- https://github.com/newton-physics/newton/actions/runs/32630202961
- https://github.com/newton-physics/newton/actions/runs/32710933750
- https://github.com/newton-physics/newton/actions/runs/32830939982

All four runs have the same primary failure:

```text
test_hydroelastic_attached_sdf_uses_padding_metadata_cuda_1
Warp CUDA error 700: an illegal memory access was encountered
```

The test passes on `cuda:0` and fails on `cuda:1`. After the illegal access poisons the CUDA context, the suite reports 66 cascading errors, including failures to allocate 12-112 bytes. The August 23 XPBD assertion also occurred after the original CUDA fault and is collateral.

This began after #3719 merged. The August 21 nightly passed before the merge, and the August 22 nightly was the first failure. Both runs used Warp `1.17.0.dev20260807`, so this is not a Warp-nightly version change.

### Root cause

The test builds its texture SDF without specifying its parameterized device:

```python
mesh.build_sdf(max_resolution=32, margin=0.3)
```

This allocates the texture SDF on Warp's default `cuda:0`. The test later finalizes the model with `device=device`, including `cuda:1`.

`ModelBuilder.finalize()` copies the `TextureSDFData` struct to the model device but retains the original device-owned texture handles. A hydroelastic sampling kernel on `cuda:1` therefore dereferences `cuda:0` texture handles and triggers CUDA error 700. The exact kernel named in the traceback varies between `scatter_iso_subblock` and `count_iso_voxel_children` because CUDA reports the fault asynchronously.

The PR GPU check did not catch this because it used a single-GPU `g7e.2xlarge`; the nightly uses a two-GPU `g7e.12xlarge`.

## Reproduction Script

Run the added test on a machine with at least two CUDA devices:

```bash
  -k test_hydroelastic_attached_sdf_uses_padding_metadata_cuda_1
```

Expected: the padding-metadata behavior is tested on `cuda:1` without a CUDA fault.

Actual: the test triggers CUDA error 700 and poisons the CUDA context.

## Proposed resolution

- Pass `device=device` to `mesh.build_sdf()` in the attached-SDF tests so the test resources and finalized model are colocated.
- Validate prebuilt SDF texture ownership in `ModelBuilder.finalize()`. If the texture device differs from the target model device, raise a clear `ValueError` instructing the caller to rebuild the SDF on the target device rather than allowing an illegal memory access.
- Add focused coverage for the device-mismatch rejection unless cross-device texture migration is intentionally supported.

## Acceptance criteria

- The focused test passes on both `cuda:0` and `cuda:1`.
- The full multi-GPU nightly passes.
- Passing a prebuilt texture SDF from one CUDA device to a model finalized on another fails safely with a clear error, or is explicitly migrated if cross-device support is implemented.

## System Information

- First bad Newton merge: `cd6a65c2` (#3719)
- Warp: `1.17.0.dev20260807`
- Python: 3.12
- Nightly runner: AWS `g7e.12xlarge`
- GPUs: 2 x NVIDIA RTX PRO 6000 Blackwell Server Edition

mzamoramora-nvidia:
## Description

Build the attached SDF test fixtures on the parameterized test device.

Without this, the `cuda:1` variants build their SDF texture data on the
default `cuda:0`, then use those texture handles in a model finalized on
`cuda:1`.

Closes #4032.

## Checklist

      changelog fragment instructions

This is a test-only correction with no user-facing behavior change, so no
changelog fragment is needed.

## Test plan

On an Ubuntu 24.04 x86_64 host with two NVIDIA L40 GPUs and CUDA peer
access unavailable:

```bash
  -k test_hydroelastic_attached_sdf_uses_padding_metadata_cuda_1

  -k test_hydroelastic_attached_sdf
```

The exact `cuda:1` regression passed, and all four affected
device-parameterized tests passed.

```bash
uvx pre-commit run -a
```

All hooks passed.

The complete test suite is also running on the two-L40 host.

## Bug fix

**Steps to reproduce:**

1. Use a machine with at least two CUDA devices.
2. Run the `cuda:1` attached-SDF hydroelastic test.
3. Without this fix, `Mesh.build_sdf()` creates the texture data on the
   default `cuda:0`, while the model is finalized and evaluated on
   `cuda:1`.

## Summary by CodeRabbit

* **Tests**
  * Updated hydroelastic tests to explicitly select the device when building mesh signed-distance fields.
  * Confirmed both attached-SDF test cases pass with the selected device.


nvtw (approved):
Thanks for the fix

PR #3708, opened 2026-07-28
Fix PR author auto-assignment

jcarius-nv:
## Description

Check whether the PR creator can be assigned to the specific pull request
instead of using the repository-wide eligibility endpoint.

The repository-wide endpoint rejects external contributors who are nevertheless
eligible as participants in their own PR. This caused the workflow to warn and
skip assignment even though GitHub allowed the author to be assigned manually.

## Checklist

- [x] New or existing tests cover these changes (live API regression check)
- [x] The documentation is up to date with these changes (not applicable)
- [x] `CHANGELOG.md` has been updated (not applicable; workflow-only change)

## Test plan

- Run `/usr/bin/env UV_CACHE_DIR=/tmp/newton-uv-cache UV_TOOL_DIR=/tmp/newton-uv-tools PRE_COMMIT_HOME=/tmp/newton-pre-commit-cache uvx --python 3.12 pre-commit run -a`.
- Verify the repository-wide assignee endpoint returns `404` for the author of
  PR #3649.
- Verify the PR-specific assignee endpoint returns `204` for the same author.

## Bug fix

**Steps to reproduce:**

1. Open a draft PR as an external contributor who lacks repository-wide
   assignment eligibility.
2. Mark the PR ready for review.
3. Observe the auto-assignment workflow warn that the author is not assignable
   and leave the PR unassigned.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pull request author eligibility checks during automatic assignment.
  * Added more reliable handling when an author cannot be assigned, including warning messages and skipping assignment when appropriate.
  * Continued verification that the author was successfully added as an assignee.

PR #3785, opened 2026-08-04
Fix particle collisions for disabled USD colliders

mmichelis:
## Description

Clear both collision flags for disabled USD colliders so physics behavior does not depend on `load_visual_shapes`. Extend the existing USD importer test to cover physics-only imports.

Closes #3784.

## Checklist

- [x] New or existing tests cover these changes
- [x] The documentation is up to date with these changes
- [x] `CHANGELOG.md` has been updated (if user-facing change)

## Test plan

```bash
python -m unittest newton.tests.test_import_usd.TestImportSampleAssetsBasic.test_disabled_static_collider_has_no_collision_flags -v
uvx pre-commit run -a
```

## Bug fix

**Steps to reproduce:**

1. Import a disabled USD collider with `load_visual_shapes=False`.
2. Observe that `COLLIDE_PARTICLES` remains set.

**Minimal reproduction:**

```python
import newton
from pxr import Usd, UsdGeom, UsdPhysics

stage = Usd.Stage.CreateInMemory()
cube = UsdGeom.Cube.Define(stage, "/Cube")
UsdPhysics.CollisionAPI.Apply(cube.GetPrim()).CreateCollisionEnabledAttr(False)
builder = newton.ModelBuilder()
shape = builder.add_usd(stage, load_visual_shapes=False)["path_shape_map"]["/Cube"]
print(builder.shape_flags[shape] & newton.ShapeFlags.COLLIDE_PARTICLES)  # 4, expected 0
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Disabled USD colliders no longer participate in shape or particle collisions.
  * Corrected collision behavior when visual shape loading is disabled.

* **Tests**
  * Added coverage verifying disabled guide colliders consistently have collision disabled.

PR #4036, opened 2026-08-25
Retry auto-assign when fork authors are not yet assignable at PR open

dsmither-lgtm:
## Description

Fixes a gap where the auto-assign workflow ran successfully but left fork PRs unassigned when GitHub\'s issue-specific assignability check returned `404` within seconds of `opened` (observed on #4007).

Changes:

- **Retry with backoff** on `opened`, `reopened`, and `ready_for_review` (4 attempts, 45s apart)
- **Re-attempt on `synchronize`** while the PR still has no assignees (1 attempt per push)
- **Do not hard-skip on preflight 404** — log it, still call `addAssignees`, and verify the response (post-flight check retained from #3106 / #3708)
- **Increase job timeout** to 5 minutes to accommodate retries
- **Comment on PR** (once) if all attempts fail on open events, so maintainers see the skip without digging into Actions logs

Closes the scenario where PR #4007 sat unassigned for ~34 hours until a maintainer assigned manually, despite the workflow logging success.

## Checklist

- [x] New or existing tests cover these changes — N/A (CI workflow only)
- [x] The documentation is up to date with these changes — N/A
- [ ] `CHANGELOG.md` has been updated (if user-facing change) — N/A (maintainer/CI-only change)

## Test plan

- [ ] Open a fork PR without an assignee → author should be assigned within the retry window or on first push
- [ ] Open a PR with an assignee already set → workflow no-ops
- [ ] Verify #4007-style case: fork contributor PR where preflight returns 404 at T+3s eventually gets assigned without manual intervention
- [ ] Confirm failed assignment posts a single bot comment with `<!-- auto-assign-pr-creator -->` marker

## Bug fix

**Steps to reproduce:**

1. Open a fork PR as an external contributor without selecting an assignee.
2. Observe the auto-assign workflow warn that the author is not assignable within seconds of `opened` and leave the PR unassigned (see #4007, [run 32622300473](https://github.com/newton-physics/newton/actions/runs/32622300473)).

**Expected after this PR:**

Assignment succeeds after retry/backoff or on the next `synchronize` while still unassigned.

## Summary by CodeRabbit

* **Improvements**
  * Pull request creator assignment is now more reliable, with automatic retries and verification.
  * Assignment processing also runs when a pull request is updated.
  * Temporary assignment lookup issues no longer prevent processing.
  * Clear success and failure status updates are provided.
  * Maintainers receive a single notification when assignment cannot be completed.

### DIFF
--- a/.github/workflows/pr_auto_assign_creator.yml
+++ b/.github/workflows/pr_auto_assign_creator.yml
@@ -6,6 +6,7 @@ on:
       - opened
       - reopened
       - ready_for_review
+      - synchronize
 
 permissions:
   pull-requests: write
@@ -16,7 +17,7 @@ jobs:
       github.repository == 'newton-physics/newton' &&
       github.event.pull_request.draft == false
     runs-on: ubuntu-latest
-    timeout-minutes: 3
+    timeout-minutes: 5
     steps:
       - name: Harden the runner (Audit all outbound calls)
         uses: step-security/harden-runner@fa2e9d605c4eeb9fcad4c99c224cee0c6c7f3594  # v2.16.0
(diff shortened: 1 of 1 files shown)


jcarius-nv:
Regarding "assignability check returned 404 within seconds of opened" . Is this a sign of a connectivity problem that is actually solved by re-trying, or is it more of a "did not find an author who can be assigned" due to permission errors or something like that, which would not be solved by re-trying? I think before we go ahead with re-tries it would be worth doing a manual re-try on a PR that has the problem and seeing if that succeeds.

jcarius-nv:
I checked this more broadly, and the evidence does not support retries as a useful solution here.

Across the last 300 Auto-assign workflow invocations (July 31 through August 26), 226 jobs executed and 55 of them emitted the exact `Author ... is not assignable ...; skipping` warning, covering 53 unique PRs. This is therefore a recurring condition rather than an isolated connectivity failure. GitHub documents the issue-specific `404` as meaning that the user cannot be assigned in the context of that request; it is not a network-error response.

The historical timing also argues against four retries at 45-second intervals:

- #3785 returned the same 404 on two separate events about 7 minutes 43 seconds apart.
- #3766 returned it again about 18 hours 35 minutes later.
- #3916 failed initially and only succeeded under the same workflow identity on a later `ready_for_review` event roughly a week afterward.
- On #4007 and this PR, the later successful assignments were self-assignments by the authors, so they do not show that the Actions token would have succeeded on retry.

Read-only checks also indicate that eligibility is caller/context-sensitive: affected PR authors can return `204` with a maintainer credential while returning `404` without that context. The best-supported conclusion is that this is GitHub assignment-eligibility behavior, possibly with inconsistent propagation around PR events, rather than ordinary connectivity. A direct assignment attempt plus response verification would be more meaningful than the preflight, but retries cannot reliably solve the observed cases.

Given that occasional missed auto-assignments are acceptable, I do not think the added complexity is worthwhile. Let us drop this PR without merging and accept that creator assignment remains best-effort.

PR #4037, opened 2026-08-26
Avoid pytorch imports during CI benchmark runs 

nvtw:
## Description

Fast benchmarks should not import pytorch since it is not needed and not available anymore.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compatibility for the fast DR Legs benchmark when PyTorch is unavailable.
  - Ensured the benchmark solver can initialize reliably with its configured simulation settings.

- **Tests**
  - Added regression coverage for solver creation without PyTorch.

### DIFF
--- a/asv/benchmarks/benchmark_kamino.py
+++ b/asv/benchmarks/benchmark_kamino.py
@@ -508,16 +508,28 @@ class DRLegsBenchmarkWorkload:
         return builder
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


mzamoramora-nvidia (approved):
Looks good to me. This removes the unnecessary PyTorch import from the fast DR Legs benchmark while preserving the existing solver configuration. Once CI is green, this should be ready to merge.

PR #4039, opened 2026-08-26
Experimental newton viewer extension support

nvtw:
## Description

Experimental newton viewer extension support

Allows to have viewers in other pip packages that can then be used directly inside newton if they are in the same venv. E. g. viewers that visualize more data through ImGui etc.

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

- **New Features**
  - Added support for third-party visualization viewer plugins.
  - Viewer plugins can be discovered and selected alongside built-in viewers.
  - Plugin factories are called without command-line arguments.
  - Conflicting or duplicate viewer names are safely ignored, allowing other valid viewers to remain available.

- **Documentation**
  - Added guidance for developing and registering third-party viewer plugins, including the required factory format.

### DIFF
--- a/docs/guide/visualization.rst
+++ b/docs/guide/visualization.rst
@@ -140,17 +140,18 @@ through the ``newton.viewers`` entry-point group. For example, its
    [project.entry-points."newton.viewers"]
    custom = "my_package.newton_viewer:create_viewer"
 
-The entry point must load a factory that accepts the parsed example arguments
-and returns a :class:`~newton.viewer.ViewerBase` instance:
+The entry point must load a zero-argument factory that returns a
+:class:`~newton.viewer.ViewerBase` instance:
 
 .. code-block:: python
 
-   def create_viewer(args):
-       return MyViewer(headless=args.headless, num_frames=args.num_frames)
+   def create_viewer():
+       return MyViewer()
 
 Newton reads installed entry-point metadata when constructing the example
 parser and imports only the selected viewer package. Third-party viewers cannot
-override built-in viewer names.
+override built-in viewer names. Conflicting entry points are ignored so they do
+not prevent other viewers from being used.
 
 Real-time Viewers
 -----------------
--- a/docs/guide/visualization.rst
+++ b/docs/guide/visualization.rst
@@ -128,6 +128,30 @@ All viewers support ``set_visible_worlds()`` to limit visualization to a subset 
     viewer.set_model(model)
     viewer.set_visible_worlds(range(4))
 
+Third-party viewers
+-------------------
+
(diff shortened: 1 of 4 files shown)


newton/examples/__init__.py
camevor, line 1: Similar to above, I'd prefer a more general place, e.g. under `newton._src.viewer`.

newton/examples/__init__.py
```diff
@@ -646,8 +661,8 @@ def create_parser():
         "--viewer",
         type=str,
         default="gl",
-        choices=["gl", "usd", "rtx", "rerun", "null", "viser"],
-        help="Viewer to use (gl, usd, rtx, rerun, null, or viser).",
+        choices=[*_BUILTIN_VIEWER_NAMES, *sorted(_get_viewer_entry_points())],
```
camevor: nit: a more fault-tolerant implementation would avoid breaking existing viewers in case of a misconfigured 3rd party package.

docs/guide/visualization.rst
```diff
@@ -128,6 +128,30 @@ All viewers support ``set_visible_worlds()`` to limit visualization to a subset
...
+through the ``newton.viewers`` entry-point group. For example, its
+``pyproject.toml`` can contain:
+
+.. code-block:: toml
+
+   [project.entry-points."newton.viewers"]
+   custom = "my_package.newton_viewer:create_viewer"
+
+The entry point must load a factory that accepts the parsed example arguments
+and returns a :class:`~newton.viewer.ViewerBase` instance:
```
camevor: The entry-point factory receives the full `newton.examples argparse.Namespace`, making the examples CLI’s field names part of the third-party viewer contract. This introduces an unnecessarily tight coupling of the 3rd party viewer to the example class.

camevor (commented):
Thanks @nvtw, I think this would be a very useful addition!

The only thing I'd revisit is the coupling to Newton's `Example` class, which does not seem necessary.

eric-heiden:
I think this feature is too hacky - if users want to use their own Viewer class, they can just import it in their runner code. I don't think we should support the example runner to be able to load arbitrary viewers when it is already quite straightforward for the user to register a new viewer in `examples/__init__.py`.

nvtw:
> I think this feature is too hacky - if users want to use their own Viewer class, they can just import it in their runner code. I don't think we should support the example runner to be able to load arbitrary viewers when it is already quite straightforward for the user to register a new viewer in `examples/__init__.py`.

Ok, closed the MR. The downside of registering a new viewer directly inside __init__.py if you don't plan to contribute it to newton is that you need to maintain your own fork and update it all the time. Doable of course but annoying.

PR #4041, opened 2026-08-26
Use current ALM normal multiplier for friction projection

jumyungc:
## Description

This is a small follow-up cleanup to the compliant-ALM update introduced in #3769 for the same release :
use the updated normal multiplier, `lambda_n+`, to define the Coulomb radius during the compliant-ALM contact dual update.

  After the normal update,

  `lambda_n+ = [s * lambda_n + k_eff * R_n]+`,

  the tangential multiplier is now projected onto the disk with radius:

  `mu * lambda_n+`.

Previously, the friction projection used a normal load computed from the updated normal multiplier and the contact residual. It now uses the updated normal multiplier directly.

No separate changelog entry is needed because this refines an unreleased feature already covered by #3769.

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rigid-contact friction handling by applying the Coulomb limit to the updated effective normal contact force.
  * Corrected tangential contact behavior for material-scaled contact parameters.

* **Tests**
  * Updated contact solver validation to reflect the corrected friction limit.


gdaviet (approved):
Thanks @jumyungc , in my opinion this is cleaner but  do you observe any practical difference?

jumyungc:
> Thanks @jumyungc , in my opinion this is cleaner but do you observe any practical difference?

Thank you for the feedback. In the current code/formulation, this actually works slightly better overall (though very slight).

PR #4043, opened 2026-08-26
Add GRU actuator drive

vidurv-nvidia:
## Description

Add the experimental public `newton.actuators.DriveNeuralGRU` drive for stateful GRU actuator inference from local ONNX checkpoints.

The implementation uses Warp and Warp-NN only; PyTorch is not required. Warp-NN 0.3.1 exposes `GRUCell` but its generic ONNX runtime does not yet execute ONNX `GRU` nodes, so this drive reads the embedded standard ONNX GRU parameters and maps them into public Warp-NN `GRUCell`, `Linear`, and optional `Tanh` modules.

Checkpoint metadata selects and orders any non-empty subset of these inputs:

- `position`
- `position_error`
- `velocity`
- `target_velocity`
- `velocity_error`
- optional `dynamic_bias`

The optional bias input is solver-neutral caller-owned Control data at `control.dynamic_bias`. When checkpoint metadata selects it, the GRU drive tells `ModelBuilder.add_actuator()` to register the `CONTROL`/`JOINT_DOF` custom attribute. `Actuator.step()` assigns the same-named array to `DriveNeuralGRU.dynamic_bias` before each evaluation. Applications overwrite the Control array with the current value before every evaluation; `Control.clear()` does not clear custom Control attributes. The drive consumes it as a network feature and never treats it as feedforward effort or adds it to predicted torque. Because it represents current per-evaluation data, it does not pass through `Delay`; target-derived features retain the existing delay behavior.

A private generic `DriveBase`/`ModelBuilder` configuration hook allows a drive to declare required model attributes. `Actuator.step()` generically binds same-named Control inputs declared by a drive. No public signatures change.

The drive preserves recurrent state through the existing double-buffered `Actuator.State` lifecycle, supports vectorized actuator groups and USD construction through `newton:modelPath`, composes with existing delays and clamping, and supports CUDA graph capture.

## Checklist

## Test plan

- `uv run --extra dev -m unittest newton.tests.test_actuators.TestDriveNeuralGRU` — 20 tests pass.
- Focused GRU, MLP, LSTM, actuator-step, and builder regression suites — 59 tests pass.
- GRU implementation coverage — 92%.
- `uvx pre-commit run -a` — passes.
- `uv run python docs/generate_api.py` — passes with generated API files current.
- Pinned Towncrier draft rendering — passes.
- Verified the generalized-feature regression tests fail against the previous implementation.

## New feature / API change

```python
import newton
from newton.actuators import DriveNeuralGRU

builder.add_actuator(
    DriveNeuralGRU,
    index=dof_index,
    model_path="motor.onnx",
)
model = builder.finalize()

state = model.state()
control = model.control()
actuator = model.actuators[0]
actuator_state_a = actuator.state()
actuator_state_b = actuator.state()

control.clear(model)
control.joint_target_q.assign(target_positions)
control.joint_target_qd.assign(target_velocities)

# Present when dynamic_bias is selected by the checkpoint.
control.dynamic_bias.assign(generalized_bias_force)

actuator.step(
    state,
    control,
    actuator_state_a,
    actuator_state_b,
    dt=sample_dt_s,
)
actuator_state_a, actuator_state_b = actuator_state_b, actuator_state_a
```


docs/concepts/actuators.rst
```diff
@@ -225,6 +225,104 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
 into a computation graph that no longer exposes it.
 
+.. _gru-actuator-support:
+
+SISO GRU
+^^^^^^^^
+
+:class:`ControllerNeuralGRU` is a deliberately narrow runtime for explicit,
+stateful, single-joint-input/single-joint-output actuator models. It accepts a
+dynamic-batch ``.pt2`` archive saved with ``torch.export.save`` or a TorchScript
```
jvonmuralt: Newton plans to remove its PyTorch dependency. Could you investigate implementing this controller through ONNX/Warp-NN instead of adding another Torch-specific controller? If Warp-NN currently lacks GRU support, it may be preferable to add the required operator there. Please document any technical blocker if ONNX is not feasible.

docs/concepts/actuators.rst
```diff
@@ -225,6 +225,104 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
+``qfrc_bias = c(q, v)``: the Coriolis, centrifugal, and gravitational forces
+for each joint DOF. It excludes passive, actuator, user-applied, and
+constraint forces. Checkpoint producers are responsible for using consistent
+semantics when collecting this feature. See the `MuJoCo equations of motion
+<https://mujoco.readthedocs.io/en/3.7.0/computation/index.html>`_.
+
+The base input columns are ``position``, ``position_error``, and ``velocity``.
+A checkpoint may additionally list ``dynamic_bias`` in ``input_columns``. When
+it does, the caller supplies this generalized bias force through
+``state.mujoco.qfrc_bias``, a one-dimensional ``wp.float32`` array on the
```
jvonmuralt: Could we make this input solver-neutral and avoid permanently extending State with mujoco:qfrc_bias for one controller? Generalized bias force is not inherently MuJoCo-specific, and some checkpoints do not request it.

docs/concepts/actuators.rst
```diff
@@ -225,6 +225,104 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
+
+For this full-torque controller, target velocity and feedforward
+``control.joint_act`` are not consumed. The generalized bias force conditions
+the network input and is not added to the predicted torque.
+
+.. code-block:: python
+
+   control.clear(model)
+   control.joint_target_q.assign(target_positions)
+   state.mujoco.qfrc_bias.assign(computed_generalized_bias_force)
```
jvonmuralt: Newton’s State represents simulation-owned quantities that advance with the solver and participate in state swapping. In contrast, dynamic_bias is a caller-supplied, per-evaluation controller input that must be recomputed for the current state. Storing it on State creates ambiguous ownership

newton/_src/actuators/actuator.py
```diff
@@ -317,6 +318,8 @@ def step(
 
         # --- 2+3. Effort mode: compute raw effort and clamp ---
         ctrl_state = current_act_state.controller_state if current_act_state else None
+        if isinstance(self.controller, ControllerNeuralGRU) and self.controller._uses_bias_force:
```
jvonmuralt: This breaks the controller abstraction and requires modifying Actuator for every new feature. Could this be replaced with a generic controller-input contract?

docs/concepts/actuators.rst
```diff
@@ -225,6 +225,104 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
+   actuator.step(state, control, actuator_state_a, actuator_state_b, dt=sample_dt_s)
```
jvonmuralt, line 290: please add a swapping line since it is a stateful controller to show the simulation loop example fully actuator_state_a, actuator_state_b = actuator_state_b, actuator_state_a

newton/_src/actuators/controllers/controller_neural_gru.py
```diff
@@ -0,0 +1,305 @@
...
+        self._bound_bias_force: wp.array[float] | None = None
+        self._next_hidden: torch.Tensor | None = None
+
+    def finalize(self, device: wp.Device, num_actuators: int) -> None:
+        import torch
+
+        self._device = device
+        self._num_actuators = num_actuators
+        self._torch_device = torch.device(f"cuda:{device.ordinal}" if device.is_cuda else "cpu")
+        self.network = self.network.to(self._torch_device)
```
jvonmuralt: After moving a TorchScript GRU to CUDA, its recurrent weights remain non-contiguous. Local execution warns that they may be repacked on every inference. Could finalize() compact/flatten the GRU parameters after the device transfer?

jvonmuralt (commented):
Thanks! 
I have two main design questions.

First, Newton plans to remove its PyTorch dependency. Could this controller use ONNX and Warp-NN instead? If Warp-NN does not currently support the required GRU operation, could we add that support there? 

Second, could dynamic_bias use a more generic controller input mechanism? This value is supplied by the caller for each evaluation. It is not simulation state advanced by the solver. Storing it as state.mujoco.qfrc_bias ties a generic actuator to MuJoCo, makes ownership unclear.

Newton already supports model-specific custom attributes on Control using Model.AttributeAssignment.CONTROL. Perhaps controllers could declare their required control inputs through a generic hook. ModelBuilder.add_actuator() could then register the required custom attributes only for models that need them, and Actuator.step() could bind those inputs without checking specifically for ControllerNeuralGRU.

For example, the checkpoint could request a solver-neutral joint-DOF input such as control.actuator.dynamic_bias. This should remain separate from joint_act, since generalized bias force is a network feature rather than feedforward effort.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 <!-- towncrier release notes start -->
 
+### Added
+
+- Add experimental `ControllerNeuralGRU` support for Anchor-compatible schema-v3 SISO `.pt` and `.pt2` Torch checkpoints, with caller-populated `state.dynamic_bias`, stateful explicit inference, and full-torque output.
+
```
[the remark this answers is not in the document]
jvonmuralt: please address this changelog comment

docs/concepts/actuators.rst
```diff
@@ -351,6 +450,8 @@ Controllers
 * :class:`ControllerPD` — proportional-derivative control law (stateless).
 * :class:`ControllerPID` — proportional-integral-derivative control law
   (stateful: integral accumulator with anti-windup clamp).
+* :class:`ControllerNeuralGRU` — SISO GRU neural-network
```
jvonmuralt: I do not think we need to emphasize here and in a couple of more places that it is SISO, all newton actuators are SISO at the moment and it is documented in the common limitations.

newton/_src/actuators/drives/drive_neural_gru.py
```diff
@@ -0,0 +1,305 @@
...
+            hidden=torch.zeros(
+                self._num_layers,
+                num_actuators,
+                self._hidden_size,
+                dtype=torch.float32,
+                device=self._torch_device,
+            )
+        )
+
+    def compute(
```
jvonmuralt, line 672: Controller.compute() already provides target velocity and its indices. Could input selection be generalized to support metadata-selected target_velocity and/or velocity_error, rather than accepting only the two exact hard-coded feature tuples? This would avoid introducing a public controller whose checkpoint contract is unnecessarily tied to one training setup.

docs/concepts/actuators.rst
vidurv-nvidia: Addressed: the GRU drive is now ONNX-only and runs entirely with Warp and Warp-NN; PyTorch is not used. Warp-NN 0.3.1 exposes GRUCell but its generic ONNX runtime does not execute ONNX GRU nodes, so Newton maps the embedded standard ONNX parameters into the public Warp-NN GRUCell, Linear, and optional Tanh modules.

newton/_src/actuators/controllers/controller_neural_gru.py
vidurv-nvidia: This is obsolete after the requested backend change: the GRU drive no longer loads or moves TorchScript modules. It is ONNX-only and executes with Warp-NN.

newton/_src/actuators/drives/drive_neural_gru.py
vidurv-nvidia: Addressed: input_columns can now select and order position, position_error, velocity, target_velocity, velocity_error, and optional dynamic_bias. Target velocity features use target_vel and target_vel_indices, including delayed target arrays.

docs/concepts/actuators.rst
vidurv-nvidia: Addressed: `dynamic_bias` is caller-owned Control data and Newton State is unchanged. The builder allocates the optional array only when requested, and the caller overwrites it with the current value before each evaluation. `Control.clear()` does not clear this custom attribute.

docs/concepts/actuators.rst
vidurv-nvidia: Addressed: the MuJoCo-specific State extension was removed. When selected by checkpoint metadata, `DriveNeuralGRU._configure_actuator()` registers a solver-neutral top-level `CONTROL`/`JOINT_DOF` attribute exposed as `control.dynamic_bias`. Checkpoints that do not select it allocate nothing.

docs/concepts/actuators.rst
vidurv-nvidia: Added the actuator-state buffer swap immediately after actuator.step in the GRU simulation-loop example.

docs/concepts/actuators.rst
```diff
@@ -225,6 +225,94 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
+The metadata must describe a schema-v3 ``gru`` with ``siso`` training
+topology, the exact four input columns above, positive ``sample_dt_s``, and
+normalization statistics for the four inputs and torque output. Descriptive
+Anchor fields not consumed by Newton are ignored. ``.pt2`` metadata also
+records ``num_layers`` and ``hidden_size``; TorchScript exposes those values
+through its ``gru`` submodule. The runtime actuator timestep must match
+``sample_dt_s``. Callers may compose a separate :class:`Delay` when additional
+runtime delay is desired. It delays target position, target velocity, and
+``joint_act``/``dynamic_bias`` together, on top of the dynamics learned by the
+network.
```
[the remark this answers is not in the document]
vidurv-nvidia: Addressed: the direct CHANGELOG edit is gone. The public addition is recorded in changelog/+drive-neural-gru-8e4c1a7d.added.md, and the pinned Towncrier draft succeeds.

docs/concepts/actuators.rst
vidurv-nvidia: Addressed: removed the GRU-specific SISO wording and the term from the changed repository content. The common limitation now simply states that actuators support one DOF per actuator.

newton/_src/actuators/actuator.py
vidurv-nvidia: Addressed with a generic private contract: `ModelBuilder.add_actuator()` invokes the drive configuration hook to register required attributes, and `Actuator.step()` copies same-named Control inputs declared by the drive before evaluation. There is no GRU import or type check in either class, and no public signature changes.

docs/concepts/actuators.rst
vidurv-nvidia: The application no longer needs to declare this attribute manually. The drive configuration hook registers `control.dynamic_bias` only when its metadata-selected inputs require it; the caller remains responsible for computing and populating the current value.

newton/_src/actuators/actuator.py
vidurv-nvidia: The final implementation keeps the integration generic and small: `DriveBase` provides a no-op `_configure_actuator()` and an empty `_control_input_attrs`; `DriveNeuralGRU` overrides/selects them for `dynamic_bias`. Other drives are unaffected.

docs/concepts/actuators.rst
vidurv-nvidia: The runtime path is now separate from feedforward: `Actuator.step()` copies the same-named `control.dynamic_bias` array onto `DriveNeuralGRU.dynamic_bias` before evaluation, and the GRU reads that member as conditioning only. `joint_act` and `control_feedforward_attr` retain their existing semantics.

newton/_src/actuators/actuator.py
vidurv-nvidia: The GRU now receives `dynamic_bias` directly rather than repurposing feedforward. The builder change is only the generic configuration-hook call, and the Actuator change is only the generic same-name Control-to-drive assignment before computation.

vidurv-nvidia:
Small update on the final input plumbing: when `dynamic_bias` is selected, the GRU drive registers `control.dynamic_bias` through the generic builder configuration hook. Before each evaluation, `Actuator.step()` copies that same-named Control array onto the drive, where it is consumed only as a network feature. This keeps State solver-owned and leaves `joint_act`/feedforward semantics unchanged; no public signatures changed.

newton/_src/actuators/drives/drive_neural_gru.py
```diff
@@ -0,0 +1,784 @@
...
+        """
+        self.model_path = os.fspath(model_path)
+        """Local path to the ONNX checkpoint."""
+        self._description = _load_network_description(self.model_path)
+        metadata = load_metadata(self.model_path)
+
+        normalization = metadata["normalization"]
+        input_normalization = normalization["inputs"]
+        self._input_feature_keys = _parse_input_feature_keys(metadata, self.model_path)
+        self.custom_control_attributes = ("dynamic_bias",) if "dynamic_bias" in self._input_feature_keys else ()
```
jvonmuralt, line 530: The name is hard-coded in three places — here, _INPUT_FEATURE_CODES, and the compute() guard. Could the checkpoint declare which of its input_columns the caller supplies, rather than the drive knowing the name? Something like a custom_inputs metadata key, with the drive deriving:

self.custom_state_attributes = tuple(n for n in self._input_feature_keys if n in caller_supplied)

Then a checkpoint trained against qfrc_bias, or against some other measured quantity, works without touching the drive, and the name stops being baked into the feature map.

newton/_src/actuators/drives/drive_neural_gru.py
```diff
@@ -0,0 +1,784 @@
...
+from itertools import pairwise
+from typing import Any, ClassVar
+
+import numpy as np
+import warp as wp
+
+from ..utils import _require_onnx, load_metadata
+from .base import DriveBase
+
+_FEATURE_POSITION = 0
```
jvonmuralt, line 18: The velocity row has raw state, raw target and error; the position row is missing target_position.

docs/concepts/actuators.rst
```diff
@@ -282,6 +282,109 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
 submodule is inspected directly, while ``torch.export`` flattens the network
 into a computation graph that no longer exposes it.
 
+.. _gru-actuator-support:
```
jvonmuralt, line 285: The docs are the other way round from the rest of the package. DriveNeuralLSTM and DriveNeuralMLP document their checkpoint contract in the class docstring — accepted formats, required metadata keys, required ONNX input and output names — and get a shared paragraph in the concept page. DriveNeuralGRU has a 6-line docstring and a 101-line concept-page section.
Could the GRU-specific parts move into the class docstring — the metadata keys, the feature names, the ONNX graph requirements, the normalization formulas? That is where someone using the class will look, and it matches LSTM.
 Separately: the custom-input mechanism is documented under the GRU heading, but custom_state_attributes / custom_control_attributes are DriveBase features available to any drive. Anyone writing a different drive is unlikely to find it there. It would sit better as its own short section next to Stateful Actuators, written without reference to the GRU.
 I would also drop the "Deferred capabilities" table.

docs/concepts/actuators.rst
```diff
@@ -282,6 +282,109 @@ may omit them: they contain the original module, whose ``torch.nn.LSTM``
...
+
+.. code-block:: python
+
+   control.clear(model)
+   control.joint_target_q.assign(target_positions)
+   control.dynamic_bias.assign(computed_generalized_bias_force)
+   actuator.step(state, control, actuator_state_a, actuator_state_b, dt=sample_dt_s)
+   actuator_state_a, actuator_state_b = actuator_state_b, actuator_state_a
+
+For models
```
jvonmuralt, line 340: the PR never mentions that Newton computes the same quantity itself.
newton.eval_inverse_dynamics_passive(model, state, gravity_force=..., coriolis_force=...) returns g(q) and C(q,q̇)q̇ as two (model.joint_dof_count,) arrays — exactly the layout the drive gathers with its velocity indices — and their sum is this feature. newton.selection.ArticulationView has the same call for the view-based workflow.
Could the docs point at that instead of at MuJoCo? As written, a user has to reach into a solver's internals for something the library already provides publicly. It would also let the "Deferred: automatic generalized-bias-force computation" row say what to call rather than only that it is deferred.

please see https://github.com/vidurv-nvidia/newton/pull/2 for suggested docs improvement

newton/_src/actuators/actuator.py
```diff
@@ -461,6 +461,11 @@ def step(
         positions = getattr(sim_state, self.state_pos_attr)
         velocities = getattr(sim_state, self.state_vel_attr)
 
+        for attr_name in self.drive.custom_state_attributes:
```
jvonmuralt, line 464: Per-step data should not be bound onto the drive. Actuator.step() binds these with setattr onto the drive, and DriveNeuralGRU.compute() then reads the attribute and immediately sets it back to None. That round trip suggests the value wants to be a compute() argument rather than drive state.

newton/_src/actuators/drives/base.py
```diff
@@ -63,6 +63,24 @@ def resolve_arguments(cls, args: dict[str, Any]) -> dict[str, Any]:
         """
         raise NotImplementedError(f"{cls.__name__} must implement resolve_arguments")
 
+    @classmethod
+    def _configure_actuator(cls, builder: Any, args: dict[str, Any]) -> None:
```
jvonmuralt, line 67: A few things about _configure_actuator:
1. It is declared on the public DriveBase, so a third-party drive has to override an underscore method. Could it be public? 
2. builder: Any - can this be typed?
3. It only works for Newton users. Could the docstring say that, and say what a user of another engine does instead to supply the custom array?
4. Newton already has a convention for this: SolverBase.register_custom_attributes(cls, builder) (solvers/solver.py:637), overridden by MuJoCo, Kamino, VBD, Style3D and MPM, and called once per builder. _configure_actuator runs from inside add_actuator(), so twenty actuators re-read the ONNX checkpoint twenty times. Could it follow the existing convention instead: a public method, generic in DriveBase (register custom_attributes), called once? That would also drop the DriveNeuralGRU override and the duplicate checkpoint parse. 

On the state/control split: attributes registered on State exist for solvers to read and write, and State is double-buffered, so a caller-supplied array registered there would have to be written into both buffers each step. I would drop the distinction between custom_state_attributes and custom_control_attributes, always register on CONTROL even where the quantity is semantically closer to state, and document that choice.

 I would also rather see this machinery on Actuator than on DriveBase. Actuator already owns the drive, the delay and the clamping components further components will be added soon), all three are constructed together in finalize() , and a clamping component could need a custom array too. Collecting the declarations at the actuator means one registration path for all of them rather than one per component type.

jvonmuralt (changes requested):

PR #4045, opened 2026-08-26
Expose the Gaussian splat BVH for refit

daniela-hase:
## Description

The `Gaussian` splat asset owns a Warp BVH built during `finalize()`, but there was no public way to refit it after the finalized Gaussian data changes. This adds a clean accessor
and an in-place refit method that mirror the scene shape BVH API (`Model.bvh_shapes` / `Model.bvh_refit_shapes`):

- `Gaussian.bvh` — read-only property exposing the finalized `wp.Bvh` (or `None` before `finalize()`), dropping the implementation-flavored `warp_` prefix the scene attributes also omit.
- `Gaussian.bvh_refit()` — recomputes per-Gaussian bounds from the finalized GPU `Data` and refits the BVH in place, keeping its existing topology. Guards with a `RuntimeError` if `finalize()` has
not been called, matching the scene refit methods.

Like the scene refit, this preserves tree topology and only updates bounds, so it suits moved/rescaled Gaussians; a change in Gaussian *count* still requires a full rebuild via `finalize()`.

## Checklist

## Test plan

Verified manually that the pre-finalize guard raises, `Gaussian.bvh` returns the finalized BVH, and bounds update in place after mutating the finalized data:

```python
import numpy as np, warp as wp, newton

g = newton.Gaussian(
    positions=np.array([[0, 0, 0], [1, 0, 0]], dtype=np.float32),
    rotations=np.array([[0, 0, 0, 1], [0, 0, 0, 1]], dtype=np.float32),
    scales=np.ones((2, 3), dtype=np.float32),
    opacities=np.ones(2, dtype=np.float32),
    sh_coeffs=np.ones((2, 3), dtype=np.float32),
)
g.finalize(device="cpu")
before = g.bvh.lowers.numpy().copy()

tr = g.warp_data.transforms.numpy(); tr[0, 0] = 5.0
g.warp_data.transforms = wp.array(tr, dtype=wp.transformf, device="cpu")
g.bvh_refit()
assert not np.allclose(before, g.bvh.lowers.numpy())

Existing Gaussian tests pass:

uv run --extra dev -m newton.tests -k gaussian

New feature / API change

import newton

gaussian = newton.Gaussian.create_from_ply("object.ply")
gaussian.finalize(device="cuda")

# ... mutate gaussian.warp_data.transforms / scales on the device ...

gaussian.bvh_refit()   # refit the BVH in place
bvh = gaussian.bvh     # access the finalized Warp BVH
```

## Summary by CodeRabbit

* **New Features**
  * Exposed the Gaussian bounding volume hierarchy (BVH) for inspection.
  * Added support for refitting the existing BVH after finalized Gaussian data changes.
  * BVH refitting now reports an error when attempted before Gaussian data is finalized.

* **Deprecations**
  * Deprecated legacy Gaussian data and BVH aliases. Use the finalized Gaussian data and `Gaussian.bvh` instead. Compatibility access remains available with deprecation warnings.

### DIFF
--- a/newton/_src/geometry/types.py
+++ b/newton/_src/geometry/types.py
@@ -2481,6 +2481,12 @@ class Gaussian:
         min_response: wp.float32
         sorting_mode: wp.int32
 
+    _WARP_DATA_DEPRECATION_MSG = (
+        "Gaussian.warp_data is deprecated in Newton 1.6; use the Gaussian.Data object returned by "
+        "Gaussian.finalize() instead."
+    )
+    _WARP_BVH_DEPRECATION_MSG = "Gaussian.warp_bvh is deprecated in Newton 1.6; use Gaussian.bvh instead."
+
     def __init__(
         self,
         positions: np.ndarray,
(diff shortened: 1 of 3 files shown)


moennen (dismissed):
Thanks !!!

newton/_src/geometry/types.py
```diff
@@ -2611,6 +2611,16 @@ def sorting_mode(self) -> SortingMode:
...
 
+    @property
+    def bvh(self) -> wp.Bvh | None:
+        """The finalized Warp BVH over the Gaussians, or ``None`` before :meth:`finalize`.
+
+        Mirrors the scene shape BVH exposed as :attr:`~newton.Model.bvh_shapes`.
+        Use :meth:`bvh_refit` to update it in place after the finalized
+        :class:`Data` arrays change.
+        """
+        return self.warp_bvh
```
eric-heiden: `Gaussian.warp_bvh` is already public - can we make this private instead and only allow this `bvh` property as getter? Otherwise, this property isn't needed at all.

daniela-hase: Good point, yeah! Done that :)

newton/_src/geometry/types.py
```diff
@@ -2554,8 +2554,8 @@ def __init__(
         self._sh_coeffs.setflags(write=False)
 
         # GPU arrays populated by finalize()
-        self.warp_bvh: wp.Bvh = None
-        self.warp_data: Gaussian.Data = None
+        self._warp_bvh: wp.Bvh = None
+        self._warp_data: Gaussian.Data = None
```
eric-heiden, line 2564: It's good you made these now private attributes. But we still need a deprecation behavior here so you need to implement `@property` attributes for these previous `warp_bvh` and `warp_data` names that raise a deprecation warning when the user tries to access them. Be sure to also include the `.. deprecated:: 1.6` in the docstring of these properties.

eric-heiden: I guess for `warp_data` we don't need a public accessor? Just asking in case this is an oversight. 🙂

daniela-hase: All done! And yeah, `warp_data` shouldn't need a public accessor :)

eric-heiden (approved):
Thanks!

PR #4056, opened 2026-08-27
Import Newton curve damping from USD

### LINKED ISSUE #3980: [REQ] Import the newton-usd-schemas curve damping attributes
### Description

`newton-usd-schemas` 0.5.0 added [`NewtonCurvesDeformableMaterialAPI`](https://github.com/newton-physics/newton-usd-schemas/blob/3d737cea0f29a39732e35a9d62f1a9f5c453b92b/newton_usd_schemas/generatedSchema.usda#L1291-L1380) with four per-mode damping attributes for curve deformables:

- `newton:curvesStretchDamping`
- `newton:curvesShearDamping`
- `newton:curvesBendDamping`
- `newton:curvesTwistDamping`

Each defaults to `-inf`, the schema's "use the engine's default" sentinel, and each damps the mode whose stiffness is the corresponding `physics:curves*Stiffness`. The schema deliberately gives them no shared derivation: the AOUSD proposal defines no damping modulus, so an unauthored damping falls back to the engine default rather than to another mode's damping.

Newton pins `newton-usd-schemas>=0.5.0`, so the schema is registered, but the importer never reads these four attributes. Authoring them has no effect and produces no diagnostic.

The cable material reader gates on `PhysicsCurvesDeformableMaterialAPI` and requests a fixed list of attribute names that contains only the `physics:` stiffness/thickness/modulus set, density, and the deprecated unprefixed names ([`_get_curve_deformable_material`](https://github.com/newton-physics/newton/blob/6d3fdf6f7885378677d9b69899aad1ee5bd6c667/newton/_src/usd/utils.py#L2184-L2213)). Names are resolved through [`read_deformable_attr`](https://github.com/newton-physics/newton/blob/6d3fdf6f7885378677d9b69899aad1ee5bd6c667/newton/_src/usd/schema_resolver.py#L302-L305), which tries `physics:` and then the resolvers' deformable compatibility namespaces — empty by default, and [`["omniphysics", "physxDeformableBody"]`](https://github.com/newton-physics/newton/blob/6d3fdf6f7885378677d9b69899aad1ee5bd6c667/newton/_src/usd/schemas.py#L315) under the PhysX resolver. `newton` is in neither list, so there is no path by which an authored `newton:curves*Damping` reaches the importer.

Authoring all four on a bound material that applies both APIs, then importing on `main`:

```
authored: [newton:curvesBendDamping, newton:curvesShearDamping,
           newton:curvesStretchDamping, newton:curvesTwistDamping]
material parsed by the importer: {curvesThickness, curvesStretchStiffness, curvesBendStiffness}
warnings mentioning damping: []
joint 0: ke=[2000.0, 1087.5, 40.0, 0.060]
joint 0: kd=[0.0, 0.0, 0.0, 0.0]
```

The values are dropped silently and every mode keeps `kd = 0.0`.

The model-side plumbing is already in place. [`add_joint_cable`](https://github.com/newton-physics/newton/blob/6d3fdf6f7885378677d9b69899aad1ee5bd6c667/newton/_src/sim/builder.py#L5209-L5237) accepts `stretch_damping`, `shear_damping`, `bend_damping` and `twist_damping`, and [`_set_joint_cable_stiffnesses`](https://github.com/newton-physics/newton/blob/6d3fdf6f7885378677d9b69899aad1ee5bd6c667/newton/_src/sim/builder.py#L5335-L5361) already reads `joint_target_kd[dof]` to infer each DOF's target mode — it just never writes it. What is missing is the USD read plus the same per-joint normalization the stiffnesses get.

### Motivation / Use Case

Cable damping is currently only reachable from Python, so a USD-authored cable cannot express it: every imported cable simulates undamped in all four modes regardless of what the material says. Damping is what makes an imported cable settle rather than ring, so this is a visible behavioral gap for anyone driving Newton from a stage rather than from `ModelBuilder`.

It is also the point of the schema. These attributes exist only to let a stage carry Newton's damping alongside the AOUSD stiffnesses, and shipping a registered schema whose attributes are silently ignored is worse than not having it: authoring them looks supported and produces no warning.

### Acceptance criteria

- A stage authoring the four `newton:curves*Damping` attributes on a material applying both `PhysicsCurvesDeformableMaterialAPI` and `NewtonCurvesDeformableMaterialAPI` imports with the corresponding `joint_target_kd` set on each cable joint DOF.
- The `-inf` sentinel is honored as "engine default" and leaves the builder default in place, consistent with the other `-inf` sentinels the deformable material reader already drops.
- Each damping is normalized per joint the same way its stiffness is, by the dual rest length `0.5 * (L_parent + L_child)`, so unevenly sampled curves stay accurate.
- An unauthored mode falls back to the engine default and not to another mode's damping, per the schema's documented intent. Note this differs from `add_joint_cable`'s Python-level convenience, where shear can inherit from stretch and twist from bend; that inheritance should not be extended to the authored path.
- Out-of-range values are diagnosed rather than dropped silently, matching the existing validation in the deformable material reader (the schema declares a soft minimum of 0).
- Automated coverage for both single cables and welded rod graphs, since the two import paths apply material stiffnesses through separate call sites.

### Alternatives Considered

Fold this into #3809. That issue is scoped to realigning with the updated AOUSD proposal — Young's modulus, Poisson ratio, the stiffness units change, and the attribute renames from OpenUSD-proposals#111. These damping attributes are a Newton extension in the `newton:` namespace rather than part of that proposal, and the two can land independently, so tracking them separately keeps #3809's scope honest.

Leave damping Python-only. That keeps the schema's attributes permanently inert, which is the current state and the problem.

For reference, the sibling case from the same 0.5.0 batch was tracked separately and closed the same way: #3861 covered importing the authored MPM schemas. #3912, which bumped the pin to 0.5.0, recorded this gap explicitly at the time ("the schemas added since 0.4.1 ... are not read by Newton yet") but no issue was opened for it.

jumyungc:
## Description
Import per-mode cable damping authored through
`NewtonCurvesDeformableMaterialAPI`.

This PR:

- imports stretch, shear, bend, and twist structural damping;
- converts each value to a per-joint gain using
  `joint_target_kd = structural_damping / (0.5 * (L_parent + L_child))`;
- applies damping consistently to ordinary cables and welded rod graphs;
- preserves independent modes, unauthored defaults, and the `-inf` sentinel;
- warns for invalid values and damping authored on curve geometry;
- keeps returned cable material metadata solver-neutral; and
- documents the matching stiffness and damping conversion semantics.

Newton damping is used only when the material also applies
`PhysicsCurvesDeformableMaterialAPI`, consistent with the existing curve
stiffness family gating.

Closes #3980

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **New Features**
  * Cable imports now support per-mode structural damping alongside stiffness.
  * Imported cable properties are converted into solver-ready material gains using each joint’s rest shape.
  * Zero damping and fallback values are handled during import.

* **Bug Fixes**
  * Improved validation and warnings for invalid or geometry-authored damping attributes.
  * Target behavior now correctly reflects resolved stiffness and damping settings.

* **Documentation**
  * Updated USD parsing documentation to describe damping conversion, units, fallbacks, validation, and deprecated attributes.


newton/_src/utils/import_usd_deformable_cable.py
```diff
@@ -754,6 +795,7 @@ def _deformable_import_cable(ctx: _DeformableImportContext, consumed_cable_curve
         _warn_unsupported_rest_fields(prim, path, ("restNormals",), deformable_read)
         _warn_dropped_velocities(prim, path)
         _warn_geometry_authored_material_attrs(prim, path, "PhysicsCurvesDeformableMaterialAPI", deformable_read)
+        _warn_geometry_authored_newton_curve_damping_attrs(prim, path)
```
mzamoramora-nvidia, line 813: Could we also run this check for curves consumed by the welded-graph pass? Those curves never reach this line, so damping authored on their geometry is ignored without a warning. A welded-curve test would help cover this path.

newton/_src/sim/builder.py
```diff
@@ -3632,7 +3632,7 @@ def add_usd(
...
                   - Mapping from prim path (str) of a soft body (a volume deformable, or a legacy bare TetMesh) to its ``[start, end)`` index ranges, keyed ``"particle"`` / ``"tet"``. Present only with ``return_deformable_results=True``.
                 * - ``"path_cable_attrs"``
-                  - Mapping from prim path (str) of a curve deformable (cable) to its as-authored, solver-neutral attributes (``material`` moduli, ``resolved_density``, ``closed``); includes moduli the imported rod cannot express (e.g. shear / twist). ``graph_component`` is present only for curves successfully welded into the same rod graph; curves in one graph share the component identifier. Present only with ``return_deformable_results=True``.
+                  - Mapping from prim path (str) of a curve deformable (cable) to its validated, solver-neutral cable import metadata (``material``, ``resolved_density``, ``closed``). ``material`` contains supported per-mode stiffness and damping values. ``graph_component`` is present only for curves successfully welded into the same rod graph; curves in one graph share the identifier. Present only with ``return_deformable_results=True``.
```
mzamoramora-nvidia: Could we include the units for the stiffness and damping values here? These are physical quantities in a public docstring, and the coding guidelines ask us to state their SI units.

newton/_src/usd/utils.py
```diff
@@ -2128,7 +2128,13 @@ def _read_physics_attr(prim: Usd.Prim, name: str, compat_namespaces: Sequence[st
 
 
 def _read_deformable_material(
-    prim: Usd.Prim, read_attr: Callable[[Usd.Prim, str], Any], api_schema: str, attr_names: Sequence[str]
+    prim: Usd.Prim,
+    read_attr: Callable[[Usd.Prim, str], Any],
+    api_schema: str,
+    attr_names: Sequence[str],
+    *,
+    attr_namespace: str = "physics",
```
mzamoramora-nvidia, line 2136: Could we add Google-style `Args:` sections here and to the new cable helpers? This would keep the new docstrings consistent with the repository guidelines.

newton/tests/test_import_usd_deformable_cable.py
```diff
@@ -501,16 +562,18 @@ def test_non_linear_curve_is_skipped(self):
         self.assertEqual(builder.body_count, 0)
 
     def test_cable_material_without_family_api_is_ignored(self):
-        """A physics-bound material lacking PhysicsCurvesDeformableMaterialAPI is not read as a cable material."""
+        """A family-less material remains ignored, including its Newton damping."""
```
mzamoramora-nvidia: Could we make this summary imperative? For example: `Verify that a family-less material is ignored, including its Newton damping.`

mzamoramora-nvidia:
Thanks for this clear and well-tested change. The core damping import looks solid, and I did not find any blocking behavior. I left a few small comments about the welded-graph warning path and documentation details.

newton/_src/utils/import_usd_deformable_cable.py
jumyungc: Done.

newton/_src/sim/builder.py
jumyungc: Done.

newton/_src/usd/utils.py
jumyungc: Done.

newton/tests/test_import_usd_deformable_cable.py
jumyungc: Done.

jumyungc:
> Thanks for this clear and well-tested change. The core damping import looks solid, and I did not find any blocking behavior. I left a few small comments about the welded-graph warning path and documentation details.

Thank you for the thorough review. I’ve applied all the suggestions

mzamoramora-nvidia (approved):

PR #4058, opened 2026-08-27
Support rigid-body/particle attachments

### LINKED ISSUE #3815: Support rigid-body/particle attachments
### Requirement

Add a first-class attachment constraint between a rigid body (including a VBD cable capsule) and a deformable particle or node in VBD cloth and solids. The attachment must transfer forces bidirectionally so coupled workflows do not require contact as a workaround.

### Acceptance criteria

- Expose a supported public builder API for attaching a VBD cloth or solid particle/node to an anchor on a rigid body.
- Apply equal-and-opposite constraint forces to the deformable and rigid body.
- Add CPU/CUDA regressions covering rigid-body attachments to both cloth and a VBD solid under load.
- Document supported solvers and current limitations.

Related: #1296 covered cable-to-rigid attachments, not rigid-to-particle attachments.

jvonmuralt:
## Description
Question about units in ADMM. 
VBD uses `stiffness` as a displacement penalty, so it is in N/m.
ADMM passes the same number as `kappa` into a velocity-level energy.
The ADMM local solve then divides by this sum:

    denom = kappa + damping + rho * W^2

`kappa` is N/m, `damping` is N·s/m, and `rho * W^2` is a mass in kg.
Written in base units, that sum is:

    kg/s^2  +  kg/s  +  kg

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **New Features**
  * Added compliant attachments between rigid bodies and deformable particles.
  * Attachments support stiffness, damping, enabled state, and body-local anchor points.
  * Supported by VBD and coupled ADMM workflows, including cable capsules and cross-entry coupling.
  * Added validation for endpoints, worlds, and externally integrated rigid bodies.

* **Deprecation**
  * Deprecated the solver-level attachment helper in favor of model-level attachment setup.

* **Documentation**
  * Added solver support details, configuration guidance, compatibility notes, and migration information.


gdaviet:
@jvonmuralt thanks for reporting the unit inconsistency in the ADMM parameters, I can try to address that in a follow-up. The solver is still marked as experimental so we should be able to do that without a full deprecation cycle

newton/_src/solvers/vbd/rigid_vbd_kernels.py
```diff
@@ -1639,6 +1639,68 @@ def evaluate_linear_constraint_force_hessian(
...
+    if damping > 0.0:
+        inv_dt = 1.0 / dt
+        force_body = force_body + damping * (C - C_prev) * inv_dt
+        hessian = hessian + (damping * inv_dt) * wp.identity(3, float)
+
+    return -force_body, hessian
+
+
+@wp.func
+def evaluate_body_particle_attachment_force_hessian(
```
AnkaChan, line 1670: maybe you can keep this one and remove that `evaluate_body_particle_attachment_particle_force_hessian`?

newton/_src/solvers/vbd/rigid_vbd_kernels.py
```diff
@@ -5353,6 +5415,73 @@ def compute_rigid_contact_forces(
...
+            attachment_damping[attachment],
+            dt,
+        )
+        force_acc += force_body
+        torque_acc += torque_body
+        h_ll_acc += h_ll
+        h_al_acc += h_al
+        h_aa_acc += h_aa
+
+    wp.atomic_add(body_forces, body_id, force_acc)
```
AnkaChan, line 5478: You launch one thread per body right? I dont think those atomic addes are needed.

AnkaChan (commented):
Looks overall good. Only some minor issues.

AnkaChan (commented):
Additional review at `dc0b20a579933b5275c0bbe86de691e34936f353`: two coupled-path behavior issues and two optional cleanups. The ADMM comment adds a quantitative reproduction of the units issue already acknowledged for follow-up. The units and enabled-state reproductions were rechecked on CPU at this head; the earlier review also exercised CPU/CUDA attachment paths. Existing inline comments are not duplicated.

newton/_src/solvers/coupled/solver_coupled_admm.py
```diff
@@ -2733,59 +2757,136 @@ def _build_admm_joint_groups(self, coupling: SolverCoupledADMM.Config) -> None:
...
+        if int(model.attachment_body_particle_count) > 0:
+            sources.append(
+                (
+                    "row",
+                    int(model.attachment_body_particle_count),
+                    model.attachment_body_particle_body,
+                    model.attachment_body_particle_particle,
+                    model.attachment_body_particle_body_point,
+                    model.attachment_body_particle_stiffness,
+                    model.attachment_body_particle_damping,
```
AnkaChan, line 2775: **[P1] Preserve the physical coefficient contract when native rows reach ADMM**

Following up on the units issue already noted in the description and the proposed experimental-solver follow-up: forwarding these new N/m and N·s/m fields unchanged makes the same model attachment have a different force law depending on whether its endpoints share a VBD entry.

I traced the timestep factors through the local solve, effective masses, target and force adapter. `W²` is an effective mass, and the proximal path adds `rho*W²` to mass. A consistently mass-scaled velocity formulation uses `k*h²` and `c*h` locally, with the exchanged impulse divided by `h` before entering the force buffers. The current adapter copies the splat directly as force, so changing only the local denominator would not be sufficient.

Concrete reproduction: two 1 kg endpoints, initial gap 0.1 m, zero initial velocities, no gravity or contacts, `k=1 N/m`, `c=0.4 N*s/m`; two SemiImplicit entries with ADMM `rho=1`, `gamma=0`, `baumgarte=1`, and 600 iterations:

| dt | Native VBD relative velocity | Physical backward-Euler prediction | ADMM relative velocity |
|---|---:|---:|---:|
| 0.1 s | 0.01818181 m/s | 0.01818182 m/s | 0.15624999 m/s |
| 0.01 s | 0.00198354 m/s | 0.00198373 m/s | 0.19455251 m/s |

The ADMM residual is below 6e-7 m/s, so this is not lack of convergence. With `C` the initial gap and `u=v_body-v_particle`, the current converged force is numerically `k*C/h - (k+c)*u`; the physical law is `k*C - (k*h+c)*u`. CPU and CUDA reproduced this, and a CPU recheck at this head gave the same results.

The exact merge-base reproduces the same ADMM numbers through the legacy helper, so this is an inherited solver issue newly exposed by the native coefficient contract. Please either convert the native coefficients and force/impulse convention consistently, or explicitly document the different cross-entry coefficient semantics while that follow-up is pending.

Relevant existing paths: [local solve](https://github.com/newton-physics/newton/blob/dc0b20a579933b5275c0bbe86de691e34936f353/newton/_src/solvers/coupled/admm_utils.py#L737-L752), [target and force splat](https://github.com/newton-physics/newton/blob/dc0b20a579933b5275c0bbe86de691e34936f353/newton/_src/solvers/coupled/admm_utils.py#L935-L971), [direct force adapter](https://github.com/newton-physics/newton/blob/dc0b20a579933b5275c0bbe86de691e34936f353/newton/_src/solvers/coupled/solver_coupled.py#L2638-L2651).

newton/_src/solvers/coupled/solver_coupled.py
```diff
@@ -1369,6 +1388,72 @@ def _compact_mimic_constraint_order(self, joint_global_to_local: dict[int, int])
...
+            rows,
+            model.attachment_body_particle_count,
+        )
+        view.attachment_body_particle_count = len(rows)
+        self._project_compact_attributes(
+            view,
+            projections,
+            exclude=set(),
+            include=lambda attribute: attribute.frequency == attachment_frequency,
+            source_model=True,
```
AnkaChan, line 1454: **[P2] Refresh attachment enabled state in coupled views and ADMM groups**

These projected attachment arrays retain their construction-time values after the parent model changes. For example, after constructing a coupler around a VBD entry containing both endpoints:

```python
model.attachment_body_particle_enabled.assign([False])
solver.notify_model_changed(newton.ModelFlags.ALL)
solver.step(state_in, state_out, control, None, 0.1)
```

The attachment still applies forces. Starting with a disabled row and enabling it after construction also does nothing. Standalone VBD correctly handles both toggles.

With two 1 kg endpoints, a 0.1 m gap, `k=1`, zero damping/initial velocity/gravity, and 30 VBD iterations, disabling before the first step gives zero relative velocity in standalone VBD but 0.01960784 m/s through the coupled VBD view; the view still reports `enabled=True`. This occurs before any warm-start history is involved. Cross-entry ADMM has the same lifecycle problem because it excludes disabled rows when groups are built.

Please include attachment properties in model-change projection and refresh or mask the ADMM groups accordingly. The current flag matcher does not select this attribute frequency even for `ModelFlags.ALL`. Cover both enabled transitions; stiffness/damping updates also need a defined refresh path. This does not require supporting arbitrary endpoint/topology edits at runtime.

Related paths: [model-change flag matching](https://github.com/newton-physics/newton/blob/dc0b20a579933b5275c0bbe86de691e34936f353/newton/_src/solvers/coupled/solver_coupled.py#L3045-L3077), [ADMM construction-time filtering](https://github.com/newton-physics/newton/blob/dc0b20a579933b5275c0bbe86de691e34936f353/newton/_src/solvers/coupled/solver_coupled_admm.py#L2816-L2821).

newton/_src/solvers/vbd/solver_vbd.py
```diff
@@ -1884,6 +1918,30 @@ def _compute_rigid_force_element_adjacency(self, model):
 
         return adjacency
 
+    def _compute_body_particle_attachment_adjacency(self, model: Model) -> tuple[wp.array, wp.array]:
```
AnkaChan, line 1921: **Optional: reuse the shared triangle/edge adjacency builder**

Could this reuse [`_build_vertex_adjacency_with_warp`](https://github.com/newton-physics/newton/blob/dc0b20a579933b5275c0bbe86de691e34936f353/newton/_src/utils/mesh.py#L995-L1025), which already handles counting, prefix sums, allocation and filling for triangle/edge adjacency?

It can take the body-index array on CPU, `body_count`, `values_per_entry=1`, and two small attachment-specific count/fill kernels. The existing triangle/edge kernels themselves are specialized: they visit several vertices and emit `(element_id, local_slot)` pairs; here each attachment visits one body and emits only its attachment ID. The shared driver returns NumPy `(indices, offsets)`, which this wrapper can upload to the solver device.

Renaming its node-count parameter/helper to describe generic element adjacency would make that reuse clearer. Preserve stable attachment-ID order, include disabled rows, and retain `body_count+1` zero offsets for empty input. This is a maintainability suggestion; I have not benchmarked or implemented the adaptation.

newton/_src/sim/builder.py
```diff
@@ -8791,6 +8832,81 @@ def add_particles(
...
+                coefficient is negative.
+        """
+        if body < 0 or body >= self.body_count:
+            raise IndexError(f"Body index {body} is out of range for {self.body_count} bodies")
+        if particle < 0 or particle >= self.particle_count:
+            raise IndexError(f"Particle index {particle} is out of range for {self.particle_count} particles")
+        if stiffness < 0.0:
+            raise ValueError("Attachment stiffness must be nonnegative")
+        if damping < 0.0:
+            raise ValueError("Attachment damping must be nonnegative")
```
AnkaChan, line 8882: **Optional: reject nonfinite attachment parameters at authoring time**

The negative-value checks accept NaN because `nan < 0` is false. I verified that `stiffness=float("nan")`, `stiffness=float("inf")`, `damping=float("nan")`, and `body_point=(float("nan"), 0, 0)` are accepted and stored. A nonfinite stiffness or anchor then reaches the force/Hessian arithmetic without a useful authoring error.

Consider checking that stiffness, damping and each body-point component are finite, in addition to the nonnegative coefficient checks, and adding validation tests. This is optional input hardening rather than a claim about ordinary finite-parameter scenes.

newton/_src/solvers/vbd/rigid_vbd_kernels.py
jvonmuralt: Thanks, I think having the particle-side kernel call the combined function would mean it computes torque and the angular Hessian blocks on every attachment evaluation and throws them away. So I did the other way around.
I removed evaluate_body_particle_attachment_force_hessian and kept evaluate_body_particle_attachment_particle_force_hessian instead. 
This mirrors how contacts already reuse a shared force-law function across particle and body kernels

Also noticed the torque/Hessian-rotation math (r × force, skew(r) * H, skew(r)ᵀ * H * skew(r)) was duplicated in three places: evaluate_linear_constraint_force_hessian, the attachment body-side kernel, and the contact body-side kernel,  so I factored it into a shared _point_force_to_body_torque_and_hessian helper and switched all three call sites to it.

newton/_src/solvers/coupled/solver_coupled.py
jvonmuralt: This is being addressed in #4183, which normalizes the ADMM stiffness/damping/friction coefficients at the solver boundary

newton/_src/solvers/coupled/solver_coupled_admm.py
jvonmuralt: This is being addressed in https://github.com/newton-physics/newton/pull/4183, which normalizes the ADMM stiffness/damping/friction coefficients at the solver boundary

PR #4063, opened 2026-08-27
Accelerate collision detection

nvtw:
## Description

Accelerate collision detection across the broad phase, narrow phase, and deterministic contact-processing pipeline.

- Select bounded tile sorting for medium SAP worlds and ordinary radix sorting for single global segments, while bypassing unnecessary identity-map and world-filter work.
- Reduce deterministic and persistent contact-sort radix passes by packing only the shape-index and contact-family bits required by each model.
- Accelerate large convex support queries with cooked directional seeds and adjacency walks, while retaining exhaustive support mapping as a safe fallback.
- Reduce mesh-convex and mesh-SDF work through wider query tiles, cached convex centers, specialized reducer modes, leaner pair-local state, one-warp split-convex launches, and exact sphere-triangle contacts.
- Cull unreachable heightfield cells and omit unused explicit mesh-contact stages.
- Add nightly benchmarks and regression coverage for the optimized paths.

## Checklist

## Test plan

- CI test suite passed.
- Full repository checks passed:
  ```
  uvx pre-commit run -a
  ```
- Regression coverage includes SAP sort selection and global-segment filtering, deterministic and persistent contact-key packing, explicit mesh-stage sizing, convex support acceleration and fallback, heightfield cell culling, mesh-SDF routing, split-convex launch sizing, and sphere contacts on coplanar triangle seams.

## Bug fix

The optimized paths also fix duplicate reduced contacts for spheres resting on coplanar mesh seams and preserve the correct deterministic reducer mode for specialized mesh-SDF kernels. The added regression tests reproduce these cases without the fixes and verify the corrected contact counts and sort behavior.

## Summary by CodeRabbit

* **Performance**
  * Improved collision-detection speed through adaptive broad-phase sorting and accelerated convex-shape processing.
  * Reduced memory usage for explicitly configured collision pairs.
  * Improved mesh, SDF, and heightfield filtering efficiency.
  * Optimized deterministic contact sorting and GPU processing.
* **Bug Fixes**
  * Eliminated duplicate sphere contacts along coplanar triangle-mesh seams.
  * Improved CUDA graph compatibility for single-world collision processing.


newton/_src/sim/model.py
```diff
@@ -851,6 +852,15 @@ def __init__(self, device: Devicelike | None = None):
...
         self.shape_source_ptr: wp.array[wp.uint64] | None = None
         """Geometry source pointers to be used inside the Warp kernels which are generated by finalizing the geometry objects, see for example :meth:`newton.Mesh.finalize`, shape [shape_count], uint64."""
+
+        self.shape_support_data: wp.array[wp.vec4i] | None = None
+        """Per-shape compact convex support metadata: LUT, offset, neighbor starts, and resolution."""
+        self.convex_support_lut: wp.array[wp.int32] | None = None
+        """Packed octahedral directional seed tables shared by convex geometries."""
+        self.convex_support_vertex_offsets: wp.array[wp.int32] | None = None
+        """Packed CSR vertex-neighbor offsets shared by convex geometries."""
+        self.convex_support_neighbors: wp.array[wp.int32] | None = None
```
nvtw: I should probably _ prefix these new members to mark them private

newton/_src/sim/builder.py
```diff
@@ -137,6 +137,114 @@ def _deduplicate_convex_collision_mesh(source: Mesh) -> Mesh:
...
+    )
+    normal_lengths = np.linalg.norm(face_normals, axis=1)
+    nondegenerate = normal_lengths > geometry_scale * geometry_scale * 1.0e-12
+    face_normals = face_normals[nondegenerate]
+    if len(face_normals) == 0:
+        return None
+    face_normals /= normal_lengths[nondegenerate, None]
+    face_offsets = np.einsum("ij,ij->i", face_normals, triangle_points[nondegenerate, 0])
+
+    # The input indices are not automatically rebuilt as a convex hull. Only
```
mzamoramora-nvidia, line 166: Could we detect incomplete hull topology while cooking, warn, and disable acceleration for that shape so it uses the existing exhaustive path? The current checks show that the faces are supporting and connected, but not that no faces are missing. I reproduced a connected, incomplete mesh that passes cooking but returns a lower support value than exhaustive search (`0.001678` difference), which can change GJK/MPR results. A documented validation error would also be reasonable if closed topology is required. Either way, we should avoid the accelerated walk when completeness cannot be established and add a regression for this case.

newton/_src/geometry/contact_reduction_global.py
```diff
@@ -2394,9 +2436,45 @@ def mesh_triangle_contacts_to_reducer_kernel(
...
+                contact_data.margin_a = margin_offset_a
+                contact_data.margin_b = margin_offset_b
+                contact_data.shape_a = shape_a
+                contact_data.shape_b = shape_b
+                contact_data.gap_sum = gap_sum
+                contact_data.sort_sub_key = (tri_idx << 1) | 1
+                write_contact_to_reducer(contact_data, reducer_data, -1)
+            continue
+
+        data_provider = SupportMapDataProvider()
```
mzamoramora-nvidia: Is it intentional that reduced mesh-convex contacts still use exhaustive support? Since `reduce_contacts=True` is the default, the common path skips the new acceleration, and the current test only covers `False`. Could we pass the acceleration data into the reducer and test both modes?

asv/benchmarks/simulation/bench_contacts.py
```diff
@@ -167,6 +169,82 @@ def _build_convex_scene(
...
+            i2 = i0 + row
+            i3 = i2 + 1
+            triangles.extend(((i0, i1, i3), (i0, i3, i2), (i3, i1, i0), (i2, i3, i0)))
+    return newton.Mesh(vertices, np.asarray(triangles, dtype=np.int32), compute_inertia=False)
+
+
+def _build_mesh_convex_scene(world_count: int, resolution: int) -> tuple[newton.Model, int]:
+    """Build replicated dense mesh-convex contacts without external assets."""
+    grid = _make_two_sided_grid(resolution, half_extent=1.0)
+    hull = newton.Mesh.create_sphere(
```
mzamoramora-nvidia, line 211: This 8x16 sphere is below the 256-vertex threshold, so this benchmark does not exercise the new support acceleration. Default contact reduction also skips that path. Could we use a larger hull and assert that acceleration is enabled?

newton/_src/geometry/narrow_phase.py
```diff
@@ -2609,6 +2729,10 @@ def launch_custom_write(
         shape_data: wp.array[wp.vec4],  # Shape data (scale xyz, margin w)
         shape_transform: wp.array[wp.transform],  # In world space
         shape_source: wp.array[wp.uint64],  # The index into the source array, type define by shape_types
+        shape_support_data: wp.array[wp.vec4i] | None = None,
```
mzamoramora-nvidia, line 2732: Could we update the `Args:` docs for the new parameters? `shape_support_data`, `support_lut`, `support_vertex_offsets`, and `support_neighbors` are missing here, and `data_provider` is missing from both updated `collision_core` docstrings. These inputs are not self-explanatory, so their purpose and expected data should be documented.

mzamoramora-nvidia (commented):
Some clean up comments below.

mzamoramora-nvidia (approved):

PR #4068, opened 2026-08-27
Add pipeline to export newton Controllers to a C++ runtime.

jeff-hough:
## Description

<!-- What does this PR change and why?
     Reference any issues closed by this PR with "Closes #1234". -->

This PR adds an APIC export pipeline for `newton.controller` Controllers, along with a C++ runtime class which can load the controller.

## Checklist

## Test plan

Currently, this is how the c++ controller runtime is built and tested. This is subject to how we decide to do this, given this is the first example of the python --> C++ pipeline through APIC.

```
  uv sync
  cmake -S newton/_src/controllers/cpp -B build/controllers                                                                                                                                                                                                                           
  cmake --build build/controllers                                                                                                                                                                                                                                                     
  ctest --test-dir build/controllers --output-on-failure  

```

## New feature / API change

Create the controller, then export it to a graph this way:

```python
from newton.controllers import ControllerJointImpedance, export_controller_graph

controller = ControllerJointImpedance(model, ...)
inputs, outputs = controller.input(), controller.output()

export_controller_graph(controller=controller, inputs=inputs, outputs=outputs, path="joint_impedance")
```

From c++, simply load the controller. The usage is intentionally very similar to how things look in python:
```cpp
#include <newton/controllers/controller.hpp>

newton::controllers::Controller controller{"joint_impedance.wrp"};
auto input = controller.input();
auto output = controller.output();

input["joint_q"] = {0.35f, -0.62f, 0.18f, 0.9f};
input["joint_q_des"] = {0.0f, -0.3f, 0.5f};

if (!controller.step(input, output, 0.01f)) {
    std::cerr << controller.last_error() << "\n";  // step never throws
    return 1;
}
// output["joint_f"] now holds one torque per controlled DOF
```

## Summary by CodeRabbit

* **New Features**
  * Added controller graph export to `.wrp` artifacts.
  * Added a standalone C++ runtime for replaying exported graphs on CUDA or CPU without Python.
  * Added named input, output, and timestep parameters, including indexed array view support.
* **Documentation**
  * Added API documentation and C++ build and usage instructions.
* **Tests**
  * Added coverage for export, replay, validation, parameters, CPU execution, and array views.


.github/workflows/aws_gpu_tests.yml
```diff
@@ -179,6 +179,14 @@ jobs:
           COVERAGE_CORE: "sysmon"
         run: uv run --extra dev --extra torch-cu13 -m newton.tests --strict-warnings --no-cache-clear --junit-report-xml rspec.xml --coverage --coverage-xml coverage.xml
 
+      - name: Build and run C++ controller runtime tests
+        run: |
+          uv tool install "cmake==4.4.2"
+          export PATH="$HOME/.local/bin:$PATH"
+          cmake -S newton/_src/controllers/cpp -B build/controllers
+          cmake --build build/controllers
+          ctest --test-dir build/controllers --output-on-failure
```
shi-eric: This looks fine to me! As a bonus, you should consider including a report in the existing job summary.

you can add the `--output-junit "$GITHUB_WORKSPACE/build/controllers/ctest.xml` option to the `ctest` command and change later in the workflow:

```
  with:
    paths: |
      rspec.xml
      build/controllers/ctest.xml
    show: "fail"
```

jeff-hough: Thank you for taking a look! I will add this in tomorrow.

jcarius-nv: I'd also be ok but some extra points to think about:

- Are we ok with just testing linux? Windows Cpp builds are always a major pain to set up due to discovering msvc versions, choosing the right compilers, etc. So if we can stay Linux-only, that'd be great
- For the Linux job, is it guaranteed that we have the necessary tools in the runner image (gcc and whatever libs you need)? From the few lines you added here it looks like so, but worth an extra check. The less setup/installation we need the better, of course!

newton/_src/controllers/cpp/cmake/FindWarp.cmake
jcarius-nv, line 1: Do I understand it correctly that this cpp folder is not only used for our tests but also by users of the controllers as helper files for deploying them? That would actually extend our test surface a bit because we currently don't validate that this compiles on all kinds of platforms, compilers, cpp versions, etc.

I'm a bit hesitant to introducing a larger cpp test matrix, but not doing that and shipping broken code is clearly worse.

newton/_src/controllers/cpp/README.md
```diff
@@ -0,0 +1,93 @@
+# C++ controller runtime
```
jcarius-nv: How do you expect to deliver this? Should users get it directly out of the cloned repository? Or extract from the wheel (which would feels a bit strange to me)?

Further down you're suggesting users to build their own cpp app with `#include <newton/controllers/controller.hpp>`. For convenience people usually expect in cmake to be able to use something like `find_package(newton-controllers)` and then the includes and linker settings are automatically handled for them. Have you thought about how this would work here?

.github/workflows/aws_gpu_tests.yml
jeff-hough: Personally, I think it would be OK to document that the C++ code here is more of an "example deployment pipeline", and not a fully supported cross-platform feature. So I would also be OK to leave this Linux only in CI.

shi-eric: Windows job: Can you make your files work for envs without a GPU? Then we can add new stages to the CPU-only GitHub jobs and get more platform coverage.

For Jan's other question, as part of evaluating Jeff's PR, I tested it on AWS on my fork, so everything should work.

jeff-hough: Yes, I had wanted to see if the export pipeline works on the CPU as well! I have adjusted that today, and tests are passing.

@shi-eric I added the tests to `ci.yml`, which runs all tests on all platforms (CUDA-based tests are guarded on CUDA availability).

Passes on Windows and on ARM :)

newton/_src/controllers/cpp/cmake/FindWarp.cmake
jeff-hough: I think the desired was to show this as an example deployment pipeline, not necessarily as something which is guaranteed to work for all compilers on all platforms.

newton/_src/controllers/cpp/src/controller.cpp
```diff
@@ -0,0 +1,468 @@
...
+// decltype(&::name): the real function's signature, copied from its
+// declaration in warp.h/apic.h, rather than retyped by hand. That means a
+// mismatch (wrong argument type, wrong order) fails to *compile* the call
+// site below, instead of compiling silently and only breaking at runtime --
+// resolve()'s reinterpret_cast trusts whatever type it's given, so a
+// hand-written function pointer type would give it nothing to check against.
+// The leading `::` forces each name to resolve to the free function in
+// warp.h/apic.h, not the struct member of the same name being declared on
+// that same line.
+struct WarpApi {
```
mmacklin, line 65: Hmm, this is a lot of machinery and a very general thing to have in a controllers header. Not sure I understand why you can't use warp.h directly, but think there is too much here that is not related to controllers at all and should either be shipped with Warp or as a general Newton APIC package.

newton/_src/controllers/cpp/src/controller.cpp
```diff
@@ -0,0 +1,468 @@
...
+    }
+    return static_cast<APICDeviceType>(header.device_type);
+}
+
+// Loads every compiled kernel a CPU-captured graph references, mirroring
+// Warp's own Python capture_load(): wp_load_obj() JITs each <path>_modules/*.o
+// into an in-memory module, then wp_lookup() resolves each kernel's forward /
+// backward symbol so wp_apic_register_loaded_cpu_kernel() can wire it into the
+// graph before replay.
+void load_cpu_kernels(const std::filesystem::path& wrp_path, APICGraph* graph) {
```
mmacklin, line 249: Feels like this should be something handled directly by Warp's APIC API as much as possible.

jeff-hough: Agreed - the loading of CPU vs GPU seems to be asymmetric. So I had to add some lines where I check what the device was in the graph metadata, and call the loading function based on whether it was CPU or GPU. I agree it would be much simpler if the Warp API handles this (maybe it already does and I just missed the function to use?)

newton/_src/controllers/cpp/src/controller.cpp
jeff-hough: @mmacklin it is only needed if we want windows support today. However, the proper fix is to have a `.lib` export in the warp `pypi` package alongside the `.dll` files which are currently there. This will allow windows to do linking "normally" and we would only require the header file in that case.

I agree it is not really the kind of code which should be included in the controllers module. It is probably better to just delay testing CI on windows.

newton/_src/controllers/cpp/cmake/FindWarp.cmake
```diff
@@ -0,0 +1,99 @@
...
+    add_library(warp::clang UNKNOWN IMPORTED GLOBAL)
+    set_target_properties(warp::clang PROPERTIES
+        IMPORTED_LOCATION "${_warp_clang_library}"
+        IMPORTED_NO_SONAME TRUE
+    )
+endif()
+
+set(WARP_ROOT "${_warp_root}" CACHE PATH "Root of the uv-installed warp package")
+set(WARP_LIBRARY "${_warp_library}" CACHE FILEPATH "Path to the Warp runtime shared library")
+set(WARP_CLANG_LIBRARY "${_warp_clang_library}" CACHE FILEPATH "Path to Warp's LLVM-JIT backend library")
```
jeff-hough, line 95: @mmacklin it would also be nicer if linking to warp through CMake was made a little easier. I have a feeling that most of this could be exported in a single `warp::warp` target, which is the more typical cmake linking pattern. I.e., linking to warp could be as simple as my doing this in the controller cmake:

```cmake
find_package(warp)

# ... make the controllers library ...

# link to warp in a single step here, which should automatically 
# pull in any .so, .dll, .a, .lib, .hpp files that we need
target_link_libraries(my_library SHARED warp::warp)

```

PR #4072, opened 2026-08-28
Add model-independent actuator parameter view

### LINKED ISSUE #4061: [REQ] Decouple actuator access from Newton Model
## Description

Newton actuators are intended to work independently of the physics backend, including with PhysX. However, actuator selection, mapping, and parameter access currently depend on the model-bound `newton.selection.ArticulationView`. This means TensorAPI-based workflows must instantiate a Newton `Model` even when Newton is not the simulation backend.

Introduce a model-independent actuator view or selection interface that can operate through an appropriate TensorAPI view without requiring a Newton `Model`. Newton's `ArticulationView` can remain model-bound and should retain its existing convenient actuator access, potentially by delegating to this interface.

The design should avoid introducing a parallel actuator-only selection implementation. Selection and mapping primitives should be shared with, or extracted from, `ArticulationView`, while its existing model-backed convenience methods remain available through composition or delegation. The model-independent layer should remain generic and avoid controller- or actuator-specific assumptions, such as filtering to one-DOF joints; consumers should validate their own constraints. The design should also clarify how TensorAPI layouts and heterogeneous fleets are represented.

## Motivation / Use Case

Preserve actuators as a physics-engine-independent feature while avoiding unnecessary Newton model construction in non-Newton simulation workflows.

The key outcome is that using Newton actuators with another physics backend does not require constructing a Newton model.

## Alternatives Considered

Continue requiring callers to construct a Newton `Model` and `ArticulationView` solely for actuator access. This unnecessarily couples otherwise standalone actuators to Newton's model representation.

## Request for input

Please add the motivating TensorAPI workflow, which view or views should be supported, and any expected API or compatibility constraints.

AntoineRichard:
## Description

Add `newton.actuators.ActuatorView`, a model-independent view over a list of actuators and selected global velocity-DOF indices. It derives and caches each actuator's parameter mapping, reads and writes component parameters by string name or component object, supports per-world masks, and exposes the mapping for downstream ownership queries.

Add cached `ArticulationView.get_actuator_view()` construction and retain the existing `ArticulationView.get_actuator_parameter()` / `set_actuator_parameter()` signatures by delegating through the standalone view. This lets PhysX and other non-Newton backends use Newton actuators without constructing a Newton `Model`.

Closes #4061.

## Checklist

- [x] For user-facing changes, a fragment has been added by following the changelog fragment instructions

## Test plan

```console
uv run --extra dev -m unittest newton.tests.test_actuators.TestActuatorSelectionAPI newton.tests.test_actuator_view
uvx pre-commit run -a
uvx --from towncrier==25.8.0 towncrier build --draft --version X.Y.Z --date 2026-08-28
```

The focused actuator suite ran 15 tests successfully. The full Newton suite ran 6,553 tests successfully with 172 skips.

## New feature / API change

```python
import newton

view = newton.actuators.ActuatorView([actuator], selected_dof_indices)
kp = view.get_actuator_parameter(actuator, "controller", "kp")
view.set_actuator_parameter(actuator, "controller", "kp", kp, mask=world_mask)

# Existing model-backed views provide the same standalone type.
view = articulation_view.get_actuator_view([actuator])
```

## Summary by CodeRabbit

* **New Features**
  * Added `ActuatorView` for model-independent access to mapped actuator parameters.
  * Added support for retrieving actuator views from articulation selections.
  * Actuator parameters can be read and updated through mapped views, including optional per-world masking.

* **Bug Fixes**
  * Added validation for invalid actuator selections, shapes, devices, masks, and duplicate degree-of-freedom mappings.

* **Documentation**
  * Added `ActuatorView` to the public actuator API documentation.
  * Added a changelog entry describing the new functionality.

### DIFF
--- a/newton/_src/actuators/actuator.py
+++ b/newton/_src/actuators/actuator.py
@@ -17,6 +17,30 @@ from .effort_mode_implicit import ImplicitOptions, ResponseOracle, _EffortModeIm
 
 
 @wp.kernel
+def _gather_parameter_kernel(src: Any, mapping: wp.array2d[int], dst: Any):
+    """Gather actuator parameters into a view-shaped array."""
+    world, dof = wp.tid()
+    index = mapping[world, dof]
+    if index >= 0:
+        dst[world, dof] = src[index]
+
+
+@wp.kernel
+def _scatter_parameter_kernel(
+    values: Any,
+    mapping: wp.array2d[int],
+    mask: wp.array[bool],
+    dst: Any,
+):
+    """Scatter view-shaped actuator parameters for selected worlds."""
+    world, dof = wp.tid()
+    if mask[world]:
+        index = mapping[world, dof]
+        if index >= 0:
+            dst[index] = values[world, dof]
+
+
+@wp.kernel
 def _scatter_add_kernel(
     forces: wp.array[float],
     computed_forces: wp.array[float],
--- a/newton/_src/actuators/actuator.py
+++ b/newton/_src/actuators/actuator.py
@@ -17,30 +17,6 @@ from .effort_mode_implicit import ImplicitOptions, ResponseOracle, _EffortModeIm
 
 
 @wp.kernel
-def _gather_parameter_kernel(src: Any, mapping: wp.array2d[int], dst: Any):
-    """Gather actuator parameters into a view-shaped array."""
-    world, dof = wp.tid()
-    index = mapping[world, dof]
-    if index >= 0:
-        dst[world, dof] = src[index]
-
-
-@wp.kernel
-def _scatter_parameter_kernel(
-    values: Any,
-    mapping: wp.array2d[int],
-    mask: wp.array[bool],
-    dst: Any,
-):
-    """Scatter view-shaped actuator parameters for selected worlds."""
-    world, dof = wp.tid()
-    if mask[world]:
-        index = mapping[world, dof]
-        if index >= 0:
-            dst[index] = values[world, dof]
-
-
-@wp.kernel
 def _scatter_add_kernel(
     forces: wp.array[float],
     computed_forces: wp.array[float],
--- a/newton/_src/actuators/view.py
+++ b/newton/_src/actuators/view.py
@@ -31,6 +31,18 @@ class ActuatorView:
         mapping = next(iter(self._mappings.values()))
         self._full_mask = wp.full(mapping.shape[0], True, dtype=bool, device=mapping.device)
 
+    def get_actuator_dof_mapping(self, actuator: Actuator) -> wp.array2d[int]:
+        """Get the view-local DOF mapping for an actuator.
+
+        Args:
+            actuator: Actuator whose mapping to return.
+
+        Returns:
+            Mapping from view DOFs to actuator parameter indices. Unmapped DOFs
+            contain ``-1``.
+        """
+        return self._mappings[actuator]
+
     @classmethod
     def from_articulation_view(cls, articulation_view: ArticulationView, actuators: list[Actuator]) -> ActuatorView:
         """Build a view from an existing articulation view.
--- a/newton/_src/actuators/view.py
+++ b/newton/_src/actuators/view.py
@@ -66,6 +66,8 @@ class ActuatorView:
             raise ValueError("Expected a two-dimensional integer dof_indices array")
         if dof_indices.shape[0] == 0:
             raise ValueError("Expected dof_indices for at least one world")
+        if dof_indices.shape[1] > 1 and any(len(set(indices)) != len(indices) for indices in dof_indices.numpy()):
+            raise ValueError("Expected dof_indices to be unique within each world")
 
         mappings = {}
         for actuator in actuators:
@@ -101,14 +103,22 @@ class ActuatorView:
         self._full_mask = wp.full(mapping.shape[0], True, dtype=bool, device=mapping.device)
 
     def get_actuator_dof_mapping(self, actuator: Actuator) -> wp.array2d[int]:
-        """Get the view-local DOF mapping for an actuator.
+        """Return the actuator parameter index for each selected velocity DOF.
+
+        The returned array has the same ``(world, DOF)`` layout as parameter
+        arrays read or written through this view. Each non-negative entry is
+        the index along the leading axis of the actuator's parameter arrays.
+        An entry of ``-1`` means that the actuator does not drive that DOF.
+
+        The mapping can therefore be used for actuator-ownership queries:
+        non-negative entries identify the view columns driven by *actuator*.
+        It is cached view metadata and must not be modified.
 
         Args:
             actuator: Actuator whose mapping to return.
 
         Returns:
-            Mapping from selected velocity-DOF columns to actuator parameter
-            indices. Unmapped columns contain ``-1``.
+            View-owned mapping shaped ``(world_count, dofs_per_world)``.
         """
         return self._mappings[actuator]
 
--- a/newton/_src/actuators/view.py
+++ b/newton/_src/actuators/view.py
@@ -1,0 +1,110 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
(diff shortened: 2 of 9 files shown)


newton/_src/actuators/view.py
```diff
@@ -0,0 +1,149 @@
...
+            values = wp.array(values, dtype=parameter.dtype, shape=expected_shape, device=mapping.device, copy=False)
+        if values.shape[:2] != expected_shape[:2]:
+            raise ValueError(f"Expected values shape {expected_shape}, got {values.shape}")
+
+        if mask is None:
+            mask = self._full_mask
+        elif not isinstance(mask, wp.array):
+            mask = wp.array(mask, dtype=bool, shape=(mapping.shape[0],), device=mapping.device, copy=False)
+        if mask.shape != (mapping.shape[0],):
+            raise ValueError(f"Expected mask shape ({mapping.shape[0]},), got {mask.shape}")
```
jvonmuralt: could you check values.device and mask.device against mapping.device before wp.launch

jvonmuralt (dismissed):
Thank you, looks good to me! Left a minor comment.

newton/_src/actuators/view.py
```diff
@@ -0,0 +1,149 @@
...
+class ActuatorView:
+    """Read and write actuator parameters through view-local DOF mappings.
+
+    This view is independent of :class:`~newton.Model`; callers provide the
+    mapping from each actuator to the view's ``(world, DOF)`` layout.
+
+    Args:
+        mappings: Mapping from each actuator to its actuator parameter indices.
+            Entries with value ``-1`` represent DOFs not driven by that actuator.
+            All arrays must have the same shape and device.
```
camevor: `mappings` is a somewhat opaque object that no public API can produce, so the public constructor's one parameter is something a user can only obtain by reaching for private functions or by reimplementing the layout logic from scratch.

We should provide a more straightforward constructor.

newton/_src/actuators/view.py
```diff
@@ -0,0 +1,149 @@
...
+    """
+
+    def __init__(self, mappings: dict[Actuator, wp.array2d[int]]) -> None:
+        self._mappings = dict(mappings)
+        mapping = next(iter(self._mappings.values()))
+        self._full_mask = wp.full(mapping.shape[0], True, dtype=bool, device=mapping.device)
+
+    def get_actuator_dof_mapping(self, actuator: Actuator) -> wp.array2d[int]:
```
camevor, line 105: Like above; as it stands, this object is opaque and it is not obvious why a user would be interested in it, other than – somewhat circularly – to construct an `ActuatorView` from an existing `ArticulationView`.

newton/_src/actuators/view.py
```diff
@@ -0,0 +1,149 @@
...
+            actuator: Actuator whose mapping to return.
+
+        Returns:
+            Mapping from view DOFs to actuator parameter indices. Unmapped DOFs
+            contain ``-1``.
+        """
+        return self._mappings[actuator]
+
+    @classmethod
+    def from_articulation_view(cls, articulation_view: ArticulationView, actuators: list[Actuator]) -> ActuatorView:
```
camevor: This method seems redundant.

newton/_src/utils/selection.py
```diff
@@ -1919,7 +1888,22 @@ def _get_actuator_dof_mapping(self, actuator: Actuator):
                 device=self.device,
             )
 
-        return mapping
+        return mapping.reshape((self.world_count, dofs_per_world))
+
+    def get_actuator_view(self, actuators: list[Actuator]) -> ActuatorView:
+        """Get a cached actuator view in this articulation view's DOF layout.
```
camevor: The term "DOF layout" is somewhat confusing here, as it refers to the dof vs coord layout convention elsewhere. However, positions (also position targets in the next release) follow the coord layout.

newton/_src/actuators/view.py
```diff
@@ -0,0 +1,149 @@
...
+        Args:
+            articulation_view: Articulation view that defines the desired DOF layout.
+            actuators: Actuators to expose through the new view.
+
+        Returns:
+            A standalone actuator view.
+        """
+        return articulation_view.get_actuator_view(actuators)
+
+    def get_actuator_parameter(self, actuator: Actuator, component_name: str, parameter_name: str) -> wp.array2d[Any]:
```
camevor: The signatures of `ActuatorView.get_actuator_parameter` and `ArticulationView.get_actuator_parameter` should ideally be identical, along with their implementation.

camevor (commented):
Thanks for factoring this, @AntoineRichard! Generally looks good to me, and the implementation sharing is an important aspect. As it stands, it reads as a middle way between composition and implementation sharing: `ArticulationView` constructs an `ActuatorView` and then calls into its `_get_parameter_array`/`_set_parameter_array`.. I recommend picking one: Either compose properly and delegate fully to `ActuatorView`'s public API, or share the implementation and avoid constructing an `ActuatorView` within `ArticulationView` calls.

newton/_src/actuators/actuator.py
```diff
@@ -16,6 +16,30 @@
 from .effort_mode_implicit import ImplicitOptions, ResponseOracle, _EffortModeImplicit
 
 
+@wp.kernel
+def _gather_parameter_kernel(src: Any, mapping: wp.array2d[int], dst: Any):
```
jvonmuralt: I think these kernels do not belong to this file. It would be better to move them to the view as it is the only place they are used.

newton/_src/actuators/view.py
AntoineRichard: The accessor remains intentionally public in 3eb6b378, but is no longer needed for construction. It exposes the derived ownership mapping for downstream integrations such as IsaacLab, which otherwise must duplicate Newton's mapping logic or use private selection internals.

newton/_src/actuators/actuator.py
AntoineRichard: Addressed in 3eb6b378: the gather/scatter kernels now live in actuators/view.py, their only consumer.

newton/_src/actuators/view.py
```diff
@@ -0,0 +1,149 @@
...
+    Args:
+        mappings: Mapping from each actuator to its actuator parameter indices.
+            Entries with value ``-1`` represent DOFs not driven by that actuator.
+            All arrays must have the same shape and device.
+    """
+
+    def __init__(self, mappings: dict[Actuator, wp.array2d[int]]) -> None:
+        self._mappings = dict(mappings)
+        mapping = next(iter(self._mappings.values()))
+        self._full_mask = wp.full(mapping.shape[0], True, dtype=bool, device=mapping.device)
```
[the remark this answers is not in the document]
AntoineRichard: Addressed in 3eb6b378: empty actuator sets now raise a clear ValueError before mapping initialization, including through ArticulationView.get_actuator_view([]).

newton/_src/actuators/view.py
AntoineRichard: Addressed in 3eb6b378: Warp values and masks are checked against mapping.device before the scatter launch. Parameter arrays are checked as well.

newton/_src/actuators/view.py
AntoineRichard: Addressed in 3eb6b378: the public constructor is now ActuatorView(actuators, dof_indices), where dof_indices is the selected 2-D global velocity-DOF index array. It derives and caches the opaque mappings internally.

newton/_src/actuators/view.py
AntoineRichard: Addressed in 3eb6b378: the redundant ActuatorView.from_articulation_view() classmethod was removed; callers use articulation_view.get_actuator_view(...) directly.

newton/_src/utils/selection.py
AntoineRichard: Addressed in 3eb6b378: the API and documentation now refer specifically to selected velocity-DOF columns instead of a generic DOF layout.

newton/_src/actuators/view.py
AntoineRichard: Addressed in 3eb6b378: ActuatorView and ArticulationView now have identical public get/set signatures, and ArticulationView delegates directly through those public methods. Component objects remain compatible, with actuator-relative string paths added.

PR #4074, opened 2026-08-28
Fix sparse topology classification test

nvtw:
## Description

Limit the forced single-thread launch to topology classification. The test previously serialized the entire sparse field update and could hang GPU test jobs while exercising unrelated kernels.

## Summary by CodeRabbit

* **Tests**
  * Updated sparse topology classification coverage to explicitly refresh field data before classification.
  * Improved test control by resetting grid counts and using a single processing thread.
  * Existing topology count expectations remain unchanged.


gdaviet (approved):
Approved, but feel free to remove the whole test instead, it looks quite superfluous actually.

PR #4078, opened 2026-08-28
Handle empty mesh decomposition results

### LINKED ISSUE #3507: Handle empty mesh decomposition results as failures
## Problem

`ModelBuilder.approximate_meshes()` silently leaves the original mesh unchanged when CoACD or V-HACD returns an empty decomposition.

The current branch is:

```python
if len(decomposition) == 0:
    continue
```

For a non-empty input mesh and an empty backend result, the method currently:

- returns an empty remeshed-shape set;
- leaves the shape as `GeoType.MESH`;
- emits no warning;
- does not raise even when `raise_on_failure=True`.

This conflicts with the public fallback contract: decomposition failure should raise when requested, or otherwise warn and fall back from CoACD/V-HACD to `convex_hull`, then to `bounding_box` if needed.

This behavior predates PR #3471. The empty-result branch originated with the original decomposition support in #407; the fallback and `raise_on_failure` contract was added in #430 without reconciling this path. It was confirmed while reviewing #3471: https://github.com/newton-physics/newton/pull/3471#discussion_r3573882920

## Reproduction

Use a stub backend that returns no parts:

```python
fake_coacd.run_coacd = lambda mesh, **kwargs: []

shape = builder.add_shape_mesh(body=-1, mesh=mesh)
result = builder.approximate_meshes(method="coacd", shape_indices=[shape])

assert result == set()
assert builder.shape_type[shape] == newton.GeoType.MESH
```

The same behavior occurs on PR #3471's base and head. V-HACD follows the same empty-result branch.

## Expected behavior

Treat an empty decomposition as a backend failure:

- with `raise_on_failure=True`, raise `RuntimeError`;
- otherwise, emit a warning and route only the affected shapes through the convex-hull fallback;
- preserve successful decompositions for other shapes in the same batch;
- if convex-hull generation fails, continue through the existing bounding-box fallback;
- return every shape successfully handled by either decomposition or fallback.

## Acceptance criteria

- Add regression coverage for empty CoACD/V-HACD results with `raise_on_failure=True` and `False`.
- Verify that mixed batches preserve successful decompositions and fall back only for empty-result shapes.
- Verify the final shape types and returned remeshed-shape set.
- Keep the documented warning and fallback behavior aligned with the implementation.

jcarius-nv:
## Description

Treat empty CoACD and V-HACD decomposition results as backend failures instead of silently leaving the source mesh unchanged. Strict callers now receive a shape-specific `RuntimeError`; non-strict callers warn and route only affected shapes through the existing convex-hull and bounding-box fallbacks.

Successful decompositions in a mixed batch are preserved, and every shape handled by decomposition or fallback is included in the returned set.

Closes #3507.

### Behavior change

| Case | Previously | With this fix |
|---|---|---|
| Backend returns one or more parts | Mesh becomes a convex decomposition | Unchanged |
| Backend throws | Raise, or warn and fall back | Unchanged |
| Empty result, `raise_on_failure=True` | Silently leaves `GeoType.MESH`; returns no shape | Raises a shape-specific `RuntimeError` |
| Empty result, default mode | Silently leaves the original mesh unchanged | Warns and tries `convex_hull` |
| Hull also fails | Never reached | Warns again and uses `bounding_box` |
| Mixed batch | Successful shapes are decomposed; empty-result shapes silently remain meshes | Successful shapes remain decomposed; only empty-result shapes use fallback |

This is an intentional correction to the existing public failure contract. Workflows that relied on the silent no-op may now observe a warning, an exception in strict mode, a changed collision approximation, and inclusion of the handled shape in the returned set. No method signatures, defaults, dependencies, or successful-decomposition paths change.

## Checklist

## Test plan

```console
uv run --extra dev -m newton.tests -k mesh_approximation
uvx pre-commit run -a
uvx --from towncrier==25.8.0 towncrier build --draft --version 1.6.0 --date 2026-08-27
git diff --check
```

- Mesh-approximation group: 15 passed.
- Broader model-filtered run: 176 passed, 109 skipped.
- Pre-commit and Towncrier draft: passed.
- `$code-review-newton`: Standards clean, Spec clean, Fit: `Fits`.

## Bug fix

**Steps to reproduce:**

1. Add a non-empty mesh shape to a `ModelBuilder`.
2. Stub CoACD or V-HACD to return no convex parts.
3. Call `approximate_meshes()` for that shape.
4. Observe that the original implementation returns an empty set, leaves the shape as `GeoType.MESH`, emits no warning, and ignores `raise_on_failure=True`.

**Minimal reproduction:**

```python
import sys
from types import SimpleNamespace
from unittest.mock import patch

import newton

mesh = newton.Mesh.create_box(1.0)
fake_coacd = SimpleNamespace(
    Mesh=lambda vertices, faces: (vertices, faces),
    run_coacd=lambda _mesh, **_kwargs: [],
)

shape = builder.add_shape_mesh(body=-1, mesh=mesh)

with patch.dict(sys.modules, {"coacd": fake_coacd}):
    result = builder.approximate_meshes(method="coacd", shape_indices=[shape])

assert result == set()
assert builder.shape_type[shape] == newton.GeoType.MESH
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Mesh approximation now correctly detects when CoACD or V-HACD produces no convex parts.
  * Configured failure behavior is now honored: operations can raise an error or continue with a warning.
  * Failed decompositions automatically fall back to convex hull generation, with bounding-box fallback available when needed.
  * Successful decompositions within the same request are preserved while only failed shapes use fallback handling.

### DIFF
--- a/changelog/3507.fixed.md
+++ b/changelog/3507.fixed.md
@@ -1,0 +1,1 @@
+Treat empty CoACD and V-HACD decompositions as failures so mesh approximation raises or uses the documented fallbacks.
--- a/newton/_src/sim/builder.py
+++ b/newton/_src/sim/builder.py
@@ -7600,6 +7600,8 @@ class ModelBuilder:
         remeshed_shapes = set()
 
         if method == "coacd" or method == "vhacd":
+            empty_decomposition_shape = None
+            decomposition_failed = False
             try:
                 if method == "coacd":
                     # convex decomposition using CoACD
@@ -7653,6 +7655,15 @@ class ModelBuilder:
                                 decomposition.extend((d["vertices"], d["faces"]) for d in component_decomposition)
                         decompositions[hash_m] = decomposition
                     if len(decomposition) == 0:
+                        if raise_on_failure:
+                            empty_decomposition_shape = shape
+                            break
+                        warnings.warn(
+                            f"Remeshing with method '{method}' failed for shape {shape}: the backend returned no "
+                            "convex parts. Falling back to convex_hull.",
+                            stacklevel=2,
+                        )
+                        decomposition_failed = True
                         continue
                     # note we need to copy the mesh to avoid modifying the original mesh
                     replacement_mesh = self.shape_source[shape].copy(
@@ -7721,6 +7732,15 @@ class ModelBuilder:
                     method = "convex_hull"
                     # kwargs were addressed to the failed decomposition method
                     remeshing_kwargs = {}
+            if empty_decomposition_shape is not None:
+                raise RuntimeError(
+                    f"Remeshing with method '{method}' failed for shape {empty_decomposition_shape}: "
+                    "the backend returned no convex parts."
+                )
+            if decomposition_failed:
+                method = "convex_hull"
+                # kwargs were addressed to the failed decomposition method
+                remeshing_kwargs = {}
 
         if method in RemeshingMethod.__args__:
             # remeshing of the individual meshes
--- a/newton/tests/test_model.py
+++ b/newton/tests/test_model.py
@@ -216,6 +216,23 @@ class TestModelBuilderBvhConstructor(unittest.TestCase):
 
 
 class TestModelMesh(unittest.TestCase):
+    class _FakeDecompositionMesh:
+        """Store mesh data passed to a fake decomposition backend."""
+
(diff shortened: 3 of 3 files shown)


eric-heiden (approved):
LGTM!

Issue #1724, opened 2026-02-22
Clean up noisy warnings in unit test output


shi-eric:
## Summary

CI test logs are cluttered with warnings and informational messages that obscure actual test results. In a representative run (8024 lines), ~2159 lines are test results and **~90 lines are actionable noise** from warnings that shouldn't be emitted during normal test execution.

This issue catalogs each noise category, its line count, and a proposed fix.

## Actionable warnings

These should be suppressed, captured, or fixed. Line counts are from a single CI run — they scale with the number of tests.

| Category | Lines | Example | Proposed fix |
|----------|------:|---------|--------------|
| Invalid inertia tensor (USD/PhysX runtime) | 29 | `Warning: The rigid body at /Robot/Base has a possibly invalid inertia tensor of {1.0, 1.0, 1.0}...` | Originates from USD runtime, not Newton. Tests should suppress this output or fix test USD stages to have valid inertia. The advice "add colliders regardless then disable the 'enable collision' property" is confusing even for end users. |
| Inertia validation corrected | 9 | `UserWarning: Inertia validation corrected 3 bodies. Set validate_inertia_detailed=True...` | Tests not specifically testing inertia validation should suppress this. Tests that *are* testing it should use `warnings.catch_warnings(record=True)` and assert on the message. |
| Custom attribute `mujoco:ctrl` | 35 | `UserWarning: Custom attribute 'mujoco:ctrl' has 0 values but frequency 'mujoco:actuator' expects 5...` | Fix the underlying issue (populate the ctrl attribute in test data) or suppress the warning in test setup. |
| NarrowPhase CPU fallback | 5 | `UserWarning: NarrowPhase running on CPU: mesh-mesh contacts will be skipped...` | Don't warn about a configuration the user explicitly chose. Only emit this when a GPU was requested but unavailable, or gate behind a verbosity level. |
| Non-manifold edge | 6 | `UserWarning: Detected non-manifold edge` | Not actionable for the user. Gate behind a verbose/debug logging level. |
| Friction NaN (mujoco_warp) | 1 | `UserWarning: geom 0: friction[0] (0.0) < MJ_MINMU (1e-05) with condim=3 may cause NaN` | Third-party warning from `mujoco_warp`. Tests triggering this should capture/suppress it. |
| Zero/negative mass | 1 | `UserWarning: parent_body 9 has zero or negative mass (0.0). This may cause unexpected behavior.` | Same as inertia warnings — capture in test or fix test data. |
[...]

PR #4080, opened 2026-08-28
Enforce strict warnings across test modules

jcarius-nv:
## Description

The fixture cleanups in #4079 and #4081 have landed. This branch now merges
the current `main`, removes `test_sdf_usd` from `_KNOWN_WARNING_DEBT`, and
removes the temporary SDF-only warning regression because the suite-wide
policy in this PR provides the lasting guard.

Make caller-attributed warnings strict for every top-level and nested `test_*`
module. Keep the remaining known warning debt visible through narrow filters
that match the warning category, message, and attributed test module. This
prevents unrelated new warnings from being accepted merely because their test
module has existing debt.

Merging current `main` exposed two additional warning sources: inertia
correction in `test_convex_support` and `test_controllers_operational_space`,
and the intentional parallel-joint warning in `test_eval_fk`. This PR records
those exact warnings as known debt so strict enforcement remains enabled for
all other warnings in those modules.

Add focused regression coverage for top-level and nested test modules, known
debt, and unrelated warnings in a debt-bearing module.

This is one incremental step toward #1724 and intentionally does not close the
issue.

## Checklist

- [x] The documentation is up to date with these changes (no user-facing
      documentation change)
- [x] No changelog fragment is needed because this changes test enforcement
      only

## Test plan

```text
uv run --extra dev -m newton.tests -k TestStrictWarnings -k test_convex_support -k test_controllers_operational_space -k test_eval_fk --strict-warnings --no-cache-clear
uv run --extra dev -m newton.tests --strict-warnings --no-cache-clear
uv run --with pre-commit pre-commit run -a
```

On the merged-main branch, the full strict-warning run executed 7,055 tests
and identified only seven newly introduced warnings (one inertia warning in
`test_controllers_operational_space` and six parallel-joint warnings in
`test_eval_fk`). After adding exact debt filters for those warnings, the
focused strict-warning run passed all 94 tests. Pre-commit also passed.


eric-heiden (dismissed):
LGTM! Just a minor nit my agent found, but this shouldn't block merging.

newton/tests/thirdparty/unittest_parallel.py
```diff
@@ -43,9 +43,43 @@
...
+            r"test_import_mjcf|test_jacobian_mass_matrix|test_kinematic_links|test_model|test_mujoco_solver|"
+            r"test_off_origin_convex_hull_contacts|test_rigid_contact)$"
+        ),
+    },
+    {
+        "message": r"Rigid body prim /World/Negative/Complete has a mirrored \(negative-determinant\) world transform\..*",
+        "category": UserWarning,
+        "module": r"test_import_usd$",
+    },
+    {
```
eric-heiden, line 71: 🔵 `_STRICT_WARNING_TEST_MODULE` deliberately tolerates a dotted package prefix (`(?:.*\.)?test_.*$`), but every `_KNOWN_WARNING_DEBT` entry matches a bare module name: `(?:test_collision_cloth|...|test_rigid_contact)$`, `test_import_usd$`, and `test_recorder$`. The `warnings` module applies a `module` filter with `re.match`, which anchors at the start of the string, so `test_recorder$` can never match a module imported as `kamino.test_recorder`.

The result is that escalation and exemption cover different sets of modules. Discovery starts at `newton/tests` by default, so `newton/tests/kamino/test_kamino_core_model.py` is imported as `kamino.test_kamino_core_model`: its warnings do get escalated to errors, but adding `test_kamino_core_model$` to the debt list would silently never fire. Nothing fails today because no module under `newton/tests/kamino/` currently emits one of these warnings, so this is latent rather than broken, but the next person who needs to book debt for a nested module has to rediscover the prefix rule.

| Module as imported | Escalated to error | Can be exempted via debt list |
| --- | --- | --- |
| `test_recorder` | yes | yes |
| `kamino.test_recorder` | yes | no |

Factoring the prefix into one place keeps the two halves in sync:

```python
_DEBT_MODULE_PREFIX = r"(?:.*\.)?"

_KNOWN_WARNING_DEBT = (
    ...
    {
        "message": r"invalid value encountered in cast",
        "category": RuntimeWarning,
        "module": rf"{_DEBT_MODULE_PREFIX}test_recorder$",
    },
)
```

A companion case in `TestStrictWarnings` asserting that a debt warning attributed to `kamino.test_recorder` is downgraded would sit well next to `test_nested_test_module_warning_is_an_error` and keep the two patterns from drifting again.

jcarius-nv: Addressed in 45a1aa17. `_TEST_MODULE_PREFIX` is now shared by strict escalation and every debt-module filter, and `test_nested_recorder_cast_warning_debt_is_not_an_error` covers `kamino.test_recorder`. The focused strict-warning/recorder tests and pre-commit pass.

eric-heiden (approved):

PR #4082, opened 2026-08-28
Fix mobile docs navigation drawer losing table of contents structure

mcrowleyNV:
## Description

On phones and other narrow viewports, the documentation navigation drawer opens on a flat, unindented list of every page instead of the captioned table of contents that wide screens show.

Below 960px, PyData Sphinx Theme moves the top navbar links into the primary sidebar (`.sidebar-header-items__center`) and renders them above the table of contents. Every Newton page is a top-level toctree entry, so those links repeat the entire table of contents — without the `User Guide`, `Concepts`, `Solvers`, … captions and their indentation — and that flat copy is what the drawer shows first.

This change hides the duplicated navbar links inside the primary sidebar, so the drawer opens on the same structured table of contents that wide screens use. The theme already hides that section above 960px, so the top navbar, wide-screen rendering, and the theme and version switchers at the top of the drawer are all unchanged.

An alternative would be `"navbar_center": []` in `html_theme_options`, which drops the navbar links at every viewport width and also stops the header links from wrapping onto a second line on desktop. Happy to switch to that if you would rather retire the top navbar links entirely.

## Checklist

- [ ] New or existing tests cover these changes — not applicable, stylesheet-only change

## Test plan

Full docs build, then the drawer inspected in Chrome against that build:

```
uv run --extra docs --extra sim sphinx-build -j auto -W -b html docs docs/_build/html   # build succeeded, exit 0
uv run docs/serve.py --port 8080
```

At 390×844 with the drawer open, the built page shows `Table of Contents`, all six captions (`User Guide`, `Concepts`, `Solvers`, `API Reference`, `Further Reading`, `Project Links`), 39 indented `toctree-l1` entries, and the theme and version switchers. Reverting just this rule in the built stylesheet brings the flat list back and fails the same check, so it tracks the change rather than passing by default.

At 1400×900, the header nav and the left sidebar table of contents are unchanged; renders of the page with and without the rule are byte-identical.

Lint:

```
uvx pre-commit run --files docs/_static/custom.css changelog/+mobile-sidebar-toc-4c7b91de.fixed.md
```

## Bug fix

**Steps to reproduce:**

1. Open https://newton-physics.github.io/newton/stable/ on a phone, or in a desktop browser narrowed to less than 960px.
2. Tap the hamburger menu.
3. The drawer lists every page as a flat, unindented list with no section captions; the captioned table of contents only appears further down, after that list.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved narrow-screen documentation navigation by displaying the captioned, indented table of contents in the navigation drawer.
  - Removed duplicated, flat table-of-contents links from the mobile sidebar.

Resulting images are attached. 
image


eric-heiden (approved):
Thanks for fixing this!

mcrowleyNV:
> Thanks for fixing this!

@eric-heiden Any time. Please let DocOps ([#omni-docs-hub](https://nvidia.enterprise.slack.com/archives/CJS9VREE5)) know if/when your Newton docs need editorial reviews, agentic readiness testing, etc. We're here to help and we'd like to be more involved.

PR #4086, opened 2026-08-29
Fix #3702: warn/raise on nconmax overflow in SolverMuJoCo

### LINKED ISSUE #3702: [BUG] Exceeding nconmax gives NaN with only a printed message
## Description

When the number of contacts exceeds `nconmax`, `SolverMuJoCo` prints a message and continues with constraints dropped. The simulation that follows is not just inaccurate, it can go to NaN, and by the time anything is visibly wrong the printed line has usually scrolled away. The same is true of `njmax` (`nefc overflow`) and of `CollisionPipeline`'s own contact capacity.

This is easy to hit accidentally because the default capacity is estimated from the initial state. A scene that is quiet at rest and busy later — a hand closing on an object, a pile collapsing, a gripper engaging — silently exceeds a budget that looked adequate when the solver was constructed. In a scene of five convex-decomposed objects plus a 16-shape hand we measured about 9,100 contacts at rest and several times that once the hand closed.

Because the message goes to stdout rather than the warnings machinery, it also cannot be turned into an error with `-W error` or captured with `warnings.catch_warnings`.

## Minimal reproduction

```python
import newton
import numpy as np
import warp as wp

def run(nconmax, n=12):
    b = newton.ModelBuilder()
    for i in range(n):
        body = b.add_body(xform=wp.transform(wp.vec3(0.0, 0.0, 0.06 + 0.105 * i), wp.quat_identity()))
        b.add_shape_box(body, hx=0.05, hy=0.05, hz=0.05)
    b.add_ground_plane()
    model = b.finalize()
    solver = newton.solvers.SolverMuJoCo(model, njmax=4 * nconmax, nconmax=nconmax, use_mujoco_contacts=False)
    pipeline = newton.CollisionPipeline(model)
    s0, s1 = model.state(), model.state()
    control = model.control()
    for _ in range(480):
        pipeline.collide(s0, contacts)
        solver.step(s0, s1, control, contacts, 1 / 240)
        s0, s1 = s1, s0
    return s0.joint_q.numpy().reshape(n, 7)[:, 2]

for nconmax in (8, 2000):
    z = run(nconmax)
    print(f"nconmax = {nconmax:5d}  box heights: {np.round(z, 3)}")
```

Output on `main` (d124e5bd) — a stack of 12 boxes settling for 2 s:

```
Number of Newton contacts (96) exceeded MJWarp limit (92). Increase nconmax.
nefc overflow - please increase njmax to 48
nconmax =     8  box heights: [nan nan nan nan nan nan nan nan nan nan nan nan]
nconmax =  2000  box heights: [0.049 0.148 0.247 0.346 0.446 0.545 0.644 0.743 0.843 0.942 1.042 1.142]
```

The run returns NaN for every body and the only indication is two lines on stdout.

## Expected behavior

- Route the overflow through `warnings.warn` so it can be filtered, escalated to an error, or captured in tests.
- Offer a strict mode, for example `SolverMuJoCo(..., strict_capacity=True)` and `CollisionPipeline(..., strict_capacity=True)`, that raises on overflow. Contact-rich scenes are exactly the ones where silently dropping constraints is least acceptable.
- Consider surfacing a counter (peak contacts seen versus `nconmax`) so callers can size the buffer from a short warm-up run rather than by trial and error.

## Validation

- Overflow raises under strict mode and warns otherwise.
- Non-overflowing scenes are unaffected and emit nothing.
- `njmax` overflow and the collision pipeline's contact capacity behave the same way.

## System information

- Newton: 1.5.0.dev0 (d124e5bd)
- Warp: 1.16.0.dev20260716
- mujoco-warp: 3.10.0.2
- Python 3.12, Ubuntu 24.04, NVIDIA RTX PRO 6000 Blackwell

aneangel:
## Summary

Closes #3702.

Exceeding `nconmax` in `SolverMuJoCo` previously caused silent contact clipping, a device-side `wp.printf` message (invisible to Python), and eventual NaN propagation. This PR surfaces that condition at the Python level.

**Two new checks in `_src/solvers/mujoco/solver_mujoco.py`:**

1. **Buffer mismatch (no GPU sync)** — in `_convert_contacts_to_mjwarp`, immediately after `naconmax = self.mjw_data.naconmax` is already computed, compare `contacts.rigid_contact_max > naconmax`. Default: `UserWarning`. `strict_capacity=True`: `RuntimeError`.

2. **Runtime overflow (GPU sync, strict mode only)** — after `_mujoco_warp_step`, read `contacts.rigid_contact_count` and `mjw_data.overflow` from the device and raise `RuntimeError` if either overflowed.

**New constructor parameter:**

```python
solver = newton.solvers.SolverMuJoCo(model, strict_capacity=True)
# raises RuntimeError on overflow instead of warning
```

The existing device-side `wp.printf` in the contact conversion kernel is left intact so users without Python warning capture still see the message.

## Test plan

- [x] `uv run --extra dev newton/tests/test_mujoco_solver.py` — 276 tests pass, 7 skipped
- [x] Manual repro: `nconmax=1` with 30 spheres → `UserWarning` fires (default), `RuntimeError` raised (`strict_capacity=True`)
- [x] Pre-commit hooks: all pass

## Summary by CodeRabbit

* **New Features**
  * Added optional per-step overflow reporting through `state.mujoco.overflow`.
  * Reports contact-pipeline, solver-capacity, and MuJoCo Warp overflow flags.
  * Overflow status resets after each simulation step and supports CUDA graph replay.
  * Added optional strict capacity validation for MuJoCo simulations.

* **Bug Fixes**
  * Corrected overflow reporting when MuJoCo performs its own collision detection.
  * Capacity mismatches continue silently by default; strict mode raises an error.


eric-heiden:
Thanks for tackling this, @aneangel! Turning a silent contact drop into an observable signal, with per-step overflow bits exposed through `State`, is a real improvement over the current behavior.

Two things look worth fixing before this lands: the new host-side capacity check warns (or raises, under `strict_capacity`) on default scenes that never actually overflow, and the overflow launch dereferences `contacts` on the path where the docstring says it is ignored.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -4509,6 +4592,16 @@ def _convert_contacts_to_mjwarp(self, model: Model, state_in: State, contacts: C
...
+        if contacts.rigid_contact_max > naconmax:
+            msg = (
+                f"contacts.rigid_contact_max ({contacts.rigid_contact_max}) exceeds "
+                f"nconmax ({naconmax}); Newton contacts beyond the cap are silently "
+                f"discarded and results may be invalid. "
+                f"Pass nconmax>={contacts.rigid_contact_max} to SolverMuJoCo to avoid this."
+            )
+            if self._strict_capacity:
+                raise RuntimeError(msg)
+            warnings.warn(msg, UserWarning, stacklevel=4)
```
eric-heiden: 🟠 This check compares two buffer *capacities*, not the number of contacts actually generated, so it fires on healthy scenes where nothing is ever dropped.

`contacts.rigid_contact_max` comes from Newton's collision pipeline, which floors capacity at `_RIGID_CONTACT_MIN_CAPACITY = 1000` for any model (`newton/_src/sim/collide.py:550,652`). `naconmax` comes from MuJoCo Warp's `_default_nconmax` heuristic, roughly `max(45, ...)` rounded to a valid size and multiplied by `nworld`. The two are sized by independent heuristics and routinely disagree.

Concretely, `python -m newton.examples robot_anymal_d` (defaults `use_mujoco_contacts=False`, `nconmax=45`, `world_count=8`, plain `newton.CollisionPipeline(self.model)`) now emits a `UserWarning` on its first step advising `nconmax>=<thousands>`, which would inflate MuJoCo's constraint buffers by roughly 20x for no reason. `example_robot_anymal_c_walk.py` is in the same shape. With `strict_capacity=True` that same false positive becomes a hard `RuntimeError`, so strict mode is unusable on these scenes.

The PR already implements the correct runtime test on the GPU (`count > naconmax` in `collect_overflow_kernel`), so real overflow is reported via `OVERFLOW_SOLVER_NCONMAX` regardless. Either drop this host-side check entirely, or scope it to `strict_capacity` only (no warning by default) and reword it as a capacity mismatch rather than "results may be invalid":

```suggestion
        if self._strict_capacity and contacts.rigid_contact_max > naconmax:
            raise RuntimeError(
                f"contacts.rigid_contact_max ({contacts.rigid_contact_max}) exceeds "
                f"nconmax ({naconmax}); contacts beyond the cap will be dropped if more "
                f"than {naconmax} contacts are actually generated. "
                f"Pass nconmax>={contacts.rigid_contact_max} to SolverMuJoCo to size for the worst case."
            )
```

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -4090,8 +4150,31 @@ def step(self, state_in: State, state_out: State, control: Control, contacts: Co
...
+                            # 1 when Newton fed contacts in; 0 when MuJoCo Warp
+                            # ran its own collision detection (bits 16-17 are
+                            # meaningless on that path — rigid_contact_count may
+                            # be stale or zero).
+                            int(not self.mjw_model.opt.run_collision_detection),
+                            self.mjw_data.overflow,
+                        ],
+                        outputs=[overflow_out],
+                        device=self.model.device,
+                    )
```
eric-heiden, line 4271: 🟠 This launch reads `contacts.rigid_contact_count` and `contacts.rigid_contact_max` unconditionally, but the docstring a few lines above states that `contacts` is "Ignored when `use_mujoco_contacts=True`", which is the constructor default. Passing `None` there is an established pattern in this repo: `example_recording.py:91`, `example_mpm_anymal.py:248` and `example_basic_dzhanibekov.py:117` all call `SolverMuJoCo.step(..., contacts=None, ...)`.

Add `builder.request_state_attributes("mujoco:overflow")` to any of those, which is the documented way to watch bits 0-15 (mujoco-warp's own NEFC/NJMAX flags, exactly the signals that path produces), and the step fails with `AttributeError: 'NoneType' object has no attribute 'rigid_contact_count'`. On `main` those calls work because nothing else in `step()` touches `contacts` when MuJoCo Warp runs its own collision detection.

Since the `newton_contacts` branch that consumes these values is 0 on that path anyway, compute the Newton inputs defensively before the launch, for example:

```py
newton_contacts = int(not self.mjw_model.opt.run_collision_detection and contacts is not None)
```

and pass a dummy int32 array plus `0, 0` for the counts and capacities when `contacts is None`, so bits 0-15 are still reported on the MuJoCo contacts path.

newton/_src/solvers/mujoco/solver_mujoco.py
aneangel: went ahead and address this based on your suggestion

newton/_src/solvers/mujoco/solver_mujoco.py
aneangel: thanks for pointing this out, I went back and took a look :)

PR #4088, opened 2026-08-30
Fix translated floating-base IK Jacobians

### LINKED ISSUE #3866: Floating-base free-joint analytic IK Jacobian drifts with distance from world origin (regression from #2539)

ooctipus:
## Description

Fix distance-dependent drift in analytic IK for translated `FREE` and `DISTANCE` joints.

The public Jacobian and mass-matrix paths need motion-subspace columns referenced to the child COM, while the current LM and L-BFGS IK integrators apply generalized-coordinate updates about the parent anchor. This change makes the shared helper's pivot explicit and has each IK optimizer request the parent-anchor pivot without changing the public COM-referenced path.

Regression coverage uses nonzero parent-anchor translation and rotation, a nonzero child COM, and base translations from 0 to 20 m. It checks LM analytic Jacobians and L-BFGS analytic gradients against AUTODIFF on CPU and CUDA.

This preserves the public twist convention from #2539 and addresses the review feedback on #3868.

Closes #3866

## Checklist

## Test plan

- [x] Verified the new translated-joint regressions fail before the fix (42 subtest failures).
- [x] `uv run --extra dev -m newton.tests -p "test_ik*.py" --strict-warnings` (111 passed).
- [x] `uv run --extra dev -m newton.tests -k floating_free_jacobian -k descendant_free_jacobian -k floating_free_mass_matrix -k descendant_free_mass_matrix` (12 passed).
- [x] `uvx pre-commit run -a`.
- [x] Towncrier draft build.
- [x] Isaac Lab's ANYmal-C FLAT sampler recovered `39.06%` IK acceptance (`1406 / 3600`), matching its historical baseline instead of the regressed `9.25%`.

## Bug fix

**Steps to reproduce:**

1. Build a model with a root `FREE` or `DISTANCE` joint and a nonzero parent anchor.
2. Evaluate a position objective at root translations of 0, 1, 5, and 20 m.
3. Compare the ANALYTIC derivative with AUTODIFF.
4. Observe that the difference grows with translation before this change.

**Minimal reproduction:**

Run the added `test_translated_*_jacobian_compare` cases against the parent commit. They fail for both joint types and pass with this change.

### DIFF
--- a/changelog/3866.fixed.md
+++ b/changelog/3866.fixed.md
@@ -1,0 +1,1 @@
+Fix distance-dependent drift in analytic IK Jacobians for FREE and DISTANCE joints.
--- a/newton/_src/sim/articulation.py
+++ b/newton/_src/sim/articulation.py
@@ -934,17 +934,15 @@ def eval_ik(
 @wp.func
 def write_free_distance_motion_subspace(
     X_pa_world: wp.transform,
-    x_child_com_world: wp.vec3,
+    pivot_world: wp.vec3,
     qd_start: int,
     # outputs
     joint_S_s: wp.array[wp.spatial_vector],
 ):
     """Write the 6 motion-subspace columns for a FREE/DISTANCE joint.
 
-    Used by both the Featherstone inverse-dynamics path (``jcalc_motion``) and
-    the IK/Jacobian path (``jcalc_motion_subspace``) so they agree on the exact
-    column layout. Linear DOFs act at the child body's COM; angular DOFs are
-    world-aligned axes expressed through ``X_pa_world``.
+    Linear DOFs follow the parent-anchor axes. Angular columns describe
+    rotations about ``pivot_world`` as world-origin spatial twists.
 
     Args:
         X_pa_world: Parent-anchor world transform (``X_wp * joint_X_p``) used
@@ -952,7 +950,7 @@ def write_free_distance_motion_subspace(
             This is *not* the classical Featherstone ``X_sc`` (spatial-to-
             child); Newton's FREE/DISTANCE joint coordinates live in the
             parent-anchor basis.
-        x_child_com_world: World-space position of the child body's COM.
+        pivot_world: World-space point about which angular columns rotate.
         qd_start: Starting velocity-DOF index for this joint.
         joint_S_s: Output spatial-vector subspace array; six slots starting at
             ``qd_start`` are overwritten.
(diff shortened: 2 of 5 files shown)


eric-heiden (approved):
Thanks, LGTM!

PR #4092, opened 2026-08-31
Expose Viser viewer extension context

### LINKED ISSUE #4091: [REQ] Expose application extension handles from ViewerViser
### Description

Expose a small, typed, non-owning context from `newton.viewer.ViewerViser` for applications that need to add Viser scene nodes and GUI controls to the existing viewer session.

The context would contain only the existing Viser scene API, GUI API, and initial-camera configuration. It would deliberately omit the `ViserServer`, so Newton remains the sole owner of the server and its lifecycle. The returned context should have stable identity for the lifetime of the viewer.

### Motivation / Use Case

Applications sometimes need controls and scene annotations that are specific to their workflow: editable waypoints, target markers, selection widgets, or status panels. These should share Newton's existing `ViewerViser` scene and camera rather than create a second viewer or take over renderer/server lifecycle.

Today the only practical ways to do that are to reach into the private `ViewerViser._server` attribute or maintain a custom viewer implementation. The private access has no API contract, while a second/custom viewer duplicates lifecycle and native visualization behavior.

This is narrower than #4039: it does not discover or load third-party Viewer classes. The application constructs the normal `ViewerViser` and adds application-owned content through capability-limited handles.

Ownership would remain explicit:

- Newton owns the Viser server and closes it with the viewer.
- The application owns every node, control, callback, and background task it creates through the context.
- The public context cannot start, stop, or otherwise expose the server.

### Alternatives Considered

- Access `ViewerViser._server`: works today, but depends on a private implementation detail and exposes lifecycle methods unnecessarily.
- Implement a custom `ViewerBase`: appropriate for a genuinely different viewer, but disproportionate when an application only needs to extend the existing Viser scene and GUI.
- Start a separate Viser server: splits simulation visualization and application controls across two sessions and duplicates lifecycle ownership.

ruziniuuuuu:
## Description

Expose a typed, non-owning context from `ViewerViser` for applications that add scene nodes and GUI controls to the existing Viser session.

The context contains only the scene API, GUI API, and initial-camera configuration. It deliberately omits the `ViserServer`: Newton continues to own the server and viewer lifecycle, while the application owns the nodes, controls, callbacks, and background work it creates.

This is intentionally narrower than #4039. It does not discover or load third-party viewer classes; callers construct the normal `ViewerViser` and extend that instance through capability-limited handles.

Closes #4091.

## Checklist

- [x] For user-facing changes, a fragment has been added by following the changelog fragment instructions

## Test plan

```console
uv run --extra dev -m newton.tests -k viewer_viser_extension_context
uv run --extra dev -m newton.tests -k viewer_layers
uvx pre-commit run -a
```

Results: 1 extension-context test passed, 22 viewer-layer tests passed, and every pre-commit hook passed.

## New feature / API change

```python
from newton.viewer import ViewerViser

viewer = ViewerViser()
context = viewer.extension_context

origin = context.scene.add_frame("/application/origin")
visible = context.gui.add_checkbox("Show origin", initial_value=True)

@visible.on_update
def _(_event):
    origin.visible = visible.value

context.initial_camera.position = (3.0, 3.0, 2.0)
```

## Summary by CodeRabbit

* **New Features**
  * Applications can now extend Viser scenes with custom scene nodes, GUI controls, and callbacks through a dedicated extension context.
  * The extension context provides access to scene, GUI, and initial-camera controls without exposing the underlying viewer server.
  * The new extension context is publicly available through the viewer API.

* **Documentation**
  * Added usage guidance and API reference documentation for application extensions.

### DIFF
--- a/newton/_src/viewer/viewer_viser.py
+++ b/newton/_src/viewer/viewer_viser.py
@@ -7,8 +7,9 @@ import collections
 import inspect
 import os
 import warnings
+from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, ClassVar
+from typing import TYPE_CHECKING, Any, ClassVar
 from urllib.parse import urlparse
 
 import numpy as np
@@ -19,6 +20,27 @@ import newton
 from ..core.types import override
 from ..utils.texture import load_texture, normalize_texture
 from .viewer import ViewerBase, is_jupyter_notebook
+
+if TYPE_CHECKING:
+    import viser
+
+
+@dataclass(frozen=True, slots=True)
+class ViewerViserExtensionContext:
+    """Public Viser handles for application-owned scene and GUI content.
+
+    The context deliberately omits the underlying Viser server so extensions
+    cannot take over its lifecycle or other viewer-owned facilities.
+    """
+
+    scene: viser.SceneApi
+    """Viser scene interface used to add and update scene nodes."""
+
+    gui: viser.GuiApi
+    """Viser GUI interface used to add and update controls."""
+
+    initial_camera: viser.InitialCameraConfig
+    """Viser initial-camera configuration for newly connected clients."""
 
 
 class ViewerViser(ViewerBase):
@@ -158,6 +180,11 @@ class ViewerViser(ViewerBase):
 
         # Initialize viser server
         self._server = viser.ViserServer(port=port, label=label or "Newton Viewer")
+        self._extension_context = ViewerViserExtensionContext(
+            scene=self._server.scene,
+            gui=self._server.gui,
+            initial_camera=self._server.initial_camera,
+        )
         self._camera_request: tuple[np.ndarray, np.ndarray, np.ndarray] | None = None
         self._pending_camera_clients: set[int] = set()
         self._server.on_client_connect(self._handle_client_connect)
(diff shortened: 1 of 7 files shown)


newton/_src/viewer/viewer_viser.py
```diff
@@ -20,6 +21,27 @@
 from ..utils.texture import load_texture, normalize_texture
 from .viewer import ViewerBase, is_jupyter_notebook
 
+if TYPE_CHECKING:
+    import viser
+
+
+@dataclass(frozen=True, slots=True)
+class ViewerViserExtensionContext:
```
eric-heiden, line 29: Thanks for your contribution, @ruziniuuuuu!

Rather than exposing another entire class on the `newton.viewer` namespace I would recommend either of these 2 approaches:
* use a nested class `ViewerViser.ExtensionContext`, or
* directly expose `viser_scene`, `viser_gui`, `viser_camera` as properties on `ViewerViser`

newton/tests/test_viewer_viser_extension_context.py
eric-heiden, line 1: I'm not a fan of these mock tests because they don't actually test any meaningful behavior except for API contracts, so I would prefer to remove this file altogether.

PR #4096, opened 2026-08-31
Default `use_coord_layout_targets` to True

### LINKED ISSUE #3621: Flip the default of newton.use_coord_layout_targets to True
## Background

Since #3617, the legacy DOF-shaped target layout emits a `DeprecationWarning` from `ModelBuilder.finalize()` for models whose joint coordinate and DOF counts differ (free/ball/distance joints — the only models where the two layouts diverge), and the test suite and examples already run under the coordinate layout. Once that warning has shipped in a release, the default can flip.

## Scope

- [ ] Flip the default of `newton.use_coord_layout_targets` to `True`; keep the flag as an opt-out for existing code.
- [ ] The `finalize()` deprecation warning keeps firing for affected models under an explicit `False`.
- [ ] Heads-up to major downstream consumers (IsaacLab in particular) while the warned release is current. If they confirm migration, consider folding the full removal of the flag and the legacy layout into the same PR (single step) instead of keeping the opt-out for a release.

Blocked until the release carrying the #3617 warning has shipped.

camevor:
## Description
`Model.joint_target_q` and `Control.joint_target_q` are now coord-shaped `(joint_coord_count,)`, matching `joint_q`, unless a caller opts back into the deprecated DOF layout with `newton.use_coord_layout_targets = False`.

resolves #3621

## Checklist

## Test plan
```
uv run --extra dev -m newton.tests --strict-warnings
```

## Summary by CodeRabbit

* **New Features**
  * Joint position targets now use the coordinate-aligned layout by default, matching joint position data.
  * Applications can temporarily restore the deprecated DOF-shaped layout by setting the configuration option before model construction.

* **Bug Fixes**
  * Models using the deprecated layout now receive a deprecation warning when coordinate and degree-of-freedom counts differ.

* **Documentation**
  * Updated API references, migration guidance, tutorials, and examples to reflect the new default behavior.

### DIFF
--- a/newton/__init__.py
+++ b/newton/__init__.py
@@ -11,18 +11,17 @@ from ._src.core import (
 )
 from ._version import __version__
 
-use_coord_layout_targets: bool = False
+use_coord_layout_targets: bool = True
 """Use :attr:`joint_q`-aligned layout for joint position targets.
 
 Controls the shape of :attr:`~newton.Model.joint_target_q` and
 :attr:`~newton.Control.joint_target_q`:
 
-- ``True``: shape ``(joint_coord_count,)``, matching
+- ``True`` (the default): shape ``(joint_coord_count,)``, matching
   :attr:`~newton.State.joint_q`.
-- ``False`` (the default): legacy shape ``(joint_dof_count,)``, which is
-  misaligned with :attr:`~newton.State.joint_q` whenever an articulation
-  contains a free, ball, or distance joint upstream of a position-controlled
-  DOF.
+- ``False``: legacy shape ``(joint_dof_count,)``, which is misaligned with
+  :attr:`~newton.State.joint_q` whenever an articulation contains a free,
+  ball, or distance joint upstream of a position-controlled DOF.
 
 :attr:`joint_target_qd` is shaped ``(joint_dof_count,)`` in both layouts,
 matching :attr:`~newton.State.joint_qd`.
@@ -34,8 +33,7 @@ it before constructing a :class:`~newton.ModelBuilder`.
     The legacy DOF-shaped layout is deprecated. In a future release the
     coordinate layout becomes the only layout and this flag is removed;
     ``finalize()`` warns when building an affected model (one whose joint
-    coordinate and DOF counts differ) under ``False``. Set the flag to
-    ``True`` now to migrate.
+    coordinate and DOF counts differ) under ``False``.
 """
 
 __all__ = [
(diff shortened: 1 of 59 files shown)


jcarius-nv (commented):
## Fit

Fits. Newton owns the target-layout contract, and the clarified lifecycle—coordinate layout by default plus a legacy opt-out throughout 1.6, with removal tracked in #4114 for 1.7—is proportionate.

## Requirements

One P1 compatibility finding is inline. Please also confirm here that IsaacLab was rechecked and that its GA integration will set `newton.use_coord_layout_targets = False` before constructing any `ModelBuilder` while it continues using the legacy layout. This confirmation matters because #3621 explicitly included downstream coordination in scope.

## Standards

No additional findings.

Focused validation at `7b060baadcb78ba47e4d2a5f229b11d9abe58087`: both joint-target layout tests passed under strict warnings, and a direct explicit-`False` check produced DOF-shaped model/control buffers with DOF-aligned start indices. The current macOS CI failure is from asset-download DNS failures; the other listed hosted and GPU checks pass.

newton/__init__.py
```diff
@@ -11,18 +11,17 @@
 )
 from ._version import __version__
 
-use_coord_layout_targets: bool = False
+use_coord_layout_targets: bool = True
```
jcarius-nv, line 14: P1: Please add a focused regression for the explicit `False` compatibility path before merging. This default flip also leaves the suite running almost entirely under the coordinate layout; the existing affected-model `False` test checks only deprecation-warning emission and attribution. It does not assert the promised DOF-shaped `Model`/`Control` targets, `joint_target_q_start == joint_qd_start`, projected FREE/BALL/DISTANCE values, or representative target consumption. IsaacLab GA will continue relying on this legacy path, so its observable behavior needs durable coverage through 1.6.

jcarius-nv: (I think just one sanity check here is fine, and we can remove it again when we get rid of the flag eventually)

camevor: Added a regression test for the legacy layout in c5d28fe1e.
Just FYI: the test suite has already been running under the new layout when run through `newton.tests`.

jcarius-nv (approved):
Rechecked at c5d28fe1. The focused legacy-layout regression covers the explicit False compatibility path and passes under strict warnings. IsaacLab PR #7449 https://github.com/isaac-sim/IsaacLab/pull/7449 pins the flag before model construction in both integration entry points and adds downstream floating-base coverage. All current checks pass.

jcarius-nv:
Thanks @camevor !

PR #4097, opened 2026-08-31
Fix camera overshoot for low fps

chschuma-disney:
## Description

The camera velocity update is dependent on the viewer frame rate, and currently not capped. For low frame rates, the update can actually flip the camera velocity when going from an accelerated to a non-accelerated state, leading to camera jitter.

This PR caps update step, so that the camera deceleration can at most cancel out the current velocity, but not reverse it.

## Checklist

## Test plan

A regression test has been added:

```
uv run --extra dev -m newton.tests -k test_camera_deceleration_is_capped
```

## Bug fix

The following example script (showing a box, with a capped frame rate) reproduces the issue. When moving the camera left or right via keyboard, the camera will briefly oscillate when the key is released.

```import time

import warp as wp

import newton
import newton.examples


class Example:
    def __init__(self, viewer, args):
        self.viewer = viewer
        self.time = 0.0

    def step(self):
        time.sleep(0.1)

    def render(self):
        self.viewer.begin_frame(self.time)
        self.viewer.log_shapes(
            name="/box_instance",
            geo_type=newton.GeoType.BOX,
            geo_scale=(0.5, 0.5, 0.5),
            xforms=wp.array([wp.transform([0.0, 0.0, 2.0], wp.quat_identity())], dtype=wp.transform),
        )
        self.viewer.end_frame()
        self.time += 1.0 / 60.0


if __name__ == "__main__":
    viewer, args = newton.examples.init()
    newton.examples.run(Example(viewer, args), args)
```

### DIFF
--- a/changelog/+camera-jitter-fix-a7355d42.fixed.md
+++ b/changelog/+camera-jitter-fix-a7355d42.fixed.md
@@ -1,0 +1,1 @@
+Cap the viewer camera's per-frame damping factor to avoid overshoot-driven oscillation at low framerates.
--- a/newton/_src/viewer/viewer_gui.py
+++ b/newton/_src/viewer/viewer_gui.py
@@ -189,7 +189,7 @@ class ViewerGui:
             desired[:] = 0.0
 
         tau = max(1.0e-4, float(self._cam_damp_tau))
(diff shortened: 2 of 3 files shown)


eric-heiden (approved):
Thanks, LGTM!

PR #4101, opened 2026-08-31
Add Actuator.State.assign() for odd-length graph capture

### LINKED ISSUE #4098: [BUG] Stateful actuator state does not advance across CUDA graph replays when the captured region has an odd step count
## Bug Description

The documented pattern for stateful actuators — allocate two `Actuator.state()` objects and swap them host-side after each `Actuator.step` — is not correct under CUDA graph capture. The swap happens while kernels are being *recorded*, so it is not part of the replayed graph. A replayed graph therefore always starts from whichever buffer was current at capture time.

The result depends on how many actuator steps the captured region contains:

- **even count** — exact, because the swaps cancel within the graph;
- **odd count > 1** — the last step's state update is discarded on every replay;
- **exactly one step per graph** — the state never advances at all. Delay rings stay empty, PID integrals stay frozen at one step's worth, LSTM hidden state never evolves.

Two statements in `docs/concepts/actuators.rst` are each true in isolation and wrong together:

> Create two state objects with `Actuator.state` and swap them after each step
> ```python
> for step in range(3):
>     control.joint_f.zero_()
>     model.actuators[0].step(state, control, state_0, state_1, dt=0.01)
>     state_0, state_1 = state_1, state_0
> ```

> `ControllerPD` and `ControllerPID` are fully graphable. […] `Actuator.is_graphable` returns `True` when all components can be captured in a CUDA graph.

`Actuator.is_graphable()` returns `True` for `ControllerPID` regardless of the captured step count, so nothing warns the caller. A CUDA graph bakes shapes and addresses, not the Python name bindings that the documented loop relies on.

This is not specific to any one drive. It affects `Delay`, `ControllerPID` (`DrivePID`), `ControllerNeuralLSTM` (`DriveNeuralLSTM`), and the SISO GRU drive in #4043 identically — anything for which `is_stateful()` is `True`.

**Prior art in this repo.** #2693 fixed exactly this class of bug one level down: `Delay.State.write_idx` was a host-side Python `int` baked into the graph, and it was moved to a device-side `wp.array` with the comment *"Current write position in the circular buffer, shape (1,). Device-side for graph capture."* (`newton/_src/actuators/delay.py:144`). The remaining host-side element is the state-object swap itself. The test added by that PR, `TestDelayGraphCapture.test_delay_graph_n_not_multiple_matches_eager`, encodes the constraint in its docstring — *"With N even and N % buf_depth != 0, the test confirms graph replay matches eager execution"*, *"making graph replay correct for any even N"* — but the even-count precondition is not stated anywhere a user reading `docs/concepts/actuators.rst` would find it, and there is no test for odd N or N = 1.

## Reproduction Script

Newton-only, single GPU, no solver step needed. A `ControllerPID` with `kp = kd = 0`, `ki = 1`, a constant position error of 1 rad and `dt = 0.01` must accumulate exactly `0.01` of integral per step, so 12 steps must give `0.12` no matter how the loop is chunked.

```python
# repro_actuator_graph_state.py
import warp as wp

import newton
from newton.actuators import ControllerPID  # DrivePID after #4054

DT = 0.01
KI = 1.0
TARGET = 1.0
TOTAL_STEPS = 12


def build(device):
    b = newton.ModelBuilder()
    link = b.add_link()
    joint = b.add_joint_revolute(parent=-1, child=link, axis=newton.Axis.Z)
    b.add_articulation([joint])
    b.add_actuator(ControllerPID, index=b.joint_qd_start[joint], kp=0.0, ki=KI, kd=0.0)
    return b.finalize(device=device)


def setup(device):
    model = build(device)
    act = model.actuators[0]
    state = model.state()
    control = model.control()
    control.joint_target_q.fill_(TARGET)  # constant error of TARGET rad, joint_q stays 0
    return model, act, state, control


def integral_of(act_state):
    return float(act_state.controller_state.integral.numpy()[0])  # drive_state after #4054


def run_eager(device, total_steps):
    _, act, state, control = setup(device)
    s0, s1 = act.state(), act.state()
    for _ in range(total_steps):
        control.joint_f.zero_()
        act.step(state, control, s0, s1, dt=DT)
        s0, s1 = s1, s0
    wp.synchronize_device(device)
    return integral_of(s0)


def run_captured(device, steps_per_graph, replays):
    _, act, state, control = setup(device)
    s0, s1 = act.state(), act.state()
    # Exactly the documented loop body, wrapped in a capture.
    with wp.ScopedCapture(device) as capture:
        for _ in range(steps_per_graph):
            control.joint_f.zero_()
            act.step(state, control, s0, s1, dt=DT)
            s0, s1 = s1, s0
    for _ in range(replays):
    wp.synchronize_device(device)
    return integral_of(s0)


def main():
    device = wp.get_device("cuda:0")
    print(f"newton {newton.__version__}  warp {wp.config.version}  device {device}")
    print(f"expected integral after {TOTAL_STEPS} steps = {KI * TARGET * DT * TOTAL_STEPS:.4f}\n")

    eager = run_eager(device, TOTAL_STEPS)
    print(f"eager, {TOTAL_STEPS} steps, no graph : integral = {eager:.4f}   <- reference")

    for n in (1, 2, 3, 4, 6):
        replays = TOTAL_STEPS // n
        got = run_captured(device, n, replays)
        status = "OK   " if abs(got - eager) < 1e-6 else "WRONG"
        print(
            f"captured, {n} step(s)/graph x {replays} replays = {TOTAL_STEPS} steps: "
            f"integral = {got:.4f}   {status}"
        )


if __name__ == "__main__":
    main()
```

Output:

```
newton 1.5.0  warp 1.16.0  device cuda:0
expected integral after 12 steps = 0.1200

eager, 12 steps, no graph : integral = 0.1200   <- reference
captured, 1 step(s)/graph x 12 replays = 12 steps: integral = 0.0100   WRONG
captured, 2 step(s)/graph x 6 replays = 12 steps: integral = 0.1200   OK
captured, 3 step(s)/graph x 4 replays = 12 steps: integral = 0.0900   WRONG
captured, 4 step(s)/graph x 3 replays = 12 steps: integral = 0.1200   OK
captured, 6 step(s)/graph x 2 replays = 12 steps: integral = 0.1200   OK
```

With one step per graph the integral is stuck at a single step's contribution (`0.01`) after twelve steps. With three steps per graph, one step in three is lost (`0.09` instead of `0.12`).

This is a realistic configuration, not a contrived one: a simulator that captures one control decimation iteration per graph — decimation 1 — hits the `N = 1` case, and any odd decimation hits the lossy case. The failure is silent; the actuator produces plausible-looking effort the whole time.

## System Information

- Newton 1.5.0 (PyPI wheel), Warp 1.16.0, MuJoCo Warp 3.11.0
- Python 3.12, Linux, CUDA 12.9 / driver 13.0, NVIDIA RTX 5000 Ada Generation Laptop GPU (sm_89), memory pools enabled

## Relationship to #4054

#4054 renames `Controller*` to `Drive*`; any fix should land on top of that naming. The references above use current `main` names with the post-#4054 name in parentheses where it differs.

The defect survives the rename unchanged. Verified from the #4054 diff (head `b295ef3a`, base `main` @ `7c677b95`):

- `Actuator.step(sim_state, sim_control, current_act_state, next_act_state, dt)` keeps its signature and its host-side double-buffer contract; the only change in the state path is `current_act_state.controller_state` → `.drive_state`.
- The `Stateful Actuators` section of `docs/concepts/actuators.rst`, including the `state_0, state_1 = state_1, state_0` loop, is carried over verbatim apart from `Controller` → `Drive` prose.
- The "fully graphable" claim is carried over as *"`DrivePD` and `DrivePID` are fully graphable"*.

The repro above runs unmodified against #4054 — `newton.actuators.ControllerPID` and `Actuator.State.controller_state` both survive as deprecated aliases that emit a `DeprecationWarning` — or warning-free with the two-line change `ControllerPID` → `DrivePID` and `.controller_state` → `.drive_state`. It was executed against Newton 1.5.0; the #4054 compatibility statement is from reading the diff, not from a second run.

## Suggested resolution

Either of these would close it; we have no stake in which.

1. **Capture-safe state advance (preferred).** Keep a single state allocation and put the buffer selection in a device-side cursor that a kernel flips, exposed as something like `Actuator.advance_state()` so the swap is recorded as part of the graph. This is the same move #2693 already made for `Delay.State.write_idx`, applied one level up, and it makes the captured step count irrelevant.

2. **Make the existing contract honest.** Document the even-count rule in the `Stateful Actuators` and `Differentiability and Graph Capture` sections, and either have `Actuator.is_graphable()` reflect it or add a `newton.tests.test_actuators` case for `N = 1` and `N = 3` that asserts the documented behavior. This is cheap but leaves decimation-1 capture unusable for every stateful drive.

We have been guarding against this downstream by refusing capture at decimation 1 and warning at odd decimation, which is a workaround for the caller, not a fix.

---

**Verified against `main` @ `7c677b955e813f31edda5ba2e00c6ae4e992c9a9`** (2026-08-31), in addition to the 1.5.0 wheel the repro was executed on. Everything load-bearing holds on tip:

- `docs/concepts/actuators.rst` — the `Stateful Actuators` section including the `state_0, state_1 = state_1, state_0` loop, and the *"`ControllerPD` and `ControllerPID` are fully graphable"* sentence, are byte-identical to the text quoted above.
- `Actuator.state()`, `Actuator.step(..., current_act_state, next_act_state, dt)`, and the `update_state` writes into `next_act_state` are unchanged. The host-side swap is still the only thing that advances state.
- `ControllerPID.update_state` still copies `_next_integral` into `next_state.integral`, and `newton/_src/actuators/delay.py:144` still carries the *"Device-side for graph capture"* comment cited above.
- The reproduction script's API surface (`ModelBuilder.add_actuator`, `Actuator.state`, `ControllerPID.State.integral`) is unchanged on `main`, so it runs as written.

One change since 1.5.0 **widens** the scope rather than narrowing it. `main` refactored the compute path into effort modes, so `Actuator.is_graphable()` now delegates to `self._effort_mode.is_graphable()`; for the default explicit mode that is `self._controller.is_graphable()`, still unconditionally `True` for `ControllerPID` regardless of captured step count. The implicit effort mode added in #3855 advances the PID integral through the same `_next_integral` → `next_state.integral` double buffer (`ControllerPID.prepare_implicit`), so implicit actuation inherits the same defect.

AntoineRichard:
## Description

Stateful actuators cannot carry their final state into the next CUDA graph
replay when a captured region contains an odd number of actuator steps and the
caller relies only on the usual host-side state swap. A graph records fixed
buffer addresses, while `state_0, state_1 = state_1, state_0` only rebinds
Python names:

- even-length regions return to their original buffer orientation and work as
  expected;
- odd-length regions leave the advanced state in the opposite buffer, so the
  next replay starts from stale state;
- a single-step region updates its destination buffer on the first replay but
  does not advance state across replays.

This affects stateful drives such as `DrivePID`, Warp-backed neural drives, and
actuator `Delay` state. Closes #4098.

## Approach

This PR adds two documented, caller-selected ways to preserve state across an
odd-length captured region.

### One graph with boundary assignment

The new public `Actuator.State.assign(other)` method copies the advanced state
back into the current state at the captured region boundary, in place of the
final swap:

```python
    for i in range(steps):
        control.joint_f.zero_()
        actuator.step(state, control, state_0, state_1, dt=0.01)
        if steps % 2 == 1 and i == steps - 1:
            state_0.assign(state_1)
        else:
```

`assign()` preserves destination storage. It copies Warp arrays with
`wp.array.assign()` and Torch tensors in place with `Tensor.copy_()`. Component
states must have the same concrete type and must either:

- implement their own `assign(other)` method; or
- be concrete dataclasses whose declared fields contain supported Warp arrays,
  Torch tensors, or `None`.

The default path supports direct fields in regular and slotted dataclasses.
Custom nested or otherwise specialized layouts can implement `assign()`.
Mismatched components, undeclared attributes, and unsupported field layouts
raise instead of being silently skipped or rebound.

Torch-backed neural drives are not CUDA-graphable, but Torch tensors remain
supported so that the public state-copy operation works for all built-in
actuator state representations.

### Zero-copy alternating graphs

Callers can instead key captured graphs by the current state-buffer identity
and alternate between the two buffer orientations. This keeps the established
double-buffer behavior, performs no boundary copy, and requires at most two
graphs.

```python
graphs = {}
for _ in range(replays):
    key = id(state_0)
    if key not in graphs:
            after = run_region(state_0, state_1)
        graphs[key] = (capture.graph, after)
    graph, (state_0, state_1) = graphs[key]
    wp.capture_launch(graph)
```

## API and compatibility impact

- Adds the public `Actuator.State.assign(other)` method.
- Does not change `Actuator.step()`, the existing double-buffer contract, or
  actuator state layout.
- Does not introduce an unconditional per-step copy. A copy occurs only when a
  caller selects boundary assignment; alternating graphs remain zero-copy.
- Does not automatically change existing graph-capture behavior. Callers with
  odd-length captured regions must adopt one of the documented patterns.
- No 1.5.x backport is required for a caller-side fix: applications on 1.5.x
  can use alternating graphs. The new `assign()` convenience API can remain a
  current-release API addition.

The user-facing API addition is recorded in `changelog/4098.added.md`.

## Checklist

## Test plan

Validated at head `e212b946`:

```text
uv run --extra dev -m newton.tests -k TestActuatorStateAssign
    8 passed

uv run --extra dev -m newton.tests -k TestDriveStateGraphCapture
    2 passed

uv run --extra dev --extra torch-cu12 -m newton.tests -k test_actuators
    112 passed

    all hooks passed

    build succeeded
```

`TestDriveStateGraphCapture` compares the closed-form PID integral with eager
execution for one-, two-, and three-step captured regions. Boundary assignment
is covered in both explicit and implicit effort modes; alternating parity
graphs are covered as the zero-copy route.

`TestActuatorStateAssign` covers built-in `Delay` and `DrivePID` state,
Warp-backed neural state, Torch tensor copying without aliasing, regular and
slotted dataclass state, custom assignment for nested/slotted state, mismatched
components, undeclared attributes, and rejection of unsupported undecorated
state classes.

## Bug fix reproduction

Use a `DrivePID` with `kp = kd = 0`, `ki = 1`, a constant position error of
`1`, and `dt = 0.01`. After 12 actuator steps the integral must be `0.12`.
Without either pattern above, replaying a graph containing one actuator step
produces `0.01`, while replaying a graph containing three steps produces
`0.09`. Eager execution and even-length captured regions produce `0.12`.

Both documented patterns produce `0.12` for the odd-length cases as well.


AntoineRichard:
Three open points I would like a maintainer's call on before taking this out of draft. The first two are also in the description; the third is new here.

**1. The publish is not free, and there is a zero-copy alternative.** As written the fix costs one `wp.copy` per state array per step: a single `(N,)` copy for `ControllerPID`, the ring for `Delay`, hidden and cell for `ControllerNeuralLSTM`. Small in absolute terms, but it is per step and it buys something the double buffer was already paying for.

The other reading of #4098's option (i) avoids it: allocate each state array with a leading dimension of 2 and select the slot with a device-side cursor that a kernel flips. No copy at all, and the same capture correctness. I did not implement it because it changes the public shape of `ControllerPID.State.integral` and its siblings from `(N,)` to `(2, N)`, touches every controller kernel, and would need a deprecation window — which seemed like the wrong trade for a bug fix that has to land on top of #4054 anyway. If you would rather have the cursor version, or want it as a follow-up once the rename settles, say so and I will do it.

**2. A single state object is now almost sufficient, and one race is all that stops it.** After this change the second state object carries no information, so `Actuator.step` could take one. It still requires two, because `Delay.update_state` writes `write_idx[0]` from thread 0 while its other threads read it in the same kernel — safe while `current` and `next` are distinct allocations, a race the moment they alias. Splitting that pointer advance into its own dim-1 launch would legalize a single state object and remove the copy from point 1 entirely, which is a better end state than either option above. Left out to keep this diff reviewable; happy to fold it in here or file it separately.

**3. An existing test contains a workaround for this bug.** `run_test_actuator_pipeline` keys its captured graphs on `(id(state_in), id(act_a))` and builds two of them, with the comment "the state and actuator buffers alternate with period two, so keying the graphs on them builds at most two and replays them thereafter." That period-two alternation is the host-side swap this PR makes redundant. The test passes unchanged and I have not touched it, but the second graph is now dead weight and someone may want to collapse it later.

jcarius-nv:
@jvonmuralt could you take a look at this please? If it's indeed a bug or unexpected behavior, might be worth still fixing before the 1.6 codefreeze

jvonmuralt:
Could we follow the existing State.assign() pattern here? Keep normal actuator-state swapping, and copy the final output back only for odd-length CUDA graph captures. This avoids copying delay/neural state every actuator step and stays consistent with simulation state handling.

newton/_src/actuators/actuator.py
```diff
@@ -366,3 +400,7 @@ def step(
                 current_act_state.delay_state,
                 next_act_state.delay_state,
             )
+
+        # --- 6. Publish the advanced state on-device ---
+        if self.is_stateful() and next_act_state is not current_act_state:
```
jcarius-nv: **P0 — Preserve the established double-buffer contract.**

This unconditionally turns `current_act_state` into another output after every eager and captured step. A caller can no longer retain the pre-step state, so the PR's statement that eager behavior is unchanged does not hold for the public current/next contract. It also copies every state array on every step; for `Delay`, that includes three complete `(buf_depth, N)` rings plus counters, and the ASV suite has no actuator benchmark exercising this cost.

Please retain ordinary state swapping instead. For a 1.5-compatible zero-copy fix, odd-length captured regions can alternate two graphs—A-to-B and B-to-A—while even-length regions reuse one. The existing actuator test's graph cache keyed by state-buffer identity already demonstrates this approach. It requires no `Actuator` API or state-layout change.

newton/_src/actuators/actuator.py
```diff
@@ -32,6 +32,32 @@ def _scatter_add_kernel(
...
+        advanced: State the step just wrote.
+    """
+    for field in fields(advanced):
+        src = getattr(advanced, field.name)
+        if src is None:
+            continue
+        dst = getattr(current, field.name)
+        if isinstance(src, wp.array):
+            wp.copy(dst, src)
+        elif is_dataclass(src):
```
jcarius-nv: **P1 — The generic third-party-state claim does not hold for all `Controller.State` subclasses.**

`Controller.State` is itself a dataclass. An undecorated subclass can add a Warp array in `__init__`, remain `is_dataclass() == True`, but have an empty `fields()` result. I reproduced that at this head: publishing an advanced value of `[1]` left the current state at `[0]`. If this helper remains, it needs an explicit state-copy contract or attribute traversal and validation like `newton.State.assign()`. The parity-graph solution avoids needing this new generic copy path for the 1.5 backport.

newton/tests/test_actuators.py
```diff
@@ -4090,6 +4090,87 @@ def _loop(
...
+        else:
+            # Module loading and lazy allocation have to happen before a capture.
+            s0, s1 = run(s0, s1, 1)
+            s0.controller_state.integral.zero_()
+            s1.controller_state.integral.zero_()
+            with wp.ScopedCapture(device) as capture:
+                s0, s1 = run(s0, s1, steps_per_graph)
+            for _ in range(self.STEPS // steps_per_graph):
+                wp.capture_launch(capture.graph)
+        wp.synchronize_device(device)
```
jcarius-nv: **P2 — Drop the explicit synchronization.**

`.numpy()` already performs the synchronous device-to-host copy; Newton's coding guidelines explicitly prohibit `wp.synchronize_device()` immediately before it.

newton/tests/test_actuators.py
```diff
@@ -4090,6 +4090,87 @@ def _loop(
...
+        return float(s0.controller_state.integral.numpy()[0])
+
+    def _assert_integral_matches(self, implicit: bool) -> None:
+        expected = self.KI * self.TARGET * self.DT * self.STEPS
+        for steps_per_graph in (None, 1, 2, 3):
+            with self.subTest(steps_per_graph=steps_per_graph):
+                self.assertAlmostEqual(self._integral_after_all_steps(implicit, steps_per_graph), expected, places=6)
+
+    def test_pid_integral_advances_per_replay_explicit(self):
```
jcarius-nv: **P1 Standards — Add test docstrings.**

Both new `test_pid_integral_advances_per_replay_*` methods need triple-double-quoted docstrings beginning with a concise imperative summary, as required by `CODING_GUIDELINES.rst`.

jcarius-nv (commented):
Thank you for finding and isolating this graph-replay defect, and for opening a focused PR with a clear reproduction. I reproduced the PID failure boundary and confirmed that this branch passes both the focused explicit/implicit graph test and the full actuator suite.

My Newton-fit verdict is **Fits if narrowed**. CUDA-graph-safe actuator state advancement belongs in Newton, but the unconditional publish should not become the actuator state contract: it changes eager double-buffer semantics and copies the complete actuator state after every step.

There are three relevant patterns:

1. Newton's simulation `State.assign()` records a final copy only at the boundary of an odd-length captured solver region. This pattern is documented in `newton/_src/sim/state.py` and used by examples such as `example_robot_policy.py`, `example_robot_anymal_c_walk.py`, and `example_nut_bolt_sdf.py`. Julia's suggestion follows this precedent. A private boundary-copy helper could avoid a public API change, but it would still copy once per odd replay.
2. Newton's actuator tests already cache graphs by the identities of their input state buffers. Capturing A-to-B and B-to-A variants and alternating them for odd-length regions needs at most two graphs, preserves ordinary swapping, and performs no state copy. Even-length regions need only one graph.
3. Device-side buffer indirection could provide a single zero-copy graph, but it would change public state shapes and controller kernels and is disproportionate for this fix.

A fixed CUDA graph cannot change its A/B array addresses between replays, so a single odd-length graph necessarily needs either a boundary copy or a second graph. For a 1.5 backport, I recommend the second option: remove `_publish_state()`, retain the existing `Actuator.step()` and `Actuator.State` API and double-buffer contract, and update the regression to alternate parity graphs for one- and three-step captures. This introduces no public API or state-layout change and avoids the new generic state-copy machinery entirely.

**Requirements:** the PID explicit/implicit graph defect is reproduced and fixed by the current branch, but eager state semantics and the claimed generic third-party coverage need adjustment.

**Standards:** the unsupported public semantic change and the two small test-style findings below remain.

**Validation:** reviewed exact head `ce31ce65`; focused graph tests passed 2/2; the full actuator suite passed 104/104 with CUDA and Torch; `git diff --check` passed. An independent no-swap PID probe reached the expected 0.12 in both state buffers. All hosted checks are currently passing.

I have not duplicated the two existing CodeRabbit threads. Both remain open, and the requested Torch documentation qualification remains valid. I recommend addressing the state-contract design issue before merge.

jcarius-nv:
@AntoineRichard, see my longer review above. It looks like this can be handled entirely by the graph-owning caller on 1.5.x: capture both state-buffer orientations (A->B and B->A), then alternate graphs after each odd-length replay. Could you confirm that this works for your downstream workload?
 
If so, we would not need to cherry-pick a runtime change to 1.5.x. On main, where additive API changes are acceptable, I would prefer following the existing State.assign() pattern: add Actuator.State.assign(), preserve normal eager swapping, and record one final state copy only for odd-length captured regions.

We should document both approaches: boundary assign() as the simpler single-graph pattern, and alternating parity graphs as the zero-copy option—which also works with 1.5.x today.

AntoineRichard:
Thanks both — I've taken the narrowing. Pushed as `c868ad8e`; the diff against `main` is now purely additive (243 insertions, 0 deletions) and `Actuator.step()` is byte-identical to `main`.

**P0, contract.** `_publish_state()` and the step-6 call are gone. The current/next double-buffer contract is restored exactly, so a caller can retain the pre-step state again and no actuator pays a per-step copy. You were right that my "eager behavior is unchanged" claim did not hold: the values the documented loop reads were unchanged, which is why the suite passed untouched, but `current_act_state` had become an output and that is a public contract change I should not have asserted away.

**P1, the generic-coverage claim.** Your counterexample reproduces exactly as you described — an undecorated `Controller.State` subclass is `is_dataclass() == True` with an empty `fields()`, so the old helper copied nothing. `Actuator.State.assign()` walks `__dict__` like `newton.State.assign()` does, which covers it, and it raises on a component or array presence mismatch rather than skipping. Both are now tested (`TestActuatorStateAssign`), including your `[0]`-stays-`[0]` case.

**P2 and standards.** Dropped the `wp.synchronize_device()` before `.numpy()`, and both graph tests now carry imperative docstrings.

**CodeRabbit.** The Torch qualification was valid and is in: `is_graphable()` describes the components, not the captured region, and `ControllerNeuralLSTM` on a Torch checkpoint keeps hidden and cell state in Torch tensors the controller rebinds host-side, so that state never advances inside a graph at all. The changelog is one imperative entry and moved to `added`. The `src is None` suggestion is moot now that the helper is gone.

**Your question about the downstream workload.** Confirmed, with a caveat worth having on the record. Parity-alternating graphs do work for us, but they are the worse fit, because we already use the boundary-assign pattern for simulation state — `NewtonManager._run_solver_substeps`:

```python
need_copy_on_last = cfg is not None and cls._num_substeps % 2 == 1
...
if need_copy_on_last and i == cls._num_substeps - 1:
    cls._state_0.assign(cls._state_1)
else:
```

We capture collision + actuators + every solver substep + sensors as one graph, so alternating parities means holding two copies of the entire physics graph to fix an actuator-state parity problem, and it would leave two different mechanisms for the same class of bug inside one loop. `Actuator.State.assign()` drops in beside the line above and costs one copy per captured region. So: no 1.5.x runtime cherry-pick needed either way, and on `main` the `assign()` route is the one we would actually adopt.

I verified both routes rather than assuming:

- the parity route passes against stock `main` with zero library changes (`def assign` count 0 in `actuator.py`), which is the property that makes it 1.5.x-compatible;
- the boundary-assign route fails on stock `main` with `AttributeError: 'State' object has no attribute 'assign'`, so it genuinely needs the new API.

Both are documented in `Stateful Actuators`, with the parity example noted as working on releases without `assign()`.

**Validation:** `test_actuators` 106/106, `test_mujoco_general_actuators` 36/36, Sphinx doctest 88 tests 0 failures under `-W`, `uvx pre-commit run -a` clean. RTX 5000 Ada, CUDA 12.9 / driver 13.0, Warp 1.17.0.dev20260807.

One note on scope: with the runtime fix removed, #4098 is resolved by API plus documentation rather than by changed default behavior — resolution (2) in the issue rather than (1). That is your call to make and I think it is the right one; a fixed graph cannot re-point its buffers, so forcing a per-step copy on everyone to spare callers a boundary line was the wrong default. Happy to update the issue's framing if you'd like it to say that explicitly.

newton/_src/actuators/actuator.py
```diff
@@ -32,6 +32,44 @@ def _scatter_add_kernel(
...
+            continue
+        if val_dst is None or val_src is None:
+            raise ValueError(f"Cannot assign '{name}.{attr}': present in one state and missing in the other.")
+        array_dst = isinstance(val_dst, wp.array)
+        if array_dst != isinstance(val_src, wp.array):
+            raise ValueError(f"Cannot assign '{name}.{attr}': a Warp array in one state and not in the other.")
+        if array_dst:
+            val_dst.assign(val_src)
+        else:
+            setattr(dst, attr, val_src)
```
jcarius-nv: **P1 — Define or enforce the copy contract for custom controller state.**

This only walks direct `__dict__` attributes, copies direct Warp arrays, and rebinds everything else. Consequently, valid custom `Controller.State` layouts can silently behave incorrectly: slotted Warp arrays are never visited; nested Warp arrays are rebound on the host, so their copy is not recorded by CUDA graph capture; and non-Warp mutable values, including Torch tensors, alias rather than copy. I reproduced all three at this head: a nested state reached `1` rather than `3` after three odd-length replays, a slotted array remained `0`, and resetting the destination Torch tensor also changed the source. The built-in flat Warp states are covered, but the public `Actuator.State.assign()` contract currently reads as a general copy.

Please either provide an explicit overridable assignment seam for component state, or reject unsupported layouts rather than silently rebinding them, and add coverage for the supported custom-state contract.

jcarius-nv (commented):
Thanks, Antoine, for the thoughtful update and for validating both caller-side approaches. Removing the unconditional publication restores the existing double-buffer contract, and the confirmation of the 1.5.x-compatible parity-graph route is especially helpful. I found one remaining issue with the generic `assign()` contract below.

newton/_src/actuators/actuator.py
```diff
@@ -42,6 +42,78 @@ def _scatter_add_kernel(
...
+
+    dst_is_warp = isinstance(dst, wp.array)
+    src_is_warp = isinstance(src, wp.array)
+    if dst_is_warp or src_is_warp:
+        if not (dst_is_warp and src_is_warp):
+            raise ValueError(f"Cannot assign '{name}': a Warp array in one state and not in the other.")
+        dst.assign(src)
+        return
+
+    dst_is_torch = type(dst).__module__.startswith("torch")
```
jvonmuralt, line 60: do you think we should still support assign on torch arrays, even though torch actuators are not graphable?

jcarius-nv: Good question. My inclination is to retain Torch support because assign() is a general public state-copy operation and as a user I would expect it works with all built-in actuator states.

jvonmuralt (approved):

newton/_src/actuators/actuator.py
jvonmuralt: ok, agree, I dont have a better idea

PR #4105, opened 2026-08-31
Port `ControllerOperationalSpace` into `newton.controllers` module.

### LINKED ISSUE #4122: [REQ] Implement `ControllerOperationalSpace`
### Description

Migrating the operational space controller (currently implemented in Isaac Lab) into a heterogeneous and graphable warp implementation in `newton.controllers` module.

### Motivation / Use Case

See #3548 

### Alternatives Considered

_No response_

jeff-hough:
## Description

Introduces a heterogeneous Operational Space Controller, i.e. see issue #4122, part of on-going task #3548) 

## Checklist

## Test plan

     ```
     ```

## New feature / API change

```python
import newton
import warp as wp
from newton.controllers import ControllerOperationalSpace

device = wp.get_device()

scene = newton.ModelBuilder()
for i in range(512):

control = model.control()

# -- pick what to control --
# Here every panda articulation (not the conveyor), and within it every panda
# arm joint (not the fingers) driving a single tool site per robot --
# fr3_hand_tcp.

controller = ControllerOperationalSpace(
    model,
    articulations="panda_*",
    joints=[f"panda_joint{k}" for k in range(1, 8)],  # don't select the finger joints.
    tool="fr3_hand_tcp",
    motion_stiffness=200.0,  # None to drive the gains live via inputs.motion_stiffness
    motion_damping=2.0 * 200.0**0.5,  # None to drive the gains live via inputs.motion_damping
    use_gravity_compensation=True,
    use_inertia_decoupling=True,
    # Press along local Z (index 2); the other five task axes stay
    # motion-controlled -- a hybrid force/motion task, e.g. each tool
    # pressing into a surface while its (x, y) position and orientation
    # are still commanded.
    use_wrench_feedforward=True,
    use_wrench_feedback=True,
    wrench_stiffness=0.5,
    motion_selection_axes=wp.spatial_vector(1.0, 1.0, 0.0, 1.0, 1.0, 1.0),
    wrench_selection_axes=wp.spatial_vector(0.0, 0.0, 1.0, 0.0, 0.0, 0.0),
)

inputs = controller.input()
outputs = controller.output()

# joint_q/joint_qd are whole-model -- an unselected joint still moves the
# bodies downstream of it, and so the mass matrix and gravity terms of the
# joints that are controlled. They are bound in the loop below, since the
# physics loop double-buffers. Every other port is per-robot, shape
# [controlled_robot_count] -- one target per tool, not per DOF.
inputs.desired_twist_operational.assign(
    [wp.spatial_vector(0.0, 0.0, 0.0, 0.0, 0.0, 0.0)] * controller.controlled_robot_count
)
inputs.desired_wrench_world.assign(
    [wp.spatial_vector(0.0, 0.0, -20.0, 0.0, 0.0, 0.0)] * controller.controlled_robot_count
)

# An indexed view scatters the compact torque command straight into the sim,
# leaving the conveyor's and the fingers' entries in control.joint_f untouched.
outputs.joint_f = control.joint_f[controller.qd_start]

# -- sim loop --
for _ in range(1000):
    inputs.desired_tool_pose_operational.assign(desired_pose)  # per-robot, shape [controlled_robot_count]

    for _ in range(sim_substeps):
        state_0.clear_forces()

```

## Summary by CodeRabbit

* **New Features**
  * Added model-based and model-free operational-space impedance controllers.
  * Added task-space motion, wrench, gravity compensation, inertia decoupling, null-space control, and configurable operational frames.
  * Added per-axis motion and force selection.
  * Added public access to the new controllers.

* **Examples & Documentation**
  * Added an interactive hybrid force/motion example featuring Franka Panda and UR10 robots.
  * Updated controller API documentation and example listings.

* **Improvements**
  * Improved numerical stability and support for heterogeneous robots, graph execution, and gradient-aware workflows.

* **Tests**
  * Added automated coverage for the new controller example.


jeff-hough:
@coderabbitai review

jeff-hough:
@coderabbitai review

eric-heiden:
Nice work on adding Operational Space Control into Newton!  The two controllers are carefully documented and they build on the existing `eval_fk`, `eval_jacobian`, and `eval_mass_matrix` helpers instead of reimplementing them. The comments below are mostly small: one Jacobian that ends up computed twice per step on the default settings, one dtype missing from the indexed-view gather table, and a few naming and docstring cleanups.

newton/_src/controllers/impl/operational_space/model_based.py
```diff
@@ -0,0 +1,898 @@
...
+                self._controlled_dofs_per_robot,
+            ],
+            outputs=[self._jacobian_tool_world],
+            device=self._device,
+        )
+
+        if self._use_inertia:
+            eval_mass_matrix(
+                self._model, self._model_state, H=self._model_mass_matrix, mask=self._controlled_robot_mask
+            )
```
eric-heiden, line 897: 🔵 `step()` already fills `self._jacobian_com_world` a few lines above with `eval_jacobian(self._model, self._model_state, J=self._jacobian_com_world, mask=self._controlled_robot_mask)`, but this call does not forward it.

With `J=None`, `eval_mass_matrix` takes its `if J is None:` branch (`newton/_src/sim/articulation.py:1661-1669`), allocates a fresh `(articulation_count, max_links * 6, max_dofs)` device array, and runs a second full `eval_jacobian` under the same mask. That happens every step on the default `use_inertia_decoupling=True` path, so every graph-captured frame pays a duplicate whole-model Jacobian pass plus a per-step device allocation inside the capture window. The results are correct either way, this is purely wasted work on the hot path.

The controller's own buffer is allocated with exactly the shape `eval_mass_matrix` would allocate, both calls use the same mask, and `eval_articulation_mass_matrix` only reads `J` for masked articulations, so forwarding the buffer is safe:

```suggestion
            eval_mass_matrix(
                self._model,
                self._model_state,
                H=self._model_mass_matrix,
                J=self._jacobian_com_world,
                mask=self._controlled_robot_mask,
            )
```

newton/_src/controllers/impl/_common.py
```diff
@@ -109,9 +132,21 @@ def _scatter_port_kernel(
...
+# dtype -> (rank -> gather kernel), the set of dtype/rank combinations any
+# controller's ports currently use. Extend this table, not _read_port itself,
+# when a controller needs a new port dtype or rank. Every rank-1 dtype shares
+# _gather_rank1_port_kernel: it's generic over dtype (Any), so Warp compiles
+# one concrete kernel per dtype the table actually uses, from a single body.
+_GATHER_KERNELS_BY_DTYPE_AND_RANK = {
+    wp.float32: {1: _gather_rank1_port_kernel, 3: _gather_mass_matrix_port_kernel},
+    wp.transform: {1: _gather_rank1_port_kernel},
+    wp.spatial_vector: {1: _gather_rank1_port_kernel},
+}
```
eric-heiden, line 145: 🔵 This table covers `wp.float32`, `wp.transform`, and `wp.spatial_vector`, but not `wp.quat`, and `step()` routes two quat ports through `_read_port`: `inputs.linear_selection_frame_operational` (`operational_space/model_free.py:1467`) and `inputs.angular_selection_frame_operational` (`:1482`).

Both ports are validated with `allow_indexed=True`, both are typed `wp.array[wp.quat] | wp.indexedarray[wp.quat]` in the `Inputs` schema, and the class docstring states that "Every port, of any dtype, may be bound either to a plain array or to an indexed view of a simulation-sized array." So a fleet user with `use_wrench_feedforward=True` who keeps per-robot selection frames in one simulation-sized array and binds `ins.linear_selection_frame_operational = sim_frames[selection]` passes validation and then fails at runtime with `TypeError: _read_port has no gather kernel for a 1-D indexed array of dtype <class 'warp.types.quat'>`.

`_gather_rank1_port_kernel` is already generic over dtype, so this is a missing table row rather than a missing kernel:

```suggestion
_GATHER_KERNELS_BY_DTYPE_AND_RANK = {
    wp.float32: {1: _gather_rank1_port_kernel, 3: _gather_mass_matrix_port_kernel},
    wp.transform: {1: _gather_rank1_port_kernel},
    wp.spatial_vector: {1: _gather_rank1_port_kernel},
    wp.quat: {1: _gather_rank1_port_kernel},
}
```

The gap is also untested: `test_transform_and_spatial_vector_inputs_accept_indexed_views` only binds transform and spatial_vector ports. Worth extending it to bind the two selection-frame ports to indexed views as well, and renaming it to match.

newton/_src/controllers/impl/operational_space/model_based.py
```diff
@@ -0,0 +1,898 @@
...
+        joint_f: wp.array[wp.float32] | wp.indexedarray[wp.float32]
+        """Joint torque command [N or N·m], compact, shape [total_controlled_dofs]."""
+
+    def __init__(
+        self,
+        model: Model,
+        *,
+        articulations: list[int | str | re.Pattern[str]] | str | re.Pattern[str] | None = None,
+        joints: list[int | str | re.Pattern[str]] | str | re.Pattern[str] | None = None,
+        tool: list[int | str | re.Pattern[str]] | str | re.Pattern[str],
```
eric-heiden: 🔵 `tool` is the only multi-entry selector in this signature that is singular, and it is the one place that drops the `site` noun used everywhere else for this primitive.

It sits directly beside `articulations` and `joints`, it accepts the same shape (a list of indexes or label patterns), and every error it raises talks about sites, for example `"tool matches no site on articulation {art}."`. Newton's other site selectors keep both the plural and the noun: `SensorIMU(model, sites=...)` in `newton/_src/sensors/sensor_imu.py` and `SensorFrameTransform(model, reference_sites=...)` in `newton/_src/sensors/sensor_frame_transform.py`. The docstring here has to compensate by spelling out "Site index(es) or label pattern(s)".

A singular name reads as one value and invites `tool=3` as the only supported form, when the argument in fact covers a whole heterogeneous fleet. That is exactly what the new example relies on, with `tool="tool_site"` resolving one site per robot across two different arms.

`tool_sites` would line this up with the sensors and with its two neighbours. Both classes are newly public in this PR, so there is no caller to migrate, and the existing error messages already read correctly against that name. The docstring and the entry in `docs/api/newton_controllers.rst` would need the same update.

newton/_src/controllers/impl/operational_space/model_based.py
```diff
@@ -0,0 +1,898 @@
...
+
+    @property
+    def qd_start(self) -> wp.array[wp.int32]:
+        """Model DOF index of each controlled joint, shape [total_controlled_dofs]."""
+        return self._qd_idx
+
+    @property
+    def tool_body(self) -> wp.array[wp.int32]:
+        """Body index of each controlled robot's tool site, shape [controlled_robot_count]."""
+        return self._tool_body
```
eric-heiden, line 675: ⚪ `tool_body` is the only piece of the resolved tool frame this class publishes, and the raw body index is not enough on its own.

The controller resolves the tool pose and twist from the tool site on every `step()` and stores them in `_tool_pose_world` and `_tool_twist_world`, while the site's body-local offset stays private in `_tool_transform_body`. A user who selects the tool by label pattern (`tool="ee_site"`, the documented usage) has no route to that offset, so recomposing the tool frame means re-resolving the site against `model.shape_label` and `model.shape_transform` by hand and keeping that second copy in sync with what the controller already computed.

Logging, plotting, or asserting on task-space tracking error is the natural thing to do with a task-space controller, and it runs into this. The data already exists and is already correct, it just is not reachable.

Read-only properties beside `tool_body` would close the gap, and exposing the resolved site offset as `tool_transform_body` would help too:

```suggestion
    @property
    def tool_body(self) -> wp.array[wp.int32]:
        """Body index of each controlled robot's tool site, shape [controlled_robot_count]."""
        return self._tool_body

    @property
    def tool_pose_world(self) -> wp.array[wp.transform]:
        """World pose of each controlled robot's tool site as of the latest ``step()``, shape [controlled_robot_count]."""
        return self._tool_pose_world

    @property
    def tool_twist_world(self) -> wp.array[wp.spatial_vector]:
        """World twist of each controlled robot's tool site as of the latest ``step()``, shape [controlled_robot_count]."""
        return self._tool_twist_world
```

newton/_src/controllers/impl/operational_space/model_based.py
```diff
@@ -0,0 +1,898 @@
...
+        use_inertia_decoupling: Premultiply the task-space spring-damper
+            term by Lambda, the operational-space mass matrix, computed
+            each step from ``model``'s own mass matrix (via
+            :func:`newton.eval_mass_matrix`) and the resolved tool Jacobian.
+        use_partial_inertia_decoupling: Compute Lambda ignoring the coupling
+            between translational and rotational inertia. Only meaningful
+            when ``use_inertia_decoupling=True``.
+        use_gravity_compensation: Add the model's own gravity generalized
+            forces — computed each step via
+            :func:`~newton.sim.inverse_dynamics.eval_inverse_dynamics_passive`
```
eric-heiden: ⚪ This line points at `newton.sim.inverse_dynamics.eval_inverse_dynamics_passive`, which is not a real path: `import newton.sim` raises `ModuleNotFoundError`. The public path is `newton.eval_inverse_dynamics_passive`, which is what `docs/api/newton.rst` lists and what the neighbouring references in this same docstring already use (`newton.eval_fk`, `newton.eval_jacobian`, `newton.eval_mass_matrix`).

`docs/conf.py` sets `nitpicky = False` and the docs workflow does not pass `-W`, so the build stays green. The reference simply renders as unlinked plain text in the published API reference, which readers of this newly public class will land on.

```suggestion
            :func:`~newton.eval_inverse_dynamics_passive`
```

newton/_src/controllers/impl/operational_space/model_free.py
```diff
@@ -0,0 +1,1821 @@
...
+            and that ``motion_stiffness``/``motion_damping``/
+            ``wrench_stiffness`` are interpreted in. Need not coincide with
+            the tool's own current orientation (e.g. a frame aligned to a
+            work surface, tracked independently of how the tool is
+            oriented). Pass a ``wp.transform`` to apply the same fixed pose
+            to every robot, an array of shape [controlled_robot_count] to
+            set them individually (fixed for the controller's lifetime), or
+            ``None`` to read ``inputs.operational_frame_pose_world`` each
+            step for a time-varying frame. Defaults to identity (coincides
+            with world frame).
```
eric-heiden: ⚪ This Args block documents `use_inertia_decoupling` before `operational_frame_pose_world`, but the constructor takes them the other way round: `motion_stiffness`, `motion_damping`, `operational_frame_pose_world`, `use_inertia_decoupling`, `use_partial_inertia_decoupling`.

`ControllerOperationalSpace` documents the identical argument set in true signature order, so the two classes list the same parameters in two different orders in the rendered API reference, and only one of them matches the code. Both classes are newly public here, so both Args blocks render side by side, and the class docstrings actively invite the comparison through their "See also" cross-reference. Google-style docstrings list args in signature order.

Moving the `use_inertia_decoupling` entry down so it sits after `operational_frame_pose_world` and before `use_partial_inertia_decoupling` matches both the signature and the model-based sibling.

eric-heiden:
**Consider passing `J` to `eval_mass_matrix` instead of letting it recompute one**

`eval_mass_matrix` accepts optional `J` and `body_I_s` arguments. When `J` is omitted it allocates a fresh `(articulation_count, max_links * 6, max_dofs)` device array and runs a full `eval_jacobian` internally, on every call. Both model-based controllers currently omit both arguments.

`newton/_src/actuators/response_oracle.py` already uses the pattern, with the buffers allocated once in `__init__`:

```python
eval_jacobian(model, fk_state, J=self._J, joint_S_s=self._joint_S_s)
eval_mass_matrix(model, fk_state, H=self._H, J=self._J, body_I_s=self._body_I_s)
```

`step()` already computes exactly the Jacobian that `eval_mass_matrix` is about to rebuild. Line 815 fills `self._jacobian_com_world`, and line 849 then drops it:

```python
eval_jacobian(self._model, self._model_state, J=self._jacobian_com_world, mask=self._controlled_robot_mask)
...
eval_mass_matrix(self._model, self._model_state, H=self._model_mass_matrix, mask=self._controlled_robot_mask)
```

Both calls use the same `mask`, and the controller's buffer already has the exact shape `eval_mass_matrix` would allocate. This happens on the default configuration (`use_inertia_decoupling=True`), so it is on the hot path for every user who does not turn inertia decoupling off, and it puts a device allocation inside the graph capture window of the new example.

Forwarding the buffer is enough:

```python
eval_mass_matrix(
    self._model,
    self._model_state,
    H=self._model_mass_matrix,
    J=self._jacobian_com_world,
    mask=self._controlled_robot_mask,
)
```

This is safe at that point in `step()`. `eval_mass_matrix` needs the COM-referenced Jacobian, and the tool shift in between writes into a separate output buffer (`self._jacobian_tool_world`), so `self._jacobian_com_world` is still the untouched COM-referenced version when line 849 runs.

newton/_src/controllers/impl/operational_space/model_based.py
```diff
@@ -0,0 +1,917 @@
...
+        if not np.array_equal(owning_joint, owning_joint_qd):
+            mismatched = int(np.flatnonzero(owning_joint != owning_joint_qd)[0])
+            raise ValueError(
+                f"joint_selection.q_start and joint_selection.qd_start disagree at entry {mismatched}: "
+                f"coordinate {int(q_idx_np[mismatched])} belongs to joint {int(owning_joint[mismatched])} "
+                f"but DOF {int(qd_idx_np[mismatched])} belongs to joint {int(owning_joint_qd[mismatched])}. "
+                f"Did you swap the two arrays?"
+            )
+
+        owning_robot = model.joint_articulation.numpy()[owning_joint]
```
eric-heiden, line 404: 🟠 This validation block never checks that each selected joint spans exactly one coordinate and one DOF, so multi-DOF joints are silently accepted.

The class docstring promises that rejection twice: "Only joints spanning a single coordinate and a single DOF can be controlled", and, for `joints`, that a joint named explicitly "still raises `ValueError` if it is not 1-coordinate/1-DOF". `select_joints` deliberately leaves the gap open for the controller to close, and `ControllerJointImpedance` closes it with this exact check in this exact slot of the same block.

On a floating-base arm (one free joint plus six revolutes, 12 DOFs), `ControllerOperationalSpace(model, joints=".*", ...)` is accepted and controls only 7 DOFs. `JointSelection` stores one entry per joint, its *starting* index, so the free joint's other 5 DOFs are dropped: their Jacobian columns still carry full-magnitude entries, and about 55 N of gravity load on them goes uncompensated even with `use_gravity_compensation=True`. No index goes out of range, so nothing raises, and the task-space law is just solved against a truncated Jacobian and mass matrix. `ControllerJointImpedance` raises on the identical call.

```suggestion

        # A joint is controllable when it spans exactly one coordinate and one
        # DOF. Derived from the model's own spans.
        joint_type_np = model.joint_type.numpy()
        coord_span = np.diff(model.joint_q_start.numpy())[owning_joint]
        dof_span = np.diff(model.joint_qd_start.numpy())[owning_joint]
        unsupported = sorted(
            {
                (int(j), JointType(joint_type_np[j]).name)
                for j, coords, dofs in zip(owning_joint, coord_span, dof_span, strict=True)
                if coords != 1 or dofs != 1
            }
        )
        if unsupported:
            raise ValueError(
                f"ControllerOperationalSpace only supports controlling joints that span a single "
                f"coordinate and a single DOF; joint_selection addresses unsupported joints: {unsupported}"
            )

        owning_robot = model.joint_articulation.numpy()[owning_joint]
```

This also needs `from newton import JointType` alongside the other top-level imports. A regression test that names a multi-DOF joint explicitly would be worth adding too, since none of the current tests in `test_controllers_operational_space.py` cover a non-1x1 joint.

eric-heiden (dismissed):
Thanks, LGTM!

newton/_src/controllers/impl/operational_space/model_based.py
```diff
@@ -0,0 +1,938 @@
...
+        if self._use_inertia:
+            eval_mass_matrix(
+                self._model,
+                self._model_state,
+                H=self._model_mass_matrix,
+                J=self._jacobian_com_world,
+                mask=self._controlled_robot_mask,
+            )
+            wp.launch(
+                _gather_mass_matrix_blocks_kernel,
```
jcarius-nv, line 899: The gathered principal block `M_CC` is the effective controlled-joint inertia only when omitted DOFs are fixed or dynamically decoupled. If an omitted movable DOF is coupled and free, the controlled response is governed by the Schur complement (equivalently `(M^-1)_CC`), so inertia decoupling and the dynamically consistent null-space calculation do not represent the full system. Since `joints` otherwise supports arbitrary subsets, could the public documentation explicitly state that `use_inertia_decoupling=True` does not support subsets that omit free, dynamically coupled DOFs?

newton/_src/controllers/impl/operational_space/model_free.py
```diff
@@ -0,0 +1,1821 @@
...
+            wp.launch(
+                _add_term_kernel,
+                dim=self._total_controlled_dofs,
+                inputs=[self._grav_buf],
+                outputs=[self._tau_buf],
+                device=self._device,
+            )
+
+        # A view needs the scatter kernel (wp.copy isn't graph-capture-safe for a non-contiguous target).
+        if isinstance(outputs.joint_f, wp.indexedarray):
```
jcarius-nv, line 1826: `ControllerOperationalSpaceModelFree` says that every port's shape and device are validated on each direct `step()`, but `outputs.joint_f` reaches this copy/scatter path without validation. For example, a two-element output bound to a one-DOF controller is accepted and only its first element is overwritten. This leaves the documented validation contract untrue and lets short, wrong-device, or wrong-dtype outputs fall through to lower-level copy/kernel behavior.

newton/_src/controllers/impl/operational_space/model_free.py
```diff
@@ -0,0 +1,1821 @@
...
+        [controlled_robot_count, ..., max_controlled_dofs]. Optional fields
+        are ``None`` when the corresponding feature is disabled at
+        construction.
+        """
+
+        tool_pose_world: wp.array[wp.transform] | wp.indexedarray[wp.transform]
+        """Current world pose of the tool frame, shape [controlled_robot_count]."""
+        tool_twist_world: wp.array[wp.spatial_vector] | wp.indexedarray[wp.spatial_vector]
+        """Current tool twist (linear, angular) in world coordinates [m/s, rad/s], shape [controlled_robot_count]."""
+        jacobian_tool_world: wp.array3d[wp.float32] | wp.indexedarray(dtype=wp.float32, ndim=3)
```
jcarius-nv, line 481: Several new public physical ports omit the SI-unit information required by `CODING_GUIDELINES.rst`. In particular, the transform ports should describe translation/rotation components, and the Jacobian should document its row- and DOF-dependent units. The same omissions appear on the corresponding model-based ports/properties.

jcarius-nv (commented):
Thanks for the substantial controller port and for working through the earlier feedback. I found three API-contract details worth addressing or documenting; none changes my overall view that the controller belongs in Newton.

newton/_src/controllers/impl/operational_space/model_based.py
jeff-hough: Thank you for noting this! I have added to the docstring, though I am not going as far as saying it is "not supported", given that I don't reject this outright. Instead, there is a note about correctness in this case. It can often be a useful approximation (i.e. if one doesn't wish to control the finger joints at the end of the robot).

jcarius-nv (approved):

PR #4119, opened 2026-09-01
Update MuJoCo and MuJoCo Warp to 3.12

camevor:
## Description

Upgrade `mujoco` and `mujoco-warp` to 3.12.0.

This PR also updates the mapping to MuJoCo's actuator bias/dynamics/gain enums, which had gotten stale, and replaces the old silent fallback-to-0 behavior with a `RuntimeWarning` naming the offending prim/actuator, the unsupported value, and the supported values when an MJCF or USD actuator attribute can't be resolved.

3.12.0 also changed how `mocap` bodies get their own weld id; the check now uses `body_dofnum` instead of `body_weldid` alone.

Also exposes a new `njmax_nnz` parameter on `SolverMuJoCo` to control sparse Jacobian nonzero capacity per world. In particular, this can be needed to work around a mujoco-warp 3.12.0 sizing issue seen on the G1-with-handsUSD dynamics comparison test. If unset, `njmax_nnz` is set to a conservative value to work around this; this workaround should be removed in the future.

## Checklist

## Test plan
```
uv run --extra dev -m newton.tests -k test_menagerie_mujoco
uv run --extra dev -m newton.tests -k test_menagerie_usd_mujoco
```

## Summary by CodeRabbit

* **New Features**
  * Added `njmax_nnz` configuration for sparse Jacobian capacity, with automatic sizing when unspecified.
* **Bug Fixes**
  * Improved filtering of contacts between immovable bodies.
  * Corrected actuator type mappings for MuJoCo compatibility.
  * Unsupported actuator settings now generate contextual warnings.
* **Documentation**
  * Documented sparse Jacobian capacity tuning and overflow handling.
  * Added guidance to reinstall the `sim` extra when upgrading from MuJoCo 3.11.
* **Chores**
  * Upgraded MuJoCo and MuJoCo Warp support to version 3.12.0.
* **Tests**
  * Expanded coverage for sparse Jacobian settings, actuator mappings, and contact filtering.


vreutskyy (dismissed):
Re-reviewed the updated head. The previous ARM Apollo regression is resolved, the affected local tests pass, and all CI checks, including GPU tests and benchmarks, are green.

jcarius-nv (commented):
Reviewed `e93cf32b..e880a6a2` with independent Fit, Requirements, and Standards passes.

**Fit:** Fits. The MuJoCo 3.12 integration, actuator ordinal repair, mocap contact adaptation, and sparse-capacity control belong in Newton's MuJoCo boundary.

**Requirements:** Pass. I found no demonstrated behavioral defect. A four-active-limit probe reproduced the upstream 3.12.0 underestimation (actual NNZ 4, upstream estimate 1) and confirmed this change resolves capacity to 5 with no overflow.

**Standards:** Does not pass yet. The new public `njmax_nnz` contract needs user-facing capacity and tuning documentation. The existing resolved numeric-ordinal thread is only partially addressed: the code fix is present, but its requested unsupported-numeric regression test is still absent. The existing unresolved migration-guidance thread also remains valid.

Local validation passed:
- `test_mujoco_solver`: 291 tests (7 skipped)
- `test_menagerie_mujoco`: 51 tests (48 skipped)
- `test_menagerie_usd_mujoco`: 44 tests (62 skipped)
- diff, lock, Ruff, and Towncrier checks

All required GitHub checks are green, including GPU tests; GPU benchmarks are also green. I would address the two P1 Standards gaps before approval.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -3716,6 +3740,7 @@ def __init__(
             model: The model to be simulated.
             separate_worlds: If True, each Newton world is mapped to a separate MuJoCo world. Defaults to `not use_mujoco_cpu`.
             njmax: Maximum number of constraints per world. If None, a default value is estimated from the initial state. Note that the larger of the user-provided value or the default value is used.
+            njmax_nnz: Sparse Jacobian nonzero capacity per world. If None, estimates it from the model.
```
jcarius-nv: P1: Please document the public `njmax_nnz` contract beyond this one-line parameter description. The MuJoCo capacity guide still covers only `nconmax` and `njmax`, so users cannot tell that this controls MuJoCo Warp sparse-Jacobian storage, how it relates to `njmax` and `nv`, which overflow signal to monitor, or that memory scales per world. `_convert_to_mjc()` also says to keep `docs/solvers/mujoco.rst` synchronized, but no documentation file is updated in this PR; the PR template's API usage example is also absent.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -7787,6 +7822,36 @@ def add_body_equality(i: int):
...
+                    for limited, joint_type in zip(self.mj_model.jnt_limited, self.mj_model.jnt_type, strict=True):
+                        if not limited:
+                            continue
+                        joint_type_value = int(joint_type)
+                        if joint_type_value == mujoco.mjtJoint.mjJNT_BALL:
+                            joint_limit_nnz += 3
+                        elif joint_type_value in (mujoco.mjtJoint.mjJNT_SLIDE, mujoco.mjtJoint.mjJNT_HINGE):
+                            joint_limit_nnz += 1
+
+                    # work around buffer under-sizing until the fix is released (mjwarp #1630)
```
jcarius-nv, line 7856: P2: This workaround needs a bounded retirement path. `google-deepmind/mujoco_warp#1630` is already merged, while `mujoco-warp~=3.12.0` will admit a later fixed 3.12.x. Once that release is selected, adding `joint_limit_nnz` here will conservatively double-count joint-limit storage in every world. The dense cap prevents an invalid allocation, but the workaround can silently erode sparse-memory savings and keeps Newton coupled to private `_src.io` helpers. Please record a version or capability boundary, or an explicit removal tracker.

newton/_src/solvers/mujoco/solver_mujoco.py
camevor: Updated the solver tuning documentation in a236623e5. `docs/solvers/mujoco.rst` does not need an update, as adding this sort of documentation there would be inconsistent.

newton/_src/solvers/mujoco/solver_mujoco.py
camevor: Created #4127 to track this. Adding a capability boundary here seems over the top; the memory savings are very small and unlikely to translate into any runtime perf improvement.

newton/_src/solvers/mujoco/solver_mujoco.py
```diff
@@ -3769,6 +3794,12 @@ def __init__(
             raise ValueError(
                 "enable_sleeping=True requires use_mujoco_contacts=True so contacts can wake sleeping bodies."
             )
+        if njmax_nnz is not None:
+            if isinstance(njmax_nnz, bool) or not isinstance(njmax_nnz, (int, np.integer)):
+                raise TypeError(f"njmax_nnz must be an integer or None, got {type(njmax_nnz).__name__}.")
+            if njmax_nnz < 0:
```
jcarius-nv, line 3798: P1: `njmax_nnz` accepts explicit capacities that are too small for the initial sparse Jacobian. The constructor only rejects values below zero, so `njmax_nnz=0` with one active hinge limit constructs with `nefc=1`, `efc_J_rownnz=[1]`, capacity 0, and no overflow bit; the first `step()` then reproduces CUDA illegal-memory-access error 700. An undersized positive capacity instead fails inside `mujoco_warp.put_data()` with a broadcast error. Could we validate or clamp explicit capacities against the initial required NNZ and add zero/undersized regression coverage? @camevor, could you evaluate whether this should block this PR?

camevor: Added a check in a46d96e60 to avoid `njmax_nnz` that is too small for `put_data`.

camevor:
Fixed the documentation inaccuracies in 38459f504.

Agree that the temporary workaround means that technically Newton is sizing `njmax_nnz`; however, it fully delegates to mjwarp's own sizing routine, simply addings a small number to avoid the upstream sizing bug. Explaining that here would be more confusing than helpful and outdated very soon. Note that the number of joint limits is very small compared to the number of potential contacts if those are enabled.

eric-heiden (approved):
Thanks, LGTM!

PR #4125, opened 2026-09-02
Improve Kamino DVI solver robustness and performance

nvtw:
## Description

Improve Kamino DVI convergence, stability, and sparse-contact scheduling while preserving the existing alternating contact solve as the default path.

This PR:

- adds an experimental opt-in Schur-complement path for dense and sparse coupled bilateral/contact solves;
- keeps the historical alternating path, relaxation, and contact warm-start defaults when Schur is disabled;
- avoids allocating the large sparse Schur response matrices unless Schur is enabled;
- enables the tuned Schur path explicitly in the G1 example;
- improves bounded-joint-friction handling and sparse CUDA contact scheduling; and
- fixes ordered CUDA contact mapping when bounded joint rows are present. Previously, large contact problems could overwrite joint-limit topology because the bounded-row offset was omitted.

### Analysis material

The comparison material currently contributes 589 lines of executable analysis code, 1,267 lines of generated HTML/Markdown/JSON results, and one 103,641-byte PNG. It is included for reproducibility while the Schur and alternating paths are reviewed. Should this material remain in this MR, or should the generated artifacts and/or scripts move to the PR discussion or a follow-up?

## Checklist

## Test plan

```text
uv run --extra dev -m newton.tests -k test_kamino_solvers_dvi
uv run --extra dev -m newton.tests -k example_robot_g1_kamino
uv run scripts/evaluate_dvi_block_convergence.py
uv run scripts/evaluate_kamino_dvi_coupling.py --method both --iterations 8 --sweeps 2 --omega 1.2 --frames 240 --settle-frames 60 --timing-frames 20 --timing-trials 5
uvx pre-commit run -a
```

All 57 DVI tests passed on CUDA across the completed suite and targeted remainder runs. The registered four-world, 500-frame Kamino G1 example also passed.

The controlled block-convergence output reproduced the committed reference exactly. In the G1 comparison, Schur retained substantially lower drift and complementarity residuals than the alternating path at comparable or better frame time. After making Schur opt-in, the exact 240-frame reference run improved lateral drift from 0.150 mm to 0.078 mm and median post-contact complementarity residual from `9.44e-6` to `7.28e-6`. A longer timing run measured a 68.03 ms minimum versus 68.02 ms before the merge-readiness changes.

## Bug fix

**Steps to reproduce:**

1. Configure sparse DVI on CUDA with ordered contact scheduling, bounded joint rows, joint limits, and contact capacity of at least 4096.
2. Prepare the sparse inequality schedule.
3. Observe contacts overwrite or misorder joint-limit topology because their indices omit the bounded-row offset.

**Minimal reproduction:**

```python
import subprocess

subprocess.run(
    ["uv", "run", "--extra", "dev", "-m", "newton.tests", "-k", "test_03g3_dvi_ordered_contacts_follow_bounded_rows_and_limits"],
    check=True,
)
```

## New feature / API change

```python
import newton

config = newton.solvers.SolverKamino.Config.from_model(
    model,
    dynamics_solver="dvi",
    sparse_dynamics=True,
    sparse_jacobian=True,
)
config.dvi.use_schur_complement = True
solver = newton.solvers.SolverKamino(model, config=config)
```

## Summary by CodeRabbit

* **New Features**
  * Added Kamino DVI solver support to the basic joints and G1 robot examples via `--solver kamino`.
  * Improved sparse and dense contact solving with fused bilateral/unilateral solves and adaptive convergence tracking.
  * Added enhanced friction handling and contact warm-start fallback behavior.

* **Improvements**
  * Updated DVI defaults for relaxation, bilateral solve frequency, and contact warm-starting.
  * Added validation for oversized sparse projection allocations.
  * Exposed consistent solver status information, including convergence and residual metrics.

* **Bug Fixes**
  * Improved preservation of immovable joint anchors during simulation.


newton/tests/kamino/test_kamino_solver_kamino.py
```diff
@@ -559,6 +559,86 @@ def test_step_with_zero_max_contacts(self):
         solver.step(state_in, state_out, None, None, dt=1.0 / 60.0)
 
 
+class TestSolverKaminoPublic(unittest.TestCase):
+    def setUp(self):
+        if not test_context.setup_done:
+            setup_tests(clear_cache=False)
+        self.default_device = wp.get_device(test_context.device)
+
+    def test_padmm_preserves_immovable_joint_anchor(self):
```
Guirec-Maloisel: Isn't this test already covered by test_kamino_body_flags.py? If not or partially, we should rather add/edit a test there rather than here.

newton/examples/basic/example_basic_joints.py
```diff
@@ -199,8 +213,17 @@ def __init__(self, viewer, args):
 
         newton.eval_fk(self.model, self.model.joint_q, self.model.joint_qd, self.state_0)
 
-        self.collision_pipeline = newton.CollisionPipeline(self.model)
-        self.contacts = self.collision_pipeline.contacts()
+        if self.solver_type == "kamino":
```
Guirec-Maloisel: Are we disabling the collision pipeline so that Kamino use its internel detection? Could be worth a comment to clarify.

newton/_src/solvers/kamino/config.py
```diff
@@ -387,9 +387,12 @@ class ConstrainedDynamicsConfig(ConfigBase):
 
     cull_speculative_contacts: bool = True
     """
-    Whether to cull speculative (= separated) contacts in the dynamics solve.
-    These contacts have occasionally led to numerical instabilities, and
-    can yield inaccurate restitutive impacts.
+    Whether to discard separated speculative contacts before the dynamics solve.
```
Guirec-Maloisel: Could you also revert the changes to the config here regarding speculative contacts culling?

Guirec-Maloisel (changes requested):
Looks good to me overall, I had an AI summarize the changes to DVI and they make sense to me. Just had small comments.

Guirec-Maloisel:
Actually let me take a step back, it feels like the Schur complement change (if I understood it correctly) is a pretty fundamental formulation change that would require more evaluation in terms of the impact on performance and accuracy. We should probably discuss this a bit more offline (also with @Milad-Rakhsha-NV)

newton/examples/basic/example_basic_joints.py
nvtw: That change slipped in, was not intended. Removed it

newton/tests/kamino/test_kamino_solver_kamino.py
nvtw: Should be addressed

newton/_src/solvers/kamino/config.py
nvtw: Should be addressed

nvtw:
> Actually let me take a step back, it feels like the Schur complement change (if I understood it correctly) is a pretty fundamental formulation change that would require more evaluation in terms of the impact on performance and accuracy. We should probably discuss this a bit more offline (also with @Milad-Rakhsha-NV)

We can discuss more. I have many tests in follow up MRs. But this one is already getting too big without them. The main reason for the changes was contact drift observed mainly in the g1 scene when the robot fell over and feet and hands were touching the ground at the same time.

Milad-Rakhsha-NV:
>  The main reason for the changes was contact drift observed mainly in the g1 scene when the robot fell over and feet and hands were touching the ground at the same time.

I think the likely issue might be either under-resolved contact forces or insufficient coupling between joint sovler and contact solver; I see the proposed changes here but if I understand it correctly it is coupling the joint+conctact solve now in a different way. The original approach was to do something like 
```
for c in coupling_iterations:
  - solve joint limits (unilateral) with the latest known joint and contact impulses
  - solve joints (bilateral) with the latest known joint-limit and contact impulses
  - solve contact (unilateral) with the latest known joint and joint-limit impulses
```
 I think what you are proposing is to update the bilateral impulses from the unilateral `delta lambda_b = D_bb^-1 D_bu  lambda_u` instead of doing the full direct solve? This works but might be a bit complicated.  Can't we reuse the bilateral factorization and do the fowrad/backward substitution with the old D_bb Schur complement? I think it should be roughly the same speed in terms of complexity if not better

Milad-Rakhsha-NV:
@nvtw @Guirec-Maloisel I have some updates on the GS coupling of different solvers as well, but I am unsure which approach is better until we do a bit more testing. Would it be okay if we hold off on merging this PR until we synchronize at RoboSim? or maybe @nvtw let's do a short write-up and touch base so we can also see how the formulation is changing, because I think this is a bit different from the write-up I shared with you

nvtw:
> @nvtw @Guirec-Maloisel I have some updates on the GS coupling of different solvers as well, but I am unsure which approach is better until we do a bit more testing. Would it be okay if we hold off on merging this PR until we synchronize at RoboSim? or maybe @nvtw let's do a short write-up and touch base so we can also see how the formulation is changing, because I think this is a bit different from the write-up I shared with you

The changes in this MR are from try and error. Test G1, find something that does not work or behave well and adjust and then repeat until it works. But it's of course possible that other approaches work well too.

@Milad-Rakhsha-NV I think you still have write access to my Newton fork. Feel free to push into this MR.

nvtw:
@Guirec-Maloisel @Milad-Rakhsha-NV The last two commits add numeric comparisons between the previous method and the Schur method from this benchmark. To me Schur looks like a significant improvement but you know the math better than me. I would go the pragmatic route and take whatever gives us the best result at the highest FPS, even if we deviate a bit from the original DVI idea.

Guirec-Maloisel:
> @Guirec-Maloisel @Milad-Rakhsha-NV The last two commits add numeric comparisons between the previous method and the Schur method from this benchmark. To me Schur looks like a significant improvement but you know the math better than me. I would go the pragmatic route and take whatever gives us the best result at the highest FPS, even if we deviate a bit from the original DVI idea.

Thanks for running these comparisons! I do think that the idea makes sense and sounds very promising (and could also make sense for PADMM, I think @camevor had a write-up at some point that suggested something similar). But it feels like a big step for a PR where it's mixed with other changes. One way to not block merging would be to introduce the option as opt-in for now (would make it easy also to compare on more examples).

Milad-Rakhsha-NV:
Yes I think having this as an optional route can make sense, since for cases where the coupling between contact and joint forces is not significant, this can add unnecessary overhead. I think we should compare this with an approach of increasing the coupling iteration count (between contact/joint forces) and see which one is faster; they are both mathematically equivalent,  but the latter is simpler and with an outer loop count of 1 the cost is basically similar to before for all scenarios. 
I will prepare a branch for comparison

nvtw:
@Guirec-Maloisel @Milad-Rakhsha-NV using the Schur collision path is now opt in

Guirec-Maloisel (approved):
Good to merge from my side now!

PR #4129, opened 2026-09-02
Port `ControllerDifferentialIK` into `newton.controllers` module

### LINKED ISSUE #4123: [REQ] Implement `ControllerDifferentialKinematics`
### Description

Migrate the differential kinematics controller from Isaac Lab implementation into graphable warp-based implementation.

### Motivation / Use Case

See #3548 

### Alternatives Considered

_No response_

jeff-hough:
## Description

Introduces a heterogeneous Differential Inverse Kinematics controller, i.e. see issue #4123 , part of on-going task #3548 )

## Checklist

## Test plan
```
``` 

## New feature / API change

```python
import newton
import warp as wp
from newton.controllers import ControllerDifferentialIK, IkMethod

device = wp.get_device()

scene = newton.ModelBuilder()
for i in range(512):

control = model.control()

# -- pick what to control --
# Every panda articulation (not the conveyor), and within it every panda arm
# joint (not the fingers) driving a single tool site per robot -- fr3_hand_tcp.
FRANKA_READY_POSE = [0.0, -0.785, 0.0, -2.356, 0.0, 1.571, 0.785]  # one entry per selected joint

controller = ControllerDifferentialIK(
    model,
    articulations="panda_*",
    joints=[f"panda_joint{k}" for k in range(1, 8)],  # don't select the finger joints.
    tool_sites="fr3_hand_tcp",
    bandwidth=5.0,  # None to drive it live via inputs.bandwidth
    damping=None,  # ADAPTIVE_DAMPING computes its own lambda each step -- must be None
    ik_method=IkMethod.ADAPTIVE_DAMPING,
    adaptive_damping_min=0.02,
    adaptive_damping_max=0.5,
    adaptive_damping_threshold=0.2,
    # This pick task only cares about the tool's position, not its final
    # orientation -- axis_weight's 6 canonical axes are world-frame (there is
    # no per-tool reorientation the way ControllerOperationalSpace's
    # operational_frame_pose_world has), so the 3 orientation axes are left
    # at 0 rather than needlessly constraining an unconstrained task. A
    # single wp.spatial_vector broadcasts the same weight to all 512 pandas.
    axis_weight=wp.spatial_vector(1.0, 1.0, 1.0, 0.0, 0.0, 0.0),
    # panda_joint7 is redundant against the resulting 3D task -- keep it
    # anchored to the arm's own ready pose instead of drifting with nothing
    # to hold it.
    use_null_space_posture_control=True,
    null_space_stiffness=2.0,
    null_space_damping=0.05,
)

inputs = controller.input()
outputs = controller.output()

# The posture target is constant for the whole run -- assigned once, not
# reassigned in the loop below. Compact, shape [total_controlled_dofs]: one
# entry per selected joint, robot-grouped, so the 7-entry ready pose repeats
# once per robot.
inputs.q_des_null.assign(FRANKA_READY_POSE * controller.controlled_robot_count)

# An indexed view scatters the compact position/velocity targets straight
# into the sim, leaving the conveyor's and the fingers' entries in
# control.joint_target_q/qd untouched.
outputs.joint_q_target = control.joint_target_q[controller.q_start]
outputs.joint_qd_target = control.joint_target_qd[controller.qd_start]

# -- sim loop --
for _ in range(1000):
    inputs.desired_tool_pose_world.assign(desired_pose)  # per-robot, shape [controlled_robot_count]

    for _ in range(sim_substeps):
        state_0.clear_forces()

```

## Summary by CodeRabbit

- **New Features**
  - Added differential inverse-kinematics controllers with multiple solving methods, joint-limit avoidance, and null-space posture control.
  - Added model-based and model-free support for heterogeneous robot configurations.
  - Exposed new controllers and IK options through the public API.
  - Added interactive examples for differential IK and hybrid force/motion control.

- **Documentation**
  - Added API documentation, changelog details, and README links for controller examples.

- **Bug Fixes**
  - Improved handling and validation of world-attached tool sites.


newton/examples/controllers/example_controller_diff_ik.py
```diff
@@ -0,0 +1,418 @@
...
+import warp as wp
+
+import newton
+import newton.examples
+import newton.solvers
+import newton.utils
+from newton import Axis, JointTargetMode
+from newton.controllers import ControllerDiffIK
+
+IkMethod = ControllerDiffIK.IkMethod
```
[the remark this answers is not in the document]
jeff-hough: left-over file from class name refactor - removed.

jeff-hough:
@coderabbitai review

jeff-hough:
@coderabbitai review

newton/_src/controllers/impl/differential_ik/model_free.py
```diff
@@ -0,0 +1,1234 @@
...
+                _validate_array(
+                    array=damping,
+                    name="damping",
+                    dtype=wp.float32,
+                    shape=(controlled_robot_count,),
+                    device=self._device,
+                    required=False,
+                )
+        elif damping is not None:
+            raise ValueError(f"damping was given but ik_method={ik_method} does not use it (pass damping=None).")
```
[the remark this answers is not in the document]
jeff-hough: Singular Jacobian causing issues inverting is expected on pure Pseudo Inverse method. If the user wants damping they should use one of the other available methods

newton/_src/controllers/impl/differential_ik/model_free.py
```diff
@@ -0,0 +1,1234 @@
...
+                device=self._device,
+            )
+            axis_weight_np = axis_weight_resolved.numpy()
+        else:
+            raise TypeError(
+                "axis_weight must be a wp.spatial_vector or a wp.array[wp.spatial_vector] of shape "
+                f"(controlled_robot_count,), got {type(axis_weight_resolved).__name__}."
+            )
+        if np.any(axis_weight_np < 0.0):
+            raise ValueError(f"axis_weight must be non-negative, got {axis_weight_np.tolist()}.")
```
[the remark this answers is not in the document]
jeff-hough: This argument can be extended to every single value. I do not think this is worth checking.

newton/_src/controllers/impl/joint_impedance/model_based.py
```diff
@@ -183,144 +182,33 @@ def __init__(
...
-        controlled_dofs_per_robot = wp.array(controlled_dofs_per_robot_np, dtype=wp.int32, device=self._device)
-        max_controlled_dofs = int(controlled_dofs_per_robot_np.max())
-        self._model_robot_index = wp.array(model_robot_index_np, dtype=wp.int32, device=self._device)
-        # FK and dynamics run per robot, and a robot with no controlled DOF
-        # contributes nothing to the ones that have them. The mask stays
-        # model-sized because that is what the eval_* entry points take.
-        mask_np = np.zeros(model_robot_count, dtype=bool)
-        mask_np[model_robot_index_np] = True
-        self._controlled_robot_mask = wp.array(mask_np, dtype=wp.bool, device=self._device)
-        # ------------------------------------------------------------------
```
jeff-hough, line 311: Note:
Since this _exact_ check was being done in all three controller families, I have moved it to a private helper function.

newton/_src/controllers/impl/operational_space/model_based.py
```diff
@@ -315,204 +313,40 @@ def __init__(
...
-        # within its articulation's own joint range, of the joint that moves
-        # the tool site's body -- so this is (that joint's index) minus
-        # (the articulation's first joint index).
-        body_to_joint_np = np.full(model.body_count, -1, dtype=np.int32)
-        body_to_joint_np[joint_child_np] = np.arange(joint_child_np.size, dtype=np.int32)
-        tool_site_joint_np = body_to_joint_np[tool_body_np]
-        articulation_start_np = model.articulation_start.numpy()
-        robot_link_idx_np = (tool_site_joint_np - articulation_start_np[model_robot_index_np]).astype(np.int32)
-        self._robot_link_idx = wp.array(robot_link_idx_np, dtype=wp.int32, device=self._device)
-        # ------------------------------------------------------------------
```
jeff-hough, line 515: Note, since this exact code had to be replicated in the differential IK, it has moved to a module private helper.

newton/_src/controllers/impl/operational_space/_common.py
```diff
@@ -96,65 +90,6 @@ def _tool_pose_and_twist_kernel(
...
-    robot_link_idx: wp.array[wp.int32],  # (robot_count,) -> row-block index of the tool's link, within its articulation
-    articulation_dof_idx_of_padded_dof_idx: wp.array2d[
-        wp.int32
-    ],  # (robot_count, max_dofs) padded_dof_idx -> articulation_dof_idx, jacobian_com_world's own column numbering
-    controlled_dofs_per_robot: wp.array[wp.int32],  # (robot_count,) number of controlled DOFs for each robot
-    # outputs
-    jacobian_tool_world: wp.array3d[
-        float
-    ],  # (robot_count, 6, max_dofs) columns are twists about the tool point, in world coords
-):
```
jeff-hough, line 118: any kernels here which had to be re-used verbatim in the Differential IK have moved.

newton/_src/controllers/impl/differential_ik/_common.py
```diff
@@ -0,0 +1,478 @@
...
+import warp as wp
+from warp.fem.linalg import symmetric_eigenvalues_qr
+
+# Tolerance passed to symmetric_eigenvalues_qr: the QR algorithm iterates
+# until the off-diagonal terms of the tridiagonalized matrix fall below this,
+# relative to the diagonal terms.
+_EIGENVALUE_QR_TOL = wp.constant(wp.float32(1.0e-6))
+
+
+class IkMethod(enum.Enum):
```
eric-heiden: Can we name this `DifferentialIKMethod` to make sure we are not confusing users of `newton.ik`? I also prefer the capitalized spelling "IK" you already use for the controller classes here.

newton/_src/controllers/impl/_common.py
```diff
@@ -11,21 +11,321 @@
...
+    ],  # (block_count, max_dim, max_dim) inverse of the top-left block_dim x block_dim submatrix; untouched elsewhere
+):
+    """Explicit inverse of a batch of small SPD matrices, via Cholesky factorization.
+
+    Column c of the inverse solves ``spd_matrix @ x = e_c`` (e_c the c'th
+    standard basis vector), found by forward-substituting ``L y = e_c`` and
+    then back-substituting ``L^T x = y``. No dense-inverse routine (cofactor
+    expansion, Gauss-Jordan) is used — this is the numerically standard way to
+    invert a small SPD matrix, and the same recipe
+    ``newton/_src/actuators/response_oracle.py`` uses for the same reason.
```
eric-heiden: 🔵 This Cholesky-based SPD inverse is a near-verbatim copy of `_inverse_block_from_mass_matrix_kernel` in `newton/_src/actuators/response_oracle.py`. Both clamp the pivot with the same `wp.max(s, _FLOAT32_EPS * wp.max(wp.abs(...), 1.0))` expression, both solve column by column with forward then back substitution written into the output buffer in place, and both take a caller-preallocated scratch factor. The docstring here already names that file as "the same recipe".

The cost of a second copy is that a numerically delicate routine has to be kept in sync by hand. This PR already shows the hazard: on `main` this kernel was plain `@wp.kernel`, and relocating it here also flipped it to `enable_backward=False`, so a change that reads as pure code movement altered `ControllerOperationalSpace`'s gradient behavior.

I do not think this PR should carry the extraction. `actuators/` and `controllers/` are sibling packages, so neither can import the other's private helper without a layering violation, and unifying them needs a third home plus a wider decision. I opened #4133 to track that.

For this PR, could you turn the existing docstring note into an explicit two-way pointer, and add the matching one in `response_oracle.py`, so whoever edits one copy finds the other?

newton/_src/controllers/impl/differential_ik/model_free.py
```diff
@@ -0,0 +1,1278 @@
...
+    if value is None:
+        return
+    if isinstance(value, (int, float)) and not isinstance(value, bool):
+        if value < 0.0:
+            raise ValueError(f"{name} must be non-negative, got {value}.")
+        return
+    value_np = value.numpy() if isinstance(value, wp.array) else value
+    if np.any(value_np < 0.0):
+        raise ValueError(f"{name} must be non-negative, got {value_np.tolist()}.")
+
```
eric-heiden, line 109: ⚪ This one looks worth folding into the shared helper module now. `newton/_src/controllers/utils.py` is already the home for cross-family validation (`_validate_array`), it is in the same package so there is no layering question, and line 59 of this file already imports from it.

The move would be: lift `_validate_non_negative_gain` into `controllers/utils.py` next to `_validate_array`, then extend the existing import to

```python
from ...utils import _validate_array, _validate_non_negative_gain
```

A non-negative gain check is not specific to differential IK. The operational-space family validates gains too, in `impl/operational_space/model_free.py`, and would be able to reuse this instead of growing a sixth private `_validate_*` wrapper of its own.

Happy for this to wait for #4133 if you would rather keep the PR tight, but it is self-contained enough to do here.

newton/_src/controllers/impl/differential_ik/_common.py
```diff
@@ -0,0 +1,478 @@
...
+# rank-deficient whenever dof_count is below the robot's own task dimension,
+# and while the Cholesky pivot floor in _invert_spd_block_kernel keeps that
+# from producing NaN, it does not produce a meaningful pseudo-inverse in
+# that regime, so IkMethod.PSEUDO_INVERSE requires every robot to have at
+# least as many controlled DOFs as its own task dimension.
+# ---------------------------------------------------------------------------
+
+
+@wp.kernel
+def _build_jjt_plus_damping_kernel(
```
eric-heiden: 🔵 `_build_jjt_plus_damping_kernel` is a strict generalization of `_jacobian_times_jacobian_transpose_kernel` in `impl/operational_space/_common.py`, so after this PR the tree has two kernels computing `J Jᵀ` that have to stay in sync.

Set `axis_weight` to ones, `active_axis_of_slot` to the identity table, and `damping` to zeros, and this kernel's inner loop reduces exactly to the older one. The rest of that pipeline is already shared: the operational-space Moore-Penrose null-space branch and the differential IK null-space branch both run `_invert_spd_block_kernel` → `_task_matrix_times_jacobian_kernel` → `_null_space_projector_kernel` out of `impl/_common.py` after this step. Building `JJᵀ` is the only stage still duplicated.

Both kernels are private, so this can move without touching public API, and the operational-space caller already has everything it needs except two constants:

| Argument | What the operational-space caller would pass |
|---|---|
| `dof_count` | `self._controlled_dofs_per_robot`, already there |
| `task_dim` | `self._task_dim`, already there as `wp.full(robot_count, 6)` |
| `damping` | a zeros array, like `_zero_damping_buf` in this family |
| `active_axis_of_slot` | an identity table, built once at construction |
| `axis_weight` | a ones array, like `_null_space_axis_weight` in this family |

Suggest promoting this kernel to `controllers/impl/_common.py` next to the three it already feeds, and deleting `_jacobian_times_jacobian_transpose_kernel`. The only other thing to update is its direct import in `newton/tests/test_controllers_operational_space.py`.

Happy for this to wait for #4133 if you would rather keep the PR tight, but `impl/_common.py` is already being reorganized here, so this is a natural place to do it.

newton/_src/controllers/impl/differential_ik/model_free.py
```diff
@@ -0,0 +1,1278 @@
...
+            if self._use_null_space_posture_control
+            else None
+        )
+        self._null_space_stiffness_buf: wp.array[wp.float32] | None = (
+            wp.zeros(total_controlled_dofs, dtype=wp.float32, device=self._device, requires_grad=requires_grad)
+            if self._use_null_space_posture_control and self._null_space_stiffness_baked is None
+            else None
+        )
+
+    def _bake(self, value: wp.array[wp.float32] | float | None, size: int) -> wp.array[wp.float32] | None:
```
eric-heiden: ⚪ Same theme as the `_validate_non_negative_gain` note above: `_bake` is the general form of two helpers that already exist, so this is a good moment to collapse all three into one.

| Helper | Location | Buffer size |
|---|---|---|
| `_bake_gain` | `impl/joint_impedance/model_free.py` | hardcoded `self._total_controlled_dofs` |
| `_bake_joint_gain` | `impl/operational_space/model_free.py` | hardcoded `self._total_controlled_dofs` |
| `_bake` (this one) | here | `size` argument |

The bodies are otherwise character-for-character the same: return `None` for a live gain, `wp.full` for a scalar, `wp.zeros` + `wp.copy` for an array, all with `device=self._device, requires_grad=self._requires_grad`. Adding the `size` parameter is the only thing that was missing, and both existing call sites collapse to `_bake(value, self._total_controlled_dofs)`.

Suggest lifting this into `controllers/utils.py` as a free function taking `device` and `requires_grad` explicitly, next to `_validate_array`, then deleting `_bake_gain` and `_bake_joint_gain` in favour of it. That is the same move as the validation helper, so the two can land together.

One thing to keep separate: operational space also has a `_bake_gain` that takes a `wp.spatial_vector` and produces a per-robot buffer. That one has a genuinely different signature and should stay where it is.

If you would rather keep this PR tight, #4133 is the right place to track it.

newton/_src/controllers/impl/differential_ik/_common.py
```diff
@@ -0,0 +1,478 @@
...
+    error can land a fully degenerate eigenvalue just below zero.
+    """
+    robot_idx = wp.tid()
+
+    local_matrix = wp.spatial_matrix()
+    for row in range(6):
+        for col in range(6):
+            local_matrix[row, col] = matrix[robot_idx, row, col]
+
+    eigenvalues, _ = symmetric_eigenvalues_qr(local_matrix, _EIGENVALUE_QR_TOL)
```
eric-heiden: 🔵 Both other `symmetric_eigenvalues_qr` call sites in the tree guard the result before using it. The two new ones here (this line and line 344 in `_truncated_pinv_matrix_kernel`) do not.

`newton/_src/solvers/implicit_mpm/implicit_mpm_solver_kernels.py`:

```python
scales, ev = fem.linalg.symmetric_eigenvalues_qr(diag_block, _QR_TOLERANCE)

# symmetric_eigenvalues_qr may return nans for small coefficients
if not (wp.ddot(ev, ev) < _NAN_THRESHOLD and wp.length_sq(scales) < _NAN_THRESHOLD):
    scales = wp.get_diag(diag_block)
    ev = wp.identity(n=scales.length, dtype=float)
```

`rheology_solver_kernels.py` has the same check with the threshold written out as `1.0e16`.

"Small coefficients" is the regime these two kernels are built to live in. `ADAPTIVE_DAMPING` and `TRUNCATED_SVD` exist precisely so the controller can be run near a singularity, where `JJᵀ`'s smallest eigenvalues go to zero. On top of that, whenever `task_dim < 6` the matrix handed to `symmetric_eigenvalues_qr` is a padded 6x6 with an exactly-zero trailing block, so it is rank-deficient by construction on every step, not just near a singularity.

If a NaN does come back, nothing catches it. `_adaptive_damping_kernel` takes `wp.sqrt` of the result, the NaN λ lands on the diagonal in `_build_jjt_plus_damping_kernel`, and the joint velocity targets come out NaN with no error raised.

Could you either add the same guard at both call sites, or add a short comment explaining why `JJᵀ` cannot reach the input regime the MPM callers are protecting against? Either one resolves it. As written a reader cannot tell whether the missing guard is a deliberate call or an oversight.

newton/_src/controllers/impl/differential_ik/_common.py
```diff
@@ -0,0 +1,478 @@
...
+
+    for row in range(6):
+        for col in range(6):
+            total = float(0.0)
+            for i in range(6):
+                eigenvalue = wp.max(eigenvalues[i], 0.0)
+                sigma = wp.sqrt(eigenvalue)
+                if sigma > threshold:
+                    total += eigenvectors_by_row[i, row] * eigenvectors_by_row[i, col] / eigenvalue
+            pinv_matrix[robot_idx, row, col] = total
```
eric-heiden: **[P1] `TRUNCATED_SVD` loses singular directions that the configured threshold says to retain.**

This derives the decomposition from `J J^T`, which squares `J`'s condition number before the float32 eigensolve. With singular values `[1, .7, .3, .1, .03, 1e-4]` and `truncated_svd_threshold=5e-5`, the `1e-4` direction should be retained. For a task error scaled so the expected `||qdot||` is approximately 1, I get approximately `6e-8` on CPU and `1.156` on CUDA; a direct float32 SVD of `J` has relative error around `1e-4` for the same input. Thus this method can drop or materially distort a finite retained direction without producing NaNs. This is separate from the existing comment about guarding non-finite `symmetric_eigenvalues_qr` output at this call site. For comparison, the [Isaac Lab reference implementation](https://isaac-sim.github.io/IsaacLab/main/_modules/isaaclab/controllers/differential_ik.html) performs SVD directly on the Jacobian.

newton/_src/controllers/impl/differential_ik/model_free.py
```diff
@@ -0,0 +1,1278 @@
...
+            ``ik_method=IkMethod.DAMPED_LEAST_SQUARES`` (the default); must
+            be ``None`` for every other :class:`IkMethod`, which has no λ to
+            set.
+        ik_method: Inverse-Jacobian solve method, an :class:`IkMethod`.
+            Defaults to ``IkMethod.DAMPED_LEAST_SQUARES``. If
+            ``requires_grad=True``, must be ``IkMethod.TRANSPOSE`` -- every
+            other method routes through a matrix inversion or
+            eigendecomposition kernel whose backward pass is currently
+            disabled (see ``_common.py``), so it would silently contribute
+            no gradient.
```
eric-heiden: **[P2] Align the `ik_method` documentation with the unconditional gradient rejection.**

This says that `requires_grad=True` is supported when `ik_method=IkMethod.TRANSPOSE`. However, the constructor unconditionally rejects every `requires_grad=True` configuration at lines 422-423, and the model-based controller repeats the same TRANSPOSE guidance even though it delegates here. The separate `requires_grad` argument documentation correctly says gradients are not supported. These public descriptions should agree so users are not directed to a configuration that always raises `ValueError`.

eric-heiden:
Some comments on the new differential IK controller example:
* The framerate is quite low, I'm getting about 10 FPS on an RTX 4090. This seems to be because it doesn't use CUDA graph capture. Note that even though you have an interactive gizmo target input, you can copy it to its persistent device array outside the captured graph and then keep replaying the graph.
* The 4-link arm is not really able to achieve the rotation target. Is this expected? I tried it from different joint configurations - when I rotate the target, the arm just moves away from the position target and doesn't achieve the desired yaw angle either. See attached video:

https://github.com/user-attachments/assets/123394ab-8a2a-411a-a0db-d9ded3cdf224

newton/_src/controllers/impl/differential_ik/_common.py
```diff
@@ -0,0 +1,478 @@
...
+
+    ``activation`` is 0 while more than ``margin`` away from both limits,
+    ramps linearly to 1 at either limit, and stays 1 beyond it — a DOF
+    already past its limit gets the full correction, not none.
+    """
+    dof = wp.tid()
+    q = joint_q[dof]
+    lower = joint_pos_lower[dof]
+    upper = joint_pos_upper[dof]
+    q_mid = 0.5 * (lower + upper)
```
jcarius-nv, line 565: **[P1] Joint-limit avoidance produces NaNs for accepted unbounded limits.**

An interval of `[-inf, +inf]` passes the constructor's `lower < upper` validation, but this midpoint evaluates to `NaN`. Even though the activation is zero away from a limit, the final expression evaluates `0 * (q - NaN)`, so the joint position and velocity outputs become `NaN`. I reproduced this on CUDA with one DOF, zero task error, DLS damping `0.1`, and joint-limit avoidance enabled.

newton/_src/controllers/impl/differential_ik/model_free.py
```diff
@@ -0,0 +1,1278 @@
...
+                    name="damping",
+                    dtype=wp.float32,
+                    shape=(controlled_robot_count,),
+                    device=self._device,
+                    required=False,
+                )
+        elif damping is not None:
+            raise ValueError(f"damping was given but ik_method={ik_method} does not use it (pass damping=None).")
+
+        if ik_method == IkMethod.PSEUDO_INVERSE:
```
jcarius-nv: **[P1] The dimensionality check does not establish the documented Moore-Penrose behavior.**

A square or tall Jacobian can still be rank-deficient. For a 6x6 Jacobian whose first two rows are identical and a task error of `[1, 0, ...]`, `PSEUDO_INVERSE` returned `qdot[0] = 1.0`; the Moore-Penrose least-norm result is `0.5`. DLS with the currently permitted `damping=0` has the same behavior because both reach the SPD/Cholesky inverse path. Thus singular configurations can silently return a pivot-floor-dependent result despite the public enum documenting a Moore-Penrose pseudo-inverse.

jcarius-nv (commented):
Overall, this looks good: it is a well-structured controller addition with broad tests, documentation, and examples, and the local validation and required checks pass. I found the two additional numerical correctness issues noted inline. I also independently reproduced Eric's existing [P1 `TRUNCATED_SVD` finding](https://github.com/newton-physics/newton/pull/4129#discussion_r3919428333) with randomized rotated, ill-conditioned Jacobians; the worst relative joint-velocity norm error I observed was 1.254 on CPU and 2.194 on CUDA. Aside from these three numerical findings, I do not have further concerns.

newton/_src/controllers/impl/differential_ik/model_free.py
jeff-hough: Just checking, and the operational space and joint impedance actually don't have these checks just yet. For now, I will leave the helper where it is since it is only used in the one controller family. However, I generally agree that 1) this could be lifted, and 2) it _should_ be used in the other controllers. I will make a followup issue to track.

newton/_src/controllers/impl/differential_ik/_common.py
jeff-hough: Indeed - I wanted to avoid implementing an SVD solver, but it is true that you loose significant precision when decomposing the `JJ^T` matrix compared to decomposing just `J`. I will investigate and see if a contained SVD solver is worthwhile, or if there is a different workaround which could prevent this accuracy issue at poorly conditions matrices.

newton/_src/controllers/impl/differential_ik/model_free.py
jeff-hough: Done, but changed the name to `_bake_optional_float_array` to be more descriptive of exactly what the shared helper does.

newton/_src/controllers/impl/differential_ik/_common.py
jeff-hough: I do think it would be odd user behavior to enable joint limit avoidance, and then pass infinite values for the joint limits - but I will add this check on these parameters none-the-less, since I could imagine a user passing infinity as a way of attempting to "disable" joint avoidance. I will add a message which points to the enable boolean.

hujc7 (commented):
Porting Isaac Lab's `DifferentialIKController` onto `ControllerDifferentialIKModelFree`. The mapping is 1:1 across all five of our IK methods, and it validated cleanly against our previous Torch implementation. One request and two notes below.

**Validated against our independent Torch implementation:**
- `axis_weight` reproduces our `orientation_weight` to **3.6e-6** on a scale of 2.2 — it scales the Jacobian rows and the task error together, exactly as ours did.
- The joint-limit-avoidance formula is identical to ours (range midpoint, linear ramp within the margin, `-gain * activation * (q - q_mid)`).
- `dt=1.0` reproduces our historical `q_target = q + Δq` contract exactly.
- `TRUNCATED_SVD` matches our `svd` (invert above the threshold, drop below), and `ADAPTIVE_DAMPING` matches our `adaptive_dls`.

Two independently written implementations agreeing across pose/position, relative/absolute and all five solvers is a good sign for both.

---

### 1. The null-space projector and the IK solve share one axis setting — `newton/_src/controllers/impl/differential_ik/model_free.py:339`

`axis_active = axis_weight > 0.0`, and the null-space projector runs on those active axes. That couples two questions we need answered differently.

Our SO-101 is a **5-DOF** arm on a 6-D pose task. It wants orientation *softly tracked in the solve* (`orientation_weight=0.5`), but *excluded from the null-space projection* — at 5 DOF there is no null space against a 6-D task (`JJᵀ` is 6×6, rank ≤ 5), while there is a healthy 2-D one against the 3-D position task. Our implementation projects the joint-limit bias against the position rows only, so limit avoidance provably never perturbs commanded position.

With `axis_weight` doing both jobs we can't express that: zeroing the orientation weights fixes the projector but stops the wrist tracking orientation at all. Damping helps but cannot close it — at `null_space_damping=0.01` the residual position perturbation is ~2e-3 versus ~0 for the 3-row projection, which is inherent rather than a tuning artifact.

**Request:** a separate `null_space_axes` selection for the projector, defaulting to `axis_weight` so the default is a no-op. A 7-DOF Franka (1 spare DOF) is unaffected either way; it's under-actuated arms that need the split. Happy to contribute it.

Not a blocker for us — we keep a small Torch helper for now, which would delete cleanly if this lands.

### 2. No public setter for the joint limits — `model_free.py:571`

The limits are snapshotted (`wp.array(joint_pos_lower.numpy(), ...)`), so the caller's array isn't aliased and there's no setter. Our `set_joint_pos_limits()` is public and callable at any time.

Not blocking: writing into `controller._joint_pos_lower` does take effect (verified — the correction went 0.955 → 0.0 after widening the limits in place), so it works today through a private member. A public setter, or reading the limits from a live input port, would make that supported rather than incidental.

### 3. `IkMethod.PSEUDO_INVERSE` outside its documented precondition

The docstring requires "at least as many controlled DOFs as its own task dimension". On a rank-deficient Jacobian that violates it, the solver stays finite (good) but deviates ~2.3% from Moore-Penrose, where `torch.linalg.pinv` is exact. Purely informational — no shipped Isaac Lab config uses `pinv` — but worth knowing the failure mode is silent drift rather than an error.

newton/_src/controllers/impl/differential_ik/_common.py
jeff-hough: A new SVD solver has been added (used for all of the inverse methods) - new tests check the results with ill-conditioned Jacobians. This should now be resolved.

newton/_src/controllers/impl/differential_ik/_common.py
jeff-hough: I have fully changed the solver over to use an SVD solver, the eigenvalue solver is no longer used.

newton/_src/controllers/impl/differential_ik/model_free.py
jeff-hough: Yes; the `n_dofs < task_dof` is only meant to guard against one particularly simple structural rank-deficiency, however even with a strict moore-penrose (i.e. not the floored cholesky solver), it only has a real advantage if the condition of the matrix is _exactly_ 0. For ill-conditioned matrices, the moore-penrose still performs quite poorly.

Each IK method has now been redone to use an SVD solver - you should see that it performs much better now at poorly conditioned Jacobians! Of course, the damped methods give much better more stable results than the pure moore-penrose pseudoinverse.

jeff-hough:
@eric-heiden
>Some comments on the new differential IK controller example:
>
>The framerate is quite low, I'm getting about 10 FPS on an RTX 4090. This seems to be because it doesn't use CUDA graph capture. Note that even though you have an interactive gizmo target input, you can copy it to its persistent device array outside the captured graph and then keep replaying the graph.
The 4-link arm is not really able to achieve the rotation target. Is this expected? I tried it from different joint configurations - when I rotate the target, the arm just moves away from the position target and doesn't achieve the desired yaw angle either. See attached video:

1. Can you tell me if the frame-rate has improved? I have added graphing
2. No, after some playing around, I discovered this was poor tuning in the joint-level PD controllers. Given that this is not a physics-example (and the controller outputs pure velocity and position), I have opted to fully remove the physics solver from the example. I think this more accurately lets the user see the behavior of the actual controller, unfiltered by system dynamics.

Another couple of changes to the example:

1. I have added an command-line argument to let the user choose different solver methods (pinv, dls, etc)
2. I have made it that the gizmos automatically only show for the task-axes you are controlling. The UR10 now is "position only" control.

newton/_src/controllers/impl/differential_ik/_common.py
jeff-hough: The more general kernel here `_build_jjt_plus_damping_kernel` no longer exists, and I am not explicitly forming `J*J^T` anymore for better performance when the Jacobian is ill-conditioned :)

newton/_src/controllers/impl/differential_ik/_common.py
```diff
@@ -0,0 +1,812 @@
...
+            vt[largest_idx] = v_tmp
+        sigma = candidate_sigma[i]
+        s_vec[i] = sigma
+        ut[i] = wp.where(sigma > A.dtype(1.0e-12), at[i] / sigma, ut[i] * zero)
+
+    return wp.transpose(ut), s_vec, wp.transpose(vt)
+
+
+@wp.kernel
+def _svd_one_sided_jacobi_kernel(
```
jeff-hough, line 781: @eric-heiden I added this SVD solver since, as far as I can tell, warp doesn't have one. If there were to be something of a `newton.linalg` package as I asked about in issue #4133, I think this kind of think would be a good candidate to lift there. I am sure there are many other places where batched SVD solvers would be useful!

jeff-hough:
@coderabbitai review

jeff-hough:
@coderabbitai review

jeff-hough:
@coderabbitai review

jcarius-nv:
Re-reviewed at `198332586b6d6a600d8d2e21305acd8a3553c9f5` after the latest commits and discussion. The previously raised numerical issues around unbounded joint limits, singular `PSEUDO_INVERSE`/zero-damping DLS behavior, and `TRUNCATED_SVD` accuracy are addressed by the updated validation and direct one-sided Jacobi SVD path, with focused CPU/CUDA coverage. The other substantive review feedback, including graph capture and example behavior, `DifferentialIKMethod` naming, `null_space_axes`, joint-limit updates, selection validation, and bounded SVD reconstruction, has also been addressed. The change fits Newton well, the required checks pass, and I do not see any remaining blocking concerns.

One accepted tradeoff is the absolute cutoff in `_damped_pinv_singular_value()`: values satisfying `sigma * sigma + lambda * lambda <= 1e-12` are dropped, so sufficiently small full-rank Jacobians or axis weights behave like a hard truncation. That is acceptable for this PR, but the cutoff and its scale dependence may be worth reconsidering or documenting in a future PR.

jcarius-nv (approved):

PR #4132, opened 2026-09-02
Add massless routed tendons

vreutskyy:
## Description

Add first-class massless routed tendons to Newton, with native support in
`SolverXPBD` and `SolverVBD`.

A tendon is an ordered route of attachment, pinhole, and circular rolling
links attached to rigid bodies. Unlike Newton’s rod-based cable model, the
tendon itself has no massive bodies: it stores routed material and applies
forces to the rigid bodies along the route.

This is a clean integration of the accepted tendon work on top of current
`main`, separated from the experimental branches used during development and
Tesla validation.

### Included

- Public `ModelBuilder.add_tendon()` and `add_tendon_link()` authoring API
- Attachment, pinhole, and circular rolling links
- Segment compliance and damping
- Finite capstan friction and material transfer
- Material conservation through rolling and dynamic-route transitions
- Isolated dynamic rollers with signed-distance switching and activation
  tolerance
- XPBD stretch and rolling-slip constraints
- VBD tendon forces and local Hessians
- Correct handling of same-body tendon spans and off-center roller torque
- Routing and diagnostics evaluated from the accepted body pose
- Material- and damping-tension diagnostics
- VBD graph-coloring integration
- CUDA graph capture and deterministic XPBD support
- Public documentation, five examples, and CPU/CUDA regression tests

### Not included

The following experimental or follow-up work is deliberately excluded:

- Nonlinear/sigmoid compliance
- XPBD mass splitting or tendon warm-starting
- Direct/global tendon solves and sequential/colored-GS experiments
- Virtual-radius handling for unsupported fixed-roller wraps
- Experimental VBD joint-limit enforcement
- Consecutive dynamic rollers, route reordering, branching, or cyclic tendons
- MuJoCo or USD tendon import, export, or lowering
- Tendon actuation through `Control`
- Standard solver-output integration
- Tendon-aware `ModelBuilder` merging and replication
- Differentiable tendon simulation

The unsupported cases are documented. Tendon-bearing builder merges are
rejected explicitly rather than silently dropping their tendon data.

Reference: Müller et al., [Cable Joints](https://doi.org/10.1111/cgf.13507),
Computer Graphics Forum 37(8), 2018.

## Checklist

- [x] The user-facing change is recorded in the `[Unreleased]` changelog

## Test plan

```text
uv run --extra dev -m newton.tests -k tendon
uvx pre-commit run -a
```

The tendon-filtered suite passed 195 tests on CPU/CUDA with two expected
CPU CUDA-graph skips. The Sphinx HTML documentation build also passed with
warnings treated as errors.

## New feature / API change

```python
import newton

# Create anchor_body, guide_body, and payload_body first.
builder.add_tendon()
builder.add_tendon_link(
    body=anchor_body,
    link_type=newton.TendonLinkType.ATTACHMENT,
)
builder.add_tendon_link(
    body=guide_body,
    link_type=newton.TendonLinkType.ROLLING,
    radius=0.05,
    mu=0.2,
    compliance=1.0e-6,
    damping=0.1,
)
builder.add_tendon_link(
    body=payload_body,
    link_type=newton.TendonLinkType.ATTACHMENT,
    compliance=1.0e-6,
    damping=0.1,
)

# VBD requires coloring after tendon connectivity is authored.
builder.color()
model = builder.finalize()
```

## Summary by CodeRabbit

* **New Features**
  * Added massless routed tendon support for XPBD and VBD solvers.
  * Supports attachments, pinholes, rolling links, capstan friction, and dynamic route switching.
  * Added public tendon link types and flags for model construction.
  * Added tendon state reset, routing diagnostics, and configurable solver controls.
  * Added five cable and tendon examples, including 3D routing and friction demonstrations.

* **Documentation**
  * Added comprehensive guidance, API references, usage examples, limitations, and cable examples for routed tendons.


jcarius-nv:
# General Review Comments

Thanks @vreutskyy  — I think this is a promising direction and that the capability fits Newton. I support landing this as an **experimental, narrowly scoped foundation for massless routed tendons**, but “experimental” should describe the maturity and supported feature set, not leave the public concepts or API ambiguous.

We already prepared with @jumyungc 

- **Cable** is the broader modeled object/application.
- A **rod** is the massive, deformable representation, appropriate when cable inertia, deformation, contact, or direct manipulation matters.
- A **tendon** is Newton’s reduced-order, massless routed cable-transmission representation, appropriate when the load path matters but the cable’s own inertia does not.

I do not expect rods and tendons to share solver storage or implementation, but they should form one coherent public modeling story. Today we have `add_rod()`, the deprecated `add_joint_cable()` alias, this new `add_tendon()`, and the provider-scoped MuJoCo tendon representation. Before freezing another public noun and API, please explain how these fit together and whether compatible MuJoCo spatial tendons should lower to the new Newton tendon model or remain a separate solver-specific representation.

For this experimental landing, I think the following should be completed in this PR:

1. **Tighten the builder API.** `add_tendon()` returns a tendon index, but `add_tendon_link()` cannot use it and instead appends to the most recently created tendon. That makes construction depend on hidden mutable builder state and prevents interleaved or reusable authoring. Most comparable builder APIs either take an explicit owner/index or receive the complete ordered collection. I would expect something like `add_tendon_link(tendon_idx, ...)`, or an atomic API that accepts the route, unless there is a strong established precedent for the current pattern.

2. **Clarify the route vocabulary.** “Link” is ambiguous in Newton because it commonly means a rigid body in an articulation. These objects are heterogeneous route elements—attachments, pinholes, and rollers—rather than links in that sense. “Anchor” only describes the attachment subtype. Please define the domain concept more clearly and consider a name such as `route point`, `routing element`, or `guide`, with corresponding API/type names.

3. **Sketch the USD authoring model.** Full USD implementation does not need to land here, but we should understand how a tendon, its ordered route, its body/geometry relationships, and its per-segment properties would be represented before stabilizing the Python API. Newton’s existing `MjcTendon` USD representation is useful design input: it gives the tendon stable prim identity, separates tendon-level properties from ordered path entries, uses relationships to modeled objects, and allows an actuator to target the tendon. The Newton schema need not copy it, but the two should be intentionally relatable.

4. **Document the MuJoCo relationship and cross-solver subset.** Please add a concise matrix showing what maps exactly, approximately, or not at all. For example, attachments and pinholes may overlap with MuJoCo sites, and some circular routes may overlap with cylinder wrapping, while Newton’s capstan material transfer and physical roller torque have additional semantics. The important outcome is to identify which authored Newton tendons could also run through `SolverMuJoCo`, and whether they share a common model/schema representation or require explicit lowering.

5. **Support builder merging and replication.** Rejecting tendon-bearing builders makes this difficult to use in normal vectorized robot-learning workflows. Unless there's a good reason to defer it, I'd consider this part of making the experimental feature usable rather than a later feature expansion.

6. **Complete the user-facing documentation.** In addition to marking the API experimental and precisely listing its current limitations, please add practical guidance for choosing a massless tendon versus a massive rod. That distinction is central to the feature, not merely terminology, so we guide users the right way.

One related API question: is there intended to be a public read API for cable/tendon force? The current documentation points users to solver-owned `tendon_seg_material_tension` and `tendon_seg_damping_tension`, with total unilateral tension reconstructed from them. That is useful diagnostically, but it is not a solver-neutral `Model`/`State` contract. I am not asking for the broader actuator/`Control` feature in this PR, but we should either provide a minimal experimental per-segment tension read contract or explicitly document that the current arrays are solver-specific diagnostics rather than settled public API.

In my opinion out of scope of this PR are arbitrary wrap geometry, nonlinear constitutive laws, or the broader future actuator feature set. Those can follow. My requested scope here is to get the foundational abstraction and API right, provide the schema and MuJoCo direction, support replication, and document clearly what this experimental representation is and when users should choose it.
Also the current feature set looks good enough for a first version. I think once we have worked out the above points it's better getting this in the hands of users and adding features based on concrete requests and usecases rather than bringing more into this PR.

PR #4143, opened 2026-09-03
[Kamino] Fix type alias on reset callbacks on python 3.10

Guirec-Maloisel:
## Description

The recent PR #4120 modified a type alias in solver_kamino_impl.py that leads to a TypeError on Python 3.10 (because there wp.array[wp.bool] is not recognized as a type). `from __future__ import annotations` does not help here because it is a type alias rather than an annotation.

Stringifying the wp.array part of the annotation (as how `from __future__ import annotations` would do it if it was an annotation) solves the issue.

## Checklist

## Bug fix

**Steps to reproduce:**

Run any unit test or example involving Kamino on a platform with Python 3.10.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a `TypeError` affecting Kamino callback type annotations when using Python 3.10.
  * Improved compatibility for callback type annotations without changing runtime behavior.

### DIFF
--- a/changelog/+kamino-type-error-008d8f13.fixed.md
+++ b/changelog/+kamino-type-error-008d8f13.fixed.md
@@ -1,0 +1,1 @@
+Fix type alias leading to TypeError on Python 3.10 when using SolverKamino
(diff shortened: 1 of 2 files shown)


chschuma-disney (dismissed):
Sorry for missing this. LGTM

chschuma-disney (approved):

PR #4155, opened 2026-09-04
Raise named error for unmappable MuJoCo shape types

### LINKED ISSUE #4064: [BUG] A CONE shape raises KeyError in SolverMuJoCo construction
`GeoType.CONE` is defined (`newton/_src/geometry/types.py:105`) and
`ModelBuilder.add_shape_cone()` accepts it, but `CONE` is absent from
`geom_type_mapping` in the MuJoCo solver
(`newton/_src/solvers/mujoco/solver_mujoco.py:5900-5910`), so building a
`SolverMuJoCo` over a model containing one dies with a bare `KeyError`.

The failure is at solver construction, so the model gets no solver at all --
it is not a degraded cone, it is no simulation.

Repro, newton 1.5.0 / warp 1.16.0 / mujoco 3.11.0:

```python
import newton, warp as wp

body = builder.add_body(xform=wp.transform(wp.vec3(0.0, 0.0, 1.0), wp.quat_identity()))
builder.add_shape_cone(body, radius=0.2, half_height=0.3)
builder.add_ground_plane()
model = builder.finalize()
print("finalize OK: %d shapes" % model.shape_count)
newton.solvers.SolverMuJoCo(model)
```

```
finalize OK: 2 shapes
KeyError: np.int32(9)
```

`9` is `GeoType.CONE`. `finalize()` succeeds, so nothing upstream warns.

MuJoCo has no cone primitive, so there may be no faithful mapping -- but the
current behaviour is an unhandled `KeyError` from inside a dict lookup, which
gives a caller nothing to act on. Even raising a named error at
`add_shape_cone` time (or at solver construction, naming the shape and
suggesting a convex mesh instead) would make this diagnosable.

Happy to send a PR for whichever you prefer -- a clear raise, or a mapping to
an approximating mesh.

Found while integrating newton as the only physics backend in an external
simulator; we exclude cones from collision geometry on our side for this exact
reason, and it took a while to attribute.

deepspace28:
﻿## Description

`GeoType.CONE` is a valid Newton shape (`ModelBuilder.add_shape_cone()`), but it has no entry in `geom_type_mapping`, so building a `SolverMuJoCo` over a model containing one crashed with a bare `KeyError: np.int32(9)` raised from inside a dict lookup at solver construction. The caller gets nothing actionable, and because the failure happens at solver construction, the model ends up with no solver at all.

This PR raises a named `ValueError` instead, stating the unsupported shape type, the shape label and index, and the supported alternatives (primitives or a convex mesh). Sites already fall back to spheres, so only collision geoms are affected.

The check is generic (membership in `geom_type_mapping`), so it also covers any future Newton shape types without a MuJoCo mapping, not just cones.

Closes #4064

## Checklist

## Test plan

- New regression test `test_unsupported_shape_type_raises` (in `TestMuJoCoValidation`) builds a model with a cone and asserts `SolverMuJoCo` raises a named `ValueError`.
- Verified the test fails without the fix (fails with `KeyError: np.int32(9)`, the exact error from the issue) and passes with it.
- Full class run: `uv run --extra dev -m newton.tests -k TestMuJoCoValidation` — 16/16 OK.

## Bug fix

**Steps to reproduce:**

1. Build a Newton model containing a cone shape (`builder.add_shape_cone(...)`).
2. Construct `newton.solvers.SolverMuJoCo(model)`.
3. Observe a bare `KeyError: np.int32(9)` from `geom_type_mapping[stype]` instead of a diagnosable error.

**Minimal reproduction:**

```python
import newton, warp as wp

body = builder.add_body(xform=wp.transform(wp.vec3(0.0, 0.0, 1.0), wp.quat_identity()))
builder.add_shape_cone(body, radius=0.2, half_height=0.3)
builder.add_ground_plane()
model = builder.finalize()
newton.solvers.SolverMuJoCo(model)  # KeyError: np.int32(9)
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Unsupported shape types in the MuJoCo solver now produce a clear `ValueError` identifying the shape and recommending a supported primitive or convex mesh.
* **Documentation**
  * Documented the updated error behavior for unsupported geometry types.
* **Tests**
  * Added coverage to verify that unsupported cone shapes raise the expected error.

### DIFF
--- a/changelog/4064.fixed.md
+++ b/changelog/4064.fixed.md
@@ -1,0 +1,1 @@
+Raise a named `ValueError` when `SolverMuJoCo` encounters a shape type it cannot map to a MuJoCo geom (e.g. cones), instead of failing with a bare `KeyError`.
--- a/newton/_src/solvers/mujoco/solver_mujoco.py
+++ b/newton/_src/solvers/mujoco/solver_mujoco.py
@@ -6361,6 +6361,13 @@ class SolverMuJoCo(SolverBase, CouplingInterface):
 
                 if stype == GeoType.PLANE and newton_body_id != -1:
                     raise ValueError("Planes can only be attached to static bodies")
+                if stype not in geom_type_mapping:
+                    raise ValueError(
+                        f"MuJoCo solver does not support shape type '{GeoType(stype).name}' "
+                        f"for shape {model.shape_label[shape]!r} (shape {shape}). "
+                        "Replace it with a supported primitive (sphere, capsule, cylinder, box, "
+                        "ellipsoid) or a convex mesh."
+                    )
                 geom_params = {
                     "type": geom_type_mapping[stype],
                     "name": name,
--- a/newton/tests/test_mujoco_solver.py
+++ b/newton/tests/test_mujoco_solver.py
@@ -5872,6 +5872,22 @@ class TestMuJoCoValidation(unittest.TestCase):
             SolverMuJoCo(model, separate_worlds=True)
         self.assertIn("shape types mismatch at position", str(ctx.exception).lower())
 
+    def test_unsupported_shape_type_raises(self):
+        """Test that a shape type without a MuJoCo geom mapping raises a named ValueError."""
+        builder = newton.ModelBuilder()
+        b1 = builder.add_link(mass=1.0, com=wp.vec3(0, 0, 0), inertia=wp.mat33(np.eye(3)))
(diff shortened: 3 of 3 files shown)


vreutskyy (approved):
Reviewed the unsupported-shape validation, surrounding geometry-routing paths, regression coverage, and changelog entry. Verified the exact head commit with the full TestMuJoCoValidation class: 16/16 tests passed.

PR #4157, opened 2026-09-04
Speed up deformable edge-edge collision detection via adaptive block size

nvtw:
## Description

The per-edge BVH traversal in deformable edge-edge self-collision detection diverges heavily within a warp. Launches too small to fill the GPU are latency bound and run fastest with 8-thread blocks, while large launches need full warps for throughput. Choose the edge-edge launch block size from the edge count and SM count (about 16 blocks per SM, clamped to [8, 32]). `TriMeshCollisionDetector(collision_detection_block_size=None)` now means automatic; an explicit value is still honored for both detection kernels.

Measured on an RTX PRO 6000 against main: cloth twist example 640 ms vs 730 ms for 100 frames; 391k-edge cloth 2.54 ms vs 3.48 ms per detection; 1.44M-edge scene (1024 replicated worlds) 7.30 ms vs 11.07 ms. A tile-stack candidate redistribution variant was evaluated and rejected: it was 15-20% slower than a plain 32-thread launch on deformed twist and Franka cloth states replicated to 64 worlds, and 2.5x slower on the single-world twist scene.

Add the `FastDeformableSelfCollision` benchmark (one large cloth and 1024 replicated worlds, full refit plus vertex-triangle and edge-edge detection) to the PR gate. Setup takes about 2.5 s per case and the timed part under 0.1 s.

## Checklist

## Test plan

```
uv run --extra dev -m newton.tests -k test_collision_cloth
cd asv && uv run --extra dev python -m unittest discover -s tests -p "test_benchmark_simulation.py"
uv run python asv/benchmarks/simulation/bench_cloth.py -b FastDeformableSelfCollision -b FastExampleClothTwist

## Summary by CodeRabbit

- **New Features**
  - Added a fast deformable cloth self-collision benchmark covering large and replicated simulation worlds.
  - Added automatic CUDA launch block-size selection for deformable collision detection.

- **Performance**
  - Improved CUDA deformable self-collision performance through adaptive workload tuning.

- **Documentation**
  - Documented automatic block-size selection when no explicit collision block size is provided.

- **Tests**
  - Added coverage for collision block-size selection across supported devices and workload sizes.

### CODE REVIEW

File: asv/benchmarks/simulation/bench_cloth.py [hunk 1]
```diff
@@ -15,10 +16,94 @@
...
+
+        resolution, world_count = case
+        builder = newton.ModelBuilder()
+        builder.replicate(_make_collision_world(resolution), world_count)
+        self.model = builder.finalize(device=device)
+        self.detector = TriMeshCollisionDetector(
+            self.model,
+            init_collision_info=True,
+            topological_contact_filter_threshold=0,
+            vertex_collision_buffer_pre_alloc=32,
```
mzamoramora-nvidia: Minor benchmark nit: could we increase `vertex_collision_buffer_pre_alloc` to 64? Both benchmark cases reach 34 vertex-triangle contacts against the current capacity of 32, so a small portion of that workload is truncated. The edge-edge path being measured does not overflow, so this is non-blocking.

mzamoramora-nvidia:
I reviewed this locally on an RTX PRO 6000. The focused cloth-collision suite and ASV benchmark tests pass on CPU/CUDA, and the adaptive launch policy improves the realistic cloth-twist case as well as both large synthetic workloads. For visual inspection, I ran the following examples:

```bash
uv run --extra examples -m newton.examples cloth_bending
uv run --extra examples -m newton.examples cloth_hanging
uv run --extra examples -m newton.examples cloth_rollers
uv run --extra examples -m newton.examples cloth_poker_cards
uv run --extra examples -m newton.examples softbody_gift
uv run --extra examples -m newton.examples softbody_dropping_to_cloth
uv run --extra examples -m newton.examples mujoco_vbd_coupled_solver
```

They all looked fine. The implementation stays within the existing internal detector. Using an AI-assisted review, I found only minor, non-blocking comments. An additional review from @jumyungc or @AnkaChan would be great before merging.

newton/_src/geometry/tri_mesh_collision.py
```diff
@@ -995,9 +997,18 @@ def edge_edge_collision_detection(
             ],
             dim=self.model.edge_count,
             device=self.model.device,
-            block_dim=self.collision_detection_block_size,
+            block_dim=self.collision_detection_block_size or self._edge_collision_block_size(),
```
mzamoramora-nvidia: Minor robustness nit: would it be clearer to distinguish `None` explicitly here and in the vertex-triangle launch? With `or`, an explicit `collision_detection_block_size=0` selects the fallback even though Warp normally treats a non-positive CUDA `block_dim` as its own default. This detector is internal and no repository caller passes zero, so I do not think this needs to block the PR; an `is None` branch or explicit positive-value validation would simply make the intended contract clearer.

changelog/+deformable-collision-block-size-6f2a9c1d.changed.md
```diff
@@ -0,0 +1 @@
+Speed up CUDA deformable edge-edge self-collision detection by choosing the launch block size from the edge count and GPU size; `TriMeshCollisionDetector(collision_detection_block_size=None)` now means automatic.
```
jumyungc: Nit: Could we also mention that this replaces the previous fixed block size of 16?

nvtw: Done

jumyungc (approved):

PR #4158, opened 2026-09-04
Run DR Legs policy with Warp-NN

nvtw:
## Description

Replace the PyTorch policy and observation path with native Warp arrays and ONNX inference. Pin the asset revision containing the converted model so the example works from a fresh checkout.

Keep floating-root coordinates out of the 94-value policy input and lazily create Torch views for other RL examples.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * The DR Legs reinforcement-learning example now runs entirely with Warp and Warp-NN using ONNX policies, without requiring PyTorch.
  * Added Warp-native simulation support for observations, actions, contacts, and state.
  * Joystick path tracking can be enabled or disabled through configuration.
  * Added validation for policy files, model inputs, outputs, and runtime data types.

* **Bug Fixes**
  * Improved asynchronous input handling for examples with custom input polling.

* **Tests**
  * Added coverage for Torch-free imports, Warp-array policies, and observation construction.


newton/_src/solvers/kamino/examples/rl/onnx_policy.py
```diff
@@ -0,0 +1,49 @@
...
+        self.input_name = self.runtime.input_names[0]
+        self.output_name = self.runtime.output_names[0]
+        output_shape = self.runtime._shapes[self.output_name]
+        expected_output_shape = (batch_size, action_width)
+        if output_shape != expected_output_shape:
+            raise ValueError(f"Policy '{path}' output shape must be {expected_output_shape}, got {output_shape}")
+
+    def __call__(self, observation: "torch.Tensor") -> "torch.Tensor":
+        """Evaluate a contiguous float32 Torch observation batch."""
+        import torch
```
Guirec-Maloisel: Question on this function: is this only needed if one wants a torch view on this policy, or is this required in general to use the onnx policy? (i.e. does the dependency towards torch truly disappear)?

nvtw: I have not fully removed the torch dependency yet because I was not sure if people want to do comparisons. But maybe it's better to completely remove the torch dependency. Will do

Guirec-Maloisel: I think it's fine if the dependency remains for the unit tests (there it makes sense to compare), but not for running the policy itself.

Guirec-Maloisel (commented):
The RL example (example_rl_drlegs.py) now loads the onnx policy, but this file still imports and uses torch, so I'm a bit confused in the end about removing the torch dependency? But could be that I misunderstood the intent of the PR.

Guirec-Maloisel (approved):
This is good to merge from my side now, sorry for the delay. Currently the policy isn't working well in practice (the robot is falling), but that is already the case on main. So we'll have to investigate separately when this broke; but I see no reason to delay the infrastructure change.

Issue #4000, opened 2026-08-21
Startup time improvements


camevor:
Epic for collecting startup-time improvements.

jcarius-nv:
Just for the record, as discussed with @camevor . The remaining work in this epic is:

- Merge already opened PRs in Newton and IsaacLab
- Figure out and ideally implement the builder.replicate_and_finalize() method
- Add some additional metrics to existing startup time benchmarks for regression coverage
- Compile a little report with findings, improvements, and potential future work.

PR #4159, opened 2026-09-04
Array-backed `ModelBuilder` attributes

camevor:
### Background

In the typical, homogeneous RL workflow, `ModelBuilder` is used in two very different ways:
1. Initially, with fairly small collections to conveniently import and set up a single world. This stage involves adding, removing, and modifying elements of collections containing on the order of hundreds of elements.
2. Subsequently, replication is run and the same world is duplicated thousands of times before it is finalized. The largest containers now contain on the order of half a million elements. In most cases, no major modifications are made after replication.

This presents some representation challenges, as for (1), Python lists (sometimes containing individual objects like Warp transforms) are convenient, while for (2), they become very inefficient compared with an array-based representation.

To give an example of typical numbers, for a 4,096-world G1 setup, (1) will handle 3,700 elements across 56 attributes (with at most 104 elements in any one attribute), while (2) will handle 15.2 million elements across the same attributes. Since #3415, `replicate()` performs its batched work using NumPy arrays. However, those have to be converted back to the Python-container representation before finalization, only to be converted back to NumPy (and ultimately device-resident Warp) arrays.

### Array-backed `ModelBuilder` attributes
This PR introduces array-backed `ModelBuilder` attributes: Many `ModelBuilder` attributes can optionally be represented as NumPy arrays instead of Python lists. They can remain as arrays across the boundary between `replicate()` and `finalize()`. An array-backed attribute can be accessed as a NumPy array within the private `_raw_array_access()` context manager, but will be converted to its normal Python list representation on first access outside such a scope.

This preserves the existing list-valued attribute interface for code outside `_raw_array_access()`. Only internal code that opts into raw access needs to handle both lists and NumPy arrays.

### Further `ModelBuilder` optimizations
A few more optimizations have been folded in:

- Disable the Python cyclic garbage collector during `replicate()` and `finalize()`. By their nature, `replicate()` and `finalize()` do not create large numbers of reference cycles. Thus, cyclic garbage collections mostly add overhead while reclaiming little. The previous gc state is restored afterward. We could not measure any deferred-gc penalty compared to main.
- Vectorize joint articulation-membership validation, and return early when every joint already belongs to an articulation.
- Convert array-backed attributes to Python lists for efficient looping where Python loops are necessary, without materializing the corresponding attribute.
- Avoid constructing world offsets and identity transforms when replication uses zero spacing.

### Results
Benchmark results are for 4096 world Robot Anymal D and Robot G1 examples (with ground plane construction moved ahead of replicate), medians of 5 runs.

| Task / schema | Startup through finalize, baseline → PR | `replicate + finalize`, baseline → PR |
|---|---:|---:|
| G1, with MuJoCo attributes | 9.790 → 6.760 s (**−30.9%, 1.45×**) | 6.621 → 3.618 s (**−45.4%, 1.83×**) |
| G1, w/o MuJoCo attrs | 8.425 → 5.550 s (**−34.1%, 1.52×**) | 5.298 → 2.438 s (**−54.0%, 2.17×**) |
| Anymal, with MuJoCo attributes | 3.386 → 2.268 s (**−33.0%, 1.49×**) | 2.091 → 0.976 s (**−53.3%, 2.14×**) |
| Anymal, w/o MuJoCo attrs | 3.319 → 2.233 s (**−32.7%, 1.49×**) | 2.030 → 0.930 s (**−54.2%, 2.18×**) |

### Follow-ups
* Custom attributes currently don't enjoy array-based replication, and their values are stored separately; hence, they do not benefit from the array-backed attributes. This should be addressed as a follow-up, as custom attribute finalization is now the dominant cost in workloads such as G1 startup.
* The documentation should be updated to mention the performance benefits of avoiding builder modification between `replicate()` and `finalize()`, and examples should be adapted accordingly.
 
Part of #4000.

## Checklist

## Test plan
```
uv run --extra dev -m newton.tests -k test_builder_replicate
```

## Summary by CodeRabbit

* **New Features**
  * Added NumPy-backed builder storage with automatic list materialization during standard access.
  * Added support for constructing rods from prepared rod objects, chains, and graphs.
  * Added rigidity-based stiffness derivation and junction collision filtering for rod assemblies.

* **Reliability**
  * Improved validation for rod stiffness, topology, world merging, and mesh preservation.
  * Garbage-collection settings are restored after builder operations, including failures.

* **Breaking Changes**
  * Removed the deprecated public shape contact-pair discovery method.
  * `add_rod_graph` now emits a deprecation warning.


eric-heiden (approved):
Nice speedup, and it is well contained: the array backing sits behind the existing attribute access, so no caller changes are needed. The one thing worth a follow up is how easily an ordinary read gives the array backing back, which cancels part of the win in the ordering most examples use.

🔵 `add_shape()` touches around 20 of the `shape_*` attributes, so a single `add_ground_plane()` after `replicate()` materializes all of them back to Python lists. That ordering is what most examples do today (`example_robot_g1.py`, `example_basic_urdf.py`, `example_robot_h1.py`, `example_robot_ur10.py`, and others), so those paths hand `finalize()` the same lists as before. The count properties are affected the same way, since `body_count` is `len(self.body_q)` and `particle_count` is `len(self.particle_q)`, so reading a count after `replicate()` turns an O(1) lookup into a full materialization of that family. It compounds when `replicate()` is called repeatedly with reads in between, because each call then re-copies a prefix that is back in list form.

End to end this is still a clear win, `replicate()` + `add_ground_plane()` + `finalize()` at 2048 worlds is about 2x faster than `main` even in the materializing order, so this is fine to handle separately. When you get to it, keeping a small tail list alongside the array and appending into that, flushed on the next raw array read, would let `add_shape()` and `add_particle()` run after `replicate()` without giving up the backing. A cheaper first step would be to have the count properties read the array length under `_raw_array_access()`, so a bare count read never materializes.

PR #4161, opened 2026-09-05
Warn when non-convex meshes hit MuJoCo hull path

### LINKED ISSUE #4065: [BUG] Non-convex MESH colliders are silently convexified on the MuJoCo solver, but the docs say they are supported
`docs/concepts/collisions.rst:446` documents `MESH` as:

> Triangle mesh (arbitrary, including non-convex)

On `SolverMuJoCo` that is not what happens. `geom_type_mapping`
(`solver_mujoco.py:5900-5910`) sends **both** `GeoType.MESH` and
`GeoType.CONVEX_MESH` to `mujoco.mjtGeom.mjGEOM_MESH`, and MuJoCo compiles
every mesh geom through its convex-hull path. The cavity of a bowl, cup, tube,
C-channel, bin or room shell does not exist at simulation time. There is no
error and no warning.

Repro, newton 1.5.0 / warp 1.16.0 / mujoco 3.11.0. A U-channel (floor + two
walls) is built as ONE triangle mesh; a sphere is dropped into the channel.

```python
import numpy as np, newton, warp as wp

def box_tris(cx, cy, cz, hx, hy, hz):
    v = np.array([[sx*hx+cx, sy*hy+cy, sz*hz+cz]
                  for sx in (-1, 1) for sy in (-1, 1) for sz in (-1, 1)], dtype=np.float32)
    f = [(0,1,3),(0,3,2),(4,6,7),(4,7,5),(0,4,5),(0,5,1),
         (2,3,7),(2,7,6),(0,2,6),(0,6,4),(1,5,7),(1,7,3)]
    return v, np.array(f, dtype=np.int32)

parts = [(0, 0, 0.05, 0.6, 0.6, 0.05),      # floor,  top z = 0.10
         (0, 0.40, 0.30, 0.6, 0.2, 0.20),   # +y wall, top z = 0.50
         (0, -0.40, 0.30, 0.6, 0.2, 0.20)]  # -y wall
V, F, off = [], [], 0
for p in parts:
    v, f = box_tris(*p); V.append(v); F.append(f + off); off += len(v)
V = np.concatenate(V); F = np.concatenate(F)

b = newton.ModelBuilder()
b.add_shape_mesh(-1, mesh=newton.Mesh(V, F.flatten()))
ball = b.add_body(xform=wp.transform(wp.vec3(0., 0., 0.9), wp.quat_identity()))
R = 0.08
b.add_shape_sphere(ball, radius=R)
m = b.finalize()

s = newton.solvers.SolverMuJoCo(m)
s0, s1 = m.state(), m.state(); c = m.control()
for _ in range(3000):
    ct = m.collide(s0); s.step(s0, s1, c, ct, 1.0/480.0); s0, s1 = s1, s0
print("rest z = %.4f" % float(s0.body_q.numpy()[ball][2]))
```

```
  inside the channel (floor 0.10 + r) = 0.1800
  on the convex hull (walls 0.50 + r) = 0.5800
  MEASURED                            = 0.5796
```

The ball rests on the wall tops, matching the hull prediction to four decimals.

Two consequences worth separating:

1. **`add_shape_convex_hull` is indistinguishable from `add_shape_mesh` on this
   path.** Both reach the same `spec.add_mesh(...)`. We measured identical rest
   heights for a hull and a mesh of the same non-convex geometry in our own
   integration, so the `CONVEX_MESH` GeoType carries no meaning here.

2. **The silence is the actual problem, not the hulling.** MuJoCo genuinely has
   no non-convex mesh primitive, so hulling is a reasonable thing to do. But at
   the same call site a *planar* mesh raises a hard `ValueError`
   (`solver_mujoco.py:330/352/367`), while a non-convex one is accepted without
   a word. A one-line warning naming the shape would have saved us a long
   misattribution, and would let a user reach for `approximate_meshes("coacd")`
   knowingly.

The doc line at `collisions.rst:446` reads as a capability guarantee, so it is
worth reconciling either way.

Happy to send a PR for the warning and the docs wording if that is welcome.

Context: we use newton as the only physics backend in an external simulator and
hit this as a "bodies rest too high" bug that took real time to attribute --
the collider looks correct everywhere you can inspect it.

deepspace28:
## Description

The MuJoCo solver compiles every mesh geom through a convex-hull path, so a non-convex `MESH` collider silently loses its cavities: a bowl, tube or C-channel behaves as its hull, and bodies can come to rest on the hull surface. This PR detects that case and warns the user instead of failing silently.

It adds `is_mesh_convex()` to `newton._src.geometry.utils` — an exact, winding-agnostic convexity test (a face plane violates convexity iff it separates the vertex set into points strictly on both sides) with a `max_face_vertex_pairs` cost guard. The solver computes the verdict once per unique mesh asset (cached in `mesh_export_cache`; convexity is scale-invariant) and warns only when MuJoCo generates the contacts, i.e. when the hull approximation actually affects the simulation. The warning points to the two resolutions: `ModelBuilder.approximate_meshes("coacd")` or `use_mujoco_contacts=False`.

Closes #4065.

## Checklist

- [x] The documentation is up to date with these changes (note added to `docs/concepts/collisions.rst`)
- [x] For user-facing changes, a fragment has been added by following the [changelog fragment instructions](https://github.com/newton-physics/newton/blob/main/changelog/README.md) (`changelog/4065.fixed.md`)

## Test plan

```
uv run --extra dev -m newton.tests -k TestIsMeshConvex
uv run --extra dev -m newton.tests -k nonconvex
uv run --extra dev -m newton.tests -k convex_hull_shape
```

- `test_nonconvex_mesh_contacts_warning` fails without the solver change (`AssertionError: UserWarning not triggered`); the `test_is_mesh_convex` module fails to import without the utils change.
- 10 unit tests + 3 solver tests pass with the change; the surrounding 44 `-k convex_hull` tests stay green.

## Bug fix

**Steps to reproduce:**

1. Build a model with a non-convex closed triangle mesh collider (e.g. a U-channel made of three joined boxes) and let the MuJoCo solver generate contacts.
2. Observe: no warning, and a ball dropped into the cavity comes to rest on the convex hull instead of the floor.

**Minimal reproduction:**

```python
import warp as wp
import newton
from newton import Mesh, ModelBuilder, SolverMuJoCo

verts, faces = u_channel_mesh()  # closed, non-convex triangle mesh
builder = ModelBuilder()
builder.add_shape_mesh(-1, mesh=Mesh(verts, faces.flatten()), label="u_channel")
ball = builder.add_body(xform=wp.transform(wp.vec3(0.0, 0.0, 0.9), wp.quat_identity()))
builder.add_shape_sphere(ball, radius=0.08)
model = builder.finalize()

solver = SolverMuJoCo(model)  # pre-PR: silent; post-PR: UserWarning about the convex hull
```

## Summary by CodeRabbit

* **Bug Fixes**
  * MuJoCo now warns when non-convex mesh colliders are exported and may be represented as convex hulls, which can omit cavities.
  * Large meshes whose convexity cannot be verified now receive a separate warning instead of being silently processed.

* **Documentation**
  * Added guidance on solver-specific mesh collision behavior and ways to address warnings, including mesh approximation, convex hulls, or Newton-managed contacts.

* **Tests**
  * Added coverage for non-convex, convex-hull, Newton-contact, and oversized mesh scenarios.


deepspace28:
Good catch, and verified before fixing: with the raw comparison, the u-channel downscaled by 1e-3 had every separating distance around 4e-10 against an eps of 1e-6, so it reported convex. Fixed in 0c849e8e - distances are now divided by the plane-normal magnitude (true point-to-plane distance), plus regression tests for the downscaled non-convex and enlarged convex cases. Note the api-changes label: the public addition of newton.geometry.is_mesh_convex was requested in review, so that advisory is expected.

nvtw:
I had a look:

I think the warning is useful and solves a real problem. However, I would prefer to keep is_mesh_convex() private for now.

The current function returns True not only for convex meshes, but also when the mesh is too large to check. The default limit is already reached by meshes with roughly 1,600 vertices. In
that case, a non-convex mesh can still be silently convexified, which is the problem this PR is trying to solve.

Could we use three internal results instead: convex, non-convex, and unknown? For an unknown result, the solver could give a softer warning. Once the function has a clear and reliable
contract, we can still make it public later if users need it.

So my preference is: keep the warning and documentation, but keep the detection helper private and do not treat a skipped check as convex.

deepspace28:
Thank you for the review — all three points are addressed in 1dbc08ca:

- **Private helper:** renamed to `newton._src.geometry.utils._is_mesh_convex` and removed from the `newton.geometry` public namespace, `__all__`, and the API docs. The `api-changes` advisory should go away with this commit.
- **Skipped checks are no longer convex:** the helper returns a three-valued result now. `True` is only for verified convexity or meshes too small to enclose a cavity (their hull equals the mesh, so hulling loses nothing). Missing indices and the cost guard both return `None`. Your 1,600-vertex case now lands in `unknown`.
- **Softer warning for unknown:** the solver keeps verdicts in a separate `mesh_convexity_cache` (so `None` is distinct from "not yet evaluated") and emits a second, hedged warning for unverified meshes — it states that cavities *may* not be simulated instead of asserting they are lost, and suggests the same two remedies.

A regression test covers the oversized path: a midpoint-subdivided U-channel (~1.7k verts x ~9.2k faces) that trips the default budget, asserting the unverified-convexity warning fires.

PR #4163, opened 2026-09-07
Fix linesearch warnings in conveyor tests

jvonmuralt:
## Description
The minimum-deps nightly failed on `test_long_running_stability_cuda_0`.  the failure came from unexpected stdout.

MuJoCo Warp 3.12.0, adopted in #4119, added device-side overflow diagnostics
(google-deepmind/mujoco_warp#1579). One of them prints when the Newton solver's
linesearch does not converge within `ls_iterations`. Newton's test harness fails
a test on any unexpected output.
The message text is also misleading: `"please increase ls_iterations to 50"`
prints the *current* value, so it asks you to increase a limit to what it already
is. Already corrected upstream in google-deepmind/mujoco_warp#1611 to say
"beyond".

### The fix

1. **`ls_iterations=100` for the conveyor tests.** The scene converges in 28-30
   iterations on an RTX 6000 Ada but exceeded the default of 50 on the CI runner.
   The linesearch breaks out as soon as it converges, so the higher cap is free.

2. **Tolerate the message via `NewtonTestCase.allowOutputRegex`.** The number of
   iterations actually required is hardware dependent and non deterministic, so the raised cap is a margin rather than a guarantee.
   The regex accepts both the current wording and the corrected upstream one.

## Checklist

## Test plan

```python
import newton

```

## Summary by CodeRabbit

* **Tests**
  * Updated conveyor-force validation to use a more robust solver configuration.
  * Restored standard test output handling and warning behavior for more representative test execution.

### DIFF
--- a/newton/tests/test_conveyor_forces.py
+++ b/newton/tests/test_conveyor_forces.py
@@ -16,7 +16,7 @@ import warp as wp
 
 import newton
 from newton.examples.basic.example_basic_conveyor_forces import ConveyorForceModel
-from newton.tests.unittest_utils import add_function_test, get_test_devices
+from newton.tests.unittest_utils import NewtonTestCase, add_function_test, get_test_devices
 
 BELT_HALF_X = 1.5
 BELT_HALF_Y = 6.0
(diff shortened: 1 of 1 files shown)


jcarius-nv:
Thanks for the quick fix. I see you're both increasing ls_iterations=100 and also ignoring the warning in the test. I was wondering why both? Isn't just increasing the limit good enough?

jvonmuralt:
@jcarius-nv  I added the description with explanation. It might be enough but since it is not deterministic I thought to be on the safe side... but I can also drop and see if it happens again, let me know what you prefer

jcarius-nv:
I think we'd rather want to see the errors. If 100 line search iterations are not good enough, something else might be off and I'd err on the side of seeing errors rather than silencing them.

jcarius-nv (approved):

Issue #4051, opened 2026-08-26
[REQ]  USD lighting and visual fidelity parity between ViewerRTX and Kit viewer


matthewtrepte:
### Description

 We'd like the Newton RTX viewer to reach visual parity with the Kit viewer for the same Isaac Lab environment. This covers USD lighting (dome light, distant light, intensity, color temperature), tone mapping, material shading, and render environment configuration. Currently ViewerRTX exposes only a single rtx_environment preset string ("default", "studio", "none") with no ability to configure individual lighting or render quality parameters. 

Here is Kit and Newton RTX visualizer for the same IL env. The lighting, sky color, and materials. Some of this is a larger scoped request, and I can create separate tickets for.  The missing velocity arrow markers in the RTX visualizer is covered by a separate Newton ticket.
https://github.com/user-attachments/assets/a0f09b08-f5eb-4635-a602-44bd5ec064b8
https://github.com/user-attachments/assets/7ef15ecb-3270-4fb8-95f1-566f52b1cf5a

### Motivation / Use Case

Isaac Lab is deprecating the Kit viewer in favor of Newton RTX as the primary interactive renderer.

jcarius-nv:
Related existing gaps: #2440 tracks RTX render-quality options, while #3625 tracks consistent GL/USD lookdev and an RTX-compatible USD lighting rig. This issue is broader Kit parity across lighting, tone mapping, materials, and environment configuration. Keeping these separate and cross-linking them because they likely share renderer and settings work.

eric-heiden:
After our discussion with @jcarius-nv, @mzamoramora-nvidia and @andrewkaufman today we will investigate how ovnewton could be used to parse non-physics properties. Rather than having to handle the visual material/shader parsing in Newton's USD importer (and then constructing a USD stage again from Model), we could rely on the original USD stage (ovstage) that we directly pass to ovrtx to retain all the visual information.

Issue #3646, opened 2026-07-24
[REQ] Preserve PBR visual materials in Model for ovrtx rendering


ruziniuuuuu:
### Description

Newton's runtime `Model` currently does not retain enough visual material data to represent fully textured PBR materials. Please add a PBR visual-material representation to `Model` and propagate it from USD import to renderer backends.

At minimum, the runtime could retain the bound `UsdMaterial` graph (or a reference to it) for each visual shape and make it available to renderers such as `ovrtx`. A fuller implementation could preserve common PBR inputs and texture bindings such as base color, metallic, roughness, normal, emissive, and opacity.

This request concerns visual materials and is separate from Newton's physics/contact materials.

### Motivation / Use Case

I would like to use Newton for physics simulation with `ovrtx` as the renderer. OpenUSD and `ovrtx` already support PBR materials, but the Newton runtime currently cannot pass the complete material information through its `Model` to the renderer.

Preserving the material graph or equivalent PBR data would allow faithful rendering without maintaining a separate USD scene and manually synchronizing transforms.

This was suggested in [urdf-usd-converter#110](https://github.com/newton-physics/urdf-usd-converter/issues/110#issuecomment-5071787579). [#2198](https://github.com/newton-physics/newton/issues/2198) is related but focuses on per-face materials through `UsdGeomSubset`.

### Alternatives Considered

- Pass the original USD stage directly to `ovrtx` and manually synchronize transforms with Newton.
- Reduce materials to diffuse textures and scalar values.
- Maintain a separate renderer-specific scene, at the cost of duplicated scene state and synchronization.

PR #4170, opened 2026-09-08
Preserve roughness textures in USD viewers

maxkra15:
## Description

Preserve roughness textures when importing USD visual meshes and forward them through Newton's viewer interface to `ViewerUSD` and `ViewerRTX`.

This change:

- Adds `roughness_texture` and `roughness_texture_influence` to `newton.Mesh` visual appearance.
- Imports roughness maps from `UsdPreviewSurface` networks and common direct MDL/OmniPBR inputs.
- Preserves OmniPBR's scalar-to-texture influence, including its disabled-by-default behavior when no influence is authored.
- Authors roughness maps as raw `UsdUVTexture` data, using the red channel and `scale`/`bias` to reconstruct `(1 - influence) * roughness + influence * texture`.
- Preserves mesh copying, appearance hashing, instancing, and the common viewer API.

This is a focused, useful subset of full PBR preservation. Base-color and roughness maps currently share Newton's existing `texture_transform`; independent per-map transforms, normal/metallic/emissive maps, and arbitrary material graphs remain follow-up work.

Partially addresses #3646. Related to #4051.

## Checklist

## Test plan

```text
uvx pre-commit run -a
uv run --extra dev -m newton.tests -k test_mesh_cache
uv run --extra dev -m newton.tests -k test_viewer_usd
uvx --from towncrier==25.8.0 towncrier build --draft --version X.Y.Z --date 2026-09-08
```

Results:

- Pre-commit: passed.
- Mesh cache: 14 passed.
- USD/RTX viewer: 23 passed.
- USD import: 566 passed, 1 skipped.
- Towncrier draft: rendered successfully.
- Imported Isaac Lab's checker ground-plane asset and verified that Newton retained its albedo map, roughness map, scalar roughness (`1.0`), texture influence (`1.0`), and metric UVs.

## Bug fix

**Steps to reproduce:**

1. Bind a visual mesh to an OmniPBR material with `reflectionroughness_texture` and `reflection_roughness_texture_influence`.
2. Import the stage with `ModelBuilder.add_usd()` and render it with `ViewerRTX`.
3. On `main`, observe that Newton discards the roughness map and renders the mesh with uniform scalar roughness.

**Minimal reproduction:**

```python
import newton

result = builder.add_usd("scene.usda")
mesh = builder.shape_source[result["path_shape_map"]["/World/VisualMesh"]]

assert mesh.roughness_texture is not None
assert mesh.roughness_texture_influence == 1.0
```

## New feature / API change

```python
import newton

mesh = newton.Mesh(
    vertices,
    indices,
    uvs=uvs,
    roughness=0.8,
    roughness_texture="roughness.png",
    roughness_texture_influence=0.25,
)
```

## Summary by CodeRabbit

* **New Features**
  * Added support for importing roughness textures from USD materials.
  * Added configurable scalar-to-texture roughness blending.
  * Added roughness-texture support for USD and RTX rendering.
  * Roughness textures preserve channels, UV transforms, sampling settings, and material metadata.
  * Added support for image arrays, paths, and URLs as texture sources.
* **Bug Fixes**
  * Improved roughness texture handling for material subsets and UV mapping.
  * Mesh copies and visual hashes now account for roughness texture data.
* **Documentation**
  * Added guidance on roughness texture support across visualization backends.

### DIFF
--- a/newton/_src/geometry/types.py
+++ b/newton/_src/geometry/types.py
@@ -182,6 +182,7 @@ class Mesh:
         roughness: float | None = None,
         metallic: float | None = None,
         texture: str | np.ndarray | None = None,
+        roughness_texture: str | np.ndarray | None = None,
         texture_transform: Sequence[Sequence[float]] | np.ndarray = ((1.0, 0.0, 0.0), (0.0, 1.0, 0.0)),
         sdf: "SDF | None" = None,
         opacity: float | None = None,
@@ -204,7 +205,8 @@ class Mesh:
             color: Optional per-mesh base color (values in [0, 1]).
             roughness: Optional mesh roughness in [0, 1].
             metallic: Optional mesh metallic in [0, 1].
-            texture: Optional texture path/URL or image data (H, W, C).
+            texture: Optional base-color texture path/URL or image data (H, W, C).
+            roughness_texture: Optional linear roughness texture path/URL or image data (H, W, C).
             texture_transform: Affine texture-coordinate transform as two rows
                 ``((m00, m01, tx), (m10, m11, ty))``. It is applied to the
                 authored UV coordinates as ``(u', v') = M @ (u, v) + t``.
@@ -224,6 +226,7 @@ class Mesh:
         self.opacity = opacity
         # Store texture lazily: strings/paths are kept as-is, arrays are normalized
         self._texture = _normalize_texture_input(texture)
+        self._roughness_texture = _normalize_texture_input(roughness_texture)
         self.texture_transform = texture_transform
         self._roughness = roughness
         self._metallic = metallic
@@ -240,6 +243,7 @@ class Mesh:
         self.maxhullvert = maxhullvert
         self._cached_hash = None
         self._texture_hash = None
+        self._roughness_texture_hash = None
         self._edges = None
         self._collision_edges: np.ndarray | None = None
         self._is_watertight: bool | None = None
@@ -810,6 +814,9 @@ class Mesh:
             texture=self._texture
             if isinstance(self._texture, str)
             else (self._texture.copy() if self._texture is not None else None),
+            roughness_texture=self._roughness_texture
+            if isinstance(self._roughness_texture, str)
+            else (self._roughness_texture.copy() if self._roughness_texture is not None else None),
             roughness=self._roughness,
             metallic=self._metallic,
             texture_transform=self._texture_transform,
@@ -1556,6 +1563,24 @@ class Mesh:
         return self._compute_texture_hash()
 
     @property
+    def roughness_texture(self) -> str | np.ndarray | None:
+        """Optional linear roughness texture as a file path or image array."""
+        return self._roughness_texture
+
+    @roughness_texture.setter
+    def roughness_texture(self, value: str | np.ndarray | None):
+        self._roughness_texture = _normalize_texture_input(value)
+        self._roughness_texture_hash = None
+        self._cached_hash = None
+
+    @property
+    def roughness_texture_hash(self) -> int:
+        """Content-based hash of the assigned roughness texture."""
+        if self._roughness_texture_hash is None:
+            self._roughness_texture_hash = compute_texture_hash(self._roughness_texture)
+        return self._roughness_texture_hash
+
+    @property
     def texture_transform(self) -> tuple[tuple[float, float, float], tuple[float, float, float]]:
         """Affine transform applied to the authored UV coordinates."""
         return self._texture_transform
@@ -1695,7 +1720,8 @@ class Mesh:
                     digest.update(int(dimension).to_bytes(8, "big"))
                 digest.update(values.tobytes())
             digest.update(bytes([bool(self.is_solid)]))
-            self._cached_hash = int.from_bytes(digest.digest()[:8], "big") ^ hash(self._compute_texture_hash())
+            texture_hashes = (self._compute_texture_hash(), self.roughness_texture_hash)
+            self._cached_hash = int.from_bytes(digest.digest()[:8], "big") ^ hash(texture_hashes)
         return self._cached_hash
 
     # ---- Factory methods ---------------------------------------------------
--- a/newton/_src/geometry/types.py
+++ b/newton/_src/geometry/types.py
@@ -183,6 +183,7 @@ class Mesh:
         metallic: float | None = None,
         texture: str | np.ndarray | None = None,
         roughness_texture: str | np.ndarray | None = None,
+        roughness_texture_influence: float = 1.0,
         texture_transform: Sequence[Sequence[float]] | np.ndarray = ((1.0, 0.0, 0.0), (0.0, 1.0, 0.0)),
         sdf: "SDF | None" = None,
         opacity: float | None = None,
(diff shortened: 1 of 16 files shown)


newton/_src/usd/utils.py
```diff
@@ -3180,13 +3208,69 @@ def _output_channel_count(type_name: Sdf.ValueTypeName) -> int:
...
+        source_shader = UsdShade.Shader(attr.GetPrim())
+        try:
+            is_uv_texture = source_shader.GetIdAttr().Get() == "UsdUVTexture"
+        except Exception:
+            is_uv_texture = False
+        if not is_uv_texture:
+            continue
+        texture = _resolve_uv_texture_asset(source_shader, prim, is_color=False)
+        if texture is not None:
+            return texture, _extract_usd_transform2d(source_shader)
```
jcarius-nv, line 3297: Could we preserve the source connection's sampling semantics rather than reducing it to only the asset and UV transform? This loses which output channel (`r`/`g`/`b`/`a`) feeds roughness, plus authored `scale`, `bias`, `fallback`, and wrap modes. ViewerUSD then reconstructs every map as red-channel/repeat with a new scale/bias formula. Packed maps can therefore sample the wrong channel, and inputs such as glossiness-to-roughness inversion change meaning.

newton/_src/geometry/types.py
```diff
@@ -204,10 +206,15 @@ def __init__(
             color: Optional per-mesh base color (values in [0, 1]).
             roughness: Optional mesh roughness in [0, 1].
             metallic: Optional mesh metallic in [0, 1].
-            texture: Optional texture path/URL or image data (H, W, C).
+            texture: Optional base-color texture path/URL or image data (H, W, C).
+            roughness_texture: Optional linear roughness texture path/URL or image data (H, W, C).
```
jcarius-nv: The advertised URL form is not currently honored by the USD viewers: `_resolve_texture_path()` sends every string through `os.path.abspath()`, so e.g. `https://example.com/rough.png` becomes a bogus local path. Could we either support URLs on this path or narrow the public contract to the inputs that actually work?

newton/_src/usd/utils.py
```diff
@@ -3148,16 +3165,27 @@ def _extract_preview_surface_properties(shader: UsdShade.Shader | None, prim: Us
...
+            roughness_texture, roughness_transform = _roughness_texture_from_input(roughness_input, prim)
+            properties["roughness_texture"] = roughness_texture
+            if roughness_texture is not None:
+                properties["roughness_texture_influence"] = 1.0
+            if properties["texture_transform"] is None and roughness_transform is not None:
+                properties["texture_transform"] = roughness_transform
+
+            # Preserve the authored input value as the renderer's fallback when
+            # the texture cannot be sampled.
+            properties["roughness"] = _coerce_float(roughness_input.Get())
```
jcarius-nv: Is it intentional to prioritize the connected destination input's authored value over the upstream scalar? Before this PR, a connected roughness input was resolved from the source shader. With this ordering, a destination value of `0.2` connected to a source shader with scalar roughness `0.8` imports as `0.2`; the base revision imports `0.8`. This changes the appearance of an unchanged asset even when no roughness texture is resolved. For backward compatibility, especially for 1.6.1, can we preserve the previous scalar resolution unless this change is agreed separately?

newton/_src/usd/utils.py
```diff
@@ -3317,7 +3423,15 @@ def _extract_material_input_properties(material: UsdShade.Material | None, prim:
...
-        if properties["texture"] is None and ("texture" in name_lower or "file" in name_lower):
+        if properties["roughness_texture"] is None and _is_roughness_texture_input_name(name):
+            roughness_texture, _ = _roughness_texture_from_input(inp, prim)
+            if roughness_texture is not None:
+                properties["roughness_texture"] = roughness_texture
+                roughness_texture_input_name = name
+                continue
+
+        is_generic_color_texture = name_lower in ("texture", "file")
+        if properties["texture"] is None and (is_generic_color_texture or _is_color_texture_input_name(name)):
```
jcarius-nv: Can we discuss the backward-compatibility impact of narrowing the existing Material-input heuristic here? Previously any unconnected input whose name contained `texture` or `file` could supply the color texture; now only exact `texture`/`file` or recognized color-name fragments do. For example, an unchanged `inputs:custom_texture = @albedo.png@` is imported on the base revision but dropped at this head. The old heuristic may be broad, but tightening it as part of this additive feature changes existing assets; could we preserve the old non-roughness behavior or handle that narrowing separately?

jcarius-nv (commented):
The overall direction makes sense to me: a roughness texture is a renderer-neutral material concept, so it can live in Newton's common model even if the first implementation only supports ViewerUSD/ViewerRTX. For this PR, please document the current backend support and degradation behavior (including what ViewerGL, ViewerRerun, and ViewerViser do), and link a follow-up issue—potentially #3646—to define the bounded portable material subset and capability matrix. Initial support need not land in every backend here, but the shared contract and fallback behavior should be explicit.

The inline comments below cover correctness and backward-compatibility concerns. I did not repeat the two existing UV-cardinality threads; those already capture the issues I found.

maxkra15:
@jcarius-nv thanks for the review, I implemented the following based on the feedback: 
- Preserved roughness sampler semantics: selected channel, scale, bias, fallback, wrap modes, and source color space
- Added HTTP(S) texture localization for USD viewers
- Restored upstream scalar precedence and the legacy material texture heuristic
- Fixed UV validation and UV duplication for both solidified-mesh paths
- Added a renderer-neutral texture descriptor and documented backend support/fallback behavior, with #3646 tracking broader PBR support
- Added regression coverage for all reported cases; local targeted tests, pre-commit, documentation, and Towncrier checks pass

jcarius-nv:
Thanks @maxkra15 for addressing the earlier concerns — I rereviewed the updated head, and the sampler semantics, URL handling, scalar and generic-material compatibility, UV handling, and backend documentation all look materially addressed.

One follow-up compatibility issue surfaced during the rereview: `ViewerBase.log_geo()` now forwards `roughness_texture` and `roughness_texture_influence` to the dynamically dispatched `log_mesh()`. An unchanged third-party `ViewerBase` subclass that implements the previous public `log_mesh()` signature therefore raises `TypeError: ... got an unexpected keyword argument 'roughness_texture'` when an existing USD asset contains a roughness map. We reproduced this at the current head.

Could we preserve compatibility for existing viewer subclasses, so unsupported backends degrade to scalar roughness rather than failing, and add a regression covering a legacy-signature subclass with a roughness-textured model path? This is particularly important for the proposed 1.6.1 delivery, where unchanged user code must continue to work.

maxkra15:
@jcarius-nv adressed

jcarius-nv:
Thanks @maxkra15 — I rereviewed the current head, and this iteration addresses the concerns raised in the earlier rounds. The roughness sampler semantics, URL handling, scalar and generic-material compatibility, UV cardinality and solidification paths, backend degradation policy, and roughness-only UV selection are now covered. So technically this is ready to go but after our discussionwe'll hold it to see if a more direct USD / ovstage passthrough to ViewerRTX can avoid introducing a lot of API changes.
If we can't land this for 1.7 then we can merge this anyways. Thanks again!

jcarius-nv:
=> re-visit once #4051 is addressed

PR #4172, opened 2026-09-08
Honor explicit MuJoCo buffer capacities

### LINKED ISSUE #4168: [BUG] Explicit SolverMuJoCo nconmax/njmax are overridden by conservative contact estimates
## Bug Description

Newton 1.6.0rc1 changes `SolverMuJoCo(use_mujoco_contacts=False)` so explicitly configured `nconmax` and `njmax` values are overridden by conservative topology-based estimates.

The regression was introduced by a2ad5d170ad540e1757fc55ba36307a265ff08f5. In the Newton-contact path, positive explicit values are now resolved as:

```python
nconmax = max(nconmax, default_nconmax)
njmax = max(njmax, default_njmax)
```

This means the arguments no longer act as user-selected capacities. The sizing happens during solver construction; this is not runtime buffer growth.

The change was intended to prevent injected Newton contacts from being truncated after the initialization-only MuJoCo contact pass was removed. However, using the collision pipeline's conservative capacity estimate as a mandatory lower bound can substantially over-allocate MJWarp buffers even when the explicitly configured values were measured and sufficient for the task.

In our Isaac Lab benchmark, upgrading from the parent revision to the culprit revision produced:

| Per-world capacity | Parent | a2ad5d1 | Increase |
| --- | ---: | ---: | ---: |
| Contacts (`nconmax`) | 200 | 1,320 | 6.6x |
| Constraint rows (`njmax`) | 300 | 5,644 | 18.8x |

The corresponding peak VRAM scaled approximately as follows on the same workload and hardware:

| Environment count | Parent | a2ad5d1 |
| ---: | ---: | ---: |
| 64 | 5.6 GiB | 6.4 GiB |
| 1,024 | 5.9 GiB | 8.6 GiB |
| 8,192 | 10.3 GiB | 29.8 GiB |

At 8,192 environments this is roughly a 19.5 GiB regression and approaches 3x the parent revision's peak VRAM.

The issue is especially visible in replicated simulation because `njmax` is per world and many MJWarp arrays scale with `nworld * njmax`. `naconmax` is aggregate across worlds, so it also scales with the resolved per-world contact capacity.

This conflicts with the practical purpose of explicit capacity settings. It also undermines the mitigation documented in #3402, where specifying an explicit collision capacity is the recommended way to control excessive automatic allocation.

### Expected behavior

- Auto-estimate `nconmax` and `njmax` when the corresponding argument is `None`.
- Honor explicitly supplied positive capacities as the requested limits.
- Report actual overflow through a warning/state bit or strict-capacity mechanism, rather than silently replacing the user's capacity with a worst-case topology estimate.
- Alternatively, provide an explicit opt-out from the lower-bound behavior if it must remain the default.

The overflow-reporting work in #4086 appears to provide the appropriate mechanism for making genuinely undersized explicit capacities observable.

## Reproduction Script

```python
import newton
import warp as wp

builder.add_ground_plane()

# Separated bodies produce few/no active contacts initially, but the topology
# estimate still reserves for the conservative set of possible shape pairs.
for i in range(60):
    body = builder.add_body(
        xform=wp.transform(
            wp.vec3(float(i % 10) * 2.0, float(i // 10) * 2.0, 10.0),
            wp.quat_identity(),
        )
    )
    builder.add_shape_sphere(body=body, radius=0.5)

model = builder.finalize()
newton.CollisionPipeline(model)

requested_nconmax = 200
requested_njmax = 300
    model,
    use_mujoco_contacts=False,
    nconmax=requested_nconmax,
    njmax=requested_njmax,
)

print("requested nconmax:", requested_nconmax)
print("resolved naconmax:", solver.mjw_data.naconmax)
print("requested njmax:", requested_njmax)
print("resolved njmax:", solver.mjw_data.njmax)

assert solver.mjw_data.naconmax > requested_nconmax
assert solver.mjw_data.njmax > requested_njmax
```

Newton's test `test_newton_contact_defaults_bound_explicit_capacities` also codifies the new behavior by passing `nconmax=16, njmax=16` and asserting that both resolved buffers exceed those explicit values.

## System Information

- Newton: `1.6.0rc1`, including a2ad5d170ad540e1757fc55ba36307a265ff08f5
- Warp: `1.17.0`
- MuJoCo: `3.12.0`
- MuJoCo Warp: `3.12.0`
- Integration: Isaac Lab `develop` after isaac-sim/IsaacLab@bc3ebea906c6e32cc3a5ec5a2fd2aad913c66fda
- GPU: NVIDIA CUDA; the capacity resolution itself is constructor/model dependent rather than GPU dependent

jcarius-nv:
## Description

Closes #4168.

Use Newton's conservative contact-topology estimates for `nconmax` and `njmax`
only when the corresponding `SolverMuJoCo` argument is `None`. Explicit values
now remain caller-selected capacities instead of minimum hints, avoiding large
per-world MuJoCo Warp allocations in replicated workloads.

PR #3825 removed an initialization-only MuJoCo contact pass that had acted as
an accidental sizing probe even when Newton generated contacts. Replacing that
probe with explicit automatic sizing was necessary, but applying the automatic
topology estimate as a mandatory lower bound to user-provided values caused the
memory regression reported in #4168.

This PR deliberately preserves the pre-existing warning-and-increase behavior
when an explicit value cannot contain contacts or constraints already present
in the initial MuJoCo state. Turning that compatibility behavior into a hard
error would expand this regression fix and break workflows that currently
continue after the warning. An opt-in hard-cap/strict behavior can instead be
considered with the overflow-reporting work in #4086.

## Checklist

## Test plan

- `uv run --extra dev newton/tests/test_mujoco_solver.py` — 298 passed, 7 skipped
- `uvx --python 3.12 pre-commit run -a` — passed
- Towncrier draft build for 1.7.0 — passed

The focused regression test was also verified to fail before the implementation
change: a requested `nconmax=200` resolved to `naconmax=3300`.

## Bug fix

**Steps to reproduce:**

1. Build a model with enough collision topology for the automatic estimate to
   exceed `nconmax=200` and `njmax=300`.
2. Construct `SolverMuJoCo` with Newton-generated contacts and those explicit
   capacities.
3. Observe that the resolved MuJoCo Warp buffers exceed both requested values.

**Minimal reproduction:**

```python
import newton
import warp as wp

builder.add_ground_plane()
for i in range(60):
    body = builder.add_body(
        xform=wp.transform(
            wp.vec3(float(i % 10) * 2.0, float(i // 10) * 2.0, 10.0),
            wp.quat_identity(),
        )
    )
    builder.add_shape_sphere(body=body, radius=0.5)

model = builder.finalize()
newton.CollisionPipeline(model)
    model,
    use_mujoco_contacts=False,
    nconmax=200,
    njmax=300,
)

assert solver.mjw_data.naconmax == 200
assert solver.mjw_data.njmax == 300
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Explicit `nconmax` and `njmax` settings are now honored instead of being silently increased based on estimated contact or constraint capacity.
  - If a configured capacity is too small for the model’s initial contacts or constraints, it is increased automatically and a warning is issued.

- **Documentation**
  - Clarified automatic capacity estimation when values are unset and when configured values may still be adjusted.

### DIFF
--- a/changelog/4168.fixed.md
+++ b/changelog/4168.fixed.md
@@ -1,0 +1,1 @@
+Honor explicit `SolverMuJoCo` `nconmax` and `njmax` capacities instead of increasing them to conservative Newton contact estimates.
--- a/docs/concepts/simulation_tuning_mujoco.rst
+++ b/docs/concepts/simulation_tuning_mujoco.rst
@@ -280,10 +280,13 @@ practice:
 ``nconmax`` and ``njmax`` size the **per-world** contact and constraint buffers.
 Set them for the busiest world, not the average: a buffer that fits a quiet
 world can truncate contacts or constraints in a heavier one, while an oversized
-buffer wastes GPU memory multiplied across every world. If left unset, they are
-estimated from the initial state; monitor overflow counters or warnings and raise
-the relevant buffer when needed. A positive gap can increase the number of
-detected contacts even though contacts outside the margin remain inactive.
+buffer wastes GPU memory multiplied across every world. If left unset, Newton
+estimates them automatically. Explicit values set fixed capacities instead of
+lower bounds for those estimates. A value that cannot hold the initial MuJoCo
+contacts or constraints may still be increased with a warning. Monitor overflow
+counters or warnings and raise the relevant buffer when needed. A positive gap
+can increase the number of detected contacts even though contacts outside the
+margin remain inactive.
 After changing gaps or upgrading margin/gap behavior, remeasure peak contacts,
 constraints, and overflow in the busiest world before compensating with
 stiffness or iterations; do not assume the previous run generated the same rows.
--- a/newton/_src/solvers/mujoco/solver_mujoco.py
+++ b/newton/_src/solvers/mujoco/solver_mujoco.py
@@ -3737,9 +3737,13 @@ class SolverMuJoCo(SolverBase, CouplingInterface):
         Args:
             model: The model to be simulated.
             separate_worlds: If True, each Newton world is mapped to a separate MuJoCo world. Defaults to `not use_mujoco_cpu`.
-            njmax: Maximum number of constraints per world. If None, a default value is estimated from the initial state. Note that the larger of the user-provided value or the default value is used.
+            njmax: Maximum number of constraints per world. If None, a default
+                value is estimated automatically. An explicit value is preserved
+                unless it cannot hold the initial MuJoCo constraints.
             njmax_nnz: Sparse constraint Jacobian nonzero capacity per world. If provided, must be non-negative and large enough for the initial sparse Jacobian. If None, derived from the model's constraint counts and njmax.
-            nconmax: Number of contact points per world. If None, a default value is estimated from the initial state. Note that the larger of the user-provided value or the default value is used.
+            nconmax: Number of contact points per world. If None, a default value
+                is estimated automatically. An explicit value is preserved unless
+                it cannot hold the initial MuJoCo contacts.
             iterations: Number of solver iterations. If None, uses model custom attribute or MuJoCo's default (100).
             ls_iterations: Number of line search iterations for the solver. If None, uses model custom attribute or MuJoCo's default (50).
             ccd_iterations: Maximum CCD iterations. If None, uses model custom attribute or MuJoCo's default (35).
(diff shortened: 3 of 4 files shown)


vreutskyy (dismissed):
Reviewed the capacity-sizing changes at a5e65365. Explicit nconmax/njmax values now bypass the conservative topology estimates while None retains automatic sizing, and the existing warning-based initial-state expansion remains intact. The three new regression tests and four adjacent automatic/per-world sizing tests passed locally. LGTM.

vreutskyy (approved):

PR #4175, opened 2026-09-08
Add USD parsing required to implement CollisionPipeline.create_from_usd()

gyeomannvidia:
## Checklist

Closes https://github.com/newton-physics/newton/pull/4175

## Test plan

```python
import newton

CollisionPipeline.create_from_usd(scene_prim, model)
```

## Summary by CodeRabbit

* **New Features**
  * Added support for creating collision pipelines from settings authored on a USD physics scene.
  * Explicit configuration overrides take precedence over USD-authored settings.
  * Supports typed USD scene objects and documented sentinel values that fall back to standard defaults.
  * Converts authored length settings from stage units to meters.
  * Reads contact reduction settings, collision matching options, and other pipeline configuration from USD.

* **Bug Fixes**
  * Added descriptive validation errors for invalid scene configuration and unsupported collision settings.
  * Out-of-range soft-contact settings now warn and use standard defaults instead of failing.

* **Documentation**
  * Added a changelog entry for USD-based collision pipeline configuration.

### DIFF
--- a/newton/_src/sim/collide.py
+++ b/newton/_src/sim/collide.py
@@ -7,7 +7,7 @@ import dataclasses
 import math
 import operator
 import warnings
-from typing import Any, Literal
+from typing import TYPE_CHECKING, Any, Literal
 
 import numpy as np
 import warp as wp
@@ -42,6 +42,9 @@ from ..geometry.types import GeoType
 from ..sim.contacts import Contacts
 from ..sim.model import Model
 from ..sim.state import State
+
+if TYPE_CHECKING:
+    from pxr import Usd, UsdPhysics
 
 
 def _shape_collide_mask(model: Model, shape_count: int | None = None) -> np.ndarray:
@@ -1871,7 +1874,9 @@ class CollisionPipeline:
         self._soft_self_contact_detector: TriMeshCollisionDetector | None = None
 
     @classmethod
-    def create_from_usd(cls, scene_prim: Any, model: Model, **overrides: Any) -> CollisionPipeline:
+    def create_from_usd(
+        cls, scene_prim: Usd.Prim | UsdPhysics.Scene, model: Model, **overrides: Any
+    ) -> CollisionPipeline:
         """Create a :class:`CollisionPipeline` from ``NewtonCollisionPipelineAPI``.
 
         Reads ``newton:collisionPipeline:*`` attributes off ``scene_prim``.
--- a/newton/_src/sim/collide.py
+++ b/newton/_src/sim/collide.py
@@ -1939,7 +1939,8 @@ class CollisionPipeline:
                 suffix = " or -1" if allow_minus_one else ""
                 raise ValueError(f"{path}: {name} must be non-negative{suffix}, got {result}.")
             if result == 0 and disallow_zero:
-                raise ValueError(f"{path}: {name} must be a positive integer or -1, got {result}.")
+                suffix = " or -1" if allow_minus_one else ""
+                raise ValueError(f"{path}: {name} must be a positive integer{suffix}, got {result}.")
             return result
 
         def token(name: str, value: Any, allowed: set[str]) -> str:
@@ -1994,7 +1995,9 @@ class CollisionPipeline:
 
         value = authored("newton:collisionPipeline:maxTrianglePairs")
         if value is not None:
-            kwargs["max_triangle_pairs"] = integer("newton:collisionPipeline:maxTrianglePairs", value)
+            kwargs["max_triangle_pairs"] = integer(
+                "newton:collisionPipeline:maxTrianglePairs", value, disallow_zero=True
+            )
 
         value = authored("newton:collisionPipeline:rigidContactMax")
         if value is not None:
@@ -2063,6 +2066,11 @@ class CollisionPipeline:
         value = authored("newton:collisionPipeline:contactReport")
         if value is not None:
             kwargs["contact_report"] = bool(value)
+            if kwargs["contact_report"] and kwargs.get("contact_matching", "disabled") == "disabled":
+                raise ValueError(
+                    f"{path}: newton:collisionPipeline:contactReport=True requires "
+                    "newton:collisionPipeline:contactMatching != 'disabled'."
+                )
 
         value = authored("newton:collisionPipeline:verifyBuffers")
         if value is not None:
--- a/newton/_src/sim/collide.py
+++ b/newton/_src/sim/collide.py
@@ -4,8 +4,10 @@
 from __future__ import annotations
 
 import dataclasses
+import math
+import operator
 import warnings
-from typing import Literal
+from typing import Any, Literal
 
 import numpy as np
 import warp as wp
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -168,6 +168,9 @@ torch = [
     { index = "pytorch-cu128", extra = "torch-cu12" },
     { index = "pytorch-cu130", extra = "torch-cu13" },
 ]
+# TODO(gyeoman): drop once NewtonCollisionPipelineAPI merges upstream and a
+# release of newton-usd-schemas ships it.
+newton-usd-schemas = { git = "https://github.com/gyeomannvidia/newton-usd-schemas", branch = "dev/gyeoman/collision-pipeline-api" }
```
gyeomannvidia, line 173: This needs to be removed before merging

newton/_src/sim/collide.py
```diff
@@ -1868,6 +1870,199 @@ def __init__(
...
+        if value is not None:
+            kwargs["reduce_contacts"] = bool(value)
+
+        value = authored("newton:collisionPipeline:softContactMax")
+        if value is not None:
+            result = integer("newton:collisionPipeline:softContactMax", value, allow_minus_one=True)
+            if result != -1:
+                kwargs["soft_contact_max"] = result
+
+        value = authored("newton:collisionPipeline:softContactGap")
```
mzamoramora-nvidia, line 2027: Could we convert these distance values from USD stage units to meters before passing them to `CollisionPipeline`? Newton uses SI units, while USD length attributes follow the stage's `metersPerUnit`. For example, a value of `2` on a centimeter stage currently becomes `2 m` instead of `0.02 m`. This also applies to `contactMatchingPosThreshold` and `speculativeMaxExtension`.

newton/_src/sim/collide.py
```diff
@@ -1868,6 +1870,199 @@ def __init__(
...
         self._soft_self_contact_detector: TriMeshCollisionDetector | None = None
 
+    @classmethod
+    def create_from_usd(cls, scene_prim: Any, model: Model, **overrides: Any) -> CollisionPipeline:
+        """Create a :class:`CollisionPipeline` from ``NewtonCollisionPipelineAPI``.
+
+        Reads ``newton:collisionPipeline:*`` attributes off ``scene_prim``.
+        Only authored USD values override the :meth:`__init__`
+        defaults, so unauthored attributes fall back to the same defaults as
+        constructing :class:`CollisionPipeline` directly. ``overrides`` take
```
mzamoramora-nvidia, line 1885: Could we add a small test for override precedence? For example, if the stage authors `broadPhase = "nxn"` and the caller passes `broad_phase="sap"`, we should verify that the resulting pipeline uses `"sap"`, as documented.

pyproject.toml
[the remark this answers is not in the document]
gyeomannvidia: This is temporary and will be removed when NewtonCollisionAPI has been pushed (https://github.com/newton-physics/newton-usd-schemas/pull/89)

newton/_src/sim/collide.py
gyeomannvidia: Added in 927c7fc679e82b5e177c7b36f0d38327ecb23dbf

newton/_src/sim/collide.py
gyeomannvidia: Test added in 00e1e57cbacbbe799726affbc74b02dea4a98dbf

newton/_src/sim/collide.py
```diff
@@ -1868,6 +1873,241 @@ def __init__(
         # _get_soft_self_contact_detector).
         self._soft_self_contact_detector: TriMeshCollisionDetector | None = None
 
+    @classmethod
+    def create_from_usd(
+        cls, scene_prim: Usd.Prim | UsdPhysics.Scene, model: Model, **overrides: Any
+    ) -> CollisionPipeline:
+        """Create a :class:`CollisionPipeline` from ``NewtonCollisionPipelineAPI``.
```
jcarius-nv, line 1880: **P1 — Mark this exact public factory as experimental.** The schema package explicitly treats its v0.x schemas as experimental, and this method's contract is the mapping from that evolving schema into a live pipeline. Newton's API guidelines require the exact callable to carry the experimental marker and relevant limitations, rather than relying on package-level status elsewhere.

newton/tests/test_collision_pipeline.py
```diff
@@ -352,6 +353,361 @@ def test_soft_contact_max_zero_disables_soft_contact_generation(self):
...
+        self.assertAlmostEqual(pipeline.soft_contact_gap, default_pipeline.soft_contact_gap)
+        scene_prim.GetAttribute("newton:collisionPipeline:softContactGap").Clear()
+
+        scene_prim.GetAttribute("newton:collisionPipeline:speculativeMaxExtension").Set(-0.1)
+        with self.assertWarnsRegex(UserWarning, r"newton:collisionPipeline:speculativeMaxExtension=-0\.\d+ is invalid"):
+            pipeline = CollisionPipeline.create_from_usd(scene_prim, model)
+        self.assertEqual(pipeline.speculative_config, default_pipeline.speculative_config)
+
+    @unittest.skipUnless(USD_AVAILABLE, "Requires usd-core")
+    def test_create_from_usd_converts_length_attributes_to_meters(self):
```
jcarius-nv, line 555: **P1 — Add an end-to-end USD-to-collision test.** This test constructs the model manually, so it cannot verify that pipeline settings use the same unit and scene-selection policy as `ModelBuilder.add_usd()`. Please exercise `stage -> add_usd -> finalize -> create_from_usd -> contacts/collide` and assert an observable contact or configuration outcome. Besides validating this first slice, that gives us a workflow test we can later extend through solver and controller construction.

newton/_src/sim/collide.py
```diff
@@ -1868,6 +1873,241 @@ def __init__(
...
+
+        value = authored("newton:collisionPipeline:speculativeMaxExtension")
+        if value is not None:
+            extension = optional_finite_float("newton:collisionPipeline:speculativeMaxExtension", value, minimum=0.0)
+            if extension is not None:
+                kwargs["speculative_config"] = cls.SpeculativeContactConfig(
+                    max_speculative_extension=extension * linear_unit
+                )
+
+        kwargs.update(overrides)
```
jcarius-nv, line 2101: **P2 — Normalize the deprecated `soft_contact_margin` override.** If USD authors `softContactGap`, `kwargs` contains `soft_contact_gap`; adding `soft_contact_margin` here then passes both names to `__init__`, which raises instead of honoring the documented override contract. Please either normalize this alias before merging the layers or explicitly exclude deprecated aliases from `overrides`.

jcarius-nv (commented):
The overall direction looks reasonable for a first experimental slice. Please also replace the placeholder/self-closing PR description and commented-out test plan with a standalone description of the supported workflow and validation command. I have not repeated the temporary schema-source concern or previously raised override-precedence discussion.

newton/_src/sim/collide.py
jcarius-nv: The conversion added for this thread exposes a second mismatch with the current rigid importer. `ModelBuilder.add_usd()` warns that non-unit linear stages are unsupported for legacy rigid/collider geometry and retains that geometry in stage units, while `create_from_usd()` now converts the pipeline distances to meters. In an end-to-end probe with `metersPerUnit=0.01`, a sphere authored with radius `100` imported with model radius `100.0`, while `softContactGap=2` became `0.02`. Please align the policies; until rigid geometry conversion is supported, rejecting the unsupported non-unit rigid path would be safer than silently constructing a mixed-scale pipeline.

PR #4177, opened 2026-09-08
Fix sharp USD terrain rendering

### LINKED ISSUE #4052: [BUG] Rough terrain mesh geometry incorrect in Newton GL and Newton RTX (rounded vs square stairs)
### Description

Stair geometry in rough terrain environments renders with smoothed/rounded edges in both Newton GL and Newton RTX. The Kit viewer correctly shows sharp square geometry for the same scene. See attached clips: Kit (correct), Newton GL (rounded), Newton RTX (rounded).

In the Newton GL and Newton RTX clips, the stairs are rounded.
https://github.com/user-attachments/assets/ae2a7a89-3b2f-49bf-83db-2f6c70ee3ec7
https://github.com/user-attachments/assets/ca7c8d42-fea2-4fc8-8ef7-1e85ab464297
https://github.com/user-attachments/assets/30e08c6d-e4a2-4712-a6e2-defded71a023

### Motivation / Use Case

Rough terrain is one of the most common Isaac Lab locomotion environments. Incorrect geometry makes it hard to verify terrain curriculum difficulty and debug foot contacts.

### Alternatives Considered

_No response_

vreutskyy:
## Description

Resolve USD mesh shading into Newton's existing triangle topology and per-vertex
normals, following the agreed renderer-neutral direction for 1.7. This replaces
the earlier subdivision-token propagation and viewer-specific flat-shading path.

- Load authored normals on untextured visual meshes too, and canonicalize the
  supported USD normal interpolation modes. Honor the caller's averaging or
  vertex-splitting policy, including its angle threshold, for uniform normals.
- Generate faceted normals for `none`/`bilinear` meshes without authored normals;
  generate smooth normals otherwise, before UV expansion to avoid shading seams.
- Preserve each component's normals when merging differently shaded USD meshes.
  Include normals and UVs in viewer prototype identity without changing the
  physics mesh hash.
- Export final polygonal meshes with explicit normals through USD/RTX. Replace
  stale shading and topology when a mesh name is reused. Do not reinterpret
  triangulated meshes as source subdivision control surfaces.
- Generate missing normals on RTX runtime mesh updates too, including topology
  changes and point-only deformation; preserve explicitly supplied normals.
- Keep source mass properties during visual expansion and avoid duplicate mesh
  loading in the collider mass fallback. Document the topology change and the
  `load_normals=False` geometry-only escape hatch in API docs and Towncrier.

No new public shading enum, raw USD token on `Mesh`, mutable viewer side channel,
or GL shader change is needed. Exact source USD topology/subdivision passthrough
is outside this PR; no separate 1.6.1 patch is included here.

Closes #4052

## Checklist

## Test plan

```text
uv run --no-sync -m unittest newton.tests.test_api newton.tests.test_negative_scaling newton.tests.test_viewer_rtx newton.tests.test_usd_mesh_loading newton.tests.test_viewer_usd newton.tests.test_import_usd newton.tests.test_mesh_cache newton.tests.test_viewer_geometry_batching newton.tests.test_viewer_layers
uvx --from towncrier==25.8.0 towncrier build --draft --version 1.7.0 --date 2026-09-17
uvx pre-commit run -a
```

- Affected test suites: 472 tests, one skipped, no failures.
- Regressions reproduced before fixing canonical shading, mixed-mesh merges,
  uniform-normal conversion policy, render-cache collisions, and reused USD
  mesh shading/topology. Tests include actual camera-sensor normal output.
- Headless OpenGL and RTX renders checked at 640x480 on an RTX 3080 Laptop GPU.
- RTX runtime regressions fail before the fix and verify attribute writes for
  smooth, split, empty, and deforming meshes. A separate headless OVRTX check
  reads back the expected generated normals from the live renderer after
  topology updates to both shared-vertex and split-vertex meshes.
- Local G1 import spot-check against the previous PR head (`c5c05477`): final
  isolated warm medians were 0.920 s before and 0.916 s after (four samples each),
  with 1,888,142 mesh vertices in both. Earlier runs had several-percent timing
  variation; this is a local sanity check, not a replacement for the CI ASV gate.

## Bug fix

**Steps to reproduce:**

1. Import a shared-vertex USD terrain mesh with no authored normals and
   `subdivisionScheme = "bilinear"`.
2. Render it with `ViewerGL` or `ViewerRTX`.
3. Without this PR, sharp edges become smooth in GL, and the USD/RTX output falls
   back to subdivision instead of retaining the terrain's intended appearance.

**Minimal reproduction:**

```python
import numpy as np
from pxr import Usd, UsdGeom

import newton
import newton.usd

mesh = UsdGeom.Mesh.Define(stage, "/Terrain")
mesh.CreatePointsAttr([(0, 0, 0), (1, 0, 0), (0, 1, 0), (0, 0, 1)])
mesh.CreateFaceVertexCountsAttr([3, 3])
mesh.CreateFaceVertexIndicesAttr([0, 1, 2, 0, 3, 1])
mesh.CreateSubdivisionSchemeAttr(UsdGeom.Tokens.bilinear)

loaded = newton.usd.get_mesh(mesh.GetPrim(), load_normals=True, compute_inertia=False)
assert loaded.normals is not None
np.testing.assert_allclose(
    loaded.normals[loaded.indices], [(0, 0, 1)] * 3 + [(0, 1, 0)] * 3
)
```


jcarius-nv:
Thanks @vreutskyy .

We're going to discuss soon with @eric-heiden about whether we can avoid doing so much USD-specific work in Newton and just pass the USD more directly to ViewerRTX, so the Newton model only contains what's truly needed for simulation and not per-viewer specific data.

Apart from that, some concrete findings from my agent below, although I suggest we wait for a moment until we have settled the above questions first:

   - P1 — Render caching ignores the new visual identity. _hash_geometry() (/tmp/newton-pr4177.7oMy3r/newton/_src/viewer/viewer.py:2270) uses the physics Mesh hash, which excludes subdivision scheme and
    normals. I reproduced two otherwise-identical none and catmullClark meshes collapsing to one cached prototype; whichever appears first determines both meshes’ rendering. This directly defeats the
    preservation claim in ordinary multi-shape models.

  - P1 — Heterogeneous merges silently restore smoothing. get_mesh(stage_or_root) (/tmp/newton-pr4177.7oMy3r/newton/_src/usd/utils.py:1198) drops the scheme when merged meshes differ. A none + bilinear probe
    produced _subdivision_scheme=None, making GL smooth again and USD/RTX fall back to catmullClark. The test claiming merge coverage only exercises homogeneous schemes.

  - P1 — ViewerUSD retains stale authored state. ViewerUSD.log_mesh() (/tmp/newton-pr4177.7oMy3r/newton/_src/viewer/viewer_usd.py:351) sets non-None schemes but never clears an existing value. Reusing a mesh
    name with bilinear followed by an unauthored mesh leaves bilinear authored, contradicting the stated round-trip requirement.
    
    - Scope concerns: The current hidden cross-layer contract—raw USD token on Mesh (/tmp/newton-pr4177.7oMy3r/newton/_src/geometry/types.py:221), mutable viewer side-channel, GL interpreting it as flat shading, and USD
  reauthoring it—is not coherent. OpenUSD’s subdivisionScheme controls the subdivision algorithm, not merely shading; Newton has already triangulated and sometimes merged the original control topology before
  reapplying it (OpenUSD Mesh API (https://openusd.org/24.08/api/class_usd_geom_mesh.html)).

jcarius-nv:
Thanks @vreutskyy — we have now settled the architectural question, so the pause I mentioned above can be lifted and you can go ahead with revising this.

Preserving final hard-edge/smooth-shading semantics belongs in Newton because it applies to multiple viewers and to sensor rendering. The shared representation should be renderer-neutral — canonical normals/topology or a portable flat/smooth shading intent — rather than treating the raw USD `subdivisionScheme` as the cross-viewer contract. Exact source-USD control-topology and subdivision fidelity can belong to the planned USD/ovstage passthrough.

The three concrete issues in my earlier comment still need addressing at the current head: render-cache visual identity, heterogeneous-merge shading loss, and stale ViewerUSD authored state. I would target the complete, clean design to 1.7 rather than add patch-specific architecture merely to qualify the full change for 1.6.1. If early delivery is still needed, a separately scoped fix for the no-authored-normal sharp-terrain case would be safer.

newton/_src/usd/utils.py
```diff
@@ -1619,37 +1609,49 @@ def get_mesh(
...
-                    f"Length of uniform normals ({len(normals)}) does not match number of faces "
-                    f"({len(counts)}) for mesh {prim_path}"
+                    f"Length of uniform normals ({len(normals)}) does not match number of faces ({len(counts)}) "
+                    f"for mesh {prim_path}"
                 )
             normals = np.repeat(normals, np.asarray(counts, dtype=np.int32), axis=0)
             normals_interpolation = UsdGeom.Tokens.faceVarying
+            # Uniform normals are explicitly constant per face. Split at every
+            # disagreement rather than applying the face-varying smoothing threshold.
+            normal_conversion = "vertex_splitting"
```
jcarius-nv: One additional compatibility point: because `parse_usd()` now loads authored normals for untextured visual meshes, unchanged `add_usd()` callers can newly reach this branch. It forces uniform normals through `vertex_splitting` with a zero-degree threshold, even when the caller requested `face_varying_normal_conversion="vertex_averaging"`; a two-face probe returned six vertices instead of the previous four.

Vertex splitting can be necessary to preserve discontinuous normals and may be a valid deliberate 1.7 behavior. The concern is that this bypasses the caller's conversion policy and changes `shape_source` topology implicitly. Could we honor the selected conversion policy and document any intentional default topology change? I am not suggesting that we introduce a separate render-topology architecture solely to make this full PR patch-safe; as currently scoped, this broader behavior is a reason not to backport it wholesale to 1.6.1.

jcarius-nv:
Thanks @vreutskyy — I like this iteration much better. Resolving the USD shading intent into canonical mesh topology and normals is aligned with the renderer-neutral direction we discussed, and the earlier cache identity, heterogeneous-merge, stale topology/normals, and uniform-normal conversion concerns look addressed.

One remaining P1 surfaced during the rereview: dynamic ViewerRTX updates can drop the new canonical normals. After ViewerRTX leaves its build phase, `log_mesh(dynamic=True, normals=None)` bypasses ViewerUSD's normal generation, queues `None`, and `_update_ovrtx_mesh_points()` writes an empty normals array to Fabric. The equivalent update therefore remains smooth in ViewerGL/ViewerUSD but can become faceted in ViewerRTX.

Could we preserve or generate the canonical normals on this render-phase path and add a ViewerRTX regression covering a dynamic topology update with omitted normals? The test should verify that generated normals reach the runtime mesh rather than an empty array. This is the one issue currently blocking my approval of the 1.7 change.

jcarius-nv (approved):
Review summary at `d6fa4e0d`:

- **Fit: Fits for 1.7.** Resolving imported USD shading into canonical triangle topology and normals is renderer-neutral and is consumed by ViewerGL, ViewerUSD/RTX, and SensorTiledCamera. A future original-USD/ovstage path remains complementary: it can preserve exact authored USD fidelity, while Newton still needs canonical geometry for non-USD viewers, sensors, and generated/runtime meshes.
- **Requirements:** The earlier cache-identity, heterogeneous-merge, stale-state, normal-conversion, and dynamic ViewerRTX update concerns are addressed at the current head. The implementation consistently carries final shading semantics through the shared mesh representation.
- **Standards/release scope:** The resulting topology and runtime behavior are broader than a micro-release fix. In particular, normal loading can duplicate vertices and affect visible collider topology, while runtime ViewerRTX updates may generate normals when omitted. Those semantics are acceptable as a documented 1.7 change, but the complete series should not be cherry-picked into 1.6.1.

I also applied the seven commits cleanly to the current `release-1.6` head as a compatibility probe: `git diff --check` passed and 473 focused tests passed. That establishes mechanical health, but does not remove the patch-release compatibility concern.

Approving for `main` / 1.7.

jcarius-nv:
just to clarify: I moved this back to 1.7 because cherry-picking seems to risky for me. This is only a "nice-to-have" 1.6.1 item for IsaacLab, so not worth the risk in my eyes.

